PR #19986, opened 2018-03-04
DOC: is confusing for ddof parameter of sem, var and std functions

### LINKED ISSUE #18903: DOC: is confusing for ddof parameter of sem, var and std functions
https://github.com/pandas-dev/pandas/blob/7bb204a05fee20c3c825e7da39ccaf39fbeb8ca5/pandas/core/generic.py#L7160-L7161

It states that _ddof_ is for degrees of freedom with default value of 1. But actually it is for "Delta degrees-of-freedom" (even the name of the parameter is abbreviation of that definition). The meaning is:  how many degrees of freedom **to adjust** relative to the population size. Default of 1 - means we need to **adjust** by one, not use one as a value of degrees of freedom (thats makes no sense). All functions are worked as expected, but documentation on _ddof_ parameter is confusing.

alinde1:
Closes #18903

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -7767,7 +7767,8 @@ level : int or level name, default None
     If the axis is a MultiIndex (hierarchical), count along a
     particular level, collapsing into a %(name1)s
 ddof : int, default 1
-    degrees of freedom
+    Delta Degrees of Freedom. The divisor used in calculations is N - ddof,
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):
Looks good to me!

jorisvandenbossche:
Thanks @alinde1 !

PR #18383, opened 2017-11-20
CLN: Replace comprehensions list/set/dict functions with corresponding symbols

mroeschke:
- [x] tests passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

xref @jreback's [comment](https://github.com/pandas-dev/pandas/pull/18335#discussion_r151851542) and @maxim-lian [comment](https://github.com/pandas-dev/pandas/pull/18335#discussion_r151843012), replace the following where appropriate:

`list(x for x in iterator)` --> `[x for x in iterator]`
`set(x for x in iterator)` --> `{x for x in iterator}`
`dict((x,x) for x in iterator)` --> `{x: x for x in iterator}`

PR #18276, opened 2017-11-14
CLN: Use generators instead of lists in built-in Python functions

mroeschke:
- [x] tests passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

Pass generators in built-in Python functions instead of passing in a list, e.g `any([value for value in iterator])` --> `any(value for value in iterator)`

PR #19989, opened 2018-03-05
CLN: Python comprehensions and generators cleanup followup

mroeschke:
xref #18276, #18383

Some multiline comprehensions and cleanups I missed in the two passes xref'd above.


jreback:
thanks! if you think we should update the lint rule (for .pyx), great.

Issue #19412, opened 2018-01-26
Remove rebasing from contributing guide


datapythonista:
If I'm not wrong, now GitHub squashes commits automatically when merging pull requests [1]. In the pandas documentation there is a section about combining commits [2], that doesn't seem relevant anymore if that is being used.

Would it make sense to delete that section, and may be add a comment about using the title of the pull request for the "right" commit message that will end up in the git history? 

1. https://help.github.com/articles/about-pull-request-merges/#squash-and-merge-your-pull-request-commits
2. https://pandas.pydata.org/pandas-docs/stable/contributing.html#combining-commits

jorisvandenbossche:
Yes, we need to update that section.

We had a discussion somewhat related to this some time ago on the pandas-dev mailing list, see https://mail.python.org/pipermail/pandas-dev/2017-November/000632.html (I just replied to that thread as well). 
So I think the explanation how to squash can be removed (or at least mentioned in what cases it could be useful), and we should add explanation about how to update a feature branch with master without doing a rebase/squash.

willingc:
Sorry folks. I didn't see that the other PR was open. Feel free to close #20190 or ask me to just resubmit the changes related to 'documentation contributions' (which are unrelated to this issue but was seen when scanning down the doc file).

jorisvandenbossche:
I would just update your PR to only include the other changes

PR #19990, opened 2018-03-05
DOC: update contributing guide regarding updating a pull request (merge master)

jorisvandenbossche:
Closes https://github.com/pandas-dev/pandas/issues/19412

Reference discussion: https://mail.python.org/pipermail/pandas-dev/2017-November/000632.html

cc @datapythonista


datapythonista:
Looks great. Is it any reason to not use the `--set-upstream` (or `-u`) in the first push to the branch, so in the other pushes they need to repeat the remote and branch again? I'd say it's more common this way, not sure if there is a reason I'm missing to not use it.

jreback:
@datapythonista actually should check out tracking branches in the first place, e.g.

```
git checkout -b shiny-new-feature --track origin/master
```

is usually the correct thing to do

jorisvandenbossche:
> I'd say it's more common this way, not sure if there is a reason I'm missing to not use it.

Yes, the only problem is then that those commands won't work if you didn't do that in the first place. So I am certainly in favor of mentioning this when creating the branch (as @jreback points out) or when doing the first push. But maybe we can still keep the general working method here? (although it being more verbose)

(I personally have just my general setting of git to push to same-named branches, so I never do `-u` or `--track origin/master`)

doc/source/contributing.rst
```diff
@@ -1045,15 +1019,43 @@ release.  To submit a pull request:
...
 
+Another reason you might need to update your pull request is to solve conflicts
+with changes that have been merged into the master branch since you opened your
+pull request. 
+
+To do this, you need to "merge upstream master" in your branch::
+
+    git checkout shiny-new-feature
+    git fetch upstream
+    git merge upstream/master
```
jschendel: Earlier in this doc, under the "Creating a Branch" section, the prescribed procedure for updating a branch is slightly different:
```
git fetch upstream
git rebase upstream/master
```
Is there a reason to prefer `rebase` in the former section, and `merge` in this one?  Since these sections are relatively far apart it might be nice to provide a quick link between the two sections, and briefly mention when to use one method vs, another?  Or could maybe just use a consistent command in each section?

jorisvandenbossche:
@jschendel did you delete your comment? I think it was relevant:

> Earlier in this doc, under the "Creating a Branch" section, the prescribed procedure for updating a branch is slightly different:
>
>git fetch upstream
>git rebase upstream/master
>
>Is there a reason to prefer rebase in the former section, and merge in this one? Since these sections are relatively far apart it might be nice to provide a quick link between the two sections, and briefly mention when to use one method vs, another? Or could maybe just use a consistent command in each section?

You could say that as long as this is *before* submitting the PR this is fine, but then it can be confusing that we give different guidelines on how to update a branch before and after you submitted a PR ? 
We could also just remove that section, since this is now covered by my new section.

jschendel:
Yeah, I deleted my comment since I missed part of the mailing list discussion thread, and finding it cleared up some of my confusion.  I'm +1 on removing the `rebase` portion of the "Creating a Branch" section.  Seems like that's the most straightforward option that would result in the least amount of confusion for new contributors, especially if we want to nudge them towards using `merge` instead of `rebase`.

jorisvandenbossche:
@jschendel @jreback @TomAugspurger or any body else -> Question: how do you typically update your master branch with upstream/master? (before creating a new feature branch)

Because that is currently not really mentioned in the contributing guide (it only mentions how to update a branch with rebase, which I will change to merge).

I have an git alias that does:

```
git fetch upstream
git merge upstream/master --ff-only
```

I have the `--ff-only` added because in principle I never committed myself on master and only in branches, and this ensures that this is the case (which is a check that `git pull upstream master` will not do I think, but of course could do `git pull upstream master --ff-only` as well)

What do others do for this step? Or what would be the best to recommend in the contributing guide?

jreback:
i have a local tracking branch: master that tracks upstream/master 

so
gilt pull

does it all for me

jorisvandenbossche:
Yes in that case `git pull upstream master` will do the same as `git pull` I think? 
So for the contributing guide, let's assume no set-up tracking branch, and give the full command that certainly works ? 

(and `git pull upstream master --ff-only` will still further do exactly the same as `git pull` if you have no commits on local master that are not present in upstream master. But again, I would like to give the command that is fully save. Somebody who knows git will know how to simplify it).

jschendel:
> Question: how do you typically update your master branch with upstream/master? (before creating a new feature branch)

I only had basic git skills when I started contributing to pandas, so I just followed the contributing guide, probably too closely, and was using the previously recommended `rebase` procedure of:
```
git fetch upstream
git rebase upstream/master
```
Definitely overkill though, as I never actually make changes on master.

jorisvandenbossche:
> Definitely overkill though, as I never actually make changes on master.

Yes, and therefore also the wrong recommendation I think, as *if* you made changes by accident to master, it will not do what you want (it should complain instead IMO)

jschendel:
> therefore also the wrong recommendation I think
Yes, didn't mean for that to sound like I was suggesting the rebase approach that was previously in the docs; was also trying to say that my approach is not the right one here. The fast-forward approach seems like the most appropriate one to me.

jorisvandenbossche:
Yes, I didn't interpret it like that! You nicely followed are guidelines :-)

doc/source/contributing.rst
```diff
@@ -1045,15 +1019,50 @@ release.  To submit a pull request:
...
+
+If there are no conflicts (or they could be fixed automatically), a file with a
+default commit message will open, and you can simply save and quit this file.
+
+If there are merge conflicts, you need to solve those conflicts. See for
+example at https://help.github.com/articles/resolving-a-merge-conflict-using-the-command-line/
```
jreback: need to make this an http ref

doc/source/contributing.rst
```diff
@@ -1045,15 +1019,50 @@ release.  To submit a pull request:
...
+for an explanation on how to do this.
+Once the conflicts are merged and the files where the conflicts were solved are
+added, you can run ``git commit`` to save those fixes.
+
+If you have uncommitted changes at the moment you want to update the branch with
+master, you will need to ``stash`` them prior to updating.  This will effectively
```
jreback: maybe add a reference to git stash here (git docs)

doc/source/contributing.rst
```diff
@@ -1045,15 +1019,50 @@ release.  To submit a pull request:
...
+store your changes and they can be reapplied after updating.
+
+After the feature branch has been update locally, you can now update your pull
+request by pushing to the branch on GitHub::
```
jreback: maybe add a comment about ``-f`` force option of push (they *shouldn't* use it)

jreback (changes requested):
lgtm. very minor comments.

doc/source/contributing.rst
jorisvandenbossche: sphinx does that automatically

doc/source/contributing.rst
jorisvandenbossche: done

jorisvandenbossche:
Any more comment?

TomAugspurger (approved):

Issue #18197, opened 2017-11-09
The use of @cache_readonly removes the docstring of attributes


jorisvandenbossche:
Related to our general investigations into the api docs (https://github.com/pandas-dev/pandas/issues/18147), there is the problem that those attributes that are decorated with `@cache_readonly` no longer have a docstring. Eg:

```
In [1]: pd.Index.is_unique?
Type:        NoneType
String form: None
Docstring:   <no docstring>
```

If I simply comment out the cache_readonly for this attribute (just as a check this is the culprit), I get:

```
In [1]: pd.Index.is_unique?
Signature: pd.Index.is_unique(self)
Docstring: return if the index has unique values 
File:      ~/scipy/pandas/pandas/core/indexes/base.py
Type:      function
```

(although it seems to think it is a function, although `inspect.isfunction` correctly gives false)

Just posting this here as a reminder to further investigate, and to see if somebody directly has an idea how to fix this (@jreback ?). The main problem is that this decorator is implemented in cython, and we probably can't use something like `functools.wraps`.

jreback:
I think we could do what wraps is doing though
which is just setting dunder doc I think

jorisvandenbossche:
> which is just setting dunder doc I think

I tried to do that back then, but that gave compile errors (to be correct: no error on compilation, but on import I got an error that looked like when you didn't compile correctly). So not sure what was going on there.

jorisvandenbossche:
OK, I forgot to declare the new `doc` property I added :-)

Trying it now correctly, I managed to do something like the following:

```patch
diff --git a/pandas/_libs/properties.pyx b/pandas/_libs/properties.pyx
index 4beb24f..37886c6 100644
--- a/pandas/_libs/properties.pyx
+++ b/pandas/_libs/properties.pyx
@@ -9,17 +9,19 @@ from cpython cimport (
 cdef class cache_readonly(object):
 
     cdef readonly:
-        object func, name, allow_setting
+        object func, name, doc, allow_setting
 
     def __init__(self, func=None, allow_setting=False):
         if func is not None:
             self.func = func
             self.name = func.__name__
+            self.doc = getattr(func, '__doc__', None)
         self.allow_setting = allow_setting
 
     def __call__(self, func, doc=None):
         self.func = func
         self.name = func.__name__
+        self.doc = getattr(func, '__doc__', None)
         return self
 
     def __get__(self, obj, typ):
[...]

PR #19991, opened 2018-03-05
CLN/DOC: cache_readonly: remove allow_setting + preserve docstring

jorisvandenbossche:
Closes https://github.com/pandas-dev/pandas/issues/18197

This kind of 'works' in that `?` or `__doc__` now give the correct docstring, but it's not yet fully ideal as it is now seen as a class (which will probably mess up sphinx' autodoc/autosummary, which I didn't check yet)

### DIFF
--- a/pandas/_libs/properties.pyx
+++ b/pandas/_libs/properties.pyx
@@ -17,8 +17,11 @@ cdef class CachedProperty(object):
         self.__doc__ = getattr(func, '__doc__', None)
 
     def __get__(self, obj, typ):
-        # Get the cache or set a default one if needed
+        if obj is None:
+            # accessed on the class, not the instance
+            return self
 
+        # Get the cache or set a default one if needed
         cache = getattr(obj, '_cache', None)
         if cache is None:
             try:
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
@jreback the `cache_readonly` property now has the `allow_setting` option, which is only used for `Index.is_unique`. But this is actually not used anymore internally (it was used for the default index, see https://github.com/pandas-dev/pandas/commit/8c08acaef77e3b901dcbe09b612de785b0c5e782, but that is now a range index). Do you mind me removing it?

jreback:
yeah i think can remove that

jorisvandenbossche:
Another thing, a test is failing for  (with `ind` being an Index):

```
ind.view(np.ndarray).view(Index)

```

but I think this stems from the time that Index was an array subclass. As this now raises an error (with this PR, because `pd.Index.dtype` no longer returns None but an instance of the cached property), while before it was just silently returning garbage.

jorisvandenbossche:
@jreback any comments ?

jorisvandenbossche:
Did some manual timeits on a cache attribute (Index.hasnans), and once it is cached seeing a small slowdown from ca 85 ns to ca 95 ns. Not sure if this is actually related to this, but since this is about 10 nanoseconds, this doesn't seem a problem.

PR #19994, opened 2018-03-05
DOC: require returns section in validation script

jorisvandenbossche:
Are there cases where do not want a "Returns" section? 

cc @datapythonista


datapythonista:
I think the only cases are when the function/method does not return anything (I would about a return None), when a class it's being documented, or when the function yields its value.

If I'm not wrong, it should be easy in Python to detect whether it's a class or a generator yielding the values. But not sure if there is an easy way to detect that a function doean't have a return, or all returns return None

jorisvandenbossche:
> I think the only cases are when the function/method does not return anything (I would about a return None), when a class it's being documented, or when the function yields its value.

I don't think we have many functions/methods in pandas that do not return anything (except for the inplace ones, but that is always an option)

PR #19996, opened 2018-03-05
DOC: updating docstring of Index.shift

DataOmbudsman:
Docstring assigned to PyData Budapest chapter of the Worldwide Documentation Sprint.

As Index.shift is an abstract method, this is more of a docstring for developers implementing a subclass than for end users.

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -2223,20 +2223,37 @@ class Index(IndexOpsMixin, PandasObject):
             can be positive or negative (default is 1).
         freq : pandas.DateOffset, pandas.Timedelta or string
             Frequency increment to shift by (default is None).
-            Offset aliases are valid strings, e.g., 'EOM'.
+            If None, the index is shifted by its own `freq` attribute.
+            Offset aliases are valid strings, e.g., 'D', 'W', 'M' etc.
 
-        See Also
+        Returns
+        -------
+        pandas.Index
+            shifted index
+
+        Examples
         --------
-        DatetimeIndex.shift : an implementation of shift.
-        PeriodIndex.shift : an implementation of shift.
-        TimedeltaIndex.shift : an implementation of shift.
+        >>> month_starts = pd.date_range('1/1/2011', periods=5, freq='MS')
+
+        Shift by 10 days.
+
+        >>> month_starts.shift(10, freq='D')
+        DatetimeIndex(['2011-01-11', '2011-02-11', '2011-03-11', '2011-04-11',
+                       '2011-05-11'],
+                      dtype='datetime64[ns]', freq=None)
+
+        The default value of `freq` is the `freq` attribute of the index;
+        in this example it is 'MS' (month start).
+
+        >>> month_starts.shift(10)
+        DatetimeIndex(['2011-11-01', '2011-12-01', '2012-01-01', '2012-02-01',
+                       '2012-03-01'],
+                      dtype='datetime64[ns]', freq='MS')
 
         Notes
         -----
-        Must be overridden in child classes.
-        h
-        Implemented in datetime-like index classes,
-        e.g., DatetimeIndex, PeriodIndex and TimedeltaIndex.
+        This method is only implemented for datetime-like index classes,
+        i.e., DatetimeIndex, PeriodIndex and TimedeltaIndex.
         """
         raise NotImplementedError("Not supported for type %s" %
                                   type(self).__name__)
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -2211,13 +2211,32 @@ class Index(IndexOpsMixin, PandasObject):
         return self.sort_values(return_indexer=True, ascending=ascending)
 
     def shift(self, periods=1, freq=None):
-        """
-        Shift Index containing datetime objects by input number of periods and
-        DateOffset
+        """Shift index by desired number of time frequency increments.
 
-        Returns
-        -------
-        shifted : Index
+        This method is for shifting the values of datetime-like indexes
+        by a specified time increment a given number of times.
+
+        Parameters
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/base.py
```diff
@@ -2211,13 +2211,32 @@ def sortlevel(self, level=None, ascending=True, sort_remaining=None):
...
+
+        See Also
+        --------
+        DatetimeIndex.shift : an implementation of shift.
+        PeriodIndex.shift : an implementation of shift.
+        TimedeltaIndex.shift : an implementation of shift.
+
+        Notes
+        -----
+        Must be overridden in child classes.
```
jorisvandenbossche: This is not needed here (not relevant information for the user, but only for the developer), although could keep it as a normal comment (`# .. `) after the docstring to clarify this.

I would maybe mention in the Notes section that "This method is only implemented for DatetimeIndex, PeriodIndex and TimedeltaIndex"

pandas/core/indexes/base.py
```diff
@@ -2211,13 +2211,32 @@ def sortlevel(self, level=None, ascending=True, sort_remaining=None):
...
+            can be positive or negative (default is 1).
+        freq : pandas.DateOffset, pandas.Timedelta or string
+            Frequency increment to shift by (default is None).
+            Offset aliases are valid strings, e.g., 'EOM'.
+
+        See Also
+        --------
+        DatetimeIndex.shift : an implementation of shift.
+        PeriodIndex.shift : an implementation of shift.
+        TimedeltaIndex.shift : an implementation of shift.
```
jorisvandenbossche: In principle it is good to add this, but since those are not included in the API docs online, the links will not work, and since this docstring is also used for those referenced methods, it is maybe not needed to include those links (mentioning in the notes that those are implemented will be OK I think)

pandas/core/indexes/base.py
```diff
@@ -2211,13 +2211,32 @@ def sortlevel(self, level=None, ascending=True, sort_remaining=None):
...
+        by a specified time increment a given number of times.
+
+        Parameters
+        ----------
+        periods : int
+            Number of periods (or increments) to shift by,
+            can be positive or negative (default is 1).
+        freq : pandas.DateOffset, pandas.Timedelta or string
+            Frequency increment to shift by (default is None).
+            Offset aliases are valid strings, e.g., 'EOM'.
```
jorisvandenbossche: We should explain here what the default of None means.

jorisvandenbossche (commented):
Can you add some examples and a returns section?

> As Index.shift is an abstract method, this is more of a docstring for developers implementing a subclass than for end users.

It is indeed implemented in the subclasses, but this docstring should still be meant for users, not developers, as users see it if they look for the api docs of Index.shift.

I think we should try to let `DatetimeIndex` et al reuse the parent docstring

jorisvandenbossche:
@DataOmbudsman btw, thanks a lot for the PR! You are a bit unlucky that this one is not fully clear what should be done with it (although there will be many like that during the sprint as well). 
One thing we can do is write this docstring fully complete for users, and then reuse the docstring for `Datetime/Period/TimedeltaIndex.shift`. You could do this by using the `@Appender` decorator

DataOmbudsman:
@jorisvandenbossche Thanks for your notes. I understand your points. When discussing this issue in the sprint mailing list my interpretation of the answer by @datapythonista was that pandas developers are the target audience of this specific docstring rather than pandas users. Maybe I misunderstood it but it seemed reasonable since it was hard for me to imagine a pandas user browsing the docstring of this abstract function.

I've been reluctant to include a Returns section since what this method returns is specified in subclasses that implement this method. I avoided the Examples for the same reason -- how can I give examples of an abstract method that doesn't implement anything? 

Nevertheless, your idea about making this docstring fully complete for users, and then reuse it for `Datetime/Period/TimedeltaIndex.shift` is also reasonable, I will look into `@Appender`.

jorisvandenbossche:
> Maybe I misunderstood it but it seemed reasonable since it was hard for me to imagine a pandas user browsing the docstring of this abstract function.

This is the docstring that is included in the online API docs. So if somebody searches for "pandas index shift", it will end up there. So even if the method itself is not implemented in the base class, the docstring should be targeted to users IMO.

> I will look into @Appender.

It should be something like

```
class DatetimeIndex

    ....

    @Appender(ibase.Index.shift.__doc__)
    def shift(...):
        ....
```

DataOmbudsman:
Reusing the docstring of `Index.shift` is pretty tricky, so I would leave that out of scope if that's acceptable. For example it should handle that parameter names and number of parameters change in the subclasses. The Returns and Notes sections should be changed in the subclasses.

pandas/core/indexes/base.py
```diff
@@ -2211,13 +2211,49 @@ def sortlevel(self, level=None, ascending=True, sort_remaining=None):
...
         Returns
         -------
-        shifted : Index
+        pandas.Index
+            shifted index
+
+        Examples
+        --------
+        >>> month_starts = pd.date_range('1/1/2011', periods=5, freq='MS')
+
```
jorisvandenbossche: Can you show the output here? (so it is easier to compare the result of the shift with the original)

jorisvandenbossche (commented):
Thanks for the updates! Looking good

> eusing the docstring of Index.shift is pretty tricky, so I would leave that out of scope if that's acceptable. For example it should handle that parameter names and number of parameters change in the subclasses

OK, that's fine. We should open an issue to actually fix this, as there is no reason that Index would use different names as DatetimeIndex. But indeed, let's keep that separate from this docstring fix.

jreback:
@jorisvandenbossche if you'd approve the PRs then easy to see what to merge :> (or just give a lgtm).

jreback:
thanks!

PR #19999, opened 2018-03-05
DOC: Updated kurt docstring (for pandas sprint)

WillAyd:
Assigned docstring for pandas documentation sprint

Please note that I also removed **kwargs from the call signature here as they were not being used, so I thought it would be strange to document. For what it's worth I started a discussion on the G[oogle groups list ](https://groups.google.com/forum/#!topic/pandas-documentation-sprint/P5_av_VuV6E)to see if that's the best way to handle or if we wanted to go a different path


pandas/core/window.py
```diff
@@ -899,11 +899,45 @@ def skew(self, **kwargs):
         return self._apply('roll_skew', 'skew',
                            check_minp=_require_min_periods(3), **kwargs)
 
-    _shared_docs['kurt'] = """Unbiased %(name)s kurtosis"""
+    _shared_docs['kurt'] = dedent("""Calculate unbiased %(name)s kurtosis.
 
-    def kurt(self, **kwargs):
+This function uses Fisher's definition of kurtosis (kurtosis of normal
+== 0.0) without bias.
```
jorisvandenbossche: I know this is copied from the DataFrame/Series one, but I personally don't find "kurtosis of normal == 0.0" very clear. I suppose it is about a normal distribution having a kurtosis of 0.0 ?

pandas/core/window.py
```diff
@@ -899,11 +899,45 @@ def skew(self, **kwargs):
...
+
+Returns
+-------
+same type as input
```
jorisvandenbossche: Again, I know this was already there ..:) But I also find this not very clear. What is the "input"? As I didn't provide anything to the function .. (it's rather the object calling rolling) 

But not directly sure how to document this very well. Maybe just "Series/DataFrame" (depending on the calling object)

pandas/core/window.py
```diff
@@ -899,11 +899,45 @@ def skew(self, **kwargs):
...
+
+See Also
+--------
+scipy.stats.kurtosis
+pandas.DataFrame.kurtosis
+pandas.Series.kurtosis
+
+Notes
+-----
+A minimum of 4 periods is required for the rolling calculation
```
jorisvandenbossche: period at and

pandas/core/window.py
```diff
@@ -899,11 +899,45 @@ def skew(self, **kwargs):
         return self._apply('roll_skew', 'skew',
                            check_minp=_require_min_periods(3), **kwargs)
 
-    _shared_docs['kurt'] = """Unbiased %(name)s kurtosis"""
+    _shared_docs['kurt'] = dedent("""Calculate unbiased %(name)s kurtosis.
 
-    def kurt(self, **kwargs):
+This function uses Fisher's definition of kurtosis (kurtosis of normal
```
jorisvandenbossche: I think by using `dedent`, you can actually keep this indented (so it doesn't look that 'off' inside the class)

(I am not even sure the `dedent` is needed, as there are many places where we don't use it, but that is something else)

pandas/core/window.py
```diff
@@ -899,11 +899,45 @@ def skew(self, **kwargs):
...
+A minimum of 4 periods is required for the rolling calculation
+
+Examples
+--------
+>>> arr = [1, 2, 3, 4, 5]
+>>> import scipy.stats
+>>> scipy.stats.kurtosis(arr, bias=False)
+-1.200000000000000
+
+>>> df = pd.DataFrame(arr)
```
jorisvandenbossche: Can you put a small explanation? Very dummy saying what the example is doing

jorisvandenbossche (commented):
Thanks! Added some comments

pandas/core/window.py
WillAyd: So when I did this originally I had:

```python
Series or DataFrame
    Like-indexed object containing the result of the kurt operation
```

However, I ended up with the above because all of the other rolling / expanding functions used this in the `_doc_template` variable (linked below). We could either break that consistency or update the `_doc_template` variable to reflect the above (perhaps substituting `kurt` for `%(name)`) if we think that's a better generally applicable description

https://github.com/pandas-dev/pandas/blob/aedbd948938f7e9230a321eb49f6c789867ab2b6/pandas/core/window.py#L51

pandas/core/window.py
WillAyd: Yep mistake on initial push. Subsequently indented in the next push

pandas/core/window.py
jorisvandenbossche: I think we should break the consistency for now, with the idea that if all those docstrings get updated during the sprint, we should strive again for consistency with the new wording?

Or indeed already update the template, then it is changed for all as well. 

In any case I think we should try to make the docstring that you are now editing, as good as possible.

pandas/core/window.py
```diff
@@ -899,11 +899,45 @@ def skew(self, **kwargs):
...
+
+    Returns
+    -------
+    same type as input
+
+    See Also
+    --------
+    scipy.stats.kurtosis
+    pandas.DataFrame.kurtosis
+    pandas.Series.kurtosis
```
datapythonista: I think it could make sense to add `.skew` too as statistically they are related, and usually appear together in literature. And technically speaking the mean and the variance could also belong here, as they are the first and second moments of a probability density (while skewness and kurtosis are the third and forth).

Also, it's probably worth to add `pandas.Series.rolling`. And not sure if in the "See Also" or somewhere else, but I think it would be very useful for the users to have a link from these functions to https://pandas.pydata.org/pandas-docs/stable/computation.html#window-functions

pandas/core/window.py
```diff
@@ -899,11 +899,45 @@ def skew(self, **kwargs):
...
+
+    Notes
+    -----
+    A minimum of 4 periods is required for the rolling calculation
```
datapythonista: Nice, did you find that the documentation about the "Notes" section made sense for this case?

pandas/core/window.py
```diff
@@ -899,11 +899,45 @@ def skew(self, **kwargs):
...
+    -1.2000000000000004
+
+    >>> df = pd.DataFrame(arr)
+    >>> df.rolling(5).kurt()
+        0
+    0  NaN
+    1  NaN
+    2  NaN
+    3  NaN
+    4 -1.2
```
datapythonista: Probably it's obvious enough, but may be having 6 values instead of 5, would give more the idea that after the 4th `NaN` all the values will be filled?

pandas/core/window.py
```diff
@@ -899,11 +899,45 @@ def skew(self, **kwargs):
...
-    _shared_docs['kurt'] = """Unbiased %(name)s kurtosis"""
+    _shared_docs['kurt'] = dedent("""Calculate unbiased %(name)s kurtosis.
 
-    def kurt(self, **kwargs):
+    This function uses Fisher's definition of kurtosis (kurtosis of normal
+    == 0.0) without bias.
+
+    Returns
+    -------
+    same type as input
```
datapythonista: It's a somehow arbitrary comment, but to me it'd look better something like:
`Series or DataFrame (same as the input) : some description`

Or if the `%(name)s` above `Series` or `DataFrame` depending on in which method is being used? Then we can use it also here.

datapythonista (commented):
Really nice, added some comments with ideas.

pandas/core/window.py
WillAyd: Yea I was thinking the same thing about `pandas.Series.rolling` and `pandas.Series.expanding`. Right now there is the `_doc_template` contained within the module that gives a generic `See Also` section, but my first attempt to override that didn't work out. 

I'll look a little closer to see if there's a scalable way to have all of the functions here reference `pandas.Series.rolling` and the Series / DataFrame methods at a minimum, while allowing for additional "see also" to be specified per-method

pandas/core/window.py
WillAyd: I added this from familiarity with the NumPy standard, not necessarily the docs that you built (although they were *very* helpful on many other aspects - kudos)

WillAyd:
So based on both of your feedback I'm wondering if we should update the `_doc_template` in this module to look as follows:

```python
_doc_template = """

Returns
-------
Series or DataFrame (matches input)
    Like-indexed object containing the result of function application

See also
--------
pandas.Series.%(name)s
pandas.DataFrame.%(name)s
%(other_see_alsos)
"""
```

This would ensure that we consistently always have `pandas.Series.rolling` and `pandas.DataFrame.rolling` as part of every function, but also give it the flexibility to add more references. 

The downside here is at the moment I don't think that there's a way to add `%(other_see_alsos)` without explicitly wrapping all of the other methods in the module with a substitution that provides it. Alternately we could update the `Substitution` class to allow for optional keyword specification, but I'd worry about a lack of explicitness there.

Anyway wanted to get your feedback before going further - do you think the above template is a good idea and if so if it's worth me going and updating all of the `@Substitution` decorators in this module to contain `other_see_also=""` prior to the sprint?

jorisvandenbossche:
@WillAyd I'm fine with both options (only updating this docstring here, or updating the template as well). 

The question is a bit how well the template will scale if all docstrings have a custom extended summary, examples, notes etc, only to share the 'returns' and 'see also' section

jreback:
thanks!

jorisvandenbossche:
@jreback please wait a bit with merging the 'test' PRs for the sprint for now. Let's leave time to comment on them.

jreback:
np

just wanted to clear backlog

i wont merge doc prs then u less u approve

jorisvandenbossche:
> i wont merge doc prs then u less u approve

It's also that for the test PRs now want to keep them open a bit so other organizers can look and review.

jorisvandenbossche:
@WillAyd feel free to open a new PR for my remaining comments

pandas/core/window.py
```diff
@@ -899,7 +899,53 @@ def skew(self, **kwargs):
...
+    -------
+    Series or DataFrame (matches input)
+        Like-indexed object containing the result of function application
+
+    See Also
+    --------
+    pandas.Series.%(name)s
+    pandas.DataFrame.%(name)s
+    pandas.Series.kurtosis
+    pandas.DataFrame.kurtosis
```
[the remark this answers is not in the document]
WillAyd: The substituted name in this case is either `rolling` or `expanding` - does that answer your question?

pandas/core/window.py
```diff
@@ -899,7 +899,53 @@ def skew(self, **kwargs):
...
+
+    Examples
+    --------
+    The below example will show a rolling calculation with a window size of
+    four matching the equivalent function call using `scipy.stats`.
+
+    >>> arr = [1, 2, 3, 4, 999]
+    >>> import scipy.stats
+    >>> print("{0:.6f}".format(scipy.stats.kurtosis(arr[:-1], bias=False)))
+    -1.200000
```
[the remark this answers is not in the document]
WillAyd: Fair enough. Let me mess around with iPython - I think we can use the %precision directive here behind the scenes

pandas/core/window.py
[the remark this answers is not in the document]
jorisvandenbossche: Ah, yes, that clarifies. 
Another point, to avoid that the See also section becomes long, we could also put the related series/dataframe versions on a single line (like `Series.kurtosis, DataFrame.kurtosis`)

pandas/core/window.py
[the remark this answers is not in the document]
jorisvandenbossche: Hmm, the annoying thing here is that scipy prints 1.99999999999, as otherwise `1.200...` would have worked (using an ellipsis to match anything on that line)

Yeah, not sure there is a better solution in  this case

pandas/core/window.py
```diff
@@ -899,7 +899,53 @@ def skew(self, **kwargs):
...
+
+    This function uses Fisher's definition of kurtosis without bias.
+
+    Parameters
+    ----------
+    kwargs : Under Review
+
+    Returns
+    -------
+    Series or DataFrame (matches input)
```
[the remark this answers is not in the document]
jorisvandenbossche: Yes, that sounds good

pandas/core/window.py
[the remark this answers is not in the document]
WillAyd: When you are worried about it becoming to long is that purely within the source or on the web as well? I believe that the HTML output styles this to display it all inline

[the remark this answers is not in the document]
jorisvandenbossche: Ah, didn't know the html output put them on a single line anyhow. Not ideal IMO, but that's numpydoc behaviour.

pandas/core/window.py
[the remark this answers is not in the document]
WillAyd: Yea I was hoping that setting `np.print_options` would help for SciPy but no luck. I kind of wonder if globally it wouldn't be preferable to just set this to 6 across all docstrings for consistency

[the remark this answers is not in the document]
datapythonista: +1 on setting it globally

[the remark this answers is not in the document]
WillAyd: Looked into it and I think that would unfortunately be pretty difficult for scalar values. `np.set_printoptions` may be worthwhile for arrays, but as far as this specific example is concerned the formatting of scalar values from what I can tell is left to interpreters like iPython. I looked at DocTests' documentation but didn't see anything there that would easily allow us to enforce that option during its testing

PR #20000, opened 2018-03-05
API & BUG: allow list-like y argument to df.plot & fix integer arg to x,y

### LINKED ISSUE #19699: Allow list-like for `y` in `DataFrame.plot`.
In pandas 0.22 and earlier, we passing multiple values for `y` worked by accident. The code clearly assumed that `y` was a scalar, but things usually worked, perhaps with an incidental warning:

```python
In [12]: df = pd.DataFrame(np.random.uniform(size=(10, 2)), columns=['a', 'b'])

In [15]: df.plot(x='c', y=['a', 'b'])
/Users/taugspurger/miniconda3/envs/pandas-0.21.0/lib/python3.6/site-packages/pandas/plotting/_core.py:1714: UserWarning: Pandas doesn't allow columns to be created via a new attribute name - see https://pandas.pydata.org/pandas-docs/stable/indexing.html#attribute-access
  series.name = label
Out[15]: <matplotlib.axes._subplots.AxesSubplot at 0x10f83b198>
```

On master, this will currently raise. I think that we should explicitly support list-likes for `y`.

Tagging for 0.23 since I think this should be a blocker.

xref https://github.com/pandas-dev/pandas/pull/18695 where we made the change.

masongallo:
- [x] closes #19699
- [x] tests added / passed
- [x] whatsnew entry

@TomAugspurger 

I added support for the user to pass a list-like to y, as discussed in #19699. The API to `df.plot` is relatively complex with lots of args, so lmk with questions / fdbck. Not sure if I covered everything for docs.


pandas/tests/plotting/test_frame.py
```diff
@@ -2172,24 +2172,33 @@ def test_invalid_kind(self):
...
             df.plot(x=x, y=y)
 
+    @pytest.mark.parametrize("y,lbl", [
+        (['B'], ['b']),
+        (['B', 'C'], ['b', 'c'])
+    ])
+    def test_y_listlike(self, y, lbl):
+        # GH 19699
+        df = DataFrame({"A": [1, 2], 'B': [3, 4], 'C': [5, 6]})
+        _check_plot_works(df.plot, x='A', y=y, label=lbl)
```
TomAugspurger: Could you get the `ax` from `df.plot` and assert that it has two lines? I think `len(ax.lines)` should work.

TomAugspurger: We should also check the color. What happens? Ideally it's the same as `df.set_index('x').plot()`, so two different colors.

pandas/tests/plotting/test_frame.py
```diff
@@ -2172,24 +2172,33 @@ def test_invalid_kind(self):
...
+
```
TomAugspurger: Could you add tests for

1. all integer columns: `x=0, y=[1, 2]`
2. Mix of int and named columns. `x=0, y=['A', 2]`

pandas/plotting/_core.py
```diff
@@ -1714,13 +1714,28 @@ def _plot(data, x=None, y=None, subplots=False,
...
+                label_kw = kwds['label'] if 'label' in kwds else False
+                new_data = data[y].copy()  # Don't modify
+
+                if isinstance(data[y], ABCSeries):
+                    label_name = label_kw or y
+                    new_data.name = label_name
+                else:
+                    match = is_list_like(label_kw) and len(label_kw) == len(y)
+                    if label_kw and not match:
+                        raise ValueError(
```
TomAugspurger: Can you add a test that raises this assertion?

TomAugspurger (commented):
Looks good thanks. Just a couple edge cases that need to be tested I think (your code should handle them already. Just for future regressions).

pandas/tests/plotting/test_frame.py
MasonGallo: > Mix of int and named columns. x=0, y=['A', 2]

Good point. Shouldn't this raise tho? If you try this on a `DataFrame` you'll get a `KeyError`. IMO we shouldn't allow users to specify a mix of int & named cols since it's unclear what you actually want.

TomAugspurger:
Yeah, raising is probably best.

MasonGallo:
I addressed #20056 and added test cases to cover it. Also added a bunch more test cases to increase coverage as requested.

MasonGallo:
Any ideas on the failure on CircleCI? I'm on a mac so it looks like the test gets skipped locally?

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -965,6 +965,8 @@ Plotting
 ^^^^^^^^
 
 - :func:`DataFrame.plot` now raises a ``ValueError`` when the ``x`` or ``y`` argument is improperly formed (:issue:`18671`)
+- :func:`DataFrame.plot` now supports multiple columns to the ``y`` argument (:issue:`19699`)
```
jreback: this first should be on Other Enhancements section

pandas/plotting/_core.py
```diff
@@ -1706,21 +1706,37 @@ def _plot(data, x=None, y=None, subplots=False,
         plot_obj = klass(data, subplots=subplots, ax=ax, kind=kind, **kwds)
     else:
         if isinstance(data, ABCDataFrame):
+            new_data = data.copy()  # don't modify until necessary
```
jreback: this is not necessary, just use 'data' don't add something else here

jreback (changes requested):

pandas/plotting/_core.py
MasonGallo: I added the copy to fix the bug from #20056 so we get correct integer indexing - what did you have in mind?

pandas/tests/plotting/test_frame.py
```diff
@@ -2170,26 +2170,51 @@ def test_invalid_kind(self):
...
+        assert len(ax.lines) == len(y)
+        self._check_colors(ax.get_lines(), linecolors=colors)
+
+    @pytest.mark.parametrize("x,y,colnames", [
+        (0, 1, ['A', 'B']),
+        (1, 0, [0, 1])
+    ])
+    def test_xy_args_integer(self, x, y, colnames):
+        # GH 20056
+        df = DataFrame({"A": [1, 2], 'B': [3, 4]})
```
jreback: we do you think this should be allowed? having *both* labels and positions is so very confusing.

jreback (changes requested):

pandas/tests/plotting/test_frame.py
MasonGallo: I agree that it's confusing and definitely don't think it should be allowed. Let's remove support for positions?

TomAugspurger:
> Any ideas on the failure on CircleCI? I'm on a mac so it looks like the test gets skipped locally?

Merge master into your branch and repush. The'll be fixed.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -965,6 +965,8 @@ Plotting
...
+- Bug in :func:`DataFrame.plot` with ``x`` or ``y`` arguments as positions (:issue:`20056`)
```
TomAugspurger: Could you describe the issue a bit more?

doc/source/whatsnew/v0.23.0.txt
TomAugspurger: Could you move this to other enhancements?

pandas/plotting/_core.py
TomAugspurger: I think it's best to convert from integer to labels as soon as possible. Can you check if you can do it at the very start of `_plot`? Then you don't have to worry about this.

MasonGallo:
@TomAugspurger good suggestion! I think it's doable depending on where we land with the API discussion -

pandas/plotting/_core.py
MasonGallo: done

pandas/plotting/_core.py
```diff
@@ -1751,21 +1751,36 @@ def _plot(data, x=None, y=None, subplots=False,
...
-                series.name = label
+                int_y = is_integer(y) or all(is_integer(c) for c in y)
+                if int_y and not data.columns.holds_integer():
+                    y = data_cols[y]
+                elif not isinstance(data[y], (ABCSeries, ABCDataFrame)):
+                    raise ValueError(
+                        "y must be a label or position or list of them"
+                    )
+                label_kw = kwds['label'] if 'label' in kwds else False
+                new_data = data[y].copy()
```
jreback: just call this data

pandas/plotting/_core.py
```diff
@@ -1751,21 +1751,36 @@ def _plot(data, x=None, y=None, subplots=False,
...
+
+                if isinstance(data[y], ABCSeries):
```
jreback: you can just test if y is_scalar

pandas/tests/plotting/test_frame.py
```diff
@@ -2170,26 +2170,51 @@ def test_invalid_kind(self):
...
         with pytest.raises(ValueError):
             df.plot(x=x, y=y)
 
+    @pytest.mark.parametrize("x,y,lbl,colors", [
+        ('A', ['B'], ['b'], ['red']),
+        ('A', ['B', 'C'], ['b', 'c'], ['red', 'blue']),
+        (0, [1, 2], ['bokeh', 'cython'], ['green', 'yellow'])
+    ])
+    def test_y_listlike(self, x, y, lbl, colors):
+        # GH 19699
```
jreback: can you give a 1-liner expln

pandas/tests/plotting/test_frame.py
```diff
@@ -2170,26 +2170,51 @@ def test_invalid_kind(self):
...
+        ax = df.plot(x=x, y=y, label=lbl, color=colors)
+        assert len(ax.lines) == len(y)
+        self._check_colors(ax.get_lines(), linecolors=colors)
+
+    @pytest.mark.parametrize("x,y,colnames", [
+        (0, 1, ['A', 'B']),
+        (1, 0, [0, 1])
+    ])
+    def test_xy_args_integer(self, x, y, colnames):
+        # GH 20056
```
jreback: same

pandas/plotting/_core.py
jreback: where is this test?

pandas/plotting/_core.py
```diff
@@ -1751,21 +1751,36 @@ def _plot(data, x=None, y=None, subplots=False,
...
-                    y = data.columns[y]
-                elif not isinstance(data[y], ABCSeries):
-                    raise ValueError("y must be a label or position")
-                label = kwds['label'] if 'label' in kwds else y
-                series = data[y].copy()  # Don't modify
-                series.name = label
+                int_y = is_integer(y) or all(is_integer(c) for c in y)
+                if int_y and not data.columns.holds_integer():
+                    y = data_cols[y]
+                elif not isinstance(data[y], (ABCSeries, ABCDataFrame)):
```
jreback: where is the test for this? (as you expanded to both Series & DataFrame)?

jreback: why are you actually selecting ``data[y]`` here what else could ``data[y]`` be?

jreback (changes requested):

pandas/plotting/_core.py
```diff
@@ -1751,21 +1751,36 @@ def _plot(data, x=None, y=None, subplots=False,
...
 
             if y is not None:
-                if is_integer(y) and not data.columns.holds_integer():
-                    y = data.columns[y]
-                elif not isinstance(data[y], ABCSeries):
-                    raise ValueError("y must be a label or position")
-                label = kwds['label'] if 'label' in kwds else y
-                series = data[y].copy()  # Don't modify
-                series.name = label
+                int_y = is_integer(y) or all(is_integer(c) for c in y)
```
TomAugspurger: This looks fragile. For better or worse (mostly worse) you can pretty much any object in pandas columns, e.g. `df = pd.DataFrame({pd.Timestamp('2017'): [1, 2]})`.

I think your code will fail for `y=pd.Timestamp('2017')`, since it's not an integer, but it also isn't iterable.

I'd recommend adding `is_list_like(y)` before the `all`

pandas/plotting/_core.py
TomAugspurger: Yeah, this can probably be removed / replaced?

pandas/plotting/_core.py
TomAugspurger: Duplicate column names may mess that up, not sure if we allow that here though.

MasonGallo: > Duplicate column names may mess that up

right, that's why I have the check for `ABCSeries`

pandas/plotting/_core.py
MasonGallo: agreed, will remove

pandas/plotting/_core.py
MasonGallo: > you can pretty much any object in pandas columns, e.g. df = pd.DataFrame({pd.Timestamp('2017'): [1, 2]})

good call

MasonGallo:
> where is this test?

@jreback not sure why github won't let me respond to that comment inline, but it's here: https://github.com/MasonGallo/pandas/blob/d6d824f3097f2ba48f8778d65422d8471a9f746f/pandas/tests/plotting/test_frame.py#L2176

pandas/plotting/_core.py
MasonGallo: https://github.com/MasonGallo/pandas/blob/d6d824f3097f2ba48f8778d65422d8471a9f746f/pandas/tests/plotting/test_frame.py#L2176

pandas/plotting/_core.py
MasonGallo: good point, removing

TomAugspurger (approved):
Looks good. May have a closer look tomorrow, but otherwise +1

TomAugspurger:
Thanks @MasonGallo!

PR #20004, opened 2018-03-05
DOC: Add `Series` type to `to_sql` docstring

gcbeltramini:


pandas/io/sql.py
```diff
@@ -404,7 +404,7 @@ def to_sql(frame, name, con, schema=None, if_exists='fail', index=True,
 
     Parameters
     ----------
-    frame : DataFrame
+    frame : DataFrame, Series
```
jreback: @jorisvandenbossche is this right? should we deprecate / rename this parameter?

jreback (changes requested):

pandas/io/sql.py
jorisvandenbossche: yes, this seems right, as a few lines below it does:

```
    if isinstance(frame, Series):
        frame = frame.to_frame()
```

but since this is mostly an internal method (most people will directly use DataFrame/Series.to_sql), I don't think we should bother with deprecating or renaming it

jreback (approved):

jreback:
thanks @gcbeltramini

PR #20005, opened 2018-03-05
DOC: fixed dynamic import mechanics of make.py

### LINKED ISSUE #20001: Unable to Build Documentation
```bash

python make.py html

  File "doc/make.py", line 23, in <module>
    import pandas
ModuleNotFoundError: No module named 'pandas'
```

I believe a slight bug was introduced by c8859b57b891701f250fb05f2cc60d2e6cae2d6b in including `import pandas` at the top of the script. Unless the user has pandas installed as a package I don't think this module would know where to find it without modifying the path or being more explicit on import

@jorisvandenbossche

WillAyd:
- [X] closes #20001 
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/doc/make.py
+++ b/doc/make.py
@@ -11,6 +11,7 @@ Usage
     $ python make.py html
     $ python make.py latex
 """
+import importlib
 import sys
 import os
 import shutil
@@ -19,8 +20,6 @@ import argparse
 from contextlib import contextmanager
 import webbrowser
 import jinja2
-
-import pandas
 
 
 DOC_PATH = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


doc/make.py
```diff
@@ -332,7 +331,7 @@ def main():
                                  'compile, e.g. "indexing", "DataFrame.join"'))
     argparser.add_argument('--python-path',
                            type=str,
-                           default=os.path.join(DOC_PATH, '..'),
+                           default=os.path.dirname(DOC_PATH),
```
WillAyd: This is purely stylistic

jorisvandenbossche:
Looks good to me. I suppose this works for you?

cc @datapythonista

datapythonista:
Yes, well spotted. Didn't realize the imported pandas wasn't being affected by the path.

WillAyd:
Checked Travis failures and none look related to this change

jorisvandenbossche:
Thanks!

PR #20006, opened 2018-03-05
TST: split series/test_indexing.py (#18614)

### LINKED ISSUE #18614: TST: split test_indexing
after #18497 ``pandas/tests/series/test_indexing.py`` and its frame equiv are getting large

we could consider splitting these out into a sub-dir, maybe similar to how we split ``pandas/tests/indexing``, which is parially by indexer type, e.g. ``.iloc/.loc`` with some split on dtype (datetimes/period).

almaleksia:
- [x] closes #18614
- [ ] tests added / passed
- [ ] whatsnew entry


almaleksia:
Hmm Travis failure seems unrelated. In "Pull Requests" tab in Travis all PRs built within last several hours fail with the same reason :/

jreback:
ok, looks nice. there are a number of things I'd like to do, but in follow-on PRs. can you create an issue / (and send PR!)

- remove the classes, IOW make all of the tests at function level. This will require removing the ``setup_method`` and making fixtures for this.
- some tests need to be moved around a bit, IOW we want to move tests out of test_indexing.py (the new one), so appropriate places (e.g. .drop should be in alter_axes). 
- want to parameterize things (e.g. anything with loops)

any/all of these can be separate PRs.

jreback:
thanks!

PR #19980, opened 2018-03-04
CLN: deprecate the pandas.tseries.plotting.tsplot function (GH18627)

GGordonGordon:
- [X] closes #18627 
- [X] tests added / passed
- [X] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [X] whatsnew entry

PR #20013, opened 2018-03-06
TST: clean deprecation warnings for xref #19980

jreback:
xfail some mpl > 2.1.2 tests

### DIFF
--- a/pandas/tests/plotting/test_datetimelike.py
+++ b/pandas/tests/plotting/test_datetimelike.py
@@ -111,12 +111,15 @@ class TestTSPlot(TestPlotBase):
 
     @pytest.mark.slow
     def test_tsplot(self):
+
         from pandas.tseries.plotting import tsplot
 
         _, ax = self.plt.subplots()
         ts = tm.makeTimeSeries()
 
-        f = lambda *args, **kwds: tsplot(s, self.plt.Axes.plot, *args, **kwds)
+        def f(*args, **kwds):
+            with tm.assert_produces_warning(FutureWarning):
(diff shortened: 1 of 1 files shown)


jreback:
@jorisvandenbossche
cc @GGordonGordon  

would have rather actually changed the usage of ``tsplot`` to the non-deprecated form in all places (except the actual deprecation test), followup PR would be great.

jorisvandenbossche:
@jreback sorry, somehow I had in my mind that we didn't use this `tsplot` method at all, which is also the case in the code, but apparently not in the tests ... and didn't check that when merging. 

I personally wouldn't bother with changing the tests to non-deprecated version (it seems the tests already test both), but rather just remove them.

jorisvandenbossche:
@jreback did you open an issue for matplotlib 2.2? As this is unrelated I think to the tsplot deprecation?

jreback:
@jorisvandenbossche no I didn't open an issue for mpl 2.2.2. the package is *only* on conda-forge and I don't think works on mac. just trying to fix the build. not even sure what's wrong.

jorisvandenbossche:
@jreback it are issues with the new matplotlib release, not specifically related to mac. Opened an issue for this: https://github.com/pandas-dev/pandas/issues/20031

jreback:
the mac issue were existing for those tests

jorisvandenbossche:
yes, i know, I justed wanted to say that the *new* failure is only related to the new mpl release and not specifically to mac. But that's how you skipped it now, so perfect. We should only actually solve it, therefore opened the issue.

Issue #19932, opened 2018-02-27
DOC: general discussion of our docstring sharing system


jorisvandenbossche:
Currently in many places we write docstrings as string values inside a `_shared_docs` dictionary, and then add this docstring to the method with a decorator like `@Appender(_shared_docs['where'])`, or possibly with substitution of variables in the template docstring with `@Appender(_shared_docs['where'] % _shared_doc_kwargs)`. 
Apart from that, we have some other custom ways to generate them, typically for a specific set of methods (eg all statistical methods like mean, sum, median, ..).

This system is important to not duplicate content too much, but also has some drawbacks:

- I think it is a rather complex system for contributors to understand. It can be hard to know where to find the method's docstring, or how to edit it appropriately, and how it will look like eventually.
- It gives rise to some errors (which can of course be solved). For example:
  - left-overs from shared_kwargs of parent docstrings that have not been substituted correctly. Eg NDFrame should never occur in the docs, but does quite a lot (eg [DataFrame.where docstring](http://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.DataFrame.where.html))
  - left-overs of the template where we forgot to substitute at all (eg [DataFrame.rename docstring](http://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.DataFrame.rename.html))
- We have examples that are not specific for the function (eg [DataFrame.mask docstring](http://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.DataFrame.mask.html) with DataFrame.where examples) or for the object (eg [Index.searchsorted docstring](http://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.Index.searchsorted.html) with Series examples). 
 And in general in can be hard/complicated to add good examples, as those are typically specific to one method, so you need to write them in separate strings and then add to the appropriate docstring.

So therefore opening this issue, to gather ideas on whether we can improve this workflow, without sacrificing the deduplication and flexibility of the current system?

jorisvandenbossche:
Some first thoughts:

* Better document the workflow for contributors (this is actually issue https://github.com/pandas-dev/pandas/issues/16446)
* Remove `% _shared_doc_kwargs` when there is nothing to substitute, and only put values inside `_shared_doc_kwargs` if they are used multiple times: if it is only for one method, write the string there where it is substituted.
[...]

PR #20016, opened 2018-03-06
DOC: add guide on shared docstrings

### LINKED ISSUE #16446: DOC: Explain how our shared_docs system works
Enhance https://github.com/pandas-dev/pandas/blob/master/doc/source/contributing.rst with a discussion of how our `_shared_docs` system works.

- Why we do it (reduce duplicate docstrings, while still having some class-specific stuff)
- How it's done (Appender, _shared_docs dict, _shared_doc_kwargs)
- How to substitute values

TomAugspurger:
closes https://github.com/pandas-dev/pandas/issues/16446

xref https://github.com/pandas-dev/pandas/issues/19932

cc @jorisvandenbossche 

The main change in best practices here is the recommendation is that `Appender` is only for appending. `Substitution` should be used for substituting values.

See the changes to `fillna` that this policy requires. Personally, I find the new way clearer than the old way.

- We write the base template normally
- I like `Appender(NDFrame.method.__doc__)` better than `Appender(_shared_docs['name'] % kwargs)`
- I like `Substitution(**kwargs)` better than `%` formatting.

The downsides is that we don't have the template in a dictionary anymore and the docstring in `NDFrame` has to be the template, it can't have its variables substituted. I think that's OK.


TomAugspurger:
Oh, I put this in contributing, but it should go in the docs file once that's ready.

jorisvandenbossche:
Thanks!

> The downsides is that we don't have the template in a dictionary anymore and the docstring in NDFrame has to be the template, it can't have its variables substituted. 

Ah, yes, that's something I didn't think about. But, as long as the templates are used in classes that are never exposed to users (like NDFrame and IndexOpsMixin), then this is not a problem. But should check that we also don't use the system in actual user classes (eg I think base Index also has shared docstrings that are substituted in the index subclasses)

doc/source/contributing.rst
```diff
@@ -1077,5 +1077,78 @@ The branch will still exist on GitHub, so to delete it there do::
...
 
+Sharing Docstrings
+==================
+
+Pandas has a system for sharing docstrings, with slight variations, between
+classes. This helps us keep docstrings consistent, while keeping things clear
+for the user reading. It comes at the cost of some complexity when writing.
+
+Each shared docstring will have a base template with variables, like
+``%(klass)s``. The variables filled in later on using the ``Substitution``
```
jorisvandenbossche: filled -> are filled

doc/source/contributing.rst
```diff
@@ -1077,5 +1077,78 @@ The branch will still exist on GitHub, so to delete it there do::
...
+.. code-block:: python
+
+   @Appender(template % _shared_doc_kwargs)
+   def my_function(self):
+       ...
+
+where ``template`` may come from a module-level ``_shared_docs`` dictionary
+mapping function names to docstrings. Wherever possible, we prefer using
+``Appender`` and ``Substitution``, since the docstring-writing processes is
+slightly closer to normal.
```
jorisvandenbossche: I don't fully understand this reasoning (both with using `%` or `Substitution` you can write a normal docstring and append it with `baseclass.method.__doc__` ?)

jorisvandenbossche (commented):
Looked at the doc guide now as well, looks good.

Maybe worth to mention that often `Appender` is used to inherit a docstring from partent class, also when no substitution is done (for Index this is eg the majority of the cases)

jorisvandenbossche:
@TomAugspurger I made a PR doing such a clean-up for Index (as this is one where we don't have always a non-user exposed base class where it is no problem to have non-substituted values): #20022

jorisvandenbossche:
@jreback you are OK with this approach? (see also discussion in #19932 and other PR #20022)

jreback:
this is just the guide which is codifying current practice which is good

maybe add something about actually putting functions in Series/Frame that render the doc strings (while the actual doc string itself is in generic.py)

jorisvandenbossche:
> this is just the guide which is codifying current practice which is good

no, it is not exactly codifying the current practice, but a slight variation of it (without `_shared_docs` dict).

There is one example in this PR (see for fillna in the diff, or #20022 for a bigger example)

jreback:
my only comment is actually on your other PR #20022. Should clarify *when* should be using _shared_docs, I agree in Index subclasses we should prob try not to use it as much as its cumbersome.  Though it *is* nice to have the correct return class rather than Index as the generic return value. Its more informative.

jorisvandenbossche:
> my only comment is actually on your other PR #20022. 

Can you then comment there?

> Should clarify when should be using _shared_docs, 

The proposal is to never use it, at least if there is a base class that can hold the template docstrings (like NDFrame). For classes like Index (where the base class is already user exposed), we might discuss further. But I show in the PR #20022 that at least currently we don't actually do this much (4 times for all Index methods).

jorisvandenbossche:
@jreback for NDFrame, we certainly still use the substitution with shared_kwargs, so are you ok with the approach in this PR for that part of the codebase? (use `NDFrame.method.__doc__` instead of `_shared_docs['method']`)

jreback:
oh I c now what you did. yes I much prefer that, its way more explict.

datapythonista:
I know that this approach is based on the current one, and may be it's too ambitious to define a brand-new one, but given that we'll be working in most of the docstrings, I think it's worth suggesting.

Correct me if I'm wrong, but if the general case it's to reuse the docstring of the parent class as a template, why doesn't the decorator get it for us by default? Also, if `_shared_doc_kwargs` is defined as standard, why don't we render the parent docstring with the parameters defined on it?

Then, I'm sure you're already use to it, but I think naming the decorator `Appender` is not very descriptive.

So, with the next example:

    class Series(NDFrame):
        @reuse_parent_docstring
        def fillna(self, ...):
            ...

I think the decorator should be able to automatically do the right thing. And for special cases, the next syntax could also be supported:

    class Series(NDFrame):
        @reuse_parent_docstring(fillna_template, klass='Series',...)
        def fillna(self, ...):
            ...

or

    class Series(NDFrame):
        @reuse_parent_docstring(NotTheParent.fillna.__doc__, **_shared_doc_kwargs)
        def fillna(self, ...):
            ...

Also, may be it's a good time to use the new formatting syntax `'{klass}'` instead of `'%(klass)s'`?

I think this adds a bit of complexity to the decorator, but the code mode clear, so I think it's worth. Am I missing something?

jorisvandenbossche:
> Then, I'm sure you're already use to it, but I think naming the decorator Appender is not very descriptive.

Agreed here, I think "AppendDocstring" or "AddDocstring" (or lower case) would already be more explicit (it is used for more than inheriting docstrings, so "reuse_parent_docstring" might also be confusing in certain cases).

> [about @reuse_parent_docstring] I think the decorator should be able to automatically do the right thing. 

I have thought about this as well, but the problem is that there are many parent classes, for example for Series it are in order: IndexOpsMixin, NDFrame, PandasObject, etc. So already in this case, it would not work out of the box for Series, as the majority of the methods needs to inherit from NDFrame, and not IndexOpsMixin. For DatetimeIndex there are even 5 parent classes before getting to `Index`. 
We could of course define a private attribute that holds this default parent class to inherit from.

> Also, may be it's a good time to use the new formatting syntax '{klass}' instead of '%(klass)s'?

The problem is that this does not work with strings where you also want to use { itself (for example for parameter type description listing the possibilities). It would also be annoying to have to do ``{{'optionA', 'optionB'}}``.  
Although of course it is a trade-off, so it is a choice we have to make, and we can also choose the other way.

TomAugspurger:
A few things:

1. I'm OK with having the contributor explicitly write the exact method that's the template. In my mind, if you're writing a shared docstring, you'll need to be modify both the parent template and the child, so writing `@Appender(NDFrame.method)` is a good thing.
2. @jreback mentioned a clear policy for when to use the two styles. Let's call them dict-style and method-style.

*Use method-style when the parent class, where the template is written, is not public*. This include NDFrame and I think much of IndexOpsMixin. I haven't looked at the groupby ones.


3. I don't have a preference for new vs. old-style format strings.

If people are ok with that policy I'll add it to this PR.

jorisvandenbossche:
> *Use method-style when the parent class, where the template is written, is not public.* This include NDFrame and I think much of IndexOpsMixin. I haven't looked at the groupby ones.

The groupby ones mostly just use variables holding a template (without storing it in a dict), which I think is good. And IndexOpsMixin can all use the method-style (as this is not public). 
The only remaining case that uses the dict-style is Index itself, for which I propose to not use substitution (https://github.com/pandas-dev/pandas/pull/20022).

doc/source/contributing.rst
```diff
@@ -1077,5 +1077,78 @@ The branch will still exist on GitHub, so to delete it there do::
 
     git push origin --delete shiny-new-feature
 
+Sharing Docstrings
```
jreback: can you add a ref here

doc/source/contributing.rst
```diff
@@ -1077,5 +1077,78 @@ The branch will still exist on GitHub, so to delete it there do::
...
+==================
+
+Pandas has a system for sharing docstrings, with slight variations, between
+classes. This helps us keep docstrings consistent, while keeping things clear
+for the user reading. It comes at the cost of some complexity when writing.
+
+Each shared docstring will have a base template with variables, like
+``%(klass)s``. The variables filled in later on using the ``Substitution``
+decorator. Finally, docstrings can be appended to with the ``Appender``
```
jreback: can we reference the Substituion/Appender via links to the code?

doc/source/contributing.rst
```diff
@@ -1077,5 +1077,78 @@ The branch will still exist on GitHub, so to delete it there do::
...
+decorator.
+
+In this example, we'll create a parent docstring normally (this is like
+``pandas.core.generic.NDFrame``. Then we'll have two children (like
```
jreback: we'll -> we will, don't use contractions :>

doc/source/contributing.rst
```diff
@@ -1077,5 +1077,78 @@ The branch will still exist on GitHub, so to delete it there do::
...
+``pandas.core.series.Series`` and ``pandas.core.frame.DataFrame``). We'll
+substitute the children's class names in this docstring.
```
jreback: childrens'

doc/source/contributing.rst
```diff
@@ -1077,5 +1077,78 @@ The branch will still exist on GitHub, so to delete it there do::
...
+   >>> print(Parent.my_function.__doc__)
+   Apply my function to %(klass)s.
+   >>> print(ChildA.my_function.__doc__)
+   Apply my function to ChildA.
+   >>> print(ChildB.my_function.__doc__)
+   Apply my function to ChildB.
+
+Notice two things:
+
+1. We "append" the parent docstring to the children docstrings, which are
```
jreback: I think we use ``*`` rather than numbering generally (though this is slightly more clear)

doc/source/contributing.rst
```diff
@@ -1077,5 +1077,78 @@ The branch will still exist on GitHub, so to delete it there do::
...
+   initially empty.
+2. Python decorators are applied inside out. So the order is Append then
```
jreback: double backticks on Append/Substitution

doc/source/contributing.rst
```diff
@@ -1077,5 +1077,78 @@ The branch will still exist on GitHub, so to delete it there do::
...
+   @Appender(template % _shared_doc_kwargs)
+   def my_function(self):
+       ...
+
+where ``template`` may come from a module-level ``_shared_docs`` dictionary
+mapping function names to docstrings. Wherever possible, we prefer using
+``Appender`` and ``Substitution``, since the docstring-writing processes is
+slightly closer to normal.
+
+See ``pandas.core.generic.NDFrame.fillna`` for an example template, and
```
jreback: can you point to actual code

jreback (changes requested):

doc/source/contributing.rst
TomAugspurger: Will do. Going to wait till the docstring guide is merged and I move it to that document so that the name makes sense.

doc/source/contributing.rst
TomAugspurger: AFAIK, there isn't a good way to link to a specific function. I think you can only do a specific line (with or without a commit). If we pin the commit things may go out of date. If don't do a specific commit then the line number will become incorrect.

doc/source/contributing.rst
TomAugspurger: Children is a plural noun not ending in "s", so children's is the correct plural possesive.

doc/source/contributing.rst
TomAugspurger: Same issue with linking specific lines / commits, though maybe this one is OK, since we're giving an example.

But unfortunately I can't link to the correct version, since the correction is in this PR :)

doc/source/contributing.rst
```diff
@@ -1077,5 +1077,78 @@ The branch will still exist on GitHub, so to delete it there do::
 
     git push origin --delete shiny-new-feature
 
+Sharing Docstrings
+==================
+
```
jreback: I think this line is not long enough

doc/source/contributing.rst
jreback: hmm, yeah never get these right :>. We'll -> we will

TomAugspurger:
Moved to the docstring guide now that it's in master.

TomAugspurger:
@jorisvandenbossche @datapythonista `contribution_docstring.rst` isn't in any toctree. Is that deliberate?

jorisvandenbossche:
Not sure, it is linked too from contributing.rst part about documentation. 

I am fine with adding it to the main index.rst, it's already a disaster anyhow :-) 
But ideally we would have a more nested contributing section of the docs.

doc/source/contributing_docstring.rst
```diff
@@ -908,3 +911,79 @@ plot will be generated automatically when building the documentation.
...
+       ...
+
+where ``template`` may come from a module-level ``_shared_docs`` dictionary
+mapping function names to docstrings. Wherever possible, we prefer using
+``Appender`` and ``Substitution``, since the docstring-writing processes is
+slightly closer to normal.
+
+See ``pandas.core.generic.NDFrame.fillna`` for an example template, and
+``pandas.core.series.Series.fillna`` and ``pandas.core.generic.frame.fillna``
+for the filled versions.
```
jorisvandenbossche: fillna is actually not a really good example, as it has an example section only targetting DataFrame .. (but anyhow, that is another issue :))

jorisvandenbossche (approved):

datapythonista:
I linked `contributing_docstring.rst` from `contributing.rst`. Personally I think `contributing.rst` is long enough to be splited. If that sounds good, we can probably leave it out of any toctree until then. I'm happy to take a looks at it when I find some time.

TomAugspurger:
Yes I’m happy with it being split.

Sphinx warns if there’s source files that aren’t in a toctree. But it’ll be a bit before our doc build is warning free.

TomAugspurger:
Anything else you'd like to see here?

jorisvandenbossche (approved):

jorisvandenbossche:
@TomAugspurger can you update with master?

TomAugspurger:
Done.

jorisvandenbossche:
Thanks!

PR #20020, opened 2018-03-06
DOC: improve groupby.DataFrameGroupBy.resample docstring

pandres:
After asking for feedback, this is the very minimal _docstring_ that we interpret that would be acceptable. 

This is in order to unlock the full list of functions we have assigned for the pandas documentation sprint.

If there is something to be improved here please tell us. Also we may make a new push of this documentation including the full scope of the _args_ and _kwargs_ parameters and more examples.

### DIFF
--- a/pandas/core/groupby.py
+++ b/pandas/core/groupby.py
@@ -1482,11 +1482,13 @@ class GroupBy(_GroupBy):
         Parameters
         ----------
         rule : str
-            Medio.
-        args
-            Hola.
-        kwargs
-            Chau.
+            The offset string or object representing target conversion.
+        *args
+            These parameters will be passed to the get_resampler_for_grouping
+            function.
+        **kwargs
+            These parameters will be passed to the get_resampler_for_grouping
+            function.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


pandas/core/groupby.py
```diff
@@ -1475,9 +1475,169 @@ def describe(self, **kwargs):
...
+        2000-01-01 00:04:00    4
+        2000-01-01 00:05:00    5
+        2000-01-01 00:06:00    6
+        2000-01-01 00:07:00    7
+        2000-01-01 00:08:00    8
+        Freq: T, dtype: int64
+
+        Downsample the series into 3 minute bins and sum the values
+        of the timestamps falling into a bin.
+        >>> series.resample('3T').sum()
```
jorisvandenbossche: This method is `resample` on a groupby object, so the examples should also show that. See here for some explanation of this: https://pandas.pydata.org/pandas-docs/stable/groupby.html#new-syntax-to-window-and-resample-operations

pandas/core/groupby.py
```diff
@@ -1475,9 +1475,169 @@ def describe(self, **kwargs):
...
+
+        Downsample the series into 3 minute bins as above, but label each
+        bin using the right edge instead of the left. Please note that the
+        value in the bucket used as the label is not included in the bucket,
+        which it labels. For example, in the original series the
+        bucket ``2000-01-01 00:03:00`` contains the value 3, but the summed
+        value in the resampled bucket with the label ``2000-01-01 00:03:00``
+        does not include 3 (if it did, the summed value would be 6, not 3).
+        To include this value close the right side of the bin interval as
+        illustrated in the example below this one.
```
jorisvandenbossche: General feedback: best to try to keep this text shorter. 
And related to my comment above: I don't think you need to explain the details here of the `resample` method (for that you can refer to DataFrame.resample and include a "See Also" section), but rather how it is used together with groupby.

jorisvandenbossche:
Ah, OK, I see you actually copied the examples from the Series/DataFrame.resample docstring, that's why they were off topic :) 
(but, best to only open a PR when you actually have changes, work in progress is certainly fine, but best that they are already relevant)

pandres:
@jorisvandenbossche, thank you for the feedback. Maybe I rushed here, we were just doing a warm-up for the sprint on Saturday. 

We do have questions though, related with the `Series` and `DataFrames` examples. I'll check your comment in https://github.com/pandas-dev/pandas/pull/20020#discussion_r172646215

jorisvandenbossche:
If something is unclear, I think it is best to ask on the gitter channel

pandres:
Not sure what's going on with the circleci failures.

pandas/core/groupby.py
```diff
@@ -1475,9 +1475,88 @@ def describe(self, **kwargs):
     @Substitution(name='groupby')
     @Appender(_doc_template)
     def resample(self, rule, *args, **kwargs):
-        """
-        Provide resampling when using a TimeGrouper
+        """Provide resampling when using a TimeGrouper.
+
         Return a new grouper with our resampler appended
+
```
datapythonista: I know this is the original description, but the goal of the sprint is to make it better. After reading these texts, it's still difficult for me to understand what this method is about. In which cases it'll be useful for me.

pandas/core/groupby.py
```diff
@@ -1475,9 +1475,88 @@ def describe(self, **kwargs):
     @Substitution(name='groupby')
     @Appender(_doc_template)
     def resample(self, rule, *args, **kwargs):
-        """
-        Provide resampling when using a TimeGrouper
+        """Provide resampling when using a TimeGrouper.
```
datapythonista: The short summary goes to the next line. Did you run the `./validate_docstrings.py` script with it? It should let you detect formatting errors like this, and save time from reviewers, on things that a script can tell you.

pandas/core/groupby.py
```diff
@@ -1475,9 +1475,88 @@ def describe(self, **kwargs):
...
+        Returns
+        -------
+        Grouper
+            Return a new grouper with our resampler appended
+
+        Examples
+        --------
+
+        Start by creating a DataFrame with 9 one minute timestamps.
+        >>> index = pd.date_range('1/1/2000', periods=9, freq='T')
```
datapythonista: I'd say that we leave a blank line between the text and the code. Does it render all right in the html?

The convention is to use `idx` instead of `index`.

pandas/core/groupby.py
```diff
@@ -1475,9 +1475,88 @@ def describe(self, **kwargs):
...
+        2000-01-01 00:01:00  0  1  2  3
+        2000-01-01 00:02:00  0  1  2  3
+        2000-01-01 00:03:00  0  1  2  3
+        2000-01-01 00:04:00  0  1  2  3
+        2000-01-01 00:05:00  0  1  2  3
+        2000-01-01 00:06:00  5  1  2  3
+        2000-01-01 00:07:00  0  1  2  3
+        2000-01-01 00:08:00  0  1  2  3
+
+        >>> series = pd.Series(range(9), index=index) # delete this
```
datapythonista: The convention is to use `s` instead of series, you can check these in the documentation guide. `# delete this` is because you forgot to delete this? it doesn't seem to add value this line.

pandas/core/groupby.py
```diff
@@ -1475,9 +1475,88 @@ def describe(self, **kwargs):
...
-        """
-        Provide resampling when using a TimeGrouper
+        """Provide resampling when using a TimeGrouper.
+
         Return a new grouper with our resampler appended
+
+        Parameters
+        ----------
+        rule : str
+            The offset string or object representing target conversion.
```
datapythonista: I find this a bit cryptic. I think we can find a better explanation. And I'm sure there is somewhere in the pandas documentation where it explains which are the different values, we could link it. Also we could provide an example.

pandas/core/groupby.py
```diff
@@ -1475,9 +1475,88 @@ def describe(self, **kwargs):
...
+        *args
+            These parameters will be passed to the get_resampler_for_grouping
+            function.
+        **kwargs
+            These parameters will be passed to the get_resampler_for_grouping
+            function.
```
datapythonista: I think it'd be good to do some research on what `get_resampler_for_grouping` does with them, which can be used, and what's the impact on this method.

datapythonista (changes requested):
Added some comments, but please:
* Make sure you read the documentation (it explains the next points there, among other things)
* Make sure somebody else (not the author) understands what the method does, and find this page useful. The (local) reviewer should try to imagine that she/he is a pandas user who wants to use this method. If this page doesn't help, is confusing... Fix it until it's "perfect".
* Run the `validate_docstrings.py` script, and make sure you don't have any of the obvious mistake the script detects.

PR #20021, opened 2018-03-06
Added .pytest_cache to gitignore

WillAyd:
After running the unit test suite `.pytest_cache/` keeps showing up in the project root. Figured it was worth adding to `.gitignore` to prevent this from being inadvertently committed


jreback:
thanks!

PR #20022, opened 2018-03-06
DOC/CLN: Index shared docs with appending actual docstrings

jorisvandenbossche:
xref https://github.com/pandas-dev/pandas/issues/19932 and https://github.com/pandas-dev/pandas/pull/20016

This removed the `_index_shared_docs` dict for the Index class. 
There were 4 methods that actually substituted values: take, get_indexer, get_indexer_non_unique, fillna

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -1868,7 +1868,7 @@ class Index(IndexOpsMixin, PandasObject):
     def take(self, indices, axis=0, allow_fill=True,
              fill_value=None, **kwargs):
         """
-        return a new %(klass)s of the values selected by the indices
+        return a new Index of the values selected by the indices
 
         For internal compatibility with numpy arrays.
 
@@ -2705,7 +2705,7 @@ class Index(IndexOpsMixin, PandasObject):
 
         Parameters
         ----------
-        target : %(target_klass)s
+        target : array-like or Index
         method : {None, 'pad'/'ffill', 'backfill'/'bfill', 'nearest'}, optional
             * default: exact matches only.
             * pad / ffill: find the PREVIOUS index value if no exact match.
@@ -2868,7 +2868,7 @@ class Index(IndexOpsMixin, PandasObject):
 
         Parameters
         ----------
-        target : %(target_klass)s
+        target : array-like or Index
 
         Returns
         -------
@@ -3914,7 +3914,7 @@ class Index(IndexOpsMixin, PandasObject):
 
         Returns
         -------
-        filled : %(klass)s
+        filled : Index of the same type
         """
         self._assert_can_do_op(value)
         if self.hasnans:
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -73,8 +73,7 @@ _unsortable_types = frozenset(('mixed', 'mixed-integer'))
 
 _index_doc_kwargs = dict(klass='Index', inplace='',
                          target_klass='Index',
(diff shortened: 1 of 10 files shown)


jorisvandenbossche:
You can see in the second commit what actually needed to be changed in the Index docstrings. 

For `take` and `fillna`: I don't think it was that much of a added value that it said the specific Index subclass (in many other docstrings we just say 'Index') 
For `get_indexer` and `get_indexer_non_unique` the substituted values actually contained more information (eg target_klass was `PeriodIndex or list of Periods`). However, this was not always correct (eg for CategoricalIndex it does not only accept CategoricalIndex, but aslo object Index/Series/array/list) or consistent (eg for PeriodIndex we do not accept an object Index of Periods (but array or list is fine), while for CategoricalIndex we do).

If we want, I can keep a template for `get_indexer` / `get_indexer_non_unique`. But personally I don't think this is worth it (and I would rather try to make it more consistent)

jreback:
> With this PR it can be quite clear: for Index subclasses we don't use template substitution, for Series/Index (from IndexOpsMixin) or Series/DataFrame (from NDFrame) we do use them (for some of the methods).

see that's the thing. Why would we do this? it just adds to the confusion. Index/Series are very very similar, so to do it in Series but not Index raises questions of why? and then there is the special case of IndexOpsMixin.

jorisvandenbossche:
> Why would we do this? it just adds to the confusion.

There is not much difference in practice:

```
@Subtitution(..)
@Appender(NDFrame.method.__doc__)
```

vs

```
@Appender(Index.method.__doc__)
```

Anyhow, even for NDFrame, is there is actually *no* substitution, we would only do `Appender` without `Substitution`, so you would get such a mixture in frame.py/series.py anyway (as you have this mixture today as well for cases that use `%` and other not).

jorisvandenbossche:
IMO this makes it more explicit. Now often there is `% _shared_docs_kwargs` without that there is actually something to replace. For me, this creates confusion. So I would personally only do `% _shared_docs_kwargs`  / `@Substitution(_shared_docs_kwargs)` when there is actually something to replace in the docstring.

jreback:
@jorisvandenbossche the problem is a new contributor will have no which one to use. we need a clear unambiguous policy. it may be obvious. e.g. if the method is *only* defined in base.py then don't use a template / shared, otherwise do.

pandas/core/indexes/base.py
```diff
@@ -1890,8 +1865,10 @@ def _concat_same_dtype(self, to_concat, name):
         # must be overridden in specific classes
         return _concat._concat_index_asobject(to_concat, name)
 
-    _index_shared_docs['take'] = """
-        return a new %(klass)s of the values selected by the indices
+    def take(self, indices, axis=0, allow_fill=True,
+             fill_value=None, **kwargs):
+        """
+        return a new Index of the values selected by the indices
```
TomAugspurger: Thoughts on phrasing this as "Return a new index of the same type ..."? I suppose "of the values" implies that you're getting the same type of index back though.

jreback: generally we need language to assert that we are getting back the same type of Index after the op (which is almost always true), though there are exceptions. This is technically easier to have the doc-strings render like this, e.g. ``PeriodIndex`` is pretty explicty as a return type for example.

jreback:

jorisvandenbossche:
I will look into rebasing this next week

jreback:
closing as stale, but reopen if can continue

Issue #19165, opened 2018-01-10
PERF: Discrepancy in groupby methods


mroeschke:
xref #8426 and [comment](https://github.com/pandas-dev/pandas/pull/19113#issue-286546244)

[issue filter for groupby & perf](https://github.com/pandas-dev/pandas/issues?q=is%3Aopen+is%3Aissue+label%3AGroupby+label%3APerformance)

- [ ] ``.describe()``
- [ ] ``.mad()``
- [x] ``.pct_change()`` (#19919)
- [x] ``.rank()`` (#15779) (#19481)
- [x] ``.all()`` and ``.any()`` (#15435)
- [x] `fillna()` (#11296) (#19673)
- [ ] `mode()` (#19254)

Some groupby methods (notably `describe`, `mad`, `pct_change`) are not as performant as others. Many of the less performant methods are pre-generated in a `_common_apply_whitelist` in `pandas/core/groupby.py`, so it may be worthwhile to revisit this implementation. 

```
asv dev -b ^groupby.GroupByMethods
· Discovering benchmarks
· Running 1 total benchmarks (1 commits * 1 environments * 1 benchmarks)
[  0.00%] ·· Building for existing-py_home_matt_anaconda_envs_pandas_dev_bin_python
[  0.00%] ·· Benchmarking existing-py_home_matt_anaconda_envs_pandas_dev_bin_python
[100.00%] ··· Running groupby.GroupByMethods.time_method                     ok
[100.00%] ···· 
               ======= ============== ========
                dtype      method             
               ------- -------------- --------
                 int        all        256ms  
                 int        any        255ms  
                 int       count       925μs  
                 int      cumcount     1.15ms 
                 int       cummax      1.13ms 
                 int       cummin      1.15ms 
                 int      cumprod      1.58ms 
                 int       cumsum      1.16ms 
                 int      describe     3.25s  
                 int       first       1.12ms 
                 int        head       1.37ms 
                 int        last       1.12ms 
                 int        mad        1.42s  
                 int        max        1.12ms 
                 int        min        1.16ms 
                 int       median      1.53ms 
                 int        mean       1.43ms 
                 int      nunique      1.40ms 
                 int     pct_change    1.56s  
                 int        prod       1.53ms 
                 int        rank       380ms  
                 int        sem        414ms  
                 int       shift       974μs  
                 int        size       858μs  
                 int        skew       414ms  
                 int        std        1.46ms 
                 int        sum        1.50ms 
[...]

Issue #11787, opened 2015-12-07
DEPR: mad (mean absolute difference) functions


aechase:
The generic function `.mad()` calculates the mean absolute difference of a set of data, but in some cases the median absolute difference is more appropriate. In `R`, the `mad()` function accepts a `center` argument to specify how the average absolute difference should be calculated. I propose to add the same to the _pandas_ function.

jreback:
we would do this as ``how='mean'`` as the default (and accept say ``median``).

shoyer:
`df.mad()` is equivalent to `abs(df - df.mean()).mean()`, which is much more explicit and obvious. Likewise, `abs(df - df.median()).mean()` would probably be better than `df.mad(how='median')` (note that from a performance perspective these are equivalent). I'm not entirely sure we really even need a separate `.mad()` method, though at this point I suppose there's little to gain from deprecating it.

StephenKappel:
In R, the `mad()` function can never be used to calculate the `mad()` as currently defined in Pandas. The outer aggregator is always the median. The`center` argument is a numeric constant and is used like this: `abs(df - center).median()`.

Having an option to use the median in Pandas seems like a good idea; robustness is a common reason to use the MAD rather than standard deviation, but the current implementation is certainly not robust.

I think `how` should control the outer aggregator and adding a separate `center` argument similar to R's could be convenient and add clarity.

`df.mad(how='median', center='median')` would evaluate as `abs(df - df.median()).median()`

`df.mad(how='median', center='mean')` would evaluate as `abs(df - df.mean()).median()`

`df.mad(how='median', center=15.4)` would evaluate as `abs(df - 15.4).median()`

`df.mad(how='mean', center='median')` would evaluate as `abs(df - df.median()).mean()`

etc....

If this seems like a reasonable end-state, I can work on this issue.

shoyer:
We could add a `center` keyword argument, but I would probably add documentation examples advising `abs(x - x.median()).median()` before adding both new keywords.

BastiaanBergman:
* R implements MAD as:

> constant * Median(abs(x - center)) 

where the constant is by default ~1.48 which makes it comparable to the standard deviation, the center is provided by the user (could be median or anything). [R manual](https://stat.ethz.ch/R-manual/R-devel/library/stats/html/mad.html)

* SAS implements MAD as:

> median absolute deviation from the median

[...]

PR #20024, opened 2018-03-07
Cythonized GroupBy mad

WillAyd:
- [ ] progress towards #19165 
- [X] tests added / passed
- [x] whatsnew entry

```bash
     [01b91c26]       [9d7f0ac9]
+         678±3μs         807±10μs     1.19  groupby.GroupByMethods.time_method('object', 'value_counts')
+       116±0.8μs          135±1μs     1.16  groupby.GroupByMethods.time_method('object', 'shift')
+        777±10μs         888±10μs     1.14  groupby.GroupByMethods.time_method('object', 'unique')
+      71.6±0.9μs       81.4±0.4μs     1.14  groupby.GroupByMethods.time_method('object', 'size')
-           730ms       2.87±0.4ms     0.00  groupby.GroupByMethods.time_method('int', 'mad')
-           1.23s       3.05±0.5ms     0.00  groupby.GroupByMethods.time_method('float', 'mad')

```


pandas/tests/groupby/test_whitelist.py
```diff
@@ -12,7 +12,7 @@
 
 AGG_FUNCTIONS = ['sum', 'prod', 'min', 'max', 'median', 'mean', 'skew',
                  'mad', 'std', 'var', 'sem']
-AGG_FUNCTIONS_WITH_SKIPNA = ['skew', 'mad']
+AGG_FUNCTIONS_WITH_SKIPNA = ['skew']
```
WillAyd: Speaking to why I did this - since `mad` uses the `mean` behind the scenes I figured it made sense to rely on `mean` to do as much of the heavy lifting as possible. Unfortunately `mean` doesn't currently handle the `skipna` parameter but I think it's worth addressing within that function and allowing that to pass through to `mad` rather than implementing specifically within `mad`.

As always glad to open an issue for that if you agree on approach

pandas/core/groupby.py
```diff
@@ -1353,6 +1353,26 @@ def var(self, ddof=1, *args, **kwargs):
...
+    def mad(self, skipna=True):
+        if not skipna:
+            raise NotImplementedError("'skipna=False' not yet implemented")
+
+        if self.axis != 0:
+            return self.apply(lambda x: x.mad(axis=self.axis))
+
+        # Wrap in a try..except to catch a TypeError with bool data
+        # Ideally this would be implemented in `mean` instead of here
+        try:
```
WillAyd: As touched on in the comments I ideally would not want this `try...except` and think instead that the `mean` application should be throwing the error. While `mean` does raise for `object` types, something like `pd.Series([True, False, True]).mean()` is entirely valid and therefore this ends up throwing a TypeError on the subsequent `np.abs` operation.

I think `mean` should raise on boolean data and have that pass through. Will open separate issue if you agree

pandas/tests/groupby/test_groupby.py
```diff
@@ -1300,17 +1301,6 @@ def test_non_cython_api(self):
         g = df.groupby('A')
         gni = df.groupby('A', as_index=False)
 
-        # mad
```
WillAyd: This test was failing with this change because it allowed for `mad` operations on `object` types, returning `NaN` rather than raising. I figured the explicit tests added elsewhere made more sense and that raising is preferable to `NaN` for particular types.

There's also some code that tests `any` further down within this test. Didn't check in detail but can probably be removed on account of changes done in #19722. Assuming I have updates I'll include in the next batch or alternately open up a separate issue

WillAyd:
I assume you are referencing #11787. FWIW it would be a pretty trivial change here to support 'median' as well. 

OK with deprecating, but I will say the one thing that we may want to consider is how users would roll this on their own. Operating on a series is for sure straightforward:

```python
abs(df['val'] - df['val'].mean()).mean()
```

But attempting the same on a GroupBy object will Raise

```python
abs(df.groupby('key') - df.groupby('key').mean()).mean()
ValueError: Unable to coerce to Series, length must be 1: given 2
```

So the user would be responsible for some heavier lifting on that side, assuming there's no built-in way for the plain `GroupBy` object to handle subtraction of its aggregated result.

jreback:
try with a group by and then transform
this is a typical pattern

WillAyd:
Assuming you mean

```python
abs(df - df.groupby('key').transform('mean')).mean()
```
It's possible but then requires the user to explicitly drop the grouped field(s) somewhere in the operation.

A similar implementation (which I used here) doesn't add the grouped fields but is definitely more verbose than what would be required for the Series / DataFrame counterparts

```python
abs(df.groupby('key').shift(0) - df.groupby('key').transform('mean')).mean()
```

jreback:
not sure why u need .shift
also you don’t need the final mean 
not sure what that is for

WillAyd:
If you remove the shift you get 

```python
ValueError: Unable to coerce to Series, length must be 1: given 2
```

The inner mean is the "centering" op and the outer is the resulting op. I think [this comment](https://github.com/pandas-dev/pandas/issues/11787#issuecomment-172399414) sums up the possible combinations pretty well

shoyer:
In xarray we support [arithmetic with GroupBy objects](http://xarray.pydata.org/en/stable/groupby.html#grouped-arithmetic) so your example would actually work:
```
abs(df.groupby('key') - df.groupby('key').mean()).mean()
``

It would be interesting to explore porting this syntax to pandas. Xarray users find it pretty useful for doing these sorts of grouped normalizations (which are common in climate science).

WillAyd:
@shoyer that could be a viable option here, and certainly make the mad operation across `DataFrame` / `Series` / `GroupBy` consistent while opening up some other possibilities (ex: easy demeaning). 

Do you know where that is implemented in xarray? Dug through the source but nothing was immediately apparent to me

shoyer:
We use some awkward machinery for defining binary ops in xarray (you'd need to figure out how to do this for pandas), but here's where the core groupby arithmetic logic is defined:
https://github.com/pydata/xarray/blob/870e4eaf1895cfeffdc27dab61ad739e67133777/xarray/core/groupby.py#L301-L332

jreback:
@WillAyd ideally like to move the groupby cython routines to pandas/core/groupby/cython before we attempt this

PR #20025, opened 2018-03-07
Replace api coverage script

datapythonista:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

This PR adds to the output of the script `validate_docstrings.py` the methods from `Series`, `DataFrame` and `Panel` that are not present in `doc/api.rst`. This makes the script `api_rst_coverage.py` obsolete, as its only functionality was to detect these methods, and the methods in `doc/api.rst` that didn't exist, which was already returned by `validate_docstrings.py`.

Besides that, the PR implements to minor changes. It adds new columns 'section' and 'subsection' to the csv returned by `validate_docstrings.py`, so the csv contains the section and subsection in `api.rst` in which each docstring is defined (besides giving context, this will be useful to divide the docstrings in similar groups for the sprint).

Also, this PR adds some extra validations, like making sure the docstrings starts in the next line after the opening triple quotes, and similar validations.

CC @jorisvandenbossche


scripts/validate_docstrings.py
```diff
@@ -75,7 +97,7 @@ class Docstring:
     def __init__(self, method_name, method_obj):
         self.method_name = method_name
         self.method_obj = method_obj
-        self.raw_doc = pydoc.getdoc(method_obj)
+        self.raw_doc = method_obj.__doc__ or ''
```
jorisvandenbossche: Is there a reason you changed this back? 
An advantage of `pydoc.getdoc` is that it correctly deals with dedenting docstrings

scripts/validate_docstrings.py
```diff
@@ -328,6 +432,13 @@ def validate_one(func_name):
         errs.append('Private classes ({}) should not be mentioned in public '
                     'docstring.'.format(mentioned_errs))
 
+    if not doc.see_also:
+        errs.append('See Also section not found')
+    else:
+        for rel_name, rel_desc in doc.see_also:
```
jorisvandenbossche: This gives me an error for ``scripts/validate_docstrings.py pandas.DataFrame.join`` (and others as well):

  File "scripts/validate_docstrings.py", line 483, in <module>
    sys.exit(main(args.function))
  File "scripts/validate_docstrings.py", line 469, in main
    return validate_one(function)
  File "scripts/validate_docstrings.py", line 438, in validate_one
    for rel_name, rel_desc in doc.see_also:
ValueError: too many values to unpack (expected 2)

scripts/validate_docstrings.py
```diff
@@ -125,18 +169,22 @@ def needs_summary(self):
...
+        return {name: (type_, ''.join(desc))
+                for name, type_, desc in self.doc['Parameters']}
 
     @property
     def signature_parameters(self):
-        if not (inspect.isfunction(self.method_obj)
-                or inspect.isclass(self.method_obj)):
+        try:
+            signature = inspect.signature(self.method_obj)
+        except ValueError:
```
jorisvandenbossche: this breaks for attributes (they raise a TypeError). Also, the test for {'dt', 'str', 'cat'} might need to be done before this one

scripts/validate_docstrings.py
```diff
@@ -125,18 +169,22 @@ def needs_summary(self):
 
     @property
     def doc_parameters(self):
-        return self.doc['Parameters']
+        return {name: (type_, ''.join(desc))
+                for name, type_, desc in self.doc['Parameters']}
```
jorisvandenbossche: best to use OrderedDict for people not on python >3.6, otherwise the order of the params is lost (and you get incorrect error messages about order not matching)

scripts/validate_docstrings.py
jorisvandenbossche: One of the reasons I made this change is because when the dedenting is not done properly, the inference of the parameters by NumpyDocString goes wrong (eg try to validate pandas.Timestamp)

scripts/validate_docstrings.py
```diff
@@ -210,101 +261,135 @@ def examples_errors(self):
...
     errs = []
+    if doc.start_blank_lines != 1:
+        errs.append('Docstring text (summary) should start in the line '
+                    'immediately after the opening quotes (not in the same '
+                    'line, or leaving a blank line in between)')
+    if doc.end_blank_lines != 1:
+        errs.append('Closing quotes should be placed in the line after '
+                    'the last text in the docstring (do not close the '
+                    'quotes in the same line as the text, or leave a '
+                    'blank line between the last text and the quotes)')
```
jorisvandenbossche: There are actually some cases where I see people deliberately leaving a blank line at the end of the docstring (see eg examples in http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html. Also numpydoc does it in the example: http://numpydoc.readthedocs.io/en/latest/format.html, although not sure if it is on purpose). 

So not sure we should be that strict here.

scripts/validate_docstrings.py
datapythonista: `pydoc.getdoc` strips the string, so we can't validate whether the user is starting the docstring after the quotes, in the next line, or 3 lines after. I thought it was only used to dedent the docstring in the print, and I added the `textwrap.dedent` there. Will take a look at what you say in `pandas.Timestamp`.

scripts/validate_docstrings.py
datapythonista: Ok, will take a look, I guess the solution is to also capture `TypeError`. With the `isfunc` and `isclass` we had before was raising an exception (I assume to everyone) in `Index.at` I think.

jreback:
note generally I would simply not touch Panel doc-strings at all, except if you need to handle shared things.

scripts/validate_docstrings.py
jorisvandenbossche: Do you mean Series.at instead of Index.at? 
For `Series.at` I now get an error, while previously it worked.

scripts/validate_docstrings.py
jorisvandenbossche: Ah, yes. We can also keep both ... but that makes it a bit more complex.
Or only do the dedented one (with pydoc.getdoc) when passing it to NumpyDocString

scripts/validate_docstrings.py
datapythonista: This is the error:
```
  File "./validate_docstrings.py", line 372, in <module>
    sys.exit(main(args.function))
  File "./validate_docstrings.py", line 356, in main
    return validate_all()
  File "./validate_docstrings.py", line 273, in validate_all
    int(doc.correct_parameters),
  File "./validate_docstrings.py", line 169, in correct_parameters
    return not bool(self.parameter_mismatches)
  File "./validate_docstrings.py", line 147, in parameter_mismatches
    signature_params = self.signature_parameters
  File "./validate_docstrings.py", line 139, in signature_parameters
    params = tuple(inspect.signature(self.method_obj).parameters.keys())
  File "/home/mgarcia/.anaconda3/envs/pandas_dev/lib/python3.6/inspect.py", line 3036, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "/home/mgarcia/.anaconda3/envs/pandas_dev/lib/python3.6/inspect.py", line 2786, in from_callable
    follow_wrapper_chains=follow_wrapped)
  File "/home/mgarcia/.anaconda3/envs/pandas_dev/lib/python3.6/inspect.py", line 2341, in _signature_from_callable
    'no signature found for builtin type {!r}'.format(obj))
ValueError: no signature found for builtin type <class 'pandas.core.indexing._AtIndexer'>
```

jorisvandenbossche:
Quick feedback: parameters are still not working due to list vs set I think(didn't look at the code). Output for DataFrame:

```
	Errors in parameters section
		Wrong parameters order. Actual: ('data', 'index', 'columns', 'dtype', 'copy'). Documented: ['data', 'index', 'columns', 'dtype', 'copy']

```

scripts/validate_docstrings.py
```diff
@@ -125,18 +175,23 @@ def needs_summary(self):
...
             # accessor classes have a signature, but don't want to show this
             return tuple()
-        params = tuple(inspect.signature(self.method_obj).parameters.keys())
+        try:
+            signature = inspect.signature(self.method_obj)
+        except (TypeError, ValueError):
+            # Some objects, mainly in C extensions do not support introspection
+            # of the signature
+            return tuple()
+        params = tuple(signature.parameters.keys())
```
jorisvandenbossche: tuple -> list ?

datapythonista: Did the opposite, but yes, fixed. :)

scripts/validate_docstrings.py
datapythonista: It's your call, but personally I think that if there is no specific reason to leave one line sometimes, and to not leave it other times, I'd define a standard way to do it.

jorisvandenbossche: Somehow I have a personal preference to leave a blank line at the end .. :-) 
But since the way you check it now is most occurring way in pandas, let's pick that.

PR #20036, opened 2018-03-07
DOC: Extend docstring pandas core index to_frame method

stijnvanhoey:
- [ ] whatsnew entry

This PR provides adaptations of the original `to_frame` docstring, taking into account the [docstring guide](https://python-sprints.github.io/pandas/guide/pandas_docstring.html).

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -1124,6 +1124,23 @@ class Index(IndexOpsMixin, PandasObject):
         -------
         DataFrame
             DataFrame containing the original Index data.
+
+        Examples
+        --------
+        >>> idx = pd.Index(['Ant', 'Bear', 'Cow'])
+        >>> idx.to_frame()
+                0
+        Ant    Ant
+        Bear  Bear
+        Cow    Cow
+
+        By default, the original Index is reused. To enforce a new Index:
+
+        >>> idx.to_frame(index=False)
+            0
+        0   Ant
+        1  Bear
+        2   Cow
         """
 
         from pandas import DataFrame
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -1117,7 +1117,7 @@ class Index(IndexOpsMixin, PandasObject):
         Parameters
         ----------
         index : boolean
-            Set the index of the returned DataFrame as the original Index 
+            Set the index of the returned DataFrame as the original Index
             (default True).
 
         Returns
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -1116,8 +1116,9 @@ class Index(IndexOpsMixin, PandasObject):
 
         Parameters
         ----------
-        index : boolean, default True
-            Set the index of the returned DataFrame as the original Index.
+        index : boolean
+            Set the index of the returned DataFrame as the original Index 
+            (default True).
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/base.py
```diff
@@ -1110,19 +1110,37 @@ def to_series(self, index=None, name=None):
...
-        Create a DataFrame with a column containing the Index.
+        """Create a DataFrame with a column containing the Index.
 
         .. versionadded:: 0.21.0
 
         Parameters
         ----------
-        index : boolean, default True
-            Set the index of the returned DataFrame as the original Index.
+        index : boolean
```
jreback: @datapythonista @jorisvandenbossche defaults on the same line, yes?

jreback (changes requested):

pandas/core/indexes/base.py
datapythonista: Yes, the documentation still says to use it the numpy way (in brackets after the description), but in pandas-dev the consensus seems to be keeping the pandas way, which was already right in this case. Will update the documentation asap and send an email with the last minute changes to the organizers

jorisvandenbossche: Yes, the docstring guide has been updated, but apparently only in the source: https://github.com/python-sprints/python-sprints.github.io/blob/master/pandas/guide/source/pandas_docstring.rst and not yet in the online version

stijnvanhoey:
There is indeed a difference between the online version and the `rst` source. I used the online guide to do the adaptations. Fixed now

pandas/core/indexes/base.py
```diff
@@ -1122,7 +1122,25 @@ def to_frame(self, index=True):
 
         Returns
         -------
-        DataFrame : a DataFrame containing the original Index data.
+        DataFrame
+            DataFrame containing the original Index data.
+
+        Examples
+        --------
+        >>> idx = pd.Index(['Ant', 'Bear', 'Cow'])
```
datapythonista: Minor thing, but it'd probably look a bit nicer if you create the index with `name='animal'`. :) For the rest looks great to me.

pandas/core/indexes/base.py
```diff
@@ -1110,13 +1110,14 @@ def to_series(self, index=None, name=None):
...
-        """Create a DataFrame with a column containing the Index.
+        """
+        Create a DataFrame with a column containing the Index.
 
         .. versionadded:: 0.21.0
 
         Parameters
         ----------
-        index : boolean (default True)
+        index : boolean, default True
```
stijnvanhoey: @jorisvandenbossche on the [docstring guide rst file](https://github.com/python-sprints/python-sprints.github.io/blob/master/pandas/guide/source/pandas_docstring.rst), the example dosctring provides: `color : str (default 'blue')`. So should it be `(default True)` or `, default True`?

jorisvandenbossche: Whoops, I was using the virtual in my head :-) That's something we need to fix, we decided on `, default True` I think

pandas/core/indexes/base.py
stijnvanhoey: Well, I do think it is interesting when creating a new index:
```
>>> idx.to_frame(index = False)
  animal
0    Ant
1   Bear
2    Cow
```
but when using the same as index and column, it does not get 'nicer':

```
>>> idx.to_frame(index = True)
       animal
animal       
Ant       Ant
Bear     Bear
Cow       Cow
```

pandas/core/indexes/base.py
stijnvanhoey: See, https://github.com/python-sprints/python-sprints.github.io/pull/62

PR #19704, opened 2018-02-14
DOC: Adding guide for the pandas documentation sprint

datapythonista:
This PR is to make it easier to review the proposal guide for the pandas documentation sprint, as discussed in pandas-dev.

PR #20037, opened 2018-03-07
DOC: add small guide on how to write examples that pass doctests

jorisvandenbossche:
Some initial attention points for getting doctests passing (this would eventually fit in the docstring guide, putting it in contributing.rst for now, will move once the other is merged).

Anybody else thinks of typical gotcha's?

cc @datapythonista

### DIFF
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -430,6 +430,70 @@ the documentation are also built by Travis-CI. These docs are then hosted `here
 <http://pandas-docs.github.io/pandas-docs-travis>`__, see also
 the :ref:`Continuous Integration <contributing.ci>` section.
 
+
+Writing examples that pass the doctest
+--------------------------------------
+
+The "Examples" section in the docstrings follows the `doctest <https://docs.python.org/3.6/library/doctest.html>`__
+format (lines beginning with ``>>>`` for code), and are also tested to ensure the
+correctness of the examples shown in the API documentation.
+
+In general, the format looks like:
+
+::
+
+    Examples
+    --------
+
+    A small example:
+
+    >>> s = pd.Series([1, 2, 3])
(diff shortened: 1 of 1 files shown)


doc/source/contributing.rst
```diff
@@ -430,6 +430,70 @@ the documentation are also built by Travis-CI. These docs are then hosted `here
 <http://pandas-docs.github.io/pandas-docs-travis>`__, see also
 the :ref:`Continuous Integration <contributing.ci>` section.
 
+
```
jreback: can you add sub-section refs

jreback (changes requested):

datapythonista:
One that comes to my mind is reading from files, like `pandas.read_csv`, that I assume it fails, as the file won't be found.

jorisvandenbossche:
> One that comes to my mind is reading from files, like pandas.read_csv, that I assume it fails, as the file won't be found.

The question is what to do in general for those IO functions. 
Do we want to show actual output? But then there needs to be an actual file to read in, as otherwise it also does not say much. 

In principle you could always create a small file to then read back in.

jorisvandenbossche:
In general, for read functions, we could do like:

```
For the example, we first create a small example csv file

>>> df = pd.DataFrame(...)
>>> df.to_csv('file.csv')

We can now read that in with read_csv:

>>> pd.read_csv('file.csv')

read with option:

>>> ...
```

for write functions that is a bit more difficult, as you don't directly see what you get. Maybe just showing the basic usage (method call) and then referring to user guide is enough for those?

jreback (approved):

jorisvandenbossche:
This was included in the merge of https://github.com/pandas-dev/pandas/pull/19704, so closing.

Issue #11042, opened 2015-09-10
DOC: document optional arguments on plot submethods


shoyer:
As a follow up on #9321:

We should start adding optional arguments that are common between multiple plot types to the signatures for plotting sub-methods, such as `figsize` and `axis`. Currently, these are only documented on the main `.plot()` methods and we use `**kwargs` for the rest. This would be nice because we could clarify for users which optional arguments work for which types of plots -- this is not at all obvious right now.

@JanSchulz [suggests](https://github.com/pydata/pandas/pull/9321#discussion_r39086063) we could perhaps do this dynamically by updating the `__signature__` object. However, I don't know yet if this will generate the correct API documentation -- maybe if Sphinx is run using Python 3?

123pm123:
Europython sprint group has a look at this

ataleb:
I would like to work on this issue, if it is still available?

Coder-Harshit:
is this issue still open ... after almost a decade ?

Dronten:
is this issue still up , can i  work on this ?

zhaohel:
take

Armaan1Gohil:
take

LeonxLJX:
I'd like to take this one.

Reading the issue, this looks like a documentation-clarity gap rather than a code defect: the current page doesn't give readers enough to self-diagnose, which is why it surfaces here as an issue instead of being answered by the docs.

Plan: locate the relevant section, confirm what the code actually does today (so the docs match behavior, not intent), tighten the wording, and scan neighboring pages for the same inconsistency — otherwise the fix just relocates the confusion.

One thing to confirm before writing: minimal correction scoped to this page, or also cross-link from related guides if the same gap shows up there?

(claiming via @LeonxLJX)

PR #20041, opened 2018-03-07
DOC: Improve pandas.Series.plot.kde docstring and kwargs rewording for whole file

dukebody:
Initial docstring assigned to the Barcelona Chapter for the documentation sprint.

Credits to @arnau126 who also contributed to this change.

Also, rewording of `kwargs` argument explanation for the whole file, where we refer to the `pandas.Series.plot` docstring for info.

### DIFF
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2532,7 +2532,8 @@ class SeriesPlotMethods(BasePlotMethods):
         Parameters
         ----------
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.Series.plot`.
 
         Returns
         -------
@@ -2556,7 +2557,8 @@ class SeriesPlotMethods(BasePlotMethods):
         Parameters
         ----------
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.Series.plot`.
 
         Returns
         -------
@@ -2571,7 +2573,8 @@ class SeriesPlotMethods(BasePlotMethods):
         Parameters
         ----------
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.Series.plot`.
 
         Returns
         -------
@@ -2586,7 +2589,8 @@ class SeriesPlotMethods(BasePlotMethods):
         Parameters
         ----------
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.Series.plot`.
 
         Returns
         -------
@@ -2603,7 +2607,8 @@ class SeriesPlotMethods(BasePlotMethods):
         bins: integer, default 10
             Number of histogram bins to be used
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.Series.plot`.
 
         Returns
         -------
@@ -2633,7 +2638,8 @@ class SeriesPlotMethods(BasePlotMethods):
             kde is evaluated at the points passed. If `ind` is an integer,
             `ind` number of equally spaced points are used.
         kwds : optional
-            Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.Series.plot`.
 
         Returns
         -------
@@ -2692,7 +2698,8 @@ class SeriesPlotMethods(BasePlotMethods):
         Parameters
         ----------
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.Series.plot`.
 
         Returns
         -------
@@ -2707,7 +2714,8 @@ class SeriesPlotMethods(BasePlotMethods):
         Parameters
         ----------
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.Series.plot`.
 
         Returns
         -------
@@ -2758,7 +2766,8 @@ class FramePlotMethods(BasePlotMethods):
         x, y : label or position, optional
             Coordinates for each point.
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
@@ -2775,7 +2784,8 @@ class FramePlotMethods(BasePlotMethods):
         x, y : label or position, optional
             Coordinates for each point.
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
@@ -2792,7 +2802,8 @@ class FramePlotMethods(BasePlotMethods):
         x, y : label or position, optional
             Coordinates for each point.
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
@@ -2809,7 +2820,8 @@ class FramePlotMethods(BasePlotMethods):
         by : string or sequence
             Column in the DataFrame to group by.
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
@@ -2828,7 +2840,8 @@ class FramePlotMethods(BasePlotMethods):
         bins: integer, default 10
             Number of histogram bins to be used
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
@@ -2853,7 +2866,8 @@ class FramePlotMethods(BasePlotMethods):
             points passed. If `ind` is an integer, `ind` number of equally
             spaced points are used.
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
@@ -2872,7 +2886,8 @@ class FramePlotMethods(BasePlotMethods):
         x, y : label or position, optional
             Coordinates for each point.
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
@@ -2889,7 +2904,8 @@ class FramePlotMethods(BasePlotMethods):
         y : label or position, optional
             Column to plot.
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
@@ -2910,7 +2926,8 @@ class FramePlotMethods(BasePlotMethods):
         c : label or position, optional
             Color of each point.
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Additional keyword arguments are documented in
+            :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


pandas/plotting/_core.py
```diff
@@ -2604,26 +2604,73 @@ def hist(self, bins=10, **kwds):
...
+        kwds : optional
             Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
 
         Returns
         -------
         axes : matplotlib.AxesSubplot or np.array of them
+
+        See also
+        --------
+        :class:`scipy.stats.gaussian_kde` : Representation of a kernel-density
```
jorisvandenbossche: Can you check if this works without the ```:class:`..` ```? Normally that should work I think (in the see also section this linking is done automatically)

pandas/plotting/_core.py
```diff
@@ -2604,26 +2604,73 @@ def hist(self, bins=10, **kwds):
...
-            are used. If `ind` is a NumPy array, the kde is evaluated at the
-            points passed. If `ind` is an integer, `ind` number of equally
-            spaced points are used.
-        `**kwds` : optional
+            Evaluation points for the estimated PDF. If None (default),
+            1000 equally spaced points are used. If `ind` is a NumPy array, the
+            kde is evaluated at the points passed. If `ind` is an integer,
+            `ind` number of equally spaced points are used.
+        kwds : optional
             Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
```
jorisvandenbossche: I am wondering if we can make this more clear, because in the end, what the user needs to know is basically that the other keywords are documented there. 
Would you find something like "Additional keyword arguments are documented in pandas.Series.plot" clearer ?

pandas/plotting/_core.py
```diff
@@ -2604,26 +2604,73 @@ def hist(self, bins=10, **kwds):
...
+
+        See also
+        --------
+        :class:`scipy.stats.gaussian_kde` : Representation of a kernel-density
+            estimate using Gaussian kernels. This is the function used
+            internally to estimate the PDF.
+
+        Examples
+        --------
+        Given a ``Series`` of points randomly sampled from an unknown
```
jorisvandenbossche: generally in docstrings we don't quote Series or DataFrame (but we don't really have a rule about it)

pandas/plotting/_core.py
```diff
@@ -2604,26 +2604,73 @@ def hist(self, bins=10, **kwds):
...
+        distribution, estimate this distribution using KDE with automatic
+        bandwidth determination and plot the results, evaluating them at
+        1000 equally spaced points (default):
+
+        .. plot::
+            :context: close-figs
+
+            >>> s = pd.Series([1, 2, 2.5, 3, 3.5, 4, 5])
+            >>> p = s.plot.kde()
```
jorisvandenbossche: The `p = ` is to ensure there is no output (repr of the matplotlib object)? (so the doctests pass?) 
An alternative is to do `s.plot.kde();` (otherwise, I would replace `p` with `ax`, as this is the more common used name for such an object)

pandas/plotting/_core.py
```diff
@@ -2604,26 +2604,73 @@ def hist(self, bins=10, **kwds):
...
+
+        .. plot::
+            :context: close-figs
+
+            >>> p = s.plot.kde(bw_method=3)
+
+        .. plot::
+            :context: close-figs
+
+            >>> p = s.plot.kde(bw_method=0.3)
```
jorisvandenbossche: maybe switch the order (first small then large value), so it matches the order it is mentioned in the text

jorisvandenbossche (commented):
Great! Added some comments

pandas/plotting/_core.py
dukebody: Yes, it works!

pandas/plotting/_core.py
dukebody: I don't have a strong opinion about the wording, but I think it should be consistent across all docstrings. Right now other docstrings use the `Keyword arguments to pass on to :py:meth:pandas.Series.plot.` wording, so I'd vote for not changing this one now, and decide if we want to change all of them in a row later after the sprint.

What do you think?

pandas/plotting/_core.py
dukebody: Yes, exactly! I will use the `ax` alternative.

dukebody:
@jorisvandenbossche Thanks for the review! I've updated the PR in response to your comments.

pandas/plotting/_core.py
jorisvandenbossche: *If* we think it is a better explanation, I would change it now, and then when updating the other docstrings during the sprint, they can take this as an example

jorisvandenbossche:
There is also issue https://github.com/pandas-dev/pandas/issues/11042 on adding the explanation of more keyword arguments.

pandas/plotting/_core.py
dukebody: Hmm I believe this can create a bit of confusion if we have to communicate to everyone in the sprint that this is the docstring they have to take as example and always change the existing wording for the new one, but you probably know better than me. :)

With regards to the explanation itself, know that I've re-read your proposal I believe that yes, it's clearer than the existing one, because the user doesn't need to know that `pandas.Series.plot.kde` calls `pandas.Series.plot` under the hood, just that `pandas.Series.plot.kde` accepts certain keyword arguments and what they mean.

dukebody:
What about the travis failures? Is it just that the tests are flaky, or the docstring changes can really break something? I.e. are docstring doctests executed by CI?

pandas/plotting/_core.py
jorisvandenbossche: you can also change it directly everywhere in this PR, that would indeed cause less confusion

jorisvandenbossche:
> What about the travis failures? Is it just that the tests are flaky, or the docstring changes can really break something? I.e. are docstring doctests executed by CI?

Don't worry about those, we have some failing tests due to the new matplotlib release we need to fix.

pandas/plotting/_core.py
dukebody: I can do this if we decide the wording is better. Whose input should we wait for?

I guess it's better to do it in a separate commit and then change the title of the PR. Or perhaps another PR?

jorisvandenbossche: It's fine to do it in this PR. 
And if you and I agree on it, I think it is fine to do the change now (although it can of course be a bit annoying to go back and forth on a change if then other people disagree).

dukebody: Done! I thought this would appear in more files, but it turns out that it's only in this one.

pandas/plotting/_core.py
```diff
@@ -2532,7 +2532,8 @@ def line(self, **kwds):
         Parameters
         ----------
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
+            Additional keyword arguments are documented in
+            :py:meth:`pandas.Series.plot`.
```
jorisvandenbossche: sorry for not saying earlier, but you can leave out the `:py` part (that makes it a bit less noisy in the plain text version)

(in principle you can also leave out the 'pandas.' part)

pandas/plotting/_core.py
```diff
@@ -2711,7 +2766,8 @@ def line(self, x=None, y=None, **kwds):
         x, y : label or position, optional
             Coordinates for each point.
         `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Additional keyword arguments are documented in
+ :py:meth:`pandas.DataFrame.plot`.
```
jorisvandenbossche: here went something wrong with the indentation (and some below as well)

dukebody: Sorry, fixing...

pandas/plotting/_core.py
dukebody: Removing the `:py` part but leaving the `pandas.` one.

Related question: What is the difference between `:meth:` and `:class:`? Do they behave differently in docstrings?

jorisvandenbossche: The end goal is to create a link to the help page of that method. But I am not sure how essential it is to use the correct prefix. From experience I know both :func: and :meth: work for methods. But anyway, for the plain text version it is good to use the correct one anyhow (the difference between :meth: and :class: might be how they are displayed in html, with or without brackets)

dukebody:
I believe this should be ready to go. @jorisvandenbossche please do a last review!

pandas/plotting/_core.py
```diff
@@ -2613,26 +2618,74 @@ def hist(self, bins=10, **kwds):
...
+        1000 equally spaced points (default):
+
+        .. plot::
+            :context: close-figs
+
+            >>> s = pd.Series([1, 2, 2.5, 3, 3.5, 4, 5])
+            >>> ax = s.plot.kde()
+
+
+        An scalar fixed bandwidth can be specified. Using a too small bandwidth
```
shivam6294: Great work @dukebody! :+1: This is Shivam from the HK Chapter of the Pandas Doc Sprint. 

Just had a minor suggestion for simplifying this paragraph: 
```
        A scalar bandwidth can be specified. Using a small bandwidth value can
        lead to overfitting, while using a large bandwidth value can result 
        in underfitting:
```

pandas/plotting/_core.py
```diff
@@ -2613,26 +2618,74 @@ def hist(self, bins=10, **kwds):
 
     def kde(self, bw_method=None, ind=None, **kwds):
         """
-        Kernel Density Estimate plot
+        Kernel Density Estimate plot using Gaussian kernels.
+
+        In statistics, kernel density estimation (KDE) is a non-parametric way
+        to estimate the probability density function (PDF) of a random
```
jreback: ideally include a reference to a wiki page (same ones that matplotlib uses?)

pandas/plotting/_core.py
```diff
@@ -2613,26 +2618,74 @@ def hist(self, bins=10, **kwds):
...
             If None (default), 'scott' is used.
             See :class:`scipy.stats.gaussian_kde` for more information.
         ind : NumPy array or integer, optional
-            Evaluation points. If None (default), 1000 equally spaced points
-            are used. If `ind` is a NumPy array, the kde is evaluated at the
-            points passed. If `ind` is an integer, `ind` number of equally
-            spaced points are used.
-        `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.Series.plot`.
+            Evaluation points for the estimated PDF. If None (default),
```
jreback: can you spell out PDF

pandas/plotting/_core.py
```diff
@@ -2613,26 +2618,74 @@ def hist(self, bins=10, **kwds):
...
 
         Returns
         -------
         axes : matplotlib.AxesSubplot or np.array of them
+
+        See also
+        --------
+        scipy.stats.gaussian_kde : Representation of a kernel-density
+            estimate using Gaussian kernels. This is the function used
+            internally to estimate the PDF.
```
jreback: spell out PDF (can also include PDF in parens)

jreback (changes requested):

pandas/plotting/_core.py
jonas-schulze: I agree, @shivam6294 version sounds a little better. Anyways, you have to change `An` to `A` as `scalar` doesn't start with a vowel.

pandas/plotting/_core.py
jonas-schulze: Both KDE and DPF are spelled out in the small summary above the parameters section. I wouldn't like to repeat that over here. @dukebody what do you think?

jreback: maybe if you put KDE and PDF in parents then its more clear (do this above)

pandas/plotting/_core.py
```diff
@@ -2613,26 +2618,74 @@ def hist(self, bins=10, **kwds):
 
     def kde(self, bw_method=None, ind=None, **kwds):
         """
-        Kernel Density Estimate plot
+        Kernel Density Estimate plot using Gaussian kernels.
```
jonas-schulze: The short summary should start with an infinitive verb. I would change it to
```
Draw Kernel Density Estimate plot using Gaussian kernels.
```

pandas/plotting/_core.py
jreback: nvm I see it.

pandas/plotting/_core.py
jorisvandenbossche: I think for plot functions it is fine to just state the plot type (the rules are to give direction, but sometimes there can be reason to deviate, so in this case the question is if "Draw .." makes it more informative)

jreback (approved):

pandas/plotting/_core.py
jonas-schulze: I agree in that adding `Draw` doesn't add much information, but I would still add a prefix since most (all?) of the doc strings created in the Doc Sprint will be written with these rules in mind. Also, the doc string proposed in [PR20113](https://github.com/pandas-dev/pandas/pull/20113/) for the `hist` function includes `Draw` as well.

On the other hand, `Generate` might be a better fit than `Draw` because out of a jupyter notebook there is nothing drawn ..

pandas/plotting/_core.py
```diff
@@ -2613,26 +2618,74 @@ def hist(self, bins=10, **kwds):
...
+            `ind` number of equally spaced points are used.
+        kwds : optional
+            Additional keyword arguments are documented in
+            :meth:`pandas.Series.plot`.
 
         Returns
         -------
         axes : matplotlib.AxesSubplot or np.array of them
+
+        See also
```
jonas-schulze: It would be nice to have an additional reference:
```
:meth:`pandas.DataFrame.plot.kde` : Generate a KDE plot for a DataFrame
```

jonas-schulze (changes requested):
Good work!

pandas/plotting/_core.py
jreback: See also -> See Also

jreback (changes requested):

jonas-schulze (changes requested):

jorisvandenbossche:
@dukebody Thanks! We will do the follow-up in https://github.com/pandas-dev/pandas/pull/20236

PR #20043, opened 2018-03-07
Refactored GroupBy ASVs

### LINKED ISSUE #19733: Add ASV Benchmarks for GroupBy Ops on Datetimelike
xref comment in #19673, we don't have any ASV benchmarks in `groupby.GroupByMethods` to measure performance on datetimelike types (timestamp / timedelta)

WillAyd:
- [X] closes #19733

I ended up mixing in some cleanup as part of this change

### DIFF
--- a/asv_bench/benchmarks/groupby.py
+++ b/asv_bench/benchmarks/groupby.py
@@ -364,8 +364,8 @@ class GroupByMethods(object):
             key = date_range('1/1/2011', periods=size, freq='s')
 
         df = DataFrame({'values': values, 'key': key})
-        self.as_group_method =  getattr(df.groupby('key')['values'], method)
-        self.as_field_method =  getattr(df.groupby('values')['key'], method)
+        self.as_group_method = getattr(df.groupby('key')['values'], method)
+        self.as_field_method = getattr(df.groupby('values')['key'], method)
 
     def time_dtype_as_group(self, dtype, method):
         self.as_group_method()
--- a/asv_bench/benchmarks/groupby.py
+++ b/asv_bench/benchmarks/groupby.py
@@ -231,7 +231,7 @@ class AggFunctions(object):
 
     goal_time = 0.2
 
-    def setup_cache(self):
+    def setup_cache():
         N = 10**5
         fac1 = np.array(['A', 'B', 'C'], dtype='O')
         fac2 = np.array(['one', 'two'], dtype='O')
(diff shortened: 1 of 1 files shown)


asv_bench/benchmarks/groupby.py
```diff
@@ -90,45 +93,6 @@ def time_series_groups(self, data, key):
         self.ser.groupby(self.ser).groups
 
 
-class FirstLast(object):
```
WillAyd: I removed this class because it is duplicative of other tests. `first` and `last` are already covered in the `GroupByMethods` class. For the `nth` tests (although they are arguably duplicated of `first`) I ended up moving them all to one shared `Nth` class that covers the various dtype permutations

asv_bench/benchmarks/groupby.py
```diff
@@ -149,39 +113,40 @@ class Nth(object):
 
     goal_time = 0.2
 
-    def setup_cache(self):
```
WillAyd: Not sure if there is a reason why we are intermixing `setup_cache` and `setup` within this module. The former seems preferable given these setups aren't super resource intensive and when refactoring this to use `setup` I didn't encounter any issues locally. May be worth a separate cleanup to consistently use `setup` throughout the module

asv_bench/benchmarks/groupby.py
```diff
@@ -243,7 +208,7 @@ def time_multi_count(self, df):
         df.groupby(['key1', 'key2']).count()
 
 
-class CountInt(object):
+class CountMultiInt(object):
```
WillAyd: `int` tests for `count` / `nunique` are already covered in `GroupByMethods` though the differentiator here is that we are testing the use of multiple fields in the grouping. It may make sense to just do that natively in `GroupByMethods` but I didn't want to rip up too many things as of yet

asv_bench/benchmarks/groupby.py
```diff
@@ -398,12 +360,18 @@ def setup(self, dtype, method):
...
-        self.df_groupby_method = getattr(df.groupby('key')['values'], method)
+        self.as_group_method =  getattr(df.groupby('key')['values'], method)
+        self.as_field_method =  getattr(df.groupby('values')['key'], method)
+
+    def time_dtype_as_group(self, dtype, method):
+        self.as_group_method()
 
-    def time_method(self, dtype, method):
-        self.df_groupby_method()
+    def time_dtype_as_field(self, dtype, method):
```
WillAyd: Before this was only testing the dtype as part of the grouping, but it makes sense to test it as an aggregated / transformed field as well

asv_bench/benchmarks/groupby.py
jreback: that's fine

jreback (commented):
maybe add .transform for methods which we support? (not sure if these are already in a separate place)

asv_bench/benchmarks/groupby.py
mroeschke: While refactoring the benchmarks, I mainly preferred `setup_cache` over `setup` to avoid running the setup n times for each method of the benchmarking class. 

I think it should be okay to change them to just `setup`, just need to monitor that it will not significantly increase the time of the ASV CI run on travis. (which I don't think it will)

WillAyd: Thanks @mroeschke that makes sense. Should be easy to revert if it causes an issue

WillAyd:
Travis failures is from `TestTSPlot` and isn't related to this change

jreback:
thanks!

PR #20044, opened 2018-03-07
DOC: Update Kurt Docstr

WillAyd:
Continuation of #19999


WillAyd:
@jorisvandenbossche

pandas/core/window.py
```diff
@@ -910,8 +910,9 @@ def skew(self, **kwargs):
 
     Returns
     -------
-    Series or DataFrame (matches input)
-        Like-indexed object containing the result of function application
+    Series or DataFrame
+        Returned object type is dictated by the caller of the %(name)s
+            calculation
```
jorisvandenbossche: align with the "Returned .." from above

pandas/core/window.py
```diff
@@ -910,8 +910,9 @@ def skew(self, **kwargs):
 
     Returns
     -------
-    Series or DataFrame (matches input)
-        Like-indexed object containing the result of function application
+    Series or DataFrame
+        Returned object type is dictated by the caller of the %(name)s
```
jorisvandenbossche: As a non-native speaker, I found your original "determined" better than "dictated" (but maybe not more correct :-)). Or simpler "Returned object is of the same type as the caller ..." ?

pandas/core/window.py
```diff
@@ -932,19 +933,20 @@ def skew(self, **kwargs):
...
     3.999946
-    >>> df = pd.DataFrame(arr)
-    >>> df.rolling(4).kurt()
-              0
-    0       NaN
-    1       NaN
-    2       NaN
-    3 -1.200000
-    4  3.999946
+    >>> ser = pd.Series(arr)
```
jorisvandenbossche: let's use `s` by default (following the guide)

jorisvandenbossche (commented):
Looks good! some minor comments

WillAyd:
Travis failures is from `TestTSPlot` and isn't related to this change

toobaz:
Looks good!

jorisvandenbossche:
@toobaz do we merge this one first, and then you update your PR?

toobaz:
Yes, I will rebase, no problem.

jorisvandenbossche:
Just made a small change to "Like" -> "Equivalent" (I *think* you cannot use "like" as an adjective like that? but may be wrong)

jorisvandenbossche:
@WillAyd Thanks for the updates!

PR #20049, opened 2018-03-08
API: PeriodIndex subtraction to return object Index of DateOffsets

### LINKED ISSUE #13077: ENH/API: Decide what to return Period - Period subtraction

Split from #13071. Currently it returns int and loses freq info.

```
pd.Period('2011-03', freq='M') - pd.Period('2011-01', freq='M') 
#2
```
#### Expected Output

Current ideas are:
1. Return `Timedelta` adding support of `Monthly` freq.
2. Return `Perioddelta` newly created.
3. Return `DateOffset`.
#### output of `pd.show_versions()`

on current master.

CC: @jreback @MaximilianR

jbrockmendel:
Implements _sub_period_array in DatetimeIndexOpsMixin.  Behavior is analogous to subtraction of Period scalar.

- [x] closes #13077
- [x] tests added / passed
- [x] whatsnew entry


pandas/core/indexes/datetimelike.py
```diff
@@ -667,6 +668,43 @@ def _sub_nat(self):
...
+            dtype is np.int64 if there are no NaTs in self or other,
+            otherwise np.float64
+        """
+        if not is_period_dtype(self):
+            raise TypeError("cannot subtract {dtype}-dtype to {cls}"
+                            .format(dtype=other.dtype,
+                                    cls=type(self).__name__))
+
+        if not len(self) == len(other):
+            raise ValueError("cannot add indices of unequal length")
```
jschendel: I might have missed it, but is there a test that hits this?  And should it say "subtract" instead of "add"?

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -710,6 +710,7 @@ Other API Changes
 - ``Categorical.fillna`` now validates its ``value`` and ``method`` keyword arguments. It now raises when both or none are specified, matching the behavior of :meth:`Series.fillna` (:issue:`19682`)
 - ``pd.to_datetime('today')`` now returns a datetime, consistent with ``pd.Timestamp('today')``; previously ``pd.to_datetime('today')`` returned a ``.normalized()`` datetime (:issue:`19935`)
 - :func:`Series.str.replace` now takes an optional `regex` keyword which, when set to ``False``, uses literal string replacement rather than regex replacement (:issue:`16808`)
+- :class:`PeriodIndex` subtraction of another ``PeriodIndex`` will now return a numeric ``Index`` instead of raising a ``TypeErrorr`` (:issue:`20049`)
```
jschendel: Extra "r" at the end of "TypeErrorr"

pandas/core/indexes/datetimelike.py
jbrockmendel: Good catches on both counts.  Will fix.

jreback:
I think we should be returning an array (Index?) of offsets rather than the integers which are untyped.

we don't really support this so an object Index would be ok I think.

jbrockmendel:
> I think we should be returning an array (Index?) of offsets rather than the integers which are untyped.

I more-or-less agree, but ATM Period - Period returns an integer and PeriodIndex - Period returns an integer array.  These three operations should all be consistent.

jreback:
> I more-or-less agree, but ATM Period - Period returns an integer and PeriodIndex - Period returns an integer array. These three operations should all be consistent.

and these should changed (both). yes its an API breaking change but much more intuitve / better. it will just have to be an object array but that's ok.

jbrockmendel:
> and these should changed (both). yes its an API breaking change but much more intuitve / better. it will just have to be an object array but that's ok.

Implementing this, the non-obvious issue that comes up is what to do with `NaT`s in a `PeriodIndex` when subtracting `Period`.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -996,6 +996,7 @@ Other API Changes
 - :func:`DataFrame.to_dict` with ``orient='index'`` no longer casts int columns to float for a DataFrame with only int and float columns (:issue:`18580`)
 - A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousins, will now *always* be passed a ``Series``, rather than a ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
 - Rolling and Expanding types raise ``NotImplementedError`` upon iteration (:issue:`11704`).
+>>>>>>> cbec58eacd8e9cd94b7f42351b8de4559c250909
```
jreback: leftover

pandas/tests/indexes/datetimes/test_arithmetic.py
```diff
@@ -740,6 +740,17 @@ def test_sub_period(self, freq):
         with pytest.raises(TypeError):
             p - idx
 
+    @pytest.mark.parametrize('op', [operator.add, ops.radd,
```
jreback: can you add (maybe another test) add/sub with a pi and another dtype (I don't think we have this test?)

pandas/tests/indexes/period/test_arithmetic.py
```diff
@@ -258,6 +258,44 @@ def test_comp_nat(self, dtype):
...
 
 class TestPeriodIndexArithmetic(object):
+    # ---------------------------------------------------------------
+    # __add__/__sub__ with PeriodIndex
+
+    def test_pi_add_iadd_pi_raises(self):
+        rng = pd.period_range('1/1/2000', freq='D', periods=5)
+        other = pd.period_range('1/6/2000', freq='D', periods=5)
+
+        # previously performed setop union, now raises TypeError (GH14164)
```
jreback: this is a pretty old comment, can you improve

pandas/tests/indexes/period/test_arithmetic.py
```diff
@@ -258,6 +258,44 @@ def test_comp_nat(self, dtype):
 
 
 class TestPeriodIndexArithmetic(object):
+    # ---------------------------------------------------------------
+    # __add__/__sub__ with PeriodIndex
+
```
jreback: can you give a 1-2 line on what are allowed ops on PI

pandas/tests/indexes/period/test_arithmetic.py
```diff
@@ -419,17 +446,20 @@ def test_pi_sub_intlike(self, five):
         exp = rng + (-five)
         tm.assert_index_equal(result, exp)
 
-    def test_pi_sub_isub_pi_raises(self):
-        # previously performed setop, now raises TypeError (GH14164)
-        # TODO needs to wait on #13077 for decision on result type
+    def test_pi_sub_isub_pi(self):
+        # GH#20049
+        # previously raised TypeError (GH#14164), before that
```
jreback: can you refresh comment

jreback (changes requested):
looks pretty good. rebase, small comments. @jschendel can you give a once-over

pandas/tests/indexes/datetimes/test_arithmetic.py
jbrockmendel: Just checking, you mean to add cases over in the period test_arithmetic file, not here, right?

pandas/core/indexes/datetimelike.py
jbrockmendel: Just pushed with fix to add-->subtract.  test_pi_sub_pi, test_pi_sub_pi_with_nat, test_pi_sub_isub_pi each go through this method, though I don't think any go through this particular line

jbrockmendel:
Removing `test_pi_sub_pi` because it is entirely redundant with `test_pi_sub_isub_pi`.  Moving the latter to where it _should_ be in the test file.

jorisvandenbossche (commented):
Shouldn't first the Period scalar substraction be changed as well?

jbrockmendel:
@jorisvandenbossche yes, thats in #21314.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -68,6 +68,7 @@ Datetimelike API Changes
 
 - For :class:`DatetimeIndex` and :class:`TimedeltaIndex` with non-``None`` ``freq`` attribute, addition or subtraction of integer-dtyped array or ``Index`` will return an object of the same class (:issue:`19959`)
 - :class:`DateOffset` objects are now immutable. Attempting to alter one of these will now raise ``AttributeError`` (:issue:`21341`)
+- :class:`PeriodIndex` subtraction of another ``PeriodIndex`` will now return an object-dtype :class:`Index` of :class:`DateOffset` objects instead of raising a ``TypeErrorr`` (:issue:`20049`)
```
jschendel: Extra "r" at the end of "TypeErrorr"

pandas/tests/indexes/period/test_arithmetic.py
```diff
@@ -258,6 +258,57 @@ def test_comp_nat(self, dtype):
...
+        result = rng - other
+        tm.assert_index_equal(result, expected)
+
+        rng -= other
+        tm.assert_index_equal(rng, expected)
+
+    def test_pi_sub_pi_with_nat(self):
+        rng = pd.period_range('1/1/2000', freq='D', periods=5)
+        other = rng[1:].insert(0, pd.NaT)
+        assert other[1:].equals(rng[1:])
```
jschendel: Is this `assert` necessary? I don't immediately see why it's needed, but could be missing something.  If it is needed, can `assert_index_equal` be used instead?

jbrockmendel: This is checking that I haven't already screwed up, i.e. that the test below is indeed testing what I think it is testing.

jbrockmendel:
gentle ping (this should be tandem with #21314)

jreback (approved):

jreback:
thanks @jbrockmendel 

also pls have a look at the period docs to see if we need updating anywhere.

PR #20050, opened 2018-03-08
DOC: Updated docstring DatetimeIndex.tz_localize

IHackPy:
Kindly review this docstring DatetimeIndex.tz_localize

### DIFF
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1973,16 +1973,15 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
 
         Examples
         --------
-        In the example below, We put the date range from 01 March 2018 to
-	08 March 2018 & convert this to US/Eastern Time zone
+        In the example below, We put the date range from 01 March 2018 to 08 March 2018 & convert this to US/Eastern Time zone
  
         >>> df = pd.date_range('2018-03-01', '2018-03-08')
-        >>> df
-    	DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
+	>>> df
+	DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
                '2018-03-05', '2018-03-06', '2018-03-07', '2018-03-08'],
               dtype='datetime64[ns]', freq='D')
-        >>> df.tz_localize(tz='US/Eastern')
-    	DatetimeIndex(['2018-03-01 00:00:00-05:00', '2018-03-02 00:00:00-05:00',
+	>>> df.tz_localize(tz='US/Eastern')
+	DatetimeIndex(['2018-03-01 00:00:00-05:00', '2018-03-02 00:00:00-05:00',
                '2018-03-03 00:00:00-05:00', '2018-03-04 00:00:00-05:00',
                '2018-03-05 00:00:00-05:00', '2018-03-06 00:00:00-05:00',
                '2018-03-07 00:00:00-05:00', '2018-03-08 00:00:00-05:00'],
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1939,8 +1939,9 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
         """
         Localize tz-naive DatetimeIndex to tz-aware DatetimeIndex.
 
-        tz_localize() takes a naive datetime object and interprets it as if it
-        is in that timezone. It does not move the time to another timezone.
+        tz_localize() takes a naive DatetimeIndex object and interprets it as
+        if it is in that timezone. It does not move the time to another
+        timezone.
         tz_localize function helps to switch b/w time zone aware and time zone
         unaware objects.
 
@@ -1984,18 +1985,25 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
         we perform reverse operation where we remove tize zone & make it
         tz-naive
 
-        >>> dti = pd.date_range('2018-03-01', '2018-03-05')
-        >>> dti
-        DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03',
+        In [2]: dti = pd.date_range('2018-03-01', '2018-03-05')
+
+        In [3]: dti
+
+        Out[3]: DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03',
         '2018-03-04', '2018-03-05'], dtype='datetime64[ns]', freq='D')
-        >>> dti.tz_localize(tz='US/Eastern')
-        DatetimeIndex(['2018-03-01 00:00:00-05:00',
+
+        In [4]: dti.tz_localize(tz='US/Eastern')
+
+        Out[4]: DatetimeIndex(['2018-03-01 00:00:00-05:00',
         '2018-03-02 00:00:00-05:00', '2018-03-03 00:00:00-05:00',
         '2018-03-04 00:00:00-05:00', '2018-03-05 00:00:00-05:00'],
         dtype='datetime64[ns, US/Eastern]', freq='D')
-        >>> dti.tz_localize(None)
-        DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03',
+
+        In [5]: dti.tz_localize('US/Eastern').tz_localize(None)
+
+        Out[5]: DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03',
         '2018-03-04', '2018-03-05'], dtype='datetime64[ns]', freq='D')
+
 
         Raises
         ------
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1973,15 +1973,16 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
 
         Examples
         --------
-        In the example below, We put the date range from 01 March 2018 to 08 March 2018 & convert this to US/Eastern Time zone
+        In the example below, We put the date range from 01 March 2018 to
+	08 March 2018 & convert this to US/Eastern Time zone
  
         >>> df = pd.date_range('2018-03-01', '2018-03-08')
         >>> df
-    DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
+    	DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
                '2018-03-05', '2018-03-06', '2018-03-07', '2018-03-08'],
               dtype='datetime64[ns]', freq='D')
         >>> df.tz_localize(tz='US/Eastern')
-    DatetimeIndex(['2018-03-01 00:00:00-05:00', '2018-03-02 00:00:00-05:00',
+    	DatetimeIndex(['2018-03-01 00:00:00-05:00', '2018-03-02 00:00:00-05:00',
                '2018-03-03 00:00:00-05:00', '2018-03-04 00:00:00-05:00',
                '2018-03-05 00:00:00-05:00', '2018-03-06 00:00:00-05:00',
                '2018-03-07 00:00:00-05:00', '2018-03-08 00:00:00-05:00'],
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1937,10 +1937,10 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-        DatetimeIndex.tz_localize used for Localize tz-naive DatetimeIndex.
-
-	tz_localize method is use to convert naive DatetimeIndex into any other localize
-	DatetimeIndex 
+	DatetimeIndex.tz_localize used for Localize tz-naive DatetimeIndex.
+	
+	tz_localize method is use to convert naive DatetimeIndex into any other 
+	localize DatetimeIndex 
 
 
         Parameters
@@ -1948,7 +1948,7 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
         tz : string, pytz.timezone, dateutil.tz.tzfile or None
             Time zone for time. Corresponding timestamps would be converted to
             time zone of the TimeSeries.
-            None will remove timezone holding local time
+            None will remove timezone holding local time.
         ambiguous : 'infer', bool-ndarray, 'NaT', default 'raise'
             - 'infer' will attempt to infer fall dst-transition hours based on
               order
@@ -1974,10 +1974,9 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
 
 	Examples
     	--------
-    	In the example below, We put the date range from 01 March 2018 to 08 March 2018
-	& convert this to US/Eastern Time zone
+    	In the example below, We put the date range from 01 March 2018 to 
+	08 March 2018 & convert this to US/Eastern Time zone
  
-    	>>> import pandas as pd
     	>>> df = pd.date_range('2018-03-01', '2018-03-08')
     	>>> df
 	DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
@@ -1993,7 +1992,7 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
         Raises
         ------
         TypeError
-            If the DatetimeIndex is tz-aware and tz is not None
+            If the DatetimeIndex is tz-aware and tz is not None.
         """
         if self.tz is not None:
             if tz is None:
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1937,10 +1937,10 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-	DatetimeIndex.tz_localize used for Localize tz-naive DatetimeIndex.
+	    Convert aware Datetime index to TimeZone aware Datetime index.
 	
-	tz_localize method is use to convert naive DatetimeIndex into any other 
-	localize DatetimeIndex 
+	    Localize timezone-naive DatetimeIndex to given time zone , or
+        remove timezone from timezone aware DatetimeIndex
 
 
         Parameters
@@ -1972,18 +1972,18 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
         -------
         localized : DatetimeIndex
 
-	Examples
+	    Examples
     	--------
     	In the example below, We put the date range from 01 March 2018 to 
-	08 March 2018 & convert this to US/Eastern Time zone
+	    08 March 2018 & convert this to US/Eastern Time zone
  
     	>>> df = pd.date_range('2018-03-01', '2018-03-08')
     	>>> df
-	DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
+	    DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
                '2018-03-05', '2018-03-06', '2018-03-07', '2018-03-08'],
               dtype='datetime64[ns]', freq='D')
     	>>> df.tz_localize(tz='US/Eastern')
-	DatetimeIndex(['2018-03-01 00:00:00-05:00', '2018-03-02 00:00:00-05:00',
+	    DatetimeIndex(['2018-03-01 00:00:00-05:00', '2018-03-02 00:00:00-05:00',
                '2018-03-03 00:00:00-05:00', '2018-03-04 00:00:00-05:00',
                '2018-03-05 00:00:00-05:00', '2018-03-06 00:00:00-05:00',
                '2018-03-07 00:00:00-05:00', '2018-03-08 00:00:00-05:00'],
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1939,7 +1939,7 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
         """
         Convert aware Datetime index to TimeZone naive Datetime index.
 
-        Localize timezone-naive DatetimeIndex to given time zone, 
+        Localize timezone-naive DatetimeIndex to given time zone,
         or remove timezone from timezone aware DatetimeIndex.
 
         Parameters
@@ -1973,18 +1973,18 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
 
         Examples
         --------
-        In the example below, We put the date range from 01 March 2018 to 08 March 2018 & convert this to US/Eastern Time zone
- 
-        >>> df = pd.date_range('2018-03-01', '2018-03-08')
-	>>> df
-	DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
-               '2018-03-05', '2018-03-06', '2018-03-07', '2018-03-08'],
+        In the example below, We put the date range from 01 March 2018
+        to 08 March 2018 & convert this to US/Eastern Time zone
+
+        >>> df = pd.date_range('2018-03-01', '2018-03-05')
+        >>> df
+        DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
+               '2018-03-05'],
               dtype='datetime64[ns]', freq='D')
-	>>> df.tz_localize(tz='US/Eastern')
-	DatetimeIndex(['2018-03-01 00:00:00-05:00', '2018-03-02 00:00:00-05:00',
-               '2018-03-03 00:00:00-05:00', '2018-03-04 00:00:00-05:00',
-               '2018-03-05 00:00:00-05:00', '2018-03-06 00:00:00-05:00',
-               '2018-03-07 00:00:00-05:00', '2018-03-08 00:00:00-05:00'],
+        >>> df.tz_localize(tz='US/Eastern')
+        DatetimeIndex(['2018-03-01 00:00:00-05:00',
+        '2018-03-02 00:00:00-05:00', '2018-03-03 00:00:00-05:00',
+        '2018-03-04 00:00:00-05:00', '2018-03-05 00:00:00-05:00'],
               dtype='datetime64[ns, US/Eastern]', freq='D')
 
         Raises
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1937,9 +1937,9 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-	    Convert aware Datetime index to TimeZone aware Datetime index.
+	Convert aware Datetime index to TimeZone aware Datetime index.
 	
-	    Localize timezone-naive DatetimeIndex to given time zone , or
+	Localize timezone-naive DatetimeIndex to given time zone , or
         remove timezone from timezone aware DatetimeIndex
 
 
@@ -1972,18 +1972,18 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
         -------
         localized : DatetimeIndex
 
-	    Examples
+	Examples
     	--------
     	In the example below, We put the date range from 01 March 2018 to 
 	    08 March 2018 & convert this to US/Eastern Time zone
  
     	>>> df = pd.date_range('2018-03-01', '2018-03-08')
     	>>> df
-	    DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
+	DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
                '2018-03-05', '2018-03-06', '2018-03-07', '2018-03-08'],
               dtype='datetime64[ns]', freq='D')
     	>>> df.tz_localize(tz='US/Eastern')
-	    DatetimeIndex(['2018-03-01 00:00:00-05:00', '2018-03-02 00:00:00-05:00',
+	DatetimeIndex(['2018-03-01 00:00:00-05:00', '2018-03-02 00:00:00-05:00',
                '2018-03-03 00:00:00-05:00', '2018-03-04 00:00:00-05:00',
                '2018-03-05 00:00:00-05:00', '2018-03-06 00:00:00-05:00',
                '2018-03-07 00:00:00-05:00', '2018-03-08 00:00:00-05:00'],
@@ -1992,7 +1992,7 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
         Raises
         ------
         TypeError
-            If the DatetimeIndex is tz-aware and tz is not None.
+        	If the DatetimeIndex is tz-aware and tz is not None.
         """
         if self.tz is not None:
             if tz is None:
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1937,10 +1937,12 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-        Convert aware to naive.
+        Localize tz-naive DatetimeIndex to tz-aware DatetimeIndex.
 
-        Localize timezone-naive DatetimeIndex to given time zone,
-        or remove timezone from timezone aware DatetimeIndex.
+        tz_localize() takes a naive datetime object and interprets it as if it
+        is in that timezone. It does not move the time to another timezone.
+        tz_localize function helps to switch b/w time zone aware and time zone
+        unaware objects.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


pandas/core/indexes/datetimes.py
```diff
@@ -1937,15 +1937,18 @@ def tz_convert(self, tz):
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-        Localize tz-naive DatetimeIndex to given time zone (using
-        pytz/dateutil), or remove timezone from tz-aware DatetimeIndex
+        DatetimeIndex.tz_localize used for Localize tz-naive DatetimeIndex.
+
+	tz_localize method is use to convert naive DatetimeIndex into any other localize
```
jorisvandenbossche: Your indentation is one space off (this line should be indented the same as the first line of the docstring)

pandas/core/indexes/datetimes.py
```diff
@@ -1937,15 +1937,18 @@ def tz_convert(self, tz):
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-        Localize tz-naive DatetimeIndex to given time zone (using
-        pytz/dateutil), or remove timezone from tz-aware DatetimeIndex
+        DatetimeIndex.tz_localize used for Localize tz-naive DatetimeIndex.
```
jorisvandenbossche: We should try to avoid to repeat the name of the method in the beginning of this summary line. What do you think of "Localize a tz-naive DatetimeIndex" ? 
At the same time, we could try to make this even clearer. For example use "timezone" instead of "tz".

pandas/core/indexes/datetimes.py
```diff
@@ -1937,15 +1937,18 @@ def tz_convert(self, tz):
...
+	DatetimeIndex 
+
 
         Parameters
         ----------
         tz : string, pytz.timezone, dateutil.tz.tzfile or None
             Time zone for time. Corresponding timestamps would be converted to
             time zone of the TimeSeries.
-            None will remove timezone holding local time.
+            None will remove timezone holding local time
```
jorisvandenbossche: Please leave this final '.' (the updated validation script will also give a message about this)

pandas/core/indexes/datetimes.py
```diff
@@ -1964,18 +1967,33 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
-        infer_dst : boolean, default False
-            .. deprecated:: 0.15.0
-               Attempt to infer fall dst-transition hours based on order
 
         Returns
         -------
         localized : DatetimeIndex
 
+	Examples
+    	--------
```
jorisvandenbossche: same comment here about the indentation (it seems you are mixing tabs and spaces. It is best to configure your editor to introduce 4 spaces when you do tab)

pandas/core/indexes/datetimes.py
```diff
@@ -1964,18 +1967,33 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+    	In the example below, We put the date range from 01 March 2018 to 08 March 2018
+	& convert this to US/Eastern Time zone
+ 
+    	>>> import pandas as pd
```
jorisvandenbossche: The import of pandas is not needed (check https://python-sprints.github.io/pandas/guide/pandas_docstring.html#conventions-for-the-examples)

typescript
```diff
@@ -0,0 +1,135 @@
+Script started on 2018-03-06 11:15:57+0530
+[1m[7m#[27m[1m[0m                                                                                ]2;root@localhost: /home/ihackpy/Panda/pandas-dir]1;..da/pandas-dir[0m[27m[24m[J[01;32m➜  [36mpandas-dir[00m [01;34mgit:([31mdocstringhead[34m) [33m✗[00m [K[?1h=[?2004hlls[?1l>[?2004l
+]2;ls --color=tty]1;lsappveyor.yml  [0m[01;34mconda.recipe[0m  [01;34mpandas[0m            setup.cfg        [01;32mtest.sh[0m
```
jorisvandenbossche: Can you remove this file again?

pandas/core/indexes/datetimes.py
```diff
@@ -1964,18 +1967,33 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+               '2018-03-03 00:00:00-05:00', '2018-03-04 00:00:00-05:00',
+               '2018-03-05 00:00:00-05:00', '2018-03-06 00:00:00-05:00',
+               '2018-03-07 00:00:00-05:00', '2018-03-08 00:00:00-05:00'],
+              dtype='datetime64[ns, US/Eastern]', freq='D')
+
         Raises
         ------
         TypeError
-            If the DatetimeIndex is tz-aware and tz is not None.
+            If the DatetimeIndex is tz-aware and tz is not None
```
jorisvandenbossche: Same here: please keep the '.'

jorisvandenbossche (commented):
Thanks for the PR! I added a first round of comments

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: okey sure

typescript
HimanshuAwasthi95: Sorry I don't know HOw this file added sorry for the inconvenience

jorisvandenbossche:
Can you do a check of PEP8 with running `git diff upstream/master -u -- "*.py" | flake8 --diff`  (see https://python-sprints.github.io/pandas/guide/pandas_pr.html#push-your-changes-to-pandas for more explanation)

pandas/core/indexes/datetimes.py
```diff
@@ -1937,8 +1937,11 @@ def tz_convert(self, tz):
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-        Localize tz-naive DatetimeIndex to given time zone (using
-        pytz/dateutil), or remove timezone from tz-aware DatetimeIndex
+	DatetimeIndex.tz_localize used for Localize tz-naive DatetimeIndex.
```
jreback: used to localize a tz-aware DatetimeIndex to tz-naive

pandas/core/indexes/datetimes.py
```diff
@@ -1937,8 +1937,11 @@ def tz_convert(self, tz):
...
+	
+	tz_localize method is use to convert naive DatetimeIndex into any other
```
jreback: tz-aware -> tz-naive

jreback (changes requested):

jorisvandenbossche:
@HimanshuAwasthi95 You are still mixing tabs and spaces. Please try to fix that first.

HimanshuAwasthi95:
Locally showing perfect I use vim for editing but don't know when I push code it shows wrong indentation is any tool which help me to remove this issue

jorisvandenbossche:
I am not familiar with vim myself, but you will somehow have to say to vim that tabs should not be introduces as tabs but as 4 spaces (it may 'look' good, but that doesn't mean it is not mixing tabs and spaces). See eg https://stackoverflow.com/questions/1878974/redefine-tab-as-4-spaces

jorisvandenbossche:
Or also see the second answer in https://stackoverflow.com/questions/426963/replace-tab-with-spaces-in-vim

jorisvandenbossche:
I am afraid it is not yet solved. 
You can check the output of `git diff upstream/master -u -- "*.py" | flake8 --diff `, it will still contain errors related to tabs/spaces. See also https://github.com/pandas-dev/pandas/pull/20050#issuecomment-371473606

jorisvandenbossche:
Tabs/spaces issues seems to be solved!

pandas/core/indexes/datetimes.py
```diff
@@ -1937,8 +1937,10 @@ def tz_convert(self, tz):
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-        Localize tz-naive DatetimeIndex to given time zone (using
-        pytz/dateutil), or remove timezone from tz-aware DatetimeIndex
+        Convert aware Datetime index to TimeZone naive Datetime index.
```
jorisvandenbossche: "Datetime index" -> DatetimeIndex 
 "TimeZone" -> time zone

And it is the other way around: you now say "convert aware to naive", but it is "localize naive to aware"

pandas/core/indexes/datetimes.py
```diff
@@ -1964,18 +1966,32 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
 
             .. versionadded:: 0.19.0
 
-        infer_dst : boolean, default False
-            .. deprecated:: 0.15.0
-               Attempt to infer fall dst-transition hours based on order
```
jorisvandenbossche: You can leave this one (although it will give an error in the validation script, but you can ignore that)

pandas/core/indexes/datetimes.py
```diff
@@ -1964,18 +1966,32 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+        Examples
+        --------
+        In the example below, We put the date range from 01 March 2018
+        to 08 March 2018 & convert this to US/Eastern Time zone
+
+        >>> df = pd.date_range('2018-03-01', '2018-03-05')
+        >>> df
+        DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
+               '2018-03-05'],
+              dtype='datetime64[ns]', freq='D')
```
jorisvandenbossche: Can you clean up the alignment a bit here? If I run this in a console, I get this:

```
>>> df
DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
               '2018-03-05'],
              dtype='datetime64[ns]', freq='D')
```

Ideally it should look the same (notice the different vertical alignment with your code)

jorisvandenbossche (commented):
Can you add an example for the `None` behaviour as well? (removing the timezone)

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: okey sure

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: so will I add this again ?

jreback: @jorisvandenbossche why adding this back? looks leftover

HimanshuAwasthi95: @jreback @jorisvandenbossche I add this again or not ???????

jorisvandenbossche: @jreback no, the deprecation is still there (the fact that the deprecation is still there is certainly a left-over, but let's remove that (both the actual deprecation and the docs) in a separate PR)

@HimanshuAwasthi95 Yes, please add back

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: @jorisvandenbossche can you illustrate more "Can you add an example for the None behaviour as well? (removing the timezone)"

pandas/core/indexes/datetimes.py
```diff
@@ -1937,8 +1937,10 @@ def tz_convert(self, tz):
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-        Localize tz-naive DatetimeIndex to given time zone (using
-        pytz/dateutil), or remove timezone from tz-aware DatetimeIndex
+        Convert aware to naive.
```
jreback: Localize a tz-naive DatetimeIndex to tz-aware DatetimeIndex

pandas/core/indexes/datetimes.py
```diff
@@ -1937,8 +1937,10 @@ def tz_convert(self, tz):
...
+
+        Localize timezone-naive DatetimeIndex to given time zone,
+        or remove timezone from timezone aware DatetimeIndex.
```
jreback: this is almost a duplicate of the top

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1970,32 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
 
+
         Returns
         -------
         localized : DatetimeIndex
 
+        Examples
+        --------
+        In the example below, We put the date range from 01 March 2018
+        to 08 March 2018 & convert this to US/Eastern Time zone
```
jreback: don't say convert. this is localizing, almost like 'setting' a time-zone.

jreback (changes requested):

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1970,32 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+
+        >>> df = pd.date_range('2018-03-01', '2018-03-05')
+        >>> df
+        DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03',
+        '2018-03-04', '2018-03-05'],
+        dtype='datetime64[ns]', freq='D')
+        >>> df.tz_localize(tz='US/Eastern')
+        DatetimeIndex(['2018-03-01 00:00:00-05:00',
+        '2018-03-02 00:00:00-05:00', '2018-03-03 00:00:00-05:00',
+        '2018-03-04 00:00:00-05:00', '2018-03-05 00:00:00-05:00'],
```
jreback: showing the reverse example is also useful, IOW ``.tz_localize(None)``, removes the timezone and makes it tz-naive.

jreback (changes requested):

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: @jreback 
can I add this line in extended summary :
tz_localize() takes a naive datetime object and interprets it as if it is in that timezone. It does not move the time to another timezone. tz_localize function helps to switch b/w time zone aware and time zone unaware objects.

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: @jreback when I add another example it shows error how I add that example in the docstring

pandas/core/indexes/datetimes.py
jreback: interprets -> sets it in that timezone with the current timestamps.

> tz_localize function helps to switch b/w time zone aware and time zone unaware objects.

not sure this last line is necessary

pandas/core/indexes/datetimes.py
jreback:
```
n [11]: pd.date_range('2018-03-01', '2018-03-05').tz_localize('US/Eastern').tz_localize(None)
Out[11]: 
DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03', '2018-03-04',
               '2018-03-05'],
              dtype='datetime64[ns]', freq='D')
```

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1970,32 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
         Returns
         -------
         localized : DatetimeIndex
 
+        Examples
+        --------
+        In the example below, We put the date range from 01 March 2018
+        to 08 March 2018 & convert this to US/Eastern Time zone
+
+        >>> df = pd.date_range('2018-03-01', '2018-03-05')
```
jreback: don't use df, dti is better

jreback (changes requested):

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: @jreback @jorisvandenbossche I made required change.

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: any thing else in summary section ???

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: @jreback If it is give more clearity then It is okey other wise I can remove ??????

pandas/core/indexes/datetimes.py
```diff
@@ -1937,8 +1937,12 @@ def tz_convert(self, tz):
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-        Localize tz-naive DatetimeIndex to given time zone (using
-        pytz/dateutil), or remove timezone from tz-aware DatetimeIndex
+        Localize tz-naive DatetimeIndex to tz-aware DatetimeIndex.
+
+        tz_localize() takes a naive datetime object and interprets it as if it
```
jreback: takes a naive DatetimeIndex

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1972,36 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+        --------
+        In the example below, We put the date range from 01 March 2018
+        to 08 March 2018 & localize this to US/Eastern Time zone & again
+        we perform reverse operation where we remove tize zone & make it
+        tz-naive
+
+        >>> dti = pd.date_range('2018-03-01', '2018-03-05')
+        >>> dti
+        DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03',
+        '2018-03-04', '2018-03-05'], dtype='datetime64[ns]', freq='D')
```
jreback: can you add a blank line between cases

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1972,36 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+        >>> dti.tz_localize(tz='US/Eastern')
+        DatetimeIndex(['2018-03-01 00:00:00-05:00',
+        '2018-03-02 00:00:00-05:00', '2018-03-03 00:00:00-05:00',
+        '2018-03-04 00:00:00-05:00', '2018-03-05 00:00:00-05:00'],
+        dtype='datetime64[ns, US/Eastern]', freq='D')
+        >>> dti.tz_localize(None)
```
jreback: this last won't work, you need to do
``dti.tz_localize('US/Eastern').tz_localize(None)``, which gets you back the original

jreback (changes requested):

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: screenshot from 2018-03-10 21-00-32
it works well

jreback: it works but its not very useful. The point is to take a tz-aware, IOW some you previously used ``tz_localize`` on, or just constructed directly and the remove the tz. 

How about you construct a new dti with a tz (IOW pass directly to ``date_range``), then use ``.tz_localize(None)`` on it.

HimanshuAwasthi95: @jreback both gives same result

HimanshuAwasthi95: oh K Will do that

HimanshuAwasthi95:
@jreback i changes example as required

HimanshuAwasthi95:
@datapythonista @jorisvandenbossche @jreback Review

pandas/core/indexes/datetimes.py
```diff
@@ -1985,23 +1985,20 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         we perform reverse operation where we remove tize zone & make it
```
datapythonista: Can you just leave the type in `Returns` (so, get rid of the "localized" name). Also, a short explanation on what is returned would be useful.

pandas/core/indexes/datetimes.py
```diff
@@ -1985,23 +1985,20 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
         '2018-03-02 00:00:00-05:00', '2018-03-03 00:00:00-05:00',
         '2018-03-04 00:00:00-05:00', '2018-03-05 00:00:00-05:00'],
         dtype='datetime64[ns, US/Eastern]', freq='D')
 
-        In [5]: dti.tz_localize('US/Eastern').tz_localize(None)
-
-        Out[5]: DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03',
+        >>> dti.tz_localize('US/Eastern').tz_localize(None)
+        DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03',
         '2018-03-04', '2018-03-05'], dtype='datetime64[ns]', freq='D')
```
datapythonista: More a personal opinion, but couple of things that I think would make the examples more clear:
- Use 3 dates instead of 5
- Add short sentences between commands on what's going on (e.g. "The values for the times are the same, but the index contains the timezone")
- Save the output of `dti.tz_localize(tz='US/Eastern')` so it can be reused in the last command, instead of doing to two localizations in the same line. May be naming `dti` `tz_naive` first, and `tz_aware` later would make them more clear too.

datapythonista (commented):

Also, in the extended summary, I'd avoid repeating the name of the method at the beginning of each parameter.

In the ambigous parameter, the accepted values should be in curly brackets (except bool-ndarray), and include 'raise'.

Btw, no need to add comments to ask for review when you push new changes. github notifies reviewers when you push.

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: @datapythonista 
I'm understand that I have to remove localized from return section & write like this :
Returns
-------
DatetimeIndex

or I should write some explanation on returns section like this:
Returns
 -------
Some Explanation 
I'm confused about it ?

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: okey sure

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: and parameter ambiguous : {'infer', 'NaT', default 'raise'}, bool-ndarray . write like this ???

datapythonista: For return, first line as you say, then another line with explanation. You can check the first example here:
https://python-sprints.github.io/pandas/guide/pandas_docstring.html#section-4-returns-or-yields.

For parameter, I didn't see a case exactly like this before, but for the convention I'd say something like:

`{'raise', 'infer', 'NaT'} or bool array-like, default 'raise'`

HimanshuAwasthi95: is there any change with summary & with example section

pandas/core/indexes/datetimes.py
```diff
@@ -1937,16 +1937,21 @@ def tz_convert(self, tz):
                      mapping={True: 'infer', False: 'raise'})
     def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         """
-        Localize tz-naive DatetimeIndex to given time zone (using
-        pytz/dateutil), or remove timezone from tz-aware DatetimeIndex
+        Localize tz-naive DatetimeIndex to tz-aware DatetimeIndex.
+
+        This method takes a naive DatetimeIndex object and interprets it as
+        if it is in that timezone. It does not move the time to another
+        timezone.
```
datapythonista: I found the interprets word a bit unclear. I think something like "Make a naive DatetimeIndex time zone aware, by attaching a time zone. It does not change the values."

Surely you can write it in a better way, but hopefully you see what I mean.

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1973,45 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+
+        #localize DatetimeIndex in US?Eastern time zone.
+        >>> tz_aware = dti.tz_localize(tz='US/Eastern')
+
+        >>> tz_aware
+        DatetimeIndex(['2018-03-01 00:00:00-05:00',
+        '2018-03-02 00:00:00-05:00', '2018-03-03 00:00:00-05:00'],
+        dtype='datetime64[ns, US/Eastern]', freq='D')
+
+        #localize aware time zone into naive time zone.
```
datapythonista: Sorry, that was my fault, but when I said to add comments to the examples, I didn't mean Python comments, but just the sentence. So, without the `#`, in capital letters, and with a blank line after it (also the one before you already have).

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1973,45 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+        we perform reverse operation where we remove tize zone & make it
+        tz-naive
+
+        >>> dti = pd.date_range('2018-03-01', '2018-03-03')
+
+        >>> dti
+        DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03'],
+        dtype='datetime64[ns]', freq='D')
+
+        #localize DatetimeIndex in US?Eastern time zone.
```
datapythonista: Typo, ? instead of / in the time zone.

datapythonista (commented):
Looking great, added couple of last comments.

pandas/core/indexes/datetimes.py
jorisvandenbossche: Can you also remove the `#` at the beginning of the line?

pandas/core/indexes/datetimes.py
```diff
@@ -1937,16 +1937,21 @@ def tz_convert(self, tz):
...
+        zone unaware objects.
 
         Parameters
         ----------
         tz : string, pytz.timezone, dateutil.tz.tzfile or None
             Time zone for time. Corresponding timestamps would be converted to
             time zone of the TimeSeries.
             None will remove timezone holding local time.
-        ambiguous : 'infer', bool-ndarray, 'NaT', default 'raise'
+        ambiguous : {'infer', 'NaT', default 'raise'}, bool-ndarray
```
jorisvandenbossche: I would make this `str {'infer', 'NaT', 'raise'} or boolean array, default 'raise'`

But that is probably too long, so you can do:

```
        ambiguous : str {'infer', 'NaT', 'raise'} or bool array, \
default 'raise'
            - 'infer': ...
```

It is a bit ugly in the code, but will render OK in plain text docstring and in html docstring.

jorisvandenbossche: A little bit below for the "errors" keyword, can you make the type description there: `errors : {'raise', 'coerce'}, default 'raise'` (so with curly braces)

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: oops

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: sure

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: errors : {'raise', 'coerce'}, default 'raise' 
we have to define str{'raise', 'coerce'} I thought its already understood

jorisvandenbossche: Yes, in this case `errors : {'raise', 'coerce'}, default 'raise'` is good (without the 'str')

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: @datapythonista hoes this otherwise I'm going to co[py paste same you suggest above :) 

"This method takes a naive DatetimeIndex object and make this time zone aware. It does not move the time to another timezone.
 Time zone localization helps to switch b/w time zone aware and time zone unaware objects."

pandas/core/indexes/datetimes.py
```diff
@@ -1991,15 +1992,17 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
         DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03'],
```
datapythonista: `Returns` should contain a description. See the example in: https://python-sprints.github.io/pandas/guide/pandas_docstring.html#section-4-returns-or-yields

pandas/core/indexes/datetimes.py
```diff
@@ -1991,15 +1992,17 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
         dtype='datetime64[ns]', freq='D')
 
-        #localize DatetimeIndex in US?Eastern time zone.
+        localize DatetimeIndex in US/Eastern time zone.
```
datapythonista: The explanations of the examples need improvements.  The first paragraph is outdated (end date is not the 8th). And I don't think you need to say that, it's clear from the code. So I would just get rid of it.

In this case, it should start with a capital `L`. May be you can add here what you were explaining before, that `dti` is naive (and also what you're already saying, that you're localizing to US/Eastern). Personally I'd rename `dti` to `tz_aware`, so it's more obvious.

pandas/core/indexes/datetimes.py
```diff
@@ -1991,15 +1992,17 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
         DatetimeIndex(['2018-03-01 00:00:00-05:00',
         '2018-03-02 00:00:00-05:00', '2018-03-03 00:00:00-05:00'],
         dtype='datetime64[ns, US/Eastern]', freq='D')
 
-        #localize aware time zone into naive time zone.
+        localize aware time zone into naive time zone.
+
         >>> tz_naive = tz_aware.tz_localize(None)
 
         >>> tz_naive
```
datapythonista: As you're not doing anything with `tz_naive` more than displaying it, I wouldn't save it to a variable first, I'd show the output direclty.

pandas/core/indexes/datetimes.py
```diff
@@ -1991,15 +1992,17 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+
         >>> tz_aware = dti.tz_localize(tz='US/Eastern')
 
         >>> tz_aware
         DatetimeIndex(['2018-03-01 00:00:00-05:00',
         '2018-03-02 00:00:00-05:00', '2018-03-03 00:00:00-05:00'],
         dtype='datetime64[ns, US/Eastern]', freq='D')
 
-        #localize aware time zone into naive time zone.
+        localize aware time zone into naive time zone.
```
datapythonista: Same as before.

datapythonista (commented):
Almost there, looks good, just couple of comments for the Return and Examples sections.

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: You are saying this : 

In [2]: tz_aware = pd.date_range('2018-03-01', '2018-03-03')

In [3]: tz_aware
Out[3]: DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03'], dtype='datetime64[ns]', freq='D')

In [4]: tz_aware.tz_localize(tz='US/Eastern')
Out[4]: 
DatetimeIndex(['2018-03-01 00:00:00-05:00', '2018-03-02 00:00:00-05:00',
               '2018-03-03 00:00:00-05:00'],
              dtype='datetime64[ns, US/Eastern]', freq='D')

In [5]: tz_aware.tz_localize(None)
Out[5]: DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03'], dtype='datetime64[ns]', freq='D')

datapythonista: Almost. But in `In [2]` that's not a `tz_aware` index, it's a naive one. And in `In [4]` you do want to save the output, as that's the aware one that you'll one to convert to naive in `In [5]`.

Make sure that the example makes sense for you too. Reviewers can also be wrong.

HimanshuAwasthi95: but thats why we added comment before where we analyse aware and naive

datapythonista: I agree, but I think the variable names can still help by being descriptive, even if you explain what you're doing. Feel free to keep `dti` if you prefer it over `tz_naive`. But for me it's more clear to see what the example does, if the variables are more descriptive. All objects in the example could be `dti`, and it doesn't help much IMO, but naming one `tz_naive` and the other `tz_aware`, makes it clear what every object used has.

Sorry if I'm not being able to explain what I mean clearer enough.

HimanshuAwasthi95: I hope this commit define exactly what you want in example section

HimanshuAwasthi95: sorry for disappointing you :(

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1974,41 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+        Localize DatetimeIndex in US/Eastern time zone.
+
+        >>> tz_aware = tz_naive.tz_localize(tz='US/Eastern')
+
+        >>> tz_aware
+        DatetimeIndex(['2018-03-01 00:00:00-05:00',
+        '2018-03-02 00:00:00-05:00', '2018-03-03 00:00:00-05:00'],
+        dtype='datetime64[ns, US/Eastern]', freq='D')
+
+        Localize aware time zone into naive time zone.
```
datapythonista: "time zone aware index" instead of "aware time zone"?

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1974,41 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
         -------
-        localized : DatetimeIndex
+        DatetimeIndex
+            It returns naive time or aware time
+
+        Examples
+        --------
+        >>> tz_naive = pd.date_range('2018-03-01', '2018-03-03')
+
+        >>> tz_naive
```
datapythonista: The blank line between them seems unnecessary. (also later in tz_aware)

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1974,41 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
...
+        >>> tz_aware.tz_localize(None)
+        DatetimeIndex(['2018-03-01', '2018-03-02', '2018-03-03'],
+        dtype='datetime64[ns]', freq='D')
+
 
         Raises
         ------
         TypeError
             If the DatetimeIndex is tz-aware and tz is not None.
+
```
datapythonista: I think we don't leave this line here, the validation script probably complains.

pandas/core/indexes/datetimes.py
```diff
@@ -1968,14 +1974,41 @@ def tz_localize(self, tz, ambiguous='raise', errors='raise'):
             .. deprecated:: 0.15.0
                Attempt to infer fall dst-transition hours based on order
 
+
         Returns
         -------
-        localized : DatetimeIndex
+        DatetimeIndex
+            It returns naive time or aware time
```
datapythonista: I'd describe it more in terms of the functionality than the types. Something like "Index converted to the specified time zone." (The "It returns" seems redundant, and I think the spec requires a period at the end)

datapythonista:
The example looks much better now. Just added couple of minor comments.

Sorry if I made you feel you were disappointing anyone. You're doing a great job, discussions in reviews are a normal and very beneficial part of open source. As you get more experience probably the discussions will be a bit shorter. But we've all been there (and we often are), in these PRs that require several iterations to look great. :)

pandas/core/indexes/datetimes.py
HimanshuAwasthi95: I copied your line "index converted to the specified time zone". because It gives perfect means

datapythonista (approved):

datapythonista:
lgtm

HimanshuAwasthi95:
Thats why I love community thanks for supporting and guiding

HimanshuAwasthi95:
Thanks alot :)

jorisvandenbossche (approved):
Did some final edits, looks good!

jorisvandenbossche:
@HimanshuAwasthi95 Thanks for the PR!

HimanshuAwasthi95:
thank @jorisvandenbossche @datapythonista @jreback

PR #20055, opened 2018-03-08
Temporary github PR template for sprint

jorisvandenbossche:
@datapythonista do you think this would be useful to repeat it once more here?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,17 +7,13 @@ an unrelated PR):
 - [ ] The html version looks good: `python doc/make.py --single <your-function-or-method>`
 - [ ] It has been proofread on language by another sprint participant
 
-Please include the output of the validation script under the details tag below:
-
-<details>
+Please include the output of the validation script below between the "```" ticks:
 
 ```
 # paste output of "scripts/validate_docstrings.py <your-function-or-method>" here
 # between the "```" (remove this comment, but keep the "```")
 
 ```
-
-</details>
 
 If the validation script still gives errors, but you think there is a good reason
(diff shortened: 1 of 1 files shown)


datapythonista:
Absolutely! I thought on that actually, I think it's great.

I'm not 100% sure, but I'd ask people to copy the output of `validate_docstrings.py` in the description. First that would avoid the temptation of some people of marking the check without running it. Then for shared docstrings it'd show the rendered one for that function. And finally, in cases when the validation shows errors that make sense ignoring, it'll be quite easy to find which ones they are.

jorisvandenbossche:
I think that's a good idea. Of course, in the back-and-forth review process, this will change, but we can always ask them to update it.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,30 @@
+Checklist for the pandas documentation sprint (ignore this if you are doing
+an unrelated PR):
+
+- [ ] PR title is "DOC: update the <your-function-or-method> docstring"
+- [ ] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
+- [ ] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
+- [ ] The html version looks good: `python doc/make.py --single <your-function-or-method>`
+- [ ] It has been proofread on language by another sprint participant
+
+Please include below the output of the validation script in the details tag:
```
gfyoung: Probably say "under" instead of "in"

jreback:
lgtm. pls make a revert issue for this as well (though maybe steel some ideas for the regular PR guide)

jorisvandenbossche:
Removed the details tag. I personally like it, but people often don't use it properly, let's keep it simple for this sprint

PR #20057, opened 2018-03-08
DOC: Updating pandas.Interval docstring

akielbowicz:
Docstring assigned to the Buenos Aires chapter for the sprint.

Hi, I have a few doubts about the changes.

First about the type of the parameters *left* and *right*, it's ok to refer to them as *orderable*?
In the implementation, the constructor takes any pair of objects that satisfy `left <= right` and the comparison returns a boolean value.
 I tried to pass two numpy arrays and it breaks because of ambiguity, but still it's so general that you can even make Intervals of Intervals, sets or booleans. And I believe this is too permissive and prone to error. 

So what it's better, put a warning about this or add an example that is a little more complex and uses this capability (I tried to think one, but nothing came up to my mind).

And last,  validate_docstrings.py returned this error

`2341, in _signature_from_callable 
   'no signature found for builtin type {!r}'.format(obj)) 
ValueError: no signature found for builtin type <class 'pandas._libs.interval.Interval'>
`

Thanks, happy to collaborate.

PS: Given that this is a Cython file I needed to recompile the project to see the changes in the documentation. I mention this because is not in the sprint guide and I don't know if I will able to add it before the event.

### DIFF
--- a/pandas/_libs/interval.pyx
+++ b/pandas/_libs/interval.pyx
@@ -81,43 +81,68 @@ cdef class Interval(IntervalMixin):
 
     Parameters
     ----------
-    left : orderable object
+    left : orderable scalar
         Left bound for the interval.
-    right : orderable object
+    right : orderable scalar
         Right bound for the interval.
     closed : {'left', 'right', 'both', 'neither'}, default 'right'
         Whether the interval is closed on the left-side, right-side, both or
         neither.
-    
+
     Notes
     -----
     You must be able to compare the parameters `left` and `right`,
-    and they must satisfy ``left <= right``. 
+    and they must satisfy ``left <= right``.
 
     Examples
     --------
-    The examples below show how you can build Intervals of various types
-    of elements. In particular a numeric interval, a time interval and
-    finally a string interval. Given an Interval you can check whether 
-    an element belongs to it.
+    It is possible to build Intervals of different types, like numeric ones
 
     >>> iv = pd.Interval(left=0, right=5)
     >>> iv
     Interval(0, 5, closed='right')
+
+    In which you can check if an element belongs to it
+
     >>> 2.5 in iv
     True
+
+    You can test the bounds
+
+    >>> 0 in iv
+    False
+    >>> 5 in iv
+    True
+
+    Calculate its length
+
+    >>> iv.length
+    5
+
+    You can operate with `+` and `*` over an Interval and the operation
+    is applied to each of its bounds, so the result depends on the type
+    of the bound elements
+
     >>> shifted_iv = iv + 3
-    >>> shifted_iv 
+    >>> shifted_iv
     Interval(3, 8, closed='right')
-    
-    >>> year_2017 = pd.Interval(pd.Timestamp('2017-01-01'),
-    ...                         pd.Timestamp('2017-12-31'), closed='both')
+    >>> extended_iv = iv * 10.0
+    >>> extended_iv
+    Interval(0.0, 50.0, closed='right')
+
+    To create a time interval you can use Timestamps as the bounds
+
+    >>> year_2017 = pd.Interval(pd.Timestamp('2017-01-01 00:00:00'),
+    ...                         pd.Timestamp('2018-01-01 00:00:00'),
+    ...                         closed='left')
     >>> pd.Timestamp('2017-01-01 00:00') in year_2017
     True
     >>> year_2017.length
-    Timedelta('364 days 00:00:00')
+    Timedelta('365 days 00:00:00')
 
-    >>> volume_1 = pd.Interval('Ant','Dog',closed='both')
+    And also you can create string intervals
+
+    >>> volume_1 = pd.Interval('Ant', 'Dog', closed='both')
     >>> 'Bee' in volume_1
     True
 
--- a/pandas/_libs/interval.pyx
+++ b/pandas/_libs/interval.pyx
@@ -81,25 +81,44 @@ cdef class Interval(IntervalMixin):
 
     Parameters
     ----------
-    left : value
-        Left bound for the interval
-    right : value
-        Right bound for the interval
+    left : orderable object
+        Left bound for the interval.
+    right : orderable object
+        Right bound for the interval.
     closed : {'left', 'right', 'both', 'neither'}, default 'right'
         Whether the interval is closed on the left-side, right-side, both or
-        neither
+        neither.
+    
+    Notes
+    -----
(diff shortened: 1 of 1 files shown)


pandas/_libs/interval.pyx
```diff
@@ -81,26 +81,45 @@ cdef class Interval(IntervalMixin):
 
     Parameters
     ----------
-    left : value
-        Left bound for the interval
-    right : value
-        Right bound for the interval
+    left : orderable object
```
jorisvandenbossche: maybe 'scalar' instead of object? To make clear an array-like is not accepted?

pandas/_libs/interval.pyx
```diff
@@ -81,26 +81,45 @@ cdef class Interval(IntervalMixin):
...
+    right : orderable object
+        Right bound for the interval.
     closed : {'left', 'right', 'both', 'neither'}, default 'right'
         Whether the interval is closed on the left-side, right-side, both or
-        neither
+        neither.
+    
+    Notes
+    -----
+    You must be able to compare the parameters **left** and **right**,
```
jorisvandenbossche: Can you use single backticks to refer to the keywords? (like `` `left` ``), and double backticks for the code snippet on the line below ?

pandas/_libs/interval.pyx
```diff
@@ -81,26 +81,45 @@ cdef class Interval(IntervalMixin):
...
+    and they must satisfy **left <= right**. 
 
     Examples
     --------
+    The examples below show how you can build Intervals of various types
+    of elements. In particular a numeric interval, a time interval and
+    finally a string interval. Given an Interval you can check whether 
+    an element belongs to it.
```
jorisvandenbossche: In general, I think it can be easier to follow to put some of those explanations between the examples, instead of first all explanation and then all examples.

jorisvandenbossche (commented):

jorisvandenbossche:
> And last, validate_docstrings.py returned this error
> 
> 2341, in _signature_from_callable 'no signature found for builtin type {!r}'.format(obj)) ValueError: no signature found for builtin type <class 'pandas._libs.interval.Interval'>

Can you try the latest version? I think this should be fixed now. 
You can update this branch by doing:

```
git fetch upstream
git merge upstream/master
```

and then try the script again.

> Given that this is a Cython file I needed to recompile the project to see the changes in the documentation. I mention this because is not in the sprint guide and I don't know if I will able to add it before the event.

Ah, indeed, that's a good catch. That's something we need to add to the guide.

jorisvandenbossche:
> So what it's better, put a warning about this or add an example that is a little more complex and uses this capability (I tried to think one, but nothing came up to my mind).

I would personally give 'good examples', as you did now. And the note you added about `left <= right` needing to work seems sufficient to me.

pandas/_libs/interval.pyx
```diff
@@ -81,26 +81,45 @@ cdef class Interval(IntervalMixin):
...
+    >>> shifted_iv = iv + 3
+    >>> shifted_iv 
+    Interval(3, 8, closed='right')
+    
     >>> year_2017 = pd.Interval(pd.Timestamp('2017-01-01'),
     ...                         pd.Timestamp('2017-12-31'), closed='both')
     >>> pd.Timestamp('2017-01-01 00:00') in year_2017
     True
+    >>> year_2017.length
+    Timedelta('364 days 00:00:00')
```
jschendel: Might be nice to change `year_2017` to 
```python
year_2017 = pd.Interval(pd.Timestamp('2017-01-01'),
                        pd.Timestamp('2018-01-01'), closed='left')
```
That seems like a more accurate representation of a year, since the current version doesn't include anything beyond `2017-12-31 00:00:00`, e.g. `2017-12-31 12:00:00` isn't included in `year_2017` as currently written.  This change would also result in `year_2017.length` being 365 days, which should be more intuitive for users.

akielbowicz:
Thanks for the comments, I tried to improve it following your guidelines.

jorisvandenbossche:
@jschendel could you give this another look?

pandas/_libs/interval.pyx
```diff
@@ -81,26 +81,70 @@ cdef class Interval(IntervalMixin):
...
+        neither.
+
+    Notes
+    -----
+    You must be able to compare the parameters `left` and `right`,
+    and they must satisfy ``left <= right``.
 
     Examples
     --------
+    It is possible to build Intervals of different types, like numeric ones
```
jreback: can you add that left & right must be of the *same* type. (add to Notes)

pandas/_libs/interval.pyx
```diff
@@ -81,26 +81,70 @@ cdef class Interval(IntervalMixin):
...
+    >>> year_2017.length
+    Timedelta('365 days 00:00:00')
+
+    And also you can create string intervals
+
+    >>> volume_1 = pd.Interval('Ant', 'Dog', closed='both')
+    >>> 'Bee' in volume_1
+    True
 
     See Also
```
jreback: can you add Period here as well (its pretty much a time based interval)

jreback (changes requested):

pandas/_libs/interval.pyx
```diff
@@ -81,26 +81,70 @@ cdef class Interval(IntervalMixin):
...
+    is applied to each of its bounds, so the result depends on the type
+    of the bound elements
+
+    >>> shifted_iv = iv + 3
+    >>> shifted_iv
+    Interval(3, 8, closed='right')
+    >>> extended_iv = iv * 10.0
+    >>> extended_iv
+    Interval(0.0, 50.0, closed='right')
+
```
jschendel: Optional: Could maybe add that an interval does not necessarily need to be bounded, e.g. `pd.Interval(-np.inf, 5)`.  Not sure if there's a ton of utility in adding this though.

jschendel (commented):
Thanks! LGTM pending @jreback's comments.  Added one comment regarding an optional potential addition that could be made.

pandas/_libs/interval.pyx
jschendel: To add a couple of points here: `Interval` is basically a more generic `Period`, in that `Period` requires a frequency and `Interval` does not.  The trade-off being that `Period` has additional frequency specific functionality that `Interval` does not.

jorisvandenbossche (approved):

jorisvandenbossche:
@akielbowicz Thanks a lot!

PR #20058, opened 2018-03-08
Small typo in apply.py

mroeschke:
`np. asarray(res)` --> `np.asarray(res)`

### DIFF
--- a/pandas/core/apply.py
+++ b/pandas/core/apply.py
@@ -191,7 +191,7 @@ class FrameApply(object):
 
         for i, col in enumerate(target.columns):
             res = self.f(target[col])
-            ares = np. asarray(res).ndim
+            ares = np.asarray(res).ndim
 
             # must be a scalar or 1d
             if ares > 1:


pandas/core/apply.py
```diff
@@ -191,7 +191,7 @@ def apply_broadcast(self, target):
 
         for i, col in enumerate(target.columns):
             res = self.f(target[col])
-            ares = np. asarray(res).ndim
```
gfyoung: Are we even hitting this block of code if we managed to get away with that!?

mroeschke: I haven't looked at the apply tests closely, but I was somewhat surprised this is line works with the whitespace...

```
In [1]: import numpy as np

In [2]: np. asarray([1, 2, 3]).ndim
Out[2]: 1

In [3]: np. asarray([1, 2, 3])
Out[3]: array([1, 2, 3])

In [4]: np. asarray([1, 2, 3]).ndim == np.asarray([1, 2, 3]).ndim
Out[4]: True
```

gfyoung: That's really odd!  I didn't know this could work.  And I thought Python was really picky about whitespace...

BTW, the easiest way to check is to deliberately raise an exception in that logic block and run tests with that bugged install.

jschendel: Looks like Python actually allows for you to put spaces after periods and will just evaluate as if the spaces weren't there, so this may or may not actually be getting hit:

```python
In [7]: np.   array([0, 2, 1]).   max()
Out[7]: 2
```

mroeschke: Looks like the following tests in `pandas/tests/frame/test_apply.py` hit this path:

```
TestDataFrameApply.test_apply_broadcast_deprecated
TestDataFrameApply.test_apply_broadcast
TestDataFrameApply.test_apply_broadcast_error
TestInferOutputShape.test_result_type

```

jreback:
thanks....guess python allows whitespace after member access, prob to allow multi-line splitting

PR #20059, opened 2018-03-08
TST: series/indexing tests parameterization + moving test methods

### LINKED ISSUE #20014: CLN: followons to refactor of series/test_indexing.py
xref #18164 

- [ ] remove the classes, IOW make all of the tests at function level. This will require removing the ``setup_method`` and making fixtures for this.
- [ ] some tests need to be moved around a bit, IOW we want to move tests out of test_indexing.py (the new one), so appropriate places (e.g. .drop should be in alter_axes). 
- [ ] want to parameterize things (e.g. anything with loops)

any/all of these can be separate PRs.

almaleksia:
Parts 2 and 3 of #20014. 
- tests parameterization
- moving some tests from test_indexing.py to other files 
- test methods reordering in test_indexing.py

- [x] closes #20014
- [x] tests added / passed

### DIFF
--- a/pandas/tests/series/indexing/test_boolean.py
+++ b/pandas/tests/series/indexing/test_boolean.py
@@ -283,34 +283,30 @@ def test_where_error():
                   [])
 
 
-def test_where_array_like():
+@pytest.mark.parametrize('klass', [list, tuple, np.array, Series])
+def test_where_array_like(klass):
     # see gh-15414
     s = Series([1, 2, 3])
     cond = [False, True, True]
     expected = Series([np.nan, 2, 3])
-    klasses = [list, tuple, np.array, Series]
 
-    for klass in klasses:
-        result = s.where(klass(cond))
-        assert_series_equal(result, expected)
+    result = s.where(klass(cond))
+    assert_series_equal(result, expected)
 
 
-def test_where_invalid_input():
+@pytest.mark.parametrize('cond', [
+    [1, 0, 1],
+    Series([2, 5, 7]),
+    ["True", "False", "True"],
+    [Timestamp("2017-01-01"), pd.NaT, Timestamp("2017-01-02")]
+])
+def test_where_invalid_input(cond):
     # see gh-15414: only boolean arrays accepted
     s = Series([1, 2, 3])
     msg = "Boolean array expected for the condition"
 
-    conds = [
-        [1, 0, 1],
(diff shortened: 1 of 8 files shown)


pandas/tests/series/indexing/test_boolean.py
```diff
@@ -403,37 +399,32 @@ def f():
...
+    [False]
+])
+@pytest.mark.parametrize('item', [
+    2.0, np.nan, np.finfo(np.float).max, np.finfo(np.float).min
+])
+def test_where_broadcast(size, mask, item):
+    selection = np.resize(mask, size)
+    # Test numpy arrays, lists and tuples as the input to be
+    # broadcast
+    for arr in [np.array([item]), [item], (item,)]:
```
jreback: could add this to the parameterization, we typically would call this ``box: [lambda x: np.array([x]), lambda x: [x], lambda x: (x,)]``

pandas/tests/series/indexing/test_boolean.py
```diff
@@ -587,27 +578,27 @@ def test_mask():
...
+    [True, False],
+    [False]
+])
+@pytest.mark.parametrize('item', [
+    2.0, np.nan, np.finfo(np.float).max, np.finfo(np.float).min
+])
+def test_mask_broadcast(size, mask, item):
+        selection = np.resize(mask, size)
+
+        for arr in [np.array([item]), [item], (item,)]:
```
jreback: same as above, though would not be averse to testing both mask and where in the same function to make this simpler

jreback (changes requested):
very nice! just a couple of minor comments. ping on green.

jreback (approved):

jreback:
very nice @almaleksia !

jreback:
if you wanted to move the JOIN_TYPES to the top-level ``pandas/conftest.py`` (and make it a fixture),  and prob rename to join_types, would be *really* helpful. note that this will percolate to lots of other modules, mainly in pandas/tests/reshape where they are actually used.

jreback:
I *think* we have an issue about this (or at least about generally moving things like this to fixtures). Also would be great to split up pandas/tests/frame/test_indexing.py along the same lines as you did here :-D

almaleksia:
@jreback thanks! I can create separate PRs for joins and frame/test_indexing refactoring

PR #19952, opened 2018-03-01
DOC/CI: run doctests on travis

jorisvandenbossche:
Trying to see if we can check the doctests on travis. We were already running them, but at this point they were not failing travis if they failed.

Closes https://github.com/pandas-dev/pandas/issues/19934 
Closes https://github.com/pandas-dev/pandas/issues/3439

Issue #18147, opened 2017-11-07
API Docs Rework 


TomAugspurger:
What do we want out of our API docs? What are the current issues?

I think we're (mostly) happy with `api.rst`: we manually list the classes and methods that we want to include the docs, grouped by topics of our choosing.

I think we're unhappy with some aspects of the auto-genrated class pages.

The [hack](https://github.com/pandas-dev/pandas/blob/c176a3c29ccabe7e471fa0b76c6cc1ceeb9bcd77/doc/sphinxext/numpydoc/numpydoc.py#L45-L48) in our vendored `numpydoc.py` "fixing" the autosummary to not include a table of methods for certain classes. This avoids a ton of redundant pages with the parent class, but we can't call out important differences like `RangeIndex.start`

We'd like a way to limit autosummary to specific methods in a class (I think this can be done with a `Methods` section in the docstring)

Tasks:

- [x] Limit API doc of classes to certain members (either extend autodoc or do it manually?) -> #18202
- [x] DOC: fix API docs generation with sphinx 1.6 #16705
- [x] DOC: documenting cython class (eg Timestamp): "cyfunction is not a python function" #5218 -> #18202
- [x] Remove vendored IPython sphinxext (#18193, #19657)
  - [x] should be OK with https://github.com/ipython/ipython/pull/10907 released)
  - [x] removed in #19657
  - [ ] We should check on Windows!
- [x] Align our numpydoc version with upstream / use upstream version: for this we somehow need to fix (or push upstream) our hacks to numpydoc ( I think this are the only relevant changes to the numpydoc source since it was last aligned with upstream in #6003) (xref #11257)):
  - [x] cf40991 initial commit for allowing classes without members table -> #18202
  - [x] 70c9d31 for None attributes
  - [x] Remove vendored numpydoc (work around https://github.com/numpy/numpydoc/pull/106#issuecomment-343220139)
- [x] Fix `cache_readonly` copying docstring (https://github.com/pandas-dev/pandas/issues/18197)
- [ ] `Index.name` and other class attributes. Either fix in numpydoc to use sphinx magic, or make into properties
- [x] `DataFrame.index` and `DataFrame.columns` (#18196)
- [x] Check performance (better than 1.6, but still slower? https://github.com/sphinx-doc/sphinx/issues/4230)

jorisvandenbossche:
I indeed think we at least want to have a way to only document certain methods/attributes on a class, to prevent including all. 

[...]

PR #20061, opened 2018-03-08
Add tests for docstring Validation Script + py27 compat

WillAyd:
- [x] closes #xxxx
- [x] tests added / passed
- [ ] whatsnew entry

@jorisvandenbossche @toobaz a few edits here should make it work for both Py3 and Py2 users


toobaz:
Tested, works great, thanks!

jreback:
could you add a test for this somewhere? (need to make sure the test runner runs this); or can you have the lint.sh script actually run (a sub-set of this). that way we know changes to this will be good.

WillAyd:
Fair point. The only issue with CI here is going to be the location of the files in the scripts folder, which I assume pytest nor the tests themselves would have access to given it's outside of the top level `pandas` package. `lint.sh` could fix that but I do think it would be better to have dedicated unit tests for the functionality.

Have to think it over but if you have any ideas on how to ideally configure the imports for testing let me know

jreback:
can just add 

pandas/tests/scripts might be ok

WillAyd:
Latest commit will fail but should be a starting point for a test module. @jorisvandenbossche @datapythonista I took these examples directly from the guide, but it looks like even the examples in the guide fail the validation script. In this case do you know if it's the script or the examples that need to be updated?

Here are the major failures quasi-summarized:
- Parameters {'kwargs'} not documented
- Unknown parameters {'**kwargs'}
- Parameter "**kwargs" has not type
- No returns section found (some methods don't return anything)
- No see also section found (docs technically say these are optional)
- No examples section found (docs suggest these may be required, but aren't explicit)
- Parameters {'letters', 'length'} not documented (this is wrong for test `random_letters`)
- Docstring text should start in the line immediately... (`head` examples in docs are wrong)

Curious to hear how you would like to approach the above

scripts/validate_docstrings.py
```diff
@@ -186,12 +185,11 @@ def signature_parameters(self):
             # accessor classes have a signature, but don't want to show this
             return tuple()
         try:
-            signature = inspect.signature(self.method_obj)
+            params = self.method_obj.__code__.co_varnames
```
jorisvandenbossche: unfortunately, this does not give the correct result (it seems to give some extra names). Eg 

```
In [156]: pd.DataFrame.apply.__code__.co_varnames
Out[156]: 
('self',
 'func',
 'axis',
 'broadcast',
 'raw',
 'reduce',
 'result_type',
 'args',
 'kwds',
 'frame_apply',
 'op')

```

the last two should not be there

toobaz:
With the new version of the script, an error of the kind:

```
Line 141, in pandas.DataFrame.apply
Failed example:
    df.apply(lambda x: Series([1, 2], index=['foo', 'bar']), axis=1)
Exception raised:
      File "/usr/lib/python3.5/doctest.py", line 1321, in __run
        compileflags, 1), test.globs)
      File "<doctest pandas.DataFrame.apply[7]>", line 1, in <module>
        df.apply(lambda x: Series([1, 2], index=['foo', 'bar']), axis=1)
      File "/home/nobackup/repo/pandas/pandas/core/frame.py", line 5004, in apply
        return op.get_result()
      File "/home/nobackup/repo/pandas/pandas/core/apply.py", line 136, in get_result
        return self.apply_standard()
      File "/home/nobackup/repo/pandas/pandas/core/apply.py", line 242, in apply_standard
        self.apply_series_generator()
      File "/home/nobackup/repo/pandas/pandas/core/apply.py", line 271, in apply_series_generator
        results[i] = self.f(v)
      File "<doctest pandas.DataFrame.apply[7]>", line 1, in <lambda>
        df.apply(lambda x: Series([1, 2], index=['foo', 'bar']), axis=1)
    NameError: ("name 'Series' is not defined", 'occurred at index 0')
```

becomes under python2

```
  File "validate_docstrings.py", line 496, in <module>
    sys.exit(main(args.function))
  File "validate_docstrings.py", line 482, in main
    return validate_one(function)
  File "validate_docstrings.py", line 459, in validate_one
    examples_errs = doc.examples_errors
  File "validate_docstrings.py", line 265, in examples_errors
    runner.run(test, out=f.write)
  File "/usr/lib/python2.7/doctest.py", line 1454, in run
    return self.__run(test, compileflags, out)
  File "/usr/lib/python2.7/doctest.py", line 1363, in __run
    self.report_failure(out, test, example, got)
  File "/usr/lib/python2.7/doctest.py", line 1228, in report_failure
    self._checker.output_difference(example, got, self.optionflags))
TypeError: unicode argument expected, got 'str'
```

(obtained with ``python validate_docstrings.py pandas.DataFrame.apply``).

toobaz:
As I stated in gitter, fixing the remaining issues for py2 support would be great, but I think changing ``#!/usr/bin/env python`` to ``#!/usr/bin/env python3`` and telling users that they must have python 3 is also perfectly acceptable.

WillAyd:
Thanks @toobaz and @jorisvandenbossche. So it looks like there are a couple of things to do here and with the sprint being tomorrow I'm wondering if it's even worth trying to get the compatibility to work and rather doing as @toobaz suggested and just explicitly requiring python3.

I think the bigger issue is that the test cases here taken from the doc don't pass the script. @jorisvandenbossche I saw your clarification on the `**kwargs` piece on the Google group, but curious what you think of the other issues?

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -0,0 +1,327 @@
...
+        >>> df.method(bar=14)
+        123
+        """
+        pass
+
+class TestValidator(object):
+
+    @pytest.fixture(autouse=True, scope="class")
+    def import_scripts(self):
+        up = os.path.dirname
```
jreback: pretty magical can you add a comment

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -0,0 +1,327 @@
+import os
```
jreback: not sure if the CI is actually running this?

jreback (changes requested):

WillAyd:
@datapythonista @jorisvandenbossche in the last commit I blended a few things together, namely:

- Added the test cases
- Fixed Py27 support
- Fixed varargs and keyword args validation AND
- Relaxed errors on some sections (namely Extended Summary and Examples)

I copied most of the test cases from the document, but even then some of them weren't passing the validation script so I took a few liberties to modify. Ideally we would build out test cases to unit test particular aspects of the validation script, but for now I've just done a blanket pass for success / failures.

scripts/validate_docstrings.py
```diff
@@ -440,24 +463,28 @@ def validate_one(func_name):
         for param_err in param_errs:
             errs.append('\t{}'.format(param_err))
 
-    if not doc.returns:
-        errs.append('No returns section found')
+    if not doc.returns and "return" in doc.method_source:
+        errs.append('No Returns section found')
+    if "yield" in doc.method_source:
```
WillAyd: This would ideally have the same structure as the returns check directly above it, but I think there's a bug in numpy doc where it doesn't parse Yields sections

jreback:
andas/tests/scripts/test_validate_docstrings.py ..............                                                                                                                                                                                                             [100%]
normal? maybe should catch this
```
================================================================================================================================ warnings summary =================================================================================================================================
pandas/tests/scripts/test_validate_docstrings.py::TestValidator::()::test_good_functions[sample_values]
  /Users/jreback/pandas/doc/sphinxext/numpydoc/docscrape.py:119: UserWarning: Unknown section Yields
    warn("Unknown section %s" % key)

-- Docs: http://doc.pytest.org/en/latest/warnings.html
```

WillAyd:
The issue with 'Yields' stems from the fact that we include a copy of `numpydocstr` directly in pandas, but it doesn't appear to be a recent version (`Yields` support was added back in numpydoc 0.6)

Is there a particular reason we decided to copy that package into pandas rather than manage as a dependency? @jorisvandenbossche

TomAugspurger:
Long story involving many awful hacks :)

https://github.com/pandas-dev/pandas/issues/18147

I think the final blocker is https://github.com/numpy/numpydoc/pull/106

WillAyd:
Woof...alright well I figure for now it's probably just easiest to port over 'Yields' support as it doesn't add that much overhead to what's already in place. Updated in latest push using https://github.com/numpy/numpydoc/pull/25/files as an implementation reference

doc/sphinxext/numpydoc/tests/test_docscrape.py
```diff
@@ -164,6 +179,36 @@ def test_returns():
     assert desc[0].startswith('This is not a real')
     assert desc[-1].endswith('anonymous return values.')
 
+def test_yields():
```
WillAyd: Doesn't look like we maintain the tests on our end as there are 4 failing on master, but figured it was worth adding these anyway. Adding the `numpydoc` directory to `PYTHONPATH` and running `pytest tests/test_docscrape.py -k "yield"` yields 3 passes here

doc/sphinxext/numpydoc/tests/test_docscrape.py
```diff
@@ -1,6 +1,7 @@
 # -*- encoding:utf-8 -*-
 from __future__ import division, absolute_import, print_function
 
+import re
```
jreback: can you move this to pandas/tests/scripts as well? (maybe tricky to get this to import correctly though)

jorisvandenbossche:
> Woof...alright well I figure for now it's probably just easiest to port over 'Yields' support as it doesn't add that much overhead to what's already in place. Updated in latest push using https://github.com/numpy/numpydoc/pull/25/files as an implementation reference

Could you do this in a separate PR? 
I would like to keep the changes to our vendored version of numpydoc as self contained as possible to keep easier track of this. 
Actually we might also just try to completely update the vendored version with upstream (and fix the few things we need)

doc/sphinxext/numpydoc/tests/test_docscrape.py
jorisvandenbossche: I would personally keep this here, it is a vendored package, not our own custom sphinx extension

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -0,0 +1,344 @@
...
+import numpy as np
+import pytest
+
+import pandas.util._test_decorators as td
+
+
+class GoodDocStrings(object):
+    """
+    Collection of good docstrings - be sure to update the tests as new
+    examples are added here
```
jorisvandenbossche: Maybe make this class docstring also an example to test? (to be sure that class docstrings also get checked correctly)

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -0,0 +1,344 @@
...
+        Also, note that the parameter descriptions do not start with a
+        capital letter, and do not finish with a dot.
+
+        Finally, the `**kwargs` parameter is missing.
+
+        Parameters
+        ----------
+
+        kind: str
+            kind of matplotlib plot
```
jorisvandenbossche: I would try to split up some of those "bad" methods so each function is focused on 1 bad element, to be sure we are testing that.

So for example, the plot one above could be split up in:

```
    def test_missing_kwargs(self, kind, **kwargs):
        """
        Kwargs argument is missing in parameters section.


        Parameters
        ----------
        kind : str
            Kind of matplotlib plot.
        """

    def test_parameter_colon_spacing(self, kind):
        """
        Space around colon is missing.

        Parameters
        ----------
        kind: str
            Kind of matplotlib plot.
        """

    def test_parameter_description_period(self, kind):
        """
        Parameter description does not end in a dot.

        Parameters
        ----------
        kind : str
            Kind of matplotlib plot
        """

    def test_parameter_description_capital(self, kind):
        """
        The parameter descriptions do not start with a capital.

        Parameters
        ----------
        kind : str
            kind of matplotlib plot.
        """

    def test_parameter_blank_line(self, kind):
        """
        Blank line between the parameters title and the first parameter. 
    
        Parameters
        ----------
        
        kind : str
            Kind of matplotlib plot.
        """
```

scripts/validate_docstrings.py
```diff
@@ -264,8 +277,7 @@ def examples_errors(self):
         error_msgs = ''
         for test in finder.find(self.raw_doc, self.method_name, globs=context):
             f = StringIO()
-            with contextlib.redirect_stdout(f):
-                runner.run(test)
```
jorisvandenbossche: This change is for py27 compat ? 
As I seem to remember that this caused another error (@toobaz do you recall?)

jorisvandenbossche (commented):
Thanks for your work on this!

scripts/validate_docstrings.py
WillAyd: Yes it is for compat. I think the other error you are referring to was present in the very first commit 4673475ec7098a7b4203ca716cdda8c029618fc6 but this is a little different and should pass for 2 and 3

pandas/tests/scripts/test_validate_docstrings.py
WillAyd: Yes this is a great idea and what I envision for this script long term (would make it clearer the purpose of the tests). 

For the initial change here I was just copying from the sprint docs, but I'll look how to integrate the above and any others that come to mind

WillAyd:
@jreback @jorisvandenbossche the latest push adds individual test cases to support specific units of the validation script. We have conflicting viewpoints about what to do with the changes I made to the `numpydoc` module to support adding `Yields`.

I'd be happy to try and add the tests for that module to the normal CI, but there are existing tests failing in there that I would need to fix that have nothing to do with this change - OK to do as a separate change? @jorisvandenbossche I also think it would be more trouble than it's worth to try and back out those changes from this change, as essentially it's bundled with the tests for the `Yield` section - OK to leave as is?

jorisvandenbossche:
> @jorisvandenbossche I also think it would be more trouble than it's worth to try and back out those changes from this change, as essentially it's bundled with the tests for the Yield section - OK to leave as is?

I would like to keep it separate, but I can also do that if you want.

jorisvandenbossche:
@WillAyd FYI, I am currently working on a full update of numpydoc, so you can for now ignore my comment about the numpydoc changes.

jorisvandenbossche:
@WillAyd PR is here: https://github.com/pandas-dev/pandas/pull/20383

jbrockmendel:
@WillAyd this looks nice, especially the tests.  Is getting this over the finish line an active goal?

WillAyd:
I figured this is less of a priority with the plans for Py27 but I'd be happy to rebase if you feel there's value in it

datapythonista:
There are still lots of docstrings that need to be fixed, and this PR adds a lot of value to the validation script. May be not so important the py2 compatibility now that the sprint is over, but surely the tests will be very useful when fixing the script and adding new validations.

jbrockmendel:
> Is by default repr-ed as follows in Py2:

```
from pandas.io.formats.printing import pprint_thing

s = {'**kwargs'}
out = 'Parameters {!s} not documented'.format(pprint_thing(s))

>>> print(out)
Parameters {**kwargs} not documented
```

Problem solved?

WillAyd:
@jbrockmendel thanks for the tip. I am under the impression that the quotes aren't that important, so I just updated the tests to assume the simple, compatible output

WillAyd:
Travis failures appear unrelated. I think this is in a good spot but as always would appreciate reviews.

@jbrockmendel @datapythonista

datapythonista:
lgtm, but I don't think the tests are running in Travis, are they?

WillAyd:
Where are you seeing that? Is there a collection output in the logs somewhere?

datapythonista:
I checked the logs, and didn't see them, so I assumed they weren't executed. But you surely now more about travis than me.

May be you can push a failing test, and we'll know for sure. ;)

WillAyd:
Haven't done much personally with Travis so I doubt that is the case :-)

Running

```sh
pytest pandas/ --collect-only | grep validate_docstrings
```

Certainly collects the new module. Does anyone else know a way of validating that this same behavior is reflected in Travis?

jbrockmendel:
> Does anyone else know a way of validating that this same behavior is reflected in Travis?

I think @datapythonista's suggestion is the simplest approach: put in a test that you expect to fail and check that Travis fails.

WillAyd:
Sorry @datapythonista I misread your post before but that approach makes a lot of sense. Trying in a bit...

WillAyd:
Not sure why I originally added a `skip_if_no('sphinx')` a few months back but that was the culprit for not running. Should be good to go now.

One failure in Travis is around clipboard tests so unrelated imo. Any other feedback lmk

WillAyd:
Any other feedback on this change? Hoping to avoid more rebasing :-)

datapythonista:
lgtm, just wanted to run it for some docstrings before merging it, and didn't have the time. But will do during the weekend.

scripts/validate_docstrings.py
```diff
@@ -181,18 +184,24 @@ def doc_parameters(self):
...
-        if (inspect.isclass(self.method_obj)
-                and self.method_name.split('.')[-1] in
-                self.method_obj._accessors):
-            # accessor classes have a signature, but don't want to show this
-            return tuple()
+        if inspect.isclass(self.method_obj):
+            if hasattr(self.method_obj, '_accessors') and (
+                    self.method_name.split('.')[-1] in
+                    self.method_obj._accessors):
+                # accessor classes have a signature but don't want to show this
```
jorisvandenbossche: This does not seem to work for me. Eg checking `pandas.Series.dt` with the script complains about parameter 'data' not being documented.

(but anyhow, it is better as on master, there it raises an attribute error somewhere in the script)

jorisvandenbossche (approved):
This looks good to me! 
Quickly skimmed through the code and changes look good to me. Some nice improvements! And tried some docstrings and seems to work nicely.


One thing (but not necessarily for this PR), when doing the check that a parameter description ends with a '.', we should check that the last line does not contain 'versionadded' (that's a typical false positive in the script that should not be hard to catch)

scripts/validate_docstrings.py
WillAyd: Good to know. I’d be happy to open that up as a follow up issue here - worth adding dedicated tests for accessors

datapythonista:
I've been doing tests, and besides what @jorisvandenbossche said, there are few other cases where the script still does not work (unrelated to this PR). For example `pandas.Timestamp.combine`, `pandas.Series.droplevel` or `pandas.Panel.transpose`.

Merging, so we can keep improving the scripts and the docs in separate PRs.

Thanks a lot @WillAyd, I know it's been a lot of work this PR, but it'll help a lot.

jorisvandenbossche:
@datapythonista what is wrong for `pandas.Series.droplevel` ?

datapythonista:
@jorisvandenbossche not sure what was the problem, I was executing the validation for all docstrings, and it failed, but probably caused by something in my changes.

I fixed the problem with `Panel.transpose` in #22408, which also allows running the validation of all docstrings, returning the result as a json. Only 11% pass all the validations. A bit more, as some will fail because of the problem with `.. versionadded::` (#22405). But seems like we're still far from having all ready.

jreback:
this broke all of the wheel building: 
https://travis-ci.org/MacPython/pandas-wheels/jobs/418057182

something wrong with the imports

jorisvandenbossche:
@jreback are the tests run there for an wheel installed version of pandas? (because I suppose that then the scripts directory is not available?)

If that is the case, easiest solution is probably to check the existence of the script, and otherwise simply skip the tests

jorisvandenbossche:
Ah, this is already discussed in https://github.com/pandas-dev/pandas/pull/22413

jorisvandenbossche:
@WillAyd the test log output on travis et al now includes output of the validation script from running it in its tests. We should maybe capture stdout for the tests here to avoid that?

WillAyd:
Makes sense - opened #22483 for that. 

On the road for the next few days so wouldn’t be able to get around to it personally until next week, though it’s out there for the community to pick up as well :-)

h-vetinari:
Since I rebased, I'm getting an error related to this PR (I believe), but only in the 3.6 travis job (https://travis-ci.org/pandas-dev/pandas/jobs/422781966 and https://travis-ci.org/pandas-dev/pandas/jobs/422883422):

```
UNEXPECTED EXCEPTION: SyntaxError('unexpected EOF while parsing', ('<doctest pandas.core.frame.DataFrame.duplicated[0]>', 1, 66, "data = {'species': ['lama', 'cow', 'lama', 'ant', 'lama', 'bee'],\n"))
```

The relevant section from the docstring is:
```
        Examples
        --------

        >>> data = {'species': ['lama', 'cow', 'lama', 'ant', 'lama', 'bee'],
                    'type': ['mammal'] * 3 + ['insect', 'mammal', 'insect']}
        >>> animals = pd.DataFrame(data, index=[1, 4, 9, 16, 25])
```

and I can't see what's wrong with it. How can one have a line-break in there, or is this just not supported / a bug?

jorisvandenbossche:
@h-vetinari it's not related to this PR, but to this one: https://github.com/pandas-dev/pandas/pull/19952 
But, this is actually the point of that test, to catch such errors. I commented on your PR what is wrong.

thoo:
When I run `scripts/validate_docstrings.py pandas.read_fwf` , I got this error.

```
Parameters {**kwds} not documented
                Unknown parameters {mangle_dupe_cols, compression, doublequote, warn_bad_lines, quotechar, usecols, na_values, converters, chunksize, skiprows, na_filter, true_values, escapechar, comment, memory_map, delim_whitespace, squeeze, low_memory, index_col, parse_dates, lineterminator, float_precision, iterator, dtype, keep_default_na, dialect, infer_datetime_format, encoding, dayfirst, decimal, verbose, delimiter, skip_blank_lines, quoting, names, tupleize_cols, error_bad_lines, header, nrows, keep_date_col, thousands, false_values, skipfooter, date_parser, prefix, skipinitialspace}
```
Should this one also fix **kwds? Thanks.

PR #20064, opened 2018-03-09
Docstring for pd.core.window.Expanding.kurt

toobaz:

Notice that I made two small changes to the ``Rolling.kurt``(by @WillAyd ) docstring too.

Namely:
- since window methods have ``kurt`` but not ``kurtosis`` (while e.g. ``Series`` has both), I guess it is best to tell users that ``kurt`` is the default in pandas - that is, use it in "See also"
- it is true that a ``DataFrame`` can be constructed directly from a list, but this is not the standard use, so I replaced with ``Series``

### DIFF
--- a/pandas/core/window.py
+++ b/pandas/core/window.py
@@ -926,28 +926,7 @@ class _Rolling_and_Expanding(_Rolling):
 
     Notes
     -----
-    A minimum of 4 periods is required for the rolling calculation.
-
-    Examples
-    --------
-    The below example will show a rolling calculation with a window size of
-    four matching the equivalent function call using `scipy.stats`.
-
-    >>> arr = [1, 2, 3, 4, 999]
-    >>> fmt = "{0:.6f}"  # limit the printed precision to 6 digits
-    >>> import scipy.stats
-    >>> print(fmt.format(scipy.stats.kurtosis(arr[:-1], bias=False)))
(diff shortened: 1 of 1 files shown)


WillAyd:
Thanks @toobaz. The merged update was actually incomplete so I have some conflicting work in #20044 which amongst other things accounts for your second point. I didn't think of your first point though. 

Probably easiest if I just include as part of mine and we close this out, otherwise we'll be doing double work to rebase. OK with that?

toobaz:
> Probably easiest if I just include as part of mine and we close this out, otherwise we'll be doing double work to rebase. OK with that?

Sure, just ping when done. Feel free to also include the "See also" descriptions (my last commit).

toobaz:
I mean: as long as @datapythonista is OK (level unlocked) ;-)

datapythonista:
lgtm :)

WillAyd:
@toobaz pushed a new commit in #20044 integrating a few things of what you had. Give it a look and let me know what you think

toobaz:
OK, I will rebase this on #20044 when merged, and update.

toobaz:
ready to merge, as far as I can tell

jreback:
lgtm. will be merged after sprints.

jorisvandenbossche (approved):

PR #20069, opened 2018-03-09
 DOC: Fixed example & description for pandas.cut

ikoevska:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

```
############################ Docstring (pandas.cut) ############################

Return indices of half-open `bins` to which each value of `x` belongs.

Use `cut` when you need to segment and sort data values into bins or
buckets of data. This function is also useful for going from a continuous
variable to a categorical variable. For example, `cut` could convert ages
to groups of age ranges.

Parameters
----------
x : array-like
    Input array to be binned. It has to be 1-dimensional.
bins : int, sequence of scalars, or pandas.IntervalIndex
    If `bins` is an int, defines the number of equal-width bins in the
    range of `x`. The range of `x` is extended by .1% on each side to
    include the min or max values of `x`.
    If `bins` is a sequence, defines the bin edges allowing for
    non-uniform bin width. No extension of the range of `x` is done.
right : bool, optional, default 'True'
    Indicates whether the `bins` include the rightmost edge or not. If
    `right == True` (the default), then the `bins` [1,2,3,4] indicate
    (1,2], (2,3], (3,4].
labels : array or bool, optional
    Used as labels for the resulting `bins`. Must be of the same length as
    the resulting `bins`. If False, returns only integer indicators of the
    `bins`.
retbins : bool, optional, default 'False'
    Whether to return the `bins` or not. Useful when `bins` is provided
    as a scalar.
precision : int, optional, default '3'
    The precision at which to store and display the `bins` labels.
include_lowest : bool, optional, default 'False'

Returns
-------
out : pandas.Categorical or Series, or array of int if `labels` is 'False'
    The return type depends on the input.
    If the input is a Series, a Series of type category is returned.
    Else - pandas.Categorical is returned. `Bins` are represented as
    categories when categorical data is returned.
bins : numpy.ndarray of floats
    Returned only if `retbins` is 'True'.

See Also
--------
qcut : Discretize variable into equal-sized buckets based on rank
    or based on sample quantiles.
pandas.Categorical : Represents a categorical variable in
    classic R / S-plus fashion.
Series : One-dimensional ndarray with axis labels (including time series).
pandas.IntervalIndex : Immutable Index implementing an ordered,
    sliceable set. IntervalIndex represents an Index of intervals that
    are all closed on the same side.

Notes
-----
Any NA values will be NA in the result. Out of bounds values will be NA in
the resulting pandas.Categorical object.

Examples
--------
>>> pd.cut(np.array([.2, 1.4, 2.5, 6.2, 9.7, 2.1]), 3, retbins=True)
... # doctest: +ELLIPSIS
([(0.19, 3.367], (0.19, 3.367], (0.19, 3.367], (3.367, 6.533], ...
Categories (3, interval[float64]): [(0.19, 3.367] < (3.367, 6.533] ...

>>> pd.cut(np.array([.2, 1.4, 2.5, 6.2, 9.7, 2.1]),
...        3, labels=["good", "medium", "bad"])
... # doctest: +SKIP
[good, good, good, medium, bad, good]
Categories (3, object): [good < medium < bad]

>>> pd.cut(np.ones(5), 4, labels=False)
array([1, 1, 1, 1, 1], dtype=int64)

Docstring for "pandas.cut" correct. :)
```

### DIFF
--- a/pandas/core/reshape/tile.py
+++ b/pandas/core/reshape/tile.py
@@ -25,6 +25,11 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
         include_lowest=False):
     """
     Return indices of half-open `bins` to which each value of `x` belongs.
+
+    Use `cut` when you need to segment and sort data values into bins or 
+    buckets of data. This function is also useful for going from a continuous
+    variable to a categorical variable. For example, `cut` could convert ages
+    to groups of age ranges.
     
     Parameters
     ----------
@@ -74,10 +79,6 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
 
     Notes
     -----
-    The `cut` function is useful for going from a continuous variable to
-    a categorical variable. For example, `cut` could convert ages to groups
-    of age ranges.
-
     Any NA values will be NA in the result. Out of bounds values will be NA in
     the resulting pandas.Categorical object.
 
@@ -95,7 +96,7 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
     Categories (3, object): [good < medium < bad]
 
     >>> pd.cut(np.ones(5), 4, labels=False)
-    array([1, 1, 1, 1, 1])
+    array([1, 1, 1, 1, 1], dtype=int64)
     """
     # NOTE: this binning code is changed a bit from histogram for var(x) == 0
 
--- a/pandas/core/reshape/tile.py
+++ b/pandas/core/reshape/tile.py
@@ -27,6 +27,11 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
         include_lowest=False):
     """
     Return indices of half-open `bins` to which each value of `x` belongs.
+
+    Use `cut` when you need to segment and sort data values into bins or 
+    buckets of data. This function is also useful for going from a continuous
+    variable to a categorical variable. For example, `cut` could convert ages
+    to groups of age ranges.
     
     Parameters
     ----------
@@ -76,10 +81,6 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
 
     Notes
     -----
-    The `cut` function is useful for going from a continuous variable to
-    a categorical variable. For example, `cut` could convert ages to groups
-    of age ranges.
-
     Any NA values will be NA in the result. Out of bounds values will be NA in
     the resulting pandas.Categorical object.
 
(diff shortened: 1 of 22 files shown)


pandas/core/reshape/tile.py
```diff
@@ -24,53 +24,64 @@
 def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
         include_lowest=False):
     """
-    Return indices of half-open bins to which each value of `x` belongs.
+    Return indices of half-open `bins` to which each value of `x` belongs.
```
jorisvandenbossche: I know it was already there, but I wondering if we can make this first sentence better. Because I have to say I have to read it very carefully to actually understand it :)

Some ideas:
 
- "Return indices" is actually not correct? As it just returns the intervals or bins itself?
- Looking at https://en.wikipedia.org/wiki/Discretization_of_continuous_features and https://en.wikipedia.org/wiki/Data_binning so maybe something with the terms "convert continuous values in discrete bins" ?

pandas/core/reshape/tile.py
```diff
@@ -24,53 +24,64 @@
...
-    right : bool, optional
-        Indicates whether the bins include the rightmost edge or not. If
-        right == True (the default), then the bins [1,2,3,4] indicate
+    bins : int, sequence of scalars, or pandas.IntervalIndex
+        If `bins` is an int, defines the number of equal-width bins in the
+        range of `x`. The range of `x` is extended by .1% on each side to
+        include the min or max values of `x`.
+        If `bins` is a sequence, defines the bin edges allowing for
+        non-uniform bin width. No extension of the range of `x` is done.
+    right : bool, optional, default 'True'
```
jorisvandenbossche: you can leave out the "optional" here (it's indeed optional to specify it, but it has a default value, so is not purely optional)

(same for the ones below where you have both 'optional' and 'default ..'

jorisvandenbossche (commented):
Thanks for the PR! Already added a few comments, will take a closer look later. 

For the examples, can you start it with a default example? So one not using `retbins=True`. And then you can after that explicitly say the difference if you do `retbins=True`. I would also show an example with Series to illustrate the return type explanation.

ikoevska:
I am closing this one as it got totally messed up after a rebase. Will open a new one with @jorisvandenbossche comments applied.

jorisvandenbossche:
@ikoevska For future reference, please update this PR. Even if you make a new branch locally, you can force push to the same branch on your fork, and this PR gets updated. But no problem for this time!

PR #20072, opened 2018-03-09
Docstring validation script: handle args/kwargs

jorisvandenbossche:
This ensures that if you use `**kwargs` in the docstring Parameters, the validation script does not complain. 
But it does not yet check it is actually written like that.

### DIFF
--- a/scripts/validate_docstrings.py
+++ b/scripts/validate_docstrings.py
@@ -175,9 +175,9 @@ class Docstring:
 
     @property
     def doc_parameters(self):
-        return collections.OrderedDict((name.strip('*'), (type_, ''.join(desc)))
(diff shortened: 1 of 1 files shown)


TomAugspurger:
FYI, we use `**kwds` in some of the plotting docs. I haven't looked, but does that mess things up? We can / should just change those to `**kwargs`.

jreback:
lgtm. maybe we could/should run this in our lint.sh script? (or as an additional item in the CI). Gives much more deedback than the doc-build .

jorisvandenbossche:
Check is included for this in https://github.com/pandas-dev/pandas/pull/20061, so closing this

PR #20074, opened 2018-03-09
DOC/CLN: clean-up shared_docs in generic.py

jorisvandenbossche:
xref https://github.com/pandas-dev/pandas/pull/20016


TomAugspurger (approved):
The diff is a bit hard to read but I think things look OK.

jreback:
cc @jorisvandenbossche

jbrockmendel:
@jorisvandenbossche needs rebase but I think the consensus is this is pretty much good to go

datapythonista (approved):
I rebased this. lgtm. Happy to have it merged. I think this simplifies things in the shared docs.

jorisvandenbossche:
@datapythonista thanks!

(don't merge yet, the failure is relevant)

jorisvandenbossche:
So changing `@Appender(_shared_docs['reindex'] % _shared_doc_kwargs)` with `@Appender(NDFrame.reindex.__doc__ % _shared_doc_kwargs)` fails if there is the optimization to remove docstrings (`__doc__` will then be None, and the `%` operation will fail). 

Are we fine with changing it to (using our `Subsitution` decorator which handles this corner case):

```
@Substitution(**_shared_doc_kwargs)
@Appender(NDFrame.reindex.__doc__)
```

(this is already used in some places)

Or do we rather want that the `Appender` can already handle this (`@Appender(NDFrame.reindex.__doc__, _shared_doc_kwargs)` or `@Appender(NDFrame.reindex.__doc__, **_shared_doc_kwargs)`) ?

Since we already do the first, I will update the PR with this

datapythonista:
+1 for the first. We can review whether `Appender` and `Substitution` can be improved in the future, but if the first works, I think it's better to use it for now.

jbrockmendel:
> [...] fails if there is the optimization to remove docstrings

Just opened #22927 that I think addresses this (although now that I look at your comment more closely, I no longer think it makes a difference)

PR #20076, opened 2018-03-09
DOC: Improved the docstring of errors.ParserWarning

joaoavf:
an unrelated PR):

```
################### Docstring (pandas.errors.ParserWarning)  ###################

Warning raised when reading a file that doesn't use the default parser.

Thrown by `pd.read_csv` and `pd.read_table` when it is necessary to
change parsers, generally from 'c' to 'python'.

It happens due to lack of support or functionality for parsing
particular attributes of a CSV file with the requested engine.

Currently, C-unsupported options include the following parameters:

1. `sep` other than a single character (e.g. regex separators)
2. `skipfooter` higher than 0
3. `sep=None` with `delim_whitespace=False`

The warning can be avoided by adding `engine='python'` as a parameter
in `pd.read_csv` and `pd.read_table` methods.

See Also
--------
pd.read_csv : Read CSV (comma-separated) file into DataFrame.
pd.read_table : Read general delimited file into DataFrame.

Examples
--------
Using a `sep` in `pd.read_csv` other than a single character:

>>> import io
>>> csv = u'''a;b;c
...           1;1,8
...           1;2,1'''
>>> df = pd.read_csv(io.StringIO(csv), sep='[;,]')
...
ParserWarning: Falling back to the 'python' engine...

Adding `engine='python'` to `pd.read_csv` removes the Warning:

>>> df = pd.read_csv(io.StringIO(csv), sep='[;,]', engine='python')
scripts/validate_docstrings.py:1: ParserWarning: Falling back to the 'python' engine because the 'c' engine does not support regex separators (separators > 1 char and different from '\s+' are interpreted as regex); you can avoid this warning by specifying engine='python'.
  #!/usr/bin/env python

Errors found:
        No returns section found
        Examples do not pass tests

Line 32, in pandas.errors.ParserWarning
Failed example:
    df = pd.read_csv(io.StringIO(csv), sep='[;,]')
Expected:
    ...
    ParserWarning: Falling back to the 'python' engine...
Got nothing

```

I am documenting a Warning and I could not find a better way to display the warning in the html example other than using a "Traceback (most recent call last):" followed by "ParserWarning: Falling back to the 'python' engine..." in the docstring.

It also says that it found errors about "No returns sections found". On what I understood this is not relevant to the docstring in hand.

### DIFF
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -84,9 +84,7 @@ class ParserWarning(Warning):
     ...           1;1,8
     ...           1;2,1'''
     >>> df = pd.read_csv(io.StringIO(csv), sep='[;,]')
-    Traceback (most recent call last):
-    ...
-    ParserWarning: Falling back to the 'python' engine...
+    ... # ParserWarning: Falling back to the 'python' engine...
 
     Adding `engine='python'` to `pd.read_csv` removes the Warning:
 
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -53,10 +53,37 @@ class EmptyDataError(ValueError):
 
 class ParserWarning(Warning):
     """
-    Warning that is raised in `pd.read_csv` whenever it is necessary
-    to change parsers (generally from 'c' to 'python') contrary to the
-    one specified by the user due to lack of support or functionality for
-    parsing particular attributes of a CSV file with the requested engine.
+    Warning raised in `pd.read_csv` and `pd.read_table` when it is
+    necessary to change parsers, generally from 'c' to 'python'.
+    
+    It happens due to lack of support or functionality for parsing
+    particular attributes of a CSV file with the requested engine.
+
+    Currently, C-unsupported options include the following parameters:
+
+    1. `sep` other than a single character (e.g. regex separators)
+    2. `skipfooter` higher than 0
+    3. `sep=None` with `delim_whitespace=False`
+
+    The warning can be avoided by adding `engine='python'` as a parameter
+    in `pd.read_csv` and `pd.read_table` methods.
+
+    Examples
+    --------
+    Using a `sep` in `pd.read_csv` other than a single character:
+
+    >>> import io
+    >>> csv = u'''a;b;c
+    ...           1;1,8
+    ...           1;2,1'''
+    >>> df = pd.read_csv(io.StringIO(csv), sep='[;,]')
+    Traceback (most recent call last):
+    ...
+    ParserWarning: Falling back to the 'python' engine...
+
+    Adding `engine='python'` to `pd.read_csv` removes the Warning:
+
+    >>> df = pd.read_csv(io.StringIO(csv), sep='[;,]', engine='python')
     """
 
 
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -55,20 +55,20 @@ class ParserWarning(Warning):
     """
     Warning raised when reading a file that doesn't use the default parser.
 
-    Thrown by `pd.read_csv` and `pd.read_table` when it is necessary to
-    change parsers, generally from 'c' to 'python'.
+    Raised by `pd.read_csv` and `pd.read_table` when it is necessary to change
+    parsers, generally from the default 'c' parser to 'python'.
 
-    It happens due to lack of support or functionality for parsing
-    particular attributes of a CSV file with the requested engine.
+    It happens due to a lack of support or functionality for parsing a
+    particular attribute of a CSV file with the requested engine.
 
-    Currently, C-unsupported options include the following parameters:
+    Currently, 'c' unsupported options include the following parameters:
 
     1. `sep` other than a single character (e.g. regex separators)
     2. `skipfooter` higher than 0
     3. `sep=None` with `delim_whitespace=False`
 
-    The warning can be avoided by adding `engine='python'` as a parameter
-    in `pd.read_csv` and `pd.read_table` methods.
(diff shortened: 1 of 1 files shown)


pandas/errors/__init__.py
```diff
@@ -53,10 +53,37 @@ class EmptyDataError(ValueError):
...
+    Using a `sep` in `pd.read_csv` other than a single character:
+
+    >>> import io
+    >>> csv = u'''a;b;c
+    ...           1;1,8
+    ...           1;2,1'''
+    >>> df = pd.read_csv(io.StringIO(csv), sep='[;,]')
+    Traceback (most recent call last):
+    ...
+    ParserWarning: Falling back to the 'python' engine...
```
datapythonista: did you check why the validation says that this test didn't pass, and that the `read_csv` returned nothing?

pandas/errors/__init__.py
```diff
@@ -53,10 +53,37 @@ class EmptyDataError(ValueError):
 
 class ParserWarning(Warning):
     """
-    Warning that is raised in `pd.read_csv` whenever it is necessary
-    to change parsers (generally from 'c' to 'python') contrary to the
-    one specified by the user due to lack of support or functionality for
-    parsing particular attributes of a CSV file with the requested engine.
+    Warning raised in `pd.read_csv` and `pd.read_table` when it is
+    necessary to change parsers, generally from 'c' to 'python'.
+
```
datapythonista: The first line needs to fit in a line. Can you write something more concise please? This paragraph is really useful, and it surely needs to be in the description, but the first line is used in some summaries that should be shorter. Something like `Warning raised when reading a table does not use the default parser`. Not sure if it's accurate or fits in one line, but to give you an idea.

pandas/errors/__init__.py
```diff
@@ -53,10 +53,37 @@ class EmptyDataError(ValueError):
...
+    Currently, C-unsupported options include the following parameters:
+
+    1. `sep` other than a single character (e.g. regex separators)
+    2. `skipfooter` higher than 0
+    3. `sep=None` with `delim_whitespace=False`
+
+    The warning can be avoided by adding `engine='python'` as a parameter
+    in `pd.read_csv` and `pd.read_table` methods.
+
```
datapythonista: I think `read_csv` and `read_table` are good candidates for a `See Also` section, as you're mentioning them.

datapythonista (commented):
Looks great, added couple of comments.

pandas/errors/__init__.py
joaoavf: When I ran the code in my console I had this warning displayed: 'ParserWarning: Falling back to the 'python' engine...'

I thought it might have something to do as it is a warning and not an error. Something along the lines that the kind of output generated by an error could be caught by Traceback but not the output of a warning.

Any ideas on how to fix and approach this?

pandas/errors/__init__.py
joaoavf: Great Mark! Thanks for the suggestion. Already commited my version of it.

pandas/errors/__init__.py
joaoavf: Added a `See Also` section with `read_csv` and `read_table`.

pandas/errors/__init__.py
```diff
@@ -53,10 +53,44 @@ class EmptyDataError(ValueError):
 
 class ParserWarning(Warning):
     """
-    Warning that is raised in `pd.read_csv` whenever it is necessary
-    to change parsers (generally from 'c' to 'python') contrary to the
-    one specified by the user due to lack of support or functionality for
-    parsing particular attributes of a CSV file with the requested engine.
+    Warning raised when reading a file that doesn't use the default parser.
+
+    Thrown by `pd.read_csv` and `pd.read_table` when it is necessary to
```
jreback: Thrown -> Raised

pandas/errors/__init__.py
```diff
@@ -53,10 +53,44 @@ class EmptyDataError(ValueError):
 
 class ParserWarning(Warning):
     """
-    Warning that is raised in `pd.read_csv` whenever it is necessary
-    to change parsers (generally from 'c' to 'python') contrary to the
-    one specified by the user due to lack of support or functionality for
-    parsing particular attributes of a CSV file with the requested engine.
+    Warning raised when reading a file that doesn't use the default parser.
```
jreback: say default is the c parser

pandas/errors/__init__.py
```diff
@@ -53,10 +53,44 @@ class EmptyDataError(ValueError):
...
+
+    Thrown by `pd.read_csv` and `pd.read_table` when it is necessary to
+    change parsers, generally from 'c' to 'python'.
+
+    It happens due to lack of support or functionality for parsing
```
jreback: due to a lack

for parsing a particular attribute

pandas/errors/__init__.py
```diff
@@ -53,10 +53,44 @@ class EmptyDataError(ValueError):
...
+    particular attributes of a CSV file with the requested engine.
+
+    Currently, C-unsupported options include the following parameters:
```
jreback: 'c' unsupported options

jreback (changes requested):

pandas/errors/__init__.py
joaoavf: Jeff, I made all the changes you requested. Thank you for the feedback.

jreback (approved):

TomAugspurger:
https://github.com/pandas-dev/pandas/issues/20309#issuecomment-373495419 for doctesting warnings. Thanks @joaoavf

PR #20077, opened 2018-03-09
DOC: fix the pandas.DataFrame.add example

robinkiplangat:
Docstring assigned to the PyData Nairobi chapter for the sprint
Checklist for the pandas documentation sprint 

- [x] PR title is "DOC: update the pandas.DataFrame.rmul docstring"
- [x] The validation script passes: `scripts/validate_docstrings.py pandas.DataFrame.rmul`
- [x] The html version looks good: `python doc/make.py --single pandas.DataFrame.rmul`

`scripts/validate_docstrings.py pandas.DataFrame.rmul`
```
###################### Docstring (pandas.DataFrame.rmul)  ######################

Multiplication of dataframe and other, element-wise (binary operator `rmul`).

Equivalent to ``other * dataframe``, but with support to substitute a fill_value for
missing data in one of the inputs.

Parameters
----------
other : Series, DataFrame, or constant
axis : {0, 1, 'index', 'columns'}
    For Series input, axis to match Series index on
level : int or name
    Broadcast across a level, matching Index values on the
    passed MultiIndex level
fill_value : None or float value, default None
    Fill existing missing (NaN) values, and any new element needed for
    successful DataFrame alignment, with this value before computation.
    If data in both corresponding DataFrame locations is missing
    the result will be missing

Notes
-----
Mismatched indices will be unioned together

Returns
-------
result : DataFrame

Examples
--------
>>> a = pd.DataFrame([1, 1, 1, np.nan], index=['a', 'b', 'c', 'd'],
...                  columns=['one'])
>>> a
   one
a  1.0
b  1.0
c  1.0
d  NaN
>>> b = pd.DataFrame(dict(one=[1, np.nan, 1, np.nan],
...                       two=[np.nan, 2, np.nan, 2]),
...                   index=['a', 'b', 'd', 'e'])
>>> b
   one  two
a  1.0  NaN
b  NaN  2.0
d  1.0  NaN
e  NaN  2.0
>>> a.add(b, fill_value=0)
   one  two
a  2.0  NaN
b  1.0  2.0
c  1.0  NaN
d  1.0  NaN
e  NaN  2.0

See also
--------
DataFrame.mul

Docstring for "pandas.DataFrame.rmul" correct. :)
```

explicitly.


pandas/core/ops.py
```diff
@@ -477,16 +477,16 @@ def _get_op_name(op, special):
...
    one
 a  1.0
 b  1.0
 c  1.0
 d  NaN
 >>> b = pd.DataFrame(dict(one=[1, np.nan, 1, np.nan],
-                          two=[np.nan, 2, np.nan, 2]),
-                     index=['a', 'b', 'd', 'e'])
+...                       two=[np.nan, 2, np.nan, 2]),
+...                 index=['a', 'b', 'd', 'e'])
```
TomAugspurger: May need one more space here, so that the "i" in "index" is below the "d" in "dict'.

pandas/core/ops.py
```diff
@@ -513,14 +513,14 @@ def _get_op_name(op, special):
 other : Series, DataFrame, or constant
 axis : {{0, 1, 'index', 'columns'}}
     For Series input, axis to match Series index on
+level : int or name
```
TomAugspurger: I think `level : int or str`, though we're inconsistent here.

@jorisvandenbossche what should our policy be on index positions or labels? In principle, `int or str` isn't quite right since index labels needn't be strings. It'd be `int or object`, which isn't that informative. `int or label`?

jorisvandenbossche: Hmm, yes, this is a dubious one. "name" is technically not a 'type', but "str" would not be correct as you say. So maybe better to keep it that way. 

If we keep the 'name' or 'label', wouldn't 'name' be better because it is here a index level name, not a label inside the index ?

TomAugspurger: Ah, yes you're right I was mistaken.

And we can declare that "name" means anything that can go in `{.index,.columns}{.name,.names}`.

jorisvandenbossche (approved):
Fixed the merge conflict.

jorisvandenbossche:
@4bic Thanks!

4bic:
Hi @jorisvandenbossche, 
`DOC:update the pandas.DataFrame.rmul docstring` was renamed as a `DOC:fix the pandas.DataFrame.add example`. 

Was the dostring the same for the [pandas.DataFrame.rmul](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.rmul.html) and **pandas.DataFrame.add example**

jorisvandenbossche:
Ah, I renamed because the *example* you fixed was for `add`, but the other changes were for both rmul and add

Issue #12863, opened 2016-04-11
Series.any() and .all() don't return bool values if dtype=object


jdavies1618:
#### Code Sample, a copy-pastable example if possible

```
s = pd.Series(index=range(5), data=['a', 'b', 'c', 'd', 'e'], dtype=object)
print [s.any(), s.all()]
```
#### Expected Output

`[True, True]`
#### Actual output

`['a', 'e']`
#### output of `pd.show_versions()`

```
INSTALLED VERSIONS
------------------
commit: None
python: 2.7.11.final.0
python-bits: 64
OS: Darwin
OS-release: 15.0.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8

pandas: None
nose: 1.3.7
pip: 8.1.1
setuptools: None
Cython: None
numpy: 1.10.1
scipy: 0.16.1
statsmodels: None
IPython: 4.0.0
sphinx: 1.3.3
patsy: None
dateutil: 2.4.2
pytz: 2015.7
blosc: None
bottleneck: None
tables: None
numexpr: None
matplotlib: 1.4.3
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
httplib2: None
apiclient: None
sqlalchemy: 1.0.11
pymysql: None
psycopg2: None
Jinja2: None
```

jreback:
what you are doing doesn't make any sense but it is correct.

http://docs.scipy.org/doc/numpy-1.10.1/reference/generated/numpy.any.html

I guess this *should* return a boolean. But to be honest I don't know what numpy is doing.

```
In [2]: np.array(['a','b','c'],dtype='object').any()
Out[2]: 'a'
```

cc @charris
cc @njsmith

charris:
Hmm..., yes, it is correct
```
In [1]: 'a' or 'b'
Out[1]: 'a'
```
I suppose we could call `bool` on the result for the logical operators.
```
In [2]: bool('a' or 'b')
Out[2]: True
```

jreback:
yep, looks like this is the case for ``dtype=object`` in general
```
In [4]: np.array([1,2,3],dtype=object).any()
Out[4]: 1
```

jreback:
ok, I think we will add a ``bool()`` if its a scalar on the pandas side for compat.

@jdavies1618 want to do a pull-request?

jdavies1618:
[...]

PR #20078, opened 2018-03-09
DOC: Improved the docstring of Series.any()

FrankDupree:
an unrelated PR):

- [x] PR title is "DOC: Improved the docstring of Series.any()"
- [ ] The validation script passes: `scripts/validate_docstrings.py Series.any()`
- [x] The html version looks good: `python doc/make.py --single Series.any()`

```
######################## Docstring (pandas.Series.any)  ########################

Return whether any element is True over requested axis.

Returns True if one (or more) elements are non-zero,
not-empty or not-False.

Parameters
----------
axis : {index (0)}
    will be NA.
    If the axis is a MultiIndex (hierarchical), count along a
    particular level, collapsing into a scalar.
bool_only : boolean, default None
    Include only boolean columns. If None, will attempt to use everything,
    then use only boolean data. Not implemented for Series.
**kwargs : Additional keywords have no effect but might be accepted for
    compatibility with numpy.

Returns
-------
any : scalar or Series (if level specified)

Examples
--------
>>> s1 = pd.Series([1, 2, 3])
>>> s1.any()
True

>>> import numpy as np
>>> s2 = pd.Series([np.NaN, np.NaN, np.NaN])
>>> s2.any()
False

Errors found:
        Errors in parameters section
                Parameter "axis" has no description
                Parameter "**kwargs" description should start with capital letter
        See Also section not found
```

Participants of the documentation sprint were told to go ahead with the description of the "**kwargs" parameter. 

Parameter "axis" has no description looks like a bug to me. Noticed some other methods show the same validation error. e.g "pandas.Series.mean" throws the same validation error.


TomAugspurger:
Does this close a specific issue?

pandas/core/generic.py
```diff
@@ -7784,25 +7784,49 @@ def _doc_parms(cls):
...
+    compatibility with numpy.
 
 Returns
 -------
-%(outname)s : %(name1)s or %(name2)s (if level specified)\n"""
+%(outname)s : %(name1)s or %(name2)s (if level specified)
+
+Examples
+--------
+>>> s1 = pd.Series([1,2,3])
```
TomAugspurger: Could you add spaces after the commas here so this is pep8 compliant?

pandas/core/generic.py
```diff
@@ -7784,25 +7784,49 @@ def _doc_parms(cls):
...
+>>> s1.any()
+True
+
+>>> import numpy as np
+>>> s2 = pd.Series([np.NaN,np.NaN,np.NaN])
+>>> s2.any()
+False
+
+>>> s3 = pd.Series([1,2,3,"Hobbit"])
```
TomAugspurger: This last one is a bug: https://github.com/pandas-dev/pandas/issues/12863 It should return True / False.

I think just remove it for now.

FrankDupree: yeah, I sensed it 👍

pandas/core/generic.py
FrankDupree: Okay, will do.

pandas/core/generic.py
FrankDupree: Done.

pandas/core/generic.py
FrankDupree: Done.

datapythonista:
@FrankDupree can you please read carefully the doc on how to submit a PR [1], and make sure you do all the steps correctly? You surely missed the point about the title of the PR, and also copying the output of the `validate_docstrings.py` in the description. Without those, it's difficult to understand what you sent and review it.

Btw, in the description of the PR, the `[ ]` are checkboxes. You should make sure you completed every listed point, and mark them with an `[X]`, so we know you did.

Thank you!

1. https://python-sprints.github.io/pandas/guide/pandas_pr.html

toobaz:
Are ``Parameter "axis" has no description`` and ``See Also section not found`` solved? If yes, could you update the description?

pandas/core/generic.py
```diff
@@ -7784,25 +7784,41 @@ def _doc_parms(cls):
...
 _bool_doc = """
-
 %(desc)s
 
+Returns True if one (or more) elements are non-zero,
+not-empty or not-False.
+
 Parameters
 ----------
 axis : %(axis_descr)s
```
toobaz: Missing description

toobaz (changes requested):

pandas/core/generic.py
toobaz: (never mind)

toobaz:
Please update the description with the validation script output

pandas/core/generic.py
```diff
@@ -7784,25 +7784,41 @@ def _doc_parms(cls):
...
-    will be NA
+    will be NA.
 level : int or level name, default None
     If the axis is a MultiIndex (hierarchical), count along a
-    particular level, collapsing into a %(name1)s
+    particular level, collapsing into a %(name1)s.
 bool_only : boolean, default None
     Include only boolean columns. If None, will attempt to use everything,
     then use only boolean data. Not implemented for Series.
+**kwargs : Additional keywords have no effect but might be accepted for
```
jreback: @datapythonista is this the language generally recommending here?

jreback (approved):

TomAugspurger:
@FrankDupree Some things have changed in master. Can you check if there's anything else to be done?

pandas/core/generic.py
datapythonista: I didn't see this earlier. I think the description is similar to the ones used in `kwargs` with similar behavior. But it should go in the next line, as in the other parameters, and in this line simply leave `**kwargs`

FrankDupree:
@TomAugspurger I will.

jorisvandenbossche:
So it seems we had a conflicting PR https://github.com/pandas-dev/pandas/pull/20217 that implemented `DataFrame.any`, and since Series and DataFrame share code, this might have been duplicate work .. Sorry about that! 

I am going to close this PR, but please check if there is anything in your PR that is not in http://pandas-docs.github.io/pandas-docs-travis/generated/pandas.Series.any.html#pandas.Series.any. 
For example, I think it would be good to add an example of what an empty Series returns (Series([]).any()). Feel free to open a PR for that.

PR #20086, opened 2018-03-09
DOC: update the Series.memory_usage() docstring

lebigot:
an unrelated PR):

Output of the validation script:

```
#################### Docstring (pandas.Series.memory_usage) ####################

Return the memory usage of the Series.

The memory usage can optionally include the contribution of
the index and of elements of `object` dtype.

Parameters
----------
index : bool, default True
    Specifies whether to include the memory usage of the Series index.
deep : bool, default False
    If True, introspect the data deeply by interrogating
    `object` dtypes for system-level memory consumption, and include
    it in the returned value.

Returns
-------
int
    Bytes of memory consumed.

See Also
--------
numpy.ndarray.nbytes : Total bytes consumed by the elements of the
    array.

Examples
--------

>>> s = pd.Series(range(3))
>>> s.memory_usage()
104

Not including the index gives the size of the rest of the data, which
is necessarily smaller:

>>> s.memory_usage(index=False)
24

The memory footprint of `object` values is ignored by default:

>>> class MyClass: pass
>>> s = pd.Series(MyClass())
>>> s
0    <__main__.MyClass object at ...>
dtype: object
>>> s.memory_usage()
88
>>> s.memory_usage(deep=True)
120

Errors found:
        Examples do not pass tests

Line 43, in pandas.Series.memory_usage
Failed example:
    s
Expected:
    0    <__main__.MyClass object at ...>
    dtype: object
Got:
    0    <__main__.MyClass object at 0x10f8076d8>
    dtype: object
```

Error due to following the recommendation about not giving non-reproducible addresses.

### DIFF
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -2705,9 +2705,9 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         Parameters
         ----------
         index : bool, default True
-            Specifies whether to include memory usage of Series index.
+            Specifies whether to include the memory usage of the Series index.
         deep : bool, default False
-            Specifies whether to introspect the data deeply, interrogate
+            If True, introspect the data deeply by interrogating
             `object` dtypes for system-level memory consumption, and include
             it in the returned value.
 
@@ -2728,7 +2728,8 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         >>> s.memory_usage()
         104
 
-        Not including the index gives the size of the rest of the data:
+        Not including the index gives the size of the rest of the data, which
+        is necessarily smaller:
 
         >>> s.memory_usage(index=False)
         24
@@ -2736,6 +2737,9 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         The memory footprint of `object` values is ignored by default:
 
         >>> s = pd.Series(object())
+        >>> s
+        0    <object object at 0x118acf230>
+        dtype: object
         >>> s.memory_usage()
         88
         >>> s.memory_usage(deep=True)  # Footprint of object() included
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -2720,6 +2720,7 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         --------
         numpy.ndarray.nbytes : Total bytes consumed by the elements of the
             array.
+        DataFrame.memory_usage : Bytes consumed by a DataFrame.
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


pandas/core/series.py
```diff
@@ -2696,28 +2696,55 @@ def reindex_axis(self, labels, axis=0, **kwargs):
...
-        are not components of the array if deep=False
+        int
+            Bytes of memory consumed.
 
         See Also
         --------
-        numpy.ndarray.nbytes
+        numpy.ndarray.nbytes : Total bytes consumed by the elements of the
+            array.
+
```
jorisvandenbossche: Maybe also DataFrame.memory_usage ?

pandas/core/series.py
```diff
@@ -2696,28 +2696,55 @@ def reindex_axis(self, labels, axis=0, **kwargs):
...
+        Not including the index gives the size of the rest of the data, which
+        is necessarily smaller:
+
+        >>> s.memory_usage(index=False)
+        24
+
+        The memory footprint of `object` values is ignored by default:
+
+        >>> class MyClass: pass
+        >>> s = pd.Series(MyClass())
```
jorisvandenbossche: I would maybe use an existing python object (something from the standard library, eg decimal.Decimal, IPAddress object, ... or your other favorite python object)

jorisvandenbossche: Or, maybe also useful, to just use some strings. Because of storing them as object (typical gotcha in pandas), there memory is not reflected well by default in this

jorisvandenbossche (commented):
Nice docstring! 
Added a few comments

pandas/core/series.py
TomAugspurger: Yes, I like the idea of using strings.

You'll need to make sure the string is is not interned by the interprepter, so don't use just `pd.Series(['a', 'b'])`

lebigot: Why not use interned strings? They are longer than their pointer, so they also are a good example:
```
>>> s = pd.Series(["a", "b"])
>>> s.memory_usage()
96
>>> s.memory_usage(deep=True)
212
```

jorisvandenbossche: so apparantly it is giving the memory of that object, so for the example that looks fine (but I would expect that if you have repeated interned strings, it might then not give the correct result)

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks a lot!

pandas/core/series.py
lebigot: A repeated interned string does not give the real memory footprint indeed:
```
>>> id("a")
4397773352
>>> id("a")
4397773352
>>> s = pd.Series(["a"]*1000)
>>> s.memory_usage()
8080
>>> s.memory_usage(deep=True)
66080
```

PR #20099, opened 2018-03-10
DOC: update the docstring for several functions and properties (Seoul).

coffeedjimmy:
Checklist for the pandas documentation sprint (ignore this if you are doing
an unrelated PR):

- [x] PR title is "DOC: update the <your-function-or-method> docstring"
- [x] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [x] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [x] It has been proofread on language by another sprint participant

Please include the output of the validation script below between the "```" ticks:

```
`pd.DataFrame.get_dtype_counts`

################################################################################
################ Docstring (pandas.DataFrame.get_dtype_counts)  ################
################################################################################

Return counts of unique dtypes in this object.

Returns
-------
dtype   Number of dtype

See Also
--------
dtypes : Return the dtypes in this object.

Examples
--------
>>> a = [['a', 1, 1.0], ['b', 2, 2.0], ['c', 3, 3.0]]
>>> df = pd.DataFrame(a, columns=['str', 'int', 'float'])
>>> df['int'].astype(int)
>>> df['float'].astype(float)
>>> df.get_dtype_counts()
float64    1
int64      1
object     1
dtype: int64

################################################################################
################################## Validation ##################################
################################################################################

Errors found:
	No extended summary found
	Examples do not pass tests

################################################################################
################################### Doctests ###################################
################################################################################

**********************************************************************
Line 16, in pandas.DataFrame.get_dtype_counts
Failed example:
    df['int'].astype(int)
Expected nothing
Got:
    0    1
    1    2
    2    3
    Name: int, dtype: int64
**********************************************************************
Line 17, in pandas.DataFrame.get_dtype_counts
Failed example:
    df['float'].astype(float)
Expected nothing
Got:
    0    1.0
    1    2.0
    2    3.0
    Name: float, dtype: float64

`pd.DataFrame.get_ftype_counts`

################################################################################
################ Docstring (pandas.DataFrame.get_ftype_counts)  ################
################################################################################

Return counts of unique ftypes in this object.

Returns
-------
dtype   Number of dtype:dense|sparse

See Also
--------
ftypes : Return
         ftypes (indication of sparse/dense and dtype) in this object.

Examples
--------
>>> a = [['a', 1, 1.0], ['b', 2, 2.0], ['c', 3, 3.0]]
[...]

PR #20100, opened 2018-03-10
DOC: update the dtypes/ftypes docstring (Seoul)

jongwony:
an unrelated PR):

```


##################### Docstring (pandas.DataFrame.dtypes)  #####################

Return the dtypes in this object.

Notes
-----
It returns a Series with the data type of each column.
If object contains data multiple dtypes in a single column,
dtypes will be chosen to accommodate all of the data types.
``object`` is the most general.

Examples
--------
>>> df = pd.DataFrame({'f': pd.np.random.rand(3),
...                    'i': 1,
...                    'd': pd.Timestamp('20180310'),
...                    'o': 'foo'})
>>> df.dtypes
f           float64
i             int64
d    datetime64[ns]
o            object
dtype: object

Errors found:
	No extended summary found
	No returns section found
	See Also section not found


##################### Docstring (pandas.DataFrame.ftypes)  #####################

Return the ftypes (indication of sparse/dense and dtype)
in this object.

Notes
-----
Sparse data should have the same dtypes as its dense representation

See Also
--------
dtypes, SparseDataFrame

Examples
--------
>>> arr = pd.np.random.randn(100, 4)
>>> arr[arr < .8] = pd.np.nan
>>> pd.DataFrame(arr).ftypes
0    float64:dense
1    float64:dense
2    float64:dense
3    float64:dense
dtype: object
>>> pd.SparseDataFrame(arr).ftypes
0    float64:sparse
1    float64:sparse
2    float64:sparse
3    float64:sparse
dtype: object

Errors found:
	No returns section found
	Missing description for See Also "dtypes" reference
	Missing description for See Also "SparseDataFrame" reference

```

explicitly.

Lastly, I left errors already occurred in the previous version without changes.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4275,7 +4275,29 @@ class NDFrame(PandasObject, SelectionMixin):
 
     @property
     def dtypes(self):
-        """Return the dtypes in this object."""
+        """
+        Return the dtypes in this object.
+
+        Notes
+        -----
+        It returns a Series with the data type of each column.
+        If object contains data multiple dtypes in a single column,
+        dtypes will be chosen to accommodate all of the data types.
+        ``object`` is the most general.
+
+        Examples
+        --------
+        >>> df = pd.DataFrame({'f': pd.np.random.rand(3),
+        ...                    'i': 1,
+        ...                    'd': pd.Timestamp('20180310'),
+        ...                    'o': 'foo'})
+        >>> df.dtypes
+        f           float64
+        i             int64
+        d    datetime64[ns]
+        o            object
+        dtype: object
+        """
         from pandas import Series
         return Series(self._data.get_dtypes(), index=self._info_axis,
                       dtype=np.object_)
@@ -4285,6 +4307,31 @@ class NDFrame(PandasObject, SelectionMixin):
         """
         Return the ftypes (indication of sparse/dense and dtype)
         in this object.
+
+        Notes
+        -----
+        Sparse data should have the same dtypes as its dense representation
+
+        See Also
+        --------
+        dtypes, SparseDataFrame
+
+        Examples
+        --------
+        >>> arr = pd.np.random.randn(100, 4)
+        >>> arr[arr < .8] = pd.np.nan
+        >>> pd.DataFrame(arr).ftypes
+        0    float64:dense
+        1    float64:dense
+        2    float64:dense
+        3    float64:dense
+        dtype: object
+        >>> pd.SparseDataFrame(arr).ftypes
+        0    float64:sparse
+        1    float64:sparse
+        2    float64:sparse
+        3    float64:sparse
+        dtype: object
         """
         from pandas import Series
         return Series(self._data.get_ftypes(), index=self._info_axis,
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4280,7 +4280,8 @@ class NDFrame(PandasObject, SelectionMixin):
 
         This returns a Series with the data type of each column.
         The result's index is the original DataFrame's columns. Columns
-        with mixed types are stored in with the ``object`` dtype.
+        with mixed types are stored with the ``object`` dtype. See
+        :ref:`the User Guide <basics.dtypes>` for more.
 
         Returns
         -------
@@ -4315,7 +4316,8 @@ class NDFrame(PandasObject, SelectionMixin):
 
         This returns a Series with the data type of each column.
         The result's index is the original DataFrame's columns. Columns
-        with mixed types are stored in with the ``object`` dtype.
+        with mixed types are stored with the ``object`` dtype.  See
+        :ref:`the User Guide <basics.dtypes>` for more.
 
         Returns
         -------
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4278,17 +4278,25 @@ class NDFrame(PandasObject, SelectionMixin):
         """
         Return the dtypes in this object.
 
-        Notes
-        -----
-        It returns a Series with the data type of each column.
+        This returns a Series with the data type of each column.
+        The result's index is original DataFrame's columns.
         If object contains data multiple dtypes in a single column,
         dtypes will be chosen to accommodate all of the data types.
         ``object`` is the most general.
 
+        Returns
+        -------
+        pandas.Series
+            The data type of each column.
+
+        See Also
+        --------
+        pandas.DataFrame.ftypes
+
         Examples
         --------
-        >>> df = pd.DataFrame({'f': pd.np.random.rand(3),
-        ...                    'i': 1,
+        >>> df = pd.DataFrame({'f': [1.0],
+        ...                    'i': [1],
         ...                    'd': pd.Timestamp('20180310'),
         ...                    'o': 'foo'})
         >>> df.dtypes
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -4275,7 +4275,29 @@ def get_ftype_counts(self):
 
     @property
     def dtypes(self):
-        """Return the dtypes in this object."""
+        """
+        Return the dtypes in this object.
+
+        Notes
```
TomAugspurger: I think you can remove the "Notes" header, and just make this the extended summary.

pandas/core/generic.py
```diff
@@ -4275,7 +4275,29 @@ def get_ftype_counts(self):
...
+        -----
+        It returns a Series with the data type of each column.
```
TomAugspurger: Maybe replace "it" with "This method". And let's say what the values and index is.

    This returns a Series with the data type of each column. The result's index is the
    original DataFrame's columns.

Let's also replace all instances of "object" with "DataFrame". I'm not sure why this is in generic.py since I think it's specific to DataFrame.

pandas/core/generic.py
```diff
@@ -4275,7 +4275,29 @@ def get_ftype_counts(self):
...
+        If object contains data multiple dtypes in a single column,
+        dtypes will be chosen to accommodate all of the data types.
+        ``object`` is the most general.
+
+        Examples
+        --------
+        >>> df = pd.DataFrame({'f': pd.np.random.rand(3),
```
TomAugspurger: Can you just write out three floating point values? I'd like to avoid random data.

And FYI in general you don't want to use `pd.np`. You'll want to import NumPy manually (it's assumed to be imported in our docstrings.)

TomAugspurger: Ah, and could you just make all these length-1 lists, just to be clearer? so `{'f': [1.0], 'i': [1], ...}`

pandas/core/generic.py
```diff
@@ -4285,6 +4307,31 @@ def ftypes(self):
         """
         Return the ftypes (indication of sparse/dense and dtype)
         in this object.
+
+        Notes
+        -----
+        Sparse data should have the same dtypes as its dense representation
```
TomAugspurger: End in a `.`

pandas/core/generic.py
```diff
@@ -4285,6 +4307,31 @@ def ftypes(self):
...
+        --------
+        >>> arr = pd.np.random.randn(100, 4)
+        >>> arr[arr < .8] = pd.np.nan
+        >>> pd.DataFrame(arr).ftypes
+        0    float64:dense
+        1    float64:dense
+        2    float64:dense
+        3    float64:dense
+        dtype: object
+        >>> pd.SparseDataFrame(arr).ftypes
```
TomAugspurger: Maybe a blank line before this to break things up a bit.

pandas/core/generic.py
```diff
@@ -4285,6 +4307,31 @@ def ftypes(self):
...
+        -----
+        Sparse data should have the same dtypes as its dense representation
+
+        See Also
+        --------
+        dtypes, SparseDataFrame
+
+        Examples
+        --------
+        >>> arr = pd.np.random.randn(100, 4)
```
TomAugspurger: `arr = np.random.RandomState(0).randn(100, 4)` for reproducibility

pandas/core/generic.py
```diff
@@ -4275,7 +4275,29 @@ def get_ftype_counts(self):
 
     @property
     def dtypes(self):
-        """Return the dtypes in this object."""
+        """
+        Return the dtypes in this object.
+
```
DataOmbudsman: Can you please add a Returns section as specified in the guide?

pandas/core/generic.py
```diff
@@ -4275,7 +4275,29 @@ def get_ftype_counts(self):
...
+        Notes
+        -----
+        It returns a Series with the data type of each column.
+        If object contains data multiple dtypes in a single column,
+        dtypes will be chosen to accommodate all of the data types.
+        ``object`` is the most general.
+
```
DataOmbudsman: I believe in a See Also section `ftypes` could be mentioned.

pandas/core/generic.py
joaoavf: Create a 'Returns' section to explain the output. Notes should be used only to explain technical details about the implementation of the algorithm or function behavior.

pandas/core/generic.py
```diff
@@ -4285,6 +4307,31 @@ def ftypes(self):
         """
         Return the ftypes (indication of sparse/dense and dtype)
```
DataOmbudsman: The docstring guide asks that the summary should fit in a single line. Could you rephrase it that way? If all information  could not fit in a single line you can then use an Extended Summary section.

pandas/core/generic.py
```diff
@@ -4275,7 +4275,29 @@ def get_ftype_counts(self):
 
     @property
     def dtypes(self):
-        """Return the dtypes in this object."""
+        """
+        Return the dtypes in this object.
```
joaoavf: Explain in more depth to a novice user that this is used to get the dtypes per column of the DataFrame.

pandas/core/generic.py
```diff
@@ -4275,7 +4275,29 @@ def get_ftype_counts(self):
...
+
+        Notes
+        -----
+        It returns a Series with the data type of each column.
+        If object contains data multiple dtypes in a single column,
+        dtypes will be chosen to accommodate all of the data types.
+        ``object`` is the most general.
```
joaoavf: It is worth explaining that `str` will be represented as `object`.

pandas/core/generic.py
```diff
@@ -4285,6 +4307,31 @@ def ftypes(self):
...
         Return the ftypes (indication of sparse/dense and dtype)
         in this object.
+
+        Notes
+        -----
+        Sparse data should have the same dtypes as its dense representation
+
+        See Also
+        --------
+        dtypes, SparseDataFrame
```
DataOmbudsman: Instead of simply `dtypes` you should use `pandas.DataFrame.dtypes`. You can even try linking to dtypes with the notation found in the guide: :meth:\`pandas.Series.sum\`

DataOmbudsman (changes requested):

pandas/core/generic.py
joaoavf: Add a 'See Also' section to contemplate common dtypes.

pandas/core/generic.py
```diff
@@ -4285,6 +4307,31 @@ def ftypes(self):
         """
```
joaoavf: It would better organized it there was pull request for dtypes and another pull request for ftypes.

pandas/core/generic.py
joaoavf: Short summary should have 1 line.

pandas/core/generic.py
```diff
@@ -4285,6 +4307,31 @@ def ftypes(self):
...
         Return the ftypes (indication of sparse/dense and dtype)
         in this object.
+
```
joaoavf: Add a return section.

pandas/core/generic.py
```diff
@@ -4285,6 +4307,31 @@ def ftypes(self):
...
+        Notes
+        -----
+        Sparse data should have the same dtypes as its dense representation
+
+        See Also
```
joaoavf: 'See Also' should go before notes.

pandas/core/generic.py
joaoavf: Make a short description for the itens cited in the summary.

joaoavf (commented):
Found some points that could be improved. Good sprint for you all!

jreback:
is this overlapping with #20099 ?

pandas/core/generic.py
```diff
@@ -4275,7 +4275,29 @@ def get_ftype_counts(self):
...
+        If object contains data multiple dtypes in a single column,
+        dtypes will be chosen to accommodate all of the data types.
+        ``object`` is the most general.
+
+        Examples
+        --------
+        >>> df = pd.DataFrame({'f': pd.np.random.rand(3),
+        ...                    'i': 1,
+        ...                    'd': pd.Timestamp('20180310'),
+        ...                    'o': 'foo'})
```
jreback: add in See Also to Series.dtype

pandas/core/generic.py
lastone9182: I'm modifying dtype `@property` of the NDFrame in generic.py
Is "This property" okay instead of "This method"?

At first, ``object`` was ``NDFrame``, but found an error.

```
```

but it seems to be better that the way you specify with "DataFrame", I will do that.

PR #20103, opened 2018-03-10
Doc: Update the DatetimeIndex.strftime docstring

Tushar987:
an unrelated PR):

```
################## Docstring (pandas.DatetimeIndex.strftime)  ##################

Convert to specified date_format.

Return an array of formatted strings specified by date_format, which
supports the same string format as the python standard library. Details
of the string format can be found in `python string format doc <https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior>`__

Parameters
----------
date_format : str
    Date format string (e.g. "%Y-%m-%d").

Returns
-------
numpy.ndarray
    n-dimensional array of formatted strings

See Also
--------
DatetimeIndex.normalize : Return DatetimeIndex with times to midnight.
DatetimeIndex.round : Round the DatetimeIndex to the specified freq.
DatetimeIndex.floor : Floor the DatetimeIndex to the specified freq.

Examples
--------
>>> data = ['2015-05-01 18:47:05.060000','2014-05-01 18:47:05.110000']
>>> df = pd.DataFrame(data, columns=['date'])
>>> df.date = pd.to_datetime(df.date)
>>> df.date[1]
Timestamp('2014-05-01 18:47:05.110000')
>>> df.date[1].strftime('%d-%m-%Y')
'01-05-2014'
>>> df.date[1].strftime('%B %d, %Y, %r')
'May 01, 2014, 06:47:05 PM'

Docstring for "pandas.DatetimeIndex.strftime" correct. :)
```

explicitly.

### DIFF
--- a/pandas/core/indexes/datetimelike.py
+++ b/pandas/core/indexes/datetimelike.py
@@ -61,6 +61,8 @@ class DatelikeOps(object):
         return np.asarray(self.format(date_format=date_format),
                           dtype=compat.text_type)
     strftime.__doc__ = """
+    Convert to specified date_format.
+
     Return an array of formatted strings specified by date_format, which
     supports the same string format as the python standard library. Details
     of the string format can be found in `python string format doc <{0}>`__
@@ -68,11 +70,30 @@ class DatelikeOps(object):
     Parameters
     ----------
     date_format : str
-        date format string (e.g. "%Y-%m-%d")
+        Date format string (e.g. "%Y-%m-%d").
 
     Returns
     -------
-    ndarray of formatted strings
+    numpy.ndarray
+        n dimensional array of formatted strings
+
+    See Also
+    --------
+    DatetimeIndex.normalize : Return DatetimeIndex with times to midnight.
+    DatetimeIndex.round : Round the DatetimeIndex to the specified freq.
+    DatetimeIndex.floor : Floor the DatetimeIndex to the specified freq.
+
+    Examples
+    --------
+    >>> data = ['2015-05-01 18:47:05.060000','2014-05-01 18:47:05.110000']
+    >>> df = pd.DataFrame(data, columns=['date'])
+    >>> df.date = pd.to_datetime(df.date)
+    >>> df.date[1]
+    Timestamp('2014-05-01 18:47:05.110000')
+    >>> df.date[1].strftime('%d-%m-%Y')
+    '01-05-2014'
+    >>> df.date[1].strftime('%B %d, %Y, %r')
+    'May 01, 2014, 06:47:05 PM'
     """.format("https://docs.python.org/3/library/datetime.html"
                "#strftime-and-strptime-behavior")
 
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/datetimelike.py
```diff
@@ -61,18 +61,39 @@ def strftime(self, date_format):
         return np.asarray(self.format(date_format=date_format),
                           dtype=compat.text_type)
     strftime.__doc__ = """
+    Convert to specified date_format.
```
jorisvandenbossche: Maybe mention it is converted to a string?

pandas/core/indexes/datetimelike.py
```diff
@@ -61,18 +61,39 @@ def strftime(self, date_format):
...
     -------
-    ndarray of formatted strings
+    numpy.ndarray
+        n-dimensional array of formatted strings
+
+    See Also
+    --------
+    DatetimeIndex.normalize : Return DatetimeIndex with times to midnight.
+    DatetimeIndex.round : Round the DatetimeIndex to the specified freq.
+    DatetimeIndex.floor : Floor the DatetimeIndex to the specified freq.
```
jorisvandenbossche: I would also link to the similar conversion but other way around: to_datetime

pandas/core/indexes/datetimelike.py
```diff
@@ -61,18 +61,39 @@ def strftime(self, date_format):
...
+    --------
+    >>> data = ['2015-05-01 18:47:05.060000','2014-05-01 18:47:05.110000']
+    >>> df = pd.DataFrame(data, columns=['date'])
+    >>> df.date = pd.to_datetime(df.date)
+    >>> df.date[1]
+    Timestamp('2014-05-01 18:47:05.110000')
+    >>> df.date[1].strftime('%d-%m-%Y')
+    '01-05-2014'
+    >>> df.date[1].strftime('%B %d, %Y, %r')
+    'May 01, 2014, 06:47:05 PM'
```
jorisvandenbossche: The docstring is from DatetimeIndex. Can you update the example to use that? (so eg create an example dataframe with DatetimeIndex with eg `pd.date_range`)

pandas/core/indexes/datetimelike.py
```diff
@@ -61,18 +61,39 @@ def strftime(self, date_format):
...
     ----------
     date_format : str
-        date format string (e.g. "%Y-%m-%d")
+        Date format string (e.g. "%Y-%m-%d").
 
     Returns
     -------
-    ndarray of formatted strings
+    numpy.ndarray
+        n-dimensional array of formatted strings
```
jorisvandenbossche: I would just say "NumPy array" instead of "n-dimensional", as it is always 1D

jorisvandenbossche (commented):

pandas/core/indexes/datetimelike.py
jreback: @jorisvandenbossche separate issue, this should actually return an ``Index`` here.

jreback (approved):

jreback:
thanks @Tushar987

PR #20104, opened 2018-03-10
DOC: Update pandas.cut docstring

ikoevska:
an unrelated PR):

- [X ] PR title is "DOC: update the <your-function-or-method> docstring"
- [X ] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [X ] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [X ] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [X ] It has been proofread on language by another sprint participant

```
############################ Docstring (pandas.cut) ############################

Bin `x` and return data about the bin to which each `x` value belongs.

Splits `x` into the specified number of equal-width half-open bins.
Based on the parameters specified and the input, returns data about
the half-open bins to which each value of `x` belongs or the bins
themselves.
Use `cut` when you need to segment and sort data values into bins. This
function is also useful for going from a continuous variable to a
categorical variable. For example, `cut` could convert ages to groups
of age ranges.

Parameters
----------
x : array-like
    The input array to be binned. Must be 1-dimensional.
bins : int, sequence of scalars, or pandas.IntervalIndex
    If int, defines the number of equal-width bins in the range of `x`.
    The range of `x` is extended by .1% on each side to include the min or
    max values of `x`.
    If a sequence, defines the bin edges allowing for non-uniform width.
    No extension of the range of `x` is done.
right : bool, default 'True'
    Indicates whether the `bins` include the rightmost edge or not. If
    `right == True` (the default), then the `bins` [1,2,3,4] indicate
    (1,2], (2,3], (3,4].
labels : array or bool, optional
    Specifies the labels for the returned bins. Must be the same length as
    the resulting bins. If False, returns only integer indicators of the
    bins.
retbins : bool, default 'False'
    Whether to return the bins or not. Useful when bins is provided
    as a scalar.
precision : int, default '3'
    The precision at which to store and display the bins labels.
include_lowest : bool, default 'False'

Returns
-------
out : pandas.Categorical, Series, or ndarray
    An array-like object representing the respective bin for each value
    of `x`. The type depends on the value of `labels`.

    * True : returns a Series for Series `x` or a pandas.Categorical for
    pandas.Categorial `x`.

    * False : returns an ndarray of integers.
bins : numpy.ndarray of floats
    Returned when `retbins` is 'True'.

See Also
--------
qcut : Discretize variable into equal-sized buckets based on rank
    or based on sample quantiles.
pandas.Categorical : Represents a categorical variable in
    classic R / S-plus fashion.
Series : One-dimensional ndarray with axis labels (including time series).
pandas.IntervalIndex : Immutable Index implementing an ordered,
    sliceable set. IntervalIndex represents an Index of intervals that
    are all closed on the same side.

Notes
-----
Any NA values will be NA in the result. Out of bounds values will be NA in
the resulting pandas.Categorical object.

Examples
--------
>>> pd.cut(np.array([1,7,5,4,6,3]), 3)
... # doctest: +ELLIPSIS
[(0.994, 3.0], (5.0, 7.0], (3.0, 5.0], (3.0, 5.0], (5.0, 7.0], ...
Categories (3, interval[float64]): [(0.994, 3.0] < (3.0, 5.0] ...

>>> pd.cut(np.array([1,7,5,4,6,3]), 3, retbins=True)
... # doctest: +ELLIPSIS
([(0.994, 3.0], (5.0, 7.0], (3.0, 5.0], (3.0, 5.0], (5.0, 7.0], ...
Categories (3, interval[float64]): [(0.994, 3.0] < (3.0, 5.0] ...
array([0.994, 3.   , 5.   , 7.   ]))

>>> pd.cut(np.array([.2, 1.4, 2.5, 6.2, 9.7, 2.1]),
...        3, labels=["good", "medium", "bad"])
... # doctest: +SKIP
[good, good, good, medium, bad, good]
Categories (3, object): [good < medium < bad]

>>> pd.cut(np.ones(5, dtype='int64'), 4, labels=False)
array([1, 1, 1, 1, 1], dtype=int64)

>>> s = pd.Series(np.array([2,4,6,8,10]), index=['a', 'b', 'c', 'd', 'e'])
>>> pd.cut(s, 3)
... # doctest: +ELLIPSIS
a    (1.992, 4.667]
b    (1.992, 4.667]
c    (4.667, 7.333]
d     (7.333, 10.0]
e     (7.333, 10.0]
dtype: category
Categories (3, interval[float64]): [(1.992, 4.667] < (4.667, ...

Docstring for "pandas.cut" correct. :)
```

Comment: Resubmitting https://github.com/pandas-dev/pandas/pull/20069 after a botched rebase.

### DIFF
--- a/pandas/core/reshape/tile.py
+++ b/pandas/core/reshape/tile.py
@@ -26,37 +26,41 @@ import numpy as np
 def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
         include_lowest=False):
     """
-    Return indices of half-open `bins` to which each value of `x` belongs.
+    Bin `x` and return data about the bin to which each `x` value belongs.
 
-    Use `cut` when you need to segment and sort data values into bins or
-    buckets of data. This function is also useful for going from a continuous
-    variable to a categorical variable. For example, `cut` could convert ages
-    to groups of age ranges.
+    This function splits `x` into the specified number of equal-width half-
+    open bins. Based on the parameters specified and the input, returns
+    information about the half-open bins to which each value of `x` belongs
+    or the bins themselves.
+    Use `cut` when you need to segment and sort data values into bins. This
+    function is also useful for going from a continuous variable to a
+    categorical variable. For example, `cut` could convert ages to groups
+    of age ranges.
 
     Parameters
     ----------
     x : array-like
-        Input array to be binned. It has to be 1-dimensional.
+        The input array to be binned. Must be 1-dimensional.
     bins : int, sequence of scalars, or pandas.IntervalIndex
-        If `bins` is an int, defines the number of equal-width bins in the
-        range of `x`. The range of `x` is extended by .1% on each side to
-        include the min or max values of `x`.
-        If `bins` is a sequence, defines the bin edges allowing for
-        non-uniform bin width. No extension of the range of `x` is done.
-    right : bool, optional, default 'True'
+        If int, defines the number of equal-width bins in the range of `x`.
+        The range of `x` is extended by .1% on each side to include the min or
+        max values of `x`.
+        If a sequence, defines the bin edges allowing for non-uniform width.
+        No extension of the range of `x` is done.
+    right : bool, default 'True'
         Indicates whether the `bins` include the rightmost edge or not. If
         `right == True` (the default), then the `bins` [1,2,3,4] indicate
         (1,2], (2,3], (3,4].
     labels : array or bool, optional
-        Used as labels for the resulting `bins`. Must be of the same length as
-        the resulting `bins`. If False, returns only integer indicators of the
-        `bins`.
-    retbins : bool, optional, default 'False'
-        Whether to return the `bins` or not. Useful when `bins` is provided
+        Specifies the labels for the returned bins. Must be the same length as
+        the resulting bins. If False, returns only integer indicators of the
+        bins.
+    retbins : bool, default 'False'
+        Whether to return the bins or not. Useful when bins is provided
         as a scalar.
-    precision : int, optional, default '3'
-        The precision at which to store and display the `bins` labels.
-    include_lowest : bool, optional, default 'False'
+    precision : int, default '3'
+        The precision at which to store and display the bins labels.
+    include_lowest : bool, default 'False'
         Whether the first interval should be left-inclusive or not.
 
     Returns
@@ -64,10 +68,10 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
     out : pandas.Categorical or Series, or array of int if `labels` is 'False'
         The return type depends on the input.
         If the input is a Series, a Series of type category is returned.
-        Else - pandas.Categorical is returned. `Bins` are represented as
+        Else - pandas.Categorical is returned. Bins are represented as
         categories when categorical data is returned.
     bins : numpy.ndarray of floats
-        Returned only if `retbins` is 'True'.
+        Returned when `retbins` is 'True'.
 
     See Also
     --------
@@ -87,10 +91,16 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
 
     Examples
     --------
-    >>> pd.cut(np.array([.2, 1.4, 2.5, 6.2, 9.7, 2.1]), 3, retbins=True)
+    >>> pd.cut(np.array([1,7,5,4,6,3]), 3)
     ... # doctest: +ELLIPSIS
-    ([(0.19, 3.367], (0.19, 3.367], (0.19, 3.367], (3.367, 6.533], ...
-    Categories (3, interval[float64]): [(0.19, 3.367] < (3.367, 6.533] ...
+    [(0.994, 3.0], (5.0, 7.0], (3.0, 5.0], (3.0, 5.0], (5.0, 7.0], ...
+    Categories (3, interval[float64]): [(0.994, 3.0] < (3.0, 5.0] ...
+
+    >>> pd.cut(np.array([1,7,5,4,6,3]), 3, retbins=True)
+    ... # doctest: +ELLIPSIS
+    ([(0.994, 3.0], (5.0, 7.0], (3.0, 5.0], (3.0, 5.0], (5.0, 7.0], ...
+    Categories (3, interval[float64]): [(0.994, 3.0] < (3.0, 5.0] ...
+    array([0.994, 3.   , 5.   , 7.   ]))
 
     >>> pd.cut(np.array([.2, 1.4, 2.5, 6.2, 9.7, 2.1]),
     ...        3, labels=["good", "medium", "bad"])
@@ -100,6 +110,17 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
 
     >>> pd.cut(np.ones(5), 4, labels=False)
     array([1, 1, 1, 1, 1], dtype=int64)
+
+    >>> s = pd.Series(np.array([2,4,6,8,10]), index=['a', 'b', 'c', 'd', 'e'])
+    >>> pd.cut(s, 3)
+    ... # doctest: +ELLIPSIS
+    a    (1.992, 4.667]
+    b    (1.992, 4.667]
+    c    (4.667, 7.333]
+    d     (7.333, 10.0]
+    e     (7.333, 10.0]
+    dtype: category
+    Categories (3, interval[float64]): [(1.992, 4.667] < (4.667, ...
     """
     # NOTE: this binning code is changed a bit from histogram for var(x) == 0
 
--- a/pandas/core/reshape/tile.py
+++ b/pandas/core/reshape/tile.py
@@ -65,11 +65,14 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
 
     Returns
     -------
-    out : pandas.Categorical or Series, or array of int if `labels` is 'False'
-        The return type depends on the input.
-        If the input is a Series, a Series of type category is returned.
-        Else - pandas.Categorical is returned. Bins are represented as
-        categories when categorical data is returned.
+    out : pandas.Categorical, Series, or ndarray
+        An array-like object representing the respective bin for each value
+        of `x`. The type depends on the value of `labels`.
+
+        * True : returns a Series for Series `x` or a pandas.Categorical for
+        pandas.Categorial `x`.
+        
+        * False : returns an ndarray of integers.
     bins : numpy.ndarray of floats
         Returned when `retbins` is 'True'.
 
--- a/pandas/core/reshape/tile.py
+++ b/pandas/core/reshape/tile.py
@@ -28,10 +28,10 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
     """
     Bin `x` and return data about the bin to which each `x` value belongs.
 
-    This function splits `x` into the specified number of equal-width half-
-    open bins. Based on the parameters specified and the input, returns
-    information about the half-open bins to which each value of `x` belongs
-    or the bins themselves.
+    Splits `x` into the specified number of equal-width half-open bins.
+    Based on the parameters specified and the input, returns data about
+    the half-open bins to which each value of `x` belongs or the bins
+    themselves.
     Use `cut` when you need to segment and sort data values into bins. This
     function is also useful for going from a continuous variable to a
     categorical variable. For example, `cut` could convert ages to groups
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Did you already update according to my comments?

ikoevska:
Nope, updating right now. :)

ikoevska:
@jorisvandenbossche Updated based on your comments and added some more examples. I've also updated the validation script output in the description of the PR.

pandas/core/reshape/tile.py
```diff
@@ -26,60 +26,81 @@
 def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
         include_lowest=False):
     """
-    Return indices of half-open bins to which each value of `x` belongs.
+    Bin `x` and return data about the bin to which each `x` value belongs.
+
+    This function splits `x` into the specified number of equal-width half-
```
jreback: can remove function.

you say 'data' above and use 'information' here. pick one.

pandas/core/reshape/tile.py
```diff
@@ -26,60 +26,81 @@
...
+    open bins. Based on the parameters specified and the input, returns
+    information about the half-open bins to which each value of `x` belongs
+    or the bins themselves.
```
jreback: if ``retbins=True``

pandas/core/reshape/tile.py
```diff
@@ -26,60 +26,81 @@
...
     -------
-    out : Categorical or Series or array of integers if labels is False
-        The return type (Categorical or Series) depends on the input: a Series
-        of type category if input is a Series else Categorical. Bins are
-        represented as categories when categorical data is returned.
-    bins : ndarray of floats
-        Returned only if `retbins` is True.
+    out : pandas.Categorical or Series, or array of int if `labels` is 'False'
+        The return type depends on the input.
+        If the input is a Series, a Series of type category is returned.
```
jreback: not sure about the formatting here, @TomAugspurger ?

pandas/core/reshape/tile.py
```diff
@@ -26,60 +26,81 @@
...
+    bins : numpy.ndarray of floats
+        Returned when `retbins` is 'True'.
+
+    See Also
+    --------
+    qcut : Discretize variable into equal-sized buckets based on rank
+        or based on sample quantiles.
+    pandas.Categorical : Represents a categorical variable in
+        classic R / S-plus fashion.
+    Series : One-dimensional ndarray with axis labels (including time series).
```
jreback: I dont' think Series is needed here

pandas/core/reshape/tile.py
```diff
@@ -88,7 +109,18 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
     Categories (3, object): [good < medium < bad]
 
     >>> pd.cut(np.ones(5), 4, labels=False)
```
jreback: specify ``dtype='int64'`` to np.ones (this is normally a platform int)

jreback (changes requested):

pandas/core/reshape/tile.py
ikoevska: I disagree with this suggested change. Currently, the complete sentence begins with "Based on the parameters specified and the input" which points to looking at parameters and other information below. If I add "if retbins=True", I will be pressed to explain all other options in the extended description as well.

pandas/core/reshape/tile.py
ikoevska: As this is something that you can input or get as an output from the command, I think this needs to be here. There is also an example with series below.

pandas/core/reshape/tile.py
ikoevska: @TomAugspurger, still waiting for more feedback on that. Thanks!

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,101 @@
...
 
     Returns
     -------
-    out : Categorical or Series or array of integers if labels is False
-        The return type (Categorical or Series) depends on the input: a Series
-        of type category if input is a Series else Categorical. Bins are
-        represented as categories when categorical data is returned.
-    bins : ndarray of floats
-        Returned only if `retbins` is True.
+    out : pandas.Categorical or Series, or array of int if `labels` is 'False'
```
TomAugspurger: I think something liek

```
out : pandas.Categorical, Series, or ndarray
    An array-like object representing the bin for each value of `x`.
    The type depends on the value of `labels`.

    * True : returns a Series for Series `x` or a Categorical for Categorical `x`.
    * False : returns an ndarray of integers.
```

TomAugspurger:
Looks like some git issues @ikoevska. LMK if you need some help.

TomAugspurger:
Did you have any changes in https://github.com/pandas-dev/pandas/pull/20104/commits/2dffb608b3949caf1b9d4a056c325deac6818390? If not, I think

```
git reset --hard d24c749b0 
git merge upstream/master
```

should do it. It'll throw away changes from that commit though.

ikoevska:
Yep, I did a sync with the upstream and then a rebase on master and this happened. Suggestions how to fix it?

ikoevska:
@jreback Can you take a look at the latest changes and my comments? Thanks!

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,104 @@
...
         bins.
-    retbins : bool, optional
-        Whether to return the bins or not. Can be useful if bins is given
+    retbins : bool, default 'False'
+        Whether to return the bins or not. Useful when bins is provided
         as a scalar.
-    precision : int, optional
-        The precision at which to store and display the bins labels
-    include_lowest : bool, optional
+    precision : int, default '3'
```
TomAugspurger: No quotes around `'3'`. Quotes makes it look like a string.

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,104 @@
...
-        this case.
-    right : bool, optional
-        Indicates whether the bins include the rightmost edge or not. If
-        right == True (the default), then the bins [1,2,3,4] indicate
+        The input array to be binned. Must be 1-dimensional.
+    bins : int, sequence of scalars, or pandas.IntervalIndex
+        If int, defines the number of equal-width bins in the range of `x`.
+        The range of `x` is extended by .1% on each side to include the min or
+        max values of `x`.
+        If a sequence, defines the bin edges allowing for non-uniform width.
```
TomAugspurger: Could format these as

```
* int : the number of equal-...
* sequence : integers defining the bin edges (refer to \`right\` parameter.)
* IntervalIndex : sequence of Intervals to use for binning...
```

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,104 @@
...
-        Whether to return the bins or not. Can be useful if bins is given
+    retbins : bool, default 'False'
+        Whether to return the bins or not. Useful when bins is provided
         as a scalar.
-    precision : int, optional
-        The precision at which to store and display the bins labels
-    include_lowest : bool, optional
+    precision : int, default '3'
+        The precision at which to store and display the bins labels.
+    include_lowest : bool, default 'False'
```
TomAugspurger: I think a single backtick around `False`, or just False. Not sure.

pandas/core/reshape/tile.py
TomAugspurger: What happens with `right` and `IntervalIndex`? Is it ignored?

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,104 @@
...
 
+    Notes
+    -----
+    Any NA values will be NA in the result. Out of bounds values will be NA in
+    the resulting pandas.Categorical object.
 
     Examples
     --------
-    >>> pd.cut(np.array([.2, 1.4, 2.5, 6.2, 9.7, 2.1]), 3, retbins=True)
+    >>> pd.cut(np.array([1,7,5,4,6,3]), 3)
```
TomAugspurger: PEP8 on all these (spaces after `,`)

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,104 @@
...
-        Indicates whether the bins include the rightmost edge or not. If
-        right == True (the default), then the bins [1,2,3,4] indicate
+        The input array to be binned. Must be 1-dimensional.
+    bins : int, sequence of scalars, or pandas.IntervalIndex
+        If int, defines the number of equal-width bins in the range of `x`.
+        The range of `x` is extended by .1% on each side to include the min or
+        max values of `x`.
+        If a sequence, defines the bin edges allowing for non-uniform width.
+        No extension of the range of `x` is done.
+    right : bool, default 'True'
```
jorisvandenbossche: similar comment here for True -> no quotes (makes it look like a string)

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,104 @@
...
-    labels : array or boolean, default None
-        Used as labels for the resulting bins. Must be of the same length as
-        the resulting bins. If False, return only integer indicators of the
+    labels : array or bool, optional
+        Specifies the labels for the returned bins. Must be the same length as
+        the resulting bins. If False, returns only integer indicators of the
         bins.
-    retbins : bool, optional
-        Whether to return the bins or not. Can be useful if bins is given
+    retbins : bool, default 'False'
```
jorisvandenbossche: similar, no quotes

pandas/core/reshape/tile.py
jorisvandenbossche: No quoting or backticks in this case

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,104 @@
...
+        An array-like object representing the respective bin for each value
+        of `x`. The type depends on the value of `labels`.
 
-    Notes
-    -----
-    The `cut` function can be useful for going from a continuous variable to
-    a categorical variable. For example, `cut` could convert ages to groups
-    of age ranges.
+        * True : returns a Series for Series `x` or a pandas.Categorical for
+        pandas.Categorial `x`.
```
jorisvandenbossche: I think a categorical is returned for any array-like that is not a Series (not only for Categorical) ?

jorisvandenbossche: Also, is it worth mentioning that it is a Categorical of Intervals?

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,104 @@
...
+
+        * False : returns an ndarray of integers.
+    bins : numpy.ndarray of floats
+        Returned when `retbins` is 'True'.
 
-    Any NA values will be NA in the result.  Out of bounds values will be NA in
-    the resulting Categorical object
+    See Also
+    --------
+    qcut : Discretize variable into equal-sized buckets based on rank
```
jorisvandenbossche: I suppose for qcut the bins will not be equal sized given it is based on quantiles?

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,104 @@
...
+        or based on sample quantiles.
+    pandas.Categorical : Represents a categorical variable in
+        classic R / S-plus fashion.
```
jorisvandenbossche: You copied this from Categorical, so that is fine, but I think we should really change this explanation to something not referring to R, many of our users are learning pandas without knowing R :)

pandas/core/reshape/tile.py
```diff
@@ -26,69 +26,104 @@
 def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
         include_lowest=False):
     """
-    Return indices of half-open bins to which each value of `x` belongs.
+    Bin `x` and return data about the bin to which each `x` value belongs.
```
jorisvandenbossche: I find "return data about the bin" a bit "wordy" without saying much. 
I suppose you want to be general because the return type can be either the bins, or the indices indexing into the bins? But I won't care about this: the summary line should be short and give an idea of the main use case. The extended summary can go more into detail (as you already do).

Starting from the above, I would cut that to the essential and just "Bin x", but that is maybe too short :-) 
What about "Convert continuous values into discrete bins" or "Discretize values in specified bins" or "Bin values in specified intervals", but not sure what is the best wording. 
@TomAugspurger any input?

jorisvandenbossche (commented):
Can you put some explanation between the different examples? (shortly explaining what the following example is doing, or how it is different from the previous one, to give some context for the reader looking at those examples)

pandas/core/reshape/tile.py
TomAugspurger: For series input it's a Series with categorical dtype.

TomAugspurger: And yes, saying Categorical of Intervals somewhere is good. NOt sure if here is best though.

pandas/core/reshape/tile.py
TomAugspurger: Yes, this should be simplified. "Bin values into discrete intervals"?

I don't 100% like "in specified intervals" as you don't have to specify the intervals.

jorisvandenbossche: > "Bin values into discrete intervals"?

\+1, perfect combination of all my attempts :-)

TomAugspurger (approved):
Updated, if anyone wants to take a look.

jorisvandenbossche (approved):

PR #20112, opened 2018-03-10
DOC: update validate_docstrings.py Validation script checks for tabs

mariocj89:
Add a check to validate the docstrings don't have tabs.
The documentation uses whitespace only, adding the check will prevent
tabs being added in the sprint or future submissions

Sample output (elided non changed parts)
```
...

Errors found:
	...
	Tabs found in the docstring, please use whitespace only
	...

```

### DIFF
--- a/scripts/validate_docstrings.py
+++ b/scripts/validate_docstrings.py
@@ -455,6 +455,12 @@ def validate_one(func_name):
             if not rel_desc:
                 errs.append('Missing description for '
                             'See Also "{}" reference'.format(rel_name))
+
+    for line in doc.raw_doc.splitlines():
+        if re.match("^ *\t", line):
+            errs.append('Tabs found at the start of line "{}", '
+                        'please use whitespace only'.format(line.lstrip()))
+
     examples_errs = ''
     if not doc.examples:
         errs.append('No examples section found')


scripts/validate_docstrings.py
```diff
@@ -455,6 +455,10 @@ def validate_one(func_name):
             if not rel_desc:
                 errs.append('Missing description for '
                             'See Also "{}" reference'.format(rel_name))
+
+    if "\t" in doc.raw_doc:
```
TomAugspurger: Can you make this check for tabs only at the start of lines? We may have cases where we include the literal `\t` in the docstring (e.g. `read_csv`).

scripts/validate_docstrings.py
```diff
@@ -455,6 +455,12 @@ def validate_one(func_name):
             if not rel_desc:
                 errs.append('Missing description for '
                             'See Also "{}" reference'.format(rel_name))
+
+    for line in doc.raw_doc.splitlines():
+        if re.match(" *\t", line):
```
TomAugspurger: I think the regex would be
    `if re.match(r"^\t", line)`

mariocj89: Indeed, missing the start of the line. Sorry about that. I was checking as well there is no spaces followed by tabs. Is it something we should check or just a plain "starts with a tab"

TomAugspurger: Ahh, I see. So maybe `^ *\t`? That's lines starting with 0 or more spaces followed by a tab.

mariocj89: Do you want me to add a test file for this script? I can create a tests folder with a test for this script that will 1) prevent future mistakes and would have made easier to review a PR like this one. What do you think?

mariocj89: Actually, There is already another PR adding the test file :)

TomAugspurger:
@datapythonista @jorisvandenbossche if you could glance over this. It looks good to me.

scripts/validate_docstrings.py
```diff
@@ -455,6 +455,12 @@ def validate_one(func_name):
             if not rel_desc:
                 errs.append('Missing description for '
                             'See Also "{}" reference'.format(rel_name))
+
+    for line in doc.raw_doc.splitlines():
+        if re.match("^ *\t", line):
```
jreback: '\t' in line is good enough, yes? we don't all *any* tabs

mariocj89: If you see the comments above @TomAugspurger requested "Starts with tab" as there might be tabs to format a table or the like I think

jreback: ok, not sure this is actually strict enough. as I said we don't allow *any* tabs

mariocj89: I'm happy to change it back to allow no tabs at all. Let me know what you think @TomAugspurger

jorisvandenbossche: @jreback There are cases where a `\t` is included in the text as literal character (for the explanation).
So I think only checking the start of a line is good.

jorisvandenbossche (approved):

scripts/validate_docstrings.py
datapythonista: Unrelated to where `\t` is included, we could use `re.search(r'^ *\t', self.raw_doc, flags=re.MULTILINE)` and avoid the loop

datapythonista: ups, sorry, I see now that you want to report the lines with the tabs...

jorisvandenbossche:
@mariocj89 thanks!

mariocj89:
Thank you!

PR #20113, opened 2018-03-10
DOC: update the pandas.DataFrame.hist docstring

DZPM:
- [x] PR title is "DOC: update the pandas.DataFrame.hist docstring"
- [x] The validation script passes: `scripts/validate_docstrings.py pandas.DataFrame.hist`
- [x] The html version looks good: `python doc/make.py --single pandas.DataFrame.hist`

```
###################### Docstring (pandas.DataFrame.hist)  ######################

Draw histogram of the DataFrame's series using matplotlib / pylab.

A histogram is a representation of the distribution of numerical data.
This function wraps the matplotlib histogram function for each serie in
the DataFrame. It returns an array with a plot for each histogram.

Parameters
----------
data : DataFrame
    The pandas object holding the data.
column : string or sequence
    If passed, will be used to limit data to a subset of columns.
by : object, optional
    If passed, then used to form histograms for separate groups.
grid : boolean, default True
    Whether to show axis grid lines.
xlabelsize : int, default None
    If specified changes the x-axis label size.
xrot : float, default None
    Rotation of x axis labels.
ylabelsize : int, default None
    If specified changes the y-axis label size.
yrot : float, default None
    Rotation of y axis labels.
ax : Matplotlib axes object, default None
    The histogram axes.
sharex : boolean, default True if ax is None else False
    In case subplots=True, share x axis and set some x axis labels to
    invisible; defaults to True if ax is None otherwise False if an ax
    is passed in; Be aware, that passing in both an ax and sharex=True
    will alter all x axis labels for all subplots in a figure!.
sharey : boolean, default False
    In case subplots=True, share y axis and set some y axis labels to
    invisible.
figsize : tuple
    The size of the figure to create in inches by default.
layout : tuple, optional
    Tuple of (rows, columns) for the layout of the histograms.
bins : integer or sequence, default 10
    Number of histogram bins to be used. If an integer is given, bins + 1
    bin edges are calculated and returned. If bins is a sequence, gives
    bin edges, including left edge of first bin and right edge of last
    bin. In this case, bins is returned unmodified.
kwds : Keyword Arguments
    All other plotting keyword arguments to be passed to
    matplotlib's boxplot function.

Returns
-------

See Also
--------
matplotlib.axes.Axes.hist : Plot a histogram using matplotlib.

Examples
--------

.. plot::
    :context: close-figs

    >>> df = pd.DataFrame({
    ...     'length': [ 1.5, 0.5, 1.2, 0.9, 3],
    ...     'width': [ 0.7, 0.2, 0.15, 0.2,  1.1]
    ...     }, index= ['pig', 'rabbit', 'duck', 'chicken', 'horse'])
    >>> hist = df.hist(bins=3)

Docstring for "pandas.DataFrame.hist" correct. :)
```

### DIFF
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2128,50 +2128,79 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
                xrot=None, ylabelsize=None, yrot=None, ax=None, sharex=False,
                sharey=False, figsize=None, layout=None, bins=10, **kwds):
     """
-    Draw histogram of the DataFrame's series using matplotlib / pylab.
+    Draw histogram of the DataFrame's columns using matplotlib.
+
+    A histogram is a representation of the distribution of data.
+    This function wraps the matplotlib histogram function for each series in
+    the DataFrame. It returns an array with a plot for each histogram.
 
     Parameters
     ----------
     data : DataFrame
+        The pandas object holding the data.
     column : string or sequence
-        If passed, will be used to limit data to a subset of columns
+        If passed, will be used to limit data to a subset of columns.
     by : object, optional
-        If passed, then used to form histograms for separate groups
+        If passed, then used to form histograms for separate groups.
     grid : boolean, default True
-        Whether to show axis grid lines
+        Whether to show axis grid lines.
     xlabelsize : int, default None
-        If specified changes the x-axis label size
+        If specified changes the x-axis label size.
     xrot : float, default None
-        rotation of x axis labels
+        Rotation of x axis labels. For example, a value of 90 displays the
+        x labels rotated 90 degrees clockwise.
     ylabelsize : int, default None
-        If specified changes the y-axis label size
+        If specified changes the y-axis label size.
     yrot : float, default None
-        rotation of y axis labels
-    ax : matplotlib axes object, default None
+        Rotation of y axis labels. For example, a value of 90 displays the
+        y labels rotated 90 degrees clockwise.
+    ax : Matplotlib axes object, default None
+        The axes to plot the histogram on.
     sharex : boolean, default True if ax is None else False
         In case subplots=True, share x axis and set some x axis labels to
         invisible; defaults to True if ax is None otherwise False if an ax
-        is passed in; Be aware, that passing in both an ax and sharex=True
-        will alter all x axis labels for all subplots in a figure!
+        is passed in.
+        Note that passing in both an ax and sharex=True will alter all x axis
+        labels for all subplots in a figure.
     sharey : boolean, default False
         In case subplots=True, share y axis and set some y axis labels to
-        invisible
+        invisible.
     figsize : tuple
-        The size of the figure to create in inches by default
+        The size in inches of the figure to create. Uses the value in
+        `matplotlib.rcParams` by default.
     layout : tuple, optional
-        Tuple of (rows, columns) for the layout of the histograms
+        Tuple of (rows, columns) for the layout of the histograms.
     bins : integer or sequence, default 10
         Number of histogram bins to be used. If an integer is given, bins + 1
         bin edges are calculated and returned. If bins is a sequence, gives
         bin edges, including left edge of first bin and right edge of last
         bin. In this case, bins is returned unmodified.
-    `**kwds` : other plotting keyword arguments
-        To be passed to hist function
+    kwds : optional
+        All other plotting keyword arguments to be passed to
+        matplotlib's boxplot function.
+
+    Returns
+    -------
+    axes : matplotlib.AxesSubplot or numpy.ndarray of them
 
     See Also
     --------
     matplotlib.axes.Axes.hist : Plot a histogram using matplotlib.
 
+    Examples
+    --------
+
+    .. plot::
+        :context: close-figs
+
+        This example draws a histogram based on the length and width of
+        some animals, displayed in three bins
+
+        >>> df = pd.DataFrame({
+        ...     'length': [1.5, 0.5, 1.2, 0.9, 3],
+        ...     'width': [0.7, 0.2, 0.15, 0.2, 1.1]
+        ...     }, index= ['pig', 'rabbit', 'duck', 'chicken', 'horse'])
+        >>> hist = df.hist(bins=3)
     """
     _converter._WARN = False
     if by is not None:
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2128,11 +2128,13 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
                xrot=None, ylabelsize=None, yrot=None, ax=None, sharex=False,
                sharey=False, figsize=None, layout=None, bins=10, **kwds):
     """
-    Draw histogram of the DataFrame's columns using matplotlib.
+    Make a histogram of the DataFrame's.
 
-    A histogram is a representation of the distribution of data.
-    This function wraps the matplotlib histogram function for each series in
-    the DataFrame. It returns an array with a plot for each histogram.
+    A `histogram`_ is a representation of the distribution of data.
+    This function calls :meth:`matplotlib.pyplot.hist`, on each series in
+    the DataFrame, resulting in one histogram per column.
+
+    .. _histogram: https://en.wikipedia.org/wiki/Histogram
 
     Parameters
     ----------
@@ -2175,9 +2177,9 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
         bin edges are calculated and returned. If bins is a sequence, gives
         bin edges, including left edge of first bin and right edge of last
         bin. In this case, bins is returned unmodified.
-    kwds : optional
+    **kwds
         All other plotting keyword arguments to be passed to
-        matplotlib's boxplot function.
+        :meth:`matplotlib.pyplot.hist`.
 
     Returns
     -------
@@ -2185,7 +2187,7 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
 
     See Also
     --------
-    matplotlib.axes.Axes.hist : Plot a histogram using matplotlib.
+    matplotlib.pyplot.hist : Plot a histogram using matplotlib.
 
     Examples
     --------


pandas/plotting/_core.py
```diff
@@ -2130,48 +2130,70 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
     """
     Draw histogram of the DataFrame's series using matplotlib / pylab.
```
TomAugspurger: Can you remove the `/ pylab`. It's been deprecated for ~5 years now :)

pandas/plotting/_core.py
```diff
@@ -2130,48 +2130,70 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
 
+    A histogram is a representation of the distribution of numerical data.
```
TomAugspurger: Strike numerical as hist can handle non-numeric data as well.

pandas/plotting/_core.py
```diff
@@ -2130,48 +2130,70 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
+    This function wraps the matplotlib histogram function for each serie in
```
TomAugspurger: serie -> series.

pandas/plotting/_core.py
```diff
@@ -2130,48 +2130,70 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
+        Rotation of x axis labels.
     ylabelsize : int, default None
-        If specified changes the y-axis label size
+        If specified changes the y-axis label size.
     yrot : float, default None
-        rotation of y axis labels
-    ax : matplotlib axes object, default None
+        Rotation of y axis labels.
+    ax : Matplotlib axes object, default None
+        The histogram axes.
```
TomAugspurger: "The axes to plot the histogram on."

pandas/plotting/_core.py
```diff
@@ -2130,48 +2130,70 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
     sharex : boolean, default True if ax is None else False
         In case subplots=True, share x axis and set some x axis labels to
         invisible; defaults to True if ax is None otherwise False if an ax
         is passed in; Be aware, that passing in both an ax and sharex=True
-        will alter all x axis labels for all subplots in a figure!
+        will alter all x axis labels for all subplots in a figure!.
```
TomAugspurger: Can remove the `.` since the period already ends in a punctuation character. Did the script complain about ending with a `!`?

pandas/plotting/_core.py
```diff
@@ -2130,48 +2130,70 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
     sharey : boolean, default False
         In case subplots=True, share y axis and set some y axis labels to
-        invisible
+        invisible.
     figsize : tuple
-        The size of the figure to create in inches by default
+        The size of the figure to create in inches by default.
```
TomAugspurger: How about

    The size in inches of the figure to create. Uses the value in `matplotlib.rcParams` by default.

pandas/plotting/_core.py
```diff
@@ -2130,48 +2130,70 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
-        Tuple of (rows, columns) for the layout of the histograms
+        Tuple of (rows, columns) for the layout of the histograms.
     bins : integer or sequence, default 10
         Number of histogram bins to be used. If an integer is given, bins + 1
         bin edges are calculated and returned. If bins is a sequence, gives
         bin edges, including left edge of first bin and right edge of last
         bin. In this case, bins is returned unmodified.
-    `**kwds` : other plotting keyword arguments
-        To be passed to hist function
+    kwds : Keyword Arguments
```
TomAugspurger: I may be wrong, but I think we're just doing these as

    kwds:
        All other plotting ...

IOW no type.

pandas/plotting/_core.py
```diff
@@ -2130,48 +2130,70 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
+        All other plotting keyword arguments to be passed to
+        matplotlib's boxplot function.
+
+    Returns
+    -------
+    axes : matplotlib.AxesSubplot or np.array of them
```
TomAugspurger: `np.array` -> `numpy.ndarray`

(np.array is the function, np.ndarray is the object).

pandas/plotting/_core.py
DataOmbudsman: I think series should be written as Series, it's more common in the docstrings.

pandas/plotting/_core.py
DZPM: Yes, the validator script complains:

	Errors in parameters section
		Parameter "sharex" description should finish with "."

I'm rewriting it to remove the `!`.

pandas/plotting/_core.py
DZPM: That raises some errors in the validator:

    Errors in parameters section
		Unknown parameters {'kwds :'}
		Parameter "kwds :" has no type

DZPM:
Fixed all isues (except the `kwargs`, as it'd fail the validation), and ammended the commit.

Thanks for the review!

pandas/plotting/_core.py
dukebody: Write `kwds : optional`.

The validation script is wrong @datapythonista team is fixing it from London. Use always `**kwargs` as parameter.

pandas/plotting/_core.py
```diff
@@ -2128,50 +2128,74 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
                xrot=None, ylabelsize=None, yrot=None, ax=None, sharex=False,
                sharey=False, figsize=None, layout=None, bins=10, **kwds):
     """
-    Draw histogram of the DataFrame's series using matplotlib / pylab.
+    Draw histogram of the DataFrame's Series using matplotlib.
```
dukebody: Perhaps this is written like this in other parts of the docs, but I personally find it easier to understand to say something like `Draw histogram of the DataFrame's columns`, instead of `DataFrame's Series`.

pandas/plotting/_core.py
```diff
@@ -2128,50 +2128,74 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
+        If passed, then used to form histograms for separate groups.
     grid : boolean, default True
-        Whether to show axis grid lines
+        Whether to show axis grid lines.
     xlabelsize : int, default None
-        If specified changes the x-axis label size
+        If specified changes the x-axis label size.
     xrot : float, default None
-        rotation of x axis labels
+        Rotation of x axis labels.
```
dukebody: Can we add some explanation like "For example, a value of `90` displays the x labels rotated 90º clockwise".

pandas/plotting/_core.py
```diff
@@ -2128,50 +2128,74 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
     ylabelsize : int, default None
-        If specified changes the y-axis label size
+        If specified changes the y-axis label size.
     yrot : float, default None
-        rotation of y axis labels
-    ax : matplotlib axes object, default None
+        Rotation of y axis labels.
```
dukebody: Same here, make clear what a value of 90 means, for example.

pandas/plotting/_core.py
```diff
@@ -2128,50 +2128,74 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
+    ax : Matplotlib axes object, default None
+        The axes to plot the histogram on.
     sharex : boolean, default True if ax is None else False
         In case subplots=True, share x axis and set some x axis labels to
         invisible; defaults to True if ax is None otherwise False if an ax
-        is passed in; Be aware, that passing in both an ax and sharex=True
-        will alter all x axis labels for all subplots in a figure!
+        is passed in.
+        Be aware: passing in both an ax and sharex=True will alter all x axis
```
dukebody: I'd say "Note" instead of "Be aware", which feels a bit more agressive.

pandas/plotting/_core.py
```diff
@@ -2128,50 +2128,74 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
     bins : integer or sequence, default 10
         Number of histogram bins to be used. If an integer is given, bins + 1
         bin edges are calculated and returned. If bins is a sequence, gives
         bin edges, including left edge of first bin and right edge of last
         bin. In this case, bins is returned unmodified.
-    `**kwds` : other plotting keyword arguments
-        To be passed to hist function
+    kwds : Keyword Arguments
+        All other plotting keyword arguments to be passed to
+        matplotlib's boxplot function.
```
dukebody: Link to this function using :method: or :class:.

pandas/plotting/_core.py
```diff
@@ -2128,50 +2128,74 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
     matplotlib.axes.Axes.hist : Plot a histogram using matplotlib.
 
+    Examples
+    --------
+
+    .. plot::
+        :context: close-figs
+
+        >>> df = pd.DataFrame({
+        ...     'length': [ 1.5, 0.5, 1.2, 0.9, 3],
```
dukebody: Mind the spacing: `[ 1.5, ...`. In next lines, there are two spaces between the 0.2 and the 1.1.

pandas/plotting/_core.py
```diff
@@ -2128,50 +2128,74 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
     --------
     matplotlib.axes.Axes.hist : Plot a histogram using matplotlib.
 
+    Examples
+    --------
+
+    .. plot::
+        :context: close-figs
+
+        >>> df = pd.DataFrame({
```
dukebody: Can you write some introduction explaining in words what the code does?

pandas/plotting/_core.py
DZPM: Ok, done!

pandas/plotting/_core.py
```diff
@@ -2128,50 +2128,74 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
...
+        ...     'length': [ 1.5, 0.5, 1.2, 0.9, 3],
+        ...     'width': [ 0.7, 0.2, 0.15, 0.2,  1.1]
+        ...     }, index= ['pig', 'rabbit', 'duck', 'chicken', 'horse'])
+        >>> hist = df.hist(bins=3)
```
dukebody: There are a lot of arguments for this method. Can you write a couple more examples trying to show how do they behave?

pandas/plotting/_core.py
```diff
@@ -2128,50 +2128,79 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
                xrot=None, ylabelsize=None, yrot=None, ax=None, sharex=False,
                sharey=False, figsize=None, layout=None, bins=10, **kwds):
     """
-    Draw histogram of the DataFrame's series using matplotlib / pylab.
+    Draw histogram of the DataFrame's columns using matplotlib.
+
+    A histogram is a representation of the distribution of data.
+    This function wraps the matplotlib histogram function for each series in
+    the DataFrame. It returns an array with a plot for each histogram.
```
jreback: can you link to wikipedia (here or in Notes is ok)

jreback (changes requested):

pandas/plotting/_core.py
dukebody: "It returns an array with a plot for each histogram" is a bit misleading. What about "A histogram is plotted for every column of the DataFrame"?

pandas/plotting/_core.py
```diff
@@ -2128,50 +2128,79 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
                xrot=None, ylabelsize=None, yrot=None, ax=None, sharex=False,
                sharey=False, figsize=None, layout=None, bins=10, **kwds):
     """
-    Draw histogram of the DataFrame's series using matplotlib / pylab.
+    Draw histogram of the DataFrame's columns using matplotlib.
```
dukebody: As in other cases, I believe that we should delay the fact that it is using matplotlib under the hood to the extended description and get it out of the short one.

TomAugspurger:
Thanks @DZPM !

PR #20114, opened 2018-03-10
DOC: update the pandas.Index.drop_duplicates and pandas.Series.drop_duplicates docstring

DaanVanHauwermeiren:
Checklist for the pandas documentation sprint (ignore this if you are doing
an unrelated PR):

- [X] PR title is "DOC: update the <your-function-or-method> docstring"
- [ ] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [X] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [X] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [ ] It has been proofread on language by another sprint participant

Please include the output of the validation script below between the "```" ticks:

Method ```pandas.Index.drop_duplicates```:

```
################################################################################
################### Docstring (pandas.Index.drop_duplicates) ###################
################################################################################

Return Index with duplicate values removed.

The drop_duplicates method can remove occurences or whole sets
of duplicated entries in a pandas.Index object.

Parameters
----------
keep : {'first', 'last', ``False``}, default 'first'
    - 'first' : Drop duplicates except for the first occurrence.
    - 'last' : Drop duplicates except for the last occurrence.
    - ``False`` : Drop all duplicates.

Returns
-------
deduplicated : Index

See Also
--------
pandas.Series.drop_duplicates : equivalent method on pandas.Series

Examples
--------
Generate an pandas.Index with duplicate values.

>>> idx = pd.Index(['lama', 'cow', 'lama', 'beetle', 'lama', 'hippo'])

With the 'keep' parameter, the selection behaviour of duplicated values
can be changed. The value 'first' keeps the first occurrence for each
set of duplicated entries. The default value of keep is 'first'.

>>> idx.drop_duplicates(keep='first')
Index(['lama', 'cow', 'beetle', 'hippo'], dtype='object')

The value 'last' keeps the last occurrence for each set of duplicated
entries.

>>> idx.drop_duplicates(keep='last')
Index(['cow', 'beetle', 'lama', 'hippo'], dtype='object')

The value ``False`` discards all sets of duplicated entries.

>>> idx.drop_duplicates(keep=False)
Index(['cow', 'beetle', 'hippo'], dtype='object')

################################################################################
################################## Validation ##################################
################################################################################

Errors found:
	Errors in parameters section
		Parameter "keep" description should start with capital letter

```

Method ```pandas.Series.drop_duplicates```:

```

################################################################################
[...]

PR #20117, opened 2018-03-10
DOC: update the pandas.Index.duplicated and pandas.Series.duplicated docstring

stijnvanhoey:
an unrelated PR):

Method `pandas.Index.duplicated`:
```
##################### Docstring (pandas.Index.duplicated)  #####################

Indicate duplicate index values.

Duplicated values are indicated as ``True`` values in the resulting
array. Either all duplicates, all except the first or all except the
last occurrence of duplicates can be indicated.

Parameters
----------
keep : {'first', 'last', False}, default 'first'
    - 'first' : Mark duplicates as ``True`` except for the first
      occurrence.
    - 'last' : Mark duplicates as ``True`` except for the last
      occurrence.
    - ``False`` : Mark all duplicates as ``True``.

Examples
--------

>>> idx = pd.Index(['lama', 'cow', 'lama', 'beetle', 'lama'])
>>> idx.duplicated()
array([False, False,  True, False,  True], dtype=bool)

which is equivalent to

>>> idx.duplicated(keep='first')
array([False, False,  True, False,  True], dtype=bool)

By using 'last', the last occurrence of each set of duplicated values
is set on False and all others on True:

>>> idx.duplicated(keep='last')
array([ True, False,  True, False, False], dtype=bool)

By setting keep on ``False``, all duplicates are True:

>>> idx.duplicated(keep=False)
array([ True, False,  True, False,  True], dtype=bool)

Returns
-------
numpy.ndarray

See Also
--------
pandas.Series.duplicated : equivalent method on pandas.Series

Errors found:
	Errors in parameters section
		Parameter "keep" description should start with capital letter
```

Method `pandas.Series.duplicated`:
```
##################### Docstring (pandas.Series.duplicated) #####################

Indicate duplicate Series values.

Duplicated values are indicated as ``True`` values in the resulting
Series. Either all duplicates, all except the first or all except the
last occurrence of duplicates can be indicated.

Parameters
----------
keep : {'first', 'last', False}, default 'first'
    - 'first' : Mark duplicates as ``True`` except for the first
      occurrence.
    - 'last' : Mark duplicates as ``True`` except for the last
      occurrence.
    - ``False`` : Mark all duplicates as ``True``.

Examples
--------

>>> animals = pd.Series(['lama', 'cow', 'lama', 'beetle', 'lama'])
>>> animals.duplicated()
0    False
1    False
2     True
3    False
4     True
dtype: bool

which is equivalent to

>>> animals.duplicated(keep='first')
0    False
1    False
2     True
3    False
4     True
dtype: bool

By using 'last', the last occurrence of each set of duplicated values
is set on False and all others on True:

>>> animals.duplicated(keep='last')
0     True
1    False
2     True
3    False
4    False
dtype: bool

By setting keep on ``False``, all duplicates are True:

>>> animals.duplicated(keep=False)
0     True
1    False
2     True
3    False
4     True
dtype: bool

Returnsinde
-------
pandas.core.series.Series

See Also
--------
pandas.Index.duplicated : equivalent method on pandas.Index

Errors found:
	Errors in parameters section
		Parameter "keep" description should start with capital letter
```

explicitly.

- We (Ghent chapter) decided that an additional line of text (with capital) was less useful than starting with explaining the list of options.

Instead of using the template-based version used before, we split out both docstrings and made a separate for the `Index` versus the `Series`. This introduces some redundancy and overlap (basically, the `keep` argument, also shared with `drop_duplicated`), but provides a cleaner option by having the examples written inside the docstring of the methods and not *somewhere* else in the code.


jorisvandenbossche:
@TomAugspurger we will have the same discussion here about sharing the docstrings or not

TomAugspurger:
FWIW I have a slight preference for sharing. Don't have a strong opinion though.

pandas/core/series.py
```diff
@@ -1320,8 +1320,76 @@ def unique(self):
...
+        4     True
+        dtype: bool
+
+        Returns
+        -------
+        pandas.core.series.Series
+
+        See Also
+        --------
+        pandas.Index.duplicated : equivalent method on pandas.Index
```
jreback: link so Series.drop_duplicates, DataFrame.duplicated

pandas/core/indexes/base.py
```diff
@@ -4021,8 +4021,56 @@ def unique(self, level=None):
     def drop_duplicates(self, keep='first'):
         return super(Index, self).drop_duplicates(keep=keep)
 
-    @Appender(base._shared_docs['duplicated'] % _index_doc_kwargs)
     def duplicated(self, keep='first'):
+        """
+        Indicate duplicate index values.
+
+        Duplicated values are indicated as ``True`` values in the resulting
```
jreback: can you coordinate text with #20114, seems some slight differences

jreback (changes requested):

pandas/core/indexes/base.py
jorisvandenbossche: Looks like they are in the zone together already :-) 
(they are sitting close to me: we removed the extended summary in the other PR as Tom asked, or are there other differences?)

jreback: great, seemed likely.

jreback:
Ideally I agree that sharing would be nice.

stijnvanhoey:
Could we maybe share the essential common element, i.e. the parameters information and load that ffrom a shared doc-variable?

jreback:
well ideally this *should* be much more shared, it ends up being a lot of duplicate text. (this means if we can fix it to make it shared would be the best), but if its too complicated for now, then ok too.

TomAugspurger:
Thanks @stijnvanhoey

PR #20118, opened 2018-03-10
DOC: improved the scatter method

DaniGate:
an unrelated PR):

```


################## Docstring (pandas.DataFrame.plot.scatter)  ##################

A scatter plot with point size *s* and color *c*.

The coordinates of each point *x,y* are defined by two dataframe
columns and filled circles are used to represent each point.

Parameters
----------
x : column name or column position
    Horizontal and vertical coordinates of each point.
y : column name or column position
    Vertical coordinates of each point.
s : scalar or array_like, optional
    Size of each point.
c : label, column name or column position, optional
    Color of each point.
kwds : optional

Returns
-------

See Also
--------
matplotlib.pyplot.scatter : scatter plot using multiple input data
    formats.

Examples
--------

.. plot::
    :context: close-figs

    >>> from sklearn.datasets import load_iris
    >>> iris = load_iris()
    >>> df = pd.DataFrame(iris.data[:,:2],
    ...                   columns=iris.feature_names[:2])
    >>> df['species'] = load_iris().target
    >>> f = df.plot.scatter(x='sepal length (cm)',
    ...                     y='sepal width (cm)',
    ...                     c='species',
    ...                     colormap='viridis')

Docstring for "pandas.DataFrame.plot.scatter" correct. :)

```

explicitly.

### DIFF
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2852,22 +2852,48 @@ class FramePlotMethods(BasePlotMethods):
 
     def scatter(self, x, y, s=None, c=None, **kwds):
         """
-        Scatter plot
+        A scatter plot with point size *s* and color *c*.
+
+        The coordinates of each point *x,y* are defined by two dataframe
+        columns and filled circles are used to represent each point.
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
+        x : column name or column position
+            Horizontal and vertical coordinates of each point.
+        y : column name or column position
+            Vertical coordinates of each point.
         s : scalar or array_like, optional
             Size of each point.
-        c : label or position, optional
+        c : label, column name or column position, optional
             Color of each point.
-        `**kwds` : optional
+        kwds : optional
             Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
         axes : matplotlib.AxesSubplot or np.array of them
+
+        See Also
+        --------
+        matplotlib.pyplot.scatter : scatter plot using multiple input data
+            formats.
+
+        Examples
+        --------
+
+        .. plot::
+            :context: close-figs
+
+            >>> from sklearn.datasets import load_iris
+            >>> iris = load_iris()
+            >>> df = pd.DataFrame(iris.data[:,:2],
+            ...                   columns=iris.feature_names[:2])
+            >>> df['species'] = load_iris().target
+            >>> f = df.plot.scatter(x='sepal length (cm)',
+            ...                     y='sepal width (cm)',
+            ...                     c='species',
+            ...                     colormap='viridis')
         """
         return self(kind='scatter', x=x, y=y, c=c, s=s, **kwds)
 
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2852,22 +2852,47 @@ class FramePlotMethods(BasePlotMethods):
 
     def scatter(self, x, y, s=None, c=None, **kwds):
         """
-        A scatter plot with point size `s` and color `c`.
+        Create a scatter plot with varying marker point size and color.
 
-        The coordinates of each point `x,y` are defined by two dataframe
-        columns and filled circles are used to represent each point.
+        The coordinates of each point are defined by two dataframe columns and
+        filled circles are used to represent each point. This kind of plot is
+        useful to see complex correlations between two variables. Points could
+        be for instance natural 2D coordinates like longitude and latitude in
+        a map or, in general, any pair of metrics that can be plotted against
+        each other.
 
         Parameters
         ----------
-        x : column name or column position
-            Horizontal coordinates of each point.
-        y : column name or column position
-            Vertical coordinates of each point.
-        s : scalar or array_like, optional
-            Size of each point.
-        c : label, column name or column position, optional
-            Color of each point.
-        kwds : optional
+        x : int, str
+            The column name or column position to be used as horizontal
+            coordinates for each point.
+        y : int, str
+            The column name or column position to be used as vertical
+            coordinates for each point.
+        s : scalar, array_like, optional
+            The size of each point. Possible values are:
+
+            - A single scalar so all points have the same size.
+
+            - A sequence of scalars, which will be used for each point's size
+            recursively. For intance [2,14] all points will be size 2 or 14,
+            alternatively.
+
+        c : str, int, array_like, optional
+            The color of each point. Possible values are:
+
+            - A single color string referred to by name, RGB or RGBA code,
+            for instance 'red' or '#a98d19'.
+
+            - A sequence of color strings referred to by name, RGB or RGBA code,
+            which will be used for each point's color recursively. For intance
+            ['green','yellow'] all points will be filled in green or yellow,
+            alternatively.
+
+            - A column name or position whose values will be used to color the
+            marker points according to a colormap.
+
+        **kwds : optional
             Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
 
         Returns
@@ -2881,6 +2906,8 @@ class FramePlotMethods(BasePlotMethods):
 
         Examples
         --------
+        Let's see how to draw a scatter plot using coordinates and color from
+        the values in three DataFrame columns.
 
         .. plot::
             :context: close-figs
@@ -2888,10 +2915,13 @@ class FramePlotMethods(BasePlotMethods):
             >>> df = pd.DataFrame([[5.1, 3.5, 0], [4.9, 3.0, 0], [7.0, 3.2, 1],
             ...                    [6.4, 3.2, 1], [5.9, 3.0, 2]],
             ...                   columns = ['length', 'width', 'species'])
-            >>> f = df.plot.scatter(x='length',
-            ...                     y='width',
-            ...                     c='species',
-            ...                     colormap='viridis')
+            >>> ax1 = df.plot.scatter(x='length',
+            ...                       y='width',
+            ...                       c='DarkBlue')
+            >>> ax2 = df.plot.scatter(x='length',
+            ...                       y='width',
+            ...                       c='species',
+            ...                       colormap='viridis')
         """
         return self(kind='scatter', x=x, y=y, c=c, s=s, **kwds)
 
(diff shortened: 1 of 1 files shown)


pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,48 @@ def pie(self, y=None, **kwds):
 
     def scatter(self, x, y, s=None, c=None, **kwds):
         """
-        Scatter plot
+        A scatter plot with point size *s* and color *c*.
```
TomAugspurger: How about the first line is "A scatter plot of two columns in the DataFrame."

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,48 @@ def pie(self, y=None, **kwds):
...
+
+        The coordinates of each point *x,y* are defined by two dataframe
```
TomAugspurger: Parameter names should be in singel backticks.

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,48 @@ def pie(self, y=None, **kwds):
...
+        columns and filled circles are used to represent each point.
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
+        x : column name or column position
```
TomAugspurger: I think `label or position` is OK. You can remove optional, as both `x` and `y` are required.

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,48 @@ def pie(self, y=None, **kwds):
...
+        matplotlib.pyplot.scatter : scatter plot using multiple input data
+            formats.
+
+        Examples
+        --------
+
+        .. plot::
+            :context: close-figs
+
+            >>> from sklearn.datasets import load_iris
```
TomAugspurger: We can't import sklearn here, since it isn't installed.

TomAugspurger: Maybe just make a small sample dataset instead.

DaniGate: It passed all checks and test and the plot was shown in the docs. But ok, I will write down the dataframe explicitely.

pandas/plotting/_core.py
DaniGate: But the description of the argument `c` specifies column name or column position. My intention was to make it consistent

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,46 @@ def pie(self, y=None, **kwds):
 
     def scatter(self, x, y, s=None, c=None, **kwds):
         """
-        Scatter plot
+        A scatter plot with point size `s` and color `c`.
```
dukebody: To be consistent with the rest of the docs from the chapter I'd start this with an infinitive verb like "Create a scatter plot", "Generate a scatter plot", "Make a scatter plot"...

dukebody:
FTR this is pandas.DataFrame.plot.scatter.

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,46 @@ def pie(self, y=None, **kwds):
...
+
+        See Also
+        --------
+        matplotlib.pyplot.scatter : scatter plot using multiple input data
+            formats.
+
+        Examples
+        --------
+
+        .. plot::
```
dukebody: Can you add some text above the code explaining what are you doing here?

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,46 @@ def pie(self, y=None, **kwds):
...
+            :context: close-figs
+
+            >>> df = pd.DataFrame([[5.1, 3.5, 0], [4.9, 3.0, 0], [7.0, 3.2, 1],
+            ...                    [6.4, 3.2, 1], [5.9, 3.0, 2]],
+            ...                   columns = ['length', 'width', 'species'])
+            >>> f = df.plot.scatter(x='length',
```
dukebody: If what is returned is an axis, better to use `ax = ...` here.

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,46 @@ def pie(self, y=None, **kwds):
...
+            ...                     y='width',
+            ...                     c='species',
+            ...                     colormap='viridis')
```
dukebody: Can you add at least one more example with different parameters to see the differences?

dukebody (changes requested):
Please add the type of the parameters and don't mention them in the description and extended description.

pandas/plotting/_core.py
dukebody: There is some confusion about using "label or position" as type. In other PR @datapythonista suggests using real types like "int or str" as the argument type, and mention it's the column name or position in the description of the parameter on next line.

datapythonista: yes, my mistake, I realized later that `label` could be any type, I guess that's why `label or position` is being used.

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,46 @@ def pie(self, y=None, **kwds):
...
-        x, y : label or position, optional
-            Coordinates for each point.
+        x : column name or column position
+            Horizontal coordinates of each point.
+        y : column name or column position
+            Vertical coordinates of each point.
         s : scalar or array_like, optional
             Size of each point.
-        c : label or position, optional
+        c : label, column name or column position, optional
```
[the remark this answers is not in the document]
TomAugspurger: I think valid options are matplotlib's + a column label, which extracts the array and passes it to matplotlib.

from plt.scatter:

```
c : color, sequence, or sequence of color, optional, default: 'b'
    `c` can be a single color format string, or a sequence of color
    specifications of length `N`, or a sequence of `N` numbers to be
    mapped to colors using the `cmap` and `norm` specified via kwargs
    (see below). Note that `c` should not be a single numeric RGB or
    RGBA sequence because that is indistinguishable from an array of
    values to be colormapped.  `c` can be a 2-D array in which the
    rows are RGB or RGBA, however, including the case of a single
    row to specify the same color for all points.
```

You'll to verify that. Check if / when a colorbar is added.

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,46 @@ def pie(self, y=None, **kwds):
...
             Color of each point.
-        `**kwds` : optional
+        kwds : optional
```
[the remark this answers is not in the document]
TomAugspurger: And remove the `: optional`. Just `**kwds`.

pandas/plotting/_core.py
TomAugspurger: Can you check if we recommend `cmap` or `colormap`?

DaniGate: @TomAugspurger They are both accepted but the one described in pandas.DataFrame.plot is colormap.

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,47 @@ def pie(self, y=None, **kwds):
 
     def scatter(self, x, y, s=None, c=None, **kwds):
         """
-        A scatter plot with point size `s` and color `c`.
+        Create a scatter plot with varying marker point size and color.
 
-        The coordinates of each point `x,y` are defined by two dataframe
-        columns and filled circles are used to represent each point.
+        The coordinates of each point are defined by two dataframe columns and
```
dukebody: Muuuuch better!!!!!!

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,47 @@ def pie(self, y=None, **kwds):
...
-        x : column name or column position
-            Horizontal coordinates of each point.
-        y : column name or column position
-            Vertical coordinates of each point.
-        s : scalar or array_like, optional
-            Size of each point.
-        c : label, column name or column position, optional
-            Color of each point.
-        kwds : optional
+        x : int, str
```
dukebody: I think this should be `int or str` instead.

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,47 @@ def pie(self, y=None, **kwds):
...
+        y : int, str
+            The column name or column position to be used as vertical
+            coordinates for each point.
+        s : scalar, array_like, optional
+            The size of each point. Possible values are:
+
+            - A single scalar so all points have the same size.
+
+            - A sequence of scalars, which will be used for each point's size
+            recursively. For intance [2,14] all points will be size 2 or 14,
```
dukebody: "For instance, using ``[2, 14]`` all points will be of size ...". Typo in "instance", the rest is a proposal to make it more prose-English.

pandas/plotting/_core.py
```diff
@@ -2852,22 +2852,47 @@ def pie(self, y=None, **kwds):
...
+            alternatively.
+
+        c : str, int, array_like, optional
+            The color of each point. Possible values are:
+
+            - A single color string referred to by name, RGB or RGBA code,
```
dukebody: Have you verified if these bullet points render nicely in the final HTML? I'm not good at restructured text so I tend to be wary about these things. :)

pandas/plotting/_core.py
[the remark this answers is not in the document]
dukebody: @TomAugspurger I believe we have used `**kwds : optionals` in other similar docstrings. Probably we should come up with an agreement on how to document the `**kwds` parameter in these cases.

dukebody:
This is almost done I believe, just a couple of style fixes and IMO it will be good to merge. :)

pandas/plotting/_core.py
```diff
@@ -2881,17 +2906,22 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
         --------
+        Let's see how to draw a scatter plot using coordinates and color from
+        the values in three DataFrame columns.
 
         .. plot::
             :context: close-figs
 
             >>> df = pd.DataFrame([[5.1, 3.5, 0], [4.9, 3.0, 0], [7.0, 3.2, 1],
             ...                    [6.4, 3.2, 1], [5.9, 3.0, 2]],
             ...                   columns = ['length', 'width', 'species'])
```
dukebody: Can you remove spaces between the equal sign? `columns=['length...`

pandas/plotting/_core.py
DaniGate: You need the blank lines to properly render the bullet points ;)
screen shot 2018-03-14 at 09 59 01

TomAugspurger:
Thanks @DaniGate and @dukebody .

Fixed the line length issue.

TomAugspurger:
OK, also split the examples in two to be a little clearer. Merging, thanks1

PR #20121, opened 2018-03-10
DOC: updated the pandas.DataFrame.plot.hexbin docstring

BielStela:
an unrelated PR):

```

################### Docstring (pandas.DataFrame.plot.hexbin) ###################

Make hexagonal binning plots.

Make an hexagonal binning plot of *x* versus *y*, where *x*,
*y* are 1-D sequences of the same length, *N*. If *C* is *None*
(the default), this is an histogram of the number of occurrences
of the observations at (x[i],y[i]).

If *C* is specified, specifies values at given coordinates
(x[i],y[i]). These values are accumulated for each hexagonal
bin and then reduced according to *reduce_C_function*, having as default
the numpy's mean function (np.mean). (If *C* is
specified, it must also be a 1-D sequence of the same length
as *x* and *y*.)

Parameters
----------
x : label or position, optional
    Coordinates for x point.
y : label or position, optional
    Coordinates for y point.
C : label or position, optional
    The value at each `(x, y)` point.
reduce_C_function : callable, optional, default `mean`
    Function of one argument that reduces all the values in a bin to
    a single number (e.g. `mean`, `max`, `sum`, `std`).
gridsize : int, optional, default 100
    The number of hexagons in the x-direction.
    The corresponding number of hexagons in the y-direction is chosen
    in a way that the hexagons are approximately regular. Alternatively,
    gridsize can be a tuple with two elements specifying the number of
    hexagons in the x-direction and the y-direction.
kwds : optional

Returns
-------
axes : matplotlib.AxesSubplot or np.array of them.

See Also
--------
matplotlib.pyplot.hexbin : hexagonal binning plot using matplotlib.

Examples
--------

.. plot::
    :context: close-figs

    >>> from  sklearn.datasets import load_iris
    >>> iris = load_iris()
    >>> df = pd.DataFrame(iris.data, columns=iris.feature_names)
    >>> hexbin = df.plot.hexbin(x='sepal length (cm)', y='sepal width (cm)',
    ...                         gridsize=10, cmap='viridis')

Docstring for "pandas.DataFrame.plot.hexbin" correct. :)


```
thanks to @Renton2017


pandas/plotting/_core.py
```diff
@@ -2874,25 +2874,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
     def hexbin(self, x, y, C=None, reduce_C_function=None, gridsize=None,
                **kwds):
         """
-        Hexbin plot
+        Make hexagonal binning plots.
+
+        Make an hexagonal binning plot of *x* versus *y*, where *x*,
```
TomAugspurger: an -> a.

Parameter names should be in backticks, so

    `x` versus `y`.

Renton2017: Well the topic related to h depends. You say An herb garden

Renton2017: It is actually depends whether the h is silent or not.  In case of hexagon I ask if it sound like herbal or like history. In case of history you say always "a history"

Renton2017: Looking for your reply.

Renton2017: Looking forward for your reply Tom, since I might be wrong.

pandas/plotting/_core.py
```diff
@@ -2874,25 +2874,62 @@ def scatter(self, x, y, s=None, c=None, **kwds):
     def hexbin(self, x, y, C=None, reduce_C_function=None, gridsize=None,
                **kwds):
         """
-        Hexbin plot
+        Make hexagonal binning plots.
+
+        Make an hexagonal binning plot of `x` versus `y`, where `x`,
+        `y` are 1-D sequences of the same length, `N`. If `C` is `None`
+        (the default), this is an histogram of the number of occurrences
```
jreback: lgtm. if you can find a referene on wikipedia might be nice to link.

jreback (approved):

pandas/plotting/_core.py
```diff
@@ -2874,25 +2874,62 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+        of the observations at (x[i],y[i]).
+
+        If `C` is specified, specifies values at given coordinates
+        (x[i],y[i]). These values are accumulated for each hexagonal
+        bin and then reduced according to `reduce_C_function`,
+        having as default
+        the numpy's mean function (np.mean). (If *C* is
+        specified, it must also be a 1-D sequence of the same length
+        as `x` and `y`.)
```
jorisvandenbossche: Can you reflow the text a bit so that each line is just less than 79 chars ?

pandas/plotting/_core.py
```diff
@@ -2874,25 +2874,62 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+            Alternatively,
+            gridsize can be a tuple with two elements specifying the number of
+            hexagons in the x-direction and the y-direction.
+        kwds : optional
             Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : matplotlib.AxesSubplot or np.array of them
+        axes : matplotlib.AxesSubplot or np.array of them.
```
jorisvandenbossche: Can you indicate here when an array is returned?

pandas/plotting/_core.py
```diff
@@ -2874,25 +2874,62 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+
+        See Also
+        --------
+        matplotlib.pyplot.hexbin : hexagonal binning plot using matplotlib.
```
jorisvandenbossche: I would say something like "the matplotlib function that is used under the hood" ?

pandas/plotting/_core.py
```diff
@@ -2874,25 +2874,62 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+
+        Examples
+        --------
+
+        .. plot::
+            :context: close-figs
+
+            >>> from  sklearn.datasets import load_iris
+            >>> iris = load_iris()
```
jorisvandenbossche: I am not fully sure if we want to rely on sklearn for the example data. In this case, I think it is actually fine to generate some random data with `np.random.randn(..)`, as that will change the generated plot

jorisvandenbossche (commented):

pandas/plotting/_core.py
jreback: yes pls don't use sklearn imports

BielStela: Ok, if there is no problem with using `np.random.randn()` I'll change the example to one similar to matplotlib's documentation.

jorisvandenbossche: yes, you can use random data in this case

pandas/plotting/_core.py
BielStela: I'm afraid there is no reference in wikipedia to hexagonal binning. The closed topic  I found in wiki is [the data binning article](https://en.wikipedia.org/wiki/Data_binning). Maybe it's a little bit to generic to include it in the hexbin docstring since it is also suitable for histogram and histogram2d

pandas/plotting/_core.py
TomAugspurger: I'm not sure if it ever is an ndarray.

pandas/plotting/_core.py
BielStela: Done!

pandas/plotting/_core.py
BielStela: It looks like the pandas wrapper can only get `matplotlib.AxesSubplot` ( I'm struggling to find a way to get `np.array` as return) , so if it's ok I'll omit this since it looks like it is a behavior of the original matplotlib function.

pandas/plotting/_core.py
BielStela: done

pandas/plotting/_core.py
dukebody: I will update the docstrings guide to explain when random data might be OK in examples, and also explain how to set a random seed to avoid changing examples. :)

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+        bin and then reduced according to `reduce_C_function`,
+        having as default the numpy's mean function (np.mean).
+        (If `C` is specified, it must also be a 1-D sequence
+        of the same length as `x` and `y`.)
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
+        x : label or position
```
dukebody: In the guide we ask people to use types like, `int`, `str` here for the parameter types. Should we accept "label or position" as well for types for column selection? This is something very common in many docstrings and we should to an agreement.

cc/ @datapythonista , @jorisvandenbossche

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
-            Additional keyword arguments are documented in
-            :meth:`pandas.DataFrame.plot`.
+        gridsize : int, optional, default 100
+            The number of hexagons in the x-direction.
+            The corresponding number of hexagons in the y-direction is
+            chosen in a way that the hexagons are approximately regular.
+            Alternatively,
+            gridsize can be a tuple with two elements specifying the number
+            of hexagons in the x-direction and the y-direction.
+        kwds : optional
```
dukebody: Should be `**kwds` instead of `kwds` according to our agreement, even if the validation script reports this as a mistake.

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+
+        See Also
+        --------
+        matplotlib.pyplot.hexbin : hexagonal binning plot using matplotlib,
+                    the matplotlib function that is used under the hood.
+
+        Examples
+        --------
+
+        .. plot::
```
dukebody: Can we add some plain English text before the example to explain what the example is about?

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
     def hexbin(self, x, y, C=None, reduce_C_function=None, gridsize=None,
                **kwds):
         """
-        Hexbin plot
+        Make hexagonal binning plot.
+
+        Make an hexagonal binning plot of `x` versus `y`. If `C` is `None`
+        (the default), this is an histogram of the number of occurrences
+        of the observations at (x[i],y[i]).
```
dukebody: I believe we should quote this using **double** backticks since it's code and not just a variable name:
```
of the observations at ``(x[i],y[i])``.
```
But I cannot find this in the guide. @datapythonista?

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+
+        If `C` is specified, specifies values at given coordinates
+        (x[i],y[i]). These values are accumulated for each hexagonal
+        bin and then reduced according to `reduce_C_function`,
+        having as default the numpy's mean function (np.mean).
```
dukebody: Can we link to the numpy function? I believe:
```
:meth:`numpy.mean`
 ```
should do the job.

File: pandas/plotting/_core.py [hunk 12]
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
-            Number of bins.
-        `**kwds` : optional
-            Additional keyword arguments are documented in
-            :meth:`pandas.DataFrame.plot`.
+        gridsize : int, optional, default 100
+            The number of hexagons in the x-direction.
+            The corresponding number of hexagons in the y-direction is
+            chosen in a way that the hexagons are approximately regular.
+            Alternatively,
+            gridsize can be a tuple with two elements specifying the number
```
dukebody: Then we should write: `gridsize : int or tuple of (int, int), optional ...`

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
 
         Returns
         -------
-        axes : matplotlib.AxesSubplot or np.array of them
+        axes : matplotlib.AxesSubplot.
+
+        See Also
+        --------
+        matplotlib.pyplot.hexbin : hexagonal binning plot using matplotlib,
+                    the matplotlib function that is used under the hood.
```
dukebody: Is the indentation level of this line OK? I thought it should be at the level of the previous one +4.

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+
+        Examples
+        --------
+
+        .. plot::
+            :context: close-figs
+
+            >>> n = 100000
+            >>> df = pd.DataFrame({'x':np.random.randn(n),
```
dukebody: For examples using random data I think we should add a seed. Otherwise the example will change everytime we build the documentation, which is a bit weird IMO.

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+            ...                    'y':np.random.randn(n)})
+            >>> hexbin = df.plot.hexbin(x='x', y='y', cmap='viridis')
```
dukebody: Can we add an example using `C`?

pandas/plotting/_core.py
datapythonista: In my opinion I'd use `int or str`, and use the description to say it's the label or the position. I don't see a reason to make an exception here.I think it's useful for the user to know that it's `int` or `str`. And who knows if in the future having the types can be useful in a way similar to mypy. :)

pandas/plotting/_core.py
datapythonista: yes please, we'll fix the script one of these days. :)

pandas/plotting/_core.py
datapythonista: We discussed the backticks really last minute, and the guide wasn't very clear, but I think that's the standard in general, yes.

As it's code, we can also respect PEP-8 and have the space after the comma. :)

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
     def hexbin(self, x, y, C=None, reduce_C_function=None, gridsize=None,
                **kwds):
         """
-        Hexbin plot
+        Make hexagonal binning plot.
```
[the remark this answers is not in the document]
jorisvandenbossche: And, something else, in the other PRs I think they used "Generate" instead of "Make" ? I don't really care which ones, but maybe best to be consistent

pandas/plotting/_core.py
jorisvandenbossche: The reason we *might* make an exception is because 'str' is too restrictive for labels, as a column labels can in principle be any hashable object ... (but ok, in most cases str will be what people use)

But I am fine here with using `int or str`

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
-        x, y : label or position, optional
-            Coordinates for each point.
+        x : label or position
+            Coordinates for x points.
+        y : label or position
+            Coordinates for y points.
         C : label or position, optional
             The value at each `(x, y)` point.
-        reduce_C_function : callable, optional
+        reduce_C_function : callable, optional, default `mean`
```
jorisvandenbossche: you can remove the 'optional' part here

(and the same below)

pandas/plotting/_core.py
```diff
@@ -2938,26 +2938,59 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
-            :meth:`pandas.DataFrame.plot`.
+        gridsize : int, optional, default 100
+            The number of hexagons in the x-direction.
+            The corresponding number of hexagons in the y-direction is
+            chosen in a way that the hexagons are approximately regular.
+            Alternatively,
+            gridsize can be a tuple with two elements specifying the number
+            of hexagons in the x-direction and the y-direction.
+        kwds : optional
+            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
```
jorisvandenbossche: Can you make this explanation into "Additional keyword arguments are documented in DataFrame.plot" ?

also, the `:py` part is not needed

pandas/plotting/_core.py
jorisvandenbossche: I think sphinx doesn't care, but let's indeed use '+4' to be consistent

pandas/plotting/_core.py
BielStela: Yes, I forgot about the type. I have a doubt with `reduce_C_function` parameter: it only accepts numpy functions in the form `np.mean` or  `np.std`  not strings like `'mean'`. Is  `function` (or `func`) the correct type reference?

TomAugspurger:
Probably callable

pandas/plotting/_core.py
dukebody: I've been told that it's OK-ish to use random data in plots without seed if the exact plot result is not important, so nevermind.

pandas/plotting/_core.py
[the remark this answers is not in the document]
BielStela: ok, Done

pandas/plotting/_core.py
BielStela: Yes, I'm working on one that is self explanatory

jorisvandenbossche (approved):

jorisvandenbossche:
Fixed the PEP8 issue and conflict with master

pandas/plotting/_core.py
```diff
@@ -3174,26 +3174,82 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+
+        Examples
+        --------
+        The following examples are generated with random data.
+
+        .. plot::
+            :context: close-figs
+
+            >>> n = 100000
+            >>> # Make a dataframe with normal distributed data
```
datapythonista: minor thing, but I haven't seen a comment in the interpreter in the rest of the documentation. I'd simply add to the previous explanation that the data is normal, and this would be a bit more compact and standard

TomAugspurger:
1 sec. Fixing some grammar.

pandas/plotting/_core.py
```diff
@@ -3174,26 +3174,82 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+            specifying the number of hexagons in the x-direction and the
+            y-direction.
+        **kwds
             Additional keyword arguments are documented in
             :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : :class:`matplotlib.axes.Axes` or numpy.ndarray of them
+        axes : matplotlib.AxesSubplot
```
datapythonista: We're using in most places just the type, and in the next line a short summary of what is returned.

pandas/plotting/_core.py
```diff
@@ -3174,26 +3174,82 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+
+        See Also
+        --------
+        matplotlib.pyplot.hexbin : hexagonal binning plot using matplotlib,
+            the matplotlib function that is used under the hood.
```
datapythonista: may be we could add `DataFrame.plot` here, as it's discussed in the parameters? Not sure what the other methods had in `See Also`.

datapythonista (commented):
Looks good, just minor comments

TomAugspurger (approved):

jorisvandenbossche:
@BielStela I made a small edit in the actual plot. The default gridsize seems rather useless, so I added one to the first plot to make it look a bit better. I also removed the cmap there (because the "inferno" has dark low values, it is not really fitting for a hexbin IMO, but added a custom cmap to the second example)

TomAugspurger:
@jorisvandenbossche you want to fixup @datapythonista's points?

jorisvandenbossche:
OK, editing via github at the same time is not a good idea :-) 
@TomAugspurger I will add your changes back, and do the ones of @datapythonista

jorisvandenbossche:
@BielStela Thanks a lot for the PR!

TomAugspurger:
:)

FYI Joris, https://github.com/github/hub is nice for checking out PRs. Just
`hub checkout <url>` and all the remotes are setup locally.

jorisvandenbossche:
Also, welcome to take a look in a few hours how the pages is looking to check everything is OK: http://pandas-docs.github.io/pandas-docs-travis/generated/pandas.DataFrame.plot.hexbin.html#pandas.DataFrame.plot.hexbin

jorisvandenbossche:
@TomAugspurger I have my alias `git pr <number>` which basically does the same I suppose (and normally I do that to test PRs, but for the doc prs now I was mainly using the github edit interface for really minor things)

BielStela:
Thanks a lot for the help and for the hard work you do!

PR #20125, opened 2018-03-10
DOC: Improved the docstring of pandas.Series.truncate

simobasso:
an unrelated PR):

```

###################### Docstring (pandas.Series.truncate) ######################

Truncate a DataFrame/Series before/after some index value.

If the axis contains only datetime values,
before/after parameters are converted to datetime values.

Parameters
----------
before : date, string, int
    Truncate all rows before this index value.
after : date, string, int
    Truncate all rows after this index value.
axis : {0 or 'index', 1 or 'columns'}
    Default is stat axis for given data type (0 for Series and
    DataFrames, 1 for Panels).
copy : boolean, default is True,
    Return a copy of the truncated section.

Returns
-------
truncated : type of caller

See Also
--------
DataFrame.truncate : Truncate a DataFrame before/after some index value.
Series.truncate : Truncate a Series before/after some index value.

Examples
--------
>>> df = pd.DataFrame({'A': ['a', 'b', 'c', 'd', 'e'],
...                    'B': ['f', 'g', 'h', 'i', 'j'],
...                    'C': ['k', 'l', 'm', 'n', 'o']},
...                    index=[1, 2, 3, 4, 5])
>>> df.truncate(before=2, after=4)
   A  B  C
2  b  g  l
3  c  h  m
4  d  i  n
>>> df = pd.DataFrame({'A': [1, 2, 3, 4, 5],
...                    'B': [6, 7, 8, 9, 10],
...                    'C': [11, 12, 13, 14, 15]},
...                    index=['a', 'b', 'c', 'd', 'e'])
>>> df.truncate(before='b', after='d')
   A  B   C
b  2  7  12
c  3  8  13
d  4  9  14

The index values in ``truncate`` can be datetimes or string
dates. Note that ``truncate`` assumes a 0 value for any unspecified
date component in a ``DatetimeIndex`` in contrast to slicing which
returns any partially matching dates.

>>> dates = pd.date_range('2016-01-01', '2016-02-01', freq='s')
>>> df = pd.DataFrame(index=dates, data={'A': 1})
>>> df.truncate('2016-01-05', '2016-01-10').tail()
                     A
2016-01-09 23:59:56  1
2016-01-09 23:59:57  1
2016-01-09 23:59:58  1
2016-01-09 23:59:59  1
2016-01-10 00:00:00  1
>>> df.loc['2016-01-05':'2016-01-10', :].tail()
                     A
2016-01-10 23:59:55  1
2016-01-10 23:59:56  1
2016-01-10 23:59:57  1
2016-01-10 23:59:58  1
2016-01-10 23:59:59  1

Docstring for "pandas.Series.truncate" correct. :)
```

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -6888,29 +6888,33 @@ class NDFrame(PandasObject, SelectionMixin):
 
     def truncate(self, before=None, after=None, axis=None, copy=True):
         """
-        Truncates a sorted DataFrame/Series before and/or after some
-        particular index value. If the axis contains only datetime values,
+        Truncate a DataFrame/Series before/after some index value.
+
+        If the axis contains only datetime values,
         before/after parameters are converted to datetime values.
 
         Parameters
         ----------
         before : date, string, int
-            Truncate all rows before this index value
+            Truncate all rows before this index value.
         after : date, string, int
-            Truncate all rows after this index value
+            Truncate all rows after this index value.
         axis : {0 or 'index', 1 or 'columns'}
-
-            * 0 or 'index': apply truncation to rows
-            * 1 or 'columns': apply truncation to columns
-
             Default is stat axis for given data type (0 for Series and
-            DataFrames, 1 for Panels)
+            DataFrames).
         copy : boolean, default is True,
-            return a copy of the truncated section
+            Return a copy of the truncated section.
 
         Returns
         -------
-        truncated : type of caller
+        type of caller
+            The truncated DataFrame/Series.
+
+        See Also
+        --------
+        DataFrame.truncate : Truncate a DataFrame before/after some index
+            value.
+        Series.truncate : Truncate a Series before/after some index value.
 
         Examples
         --------
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -6888,10 +6888,10 @@ class NDFrame(PandasObject, SelectionMixin):
 
     def truncate(self, before=None, after=None, axis=None, copy=True):
         """
-        Truncate a DataFrame/Series before/after some index value.
+        Truncate a Series or DataFrame before and after some index value.
 
-        If the axis contains only datetime values,
-        before/after parameters are converted to datetime values.
+        This is a useful shorthand for boolean indexing based on index
+        values above or below certain thresholds.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -6888,30 +6888,33 @@ def tshift(self, periods=1, freq=None, axis=0):
...
             Default is stat axis for given data type (0 for Series and
-            DataFrames, 1 for Panels)
+            DataFrames, 1 for Panels).
         copy : boolean, default is True,
-            return a copy of the truncated section
+            Return a copy of the truncated section.
 
         Returns
         -------
         truncated : type of caller
```
arnau126: If function returns a single value, name is no necessary.

arnau126: It should be, for example:
```
type of caller
        The truncated DataFrame/Series.
```

pandas/core/generic.py
```diff
@@ -6888,30 +6888,33 @@ def tshift(self, periods=1, freq=None, axis=0):
...
 
+        See Also
+        --------
+        DataFrame.truncate :
+            Truncate a DataFrame before/after some index value.
```
arnau126: Description must start in the same line:
```
DataFrame.truncate : Truncate a DataFrame before/after some index value.
```

arnau126: If more than 80 chars:
```
        DataFrame.truncate : Truncate a DataFrame before/after some index
            value.
```

simobasso:
@arnau126 thanks for review!

pandas/core/generic.py
```diff
@@ -6888,29 +6888,33 @@ def tshift(self, periods=1, freq=None, axis=0):
...
         after : date, string, int
-            Truncate all rows after this index value
+            Truncate all rows after this index value.
         axis : {0 or 'index', 1 or 'columns'}
-
-            * 0 or 'index': apply truncation to rows
-            * 1 or 'columns': apply truncation to columns
-
             Default is stat axis for given data type (0 for Series and
-            DataFrames, 1 for Panels)
```
jreback: you can remove the default for Panel

jreback (changes requested):

pandas/core/generic.py
simobasso: done thanks!

pandas/core/generic.py
```diff
@@ -6888,29 +6888,33 @@ def tshift(self, periods=1, freq=None, axis=0):
...
 
         Returns
         -------
-        truncated : type of caller
+        type of caller
+            The truncated DataFrame/Series.
+
+        See Also
+        --------
+        DataFrame.truncate : Truncate a DataFrame before/after some index
```
jreback: we don't have these defined in Series/DataFrame, so not sure these can be here @jorisvandenbossche

jreback (changes requested):

pandas/core/generic.py
```diff
@@ -6888,29 +6888,33 @@ def tshift(self, periods=1, freq=None, axis=0):
...
         before : date, string, int
-            Truncate all rows before this index value
+            Truncate all rows before this index value.
         after : date, string, int
-            Truncate all rows after this index value
+            Truncate all rows after this index value.
         axis : {0 or 'index', 1 or 'columns'}
-
-            * 0 or 'index': apply truncation to rows
-            * 1 or 'columns': apply truncation to columns
```
jorisvandenbossche: I think this list was still informative? So I would keep it.

pandas/core/generic.py
jorisvandenbossche: Yes, since the docstring is shared for both Series and DataFrame, those "see also" are referring to themselves. So indeed not needed in this case (and you ignore the error in the validation script about missing "See Also")

TomAugspurger (approved):
Updated.

I think the original doc for `axis` was wrong. `rows` is always truncated by default (which makes more sense to me). See the examples.

jorisvandenbossche (approved):

jorisvandenbossche:
@simobasso Thanks for the PR!

PR #20126, opened 2018-03-10
DOC: update the pandas.DataFrame.to_sql docstring

jazzmuesli:
an unrelated PR):

```

##################### Docstring (pandas.DataFrame.to_sql)  #####################

Write records stored in a DataFrame to a SQL database.

This function inserts all rows of the dataframe into the given
 table and recreates it if if_exists='replace'. Databases supported by
 SQLAlchemy or DBAPI2 are also supported.

Parameters
----------
name : string
    Name of SQL table.
con : SQLAlchemy engine or DBAPI2 connection (legacy mode)
    Using SQLAlchemy makes it possible to use any DB supported by that
    library. If a DBAPI2 object, only sqlite3 is supported.
schema : string, default None
    Specify the schema (if database flavor supports this). If None, use
    default schema.
if_exists : {'fail', 'replace', 'append'}, default 'fail'
    Accepted values:
    - fail: If table exists, do nothing.
    - replace: If table exists, drop it, recreate it, and insert data.
    - append: If table exists, insert data. Create if does not exist.
index : boolean, default True
    Write DataFrame index as a column.
index_label : string or sequence, default None
    Column label for index column(s). If None is given (default) and
    `index` is True, then the index names are used.
    A sequence should be given if the DataFrame uses MultiIndex.
chunksize : int, default None
    If not None, then rows will be written in batches of this size at a
    time.  If None, all rows will be written at once.
dtype : dict of column name to SQL type, default None
    Optional specifying the datatype for columns. The SQL type should
    be a SQLAlchemy type, or a string for sqlite3 fallback connection.

Returns
--------
    None

See Also
--------
pandas.read_sql_query : read a DataFrame from a table

Examples
--------
>>> from sqlalchemy import create_engine
>>> engine = create_engine('sqlite:///example.db', echo=False)
>>> df = pd.DataFrame({'name' : ['User 1', 'User 2', 'User 3']})
>>> # create a table from scratch with 3 rows
>>> df.to_sql('users', con=engine, if_exists='replace')
>>> df1 = pd.DataFrame({'name' : ['User 4', 'User 5']})
>>> # 2 new rows inserted
>>> df1.to_sql('users', con=engine, if_exists='append')
>>> # table will be recreated and 5 rows inserted
>>> df = pd.concat([df, df1], ignore_index=True)
>>> df.to_sql('users', con=engine, if_exists='replace')
>>> pd.read_sql_query("select * from users",con=engine)
   index    name
0      0  User 1
1      1  User 2
2      2  User 3
3      3  User 4
4      4  User 5

Docstring for "pandas.DataFrame.to_sql" correct. :)

```

explicitly.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1913,16 +1913,19 @@ class NDFrame(PandasObject, SelectionMixin):
         --------
         >>> from sqlalchemy import create_engine
         >>> engine = create_engine('sqlite://', echo=False)
-        >>> df = pd.DataFrame({'name' : ['User 1', 'User 2', 'User 3']})
 
         **create a table from scratch with 3 rows**
+
+        >>> df = pd.DataFrame({'name' : ['User 1', 'User 2', 'User 3']})
         >>> df.to_sql('users', con=engine, if_exists='replace')
 
         **2 new rows inserted**
+
         >>> df1 = pd.DataFrame({'name' : ['User 4', 'User 5']})
         >>> df1.to_sql('users', con=engine, if_exists='append')
 
         **table will be recreated and 5 rows inserted**
+
         >>> df = pd.concat([df, df1], ignore_index=True)
         >>> df.to_sql('users', con=engine, if_exists='replace')
         >>> pd.read_sql_query("select * from users",con=engine)
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1865,79 +1865,108 @@ class NDFrame(PandasObject, SelectionMixin):
         """
         Write records stored in a DataFrame to a SQL database.
 
-        This function inserts all rows of the dataframe into the given
-         table and recreates it if if_exists='replace'. Databases supported by
-         SQLAlchemy or DBAPI2 are also supported.
+        Databases supported by SQLAlchemy [1]_ are supported. Tables can be
+        newly created, appended to, or overwritten.
 
         Parameters
         ----------
         name : string
             Name of SQL table.
-        con : SQLAlchemy engine or DBAPI2 connection (legacy mode)
+        con : sqlalchemy.engine.Engine or sqlite3.Connection
             Using SQLAlchemy makes it possible to use any DB supported by that
-            library. If a DBAPI2 object, only sqlite3 is supported.
-        schema : string, default None
+            library. Legacy support is provided for sqlite3.Connection objects.
+        schema : string, optional
             Specify the schema (if database flavor supports this). If None, use
             default schema.
         if_exists : {'fail', 'replace', 'append'}, default 'fail'
-            Accepted values:
-            - fail: If table exists, do nothing.
-            - replace: If table exists, drop it, recreate it, and insert data.
-            - append: If table exists, insert data. Create if does not exist.
+            How to behave if the table already exists.
+
+            * fail: Raise a ValueError.
+            * replace: Drop the table before inserting new values.
+            * append: Insert new values to the existing table.
+
         index : boolean, default True
-            Write DataFrame index as a column.
+            Write DataFrame index as a column. Uses `index_label` as the column
+            name in the table.
         index_label : string or sequence, default None
             Column label for index column(s). If None is given (default) and
             `index` is True, then the index names are used.
             A sequence should be given if the DataFrame uses MultiIndex.
-        chunksize : int, default None
-            If not None, then rows will be written in batches of this size at a
-            time.  If None, all rows will be written at once.
-        dtype : dict of column name to SQL type, default None
-            Optional specifying the datatype for columns. The SQL type should
-            be a SQLAlchemy type, or a string for sqlite3 fallback connection.
+        chunksize : int, optional
+            Rows will be written in batches of this size at a time. By default,
+            all rows will be written at once.
+        dtype : dict, optional
+            Specifying the datatype for columns. The keys should be the column
+            names and the values should be the SQLAlchemy types or strings for
+            the sqlite3 legacy mode.
 
-        Returns
-        --------
-            None
+        Raises
+        ------
+        ValueError
+            When the table already exists and `if_exists` is 'fail' (the
+            default).
 
         See Also
         --------
         pandas.read_sql : read a DataFrame from a table
 
+        References
+        ----------
+        .. [1] http://docs.sqlalchemy.org
+        .. [2] https://www.python.org/dev/peps/pep-0249/
+
         Examples
         --------
 
-        **create an in-memory database using [1]**
+        Create an in-memory SQLite database.
 
         >>> from sqlalchemy import create_engine
         >>> engine = create_engine('sqlite://', echo=False)
 
-        **create a table from scratch with 3 rows**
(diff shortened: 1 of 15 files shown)


pandas/core/generic.py
```diff
@@ -1892,6 +1896,30 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
...
+        --------
+            None
+
+        See Also
+        --------
+        pandas.io.sql.to_sql : this function will be called.
+
+        Examples
+        --------
+        >>> import pandas as pd
```
acidburnburn: pandas import is not recommended in examples.

pandas/core/generic.py
```diff
@@ -1865,17 +1865,21 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
         """
         Write records stored in a DataFrame to a SQL database.
 
+        This function inserts all rows of the dataframe into the given
```
acidburnburn: would be good to know which databases are supported and how.

pandas/core/generic.py
```diff
@@ -1892,6 +1896,30 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
...
+        Examples
+        --------
+        >>> import pandas as pd
+        >>> from sqlalchemy import create_engine
+        >>> engine = create_engine('sqlite:///example.db', echo=False)
+        >>> gen_names = lambda ids: ["User" + str(x) for x in ids]
+        >>> gen_users = lambda ids: {"id": ids, "name" : gen_names(ids)}
+        >>> df=pd.DataFrame(gen_users(list(range(3))))
+        >>> # create a table from scratch
+        >>> df.to_sql('users',con=engine,if_exists='replace')
```
acidburnburn: some example lines are not compatible with PEP8

pandas/core/generic.py
```diff
@@ -1892,6 +1896,30 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
...
+        pandas.io.sql.to_sql : this function will be called.
+
+        Examples
+        --------
+        >>> import pandas as pd
+        >>> from sqlalchemy import create_engine
+        >>> engine = create_engine('sqlite:///example.db', echo=False)
+        >>> gen_names = lambda ids: ["User" + str(x) for x in ids]
+        >>> gen_users = lambda ids: {"id": ids, "name" : gen_names(ids)}
+        >>> df=pd.DataFrame(gen_users(list(range(3))))
```
acidburnburn: maybe it's a bit more straightforward to generate dataframe without other functions:
i.e. df=pd.DataFrame(['User 1', 'User 2', 'User 3'])

pandas/core/generic.py
```diff
@@ -1865,17 +1865,22 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
...
-            Name of SQL table
+            Name of SQL table.
         con : SQLAlchemy engine or DBAPI2 connection (legacy mode)
             Using SQLAlchemy makes it possible to use any DB supported by that
             library. If a DBAPI2 object, only sqlite3 is supported.
         schema : string, default None
             Specify the schema (if database flavor supports this). If None, use
             default schema.
         if_exists : {'fail', 'replace', 'append'}, default 'fail'
+            Accepted values:
```
TomAugspurger: FWIW, I find it clearer without the line here, though I know the script complains when it isn't present :)  cc @datapythonista

pandas/core/generic.py
```diff
@@ -1892,6 +1897,29 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
             Optional specifying the datatype for columns. The SQL type should
             be a SQLAlchemy type, or a string for sqlite3 fallback connection.
 
+        Returns
```
TomAugspurger: The Returns section can be ommitted if ther's no return value.

pandas/core/generic.py
```diff
@@ -1892,6 +1897,29 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
...
+        --------
+            None
+
+        See Also
+        --------
+        pandas.io.sql.to_sql : this function will be called.
```
TomAugspurger: I don't think that method is public. I think just link to `pandas.read_sql`

pandas/core/generic.py
```diff
@@ -1892,6 +1897,29 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
...
+
+        Examples
+        --------
+        >>> from sqlalchemy import create_engine
+        >>> engine = create_engine('sqlite:///example.db', echo=False)
+        >>> gen_names = lambda ids: ["User" + str(x) for x in ids]
```
TomAugspurger: Any chance you could simplify this? It's a bit dense.

Maybe just make a simple df, display it. then show the insert. And maybe show a `pd.read_sql` to view the result.

pandas/core/generic.py
jazzmuesli: python scripts/validate_docstrings.py pandas.DataFrame.to_sql complains:
Errors found:
	No returns section found

pandas/core/generic.py
```diff
@@ -1865,17 +1865,22 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
         """
         Write records stored in a DataFrame to a SQL database.
 
+        This function inserts all rows of the dataframe into the given
+         table and recreates it if if_exists='replace'. Databases supported by
```
jreback: can add a link to SQLAlchemy in References

pandas/core/generic.py
```diff
@@ -1892,6 +1897,34 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
...
+        --------
+        pandas.read_sql_query : read a DataFrame from a table
+
+        Examples
+        --------
+        >>> from sqlalchemy import create_engine
+        >>> engine = create_engine('sqlite://', echo=False)
+        >>> df = pd.DataFrame({'name' : ['User 1', 'User 2', 'User 3']})
+        >>> # create a table from scratch with 3 rows
+        >>> df.to_sql('users', con=engine, if_exists='replace')
```
jreback: can you add blank lines between cases
also for comments, don't use the leading '>>>'

jreback (changes requested):

pandas/core/generic.py
```diff
@@ -1892,6 +1896,34 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
             Optional specifying the datatype for columns. The SQL type should
             be a SQLAlchemy type, or a string for sqlite3 fallback connection.
 
+        Returns
+        --------
+            None
+
+        See Also
+        --------
+        pandas.read_sql_query : read a DataFrame from a table
```
jreback: just to ``pandas.read_sql``

jreback (changes requested):

TomAugspurger:
Made some updates if anyone has a chance to review @jazzmuesli.

TomAugspurger:
Thanks @jazzmuesli !

PR #20133, opened 2018-03-10
DOC: update the pandas.DataFrame.plot.pie docstring

dim1tri:
an unrelated PR):

- [x] The validation script passes: `scripts/validate_docstrings.py pandas.DataFrame.plot.pie`

```
Docstring for "pandas.DataFrame.plot.pie" correct. :)
```


Renton2017 (commented):
Well the code is OK. I have a specific doubt on `**kwds` : optional. I would delete the asterisks and leave like `kwds`

Renton2017 (approved):
Well the code is OK. I have a specific doubt on `**kwds` : optional. I would delete the asterisks and leave like `kwds`

dim1tri:
@Renton2017, could you please make a reference to a line number?

dukebody:
The keyword arguments must be mentioned in the docstring as `**kwargs`.

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,37 @@ def area(self, x=None, y=None, **kwds):
...
 
         Parameters
         ----------
-        y : label or position, optional
-            Column to plot.
-        `**kwds` : optional
+        y : str or int, optional
+            Label or position.
+            If not provided, ``subplots=True`` argument must be passed.
+        kwds : optional
```
dukebody: Use `**kwds : optional`, not `kwds : optional`, even if the validation script complains about it.

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,37 @@ def area(self, x=None, y=None, **kwds):
...
+        function for specified column.
 
         Parameters
         ----------
-        y : label or position, optional
-            Column to plot.
-        `**kwds` : optional
+        y : str or int, optional
+            Label or position.
+            If not provided, ``subplots=True`` argument must be passed.
```
dukebody: Really? What happens if we don't pass `subplots=True` and we don't pass a `y` argument? Does it raise an error? What does the `subplots=True` argument change?

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,37 @@ def area(self, x=None, y=None, **kwds):
...
             Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : matplotlib.AxesSubplot or np.array of them
+        axes : matplotlib.AxesSubplot or np.array of them.
+
+        See Also
+        --------
+        :meth:`pandas.Series.plot.pie` : Generate a pie plot for a Serie.
```
dukebody: `pie plot for a Series`, in plural

dukebody:
Can you include a screenshot of the generated html?

Renton2017:
Hi Dimitri,

on line 2844 `**kwds` : optional 
meanwhile on line 2849 kwds  : optional 

They cannot be true both!

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,37 @@ def area(self, x=None, y=None, **kwds):
...
+        in a DataFrame column. This function wraps the `matplotlib.pyplot.pie`
+        function for specified column.
 
         Parameters
         ----------
-        y : label or position, optional
-            Column to plot.
-        `**kwds` : optional
+        y : str or int, optional
+            Label or position.
```
dukebody: I find the `Column to plot` part important. What about `Label of position indicating the column to plot`?

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,37 @@ def area(self, x=None, y=None, **kwds):
...
+        axes : matplotlib.AxesSubplot or np.array of them.
+
+        See Also
+        --------
+        :meth:`pandas.Series.plot.pie` : Generate a pie plot for a Serie.
+
+        Examples
+        --------
+
+        .. plot::
```
dukebody: Can you add some text above explaining what is the example about in plain English? Also, perhaps you want to create different plots for different `y` values to show how the plot changes when changing the `y` argument.

pandas/plotting/_core.py
dim1tri: Yes. This error is raised:
```ValueError: pie requires either y column or 'subplots=True'```
with subplots=True, pie() returns an array of plots for all columns

File: pandas/plotting/_core.py [hunk 6]
```diff
@@ -2835,18 +2835,42 @@ def area(self, x=None, y=None, **kwds):
 
     def pie(self, y=None, **kwds):
         """
-        Pie chart
+        Generate a pie plot.
+
+        A pie plot is a representation of the distribution of numerical data
+        in a DataFrame column. This function wraps the `matplotlib.pyplot.pie`
+        function for specified column. If no column reference is passed and
```
jreback: for the specified column

jreback: refer to 'y' here (its the column reference)

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,42 @@ def area(self, x=None, y=None, **kwds):
...
 
         Parameters
         ----------
-        y : label or position, optional
-            Column to plot.
-        `**kwds` : optional
+        y : str or int, optional
+            Label or position of the column to plot.
+            If not provided, ``subplots=True`` argument must be passed.
+        **kwds : optional
```
jreback: no ** on kwds

jreback (changes requested):

pandas/plotting/_core.py
dukebody: @jreback there is a huge confusion here. @datapythonista said we DO want ** on kwds and they are modifying the validation script to accept it. Now you say we don't want it. Can we make a decision about this?

jorisvandenbossche: @jreback we decided to keep them

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,42 @@ def area(self, x=None, y=None, **kwds):
 
     def pie(self, y=None, **kwds):
         """
-        Pie chart
+        Generate a pie plot.
+
+        A pie plot is a representation of the distribution of numerical data
+        in a DataFrame column. This function wraps the `matplotlib.pyplot.pie`
+        function for specified column. If no column reference is passed and
+        ``subplots`` argument is set to ``True``, an np.array of plots for
```
jorisvandenbossche: you can use single backticks for 'subplots' (because it is a reference to a parameter)

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,42 @@ def area(self, x=None, y=None, **kwds):
...
+        all DataFrame columns will be returned.
```
dukebody: `for all the DataFrame columns`

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,42 @@ def area(self, x=None, y=None, **kwds):
...
 
         Parameters
         ----------
-        y : label or position, optional
-            Column to plot.
-        `**kwds` : optional
+        y : str or int, optional
```
jorisvandenbossche: It's a bit a corner case as well, but in another PR we decide to keep the 'label', as 'str' can actually be too strict on the column name (column names can be other things as strings) ..

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,42 @@ def area(self, x=None, y=None, **kwds):
...
+            Label or position of the column to plot.
+            If not provided, ``subplots=True`` argument must be passed.
+        **kwds : optional
             Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : matplotlib.AxesSubplot or np.array of them
+        axes : matplotlib.AxesSubplot or np.array of them.
```
jorisvandenbossche: Can you explain here when an array is returned?

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,42 @@ def area(self, x=None, y=None, **kwds):
...
+
+        See Also
+        --------
+        :meth:`pandas.Series.plot.pie` : Generate a pie plot for a Series.
```
jorisvandenbossche: You can leave out the ``:meth:` `` part here (the link is automatically made by numpydoc

I would also add the see also to matplotlib.pyplot.pie

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,42 @@ def area(self, x=None, y=None, **kwds):
...
+        In the example below we have a DataFrame with the information about
+        planet's mass and radius. We pass the the 'mass' column to the
+        pie function to get a pie plot.
+
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({'mass': [0.330, 4.87 , 5.97],
+            ...                    'radius': [2439.7, 6051.8, 6378.1]})
+            >>> plot = df.plot.pie(y='mass', labels=['Mercury', 'Venus', 'Earth'])
```
jorisvandenbossche: Maybe also add an example with the subplots argument?

jorisvandenbossche (commented):

pandas/plotting/_core.py
datapythonista: @jreback  I thought the consensus was to have them. We've got a bug in the script that was validating the opposite, which created some confusion, but in the documentation we've got that they should have the `**` in the name.

If there is no final decision, I'd accept both. I think it should be quite easy to use `validate_docstrings.py` to detect all the ones that have, or not have them. And we can quickly change all them later on.

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,42 @@ def area(self, x=None, y=None, **kwds):
...
+        all DataFrame columns will be returned.
 
         Parameters
         ----------
-        y : label or position, optional
-            Column to plot.
-        `**kwds` : optional
+        y : str or int, optional
+            Label or position of the column to plot.
+            If not provided, ``subplots=True`` argument must be passed.
```
dukebody: Improvement proposal: `If not provided, the ``subplots=True`` argument must be passed. In this case, a subplot will be created for every column of the DataFrame`.

pandas/plotting/_core.py
```diff
@@ -2835,18 +2835,42 @@ def area(self, x=None, y=None, **kwds):
...
-        axes : matplotlib.AxesSubplot or np.array of them
+        axes : matplotlib.AxesSubplot or np.array of them.
+
+        See Also
+        --------
+        :meth:`pandas.Series.plot.pie` : Generate a pie plot for a Series.
+
+        Examples
+        --------
+        In the example below we have a DataFrame with the information about
```
dukebody: Perhaps a matter of style, but I'd say:
```
Given a DataFrame with information about some planets mass and radius, create a pie plot of their masses:
```
Because the fact that it is an example can be determined from the section it is in, and the fact that we pass the 'mass' column to the `pie` method can be seen in the code itself.

What about another example without specifying the `y` argument?

dukebody:
@dim1tri please ping me if you have issues fixing the merge conflicts.

dukebody:
@dim1tri is there something we can do to help you with this PR?

TomAugspurger:
Updated

Fixed the aspect ratio on the plots, added a `subplots=True` example, and used the `index` instead of labels.

fireshot capture 006 - pandas dataframe plot pie pandas 0 _ - file____users_taugspurger_sandbox_

TomAugspurger (approved):

jorisvandenbossche (approved):

jorisvandenbossche:
@dim1tri Thanks!

SimonBaron:
I have wanted to ask for a while, but never sure where.. Is this documentation page a joke? Is it intentionally trolling? It is, right?

PR #20134, opened 2018-03-10
DOC: update the pandas.core.generic.NDFrame.to_clipboard docstring

dandxy89:
an unrelated PR):

- [X] The validation script passes: `scripts/validate_docstrings.py pandas.core.generic.NDFrame.to_clipboard`
- [X] The html version looks good: `python doc/make.py --single pandas.core.generic.NDFrame.to_clipboard`

```
############# Docstring (pandas.core.generic.NDFrame.to_clipboard) #############

Copy object to the system clipboard.

Write a text representation of object to the system clipboard.
This can be pasted into Excel, for example.

Parameters
----------
excel : bool
    Default setting for this argument is True.
    If True, use the provided separator, writing in a csv format for
    allowing easy pasting into excel.
    If False, write a string representation of the object to the
    clipboard.
sep : str, default tab
    Field delimiter.
kwargs : optional
    These parameters will be passed to either
    :py:meth:`pandas.DataFrame.to_csv`
    or :py:meth:`pandas.Series.to_csv` methods depending on the
    Object type.

See Also
--------
pandas.core.frame.DataFrame.to_csv : Write a DataFrame to a
    comma-separated values (csv) file.
pandas.core.series.Series.to_csv : Write a Series to a
    comma-separated values (csv) file.

Notes
-----
Requirements for your platform.
  - Linux : `xclip`, or `xsel` (with `gtk` or `PyQt4` modules)
  - Windows : none
  - OS X : none

Examples
--------
Copy the contents of a DataFrame to the clipboard.

>>> df = pd.DataFrame([[1, 2, 3], [4, 5, 6]], columns=['A', 'B', 'C'])
>>> df.to_clipboard()

We can omit the the index by passing the keyword 'index' and setting it to false.

>>> df.to_clipboard(index=False)

```

explicitly.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1951,10 +1951,6 @@ class NDFrame(PandasObject, SelectionMixin):
             or :py:meth:`pandas.Series.to_csv` methods depending on the
             Object type.
 
-        Returns
-        -------
-            None
-
         See Also
         --------
         pandas.core.frame.DataFrame.to_csv : Write a DataFrame to a
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1943,7 +1943,7 @@ class NDFrame(PandasObject, SelectionMixin):
             - False, write a string representation of the object to the
               clipboard.
 
-        sep : str, ``\t``
+        sep : str, default ``'\t'``
             Field delimiter.
         **kwargs
             These parameters will be passed to DataFrame.to_csv.
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1929,7 +1929,7 @@ class NDFrame(PandasObject, SelectionMixin):
                          protocol=protocol)
 
     def to_clipboard(self, excel=True, sep=None, **kwargs):
-        """
+        r"""
         Copy object to the system clipboard.
 
         Write a text representation of object to the system clipboard.
@@ -1939,23 +1939,25 @@ class NDFrame(PandasObject, SelectionMixin):
         ----------
         excel : bool, default True
             - True, use the provided separator, writing in a csv format for
-            allowing easy pasting into excel.
+              allowing easy pasting into excel.
             - False, write a string representation of the object to the
-            clipboard.
-        sep : str, default tab
+              clipboard.
+
+        sep : str, ``\t``
             Field delimiter.
-        kwargs : optional -> **kwargs
-            These parameters will be passed to meth:`pandas.DataFrame.to_csv`
+        **kwargs
+            These parameters will be passed to DataFrame.to_csv.
 
         See Also
         --------
         DataFrame.to_csv : Write a DataFrame to a comma-separated values
-        (csv) file.
+            (csv) file.
         read_clipboard : Read text from clipboard and pass to read_table.
 
         Notes
         -----
         Requirements for your platform.
+
           - Linux : `xclip`, or `xsel` (with `gtk` or `PyQt4` modules)
           - Windows : none
           - OS X : none
@@ -1965,12 +1967,20 @@ class NDFrame(PandasObject, SelectionMixin):
         Copy the contents of a DataFrame to the clipboard.
 
         >>> df = pd.DataFrame([[1, 2, 3], [4, 5, 6]], columns=['A', 'B', 'C'])
-        >>> df.to_clipboard()
+        >>> df.to_clipboard(sep=',')
+        ... # Wrote the following to the system clipboard:
+        ... # ,A,B,C
+        ... # 0,1,2,3
+        ... # 1,4,5,6
 
         We can omit the the index by passing the keyword `index` and setting
         it to false.
 
-        >>> df.to_clipboard(index=False)
+        >>> df.to_clipboard(sep=',', index=False)
+        ... # Wrote the following to the system clipboard:
+        ... # A,B,C
+        ... # 1,2,3
+        ... # 4,5,6
         """
         from pandas.io import clipboards
         clipboards.to_clipboard(self, excel=excel, sep=sep, **kwargs)
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1972,7 +1972,8 @@ class NDFrame(PandasObject, SelectionMixin):
         >>> df = pd.DataFrame([[1, 2, 3], [4, 5, 6]], columns=['A', 'B', 'C'])
         >>> df.to_clipboard()
 
-        We can omit the the index by passing the keyword 'index' and setting it to false.
+        We can omit the the index by passing the keyword 'index' and setting
+         it to false.
 
         >>> df.to_clipboard(index=False)
         """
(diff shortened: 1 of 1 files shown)


benman1:
```
################### Docstring (pandas.DataFrame.to_parquet)  ###################

Write a DataFrame to the binary parquet format.

.. versionadded:: 0.21.0

Requires either fastparquet or pyarrow libraries.

Parameters
----------
fname : str
    String file path.
engine : {'auto', 'pyarrow', 'fastparquet'}, default 'auto'
    Parquet library to use. If 'auto', then the option
    ``io.parquet.engine`` is used. The default ``io.parquet.engine``
    behavior is to try 'pyarrow', falling back to 'fastparquet' if
    'pyarrow' is unavailable.
compression : {'snappy', 'gzip', 'brotli', None}, default 'snappy'
    Name of the compression to use. Use ``None`` for no compression.
kwargs : dict
    Additional keyword arguments passed to the engine.

Examples
----------
>>> df = pd.DataFrame(data={'col1': [1, 2], 'col2': [3, 4]})
>>> df.to_parquet('df.parquet.gzip', compression='gzip')

Returns
----------
None

See Also
--------
DataFrame.to_csv : write a csv file.
DataFrame.to_sql : write to a sql table.
DataFrame.to_hdf : write to hdf.

Docstring for "pandas.DataFrame.to_parquet" correct. :)
```

pandas/core/generic.py
```diff
@@ -1930,25 +1930,52 @@ def to_pickle(self, path, compression='infer',
...
-        excel : boolean, defaults to True
-                if True, use the provided separator, writing in a csv
-                format for allowing easy pasting into excel.
-                if False, write a string representation of the object
-                to the clipboard
-        sep : optional, defaults to tab
-        other keywords are passed to to_csv
+        excel : bool
+            Default setting for this argument is True.
+            If True, use the provided separator, writing in a csv format for
```
jreback: @jorisvandenbossche I think we indenting or '-' here to separate these cases?

pandas/core/generic.py
```diff
@@ -1930,25 +1930,52 @@ def to_pickle(self, path, compression='infer',
...
+            allowing easy pasting into excel.
+            If False, write a string representation of the object to the
+            clipboard.
+        sep : str, default tab
+            Field delimiter.
+        kwargs : optional
+            These parameters will be passed to either
+            :py:meth:`pandas.DataFrame.to_csv`
```
jreback: don't need the ':py:' prefix

jreback (changes requested):

pandas/core/generic.py
```diff
@@ -1930,25 +1930,52 @@ def to_pickle(self, path, compression='infer',
...
         Parameters
         ----------
-        excel : boolean, defaults to True
-                if True, use the provided separator, writing in a csv
-                format for allowing easy pasting into excel.
-                if False, write a string representation of the object
-                to the clipboard
-        sep : optional, defaults to tab
-        other keywords are passed to to_csv
+        excel : bool
```
jorisvandenbossche: please leave the default: `bool, default True`

pandas/core/generic.py
jorisvandenbossche: Depends on the context, in this case for True/False I think it is fine as sentences (but if there are more options certainly better to explicitly use lists)

pandas/core/generic.py
```diff
@@ -1930,25 +1930,52 @@ def to_pickle(self, path, compression='infer',
...
+            Default setting for this argument is True.
+            If True, use the provided separator, writing in a csv format for
+            allowing easy pasting into excel.
+            If False, write a string representation of the object to the
+            clipboard.
+        sep : str, default tab
+            Field delimiter.
+        kwargs : optional
```
jorisvandenbossche: `kwargs : optional` -> `**kwargs`

pandas/core/generic.py
TomAugspurger: Default would go on this line. so `bool, default True`.

pandas/core/generic.py
```diff
@@ -1930,25 +1930,52 @@ def to_pickle(self, path, compression='infer',
...
+            These parameters will be passed to either
+            :py:meth:`pandas.DataFrame.to_csv`
+            or :py:meth:`pandas.Series.to_csv` methods depending on the
+            Object type.
```
jorisvandenbossche: Since they actually have the same optional keyword, I would just refer to DataFrame.to_csv to keep it a bit shorter

pandas/core/generic.py
TomAugspurger: Don't think we have a policy. But I think

    - True : use the provided separator,...
    - False : ...

looks nice.

pandas/core/generic.py
```diff
@@ -1930,25 +1930,52 @@ def to_pickle(self, path, compression='infer',
...
+
+        See Also
+        --------
+        pandas.core.frame.DataFrame.to_csv : Write a DataFrame to a
```
TomAugspurger: Just DataFrame.to_csv

jorisvandenbossche: `pandas.core.frame.DataFrame.to_csv` -> `DataFrame.to_csv` (same below)

would also refer to `read_clipboard`

pandas/core/generic.py
```diff
@@ -1930,25 +1930,52 @@ def to_pickle(self, path, compression='infer',
...
+            comma-separated values (csv) file.
+        pandas.core.series.Series.to_csv : Write a Series to a
```
TomAugspurger: Just Series.to_csv

jorisvandenbossche (commented):
Thanks for the PR! Added some more comments

pandas/core/generic.py
```diff
@@ -1930,25 +1930,52 @@ def to_pickle(self, path, compression='infer',
...
+
+        Examples
+        --------
+        Copy the contents of a DataFrame to the clipboard.
+
+        >>> df = pd.DataFrame([[1, 2, 3], [4, 5, 6]], columns=['A', 'B', 'C'])
+        >>> df.to_clipboard()
+
+        We can omit the the index by passing the keyword 'index' and setting
+         it to false.
```
TomAugspurger: Remove the leading space here.

pandas/core/generic.py
```diff
@@ -1930,25 +1930,52 @@ def to_pickle(self, path, compression='infer',
...
+          - OS X : none
+
+        Examples
+        --------
+        Copy the contents of a DataFrame to the clipboard.
+
+        >>> df = pd.DataFrame([[1, 2, 3], [4, 5, 6]], columns=['A', 'B', 'C'])
+        >>> df.to_clipboard()
+
+        We can omit the the index by passing the keyword 'index' and setting
```
TomAugspurger: backticks for "index" instead of quotes

dandxy89:
@jreback ready for a review.

TomAugspurger:
Used the same trick as DtypeWarning for showing the "output" of to_clipboard.

Thanks @dandxy89!

PR #20136, opened 2018-03-10
DOC: update the DataFrame.applymap docstring

orereta:
an unrelated PR):

```
python scripts/validate_docstrings.py pandas.DataFrame.applymap

#################### Docstring (pandas.DataFrame.applymap)  ####################

Apply a function to a Dataframe elementwise.

This method applies a function to a DataFrame that is intended to
operate elementwise, i.e. like doing `map(func, series)` for each
series in the DataFrame.

Parameters
----------
func : function
    Python function, returns a single value from a single value.

Returns
-------
pandas.DataFrame
    A transformed DataFrame.

See also
--------
DataFrame.apply : Apply a function along input axis of DataFrame

Examples
--------
>>> df = pd.DataFrame([[1.,2.], [3.,4.]])
>>> df
     0    1
0  1.0  2.0
1  3.0  4.0
>>> df = df.applymap(lambda x: x**2)
>>> df
     0     1
0  1.0   4.0
1  9.0  16.0

Docstring for "pandas.DataFrame.applymap" correct. :)
```

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -5007,19 +5007,18 @@ class DataFrame(NDFrame):
         """
         Apply a function to a Dataframe elementwise.
 
-        This method applies a function to a DataFrame that is intended to
-        operate elementwise, i.e. like doing `map(func, series)` for each
-        series in the DataFrame.
+        This method applies a function that accepts and returns a scalar
+        to every element of a DataFrame.
 
         Parameters
         ----------
-        func : function
+        func : callable
             Python function, returns a single value from a single value.
 
         Returns
         -------
-        pandas.DataFrame
-            A transformed DataFrame.
+        DataFrame
+            Transformed DataFrame.
 
         See also
         --------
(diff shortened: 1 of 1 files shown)


pandas/core/frame.py
```diff
@@ -5005,39 +5005,41 @@ def apply(self, func, axis=0, broadcast=None, raw=False, reduce=None,
 
     def applymap(self, func):
         """
-        Apply a function to a DataFrame that is intended to operate
-        elementwise, i.e. like doing map(func, series) for each series in the
-        DataFrame
+        Apply a function to a Dataframe elementwise.
```
rth: Maybe a "elementwise function" or a "scalar function", I'm not sure.

pandas/core/frame.py
```diff
@@ -5005,39 +5005,41 @@ def apply(self, func, axis=0, broadcast=None, raw=False, reduce=None,
...
+
+        This method applies a function to a DataFrame that is intended to
+        operate elementwise, i.e. like doing map(func, series) for each series
```
rth: We should probably mark it as code  \``map(func, series)`\`

pandas/core/frame.py
```diff
@@ -5005,39 +5005,41 @@ def apply(self, func, axis=0, broadcast=None, raw=False, reduce=None,
...
-        0  -0.03      1.08       1.28
-        1   0.65      0.83      -1.55
-        2   0.51     -0.88       0.20
+            Python function, returns a single value from a single value.
 
         Returns
         -------
-        applied : DataFrame
+        pandas.DataFrame
+            A transformed DataFrame.
```
rth: Possibly, remove the "A" .

pandas/core/frame.py
```diff
@@ -5005,39 +5005,41 @@ def apply(self, func, axis=0, broadcast=None, raw=False, reduce=None,
...
 
         See also
         --------
-        DataFrame.apply : For operations on rows/columns
+        DataFrame.apply : Apply a function along input axis of DataFrame
+
+        Examples
+        --------
 
+        >>> df = pd.DataFrame(np.random.RandomState(0).randn(3, 3))
```
cmdelatorre: In the [Conventions for the examples](https://python-sprints.github.io/pandas/guide/pandas_docstring.html#conventions-for-the-examples) mentions:
> avoid using data without interpretation, like a matrix of random numbers with columns A, B, C, D… And instead use a meaningful example, which makes it easier to understand the concept

In fact, this one looks like the **Bad** example :-P

I'd suggest improving this example.

cmdelatorre (changes requested):

pandas/core/frame.py
orereta: thanks for the comment! We upgraded the original example to a deterministic one (adding the hardcoded seed in RandomState),  .. any idea to improve the example?

orereta: thanks for the comment, actually we updated previous random example to a deterministic one by hardcoding the random seed through RandomState(0). Since we are illustrating an elementwise mapping it might be easier to use a matrix [[1,2],[3,4]] and square it.

pandas/core/frame.py
orereta: thanks! addressed

TomAugspurger:
Thanks @orereta !

PR #20137, opened 2018-03-10
DOC: update pandas.Series.rename_axis

xrmx:
an unrelated PR):

```
############# Docstring (pandas.core.generic.NDFrame.rename_axis)  #############

Alter the name of the index or columns.

This function alters the name of the index or columns.

Parameters
----------
mapper : scalar, list-like, optional
    Value to set the axis name attribute.
axis : int, str, default 0
    The index or the name of the axis.
copy : boolean, default True
    Also copy underlying data.
    Modifies the mapper in place.

Returns
-------
type of caller or None if inplace=True
    Renamed

Notes
-----
Prior to version 0.21.0, ``rename_axis`` could also be used to change
the axis *labels* by passing a mapping or scalar. This behavior is
deprecated and will be removed in a future version. Use ``rename``
instead.

See Also
--------
pandas.Series.rename : Alter Series index labels or name
pandas.DataFrame.rename : Alter DataFrame index labels or name
pandas.Index.rename : Set new names on index

Examples
--------
>>> df = pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})
>>> df.rename_axis("foo")
     A  B
foo
0    1  4
1    2  5
2    3  6

>>> df.rename_axis("bar", axis="columns")
bar  A  B
0    1  4
1    2  5
2    3  6

Docstring for "pandas.core.generic.NDFrame.rename_axis" correct. :)
```

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -919,18 +919,19 @@ class NDFrame(PandasObject, SelectionMixin):
         Parameters
         ----------
         mapper : scalar, list-like, optional
-            Value to set the axis name attribute.
+            Value to set as the axis name attribute.
         axis : {0 or 'index', 1 or 'columns'}, default 0
             The index or the name of the axis.
         copy : boolean, default True
             Also copy underlying data.
         inplace : boolean, default False
-            Modifies the mapper in place.
+            Modifies the object directly, instead of creating a new Series
+            or DataFrame.
 
         Returns
         -------
         renamed : Series, DataFrame, or None
-            The same type as the caller or None ifa` inplace` is True.
+            The same type as the caller or None if `inplace` is True.
 
         Notes
         -----
@@ -947,6 +948,18 @@ class NDFrame(PandasObject, SelectionMixin):
 
         Examples
         --------
+        **Series**
+
+        >>> s = pd.Series([1, 2, 3])
+        >>> s.rename_axis("foo")
+        foo
+        0    1
+        1    2
+        2    3
+        dtype: int64
+
+        **DataFrame**
+
         >>> df = pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})
         >>> df.rename_axis("foo")
              A  B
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -913,20 +913,26 @@ def f(x):
     rename.__doc__ = _shared_docs['rename']
 
     def rename_axis(self, mapper, axis=0, copy=True, inplace=False):
-        """Alter the name of the index or columns.
+        """
+        Alter the name of the index or columns.
+
+        This function alters the name of the index or columns.
```
TomAugspurger: Since this is just restating the short summary I think it can be removed.

pandas/core/generic.py
```diff
@@ -913,20 +913,26 @@ def f(x):
...
 
         Parameters
         ----------
         mapper : scalar, list-like, optional
             Value to set the axis name attribute.
-        axis : int or string, default 0
+        axis : int, str, default 0
+            The index or the name of the axis.
```
TomAugspurger: Where inconsistent with axis. But

seems to be the most common.

pandas/core/generic.py
```diff
@@ -913,20 +913,26 @@ def f(x):
...
         copy : boolean, default True
-            Also copy underlying data
+            Also copy underlying data.
         inplace : boolean, default False
+            Modifies the mapper in place.
 
         Returns
         -------
-        renamed : type of caller or None if inplace=True
+        type of caller or None if inplace=True
```
TomAugspurger: I think this should be

    renamed : Series, DataFrame, or None
        The same type as the caller or None if `inplace` is True.

pandas/core/generic.py
xrmx: Done.

pandas/core/generic.py
xrmx: Done

pandas/core/generic.py
xrmx: Done

xrmx:
Updated PR after review, thanks!

TomAugspurger:
Added a series example. Thanks @xrmx!

PR #20140, opened 2018-03-10
DOC: improved the docstring of pandas.Index.min()

juanhuguetgarcia:
an unrelated PR):

```
Docstring for "pandas.Index.min" correct. :)
```

explicitly.

### DIFF
--- a/pandas/core/base.py
+++ b/pandas/core/base.py
@@ -812,8 +812,8 @@ class IndexOpsMixin(object):
         See Also
         --------
         Index.max : Return the maximum value of the object.
-        Series.max : Return the maximum value in a Series.
-        DataFrame.max : Return the maximum values in a DataFrame.
+        Series.min : Return the minimum value in a Series.
+        DataFrame.min : Return the minimum values in a DataFrame.
 
         Examples
         --------
--- a/pandas/core/base.py
+++ b/pandas/core/base.py
@@ -802,19 +802,18 @@ class IndexOpsMixin(object):
 
     def min(self):
         """
-        Return the minimum value of the Index object.
-
-        Return the minimum value of the object within the same type.
-        The Index of a DataFrame can be accessed as pandas.DataFrame.index.
+        Return the minimum value of the Index.
 
         Returns
         -------
-        scalar or object
+        scalar
             Minimum value.
 
         See Also
         --------
         Index.max : Return the maximum value of the object.
+        Series.max : Return the maximum value in a Series.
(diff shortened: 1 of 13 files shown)


caperutxa:
Change it

pandas/core/base.py
```diff
@@ -801,7 +801,31 @@ def argmax(self, axis=None):
         return nanops.nanargmax(self.values)
 
     def min(self):
-        """ The minimum value of the object """
+        """
+        Return the minimum value of the Index object.
+
+        Return the maximum value of the object within the same type.
```
arnau126: minimum, not maximum

juanhuguetgarcia: thanks

caperutxa:
Looks fine, as index.max ;)

pandas/core/base.py
```diff
@@ -801,7 +801,31 @@ def argmax(self, axis=None):
...
+        """
+        Return the minimum value of the Index object.
+
+        Return the minimum value of the object within the same type.
+        The Index of a DataFrame can be accessed as pandas.DataFrame.index.
+
+        Returns
+        -------
+        scalar or object
+            minimum value
```
villasv: The description must start with a capital letter, and finish with a dot.

pandas/core/base.py
```diff
@@ -801,7 +801,31 @@ def argmax(self, axis=None):
...
+        """
+        Return the minimum value of the Index object.
+
+        Return the minimum value of the object within the same type.
+        The Index of a DataFrame can be accessed as pandas.DataFrame.index.
+
+        Returns
+        -------
+        Scalar or object.
+            minimum value
```
villasv: Sorry, perhaps I wasn't clear before.
```
scalar or object
            Minimum value.
```

I'll quote the docs:

> The documentation of the return is also similar to the parameters. But in this case, no name will be provided, unless the method returns or yields more than one value (a tuple of values).

> The parameters are defined by their name, followed by a space, a colon, another space, and the type (or types). Note that the space between the name and the colon is important. Types are not defined for *args and **kwargs, but must be defined for all other parameters. After the parameter definition, it is required to have a line with the parameter description, which is indented, and can have multiple lines. The description must start with a capital letter, and finish with a dot.

villasv (changes requested):

pandas/core/base.py
juanhuguetgarcia: I think now should be fine, thanks!

TomAugspurger:
Thanks everyone!

PR #20141, opened 2018-03-10
DOC: Improved the docstring of str.extract() (Delhi)

arjunsharma97:
an unrelated PR):

```
#################### Docstring (pandas.Series.str.extract)  ####################

Return the match object corresponding to regex `pat`.

For each subject string in the Series, extract groups from the
first match of regular expression `pat`.

Parameters
----------
pat : string
    Re module flags, e.g. re.IGNORECASE.
expand : bool, default True
    If True, return DataFrame, else return Series/Index/DataFrame.

    .. versionadded:: 0.18.0.

Returns
-------
DataFrame with one row for each subject string, and one column for
each group. Any capture group names in regular expression pat will
be used for column names; otherwise capture group numbers will be
used. The dtype of each result column is always object, even when
no match is found. If expand=False and pat has only one capture group,
then return a Series (if subject is a Series) or Index (if subject
is an Index).

See Also
--------
extractall : returns all matches (not just the first match)

Examples
--------
A pattern with two groups will return a DataFrame with two columns.
Non-matches will be NaN.

>>> s = pd.Series(['a1', 'b2', 'c3'])
>>> s.str.extract(r'([ab])(\d)')
     0    1
0    a    1
1    b    2
2  NaN  NaN

A pattern may contain optional groups.

>>> s.str.extract(r'([ab])?(\d)')
     0  1
0    a  1
1    b  2
2  NaN  3

Named groups will become column names in the result.

>>> s.str.extract(r'(?P<letter>[ab])(?P<digit>\d)')
  letter digit
0      a     1
1      b     2
2    NaN   NaN

A pattern with one group will return a DataFrame with one column
if expand=True.

>>> s.str.extract(r'[ab](\d)', expand=True)
     0
0    1
1    2
2  NaN

A pattern with one group will return a Series if expand=False.

>>> s.str.extract(r'[ab](\d)', expand=False)
0      1
1      2
2    NaN
dtype: object

Docstring for "pandas.Series.str.extract" correct. :)

```

explicitly.


pandas/core/strings.py
```diff
@@ -633,21 +633,21 @@ def _str_extract_frame(arr, pat, flags=0):
 
 def str_extract(arr, pat, flags=0, expand=True):
     r"""
+    Return the match object corresponding to regex `pat`.
```
rafarui: Is this `r` before the doc string right?

jorisvandenbossche: Yes, it is to ensure correct rendering for used `\` in the examples

pandas/core/strings.py
```diff
@@ -633,21 +633,21 @@ def _str_extract_frame(arr, pat, flags=0):
...
 
     Parameters
     ----------
     pat : string
-        Regular expression pattern with capturing groups
+        Regular expression pattern with capturing groups.
     flags : int, default 0 (no flags)
-        re module flags, e.g. re.IGNORECASE
-
+        Re module flags, e.g. re.IGNORECASE.
```
jorisvandenbossche: In this case it is fine to not capitalize the first word, as it refers to a python module. Therefore, would you actually quote it like ``` ``re`` ``` ? (and the same for `re.IGNORECASE`)

pandas/core/strings.py
```diff
@@ -898,94 +897,23 @@ def str_join(arr, sep):
 
 def str_findall(arr, pat, flags=0):
     """
-    Find all occurrences of pattern or regular expression in the Series/Index.
-
-    Equivalent to applying :func:`re.findall` to all the elements in the
-    Series/Index.
+    Find all occurrences of pattern or regular expression in the
+    Series/Index. Equivalent to :func:`re.findall`.
```
jorisvandenbossche: The summary line should only be one line. Can you try to condense it or split it up?

pandas/core/strings.py
```diff
@@ -898,94 +897,23 @@ def str_join(arr, sep):
...
     Parameters
     ----------
     pat : string
-        Pattern or regular expression.
-    flags : int, default 0
-        ``re`` module flags, e.g. `re.IGNORECASE` (default is 0, which means
-        no flags).
+        Pattern or regular expression
+    flags : int, default 0 (no flags)
+        re module flags, e.g. re.IGNORECASE
```
jorisvandenbossche: similar comment here as above

pandas/core/strings.py
```diff
@@ -898,94 +897,23 @@ def str_join(arr, sep):
...
-        of pattern or regular expression in string, as a list of strings.
-
-    Examples
-    --------
-
-    >>> s = pd.Series(['Lion', 'Monkey', 'Rabbit'])
-
-    The search for the pattern 'Monkey' returns one match:
-
-    >>> s.str.findall('Monkey')
```
jorisvandenbossche: Is there a reason you removed those examples ?

jorisvandenbossche (commented):

pandas/core/strings.py
TomAugspurger: This description isn't quite right, is it? We aren't returning a `Match` object.

Maybe

    Extract capture groups in the regex `pat` as columns in a DataFrame.

pandas/core/strings.py
```diff
@@ -693,21 +693,21 @@ def _str_extract_frame(arr, pat, flags=0):
...
 
     Parameters
     ----------
     pat : string
-        Regular expression pattern with capturing groups
+        Regular expression pattern with capturing groups.
     flags : int, default 0 (no flags)
-        re module flags, e.g. re.IGNORECASE
-
+        ``re`` module flags, e.g. ``re.IGNORECASE``.
```
TomAugspurger: Add a llink to https://docs.python.org/3/library/re.html#contents-of-module-re ?

pandas/core/strings.py
```diff
@@ -693,21 +693,21 @@ def _str_extract_frame(arr, pat, flags=0):
...
     expand : bool, default True
-        * If True, return DataFrame.
-        * If False, return Series/Index/DataFrame.
+        If True, return DataFrame, else return Series/Index/DataFrame.
```
TomAugspurger: When is a DataFrame returned if `expand=False`?

pandas/core/strings.py
jorisvandenbossche: I think this can be done by using `` :mod:`re` ``

pandas/core/strings.py
jorisvandenbossche: It still returns a DataFrame when you have multiple groups (only if there is only one group, it gives a Series). Probably good to explain this a bit more.

(I find this a bit strange behaviour, but that's what it currently is. In other case we then return a Series with lists I think)

PR #20145, opened 2018-03-10
DOC: Added examples to the IndexOpsmixin.value_counts() docstring

marcosrullan:
an unrelated PR):

```

###############################################################################
#################### Docstring (pandas.Index.value_counts)  ####################

Return object containing counts of unique values.

The resulting object will be in descending order so that the
first element is the most frequently-occurring element.
Excludes NA values by default.

Parameters
----------
normalize : boolean, default False
    If True then the object returned will contain the relative
    frequencies of the unique values.
sort : boolean, default True
    Sort by values.
ascending : boolean, default False
    Sort in ascending order.
bins : integer, optional
    Rather than count values, group them into half-open bins,
    a convenience for pd.cut, only works with numeric data.
dropna : boolean, default True
    Don't include counts of NaN.

Returns
-------
counts : Series

Examples
--------

>>> index = pd.Index([3, 1, 2, 3, 4, np.nan])
>>> index.value_counts()
3.0    2
4.0    1
2.0    1
1.0    1
dtype: int64

With `normalize` set to `True`, returns the relative frequency by
dividing all values by the sum of values.

>>> s = pd.Series([3, 1, 2, 3, 4, np.nan])
>>> s.value_counts(normalize=True)
3.0    0.4
4.0    0.2
2.0    0.2
1.0    0.2
dtype: float64

**bins**

Bins can be useful for going from a continuous variable to a
categorical variable; instead of counting unique
apparitions of values, divide the index in the specified
number of half-open bins.

>>> s.value_counts(bins=3)
(2.0, 3.0]      2
(0.996, 2.0]    2
(3.0, 4.0]      1
dtype: int64

**dropna**

With `dropna` set to `False` we can also see NaN index values.

>>> s.value_counts(dropna=False)
3.0    2
NaN     1
4.0    1
2.0    1
1.0    1
dtype: int64

Errors found:
	See Also section not found

```
The validation is not passing because we lack an see also section, but we think that is not relevant for this method.


pandas/core/base.py
```diff
@@ -943,18 +943,65 @@ def value_counts(self, normalize=False, sort=True, ascending=False,
...
         dropna : boolean, default True
             Don't include counts of NaN.
 
         Returns
         -------
         counts : Series
+
+        Examples
+        --------
+
```
cemsbr: I've noticed there's no blank line after section headers in https://python-sprints.github.io/pandas/guide/pandas_docstring.html

cemsbr (changes requested):

caperutxa:
Pa que tocas

pandas/core/base.py
marcosrullan: Thanks, I have removed the unnecessary new line!

cemsbr (approved):

jreback:
thanks @marcosrullan and @mroeschke for the fixups!

PR #20149, opened 2018-03-10
DOC: update the to_json() docstring

robmarkcole:
Fixes validation errors and adds an example for each orientation.

- [ x] PR title is "DOC: update the <your-function-or-method> docstring"
- [ x] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [ x] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x ] The html version looks good: `python doc/make.py --single <your-function-or-method>`

```
Errors found:
	Missing description for See Also "pandas.read_json" reference
	Examples do not pass tests
```

I'm not able to make this pass.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1694,7 +1694,7 @@ class NDFrame(PandasObject, SelectionMixin):
                 describing the data, and the data component is
                 like ``orient='records'``.
 
-                .. versionchanged:: 0.20.0.
+                .. versionchanged:: 0.20.0
 
         date_format : {None, 'epoch', 'iso'}
             Type of date conversion. 'epoch' = epoch milliseconds,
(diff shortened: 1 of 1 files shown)


robmarkcole:
@samuelsinayoko
@gioiab

pandas/core/generic.py
```diff
@@ -1665,8 +1665,9 @@ def to_json(self, path_or_buf=None, orient=None, date_format=None,
         Parameters
         ----------
         path_or_buf : the path or buffer to write the result string
```
jorisvandenbossche: The "the path or buffer to write the result string" is actually the explanation, so should go on the next line. On this line comes the type description, you can take a look at eg DataFrame.to_csv how this is typically explained.

jorisvandenbossche (commented):
Thanks for the PR!

Added an inline comment. And two additional ones:

- In the See Also section, can you change `pd.read_json` to `pandas.read_json` ?
- In the "Parameters" section, in the description of the 'orient', in the bullet point "The format of the JSON string", can you quote the options? (so eg `- 'split': ...` instead of `- split: ...`), to make it clear it are strings that needs to be passed.

pandas/core/generic.py
robmarkcole: Cheers

robmarkcole (commented):
@jorisvandenbossche `pandas.read_json` also gives an error I'm afraid, but it is consistent with the other docstrings

jorisvandenbossche (approved):
added a small edit of some quotes, and removed the return statement (was already there, but actually wrong)

jorisvandenbossche:
@robmarkcole Thanks for the PR!

PR #20150, opened 2018-03-10
DOC: update the pandas.Series.dt.is_leap_year docstring

rok:
an unrelated PR):

- [x ] PR title is "DOC: update the <your-function-or-method> docstring"
- [x ] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [x ] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x ] The html version looks good: `python doc/make.py --single <your-function-or-method>`

```
################## Docstring (pandas.Series.dt.is_leap_year)  ##################

Return a boolean indicating if the date belongs to a leap year.

A leap year is a year, occurring every four years, which has 366 days
(instead of 365) including 29th of February as an intercalary day.

Returns
-------
is_leap_year : Series of boolean

Examples
--------
>>> import pandas as pd
>>> dates = pd.date_range("2012-01-01", "2015-01-01", freq="Y")
>>> dates_series = pd.Series(dates)
>>> dates_series
0   2012-12-31
1   2013-12-31
2   2014-12-31
dtype: datetime64[ns]
>>> dates_series.dt.is_leap_year
0     True
1    False
2    False
dtype: bool

Errors found:
	See Also section not found
```

explicitly.

See Also is not relevant for this function as there is no other leap year related functions.


pandas/core/indexes/datetimes.py
```diff
@@ -1753,7 +1753,32 @@ def freq(self, value):
...
-        "Logical indicating if the date belongs to a leap year")
+        """
+        Return a boolean indicating if the date belongs to a leap year.
+
+        A leap year is a year, occurring every four years, which has 366 days
+        (instead of 365) including 29th of February as an intercalary day.
+
+        Returns
+        -------
+        is_leap_year : Series of boolean
```
villasv: There's no need to name the return value, only it's type. And it's missing the description.

> The documentation of the return is also similar to the parameters. But in this case, no name will be provided, unless the method returns or yields more than one value (a tuple of values).

> The parameters are defined by their name, followed by a space, a colon, another space, and the type (or types). Note that the space between the name and the colon is important. Types are not defined for *args and **kwargs, but must be defined for all other parameters. After the parameter definition, it is required to have a line with the parameter description, which is indented, and can have multiple lines. The description must start with a capital letter, and finish with a dot.

villasv (changes requested):

pandas/core/indexes/datetimes.py
```diff
@@ -1753,7 +1753,32 @@ def freq(self, value):
     is_leap_year = _field_accessor(
         'is_leap_year',
         'is_leap_year',
-        "Logical indicating if the date belongs to a leap year")
+        """
+        Return a boolean indicating if the date belongs to a leap year.
+
+        A leap year is a year, occurring every four years, which has 366 days
```
amuta: There are exceptions for this four year rule, years divisible by 100 are not leap years except if they are also divisible by 400.
From wiki:
   "Every year that is exactly divisible by four is a leap year, except for years that are exactly divisible by 100, but these centurial years are leap years if they are exactly divisible by 400. For example, the years 1700, 1800, and 1900 were not leap years, but the years 1600 and 2000 were."

amuta (changes requested):

pandas/core/indexes/datetimes.py
```diff
@@ -1753,7 +1753,32 @@ def freq(self, value):
...
+        (instead of 365) including 29th of February as an intercalary day.
+
+        Returns
+        -------
+        is_leap_year : Series of boolean
+
+        Examples
+        --------
+        >>> import pandas as pd
+        >>> dates = pd.date_range("2012-01-01", "2015-01-01", freq="Y")
```
jazzmuesli: no need to import pandas or numpy

jorisvandenbossche (approved):
Made a small edit to make it generic for Index/Series (the docstring is shared between Series.dt and DatetimeIndex)

jorisvandenbossche:
Thanks all!

PR #20154, opened 2018-03-10
DOC: Update the pandas.Series.str.count() docstring (Delhi)

Ridhwanluthra:
an unrelated PR):

```
##################### Docstring (pandas.Series.str.count)  #####################

Count occurrences of pattern in each string of the Series/Index.

This function is used to count the number of times a particular regex
pattern is repeated in each of the string elements of the
:class:`~pandas.Series`.

Parameters
----------
pat : str
    Valid regular expression.
flags : int, default 0, meaning no flags
    Flags for re module, e.g. re.IGNORECASE.

Returns
-------
counts : Series/Index of integer values

Notes
-----
Some characters need to be escaped when passing in pat.
eg. '$' has a special meaning in regex and must be escaped when finding
specifically this char.

Examples
--------
Take a look at
`this link <https://docs.python.org/3/howto/regex.html#compilation-flags>`_
for the list of all possible flags that can be used.

>>> s = pd.Series(['A', 'B', 'Aaba', 'Baca', np.nan, 'CABA', 'cat'])
>>> s.str.count('a')
0    0.0
1    0.0
2    2.0
3    2.0
4    NaN
5    0.0
6    1.0
dtype: float64

Errors found:
	Errors in parameters section
	See Also section not found

```
As discussed in the gitter channel it was suggested to ignore kwargs for now.
Not sure what should be a part of see also here.

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -202,15 +202,43 @@ def str_count(arr, pat, flags=0):
     """
     Count occurrences of pattern in each string of the Series/Index.
 
+    This function is used to count the number of times a particular regex
+    pattern is repeated in each of the string elements of the
+    :class:`~pandas.Series`.
+
     Parameters
     ----------
-    pat : string, valid regular expression
-    flags : int, default 0 (no flags)
-        re module flags, e.g. re.IGNORECASE
+    pat : str
+        Valid regular expression.
+    flags : int, default 0, meaning no flags
+        Flags for re module, e.g. re.IGNORECASE.
 
     Returns
     -------
     counts : Series/Index of integer values
+
+    Notes
+    -----
+    Some characters need to be escaped when passing in pat.
+    eg. '$' has a special meaning in regex and must be escaped when finding
+    specifinally this char.
+
+    Examples
+    --------
+    Take a look at
+    `this link <https://docs.python.org/3/howto/regex.html#compilation-flags>`_ 
+    for the list of all possible flags that can be used.
+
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -202,15 +202,43 @@ def str_count(arr, pat, flags=0):
...
+
+    Notes
+    -----
+    Some characters need to be escaped when passing in pat.
+    eg. '$' has a special meaning in regex and must be escaped when finding
+    specifically this char.
+
+    Examples
+    --------
+    Take a look at
```
TomAugspurger: Maybe move this link to be under the flags parameter?

pandas/core/strings.py
```diff
@@ -202,15 +202,43 @@ def str_count(arr, pat, flags=0):
...
+    >>> s = pd.Series(['A', 'B', 'Aaba', 'Baca', np.nan, 'CABA', 'cat'])
+    >>> s.str.count('a')
+    0    0.0
+    1    0.0
+    2    2.0
+    3    2.0
+    4    NaN
+    5    0.0
+    6    1.0
+    dtype: float64
```
TomAugspurger: Add an examle with special characters like `$`?

Ridhwanluthra: should I remove the note in the doc?

pandas/core/strings.py
Ridhwanluthra: I like that idea, will do

Ridhwanluthra:
@TomAugspurger please take a look at the update

TomAugspurger:
Thanks!

PR #20155, opened 2018-03-10
DOC: update the pandas.DataFrame.plot.hist docstring

liopic:
an unrelated PR):

```
#################### Docstring (pandas.DataFrame.plot.hist) ####################

Draw one histogram of the DataFrame's Series using matplotlib.

A histogram is a representation of the distribution of data.
This function groups the values of all given Series in the DataFrame
into bins, and draws all bins in only one matplotlib.AxesSubplot. This
is useful when the DataFrame's Series are in a similar scale.

Parameters
----------
by : str
    Column in the DataFrame to group by.
bins : int, default 10
    Number of histogram bins to be used.
**kwds : optional
    Parameters to pass on to :meth:`pandas.DataFrame.plot`.

Returns
-------

See Also
--------
:meth:`pandas.DataFrame.hist` : Draw histograms per DataFrame's Series.
:meth:`pandas.Series.hist` : Draw a histogram with Series' data.

Examples
--------

When using values between 0 and 3, calling hist() with bins = 3 will
create three bins: one that groups values between 0 and 1, another
for values between 1 and 2, and another for values between 2 and 3.
We use alpha parameter to be able to see overlapping columns.

.. plot::
    :context: close-figs

    >>> df = pd.DataFrame({
    ...     'length': [ 1.5, 0.5, 1.2, 0.9, 3],
    ...     'width': [ 0.7, 0.2, 0.15, 0.2,  1.1]
    ...     }, index= ['pig', 'rabbit', 'duck', 'chicken', 'horse'])
    >>> hist = df.plot.hist(bins = 3, xticks = range(4), alpha = 0.5)

Docstring for "pandas.DataFrame.plot.hist" correct. :)


```

### DIFF
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2981,7 +2981,8 @@ class FramePlotMethods(BasePlotMethods):
         --------
         When we draw a dice 6000 times, we expect to get each value around 1000
         times. But when we draw two dices and sum the result, the distribution
-        is going to be quite different. Let's display it.
+        is going to be quite different. A histogram illustrates those
+        distributions.
 
         .. plot::
             :context: close-figs
@@ -2990,7 +2991,7 @@ class FramePlotMethods(BasePlotMethods):
             ...     np.random.randint(1, 7, 6000),
             ...     columns = ['one'])
             >>> df['two'] = df['one'] + np.random.randint(1, 7, 6000)
-            >>> hist = df.plot.hist(bins = 12, alpha = 0.5)
+            >>> ax = df.plot.hist(bins=12, alpha=0.5)
         """
         return self(kind='hist', by=by, bins=bins, **kwds)
 
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2979,19 +2979,18 @@ class FramePlotMethods(BasePlotMethods):
 
         Examples
         --------
-        When using values between 0 and 3, calling hist() with bins = 3 will
-        create three bins: one that groups values between 0 and 1, another
-        for values between 1 and 2, and another for values between 2 and 3.
-        We use alpha parameter to be able to see overlapping columns.
+        When we draw a dice 6000 times, we expect to get each value around 1000
+        times. But when we draw two dices and sum the result, the distribution
+        is going to be quite different. Let's display it.
 
         .. plot::
             :context: close-figs
 
-            >>> df = pd.DataFrame({
-            ...     'length': [ 1.5, 0.5, 1.2, 0.9, 3],
-            ...     'width': [ 0.7, 0.2, 0.15, 0.2,  1.1]
-            ...     }, index = ['pig', 'rabbit', 'duck', 'chicken', 'horse'])
-            >>> hist = df.plot.hist(bins = 3, xticks = range(4), alpha = 0.5)
+            >>> df = pd.DataFrame(
+            ...     np.random.randint(1, 7, 6000),
+            ...     columns = ['one'])
+            >>> df['two'] = df['one'] + np.random.randint(1, 7, 6000)
+            >>> hist = df.plot.hist(bins = 12, alpha = 0.5)
         """
         return self(kind='hist', by=by, bins=bins, **kwds)
 
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2955,8 +2955,8 @@ class FramePlotMethods(BasePlotMethods):
 
         A histogram is a representation of the distribution of data.
         This function groups the values of all given Series in the DataFrame
-        into bins, and draws all bins in only one matplotlib.AxesSubplot. This
-        is useful when the DataFrame's Series are in a similar scale.
+        into bins, and draws all bins in only one :ref:`matplotlib.axes.Axes`.
+        This is useful when the DataFrame's Series are in a similar scale.
 
         Parameters
         ----------
@@ -2964,7 +2964,7 @@ class FramePlotMethods(BasePlotMethods):
             Column in the DataFrame to group by.
         bins : int, default 10
             Number of histogram bins to be used.
-        **kwds : optional
+        **kwds
             Additional keyword arguments are documented in
             :meth:`pandas.DataFrame.plot`.
 
@@ -2974,8 +2974,8 @@ class FramePlotMethods(BasePlotMethods):
 
         See Also
         --------
-        pandas.DataFrame.hist : Draw histograms per DataFrame's Series.
-        pandas.Series.hist : Draw a histogram with Series' data.
+        DataFrame.hist : Draw histograms per DataFrame's Series.
+        Series.hist : Draw a histogram with Series' data.
 
         Examples
         --------
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2786,11 +2786,11 @@ class FramePlotMethods(BasePlotMethods):
         bins : int, default 10
             Number of histogram bins to be used.
         **kwds : optional
-            Parameters to pass on to :meth:`pandas.DataFrame.plot`.
+            Keyword arguments to pass on to :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : matplotlib.AxesSubplot or np.array of them.
+        axes : matplotlib.AxesSubplot histogram.
 
         See Also
         --------
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2799,7 +2799,6 @@ class FramePlotMethods(BasePlotMethods):
 
         Examples
         --------
-
         When using values between 0 and 3, calling hist() with bins = 3 will
         create three bins: one that groups values between 0 and 1, another
         for values between 1 and 2, and another for values between 2 and 3.
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2951,7 +2951,7 @@ class FramePlotMethods(BasePlotMethods):
 
     def hist(self, by=None, bins=10, **kwds):
         """
-        Draw one histogram of the DataFrame's Series using matplotlib.
+        Draw one histogram of the DataFrame's columns.
 
         A histogram is a representation of the distribution of data.
         This function groups the values of all given Series in the DataFrame
@@ -2960,7 +2960,7 @@ class FramePlotMethods(BasePlotMethods):
 
         Parameters
         ----------
-        by : str or sequence
+        by : str or sequence, optional
             Column in the DataFrame to group by.
         bins : int, default 10
             Number of histogram bins to be used.
@@ -2973,8 +2973,8 @@ class FramePlotMethods(BasePlotMethods):
 
         See Also
         --------
-        :meth:`pandas.DataFrame.hist` : Draw histograms per DataFrame's Series.
-        :meth:`pandas.Series.hist` : Draw a histogram with Series' data.
+        pandas.DataFrame.hist : Draw histograms per DataFrame's Series.
+        pandas.Series.hist : Draw a histogram with Series' data.
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


igorcadelima:
Please. remove the blank line after
```
Examples
--------
```

igorcadelima (changes requested):
Please, remove the blank line after
```
Examples
--------
```
For the comment on `**kwds`, remove the backticks around **kwds and replace the description with "Keyword arguments to pass on to :meth:`pandas.DataFrame.plot`.".

For the "Returns" section, indicate the type the method returns and provide a clear and brief explanation. Avoid the use of "them" if not referring to the subject of the sentence.

pandas/plotting/_core.py
```diff
@@ -2772,20 +2772,46 @@ def box(self, by=None, **kwds):
...
+
+        A histogram is a representation of the distribution of data.
+        This function groups the values of all given Series in the DataFrame
+        into bins, and draws all bins in only one matplotlib.AxesSubplot. This
+        is useful when the DataFrame's Series are in a similar scale.
 
         Parameters
         ----------
-        by : string or sequence
+        by : str
```
mabelvj: Why removing sequence? A sequence of string can be passed to select the columns to be plotted.

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,46 @@ def box(self, by=None, **kwds):
 
     def hist(self, by=None, bins=10, **kwds):
         """
-        Histogram
+        Draw one histogram of the DataFrame's Series using matplotlib.
```
jorisvandenbossche: "the DataFrame's Series" -> "the DataFrame's columns" ?

The "using matplotlib" is not needed in this summary line, as all our plotting machinery is based on matplotlb

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,46 @@ def box(self, by=None, **kwds):
...
+
+        A histogram is a representation of the distribution of data.
+        This function groups the values of all given Series in the DataFrame
+        into bins, and draws all bins in only one matplotlib.AxesSubplot. This
+        is useful when the DataFrame's Series are in a similar scale.
 
         Parameters
         ----------
-        by : string or sequence
+        by : str or sequence
```
jorisvandenbossche: can you add `, optional`

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,46 @@ def box(self, by=None, **kwds):
...
             Column in the DataFrame to group by.
```
jorisvandenbossche: Would be good to indicate how this affects the resulting plot

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,46 @@ def box(self, by=None, **kwds):
...
-            :meth:`pandas.DataFrame.plot`.
+        bins : int, default 10
+            Number of histogram bins to be used.
+        **kwds : optional
+            Keyword arguments to pass on to :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : :class:`matplotlib.axes.Axes` or numpy.ndarray of them
+        axes : matplotlib.AxesSubplot histogram.
```
jorisvandenbossche: The "or ndarray of them" was not correct?

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,46 @@ def box(self, by=None, **kwds):
...
+
+        See Also
+        --------
+        :meth:`pandas.DataFrame.hist` : Draw histograms per DataFrame's Series.
+        :meth:`pandas.Series.hist` : Draw a histogram with Series' data.
```
jorisvandenbossche: you don't need the `` :meth`..` `` part here, as sphinx will make those links automatically

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,46 @@ def box(self, by=None, **kwds):
...
+        We use alpha parameter to be able to see overlapping columns.
+
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({
+            ...     'length': [ 1.5, 0.5, 1.2, 0.9, 3],
+            ...     'width': [ 0.7, 0.2, 0.15, 0.2,  1.1]
+            ...     }, index= ['pig', 'rabbit', 'duck', 'chicken', 'horse'])
+            >>> hist = df.plot.hist(bins = 3, xticks = range(4), alpha = 0.5)
```
jorisvandenbossche: inside the function call, no spaces around `=` for PEP8

jorisvandenbossche (commented):

pandas/plotting/_core.py
liopic: Actually this parameter is passed on to DataFrame.plot(), which is passed on to matplotlib's plot() later... but apparently it's not used in any place at all. I tried with different examples, but in all cases this parameter is ignored. So I'm keeping the old documentation just for the sake of it.

pandas/plotting/_core.py
liopic: No, it was not correct. This function always returns one matplotlib.AxesSubplot. It's different than DataFrame.hist(), which returns an ndarray.

pandas/plotting/_core.py
jorisvandenbossche: Hmm, yes, you seem to be correct. Will open an issue to discuss in general.

pandas/plotting/_core.py
jorisvandenbossche: Just looked, and the key to get this is passing `subplots=True`, then the different columns are each plotted in a subplot in instead of in overlay, en then you get an ndarray of axes

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,46 @@ def box(self, by=None, **kwds):
...
             Column in the DataFrame to group by.
-        bins: integer, default 10
-            Number of histogram bins to be used
-        `**kwds` : optional
-            Additional keyword arguments are documented in
-            :meth:`pandas.DataFrame.plot`.
+        bins : int, default 10
+            Number of histogram bins to be used.
+        **kwds : optional
+            Keyword arguments to pass on to :meth:`pandas.DataFrame.plot`.
```
jorisvandenbossche: Can you keep the original sentence here? (it was originally how you have it now, but we changed it in one of the previous PRs to say that ""Additional keywords are documented in DataFrame.plot"

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,46 @@ def box(self, by=None, **kwds):
...
+        We use alpha parameter to be able to see overlapping columns.
+
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({
+            ...     'length': [ 1.5, 0.5, 1.2, 0.9, 3],
+            ...     'width': [ 0.7, 0.2, 0.15, 0.2,  1.1]
+            ...     }, index = ['pig', 'rabbit', 'duck', 'chicken', 'horse'])
+            >>> hist = df.plot.hist(bins = 3, xticks = range(4), alpha = 0.5)
```
jorisvandenbossche: Although I like the fact you used some more "real world" data, the problem is that with only 5 values, the histrogram will look rather strange. So I would maybe still generate some more random data in this case.

jorisvandenbossche (commented):
Thanks for the updates! Added few more comments, almost there!

pandas/plotting/_core.py
jorisvandenbossche: @dukebody ah, yes, thanks for linking that

pandas/plotting/_core.py
dukebody: @jorisvandenbossche We will have to revise this after all the sprint PRs have been merged, since people have been changing this and it might not be consistent anymore. But it will be fast and easy.

See https://github.com/pandas-dev/pandas/issues/20348

pandas/plotting/_core.py
liopic: I see. However this is actually a parameter that is passed on to the plot() function, but it's not part of the default plot.hist() parameters... so IMHO it's ok to say that the return will be just one. Don't you think so?

pandas/plotting/_core.py
liopic: During the sprint some of us actually discussed about it: if you generate random data, each time the histogram will be different, and it will be kind of difficult to understand and to explain. So I tried to create some "real" data that can be small enough to generate a clear histogram that shows how it works. But if you think this data is not enough, I could try to improve it.

pandas/plotting/_core.py
jorisvandenbossche: Yes, we should still document it here, since this is what the return value of this function can be. 

(btw, you are welcome to add `subplots` to the list of Parameters, we have a general issue about that we should document more of those common parameters in each of the methods)

pandas/plotting/_core.py
jorisvandenbossche: If you want reproducible data, you can set a random seed. Also, if you use quite some data eg from a normal distrubution, the plot will look very similar each time.

I plotted the data above, and indeed I think it is too small, IMO it gives not a very good "sense" of what a good/useful histogram looks like.

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,47 @@ def box(self, by=None, **kwds):
 
     def hist(self, by=None, bins=10, **kwds):
         """
-        Histogram
+        Draw one histogram of the DataFrame's columns.
+
+        A histogram is a representation of the distribution of data.
+        This function groups the values of all given Series in the DataFrame
+        into bins, and draws all bins in only one matplotlib.AxesSubplot. This
```
TomAugspurger: `matplotlib.AxesSubplot` isn't a think. I'd say either "matplotlib axes" or :ref:\`matplotlib.axes.Axes\`.

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,47 @@ def box(self, by=None, **kwds):
...
         ----------
-        by : string or sequence
+        by : str or sequence, optional
             Column in the DataFrame to group by.
-        bins: integer, default 10
-            Number of histogram bins to be used
-        `**kwds` : optional
+        bins : int, default 10
+            Number of histogram bins to be used.
+        **kwds : optional
```
TomAugspurger: Remove ` : optional`.

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,47 @@ def box(self, by=None, **kwds):
...
             :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : :class:`matplotlib.axes.Axes` or numpy.ndarray of them
+        axes : matplotlib.AxesSubplot histogram.
+
+        See Also
+        --------
+        pandas.DataFrame.hist : Draw histograms per DataFrame's Series.
```
TomAugspurger: remove `pandas.`

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,47 @@ def box(self, by=None, **kwds):
...
+        pandas.Series.hist : Draw a histogram with Series' data.
```
TomAugspurger: Remove `pandas.`

pandas/plotting/_core.py
```diff
@@ -2951,21 +2951,47 @@ def box(self, by=None, **kwds):
...
+
```
TomAugspurger: Add `DataFrame.plot`

pandas/plotting/_core.py
TomAugspurger: +1 for using random data, as it imparts when histograms are useful (summarizing the distribution of a larger dataset).

liopic: I have been thinking about it but I can't imagine an easy way to generate random data that, at the same time, it's __easy to understand__ in a compact form (this is just an Example section) and also shows the __direct relation__ with the resulting histogram. Could you tell me any suggestion?

On the other hand, in pandas.DataFrame.hist() docstring they used another "toy" example (which actually was my work too) and it's already merged...

jorisvandenbossche:
```
In [64]: df = pd.DataFrame({'a': np.random.randn(1000), 'b': np.random.randn(1000)*1.5 + 2})

In [65]: df.plot.hist(alpha=0.5)
Out[65]: <matplotlib.axes._subplots.AxesSubplot at 0x7fbaadcea6d8>
```

gives me

figure_1-1

which I think is a reasonable plot to show functionality of hist ?

> On the other hand, in pandas.DataFrame.hist() docstring they used another "toy" example (which actually was my work too) and it's already merged...

That's true, but IMO we should change it there as well.

liopic: Ok, I found a random example that sounds quite natural and understandable. I hope you'll like it.

jorisvandenbossche (approved):
Thanks, nice example!

(made a small amendment mainly for PEP8 fixup)

jorisvandenbossche:
@liopic Thanks for the PR :+1:

PR #20159, opened 2018-03-10
DOC: update the pandas.Series.dt.total_seconds docstring

ijmbarr:
an unrelated PR):

```
################## Docstring (pandas.Series.dt.total_seconds) ##################

Return total duration of each element expressed in seconds.

Return a series with the same length and index as the original,
containing the length of each element expressed in seconds.

Returns
-------
s : pandas.Series
    a series of type `float64`.

Examples
--------
>>> s = pd.Series(pd.to_timedelta(np.arange(5), unit='d'))
>>> s
0   0 days
1   1 days
2   2 days
3   3 days
4   4 days
dtype: timedelta64[ns]
>>> s.dt.total_seconds()
0         0.0
1     86400.0
2    172800.0
3    259200.0
4    345600.0
dtype: float64

Errors found:
        Errors in parameters section
        See Also section not found

```

explicitly.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/indexes/timedeltas.py
+++ b/pandas/core/indexes/timedeltas.py
@@ -501,7 +501,33 @@ class TimedeltaIndex(DatetimeIndexOpsMixin, TimelikeOps, Int64Index):
 
     def total_seconds(self):
         """
-        Total duration of each element expressed in seconds.
+        Return total duration of each element expressed in seconds.
+
+        Return a series with the same length and index as the original,
+        containing the length of each element expressed in seconds.
+
+        Returns
+        -------
+        s : pandas.Series
+            a series of type `float64`.
+
+        Examples
+        --------
+        >>> s = pd.Series(pd.to_timedelta(np.arange(5), unit='d'))
+        >>> s
+        0   0 days
+        1   1 days
+        2   2 days
+        3   3 days
+        4   4 days
+        dtype: timedelta64[ns]
+        >>> s.dt.total_seconds()
+        0         0.0
+        1     86400.0
+        2    172800.0
+        3    259200.0
+        4    345600.0
+        dtype: float64
         """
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/timedeltas.py
```diff
@@ -500,7 +500,33 @@ def f(x):
...
+        Return total duration of each element expressed in seconds.
+
+        Return a series with the same length and index as the original,
+        containing the length of each element expressed in seconds.
+
+        Returns
+        -------
+        s : pandas.Series
+            a series of type `float64`.
+
```
mabelvj: `See Also`section is missing. Maybe it could be useful to reference [`datetime` documentation](https://docs.python.org/3/library/datetime.html#datetime.timedelta.total_seconds).

ijmbarr: Added.

pandas/core/indexes/timedeltas.py
```diff
@@ -500,7 +500,37 @@ def f(x):
...
-        Total duration of each element expressed in seconds.
+        Return total duration of each element expressed in seconds.
+
+        Return a series with the same length and index as the original,
+        containing the length of each element expressed in seconds.
+
+        Returns
+        -------
+        s : pandas.Series
+            a series of type `float64`.
```
jorisvandenbossche: "a series" -> "Series"

jorisvandenbossche (commented):
Looks good to me! (added minor comment)

@TomAugspurger we have again the problem of index vs .dt accessor. Did we already have a solution for this? (the current PR now documents it like .dt)

TomAugspurger:
Made this generic for TimedeltaIndex and Series[timedelta64].

jorisvandenbossche:
Yes, that looks like a good solution

PR #20163, opened 2018-03-10
DOC: update the DataFrame.plot.line docstring

ifosch:
an unrelated PR):

```❯ python scripts/validate_docstrings.py pandas.DataFrame.plot.line

#################### Docstring (pandas.DataFrame.plot.line) ####################

Plot DataFrame columns as lines.

This function is useful to plot lines using DataFrame's values
as coordinates.

Parameters
----------
x : int or str, optional
    Columns to use for the horizontal axis.
    Either the location or the label of the columns to be used.
    By default, it will use the `DataFrame` indices.
y : int, str, or list of them, optional
    The values to be plotted.
    Either the location or the label of the columns to be used.
    By default, it will use the remaining DataFrame numeric columns.
kwds : optional
    Keyword arguments to pass on to :py:meth:pandas.DataFrame.plot.

Returns
-------
axes : :class:matplotlib.AxesSubplot or :class:numpy.ndarray of
       them.

See Also
--------
matplotlib.pyplot.plot : Plot y versus x as lines and/or markers.

Examples
--------

.. plot::
    :context: close-figs

    The following example shows the populations for some animals
    over the years.

    >>> df = pd.DataFrame({
    ...    'pig': [20, 18, 489, 675, 1776],
    ...    'horse': [4, 25, 281, 600, 1900]
    ...    }, index=[1990, 1997, 2003, 2009, 2014])
    >>> lines = df.plot.line()

.. plot::
    :context: close-figs

    The following example shows the relationship between both
    populations.

    >>> lines = df.plot.line(x='pig', y='horse')

Docstring for "pandas.DataFrame.plot.line" correct. :)
```

explicitly.

### DIFF
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2704,18 +2704,55 @@ class FramePlotMethods(BasePlotMethods):
 
     def line(self, x=None, y=None, **kwds):
         """
-        Line plot
+        Plot DataFrame columns as lines.
+
+        This function is useful to plot lines using DataFrame's values
+        as coordinates.
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
-        `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+        x : int or str, optional
+            Columns to use for the horizontal axis.
+            Either the location or the label of the columns to be used.
+            By default, it will use the `DataFrame` indices.
+        y : int, str, or list of them, optional
+            The values to be plotted.
+            Either the location or the label of the columns to be used.
+            By default, it will use the remaining DataFrame numeric columns.
+        kwds : optional
+            Keyword arguments to pass on to :py:meth:pandas.DataFrame.plot.
 
         Returns
         -------
-        axes : matplotlib.AxesSubplot or np.array of them
+        axes : :class:matplotlib.AxesSubplot or :class:numpy.ndarray of
+               them.
+
+        See Also
+        --------
+        matplotlib.pyplot.plot : Plot y versus x as lines and/or markers.
+
+        Examples
+        --------
+
+        .. plot::
+            :context: close-figs
+
+            The following example shows the populations for some animals
+            over the years.
+
+            >>> df = pd.DataFrame({
+            ...    'pig': [20, 18, 489, 675, 1776],
+            ...    'horse': [4, 25, 281, 600, 1900]
+            ...    }, index=[1990, 1997, 2003, 2009, 2014])
+            >>> lines = df.plot.line()
+
+        .. plot::
+            :context: close-figs
+
+            The following example shows the relationship between both
+            populations.
+
+            >>> lines = df.plot.line(x='pig', y='horse')
         """
         return self(kind='line', x=x, y=y, **kwds)
 
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2787,7 +2787,7 @@ class FramePlotMethods(BasePlotMethods):
             The values to be plotted.
             Either the location or the label of the columns to be used.
             By default, it will use the remaining DataFrame numeric columns.
-        **kwds : optional
+        **kwds
             Keyword arguments to pass on to :meth:`pandas.DataFrame.plot`.
 
         Returns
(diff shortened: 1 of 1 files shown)


pandas/plotting/_core.py
```diff
@@ -2704,18 +2704,56 @@ def __call__(self, x=None, y=None, kind='line', ax=None,
 
     def line(self, x=None, y=None, **kwds):
         """
-        Line plot
+        Plot `DataFrame` columns as lines.
+
+        This function is useful to plot lines using `DataFrame`'s values
```
TomAugspurger: I think sphinx complains if you do you have something right after a \`. Just do DataFrame without the backticks.

pandas/plotting/_core.py
```diff
@@ -2704,18 +2704,56 @@ def __call__(self, x=None, y=None, kind='line', ax=None,
...
+        as coordinates.
+        It allows to specify which columns to use for the axes.
```
TomAugspurger: Can remove this line since we discuess that in the parameters section.

pandas/plotting/_core.py
```diff
@@ -2704,18 +2704,56 @@ def __call__(self, x=None, y=None, kind='line', ax=None,
...
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
-        `**kwds` : optional
+        x : int, str, list of int or list of str, optional
```
TomAugspurger: I think x has to be an int or str. Lists are not allowed for x, just y.

pandas/plotting/_core.py
```diff
@@ -2704,18 +2704,56 @@ def __call__(self, x=None, y=None, kind='line', ax=None,
...
+            It defines the X axis coordinates.
```
TomAugspurger: Remove "it". Just have "Column to use for the horizontal axis." and then the rest of your description.

pandas/plotting/_core.py
```diff
@@ -2704,18 +2704,56 @@ def __call__(self, x=None, y=None, kind='line', ax=None,
...
+            Either the location or the label of the columns to be used.
+            By default, it will use the `DataFrame` indices.
+        y : int, str, list of int or list of str, optional
```
TomAugspurger: I thin we write this as `int, str, or list of them, optional`.

pandas/plotting/_core.py
```diff
@@ -2704,18 +2704,56 @@ def __call__(self, x=None, y=None, kind='line', ax=None,
...
+            It defines the values to be plotted.
```
TomAugspurger: "The values to be plotted."

pandas/plotting/_core.py
```diff
@@ -2704,18 +2704,56 @@ def __call__(self, x=None, y=None, kind='line', ax=None,
...
+            Either the location or the label of the columns to be used.
+            By default, it will use the remaining `DataFrame` columns.
```
TomAugspurger: No backticks. Specify that it's the remaining *numeric* columns.

pandas/plotting/_core.py
```diff
@@ -2704,18 +2704,56 @@ def __call__(self, x=None, y=None, kind='line', ax=None,
...
+        kwds : optional
             Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : matplotlib.AxesSubplot or np.array of them
+        axes : :class:`matplotlib.AxesSubplot` or :meth:`np.array` of
```
TomAugspurger: :meth:`np.array` -> ndarray. I don't care either way about the sphinx markup so either

    :class:numpy.ndarray

or

    numpy.ndarray

pandas/plotting/_core.py
```diff
@@ -2704,18 +2704,56 @@ def __call__(self, x=None, y=None, kind='line', ax=None,
...
+            ...    }, index=[1990, 1997, 2003, 2009, 2014])
+            >>> lines = df.plot.line()
+
+        .. plot::
+            :context: close-figs
+
+            The following example shows the relationship between both
+            populations.
+
+            >>> lines = df.plot.line('pig', 'horse')
```
TomAugspurger: Specify `x='pig', y='horse'`

pandas/plotting/_core.py
ifosch: We tested on Jupyter and these seem to be allowed. However, the result makes not much sense. I'm changing this, anyway.

ifosch:
Fixes are up now

TomAugspurger (approved):
Pushed a couple small updates.

Will wait to see if anyone else has thoughts, otherwise ready to merge.

ifosch:
Thanks @TomAugspurger !

TomAugspurger:
Thank you!

PR #20164, opened 2018-03-10
DOC: update the pandas.Series.set_axis docstring

astrastefania:
an unrelated PR):

```


###################### Docstring (pandas.Series.set_axis) ######################

Assign desired index to given axis.

Indexes for columns or rows can be changed by assigning
a list-like or Index.

Parameters
----------
labels : list-like, Index
    The values for the new index.

axis : int, default 0
    The value 0 indentify the rows, 1 identify the columns.

inplace : boolean, default None
    Whether to return a new NDFrame instance.

    WARNING: inplace=None currently falls back to to True, but
    in a future version, will default to False.  Use inplace=True
    explicitly rather than relying on the default.

    .. versionadded:: 0.21.0
        The signature is make consistent to the rest of the API.
        Previously, the "axis" and "labels" arguments were respectively
        the first and second positional arguments.

Returns
-------
renamed : NDFrame or None
    An object of same type as caller if inplace=False, None otherwise.

See Also
--------
pandas.DataFrame.rename_axis : Alter the name of the index or columns.

Examples
--------
>>> s = pd.Series([1, 2, 3])
>>> s
0    1
1    2
2    3
dtype: int64
>>> s.set_axis(['a', 'b', 'c'], axis=0, inplace=False)
a    1
b    2
c    3
dtype: int64
>>> df = pd.DataFrame({"A": [1, 2, 3], "B": [4, 5, 6]})
>>> df.set_axis(['a', 'b', 'c'], axis=0, inplace=False)
   A  B
a  1  4
b  2  5
c  3  6
>>> df.set_axis(['I', 'II'], axis=1, inplace=False)
   I  II
0  1   4
1  2   5
2  3   6
>>> df.set_axis(['i', 'ii'], axis=1, inplace=True)
>>> df
   i  ii
0  1   4
1  2   5
2  3   6

Errors found:
```

explicitly.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -487,11 +487,19 @@ class NDFrame(PandasObject, SelectionMixin):
 
         return new_axes
 
-    _shared_docs['set_axis'] = """
+    def set_axis(self, labels, axis=0, inplace=None):
+        """
         Assign desired index to given axis.
 
-        Indexes for columns or rows can be changed by assigning
+        Indexes for column or row labels can be changed by assigning
         a list-like or Index.
+
+        .. versionchanged:: 0.21.0
+
+           The signature is now `labels` and `axis`, consistent with
+           the rest of pandas API. Previously, the `axis` and `labels`
+           arguments were respectively the first and second positional
+           arguments.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -487,24 +487,30 @@ def _expand_axes(self, key):
...
 
         Parameters
         ----------
-        labels: list-like or Index
-            The values for the new index
-        axis : int or string, default 0
+        labels : list-like, Index
+            The values for the new index.
+
+        axis : int, default 0
```
TomAugspurger: axis : {0 or 'index', 1 or 'columns'}, default 0

TomAugspurger: And modify the description to say that the string forms are also acepted.

astrastefania: Thanks, so axis : {0 or 'index', 1 or 'columns'}, default 0 is enough or should I mentioned `int, str` too?

astrastefania:
OK, I added the fixes mentioned :)

TomAugspurger:
Pushed an update.

We weren't actually using the `set_axis` in a shared doc (I think). I converted it to a regular docstring. Going to let the tests finish before merging, in case I missed something. Running locally as well.

@astrastefania let me know if you see that the test finish before I do.

TomAugspurger (approved):

astrastefania:
Hey @TomAugspurger, should I update anything in here? 
I'm happy to understand better how CircleCI works. :)

TomAugspurger:
I think those failures are unrelated. Will look more closely tomorrow.

TomAugspurger:
@astrastefania those failures were fixed by https://github.com/pandas-dev/pandas/pull/20288. It failed on CircleCI since it tests your exact branch, which doesn't include those changes. It passed on TravisCI since it tests a merge commit, `master + your changes`, which does include those change (since they're on master).

We could merge master to this branch and repush and things should pass. But I'm just going to merge. Thanks!

PR #20170, opened 2018-03-10
DOC: update the pandas.DataFrame.plot.area docstring

DZPM:
- [x] PR title is "DOC: update the pandas.DataFrame.plot.area docstring"
- [x] The validation script passes: `scripts/validate_docstrings.py pandas.DataFrame.plot.area`
- [x] The html version looks good: `python doc/make.py --single pandas.DataFrame.plot.area`

```
#################### Docstring (pandas.DataFrame.plot.area) ####################

Draw an area plot of the input series using matplotlib.

An area plot displays graphically quantitative data.
This function wraps the matplotlib area function.

Parameters
----------
x : label or position, optional
    Coordinates for X point.
y : label or position, optional
    Coordinates for Y point.
kwds : optional

Returns
-------
matplotlib.AxesSubplot or numpy.ndarray
    Area plot generated.

See Also
--------
pandas.DataFrame.plot : Draw a plot using matplotlib.

Examples
--------
.. plot::
    :context: close-figs

    This example draws an area plot based on the length and width of
    some animals, displayed in three bins

    >>> df = pd.DataFrame({
    ...     'sales': [3, 2, 3, 9, 10, 6],
    ...     'visits': [20, 42, 28, 62, 81, 50],
    ... })
    >>> area = df.plot.area()

Docstring for "pandas.DataFrame.plot.area" correct. :)
```

### DIFF
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -3345,9 +3345,9 @@ class FramePlotMethods(BasePlotMethods):
         Parameters
         ----------
         x : label or position, optional
-            Index, coordinates for X axis.
+            Coordinates for the X axis. By default uses the index.
         y : label or position, optional
-            Index, coordinates for Y axis.
+            Column to plot. By default uses all columns.
         stacked : boolean, default True
             Area plots are stacked by default. Set to False to create a
             unstacked plot.
@@ -3375,7 +3375,8 @@ class FramePlotMethods(BasePlotMethods):
             ...     'sales': [3, 2, 3, 9, 10, 6],
             ...     'signups': [5, 5, 6, 12, 14, 13],
             ...     'visits': [20, 42, 28, 62, 81, 50],
-            ... })
+            ... }, index=pd.date_range(start='2018/01/01', end='2018/07/01',
+            ...                        freq='M'))
             >>> ax = df.plot.area()
 
         Area plots are stacked by default. To produce an unstacked plot,
@@ -3388,7 +3389,8 @@ class FramePlotMethods(BasePlotMethods):
             ...     'sales': [3, 2, 3, 9, 10, 6],
             ...     'signups': [5, 5, 6, 12, 14, 13],
             ...     'visits': [20, 42, 28, 62, 81, 50],
-            ... })
+            ... }, index=pd.date_range(start='2018/01/01', end='2018/07/01',
+            ...                        freq='M'))
             >>> ax = df.plot.area(stacked=False)
 
         Draw an area plot for each metric:
@@ -3400,10 +3402,9 @@ class FramePlotMethods(BasePlotMethods):
             ...     'sales': [3, 2, 3, 9, 10, 6],
             ...     'signups': [5, 5, 6, 12, 14, 13],
             ...     'visits': [20, 42, 28, 62, 81, 50],
-            ... })
+            ... }, index=pd.date_range(start='2018/01/01', end='2018/07/01',
+            ...                        freq='M'))
             >>> ax = df.plot.area(y='sales')
-            >>> ax = df.plot.area(y='signups')
-            >>> ax = df.plot.area(y='visits')
 
         Draw with a different `x`:
 
@@ -3414,7 +3415,8 @@ class FramePlotMethods(BasePlotMethods):
             ...     'sales': [3, 2, 3],
             ...     'visits': [20, 42, 28],
             ...     'day': ['Monday', 'Tuesday', 'Wednesday'],
-            ... })
+            ... }, index=pd.date_range(start='2018/01/01', end='2018/07/01',
+            ...                        freq='M'))
             >>> ax = df.plot.area(x='day')
         """
         return self(kind='area', x=x, y=y, **kwds)
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -3348,7 +3348,7 @@ class FramePlotMethods(BasePlotMethods):
             Coordinates for the X axis. By default uses the index.
         y : label or position, optional
             Column to plot. By default uses all columns.
-        stacked : boolean, default True
+        stacked : bool, default True
             Area plots are stacked by default. Set to False to create a
             unstacked plot.
         **kwds : optional
@@ -3362,7 +3362,7 @@ class FramePlotMethods(BasePlotMethods):
 
         See Also
         --------
-        pandas.DataFrame.plot : Draw plots.
+        DataFrame.plot : Make plots of DataFrame using matplotlib / pylab.
 
         Examples
         --------
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -3337,19 +3337,85 @@ class FramePlotMethods(BasePlotMethods):
 
     def area(self, x=None, y=None, **kwds):
         """
-        Area plot
+        Draw a stacked area plot.
+
+        An area plot displays quantitative data visually.
+        This function wraps the matplotlib area function.
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
-        `**kwds` : optional
+        x : label or position, optional
+            Index, coordinates for X axis.
+        y : label or position, optional
+            Index, coordinates for Y axis.
+        stacked : boolean, default True
+            Area plots are stacked by default. Set to False to create a
+            unstacked plot.
+        **kwds : optional
             Additional keyword arguments are documented in
             :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : :class:`matplotlib.axes.Axes` or numpy.ndarray of them
+        matplotlib.axes.Axes or numpy.ndarray
+            Area plot, or array of area plots if subplots is True
+
+        See Also
+        --------
+        pandas.DataFrame.plot : Draw plots.
+
+        Examples
+        --------
+        Draw an area plot based on basic business metrics:
+
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({
+            ...     'sales': [3, 2, 3, 9, 10, 6],
+            ...     'signups': [5, 5, 6, 12, 14, 13],
+            ...     'visits': [20, 42, 28, 62, 81, 50],
+            ... })
+            >>> ax = df.plot.area()
+
+        Area plots are stacked by default. To produce an unstacked plot,
+        pass ``stacked=False``:
+
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({
+            ...     'sales': [3, 2, 3, 9, 10, 6],
+            ...     'signups': [5, 5, 6, 12, 14, 13],
+            ...     'visits': [20, 42, 28, 62, 81, 50],
+            ... })
+            >>> ax = df.plot.area(stacked=False)
+
+        Draw an area plot for each metric:
+
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({
+            ...     'sales': [3, 2, 3, 9, 10, 6],
+            ...     'signups': [5, 5, 6, 12, 14, 13],
+            ...     'visits': [20, 42, 28, 62, 81, 50],
+            ... })
+            >>> ax = df.plot.area(y='sales')
+            >>> ax = df.plot.area(y='signups')
+            >>> ax = df.plot.area(y='visits')
+
+        Draw with a different `x`:
+
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({
+            ...     'sales': [3, 2, 3],
+            ...     'visits': [20, 42, 28],
+            ...     'day': ['Monday', 'Tuesday', 'Wednesday'],
+            ... })
+            >>> ax = df.plot.area(x='day')
         """
         return self(kind='area', x=x, y=y, **kwds)
 


pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
 
     def area(self, x=None, y=None, **kwds):
         """
-        Area plot
+        Draw an area plot of the input series using matplotlib.
+
+        An area plot displays graphically quantitative data.
```
kynan: displays quantitative data visually

pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+        This function wraps the matplotlib area function.
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
-        `**kwds` : optional
+        x : label or position, optional
+            Coordinates for X point.
```
kynan: X axis

pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+        y : label or position, optional
+            Coordinates for Y point.
```
kynan: Y axis

pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
 
         Returns
         -------
-        axes : matplotlib.AxesSubplot or np.array of them
+        matplotlib.AxesSubplot or numpy.ndarray
+            Area plot generated.
+
+        See Also
+        --------
+        pandas.DataFrame.plot : Draw a plot using matplotlib.
```
jorisvandenbossche: The explanation here is a bit strange, as it would seem to say that this one is not plotting with matplotlib

pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+        y : label or position, optional
+            Coordinates for Y point.
+        kwds : optional
             Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : matplotlib.AxesSubplot or np.array of them
+        matplotlib.AxesSubplot or numpy.ndarray
+            Area plot generated.
```
jorisvandenbossche: Can you explain something about when ndarray is returned?

pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
 
     def area(self, x=None, y=None, **kwds):
         """
-        Area plot
+        Draw an area plot of the input series using matplotlib.
```
jorisvandenbossche: "input series" -> input is a bit strange wording as it is the calling object (not passed as input) + it should be DataFrame instead of Series ?

jorisvandenbossche (commented):
Thanks! added some comments

dukebody:
Can you post a screenshot of the rendered HTML? It will help reviewing without having to check out your branch and compiling it.

pandas/plotting/_core.py
dukebody: I don't know if we want to mention the underlying library this is using at this level. Does the user really care so much? In the extended description is more OK IMO.

dukebody: Also, as @jorisvandenbossche says, the description doesn't make sense for dataframes. Probably what the function does is to display a stacked area plot?

pandas/plotting/_core.py
dukebody: It would be great to add some info about why this function is useful or some basic use cases. `Displays quantitative data visually` doesn't really give me very interesting information about what is the shape or usage of the plot.

pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+
+        An area plot displays graphically quantitative data.
+        This function wraps the matplotlib area function.
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
-        `**kwds` : optional
+        x : label or position, optional
```
dukebody: This should follow the guide: https://python-sprints.github.io/pandas/guide/pandas_docstring.html#section-3-parameters

So it should indicate the type of the parameter in the first line (plus optional), and later the meaning of the parameter.

pandas/plotting/_core.py
dukebody: From this description of the arguments `x` and `y` I really don't understand what they change in the plot. Can we make the parameter description a bit longer explaining what do they mean?

pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+            Coordinates for X point.
+        y : label or position, optional
+            Coordinates for Y point.
+        kwds : optional
```
dukebody: As mentioned in other PRs, you should use `**kwds` instead. Sorry for the great big confusion here.

pandas/plotting/_core.py
dukebody: Yep. In this case the existing explanation `Area plot generated` doesn't tell a lot. How is a `numpy.ndarray` the area plot generated?

pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+        See Also
+        --------
+        pandas.DataFrame.plot : Draw a plot using matplotlib.
+
+        Examples
+        --------
+        .. plot::
+            :context: close-figs
+
+            This example draws an area plot based on the length and width of
```
dukebody: 1) I think is better to put the text before the start of the `plot` block, and leave the block only for the code that constructs the plot.
2) The example this sentence talks about doesn't correspond to the dataframe below.
3) Don't write in 3rd person `This example does this`. Better write directly in imperative or present form, like `Draw an area plot based on` or `We can draw an area plot based on...`, but IMO mentioning `this example` is somewhat redundant when we are already in the examples section. :)

pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+            some animals, displayed in three bins
```
dukebody: In three bins? Where does the three come from?

pandas/plotting/_core.py
```diff
@@ -2818,18 +2818,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+
+            >>> df = pd.DataFrame({
+            ...     'sales': [3, 2, 3, 9, 10, 6],
+            ...     'visits': [20, 42, 28, 62, 81, 50],
+            ... })
+            >>> area = df.plot.area()
```
dukebody: I'm not sure about this, but if the result of this method is an axis object, perhaps it's better to assign it to a variable named `ax` like in other examples.

Can you add other examples where the usage of the `x` and `y` parameters is shown? Also an example where a `numpy.ndarray` is returned.

dukebody (changes requested):
I believe this needs some better explanations of the extended description, arguments and more examples that help to understand how this graph works.

dukebody:
Is there anything we can do to help finishing the work on this PR?

DZPM:
I'll update the PR later this week, just need some time :+1:

pandas/plotting/_core.py
DZPM: Using `**kwds` raises errors in the current version of the validator:

	Errors in parameters section
		Unknown parameters {'**kwds'}

I think it's important to be consistent between the validator and the code.

Of course, once the validator is updated, I agree this must be changed.

pandas/plotting/_core.py
DZPM: Yes, it's wrong, copied it from another example. Removed it.

pandas/plotting/_core.py
DZPM: Not sure about how to define a "usable" example, a little help would be appreciated!

DZPM:
Fixed most of the suggestions, except for:
- writing `**kwds` instead of `kwds`. That would break the validation.
- more/better examples: I ask for inspiration :)

Also, a screenshot of the example:
image

pandas/plotting/_core.py
jorisvandenbossche: You can ignore that error in the validation script for now.

pandas/plotting/_core.py
```diff
@@ -3115,19 +3115,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
 
     def area(self, x=None, y=None, **kwds):
         """
-        Area plot
+        Draw an stacked area plot.
```
jorisvandenbossche: an -> a

pandas/plotting/_core.py
DZPM: Ok, updated!

pandas/plotting/_core.py
```diff
@@ -3115,19 +3115,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+
+        An area plot displays quantitative data visually.
+        This function wraps the matplotlib area function.
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
-        `**kwds` : optional
+        x : label or position, optional, default None
```
jorisvandenbossche: you can leave out the ", default None" part (saying it is "optional" should be informative enough"). 

But we should specify here what it used for `x` by default. I suppose this is the index?

(and the same comment for `y`

pandas/plotting/_core.py
```diff
@@ -3115,19 +3115,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+            Coordinates for X axis.
+        y : label or position, optional, default None
+            Coordinates for Y axis.
+        **kwds : optional
```
jorisvandenbossche: Can you also document the `stacked` keyword? I know it is not in the signature, but it is a relevant one for this type of plot, so I would add it. 
See http://pandas.pydata.org/pandas-docs/stable/visualization.html#area-plot

(the validation script will complain about it, but you can ignore that again)

pandas/plotting/_core.py
```diff
@@ -3115,19 +3115,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
             Additional keyword arguments are documented in
             :meth:`pandas.DataFrame.plot`.
 
         Returns
         -------
-        axes : :class:`matplotlib.axes.Axes` or numpy.ndarray of them
+        matplotlib.axes.Axes or numpy.ndarray
+            Area plot, or array of area plots.
```
jorisvandenbossche: It would be good to add when an array is returned, which is when `subplots=True` is specified

pandas/plotting/_core.py
```diff
@@ -3115,19 +3115,42 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+        Draw an area plot based on the length and width of some animals:
+
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({
+            ...     'sales': [3, 2, 3, 9, 10, 6],
+            ...     'visits': [20, 42, 28, 62, 81, 50],
+            ... })
+            >>> ax = df.plot.area()
```
jorisvandenbossche: Some ideas for the examples:

- I would also show one where you use the `x` and `y` keywords
- I would show one with `stacked=False`
- I think it is typical to use area plots with time series data. So maybe it would be nice to add a DatetimeIndex to the `df`.

jorisvandenbossche (commented):
Thanks for the updates! Added some more comments

pandas/plotting/_core.py
TomAugspurger: Yes I believe it's the index by default.

TomAugspurger:
> writing **kwds instead of kwds. That would break the validation.

You can ignore this, `**kwargs` is correct, the script is incorrect :)

TomAugspurger:
http://pandas.pydata.org/pandas-docs/stable/visualization.html#area-plot has an example. The same thing with a datetimeidnex from `pd.date_range` would be nice (maybe a bit wider of an aspect ratio).

DZPM:
Hi,

I've updated with all the proposals, and rebased.

Screenshots of the rendered example section:
image
image
image
image
image

It looks much better :) Thanks for the feedback!

DZPM:
Ping @kynan @jorisvandenbossche @TomAugspurger @dukebody 

Feel free to keep reviewing, or do the merge :)

jorisvandenbossche:
@DZPM Thanks for the reminder! 
Can you add a datetime index to the example frame? See https://github.com/pandas-dev/pandas/pull/20170#issuecomment-373364136

pandas/plotting/_core.py
```diff
@@ -3188,19 +3188,85 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+        An area plot displays quantitative data visually.
+        This function wraps the matplotlib area function.
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
-        `**kwds` : optional
+        x : label or position, optional
+            Index, coordinates for X axis.
```
jorisvandenbossche: The "Index, " part is a bit confusing I think. I would reword this as : `Coordinates for the Y axis. By default uses the index.`

pandas/plotting/_core.py
```diff
@@ -3188,19 +3188,85 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+        y : label or position, optional
+            Index, coordinates for Y axis.
```
jorisvandenbossche: Here the default is not the index, but all columns. So maybe something like "Column to plot. By default uses all columns." ?

pandas/plotting/_core.py
```diff
@@ -3188,19 +3188,85 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({
+            ...     'sales': [3, 2, 3, 9, 10, 6],
+            ...     'signups': [5, 5, 6, 12, 14, 13],
+            ...     'visits': [20, 42, 28, 62, 81, 50],
+            ... })
+            >>> ax = df.plot.area(y='sales')
+            >>> ax = df.plot.area(y='signups')
```
jorisvandenbossche: I think showing only the first is fine to illustrate the `y` keyword (to save some vertical screen estate)

(if you want to have all of them on a separate figure, one could also use `subplots=True`)

jorisvandenbossche (commented):
Added a few more comments

BTW, for future reference, if you add new commits to the branch with additional changes instead of squashing everything into a single commit, it is a bit easier to see what has been updated since the last review

datapythonista:
@DZPM can you please update this PR based on the previous feedback

DZPM:
Sure, will do. (Sorry, I've been busy and the PR slipped my mind...)

DZPM:
Done. Added the changes in a new commit, and rebased.

pandas/plotting/_core.py
```diff
@@ -3337,19 +3337,87 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({
+            ...     'sales': [3, 2, 3, 9, 10, 6],
+            ...     'signups': [5, 5, 6, 12, 14, 13],
+            ...     'visits': [20, 42, 28, 62, 81, 50],
+            ... }, index=pd.date_range(start='2018/01/01', end='2018/07/01',
+            ...                        freq='M'))
```
datapythonista: is the `.. plot::` directive resetting the namespace? Otherwise, It'd be good to reuse the DataFrame instead of repeating it in every example.

pandas/plotting/_core.py
```diff
@@ -3337,19 +3337,87 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
 
         Returns
         -------
-        axes : :class:`matplotlib.axes.Axes` or numpy.ndarray of them
+        matplotlib.axes.Axes or numpy.ndarray
+            Area plot, or array of area plots if subplots is True
+
+        See Also
+        --------
+        pandas.DataFrame.plot : Draw plots.
```
datapythonista: You can get rid of `pandas.` and just leave `DataFrame.plot`. Something a bit more descriptive than "Draw plots." could be useful.

Not sure if in similar docstrings we also added the matplotlib equivalent method. Can you check the docstring for `DataFrame.kde` for example, and be consistent with it?

pandas/plotting/_core.py
```diff
@@ -3337,19 +3337,87 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+        This function wraps the matplotlib area function.
 
         Parameters
         ----------
-        x, y : label or position, optional
-            Coordinates for each point.
-        `**kwds` : optional
+        x : label or position, optional
+            Coordinates for the X axis. By default uses the index.
+        y : label or position, optional
```
datapythonista: As @dukebody says, for both `x` and `y`, in the first line we should have the type, more than a short description. It's a bit tricky, as `label` can be anything, but I think we're using something like `object or int, optional`. Then, we can sure clarify that it's the `label` or the `position` in the description.

pandas/plotting/_core.py
```diff
@@ -3337,19 +3337,87 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+            Column to plot. By default uses all columns.
+        stacked : boolean, default True
```
datapythonista: Happy to leave like this now, but as `stacked` is not in the signature, may be we should add it there, or document it under `kwds`. But if we leave it like this, can you replace `boolean` by the Python type `bool` please?

datapythonista (commented):
Looks great, just few minor things. Thanks for the changes!

pandas/plotting/_core.py
DZPM: Had some issues with the namespace, so I kept it in different places.

Don't have the full environment right now, can't check if it also works the other way.

pandas/plotting/_core.py
DZPM: Added the `DataFrame.plot` description.

pandas/plotting/_core.py
DZPM: I kept it consistent with  `DataFrame.plot`, which is the more complete docstring example I could find.

DZPM:
Changes done, and rebased. Hope it's ok!

TomAugspurger:
LGTM, thanks!

If you want, check the doc build when it finishes (probably a few hours).

pandas/plotting/_core.py
jorisvandenbossche: Recreating the dataframe each time should not be needed

pandas/plotting/_core.py
```diff
@@ -3337,19 +3337,87 @@ def kde(self, bw_method=None, ind=None, **kwds):
...
+
+        .. plot::
+            :context: close-figs
+
+            >>> df = pd.DataFrame({
+            ...     'sales': [3, 2, 3],
+            ...     'visits': [20, 42, 28],
+            ...     'day': ['Monday', 'Tuesday', 'Wednesday'],
+            ... }, index=pd.date_range(start='2018/01/01', end='2018/07/01',
+            ...                        freq='M'))
```
jorisvandenbossche: This example is not running (the index is not correct). And also the labels to not show on the actual plot once this is fixed.

jorisvandenbossche (commented):
Some follow-up in https://github.com/pandas-dev/pandas/pull/22307

PR #20173, opened 2018-03-10
DOC: update the pandas.DataFrame.to_string docstring

Tomasz-Kluczkowski:
Had to correct parameters descriptions in common docstrings (format.py) as well as in frame.py.

an unrelated PR):

```
#################### Docstring (pandas.DataFrame.to_string) ####################

Render a DataFrame to a console-friendly tabular output.

Convert DataFrame object into a string (or unicode) representation
which can be shown in command line interface.

See Also
--------
pandas.DataFrame.to_html
    Convert dataframe to a html file.

Examples
--------
>>> d = {'col1' : [1, 2, 3], 'col2' : [4, 5, 6]}
>>> df = pd.DataFrame(d)
>>> print(df.to_string())
   col1  col2
0     1     4
1     2     5
2     3     6

Parameters
----------
buf : StringIO-like, optional
    Buffer to write to.
columns : sequence, optional, default None
    The subset of columns to write. The default None writes all columns.
col_space : int, optional
    The minimum width of each column.
header : bool, optional
    Write out the column names. If a list of strings is given, it is assumed to be aliases for the column names.
index : bool, optional, default True
    Whether to print index (row) labels.
na_rep : str, optional, default 'NaN'
    String representation of NAN to use.
formatters : list or dict of one-param. functions, optional, default None
    Formatter functions to apply to columns' elements by position or name.
    The result of each function must be a unicode string.
    List must be of length equal to the number of columns.
float_format : one-parameter function, optional, default None
    Formatter function to apply to columns' elements if they are floats.
    The result of this function must be a unicode string.
sparsify : bool, optional, default True
    Set to False for a DataFrame with a hierarchical index to print every
    multiindex key at each row.
index_names : bool, optional, default True
    Prints the names of the indexes.
line_width : int, optional, default no wrap
    Width to wrap a line in characters.
max_rows : int, optional
    Maximum number of rows to display in the console.
max_cols : int, optional
    Maximum number of columns to display in the console.
show_dimensions : bool, default False
    Display dataframe's dimensions (number of rows by number of columns).

justify : str, default None
    How to justify the column labels. If None uses the option from
    the print configuration (controlled by set_option), 'right' out
    of the box. Valid values are

    * left
    * right
    * center
    * justify
    * justify-all
    * start
    * end
    * inherit
    * match-parent
    * initial
    * unset.

Returns
-------
formatted : str (or unicode, depending on data and options)

Docstring for "pandas.DataFrame.to_string" correct. :)
```

explicitly.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


pandas/io/formats/format.py
```diff
@@ -53,35 +53,38 @@
...
     header : bool, optional
-        %(header)s
-    index : bool, optional
-        whether to print index (row) labels, default True
-    na_rep : string, optional
-        string representation of NAN to use, default 'NaN'
-    formatters : list or dict of one-parameter functions, optional
-        formatter functions to apply to columns' elements by position or name,
-        default None. The result of each function must be a unicode string.
+        %(header)s.
```
kynan: Substituting the header here won't break lines at 79 characters. Not sure this is worth fixing though.

kynan (approved):

pandas/io/formats/format.py
```diff
@@ -53,35 +53,38 @@
     Parameters
```
hissashirocha: Parameters should go before the See Also section, right after the Extended Summary.

pandas/io/formats/format.py
```diff
@@ -103,14 +106,14 @@
         * inherit
         * match-parent
         * initial
-        * unset
+        * unset.
 """
 
 return_docstring = """
-
     Returns
```
hissashirocha: The Returns section should go after the Parameters section and before the See Also section.

pandas/io/formats/format.py
```diff
@@ -103,14 +106,14 @@
...
     -------
-    formatted : string (or unicode, depending on data and options)"""
+    formatted : str (or unicode, depending on data and options)
```
hissashirocha: The return explanation is missing. It should go in the next line, indented.

Tomasz-Kluczkowski:
Hi,

I have a bit of a problem here as the docstring for this method is a result of concatenation of common docstrings, method’s docstring and a return docstring.
Therefore it looks like it is missing some parts or they are not in the right order but I believe the deciding factor here should be the actual html built by sphinx and it looks ok to me.
I shall add a return description though and recheck this again and post a result of html build by sphinx to confirm the final result.
Thx for reviewing.

Tomasz-Kluczkowski:
As an attachement there is a zipped html rendering of the command:
python make.py html --single pandas.DataFrame.to_string 
for modified docstring for pandas.DataFrame.to_string.


[pandas.DataFrame.to_string.zip](https://github.com/pandas-dev/pandas/files/1808817/pandas.DataFrame.to_string.zip)

It looks ok to me.
We have to remember that under the method itself there will be only See Also and Examples as the rest is pulled automatically from the common docstrings in format.py.

WillAyd:
Taking a look at this one!

WillAyd:
Ended up being a rather complicated fixup - @datapythonista could use a second set of eyes if you have time

WillAyd:
Thanks @Tomasz-Kluczkowski !

PR #20179, opened 2018-03-10
DOC: Improve the docstring of Series.take

gmacario:
an unrelated PR):

```
######################## Docstring (pandas.Series.take) ########################

Return the elements in the given *positional* indices along an axis.

This means that we are not indexing according to actual values in
the index attribute of the object. We are indexing according to the
actual position of the element in the object.

Parameters
----------
indices : array-like
    An array of ints indicating which positions to take.
axis : int, default 0
    The axis on which to select elements. "0" means that we are
    selecting rows, "1" means that we are selecting columns, etc.
convert : bool, default True
    Whether to convert negative indices into positive ones.
    For example, ``-1`` would map to the ``len(axis) - 1``.
    The conversions are similar to the behavior of indexing a
    regular Python list.
    .. deprecated:: 0.21.0
       In the future, negative indices will always be converted.
is_copy : bool, default True
    Whether to return a copy of the original object or not.
kwargs : xxx, default xxx
    TODO MUST ADD A REASONABLE DESCRIPTION.

Returns
-------
taken : type of caller
    An array-like containing the elements taken from the object.

See Also
--------
numpy.ndarray.take : TODO
numpy.take : TODO

Examples
--------
>>> df = pd.DataFrame([('falcon', 'bird',    389.0),
...                    ('parrot', 'bird',     24.0),
...                    ('lion',   'mammal',   80.5),
...                    ('monkey', 'mammal', np.nan)],
...                    columns=('name', 'class', 'max_speed'),
...                    index=[0, 2, 3, 1])
>>> df
     name   class  max_speed
0  falcon    bird      389.0
2  parrot    bird       24.0
3    lion  mammal       80.5
1  monkey  mammal        NaN

Take elements at positions 0 and 3 along the axis 0 (default).

Note how the actual indices selected (0 and 1) do not correspond to
our selected indices 0 and 3. That's because we are selecting the 0th
and 3rd rows, not rows whose indices equal 0 and 3.

>>> df.take([0, 3])
     name   class  max_speed
0  falcon    bird      389.0
1  monkey  mammal        NaN

Take elements at indices 1 and 2 along the axis 1 (column selection).

>>> df.take([1, 2], axis=1)
    class  max_speed
0    bird      389.0
2    bird       24.0
3  mammal       80.5
1  mammal        NaN

We may take elements using negative integers for positive indices,
starting from the end of the object, just like with Python lists.

>>> df.take([-1, -2])
     name   class  max_speed
1  monkey  mammal        NaN
3    lion  mammal       80.5

Docstring for "pandas.Series.take" correct. :)
```

explicitly.

Signed-off-by: Gianpaolo Macario <gmacario@gmail.com>

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -2525,24 +2525,35 @@ class NDFrame(PandasObject, SelectionMixin):
             The axis on which to select elements. "0" means that we are
             selecting rows, "1" means that we are selecting columns, etc.
         convert : bool, default True
-            .. deprecated:: 0.21.0
-               In the future, negative indices will always be converted.
-
             Whether to convert negative indices into positive ones.
             For example, ``-1`` would map to the ``len(axis) - 1``.
             The conversions are similar to the behavior of indexing a
             regular Python list.
+            .. deprecated:: 0.21.0
+               In the future, negative indices will always be converted.
         is_copy : bool, default True
             Whether to return a copy of the original object or not.
+        kwargs : xxx, default xxx
+            TODO MUST ADD A REASONABLE DESCRIPTION.
+
+        Returns
+        -------
+        taken : type of caller
+            An array-like containing the elements taken from the object.
+
+        See Also
+        --------
+        numpy.ndarray.take : TODO
+        numpy.take : TODO
 
         Examples
         --------
         >>> df = pd.DataFrame([('falcon', 'bird',    389.0),
-                               ('parrot', 'bird',     24.0),
-                               ('lion',   'mammal',   80.5),
-                               ('monkey', 'mammal', np.nan)],
-                              columns=('name', 'class', 'max_speed'),
-                              index=[0, 2, 3, 1])
+        ...                    ('parrot', 'bird',     24.0),
+        ...                    ('lion',   'mammal',   80.5),
+        ...                    ('monkey', 'mammal', np.nan)],
+        ...                    columns=('name', 'class', 'max_speed'),
+        ...                    index=[0, 2, 3, 1])
         >>> df
              name   class  max_speed
         0  falcon    bird      389.0
@@ -2557,6 +2568,7 @@ class NDFrame(PandasObject, SelectionMixin):
         and 3rd rows, not rows whose indices equal 0 and 3.
 
         >>> df.take([0, 3])
+             name   class  max_speed
         0  falcon    bird      389.0
         1  monkey  mammal        NaN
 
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -2525,24 +2525,35 @@ def _take(self, indices, axis=0, convert=True, is_copy=True):
...
             Whether to convert negative indices into positive ones.
             For example, ``-1`` would map to the ``len(axis) - 1``.
             The conversions are similar to the behavior of indexing a
             regular Python list.
+            .. deprecated:: 0.21.0
+               In the future, negative indices will always be converted.
         is_copy : bool, default True
             Whether to return a copy of the original object or not.
+        kwargs : xxx, default xxx
+            TODO MUST ADD A REASONABLE DESCRIPTION.
```
kynan: optional keyword arguments to ``:meth:`pandas.compat.numpy.function.validate_take`

pandas/core/generic.py
```diff
@@ -2525,24 +2525,35 @@ def _take(self, indices, axis=0, convert=True, is_copy=True):
...
+
+        Returns
+        -------
+        taken : type of caller
+            An array-like containing the elements taken from the object.
+
+        See Also
+        --------
+        numpy.ndarray.take : TODO
+        numpy.take : TODO
```
kynan: Take elements from an array along an axis

pandas/core/generic.py
```diff
@@ -2525,24 +2525,35 @@ def _take(self, indices, axis=0, convert=True, is_copy=True):
...
+            TODO MUST ADD A REASONABLE DESCRIPTION.
+
+        Returns
+        -------
+        taken : type of caller
+            An array-like containing the elements taken from the object.
+
+        See Also
+        --------
+        numpy.ndarray.take : TODO
```
kynan: Return an array formed from the elements at given indices

kynan (changes requested):

pandas/core/generic.py
gmacario: Fix in later commit

pandas/core/generic.py
gmacario: Fix in later commit

pandas/core/generic.py
gmacario: Fix in later commit

gmacario:
Hello @kynan

I believe I have addressed all your review feedbacks with my recent commits.
Would you mind giving it a look again?

Thanks!

TomAugspurger:
Thanks @gmacario.

PR #20180, opened 2018-03-10
DOC: update the pandas.Index.notna docstring

datadonK23:
an unrelated PR):

```
######################## Docstring (pandas.Index.notna) ########################

Detect existing (non-missing) values.

Return a boolean same-sized object indicating if the values are not NA.
Non-missing values get mapped to True. Characters such as empty
strings `''` or :attr:`numpy.inf` are not considered NA values
(unless you set :attr:`pandas.options.mode.use_inf_as_na` `= True`).
NA values, such as None or :attr:`numpy.NaN`, get mapped to False
values.

.. versionadded:: 0.20.0

Returns
-------
bool of type numpy.ndarray
    Boolean array to indicate which entries are not NA.

See also
--------
notnull : alias of notna
isna: inverse of notna
pandas.notna : top-level notna

Examples
--------
Show which entries in a pandas.Index are not NA. The result is an
array.

>>> idx = pd.Index([5.2, 6.0, np.NaN])
>>> idx
Float64Index([5.2, 6.0, nan], dtype='float64')
>>> idx.notna()
array([ True,  True, False])

Empty strings are not considered NA values. None is considered a NA
value.

>>> idx = pd.Index(['black', '', 'red', None])
>>> idx
Index(['black', '', 'red', None], dtype='object')
>>> idx.notna()
array([ True,  True,  True, False])

Docstring for "pandas.Index.notna" correct. :)
```

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -2045,7 +2045,7 @@ class Index(IndexOpsMixin, PandasObject):
         Return a boolean same-sized object indicating if the values are not NA.
         Non-missing values get mapped to ``True``. Characters such as empty
         strings ``''`` or :attr:`numpy.inf` are not considered NA values
-        (unless you set :attr:`pandas.options.mode.use_inf_as_na` `= True`).
+        (unless you set ``pandas.options.mode.use_inf_as_na = True``).
         NA values, such as None or :attr:`numpy.NaN`, get mapped to ``False``
         values.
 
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/base.py
```diff
@@ -2040,18 +2040,47 @@ def isna(self):
 
     def notna(self):
         """
-        Inverse of isna
+        Detect existing (non-missing) values.
+
+        Return a boolean same-sized object indicating if the values are not NA.
+        Non-missing values get mapped to True. Characters such as empty
```
kynan: ``True``

pandas/core/indexes/base.py
```diff
@@ -2040,18 +2040,47 @@ def isna(self):
...
+        strings `''` or :attr:`numpy.inf` are not considered NA values
```
kynan: ``''``

pandas/core/indexes/base.py
```diff
@@ -2040,18 +2040,47 @@ def isna(self):
...
+        (unless you set :attr:`pandas.options.mode.use_inf_as_na` `= True`).
+        NA values, such as None or :attr:`numpy.NaN`, get mapped to False
```
kynan: ``False``

pandas/core/indexes/base.py
donK23: Thx for the trick with the double quotes. I'm using it now.

jorisvandenbossche (approved):

jorisvandenbossche:
@donK23 made a minor edit, thanks a lot!

PR #20185, opened 2018-03-10
DOC: update the to_excel docstring

robmarkcole:
an unrelated PR):

- [ x] PR title is "DOC: update the <your-function-or-method> docstring"
- [x ] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [x ] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ x] The html version looks good: `python doc/make.py --single <your-function-or-method>`

```
Errors found:
	Missing description for See Also "pd.read_excel" reference
```


cmdelatorre (changes requested):
What about the docstring validtion script output? There are many things to fix there (No extended summary, no examples...)

robmarkcole:
@cmdelatorre I'm aware this is an incomplete PR and I'm sorry about that but we ran out of time on the sprint. I am getting errors due to my examples and `also` section and I will take some more time to fix these.

pandas/core/generic.py
```diff
@@ -1591,62 +1591,88 @@ def _repr_latex_(self):
...
     engine : string, default None
-        write engine to use - you can also set this via the options
+        Write engine to use - you can also set this via the options
         ``io.excel.xlsx.writer``, ``io.excel.xls.writer``, and
         ``io.excel.xlsm.writer``.
     merge_cells : boolean, default True
         Write MultiIndex and Hierarchical Rows as merged cells.
-    encoding: string, default None
-        encoding of the resulting excel file. Only necessary for xlwt,
+    encoding : string, default 'ascii'
```
dukebody: How did you figure this out? Reading the code of `DataFrame.to_excel` it looks like this parameter is ignored!

pandas/core/generic.py
```diff
@@ -1591,62 +1591,88 @@ def _repr_latex_(self):
...
+        File path or existing ExcelWriter.
     sheet_name : string, default 'Sheet1'
-        Name of sheet which will contain DataFrame
+        Name of sheet which will contain DataFrame.
     na_rep : string, default ''
-        Missing data representation
+        Missing data representation.
     float_format : string, default None
-        Format string for floating point numbers
+        Format string for floating point numbers.
```
dukebody: Can you mention some examples and/or link to where the format of this parameter is documented?

pandas/core/generic.py
```diff
@@ -1591,62 +1591,88 @@ def _repr_latex_(self):
...
     columns : sequence, optional
```
dukebody: Sequence of what? I think this should be "list of str or list of int" to select columns.

Also, add "If not specified, all columns will be written".

dukebody: BTW I think this parameter probably makes no sense for `Series.to_excel` and the docstring is shared. We should find a way to not show this parameter in the docs in this case.

pandas/core/generic.py
```diff
@@ -1591,62 +1591,88 @@ def _repr_latex_(self):
...
-        .. versionadded:: 0.20.0
+        .. versionadded:: 0.20.0.
+
+    Returns
+    -------
+    Nothing returned.
+
+    See Also
+    --------
+    pd.read_excel
```
dukebody: fix formatting: https://python-sprints.github.io/pandas/guide/pandas_docstring.html#section-5-see-also

pandas/core/generic.py
```diff
@@ -1591,62 +1591,88 @@ def _repr_latex_(self):
...
-        upper left cell row to dump data frame
-    startcol :
-        upper left cell column to dump data frame
+    startrow : integer, default 0
+        Upper left cell row to dump data frame.
+    startcol : integer, default 0
+        Upper left cell column to dump data frame.
     engine : string, default None
-        write engine to use - you can also set this via the options
+        Write engine to use - you can also set this via the options
```
dukebody: Which options? I understand the string is the fully qualified import part for the write engine class. Are the ones shown there examples? Can we specify that in the docs?

pandas/core/generic.py
```diff
@@ -1591,62 +1591,88 @@ def _repr_latex_(self):
...
+        Display more information in the error logs.
     freeze_panes : tuple of integer (length 2), default None
         Specifies the one-based bottommost row and rightmost column that
-        is to be frozen
+        is to be frozen.
 
-        .. versionadded:: 0.20.0
+        .. versionadded:: 0.20.0.
+
+    Returns
```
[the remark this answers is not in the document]
robmarkcole: OK

pandas/core/generic.py
robmarkcole: `pandas.read_excel`

pandas/core/generic.py
robmarkcole: You are correct

pandas/core/generic.py
robmarkcole: Formatting takes place [here](https://github.com/pandas-dev/pandas/blob/79200d3929f46fb5fd2e43ebf244658b3956ee00/pandas/io/formats/excel.py#L395) using the string modulo operator which is described [here](https://www.python-course.eu/python3_formatted_output.php). I've added an example.

pandas/core/generic.py
robmarkcole: Added `list of strings`

pandas/core/generic.py
robmarkcole: [It appears](https://github.com/pandas-dev/pandas/blob/aa59954a217c8f856bb0980265520d37b85a80af/asv_bench/benchmarks/io/excel.py#L12) the options are `'openpyxl', 'xlsxwriter', 'xlwt'`

pandas/core/generic.py
```diff
@@ -1591,62 +1591,87 @@ def _repr_latex_(self):
...
+    specifying unique sheet_name. With all data written to the file it is
+    necessary to save the changes. Note that creating an ExcelWriter object
+    with a file name that already exists will result in the contents of the
+    existing file being erased.
 
     Parameters
     ----------
     excel_writer : string or ExcelWriter object
-        File path or existing ExcelWriter
+        File path or existing ExcelWriter.
```
dukebody: Can we link to the class? I believe
```
:class:`pandas.ExcelWriter`
``` 
should work.

dukebody: Also, in the extended description we say that an `ExcelWriter` object is necessary to be able to use `.to_excel`, but in the parameter description we say that we also accept a "file path". So the `ExcelWriter` is not really needed? What happens if we specify a file path? I guess that a new `ExcelWriter` gets automatically created with default options.

pandas/core/generic.py
```diff
@@ -1591,62 +1591,87 @@ def _repr_latex_(self):
     # I/O Methods
 
     _shared_docs['to_excel'] = """
-    Write %(klass)s to an excel sheet
-    %(versionadded_to_excel)s
+    Write %(klass)s to an excel sheet.
+
+    To write a %(klass)s to an excel .xlsx file it is necessary to first create
+    an ExcelWriter object with a target file name, and specify a sheet in the
```
dukebody: Since ExcelWriter is a class I believe it should be between backticks to make it itallic like
```
`ExcelWriter`
```

pandas/core/generic.py
```diff
@@ -1591,62 +1591,87 @@ def _repr_latex_(self):
...
+
+    Examples
+    --------
+
+    >>> df1 = pd.DataFrame([['a', 'b'], ['c', 'd']],
+    ...                   index=['row 1', 'row 2'],
+    ...                   columns=['col 1', 'col 2'])
+    >>> writer = pd.ExcelWriter('output.xlsx', engine='xlsxwriter')
+    >>> df1.to_excel(writer, sheet_name='Sheet1')
+    >>> writer.save()
```
dukebody: I've observed this is only needed if we are creating an `ExcelWriter` explicitly. If one specifies a file path, like `df.to_excel('my_file.xml')` it is not needed.

Perhaps it makes sense to show first an example with a file path, which is simpler, and then an example with an `ExcelWriter` where we can specify more custom options?

pandas/core/generic.py
```diff
@@ -1591,62 +1591,87 @@ def _repr_latex_(self):
...
+    >>> writer2 = pd.ExcelWriter('output2.xlsx', engine='xlsxwriter')
+    >>> df1.to_excel(writer2, sheet_name='Sheet1')
+    >>> df2 = df1.copy()
+    >>> df2.to_excel(writer2, sheet_name='Sheet2')
+    >>> writer2.save()
+
+    Limit floats to a fixed precision using float_format. For example
+    float_format="%.2f" will format 0.1234 to 0.12.
 
     For compatibility with to_csv, to_excel serializes lists and dicts to
```
dukebody: Can we add a link to the method `pandas.DataFrame.to_csv` here as well?

pandas/core/generic.py
```diff
@@ -1591,62 +1591,87 @@ def _repr_latex_(self):
...
 
     _shared_docs['to_excel'] = """
-    Write %(klass)s to an excel sheet
-    %(versionadded_to_excel)s
+    Write %(klass)s to an excel sheet.
+
+    To write a %(klass)s to an excel .xlsx file it is necessary to first create
+    an ExcelWriter object with a target file name, and specify a sheet in the
+    file to write to. Multiple sheets may be written to by
+    specifying unique sheet_name. With all data written to the file it is
```
dukebody: Sheet name between backquotes since it's a variable name, see http://numpydoc.readthedocs.io/en/latest/format.html#sections

robmarkcole:
I've completely wiped by env to try and fix this error but still it persists. Means I am not able to run the validation script:
```
(pandas_dev) Robins-MacBook-Air:pandas robincole$ python scripts/validate_docstrings.py pandas.DataFrame.to_excel
  File "scripts/validate_docstrings.py", line 37, in <module>
    import pandas
  File "/Users/robincole/Documents/Github/pandas/pandas/__init__.py", line 42, in <module>
    from pandas.core.api import *
  File "/Users/robincole/Documents/Github/pandas/pandas/core/api.py", line 10, in <module>
    from pandas.core.groupby import Grouper
  File "/Users/robincole/Documents/Github/pandas/pandas/core/groupby.py", line 48, in <module>
  File "/Users/robincole/Documents/Github/pandas/pandas/core/frame.py", line 75, in <module>
    from pandas.core.series import Series
  File "/Users/robincole/Documents/Github/pandas/pandas/core/series.py", line 115, in <module>
    class Series(base.IndexOpsMixin, generic.NDFrame):
  File "/Users/robincole/Documents/Github/pandas/pandas/core/series.py", line 2942, in Series
    @Appender(generic._shared_docs['to_excel'] % _shared_doc_kwargs)
TypeError: not enough arguments for format string
```

pandas/core/generic.py
robmarkcole: Yes this is allowed `df1.to_excel("example.xlsx", sheet_name='Sheet3')`

pandas/core/generic.py
robmarkcole: ok

pandas/core/generic.py
robmarkcole: Added

pandas/core/generic.py
robmarkcole: ok

pandas/core/generic.py
robmarkcole: ok

pandas/core/generic.py
jorisvandenbossche: Yes, the ExcelWriter is not required

pandas/core/generic.py
```diff
@@ -1591,65 +1591,92 @@ def _repr_latex_(self):
     # I/O Methods
 
     _shared_docs['to_excel'] = """
-    Write %(klass)s to an excel sheet
-    %(versionadded_to_excel)s
+    Write %(klass)s to an excel sheet.
+
+    To write a %(klass)s to an excel .xlsx file it is necessary to first create
+    an `ExcelWriter` object with a target file name, and specify a sheet in the
```
jorisvandenbossche: As discussed, this is not required. 
But I think it is needed if you want to write multiple sheets, so that would be good for in the examples.

See http://pandas-docs.github.io/pandas-docs-travis/io.html#writing-excel-files-to-disk

pandas/core/generic.py
```diff
@@ -1591,65 +1591,92 @@ def _repr_latex_(self):
...
 
-    Notes
-    -----
-    If passing an existing ExcelWriter object, then the sheet will be added
-    to the existing workbook.  This can be used to save different
-    DataFrames to one workbook:
+    See Also
+    --------
+    pandas.read_excel
+
```
jorisvandenbossche: Can you add pandas.ExcelWriter ?

pandas/core/generic.py
```diff
@@ -1591,65 +1591,92 @@ def _repr_latex_(self):
...
 
-    For compatibility with to_csv, to_excel serializes lists and dicts to
-    strings before writing.
+    Note that once a workbook has been saved it is not possible write further
+    data without rewriting the whole workbook.
+
+    Limit floats to a fixed precision using float_format. For example
+    float_format="%.2f" will format 0.1234 to 0.12.
+
+    For compatibility with `to_csv <https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.to_csv.html>`__ ,
```
jorisvandenbossche: you can make this `` :meth:`~DataFrame.to_csv` `` to auto-generate the link

pandas/core/generic.py
```diff
@@ -1591,65 +1591,92 @@ def _repr_latex_(self):
...
+    >>> df2.to_excel(writer, sheet_name='Sheet2')
     >>> writer.save()
 
-    For compatibility with to_csv, to_excel serializes lists and dicts to
-    strings before writing.
+    Note that once a workbook has been saved it is not possible write further
+    data without rewriting the whole workbook.
+
+    Limit floats to a fixed precision using float_format. For example
+    float_format="%.2f" will format 0.1234 to 0.12.
```
jorisvandenbossche: So I think this one is generating the problem you have, because due to the `%` the templating / shared_docs system thinks it has to fill in something here, which it can't ... 
You can fix it by doing `"%%.2f"` (and in the actual docstring it will look like a single %)

TomAugspurger:
@robmarkcole were you able to sort out your issues? That error means that you're failing to pass a keyword to some other place that's using this shared docstring. Do you have time to finish this off?

robmarkcole:
@tomaugspurger I want able to solve and an not sure how I will. I am on holiday this week without computer so will have to take a look next week. Sorry for the delay

jorisvandenbossche:
See my comment here https://github.com/pandas-dev/pandas/pull/20185#discussion_r174501036 for a solution. 

> Sorry for the delay

No problem!

pandas/core/generic.py
robmarkcole: Updated

pandas/core/generic.py
robmarkcole: OK

pandas/core/generic.py
robmarkcole: OK

pandas/core/generic.py
robmarkcole: OK

robmarkcole:
@TomAugspurger @jorisvandenbossche sorry this fell off my radar after my holiday, have addressed your comments now. I'm still unable to run the validation script but hopefully all changes are OK

pandas/core/generic.py
```diff
@@ -1591,65 +1591,98 @@ def _repr_latex_(self):
     # I/O Methods
 
     _shared_docs['to_excel'] = """
-    Write %(klass)s to an excel sheet
-    %(versionadded_to_excel)s
+    Write %(klass)s to an excel sheet.
+
+    To write a single %(klass)s to an excel .xlsx file it is only necessary to
```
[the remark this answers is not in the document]
gfyoung: It is a little verbose, but I wouldn't cut this down that drastically IMO.

TomAugspurger:
Thanks @robmarkcole!

PR #20186, opened 2018-03-10
DOC: update the DataFrame.to_hdf() docstirng

ghost:
an unrelated PR):

```
##################### Docstring (pandas.DataFrame.to_hdf)  #####################

Write the contained data to an HDF5 file using HDFStore.

Hierarchical Data Format (HDF) is self-describing, allowing an
application to interpret the structure and contents of a file with
no outside information. One HDF file can hold a mix of related objects
which can be accessed as a group or as individual objects.

In order to add another :class:`~pandas.DataFrame` or
:class:`~pandas.Series` to an existing HDF file please use append mode
and different a key.

Parameters
----------
path_or_buf : str or pandas.HDFStore
    File path or HDFStore object.
key : str
    Identifier for the group in the store.
mode : {'a', 'w', 'r+'}, default is 'a'
    Mode to open file:
        - ``'w'``: write, a new file is created (an existing file with
            the same name would be deleted).
        - ``'a'``: append, an existing file is opened for reading and
            writing, and if the file does not exist it is created.
        - `'r+'`: similar to ``'a'``, but the file must already exist.
format : {'fixed', 'table'}, default is 'fixed'
    Possible values:
        - fixed: Fixed format. Fast writing/reading. Not-appendable,
            nor searchable.
        - table: Table format. Write as a PyTables Table structure
            which may perform worse but allow more flexible operations
            like searching / selecting subsets of the data.
append : boolean, default False
    For Table formats, append the input data to the existing.
data_columns :  list of columns or True, optional
    List of columns to create as indexed data columns for on-disk
    queries, or True to use all columns. By default only the axes
    of the object are indexed. See `here
    <http://pandas.pydata.org/pandas-docs/stable/io.html#query-via-data-columns>`__.
    Applicable only to format='table'.
complevel : {0-9}, optional
    Specifies a compression level for data.
    A value of 0 disables compression.
complib : {'zlib', 'lzo', 'bzip2', 'blosc'}, default 'zlib'
    Specifies the compression library to be used.
    As of v0.20.2 these additional compressors for Blosc are supported
    (default if no compressor specified: 'blosc:blosclz'):
    {'blosc:blosclz', 'blosc:lz4', 'blosc:lz4hc', 'blosc:snappy',
    'blosc:zlib', 'blosc:zstd'}.
    Specifying a compression library which is not available issues
    a ValueError.
fletcher32 : bool, default False
    If applying compression use the fletcher32 checksum.
dropna : bool, default False
    If true, ALL nan rows will not be written to store.

See Also
--------
DataFrame.to_csv : write out to a csv file.
DataFrame.to_sql : write to a sql table.
DataFrame.to_feather : write out feather-format for DataFrames.

Examples
--------
>>> df = pd.DataFrame({'A': [1, 2, 3], 'B': [4, 5, 6]},
...                   index=['a', 'b', 'c'])
>>> df.to_hdf('data.h5', key='df', mode='w')

We can append another object to the same file:

>>> s = pd.Series([1, 2, 3, 4])
>>> s.to_hdf('data.h5', key='s')

Reading from HDF file:

>>> pd.read_hdf('data.h5', 'df')
A  B
a  1  4
b  2  5
c  3  6
>>> pd.read_hdf('data.h5', 's')
0    1
1    2
2    3
3    4
dtype: int64

Notes
-----
Learn more about `Hierarchical Data Format (HDF)
<https://support.hdfgroup.org/HDF5/whatishdf5.html>`__.

Errors found:
	Errors in parameters section
		Unknown parameters {'format', 'complevel', 'mode', 'append', 'complib', 'dropna', 'data_columns', 'fletcher32'}
	No returns section found
```

explicitly.

- **kwargs are actually misleading and should be put as args, will do it in separate pull request.
- returns None

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1796,9 +1796,9 @@ class NDFrame(PandasObject, SelectionMixin):
 
         In order to add another :class:`~pandas.DataFrame` or
         :class:`~pandas.Series` to an existing HDF file please use append mode
-        and different a key.
+        and a different a key.
 
-        For more information see the :ref:`user guide <io.hdf5>`.
+        For more information see the :ref:`user guide <io.html#io-hdf5>`.
 
         Parameters
         ----------
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1794,9 +1794,8 @@ class NDFrame(PandasObject, SelectionMixin):
         no outside information. One HDF file can hold a mix of related objects
         which can be accessed as a group or as individual objects.
 
-        In order to add another :class:`~pandas.DataFrame` or
-        :class:`~pandas.Series` to an existing HDF file please use append mode
-        and a different a key.
+        In order to add another DataFrame or Series to an existing HDF file
+        please use append mode and a different a key.
 
         For more information see the :ref:`user guide <io.html#io-hdf5>`.
 
@@ -1806,27 +1805,23 @@ class NDFrame(PandasObject, SelectionMixin):
             File path or HDFStore object.
         key : str
             Identifier for the group in the store.
-        mode : {'a', 'w', 'r+'}, default is 'a'
+        mode : {'a', 'w', 'r+'}, default 'a'
             Mode to open file:
             
             - 'w': write, a new file is created (an existing file with
             the same name would be deleted).
-            
             - 'a': append, an existing file is opened for reading and
             writing, and if the file does not exist it is created.
-            
             - 'r+': similar to 'a', but the file must already exist.
         format : {'fixed', 'table'}, default 'fixed'
-            
             Possible values:
             
             - 'fixed': Fixed format. Fast writing/reading. Not-appendable,
             nor searchable.
-            
             - 'table': Table format. Write as a PyTables Table structure
             which may perform worse but allow more flexible operations
             like searching / selecting subsets of the data.
-        append : boolean, default False
+        append : bool, default False
             For Table formats, append the input data to the existing.
         data_columns :  list of columns or True, optional
             List of columns to create as indexed data columns for on-disk
@@ -1852,11 +1847,11 @@ class NDFrame(PandasObject, SelectionMixin):
 
         See Also
         --------
-        DataFrame.read_hdf : read from HDF file.
-        DataFrame.to_parquet : write a DataFrame to the binary parquet format.
-        DataFrame.to_sql : write to a sql table.
-        DataFrame.to_feather : write out feather-format for DataFrames.
-        DataFrame.to_csv : write out to a csv file.
+        DataFrame.read_hdf : Read from HDF file.
+        DataFrame.to_parquet : Write a DataFrame to the binary parquet format.
+        DataFrame.to_sql : Write to a sql table.
+        DataFrame.to_feather : Write out feather-format for DataFrames.
+        DataFrame.to_csv : Write out to a csv file.
 
         Examples
         --------
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1798,6 +1798,8 @@ class NDFrame(PandasObject, SelectionMixin):
         :class:`~pandas.Series` to an existing HDF file please use append mode
         and different a key.
 
+        For more information see the :ref:`user guide <io.hdf5>`.
+
         Parameters
         ----------
         path_or_buf : str or pandas.HDFStore
@@ -1806,18 +1808,24 @@ class NDFrame(PandasObject, SelectionMixin):
             Identifier for the group in the store.
         mode : {'a', 'w', 'r+'}, default is 'a'
             Mode to open file:
-                - ``'w'``: write, a new file is created (an existing file with
-                    the same name would be deleted).
-                - ``'a'``: append, an existing file is opened for reading and
-                    writing, and if the file does not exist it is created.
-                - `'r+'`: similar to ``'a'``, but the file must already exist.
-        format : {'fixed', 'table'}, default is 'fixed'
+            
+            - 'w': write, a new file is created (an existing file with
+            the same name would be deleted).
+            
+            - 'a': append, an existing file is opened for reading and
+            writing, and if the file does not exist it is created.
+            
+            - 'r+': similar to 'a', but the file must already exist.
+        format : {'fixed', 'table'}, default 'fixed'
+            
             Possible values:
-                - fixed: Fixed format. Fast writing/reading. Not-appendable,
-                    nor searchable.
-                - table: Table format. Write as a PyTables Table structure
-                    which may perform worse but allow more flexible operations
-                    like searching / selecting subsets of the data.
+            
+            - 'fixed': Fixed format. Fast writing/reading. Not-appendable,
+            nor searchable.
+            
+            - 'table': Table format. Write as a PyTables Table structure
+            which may perform worse but allow more flexible operations
+            like searching / selecting subsets of the data.
         append : boolean, default False
             For Table formats, append the input data to the existing.
         data_columns :  list of columns or True, optional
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -1834,6 +1835,42 @@ def to_hdf(self, path_or_buf, key, **kwargs):
             If applying compression use the fletcher32 checksum
         dropna : boolean, default False.
             If true, ALL nan rows will not be written to store.
+
+        See Also
+        --------
+        DataFrame.to_csv : write out to a csv file.
+        DataFrame.to_sql : write to a sql table.
+        DataFrame.to_feather : write out feather-format for DataFrames.
```
jreback: add DataFrame.to_parquet, read_hdf

pandas/core/generic.py
```diff
@@ -1834,6 +1835,42 @@ def to_hdf(self, path_or_buf, key, **kwargs):
...
+        b  2  5
+        c  3  6
+        >>> pd.read_hdf('data.h5', 's'))
+        0    1
+        1    2
+        2    3
+        3    4
+        dtype: int64
+
+        Notes
```
jreback: I think Notes are before examples.

If you want to put a link, link to the section in the io.rst docs. You *can* add the HDF5 link in there if you want (in an appropriate location)

jreback (changes requested):

pandas/core/generic.py
acidburnburn: Actually they already have link to the same manual there, so just removed this section.
Thanks!

acidburnburn: Also don't know why, but previously not all changes were pushed here, some remained locally, fixed it. Sorry for that.

pandas/core/generic.py
```diff
@@ -1786,40 +1786,47 @@ def to_json(self, path_or_buf=None, orient=None, date_format=None,
...
+        Write the contained data to an HDF5 file using HDFStore.
+
+        Hierarchical Data Format (HDF) is self-describing, allowing an
+        application to interpret the structure and contents of a file with
+        no outside information. One HDF file can hold a mix of related objects
+        which can be accessed as a group or as individual objects.
+
+        In order to add another :class:`~pandas.DataFrame` or
+        :class:`~pandas.Series` to an existing HDF file please use append mode
+        and different a key.
```
jorisvandenbossche: Can you add here a link to the user guide (because there is a lot more information there). You can use something like `` For more information see the :ref:`user guide <io.hdf5>`  ``

pandas/core/generic.py
```diff
@@ -1786,40 +1786,47 @@ def to_json(self, path_or_buf=None, orient=None, date_format=None,
...
-                       Write as a PyTables Table structure which may perform
-                       worse but allow more flexible operations like searching
-                       / selecting subsets of the data
+        path_or_buf : str or pandas.HDFStore
+            File path or HDFStore object.
+        key : str
+            Identifier for the group in the store.
+        mode : {'a', 'w', 'r+'}, default is 'a'
+            Mode to open file:
+                - ``'w'``: write, a new file is created (an existing file with
```
jorisvandenbossche: It's good to make this a list, but for sphinx , no indentation is needed (compared to "Mode .." on the line above), but, it needs a blank line between both lines (rst syntax details ...)

pandas/core/generic.py
```diff
@@ -1786,40 +1786,47 @@ def to_json(self, path_or_buf=None, orient=None, date_format=None,
...
+                    the same name would be deleted).
+                - ``'a'``: append, an existing file is opened for reading and
+                    writing, and if the file does not exist it is created.
+                - `'r+'`: similar to ``'a'``, but the file must already exist.
+        format : {'fixed', 'table'}, default is 'fixed'
+            Possible values:
+                - fixed: Fixed format. Fast writing/reading. Not-appendable,
```
jorisvandenbossche: same here about indentation / blank line

pandas/core/generic.py
```diff
@@ -1786,40 +1786,47 @@ def to_json(self, path_or_buf=None, orient=None, date_format=None,
...
+        key : str
+            Identifier for the group in the store.
+        mode : {'a', 'w', 'r+'}, default is 'a'
+            Mode to open file:
+                - ``'w'``: write, a new file is created (an existing file with
+                    the same name would be deleted).
+                - ``'a'``: append, an existing file is opened for reading and
+                    writing, and if the file does not exist it is created.
+                - `'r+'`: similar to ``'a'``, but the file must already exist.
+        format : {'fixed', 'table'}, default is 'fixed'
```
jorisvandenbossche: "default is 'fixed' " -> "default 'fixed' "

pandas/core/generic.py
jorisvandenbossche: also, can you add single quotes around fixed (like `'fixed'`), to make it clear it is a string

(and same for table below)

pandas/core/generic.py
```diff
@@ -1786,40 +1786,47 @@ def to_json(self, path_or_buf=None, orient=None, date_format=None,
...
+            Possible values:
+                - fixed: Fixed format. Fast writing/reading. Not-appendable,
+                    nor searchable.
+                - table: Table format. Write as a PyTables Table structure
+                    which may perform worse but allow more flexible operations
```
jorisvandenbossche: another identation issue. Here, the "which ..." needs to align with "table: .." on the line above

pandas/core/generic.py
```diff
@@ -1831,11 +1838,43 @@ def to_hdf(self, path_or_buf, key, **kwargs):
...
+        a  1  4
+        b  2  5
+        c  3  6
+        >>> pd.read_hdf('data.h5', 's')
+        0    1
+        1    2
+        2    3
+        3    4
+        dtype: int64
+        """
```
jorisvandenbossche: Can you add in the end here a code block with

```
>>> import os
>>> os.remove('data.h5')
```

(so running the doctests does not leave behind files)

pandas/core/generic.py
```diff
@@ -1831,11 +1838,43 @@ def to_hdf(self, path_or_buf, key, **kwargs):
...
+        DataFrame.to_feather : write out feather-format for DataFrames.
+        DataFrame.to_csv : write out to a csv file.
+
+        Examples
+        --------
+        >>> df = pd.DataFrame({'A': [1, 2, 3], 'B': [4, 5, 6]},
+        ...                   index=['a', 'b', 'c'])
+        >>> df.to_hdf('data.h5', key='df', mode='w')
+
+        We can append another object to the same file:
```
jorisvandenbossche: I would use "add" here instead of "append", because "append" is also a keyword with a different behaviour (appending rows to the same table, not the same file)

jorisvandenbossche (commented):
Thanks for the PR! 
Added some more comments

pandas/core/generic.py
shelvinskyi: Thanks for the hint! This one is hard for me, because i'm not an expert in rst :( I put the line  as you suggested, with the right subsection of that manual. when i generate html with 'make.py html' it doesnt' have a link. Is it ok?

pandas/core/generic.py
shelvinskyi: done

pandas/core/generic.py
shelvinskyi: done

pandas/core/generic.py
shelvinskyi: done

pandas/core/generic.py
shelvinskyi: done

pandas/core/generic.py
shelvinskyi: done

pandas/core/generic.py
shelvinskyi: done

Many thanks for you comments, they are really useful.

jorisvandenbossche (approved):
Thanks for the updates!

pandas/core/generic.py
jorisvandenbossche: Yes, that is fine (when only building the docstring, the full user guide is not built, and therefore the link does not seem to work)

jorisvandenbossche:
Thanks @acidburnburn !

PR #20189, opened 2018-03-10
DOC: update the Series.dt.strftime docstring

mircea-cosbuc:
an unrelated PR):

```
#################### Docstring (pandas.Series.dt.strftime)  ####################

Format series values with a date format string.

This function formats series values as specified by `date_format`, which
supports the same string format as the python standard library. Details
of the string format can be found in `python string format doc <https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior>`__.

Parameters
----------
date_format : str
    Date format string (e.g. "%Y-%m-%d").

Returns
-------
ndarray of formatted strings

Examples
--------
>>> rng = pd.date_range('10/3/2018 16:20:10', periods=2, freq='H')
>>> rng.strftime("%Y-%m-%d")
array(['2018-10-03', '2018-10-03'], dtype='<U10')

Errors found:
        Errors in parameters section
                Unknown parameters {'date_format'}
        See Also section not found
```

explicitly.

The method is injected with `_add_delegate_accessors` and the resulting signature on the object has `*args` and `**kwargs` instead of the actual `date_format`

### DIFF
--- a/pandas/core/indexes/datetimelike.py
+++ b/pandas/core/indexes/datetimelike.py
@@ -61,18 +61,26 @@ class DatelikeOps(object):
         return np.asarray(self.format(date_format=date_format),
                           dtype=compat.text_type)
     strftime.__doc__ = """
-    Return an array of formatted strings specified by date_format, which
+    Format series values with a date format string.
+
+    This function formats series values as specified by `date_format`, which
     supports the same string format as the python standard library. Details
-    of the string format can be found in `python string format doc <{0}>`__
+    of the string format can be found in `python string format doc <{0}>`__.
 
     Parameters
     ----------
     date_format : str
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/datetimelike.py
```diff
@@ -61,18 +61,26 @@ def strftime(self, date_format):
...
 
     Parameters
     ----------
     date_format : str
-        date format string (e.g. "%Y-%m-%d")
+        Date format string (e.g. "%Y-%m-%d").
 
     Returns
     -------
     ndarray of formatted strings
```
TomAugspurger: I think it should be 

    formatted: ndarray
        NumPy array strings formatted according to `date_format`.

jreback:
this is a duplicate of #20103 pls coordinate

PR #20193, opened 2018-03-10
DOC: update the pandas.DataFrame.to_sparse docstring

gioiab:
Updates the docstring for the to_sparse function.

Here is the output of the validation script:
```

#################### Docstring (pandas.DataFrame.to_sparse) ####################

Convert to SparseDataFrame.

Implement the sparse version of the DataFrame meaning that any data matching
a specific value it's omitted in the representation. The sparse DataFrame takes
less memory on disk when pickled and in the Python interpreter.

Parameters
----------
fill_value : float, default NaN
    The specific value that should be omitted in the representation.
kind : {'block', 'integer'}
    The kind of the SparseIndex tracking where data has been omitted.
    The block kind is recommended since it’s more memory efficient:
    it tracks just the locations and sizes of the blocks of data that
    are not equal to the fill value while the integer kind keeps an
    array with all those locations.

Returns
-------
y : SparseDataFrame

See Also
--------
pandas.DataFrame.to_dense: converts the DataFrame back to the its dense form

Examples
--------

Compressing on the zero value.

>>> df = pd.DataFrame(np.random.randn(1000, 4))
>>> df.iloc[:995] = 0.
>>> sdf = df.to_sparse(fill_value=0.)
>>> sdf.density
0.005

Docstring for "pandas.DataFrame.to_sparse" correct. :)
```


pandas/core/frame.py
```diff
@@ -1418,16 +1418,41 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
 
     def to_sparse(self, fill_value=None, kind='block'):
         """
-        Convert to SparseDataFrame
+        Convert to SparseDataFrame.
+
+        Implement the sparse version of the DataFrame meaning that any data matching
+        a specific value it's omitted in the representation. The sparse DataFrame takes
+        less memory on disk when pickled and in the Python interpreter.
```
jreback: You can just say its has more efficient storage.

pandas/core/frame.py
```diff
@@ -1418,16 +1418,41 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
 
         Parameters
         ----------
         fill_value : float, default NaN
+            The specific value that should be omitted in the representation.
         kind : {'block', 'integer'}
+            The kind of the SparseIndex tracking where data has been omitted.
+            The block kind is recommended since it’s more memory efficient:
+            it tracks just the locations and sizes of the blocks of data that
```
jreback: can you break these into 2 bullet points

pandas/core/frame.py
```diff
@@ -1418,16 +1418,41 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
+            are not equal to the fill value while the integer kind keeps an
+            array with all those locations.
 
         Returns
         -------
         y : SparseDataFrame
+
+        See Also
+        --------
+        pandas.DataFrame.to_dense: converts the DataFrame back to the its dense form
```
jreback: pandas.SparseDataFrame.to_dense

jreback (changes requested):

gioiab:
@jreback I've implemented the changes you requested, I'm just waiting the CI to finish. Can you get a final look?

gioiab:
@jreback the Appveyor build failed, together with at least 50 other ones, all with the same error message: `Command executed with exception: Cannot index into a null array.`

I don't have a retry button to queue another build. Could you please help me on this? Thanks!

pandas/core/frame.py
```diff
@@ -1418,16 +1418,45 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
         Parameters
         ----------
         fill_value : float, default NaN
+            The specific value that should be omitted in the representation.
         kind : {'block', 'integer'}
+            The kind of the SparseIndex tracking where data is not equal to
+            the fill value:
+
+            - 'block' tracks only the locations and sizes of blocks of data;
+
```
jreback: no blank lines between cases

pandas/core/frame.py
```diff
@@ -1418,16 +1418,45 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
+            The kind 'block' is recommended since it's more memory efficient.
 
         Returns
         -------
         y : SparseDataFrame
+
+        See Also
+        --------
+        pandas.SparseDataFrame.to_dense :
```
jreback: don't need pandas. here

jreback (approved):
minor comments. lgtm.

pandas/core/frame.py
gioiab: fixed

pandas/core/frame.py
gioiab: I changed this to `DataFrame.to_dense` instead of `pandas.SparseDataFrame.to_dense`: I've built the entire html documentation and the proper hyperlink is generated correctly in this way.

gioiab:
@jreback can I help you in some way in closing this? :)

jreback:
@gioiab can you rebase

@datapythonista if you'd have a look

pandas/core/frame.py
```diff
@@ -1570,16 +1570,44 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
+            The kind 'block' is recommended since it's more memory efficient.
 
         Returns
         -------
         y : SparseDataFrame
+
+        See Also
+        --------
+        DataFrame.to_dense :
+            converts the DataFrame back to the its dense form
```
datapythonista: Can you capitalize the first letter and add a period at the end?

I think the description should start in the same line as `to_dense` and continue to the next line indented when it doesn't fit.

pandas/core/frame.py
```diff
@@ -1570,16 +1570,44 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
+            the fill value:
+
+            - 'block' tracks only the locations and sizes of blocks of data;
+            - 'integer' keeps an array with all the locations of the data.
+
+            The kind 'block' is recommended since it's more memory efficient.
 
         Returns
         -------
         y : SparseDataFrame
```
datapythonista: You can get rid of the `y` and just leave the type. Also, may be it's a bit redundant in this case, but for consistency I'd add the description of what is returned.

pandas/core/frame.py
```diff
@@ -1570,16 +1570,44 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
 
         Parameters
         ----------
         fill_value : float, default NaN
+            The specific value that should be omitted in the representation.
         kind : {'block', 'integer'}
+            The kind of the SparseIndex tracking where data is not equal to
+            the fill value:
+
+            - 'block' tracks only the locations and sizes of blocks of data;
```
datapythonista: Not sure if the semi-colon at the end is intentional. If you find another docstring with a list, I'd use the same convention.

pandas/core/frame.py
```diff
@@ -1570,16 +1570,44 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
+        Examples
+        --------
+
+        Compressing on the zero value.
+
+        >>> df = pd.DataFrame(np.random.randn(1000, 4))
+        >>> df.iloc[:995] = 0.
+        >>> sdf = df.to_sparse(fill_value=0.)
+        >>> sdf.density
+        0.005
```
datapythonista: Nice example. It's just an opinion, feel free to leave it like this, if you think that being somehow more realistic is better, but I'd have something like: `pd.DataFrame([np.nan, np.nan, 1., np.nan])`. You can make it sparse with default arguments, then create another example with zeros instead of `NaN` and use `fill_value`. And if you find a nice way to illustrate the difference, I'd add an example with `kind='integer'`.

pandas/core/frame.py
```diff
@@ -1570,16 +1570,44 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
+
+        Implement the sparse version of the DataFrame meaning that any data
+        matching a specific value it's omitted in the representation.
+        The sparse DataFrame allows for a more efficient storage.
 
         Parameters
         ----------
         fill_value : float, default NaN
+            The specific value that should be omitted in the representation.
         kind : {'block', 'integer'}
```
datapythonista: The default is missing.

pandas/core/frame.py
```diff
@@ -1570,16 +1570,44 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
+            The kind of the SparseIndex tracking where data is not equal to
+            the fill value:
+
+            - 'block' tracks only the locations and sizes of blocks of data;
+            - 'integer' keeps an array with all the locations of the data.
+
+            The kind 'block' is recommended since it's more memory efficient.
```
datapythonista: I'd say the same, but in a way that doesn't sound like 'block' is always better. Something like "In most cases, the default 'block' is preferred for being more memory efficient.".

Not a big difference, but I'd prefer to avoid giving the idea that 'integer' is never the best option, which is not true.

datapythonista (commented):
Sorry for the late review. Great changes, I added some comments with formatting things and some ideas.

datapythonista:
Thanks @gioiab for the contribution. And sorry it took a while to merge it.

PR #20199, opened 2018-03-10
DOC: update the pandas.Series.str.rpartition() docstring (Delhi)

schopra6:
an unrelated PR):

```
################### Docstring (pandas.Series.str.rpartition) ###################

Split the string at the last occurrence of `sep`.

This method splits the string at the last occurrence of `sep`,
and return 3 elements containing the part before the separator,
the separator itself, and the part after the separator.
If the separator is not found, return 3 elements containing two empty strings, followed by the string itself.

Parameters
----------
pat : string, default whitespace
    String to split on.
expand : bool, default True
    If True, return DataFrame/MultiIndex expanding dimensionality.
    If False, return Series/Index.

Returns
-------
split : DataFrame/MultiIndex or Series/Index of objects

See Also
--------
partition : Split the string at the first occurrence of `sep`

Examples
--------

>>> s = pd.Series(['A_B_C', 'D_E_F', 'X'])
>>> s
0    A_B_C
1    D_E_F
2        X
dtype: object

>>> s.str.partition('_')
   0  1    2
0  A  _  B_C
1  D  _  E_F
2  X

>>> s.str.rpartition('_')
     0  1  2
0  A_B  _  C
1  D_E  _  F
2          X

Docstring for "pandas.Series.str.rpartition" correct. :)

```

explicitly.


cmdelatorre (approved):

pandas/core/strings.py
```diff
@@ -1611,18 +1611,20 @@ def rsplit(self, pat=None, n=-1, expand=False):
         return self._wrap_result(result, expand=expand)
 
     _shared_docs['str_partition'] = ("""
-    Split the string at the %(side)s occurrence of `sep`, and return 3 elements
-    containing the part before the separator, the separator itself,
-    and the part after the separator.
+    Split the string at the %(side)s occurrence of `sep`.
+
+    This method splits the string at the %(side)s occurrence of `sep`,
+    and return 3 elements containing the part before the separator,
```
TomAugspurger: "return" -> "returns"

pandas/core/strings.py
```diff
@@ -1635,7 +1637,8 @@ def rsplit(self, pat=None, n=-1, expand=False):
     Examples
     --------
 
-    >>> s = Series(['A_B_C', 'D_E_F', 'X'])
+    >>> s = pd.Series(['A_B_C', 'D_E_F', 'X'])
```
TomAugspurger: This docstring is shared with the Index. Could you add a short example with `Index` as well? (can be the same data as this one, just in `Index` instead of Series.

jorisvandenbossche:
@darke-spirits do you have time do update the PR based on the feedback?

PR #20200, opened 2018-03-10
DOC: update the NDFrame.transform docstring

oscar6echo:
an unrelated PR):

```
Docstring for "pandas.DataFrame.transform" correct. :)
```


oscar6echo:
We did this doc work with @DandanW924

pandas/core/generic.py
```diff
@@ -7660,6 +7618,204 @@ def nanptp(values, axis=0, skipna=True):
             ``numpy.ndarray`` method ``ptp``.""",
             nanptp)
 
+    _shared_docs['transform'] = ("""
+    Call function on a Series resp. DataFrame
```
jreback: summary needs to be a single line, not sure what 'resp.' is.

jreback: you can add more in the Extended Summary

pandas/core/generic.py
```diff
@@ -7660,6 +7618,204 @@ def nanptp(values, axis=0, skipna=True):
...
+    producing a like-indexed Series resp.
+    DataFrame with the transformed values.
+
+    .. versionadded:: 0.20.0
+
+    Parameters
+    ----------
+    func : Function, dictionary, or list of functions
+
+        Function: Function to apply to the DataFrame/Series.
```
jreback: no blank line here

pandas/core/generic.py
```diff
@@ -7660,6 +7618,204 @@ def nanptp(values, axis=0, skipna=True):
...
+    2000-01-05   -0.889050
+    2000-01-06    0.902506
+    2000-01-07   -1.212706
+    2000-01-08    1.183826
+    2000-01-09   -1.446060
+    2000-01-10    1.364543
+    Freq: D, Name: A, dtype: float64
+
+    See also
+    --------
```
jreback: nice examples, I am not sure this will render correctly, can you post a version of this?

jreback (changes requested):

pandas/core/generic.py
oscar6echo: Ok. Summary reduced to one line.  
'resp.' means respectively - but removed to be more compact.  
By 'extended summary' you meant 'Notes' ?

pandas/core/generic.py
oscar6echo: blank line 7631 removed.   
(But I think the rendering is not as nice)

pandas/core/generic.py
oscar6echo: Added Notes for clarification

pandas/core/generic.py
oscar6echo: They do render correctly.  
At least if the `python make.py  html --single pandas.DataFrame.transform` command generates the page which will be displayed (see screenshot below - simpler).  
(And I added one example to cover the case where the index is modified)

screenshot-2018-3-11-pandas dataframe transform documentation

pandas/core/generic.py
```diff
@@ -7660,6 +7618,231 @@ def nanptp(values, axis=0, skipna=True):
...
+    If applied to a DataFrame the function is applied on each row or column. 
+    Note that if the function modifies the index it must do so consistently
+    in the case of a DataFrame.
+
+    Examples
+    --------
+    >>> df = pd.DataFrame(np.arange(30).reshape(10, 3),
+    ...                   columns=['A', 'B', 'C'],
+    ...                   index=pd.date_range('1/1/2000', periods=10))
+    >>> df.iloc[3:7] = np.nan
```
TomAugspurger: Could you explain why you use missing values here? I'd prefer a shorter example (say 4-5 rows). The missing values don't seem to play a role, so I think they can be removed.

pandas/core/generic.py
```diff
@@ -8564,6 +8523,246 @@ def nanptp(values, axis=0, skipna=True):
...
+    2000-01-03 -1.0  0.0  1.0
+    2000-01-04  NaN  NaN  NaN
+    2000-01-05  NaN  NaN  NaN
+    2000-01-06  NaN  NaN  NaN
+    2000-01-07  NaN  NaN  NaN
+    2000-01-08 -1.0  0.0  1.0
+    2000-01-09 -1.0  0.0  1.0
+    2000-01-10 -1.0  0.0  1.0
+
+    >>> def f(s):
```
TomAugspurger: Can you explain the motivation for this example? What's it showing that the previous one didn't?

pandas/core/generic.py
```diff
@@ -8564,6 +8523,246 @@ def nanptp(values, axis=0, skipna=True):
...
+    2000-01-04   NaN   NaN   NaN
+    2000-01-05   NaN   NaN   NaN
+    2000-01-06   NaN   NaN   NaN
+    2000-01-07   NaN   NaN   NaN
+    2000-01-08  21.0  23.0  25.0
+    2000-01-09  24.0  26.0  28.0
+    2000-01-10  27.0  29.0  31.0
+
+    >>> def g(s):
+    ...     "input is a Series whose index may be modified"
```
TomAugspurger: I wonder whether this is officially supported, or whether it just happens to work... I'm struggling to find a reason why this would be the best way to do things.

pandas/core/generic.py
```diff
@@ -8564,6 +8523,246 @@ def nanptp(values, axis=0, skipna=True):
...
+    2000-01-03  2.772589   -0.635001  2.833213   -0.635001
+    2000-01-04       NaN         NaN       NaN         NaN
+    2000-01-05       NaN         NaN       NaN         NaN
+    2000-01-06       NaN         NaN       NaN         NaN
+    2000-01-07       NaN         NaN       NaN         NaN
+    2000-01-08  3.433987    0.635001  3.465736    0.635001
+    2000-01-09  3.526361    0.889001  3.555348    0.889001
+    2000-01-10  3.610918    1.143001  3.637586    1.143001
+
+    >>> s = pd.Series(np.sin([3*x for x in range(10)]), name='A',
```
TomAugspurger: Again, I think a shorter example would demonstrate the point better.

pandas/core/generic.py
```diff
@@ -8564,6 +8523,246 @@ def nanptp(values, axis=0, skipna=True):
...
+
+        - function : Applied to the Series, or to each column of the
+          DataFrame independently.
+        - list of functions : Each function is applied to the Series
+          or to each column of the DataFrame independently.
+        - dict of column names -> functions or list of functions :
+          For DataFrames only, apply the function or functions for a
+          given key to a specific column.
+
+    axis : {0 or 'index', 1 or 'columns'}, default 'index'
```
TomAugspurger: @jorisvandenbossche @datapythonista what to do here, do you think? `axis` isn't a keyword of `transform`, but it's intercepted by the implementation, and not passed through to `func`. I've chosen to document it. I think the best thing to do is to include it in the function signature as well (though in a separate PR).

TomAugspurger (commented):
Reformatted the Parameters section, and added some text to the examples.

I had a few questions about the examples inline.

pandas/core/generic.py
oscar6echo: The missing values were present in the [original example](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.transform.html). So I left them.  
They do not play a key role except maybe that nan do not crash transform.  
If you still want to remove them, I will.

pandas/core/generic.py
oscar6echo: It shows that the input to the function is a pandas.Series, as opposed to an unspecified kind of vector. I remember this is something that was not clear to me when I started using pandas. So I figured it might help others too.

pandas/core/generic.py
oscar6echo: Just expanding on the fact that the input is a Pandas.Series: All relevant manipulation works.  
I did not try and think if that was purposefully intended or is a consequence of implementation.  
Just to help a user, like me, see in a glance what you can do.

pandas/core/generic.py
TomAugspurger: I think the expectation is that pandas methods gracefully handle NaN, so in this case removing them is good.

pandas/core/generic.py
TomAugspurger: Yes, that's good (though `ndarray` has size as well). Perhaps something like

```
In [7]: def cell_name(series):
   ...:     return ['{}-{}'.format(series.name, idx) for idx in series.index]
   ...:

In [8]: df = pd.DataFrame({"A": [10, 20, 30], "B": [1, 2, 3]})

In [9]: df.transform(cn)
Out[9]:
     A    B
0  A-0  B-0
1  A-1  B-1
2  A-2  B-2
```

pandas/core/generic.py
TomAugspurger: >  Just to help a user, like me, see in a glance what you can do.

What you can do *currently* :) I could imagine a future change breaking this.

If you're interested, you could go through the NDFrame.transform tests to see if this is tested. If it's tested then I'd say it's supported, otherwise, I'd say it's not.

pandas/core/generic.py
oscar6echo: The reason I did not use random is that the validation script does not pass (random seed issue I guess). So I wanted 'random looking' values hence the sin().  
Anyway I simplified (a bit) and shortened the series.

pandas/core/generic.py
oscar6echo: **showing** that nan do not crash transform

pandas/core/generic.py
oscar6echo: For now I removed the index modifying example.

pandas/core/generic.py
oscar6echo: I removed NaN from the examples.

pandas/core/generic.py
oscar6echo: I put your example instead of mine.  
It is undoubtedly better.

oscar6echo:
Finally, do you plan to merge it or abandon it ? Are there other changes required ? Pls advise.

pandas/core/generic.py
datapythonista: Sorry, didn't see that during the sprint. As the sprint is over, and this PR changes code, I think this could be addressed here, right?

pandas/core/generic.py
```diff
@@ -8564,6 +8523,179 @@ def nanptp(values, axis=0, skipna=True):
...
+
+        * 0 or 'index' : apply along the index, i.e. the function
+          gets each column.
+        * 1 or 'columns' : apply along the columns, i.e. the function
+          gets each row.
+
+    *args
+        Additional positional arguments are passed to `func`.
+    **kwargs
+        Additional keyword arguments are passed to `func`.
```
datapythonista: You can merge both in a single line.

pandas/core/generic.py
```diff
@@ -8564,6 +8523,179 @@ def nanptp(values, axis=0, skipna=True):
...
+
+    Returns
+    -------
+    transformed : Series or DataFrame
```
datapythonista: You can leave only the types, the name `transformed` doesn't add much value here.

pandas/core/generic.py
```diff
@@ -8564,6 +8523,179 @@ def nanptp(values, axis=0, skipna=True):
...
+    Notes
+    ------
+    The function performs the transformation.
+    It takes a Series as input and must return a same length Series.
+    Note that the function may modify the Series index.
+
+    If applied to a Series the function is applied once.
+    If applied to a DataFrame the function is applied on each row or column.
+    Note that if the function modifies the index it must do so consistently
+    in the case of a DataFrame.
```
datapythonista: I'd move that to the extended summary. And I would probably get rid of the first sentence.

pandas/core/generic.py
```diff
@@ -8564,6 +8523,179 @@ def nanptp(values, axis=0, skipna=True):
...
+
+    See also
+    --------
+    pandas.DataFrame.aggregate : Perform aggregate type of operations.
+    pandas.DataFrame.apply : Apply general functions.
+
+    Examples
+    --------
+
+    transform() is applicable to Dataframes.
```
datapythonista: You can get rid of this.

pandas/core/generic.py
```diff
@@ -8564,6 +8523,179 @@ def nanptp(values, axis=0, skipna=True):
...
+
+    If applied to a Series the function is applied once.
+    If applied to a DataFrame the function is applied on each row or column.
+    Note that if the function modifies the index it must do so consistently
+    in the case of a DataFrame.
+
+    See also
+    --------
+    pandas.DataFrame.aggregate : Perform aggregate type of operations.
+    pandas.DataFrame.apply : Apply general functions.
```
datapythonista: I think you don't need the pandas prefix.

Would it make sense to have `map` and `applymap` too?

pandas/core/generic.py
```diff
@@ -8564,6 +8523,179 @@ def nanptp(values, axis=0, skipna=True):
...
+                A   B   C
+    2000-01-01   0   1   2
+    2000-01-02   3   4   5
+    2000-01-03   6   7   8
+    2000-01-04   9  10  11
+    2000-01-05  12  13  14
+
+    By default, `func` is applied to each column independently.
+
+    >>> df.transform(lambda x: (x - x.mean()) / x.std())
```
datapythonista: I'd try to find a more real-world example. I think for beginners it's a bit difficult to follow with meaningless data (A, B, C...) and lambdas.

Also, you may want to take a look at this document: http://pbpython.com/pandas_transform.html (or some other on how transform is usually used).

pandas/core/generic.py
oscar6echo: This PR is only about documentation. No change whatsoever in code. It was initiated during the sprint - and I was wondering why it was left hanging..

Done: I merged both lines into a single one.

pandas/core/generic.py
oscar6echo: Done.

pandas/core/generic.py
oscar6echo: Removed first sentence.  
I don't know what/where the extended summary is... Notes seemed to me the most complementary section available. I could just delete the Notes section ?

pandas/core/generic.py
oscar6echo: Yes. Done.

pandas/core/generic.py
datapythonista: There are code changes, mainly replacing lambdas by named functions. May be your IDE did it for you?

pandas/core/generic.py
oscar6echo: Frankly I disagree.  
(1) There is no clearer column names than A, B, C (Excel like) (2) the lambda example is taken from current doc about transform (3) I wrote other examples without lambda (4) at this stage of the example it is just to show it is applied by column and it gives the same result because the data is just shifted from one column to the next. (5) you're the 3rd person (legitimately) giving opinion about the examples and (naturally) each has a different one, often in contradiction... it going in circles...

The page you mention mixes groupby and transform so in my view is more complex and not what I, as a (heavy) user, would expect in the basic doc.

pandas/core/generic.py
oscar6echo: Argh... Yes I'm afraid. Sorry about that. Unintended.

pandas/core/generic.py
oscar6echo: Just to be clear.

I'll understand perfectly if you prefer to drop the PR - if only because there are some involuntary code change (bloody IDE !) - or the examples do not suit you.

Please then just let me know so I can close it.

Pandas is great and I'm a heavy almost daily user - and increasingly more people in my company. Congratulations !

PR #20201, opened 2018-03-10
DOC: update the DataFrame.update() docstring

kantologist:
an unrelated PR):

```
##################### Docstring (pandas.DataFrame.update)  #####################

Modify DataFrame in place using non-NA values from passed
DataFrame.

Aligns on indices.

Parameters
----------
other : DataFrame, or object coercible into a DataFrame
    Index should be similar to one of the columns in this one. If a
    Series is passed, its name attribute must be set, and that will be
    used as the column name in the resulting joined DataFrame.
join : {'left'}, default 'left'
    Indicates which column values overwrite.
overwrite : boolean, default True
    If True then overwrite values for common keys in the calling frame.
filter_func : callable(1d-array) -> 1d-array<boolean>, default None
    Can choose to replace values other than NA. Return True for values
    that should be updated.
raise_conflict : boolean
    If True, will raise an error if the DataFrame and other both
    contain data in the same place.

Examples
--------
>>> df = pd.DataFrame({'A': [1, 2, 3],
...                    'B': [400, 500, 600]})
>>> new_df = pd.DataFrame({'B': [4, 5, 6],
...                        'C': [7, 8, 9]})
>>> df.update(new_df)
>>> df
   A  B
0  1  4
1  2  5
2  3  6

>>> df = pd.DataFrame({'A': ['a', 'b', 'c'],
...                    'B': ['x', 'y', 'z']})
>>> new_df = pd.DataFrame({'B': ['d', 'e', 'f', 'g', 'h', 'i']})
>>> df.update(new_df)
>>> df
   A  B
0  a  d
1  b  e
2  c  f

>>> df = pd.DataFrame({'A': ['a', 'b', 'c'],
...                    'B': ['x', 'y', 'z']})
>>> new_column = pd.Series(['d', 'e'], name='B', index=[0, 2])
>>> df.update(new_column)
>>> df
   A  B
0  a  d
1  b  y
2  c  e
>>> df = pd.DataFrame({'A': ['a', 'b', 'c'],
...                    'B': ['x', 'y', 'z']})
>>> new_df = pd.DataFrame({'B': ['d', 'e']}, index=[1, 2])
>>> df.update(new_df)
>>> df
   A  B
0  a  x
1  b  d
2  c  e

If ``other`` contains NaNs the corresponding values are not updated
in the original dataframe.

>>> df = pd.DataFrame({'A': [1, 2, 3],
...                    'B': [400, 500, 600]})
>>> new_df = pd.DataFrame({'B': [4, np.nan, 6]})
>>> df.update(new_df)
>>> df
   A      B
0  1    4.0
1  2  500.0
2  3    6.0

See also
--------
DataFrame.merge : For column(s)-on-columns(s) operations

Returns
-------
updated : DataFrame

```

explicitly.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4217,7 +4217,8 @@ class DataFrame(NDFrame):
             Series is passed, its name attribute must be set, and that will be
             used as the column name in the resulting joined DataFrame.
         join : {'left'}, default 'left'
-            Indicates which column values overwrite.
+            Only left join is implemented,
+            keeping the index and columns of the original object.
         overwrite : boolean, default True
             If True then overwrite values for common keys in the calling frame.
         filter_func : callable(1d-array) -> 1d-array<boolean>, default None
@@ -4248,6 +4249,8 @@ class DataFrame(NDFrame):
         1  2  5
         2  3  6
 
+        The DataFrame's length does not increase as a result of the update.
+
         >>> df = pd.DataFrame({'A': ['a', 'b', 'c'],
         ...                    'B': ['x', 'y', 'z']})
         >>> new_df = pd.DataFrame({'B': ['d', 'e', 'f', 'g', 'h', 'i']})
@@ -4257,6 +4260,8 @@ class DataFrame(NDFrame):
         0  a  d
         1  b  e
         2  c  f
+
+        For Series, it's name attribute must be set.
 
         >>> df = pd.DataFrame({'A': ['a', 'b', 'c'],
         ...                    'B': ['x', 'y', 'z']})
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4328,8 +4328,9 @@ class DataFrame(NDFrame):
         Parameters
         ----------
         other : DataFrame, or object coercible into a DataFrame
-            Index should be similar to one of the columns in this one. If a
-            Series is passed, its name attribute must be set, and that will be
+            Should have at least one matching index/column label
+            with the original DataFrame. If a Series is passed,
+            its name attribute must be set, and that will be
             used as the column name in the resulting joined DataFrame.
         join : {'left'}, default 'left'
             Only left join is implemented, keeping the index and columns of the
@@ -4337,10 +4338,13 @@ class DataFrame(NDFrame):
         overwrite : boolean, default True
             How to handle non-NA values for overlapping keys.
 
-            * True : overwrite values in `self` with values from `other`.
-            * False : only update values that are NA in `self`.
+            * True : overwrite original DataFrame's values
+                    with values from `other`.
+            * False : only update values that are NA in
+                    the original DataFrame.
 
-        filter_func : callable(1d-array) -> 1d-array<boolean>, default None
+        filter_func : callable(1d-array) -> 1d-array<boolean>,\
+        "default None" -> "optional"
             Can choose to replace values other than NA. Return True for values
             that should be updated.
         raise_conflict : boolean
@@ -4370,7 +4374,8 @@ class DataFrame(NDFrame):
         1  2  5
         2  3  6
 
-        The DataFrame's length does not increase as a result of the update.
+        The DataFrame's length does not increase as a result of the update,
+        only values at matching index/column labels are updated.
 
         >>> df = pd.DataFrame({'A': ['a', 'b', 'c'],
         ...                    'B': ['x', 'y', 'z']})
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4206,8 +4206,7 @@ class DataFrame(NDFrame):
     def update(self, other, join='left', overwrite=True, filter_func=None,
                raise_conflict=False):
         """
-        Modify DataFrame in place using non-NA values from passed
-        DataFrame.
+        Modify in place using non-NA values from other DataFrame.
 
         Aligns on indices.
 
@@ -4227,6 +4226,14 @@ class DataFrame(NDFrame):
         raise_conflict : boolean
             If True, will raise an error if the DataFrame and other both
             contain data in the same place.
+
+        Returns
+        -------
+        updated : DataFrame
+
+        See Also
+        --------
+        DataFrame.merge : For column(s)-on-columns(s) operations
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


pandas/core/frame.py
```diff
@@ -4207,17 +4207,23 @@ def update(self, other, join='left', overwrite=True, filter_func=None,
                raise_conflict=False):
         """
         Modify DataFrame in place using non-NA values from passed
```
TomAugspurger: May have to reword this so the first line isn't too long? Does the validation script complain about this?

pandas/core/frame.py
```diff
@@ -4207,17 +4207,23 @@ def update(self, other, join='left', overwrite=True, filter_func=None,
...
+        Aligns on indices.
 
         Parameters
         ----------
         other : DataFrame, or object coercible into a DataFrame
+            Index should be similar to one of the columns in this one. If a
+            Series is passed, its name attribute must be set, and that will be
+            used as the column name in the resulting joined DataFrame.
         join : {'left'}, default 'left'
+            Indicates which column values overwrite.
```
TomAugspurger: Is `join='right'` valid? Can you do some exploration on what happens?

pandas/core/frame.py
```diff
@@ -4276,6 +4282,14 @@ def update(self, other, join='left', overwrite=True, filter_func=None,
         0  1    4.0
         1  2  500.0
         2  3    6.0
+
+        See also
```
TomAugspurger: See Also (note capitalization)

should go before the examples.

pandas/core/frame.py
```diff
@@ -4276,6 +4282,14 @@ def update(self, other, join='left', overwrite=True, filter_func=None,
...
+        --------
+        DataFrame.merge : For column(s)-on-columns(s) operations
+
+        Returns
+        -------
+        updated : DataFrame
```
TomAugspurger: Rerturns goes right after the Parameters section.

TomAugspurger (changes requested):

pandas/core/frame.py
kantologist: It's actually not valid. I tried it.

pandas/core/frame.py
kantologist: The validation script actually did not complain, but I'll try to adjust it.

pandas/core/frame.py
```diff
@@ -4206,22 +4206,35 @@ def combiner(x, y, needs_i8_conversion=False):
...
             Can choose to replace values other than NA. Return True for values
-            that should be updated
+            that should be updated.
         raise_conflict : boolean
             If True, will raise an error if the DataFrame and other both
             contain data in the same place.
 
+        Returns
+        -------
+        updated : DataFrame
```
jorisvandenbossche: It actually does not return a DataFrame, as the update is in place.

pandas/core/frame.py
jorisvandenbossche: I think you can say "Only left join is implemented, keeping the index and columns of the original object"

pandas/core/frame.py
```diff
@@ -4206,22 +4206,35 @@ def combiner(x, y, needs_i8_conversion=False):
...
         ----------
         other : DataFrame, or object coercible into a DataFrame
+            Index should be similar to one of the columns in this one. If a
+            Series is passed, its name attribute must be set, and that will be
+            used as the column name in the resulting joined DataFrame.
         join : {'left'}, default 'left'
+            Indicates which column values overwrite.
         overwrite : boolean, default True
-            If True then overwrite values for common keys in the calling frame
+            If True then overwrite values for common keys in the calling frame.
```
jorisvandenbossche: and what if False?

jorisvandenbossche: I think it that case only NA values in the calling object are updated

jorisvandenbossche (commented):
Can you add some more explanation between the different examples ?

TomAugspurger:
Made a couple updates. Mind taking a look @kantologist?

kantologist:
So the only issue is the doc validation script complaining about the absent of a return section. by the way I observed the same pattern of return section for other inlace functions too.

pandas/core/frame.py
```diff
@@ -4321,21 +4321,41 @@ def combiner(x, y, needs_i8_conversion=False):
...
-        Modify DataFrame in place using non-NA values from passed
-        DataFrame. Aligns on indices
+        Modify in place using non-NA values from another DataFrame.
+
+        Aligns on indices. There is no return value.
 
         Parameters
         ----------
         other : DataFrame, or object coercible into a DataFrame
+            Index should be similar to one of the columns in this one. If a
```
jorisvandenbossche: I don't really understand "Index should be similar to one of the columns in this one". What do you want to say here?

pandas/core/frame.py
```diff
@@ -4321,21 +4321,41 @@ def combiner(x, y, needs_i8_conversion=False):
...
+            Series is passed, its name attribute must be set, and that will be
+            used as the column name in the resulting joined DataFrame.
         join : {'left'}, default 'left'
+            Only left join is implemented, keeping the index and columns of the
+            original object.
         overwrite : boolean, default True
-            If True then overwrite values for common keys in the calling frame
+            How to handle non-NA values for overlapping keys.
+
+            * True : overwrite values in `self` with values from `other`.
```
jorisvandenbossche: I think we should try to avoid using 'self', as not necessarily every pandas user knows how classes are written. 

There is not always an ideal alternative, but I would use "original object" or "calling object" (or use DataFrame instead of object)

pandas/core/frame.py
```diff
@@ -4321,21 +4321,41 @@ def combiner(x, y, needs_i8_conversion=False):
...
+            * False : only update values that are NA in `self`.
+
         filter_func : callable(1d-array) -> 1d-array<boolean>, default None
```
jorisvandenbossche: "default None" -> "optional"

pandas/core/frame.py
```diff
@@ -4350,6 +4370,8 @@ def update(self, other, join='left', overwrite=True, filter_func=None,
         1  2  5
         2  3  6
 
+        The DataFrame's length does not increase as a result of the update.
```
jorisvandenbossche: maybe add something like ", only values at matching index/column labels are updated"

jorisvandenbossche (commented):
> So the only issue is the doc validation script complaining about the absent of a return section. by the way I observed the same pattern of return section for other inplace functions too.

It's fine to ignore those errors in this case (the methods that don't return something are a minority)

jorisvandenbossche (approved):
Thanks for the updates! Fixed up some minor formatting issues

jorisvandenbossche:
Thanks for the PR @kantologist !

PR #20202, opened 2018-03-10
DOC: update the pandas.DataFrame.apply docstring

mdeboc:
an unrelated PR):

```
###################### Docstring (pandas.DataFrame.apply) ######################

Apply a function along an axis of the `Series`.

Objects passed to the function are `Series` objects having as index
either the DataFrame's index (`axis=0`)
or the DataFrame's columns (`axis=1`).
If `result_type` is None, the final return type is the return
type of the applied function.
Otherwise, it depends on the `result_type` argument.

Parameters
----------
func : function
    Function to apply to each column or row.
    Axis along which the function is applied:

    * 0 or 'index': apply function to each column.
    * 1 or 'columns': apply function to each row.
broadcast : boolean, optional
    Only relevant for aggregation functions:

    * `False` or `None` : returns a `Series` whose length is the length
      of the index or the number of columns (based on the `axis`
      parameter)
    * `True` : results will be broadcast to the original shape
      of the frame, the original index and columns will be retained.

    .. deprecated:: 0.23.0
       This argument will be removed in a future version, replaced
       by result_type='broadcast'.

raw : boolean, default False
    * `False` : passes each row or column into a `Series` to the
      function.
    * `True` : the passed function will receive ndarray objects
      instead.
    If you are just applying a NumPy reduction function this will
    achieve much better performance.
reduce : boolean or None, default None
    Try to apply reduction procedures. If the `DataFrame` is empty,
    :meth:`apply` will use reduce to determine whether the result
    should be a `Series` or a `DataFrame`. If reduce is None (the
    default), :meth:`apply`'s return value will be guessed by calling
    func on an empty `Series`
    (note: while guessing, exceptions raised by `func` will be
    ignored).
    If reduce is True a `Series` will always be returned, and if
    `False` a `DataFrame` will always be returned.

    .. deprecated:: 0.23.0.
       This argument will be removed in a future version, replaced
       by result_type='reduce'.

result_type : {'expand', 'reduce', 'broadcast', None}
    These only act when `axis=1` (columns):

    * 'expand' : list-like results will be turned into columns.
    * 'reduce' : returns a `Series` if possible rather than expanding
      list-like results. This is the opposite of 'expand'.
    * 'broadcast' : results will be broadcast to the original shape
      of the `DataFrame`, the original index and columns will be
      retained.

    The default behaviour (`None`) depends on the return value of the
    applied function: list-like results will be returned as a `Series`
    of those. However if the apply function returns a `Series` these
    are expanded to columns.

    .. versionadded:: 0.23.0.

args : tuple
    Positional arguments to pass to function in addition to the
    array/series.
kwds :
    Additional keyword arguments to pass as keywords to the function.

Notes
-----
In the current implementation apply calls func twice on the
first column/row to decide whether it can take a fast or slow
code path. This can lead to unexpected behavior if func has
side-effects, as they will take effect twice for the first
column/row.

Examples
--------

We use this DataFrame to illustrate

>>> df = pd.DataFrame(np.tile(np.arange(3), 6).reshape(6, -1) + 1,
...                   columns=['A', 'B', 'C'])
>>> df
   A  B  C
0  1  2  3
1  1  2  3
2  1  2  3
3  1  2  3
4  1  2  3
5  1  2  3

Using a numpy universal function (in this case the same as
``np.sqrt(df)``):

>>> df.apply(np.sqrt)
     A         B         C
0  1.0  1.414214  1.732051
1  1.0  1.414214  1.732051
2  1.0  1.414214  1.732051
3  1.0  1.414214  1.732051
4  1.0  1.414214  1.732051
5  1.0  1.414214  1.732051

Using a reducing function on either axis

>>> df.apply(np.sum, axis=0)
A     6
B    12
C    18
dtype: int64

>>> df.apply(np.sum, axis=1)
0    6
1    6
2    6
3    6
4    6
5    6
dtype: int64

Retuning a list-like will result in a Series

>>> df.apply(lambda x: [1, 2], axis=1)
0    [1, 2]
1    [1, 2]
2    [1, 2]
3    [1, 2]
4    [1, 2]
5    [1, 2]
dtype: object

Passing result_type='expand' will expand list-like results
to columns of a Dataframe

>>> df.apply(lambda x: [1, 2], axis=1, result_type='expand')
   0  1
0  1  2
1  1  2
2  1  2
3  1  2
4  1  2
5  1  2

Returning a Series inside the function is similar to passing
``result_type='expand'``. The resulting column names
will be the Series index.

>>> df.apply(lambda x: pd.Series([1, 2], index=['foo', 'bar']), axis=1)
   foo  bar
0    1    2
1    1    2
2    1    2
3    1    2
4    1    2
5    1    2

Passing ``result_type='broadcast'`` will ensure the same shape
result, whether list-like or scalar is returned by the function,
and broadcast it along the axis. The resulting column names will
be the originals.

>>> df.apply(lambda x: [1, 2, 3], axis=1, result_type='broadcast')
   A  B  C
0  1  2  3
1  1  2  3
2  1  2  3
3  1  2  3
4  1  2  3
5  1  2  3

See also
--------
DataFrame.applymap: For elementwise operations
DataFrame.aggregate: only perform aggregating type operations
DataFrame.transform: only perform transformating type operations

Returns
-------
applied : Series or DataFrame

Errors found:
	Errors in parameters section
		Unknown parameters {'kwds :'}
		Parameter "raw" description should start with capital letter
		Parameter "kwds :" has no type
```

explicitly.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4859,18 +4859,18 @@ class DataFrame(NDFrame):
               achieve much better performance.
         reduce : bool or `None`, default `None`
             Try to apply reduction procedures. If the `DataFrame` is empty,
-            :meth:`apply` will use reduce to determine whether the result
+            :meth:`apply` will use `reduce` to determine whether the result
             should be a Series or a `DataFrame`. If ``reduce is None`` (the
             default), :meth:`apply`'s return value will be guessed by calling
-            func on an empty Series
+            `func` on an empty Series
             (note: while guessing, exceptions raised by `func` will be
             ignored).
-            If reduce is True a Series will always be returned, and if
-            `False` a `DataFrame` will always be returned.
+            If ``reduce is True`` a Series will always be returned, and if
+            ``reduce is False`` a `DataFrame` will always be returned.
 
             .. deprecated:: 0.23.0.
                This argument will be removed in a future version, replaced
-               by result_type='reduce'.
+               by ``result_type='reduce'``.
 
         result_type : {'expand', 'reduce', 'broadcast', `None`}
             These only act when ``axis=1`` (columns):
@@ -4898,9 +4898,9 @@ class DataFrame(NDFrame):
 
         Notes
         -----
-        In the current implementation apply calls func twice on the
+        In the current implementation apply calls `func` twice on the
         first column/row to decide whether it can take a fast or slow
-        code path. This can lead to unexpected behavior if func has
+        code path. This can lead to unexpected behavior if `func` has
         side-effects, as they will take effect twice for the first
         column/row.
 
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4855,8 +4855,8 @@ class DataFrame(NDFrame):
               function.
             * `True` : the passed function will receive ndarray objects
               instead.
-            If you are just applying a NumPy reduction function this will
-            achieve much better performance.
+              If you are just applying a NumPy reduction function this will
+              achieve much better performance.
         reduce : bool or `None`, default `None`
             Try to apply reduction procedures. If the `DataFrame` is empty,
             :meth:`apply` will use reduce to determine whether the result
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4823,9 +4823,9 @@ class DataFrame(NDFrame):
 
         Objects passed to the function are Series objects whose index is
         either the DataFrame's index (``axis=0``) or the DataFrame's columns
-        (``axis=1``). If ``result_type is None``, the final return type is the
-        return type of the applied function. Otherwise, it depends on the
-        `result_type` argument.
+        (``axis=1``). If ``result_type is None``, the final return type is
+        inferred from the return type of the applied function. Otherwise, it
+        depends on the `result_type` argument.
 
         Parameters
         ----------
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4892,7 +4892,7 @@ class DataFrame(NDFrame):
         args : tuple
             Positional arguments to pass to function in addition to the
             array/series.
-        kwds :
+        **kwds
             Additional keyword arguments to pass as keywords to the function.
 
         Notes
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4822,8 +4822,8 @@ class DataFrame(NDFrame):
         Apply a function along an axis of the DataFrame.
 
         Objects passed to the function are Series objects having as index
-        either the DataFrame's index (`axis=0`)
-        or the DataFrame's columns (`axis=1`).
+        either the DataFrame's index (``axis=0``)
+        or the DataFrame's columns (``axis=1``).
         If `result_type` is None, the final return type is the return
         type of the applied function.
         Otherwise, it depends on the `result_type` argument.
@@ -4873,7 +4873,7 @@ class DataFrame(NDFrame):
                by result_type='reduce'.
 
         result_type : {'expand', 'reduce', 'broadcast', None}
-            These only act when `axis=1` (columns):
+            These only act when ``axis=1`` (columns):
 
             * 'expand' : list-like results will be turned into columns.
             * 'reduce' : returns a Series if possible rather than expanding
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4821,7 +4821,7 @@ class DataFrame(NDFrame):
         """
         Apply a function along an axis of the DataFrame.
 
-        Objects passed to the function are Series objects having as index
+        Objects passed to the function are Series objects whose index is
         either the DataFrame's index (``axis=0``)
         or the DataFrame's columns (``axis=1``).
         If ``result_type is None``, the final return type is the return
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4858,15 +4858,15 @@ class DataFrame(NDFrame):
               If you are just applying a NumPy reduction function this will
               achieve much better performance.
         reduce : bool or `None`, default `None`
-            Try to apply reduction procedures. If the `DataFrame` is empty,
+            Try to apply reduction procedures. If the DataFrame is empty,
             :meth:`apply` will use `reduce` to determine whether the result
-            should be a Series or a `DataFrame`. If ``reduce is None`` (the
+            should be a Series or a DataFrame. If ``reduce is None`` (the
             default), :meth:`apply`'s return value will be guessed by calling
             `func` on an empty Series
             (note: while guessing, exceptions raised by `func` will be
             ignored).
             If ``reduce is True`` a Series will always be returned, and if
-            ``reduce is False`` a `DataFrame` will always be returned.
+            ``reduce is False`` a DataFrame will always be returned.
 
             .. deprecated:: 0.23.0.
                This argument will be removed in a future version, replaced
@@ -4879,7 +4879,7 @@ class DataFrame(NDFrame):
             * 'reduce' : returns a Series if possible rather than expanding
               list-like results. This is the opposite of 'expand'.
             * 'broadcast' : results will be broadcast to the original shape
-              of the `DataFrame`, the original index and columns will be
+              of the DataFrame, the original index and columns will be
               retained.
 
             The default behaviour (`None`) depends on the return value of the
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4859,13 +4859,13 @@ class DataFrame(NDFrame):
         reduce : bool or `None`, default `None`
             Try to apply reduction procedures. If the DataFrame is empty,
             `apply` will use `reduce` to determine whether the result
-            should be a Series or a DataFrame. If ``reduce is None`` (the
+            should be a Series or a DataFrame. If ``reduce=None`` (the
             default), `apply`'s return value will be guessed by calling
             `func` on an empty Series
             (note: while guessing, exceptions raised by `func` will be
             ignored).
-            If ``reduce is True`` a Series will always be returned, and if
-            ``reduce is False`` a DataFrame will always be returned.
+            If ``reduce=True`` a Series will always be returned, and if
+            ``reduce=False`` a DataFrame will always be returned.
 
             .. deprecated:: 0.23.0
                This argument will be removed in a future version, replaced
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4867,7 +4867,7 @@ class DataFrame(NDFrame):
             If ``reduce is True`` a Series will always be returned, and if
             ``reduce is False`` a DataFrame will always be returned.
 
-            .. deprecated:: 0.23.0.
+            .. deprecated:: 0.23.0
                This argument will be removed in a future version, replaced
                by ``result_type='reduce'``.
 
@@ -4886,7 +4886,7 @@ class DataFrame(NDFrame):
             of those. However if the apply function returns a Series these
             are expanded to columns.
 
-            .. versionadded:: 0.23.0.
+            .. versionadded:: 0.23.0
 
         args : tuple
             Positional arguments to pass to `func` in addition to the
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4890,10 +4890,11 @@ class DataFrame(NDFrame):
             .. versionadded:: 0.23.0.
 
         args : tuple
-            Positional arguments to pass to function in addition to the
+            Positional arguments to pass to `func` in addition to the
             array/series.
         **kwds
-            Additional keyword arguments to pass as keywords to the function.
+            Additional keyword arguments to pass as keywords arguments to
+            `func`.
 
         Notes
         -----
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4823,9 +4823,9 @@ class DataFrame(NDFrame):
 
         Objects passed to the function are Series objects whose index is
         either the DataFrame's index (``axis=0``) or the DataFrame's columns
-        (``axis=1``). If ``result_type is None``, the final return type is
-        inferred from the return type of the applied function. Otherwise, it
-        depends on the `result_type` argument.
+        (``axis=1``). By default (``result_type=None``), the final return type
+        is inferred from the return type of the applied function. Otherwise,
+        it depends on the `result_type` argument.
 
         Parameters
         ----------
@@ -4839,10 +4839,10 @@ class DataFrame(NDFrame):
         broadcast : bool, optional
             Only relevant for aggregation functions:
 
-            * `False` or `None` : returns a Series whose length is the length
+            * ``False`` or ``None`` : returns a Series whose length is the length
               of the index or the number of columns (based on the `axis`
               parameter)
-            * `True` : results will be broadcast to the original shape
+            * ``True`` : results will be broadcast to the original shape
               of the frame, the original index and columns will be retained.
 
             .. deprecated:: 0.23.0
@@ -4850,9 +4850,9 @@ class DataFrame(NDFrame):
                by result_type='broadcast'.
 
         raw : bool, default False
-            * `False` : passes each row or column as a Series to the
+            * ``False`` : passes each row or column as a Series to the
               function.
-            * `True` : the passed function will receive ndarray objects
+            * ``True`` : the passed function will receive ndarray objects
               instead.
               If you are just applying a NumPy reduction function this will
               achieve much better performance.
@@ -4871,7 +4871,7 @@ class DataFrame(NDFrame):
                This argument will be removed in a future version, replaced
                by ``result_type='reduce'``.
 
-        result_type : {'expand', 'reduce', 'broadcast', `None`}
+        result_type : {'expand', 'reduce', 'broadcast', None}, default None
             These only act when ``axis=1`` (columns):
 
             * 'expand' : list-like results will be turned into columns.
@@ -4881,7 +4881,7 @@ class DataFrame(NDFrame):
               of the DataFrame, the original index and columns will be
               retained.
 
-            The default behaviour `None` depends on the return value of the
+            The default behaviour (None) depends on the return value of the
             applied function: list-like results will be returned as a Series
             of those. However if the apply function returns a Series these
             are expanded to columns.
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4882,7 +4882,7 @@ class DataFrame(NDFrame):
               of the DataFrame, the original index and columns will be
               retained.
 
-            The default behaviour (`None`) depends on the return value of the
+            The default behaviour `None` depends on the return value of the
             applied function: list-like results will be returned as a Series
             of those. However if the apply function returns a Series these
             are expanded to columns.
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4819,9 +4819,9 @@ class DataFrame(NDFrame):
     def apply(self, func, axis=0, broadcast=None, raw=False, reduce=None,
               result_type=None, args=(), **kwds):
         """
-        Apply a function along an axis of the `Series`.
+        Apply a function along an axis of the DataFrame.
 
-        Objects passed to the function are `Series` objects having as index
+        Objects passed to the function are Series objects having as index
         either the DataFrame's index (`axis=0`)
         or the DataFrame's columns (`axis=1`).
         If `result_type` is None, the final return type is the return
@@ -4840,7 +4840,7 @@ class DataFrame(NDFrame):
         broadcast : boolean, optional
             Only relevant for aggregation functions:
 
-            * `False` or `None` : returns a `Series` whose length is the length
+            * `False` or `None` : returns a Series whose length is the length
               of the index or the number of columns (based on the `axis`
               parameter)
             * `True` : results will be broadcast to the original shape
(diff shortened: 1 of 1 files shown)


pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
 
     def apply(self, func, axis=0, broadcast=None, raw=False, reduce=None,
               result_type=None, args=(), **kwds):
-        """Applies function along an axis of the DataFrame.
+        """
+        Apply a function along an axis of the `Series`.
```
datapythonista: Why did you change `DataFrame` by `Series`? The file is `frame.py` I think it's apply for `DataFrame`, isn't it?

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
 
-        Objects passed to functions are Series objects having index
-        either the DataFrame's index (axis=0) or the columns (axis=1).
-        Final return type depends on the return type of the applied function,
-        or on the `result_type` argument.
+        Objects passed to the function are `Series` objects having as index
+        either the DataFrame's index (`axis=0`)
+        or the DataFrame's columns (`axis=1`).
+        If `result_type` is None, the final return type is the return
```
datapythonista: `None` in backticks

pandas/core/frame.py
jorisvandenbossche: This is the dataframe docs, so Series -> DataFrame

(also it is not needed to put single backticks around Series and DataFrame, same for the rest of the docstring)

TomAugspurger: \`Series\` -> DataFrame

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
               result_type=None, args=(), **kwds):
-        """Applies function along an axis of the DataFrame.
+        """
+        Apply a function along an axis of the `Series`.
 
-        Objects passed to functions are Series objects having index
-        either the DataFrame's index (axis=0) or the columns (axis=1).
-        Final return type depends on the return type of the applied function,
-        or on the `result_type` argument.
+        Objects passed to the function are `Series` objects having as index
```
TomAugspurger: "having as" -> "whose index is either"

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
+        either the DataFrame's index (`axis=0`)
```
TomAugspurger: double backticks around `axis=0` and `axis=1` since they're code snippets.

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
         broadcast : boolean, optional
-            For aggregation functions, return object of same size with values
-            propagated
+            Only relevant for aggregation functions:
+
+            * `False` or `None` : returns a `Series` whose length is the length
+              of the index or the number of columns (based on the `axis`
+              parameter)
+            * `True` : results will be broadcast to the original shape
+              of the frame, the original index and columns will be retained.
```
jorisvandenbossche: Since this keyword is deprecated, I am not sure it is that important to add such a detailed description (I would rather expand the explanation of the new keyword `result_type`)

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
-            Function to apply to each column/row
+            Function to apply to each column or row.
         axis : {0 or 'index', 1 or 'columns'}, default 0
-            * 0 or 'index': apply function to each column
-            * 1 or 'columns': apply function to each row
+            Axis along which the function is applied:
+
+            * 0 or 'index': apply function to each column.
+            * 1 or 'columns': apply function to each row.
         broadcast : boolean, optional
```
datapythonista: `bool` instead of `boolean` (also in the next 2)

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
             .. deprecated:: 0.23.0
                This argument will be removed in a future version, replaced
                by result_type='broadcast'.
 
         raw : boolean, default False
-            If False, convert each row or column into a Series. If raw=True the
-            passed function will receive ndarray objects instead. If you are
-            just applying a NumPy reduction function this will achieve much
-            better performance
+            * `False` : passes each row or column into a `Series` to the
```
jorisvandenbossche: "passes each row or column into a `Series`" -> "passes each row or column as a `Series`"

TomAugspurger: Maybe "passes each row or column as a Series to the function"

pandas/core/frame.py
jorisvandenbossche: can you use double backticks around axis=0 (as ``` ``axis=0`` ```) (because it is a small code snippet)

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
+              function.
+            * `True` : the passed function will receive ndarray objects
+              instead.
+            If you are just applying a NumPy reduction function this will
```
TomAugspurger: This applies to `raw=Ture` so keep it under the item formed by that `*`

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
-            of those. However if the apply function returns a Series these
+            The default behaviour (`None`) depends on the return value of the
+            applied function: list-like results will be returned as a `Series`
+            of those. However if the apply function returns a `Series` these
             are expanded to columns.
 
-            .. versionadded:: 0.23.0
+            .. versionadded:: 0.23.0.
 
         args : tuple
```
datapythonista: `args` is always a tuple, so we remove it. The validation script will probably complain, but the convention is to use `*args` and `**kwargs` (with the stars)

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
             Positional arguments to pass to function in addition to the
-            array/series
-        Additional keyword arguments will be passed as keywords to the function
+            array/series.
+        kwds :
```
jorisvandenbossche: kwds -> `**kwds` (and the colon is not needed)

jorisvandenbossche (commented):
Thanks for the PR! Added some first comments

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
-            Try to apply reduction procedures. If the DataFrame is empty,
-            apply will use reduce to determine whether the result should be a
-            Series or a DataFrame. If reduce is None (the default), apply's
-            return value will be guessed by calling func an empty Series (note:
-            while guessing, exceptions raised by func will be ignored). If
-            reduce is True a Series will always be returned, and if False a
-            DataFrame will always be returned.
-
-            .. deprecated:: 0.23.0
+            Try to apply reduction procedures. If the `DataFrame` is empty,
```
TomAugspurger: Don't need backticsk around Series / DataFrame

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
+            :meth:`apply` will use reduce to determine whether the result
```
TomAugspurger: I think this resolves to `pandas.apply`, which doesn't exist. SEe if you can rewreite to avoid the reference to apply

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
+            should be a `Series` or a `DataFrame`. If reduce is None (the
```
TomAugspurger: backticsk around parameter names (`reduce`)

pandas/core/frame.py
```diff
@@ -4818,66 +4818,82 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
-            .. versionadded:: 0.23.0
+            .. versionadded:: 0.23.0.
 
         args : tuple
             Positional arguments to pass to function in addition to the
-            array/series
-        Additional keyword arguments will be passed as keywords to the function
+            array/series.
+        kwds :
+            Additional keyword arguments to pass as keywords to the function.
```
datapythonista: A personal opinion, but for me it'd be a bit more explicit to say "to pass as keyword arguments to `func`". I think saying "the function" should be clear enough for most people, but specifying which functions avoids any ambiguity. Also por `args`.

datapythonista (commented):
Looks quite nice, added some comments.

pandas/core/frame.py
TomAugspurger: Just updated our docs on this, we're doing `**kwds`

pandas/core/frame.py
jorisvandenbossche: Ai, we were all doing the review at the same time :-)

pandas/core/frame.py
jorisvandenbossche: In this case we need to keep it as `args` and the 'tuple' type description is correct. This is a special case because it actually is a *keyword* with the name `args`, not `*args` in the signature

pandas/core/frame.py
mdeboc: add PR review suggestions from @jorisvandenbossche into commit c593a70c5

pandas/core/frame.py
mdeboc: @jorisvandenbossche OK. But <<results will be broadcast to the original shape of the frame, the original index and columns will be retained.>> is already copied from the explanation of the new keyword `result_type`. So nothing to add in the paragraph of this new parameter. 

Shall I revert to the previous wording for `broadcast` parameter, _i.e._:
```
'''
broadcast : boolean, optional
    For aggregation functions, return object of same size with values
    propagated
'''
```
(...which is rather fuzzy),
or not ?

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit 8caf5f615

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit c1214b42c

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit 2432ff573

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit c593a70c5

pandas/core/frame.py
mdeboc: add PR review suggestions from @datapythonista into commit da64c864c
Maybe ``result_type is None`` is better there, because it is a small code snippet.

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit dbd90c1b8

pandas/core/frame.py
mdeboc: Ok, so I let the dot at the end.

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit 6cbadd8ed

pandas/core/frame.py
mdeboc: Yep. Sorry guys :)

pandas/core/frame.py
mdeboc: add PR review suggestions from @TomAugspurger into commit eaf144156

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit c1214b42c

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit 8caf5f615

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit c593a70c5

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit 2432ff573

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit da64c86
Maybe ``reduce is None`` is better there, because it is a small code snippet

pandas/core/frame.py
mdeboc: add this PR review suggestions into commit 3794d1989

pandas/core/frame.py
mdeboc: After running 
```bash
python make.py html --single pandas.DataFrame.apply
```
the :meth:`apply`  link resolves to `pandas.DataFrame.apply`

@TomAugspurger, if it is not the case for you, I shall revert to a wording more like: `apply()` or `DataFrame.apply()`... Let me know.

pandas/core/frame.py
mdeboc: @TomAugspurger I took into account your suggestion in the commit c593a70

pandas/core/frame.py
```diff
@@ -4818,72 +4818,89 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
-            return value will be guessed by calling func an empty Series (note:
-            while guessing, exceptions raised by func will be ignored). If
-            reduce is True a Series will always be returned, and if False a
-            DataFrame will always be returned.
-
-            .. deprecated:: 0.23.0
+            :meth:`apply` will use `reduce` to determine whether the result
+            should be a Series or a DataFrame. If ``reduce is None`` (the
+            default), :meth:`apply`'s return value will be guessed by calling
+            `func` on an empty Series
```
jreback: @jorisvandenbossche can you look at this formatting

pandas/core/frame.py
```diff
@@ -4818,72 +4818,89 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
-              list-like results. This is the opposite to 'expand'.
+            * 'reduce' : returns a Series if possible rather than expanding
+              list-like results. This is the opposite of 'expand'.
             * 'broadcast' : results will be broadcast to the original shape
-              of the frame, the original index & columns will be retained.
+              of the DataFrame, the original index and columns will be
+              retained.
 
-            The default behaviour (None) depends on the return value of the
+            The default behaviour (`None`) depends on the return value of the
```
jreback: no parens here

jreback (changes requested):

pandas/core/frame.py
mdeboc: PR review suggestions from @jreback was added into commit c4da1ea72

pandas/core/frame.py
jorisvandenbossche: @mdeboc Although `` :meth:`apply` `` works correctly in sphinx rendering, I would not use the referencing syntax, as it is then just adding a reference to itself ? 
I would just do `` `apply` `` in this case

mdeboc: @jorisvandenbossche I add this PR review suggestions into commit ee6919efd

pandas/core/frame.py
```diff
@@ -4818,72 +4818,89 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
-        Objects passed to functions are Series objects having index
-        either the DataFrame's index (axis=0) or the columns (axis=1).
-        Final return type depends on the return type of the applied function,
-        or on the `result_type` argument.
+        Objects passed to the function are Series objects whose index is
+        either the DataFrame's index (``axis=0``)
+        or the DataFrame's columns (``axis=1``).
+        If ``result_type is None``, the final return type is the return
+        type of the applied function.
+        Otherwise, it depends on the `result_type` argument.
```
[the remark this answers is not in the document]
mdeboc: add this PR review suggestions into commit 23be01b34

pandas/core/frame.py
```diff
@@ -4818,72 +4818,89 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
 
-        Objects passed to functions are Series objects having index
-        either the DataFrame's index (axis=0) or the columns (axis=1).
-        Final return type depends on the return type of the applied function,
-        or on the `result_type` argument.
+        Objects passed to the function are Series objects whose index is
+        either the DataFrame's index (``axis=0``)
+        or the DataFrame's columns (``axis=1``).
+        If ``result_type is None``, the final return type is the return
+        type of the applied function.
```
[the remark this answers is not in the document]
mdeboc: add this PR review suggestions into commit 51ad0cb6c

pandas/core/frame.py
```diff
@@ -4818,72 +4818,88 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
+              retained.
 
-            The default behaviour (None) depends on the return value of the
+            The default behaviour `None` depends on the return value of the
             applied function: list-like results will be returned as a Series
             of those. However if the apply function returns a Series these
             are expanded to columns.
 
-            .. versionadded:: 0.23.0
+            .. versionadded:: 0.23.0.
```
datapythonista: There is a bug in the script, you don't need the period at the end.

pandas/core/frame.py
datapythonista: yes, my fault, sorry.

pandas/core/frame.py
```diff
@@ -4818,72 +4818,88 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
 
         Notes
         -----
-        In the current implementation apply calls func twice on the
+        In the current implementation apply calls `func` twice on the
         first column/row to decide whether it can take a fast or slow
-        code path. This can lead to unexpected behavior if func has
+        code path. This can lead to unexpected behavior if `func` has
         side-effects, as they will take effect twice for the first
         column/row.
```
datapythonista: Can you get rid of the "We use this DataFrame to illustrate" at the beginning of the examples? I don't think it adds any value.

Also, I'd personally prefer to create a smaller (e.g 3 rows, 2 cols) dataframe from a Python list, instead of all that numpy machinery. :) And if the example is with `sqrt`, I'd use numbers like `9` so the user can quickly see what `apply` does, which is the goal of this example.

datapythonista (commented):
Good work, added some comments, mainly to parts of the original docstring that can be improved.

pandas/core/frame.py
mdeboc: Ok. see cf2f2d06b

pandas/core/frame.py
mdeboc: Yes, it's simpler. See d52ac81e7

mdeboc: @datapythonista In commit 3160fb805, I suppress an unnecessary test, unintentionally introduced in previous commit (d52ac81)

pandas/core/frame.py
```diff
@@ -4818,118 +4818,121 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
-            while guessing, exceptions raised by func will be ignored). If
-            reduce is True a Series will always be returned, and if False a
-            DataFrame will always be returned.
+            `apply` will use `reduce` to determine whether the result
+            should be a Series or a DataFrame. If ``reduce is None`` (the
+            default), `apply`'s return value will be guessed by calling
+            `func` on an empty Series
+            (note: while guessing, exceptions raised by `func` will be
+            ignored).
+            If ``reduce is True`` a Series will always be returned, and if
```
jorisvandenbossche: "reduce is True" -> "reduce=True" (same on line below and few lines above)

pandas/core/frame.py
```diff
@@ -4818,118 +4818,121 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
-            better performance
-        reduce : boolean or None, default None
+        raw : bool, default False
+            * `False` : passes each row or column as a Series to the
+              function.
+            * `True` : the passed function will receive ndarray objects
+              instead.
+              If you are just applying a NumPy reduction function this will
+              achieve much better performance.
+        reduce : bool or `None`, default `None`
```
jorisvandenbossche: no backticks are needed inside the type description part

pandas/core/frame.py
```diff
@@ -4950,36 +4951,27 @@ def apply(self, func, axis=0, broadcast=None, raw=False, reduce=None,
...
-        4  1  2  3
-        5  1  2  3
+        >>> df.apply(lambda x: [1, 2], axis=1, result_type='broadcast')
+           A  B
+        0  1  2
+        1  1  2
+        2  1  2
 
         See also
         --------
```
jorisvandenbossche: Can you move the Returns and See Also sections to just after the Parameters section?

jorisvandenbossche (commented):
Added some minor comments. Looking good for the rest!

pandas/core/frame.py
mdeboc: @jorisvandenbossche Changes in dfa0f5571, but don't forget that this keyword is deprecated, as you mentioned Saturday. :)

pandas/core/frame.py
mdeboc: Sure
changes in d757960db
@jorisvandenbossche  maybe the teams working with functions:  `combine`, `combine_first`, `applymap`, `join` and `round` should do the same...

TomAugspurger (approved):

jorisvandenbossche (approved):

jorisvandenbossche:
@mdeboc Thanks a lot!

PR #20204, opened 2018-03-10
DOC: update the pandas.Series.dropna and pandas.DataFrame.dropna docstring

datadonK23:
an unrelated PR):

- [x]  The html version looks good: `python doc/make.py --single <your-function-or-method>`

Parameter **kwargs causes the only warning in validate script
```
####################### Docstring (pandas.Series.dropna) #######################

Remove missing values.

Return Series without NA values.

Parameters
----------
axis : {0 or 'index'}, default 0
    There is only one axis to drop values from.
    If True, do operation inplace and return None.
**kwargs
    Not in use.

Returns
-------
Series
    Series with NA entries dropped from it.

See also
--------
isna: show missing values
notna : show existing (non-missing) values

Examples
--------
>>> ser = pd.Series([np.NaN, 2, pd.NaT, '', None, 'I stay'])
>>> ser
0       NaN
1         2
2       NaT
3
4      None
5    I stay
dtype: object

Drop NA values from a Series.

>>> ser.dropna()
1         2
3
5    I stay
dtype: object

Drop NA values from a Series and save it in the same variable.

>>> ser.dropna(inplace=True)
>>> ser
1         2
3
5    I stay
dtype: object

Errors found:
        Errors in parameters section
```

```
##################### Docstring (pandas.DataFrame.dropna)  #####################

Remove missing values.

Return object with labels on given axis omitted where alternately any
or all of the data are missing.

Parameters
----------
axis : {0 or 'index', 1 or 'columns'}, or tuple/list thereof
    Pass tuple or list to drop on multiple axes.
how : {'any', 'all'}, default 'any'
    Determine if label is removed when we have at least one NA or all
    NA.
    * any : If any NA values are present, drop that label.
    * all : If all values are NA, drop that label.
thresh : int, default None
    Require that many non-NA values.
subset : array-like
    Labels along other axis to consider, e.g. if you are dropping rows
    these would be a list of columns to include.
    If True, do operation inplace and return None.

Returns
-------
DataFrame
    DataFrame with NA entries dropped from it.

See also
--------
isna: show missing values
notna : show existing (non-missing) values

Examples
--------
>>> df = pd.DataFrame([[np.nan, 2, np.nan, 0], [3, 4, np.nan, 1],
...                    [np.nan, np.nan, np.nan, 5]],
...                   columns=list('ABCD'))
>>> df
     A    B   C  D
0  NaN  2.0 NaN  0
1  3.0  4.0 NaN  1
2  NaN  NaN NaN  5

Drop the columns where all elements are nan:

>>> df.dropna(axis=1, how='all')
     A    B  D
0  NaN  2.0  0
1  3.0  4.0  1
2  NaN  NaN  5

Drop the columns where any of the elements is nan

>>> df.dropna(axis=1, how='any')
   D
0  0
1  1
2  5

Drop the rows where all of the elements are nan
(there is no row to drop, so df stays the same):

>>> df.dropna(axis=0, how='all')
     A    B   C  D
0  NaN  2.0 NaN  0
1  3.0  4.0 NaN  1
2  NaN  NaN NaN  5

Keep only the rows with at least 2 non-na values:

>>> df.dropna(thresh=2)
     A    B   C  D
0  NaN  2.0 NaN  0
1  3.0  4.0 NaN  1

Docstring for "pandas.DataFrame.dropna" correct. :)
```

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3503,8 +3503,8 @@ class DataFrame(NDFrame):
         """
         Remove missing values.
 
-        Return object with labels on given axis omitted where either any
-        or all of the data are missing.
+        See the :ref:`User Guide <missing_data>` for more on which values are
+        considered missing, and how to work with missing data.
 
         Parameters
         ----------
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3539,17 +3539,17 @@ class DataFrame(NDFrame):
         --------
         DataFrame.isna: show missing values
         DataFrame.notna : show existing (non-missing) values
+        DataFrame.fillna : replace missing values
         Series.dropna : drop missing values
-        isna : top-level isna
-        notna : top-level notna
+        Index.dropna : drop missing indices
 
         Examples
         --------
         >>> df = pd.DataFrame([['Alfred', np.nan, pd.NaT],
-        ...                 ['Batman', 'Batmobile', pd.Timestamp('1940-04-25')],
-        ...                 ['Catwoman', 'Bullwhip', pd.NaT],
-        ...                 [np.nan, np.nan, pd.NaT]],
-        ...                 columns=['name', 'toy', 'born'])
+        ...                ['Batman', 'Batmobile', pd.Timestamp('1940-04-25')],
+        ...                ['Catwoman', 'Bullwhip', pd.NaT],
+        ...                [np.nan, np.nan, pd.NaT]],
+        ...                columns=['name', 'toy', 'born'])
         >>> df
                name        toy       born
         0    Alfred        NaN        NaT
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3537,11 +3537,11 @@ class DataFrame(NDFrame):
 
         See Also
         --------
-        DataFrame.isna: show missing values
-        DataFrame.notna : show existing (non-missing) values
-        DataFrame.fillna : replace missing values
-        Series.dropna : drop missing values
-        Index.dropna : drop missing indices
+        DataFrame.isna: Indicate missing values.
+        DataFrame.notna : Indicate existing (non-missing) values.
+        DataFrame.fillna : Replace missing values.
+        Series.dropna : Drop missing values.
+        Index.dropna : Drop missing indices.
 
         Examples
         --------
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3535,46 +3535,70 @@ class DataFrame(NDFrame):
 
         Examples
         --------
-        >>> df = pd.DataFrame([[np.nan, 2, np.nan, 0], [3, 4, np.nan, 1],
-        ...                    [np.nan, np.nan, np.nan, 5]],
+        Drop the rows where at least one element is missing.
+
+        >>> df = pd.DataFrame([[1., 2.], [3., np.nan]], columns=list('AB'))
+        >>> df
+             A    B
+        0  1.0  2.0
+        1  3.0  NaN
+        >>> df.dropna()
+             A    B
+        0  1.0  2.0
+
+        Drop the columns where at least one element is missing.
+
+        >>> df.dropna(axis=1)
+             A
+        0  1.0
+        1  3.0
+
+        Drop the rows where all elements are missing.
+
+        >>> df = pd.DataFrame([[1., 2.], [np.nan, np.nan], [5., np.nan]],
+        ...                   columns=list('AB'))
+        >>> df
+             A    B
+        0  1.0  2.0
+        1  NaN  NaN
+        2  5.0  NaN
+        >>> df.dropna(how='all')
+             A    B
+        0  1.0  2.0
+        2  5.0  NaN
+
+        Keep only the rows with at least 2 non-NA values.
+
+        >>> df = pd.DataFrame([[np.nan, 2., np.nan, 4.], [5., 6., np.nan, 8.],
+        ...                    [np.nan, np.nan, np.nan, 12.]],
         ...                   columns=list('ABCD'))
         >>> df
-             A    B   C  D
-        0  NaN  2.0 NaN  0
-        1  3.0  4.0 NaN  1
-        2  NaN  NaN NaN  5
-
-        Drop the columns where all elements are nan:
-
-        >>> df.dropna(axis=1, how='all')
-             A    B  D
-        0  NaN  2.0  0
-        1  3.0  4.0  1
-        2  NaN  NaN  5
-
-        Drop the columns where any of the elements is nan
-
-        >>> df.dropna(axis=1, how='any')
-           D
-        0  0
-        1  1
-        2  5
-
-        Drop the rows where all of the elements are nan
-        (there is no row to drop, so df stays the same):
-
-        >>> df.dropna(axis=0, how='all')
-             A    B   C  D
-        0  NaN  2.0 NaN  0
-        1  3.0  4.0 NaN  1
-        2  NaN  NaN NaN  5
-
-        Keep only the rows with at least 2 non-na values:
-
+             A    B   C     D
+        0  NaN  2.0 NaN   4.0
+        1  5.0  6.0 NaN   8.0
+        2  NaN  NaN NaN  12.0
         >>> df.dropna(thresh=2)
-             A    B   C  D
-        0  NaN  2.0 NaN  0
-        1  3.0  4.0 NaN  1
+             A    B   C    D
+        0  NaN  2.0 NaN  4.0
+        1  5.0  6.0 NaN  8.0
+
+        Define in which columns to look for missing values.
+
+        >>> df.dropna(subset=['A', 'B'])
+             A    B   C    D
+        1  5.0  6.0 NaN  8.0
+
+        Keep the DataFrame with valid entries in the same variable.
+
+        >>> df = pd.DataFrame([[1., 2.], [3., np.nan]], columns=list('AB'))
+        >>> df
+             A    B
+        0  1.0  2.0
+        1  3.0  NaN
+        >>> df.dropna(inplace=True)
+        >>> df
+             A    B
+        0  1.0  2.0
         """
         inplace = validate_bool_kwarg(inplace, 'inplace')
         if isinstance(axis, (tuple, list)):
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3515,12 +3515,12 @@ class DataFrame(NDFrame):
             NA.
             * any : If any NA values are present, drop that label.
             * all : If all values are NA, drop that label.
-        thresh : int, default None
+        thresh : int or None, default None
             Require that many non-NA values.
         subset : array-like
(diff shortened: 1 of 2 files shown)


pandas/core/frame.py
```diff
@@ -3501,27 +3501,37 @@ def notnull(self):
...
         inplace : boolean, default False
             If True, do operation inplace and return None.
 
         Returns
         -------
-        dropped : DataFrame
+        DataFrame
+            DataFrame with NA entries dropped from it.
+
+        See also
```
DataOmbudsman: According to the guide this should be written as See Also.

pandas/core/series.py
```diff
@@ -3039,13 +3039,57 @@ def notnull(self):
...
 
         Returns
         -------
-        valid : Series
-        inplace : boolean, default False
-            Do operation in place.
+        Series
+            Series with NA entries dropped from it.
+
+        See also
```
DataOmbudsman: According to the guide this should be written as See Also.

DataOmbudsman:
Very nice docstrings IMO, thanks!

pandas/core/frame.py
jorisvandenbossche: indeed

pandas/core/frame.py
```diff
@@ -3501,27 +3501,37 @@ def notnull(self):
     def dropna(self, axis=0, how='any', thresh=None, subset=None,
                inplace=False):
         """
+        Remove missing values.
+
         Return object with labels on given axis omitted where alternately any
-        or all of the data are missing
+        or all of the data are missing.
```
jorisvandenbossche: It was already there, but I am not fully sure what is meant with "alternately" ? (that implies something alternating)

pandas/core/frame.py
```diff
@@ -3501,27 +3501,37 @@ def notnull(self):
...
-        how : {'any', 'all'}
-            * any : if any NA values are present, drop that label
-            * all : if all values are NA, drop that label
+            Pass tuple or list to drop on multiple axes.
+        how : {'any', 'all'}, default 'any'
+            Determine if label is removed when we have at least one NA or all
+            NA.
+            * any : If any NA values are present, drop that label.
+            * all : If all values are NA, drop that label.
         thresh : int, default None
```
datapythonista: we usually write `int or None, default None`

pandas/core/frame.py
```diff
@@ -3501,27 +3501,37 @@ def notnull(self):
...
-            int value : require that many non-NA values
+            Require that many non-NA values.
         subset : array-like
             Labels along other axis to consider, e.g. if you are dropping rows
-            these would be a list of columns to include
+            these would be a list of columns to include.
         inplace : boolean, default False
```
datapythonista: `bool` instead of `boolean`

pandas/core/frame.py
```diff
@@ -3501,27 +3501,37 @@ def notnull(self):
...
+        or all of the data are missing.
 
         Parameters
         ----------
         axis : {0 or 'index', 1 or 'columns'}, or tuple/list thereof
-            Pass tuple or list to drop on multiple axes
-        how : {'any', 'all'}
-            * any : if any NA values are present, drop that label
-            * all : if all values are NA, drop that label
+            Pass tuple or list to drop on multiple axes.
```
jorisvandenbossche: Can you also explain the other keyword -> drop rows (0 or 'index') or columns (1, or 'columns')

pandas/core/frame.py
```diff
@@ -3501,27 +3501,37 @@ def notnull(self):
...
+        how : {'any', 'all'}, default 'any'
+            Determine if label is removed when we have at least one NA or all
```
jorisvandenbossche: I would maybe replace 'label' with 'row or column', as in the end it is the row/column that you remove (and of course, that means that label is also removed in the index)

pandas/core/frame.py
```diff
@@ -3501,27 +3501,37 @@ def notnull(self):
...
+            NA.
+            * any : If any NA values are present, drop that label.
```
jorisvandenbossche: can you put quotes around any (and all below) like `'any'` (to clearly indicate this are string values to pass)

pandas/core/frame.py
jorisvandenbossche: `default None` -> `optional`

pandas/core/frame.py
```diff
@@ -3501,27 +3501,37 @@ def notnull(self):
...
+            * all : If all values are NA, drop that label.
         thresh : int, default None
-            int value : require that many non-NA values
+            Require that many non-NA values.
         subset : array-like
```
jorisvandenbossche: can you add `, optional` at the end

pandas/core/frame.py
```diff
@@ -3565,7 +3575,6 @@ def dropna(self, axis=0, how='any', thresh=None, subset=None,
              A    B   C  D
         0  NaN  2.0 NaN  0
         1  3.0  4.0 NaN  1
```
datapythonista: It would be nice to add some more examples. Starting by creating a simple dataframe, using `dropna` without parameters, and then explain and show examples using the parameters. In the documentation for the sprint you should have examples.

datapythonista (commented):
Nice PR, added some comments (some apply to both docstrings)

pandas/core/frame.py
```diff
@@ -3501,27 +3501,37 @@ def notnull(self):
...
 
         Returns
         -------
-        dropped : DataFrame
+        DataFrame
+            DataFrame with NA entries dropped from it.
+
+        See also
+        --------
+        isna: show missing values
```
jorisvandenbossche: Can you make this DataFrame.isna (the other also exists, but I think the method version is more appropriate here to link to) 

same for notna below

jorisvandenbossche (commented):

Additional comment about the examples: it would maybe be good to add an example using `subset` (as variation on the existing ones), as this is very useful as well

pandas/core/frame.py
donK23: Think it should mean "alternative". But will change it to "either", because should be easier to understand.

pandas/core/frame.py
```diff
@@ -3501,71 +3501,113 @@ def notnull(self):
...
 
         Returns
         -------
-        dropped : DataFrame
+        DataFrame
+            DataFrame with NA entries dropped from it.
+
+        See Also
+        --------
+        DataFrame.isna: show missing values
```
jreback: add Series.dropna

pandas/core/series.py
```diff
@@ -3039,13 +3039,57 @@ def notnull(self):
...
+
+        Return Series without NA values.
+
+        Parameters
+        ----------
+        axis : {0 or 'index'}, default 0
+            There is only one axis to drop values from.
+        inplace : bool, default False
+            If True, do operation inplace and return None.
+        **kwargs
```
jreback: so we should just remove kwargs entirely (and need to remove the check in the code as well). This was originally fro compat with DataFrame but I think its ok (its possible that we are testing this though, so may need to modify a test). Could do this in a followup. cc @jorisvandenbossche

pandas/core/series.py
```diff
@@ -3039,13 +3039,57 @@ def notnull(self):
...
-        valid : Series
-        inplace : boolean, default False
-            Do operation in place.
+        Series
+            Series with NA entries dropped from it.
+
+        See Also
+        --------
+        isna: show missing values
+        notna : show existing (non-missing) values
```
jreback: add DataFrame.dropna

jreback (changes requested):

pandas/core/frame.py
```diff
@@ -3503,78 +3503,111 @@ def dropna(self, axis=0, how='any', thresh=None, subset=None,
...
-        0  0
-        1  1
-        2  5
+        >>> df = pd.DataFrame([[1., 2.], [np.nan, np.nan], [5., np.nan]],
+        ...                   columns=list('AB'))
+        >>> df
+             A    B
+        0  1.0  2.0
+        1  NaN  NaN
+        2  5.0  NaN
```
datapythonista: Personally, I'd define this dataframe at the beginning (before the first explanations), and reuse it for all the examples. Or one with 3 columns, if you think it's not descriptive enough when using subset.

I think reusing the same data will make much more obvious the differences about every case.

datapythonista (commented):
Looks great, just added a comment that I think should make the examples better.

pandas/core/frame.py
donK23: Agreed, will rework the examples. Was uncertain, whether to write simpler & specific ones for each parameter or just a general one, that might be more complex at the first glance but can be reused. So will take your proposed approach.

pandas/core/series.py
donK23: So should I delete the **kwargs from the docstring already, or later when the code is fixed? And can I do the follow-up PR on this issue in the code base, or will you do it when the testing is clarified?

pandas/core/frame.py
```diff
@@ -3503,78 +3503,111 @@ def dropna(self, axis=0, how='any', thresh=None, subset=None,
...
         DataFrame
             DataFrame with NA entries dropped from it.
 
-        See also
+        See Also
         --------
-        isna: show missing values
-        notna : show existing (non-missing) values
+        DataFrame.isna: show missing values
+        DataFrame.notna : show existing (non-missing) values
```
jorisvandenbossche: maybe also add `DataFrame.fillna`

pandas/core/frame.py
```diff
@@ -3503,78 +3503,111 @@ def dropna(self, axis=0, how='any', thresh=None, subset=None,
...
+        isna : top-level isna
+        notna : top-level notna
```
jorisvandenbossche: Since we already have quite some 'see also entries' (certainly with the comments), I would leave out those two top-level ones (they will be linked in the isna/notna methods)

pandas/core/frame.py
jorisvandenbossche: Yes, I think this is a good idea

pandas/core/series.py
jorisvandenbossche: Let's do this in a follow-up, and keep this PR purely docstring. So @donK23 you can leave it as is for now. 
And certainly welcome to do such a follow-up PR

jorisvandenbossche (commented):
For the examples in Series.dropna: the current example shows some corner cases (empty string, None) to show how they are or are not regarded as missing values. It's good to keep that, but I would add an 'normal' examples to start with

donK23:
Implemented proposed changes. I'm not absolutely happy with the axis=1 example in DataFrame.dropna (returns empty df), but otherwise there would a second df be needed, which would complicate the section again.

TomAugspurger:
@donK23 can you take a look at the change I made in  979ac8e? That removed your last row of all NA values. Is there a reason you wanted an all-NA row? Removing it solves your issue of the `df.dropna(axis='columns')` example being empty.

TomAugspurger (approved):

TomAugspurger:
Thanks @donK23

PR #20208, opened 2018-03-10
DOC: update the pandas.errors.DtypeWarning docstring

hissashirocha:
an unrelated PR):

```
#################### Docstring (pandas.errors.DtypeWarning) ####################

Warning raised when reading different dtypes in a column from a file.

Raised for a dtype incompatibility. This can happen whenever `read_csv`
or `read_table` encounter non-uniform dtypes in a column(s) of a given
CSV file.

See Also
--------
pandas.read_csv : Read CSV (comma-separated) file into a DataFrame.
pandas.read_table : Read general delimited file into a DataFrame.

Notes
-----
This warning is issued when dealing with larger files because the dtype
checking happens per chunk read.

Despite the warning, the CSV file is read with mixed types in a single
column which will be an object type. See the examples below to better
understand this issue.

Examples
--------
This example creates and reads a large CSV file with a column that contains
`int` and `str`.

>>> df = pd.DataFrame({'a':['1']*100000 + ['X']*100000 + ['1']*100000,
...                    'b':['b']*300000})
>>> df.to_csv('test.csv', index=False)
>>> df2 = pd.read_csv('test.csv')
>>> DtypeWarning: Columns (0) have mixed types... # doctest: +SKIP

Important to notice that df2 will contain both `str` and `int` for the
same input, '1'.

>>> df2.iloc[262140, 0]
'1'
>>> type(df2.iloc[262140, 0])
<class 'str'>
>>> df2.iloc[262150, 0]
1
>>> type(df2.iloc[262150, 0])
<class 'int'>

One way to solve this issue is using the parameter `converters` in the
`read_csv` and `read_table` functions to explicit the conversion:

>>> df2 = pd.read_csv('test', sep=' ', converters={'a': str})
scripts/validate_docstrings.py:268: DtypeWarning: Columns (0) have mixed types. Specify dtype option on import or set low_memory=False.
  runner.run(test)

Errors found:
	No returns section found
```

### DIFF
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -63,7 +63,6 @@ class DtypeWarning(Warning):
     This example creates and reads a large CSV file with a column that contains
     `int` and `str`.
 
-    >>> import os
     >>> df = pd.DataFrame({'a': (['1'] * 100000 + ['X'] * 100000 +
     ...                          ['1'] * 100000),
     ...                    'b': ['b'] * 300000})
@@ -91,6 +90,7 @@ class DtypeWarning(Warning):
 
     No warning was issued.
 
+    >>> import os
     >>> os.remove('test.csv')
     """
 
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -13,12 +13,14 @@ class PerformanceWarning(Warning):
     performance impact.
     """
 
+
 class UnsupportedFunctionCall(ValueError):
     """
     Exception raised when attempting to call a numpy function
     on a pandas object, but that function is not supported by
     the object e.g. ``np.cumsum(groupby_object)``.
     """
+
 
 class UnsortedIndexError(KeyError):
     """
@@ -38,9 +40,53 @@ class ParserError(ValueError):
 
 class DtypeWarning(Warning):
     """
-    Warning that is raised for a dtype incompatibility. This
-    can happen whenever `pd.read_csv` encounters non-
-    uniform dtypes in a column(s) of a given CSV file.
+    Warning raised when importing different dtypes in a column from a file.
+
+    Raised for a dtype incompatibility. This can happen whenever `pd.read_csv`
+    or `pd.read_table` encounter non-uniform dtypes in a column(s) of a given
+    CSV file.
+
+    It only happens when dealing with larger files.
+
+    See Also
+    --------
+    pd.read_csv : Read CSV (comma-separated) file into a DataFrame.
+    pd.read_table : Read general delimited file into a DataFrame.
+
+    Notes
+    -----
+    Despite the warning, the CSV file is imported with mixed types in a single
+    column. See the examples below to better understand this issue.
+
+    Examples
+    --------
+    This example creates and reads a large CSV file with a column that contains
+    `int` and `str`.
+
+    >>> df = pd.DataFrame({'a':['1']*100000 + ['X']*100000 + ['1']*100000,
+    ...                    'b':['b']*300000})
+    >>> df.to_csv('test', sep='\t', index=False, na_rep='NA')
+    >>> df2 = pd.read_csv('test', sep='\t')
+    Traceback (most recent call last):
+    ...
+    DtypeWarning: Columns (0) have mixed types...
+
+    Important to notice that df2 will contain both `str` and `int` for the
+    same input, '1'.
+
+    >>> df2.iloc[262140,0]
+    '1'
+    >>> type(df2.iloc[262140,0])
+    <class 'str'>
+    >>> df2.iloc[262150,0]
+    1
+    >>> type(df2.iloc[262150,0])
+    <class 'int'>
+
+    One way to solve this issue is using the parameter `converters` in the
+    `read_csv` and `read_table` functions to explicit the conversion:
+
+    >>> df2 = pd.read_csv('test', sep='\t', converters={'a': str})
     """
 
 
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -13,14 +13,12 @@ class PerformanceWarning(Warning):
     performance impact.
     """
 
-
 class UnsupportedFunctionCall(ValueError):
     """
     Exception raised when attempting to call a numpy function
     on a pandas object, but that function is not supported by
     the object e.g. ``np.cumsum(groupby_object)``.
     """
-
 
 class UnsortedIndexError(KeyError):
     """
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -67,7 +67,6 @@ class DtypeWarning(Warning):
     ...                    'b':['b']*300000})
     >>> df.to_csv('test', sep='\t', index=False, na_rep='NA')
     >>> df2 = pd.read_csv('test', sep='\t')
-    Traceback (most recent call last):
     ...
     DtypeWarning: Columns (0) have mixed types...
 
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -63,13 +63,16 @@ class DtypeWarning(Warning):
     This example creates and reads a large CSV file with a column that contains
     `int` and `str`.
 
-    >>> df = pd.DataFrame({'a':['1']*100000 + ['X']*100000 + ['1']*100000,
-    ...                    'b':['b']*300000})
+    >>> import os
+    >>> df = pd.DataFrame({'a': (['1'] * 100000 + ['X'] * 100000 +
+    ...                          ['1'] * 100000),
+    ...                    'b': ['b'] * 300000})
     >>> df.to_csv('test.csv', index=False)
     >>> df2 = pd.read_csv('test.csv')
-    >>> DtypeWarning: Columns (0) have mixed types... # doctest: +SKIP
 
-    Important to notice that df2 will contain both `str` and `int` for the
+        DtypeWarning: Columns (0) have mixed types
+
+    Important to notice that ``df2`` will contain both `str` and `int` for the
     same input, '1'.
 
     >>> df2.iloc[262140, 0]
@@ -81,10 +84,14 @@ class DtypeWarning(Warning):
     >>> type(df2.iloc[262150, 0])
     <class 'int'>
 
-    One way to solve this issue is using the parameter `converters` in the
+    One way to solve this issue is using the `dtype` parameter in the
     `read_csv` and `read_table` functions to explicit the conversion:
 
-    >>> df2 = pd.read_csv('test', sep='\t', converters={'a': str})
+    >>> df2 = pd.read_csv('test.csv', sep=',', dtype={'a': str})
+
+    No warning was issued.
+
+    >>> os.remove('test.csv')
     """
 
 
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -38,23 +38,25 @@ class ParserError(ValueError):
 
 class DtypeWarning(Warning):
     """
-    Warning raised when importing different dtypes in a column from a file.
+    Warning raised when reading different dtypes in a column from a file.
 
-    Raised for a dtype incompatibility. This can happen whenever `pd.read_csv`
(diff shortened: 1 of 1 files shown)


pandas/errors/__init__.py
```diff
@@ -38,9 +38,53 @@ class ParserError(ValueError):
...
+
+    Raised for a dtype incompatibility. This can happen whenever `pd.read_csv`
+    or `pd.read_table` encounter non-uniform dtypes in a column(s) of a given
+    CSV file.
+
+    It only happens when dealing with larger files.
+
+    See Also
+    --------
+    pd.read_csv : Read CSV (comma-separated) file into a DataFrame.
```
TomAugspurger: pd -> pandas I think.

pandas/errors/__init__.py
```diff
@@ -38,9 +38,53 @@ class ParserError(ValueError):
...
+    >>> df.to_csv('test', sep='\t', index=False, na_rep='NA')
+    >>> df2 = pd.read_csv('test', sep='\t')
+    Traceback (most recent call last):
+    ...
+    DtypeWarning: Columns (0) have mixed types...
+
+    Important to notice that df2 will contain both `str` and `int` for the
+    same input, '1'.
+
+    >>> df2.iloc[262140,0]
```
TomAugspurger: PEP8: space after comma

pandas/errors/__init__.py
```diff
@@ -38,9 +38,53 @@ class ParserError(ValueError):
...
+    Examples
+    --------
+    This example creates and reads a large CSV file with a column that contains
+    `int` and `str`.
+
+    >>> df = pd.DataFrame({'a':['1']*100000 + ['X']*100000 + ['1']*100000,
+    ...                    'b':['b']*300000})
+    >>> df.to_csv('test', sep='\t', index=False, na_rep='NA')
+    >>> df2 = pd.read_csv('test', sep='\t')
+    Traceback (most recent call last):
```
TomAugspurger: I think if yyou add a line like

    >>> df2 = pd.read_csv('test', sep='\t')
    ... # doctest: +SKIP

The doctest might work.

pandas/errors/__init__.py
```diff
@@ -38,9 +38,53 @@ class ParserError(ValueError):
 
 class DtypeWarning(Warning):
     """
-    Warning that is raised for a dtype incompatibility. This
-    can happen whenever `pd.read_csv` encounters non-
-    uniform dtypes in a column(s) of a given CSV file.
+    Warning raised when importing different dtypes in a column from a file.
+
+    Raised for a dtype incompatibility. This can happen whenever `pd.read_csv`
```
jorisvandenbossche: Can you also use `read_csv` here instead of `pd.read_csv` ? (same below for read_table)

pandas/errors/__init__.py
```diff
@@ -38,9 +38,53 @@ class ParserError(ValueError):
...
+    column. See the examples below to better understand this issue.
+
+    Examples
+    --------
+    This example creates and reads a large CSV file with a column that contains
+    `int` and `str`.
+
+    >>> df = pd.DataFrame({'a':['1']*100000 + ['X']*100000 + ['1']*100000,
+    ...                    'b':['b']*300000})
+    >>> df.to_csv('test', sep='\t', index=False, na_rep='NA')
```
jorisvandenbossche: can you add '.csv' for the temp file ? 
I would also leave out the `sep='\t'` and `na_rep='NA'`, as they are not relevant to show illustrate the warning

pandas/errors/__init__.py
jorisvandenbossche: Yeah, I don't think it will be possible to get Warnings to work with doctest, so adding a `# doctest: +SKIP` is fine

pandas/errors/__init__.py
```diff
@@ -38,9 +38,53 @@ class ParserError(ValueError):
 
 class DtypeWarning(Warning):
     """
-    Warning that is raised for a dtype incompatibility. This
-    can happen whenever `pd.read_csv` encounters non-
-    uniform dtypes in a column(s) of a given CSV file.
+    Warning raised when importing different dtypes in a column from a file.
```
jreback: when reading (not importing)

pandas/errors/__init__.py
```diff
@@ -38,9 +38,53 @@ class ParserError(ValueError):
...
+
+    Raised for a dtype incompatibility. This can happen whenever `pd.read_csv`
+    or `pd.read_table` encounter non-uniform dtypes in a column(s) of a given
+    CSV file.
+
+    It only happens when dealing with larger files.
```
jreback: instead of this comment, you can add in the Notes section, that this can happen in larger files. Its because the dtype checking happens per chunk that is read.

pandas/errors/__init__.py
```diff
@@ -38,9 +38,53 @@ class ParserError(ValueError):
...
+
+    See Also
+    --------
+    pd.read_csv : Read CSV (comma-separated) file into a DataFrame.
+    pd.read_table : Read general delimited file into a DataFrame.
+
+    Notes
+    -----
+    Despite the warning, the CSV file is imported with mixed types in a single
+    column. See the examples below to better understand this issue.
```
jreback: imported -> read. The dtype of the column will be object.

jreback (changes requested):

pandas/errors/__init__.py
hissashirocha: OK!

pandas/errors/__init__.py
hissashirocha: Done!

pandas/errors/__init__.py
hissashirocha: I'm sorry but I couldn't understand what the doctest: +SKIP is supposed to do. I've added it right after the read_csv and i just got an error. Can you help me?

Line 28, in pandas.errors.DtypeWarning
Failed example:
    df2 = pd.read_csv('test', sep=' ')
Expected:
    # doctest: +SKIP
    DtypeWarning: Columns (0) have mixed types...
Got nothing

pandas/errors/__init__.py
hissashirocha: It's done!

pandas/errors/__init__.py
hissashirocha: Done!

pandas/errors/__init__.py
hissashirocha: Tks!

pandas/errors/__init__.py
hissashirocha: Tks!

pandas/errors/__init__.py
```diff
@@ -38,9 +38,53 @@ class ParserError(ValueError):
...
+    <class 'str'>
+    >>> df2.iloc[262150, 0]
+    1
+    >>> type(df2.iloc[262150, 0])
+    <class 'int'>
+
+    One way to solve this issue is using the parameter `converters` in the
+    `read_csv` and `read_table` functions to explicit the conversion:
+
+    >>> df2 = pd.read_csv('test', sep='\t', converters={'a': str})
```
TomAugspurger: FYI, this is still `test` not `test.csv`

TomAugspurger (commented):
@jorisvandenbossche running this doctest will leave behind a `test.csv` file, which isn't ideal.

We can

- convert this to a regular code-block
- Use `io.StringIO`, which is uglier but doesn't leave behind a file
- Add a bit to `os.remove('test.csv')`

Preferences?

jorisvandenbossche:
Hmm, yes, that's a general problem though, so let's discuss that somewhere else

jorisvandenbossche:
@TomAugspurger see https://github.com/pandas-dev/pandas/issues/20302

pandas/errors/__init__.py
```diff
@@ -38,9 +38,53 @@ class ParserError(ValueError):
...
+    Examples
+    --------
+    This example creates and reads a large CSV file with a column that contains
+    `int` and `str`.
+
+    >>> df = pd.DataFrame({'a':['1']*100000 + ['X']*100000 + ['1']*100000,
+    ...                    'b':['b']*300000})
+    >>> df.to_csv('test.csv', index=False)
+    >>> df2 = pd.read_csv('test.csv')
+    >>> DtypeWarning: Columns (0) have mixed types... # doctest: +SKIP
```
jorisvandenbossche: Can you remove the `>>>` on this line? (it might be you need to move the "# doctest: +SKIP" to the line above after the read_csv)

pandas/errors/__init__.py
jorisvandenbossche: Another thing: I think we should recommend `dtype={'a': str}` instead ?

jorisvandenbossche (commented):
Small comment in addition to the one of @TomAugspurger 

Looks good for the rest!

pandas/errors/__init__.py
hissashirocha: @jorisvandenbossche @TomAugspurger I've added "import os" and an "os.remove('test.csv')" at the end of the example, just as you've agreed on #20302 , and also put "# doctest: +SKIP" right before the warning, like this: 

    >>> import os
    >>> df = pd.DataFrame({'a':['1']*100000 + ['X']*100000 + ['1']*100000,
    ...                    'b':['b']*300000})
    >>> df.to_csv('test.csv', index=False)
    >>> df2 = pd.read_csv('test.csv')
    >>> os.remove('test.csv')
    # doctest: +SKIP
    DtypeWarning: Columns (0) have mixed types...

But I got this message on validation:

Line 32, in pandas.errors.DtypeWarning
Failed example:
    os.remove('test.csv')
Expected:
    # doctest: +SKIP
    DtypeWarning: Columns (0) have mixed types...
Got nothing


And also, removing ".csv" from to_csv and read_csv, like Tom suggested, raises two possible errors:
>>> os.remove('test.csv'): FileNotFoundError: [Errno 2] No such file or directory: 'test.csv'
>>> os.remove('test'): FileNotFoundError: File b'test' does not exist

What do you guys recommend?

TomAugspurger (approved):

TomAugspurger:
I went back and forth on where to put the import :)

jorisvandenbossche:
@TomAugspurger the only problem now is that the actual warning is separated from the code block, which looks a bit strange in the html output .. 
But if don't leave the blank line, you get a failing doctest ..

The alternative you mentioned would be

```
>>> import warnings
>>> with warnings.catch_warnings(record=True) as w:
...     df2 = pd.read_csv('test.csv')
>>> w[0].message
pandas.errors.DtypeWarning('Columns (0) have mixed types. Specify dtype option on import or set low_memory=False.')
```

But I agree this boilerplate is a bit unfortunate and distracts from what we actually want to show

jorisvandenbossche:
Let's go ahead with the current solution for now. We can always later look into adding our own doctest-extensions (it should be possible to add doctest options while running our own doctests: eg https://github.com/astropy/pytest-doctestplus/blob/master/pytest_doctestplus/output_checker.py)

jorisvandenbossche (approved):

jorisvandenbossche:
@hissashirocha Thanks for the PR!

TomAugspurger:
Ah was just going to push this

```
diff --git a/pandas/errors/__init__.py b/pandas/errors/__init__.py
index 16654f022..ff34df64c 100644
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -68,8 +68,7 @@ class DtypeWarning(Warning):
     ...                    'b': ['b'] * 300000})
     >>> df.to_csv('test.csv', index=False)
     >>> df2 = pd.read_csv('test.csv')
-
-        DtypeWarning: Columns (0) have mixed types
+    ... # DtypeWarning: Columns (0) have mixed types
```

Looks OK. I'll make a new PR.

jorisvandenbossche:
Ah, that's a good one :)

hissashirocha:
It was a pleasure! This is my first time contributing to pandas and I really enjoyed it. Thanks for your help!

PR #20210, opened 2018-03-10
DOC: update the Index.get_level_values docstring

bkil-syslogng:
an unrelated PR):

```
################## Docstring (pandas.Index.get_level_values)  ##################

Return an Index of values for requested level.

Return an Index of values for requested level, equal to the length
of the index.

Parameters
----------
level : int or str
    It is either the integer position of the level in the
    MultiIndex, or the `name` of the level.
    If given as an integer, it must be between 0 and the
    number of levels.

Returns
-------
values : Index
    `self`, as there is only one level in the Index.

See also
---------
pandas.MultiIndex.get_level_values : get values for a level of a
                                     MultiIndex

Notes
---------
For `Index`, level should be 0, since there are no multiple levels.

Examples
---------

Create an Index:

>>> idx = pd.Index(list('abc'))

Get level value by supplying level as integer:

>>> idx.get_level_values(0)
Index(['a', 'b', 'c'], dtype='object')

Create a MultiIndex:

>>> mi = pd.MultiIndex.from_arrays((list('abc'), list('def')))
>>> mi.names = ['level_1', 'level_2']

Get level values by supplying level as name:

>>> mi.get_level_values('level_2')
Index(['d', 'e', 'f'], dtype='object', name='level_2')

Docstring for "pandas.Index.get_level_values" correct. :)
```


pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,55 @@ def set_value(self, arr, key, value):
...
         Return an Index of values for requested level, equal to the length
         of the index.
 
         Parameters
         ----------
         level : int or str
-            ``level`` is either the integer position of the level in the
-            MultiIndex, or the name of the level.
+            It is either the integer position of the level in the
+            MultiIndex, or the `name` of the level.
```
TomAugspurger: Don't think name has to be in backticks, since it isn't a parameter

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,55 @@ def set_value(self, arr, key, value):
...
+            If given as an integer, it must be between 0 and the
+            number of levels.
```
TomAugspurger: Number of levels - 1?

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,55 @@ def set_value(self, arr, key, value):
...
-        """
 
+        Notes
+        ---------
+        For `Index`, level should be 0, since there are no multiple levels.
+
+        Examples
+        ---------
+
+        Create an Index:
```
TomAugspurger: Probalby don't need this line. I'd remove it and add a line showing `idx`

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,55 @@ def set_value(self, arr, key, value):
 
     def _get_level_values(self, level):
         """
+        Return an Index of values for requested level.
```
joaoavf: Return an Index of values for requested level. --> Return an Index of values for requested level, specially useful for MultiIndex.

It is important to make it clear somehow that this is most useful when dealing with a MultiIndex.

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,55 @@ def set_value(self, arr, key, value):
...
+        >>> idx = pd.Index(list('abc'))
+
+        Get level value by supplying level as integer:
+
+        >>> idx.get_level_values(0)
+        Index(['a', 'b', 'c'], dtype='object')
+
+        Create a MultiIndex:
+
+        >>> mi = pd.MultiIndex.from_arrays((list('abc'), list('def')))
```
TomAugspurger: can pass `names=` to `from_arrays`

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,55 @@ def set_value(self, arr, key, value):
 
     def _get_level_values(self, level):
         """
+        Return an Index of values for requested level.
+
         Return an Index of values for requested level, equal to the length
```
TomAugspurger: I would add that this is mainly for compatibility with MultiIndex.

pandas/core/indexes/base.py
TomAugspurger: Since MultiIndex has its own docstring, you probably don't have to reference it here.

TomAugspurger:
Reading through this, it may make sense to say that this isn't very useful for non-MultiIndex indexes. A small example should be OK, but the most important bit is the See Also to MultiIndex.get_level_values.

joaoavf (commented):
Very good, I would just make it clear in the summary and extended summary that this method is specially useful to deal with MultiIndex.

pandas/core/indexes/base.py
bkil-syslogng: Although line wrapping would be needed and this case and it would still end up in two lines, which isn't as nice in my opinion. Although I'll change it as you've requested.

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,55 @@ def set_value(self, arr, key, value):
...
 
         See also
         ---------
         pandas.MultiIndex.get_level_values : get values for a level of a
                                              MultiIndex
-        """
 
+        Notes
+        ---------
+        For `Index`, level should be 0, since there are no multiple levels.
```
jreback: no backticks

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,55 @@ def set_value(self, arr, key, value):
...
+
+        Examples
+        ---------
+
+        Create an Index:
+
+        >>> idx = pd.Index(list('abc'))
+
```
jreback: I wouldn't bother show this for an Index.

jreback (changes requested):

pandas/core/indexes/base.py
bkil-syslogng: Please elaborate what kind of change you would like to us to make. MultiIndex already mentions the other example. In our view, working on an Index is also useful to see for users somewhere, and this would be a proper place for it.

bkil-syslogng:
Most of the notes have been addressed and force pushed.

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,45 @@ def set_value(self, arr, key, value):
 
     def _get_level_values(self, level):
         """
+        Return an Index of values for requested level, specially useful for
```
jreback: need to be 1 line. you can remove after the comma.

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,45 @@ def set_value(self, arr, key, value):
...
+            `self`, as there is only one level in the Index.
 
         See also
         ---------
         pandas.MultiIndex.get_level_values : get values for a level of a
                                              MultiIndex
-        """
 
+        Notes
+        ---------
```
jreback: make the same length as Notes

jreback (changes requested):

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,45 @@ def set_value(self, arr, key, value):
 
     def _get_level_values(self, level):
         """
+        Return an Index of values for requested level, specially useful for
+        MultiIndex.
+
         Return an Index of values for requested level, equal to the length
```
TomAugspurger: This is pretty repetitive with the summary. I think just the bit about MultiIndex is useful.

    The is primarily useful to get an individual level of values from a MultiIndex,
    but is provided on Index as well for compatability.

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,45 @@ def set_value(self, arr, key, value):
...
-        of the index.
+        of the index. This is mainly useful for MultiIndex.
 
         Parameters
         ----------
         level : int or str
-            ``level`` is either the integer position of the level in the
-            MultiIndex, or the name of the level.
+            It is either the integer position or the name of the level.
+            If given as an integer, it must be between 0 and the
```
TomAugspurger: Ah this isn't quite right. Negative indices are accepted as well. Probably OK to just remove the second sentence, as to most Python programmers, "integer position" is clear enough.

pandas/core/indexes/base.py
```diff
@@ -2788,26 +2788,45 @@ def set_value(self, arr, key, value):
...
+        For Index, level should be 0, since there are no multiple levels.
+
+        Examples
+        ---------
+
+        >>> idx = pd.Index(list('abc'))
+        >>> idx
+        Index(['a', 'b', 'c'], dtype='object')
+
+        Get level value by supplying level as integer:
```
TomAugspurger: backtick around \`level`. Add "an" between "as integer"

jorisvandenbossche:
@bkil-syslogng Do you have time to update the PR based on the last comments?

jorisvandenbossche:
Rebased and added final fixes.

@bkil-syslogng Thanks!

PR #20211, opened 2018-03-10
DOC: update the pd.Index.contains docstring

dajcs:
an unrelated PR):

```

###################### Docstring (pandas.Index.contains)  ######################

Return a boolean if key is in the index.

This function returns True if key is in the index and
returns False if key is not in index.

Parameters
----------
key : immutable object
    Key to be searched.

Returns
-------
bool
    Result of the key search.

Notes
-----
    CategoricalIndex is a sub-class of Index.

Examples
--------
>>> pd.CategoricalIndex([2000, 2001, 2012]).contains(2001)
True
>>> pd.CategoricalIndex([2000, 2001, 2012]).contains(2222)
False

Errors found:
        See Also section not found
```

explicitly.

See Also missing because we haven't found similar functions

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -1832,15 +1832,31 @@ class Index(IndexOpsMixin, PandasObject):
             return False
 
     _index_shared_docs['contains'] = """
-        return a boolean if this key is IN the index
+        Return a boolean if key is in the index.
+
+        This function returns True if key is in the index and
+        returns False if key is not in index.
 
         Parameters
         ----------
-        key : object
+        key : immutable object
+            Key to be searched.
 
         Returns
         -------
-        boolean
+        bool
+            Result of the key search.
+
+        Notes
+        -----
+            CategoricalIndex is a sub-class of Index.
+
+        Examples
+        --------
+        >>> pd.CategoricalIndex([2000, 2001, 2012]).contains(2001)
+        True
+        >>> pd.CategoricalIndex([2000, 2001, 2012]).contains(2222)
+        False
         """
 
     @Appender(_index_shared_docs['contains'] % _index_doc_kwargs)


pandas/core/indexes/base.py
```diff
@@ -1832,15 +1832,31 @@ def __contains__(self, key):
             return False
 
     _index_shared_docs['contains'] = """
-        return a boolean if this key is IN the index
+        Return a boolean if key is in the index.
+
+        This function returns True if key is in the index and
```
TomAugspurger: I think this extended summary can be removed.

pandas/core/indexes/base.py
```diff
@@ -1832,15 +1832,31 @@ def __contains__(self, key):
...
+        returns False if key is not in index.
 
         Parameters
         ----------
-        key : object
+        key : immutable object
```
TomAugspurger: can remove `immutable`, it could be anything.

pandas/core/indexes/base.py
```diff
@@ -1832,15 +1832,31 @@ def __contains__(self, key):
...
 
         Returns
         -------
-        boolean
+        bool
+            Result of the key search.
+
+        Notes
+        -----
+            CategoricalIndex is a sub-class of Index.
```
TomAugspurger: Don't think you need this Notes section.

TomAugspurger: Could you add a See Also with Index.isin

pandas/core/indexes/base.py
```diff
@@ -1832,15 +1832,31 @@ def __contains__(self, key):
...
+
+        Examples
+        --------
+        >>> pd.CategoricalIndex([2000, 2001, 2012]).contains(2001)
```
TomAugspurger: Best to just use a regular Index here, not a CategoricalIndex.

pandas/core/indexes/base.py
dajcs: can't be anything:
pd.CategoricalIndex([2000, 2001, 2012]).contains([1,2])
TypeError: unhashable type: 'list'

pandas/core/indexes/base.py
dajcs: pd.Index.contains and pd.CategoricalIndex.contains have shared documentation 
and a CategoricalIndex example is valid for Index but not the other way around since CategoricalIndex is a sub-class of Index

pandas/core/indexes/base.py
dajcs: this note is because of shared documentation between Index and CategoricalIndex

pandas/core/indexes/base.py
jreback: yeah for this particular case I would separate the doc-strings, IOW 1 more Index and a separate one for CategoricalIndex. Can use a shared doc if that works (with templates)

jreback (changes requested):

jorisvandenbossche:
@dajcs do you have time to update the PR based on the feedback?

Tanya-Jain:
I am working to update this PR in accordance with the reviews made to it and match it to the Pandas docstring standards.

datapythonista:
Superseded by #23100 and 23107

PR #20212, opened 2018-03-10
DOC: update the pandas.DataFrame.clip docstring

Dpananos:
an unrelated PR):

```

Errors found:
	Errors in parameters section
		Unknown parameters {'*args, **kwargs'}
		Parameter "*args, **kwargs" has no type
(pandas_dev) Demetris-Retina-MacBook-Pro:pandas demetri$ git diff upstream/master -u -- "*.py" | flake8 --diff
(pandas_dev) Demetris-Retina-MacBook-Pro:pandas demetri$ python scripts/validate_docstrings.py pandas.DataFrame.clip

###################### Docstring (pandas.DataFrame.clip)  ######################

Trim values at input threshold(s).

Elements above/below the upper/lower thresholds will be changed to
upper/lower thresholds. Clipping data is a method for dealing with
out-of-range elements. If some elements are too large or too small,
clipping is one way to transform the data into a reasonable range.

Parameters
----------
lower : float, array-like or None, default None
    Lower threshold for clipping.  Values smaller than `lower` will be
    converted to `lower`.
upper : float, array-like or None, default None
    Upper threshold for clipping.  Values larger than `upper` will be
    converted to `upper`.
axis : {0 or 'index', 1 or 'columns', None}, default None
    Apply clip by index (i.e. by rows) or columns.
    Whether to perform the operation in place on the data
        .. versionadded:: 0.21.0.
*args, **kwargs
    Additional keywords have no effect but might be accepted
    for compatibility with numpy.

Returns
-------
`Series` or `DataFrame`.
    Original input with those values above/below the
    `upper`/`lower` thresholds set to the threshold values.

References
-----
.. [1] Tukey, John W. "The future of data analysis." The annals of
    mathematical statistics 33.1 (1962): 1-67.

See Also
--------
DataFrame.clip : Trim values at input threshold(s).
Series.clip : Trim values at input threshold(s).
Series.clip_lower : Return copy of the input with values below given
    value(s) truncated.
Series.clip_upper : Return copy of input with values above given
    value(s) truncated.
DataFrame.clip_lower : Return copy of the input with values below given
    value(s) truncated.
DataFrame.clip_upper : Return copy of input with values above given
    value(s) truncated.
DataFrame.quantile : Return values at the given quantile over requested
    axis, a la numpy.percentile.

Examples
--------
>>> df = pd.DataFrame({'a': [-1, -2, -100],
...                    'b': [1, 2, 100]},
...                    index=['foo', 'bar', 'foobar'])
>>> df
     a  b
foo -1  1
bar -2  2
foobar  -100    100

>>> df.clip(lower=-10, upper=10)
     a  b
foo -1  1
bar -2  2
foobar  -10 10

You can clip each column or row with different thresholds by passing
a ``Series`` to the lower/upper argument. Use the axis argument to clip
by column or rows.

>>> col_thresh = pd.Series({'a': -5, 'b': 5})
>>> df.clip(lower=col_thresh, axis='columns')
     a  b
foo -1  5
bar -2  5
foobar  -5  100

Clip the foo, bar, and foobar rows with lower thresholds 5, 7, and 10.

>>> row_thresh = pd.Series({'foo': 0, 'bar': 1, 'foobar': 10})
>>> df.clip(lower=row_thresh, axis='index')
    a   b
foo 0   1
bar 1   2
foobar  10  100

Winsorizing [1]_ is a related method, whereby the data are clipped at
the 5th and 95th percentiles. The ``DataFrame.quantile`` method returns
a ``Series`` with column names as index and the quantiles as values.
Use ``axis='columns'`` to apply clipping to columns.

>>> lower, upper = df.quantile(0.05), df.quantile(0.95)
>>> df.clip(lower=lower, upper=upper, axis='columns')
     a  b
foo -1.1    1.1
bar -2.0    2.0
foobar  -90.2   90.2

Errors found:
	Errors in parameters section
		Unknown parameters {'*args, **kwargs'}
		Parameter "*args, **kwargs" has no type
```

Changes:

* `*args` and `**kwargs` description taken from #20256 in the name of consistency

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -5629,6 +5629,12 @@ class NDFrame(PandasObject, SelectionMixin):
             Original input with those values above/below the
             `upper`/`lower` thresholds set to the threshold values.
 
+        Notes
+        -----
+
+        .. [1] Tukey, John W. "The future of data analysis." The annals of
+            mathematical statistics 33.1 (1962): 1-67.
+
         See Also
         --------
         DataFrame.clip : Trim values at input threshold(s).
@@ -5650,13 +5656,13 @@ class NDFrame(PandasObject, SelectionMixin):
         ...                    'b': [1, 2, 100]},
         ...                    index=['foo', 'bar', 'foobar'])
         >>> df
-            a   b
+             a  b
         foo -1  1
         bar -2  2
         foobar  -100    100
 
         >>> df.clip(lower=-10, upper=10)
-            a   b
+             a  b
         foo -1  1
         bar -2  2
         foobar  -10 10
@@ -5667,7 +5673,7 @@ class NDFrame(PandasObject, SelectionMixin):
 
         >>> col_thresh = pd.Series({'a': -5, 'b': 5})
         >>> df.clip(lower=col_thresh, axis='columns')
-            a   b
+             a  b
         foo -1  5
         bar -2  5
         foobar  -5  100
@@ -5681,15 +5687,14 @@ class NDFrame(PandasObject, SelectionMixin):
         bar 1   2
         foobar  10  100
 
-        `Winsorizing <https://en.wikipedia.org/wiki/Winsorizing>`__ is a
-        related method, whereby the data are clipped at
+        Winsorizing [1]_ is a related method, whereby the data are clipped at
         the 5th and 95th percentiles. The ``DataFrame.quantile`` method returns
         a ``Series`` with column names as index and the quantiles as values.
         Use ``axis='columns'`` to apply clipping to columns.
 
         >>> lower, upper = df.quantile(0.05), df.quantile(0.95)
         >>> df.clip(lower=lower, upper=upper, axis='columns')
-            a   b
+             a  b
         foo -1.1    1.1
         bar -2.0    2.0
         foobar  -90.2   90.2
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -5601,6 +5601,9 @@ class NDFrame(PandasObject, SelectionMixin):
         """
         Trim values at input threshold(s).
 
+        Elements above the upper threshold will be changed to upper threshold.
+        Elements below the lower threshold will be changed to lower threshold.
+
         Parameters
         ----------
         lower : float or array_like, default None
@@ -5609,53 +5612,69 @@ class NDFrame(PandasObject, SelectionMixin):
             Align object with lower and upper along the given axis.
         inplace : boolean, default False
             Whether to perform the operation in place on the data
-                .. versionadded:: 0.21.0
+                .. versionadded:: 0.21.0.
 
         Returns
         -------
         clipped : Series
 
+        Notes
+        -----
+        Clipping data is a method for dealing with dubious elements.
+        If some elements are too large or too small, clipping is one way to
+        transform the data into a reasonable range.
+
+        See Also
+        --------
+        pandas.DataFrame.clip_upper : Return copy of input with values
+            above given value(s) truncated.
+        pandas.DataFrame.clip_lower : Return copy of input with values
+            below given value(s) truncated.
+
         Examples
         --------
+        >>> df = pd.DataFrame({'a':[1, 2, 3], 'b':[4, 5, 6], 'c':[7, 8, 9001]})
         >>> df
             a   b   c
         0   1   4   7
         1   2   5   8
         2   3   6   9001
 
-
-        >>> df.clip( lower = 1 , upper = 9 )
+        >>> df.clip(lower = 1, upper = 9)
             a   b   c
         0   1   4   7
         1   2   5   8
         2   3   6   9
 
-        
-        You can clip each column with different thresholds by passing a Series to the lower/upper argument.
+        You can clip each column with different thresholds by passing
+        a Series to the lower/upper argument.
 
+        >>> some_data = {'A':[-19, 12, -5],'B':[1, 100, -5]}
+        >>> df = pd.DataFrame(data = some_data, index = ['foo', 'bar', 'bizz'])
         >>> df
               A   B
         foo  -19  1
         bar   12 100
         bizz -5  -5
-        
-        >>> df.clip( lower = pd.Series({'A': -10, 'B': 10}) , axis = 1 )
 
+        Use the axis argument to clip by column or rows.  Clip column A with
+        lower threshold of -10 and column B has lower threshold of 10.
+
+        >>> df.clip(lower = pd.Series({'A':-10, 'B':10}), axis = 1)
               A   B
         foo  -10 10
         bar   12 100
         bizz -5  10
 
-        Use the axis argument to clip by column or rows.
+        Clip the foo, bar, and bizz rows with lower thresholds -10, 0, and 10.
 
-        >>> df.clip( lower = pd.Series({'foo':-10,'bar':0, 'bizz':10}) , axis = 0 )
+        >>> row_thresh = pd.Series({'foo':-10, 'bar':0, 'bizz':10})
+        >>> df.clip(lower = row_thresh ,axis = 0)
              A   B
         foo  -10 1
         bar   12 100
         bizz  10 10
-
         """
-        
         if isinstance(self, ABCPanel):
             raise NotImplementedError("clip is not supported yet for panels")
 
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -5613,7 +5613,7 @@ class NDFrame(PandasObject, SelectionMixin):
             Upper threshold for clipping.  Values larger than `upper` will be
             converted to `upper`.
         axis : {0 or 'index', 1 or 'columns', None}, default None
-            Align object with lower and upper along the given axis.
+            Apply clip by index (i.e. by rows) or columns.
         inplace : boolean, default False
             Whether to perform the operation in place on the data
                 .. versionadded:: 0.21.0.
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -5631,11 +5631,18 @@ class NDFrame(PandasObject, SelectionMixin):
 
         See Also
         --------
+        DataFrame.clip : Trim values at input threshold(s).
         Series.clip : Trim values at input threshold(s).
-        DataFrame.clip_upper : Return copy of input with values above given
+        Series.clip_lower : Return copy of the input with values below given
+            value(s) truncated.
+        Series.clip_upper : Return copy of input with values above given
             value(s) truncated.
         DataFrame.clip_lower : Return copy of the input with values below given
             value(s) truncated.
+        DataFrame.clip_upper : Return copy of input with values above given
+            value(s) truncated.
+        DataFrame.quantile : Return values at the given quantile over requested
+            axis, a la numpy.percentile.
 
         Examples
         --------
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -5617,6 +5617,8 @@ class NDFrame(PandasObject, SelectionMixin):
         Returns
         -------
         clipped : DataFrame/Series
+            Elements above or below the upper and lower thresholds converted to
+            threshold values.
 
         Notes
         -----
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -5635,7 +5635,7 @@ class NDFrame(PandasObject, SelectionMixin):
         Examples
         --------
         >>> some_data = {'a': [1, 2, 3], 'b': [4, 5, 6], 'c': [7, 8, 9001]}
-        >>> df = pd.DataFrame(some_data, index = ['foo','bar','foobar'])
+        >>> df = pd.DataFrame(some_data, index = ['foo', 'bar', 'foobar'])
         >>> df
             a   b   c
         foo 1   4   7
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -5629,9 +5629,8 @@ class NDFrame(PandasObject, SelectionMixin):
             Original input with those values above/below the
             `upper`/`lower` thresholds set to the threshold values.
 
-        Notes
+        References
         -----
-
         .. [1] Tukey, John W. "The future of data analysis." The annals of
             mathematical statistics 33.1 (1962): 1-67.
 
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -5602,7 +5602,9 @@ class NDFrame(PandasObject, SelectionMixin):
         Trim values at input threshold(s).
 
         Elements above/below the upper/lower thresholds will be changed to
-        upper/lower thresholds.
+        upper/lower thresholds. Clipping data is a method for dealing with
+        out-of-range elements. If some elements are too large or too small,
+        clipping is one way to transform the data into a reasonable range.
 
         Parameters
         ----------
@@ -5617,67 +5619,70 @@ class NDFrame(PandasObject, SelectionMixin):
         inplace : boolean, default False
             Whether to perform the operation in place on the data
                 .. versionadded:: 0.21.0.
-        *args : Additional keywords have no effect but might be accepted
-            for compatibility with numpy.
-        **kwargs : Additional keywords have no effect but might be accepted
+        *args, **kwargs
+            Additional keywords have no effect but might be accepted
             for compatibility with numpy.
 
         Returns
         -------
         `Series` or `DataFrame`.
-            DataFrame is returned with those values above/below the
+            Original input with those values above/below the
             `upper`/`'lower` thresholds set to the threshold values.
 
         See Also
         --------
-        pandas.Series.clip : Trim values at input threshold(s).
+        Series.clip : Trim values at input threshold(s).
+        DataFrame.clip_upper : Return copy of input with values above given
+            value(s) truncated.
+        DataFrame.clip_lower : Return copy of the input with values below given
+            value(s) truncated.
 
         Examples
         --------
-        >>> some_data = {'a': [1, 2, 3], 'b': [4, 5, 6], 'c': [7, 8, 9001]}
-        >>> df = pd.DataFrame(some_data, index = ['foo', 'bar', 'foobar'])
+        >>> some_data = {'a': [-1, -2, -100], 'b': [1, 2, 100]}
+        >>> df=pd.DataFrame(some_data, index = ['foo', 'bar', 'foobar'])
         >>> df
-            a   b   c
-        foo 1   4   7
-        bar 2   5   8
-        foobar  3   6   9001
+            a   b
+        foo -1  1
+        bar -2  2
+        foobar  -100    100
 
-        >>> df.clip(lower=1, upper=9)
-            a   b   c
-        foo 1   4   7
-        bar 2   5   8
-        foobar  3   6   9
+        >>> df.clip(lower=-10, upper=10)
+            a   b
+        foo -1  1
+        bar -2  2
+        foobar  -10 10
 
         You can clip each column or row with different thresholds by passing
         a ``Series`` to the lower/upper argument. Use the axis argument to clip
         by column or rows.
 
-        >>> col_thresh = pd.Series({'a':4, 'b':5, 'c':6})
+        >>> col_thresh=pd.Series({'a':-5, 'b':5})
         >>> df.clip(lower=col_thresh, axis='columns')
-            a   b   c
-        foo 4   5   7
-        bar 4   5   8
-        foobar  4   6   9001
+            a   b
+        foo -1  5
+        bar -2  5
+        foobar  -5  100
 
         Clip the foo, bar, and foobar rows with lower thresholds 5, 7, and 10.
 
-        >>> row_thresh = pd.Series({'foo': 5, 'bar': 7, 'foobar': 10})
+        >>> row_thresh=pd.Series({'foo': 0, 'bar': 1, 'foobar': 10})
         >>> df.clip(lower=row_thresh, axis='index')
-            a   b   c
-        foo 5   5   7
-        bar 7   7   8
-        foobar  10  10  9001
+            a   b
+        foo 0   1
+        bar 1   2
+        foobar  10  100
 
-        Clipping data is a method for dealing with out-of-range elements.
-        If some elements are too large or too small, clipping is one way to
-        transform the data into a reasonable range.
         `Winsorizing <https://en.wikipedia.org/wiki/Winsorizing>`__ is a
         related method, whereby the data are clipped at
         the 5th and 95th percentiles.
 
-        >>> lwr_thresh = df.quantile(0.05)
-        >>> upr_thresh = df.quantile(0.95)
-        >>> dfw = df.clip(lower=lwr_thresh, upper=upr_thresh, axis='columns')
+        >>> lower, upper = df.quantile(0.05), df.quantile(0.95)
+        >>> df.clip(lower=lower, upper=upper, axis='columns')
+            a   b
+        foo -1.1    1.1
+        bar -2.0    2.0
+        foobar  -90.2   90.2
         """
         if isinstance(self, ABCPanel):
             raise NotImplementedError("clip is not supported yet for panels")
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -5692,7 +5692,7 @@ class NDFrame(PandasObject, SelectionMixin):
         >>> x = np.random.normal(size=(1000,3))
         >>> U = df.quantile(0.95)
         >>> L = df.quantile(0.5)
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -5601,53 +5601,99 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+
         Parameters
         ----------
         lower : float or array_like, default None
+            Lower threshold for clipping.  Values smaller than upper will be
+            converted to lower.
         upper : float or array_like, default None
+            Upper threshold for clipping.  Values larger than upper will be
+            converted to upper.
         axis : int or string axis name, optional
```
datapythonista: We added a standard for axis to the documentation, at the end of the `Paramters` section. Basically it's `axis : {0 or 'index', 1 or 'columns', None}, default None`

pandas/core/generic.py
```diff
@@ -5601,53 +5601,99 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
             Align object with lower and upper along the given axis.
         inplace : boolean, default False
             Whether to perform the operation in place on the data
-                .. versionadded:: 0.21.0
+                .. versionadded:: 0.21.0.
+        args : dictionary of arguments arguments passed to pandas.compat.numpy
+        kwargs : dictionary of keyword arguments passed to pandas.compat.numpy
```
datapythonista: `*args` and `**kwargs` with the stars (even if the validate script complaints), and without the type (the description part goes in the nest line)

pandas/core/generic.py
```diff
@@ -5601,53 +5601,99 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
 
         Returns
         -------
-        clipped : Series
+        clipped : DataFrame/Series
```
datapythonista: `Series` or `DataFrame`

pandas/core/generic.py
```diff
@@ -5601,53 +5601,99 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        See Also
+        --------
+        pandas.DataFrame.clip_upper : Return copy of input with values
+            above given value(s) truncated.
+        pandas.DataFrame.clip_lower : Return copy of input with values
+            below given value(s) truncated.
 
         Examples
         --------
+        >>> df=pd.DataFrame({'a':[1, 2, 3], 'b':[4, 5, 6], 'c':[7, 8, 9001]})
```
datapythonista: This have several PEP-8 issues because of missing spaces

pandas/core/generic.py
```diff
@@ -5601,53 +5601,99 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+            a   b   c
+        0   1   4   7
+        1   2   5   8
+        2   3   6   9
+
+        You can clip each column or row with different thresholds by passing
+        a ``Series`` to the lower/upper argument.
+
+        >>> some_data={'A':[-19, 12, -5],'B':[1, 100, -5]}
+        >>> df=pd.DataFrame(data=some_data, index=['foo', 'bar', 'bizz'])
```
datapythonista: more PEP-8

datapythonista (commented):
Goo contribution, added some comments

cmdelatorre (changes requested):
`args` and `kwargs` are passed to `validate_clip_with_axis`, that is not trivial. It means that `args` and `kwargs` actually make sense. These should be clear in the docstring. Check here: https://github.com/pandas-dev/pandas/blob/master/pandas/compat/numpy/function.py#L137

pandas/core/generic.py
```diff
@@ -5601,53 +5601,99 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
         Returns
         -------
-        clipped : Series
+        clipped : DataFrame/Series
+            Elements above or below the upper and lower thresholds converted to
+            threshold values.
+
+        Notes
+        -----
+        Clipping data is a method for dealing with dubious elements.
```
jreback: prefer, out-of-range to dubious

pandas/core/generic.py
```diff
@@ -5601,53 +5601,99 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        If some elements are too large or too small, clipping is one way to
+        transform the data into a reasonable range.
+
+        See Also
+        --------
+        pandas.DataFrame.clip_upper : Return copy of input with values
+            above given value(s) truncated.
```
jreback: add Series.clip

pandas/core/generic.py
```diff
@@ -5601,53 +5601,99 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+             A   B
+        foo  -10 1
+        bar   12 100
+        bizz  10 10
+
+        `Winsorizing <https://en.wikipedia.org/wiki/Winsorizing>`__ is a way
+        of removing outliers from data.  Columns of a DataFrame can be
+        winsorized by using clip.
+
+        >>> import numpy as np
```
jreback: don't need the numpy import

pandas/core/generic.py
```diff
@@ -5601,53 +5601,99 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        >>> x=np.random.normal(size=(1000,3))
+        >>> df=pd.DataFrame(x, columns=['a','b','c'])
+        >>> #Winsorize columns at 5% and 95%
```
jreback: you can add this as part of the text above

pandas/core/generic.py
```diff
@@ -5601,53 +5601,99 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        >>> U=df.quantile(0.95)
```
jreback: spaces around equals

jreback (changes requested):

Dpananos:
@cmdelatorre `validate_clip_with_axis` is going over my head.  I'm not sure what `validate_clip_with_axis` does.  Open to suggestions on what to add to docstring, or would appreciate reading suggestions.

pandas/core/generic.py
```diff
@@ -5601,53 +5601,98 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+            converted to `lower`.
+        upper : float, array-like or None, default None
+            Upper threshold for clipping.  Values larger than `upper` will be
+            converted to `upper`.
+        axis : {0 or 'index', 1 or 'columns', None}, default None
             Align object with lower and upper along the given axis.
         inplace : boolean, default False
             Whether to perform the operation in place on the data
-                .. versionadded:: 0.21.0
+                .. versionadded:: 0.21.0.
```
datapythonista: I find the description of axis a bit complex.

pandas/core/generic.py
```diff
@@ -5601,53 +5601,98 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        *args : arguments passed to pandas.compat.numpy
+        **kwargs : keyword arguments passed to pandas.compat.numpy
```
datapythonista: It's not in the documentation yet, but after some discussion some minutes ago, we'll have `*args, **kwargs` in a single line (no colon), and a single description in the next, as it's rarely different between them.

pandas/core/generic.py
```diff
@@ -5601,53 +5601,98 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
 
         Returns
         -------
-        clipped : Series
+        clipped : `Series` or `DataFrame`.
+            Elements above or below the upper and lower thresholds converted to
+            threshold values.
```
datapythonista: We can just have the type in the first row of Returns, providing a name doesn't add much value.

The description sounds a bit like if we could be returning only part of the original values.

pandas/core/generic.py
```diff
@@ -5601,53 +5601,98 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+
+        Notes
+        -----
+        Clipping data is a method for dealing with out-of-range elements.
+        If some elements are too large or too small, clipping is one way to
+        transform the data into a reasonable range.
```
datapythonista: This sounds more like part of the extended summary to me, than Notes, which is usually left for details on the implementaiton (e.g. calling this function makes a copy of the data)

pandas/core/generic.py
datapythonista: this is in generic, isn't it being reused by Series.clip?

pandas/core/generic.py
```diff
@@ -5601,53 +5601,98 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+            a   b   c
+        0   1   4   7
+        1   2   5   8
+        2   3   6   9
+
+        You can clip each column or row with different thresholds by passing
+        a ``Series`` to the lower/upper argument.
+
+        >>> some_data = {'A': [-19, 12, -5], 'B': [1, 100, -5]}
+        >>> df = pd.DataFrame(data=some_data, index=['foo', 'bar', 'bizz'])
```
datapythonista: I would create `df` just once at the beginning (with 2 cols), and reuse it all the time.

I've usually seen `foo`, `bar` and `foobar`, not important, but I'd use that unless there is another convention that uses `bizz`

pandas/core/generic.py
```diff
@@ -5601,53 +5601,98 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        >>> df
+              A   B
+        foo  -19  1
+        bar   12 100
+        bizz -5  -5
+
+        Use the axis argument to clip by column or rows.  Clip column A with
+        lower threshold of -10 and column B has lower threshold of 10.
```
datapythonista: two spaces after the dot?

pandas/core/generic.py
```diff
@@ -5601,53 +5601,98 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        bizz  10 10
+
+        `Winsorizing <https://en.wikipedia.org/wiki/Winsorizing>`__ is a way
+        of removing outliers from data.  Columns of a DataFrame can be
+        winsorized at 5th and 95th percentile by using clip.
+
+        >>> x = np.random.normal(size=(1000,3))
+        >>> U = df.quantile(0.95)
+        >>> L = df.quantile(0.5)
+        >>> winsorized_df = df.clip(lower=L, upper=U, axis=1)
```
datapythonista: I like having this here. But couple of thoughts (feel free to disagree in both, they are open questions):
- Should we add `quantile` to See Also, if we consider this a common case?
- I'd probably like more to use the previous `df` even if this makes more sense with normally distributed data

Regardless of these two points, this example has some typos:
- Missing space after the comma in the `size` tuple
- `x` is defined but later you're using `df`
- I don't see a reason to use capital letters for `L` and `U`, I don't see them as constants even if they are not modified again in the example
- I would use descriptive names, not a single letter
- Makes more sense to define lower first

datapythonista (commented):
I really like this contribution. Made some comments.

pandas/core/generic.py
Dpananos: Yep, bad habit.

pandas/core/generic.py
Dpananos: Sure thing.

pandas/core/generic.py
Dpananos: Should I just remove it then?

datapythonista: I think it's an interesting comment. May be you can also mention about the outlier thing you show in the example. But I'd have it in the extended summary. After moving it, make sure the whole summary makes sense and doesn't sound repetitive. Usually happens when you move blocks.

pandas/core/generic.py
Dpananos: I'll work on these tomorrow and update the PR then.  Thanks for the feedback.

pandas/core/generic.py
```diff
@@ -5601,53 +5601,83 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        axis : {0 or 'index', 1 or 'columns', None}, default None
+            Apply clip by index (i.e. by rows) or columns.
         inplace : boolean, default False
             Whether to perform the operation in place on the data
-                .. versionadded:: 0.21.0
+                .. versionadded:: 0.21.0.
+        *args : Additional keywords have no effect but might be accepted
+            for compatibility with numpy.
+        **kwargs : Additional keywords have no effect but might be accepted
+            for compatibility with numpy.
```
datapythonista: It was some change to the convention during the sprint, and args and kwargs should be in the same line `*args, **kwargs` and a common description in the next line. We didn't find any case where args and kwargs have different description, and it's repetitive the way it was originally defined.

Also, note that `*args` are positional arguments, I don't think it's correct to name they keywords.

pandas/core/generic.py
```diff
@@ -5601,53 +5601,83 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
 
         Returns
         -------
-        clipped : Series
+        `Series` or `DataFrame`.
+            DataFrame is returned with those values above/below the
+            `upper`/`'lower` thresholds set to the threshold values.
```
datapythonista: If return can be `Series`, the description is not accurate, saying the `DataFrame` is returned. May be something like "Original input with values above/below..."?

pandas/core/generic.py
```diff
@@ -5601,53 +5601,83 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+
+        See Also
+        --------
+        pandas.Series.clip : Trim values at input threshold(s).
```
datapythonista: I think the prefix `pandas` is not needed, just `Series.clip`. Also, not sure if that was already discussed, but don't we want `clip_lower` and `clip_upper` here?

pandas/core/generic.py
```diff
@@ -5601,53 +5601,83 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
 
         Examples
         --------
+        >>> some_data = {'a': [1, 2, 3], 'b': [4, 5, 6], 'c': [7, 8, 9001]}
+        >>> df = pd.DataFrame(some_data, index = ['foo', 'bar', 'foobar'])
```
datapythonista: The spaces around `=` are not PEP-8 compliant. Sorry the validation script doesn't check it yet.

pandas/core/generic.py
```diff
@@ -5601,53 +5601,83 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+            a   b   c
+        foo 1   4   7
+        bar 2   5   8
+        foobar  3   6   9001
+
+        >>> df.clip(lower=1, upper=9)
+            a   b   c
+        foo 1   4   7
+        bar 2   5   8
+        foobar  3   6   9
```
datapythonista: I think this example could be more concise, and help people understand the concept faster.

I think 2 columns is enough, have 3 doesn't add much value, but takes some extra time to check what's going on with the example.

Also, the lower trimming by 1 not having effect, makes me a bit confused, on whether I'm understanding it or not. Also it's somethimg really minor, but I found the 9001 unnecessarily big and arbitrary (using `100` wouldn't make me thing there is any special on it..

Feel free to use the example you want, but using something like -1, -2, -100, 1, 2, 100 and trimming at -10 and 10 would make it more clear and straight to the point.

pandas/core/generic.py
```diff
@@ -5601,53 +5601,83 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+
+        You can clip each column or row with different thresholds by passing
+        a ``Series`` to the lower/upper argument. Use the axis argument to clip
+        by column or rows.
+
+        >>> col_thresh = pd.Series({'a':4, 'b':5, 'c':6})
```
datapythonista: spaces after colons for PEP-8

pandas/core/generic.py
```diff
@@ -5601,53 +5601,83 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        Clipping data is a method for dealing with out-of-range elements.
+        If some elements are too large or too small, clipping is one way to
+        transform the data into a reasonable range.
+        `Winsorizing <https://en.wikipedia.org/wiki/Winsorizing>`__ is a
+        related method, whereby the data are clipped at
+        the 5th and 95th percentiles.
+
+        >>> lwr_thresh = df.quantile(0.05)
+        >>> upr_thresh = df.quantile(0.95)
+        >>> dfw = df.clip(lower=lwr_thresh, upper=upr_thresh, axis='columns')
```
datapythonista: as you're using df now, I think you should show the output, instead of saving it to a variable.

Also, it's more a personal taste, feel free to ignore the commend, but these abbreviations feel a bit cryptic. `lower_threshold` would be better, but I think just `lower` it's probably even better.

And just an idea, in case you like it (it's how I'd write it and find more readable, but feel free to leave it as it):

`>>> lower, upper = df.quantile(0.05), df.quantile(0.95)`

datapythonista (commented):
Looking great, I few more comments.

pandas/core/generic.py
Dpananos: I think you previously mentioned that those were generic.  Did we want them included still?

datapythonista: Sorry, I think I wasn't clear enough. You should check it, but as this docstring is in `generic.py` I assume it'll be used by both `Series` and `DataFrame`. That's why I said it was (the docstring) was generic.

So, as your assigned docstring was `DataFrame.clip` (and it's what it's in the title of the PR) but you're actually working also in `Series.clip`, I don't think it makes sense to only add `Series.clip` as @jreback suggested. I'd say that we probably want also `DataFrame.clip`. Even if it'll be a bit weird to have in the `See Also` the same method which is being documented, but I think it's all right. In a separate PR we could use a template and just have the right one.

So, summarizing, the see also should contain `Series.clip`, `DataFrame.clip`, and the `clip_lower` and `clip_upper` for each of them (I assume both classes have it). And assuming that `clip` is commonly used with `quantile` as in your example (I don't know), I would also add it.

@jreback do you agree?

Dpananos: Ah, yes I see, I was mistaken.  Sure, I can add all those.

pandas/core/generic.py
```diff
@@ -5601,53 +5601,95 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+            value(s) truncated.
+        DataFrame.clip_upper : Return copy of input with values above given
+            value(s) truncated.
+        DataFrame.quantile : Return values at the given quantile over requested
+            axis, a la numpy.percentile.
 
         Examples
         --------
+        >>> some_data = {'a': [-1, -2, -100], 'b': [1, 2, 100]}
+        >>> df = pd.DataFrame(some_data, index = ['foo', 'bar', 'foobar'])
```
datapythonista: PEP-8 issue with spaces around `=`.

Also, as you need to change this, I'd directly have the dictionary in the constructor, and avoid creating `some_data`. Just for the sake of consistency, as all examples I've seen use this way. With one key of the dictionary in one line, and the index in a third one, I think line width should be a problem.

pandas/core/generic.py
```diff
@@ -5601,53 +5601,95 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        foo 0   1
+        bar 1   2
+        foobar  10  100
+
+        `Winsorizing <https://en.wikipedia.org/wiki/Winsorizing>`__ is a
+        related method, whereby the data are clipped at
+        the 5th and 95th percentiles.
+
+        >>> lower, upper = df.quantile(0.05), df.quantile(0.95)
+        >>> df.clip(lower=lower, upper=upper, axis='columns')
```
datapythonista: Not sure if it's just because I'm too tired, but is `axis` useful here? If the threshold is a scalar, doesn't seem to have an effect, right?

datapythonista (commented):
Looks great, just a small PEP-8 issue, and I think it's perfect.

pandas/core/generic.py
Dpananos: Axis is required here since `df.quantile` returns a series with the quantiles for each column.  Running without the axis argument results in an error.

datapythonista: I'd find useful having this comment in the explanations before the test. :)

datapythonista (approved):
lgtm.

Really nice documentation, I find it very useful.

Dpananos (commented):
Edited

pandas/core/generic.py
```diff
@@ -5601,53 +5601,98 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+        `Series` or `DataFrame`.
+            Original input with those values above/below the
+            `upper`/`lower` thresholds set to the threshold values.
+
+        See Also
+        --------
+        DataFrame.clip : Trim values at input threshold(s).
+        Series.clip : Trim values at input threshold(s).
+        Series.clip_lower : Return copy of the input with values below given
+            value(s) truncated.
```
jreback: would just mention Series.clip here (not upper/lower)

pandas/core/generic.py
```diff
@@ -5601,53 +5601,98 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
-
-        >>> df.clip(t, t + 1, axis=0)
-                  0         1
-        0  0.335232 -0.300000
-        1 -0.200000  0.746646
-        2  0.027753 -0.100000
-        3  0.230930  0.000000
-        4  1.100000  0.570967
+            a   b
+        foo -1  1
```
jreback: alignment doesn’t look right here

pandas/core/generic.py
```diff
@@ -5601,53 +5601,98 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
...
+
+        >>> row_thresh = pd.Series({'foo': 0, 'bar': 1, 'foobar': 10})
+        >>> df.clip(lower=row_thresh, axis='index')
+            a   b
+        foo 0   1
+        bar 1   2
+        foobar  10  100
+
+        `Winsorizing <https://en.wikipedia.org/wiki/Winsorizing>`__ is a
+        related method, whereby the data are clipped at
```
jreback: put this reference in the Notes section

jreback (changes requested):

pandas/core/generic.py
datapythonista: Why don't you find `clip_upper` and `clip_lower` related? I suggested them, but actually I was wondering if I was missing something, or if it was being considered deprecating them. It seems to me that `clip(upper=X)` could be used for `clip_upper`.

jreback: going to deprecate lower/upper but that’s not the reason
when showing a DataFrame method only like to show 1 Series method and and any related DataFrame method 

just a preference to not have too many things in See Also

datapythonista: Makes sense to me.

But if I'm not wrong, the title is misleading but this docstring is for the Series.clip method too.

jreback: ahh ok in that case this is ok

pandas/core/generic.py
Dpananos: Do you want me to move the entire part about winsorization or just the link?

datapythonista: Just the link, you can see the format in the `13. References` section of this document: http://numpydoc.readthedocs.io/en/latest/format.html

pandas/core/generic.py
```diff
@@ -5629,6 +5629,12 @@ def clip(self, lower=None, upper=None, axis=None, inplace=False,
             Original input with those values above/below the
             `upper`/`lower` thresholds set to the threshold values.
 
+        Notes
```
datapythonista: I think it should be in `References` and not in `Notes`. It wasn't in the documentation for the sprint, for simplicity, but you can check this document: http://numpydoc.readthedocs.io/en/latest/format.html

Dpananos: Yea, I agree.  Done.

datapythonista (approved):
lgtm

jorisvandenbossche:
@Dpananos I merged the other PR that also edited the same docstring (https://github.com/pandas-dev/pandas/pull/20256). This will cause conflicts for this one, so you will need to resolve them (sorry for that trouble, but wanted to acknowledge both PRs, and yours adds some extra than the other).

If you need any help with updating the PR (merge upstream/master), let me know!

Dpananos:
@jorisvandenbossche  I'm not really certain what happened.  So you merged another docstring and now I have a conflict?  How should I go about updating the PR?

jorisvandenbossche:
So if you do:

```
git checkout docstring_clip
git fetch upstream
git merge upstream/master
```

you will have conflicts in the `generic.py` file. 

Solving these conflicts will mean merging somehow the two docstring (you can just use your judgement on which of the two versions you find best for each element of the docstring, certainly your examples were more elaborate I think). 
Once you solved the conflicts, you need to do:

```
git add pandas/core/generic.py
git commit
```

and push again to the branch on origin.

I would say: try it out, and if you have troubles with, I can also do it.

PR #20213, opened 2018-03-10
DOC: Updated docstring DatetimeIndex.ceil

vikrantkeshari:
completed the DatetimeIndex.ceil Docstring.

an unrelated PR):

```

```

explicitly.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/indexes/datetimelike.py
+++ b/pandas/core/indexes/datetimelike.py
@@ -122,20 +122,21 @@ class TimelikeOps(object):
     @Appender(_round_doc % "ceil")
     def ceil(self, freq):
         """
-        Ceil the index to the specified freq.
+        Floor the index to the specified freq.
 
-        Links to the available `frequency strings <http://pandas.
-        pydata.org/pandas-docs/stable/timeseries.html#offset-aliases>`_
-        and `frequency objects <http://pandas.pydata.org/pandas-docs/
-        stable/timeseries.html#dateoffset-objects>`_ .
+        Links to the available [frequency strings](http://pandas.
+        pydata.org/pandas-docs/stable/timeseries.html#offset-aliases)
+        and [frequency objects](http://pandas.pydata.org/pandas-docs/
+        stable/timeseries.html#dateoffset-objects).
 
         Parameters
         ----------
-        freq : freq string/object
+        freq : str, object
+            String or object specifying the frequency to round down to.
 
         Returns
         -------
-        ceil : DatetimeIndex
+        index of same type
 
         See Also
         --------
@@ -145,17 +146,16 @@ class TimelikeOps(object):
         Examples
         --------
 
-        In [2]: df = pd.DatetimeIndex(start='2014-08-01 09:23:41.321211',
+        >>> dti = pd.DatetimeIndex(start='2014-08-01 09:23:41.321211',
         freq = 'H', periods = 3)
 
-        In [3]: df
-
-        Out[3]: DatetimeIndex(['2014-08-01 09:23:41.321211',
+        >>> dti
+        DatetimeIndex(['2014-08-01 09:23:41.321211',
         '2014-08-01 10:23:41.321211', '2014-08-01 11:23:41.321211'],
         dtype='datetime64[ns]', freq='H')
 
-        In [4]: df.ceil('H')
-        Out[4]: DatetimeIndex(['2014-08-01 10:00:00', '2014-08-01 11:00:00',
+        >>> dti.ceil('H')
+        DatetimeIndex(['2014-08-01 10:00:00', '2014-08-01 11:00:00',
                '2014-08-01 12:00:00'],
               dtype='datetime64[ns]', freq=None)
 
--- a/pandas/core/indexes/datetimelike.py
+++ b/pandas/core/indexes/datetimelike.py
@@ -132,11 +132,12 @@ class TimelikeOps(object):
         Parameters
         ----------
         freq : str, object
-            String or object specifying the frequency to round down to.
+			String or object specifying the frequency to round down to.
 
         Returns
         -------
-        index of same type
+        DatetimeIndex
+            Index of the same time is rounded off.
 
         See Also
         --------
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,51 @@ def floor(self, freq):
 
     @Appender(_round_doc % "ceil")
     def ceil(self, freq):
+        """
+        Ceil the index to the specified freq.
+
+        Links to the available `frequency strings <http://pandas.
+        pydata.org/pandas-docs/stable/timeseries.html#offset-aliases>`_
```
datapythonista: sphinx creates the links for us

pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,51 @@ def floor(self, freq):
...
+        and `frequency objects <http://pandas.pydata.org/pandas-docs/
+        stable/timeseries.html#dateoffset-objects>`_ .
+
+        Parameters
+        ----------
+        freq : freq string/object
```
datapythonista: needs a description, type of parameters doesn't seem very standard.

pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,51 @@ def floor(self, freq):
...
+
+        Returns
+        -------
+        ceil : DatetimeIndex
```
datapythonista: only type if the function returns a single value and not a tuple

pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,51 @@ def floor(self, freq):
...
+
+        See Also
+        --------
+        DatetimeIndex.floor : Floor the DatetimeIndex to the specified freq.
+        DatetimeIndex.round : Round the DatetimeIndex to the specified freq.
+
+        Examples
+        --------
+
+        In [2]: df = pd.DatetimeIndex(start='2014-08-01 09:23:41.321211',
```
datapythonista: we use `>>>` and not ipython syntax.

datapythonista (commented):
Please, follow the points here. There are many mistakes that you can avoid by following the documentation.
https://python-sprints.github.io/pandas/dashboard.html

pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,45 @@ def floor(self, freq):
 
     @Appender(_round_doc % "ceil")
     def ceil(self, freq):
+        """
+        Ceil the index to the specified freq.
```
jreback: explain what ceil is

pandas/core/indexes/datetimelike.py
jreback: these can be string or pandas.DateOffset objects

pandas/core/indexes/datetimelike.py
jreback: call this dti instead of df

jreback (changes requested):

pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,45 @@ def floor(self, freq):
...
+        stable/timeseries.html#dateoffset-objects).
+
+        Parameters
+        ----------
+        freq : str, object
+            String or object specifying the frequency to round down to.
+
+        Returns
+        -------
+        index of same type
```
datapythonista: Is it a `DatetimeInidex`? Can you add it and a short description on what it returns?

pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,45 @@ def floor(self, freq):
...
+
+        Links to the available [frequency strings](http://pandas.
+        pydata.org/pandas-docs/stable/timeseries.html#offset-aliases)
+        and [frequency objects](http://pandas.pydata.org/pandas-docs/
+        stable/timeseries.html#dateoffset-objects).
+
+        Parameters
+        ----------
+        freq : str, object
+            String or object specifying the frequency to round down to.
```
datapythonista: Not sure what `object` means in this case. In the examples it seems that it can also be `None`. Could you check exactly what it accepts, and make it a bit more clear please? Didn't use this method myself, and with this description I'm still confused on what should I use for frequency.

pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,45 @@ def floor(self, freq):
...
+        See Also
+        --------
+        DatetimeIndex.floor : Floor the DatetimeIndex to the specified freq.
+        DatetimeIndex.round : Round the DatetimeIndex to the specified freq.
+
+        Examples
+        --------
+
+        >>> dti = pd.DatetimeIndex(start='2014-08-01 09:23:41.321211',
+        freq = 'H', periods = 3)
```
datapythonista: `idx` is more standard than `dti` in the docs. Also, the format should be:
```
>>> dti = pd.DatetimeIndex(start='2014-08-01 09:23:41.321211',
...                        freq='H', periods=3)
```
Note the `...` at the beginning of the second line, its alignment, and that I removed the spaces around `=`, which are not PEP-8 valid.

pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,45 @@ def floor(self, freq):
...
+
```
datapythonista: I think the html version looks better if you don't leave this blank line.

pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,45 @@ def floor(self, freq):
...
+        >>> dti
+        DatetimeIndex(['2014-08-01 09:23:41.321211',
+        '2014-08-01 10:23:41.321211', '2014-08-01 11:23:41.321211'],
+        dtype='datetime64[ns]', freq='H')
+
+        >>> dti.ceil('H')
+        DatetimeIndex(['2014-08-01 10:00:00', '2014-08-01 11:00:00',
+               '2014-08-01 12:00:00'],
+              dtype='datetime64[ns]', freq=None)
+
```
datapythonista: Unnecessary blank line.

pandas/core/indexes/datetimelike.py
```diff
@@ -121,6 +121,45 @@ def floor(self, freq):
 
     @Appender(_round_doc % "ceil")
     def ceil(self, freq):
+        """
+        Floor the index to the specified freq.
+
+        Links to the available [frequency strings](http://pandas.
+        pydata.org/pandas-docs/stable/timeseries.html#offset-aliases)
+        and [frequency objects](http://pandas.pydata.org/pandas-docs/
+        stable/timeseries.html#dateoffset-objects).
```
datapythonista: Can you use this long summary to explain what flooring the index means? The links look great, but if you check the source code of that pages, you should find a reference before them, that you can use here with a `:ref:`. Sphinx will build the links automatically with it.

Let me know if you need help with this.

datapythonista (commented):
Looks quite nice, added some comments.

pandas/core/indexes/datetimelike.py
vikrantkeshari: yes i need a help regarding  this :ref: @datapythonista

IHackPy (changes requested):
kindly test your code with flake8 then validate your docstring

pandas/core/indexes/datetimelike.py
datapythonista: At the beginning of this document [1] you can see this sentence: "See the :ref:\`cookbook<cookbook.missing_data>\` for some advanced strategies."

When the html documentation is generated with sphinx, this will look into the documentation for a bookmark `.. _cookbook.missing_data:` and create a link to it. That has some advantages. For example, it'll link to the same version of the documentation you're using, and not always the `stable` as your original implementation.

So, you should look into the source code of those links what's the name of the bookmark, and create them in the same way. If I'm not wrong, using the `--single` parameter when building the documentation something fails to create the links. It'll take longer, but build the whole documentation if that's the case.

datapythonista (changes requested):
@vikrantkeshari do you have time to fix the comments in this PR?

datapythonista:
Closing as stale.

PR #20216, opened 2018-03-10
DOC: update the pandas.DataFrame.all docstring

mlaforet:
an unrelated PR):

```
####################### Docstring (pandas.DataFrame.all) #######################

Return whether all elements are True over requested axis.

Returns True if all elements along a dataframe
axis are non-zero, not-empty or not-False.
Also note that a series consisting of different data
types returns the first occurence of the non-zero, not-empty
or not-False element.

Parameters
----------
axis : {index (0), columns (1)}
    will be NA.
    If the axis is a MultiIndex (hierarchical), count along a
    particular level, collapsing into a Series.
bool_only : boolean, default None
    Include only boolean columns. If None, will attempt to use everything,
    then use only boolean data. Not implemented for Series.
**kwargs : any, default None
    Additional keywords have no fect but might be accepted for
    compatibility with numpy.

Returns
-------
all : Series or DataFrame (if level specified)

Examples
--------
First create a pandas dataframe::
    >>> d = {'col1': [True, True], 'col2': [True, False]}
    >>> df = pd.DataFrame(data=d)
    >>> df
       col1   col2
    0  True   True
    1  True  False

Default behaviour checks if column-wise values all return True
    >>> df.all()
    col1     True
    col2    False
    dtype: bool

Adding axis=1 will check row-wise values
    >>> df.all(axis=1)
    0     True
    1    False
    dtype: bool

See also
--------
pandas.DataFrame.any : Checks if one (or more) items return True

Errors found:
	Errors in parameters section
		Parameter "axis" has no description

```

explicitly.
The axis description is a bug in the string interpollation of %(axis_descr)s
The **kwargs error is a bug in the parameter evaluation

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -7869,7 +7869,7 @@ _all_doc = """\
 Return whether all elements are True over series or dataframe axis.
 
 Returns True if all elements within a series or along a dataframe
-axis or are non-zero, not-empty or not-False."""
+axis are non-zero, not-empty or not-False."""
 
 _all_examples = """\
 Examples
@@ -7909,8 +7909,8 @@ dtype: bool
 _all_see_also = """\
 See also
 --------
-pandas.Series.all : Return if all elements are True
-pandas.DataFrame.any : Return if one (or more) elements are True
+pandas.Series.all : Return True if all elements are True
+pandas.DataFrame.any : Return True if one (or more) elements are True
 """
 
 _cnum_doc = """
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4295,7 +4295,7 @@ class NDFrame(PandasObject, SelectionMixin):
 
         e.g. If the dtypes are float16 and float32, dtype will be upcast to
         float32.  If dtypes are int32 and uint8, dtype will be upcast to
-        int32. By By :func:`numpy.find_common_type` convention, mixing int64
+        int32. By :func:`numpy.find_common_type` convention, mixing int64
         and uint64 will result in a float64 dtype.
 
         See Also
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -7802,7 +7802,7 @@ bool_only : boolean, default None
     Include only boolean columns. If None, will attempt to use everything,
     then use only boolean data. Not implemented for Series.
 **kwargs : any, default None
-    Additional keywords have no fect but might be accepted for
+    Additional keywords have no affect but might be accepted for
     compatibility with numpy.
 
 Returns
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -7576,11 +7523,10 @@ def _add_numeric_operations(cls):
         cls.any = _make_logical_function(
             cls, 'any', name, name2, axis_descr,
             'Return whether any element is True over requested axis',
-            nanops.nanany)
+            nanops.nanany, '', '')
```
TomAugspurger: Maybe remove these so they don't conflict with the other PR.

pandas/core/generic.py
```diff
@@ -7837,25 +7783,69 @@ def _doc_parms(cls):
...
+    will be NA.
 level : int or level name, default None
     If the axis is a MultiIndex (hierarchical), count along a
-    particular level, collapsing into a %(name1)s
+    particular level, collapsing into a %(name1)s.
 bool_only : boolean, default None
     Include only boolean columns. If None, will attempt to use everything,
     then use only boolean data. Not implemented for Series.
+**kwargs : any, default None
+    Additional keywords have no fect but might be accepted for
```
TomAugspurger: fect -> affect.

pandas/core/generic.py
```diff
@@ -7837,25 +7783,69 @@ def _doc_parms(cls):
...
+
+%(examples)s
+%(see_also)s"""
+
+_all_doc = """\
+Return whether all elements are True over requested axis.
+
+Returns True if all elements along a dataframe
+axis are non-zero, not-empty or not-False.
+Also note that a series consisting of different data
```
TomAugspurger: You can remove this second sentence. THe fact that this happens is a bug, so we don't want to document it.

pandas/core/generic.py
```diff
@@ -7837,25 +7783,69 @@ def _doc_parms(cls):
...
+types returns the first occurence of the non-zero, not-empty
+or not-False element."""
+
+_all_examples = """\
+Examples
+--------
+First create a pandas dataframe::
```
TomAugspurger: No dobule colon.

pandas/core/generic.py
```diff
@@ -7837,25 +7783,69 @@ def _doc_parms(cls):
...
+    >>> d = {'col1': [True, True], 'col2': [True, False]}
```
TomAugspurger: No indent.

TomAugspurger (commented):
Seems like some unrelated changes slipped in here. Could you

1. Fetch from upstream
2. merge upstream/master
3. Make sure that the only changes are contained to the docstring for `all`

Otherwise we'll get a bunch of merge conflicts later on.

jorisvandenbossche:
Be sure to also check the comments on the `DataFrame.any` PR https://github.com/pandas-dev/pandas/pull/20217 as it is a very related method, so both docstring are a bit consistent

pandas/core/generic.py
mlafore3: @TomAugspurger If I do this an error 
```python ../scripts/validate_docstrings.py pandas.DataFrame.all
  File "../scripts/validate_docstrings.py", line 37, in <module>
    import pandas
  File "/Users/marconlaforet/pandas-dev/pandas/__init__.py", line 42, in <module>
    from pandas.core.api import *
  File "/Users/marconlaforet/pandas-dev/pandas/core/api.py", line 10, in <module>
    from pandas.core.groupby import Grouper
  File "/Users/marconlaforet/pandas-dev/pandas/core/groupby.py", line 48, in <module>
  File "/Users/marconlaforet/pandas-dev/pandas/core/frame.py", line 75, in <module>
    from pandas.core.series import Series
  File "/Users/marconlaforet/pandas-dev/pandas/core/series.py", line 3202, in <module>
    Series._add_numeric_operations()
  File "/Users/marconlaforet/pandas-dev/pandas/core/generic.py", line 7530, in _add_numeric_operations
    nanops.nanany)
TypeError: _make_logical_function() missing 2 required positional arguments: 'examples' and 'see_also'```

Comment (jorisvandenbossche) [reply pandas/core/generic.py]: Yes, you can keep this. Just depending on which one of the two PRs is merged first, you might need to solve the conflict on the other. But that should be no problem

**Comment by mlafore3** [comment]
I have made the requested changes and added docs/examples for series.all(). This PR is ready for review again. Let me know if there is anything else.

File: pandas/core/generic.py [hunk 6]
```diff
@@ -7841,25 +7840,77 @@ def _doc_parms(cls):
...
+0     True
+1    False
+dtype: bool
+"""
+
+_all_see_also = """\
+See also
+--------
+pandas.DataFrame.any : Return if one (or more) elements are True
+"""
```
jreback: add Series.all

jreback (changes requested):

pandas/core/generic.py
mlafore3: @jreback its been added

jreback (approved):

jreback:
thanks!

PR #20219, opened 2018-03-10
DOC: update the DataFrame.iat[] docstring

akosel:
an unrelated PR):

```

####################### Docstring (pandas.DataFrame.iat) #######################

Selects a single value for a row/column pair by integer position.

Useful if performance is a major concern and you only need to
get or set a value at a particular row/column.

See Also
--------
at : Selects a single value for a row/column label pair
loc : Selects a group of rows and columns by label(s)
iloc : Selects group of rows and columns by integer position(s)

Returns
-------
    None

Examples
--------
>>> df = pd.DataFrame([[0,2,3], [0,4,1], [10,20,30]])
>>> df
    0   1   2
0   0   2   3
1   0   4   1
2  10  20  30
>>> df.iat[1, 2]
1
>>> df.iat[1, 2] = 10
>>> df.iat[1, 2]
10

Docstring for "pandas.DataFrame.iat" correct. :)

```

explicitly.

### DIFF
--- a/pandas/core/indexing.py
+++ b/pandas/core/indexing.py
@@ -1921,27 +1921,30 @@ class _AtIndexer(_ScalarAccessIndexer):
 
 class _iAtIndexer(_ScalarAccessIndexer):
     """
-    Selects a single value for a row/column pair by integer position.
+    Access a single value for a row/column pair by integer position.
 
-    Useful if performance is a major concern and you only need to
-    get or set a value at a particular row/column.
+    Similar to ``iloc``, in that both provided integer-based lookups. Use
+    ``iat`` if you only need to get or set a single value in a DataFrame.
 
     See Also
     --------
-    at : Selects a single value for a row/column label pair
-    loc : Selects a group of rows and columns by label(s)
-    iloc : Selects group of rows and columns by integer position(s)
+    DataFrame.at : Access a single value for a row/column label pair
+    DataFrame.loc : Access a group of rows and columns by label(s)
+    DataFrame.iloc : Access a group of rows and columns by integer position(s)
 
     Examples
     --------
-    >>> df = pd.DataFrame([[0, 2, 3], [0, 4, 1], [10, 20, 30]])
+    >>> df = pd.DataFrame([[0, 2, 3], [0, 4, 1], [10, 20, 30]],
+    ...      columns=['A', 'B', 'C'])
     >>> df
-        0   1   2
+        A   B   C
     0   0   2   3
     1   0   4   1
     2  10  20  30
+
     >>> df.iat[1, 2]
     1
+
     >>> df.iat[1, 2] = 10
     >>> df.iat[1, 2]
     10
--- a/pandas/core/indexing.py
+++ b/pandas/core/indexing.py
@@ -1924,7 +1924,8 @@ class _iAtIndexer(_ScalarAccessIndexer):
     Access a single value for a row/column pair by integer position.
 
     Similar to ``iloc``, in that both provide integer-based lookups. Use
-    ``iat`` if you only need to get or set a single value in a DataFrame.
+    ``iat`` if you only need to get or set a single value in a ``DataFrame``
+    or ``Series``.
 
     See Also
     --------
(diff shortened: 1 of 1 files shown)


pandas/core/indexing.py
```diff
@@ -1920,11 +1920,35 @@ def _convert_key(self, key, is_setter=False):
...
+    Useful if performance is a major concern and you only need to
+    get or set a value at a particular row/column.
 
+    See Also
+    --------
+    at : Selects a single value for a row/column label pair
+    loc : Selects a group of rows and columns by label(s)
+    iloc : Selects group of rows and columns by integer position(s)
+
+    Returns
```
TomAugspurger: You can omit the returns  section if there's no return value

pandas/core/indexing.py
```diff
@@ -1920,11 +1920,35 @@ def _convert_key(self, key, is_setter=False):
...
+    -------
+        None
+
+    Examples
+    --------
+    >>> df = pd.DataFrame([[0,2,3], [0,4,1], [10,20,30]])
```
TomAugspurger: pep8: spaces after comma.

pandas/core/indexing.py
```diff
@@ -1920,11 +1920,35 @@ def _convert_key(self, key, is_setter=False):
 
 
 class _iAtIndexer(_ScalarAccessIndexer):
-    """Fast integer location scalar accessor.
+    """
+    Selects a single value for a row/column pair by integer position.
```
jorisvandenbossche: Selects -> Select

pandas/core/indexing.py
```diff
@@ -1920,11 +1920,35 @@ def _convert_key(self, key, is_setter=False):
...
 
-    Similarly to ``iloc``, ``iat`` provides **integer** based lookups.
-    You can also set using these indexers.
+    Useful if performance is a major concern and you only need to
+    get or set a value at a particular row/column.
 
+    See Also
+    --------
+    at : Selects a single value for a row/column label pair
```
jorisvandenbossche: I think we will need to do `DataFrame.at` to make the automatic link work (and the same for the others below)

pandas/core/indexing.py
```diff
@@ -1920,11 +1920,35 @@ def _convert_key(self, key, is_setter=False):
 
 
 class _iAtIndexer(_ScalarAccessIndexer):
-    """Fast integer location scalar accessor.
+    """
+    Selects a single value for a row/column pair by integer position.
 
-    Similarly to ``iloc``, ``iat`` provides **integer** based lookups.
-    You can also set using these indexers.
+    Useful if performance is a major concern and you only need to
```
jorisvandenbossche: I would be a bit more careful with "major concern", as if performance is really important, you typically should not use this indexer many times. But, it is certainly useful to mention that `iat` is faster than `iloc` when you know you want to access only a single scalar.

pandas/core/indexing.py
jorisvandenbossche: BTW, I think "Select" is not the correct term, as you can also use this to set values

pandas/core/indexing.py
jorisvandenbossche: Also, can you specify string column names? eg `columns=['A', 'B', 'C']` 
to make it clear that it is position based (with the integer indexes it is dubious)

jorisvandenbossche (commented):
Thanks for the PR! Added some more comments

jorisvandenbossche:
Thanks for the updates!

pandas/core/indexing.py
```diff
@@ -1920,11 +1920,34 @@ def _convert_key(self, key, is_setter=False):
 
 
 class _iAtIndexer(_ScalarAccessIndexer):
-    """Fast integer location scalar accessor.
+    """
+    Access a single value for a row/column pair by integer position.
 
-    Similarly to ``iloc``, ``iat`` provides **integer** based lookups.
-    You can also set using these indexers.
+    Similar to ``iloc``, in that both provide integer-based lookups. Use
```
jreback: this also works on a Series, can you re-word to accomodate

jreback (changes requested):

pandas/core/indexing.py
akosel: Great point, I specify that `iat` works for both a `DataFrame` and a `Series`. I also added an example for using it on `Series`.

jorisvandenbossche (approved):

TomAugspurger:
Thanks @akosel !

PR #20220, opened 2018-03-10
DOC: update the Series.view docstring

villasv:

```
######################## Docstring (pandas.Series.view) ########################

Create a new view of the Series.

This function will return a new Series with a view of the same
underlying values in memory, optionally reinterpreted with a new data
type. The new data type must preserve the same size in bytes as to not
cause index misalignment.

Parameters
----------
dtype : data type
    Data type object or one of their string representations.

Returns
-------
Series
    A new Series object as a view of the same data in memory.

See Also
--------
numpy.ndarray.view : Equivalent numpy function to create a new view of
    the same data in memory.

Notes
-----
Series are instantiated with `dtype=float64` by default. While
`numpy.ndarray.view()` will return a view with the same data type as
the original array, Series.view() will try using `float64` and may fail
if the original data type size in bytes is not the same.

Examples
--------
>>> s = pd.Series([-2, -1, 0, 1, 2], dtype='int8')
>>> s
0   -2
1   -1
2    0
3    1
4    2
dtype: int8

The 8 bit signed integer representation of `-1` is `0b11111111`, but
the same bytes represent 255 if read as an 8 bit unsigned integer:

>>> us = s.view('uint8')
>>> us
0    254
1    255
2      0
3      1
4      2
dtype: uint8

The views share the same underlying values:

>>> us[0] = 128
>>> s
0   -128
1     -1
2      0
3      1
4      2
dtype: int8

Docstring for "pandas.Series.view" correct. :)
```

### DIFF
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -555,10 +555,6 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         type. The new data type must preserve the same size in bytes as to not
         cause index misalignment.
 
-        Series are instantiated with `dtype=float64` by default, so unlike
-        `numpy.array.view()` this function will not try to preserve the current
-        Series data type.
-
         Parameters
         ----------
         dtype : data type
@@ -571,7 +567,15 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         See Also
         --------
-        numpy.ndarray.view : Return a new view of the same data in memory.
+        numpy.ndarray.view : Equivalent numpy function to create a new view of
+            the same data in memory.
+
+        Notes
+        -----
+        Series are instantiated with `dtype=float64` by default. While
+        `numpy.ndarray.view()` will return a view with the same data type as
+        the original array, Series.view() will try using `float64` and may fail
+        if the original data type size in bytes is not the same.
 
         Examples
         --------
@@ -584,7 +588,7 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         4    2
         dtype: int8
 
-        The 8 bit signed integer representation of `-1` is `0b10000001`, but
+        The 8 bit signed integer representation of `-1` is `0b11111111`, but
         the same bytes represent 255 if read as an 8 bit unsigned integer:
 
         >>> us = s.view('uint8')
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -572,10 +572,11 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         Notes
         -----
-        Series are instantiated with `dtype=float64` by default. While
-        `numpy.ndarray.view()` will return a view with the same data type as
-        the original array, Series.view() will try using `float64` and may fail
-        if the original data type size in bytes is not the same.
+        Series are instantiated with ``dtype=float64`` by default. While
+        ``numpy.ndarray.view()`` will return a view with the same data type as
+        the original array, ``Series.view()`` (without specified dtype)
+        will try using ``float64`` and may fail if the original data type size
+        in bytes is not the same.
 
         Examples
         --------
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -547,6 +547,66 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         return len(self._data)
 
     def view(self, dtype=None):
+        """
+        Create a new view of the Series.
+
+        This function will return a new Series with a view of the same
+        underlying values in memory, optionally reinterpreted with a new data
+        type. The new data type must preserve the same size in bytes as to not
+        cause index misalignment.
+
+        Series are instantiated with `dtype=float64` by default, so unlike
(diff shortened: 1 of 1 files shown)


villasv:
As I noted in the documentation:

> Series are instantiated with `dtype=float64` by default, so unlike `numpy.array.view()` this function will not try to preserve the current Series data type.

This is probably unintended behavior. If I simply try to create a view:
```
>>> pd.Series([1,2,3], dtype='int8').view()
...
ValueError: When changing to a larger dtype, its size must be a divisor of the total size in bytes of the last axis of the array.
```

This happens because when `dtype=None` the constructor assumes `float64`.

pandas/core/series.py
```diff
@@ -547,6 +547,66 @@ def __len__(self):
...
+        Create a new view of the Series.
+
+        This function will return a new Series with a view of the same
+        underlying values in memory, optionally reinterpreted with a new data
+        type. The new data type must preserve the same size in bytes as to not
+        cause index misalignment.
+
+        Series are instantiated with `dtype=float64` by default, so unlike
+        `numpy.array.view()` this function will not try to preserve the current
+        Series data type.
```
jorisvandenbossche: This part I don't fully understand. Can you try to explain more what you exactly want to say, maybe with an example? (here in the comments)

pandas/core/series.py
```diff
@@ -547,6 +547,66 @@ def __len__(self):
...
+            Data type object or one of their string representations.
+
+        Returns
+        -------
+        Series
+            A new Series object as a view of the same data in memory.
+
+        See Also
+        --------
+        numpy.ndarray.view : Return a new view of the same data in memory.
```
jorisvandenbossche: I would say in the explanation here that this is the equivalent numpy method to take a view of an array

jorisvandenbossche (commented):
Nice docstring! Added some comments

pandas/core/series.py
villasv: It's basically what I said in the first comment of this PR. `pd.Series([1,2,3], dtype='int8').view()` will raise, but the equivalent in numpy would not.

pandas/core/series.py
villasv: But they are not *exactly* equivalent because numpy is a little bit more relaxed in this operation since it doesn't need to preserve the amount of data points. But you mean that perhaps I should write:

`numpy.ndarray.view : The equivalent function to create a view of an array.` ?

I think the current way is more formal and clear, *specially* because these functions don't behave exactly alike.

pandas/core/series.py
jorisvandenbossche: Ah, yes, thanks for the explanation. In principle, doing `view` without specifying a dtype is not that useful. So I would keep this explanation for the notes or so, and not include it here in the extended summary.

villasv: Notes looks like a good fit. On it.

pandas/core/series.py
jorisvandenbossche: > But you mean that perhaps I should write:
>
> numpy.ndarray.view : The equivalent function to create a view of an array. ?

Yes, that is what I meant (I would say "equivalent numpy function" though).  
I think a see also should also give a sense of why we mention the other function. And in this case, because it is the equivalent function of numpy (yes, they don't behave completely the same, but I would keep this for the notes)

villasv: Makes sense. I'll try to be a bit more explicit on the difference between them in the Notes section.

jorisvandenbossche (approved):

pandas/core/series.py
```diff
@@ -547,6 +547,71 @@ def __len__(self):
...
+        numpy.ndarray.view : Equivalent numpy function to create a new view of
+            the same data in memory.
+
+        Notes
+        -----
+        Series are instantiated with ``dtype=float64`` by default. While
+        ``numpy.ndarray.view()`` will return a view with the same data type as
+        the original array, ``Series.view()`` (without specified dtype)
+        will try using ``float64`` and may fail if the original data type size
+        in bytes is not the same.
```
jorisvandenbossche: I don't really like this note (I mean, the explanation is fine and correct, but we should just correct the code to not do this), but fine for now and I will open an issue to eg disallow not specifying a dtype, or if dtype=None to use `self.dtype`

jreback: see #20251 yes let's remove most of this text. This method should be deprecated anyhow.

jorisvandenbossche: The current text is correct, so let's remove *when* we deprecate it.

jorisvandenbossche:
@villasv Thanks a lot for the contribution!

Let's discuss the actual method itself further in https://github.com/pandas-dev/pandas/issues/20251 and  https://github.com/pandas-dev/pandas/issues/20281

pandas/core/series.py
villasv: I don't like this odd behavior either, but I don't think it's necessary to deprecate the function altogether. Wouldn't simply passing the current `dtype` (if the argument is `None`) to the constructor solve the oddity?

PR #20227, opened 2018-03-10
Docstring pandas.data frame.diff

adatasetaday:
an unrelated PR):

```

Docstring for "pandas.DataFrame.diff" correct. :)
```

explicitly.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4736,7 +4736,7 @@ class DataFrame(NDFrame):
         axis : {0 or 'index', 1 or 'columns'}, default 0
             Take difference over rows (0) or columns (1).
 
-            .. versionadded:: 0.16.1
+            .. versionadded:: 0.16.1.
 
         Returns
         -------
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4744,7 +4744,8 @@ class DataFrame(NDFrame):
 
         See Also
         --------
-        Series.diff: 1st discrete difference of object
+        DataFrame.pct_change: Percent change over given number of periods.
+        Series.diff: First discrete difference of object
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


TomAugspurger:
I think we just need a `.` at the end of this line https://github.com/pandas-dev/pandas/pull/20227/files#diff-1e79abbbdd150d4771b91ea60a4e1cc7R4739

LGTM otherwise.

adatasetaday:
I can definitely do that, but it won't be consistent with the other functions' documentation.

I'm also adding a "See Also" to DataFrame.pct_change() which I didn't realize was pretty much the same function using % instead of absolute value.

adatasetaday:
As dicsussed, the "." was added at the end of the versionadded and passed validation without any errors.

pandas/core/frame.py
```diff
@@ -4722,20 +4722,87 @@ def melt(self, id_vars=None, value_vars=None, var_name=None,
...
 
         Returns
         -------
         diffed : DataFrame
+
+        See Also
+        --------
+        DataFrame.pct_change: Percent change over given number of periods.
+        Series.diff: First discrete difference of object
+
```
jreback: add DataFrame.shift

pandas/core/frame.py
```diff
@@ -4722,20 +4722,87 @@ def melt(self, id_vars=None, value_vars=None, var_name=None,
...
+        ...                    'c': [1, 4, 9, 16, 25, 36]})
+        >>> df
+           a  b   c
+        0  1  1   1
+        1  2  1   4
+        2  3  2   9
+        3  4  3  16
+        4  5  5  25
+        5  6  8  36
+        >>> df.diff()
```
jreback: add a blank line between cases

jreback (changes requested):

adatasetaday (commented):
As requested added blank line between cases and added reference to DataFrame.shift in "See Also"

TomAugspurger:
Thanks @adatasetaday!

PR #20232, opened 2018-03-10
DOC: update the pd.Index.argsort docstring

dajcs:
an unrelated PR):

```
####################### Docstring (pandas.Index.argsort) #######################

Return the order of the indices.

We have an object with values and index. Using argsort method
returns a series of indices that would sort the index.

Parameters
----------
*args
    Passed to `numpy.ndarray.argsort`.
**kwargs
    Passed to `numpy.ndarray.argsort`.

Returns
-------
numpy.ndarray : Argsorted indices of the index

See also
--------
numpy.ndarray.argsort : Returns the indices that would sort this array.

Examples
--------
>>> s = pd.Series(data=[4,3,2,1],index=['c','b','a','d'])
>>> s.index.argsort()
array([2, 1, 0, 3], dtype=int64)

Errors found:
        Errors in parameters section
                Unknown parameters {'**kwargs', '*args'}
                Parameter "*args" has no type
```

explicitly.

We did our best to explain *args and **kwargs

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -2334,7 +2334,7 @@ class Index(IndexOpsMixin, PandasObject):
         See also
         --------
         numpy.argsort : Similar method for NumPy arrays.
-        Index.sort_values : Return sorted copy of Index
+        Index.sort_values : Return sorted copy of Index.
 
         Examples
         --------
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -2316,7 +2316,7 @@ class Index(IndexOpsMixin, PandasObject):
 
     def argsort(self, *args, **kwargs):
         """
-        Return the order of the indices that would sort the index.
+        Return the integer indicies that would sort the index.
 
         Parameters
         ----------
@@ -2328,7 +2328,7 @@ class Index(IndexOpsMixin, PandasObject):
         Returns
         -------
         numpy.ndarray
-            indicies that would sort the index if used as
+            Integer indicies that would sort the index if used as
             an indexer.
 
         See also
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -2328,24 +2328,26 @@ class Index(IndexOpsMixin, PandasObject):
         Returns
         -------
         numpy.ndarray
-            Argsorted indices of the index
+            indicies that would sort the index if used as
+            an indexer.
 
         See also
         --------
-        numpy.ndarray.argsort : Similar method for NumPy arrays.
-        pd.Index.sort_values : Return sorted copy of Index
+        numpy.argsort : Similar method for NumPy arrays.
+        Index.sort_values : Return sorted copy of Index
 
         Examples
         --------
-        >>> pd.Index(['b','a','d','c']).argsort()
-        array([1, 0, 3, 2], dtype=int64)
+        >>> idx = pd.Index(['b', 'a', 'd', 'c'])
+        >>> idx
+        Index(['b', 'a', 'd', 'c'], dtype='object')
 
-        When applying argsort to a Series object then the result won't
-        be affected by Series values only by Series index.
+        >>> order = idx.argsort()
+        >>> order
+        array([1, 0, 3, 2])
 
-        >>> s = pd.Series(data=[4, 3, 2, 1], index=['c', 'b', 'a', 'd'])
-        >>> s.index.argsort()
-        array([2, 1, 0, 3], dtype=int64)
+        >>> idx[order]
+        Index(['a', 'b', 'c', 'd'], dtype='object')
         """
         result = self.asi8
         if result is None:
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -2316,16 +2316,36 @@ class Index(IndexOpsMixin, PandasObject):
 
     def argsort(self, *args, **kwargs):
         """
-        Returns the indices that would sort the index and its
-        underlying data.
+        Return the order of the indices that would sort the index.
+
+        Parameters
+        ----------
+        *args
+            Passed to `numpy.ndarray.argsort`.
+        **kwargs
+            Passed to `numpy.ndarray.argsort`.
 
         Returns
         -------
-        argsorted : numpy array
+        numpy.ndarray
+            Argsorted indices of the index
 
         See also
         --------
-        numpy.ndarray.argsort
+        numpy.ndarray.argsort : Similar method for NumPy arrays.
+        pd.Index.sort_values : Return sorted copy of Index
+
+        Examples
+        --------
+        >>> pd.Index(['b','a','d','c']).argsort()
+        array([1, 0, 3, 2], dtype=int64)
+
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,31 @@ def shift(self, periods=1, freq=None):
 
     def argsort(self, *args, **kwargs):
         """
-        Returns the indices that would sort the index and its
-        underlying data.
+        Return the order of the indices.
```
TomAugspurger: Hmm I think the "would sort" is important. Does "Return the order of the indices that would sort the index" sound OK and fit on a line?

pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,31 @@ def shift(self, periods=1, freq=None):
...
+
+        We have an object with values and index. Using argsort method
```
TomAugspurger: Probably don't need this section.

pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,31 @@ def shift(self, periods=1, freq=None):
...
 
         Returns
         -------
-        argsorted : numpy array
+        numpy.ndarray : Argsorted indices of the index
 
         See also
         --------
-        numpy.ndarray.argsort
+        numpy.ndarray.argsort : Returns the indices that would sort this array.
```
TomAugspurger: Say "similar method for NumPy arrays."

pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,31 @@ def shift(self, periods=1, freq=None):
 
     def argsort(self, *args, **kwargs):
         """
-        Returns the indices that would sort the index and its
-        underlying data.
+        Return the order of the indices.
+
+        We have an object with values and index. Using argsort method
+        returns a series of indices that would sort the index.
```
jorisvandenbossche: series -> array ?

pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,31 @@ def shift(self, periods=1, freq=None):
...
+        numpy.ndarray : Argsorted indices of the index
 
         See also
         --------
-        numpy.ndarray.argsort
+        numpy.ndarray.argsort : Returns the indices that would sort this array.
+
+        Examples
+        --------
+        >>> s = pd.Series(data=[4,3,2,1],index=['c','b','a','d'])
```
TomAugspurger: PEP8: spacing after commas.

jorisvandenbossche: I think you can create the Index directly like `idx = pd.Index(['c', ..])` and then use that

jorisvandenbossche (commented):
Don't worry about the validation script erroring on the args/kwargs

pandas/core/indexes/base.py
dajcs: ok

pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,29 @@ def shift(self, periods=1, freq=None):
...
+            Argsorted indices of the index
 
         See also
         --------
-        numpy.ndarray.argsort
+        numpy.ndarray.argsort : Similar method for NumPy arrays.
+
+        Examples
+        --------
+        >>> s = pd.Series(data=[4,3,2,1], index=['c','b','a','d'])
```
jreback: can you put spaces between the commas

pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,29 @@ def shift(self, periods=1, freq=None):
 
     def argsort(self, *args, **kwargs):
         """
-        Returns the indices that would sort the index and its
-        underlying data.
+        Return the order of the indices that would sort the index.
+
+        Parameters
+        ----------
+        *args
```
jreback: @jorisvandenbossche this is the correct format for args/kwargs?

pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,29 @@ def shift(self, periods=1, freq=None):
...
         Returns
         -------
-        argsorted : numpy array
+        numpy.ndarray
+            Argsorted indices of the index
 
         See also
         --------
-        numpy.ndarray.argsort
+        numpy.ndarray.argsort : Similar method for NumPy arrays.
```
jreback: see also Index.sort_values

jreback (changes requested):

pandas/core/indexes/base.py
dajcs: ok, I will make two examples, one with series and one with index.

pandas/core/indexes/base.py
dajcs: fixed

pandas/core/indexes/base.py
jorisvandenbossche: Yes, I was only thinking it might actually make more sense to list the ones that are useful. I think it is only `kind` ? (`axis` certainly not, and `order` not as well?)

pandas/core/indexes/base.py
jorisvandenbossche: @TomAugspurger I think "order of the indices" is a bit strange no? I think it would be either "Return the order of the index" or "Return the indices that would sort the index", but not the combination

pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,36 @@ def shift(self, periods=1, freq=None):
 
     def argsort(self, *args, **kwargs):
         """
-        Returns the indices that would sort the index and its
-        underlying data.
+        Return the order of the indices that would sort the index.
```
jorisvandenbossche: See https://github.com/pandas-dev/pandas/pull/20232#discussion_r173637958

pandas/core/indexes/base.py
jreback: yes we do this for Series.argsort now. so should replicate here. or shared these doc-strings (best)

pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,36 @@ def shift(self, periods=1, freq=None):
...
         -------
-        argsorted : numpy array
+        numpy.ndarray
+            Argsorted indices of the index
 
         See also
         --------
-        numpy.ndarray.argsort
+        numpy.ndarray.argsort : Similar method for NumPy arrays.
+        pd.Index.sort_values : Return sorted copy of Index
```
jreback: don't need the pd.

pandas/core/indexes/base.py
```diff
@@ -2316,16 +2316,36 @@ def shift(self, periods=1, freq=None):
...
+
+        Examples
+        --------
+        >>> pd.Index(['b','a','d','c']).argsort()
+        array([1, 0, 3, 2], dtype=int64)
+
+        When applying argsort to a Series object then the result won't
```
jreback: I don't think from here down is relevant. Series has its own example.

jreback (changes requested):

TomAugspurger:
Thanks @dajcs!

PR #20233, opened 2018-03-10
DOC: update the Rolling.var docstring

mmaybeno:
an unrelated PR):

```
################## Docstring (pandas.core.window.Rolling.var) ##################

Calculate unbiased rolling variance.

Normalized by N-1 by default. This can be changed using the ddof argument.

Parameters
----------
ddof : int, default 1
    Delta Degrees of Freedom.  The divisor used in calculations
    is ``N - ddof``, where ``N`` represents the number of elements.
args
    Under Review.
kwargs
    Under Review.

Returns
-------
Series or DataFrame
    Returned object type is determined by the caller of the rolling
    calculation

See Also
--------
Series.var : Equivalent method for Series
DataFrame.var : Equivalent method for DataFrame
numpy.var : Equivalent method for Numpy array

Notes
-----
A minimum of 1 periods is required for the rolling calculation.

Examples
--------
The below example will show a rolling example

>>> s = pd.Series((5,5,5,5,6,7,9,10,5,5,5,5))
>>> s.rolling(3).var(1)
0          NaN
1          NaN
2     0.000000
3     0.000000
4     0.333333
5     1.000000
6     2.333333
7     2.333333
8     7.000000
9     8.333333
10    0.000000
11    0.000000
dtype: float64

Errors found:
        Errors in parameters section
                Parameter "args" has no type


################# Docstring (pandas.core.window.Expanding.var) #################

Calculate unbiased expanding variance.

Normalized by N-1 by default. This can be changed using the ddof argument.

Parameters
----------
ddof : int, default 1
    Delta Degrees of Freedom.  The divisor used in calculations
    is ``N - ddof``, where ``N`` represents the number of elements.
args
    Under Review.
kwargs
    Under Review.

Returns
-------
Series or DataFrame
    Returned object type is determined by the caller of the expanding
    calculation

See Also
--------
Series.expanding : Calling object with Series data
DataFrame.expanding : Calling object with DataFrames
Series.var : Equivalent method for Series
DataFrame.var : Equivalent method for DataFrame
numpy.var : Equivalent method for Numpy array

Notes
-----
A minimum of 1 periods is required for the rolling calculation.

Examples
--------
The below example will show a rolling example

>>> s = pd.Series((5,5,5,5,6,7,9,10,5,5,5,5))
>>> s.rolling(3).var(1)
0          NaN
1          NaN
2     0.000000
3     0.000000
4     0.333333
5     1.000000
6     2.333333
7     2.333333
8     7.000000
9     8.333333
10    0.000000
11    0.000000
dtype: float64

Errors found:
        Errors in parameters section
                Parameter "args" has no type
```

explicitly.

### DIFF
--- a/pandas/core/window.py
+++ b/pandas/core/window.py
@@ -879,13 +879,58 @@ class _Rolling_and_Expanding(_Rolling):
                            ddof=ddof, **kwargs)
 
     _shared_docs['var'] = dedent("""
-    %(name)s variance
-
+    Calculate unbiased %(name)s variance.
+    
+    Normalized by N-1 by default. This can be changed using the ddof argument.
+    
     Parameters
     ----------
     ddof : int, default 1
         Delta Degrees of Freedom.  The divisor used in calculations
-        is ``N - ddof``, where ``N`` represents the number of elements.""")
+        is ``N - ddof``, where ``N`` represents the number of elements.
+    args
+        Under Review.
+    kwargs
+        Under Review.
+      
+    Returns
+    -------
+    Series or DataFrame
+        Returned object type is determined by the caller of the %(name)s
+        calculation
+        
+    See Also
+    --------
+    Series.%(name)s : Calling object with Series data
+    DataFrame.%(name)s : Calling object with DataFrames
+    Series.var : Equivalent method for Series
+    DataFrame.var : Equivalent method for DataFrame
+    numpy.var : Equivalent method for Numpy array
+    
+    Notes
+    -----
+    A minimum of 1 periods is required for the rolling calculation.
+      
+    Examples
+    --------
+    The below example will show a rolling example 
+    
+    >>> s = pd.Series((5,5,5,5,6,7,9,10,5,5,5,5))
+    >>> s.rolling(3).var(1)
+    0          NaN
+    1          NaN
+    2     0.000000
+    3     0.000000
+    4     0.333333
+    5     1.000000
+    6     2.333333
+    7     2.333333
+    8     7.000000
+    9     8.333333
+    10    0.000000
+    11    0.000000
+    dtype: float64
+    """)
 
     def var(self, ddof=1, *args, **kwargs):
         nv.validate_window_func('var', args, kwargs)
@@ -1257,7 +1302,6 @@ class Rolling(_Rolling_and_Expanding):
         return super(Rolling, self).std(ddof=ddof, **kwargs)
 
     @Substitution(name='rolling')
-    @Appender(_doc_template)
     @Appender(_shared_docs['var'])
     def var(self, ddof=1, *args, **kwargs):
         nv.validate_rolling_func('var', args, kwargs)
@@ -1496,7 +1540,6 @@ class Expanding(_Rolling_and_Expanding):
         return super(Expanding, self).std(ddof=ddof, **kwargs)
 
     @Substitution(name='expanding')
-    @Appender(_doc_template)
     @Appender(_shared_docs['var'])
     def var(self, ddof=1, *args, **kwargs):
         nv.validate_expanding_func('var', args, kwargs)
--- a/pandas/core/window.py
+++ b/pandas/core/window.py
@@ -881,23 +881,23 @@ class _Rolling_and_Expanding(_Rolling):
     _shared_docs['var'] = dedent("""
     Calculate unbiased %(name)s variance.
 
-    Normalized by N-1 by default. This can be changed using the ddof argument.
+    Normalized by N-1 by default. This can be changed using the `ddof`
+    argument.
 
     Parameters
     ----------
     ddof : int, default 1
         Delta Degrees of Freedom.  The divisor used in calculations
         is ``N - ddof``, where ``N`` represents the number of elements.
-    args
+    *args
         Under Review.
-    kwargs
+    **kwargs
         Under Review.
 
     Returns
     -------
-    Series or DataFrame
-        Returned object type is determined by the caller of the %(name)s
-        calculation
+    Returns the same object type as determined by the caller of the %(name)s
+    calculation.
 
     See Also
     --------
@@ -913,9 +913,7 @@ class _Rolling_and_Expanding(_Rolling):
 
     Examples
     --------
-    The below example will show a rolling example
-
-    >>> s = pd.Series((5,5,5,5,6,7,9,10,5,5,5,5))
+    >>> s = pd.Series((5, 5, 5, 5, 6, 7, 9, 10, 5, 5, 5, 5))
     >>> s.rolling(3).var(1)
     0          NaN
     1          NaN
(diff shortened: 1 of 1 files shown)


pandas/core/window.py
```diff
@@ -879,13 +879,58 @@ def f(arg, *args, **kwargs):
                            ddof=ddof, **kwargs)
 
     _shared_docs['var'] = dedent("""
-    %(name)s variance
+    Calculate unbiased %(name)s variance.
+
+    Normalized by N-1 by default. This can be changed using the ddof argument.
```
TomAugspurger: single backtics around `ddof` since it's a parameter.

datapythonista: Can you add backticks around `ddof` please?

pandas/core/window.py
```diff
@@ -879,13 +879,58 @@ def f(arg, *args, **kwargs):
...
+    4     0.333333
+    5     1.000000
+    6     2.333333
+    7     2.333333
+    8     7.000000
+    9     8.333333
+    10    0.000000
+    11    0.000000
+    dtype: float64
+    """)
```
TomAugspurger: Can you show `s.expanding(3).var(1)` as well?

pandas/core/window.py
```diff
@@ -879,13 +879,58 @@ def f(arg, *args, **kwargs):
...
+
+    Notes
+    -----
+    A minimum of 1 periods is required for the rolling calculation.
+
+    Examples
+    --------
+    The below example will show a rolling example
+
+    >>> s = pd.Series((5,5,5,5,6,7,9,10,5,5,5,5))
```
datapythonista: Can you add spaces after the commas, so this passes PEP-8

pandas/core/window.py
```diff
@@ -879,13 +879,58 @@ def f(arg, *args, **kwargs):
...
+    DataFrame.var : Equivalent method for DataFrame
+    numpy.var : Equivalent method for Numpy array
+
+    Notes
+    -----
+    A minimum of 1 periods is required for the rolling calculation.
+
+    Examples
+    --------
+    The below example will show a rolling example
```
datapythonista: I'd take this sentence out, I think users will assume that.

pandas/core/window.py
```diff
@@ -879,13 +879,58 @@ def f(arg, *args, **kwargs):
...
+
+    Normalized by N-1 by default. This can be changed using the ddof argument.
 
     Parameters
     ----------
     ddof : int, default 1
         Delta Degrees of Freedom.  The divisor used in calculations
-        is ``N - ddof``, where ``N`` represents the number of elements.""")
+        is ``N - ddof``, where ``N`` represents the number of elements.
+    args
```
TomAugspurger: I don't think the underlying function takes any additional arguments. Can remove (I think).

mmaybeno:
Also updates docs for expanding window

pandas/core/window.py
```diff
@@ -879,13 +879,58 @@ def f(arg, *args, **kwargs):
...
+        Under Review.
+    kwargs
+        Under Review.
+
+    Returns
+    -------
+    Series or DataFrame
+        Returned object type is determined by the caller of the %(name)s
```
TomAugspurger: It's always the same type as the original object, right? Series.rolling().var is always a Series? Let's say that.

pandas/core/window.py
```diff
@@ -879,13 +879,58 @@ def f(arg, *args, **kwargs):
...
     Parameters
     ----------
     ddof : int, default 1
         Delta Degrees of Freedom.  The divisor used in calculations
-        is ``N - ddof``, where ``N`` represents the number of elements.""")
+        is ``N - ddof``, where ``N`` represents the number of elements.
+    args
+        Under Review.
+    kwargs
+        Under Review.
```
datapythonista: `*args` and `**kwargs` should have the stars. We changed the convention last minute, and the validation script is wrong, ignore the messages.

datapythonista (commented):
Nice one, added some comments.

pandas/core/window.py
mmaybeno: I can remove it, it looks like it passes them through, but I didn't dive into it's usage.. I'll remove for now.

pandas/core/window.py
jorisvandenbossche: But DataFrame.rolling().var() is a DataFrame, so that is what the above sentences tries to say. What do you want to change? 

"is determined by " -> "same type as" ?

pandas/core/window.py
```diff
@@ -879,13 +879,70 @@ def f(arg, *args, **kwargs):
...
-        is ``N - ddof``, where ``N`` represents the number of elements.""")
+        is ``N - ddof``, where ``N`` represents the number of elements.
+    *args
+        Under Review.
+    **kwargs
+        Under Review.
+
+    Returns
+    -------
+    Returns the same object type as determined by the caller of the %(name)s
```
jorisvandenbossche: I would keep the `Series or DataFrame` on the line above (as you had it before), that makes it clearer it is one of those.

If you use "same object type", then you can omit "as determined by" I think

pandas/core/window.py
```diff
@@ -879,13 +879,70 @@ def f(arg, *args, **kwargs):
...
+    DataFrame.var : Equivalent method for DataFrame
+    numpy.var : Equivalent method for Numpy array
+
+    Notes
+    -----
+    A minimum of 1 periods is required for the rolling calculation.
+
+    Examples
+    --------
+    >>> s = pd.Series((5, 5, 5, 5, 6, 7, 9, 10, 5, 5, 5, 5))
```
jorisvandenbossche: See comment on other PR, can you use a list here instead of tuple? 
I also think it can maybe a bit shorter? Or is there something specific you want to illustrate with this exact sequence of numbers ?

pandas/core/window.py
```diff
@@ -879,13 +879,70 @@ def f(arg, *args, **kwargs):
...
+    See Also
+    --------
+    Series.%(name)s : Calling object with Series data
+    DataFrame.%(name)s : Calling object with DataFrames
+    Series.var : Equivalent method for Series
+    DataFrame.var : Equivalent method for DataFrame
+    numpy.var : Equivalent method for Numpy array
+
+    Notes
+    -----
```
jreback: so in both this one and the std one, should put in the Notes that the ddof default is *different* than numpy. (we use 1, they use 0). This is in the Series.var/std strings already.

jreback (changes requested):

pandas/core/window.py
mmaybeno: I can do that, no specific reason why I used such a long tuple.

pandas/core/window.py
mmaybeno: 👍

pandas/core/window.py
```diff
@@ -879,13 +879,64 @@ def f(arg, *args, **kwargs):
...
+    argument.
 
     Parameters
     ----------
     ddof : int, default 1
         Delta Degrees of Freedom.  The divisor used in calculations
-        is ``N - ddof``, where ``N`` represents the number of elements.""")
+        is ``N - ddof``, where ``N`` represents the number of elements.
+    *args
+        Under Review.
```
jreback: @jorisvandenbossche ?

mmaybeno: It was mentioned that the exact content for what should be documented for args/kwargs was to be determined. If you'd rather me add the documentation I can do that too.

TomAugspurger: I think we're noting that they're accepted, but not used.

TomAugspurger (approved):

TomAugspurger:
Thanks again @mmaybeno.

PR #20234, opened 2018-03-10
DOC: update the PeriodIndex docstring

costrouc:
year, month, quarter, day, hour, minute, second documentation
improved. Detailed in depth the requirements of each parameter.

Documentation is still confusing for allowable freq arguments

an unrelated PR):

```

Errors found:
	Errors in parameters section
		Parameters {'fields', 'name', 'ordinal'} not documented
		Unknown parameters {'second', 'month', 'year', 'minute', 'quarter', 'day', 'hour'}
		Parameter "dtype" has no description
	No returns section found
```

A few errors are still reported but are outside of the scope of the doc sprint (fields, name, ordinal).

### DIFF
--- a/pandas/core/indexes/period.py
+++ b/pandas/core/indexes/period.py
@@ -148,12 +148,26 @@ class PeriodIndex(DatelikeOps, DatetimeIndexOpsMixin, Int64Index):
         If periods is none, generated index will extend to first conforming
         period on or just past end argument
     year : int, array, or Series, default None
+        If not None requires a freq to be specified and either
+        month or quarter to be specified
     month : int, array, or Series, default None
+        If not None requires a freq and year to be specified
     quarter : int, array, or Series, default None
+        If quarter is not None the freq is required to be
+        quarterly Q or Q-DEC. it will be set to Q-DEC if not specified.
+        day, hour, minute, second are not used
     day : int, array, or Series, default None
+        If not None requires a freq, month, and year to be specified.
+        day defaults to 1 if year and month specified
     hour : int, array, or Series, default None
+        If not None requires a freq, month, and year to be specified.
+        hour will default to 0 if year and month specified
     minute : int, array, or Series, default None
+        If not None requires a freq, month, and year to be specified.
+        minute will default to 0 if year and month specified
     second : int, array, or Series, default None
+        If not None requires a freq, month, and year to be specified.
+        second will default to 0 if year and month specified
     tz : object, default None
         Timezone for converting datetime64 data to Periods
     dtype : str or PeriodDtype, default None
@@ -194,6 +208,8 @@ class PeriodIndex(DatelikeOps, DatetimeIndexOpsMixin, Int64Index):
     >>> idx = PeriodIndex(year=year_arr, quarter=q_arr)
 
     >>> idx2 = PeriodIndex(start='2000', end='2010', freq='A')
+
+    >>> idx3 = PeriodIndex(year=[2000, 2001], month=[1, 2], second=[1, 2], freq='S')
 
     See Also
     ---------
--- a/pandas/core/indexes/period.py
+++ b/pandas/core/indexes/period.py
@@ -133,43 +133,43 @@ class PeriodIndex(DatelikeOps, DatetimeIndexOpsMixin, Int64Index):
     Parameters
     ----------
     data : array-like (1-dimensional), optional
-        Optional period-like data to construct index with
+        Optional period-like data to construct index with.
     copy : bool
-        Make a copy of input ndarray
+        Make a copy of input ndarray.
     freq : string or period object, optional
-        One of pandas period strings or corresponding objects
+        One of pandas period strings or corresponding objects.
     start : starting value, period-like, optional
         If data is None, used as the start point in generating regular
         period data.
     periods : int, optional, > 0
         Number of periods to generate, if generating index. Takes precedence
-        over end argument
+        over end argument.
     end : end value, period-like, optional
         If periods is none, generated index will extend to first conforming
-        period on or just past end argument
+        period on or just past end argument.
     year : int, array, or Series, default None
         If not None requires a freq to be specified and either
-        month or quarter to be specified
+        month or quarter to be specified.
     month : int, array, or Series, default None
-        If not None requires a freq and year to be specified
+        If not None requires a freq and year to be specified.
     quarter : int, array, or Series, default None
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/period.py
```diff
@@ -191,9 +205,11 @@ class PeriodIndex(DatelikeOps, DatetimeIndexOpsMixin, Int64Index):
 
     Examples
     --------
-    >>> idx = PeriodIndex(year=year_arr, quarter=q_arr)
+    >>> idx = pd.PeriodIndex(year=[2000, 2001, 2002], quarter=[1, 2, 1])
```
jorisvandenbossche: Can you show the output of each of those examples?

pandas/core/indexes/period.py
```diff
@@ -133,29 +133,43 @@ class PeriodIndex(DatelikeOps, DatetimeIndexOpsMixin, Int64Index):
...
         period data.
     periods : int, optional, > 0
         Number of periods to generate, if generating index. Takes precedence
-        over end argument
+        over end argument.
     end : end value, period-like, optional
         If periods is none, generated index will extend to first conforming
-        period on or just past end argument
+        period on or just past end argument.
     year : int, array, or Series, default None
```
jorisvandenbossche: general note: can you change `default None` in `optional` where relevant? (I think it are most of the keywords)

jorisvandenbossche (commented):
Thanks for the PR!

Can you try to make a summary line in the beginning of the docstring (one that fits on one line)? 
Also, some of the paramters that are not documented as indicated by the validation script (eg `name`), you could add those to the Parameter list

pandas/core/indexes/period.py
```diff
@@ -191,9 +205,11 @@ class PeriodIndex(DatelikeOps, DatetimeIndexOpsMixin, Int64Index):
...
-    >>> idx = PeriodIndex(year=year_arr, quarter=q_arr)
+    >>> idx = pd.PeriodIndex(year=[2000, 2001, 2002], quarter=[1, 2, 1])
 
-    >>> idx2 = PeriodIndex(start='2000', end='2010', freq='A')
+    >>> idx2 = pd.PeriodIndex(start='2000', end='2010', freq='A')
+
+    >>> idx3 = pd.PeriodIndex(year=[2000, 2001], month=[1, 2], second=[1, 2], freq='S')
 
     See Also
     ---------
```
jschendel: Can you add `Interval` and `IntervalIndex` to the See Also section? Seems relevant as they can be used for more generic intervals of time in that they don't require a specific frequency.

pandas/core/indexes/period.py
```diff
@@ -133,29 +133,43 @@ class PeriodIndex(DatelikeOps, DatetimeIndexOpsMixin, Int64Index):
...
+        If not None requires a freq and year to be specified.
     quarter : int, array, or Series, default None
+        If quarter is not None the freq is required to be
+        quarterly Q or Q-DEC. it will be set to Q-DEC if not specified.
+        day, hour, minute, second are not used.
     day : int, array, or Series, default None
+        If not None requires a freq, month, and year to be specified.
+        day defaults to 1 if year and month specified.
     hour : int, array, or Series, default None
+        If not None requires a freq, month, and year to be specified.
```
jreback: these also require a day from hour on down

jreback (changes requested):

pandas/core/indexes/period.py
```diff
@@ -133,29 +133,43 @@ class PeriodIndex(DatelikeOps, DatetimeIndexOpsMixin, Int64Index):
...
     ----------
     data : array-like (1-dimensional), optional
-        Optional period-like data to construct index with
+        Optional period-like data to construct index with.
     copy : bool
-        Make a copy of input ndarray
+        Make a copy of input ndarray.
     freq : string or period object, optional
-        One of pandas period strings or corresponding objects
+        One of pandas period strings or corresponding objects.
```
jreback: I think we have a reference somewhere to the offsets page:http://pandas.pydata.org/pandas-docs/stable/timeseries.html#offset-aliases, can you add

TomAugspurger: :ref:`offset aliases <timeseries.offset_aliases>`

pandas/core/indexes/period.py
costrouc: No a day is not required from hour since day will take a default value of 1. This is why I specified that a different default value is taken if year and month are specified.

pandas/core/indexes/period.py
```diff
@@ -133,29 +133,43 @@ class PeriodIndex(DatelikeOps, DatetimeIndexOpsMixin, Int64Index):
     Parameters
     ----------
     data : array-like (1-dimensional), optional
-        Optional period-like data to construct index with
+        Optional period-like data to construct index with.
     copy : bool
```
jorisvandenbossche: add ', default False'

pandas/core/indexes/period.py
costrouc: So I was looking at that but some of the strings do not work. For instance `BQS` is not a valid frequenct. Also the offset alias does not mention anything about strides which frequency also supports. Is this documented anywhere?

costrouc:
Ok so I am sort of lost with all the feedback. 

I hear the requests for changes but many of these parameters require multiple sentences to describe. The `PeriodIndex` class has three ways of initializing each one requiring a subset of the arguments. The `_generate_range` and `_range_from_fields` methods of the class show this. Here are the three ways:

 1. `PeriodIndex(start, end, periods, freq)`
 2. `PeriodIndex(year, quarter, freq)`
 3. `PeriodIndex(year, month, day, hour, minute, second, freq)`

Case 2 requires that the freq must match Q or Q-DEC.

Additionally while year, month, day, hour, minute and second all do default to None for cases 1 and 2. Case 3 has a completely different default behavior (day=1, hour=0, minute=0, second=0) while year and month are required.

pandas/core/indexes/period.py
TomAugspurger: Periods are time *spans* so things like month-start vs. month-end don't make sense. It's the etnire month.

I'm not sure if that's explained anywhere in http://pandas.pydata.org/pandas-docs/stable/timeseries.html. If not it'd be good to add.

What do you mean by strides?

TomAugspurger:
Thanks for listing out the cases. I think as long as each of those has at least one example we're OK.

jreback:

PR #20235, opened 2018-03-10
DOC: update Rolling.std docstring

mmaybeno:
an unrelated PR):

```
################## Docstring (pandas.core.window.Rolling.std) ##################

Calculate rolling standard deviation.

Normalized by N-1 by default. This can be changed using the ddof argument.

Parameters
----------
ddof : int, default 1
    Delta Degrees of Freedom.  The divisor used in calculations
    is ``N - ddof``, where ``N`` represents the number of elements.
args
    Under Review.
kwargs
    Under Review.

Returns
-------
Series or DataFrame
    Returned object type is determined by the caller of the rolling
    calculation

See Also
--------
Series.std : Equivalent method for Series
DataFrame.std : Equivalent method for DataFrame
numpy.std : Equivalent method for Numpy array

Notes
-----
A minimum of 1 periods is required for the rolling calculation.

Examples
--------
The below example will show a rolling example

>>> s = pd.Series((5,5,5,5,6,7,9,10,5,5,5,5))
>>> s.rolling(3).std(1)
0          NaN
1          NaN
2     0.000000
3     0.000000
4     0.577350
5     1.000000
6     1.527525
7     1.527525
8     2.645751
9     2.886751
10    0.000000
11    0.000000
dtype: float64

Errors found:
        Errors in parameters section
                Parameter "args" has no type

################# Docstring (pandas.core.window.Expanding.std) #################

Calculate expanding standard deviation.

Normalized by N-1 by default. This can be changed using the ddof argument.

Parameters
----------
ddof : int, default 1
    Delta Degrees of Freedom.  The divisor used in calculations
    is ``N - ddof``, where ``N`` represents the number of elements.
args
    Under Review.
kwargs
    Under Review.

Returns
-------
Series or DataFrame
    Returned object type is determined by the caller of the expanding
    calculation

See Also
--------
Series.expanding : Calling object with Series data
DataFrame.expanding : Calling object with DataFrames
Series.std : Equivalent method for Series
DataFrame.std : Equivalent method for DataFrame
numpy.std : Equivalent method for Numpy array

Notes
-----
A minimum of 1 periods is required for the rolling calculation.

Examples
--------
The below example will show a rolling example

>>> s = pd.Series((5,5,5,5,6,7,9,10,5,5,5,5))
>>> s.rolling(3).std(1)
0          NaN
1          NaN
2     0.000000
3     0.000000
4     0.577350
5     1.000000
6     1.527525
7     1.527525
8     2.645751
9     2.886751
10    0.000000
11    0.000000
dtype: float64

Errors found:
        Errors in parameters section
                Parameter "args" has no type
```

explicitly.

### DIFF
--- a/pandas/core/window.py
+++ b/pandas/core/window.py
@@ -885,10 +885,10 @@ class _Rolling_and_Expanding(_Rolling):
 
     Notes
     -----
-    The default `ddof` of 1 used in Series.var is different than the default
-    `ddof` of 0 in numpy.var.
+    The default `ddof` of 1 used in Series.std is different than the default
+    `ddof` of 0 in numpy.std.
 
-    A minimum of 1 periods is required for the rolling calculation.
+    A minimum of one period is required for the rolling calculation.
 
     Examples
     --------
--- a/pandas/core/window.py
+++ b/pandas/core/window.py
@@ -867,15 +867,13 @@ class _Rolling_and_Expanding(_Rolling):
     ddof : int, default 1
         Delta Degrees of Freedom.  The divisor used in calculations
         is ``N - ddof``, where ``N`` represents the number of elements.
-    *args
-        Under Review.
-    **kwargs
-        Under Review.
+    *args, **kwargs
+        For NumPy compatibility. No additional arguments are used.
 
     Returns
     -------
-    Returns the same object type as determined by the caller of the %(name)s
-    calculation.
+    Series or DataFrame
+        Returns the same object type as the caller of the %(name)s calculation.
 
     See Also
     --------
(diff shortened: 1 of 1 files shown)


pandas/core/window.py
```diff
@@ -857,13 +857,58 @@ def median(self, **kwargs):
         return self._apply('roll_median_c', 'median', **kwargs)
 
     _shared_docs['std'] = dedent("""
-    %(name)s standard deviation
+    Calculate %(name)s standard deviation.
+
+    Normalized by N-1 by default. This can be changed using the ddof argument.
```
jorisvandenbossche: Can you put single backticks around ddof (like `` `ddof` ``) (because it is a parameter name)

pandas/core/window.py
```diff
@@ -857,13 +857,58 @@ def median(self, **kwargs):
...
+    Notes
+    -----
+    A minimum of 1 periods is required for the rolling calculation.
+
+    Examples
+    --------
+    The below example will show a rolling example
+
+    >>> s = pd.Series((5,5,5,5,6,7,9,10,5,5,5,5))
+    >>> s.rolling(3).std(1)
```
jorisvandenbossche: if you use the default, I would leave out the `1` (just do `std()`)

pandas/core/window.py
```diff
@@ -857,13 +857,58 @@ def median(self, **kwargs):
...
+
+    Notes
+    -----
+    A minimum of 1 periods is required for the rolling calculation.
+
+    Examples
+    --------
+    The below example will show a rolling example
+
+    >>> s = pd.Series((5,5,5,5,6,7,9,10,5,5,5,5))
```
jorisvandenbossche: Can you make this PEP8: spaces after the comma. 
Also: I would maybe make it a little bit shorter, and it is more typical to use a list to construct the data instead of a tuple.

jorisvandenbossche (commented):
Thanks for the nice PR! Added few comments

pandas/core/window.py
jorisvandenbossche: Can you address the second part of the comment as well?

pandas/core/window.py
```diff
@@ -857,13 +857,70 @@ def median(self, **kwargs):
...
-        is ``N - ddof``, where ``N`` represents the number of elements.""")
+        is ``N - ddof``, where ``N`` represents the number of elements.
+    *args
+        Under Review.
+    **kwargs
+        Under Review.
+
+    Returns
+    -------
+    Returns the same object type as determined by the caller of the %(name)s
```
jorisvandenbossche: Same comment here as on the other PR https://github.com/pandas-dev/pandas/pull/20233

pandas/core/window.py
```diff
@@ -857,13 +857,70 @@ def median(self, **kwargs):
...
+    4     0.577350
+    5     1.000000
+    6     1.527525
+    7     1.527525
+    8     2.645751
+    9     2.886751
+    10    0.000000
+    11    0.000000
+    dtype: float64
+    >>> s.expanding(3).std()
```
jreback: blank line between cases

jreback (changes requested):

TomAugspurger:
Changed a `var` reference to `std`. Merging. Thanks @mmaybeno!

mmaybeno:
Ah good catch. Copy pasta fail. Thanks again!

PR #20236, opened 2018-03-10
DOC: update the pandas.DataFrame.plot.density docstring

jonas-schulze:
This docstring is quite similar to the one of #20041.

Unfortunately, I was not able to compute a kernel estimate of a
two-dimensional random variable. Hence, the example is more of an
analysis of some independent data series.

an unrelated PR):

- [ ] It has been proofread on language by another sprint participant (sort of; it's based on PR20041)

```
################## Docstring (pandas.DataFrame.plot.density)  ##################

Generate Kernel Density Estimate plot using Gaussian kernels.

In statistics, `kernel density estimation`_ (KDE) is a non-parametric
way to estimate the probability density function (PDF) of a random
variable. This function uses Gaussian kernels and includes automatic
bandwith determination.

.. _kernel density estimation:
    https://en.wikipedia.org/wiki/Kernel_density_estimation

Parameters
----------
bw_method : str, scalar or callable, optional
    The method used to calculate the estimator bandwidth. This can be
    'scott', 'silverman', a scalar constant or a callable.
    If None (default), 'scott' is used.
    See :class:`scipy.stats.gaussian_kde` for more information.
ind : NumPy array or integer, optional
    Evaluation points for the estimated PDF. If None (default),
    1000 equally spaced points are used. If `ind` is a NumPy array, the
    KDE is evaluated at the points passed. If `ind` is an integer,
    `ind` number of equally spaced points are used.
**kwds : optional

Returns
-------

See Also
--------
scipy.stats.gaussian_kde : Representation of a kernel-density
    estimate using Gaussian kernels. This is the function used
    internally to estimate the PDF.
Series.plot.kde : Generate a KDE plot for a Series.

Examples
--------
Given several Series of points randomly sampled from unknown
distributions, estimate their distribution using KDE with automatic
bandwidth determination and plot the results, evaluating them at
1000 equally spaced points (default):

.. plot::
    :context: close-figs

    >>> df = pd.DataFrame({
    ...     'x': [1, 2, 2.5, 3, 3.5, 4, 5],
    ...     'y': [4, 4, 4.5, 5, 5.5, 6, 6],
    ... })
    >>> ax = df.plot.kde()

A scalar bandwidth can be specified. Using a small bandwidth value can
lead to overfitting, while using a large bandwidth value may result
in underfitting:

.. plot::
    :context: close-figs

    >>> ax = df.plot.kde(bw_method=0.3)

.. plot::
    :context: close-figs

    >>> ax = df.plot.kde(bw_method=3)

Finally, the `ind` parameter determines the evaluation points for the
plot of the estimated PDF:

.. plot::
    :context: close-figs

    >>> ax = df.plot.kde(ind=[1, 2, 3, 4, 5, 6])

Errors found:
	Errors in parameters section
		Unknown parameters {'**kwds'}
```

I left the error because David and Israel mentioned doing it either way on Gitter ..

### DIFF
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -2618,12 +2618,15 @@ class SeriesPlotMethods(BasePlotMethods):
 
     def kde(self, bw_method=None, ind=None, **kwds):
         """
-        Kernel Density Estimate plot using Gaussian kernels.
+        Generate Kernel Density Estimate plot using Gaussian kernels.
 
-        In statistics, kernel density estimation (KDE) is a non-parametric way
-        to estimate the probability density function (PDF) of a random
+        In statistics, `kernel density estimation`_ (KDE) is a non-parametric
+        way to estimate the probability density function (PDF) of a random
         variable. This function uses Gaussian kernels and includes automatic
         bandwith determination.
+
+        .. _kernel density estimation:
+            https://en.wikipedia.org/wiki/Kernel_density_estimation
 
         Parameters
         ----------
@@ -2635,9 +2638,9 @@ class SeriesPlotMethods(BasePlotMethods):
         ind : NumPy array or integer, optional
             Evaluation points for the estimated PDF. If None (default),
             1000 equally spaced points are used. If `ind` is a NumPy array, the
-            kde is evaluated at the points passed. If `ind` is an integer,
+            KDE is evaluated at the points passed. If `ind` is an integer,
             `ind` number of equally spaced points are used.
-        kwds : optional
+        **kwds : optional
             Additional keyword arguments are documented in
             :meth:`pandas.Series.plot`.
 
@@ -2645,16 +2648,17 @@ class SeriesPlotMethods(BasePlotMethods):
         -------
         axes : matplotlib.AxesSubplot or np.array of them
 
-        See also
+        See Also
         --------
         scipy.stats.gaussian_kde : Representation of a kernel-density
             estimate using Gaussian kernels. This is the function used
             internally to estimate the PDF.
+        DataFrame.plot.kde : Generate a KDE plot for a DataFrame.
 
         Examples
         --------
         Given a Series of points randomly sampled from an unknown
-        distribution, estimate this distribution using KDE with automatic
+        distribution, estimate its distribution using KDE with automatic
         bandwidth determination and plot the results, evaluating them at
         1000 equally spaced points (default):
 
@@ -2664,10 +2668,9 @@ class SeriesPlotMethods(BasePlotMethods):
             >>> s = pd.Series([1, 2, 2.5, 3, 3.5, 4, 5])
             >>> ax = s.plot.kde()
 
-
-        An scalar fixed bandwidth can be specified. Using a too small bandwidth
-        can lead to overfitting, while a too large bandwidth can result in
-        underfitting:
+        A scalar bandwidth can be specified. Using a small bandwidth value can
+        lead to overfitting, while using a large bandwidth value may result
+        in underfitting:
 
         .. plot::
             :context: close-figs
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Damn, we again have the problem of miscommunication that this one was already take one of the previous days: https://github.com/pandas-dev/pandas/pull/20041 
Sorry about that!

cc @dukebody it would be good if one of both of you can combine both PRs and take the best bits of both PRs

TomAugspurger:
If either of you are comfortable with git, you should be able to cherry pick the other's commit so you'll both show up as authors.

LMK if you need help with the git stuff.

jonas-schulze:
@jorisvandenbossche and @TomAugspurger thanks for the quick responses! I noticed the other PR during the sprint, but that one improved the function of Series and not DataFrame. I'm not that comfortable in git, I'd appreciate any help. What about the requested changes in the other PR?

jorisvandenbossche:
Hmm, OK, didn't see it was DataFrame vs Series. 
We should probably share the docstrings somehow, but let's do that in a follow-up issue. Can you for now see if can 'steal' ideas of the other one?

jonas-schulze:
I agree; I had the same idea with sharing the docstrings between DataFrames and Series. Most DataFrame plotting functions just seem to "pass the call" to the Series it's containing.

What do you mean by 'steal ideas of the other one'? I basically copied the Series docstring, included all (but `See Also`) requested changes and extended the examples Series to 2-column DataFrames.

jorisvandenbossche:
Ah, then you already 'stole' the ideas :-) 
No, I assumed you wrote it completely independent, and then it would have been nice to combine the different enhancements in both PRs here. But that already happened then

jorisvandenbossche:
So I think the main difference between the frame vs series methods is that the frame version has `x, y` as keywords?

jreback:
this dupes #20041

jorisvandenbossche:
@jreback see discussion we are having. It's not an exact duplciate, as it is frame vs series. But I agree the docstring is certainly duplicative and can be shared. 
We only need to figure out how to exactly do that.

jreback:
@jorisvandenbossche yep, just noticed. nvm.

jorisvandenbossche:
OK, I merged the other PR. 
@jonas-schulze the comments that were left there (and the ones you added yourself), can you address them here?

pandas/plotting/_core.py
```diff
@@ -2791,26 +2791,77 @@ def hist(self, by=None, bins=10, **kwds):
...
-            spaced points are used.
-        `**kwds` : optional
-            Keyword arguments to pass on to :py:meth:`pandas.DataFrame.plot`.
+            Evaluation points for the estimated PDF. If None (default),
+            1000 equally spaced points are used. If `ind` is a NumPy array, the
+            kde is evaluated at the points passed. If `ind` is an integer,
+            `ind` number of equally spaced points are used.
+        **kwds : optional
+            Additional keyword arguments are documented in
+            :meth:`pandas.Series.plot`.
```
jonas-schulze: Should I change this one to `pandas.DataFrame.plot` (copy-and-paste 🙈)? I tried to understand the code but it's not yet clear to me. Both functions eventually call `MPLPlot._plot()`.

jonas-schulze (commented):
Yep, I'll do that and squash the commits if I am able to.

It does provide `x` and `y` as it passes them to `Series.plot` or `DataFrame.plot` (which one? depending on the answer, I have to fix the `See Also` section), but I wasn't able to figure out what `x` does ..

pandas/plotting/_core.py
TomAugspurger: yeah, `pandas.DataFrame.plot` is fine.

pandas/plotting/_core.py
```diff
@@ -2791,26 +2791,77 @@ def hist(self, by=None, bins=10, **kwds):
...
         Returns
         -------
         axes : matplotlib.AxesSubplot or np.array of them
+
+        See also
+        --------
+        scipy.stats.gaussian_kde : Representation of a kernel-density
+            estimate using Gaussian kernels. This is the function used
+            internally to estimate the PDF.
+        :meth:`pandas.Series.plot.kde` : Generate a KDE plot for a Series
```
TomAugspurger: Just

    Series.plot.kde : Generate a KDE plot for a Series.

jonas-schulze:
I included the changes and updated the very first comment to include the new output of the validation script.

jonas-schulze:
What do you think of my concern with the docstring / the DataFrame density function?
```
Unfortunately, I was not able to compute a kernel estimate of a
two-dimensional random variable. Hence, the example is more of an
analysis of some independent data series.
```

jorisvandenbossche:
Updates to the existing docstring look good!

What is the difference now exactly between both docstrings? Only the examples section?

jonas-schulze:
Unless I missed something, that's the only difference.

jorisvandenbossche:
Can you then maybe pull out the common part into a variable, and use `@Appender(_kde_docstring % _kde_series_example)` to add it to both functions?

jorisvandenbossche:
(if that was very cryptic, say so :-))

jonas-schulze:
I just noticed two more differences:

1. the `kwds` arguments include Series and DataFrame
2. the `See Also` sections references the other command

So there would be some more processing required than just appending the examples .. shall I still proceed? I think it might be better to come up with a more general solution to this phenomenon of "sibling functions" between Series and DataFrame. It should then be applied to all the other plotting functions as well once their PRs got merged.

jonas-schulze:
The new commit for the unified docstrings introduces two new errors in the validation script:
```
Errors found:
```
I have no idea how to fix them .. can you help me please?

pandas/plotting/_core.py
```diff
@@ -2616,45 +2660,12 @@ def hist(self, bins=10, **kwds):
...
-
-        Examples
-        --------
+    @Appender(_kde_docstring % {
+        'this-datatype': 'Series',
+        'sibling-datatype': 'DataFrame',
+        'examples': """
         Given a Series of points randomly sampled from an unknown
-        distribution, estimate this distribution using KDE with automatic
+        distribution, estimate its distribution using KDE with automatic
```
dukebody: I think the original was right: what we estimate is the distribution of the random sample. If you say "estimate its distribution" it looks like we are estimating the distribution of the distribution. But I'm not a native English speaker. What do you think?

dukebody: Saw your comment in my PR about this.

I personally prefer "this" because I feel it makes clearer that what is estimated is the unknown distribution. Otherwise "its" refers to the Series, which is early in the sentence, and is more confusing to me.

But I'm not a native English speaker. I'm fine with both options, don't like bikeshedding too much :)

dukebody:
@jonas-schulze Great work on combining both docstrings and addressing the comments on my PR! LGTM.

Regarding the issue raised by the validator, can you check the output of the compiled docstring when running it in both methods? It's probably a silly formatting issue when using the Appender.

dukebody (approved):
@jonas-schulze Great work on combining both docstrings and addressing the comments on my PR! LGTM.

Regarding the issue raised by the validator, can you check the output of the compiled docstring when running it in both methods? It's probably a silly formatting issue when using the Appender.

pandas/plotting/_core.py
```diff
@@ -2616,45 +2660,12 @@ def hist(self, bins=10, **kwds):
...
-            estimate using Gaussian kernels. This is the function used
-            internally to estimate the PDF.
-
-        Examples
-        --------
+    @Appender(_kde_docstring % {
+        'this-datatype': 'Series',
+        'sibling-datatype': 'DataFrame',
+        'examples': """
         Given a Series of points randomly sampled from an unknown
```
jorisvandenbossche: You need to give this one level of indentation less (now it 2, should only be 1 to match the level of identation you gave to the template)

(and the same for the dataframe ones)

jorisvandenbossche (commented):
Perfect!

Added small comment regarding the indentation

> The new commit for the unified docstrings introduces two new errors in the validation script:

You can ignore those I think. I don't see the two blank lines, and the issue of the closing quotes is mainly for normal in-code docstrings. For those templated ones, it doesn't really matter (it doesn't affect rendering in sphinx)

jonas-schulze:
Adding a `.strip()` to the dictionaries fixes the validation errors. In order to keep the amount of changes low, and because the strings would look very weird otherwise, I added 1 level of indent to the common docstring above `class KdePlot`.

pandas/plotting/_core.py
jonas-schulze: I added some indentation to the common part instead. Removing the indentation at this place just doesn't look right. 🙈

pandas/plotting/_core.py
jonas-schulze: Me neither .. just because we are already thinking of it: changing the whole part to `estimate its PDF using KDE` might be an even better fit, as this is what we are really doing.

jonas-schulze: I just changed it to that.

jonas-schulze:
Off topic: why doesn't pandas have a panda as its logo/mascot?

dukebody (approved):
LGTM

TomAugspurger:
Thanks!

PR #20245, opened 2018-03-10
DOC: update the DataFrame.cov docstring

sechilds:
Update the docstring with some examples from
elsewhere in the pandas documentation.

Some of the examples use randomly generated time series
because we need to get covariance between long series.
Used a random seed to ensure that the results are the
same each time.

an unrelated PR):

```

Docstring for "pandas.DataFrame.cov" correct. :)
```

explicitly.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -5501,7 +5501,22 @@ class DataFrame(NDFrame):
 
     def cov(self, min_periods=None):
         """
-        Compute pairwise covariance of columns, excluding NA/null values
+        Compute pairwise covariance of columns, excluding NA/null values.
+
+        Compute the pairwise covariance among the series of a DataFrame.
+        The returned data frame is the `covariance matrix
+        <https://en.wikipedia.org/wiki/Covariance_matrix>`__ of the columns
+        of the DataFrame.
+
+        Both NA and null values are automatically excluded from the
+        calculation. (See the note below about bias from missing values.)
+        A threshold can be set for the minimum number of
+        observations for each value created. Comparisons with observations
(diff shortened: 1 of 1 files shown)


pandas/core/frame.py
```diff
@@ -5511,12 +5526,71 @@ def cov(self, min_periods=None):
...
         Returns
         -------
-        y : DataFrame
+        DataFrame
+            The covariance matrix of the series of the DataFrame.
+
+        See Also
+        --------
+        Series.cov : compute covariance with another Series
+        core.window.EWM.cov: expoential weighted sample covariance
```
TomAugspurger: These should all start with `pandas.` so that the are linked correctly.

pandas/core/frame.py
```diff
@@ -5511,12 +5526,71 @@ def cov(self, min_periods=None):
...
+        a  0.998438 -0.020161  0.059277 -0.008943  0.014144
+        b -0.020161  1.059352 -0.008543 -0.024738  0.009826
+        c  0.059277 -0.008543  1.010670 -0.001486 -0.000271
+        d -0.008943 -0.024738 -0.001486  0.921297 -0.013692
+        e  0.014144  0.009826 -0.000271 -0.013692  0.977795
+
+        **Minimum number of periods**
+
+        This method also supports an optional ``min_periods`` keyword
+        that specifies the required minimum number of observations for each
```
TomAugspurger: maybe non-NA observations, just to be 100% clear :)

TomAugspurger (commented):
Minor comment about the See Also. LGTM otherwise.

TomAugspurger:
This looks very good. Thanks!

PR #20246, opened 2018-03-10
DOC: update `pandas/core/ops.py` docstring template to accept examples

alysivji:
Checklist for the pandas documentation sprint

The Chicago assigned docstrings uses a template with string formatting. To add examples for the various DataFrame operators, I modified the template to allow examples for each operator.

Opening this PR to get comments on this workflow.

I will work with all the ChiPy contributors to make sure we can get their examples in once we figure out the best way forward.

### DIFF
--- a/pandas/core/ops.py
+++ b/pandas/core/ops.py
@@ -343,7 +343,7 @@ def _get_op_name(op, special):
 # -----------------------------------------------------------------------------
 # Docstring Generation and Templates
 
-_add_doc_FRAME = """
+_add_example_FRAME = """
 >>> a = pd.DataFrame([1, 1, 1, np.nan], index=['a', 'b', 'c', 'd'],
                      columns=['one'])
 >>> a
@@ -375,7 +375,7 @@ _op_descriptions = {
     'add': {'op': '+',
             'desc': 'Addition',
             'reverse': 'radd',
-            'df_examples': _add_doc_FRAME},
+            'df_examples': _add_example_FRAME},
     'sub': {'op': '-',
             'desc': 'Subtraction',
             'reverse': 'rsub',
--- a/pandas/core/ops.py
+++ b/pandas/core/ops.py
@@ -343,50 +343,93 @@ def _get_op_name(op, special):
 # -----------------------------------------------------------------------------
 # Docstring Generation and Templates
 
+_add_doc_FRAME = """
+>>> a = pd.DataFrame([1, 1, 1, np.nan], index=['a', 'b', 'c', 'd'],
+                     columns=['one'])
+>>> a
+   one
+a  1.0
+b  1.0
+c  1.0
+d  NaN
+>>> b = pd.DataFrame(dict(one=[1, np.nan, 1, np.nan],
+                          two=[np.nan, 2, np.nan, 2]),
+                     index=['a', 'b', 'd', 'e'])
+>>> b
+   one  two
+a  1.0  NaN
+b  NaN  2.0
+d  1.0  NaN
+e  NaN  2.0
+>>> a.add(b, fill_value=0)
+   one  two
+a  2.0  NaN
+b  1.0  2.0
+c  1.0  NaN
+d  1.0  NaN
+e  NaN  2.0
+"""
+
 _op_descriptions = {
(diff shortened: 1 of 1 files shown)


TomAugspurger:
This one is tricky, since we have so many potential examples. I think ideally we would have a docstring stub that creates the dataframes, and then examples stubs for each of the arithmetic methods.

TomAugspurger:
Something like (untested)


```python
_arith_frames = """\
>>> a = pd.DataFrame([1, 1, 1, np.nan], index=['a', 'b', 'c', 'd'],
                     columns=['one'])
>>> a
   one
a  1.0
b  1.0
c  1.0
d  NaN
>>> b = pd.DataFrame(dict(one=[1, np.nan, 1, np.nan],
                          two=[np.nan, 2, np.nan, 2]),
                     index=['a', 'b', 'd', 'e'])
>>> b
   one  two
a  1.0  NaN
b  NaN  2.0
d  1.0  NaN
e  NaN  2.0
"""

_add_doc_frame = """\
Examples
--------
{create}
>>> a.add(b, fill_value=0)
   one  two
a  2.0  NaN
b  1.0  2.0
c  1.0  NaN
d  1.0  NaN
e  NaN  2.0 
""".format(create=_arith_frames)
```

jorisvandenbossche:
@alysivji Do you want us to merge this one? So it is easier for the other PRs to add the examples?

pandas/core/ops.py
```diff
@@ -343,50 +343,93 @@ def _get_op_name(op, special):
 # -----------------------------------------------------------------------------
 # Docstring Generation and Templates
 
+_add_doc_FRAME = """
```
jorisvandenbossche: maybe use "example" in the name?

alysivji:
@TomAugspurger each team did different examples.

@jorisvandenbossche if we can come up with a best practice i think that would be best. i can help get all the examples in once we figure out the best workflow.

TomAugspurger:
@alysivji I think you may know best here, since you've seen some of the other examples I think.

I suspect that the PRs would have many conflicts (removing the old example, updating the kwargs, etc.) and that the only distinct bits are the actual new example and updating the `df_examples` value. If that's the case, let's merge this and then start taking PRs that have an updated copy of master + the new example.

jorisvandenbossche:
Yes, I think it is fine to merge this.

pandas/core/ops.py
```diff
@@ -343,50 +343,93 @@ def _get_op_name(op, special):
...
+_add_example_FRAME = """
+>>> a = pd.DataFrame([1, 1, 1, np.nan], index=['a', 'b', 'c', 'd'],
+                     columns=['one'])
+>>> a
+   one
+a  1.0
+b  1.0
+c  1.0
+d  NaN
+>>> b = pd.DataFrame(dict(one=[1, np.nan, 1, np.nan],
```
jreback: blank line between ases

pandas/core/ops.py
```diff
@@ -343,50 +343,93 @@ def _get_op_name(op, special):
...
+                          two=[np.nan, 2, np.nan, 2]),
+                     index=['a', 'b', 'd', 'e'])
+>>> b
+   one  two
+a  1.0  NaN
+b  NaN  2.0
+d  1.0  NaN
+e  NaN  2.0
+>>> a.add(b, fill_value=0)
```
jreback: could add some text here to indicate what the filling is doing

jreback (approved):
ok to merge, my comments can be addressed in the next iteration.

jreback:
thanks @alysivji

pandas/core/ops.py
kurtiskerstein: @jreback I am about to put in a PR for another one of the examples.  Where should there be a blank line? Between what would be "cell" executions?

kurtiskerstein: @jreback I am about to put in a PR for one of the examples.  Where should there be a blank line? Between what would be "cell" executions?

jorisvandenbossche: @kurtiskerstein I think there should be only blank lines around "blocks" of code that you want to have appear together in one code block

PR #20250, opened 2018-03-10
DOC: update the DataFrame.pivot	docstring

amuta:
an unrelated PR):

```
###################### Docstring (pandas.DataFrame.pivot) ######################

Return reshaped DataFrame summarized by given index / column values.

Reshape data (produce a "pivot" table) based on column values. Uses
unique values from index / columns to form axes of the resulting
DataFrame. This function does not support data aggregation, multiple
values will result in hierarchically indexed columns.

Parameters
----------
index : string or object, optional
    Column name to use to make new frame's index. If None, uses
    existing index.
columns : string or object, optional
    Column name to use to make new frame's columns.
values : string or object, optional
    Column name to use for populating new frame's values. If not
    specified, all remaining columns will be used and the result will
    have hierarchically indexed columns.

Returns
-------
DataFrame
    Returns reshaped DataFrame.

See Also
--------
DataFrame.pivot_table : generalization of pivot that can handle
    duplicate values for one index/column pair
DataFrame.unstack : pivot based on the index values instead of a
    column

Notes
-----
For finer-tuned control, see hierarchical indexing documentation along
with the related stack/unstack methods

Examples
--------

>>> df = pd.DataFrame({'foo': ['one','one','one','two','two','two'],
...                    'bar': ['A', 'B', 'C', 'A', 'B', 'C'],
...                    'baz': [1, 2, 3, 4, 5, 6]})
>>> df
    foo   bar  baz
0   one   A    1
1   one   B    2
2   one   C    3
3   two   A    4
4   two   B    5
5   two   C    6

>>> df.pivot(index='foo', columns='bar', values='baz')
bar  A   B   C
foo
one  1   2   3
two  4   5   6

>>> df.pivot(index='foo', columns='bar')['baz']
bar  A   B   C
foo
one  1   2   3
two  4   5   6

Docstring for "pandas.DataFrame.pivot" correct. :)
```

explicitly.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4346,7 +4346,8 @@ class DataFrame(NDFrame):
         Reshape data (produce a "pivot" table) based on column values. Uses
         unique values from specified `index` / `columns` to form axes of the resulting
         DataFrame. This function does not support data aggregation, multiple
-        values will result in a MultiIndex in the columns.
+        values will result in a MultiIndex in the columns. See the
+        :ref:`User Guide <reshaping>` for more on reshaping.
 
         Parameters
         ----------
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4345,7 +4345,7 @@ class DataFrame(NDFrame):
 
         Reshape data (produce a "pivot" table) based on column values. Uses
         unique values from index / columns to form axes of the resulting
-        DataFrame. This function don't support data aggregation, multiple
+        DataFrame. This function does not support data aggregation, multiple
         values will result in hierarchically indexed columns.
 
         Parameters
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4341,27 +4341,31 @@ class DataFrame(NDFrame):
 
     def pivot(self, index=None, columns=None, values=None):
         """
+        Return reshaped DataFrame summarized by given index / column values.
+
         Reshape data (produce a "pivot" table) based on column values. Uses
         unique values from index / columns to form axes of the resulting
-        DataFrame.
+        DataFrame. This function don't support data aggregation, multiple
+        values will result in hierarchically indexed columns.
 
         Parameters
         ----------
         index : string or object, optional
             Column name to use to make new frame's index. If None, uses
             existing index.
-        columns : string or object
-            Column name to use to make new frame's columns
+        columns : string or object, optional
+            Column name to use to make new frame's columns.
         values : string or object, optional
             Column name to use for populating new frame's values. If not
             specified, all remaining columns will be used and the result will
-            have hierarchically indexed columns
+            have hierarchically indexed columns.
 
         Returns
         -------
-        pivoted : DataFrame
+        DataFrame
+            Returns reshaped DataFrame.
 
-        See also
+        See Also
         --------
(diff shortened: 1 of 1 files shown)


pandas/core/frame.py
```diff
@@ -4341,27 +4341,31 @@ def last_valid_index(self):
 
     def pivot(self, index=None, columns=None, values=None):
         """
+        Return reshaped DataFrame summarized by given index / column values.
```
jorisvandenbossche: I think "summarized" is misleading here, as there is no "summary" (aggregation) calculation happening. 

(I don't directly have the "better" wording in mind though. And the "reshaping" is important, as you already mention)

pandas/core/frame.py
```diff
@@ -4341,27 +4341,31 @@ def last_valid_index(self):
...
+
         Reshape data (produce a "pivot" table) based on column values. Uses
         unique values from index / columns to form axes of the resulting
```
jorisvandenbossche: can you add single backtick around index and column ? (to reference them as parameters). 

I would maybe also say "from specified index / columns" to make it clear it is not the existing index / columns

TomAugspurger (approved):

pandas/core/frame.py
```diff
@@ -4341,27 +4341,31 @@ def last_valid_index(self):
...
 
         Parameters
         ----------
         index : string or object, optional
             Column name to use to make new frame's index. If None, uses
             existing index.
-        columns : string or object
-            Column name to use to make new frame's columns
+        columns : string or object, optional
+            Column name to use to make new frame's columns.
```
jorisvandenbossche: What happens in this case if you do not specify it?

pandas/core/frame.py
```diff
@@ -4341,27 +4341,31 @@ def last_valid_index(self):
...
         Reshape data (produce a "pivot" table) based on column values. Uses
         unique values from index / columns to form axes of the resulting
-        DataFrame.
+        DataFrame. This function does not support data aggregation, multiple
+        values will result in hierarchically indexed columns.
 
         Parameters
         ----------
         index : string or object, optional
             Column name to use to make new frame's index. If None, uses
```
jorisvandenbossche: Would it be clearer to just say "Column" instead of "Column name". I know the value you specify is the name, but what you actually use to make the new frame's index is the actual column, not the column name ...

jorisvandenbossche (commented):
Thanks! Added a few comments

Can you add a "Raises" section? As I think this is commonly encountered (so the case where you have duplicate values in index / column combinations, so when you need to use `pivot_table` instead)

pandas/core/frame.py
```diff
@@ -4341,44 +4341,48 @@ def last_valid_index(self):
...
     def pivot(self, index=None, columns=None, values=None):
         """
+        Return reshaped DataFrame organized by given index / column values.
+
         Reshape data (produce a "pivot" table) based on column values. Uses
-        unique values from index / columns to form axes of the resulting
-        DataFrame.
+        unique values from specified `index` / `columns` to form axes of the resulting
+        DataFrame. This function does not support data aggregation, multiple
+        values will result in hierarchically indexed columns.
```
jreback: in multiindexed columns.

jreback (changes requested):

PR #20253, opened 2018-03-10
DOC: update the `to_pickle` & `read_pickle` docstring

minggli:
an unrelated PR):

```
#################### Docstring (pandas.DataFrame.to_pickle) ####################

Pickle (serialize) object to input file path.

Parameters
----------
path : str
    File path where the pickled object will be stored.
compression : {'infer', 'gzip', 'bz2', 'xz', None}, default 'infer'
    A string representing the compression to use in the output file. By
    default, infers from the file extension in specified path.

    .. versionadded:: 0.20.0
protocol : int
    Int which indicates which protocol should be used by the pickler,
    default HIGHEST_PROTOCOL (see [1], paragraph 12.1.2). The possible
    values for this parameter depend on the version of Python. For
    Python 2.x, possible values are 0, 1, 2. For Python>=3.0, 3 is a
    valid value. For Python >= 3.4, 4 is a valid value. A negative
    value for the protocol parameter is equivalent to setting its value
    to HIGHEST_PROTOCOL.

    .. [1] https://docs.python.org/3/library/pickle.html
    .. versionadded:: 0.21.0

See Also
--------
read_pickle : Load pickled pandas object (or any object) from file.
DataFrame.to_hdf : Write the contained data to an HDF5 file using
    HDFStore.
DataFrame.to_sql : Write records stored in a DataFrame to a SQL
    database.
DataFrame.to_parquet : Write a DataFrame to the binary parquet format.

Examples
--------
>>> original_df = pd.DataFrame({"foo": range(5), "bar": range(5, 10)})
>>> original_df
   foo  bar
0    0    5
1    1    6
2    2    7
3    3    8
4    4    9
>>> original_df.to_pickle("./dummy.pkl")

>>> unpickled_df = pd.read_pickle("./dummy.pkl")
>>> unpickled_df
   foo  bar
0    0    5
1    1    6
2    2    7
3    3    8
4    4    9

>>> import os
>>> os.remove("./dummy.pkl")

Errors found:
	No extended summary found
	Errors in parameters section
		Parameter "compression" description should finish with "."
		Parameter "protocol" description should finish with "."
	No returns section found
```

description already finishes with "."
.. versionadded:: tag should not end with "."

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1698,10 +1698,11 @@ class NDFrame(PandasObject, SelectionMixin):
 
     See Also
     --------
-    pandas.read_pickle
-    pandas.to_hdf
-    pandas.to_sql
-    pandas.to_parquet
+    pandas.read_pickle : Load pickled pandas object (or any other pickled
+    object) from the specified file path.
+    pandas.to_hdf : Write the contained data to an HDF5 file using HDFStore.
+    pandas.to_sql : Write records stored in a DataFrame to a SQL database.
+    pandas.to_parquet : Write a DataFrame to the binary parquet format.
     """
 
     def to_json(self, path_or_buf=None, orient=None, date_format=None,
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1699,10 +1699,13 @@ class NDFrame(PandasObject, SelectionMixin):
     See Also
     --------
     pandas.read_pickle : Load pickled pandas object (or any other pickled
-    object) from the specified file path.
-    pandas.to_hdf : Write the contained data to an HDF5 file using HDFStore.
-    pandas.to_sql : Write records stored in a DataFrame to a SQL database.
-    pandas.to_parquet : Write a DataFrame to the binary parquet format.
+                         object) from the specified file path.
+    pandas.DataFrame.to_hdf : Write the contained data to an HDF5 file using
+                              HDFStore.
+    pandas.DataFrame.to_sql : Write records stored in a DataFrame to a SQL
+                              database.
+    pandas.DataFrame.to_parquet : Write a DataFrame to the binary parquet
+                                  format.
     """
 
     def to_json(self, path_or_buf=None, orient=None, date_format=None,
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1660,7 +1660,7 @@ class NDFrame(PandasObject, SelectionMixin):
     path : string
         File path where the pickled %(klass)s object will be stored.
     compression : {'infer', 'gzip', 'bz2', 'xz', None}, default 'infer'
-        a string representing the compression to use in the output file
+        A string representing the compression to use in the output file.
 
         .. versionadded:: 0.20.0
     protocol : int
@@ -1677,24 +1677,25 @@ class NDFrame(PandasObject, SelectionMixin):
 
     Examples
     --------
+    >>> from pandas import DataFrame, read_pickle
     >>> original_df = DataFrame({"foo": range(5), "bar": range(5, 10)})
     >>> original_df
-       bar  foo
-    0    5    0
-    1    6    1
-    2    7    2
-    3    8    3
-    4    9    4
+       foo  bar
+    0    0    5
+    1    1    6
+    2    2    7
+    3    3    8
+    4    4    9
     >>> original_df.to_pickle("./dummy.pkl")
 
     >>> unpickled_df = read_pickle("./dummy.pkl")
     >>> unpickled_df
-       bar  foo
-    0    5    0
-    1    6    1
-    2    7    2
-    3    8    3
-    4    9    4
+       foo  bar
+    0    0    5
+    1    1    6
+    2    2    7
+    3    3    8
+    4    4    9
     """
 
     def to_json(self, path_or_buf=None, orient=None, date_format=None,
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1652,6 +1652,51 @@ class NDFrame(PandasObject, SelectionMixin):
     strings before writing.
     """
 
+    _shared_docs['to_pickle'] = """
+    Pickle (serialize) %(klass)s object to input file path.
+
+    Parameters
+    ----------
+    path : string
+        File path where the pickled %(klass)s object will be stored.
+    compression : {'infer', 'gzip', 'bz2', 'xz', None}, default 'infer'
+        a string representing the compression to use in the output file
+
+        .. versionadded:: 0.20.0
+    protocol : int
+        Int which indicates which protocol should be used by the pickler,
+        default HIGHEST_PROTOCOL (see [1], paragraph 12.1.2). The possible
+        values for this parameter depend on the version of Python. For
+        Python 2.x, possible values are 0, 1, 2. For Python>=3.0, 3 is a
+        valid value. For Python >= 3.4, 4 is a valid value.A negative value
+        for the protocol parameter is equivalent to setting its value to
+        HIGHEST_PROTOCOL.
+
+        .. [1] https://docs.python.org/3/library/pickle.html
+        .. versionadded:: 0.21.0
+
+    Examples
+    --------
+    >>> original_df = DataFrame({"foo": range(5), "bar": range(5, 10)})
+    >>> original_df
+       bar  foo
+    0    5    0
+    1    6    1
+    2    7    2
+    3    8    3
+    4    9    4
+    >>> original_df.to_pickle("./dummy.pkl")
+
+    >>> unpickled_df = read_pickle("./dummy.pkl")
+    >>> unpickled_df
+       bar  foo
+    0    5    0
+    1    6    1
+    2    7    2
+    3    8    3
+    4    9    4
+        """
+
     def to_json(self, path_or_buf=None, orient=None, date_format=None,
                 double_precision=10, force_ascii=True, date_unit='ms',
                 default_handler=None, lines=False, compression=None,
@@ -1900,49 +1945,6 @@ class NDFrame(PandasObject, SelectionMixin):
 
     def to_pickle(self, path, compression='infer',
                   protocol=pkl.HIGHEST_PROTOCOL):
-        """
-        Pickle (serialize) object to input file path.
-
-        Parameters
-        ----------
-        path : string
-            File path where the pickled pandas object will be stored.
-        compression : {'infer', 'gzip', 'bz2', 'xz', None}, default 'infer'
-            a string representing the compression to use in the output file
-
-            .. versionadded:: 0.20.0
-        protocol : int
-            Int which indicates which protocol should be used by the pickler,
-            default HIGHEST_PROTOCOL (see [1], paragraph 12.1.2). The possible
-            values for this parameter depend on the version of Python. For
-            Python 2.x, possible values are 0, 1, 2. For Python>=3.0, 3 is a
-            valid value. For Python >= 3.4, 4 is a valid value.A negative value
-            for the protocol parameter is equivalent to setting its value to
-            HIGHEST_PROTOCOL.
-
-            .. [1] https://docs.python.org/3/library/pickle.html
-            .. versionadded:: 0.21.0
-
-        Examples
-        --------
-        >>> original_df = DataFrame({"foo": range(5), "bar": range(5, 10)})
-        >>> original_df
-           bar  foo
-        0    5    0
-        1    6    1
-        2    7    2
-        3    8    3
-        4    9    4
-        >>> original_df.to_pickle("./dummy.pkl")
-        >>> unpickled_df = read_pickle("./dummy.pkl")
-        >>> unpickled_df
-           bar  foo
-        0    5    0
-        1    6    1
-        2    7    2
-        3    8    3
-        4    9    4
-        """
         from pandas.io.pickle import to_pickle
         return to_pickle(self, path, compression=compression,
                          protocol=protocol)
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1906,7 +1906,7 @@ class NDFrame(PandasObject, SelectionMixin):
         Parameters
         ----------
         path : string
-            File path
+            File path where the pickled pandas object will be stored.
         compression : {'infer', 'gzip', 'bz2', 'xz', None}, default 'infer'
             a string representing the compression to use in the output file
 
@@ -1923,6 +1923,25 @@ class NDFrame(PandasObject, SelectionMixin):
             .. [1] https://docs.python.org/3/library/pickle.html
             .. versionadded:: 0.21.0
 
+        Examples
+        --------
+        >>> original_df = DataFrame({"foo": range(5), "bar": range(5, 10)})
+        >>> original_df
+           bar  foo
+        0    5    0
+        1    6    1
+        2    7    2
+        3    8    3
+        4    9    4
+        >>> original_df.to_pickle("./dummy.pkl")
+        >>> unpickled_df = read_pickle("./dummy.pkl")
+        >>> unpickled_df
+           bar  foo
+        0    5    0
+        1    6    1
+        2    7    2
+        3    8    3
+        4    9    4
         """
         from pandas.io.pickle import to_pickle
         return to_pickle(self, path, compression=compression,
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1677,8 +1677,7 @@ class NDFrame(PandasObject, SelectionMixin):
 
     Examples
     --------
-    >>> from pandas import DataFrame, read_pickle
-    >>> original_df = DataFrame({"foo": range(5), "bar": range(5, 10)})
+    >>> original_df = pd.DataFrame({"foo": range(5), "bar": range(5, 10)})
     >>> original_df
        foo  bar
     0    0    5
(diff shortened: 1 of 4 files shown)


pandas/io/pickle.py
```diff
@@ -33,7 +34,27 @@ def to_pickle(obj, path, compression='infer', protocol=pkl.HIGHEST_PROTOCOL):
         .. [1] https://docs.python.org/3/library/pickle.html
         .. versionadded:: 0.21.0
 
-
+    Examples
+    --------
+    >>> from pandas import DataFrame, read_pickle
```
jreback: you don't need these imports, instead use ``pd.DataFrame`` and ``pd.read_pickle``

pandas/io/pickle.py
```diff
@@ -33,7 +34,27 @@ def to_pickle(obj, path, compression='infer', protocol=pkl.HIGHEST_PROTOCOL):
...
+    >>> original_df = DataFrame({"foo": range(5), "bar": range(5, 10)})
+    >>> original_df
+       foo  bar
+    0    0    5
+    1    1    6
```
jreback: can you add a See Also and ref read_pickle (to_hdf, to_sql, to_parquet) also good

pandas/io/pickle.py
```diff
@@ -72,6 +93,28 @@ def read_pickle(path, compression='infer'):
     Returns
     -------
     unpickled : type of object stored in file
```
jreback: add a See Also and ref DataFrame.to_pickle, pd.read_hdf, pd.read_sql, pd.read_parquet

jreback (changes requested):

pandas/io/pickle.py
minggli: added.

pandas/io/pickle.py
minggli: fixed 👍

pandas/io/pickle.py
minggli: added 💯

pandas/core/generic.py
```diff
@@ -1652,6 +1652,55 @@ def _repr_latex_(self):
...
+    protocol : int
+        Int which indicates which protocol should be used by the pickler,
+        default HIGHEST_PROTOCOL (see [1], paragraph 12.1.2). The possible
+        values for this parameter depend on the version of Python. For
+        Python 2.x, possible values are 0, 1, 2. For Python>=3.0, 3 is a
+        valid value. For Python >= 3.4, 4 is a valid value.A negative value
+        for the protocol parameter is equivalent to setting its value to
+        HIGHEST_PROTOCOL.
+
+        .. [1] https://docs.python.org/3/library/pickle.html
```
jreback: I think we are putting these in a Reference section @jorisvandenbossche ?

pandas/io/pickle.py
```diff
@@ -72,6 +101,35 @@ def read_pickle(path, compression='infer'):
...
+    1    1    6
+    2    2    7
+    3    3    8
+    4    4    9
+
+    See Also
+    --------
+    pandas.DataFrame.to_pickle
+    pandas.Series.to_pickle
+    pandas.read_hdf,
```
jreback: don't put commas at the end

jreback: I think we add text describing each of these

pandas/io/pickle.py
```diff
@@ -33,7 +34,33 @@ def to_pickle(obj, path, compression='infer', protocol=pkl.HIGHEST_PROTOCOL):
...
+    0    0    5
+    1    1    6
+    2    2    7
+    3    3    8
+    4    4    9
+
+    See Also
+    --------
+    pandas.read_pickle
+    pandas.to_hdf
```
jreback: same

pandas/core/generic.py
```diff
@@ -1652,6 +1652,55 @@ def _repr_latex_(self):
...
+       foo  bar
+    0    0    5
+    1    1    6
+    2    2    7
+    3    3    8
+    4    4    9
+
+    See Also
+    --------
+    pandas.read_pickle
```
jreback: can you add similar to what you adding in other See Also below

jreback (changes requested):

pandas/io/pickle.py
minggli: corrected.

pandas/io/pickle.py
minggli: added.

pandas/core/generic.py
minggli: added.

pandas/core/generic.py
```diff
@@ -1652,6 +1652,62 @@ def _repr_latex_(self):
...
+    compression : {'infer', 'gzip', 'bz2', 'xz', None}, default 'infer'
+        A string representing the compression to use in the output file.
+
+        .. versionadded:: 0.20.0
+    protocol : int
+        Int which indicates which protocol should be used by the pickler,
+        default HIGHEST_PROTOCOL (see [1], paragraph 12.1.2). The possible
+        values for this parameter depend on the version of Python. For
+        Python 2.x, possible values are 0, 1, 2. For Python>=3.0, 3 is a
+        valid value. For Python >= 3.4, 4 is a valid value.A negative value
```
jorisvandenbossche: space after '.'

pandas/core/generic.py
```diff
@@ -1652,6 +1652,62 @@ def _repr_latex_(self):
...
+       foo  bar
+    0    0    5
+    1    1    6
+    2    2    7
+    3    3    8
+    4    4    9
+
+    See Also
+    --------
+    pandas.read_pickle : Load pickled pandas object (or any other pickled
```
jorisvandenbossche: You can leave out the 'pandas' in all of those references

pandas/core/generic.py
jorisvandenbossche: That's the numpydoc way yes, but currently we have no docstring using that, so I would leave it like this for now

pandas/core/generic.py
```diff
@@ -1652,6 +1652,62 @@ def _repr_latex_(self):
...
 
+    _shared_docs['to_pickle'] = """
+    Pickle (serialize) %(klass)s object to input file path.
+
+    Parameters
+    ----------
+    path : string
+        File path where the pickled %(klass)s object will be stored.
+    compression : {'infer', 'gzip', 'bz2', 'xz', None}, default 'infer'
+        A string representing the compression to use in the output file.
```
jorisvandenbossche: Can you add that the default 'infer' infers it from the specified path?

jorisvandenbossche (commented):
Thanks for the PR!

I am not sure that I find it worth to use a shared_doc system here. It's used only for substituting Series/DataFrame in the summary line and path parameter explanation. I don't think that's worth the added complexity. I would rather make the docstring generic for both. It's not that the Series version has different options or behaviours.

pandas/core/generic.py
minggli: done.

pandas/core/generic.py
minggli: removed 👍

pandas/core/generic.py
minggli: added. 💯

minggli:
@jorisvandenbossche thanks for the comments. apply changes including reverting to generic docstring. agree that for the benefits, not necessary to use _shared_docs.

pandas/core/generic.py
```diff
@@ -1906,23 +1906,54 @@ def to_pickle(self, path, compression='infer',
...
+        >>> original_df
+           foo  bar
+        0    0    5
+        1    1    6
+        2    2    7
+        3    3    8
+        4    4    9
+        >>> original_df.to_pickle("./dummy.pkl")
+
+        >>> unpickled_df = pd.read_pickle("./dummy.pkl")
```
jreback: @jorisvandenbossche how handling file paths in doc-strings?

jreback (approved):

jreback:
lgtm. @jorisvandenbossche @datapythonista

pandas/core/generic.py
```diff
@@ -1906,23 +1906,54 @@ def to_pickle(self, path, compression='infer',
...
+        >>> unpickled_df
+           foo  bar
+        0    0    5
+        1    1    6
+        2    2    7
+        3    3    8
+        4    4    9
+
```
jorisvandenbossche: Can you add here a 

```
>>> import os
>>> os.remove("./dummy.pkl")
```

pandas/core/generic.py
```diff
@@ -1906,23 +1906,54 @@ def to_pickle(self, path, compression='infer',
...
+        See Also
+        --------
+        read_pickle : Load pickled pandas object (or any other pickled object)
+                      from the specified file path.
```
jorisvandenbossche: Can you only indent this line with 4 spaces? Like

```
read_pickle : Load pickled pandas object (or any other pickled object)
    from the specified file path.
```

Both work for sphinx, but we mainly use this pattern, so it's better to be consistent in this.

(same for the other ones below)

pandas/io/pickle.py
```diff
@@ -52,15 +81,17 @@ def to_pickle(obj, path, compression='infer', protocol=pkl.HIGHEST_PROTOCOL):
 def read_pickle(path, compression='infer'):
     """
     Load pickled pandas object (or any other pickled object) from the specified
-    file path
+    file path.
```
jorisvandenbossche: Can you try to fit this on one line? 

Maybe the "or any other pickled object" is not needed in the summary line and can go in an extended summary, as the typical use case should be pandas objects. 
Or, maybe the "from specified file path" can be shortened to "from file."

jorisvandenbossche (commented):
Thanks for the updates! Looking good

Added some minor additional comments. One more general comment: it might be nice to add an extended summary very briefly explaining what "pickle" is and linking to the python docs about it.

pandas/core/generic.py
jorisvandenbossche: @jreback commented about it (using os.remove to remove remaining file), issue to discuss this is here: https://github.com/pandas-dev/pandas/issues/20302

jreback: great, see that now.

pandas/core/generic.py
```diff
@@ -1906,23 +1906,54 @@ def to_pickle(self, path, compression='infer',
         Parameters
         ----------
         path : string
```
datapythonista: `str` instead of `string`

pandas/core/generic.py
```diff
@@ -1906,23 +1906,54 @@ def to_pickle(self, path, compression='infer',
...
-            File path
+            File path where the pickled object will be stored.
         compression : {'infer', 'gzip', 'bz2', 'xz', None}, default 'infer'
-            a string representing the compression to use in the output file
+            A string representing the compression to use in the output file. By
+            default, infers from the specified path.
```
datapythonista: "infers from the specified file extension" may be?

pandas/core/generic.py
```diff
@@ -1906,23 +1906,54 @@ def to_pickle(self, path, compression='infer',
...
             values for this parameter depend on the version of Python. For
             Python 2.x, possible values are 0, 1, 2. For Python>=3.0, 3 is a
-            valid value. For Python >= 3.4, 4 is a valid value.A negative value
-            for the protocol parameter is equivalent to setting its value to
-            HIGHEST_PROTOCOL.
+            valid value. For Python >= 3.4, 4 is a valid value. A negative
+            value for the protocol parameter is equivalent to setting its value
+            to HIGHEST_PROTOCOL.
 
             .. [1] https://docs.python.org/3/library/pickle.html
```
datapythonista: I think is more standard to have this in a `References` section.

pandas/io/pickle.py
```diff
@@ -33,7 +35,34 @@ def to_pickle(obj, path, compression='infer', protocol=pkl.HIGHEST_PROTOCOL):
...
+    >>> unpickled_df = pd.read_pickle("./dummy.pkl")
+    >>> unpickled_df
+       foo  bar
+    0    0    5
+    1    1    6
+    2    2    7
+    3    3    8
+    4    4    9
+
+    See Also
```
datapythonista: `See Also` should go before the `Examples`.

datapythonista (commented):

pandas/core/generic.py
minggli: done.

pandas/core/generic.py
minggli: added. :)

pandas/core/generic.py
minggli: thanks guys for looking into it.

pandas/core/generic.py
minggli: yes pythonista.

pandas/core/generic.py
minggli: done 👍

pandas/io/pickle.py
minggli: corrected.

pandas/core/generic.py
minggli: @jorisvandenbossche your opinion?

pandas/io/pickle.py
minggli: shortened.

pandas/io/pickle.py
jorisvandenbossche: Did you push your latest changes?

minggli: yes.

jorisvandenbossche: Just because you mentioned you corrected it, but the see also is still after the examples?

pandas/io/pickle.py
```diff
@@ -33,7 +35,33 @@ def to_pickle(obj, path, compression='infer', protocol=pkl.HIGHEST_PROTOCOL):
...
+    --------
+    read_pickle : Load pickled pandas object (or any object) from file.
+    DataFrame.to_hdf : Write the contained data to an HDF5 file using HDFStore.
```
jorisvandenbossche: I would make this a bit simpler (eg no need to mention HDFStore): "Write DataFame to HDF5 file"

pandas/io/pickle.py
```diff
@@ -33,7 +35,33 @@ def to_pickle(obj, path, compression='infer', protocol=pkl.HIGHEST_PROTOCOL):
...
+    DataFrame.to_sql : Write records stored in a DataFrame to a SQL database.
```
jorisvandenbossche: To be a bit more consistent in the wording here, maybe just "Write DataFrame to a SQL database" ?

pandas/io/pickle.py
minggli: oh sorry, just missed this one and did others.

pandas/io/pickle.py
```diff
@@ -72,6 +101,40 @@ def read_pickle(path, compression='infer'):
     Returns
     -------
     unpickled : type of object stored in file
+
+    See Also
+    --------
+    DataFrame.to_pickle : Pickle (serialize) DataFrame object to input file
+        path.
+    Series.to_pickle : Pickle (serialize) Series object to input file path.
+    read_hdf : read from the store, close it if we opened it.
```
jorisvandenbossche: I think the explanation about closing the file is too detailed for the See Also explanation. Maybe something like "Read HDF5 file into a DataFrame" ?

pandas/io/pickle.py
```diff
@@ -72,6 +101,40 @@ def read_pickle(path, compression='infer'):
...
+    read_sql : Read SQL query or database table into a DataFrame.
+    read_parquet : Load a parquet object from the file path, returning a
```
jorisvandenbossche: Similar comment here, "from the file path" is rather specific, would leave out that part

pandas/io/pickle.py
```diff
@@ -10,15 +10,17 @@
 
 def to_pickle(obj, path, compression='infer', protocol=pkl.HIGHEST_PROTOCOL):
     """
-    Pickle (serialize) object to input file path
+    Pickle (serialize) object to input file path.
```
jorisvandenbossche: I find it a bit strange to say we pickle the object the a "file path", it's actually to a file (located at the file path), so maybe simplify to "Pickle (serialize) object to file." ?

jorisvandenbossche (commented):
Added some remaining comment about the "See Also" descriptions. 
Looking good for the rest!

pandas/io/pickle.py
minggli: amended 👍

pandas/io/pickle.py
minggli: amended 👍

pandas/io/pickle.py
minggli: amended 👍

pandas/io/pickle.py
minggli: amended 👍

pandas/io/pickle.py
minggli: amended 👍

jreback:
this lgtm. do we use Reference in other doc-strings?

pandas/core/generic.py
jorisvandenbossche: I previously said it was fine, because we don't use `References` sections in many cases (most of the time we use inline links), another thing we can discuss in further improving the guidelines.

jorisvandenbossche (approved):

jorisvandenbossche:
@minggli Thanks a lot for the PR!

PR #20254, opened 2018-03-10
DOC: Improve the docstring of DataFrame.transpose()

igorcadelima:
an unrelated PR):

```

Errors found:
	Errors in parameters section
		Unknown parameters {'*args', '**kwargs'}
		Parameter "*args" has no type
```

The errors are well-known false positives.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -2315,22 +2315,49 @@ class DataFrame(NDFrame):
 
         Examples
         --------
-        >>> d = {'col1': [1, 2], 'col2': [3, 4]}
-        >>> df = pd.DataFrame(data=d)
-        >>> df
+        >>> d1 = {'col1': [1, 2], 'col2': [3, 4]}
+        >>> df1 = pd.DataFrame(data=d1)
+        >>> df1
            col1  col2
         0     1     3
         1     2     4
 
-        >>> df.T
+        >>> df1.T
               0  1
         col1  1  2
         col2  3  4
 
-        >>> df.transpose()
+        >>> df1.transpose()
               0  1
         col1  1  2
         col2  3  4
+
+        **Non-square DataFrame and mixed data types**
+
+        >>> d2 = {'name': ['Alice', 'Bob'],
+        ...       'score': [9.5, 8],
+        ...       'employed': [False, True],
+        ...       'kids': [0, 0]}
+
+        >>> df2 = pd.DataFrame(data=d2)
+        >>> df2
+            name  score  employed  kids
+        0  Alice    9.5     False     0
+        1    Bob    8.0      True     0
+
+        >>> df2.T
+                      0     1
+        name      Alice   Bob
+        score       9.5     8
+        employed  False  True
+        kids          0     0
+
+        >>> df2.transpose()
+                      0     1
+        name      Alice   Bob
+        score       9.5     8
+        employed  False  True
+        kids          0     0
         """
         nv.validate_transpose(args, dict())
         return super(DataFrame, self).transpose(1, 0, **kwargs)
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -2305,7 +2305,7 @@ class DataFrame(NDFrame):
 
         Returns
         -------
-        :class:`~pandas.DataFrame`
+        DataFrame
             The transposed DataFrame.
 
         See Also
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -2293,7 +2293,8 @@ class DataFrame(NDFrame):
         Transpose index and columns.
 
         Reflect the dataframe over its main diagonal by writing rows as columns
-        and vice-versa.
+        and vice-versa. The property :attr:`.T` is an accessor to the method
+        :meth:`transpose`.
 
         Parameters
         ----------
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -2289,7 +2289,48 @@ class DataFrame(NDFrame):
         return result
 
     def transpose(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


pandas/core/frame.py
```diff
@@ -1978,7 +1978,48 @@ def memory_usage(self, index=True, deep=False):
...
+
+        Returns
+        -------
+        DataFrame
+            The transposed DataFrame.
+
+        See Also
+        --------
+        numpy.transpose : Permute the dimensions of a given array.
+
```
math-and-data: Another recommendation to include in the *See Also* section:
DataFrame.T: alias for DataFrame.transpose

pandas/core/frame.py
```diff
@@ -1978,7 +1978,48 @@ def memory_usage(self, index=True, deep=False):
...
+        >>> df.T
+              0  1
+        col1  1  2
+        col2  3  4
+
+        >>> df.transpose()
+              0  1
+        col1  1  2
+        col2  3  4
+        """
```
math-and-data: Fantastic to see the usage of both T and transpose!
If possible, please add an example with a non-square DataFrame, e.g. 2 rows and 5 columns to show off some functionality - or maybe add some information about the fact it can be applied to square and non-square DataFrames?

math-and-data (changes requested):

pandas/core/frame.py
```diff
@@ -1978,7 +1978,48 @@ def memory_usage(self, index=True, deep=False):
...
+        Transpose index and columns.
+
+        Reflect the dataframe over its main diagonal by writing rows as columns
+        and vice-versa.
+
+        Parameters
+        ----------
+        *args
+            These arguments will be passed to
+            :func:`pandas.compat.numpy.function.validate_transpose`.
```
jreback: @jorisvandenbossche I think standard for this has changed? we don't need to mention the internal functions here

pandas/core/frame.py
```diff
@@ -1978,7 +1978,48 @@ def memory_usage(self, index=True, deep=False):
...
+        **kwargs
+            Keyword arguments used to create the transposed DataFrame.
+
+        Returns
```
jreback: can you also add an example with a mixed dtypes, and show the dtypes before/after. and add in the Notes section how transposing can change the dtypes.

jreback (changes requested):

pandas/core/frame.py
jorisvandenbossche: Yes, the explanation we are using in other PRs is "Additional keywords have no effect but might be accepted for compatibility with numpy". 

@igorcadelima and you can put both on a single line, like:

```
*args, **kwargs
    Additional keywords have no effect but might be accepted for
    compatibility with numpy
```

pandas/core/frame.py
jorisvandenbossche: That's a good idea, the only problem with that is that we use the same docstring for `.T`, so then it would be pointing to itself. 

So I would just reflect the full docstring for both, like you already did in the examples.

@igorcadelima maybe add in the extended summary that it is both as method transpose() and property T

pandas/core/frame.py
igorcadelima: Thanks for the comments! The reason I didn't just use "Additional keywords have no effect but might be accepted for compatibility with numpy" is because there will be noticeable effects if we pass 'copy=True`.

```
In [1]: import pandas as pd

In [2]: d = {'col1': [1, 2], 'col2': [3, 4]}

In [3]: df = pd.DataFrame(d)

In [4]: df
Out[4]: 
   col1  col2
0     1     3
1     2     4

In [5]: df2 = df.transpose() # or df.T

In [6]: df2
Out[6]: 
      0  1
col1  1  2
col2  3  4

In [7]: df3 = df.transpose(copy=True)

In [8]: df3
Out[8]: 
      0  1
col1  1  2
col2  3  4

In [9]: df['col1'] = 10

In [10]: df
Out[10]: 
   col1  col2
0    10     3
1    10     4

In [11]: df2
Out[11]: 
       0   1
col1  10  10
col2   3   4

In [12]: df3
Out[12]: 
      0  1
col1  1  2
col2  3  4
```

`copy` is the only keyword I have found to have any effect on the transpose operation. What do you think I should do in this case, @jorisvandenbossche @jreback ?

jorisvandenbossche: But copy is not a numpy keyword? https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.transpose.html 
So I think we should actually just document `copy` as an actual keyword (although it is not in the signature, for technical reasons, and so the validation script will complain about that, but you can ignore that)

igorcadelima: Not really. What happens is that when `copy=True`, the method [`copy`](https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.ndarray.copy.html#numpy.ndarray.copy) is called on the DataFrame's values, as can be confirmed [here](https://github.com/pandas-dev/pandas/blob/master/pandas/core/generic.py#L688).

I just incorporated your comments.

pandas/core/frame.py
igorcadelima: Done. How do you like it now? :)

pandas/core/frame.py
igorcadelima: Hi, @jreback . I've made some changes to incorporate your suggestion.

jorisvandenbossche (approved):
Very nice updates! Looking good

jorisvandenbossche:
(I only made a small clarification of the `copy` keyword, that even if `copy=False` it is not *always* no copy)

jorisvandenbossche:
@igorcadelima @cemsbr Thanks for the PR!

PR #20255, opened 2018-03-10
DOC: Improve the docstring of DataFrame.nlargest

cemsbr:
Co-authored-by: Igor C. A. de Lima <igorcadelima@gmail.com>
Signed-off-by: Carlos Eduardo Moreira dos Santos <cems@cemshost.com.br>
Signed-off-by: Igor C. A. de Lima <igorcadelima@gmail.com>

an unrelated PR):

```

Docstring for "pandas.core.frame.DataFrame.nlargest" correct. :)
```

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3836,17 +3836,21 @@ class DataFrame(NDFrame):
 
     def nlargest(self, n, columns, keep='first'):
         """
-        Return the `n` first rows ordered by `columns` in descending order.
+        Return the first `n` rows ordered by `columns` in descending order.
 
-        Return the `n` first rows with the largest values in `columns`, in
+        Return the first `n` rows with the largest values in `columns`, in
         descending order. The columns that are not specified are returned as
         well, but not used for ordering.
+
+        This method is equivalent to
+        ``df.sort_values(columns, ascending=False).head(n)``, but more
+        performant.
 
         Parameters
         ----------
         n : int
             Number of rows to return.
-        columns : iterable or single value
+        columns : label or list of labels
             Column label(s) to order by.
         keep : {'first', 'last'}, default 'first'
             Where there are duplicate values:
@@ -3857,13 +3861,15 @@ class DataFrame(NDFrame):
         Returns
         -------
         DataFrame
-            The `n` first rows ordered by the given columns in descending
+            The first `n` rows ordered by the given columns in descending
             order.
 
         See Also
         --------
-        DataFrame.nsmallest : Return the `n` first rows ordered by `columns` in
+        DataFrame.nsmallest : Return the first `n` rows ordered by `columns` in
             ascending order.
+        DataFrame.sort_values : Sort DataFrame by the values
+        DataFrame.head : Return the first `n` rows without re-ordering.
 
         Notes
         -----
@@ -3900,7 +3906,7 @@ class DataFrame(NDFrame):
         3  10  c  3.0
         1  10  b  2.0
         2   8  d  NaN
-        
+
         To order by the largest values in column "a" and then "c", we can
         specify multiple columns like in the next example.
 
@@ -3910,12 +3916,12 @@ class DataFrame(NDFrame):
         1  10  b  2.0
         2   8  d  NaN
 
-        The dtype of column "b" is `object` and attempting to get its largest
-        values raises a ``TypeError`` exception:
+        Attempting to use ``nlargest`` on non-numeric dtypes will raise a
+        ``TypeError``:
 
         >>> df.nlargest(3, 'b')
         Traceback (most recent call last):
-        TypeError: Column 'b' has dtype object, cannot use method 'nlargest' with this dtype
+        TypeError: Column 'b' has dtype object, cannot use method 'nlargest'
         """
         return algorithms.SelectNFrame(self,
                                        n=n,
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3836,33 +3836,40 @@ class DataFrame(NDFrame):
 
     def nlargest(self, n, columns, keep='first'):
         """
-        Return the `n` largest rows ordered by `columns`.
+        Return the `n` first rows ordered by `columns` in descending order.
 
-        Return the `n` largest rows of `columns` in descending order. The
-        remaining columns, although not used for ordering, are returned as
-        well.
+        Return the `n` first rows with the largest values in `columns`, in
+        descending order. The columns that are not specified are returned as
+        well, but not used for ordering.
 
         Parameters
         ----------
         n : int
-            Number of items to retrieve.
-        columns : list or str
-            Column name or names to retrieve values from.
+            Number of rows to return.
+        columns : iterable or single value
+            Column label(s) to order by.
         keep : {'first', 'last'}, default 'first'
             Where there are duplicate values:
-            - `first` : take the first occurrence;
-            - `last` : take the last occurrence.
+
+            - `first` : prioritize the first occurrence(s)
+            - `last` : prioritize the last occurrence(s)
 
         Returns
         -------
         DataFrame
-            The `n` largest rows in the DataFrame, ordered by the given columns
-            in descending order.
+            The `n` first rows ordered by the given columns in descending
+            order.
 
         See Also
         --------
-        DataFrame.nsmallest : Return the `n` smallest rows ordered by the given
-            columns.
+        DataFrame.nsmallest : Return the `n` first rows ordered by `columns` in
+            ascending order.
+
+        Notes
+        -----
+        This function cannot be used with all column types. For example, when
+        specifying columns with `object` or `category` dtypes, ``TypeError`` is
+        raised.
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


pandas/core/frame.py
```diff
@@ -3835,34 +3835,44 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
-            Number of items to retrieve
+            Number of items to retrieve.
         columns : list or str
-            Column name or names to order by
+            Column name or names to retrieve values from.
         keep : {'first', 'last'}, default 'first'
             Where there are duplicate values:
-            - ``first`` : take the first occurrence.
-            - ``last`` : take the last occurrence.
+            - `first` : take the first occurrence;
```
WillAyd: Shouldn't need punctuation at the end of either bullet point

pandas/core/frame.py
```diff
@@ -3835,34 +3835,44 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
+        --------
+        DataFrame.nsmallest : Return the `n` smallest rows sorted by given
+            columns.
 
         Examples
         --------
         >>> df = pd.DataFrame({'a': [1, 10, 8, 11, -1],
         ...                    'b': list('abdce'),
         ...                    'c': [1.0, 2.0, np.nan, 3.0, 4.0]})
         >>> df.nlargest(3, 'a')
```
WillAyd: Since the DataFrame constructor here is not trivial can we add a line to print the DataFrame as is, giving visual contrast to the examples?

pandas/core/frame.py
```diff
@@ -3835,34 +3835,44 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
         DataFrame
+            The `n` largest rows in the DataFrame, sorted by the given columns
+            in descending order.
+
+        See Also
+        --------
+        DataFrame.nsmallest : Return the `n` smallest rows sorted by given
+            columns.
 
         Examples
```
WillAyd: Can you add an example to differentiate between "first" and "last" `keep` parameters?

WillAyd (changes requested):

pandas/core/frame.py
```diff
@@ -3835,34 +3835,44 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
                                inplace=inplace, sort_remaining=sort_remaining)
 
     def nlargest(self, n, columns, keep='first'):
-        """Get the rows of a DataFrame sorted by the `n` largest
-        values of `columns`.
+        """
+        Return the `n` largest rows sorted by `columns`.
+
+        Sort the DataFrame by `columns` in descending order and return the top
```
jreback: we don't actually sort, can you re-word, we just pick the largest n values

jreback (changes requested):

pandas/core/frame.py
cemsbr: First, I removed them, but the validator complained about the missing full stop. Because of that, I kept this way. If I should remove them and ignore the validation error, let me know.

pandas/core/frame.py
cemsbr: Thanks for the review. Done.

pandas/core/frame.py
cemsbr: Nice suggestion. Done.

pandas/core/frame.py
cemsbr: Thanks for the review. I was not aware of the internals. Done.

pandas/core/frame.py
```diff
@@ -3835,34 +3835,59 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
 
         Parameters
         ----------
         n : int
-            Number of items to retrieve
+            Number of items to retrieve.
         columns : list or str
-            Column name or names to order by
+            Column name or names to retrieve values from.
         keep : {'first', 'last'}, default 'first'
```
WillAyd: Default value is implied for first element in the possible values, so no need to explicitly say `default 'first'` again

pandas/core/frame.py
WillAyd: Hmm OK now I know why I've seen this a few times. @datapythonista @jorisvandenbossche I would suggest that we ignore errors for a lack of punctuation at the end of bulleted lists - any objection to that?

pandas/core/frame.py
```diff
@@ -3835,34 +3835,59 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
             Where there are duplicate values:
-            - ``first`` : take the first occurrence.
-            - ``last`` : take the last occurrence.
+            - `first` : take the first occurrence;
```
WillAyd: After looking at your example do you think it makes more sense to word this as "prioritize the first occurrence(s)"? I assume it's more or less doing that as opposed to a full take right? 

Put in other words, if `n` is 2 and you have three duplicate values it wouldn't return all three but rather prioritize the 2 that come first, right?

pandas/core/frame.py
```diff
@@ -3835,34 +3835,59 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
-            a  b   c
-        3  11  c   3
-        1  10  b   2
-        2   8  d NaN
+            a  b    c
+        1  10  b  2.0
+        3  10  c  3.0
+        2   8  d  NaN
+
+        >>> df.nlargest(3, 'a', keep='last')
```
WillAyd: Great thanks! I would just add a short blurb to direct the users attention to what's important here and in the example above. So maybe before the previous example say "In the following example, we will use `nlargest` to select the three rows having the largest values in column 'a'". Then for the next example something like "When using `keep='last'` ties are resolved in reverse order". Doesn't need to be exactly those words

pandas/core/frame.py
```diff
@@ -3835,34 +3835,59 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
+            a  b    c
+        3  10  c  3.0
+        1  10  b  2.0
+        2   8  d  NaN
```
WillAyd: Can you add one more example selecting multiple columns?

WillAyd (changes requested):

pandas/core/frame.py
```diff
@@ -3835,34 +3835,59 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
                                inplace=inplace, sort_remaining=sort_remaining)
 
     def nlargest(self, n, columns, keep='first'):
-        """Get the rows of a DataFrame sorted by the `n` largest
-        values of `columns`.
+        """
+        Return the `n` largest rows ordered by `columns`.
+
+        Return the `n` largest rows of `columns` in descending order. The
```
jorisvandenbossche: I find "largest rows" a strange wording, as a row itself cannot be "large", only the values in it. Here I would use "n first rows" based on sorting 'columns' in descending order

pandas/core/frame.py
```diff
@@ -3835,34 +3835,59 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
+        well.
 
         Parameters
         ----------
         n : int
-            Number of items to retrieve
+            Number of items to retrieve.
         columns : list or str
-            Column name or names to order by
+            Column name or names to retrieve values from.
```
jorisvandenbossche: I think "to order by" is more correct, as otherwise it seems that you only return the values from those columns

pandas/core/frame.py
jorisvandenbossche: @WillAyd I think we changed the guide on this (some time last week) .. :) 
For explicitness, I think it is good to keep the " default 'first' ". A new reader cannot know the rule of "default is fist value in the set of options"

pandas/core/frame.py
jorisvandenbossche: Yes, we can certainly ignore errors in such cases of bullet points

cemsbr:
As suggested, the changes result in the expected validation error below.

```

Errors found:
        Errors in parameters section
                Parameter "keep" description should finish with "."
```

pandas/core/frame.py
```diff
@@ -3836,33 +3836,40 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
+
+            - `first` : prioritize the first occurrence(s)
+            - `last` : prioritize the last occurrence(s)
 
         Returns
         -------
         DataFrame
-            The `n` largest rows in the DataFrame, ordered by the given columns
-            in descending order.
+            The `n` first rows ordered by the given columns in descending
```
WillAyd: Minor stylistic comment but "first `n`" sounds more natural than "`n` first"

pandas/core/frame.py
```diff
@@ -3877,17 +3884,38 @@ def nlargest(self, n, columns, keep='first'):
...
 
+        When using ``keep='last'``, ties are resolved in reverse order:
+
         >>> df.nlargest(3, 'a', keep='last')
             a  b    c
         3  10  c  3.0
         1  10  b  2.0
         2   8  d  NaN
+        
+        To order by the largest values in column "a" and then "c", we can
```
WillAyd: Great example. I would also add something to Notes touching on this behavior - something like "When `columns` is iterable the order of the iterable's elements dictates the order in which columns in the DataFrame are prioritized"

pandas/core/frame.py
```diff
@@ -3877,17 +3884,38 @@ def nlargest(self, n, columns, keep='first'):
...
+        specify multiple columns like in the next example.
+
+        >>> df.nlargest(3, ['a', 'c'])
+            a  b    c
+        3  10  c  3.0
+        1  10  b  2.0
+        2   8  d  NaN
+
+        The dtype of column "b" is `object` and attempting to get its largest
```
WillAyd: Just to generalize could instead say "Attempting to use ``nlargest`` on an non-numeric dtypes will raise a ``TypeError``"

pandas/core/frame.py
```diff
@@ -3836,33 +3836,40 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
+        well, but not used for ordering.
 
         Parameters
         ----------
         n : int
-            Number of items to retrieve.
-        columns : list or str
-            Column name or names to retrieve values from.
+            Number of rows to return.
+        columns : iterable or single value
```
WillAyd: str or iterable

WillAyd (changes requested):
Looking good - couple more minor things

pandas/core/frame.py
```diff
@@ -3835,34 +3835,87 @@ def sortlevel(self, level=0, axis=0, ascending=True, inplace=False,
...
 
     def nlargest(self, n, columns, keep='first'):
-        """Get the rows of a DataFrame sorted by the `n` largest
-        values of `columns`.
+        """
+        Return the `n` first rows ordered by `columns` in descending order.
+
+        Return the `n` first rows with the largest values in `columns`, in
+        descending order. The columns that are not specified are returned as
+        well, but not used for ordering.
```
datapythonista: If I'm right, this is equivalent to using `df.sort_values(columns, ascending=False).head(n)`, isn't it? If that's the case, I'd explicitly say it in the extended summary, and I'd add both methods to the `See Also` section.

jreback: it’s an equivalent result, but this is much more performany

jorisvandenbossche (approved):
Edited for the last minor updates.

jorisvandenbossche:
@cemsbr Thanks a lot :+1:

Issue #19505, opened 2018-02-02
Documation: copy documenation is lacking detail for when deep=False


charlie0389:
The DataFrame.copy documentation (for pandas 0.22.0) states:

```
deep : boolean or string, default True
Make a deep copy, including a copy of the data and the indices. 
With deep=False neither the indices or the data are copied.
Note that when deep=True data is copied, actual python objects 
will not be copied recursively, only the reference to the object. 
This is in contrast to copy.deepcopy in the Standard Library,
which recursively copies object data.
```
#### Problem description

The documentation does not state what *is* copied when ``deep=False`` (it doesn't copy data, it doesn't copy indices, what does it copy then?).

Searching for 'copy' suggests this has not been reported as an issue.

ZhuBaohe:
The reference of   the data is copied with deep=False.

```python
df=pd.DataFrame([1,2,3])

df1=df.copy(deep=False)

df1.iat[0,0]=100

df1.index=[10,20,30]

df1
Out[188]: 
      0
10  100
20    2
30    3

df
Out[189]: 
     0
0  100
1    2
2    3
```

MIsmailKhan:
Hi, I'd be happy to take this up! 

Would this suffice?

"
deep : boolean or string, default True
Make a deep copy, including a copy of the data and the indices. **With ``deep=False``(i.e. shallow copy), the data reference is copied.** Neither the indices or data are copied.
Note that when deep=True data is copied, actual python objects 
will not be copied recursively, only the reference to the object. 
This is in contrast to copy.deepcopy in the Standard Library,
which recursively copies object data.
"

charlie0389:
Ok, I'm a little confused. What does a copy of the reference to the object mean? Isn't this the same operation as tagging an object  with a variable? i.e. 
Isn't
```
df2 = df1
```
the same as
```
df2 = df1.copy(deep=False)
```
?
Obviously if I was right this would be a little redundant, so I assume I'm not. But I think the documentation could outline with an example of when  ``deep=False`` is useful.

MIsmailKhan:
I think this should cover the question: 
https://stackoverflow.com/questions/46327494/python-pandas-dataframe-copydeep-false-vs-copydeep-true-vs

As for whether df2 = df1 and df2 =df1.copy(deep=False) are the same. I wrote up a couple of lines and checked what happened to the above if I modified the original database(i.e. df2). They both yielded the same output so I'm guessing that they are the same.

[...]

Issue #19862, opened 2018-02-23
DataFrame.copy(deep=True) is not a deep copy of the index


skaae:
#### Code Sample, a copy-pastable example if possible

```python
df1 = pd.DataFrame(index=['a', 'b'], columns=['foo', 'muu'])
df1.index.name = "foo"
print(df1)

# create deep copy of df1 and change a value in the index
df2 = df1.copy(deep=True)
df2.index.name = "bar"
df2.index.values[0] = 'c'  # changes both df1 and df2

print(df1)
print(df2)

```
#### Problem description

DataFrame.copy(deep=True) is not a deep copy of the index.

In  

https://github.com/pandas-dev/pandas/blob/a00154dcfe5057cb3fd86653172e74b6893e337d/pandas/core/indexes/base.py#L787

maybe deep should be set to True?

#### Expected Output

```
     foo  muu
foo          
a    NaN  NaN
b    NaN  NaN
     foo  muu
foo          
c    NaN  NaN
b    NaN  NaN
     foo  muu
bar          
c    NaN  NaN
b    NaN  NaN
```
#### Output of ``pd.show_versions()``

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.0-53-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.21.0
pytest: 3.2.1
pip: 9.0.1
setuptools: 36.5.0.post20170921
Cython: 0.26.1
numpy: 1.13.1
scipy: 0.19.1
pyarrow: 0.8.0
xarray: 0.9.6
IPython: 6.1.0
sphinx: 1.6.3
patsy: 0.4.1
dateutil: 2.6.1
pytz: 2017.2
blosc: None
bottleneck: 1.2.1
tables: 3.4.2
numexpr: 2.6.2
feather: None
matplotlib: 2.0.2
openpyxl: 2.4.8
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 0.9.8
lxml: 3.8.0
bs4: 4.6.0
html5lib: 0.999999999
sqlalchemy: 1.1.13
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: 0.5.0

</details>

[...]

PR #20261, opened 2018-03-10
DOC: Update pandas.Series.copy docstring

thismakessand:
Closes https://github.com/pandas-dev/pandas/issues/19505

- Add extended summary
- Add examples

an unrelated PR):

```

######################## Docstring (pandas.Series.copy) ########################

Make a copy of this objects indices and data.

Deep copy (default) will create a new object with a copy of the objects
data and indices. Modifications to the data or indices of the copy
will not be reflected in the original object.

Shallow copy (``deep=False``) will create a new object without copying
the data or indices. Any changes to the index or data of the original
will be reflected in the shallow copy (and vice versa).

Parameters
----------
deep : boolean or string, default True
    Make a deep copy, including a copy of the data and the indices.
    With ``deep=False`` neither the indices or the data are copied.

    Note that when ``deep=True`` data is copied, actual python objects
    will not be copied recursively, only the reference to the object.
    This is in contrast to ``copy.deepcopy`` in the Standard Library,
    which recursively copies object data.

Returns
-------
copy : type of caller

Examples
--------
>>> s = pd.Series([1,2], dtype="int32", index=["a", "b"])
>>> s
a    1
b    2
dtype: int32
>>> s_copy = s.copy()
>>> s_copy
a    1
b    2
dtype: int32

Shallow copy versus default (deep) copy:

In a shallow copy the index and data is a reference to the original
objects'.

>>> s = pd.Series([1,2], dtype="int32", index=["a", "b"])
>>> deep_copy = s.copy()
>>> shallow_copy = s.copy(deep=False)
>>> id(s) == id(shallow_copy)
False
>>> id(s.index) == id(shallow_copy.index)
True
>>> id(s.values) == id(shallow_copy.values)
True
>>> id(s) == id(deep_copy)
False
>>> id(s.index) == id(deep_copy.index)
False
>>> id(s.values) == id(deep_copy.values)
False
>>> s[0] = 3
>>> s
a    3
b    2
dtype: int32
>>> shallow_copy
a    3
b    2
dtype: int32
>>> deep_copy
a    1
b    2
dtype: int32
>>> shallow_copy[0] = 4
>>> s
a    4
b    2
dtype: int32
>>> shallow_copy
a    4
b    2
dtype: int32
>>> deep_copy
a    1
b    2
dtype: int32

When copying object containing python objects, deep copy will
copy the indices and data but will not do so recursively.

>>> class Point(object):
...     def __init__(self, x, y):
...         self.x = x
...         self.y = y
...     def __repr__(self):
...         return "Point(%d,%d)" % (self.x, self.y)
...
>>> p1 = Point(0, 1)
>>> s = pd.Series([p1], dtype="object")
>>> s
0    Point(0,1)
dtype: object
>>> s_copy = s.copy()
>>> s_copy
0    Point(0,1)
dtype: object
>>> id(s[0]) == id(s_copy[0])
True
>>> p1.x = 1
>>> s
0    Point(1,1)
dtype: object
>>> s_copy
0    Point(1,1)
dtype: object
>>>

For deep-copying python objects, the following can be used:

>>> import copy
>>> deep_deep_copy = pd.Series(copy.deepcopy(s.values),
...                            index=copy.deepcopy(s.index))

Errors found:
        See Also section not found

```

explicitly.

Not sure if anything should be added to the See Also section

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4508,21 +4508,21 @@ class NDFrame(PandasObject, SelectionMixin):
         """
         Make a copy of this object's indices and data.
 
-        When `deep=True` (default), a new object will be created with a
+        When ``deep=True`` (default), a new object will be created with a
         copy of the calling object's data and indices. Modifications to
         the data or indices of the copy will not be reflected in the
         original object (see notes below).
 
-        When `deep=False`, a new object will be created without copying
+        When ``deep=False``, a new object will be created without copying
         the calling object's data or index (only references to the data
         and index are copied). Any changes to the data of the original
         will be reflected in the shallow copy (and vice versa).
 
         Parameters
         ----------
-        deep : boolean or str, default `True`
+        deep : bool, default True
             Make a deep copy, including a copy of the data and the indices.
-            With `deep=False` neither the indices nor the data are copied.
+            With ``deep=False`` neither the indices nor the data are copied.
 
         Returns
         -------
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4516,8 +4516,7 @@ class NDFrame(PandasObject, SelectionMixin):
         When `deep=False`, a new object will be created without copying
         the calling object's data or index (only references to the data
         and index are copied). Any changes to the data of the original
-        will be reflected in the shallow copy (and vice versa). Changes
-        to the index will result in a new index as indices are immutable.
+        will be reflected in the shallow copy (and vice versa).
 
         Parameters
         ----------
@@ -4532,10 +4531,15 @@ class NDFrame(PandasObject, SelectionMixin):
 
         Notes
         -----
-        When `deep=True`, data is copied but actual Python objects
+        When ``deep=True``, data is copied but actual Python objects
         will not be copied recursively, only the reference to the object.
         This is in contrast to `copy.deepcopy` in the Standard Library,
         which recursively copies object data (see examples below).
+
+        While ``Index`` objects are copied when ``deep=True``, the underlying
+        numpy array is not copied for performance reasons. Since ``Index`` is
+        immutable, the underlying data can be safely shared and a copy
+        is not needed.
 
         Examples
         --------
@@ -4589,9 +4593,9 @@ class NDFrame(PandasObject, SelectionMixin):
         b    2
         dtype: int64
 
-        When copying an object containing Python objects, a deep copy will
-        copy the data, but will not do so recursively. Updating a nested data
-        object will be reflected in the deep copy.
+        Note that when copying an object containing Python objects, a deep copy
+        will copy the data, but will not do so recursively. Updating a nested
+        data object will be reflected in the deep copy.
 
         >>> s = pd.Series([[1, 2], [3, 4]])
         >>> deep = s.copy()
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4506,7 +4506,15 @@ class NDFrame(PandasObject, SelectionMixin):
 
     def copy(self, deep=True):
         """
-        Make a copy of this objects data.
+        Make a copy of this objects indices and data.
+
+        Deep copy (default) will create a new object with a copy of the objects
+        data and indices. Modifications to the data or indices of the copy
+        will not be reflected in the original object.
+
+        Shallow copy (``deep=False``) will create a new object without copying
+        the data or indices. Any changes to the index or data of the original
+        will be reflected in the shallow copy (and vice versa).
 
         Parameters
         ----------
@@ -4522,6 +4530,102 @@ class NDFrame(PandasObject, SelectionMixin):
         Returns
         -------
         copy : type of caller
+
+        Examples
+        --------
+        >>> s = pd.Series([1,2], dtype="int32", index=["a", "b"])
+        >>> s
+        a    1
+        b    2
+        dtype: int32
+        >>> s_copy = s.copy()
+        >>> s_copy
+        a    1
+        b    2
+        dtype: int32
+
+        Shallow copy versus default (deep) copy:
+
+        In a shallow copy the index and data is a reference to the original
+        objects'.
+
+        >>> s = pd.Series([1,2], dtype="int32", index=["a", "b"])
+        >>> deep_copy = s.copy()
+        >>> shallow_copy = s.copy(deep=False)
+        >>> id(s) == id(shallow_copy)
+        False
+        >>> id(s.index) == id(shallow_copy.index)
+        True
+        >>> id(s.values) == id(shallow_copy.values)
+        True
+        >>> id(s) == id(deep_copy)
+        False
+        >>> id(s.index) == id(deep_copy.index)
+        False
+        >>> id(s.values) == id(deep_copy.values)
+        False
+        >>> s[0] = 3
+        >>> s
+        a    3
+        b    2
+        dtype: int32
+        >>> shallow_copy
+        a    3
+        b    2
+        dtype: int32
+        >>> deep_copy
+        a    1
+        b    2
+        dtype: int32
+        >>> shallow_copy[0] = 4
+        >>> s
+        a    4
+        b    2
+        dtype: int32
+        >>> shallow_copy
+        a    4
+        b    2
+        dtype: int32
+        >>> deep_copy
+        a    1
+        b    2
+        dtype: int32
+
+        When copying object containing python objects, deep copy will
+        copy the indices and data but will not do so recursively.
+
+        >>> class Point(object):
+        ...     def __init__(self, x, y):
+        ...         self.x = x
+        ...         self.y = y
+        ...     def __repr__(self):
+        ...         return "Point(%d,%d)" % (self.x, self.y)
+        ...
+        >>> p1 = Point(0, 1)
+        >>> s = pd.Series([p1], dtype="object")
+        >>> s
+        0    Point(0,1)
+        dtype: object
+        >>> s_copy = s.copy()
+        >>> s_copy
+        0    Point(0,1)
+        dtype: object
+        >>> id(s[0]) == id(s_copy[0])
+        True
+        >>> p1.x = 1
+        >>> s
+        0    Point(1,1)
+        dtype: object
+        >>> s_copy
+        0    Point(1,1)
+        dtype: object
+        >>>
+
+        For deep-copying python objects, the following can be used:
+
+        >>> import copy
+        >>> deep_deep_copy = pd.Series(copy.deepcopy(s.values),
+        ...                            index=copy.deepcopy(s.index))
         """
         data = self._data.copy(deep=deep)
         return self._constructor(data).__finalize__(self)
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4514,26 +4514,28 @@ class NDFrame(PandasObject, SelectionMixin):
         original object (see notes below).
 
         When `deep=False`, a new object will be created without copying
-        the calling object's data (only a reference to the data is
-        copied). Any changes to the data of the original will be reflected
-        in the shallow copy (and vice versa).
+        the calling object's data or index (only references to the data
+        and index are copied). Any changes to the data of the original
+        will be reflected in the shallow copy (and vice versa). Changes
+        to the index will result in a new index as indices are immutable.
 
         Parameters
         ----------
-        deep : boolean or string, default True
+        deep : boolean or str, default `True`
             Make a deep copy, including a copy of the data and the indices.
-            With `deep=False` neither the indices or the data are copied.
+            With `deep=False` neither the indices nor the data are copied.
 
         Returns
         -------
         copy : type of caller
+            Object type matches caller.
 
         Notes
         -----
-        When `deep=True`, data is copied but actual python objects
+        When `deep=True`, data is copied but actual Python objects
         will not be copied recursively, only the reference to the object.
         This is in contrast to `copy.deepcopy` in the Standard Library,
-        which recursively copies object data.  (See examples below).
+        which recursively copies object data (see examples below).
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -4522,6 +4530,102 @@ def copy(self, deep=True):
...
+        >>> shallow_copy
+        a    4
+        b    2
+        dtype: int32
+        >>> deep_copy
+        a    1
+        b    2
+        dtype: int32
+
+        When copying object containing python objects, deep copy will
```
WillAyd: "an object"

pandas/core/generic.py
```diff
@@ -4506,7 +4506,15 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
 
     def copy(self, deep=True):
         """
-        Make a copy of this objects data.
+        Make a copy of this objects indices and data.
```
WillAyd: For the sake of explicitness can you change "this objects" to "the calling object's" or "the caller's"

pandas/core/generic.py
```diff
@@ -4506,7 +4506,15 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
+
+        Deep copy (default) will create a new object with a copy of the objects
```
WillAyd: Given these start to dive into the parameters and implementations I think they are better served in a dedicated Notes section

pandas/core/generic.py
```diff
@@ -4506,7 +4506,15 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
+        data and indices. Modifications to the data or indices of the copy
+        will not be reflected in the original object.
+
+        Shallow copy (``deep=False``) will create a new object without copying
```
WillAyd: Single back ticks

pandas/core/generic.py
WillAyd: object's

pandas/core/generic.py
```diff
@@ -4522,6 +4530,102 @@ def copy(self, deep=True):
         Returns
         -------
         copy : type of caller
+
+        Examples
+        --------
+        >>> s = pd.Series([1,2], dtype="int32", index=["a", "b"])
```
WillAyd: space after comma

pandas/core/generic.py
```diff
@@ -4522,6 +4530,102 @@ def copy(self, deep=True):
...
+        >>> s
```
WillAyd: This is simple enough that I don't think you need to print

pandas/core/generic.py
```diff
@@ -4522,6 +4530,102 @@ def copy(self, deep=True):
...
+        a    1
+        b    2
+        dtype: int32
+        >>> s_copy = s.copy()
+        >>> s_copy
+        a    1
+        b    2
+        dtype: int32
+
+        Shallow copy versus default (deep) copy:
```
WillAyd: Since this is a "section header" can you get it to render in bold (i.e. between asterisks)?

pandas/core/generic.py
```diff
@@ -4522,6 +4530,102 @@ def copy(self, deep=True):
...
+
+        In a shallow copy the index and data is a reference to the original
```
WillAyd: Maybe just "the index and data are shared with the original object"

pandas/core/generic.py
WillAyd: Is the `dtype` argument provided because you want it to be clear that type is copied over and not automatically upcast to int64? I think that's what you are going for but it wasn't immediately obvious to me, so I'd try and make it more explicit if that is your intention

pandas/core/generic.py
```diff
@@ -4522,6 +4530,102 @@ def copy(self, deep=True):
...
+        objects'.
+
+        >>> s = pd.Series([1,2], dtype="int32", index=["a", "b"])
+        >>> deep_copy = s.copy()
+        >>> shallow_copy = s.copy(deep=False)
+        >>> id(s) == id(shallow_copy)
```
WillAyd: These examples are certainly comprehensive but perhaps too verbose. Can you refactor to make the index / values comparisons fit on one line? Similar comment with the value assignment - any consolidation will make it more readable

pandas/core/generic.py
```diff
@@ -4522,6 +4530,102 @@ def copy(self, deep=True):
...
+        dtype: int32
+        >>> deep_copy
+        a    1
+        b    2
+        dtype: int32
+
+        When copying object containing python objects, deep copy will
+        copy the indices and data but will not do so recursively.
+
+        >>> class Point(object):
```
WillAyd: Is it required to put a class in this docstring? Again for the sake of being concise wonder if you can just use lists to illustrate

WillAyd (changes requested):

pandas/core/generic.py
thismakessand: Thanks!  I made this change plus the other requested changes.

thismakessand:
I removed the wording/examples about the index not being copied when deep=False because while it's not copied when .copy() is called, if the index is changed on the shallow copy it does not affect the index of the original (in contrast with the data values, which are also not copied, but changes to values do affect the original).

pandas/core/generic.py
```diff
@@ -4506,120 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
-        the data or indices. Any changes to the index or data of the original
-        will be reflected in the shallow copy (and vice versa).
+        When `deep=False`, a new object will be created without copying
+        the calling object's data (only a reference to the data is
+        copied). Any changes to the data of the original will be reflected
+        in the shallow copy (and vice versa).
 
         Parameters
         ----------
         deep : boolean or string, default True
```
WillAyd: \`True\` instead of just True

WillAyd: str instead of string

pandas/core/generic.py
```diff
@@ -4506,120 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
-
-            Note that when ``deep=True`` data is copied, actual python objects
-            will not be copied recursively, only the reference to the object.
-            This is in contrast to ``copy.deepcopy`` in the Standard Library,
-            which recursively copies object data.
+            With `deep=False` neither the indices or the data are copied.
 
         Returns
         -------
         copy : type of caller
```
WillAyd: Move "type of caller" down as the description and say something like "Object type matches caller."

pandas/core/generic.py
```diff
@@ -4506,120 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
+        dtype: int64
 
         Shallow copy versus default (deep) copy:
 
-        In a shallow copy the index and data is a reference to the original
-        objects'.
+        In a shallow copy, the data is shared with the original object.
 
-        >>> s = pd.Series([1,2], dtype="int32", index=["a", "b"])
+        >>> s = pd.Series([1,2], index=["a", "b"])
```
WillAyd: Space after first comma

pandas/core/generic.py
```diff
@@ -4506,120 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
+        dtype: int64
         >>> deep_copy
         a    1
         b    2
-        dtype: int32
+        dtype: int64
 
-        When copying object containing python objects, deep copy will
-        copy the indices and data but will not do so recursively.
+        When copying an object containing python objects, deep copy will
```
WillAyd: Capitalize Python and maybe say "a deep copy"

pandas/core/generic.py
```diff
@@ -4506,120 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
 
-        >>> s = pd.Series([1,2], dtype="int32", index=["a", "b"])
+        >>> s = pd.Series([1,2], index=["a", "b"])
         >>> deep_copy = s.copy()
         >>> shallow_copy = s.copy(deep=False)
         >>> id(s) == id(shallow_copy)
         False
-        >>> id(s.index) == id(shallow_copy.index)
-        True
         >>> id(s.values) == id(shallow_copy.values)
```
WillAyd: Just again wondering if these can be more concise. Perhaps choosing a different variable name will allow you to do `id(s.values) == id(shlw.values) and id(s.index) == id(shlw.values)`

pandas/core/generic.py
```diff
@@ -4506,120 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
         Returns
         -------
         copy : type of caller
 
+        Notes
+        -----
+        When `deep=True`, data is copied but actual python objects
+        will not be copied recursively, only the reference to the object.
+        This is in contrast to `copy.deepcopy` in the Standard Library,
+        which recursively copies object data.  (See examples below).
```
WillAyd: "...copies object data (see examples below)." - reduces some punctuation and saves space

pandas/core/generic.py
```diff
@@ -4506,120 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
         ----------
         deep : boolean or string, default True
             Make a deep copy, including a copy of the data and the indices.
-            With ``deep=False`` neither the indices or the data are copied.
-
-            Note that when ``deep=True`` data is copied, actual python objects
-            will not be copied recursively, only the reference to the object.
-            This is in contrast to ``copy.deepcopy`` in the Standard Library,
-            which recursively copies object data.
+            With `deep=False` neither the indices or the data are copied.
```
WillAyd: "neither...nor" instead of "or"

pandas/core/generic.py
```diff
@@ -4506,120 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
-        Make a copy of this objects indices and data.
+        Make a copy of this object's indices and data.
 
-        Deep copy (default) will create a new object with a copy of the objects
-        data and indices. Modifications to the data or indices of the copy
-        will not be reflected in the original object.
+        When `deep=True` (default), a new object will be created with a
+        copy of the calling object's data and indices. Modifications to
+        the data or indices of the copy will not be reflected in the
+        original object (see notes below).
```
WillAyd: Unless you want to have a Notes section I would say "see Examples below" as a reference to the section actually showing that

pandas/core/generic.py
```diff
@@ -4506,120 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
         a    3
         b    2
-        dtype: int32
+        dtype: int64
         >>> deep_copy
         a    1
         b    2
-        dtype: int32
+        dtype: int64
         >>> shallow_copy[0] = 4
```
WillAyd: Instead of having two different assignments I think you can have `s[0] = 3` and then this line both before printing any of the objects. Gets you the same result and covers the same concepts in less lines

WillAyd (changes requested):

WillAyd:
@thismakessand not sure I fully understand why you removed the mention of indices. Perhaps its related to the conversation in https://github.com/pandas-dev/pandas/issues/19862? Referencing that I think it's preferable to use `is` for comparing the indices instead of using the equality operator. Perhaps that will help clarify how the copy is actually working behind the scenes.

Lot's of edits but they are mostly minor - thanks and keep up the great work!

pandas/core/generic.py
```diff
@@ -4506,22 +4506,110 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
     def copy(self, deep=True):
         """
-        Make a copy of this objects data.
+        Make a copy of this object's indices and data.
+
+        When `deep=True` (default), a new object will be created with a
+        copy of the calling object's data and indices. Modifications to
+        the data or indices of the copy will not be reflected in the
+        original object (see notes below).
+
```
jreback: indices are immutable, mention this

pandas/core/generic.py
```diff
@@ -4506,22 +4506,110 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
+        which recursively copies object data.  (See examples below).
+
+        Examples
+        --------
+        >>> s = pd.Series([1, 2], index=["a", "b"])
+        >>> s
+        a    1
+        b    2
+        dtype: int64
+        >>> s_copy = s.copy()
```
jreback: blank line between cases (if they are separate / distinct)

pandas/core/generic.py
```diff
@@ -4506,22 +4506,110 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
+        >>> shallow_copy = s.copy(deep=False)
+        >>> id(s) == id(shallow_copy)
+        False
+        >>> id(s.values) == id(shallow_copy.values)
+        True
+        >>> id(s) == id(deep_copy)
+        False
+        >>> id(s.values) == id(deep_copy.values)
+        False
+        >>> s[0] = 3
```
jreback: break this up with some text describing what you are showing

pandas/core/generic.py
```diff
@@ -4506,22 +4506,110 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
+        >>> deep_copy
+        a    1
+        b    2
+        dtype: int64
+
+        When copying an object containing python objects, deep copy will
+        copy the data, but will not do so recursively.
+
+        >>> s = pd.Series([[1, 2], [3, 4]])
+        >>> s_copy = s.copy()
```
jreback: show this with a deep copy

pandas/core/generic.py
```diff
@@ -4506,22 +4506,110 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
+        >>> s
+        0    [10, 2]
+        1     [3, 4]
+        dtype: object
+        >>> s_copy
+        0    [10, 2]
+        1     [3, 4]
+        dtype: object
+
+        For deep-copying python objects, the following can be used:
```
jreback: this last example is not necessary

jreback (changes requested):

pandas/core/generic.py
thismakessand: I wasn't sure how to address this comment - do you mean to show the behavior of 
` .copy(deep=True)` versus the std python `copy.deepcopy()`?   I was trying to show an example of how a `(deep=True)` copy isn't necessarily a deep copy in the standard python sense, but maybe I should remove this example entirely since it's already mentioned as a caveat in the Notes section?

thismakessand:
Thanks - I added more clarification regarding the index and made the examples simpler and hopefully clearer.

pandas/core/generic.py
```diff
@@ -4506,22 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
-
-            Note that when ``deep=True`` data is copied, actual python objects
-            will not be copied recursively, only the reference to the object.
-            This is in contrast to ``copy.deepcopy`` in the Standard Library,
-            which recursively copies object data.
+            With `deep=False` neither the indices nor the data are copied.
 
         Returns
         -------
         copy : type of caller
```
WillAyd: Still need to clean up the line that says `copy : type of caller`. I would explicitly list the objects that could be a caller of this function, which I assume would make this `Series or DataFrame` but if there's anything else be sure to add that.

Otherwise nice job on all the edits - changes I requested all lgtm

pandas/core/generic.py
```diff
@@ -4506,22 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
+            Object type matches caller.
+
+        Notes
+        -----
+        When `deep=True`, data is copied but actual Python objects
```
jorisvandenbossche: double backticks around `deep=True`

pandas/core/generic.py
jorisvandenbossche: Yes, I would make this "Series or DataFrame"

pandas/core/generic.py
```diff
@@ -4506,22 +4506,104 @@ def astype(self, dtype, copy=True, errors='raise', **kwargs):
...
+        When `deep=True` (default), a new object will be created with a
+        copy of the calling object's data and indices. Modifications to
+        the data or indices of the copy will not be reflected in the
+        original object (see notes below).
+
+        When `deep=False`, a new object will be created without copying
+        the calling object's data or index (only references to the data
+        and index are copied). Any changes to the data of the original
+        will be reflected in the shallow copy (and vice versa). Changes
+        to the index will result in a new index as indices are immutable.
```
jorisvandenbossche: I think this line "Changes to the index will result in a new index as indices are immutable" is a bit out of place here, as this is not related to `deep=False`

For the fact that indices are immutable, I would add a short explanation about that in the Notes section (it is a bit too detailed for the extended summary at the beginning of the docstring). 
The main thing is that, even for `deep=True`, the index *data* (the underlying numpy array) is not copied, only the object (the Index class holding the array data). This is done for performance reasons: because Index is immutable, it is not needed to copy the data.
(see https://github.com/pandas-dev/pandas/issues/19862 for discussion on it)

jorisvandenbossche (commented):
Very nice docstring! 
Added few more comments

jorisvandenbossche:
@thismakessand Do you have time to update this PR based on the remaining feedback ?

thismakessand:
@jorisvandenbossche Yes, it's definitely still on my radar, hoping to find some time later today or tomorrow.

thismakessand:
I read through the linked issues and added the details you mentioned.  Looking at the linked issues and other stuff on stack overflow, etc it seems like most of the confusion is around 
1. deep copy not being a "true" deep copy when the series/dataframe contains python objects and 
2. the underlying data of the Index isn't copied even when deep=True. 

I think both of these are covered now.  Originally I had more details but  I was worried going too into the caveats would just be unnecessarily confusing so I cut a bit out.  I can add more if it's still unclear.

jorisvandenbossche (approved):
Updates are looking good!

jorisvandenbossche:
@thismakessand Thanks a lot!

jorisvandenbossche:
There is one remaining thing: the type description for `deep` also said 'str'. I removed this for now, as it is not explained, but apparently `'all'` is accepted to also copy the Index underlying data. 
But will open a new issue for that to see if we actually want to expose this publicly.

PR #20263, opened 2018-03-11
DOC: update the window.Rolling.min docstring

gojandrooo:
Checklist for the pandas documentation sprint (ignore this if you are doing
an unrelated PR):

- [ ] PR title is "DOC: update the <your-function-or-method> docstring"
- [ ] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [ ] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [ ] It has been proofread on language by another sprint participant

Please include the output of the validation script below between the "```" ticks:

```
# paste output of "scripts/validate_docstrings.py <your-function-or-method>" here
# between the "```" (remove this comment, but keep the "```")

```

If the validation script still gives errors, but you think there is a good reason
to deviate in this case (and there are certainly such cases), please state this
explicitly.

Checklist for other PRs (remove this part if you are doing a PR for the pandas documentation sprint):

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] whatsnew entry

PR #20264, opened 2018-03-11
DOC" update the Pandas core window rolling count docstring"

tommy-stone:
an unrelated PR):

```
################# Docstring (pandas.core.window.Rolling.count) #################

The rolling sum if it is a non-Nan value inside the window.

Returns
-------
Returns the object type with the sum of the window values
in the current row.

See Also
--------

Examples
--------
>>> s = pd.Series([2, 3, np.nan, 'values'])
>>> s.rolling(2).count()
0    1.0
1    2.0
2    1.0
3    1.0
dtype: float64
>>> s.rolling(3).count()
0    1.0
1    2.0
2    2.0
3    2.0
dtype: float64
>>> s.rolling(4).count()
0    1.0
1    2.0
2    2.0
3    3.0
dtype: float64

Errors found:
	No extended summary found

```

explicitly.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/window.py
+++ b/pandas/core/window.py
@@ -781,17 +781,19 @@ class _Rolling(_Window):
 class _Rolling_and_Expanding(_Rolling):
 
     _shared_docs['count'] = dedent(r"""
-    The %(name)s count of any non-Nan observations inside the window.
+    The %(name)s count of any non-NaN observations inside the window.
 
     Returns
     -------
-    Returned object type is determined by the caller of the %(name)s
-    calculation.
+    Series or DataFrame
+        Returned object type is determined by the caller of the %(name)s
+        calculation.
 
     See Also
     --------
     pandas.Series.%(name)s : Calling object with Series data
     pandas.DataFrame.%(name)s : Calling object with DataFrames
+    pandas.DataFrame.count : Count of the full DataFrame
 
     Examples
     --------
@@ -1475,7 +1477,6 @@ class Expanding(_Rolling_and_Expanding):
     agg = aggregate
 
     @Substitution(name='expanding')
-    @Appender(_doc_template)
     @Appender(_shared_docs['count'])
     def count(self, **kwargs):
         return super(Expanding, self).count(**kwargs)
--- a/pandas/core/window.py
+++ b/pandas/core/window.py
@@ -781,17 +781,17 @@ class _Rolling(_Window):
 class _Rolling_and_Expanding(_Rolling):
 
     _shared_docs['count'] = dedent(r"""
-    The %(name)'s count of any non-Nan values inside the window.  
+    The %(name)s count of any non-Nan observations inside the window.
 
     Returns
     -------
-    Series or Dataframe
-        Returned object type is determined by the caller of %(name)
+    Returned object type is determined by the caller of the %(name)s
+    calculation.
 
     See Also
     --------
-    pandas.Series.%(name)
-    pandas.DataFrame.%(name)
+    pandas.Series.%(name)s : Calling object with Series data
+    pandas.DataFrame.%(name)s : Calling object with DataFrames
 
     Examples
     --------
--- a/pandas/core/window.py
+++ b/pandas/core/window.py
@@ -781,21 +781,21 @@ class _Rolling(_Window):
 class _Rolling_and_Expanding(_Rolling):
 
     _shared_docs['count'] = dedent(r"""
-    The %(name)s sum if it is a non-Nan value inside the window.
+    The %(name)'s count of any non-Nan values inside the window.  
 
     Returns
     -------
-    Returns the object type with the sum of the window values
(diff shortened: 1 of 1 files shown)


pandas/core/window.py
```diff
@@ -780,8 +780,41 @@ def calc(x):
 
 class _Rolling_and_Expanding(_Rolling):
 
-    _shared_docs['count'] = """%(name)s count of number of non-NaN
-    observations inside provided window."""
+    _shared_docs['count'] = dedent(r"""
+    The %(name)s sum if it is a non-Nan value inside the window.
```
WillAyd: Using the word `sum` is confusing here for a `count`. Can you reword to the effect of "count of non-`NaN` observations..."?

pandas/core/window.py
```diff
@@ -780,8 +780,41 @@ def calc(x):
...
+
+    Returns
+    -------
+    Returns the object type with the sum of the window values
+    in the current row.
+
+    See Also
+    --------
+    Series.%(name)s : Calling object with Series data
+    DataFrame.%(name)s : Calling object with DataFrames
```
WillAyd: Can you link to `Series` / `DataFrame` sum methods?

pandas/core/window.py
```diff
@@ -780,8 +780,41 @@ def calc(x):
...
 class _Rolling_and_Expanding(_Rolling):
 
-    _shared_docs['count'] = """%(name)s count of number of non-NaN
-    observations inside provided window."""
+    _shared_docs['count'] = dedent(r"""
+    The %(name)s sum if it is a non-Nan value inside the window.
+
+    Returns
+    -------
+    Returns the object type with the sum of the window values
```
WillAyd: Can you make note of what the return type is? Also similar comment as above - referring to `sum` within a `count` function is confusing. I'm not sure `current row` is the right concept here, though I understand what you are trying to get at

WillAyd (changes requested):

pandas/core/window.py
jorisvandenbossche: You can take a look at how the other PRs did this, see eg https://github.com/pandas-dev/pandas/pull/19999/files

pandas/core/window.py
```diff
@@ -780,8 +780,41 @@ def calc(x):
...
+    in the current row.
+
+    See Also
+    --------
+    Series.%(name)s : Calling object with Series data
+    DataFrame.%(name)s : Calling object with DataFrames
+
+    Examples
+    --------
+    >>> s = pd.Series([2, 3, np.nan, 'values'])
```
jorisvandenbossche: It's a bit strange to give an example with mixed floats / strings (I mean, in general that's not really good practice to have data like that). I think only using floats (with nans) is fine.

pandas/core/window.py
jorisvandenbossche: You can use a similar wording as here: https://github.com/pandas-dev/pandas/pull/20263/files

jorisvandenbossche:
@scriptomation do you have time to update this based on the feedback?

scriptomation:
@jorisvandenbossche Thank you for the feed back.  Updating it this weekend and will resubmit.

pandas/core/window.py
scriptomation: "count of non-Nan observations inside the rolling window" would work?

pandas/core/window.py
scriptomation: Sounds good, I will update that information.

pandas/core/window.py
jorisvandenbossche: Yes

pandas/core/window.py
scriptomation: That makes sense.  How does this work?

Returns
    -------
    Series or Dataframe
        Returned object type is determined by the caller of %(name)

pandas/core/window.py
scriptomation: I wanted to show an example of mixed values being counted, in case a mistake is made by having different types of data.  I see what you are saying though since it's already stated in the description.  I will update to only use floats and nans.  Thanks!

pandas/core/window.py
WillAyd: I would just copy verbatim, so `Returned object type is determined by the caller of the %(name)s calculation`

pandas/core/window.py
WillAyd: @scriptomation be sure that you actually surround your reference to `NaN` in back ticks, so literally you would write non-\`NaN\` (capitalization matters too)

jorisvandenbossche (approved):
Thanks for the update!

(added a small fixup to also have this working with `expanding`)

jorisvandenbossche:
Thanks for the PR!

PR #20267, opened 2018-03-11
DOC: Improve the docstring of pandas.DataFrame.append()

Triple0:
an unrelated PR):

```
##################### Docstring (pandas.DataFrame.append)  #####################

Append rows of `other` to the end of `caller`, returning a new object.

Columns not in other are added as new columns.

Parameters
----------
other : DataFrame or Series/dict-like object, or list of these
    The data to append.
ignore_index : boolean, default False
    If True, do not use the index labels.
verify_integrity : boolean, default False
    If True, raise ValueError on creating index with duplicates.

Returns
-------
appended : DataFrame

Notes
-----
If a list of dict/series is passed and the keys are all contained in
the DataFrame's index, the order of the columns in the resulting
DataFrame will be unchanged.

Iteratively appending rows to a DataFrame can be more computationally
intensive than a single concatenate. A better solution is to append
those rows to a list and then concatenate the list with the original
DataFrame all at once.

See also
--------
pandas.concat : General function to concatenate DataFrame, Series
    or Panel objects

Examples
--------

>>> df = pd.DataFrame([[1, 2], [3, 4]], columns=list('AB'))
>>> df
   A  B
0  1  2
1  3  4
>>> df2 = pd.DataFrame([[5, 6], [7, 8]], columns=list('AB'))
>>> df.append(df2)
   A  B
0  1  2
1  3  4
0  5  6
1  7  8

With `ignore_index` set to True:

>>> df.append(df2, ignore_index=True)
   A  B
0  1  2
1  3  4
2  5  6
3  7  8

The following, while not recommended methods for generating DataFrames,
show two ways to generate a DataFrame from multiple data sources.

Less efficient:

>>> df = pd.DataFrame(columns=['A'])
>>> for i in range(5):
...     df = df.append({'A': i}, ignore_index=True)
>>> df
   A
0  0
1  1
2  2
3  3
4  4

More efficient:

>>> pd.concat([pd.DataFrame([i], columns=['A']) for i in range(5)],
...           ignore_index=True)
   A
0  0
1  1
2  2
3  3
4  4

```

explicitly.


pandas/core/frame.py
```diff
@@ -5054,7 +5054,9 @@ def infer(x):
     def append(self, other, ignore_index=False, verify_integrity=False):
         """
         Append rows of `other` to the end of this frame, returning a new
```
WillAyd: This needs to fit on a single line - can you consolidate?

pandas/core/frame.py
```diff
@@ -5054,7 +5054,9 @@ def infer(x):
...
-        object. Columns not in this frame are added as new columns.
+        object.
+
+        Columns not in this frame are added as new columns.
```
WillAyd: Instead of "this frame" can you say "the caller"? Also while conceptually simple the usage of the word "Columns" mixes reference to the calling object and the object being appended, so differentiating between those would be preferable

WillAyd (changes requested):

pandas/core/frame.py
```diff
@@ -5053,8 +5053,9 @@ def infer(x):
 
     def append(self, other, ignore_index=False, verify_integrity=False):
         """
-        Append rows of `other` to the end of this frame, returning a new
-        object. Columns not in this frame are added as new columns.
+        Append rows of `other` to the end of `caller`, returning a new object.
+
+        Columns not in the caller are added as new columns.
```
WillAyd: Better but for explicitness say "Columns in `other` not in `caller` are ..."

pandas/core/frame.py
```diff
@@ -5053,8 +5053,9 @@ def infer(x):
 
     def append(self, other, ignore_index=False, verify_integrity=False):
         """
-        Append rows of `other` to the end of this frame, returning a new
-        object. Columns not in this frame are added as new columns.
+        Append rows of `other` to the end of `caller`, returning a new object.
```
jorisvandenbossche: 'caller' is not needed to quote (it's not an argument). 

Also given this is the docstring specific to DataFrame (not shared with Series), I think we can be more specific. So maybe "calling DataFrame" ? (or is it then too long ?)

pandas/core/frame.py
```diff
@@ -5053,8 +5053,9 @@ def infer(x):
...
+
+        Columns not in other are added as new columns.
```
jorisvandenbossche: It's not the columns in `other`, but in "calling DataFrame" (or more explicit: ```Column in `other` that are not in the calling DataFrame are added as new columns```

jorisvandenbossche (commented):
Thanks for the PR!

Other comments:

- In the Notes: " A better solution is to append those rows to a list and then concatenate the list with the original DataFrame all at once." -> can you add `` .. using :func:`pandas.concat` `` at the end of that sentence ?

- for the Parameter section: 'boolean' -> 'bool'

pandas/core/frame.py
```diff
@@ -143,14 +143,16 @@
 ----------%s
 right : DataFrame
 how : {'left', 'right', 'outer', 'inner'}, default 'inner'
+    How to handle the operation of the two objects.
+
     * left: use only keys from left frame, similar to a SQL left outer join;
-      preserve key order
+      preserve key order.
```
WillAyd: Shouldn't need periods at the end of bullet points

pandas/core/frame.py
```diff
@@ -140,9 +140,11 @@
 columns, the index will be passed on.
 
 Parameters
-----------%s
-right : DataFrame
+----------
+right : DataFrame or Series/dict-like object
```
WillAyd: Since `Series` and `dict` are distinct types I'd separate these out as `DataFrame`, `Series` or `dict`. Update for append as well

pandas/core/frame.py
```diff
@@ -166,18 +168,18 @@
...
     Sort the join keys lexicographically in the result DataFrame. If False,
-    the order of the join keys depends on the join type (how keyword)
+    the order of the join keys depends on the join type (how keyword).
 suffixes : 2-length sequence (tuple, list, ...)
     Suffix to apply to overlapping column names in the left and right
-    side, respectively
+    side, respectively.
 copy : boolean, default True
-    If False, do not copy data unnecessarily
+    If False, do not copy data unnecessarily.
```
WillAyd: Generally put built-ins in backticks, so \`False\` wherever applicable

pandas/core/frame.py
```diff
@@ -199,7 +201,7 @@
       dataset.
     * "many_to_many" or "m:m": allowed, but does not result in checks.
 
-    .. versionadded:: 0.21.0
+    .. versionadded:: 0.21.0.
```
WillAyd: Don't put a period at the end of the `versionadded` directive

WillAyd (changes requested):
Couple minor edits. You can also clean up the `Returns` section where applicable to simply show that this returns a DataFrame

pandas/core/frame.py
```diff
@@ -2719,9 +2730,13 @@ def assign(self, **kwargs):
         or modified columns. All items are computed first, and then assigned
         in alphabetical order.
 
-        .. versionchanged :: 0.23.0
+        .. versionchanged :: 0.23.0.
 
-            Keyword argument order is maintained for Python 3.6 and later.
+        Keyword argument order is maintained for Python 3.6 and later.
```
jorisvandenbossche: The indentation was needed for having it part of the versionchanged directive

pandas/core/frame.py
```diff
@@ -2719,9 +2730,13 @@ def assign(self, **kwargs):
...
+
+        See Also
+        --------
+        DataFrame.assign: For column(s)-on-DataFrame operations
```
jorisvandenbossche: This is the docstring of DataFrame.assign I think? So then not needed to refer to itself

jorisvandenbossche (commented):
@Triple0 Can you try to limit the changes to a single docstring? 
(I don't know to what extens other people are working on the merge / assign docstrings. It's fine to keep the changes you have in those, but please focus for the rest on doing edits to the `append` docstring)

PR #20268, opened 2018-03-11
DOC: add documentation to core.window.corr

theandygross:
an unrelated PR):

```
################# Docstring (pandas.core.window.Rolling.corr)  #################

Calculate rolling correlation.

This function uses Pearson's definition of correlation.

Parameters
----------
other : Series, DataFrame, or ndarray, optional
    If not supplied then will default to self.
pairwise : bool, default None
    Calculate pairwise combinations of columns within a
    DataFrame. If other is not specified, defaults to True,
    otherwise defaults to False. Not relevant for Series.
    See notes.
**kwargs
    Under Review.

Returns
-------
Series or DataFrame
    Returned object type is determined by the caller of the
    rolling calculation.

See Also
--------
Series.corr : Equivalent method for Series
DataFrame.corr : Equivalent method for DataFrame
rolling.cov : Similar method to calculate covariance
numpy.corrcoef : NumPy Pearson's correlation calculation

Notes
-----
Other should be always be specified, except for DataFrame inputs with
pairwise set to `True`. All other input combinations will return all 1's.

Function will return `NaN`s for correlations of equal valued sequences;
this is the result of a 0/0 division error.

When pairwise is set to `False`, only matching columns between self and
other will be used.

When pairwise is set to `True`, the output will be a MultiIndex DataFrame
with the original index on the first level, and the "other" DataFrame
columns on the second level.

In the case of missing elements, only complete pairwise observations
will be used.

Examples
--------
The below example shows a rolling calculation with a window size of
four matching the equivalent function call using `numpy.corrcoef`.

>>> v1 = [3, 3, 3, 5, 8]
>>> v2 = [3, 4, 4, 4, 8]
>>> fmt = "{0:.6f}"  # limit the printed precision to 6 digits
>>> import numpy as np
>>> # numpy returns a 2X2 array, the correlation coefficient
>>> # is the number at entry [0][1]
>>> print(fmt.format(np.corrcoef(v1[:-1], v2[:-1])[0][1]))
0.333333
>>> print(fmt.format(np.corrcoef(v1[1:], v2[1:])[0][1]))
0.916949
>>> s1 = pd.Series(v1)
>>> s2 = pd.Series(v2)
>>> s1.rolling(4).corr(s2)
0         NaN
1         NaN
2         NaN
3    0.333333
4    0.916949
dtype: float64

The below example shows a similar rolling calculation on a
DataFrame using the pairwise option.

>>> matrix = np.array([[51., 35.], [49., 30.], [47., 32.],    [46., 31.], [50., 36.]])
>>> print(np.corrcoef(matrix[:-1,0], matrix[:-1,1]).round(7))
[[1.         0.6263001]
 [0.6263001  1.       ]]
>>> print(np.corrcoef(matrix[1:,0], matrix[1:,1]).round(7))
[[1.         0.5553681]
 [0.5553681  1.        ]]
>>> df = pd.DataFrame(matrix, columns=['X','Y'])
>>> df
      X     Y
0  51.0  35.0
1  49.0  30.0
2  47.0  32.0
3  46.0  31.0
4  50.0  36.0
>>> df.rolling(4).corr(pairwise=True)
            X         Y
0 X       NaN       NaN
  Y       NaN       NaN
1 X       NaN       NaN
  Y       NaN       NaN
2 X       NaN       NaN
  Y       NaN       NaN
3 X  1.000000  0.626300
  Y  0.626300  1.000000
4 X  1.000000  0.555368
  Y  0.555368  1.000000

Errors found:
	Errors in parameters section
```

explicitly.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


pandas/core/window.py
```diff
@@ -1028,19 +1028,112 @@ def _get_cov(X, Y):
...
+        If not supplied then will default to self.
     pairwise : bool, default None
-        If False then only matching columns between self and other will be
-        used and the output will be a DataFrame.
-        If True then all pairwise combinations will be calculated and the
-        output will be a MultiIndex DataFrame in the case of DataFrame inputs.
-        In the case of missing elements, only complete pairwise observations
-        will be used.""")
+        Calculate pairwise combinations of columns within a
+        DataFrame. If other is not specified, defaults to True,
```
WillAyd: Put back ticks around parameters and built ins, so \`other\`, \`True\`, and \`False\` here.

pandas/core/window.py
```diff
@@ -1028,19 +1028,112 @@ def _get_cov(X, Y):
...
+        otherwise defaults to False. Not relevant for Series.
```
WillAyd: :class:\`~pandas.Series\`

pandas/core/window.py
```diff
@@ -1028,19 +1028,112 @@ def _get_cov(X, Y):
...
+    Series.%(name)s : Calling object with Series data
+    DataFrame.%(name)s : Calling object with DataFrames
+    Series.corr : Equivalent method for Series
+    DataFrame.corr : Equivalent method for DataFrame
+    %(name)s.cov : Similar method to calculate covariance
+    numpy.corrcoef : NumPy Pearson's correlation calculation
+
+    Notes
+    -----
+    Other should be always be specified, except for DataFrame inputs with
```
WillAyd: \`other\` and :class:~\`pandas.DataFrame\`

pandas/core/window.py
```diff
@@ -1028,19 +1028,112 @@ def _get_cov(X, Y):
...
+    pairwise set to `True`. All other input combinations will return all 1's.
+
+    Function will return `NaN`s for correlations of equal valued sequences;
+    this is the result of a 0/0 division error.
+
+    When pairwise is set to `False`, only matching columns between self and
```
WillAyd: \`self\` and \`other\`

WillAyd (changes requested):

pandas/core/window.py
```diff
@@ -1028,19 +1028,112 @@ def _get_cov(X, Y):
                                    _get_cov, pairwise=bool(pairwise))
 
     _shared_docs['corr'] = dedent("""
-    %(name)s sample correlation
+    Calculate %(name)s correlation.
+
+    This function uses Pearson's definition of correlation.
```
dukebody: Can you add a link to Wikipedia or similar here?

pandas/core/window.py
```diff
@@ -1028,19 +1028,112 @@ def _get_cov(X, Y):
...
+    DataFrame.%(name)s : Calling object with DataFrames
+    Series.corr : Equivalent method for Series
+    DataFrame.corr : Equivalent method for DataFrame
+    %(name)s.cov : Similar method to calculate covariance
+    numpy.corrcoef : NumPy Pearson's correlation calculation
+
+    Notes
+    -----
+    Other should be always be specified, except for DataFrame inputs with
+    pairwise set to `True`. All other input combinations will return all 1's.
```
dukebody: "pairwise" in single back-ticks

pandas/core/window.py
```diff
@@ -1028,19 +1028,112 @@ def _get_cov(X, Y):
...
+
+    Function will return `NaN`s for correlations of equal valued sequences;
```
dukebody: I don't understand this. If the sequences are equally valued, like in the case of non specifying `other` and `pairwise=False`, the correlation of each column with itself should be all 1's. Am I wrong?

pandas/core/window.py
```diff
@@ -1028,19 +1028,112 @@ def _get_cov(X, Y):
...
+    this is the result of a 0/0 division error.
+
+    When pairwise is set to `False`, only matching columns between self and
+    other will be used.
+
+    When pairwise is set to `True`, the output will be a MultiIndex DataFrame
+    with the original index on the first level, and the "other" DataFrame
```
dukebody: other in single back-ticks I believe

pandas/core/window.py
```diff
@@ -1028,19 +1028,112 @@ def _get_cov(X, Y):
...
+    columns on the second level.
+
+    In the case of missing elements, only complete pairwise observations
```
dukebody: I understand the correlation of "non-complete" elements will be set to `NaN`? Can we write this in the explanation if so?

pandas/core/window.py
```diff
@@ -1028,19 +1028,112 @@ def _get_cov(X, Y):
...
+
+    Examples
+    --------
+    The below example shows a rolling calculation with a window size of
+    four matching the equivalent function call using `numpy.corrcoef`.
+
+    >>> v1 = [3, 3, 3, 5, 8]
+    >>> v2 = [3, 4, 4, 4, 8]
+    >>> fmt = "{0:.6f}"  # limit the printed precision to 6 digits
+    >>> import numpy as np
```
dukebody: You don't need to import numpy, it's automatically imported for each docstring, see https://python-sprints.github.io/pandas/guide/pandas_docstring.html#conventions-for-the-examples

pandas/core/window.py
theandygross: This is true, but trivial. I'm rewording for clarity.

pandas/core/window.py
```diff
@@ -1028,19 +1028,113 @@ def _get_cov(X, Y):
                                    _get_cov, pairwise=bool(pairwise))
 
     _shared_docs['corr'] = dedent("""
-    %(name)s sample correlation
+    Calculate %(name)s correlation.
+
+    This function uses Pearson's definition of correlation
```
jreback: can you move to Notes

pandas/core/window.py
```diff
@@ -1028,19 +1028,113 @@ def _get_cov(X, Y):
...
-        used and the output will be a DataFrame.
-        If True then all pairwise combinations will be calculated and the
-        output will be a MultiIndex DataFrame in the case of DataFrame inputs.
-        In the case of missing elements, only complete pairwise observations
-        will be used.""")
+        Calculate pairwise combinations of columns within a
+        DataFrame. If `other` is not specified, defaults to `True`,
+        otherwise defaults to `False`. Not relevant for :class:`~pandas.Series`.
+        See notes.
+    **kwargs
```
jreback: IIRC we remove this. @TomAugspurger

jreback (changes requested):

pandas/core/window.py
TomAugspurger: This line is fine.

For the explanation, you can put "For compatibility with other %(name)s methods. Not used."

WillAyd:
Thanks @theandygross !

PR #20269, opened 2018-03-11
DOC: update NDFrame.squeeze docstring

villasv:
## Series.squeeze and DataFrame.squeeze are inherited from NDFrame 

```
##################### Docstring (pandas.DataFrame.squeeze) #####################

Squeeze 1 dimensional axis objects into scalars.

If the given axis consists of one dimensional objects, they are turned
into scalars. In case no axis is specified, all axes are subject to
squeezing. In any case, objects in axes that can't be squeezed are
left unchanged.

Parameters
----------
axis : integer or string, optional
    A specific axis to squeeze.

    .. versionadded:: 0.20.0

Returns
-------
DataFrame, Series or scalar
    The projection after squeezing axis.

See Also
--------
Series.iloc : Integer-location based indexing for selecting scalars
DataFrame.iloc : Integer-location based indexing for selecting Series

Examples
--------
>>> primes = pd.Series([2, 3, 5, 7])

Slicing might produce a Series with a single value:

>>> even_primes = primes[primes % 2 == 0]
>>> even_primes
0    2
dtype: int64
>>> even_primes.squeeze()
2

Squeezing objects with more than 1 dimension does nothing:

>>> odd_primes = primes[primes % 2 == 1]
>>> odd_primes
1    3
2    5
3    7
dtype: int64
>>> odd_primes.squeeze()
1    3
2    5
3    7
dtype: int64

Squeezing is even more effective when used with DataFrames.

>>> df = pd.DataFrame([[1,2], [3, 4]], columns=['a', 'b'])
>>> df
   a  b
0  1  2
1  3  4

Slicing a single column will produce a DataFrame with one of the
axis having only 1 dimension:

>>> df_a = df[['a']]
>>> df_a
   a
0  1
1  3

Objects along the column are 1 dimensional, so they can be squeezed
into scalars:

>>> df_a.squeeze('columns')
0    1
1    3
Name: a, dtype: int64

Slicing a single row from a single column will produce a single
scalar DataFrame:

>>> df_0a = df[['a']].iloc[[0]]
>>> df_0a
   a
0  1

Squeezing along the rows produces a single scalar Series:

>>> df_0a.squeeze('rows')
a    1
Name: 0, dtype: int64

Squeezing all axes wil project directly into a scalar:

>>> df_0a.squeeze()
1

Errors found:
        Errors in parameters section
                Parameter "axis" description should finish with "."
```

> If the validation script still gives errors, but you think there is a good reason
explicitly.

The `.. versionadded` rst macro already adds the dot, so I can't just put another one after it.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -699,18 +699,100 @@ class NDFrame(PandasObject, SelectionMixin):
 
     def squeeze(self, axis=None):
         """
-        Squeeze length 1 dimensions.
+        Squeeze 1 dimensional axis objects into scalars.
+
+        If the given axis consists of one dimensional objects, they are turned
+        into scalars. In case no axis is specified, all axes are subject to
+        squeezing. In any case, objects in axes that can't be squeezed are
+        left unchanged.
 
         Parameters
         ----------
-        axis : None, integer or string axis name, optional
-            The axis to squeeze if 1-sized.
+        axis : integer or string, optional
+            A specific axis to squeeze.
 
             .. versionadded:: 0.20.0
 
         Returns
         -------
-        scalar if 1-sized, else original object
+        DataFrame, Series or scalar
+            The projection after squeezing axis.
+
+        See Also
+        --------
+        Series.iloc : Integer-location based indexing for selecting scalars
+        DataFrame.iloc : Integer-location based indexing for selecting Series
+
+        Examples
+        --------
+        >>> primes = pd.Series([2, 3, 5, 7])
+
+        Slicing might produce a Series with a single value:
+
+        >>> even_primes = primes[primes % 2 == 0]
+        >>> even_primes
+        0    2
+        dtype: int64
+        >>> even_primes.squeeze()
+        2
+
+        Squeezing objects with more than 1 dimension does nothing:
+
+        >>> odd_primes = primes[primes % 2 == 1]
+        >>> odd_primes
+        1    3
+        2    5
+        3    7
+        dtype: int64
+        >>> odd_primes.squeeze()
+        1    3
+        2    5
+        3    7
+        dtype: int64
+
+        Squeezing is even more effective when used with DataFrames.
+
+        >>> df = pd.DataFrame([[1,2], [3, 4]], columns=['a', 'b'])
+        >>> df
+           a  b
+        0  1  2
+        1  3  4
+
+        Slicing a single column will produce a DataFrame with one of the
+        axis having only 1 dimension:
+
+        >>> df_a = df[['a']]
+        >>> df_a
+           a
+        0  1
+        1  3
+
+        Objects along the column are 1 dimensional, so they can be squeezed
+        into scalars:
+
+        >>> df_a.squeeze('columns')
+        0    1
+        1    3
+        Name: a, dtype: int64
+
+        Slicing a single row from a single column will produce a single
+        scalar DataFrame:
+
+        >>> df_0a = df[['a']].iloc[[0]]
+        >>> df_0a
+           a
+        0  1
+
+        Squeezing along the rows produces a single scalar Series:
+
+        >>> df_0a.squeeze('rows')
+        a    1
+        Name: 0, dtype: int64
+
+        Squeezing all axes wil project directly into a scalar:
+
+        >>> df_0a.squeeze()
+        1
         """
         axis = (self._AXIS_NAMES if axis is None else
                 (self._get_axis_number(axis),))
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -701,15 +701,20 @@ class NDFrame(PandasObject, SelectionMixin):
         """
         Squeeze 1 dimensional axis objects into scalars.
 
-        If the given axis consists of one dimensional objects, they are turned
-        into scalars. In case no axis is specified, all axes are subject to
-        squeezing. In any case, objects in axes that can't be squeezed are
-        left unchanged.
+        Series or DataFrames with a single element are squeezed to a scalar.
+        DataFrames with a single column or a single row are squeezed to a
+        Series. Otherwise the object is unchanged.
+
+        This method is most useful when you don't know if your
+        object is a Series or DataFrame, but you do know it has just a single
+        column. In that case you can safely call `squeeze` to ensure you have a
+        Series.
 
         Parameters
         ----------
         axis : integer or string, optional
-            A specific axis to squeeze.
+            A specific axis to squeeze. By default, all length-1 axes are
+            squeezed.
 
             .. versionadded:: 0.20.0
 
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -712,7 +712,7 @@ class NDFrame(PandasObject, SelectionMixin):
 
         Parameters
         ----------
-        axis : integer or string, optional
+        axis : axis : {0 or ‘index’, 1 or ‘columns’, None}, default None
             A specific axis to squeeze. By default, all length-1 axes are
             squeezed.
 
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -699,18 +699,100 @@ def pop(self, item):
...
+        dtype: int64
+        >>> odd_primes.squeeze()
+        1    3
+        2    5
+        3    7
+        dtype: int64
+
+        Squeezing is even more effective when used with DataFrames.
+
+        >>> df = pd.DataFrame([[1,2], [3, 4]], columns=['a', 'b'])
```
WillAyd: Minor edit - need space after first comma

pandas/core/generic.py
```diff
@@ -699,18 +699,100 @@ def pop(self, item):
...
+        >>> df
+           a  b
+        0  1  2
+        1  3  4
+
+        Slicing a single column will produce a DataFrame with one of the
+        axis having only 1 dimension:
```
WillAyd: axes instead of axis

pandas/core/generic.py
```diff
@@ -699,18 +699,100 @@ def pop(self, item):
...
+        If the given axis consists of one dimensional objects, they are turned
+        into scalars. In case no axis is specified, all axes are subject to
+        squeezing. In any case, objects in axes that can't be squeezed are
+        left unchanged.
 
         Parameters
         ----------
-        axis : None, integer or string axis name, optional
-            The axis to squeeze if 1-sized.
+        axis : integer or string, optional
```
WillAyd: Convention here is `axis : {0 or ‘index’, 1 or ‘columns’, None}, default None`

pandas/core/generic.py
```diff
@@ -699,18 +699,100 @@ def pop(self, item):
...
+        >>> df_a = df[['a']]
+        >>> df_a
+           a
+        0  1
+        1  3
+
+        Objects along the column are 1 dimensional, so they can be squeezed
+        into scalars:
+
+        >>> df_a.squeeze('columns')
```
WillAyd: I think the first example is great but this one could use a little work. With how it's presented it wouldn't make sense to use `squeeze` instead of just selecting the desired `Series`. Perhaps using a predicate here of say `'a' == 1` and squeezing the result of that would showcase the utility better?

pandas/core/generic.py
```diff
@@ -699,18 +699,100 @@ def pop(self, item):
...
+        0    1
+        1    3
+        Name: a, dtype: int64
+
+        Slicing a single row from a single column will produce a single
+        scalar DataFrame:
+
+        >>> df_0a = df[['a']].iloc[[0]]
```
WillAyd: Stylistically I think using a predicate than selecting a set of columns would be clearer, so something like `df[df.index < 1][['a']]`

WillAyd (changes requested):

pandas/core/generic.py
villasv: Makes sense, in most cases squeeze will be necessary because of indirect slicing rather than specific axis selections.

pandas/core/generic.py
villasv: Yep, basically the same point as the one before. Will work on it.

pandas/core/generic.py
```diff
@@ -699,18 +699,100 @@ def pop(self, item):
 
     def squeeze(self, axis=None):
         """
-        Squeeze length 1 dimensions.
+        Squeeze 1 dimensional axis objects into scalars.
+
+        If the given axis consists of one dimensional objects, they are turned
+        into scalars. In case no axis is specified, all axes are subject to
+        squeezing. In any case, objects in axes that can't be squeezed are
+        left unchanged.
```
datapythonista: I think it's a good description, but personally I find that being a bit less technical would make it easier to understand. For example "If applied in a 1x1 DataFrame, it returns the contained value. In a DataFrame with one column, it converts it to a Series...". Feel free to disagree, but IMO something like this is a bit faster to understand.

pandas/core/generic.py
```diff
@@ -699,18 +699,100 @@ def pop(self, item):
...
-            The axis to squeeze if 1-sized.
+        axis : integer or string, optional
+            A specific axis to squeeze.
 
             .. versionadded:: 0.20.0
 
         Returns
         -------
-        scalar if 1-sized, else original object
+        DataFrame, Series or scalar
```
datapythonista: Not sure if scalar is what we use when it can return anything contained in the values of the dataframe. It's ok with me, just pointing out in case someone knows of another terminology.

pandas/core/generic.py
```diff
@@ -699,18 +699,100 @@ def pop(self, item):
...
+            The projection after squeezing axis.
+
+        See Also
+        --------
+        Series.iloc : Integer-location based indexing for selecting scalars
+        DataFrame.iloc : Integer-location based indexing for selecting Series
```
datapythonista: Does it make sense to add `DataFrame.to_series`? I think they do the same in the case of 1 column DataFrame, right?

pandas/core/generic.py
```diff
@@ -699,18 +699,100 @@ def pop(self, item):
...
+        Slicing might produce a Series with a single value:
+
+        >>> even_primes = primes[primes % 2 == 0]
+        >>> even_primes
+        0    2
+        dtype: int64
+        >>> even_primes.squeeze()
+        2
+
+        Squeezing objects with more than 1 dimension does nothing:
```
datapythonista: technically speaking "squeezing `Series` with more than 1 dimension does nothing", squeezing a dataframe with 1D in one of the axis does something.

pandas/core/generic.py
```diff
@@ -699,18 +699,100 @@ def pop(self, item):
...
+        Squeezing along the rows produces a single scalar Series:
+
+        >>> df_0a.squeeze('rows')
+        a    1
+        Name: 0, dtype: int64
+
+        Squeezing all axes wil project directly into a scalar:
+
+        >>> df_0a.squeeze()
+        1
```
datapythonista: I feel like this example using `df` covers what it's shown first with the primes. I'd leave just this one, personally I find it really good, and enough to not have to list the previous.

datapythonista (commented):
Great pull request, added some comments/ideas.

pandas/core/generic.py
villasv: It is somewhat, I think squeezing is most useful in slicing scenarios but perhaps someone might find that a direct conversion is what they really wanted. Will add it too.

pandas/core/generic.py
villasv: The hard part of describing squeeze it's that it's too general, but perhaps I lost touch with practicality because I was thinking of squeezing N-dimensional objects, even though pandas deals mainly with 1D (Series) and 2D (Frames).

I hoped to make it concrete in the examples, but I'm afraid people will just give up reading after a paragraph like this.

pandas/core/generic.py
villasv: Sorry, "objects" became ambiguous. The objects inside that DataFrame axis would be 1D, that's what I meant. Will try to clarify.

pandas/core/generic.py
villasv: I wanted to have examples with both Series and DataFrames because both classes share this docstring, so it would be a bit weird to read the docs from `Series.squeeze` and find examples only of `DataFrame.squeeze`. But I think I could "chain" those examples, since in the middle of the `df` example I may squeeze some `Series` as well.

pandas/core/generic.py
villasv: I couldn't find anything on the guidelines about it, but I'm taking other docs as guide: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.sum.html

TomAugspurger:
Updated, if anyone wants to take a look.  Merging tomorrow otherwise.

I went with @datapythonista's suggestion on how to phrase the extended summary. Pandas only has to worry about `n <= 2`, so let's take advantage of that in the docs.

villasv:
I was going to implement a few of the discussed changes next Saturday, but you can merge this one and if I actually get to it I'll make a new PR.

TomAugspurger:
No rush. Feel free to pull my changes and push them here.

pandas/core/generic.py
jorisvandenbossche: I don't think `DataFrame.to_series` exists?

datapythonista: hehe, that's a good reason to not add it... ;) not sure what I was thinking about, I think I got confused with `Index.to_series`, sorry

jschendel:
Merging this as-is, since it looks like this is good.

@villasv : Thanks! Happy to take any additional updates as a separate PR, if you're still interested in implementing additional changes.

PR #20273, opened 2018-03-11
DOC: update the pandas.Series.str.slice_replace docstring

techedlaksh:
an unrelated PR):

```
################# Docstring (pandas.Series.str.slice_replace)  #################

Replace a sliced string.

Replace a slice of each string in the Series/Index with another
string.

Parameters
----------
start : int or None
    Left edge index.
stop : int or None
    Right edge index.
repl : str or None
    String for replacement.

Returns
-------
replaced : Series/Index of objects

Examples
--------
>>> s = pd.Series(['This is a Test 1', 'This is a Test 2'])
>>> s
0    This is a Test 1
1    This is a Test 2
dtype: object
>>> s = s.str.slice_replace(8, 14, 'an Example')
>>> s
0    This is an Example 1
1    This is an Example 2
dtype: object

Errors found:
	See Also section not found

```


pandas/core/strings.py
```diff
@@ -1180,19 +1180,36 @@ def str_slice(arr, start=None, stop=None, step=None):
 
 def str_slice_replace(arr, start=None, stop=None, repl=None):
     """
+    Replace a sliced string.
```
TomAugspurger: Maybe it's implied, but I'd like to see this say what the sliced bit is replaced with

    Replace a sliced string with another value.

    Like all of pandas string methods, this operates elementwise on each string
    in the Series or Index.

and then the rest of the docstring.

pandas/core/strings.py
```diff
@@ -1180,19 +1180,36 @@ def str_slice(arr, start=None, stop=None, step=None):
...
     stop : int or None
+        Right edge index.
     repl : str or None
-        String for replacement
+        String for replacement.
 
     Returns
     -------
     replaced : Series/Index of objects
+
```
TomAugspurger: For a See Also, `str.replace` to link to https://docs.python.org/3/library/stdtypes.html#str.replace would be good.

TomAugspurger (approved):
Minor suggestions, if you have a chance to update it'd be great. Looks good though!

pandas/core/strings.py
jorisvandenbossche: I think "a slice of a string" is easier to understand than "sliced string"

pandas/core/strings.py
```diff
@@ -1180,19 +1180,36 @@ def str_slice(arr, start=None, stop=None, step=None):
...
     """
+    Replace a sliced string.
+
     Replace a slice of each string in the Series/Index with another
     string.
 
     Parameters
     ----------
     start : int or None
+        Left edge index.
```
jorisvandenbossche: Can you indicate what None means? (I suppose this is the from the start of the string)

Can you also change "int or None" into "int, optional"? And then maybe you can say for the above "If not specified, slice from the start of each string"

(and same for end, but then "slice until the end of each string"

pandas/core/strings.py
jorisvandenbossche: And we also have `replace` as a string method in `Series.str`, so I would rather add a link to that?

jorisvandenbossche (commented):
Thanks! Added some additional comments

pandas/core/strings.py
```diff
@@ -1180,19 +1180,36 @@ def str_slice(arr, start=None, stop=None, step=None):
...
     replaced : Series/Index of objects
+
+    Examples
+    --------
+    >>> s = pd.Series(['This is a Test 1', 'This is a Test 2'])
+    >>> s
+    0    This is a Test 1
+    1    This is a Test 2
+    dtype: object
+    >>> s = s.str.slice_replace(8, 14, 'an Example')
```
jreback: blank line between cases

jreback (changes requested):

TomAugspurger:
Updated if you want to take a look.

I'm not sure how, but apparently changing the default `repl` to `''` broke tests.  https://github.com/pandas-dev/pandas/pull/20273/commits/7a0cf0c93ad4778c78a6c7a477f9309cc6e13403 reverted that to be how it was before. Will let the tests run on this before merging.

pandas/core/strings.py
```diff
@@ -1180,35 +1180,74 @@ def str_slice(arr, start=None, stop=None, step=None):
...
     Parameters
     ----------
-    start : int or None
-        Left edge index.
-    stop : int or None
-        Right edge index.
-    repl : str or None
-        String for replacement.
+    start : int, optional
+        Left index position to use for the slice. The default of None
```
jorisvandenbossche: General question: do we more want to use the phrase of "If not specified" instead of "The default of None" if we say something is optional? 
On the one hand I think this can be clearer as that is what you do in practice, you don't specify it. But on the other hand of course the signature shows None.

TomAugspurger: Yeah, if not specified is probably best.

TomAugspurger: Went with `if not specified (None)`

jorisvandenbossche: Ah, `If not specified (None)`, is a good combination of both, we should use that pattern more

PR #20274, opened 2018-03-11
DOC: update the pandas.Series.dt.is_year_end docstring

janelle:
an unrelated PR):

```
################### Docstring (pandas.Series.dt.is_year_end) ###################

Return a boolean indicating whether the date is the last day of the year.

Returns
-------
is_year_end : Series of boolean.

See Also
--------
is_year_start : Return a boolean indicating whether the date is the first day of the year.

Examples
--------
>>> dates = pd.Series(pd.date_range("2017-12-30", periods=3))
>>> dates
0   2017-12-30
1   2017-12-31
2   2018-01-01
dtype: datetime64[ns]
>>> dates.dt.is_year_end
0    False
1     True
2    False
dtype: bool

Errors found:
	No extended summary found

```

explicitly.

* No extended summary required


pandas/core/indexes/datetimes.py
```diff
@@ -1747,7 +1747,33 @@ def freq(self, value):
     is_year_end = _field_accessor(
         'is_year_end',
         'is_year_end',
-        "Logical indicating if last day of year (defined by frequency)")
+        """
+        Return a boolean indicating whether the date is the last day of the
+        year.
```
TomAugspurger: This should be a single line. Does

    Return a boolean array of whether the date is the year's last day.

fit on a single line?

pandas/core/indexes/datetimes.py
```diff
@@ -1747,7 +1747,33 @@ def freq(self, value):
...
+        0   2017-12-30
+        1   2017-12-31
+        2   2018-01-01
+        dtype: datetime64[ns]
+        >>> dates.dt.is_year_end
+        0    False
+        1     True
+        2    False
+        dtype: bool
+        """)
```
TomAugspurger: Could you add an example with just a DatetimeIndex? Can be the same values, and don't have to show the values again,

```
>>> pd.date_range('2017-12-30", periods=3).is_year_end
```

and show the output

pandas/core/indexes/datetimes.py
```diff
@@ -1747,7 +1747,33 @@ def freq(self, value):
...
         'is_year_end',
         'is_year_end',
-        "Logical indicating if last day of year (defined by frequency)")
+        """
+        Return a boolean indicating whether the date is the last day of the
+        year.
+
+        Returns
+        -------
+        is_year_end : Series of boolean.
```
TomAugspurger: Ah, this return type depends on the type of self. So I think

is_year_end : Series or DatetimeIndex
    The same type as the original data. Series will have the same
    name and index. DatetimeIndex will have the same name.

pandas/core/indexes/datetimes.py
```diff
@@ -1747,7 +1747,33 @@ def freq(self, value):
...
+
+        See Also
+        --------
+        is_year_start : Return a boolean indicating whether the date is the
```
TomAugspurger: See comment https://github.com/pandas-dev/pandas/pull/20275#discussion_r173651295 (once we figure out how to handle these)

pandas/core/indexes/datetimes.py
jorisvandenbossche: Boolean array is not correct in case of Series ... 

(but I agree that just "boolean" does not reflect is returns a boolean for each element)

TomAugspurger: Perhaps just

    Return an indicator for whether each date is the last day of the year.

jorisvandenbossche: Yes, that's better. Or maybe even shorter "Indicate whether ..." instead of "Return an indicator whether .." ?

TomAugspurger:
Updated. Fixed some some See Alsos in the other docstrings where we said `method` instead of `property`.

Thanks again @janelle!

PR #20275, opened 2018-03-11
DOC: update the pandas.Series.dt.is_year_start docstring

janelle:
an unrelated PR):

```
################## Docstring (pandas.Series.dt.is_year_start) ##################

Return a boolean indicating whether the date is the first day of the
year.

Returns
-------
is_year_start : Series of boolean.

See Also
--------
is_year_end : Return a boolean indicating whether the date is the
    last day of the year.

Examples
--------
>>> dates = pd.Series(pd.date_range("2017-12-30", periods=3))
>>> dates
0   2017-12-30
1   2017-12-31
2   2018-01-01
dtype: datetime64[ns]
>>> dates.dt.is_year_start
0    False
1    False
2    True
dtype: bool

Errors found:

```

explicitly.

* short summary exceeds one line by one word, hope that is ok

### DIFF
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1764,7 +1764,7 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
         1   2017-12-31
         2   2018-01-01
         dtype: datetime64[ns]
-        >>> dates.dt.is_year_end
+        >>> dates.dt.is_year_start
         0    False
         1    False
         2    True
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1744,31 +1744,44 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
         'is_year_start',
         'is_year_start',
         """
-        Return a boolean indicating whether the date is the first day of the
-        year.
+        Indicate whether the date is the first day of a year.
 
         Returns
         -------
-        is_year_start : Series of boolean.
+        Series or DatetimeIndex
+            The same type as the original data with boolean values. Series will
+            have the same name and index. DatetimeIndex will have the same
+            name.
 
         See Also
         --------
-        is_year_end : Return a boolean indicating whether the date is the
-            last day of the year.
+        is_year_end : Similar method indicating the last day of the year.
 
         Examples
         --------
+        This method is available on Series with datetime values under
+        the ``.dt`` accessor, and directly on DatetimeIndex.
+
         >>> dates = pd.Series(pd.date_range("2017-12-30", periods=3))
         >>> dates
         0   2017-12-30
         1   2017-12-31
         2   2018-01-01
         dtype: datetime64[ns]
+
         >>> dates.dt.is_year_start
         0    False
         1    False
         2    True
         dtype: bool
+
+        >>> idx = pd.date_range("2017-12-30", periods=3)
+        >>> idx
+        DatetimeIndex(['2017-12-30', '2017-12-31', '2018-01-01'],
+                      dtype='datetime64[ns]', freq='D')
+
+        >>> idx.is_year_start
+        array([False, False,  True])
         """)
     is_year_end = _field_accessor(
         'is_year_end',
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/datetimes.py
```diff
@@ -1743,7 +1743,33 @@ def freq(self, value):
     is_year_start = _field_accessor(
         'is_year_start',
         'is_year_start',
-        "Logical indicating if first day of year (defined by frequency)")
+        """
+        Return a boolean indicating whether the date is the first day of the
```
TomAugspurger: Same thing about the single line.

pandas/core/indexes/datetimes.py
```diff
@@ -1743,7 +1743,33 @@ def freq(self, value):
...
+        year.
+
+        Returns
+        -------
+        is_year_start : Series of boolean.
```
TomAugspurger:
```
is_year_end : Series or DatetimeIndex
    The same type as the original data. Series will have the same
    name and index. DatetimeIndex will have the same name.
```

pandas/core/indexes/datetimes.py
```diff
@@ -1743,7 +1743,33 @@ def freq(self, value):
...
+        >>> dates
+        0   2017-12-30
+        1   2017-12-31
+        2   2018-01-01
+        dtype: datetime64[ns]
+        >>> dates.dt.is_year_start
+        0    False
+        1    False
+        2    True
+        dtype: bool
```
TomAugspurger: Second example showing DatetimeIndex.is_year_start.

pandas/core/indexes/datetimes.py
```diff
@@ -1743,7 +1743,33 @@ def freq(self, value):
...
+        Return a boolean indicating whether the date is the first day of the
+        year.
+
+        Returns
+        -------
+        is_year_start : Series of boolean.
+
+        See Also
+        --------
+        is_year_end : Return a boolean indicating whether the date is the
```
TomAugspurger: I'm not sure this will resolve correctly, @jorisvandenbossche do you know?

I think it should be Series.dt.is_year_start

or DatetimeIndex.is_year_start. Don't think we've decided what to do here...

jorisvandenbossche: Well, I was just testing it :-) 
Turns out this works (because it is the same class, it resolves OK), but I wanted to comment that for clarity I would write it in full like `DatetimeIndex.is_year_start`, but on the other hand, not writing it full ensures it links correctly in both Series.dt and DatetimeIndex case

pandas/core/indexes/datetimes.py
jorisvandenbossche: General question here: for attributes, do we want to follow our rule of starting the summary line with a verb? 

I have the feeling that if all those attributes start with "Return", that's rather superfluous

jorisvandenbossche: For example, this gives now a nice overview table with short explanations: http://pandas.pydata.org/pandas-docs/stable/api.html#time-date-components

Personally, I would say that if all those "The year of the datetime" become "Return the year of the datetime" etc, this is not necessarily an improvement

cc @datapythonista your view here is very appreciated

pandas/core/indexes/datetimes.py
jorisvandenbossche: Can you then also write something like "This attribute is both available on a DatetimeIndex as a Series with datetime values" or for the Series example: "On a Series with datetime values, the attribute can be accessed with `.dt`"

pandas/core/indexes/datetimes.py
TomAugspurger: Ah, in that case I'm ok with leaving it "ambiguous" since in the HTML docs you can click through and get the correct one. In the REPL you have tab completion.

pandas/core/indexes/datetimes.py
datapythonista: I agree, I think that policy only makes sense to functions and methods, not to classes, properties or attributes.

If I'm not wrong, that is said in the documentation, and also the validation avoids the check for the infinitive verb.

pandas/core/indexes/datetimes.py
```diff
@@ -1743,7 +1743,33 @@ def freq(self, value):
...
+
+        Examples
+        --------
+        >>> dates = pd.Series(pd.date_range("2017-12-30", periods=3))
+        >>> dates
+        0   2017-12-30
+        1   2017-12-31
+        2   2018-01-01
+        dtype: datetime64[ns]
+        >>> dates.dt.is_year_start
```
jreback: blank line between cases

jreback (changes requested):

TomAugspurger:
And thanks again :)

PR #20276, opened 2018-03-11
DOC: update the aggregate docstring

albertvillanova:
an unrelated PR):

```
#################### Docstring (pandas.DataFrame.aggregate) ####################

Aggregate using one or multiple operations along the specified axis.

.. versionadded:: 0.20.0

Parameters
----------
func : function, string, dictionary, or list of string/functions
    Function to use for aggregating the data. If a function, must either
    work when passed a DataFrame or when passed to DataFrame.apply. For
    a DataFrame, can pass a dict, if the keys are DataFrame column names.

    Accepted combinations are:

    - string function name.
    - function.
    - list of functions.
    - dict of column names -> functions (or list of functions).
    - 0 or 'index': apply function to each column.
    - 1 or 'columns': apply function to each row.
args
    Optional positional arguments to pass to the function.
kwargs
    Optional keyword arguments to pass to the function.

Returns
-------
aggregated : DataFrame

Notes
-----
`agg` is an alias for `aggregate`. Use the alias.

Notes
-----
The default behavior of aggregating over the axis 0 is different from
`numpy` functions `mean`/`median`/`prod`/`sum`/`std`/`var`, where the
default is to compute the aggregation of the flattened array (e.g.,
`numpy.mean(arr_2d)` as opposed to `numpy.mean(arr_2d, axis=0)`).

`agg` is an alias for `aggregate`. Use the alias.

Examples
--------

>>> df = df = pd.DataFrame([[1,2,3],
...                         [4,5,6],
...                         [7,8,9],
...                         [np.nan, np.nan, np.nan]],
...                        columns=['A', 'B', 'C'])

Aggregate these functions across all columns

>>> df.aggregate(['sum', 'min'])
        A     B     C
sum  12.0  15.0  18.0
min   1.0   2.0   3.0

Different aggregations per column

>>> df.aggregate({'A' : ['sum', 'min'], 'B' : ['min', 'max']})
        A    B
max   NaN  8.0
min   1.0  2.0
sum  12.0  NaN

See also
--------
pandas.DataFrame.apply : Perform any type of operations.
pandas.DataFrame.transform : Perform transformation type operations.
pandas.DataFrame.groupby.aggregate : Perform aggregation type operations
    over groups.
pandas.DataFrame.resample.aggregate : Perform aggregation type operations
    over resampled bins.
pandas.DataFrame.rolling.aggregate : Perform aggregation type operations
    over rolling window.

Errors found:
	Errors in parameters section
		Parameter "axis" description should start with capital letter
		Parameter "args" has no type

```

explicitly.

- args and kwargs have no type
- axis : {0 or 'index', 1 or 'columns'}, default 0

This PR has been made in collaboration with @rochamatcomp

Question to reviewers: we have added to the docstring the parameter `axis` because this is appropriate for DataFrame. However this dosctring is generic (shared with Series, groupby, window and resample) and the parameter `axis` is wrong for them (for Series the parameter has only one possible value). What should be the right way to fix this?

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -107,6 +107,10 @@ from pandas.core.config import get_option
 _shared_doc_kwargs = dict(
     axes='index, columns', klass='DataFrame',
     axes_single_arg="{0 or 'index', 1 or 'columns'}",
+    axis="""
+    axis : {0 or 'index', 1 or 'columns'}, default 0
+        - 0 or 'index': apply function to each column.
+        - 1 or 'columns': apply function to each row.""",
     optional_by="""
         by : str or list of str
             Name or list of names to sort by.
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4463,10 +4463,10 @@ class DataFrame(NDFrame):
         Return reshaped DataFrame organized by given index / column values.
 
         Reshape data (produce a "pivot" table) based on column values. Uses
-        unique values from specified `index` / `columns` to form axes of the resulting
-        DataFrame. This function does not support data aggregation, multiple
-        values will result in a MultiIndex in the columns. See the
-        :ref:`User Guide <reshaping>` for more on reshaping.
+        unique values from specified `index` / `columns` to form axes of the
+        resulting DataFrame. This function does not support data
+        aggregation, multiple values will result in a MultiIndex in the
+        columns. See the :ref:`User Guide <reshaping>` for more on reshaping.
 
         Parameters
         ----------
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4764,36 +4764,49 @@ class DataFrame(NDFrame):
         return self[key]
 
     _agg_doc = dedent("""
+    Notes
+    -----
+    The default behavior of aggregating over the axis 0 is different from
+    `numpy` functions `mean`/`median`/`prod`/`sum`/`std`/`var`, where the
+    default is to compute the aggregation of the flattened array (e.g.,
+    `numpy.mean(arr_2d)` as opposed to `numpy.mean(arr_2d, axis=0)`).
+
+    `agg` is an alias for `aggregate`. Use the alias.
+
     Examples
     --------
 
-    >>> df = pd.DataFrame(np.random.randn(10, 3), columns=['A', 'B', 'C'],
-    ...                   index=pd.date_range('1/1/2000', periods=10))
-    >>> df.iloc[3:7] = np.nan
+    >>> df = df = pd.DataFrame([[1,2,3],
+    ...                         [4,5,6],
+    ...                         [7,8,9],
+    ...                         [np.nan, np.nan, np.nan]],
+    ...                        columns=['A', 'B', 'C'])
 
     Aggregate these functions across all columns
 
-    >>> df.agg(['sum', 'min'])
-                A         B         C
-    sum -0.182253 -0.614014 -2.909534
-    min -1.916563 -1.460076 -1.568297
+    >>> df.aggregate(['sum', 'min'])
+            A     B     C
+    sum  12.0  15.0  18.0
+    min   1.0   2.0   3.0
 
     Different aggregations per column
 
-    >>> df.agg({'A' : ['sum', 'min'], 'B' : ['min', 'max']})
-                A         B
-    max       NaN  1.514318
-    min -1.916563 -1.460076
-    sum -0.182253       NaN
+    >>> df.aggregate({'A' : ['sum', 'min'], 'B' : ['min', 'max']})
+            A    B
+    max   NaN  8.0
+    min   1.0  2.0
+    sum  12.0  NaN
 
     See also
     --------
-    pandas.DataFrame.apply
-    pandas.DataFrame.transform
-    pandas.DataFrame.groupby.aggregate
-    pandas.DataFrame.resample.aggregate
-    pandas.DataFrame.rolling.aggregate
-
+    pandas.DataFrame.apply : Perform any type of operations.
+    pandas.DataFrame.transform : Perform transformation type operations.
+    pandas.DataFrame.groupby.aggregate : Perform aggregation type operations
+        over groups.
+    pandas.DataFrame.resample.aggregate : Perform aggregation type operations
+        over resampled bins.
+    pandas.DataFrame.rolling.aggregate : Perform aggregation type operations
+        over rolling window.
     """)
 
     @Appender(_agg_doc)
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4997,27 +4997,35 @@ class DataFrame(NDFrame):
 
     Examples
     --------
-
-    >>> df = pd.DataFrame([[1,2,3],
-    ...                    [4,5,6],
-    ...                    [7,8,9],
+    >>> df = pd.DataFrame([[1, 2, 3],
+    ...                    [4, 5, 6],
+    ...                    [7, 8, 9],
     ...                    [np.nan, np.nan, np.nan]],
     ...                   columns=['A', 'B', 'C'])
 
-    Aggregate these functions across all columns
+    Aggregate these functions over the rows.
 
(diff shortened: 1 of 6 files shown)


pandas/core/series.py
```diff
@@ -2353,6 +2353,13 @@ def _gotitem(self, key, ndim, subset=None):
         return self
 
     _agg_doc = dedent("""
+    Notes
+    -----
```
TomAugspurger: Can remove this section. Move the axis discussion to a substitution in parameters.

pandas/core/frame.py
```diff
@@ -4764,36 +4764,49 @@ def _gotitem(self, key, ndim, subset=None):
...
+    ...                         [np.nan, np.nan, np.nan]],
+    ...                        columns=['A', 'B', 'C'])
 
     Aggregate these functions across all columns
 
-    >>> df.agg(['sum', 'min'])
-                A         B         C
-    sum -0.182253 -0.614014 -2.909534
-    min -1.916563 -1.460076 -1.568297
+    >>> df.aggregate(['sum', 'min'])
```
TomAugspurger: I think we recommend using `.agg`

pandas/core/frame.py
```diff
@@ -4764,36 +4764,49 @@ def _gotitem(self, key, ndim, subset=None):
         return self[key]
 
     _agg_doc = dedent("""
+    Notes
+    -----
+    The default behavior of aggregating over the axis 0 is different from
+    `numpy` functions `mean`/`median`/`prod`/`sum`/`std`/`var`, where the
```
TomAugspurger: Does sphinx complain about having a character immediately after the backticks? May need spaces.

pandas/core/frame.py
```diff
@@ -4764,36 +4764,49 @@ def _gotitem(self, key, ndim, subset=None):
...
+    default is to compute the aggregation of the flattened array (e.g.,
+    `numpy.mean(arr_2d)` as opposed to `numpy.mean(arr_2d, axis=0)`).
```
TomAugspurger: double-ticks around the code snippets.

pandas/core/frame.py
```diff
@@ -4764,36 +4764,49 @@ def _gotitem(self, key, ndim, subset=None):
...
 
     See also
     --------
-    pandas.DataFrame.apply
-    pandas.DataFrame.transform
-    pandas.DataFrame.groupby.aggregate
-    pandas.DataFrame.resample.aggregate
-    pandas.DataFrame.rolling.aggregate
-
+    pandas.DataFrame.apply : Perform any type of operations.
```
TomAugspurger: You can remove the `pandas.` from these, save a bit of space.

pandas/core/frame.py
```diff
@@ -4764,36 +4764,49 @@ def _gotitem(self, key, ndim, subset=None):
...
+    pandas.DataFrame.transform : Perform transformation type operations.
+    pandas.DataFrame.groupby.aggregate : Perform aggregation type operations
```
TomAugspurger: This should be pandas.core.groupby.GroupBy.transform.

TomAugspurger: Actually, best to just remove the tranform / aggregate bits and just link to

pandas.core.groupby.GroupBy
pandas.core.resample.Resampler
pandas.core.window.Rolling
pandas.core.window.Expanding

pandas/core/generic.py
```diff
@@ -3929,36 +3929,38 @@ def pipe(self, func, *args, **kwargs):
         return com._pipe(self, func, *args, **kwargs)
 
     _shared_docs['aggregate'] = ("""
-    Aggregate using callable, string, dict, or list of string/callables
+    Aggregate using one or multiple operations along the specified axis.
```
TomAugspurger: multiple -> more

TomAugspurger: along -> over I think.

Does along mean apply to each element in an axis? And over means collapse an axis?

TomAugspurger (changes requested):
Regarding `axis`, I would add a variable in the shared doc `%(axis)s`.

In core/frame you can sub `axis : {0 or 'index', 1 or 'columns'}, default 0` and an explanation

In core/series you can sub `axis: {0 or 'index}` and note that it's for compatibility.

pandas/core/frame.py
```diff
@@ -4764,36 +4764,49 @@ def _gotitem(self, key, ndim, subset=None):
...
 
     _agg_doc = dedent("""
+    Notes
+    -----
+    The default behavior of aggregating over the axis 0 is different from
+    `numpy` functions `mean`/`median`/`prod`/`sum`/`std`/`var`, where the
+    default is to compute the aggregation of the flattened array (e.g.,
+    `numpy.mean(arr_2d)` as opposed to `numpy.mean(arr_2d, axis=0)`).
+
+    `agg` is an alias for `aggregate`. Use the alias.
```
jreback: I am not sure this is relevant, we *always* specify an axis, so by-definition you always have column-by-column or row-by-row aggregations, ``axis=None`` is simply not possible.

jreback: maybe adding some expl would help here.

pandas/core/frame.py
```diff
@@ -4764,36 +4764,49 @@ def _gotitem(self, key, ndim, subset=None):
...
+
     Examples
     --------
 
-    >>> df = pd.DataFrame(np.random.randn(10, 3), columns=['A', 'B', 'C'],
-    ...                   index=pd.date_range('1/1/2000', periods=10))
-    >>> df.iloc[3:7] = np.nan
+    >>> df = df = pd.DataFrame([[1,2,3],
```
jreback: extra df =

jreback (changes requested):

albertvillanova:
I have read that there has been a change to the convention during the sprint, and `args` and `kwargs` should be in the same line and a common description in the next line. Should we implement this?

pandas/core/generic.py
```diff
@@ -3929,36 +3929,36 @@ def pipe(self, func, *args, **kwargs):
...
-    mimicking the default Numpy behavior (e.g., np.mean(arr_2d)).
+        Accepted combinations are:
 
-    `agg` is an alias for `aggregate`. Use the alias.
+        - string function name.
+        - function.
+        - list of functions.
+        - dict of column names -> functions (or list of functions).
+    %(axis)s
+    args
```
TomAugspurger: In this case, we'll document them separately.

This should be `*args`.

pandas/core/generic.py
```diff
@@ -3929,36 +3929,36 @@ def pipe(self, func, *args, **kwargs):
...
+        Optional positional arguments to pass to the function.
```
TomAugspurger: Strike optional. The agg func could have additional required parameters.

pandas/core/generic.py
```diff
@@ -3929,36 +3929,36 @@ def pipe(self, func, *args, **kwargs):
...
+    kwargs
```
TomAugspurger: This should be `**kwargs`.

TomAugspurger (commented):
In this case we'll document them separately.

pandas/core/frame.py
TomAugspurger: Still have this @albertvillanova

pandas/core/frame.py
albertvillanova: @TomAugspurger  Also pandas.core.window.EWM?

TomAugspurger:
Sure

pandas/core/frame.py
albertvillanova: Is it better now?

pandas/core/frame.py
albertvillanova: I have added commas instead.

TomAugspurger:
Looks like there's a merge conflict @albertvillanova. Can you update? LMK if you need help with the git stuff.

TomAugspurger (approved):
Added an example with `axis='columns'`.

albertvillanova:
Please, note that this PR has been made in collaboration with @rochamatcomp

PR #20277, opened 2018-03-11
DOC: update the Period.dayofyear docstring

techytushar:
Signed-off-by: Tushar Mittal <chiragmittal.mittal@gmail.com>

Checklist for the pandas documentation sprint (ignore this if you are doing
an unrelated PR):

- [x] PR title is "DOC: update the <your-function-or-method> docstring"
- [x] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [x] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [x] It has been proofread on language by another sprint participant

Please include the output of the validation script below between the "```" ticks:

```
################################################################################
##################### Docstring (pandas.Period.dayofyear)  #####################
################################################################################

Return the day of the year.

This attribute returns the day of the year on which the particular
date occurs. The return value ranges between 1 to 365 for regular
years and 1 to 366 for leap years.

Returns
-------
int
    The day of year.

See Also
--------
Period.dayofweek : Return the day of week.
Period.daysinmonth : Return the days in that month.
PeriodIndex.dayofyear : Return the day of year of all indexes. 

Examples
--------
>>> period = pd.Period("2015-10-23", freq='H')
>>> period.dayofyear
296
>>> period = pd.Period("2012-12-31", freq='D')
>>> period.dayofyear
366
>>> period = pd.Period("2013-01-01", freq='D')
>>> period.dayofyear
1

################################################################################
################################## Validation ##################################
################################################################################

Docstring for "pandas.Period.dayofyear" correct. :)
```

If the validation script still gives errors, but you think there is a good reason
to deviate in this case (and there are certainly such cases), please state this
explicitly.

PR #20280, opened 2018-03-11
DOC: update the Period.dayofweek attribute docstring

hammadmashkoor:
an unrelated PR):

```
(pandas_dev) root@kalih:~/pythonpanda/pandas# python scripts/validate_docstrings.py pandas.Period.dayofweek

##################### Docstring (pandas.Period.dayofweek)  #####################

Return the day of the week.

This attribute returns the day of the week on which the particular
starting date for the given period occurs with Monday=0, Sunday=6.

Returns
-------
Int
    Range of 0 to 6

See also
--------
Period.dayofyear
    Return the day of year.
Period.daysinmonth
    Return the days in that month.

Examples
--------
>>> period1 = pd.Period('2012-1-1 19:00', freq='H')
>>> period1
Period('2012-01-01 19:00', 'H')
>>> period1.dayofweek
6

>>> period2 = pd.Period('2013-1-9 11:00', freq='H')
>>> period2
Period('2013-01-09 11:00', 'H')
>>> period2.dayofweek
2

Docstring for "pandas.Period.dayofweek" correct. :)
```

explicitly.

### DIFF
--- a/pandas/_libs/tslibs/period.pyx
+++ b/pandas/_libs/tslibs/period.pyx
@@ -1260,8 +1260,8 @@ cdef class _Period(object):
 
         See also
         --------
-        Period.dayofyear : Return the day of year.
-        Period.daysinmonth : Return the days in that month.
+        Period.dayofyear : Return the day of the year.
+        Period.daysinmonth : Return the number of days in that month.
 
         Examples
         --------
--- a/pandas/_libs/tslibs/period.pyx
+++ b/pandas/_libs/tslibs/period.pyx
@@ -1250,19 +1250,18 @@ cdef class _Period(object):
         Return the day of the week.
 
         This attribute returns the day of the week on which the particular
-        starting date for the given period occurs with Monday=0, Sunday=6.
+        date for the given period occurs depending on the frequency with
+        Monday=0, Sunday=6.
 
         Returns
         -------
         Int
-            Range from 0 to 6(included).
+            Range from 0 to 6 (included).
 
         See also
         --------
-        Period.dayofyear
-            Return the day of year.
-        Period.daysinmonth
-            Return the days in that month.
+        Period.dayofyear : Return the day of year.
+        Period.daysinmonth : Return the days in that month.
 
         Examples
         --------
--- a/pandas/_libs/tslibs/period.pyx
+++ b/pandas/_libs/tslibs/period.pyx
@@ -1255,7 +1255,7 @@ cdef class _Period(object):
         Returns
         -------
         Int
-            Range of 0 to 6
+            Range from 0 to 6(included).
 
         See also
         --------
--- a/pandas/_libs/tslibs/period.pyx
+++ b/pandas/_libs/tslibs/period.pyx
@@ -1246,6 +1246,37 @@ cdef class _Period(object):
 
     @property
     def dayofweek(self):
+        """
+        Return the day of the week.
+
+        This attribute returns the day of the week on which the particular
+        date occurs with Monady=0, Sunday=6.
+
+        Returns
+        -------
+        Int:Range of 0 to 6
+
+        See also
+        --------
+        Period.dayofyear
+            Return the day of year.
+        Period.daysinmonth
+            Return the days in that month.
+
+        Examples
+        --------
+        >>> p=pd.Period('2012-1-1 19:00', freq='H')
+        >>> p
+        Period('2012-01-01 19:00', 'H')
+        >>> p.dayofweek
+        6
+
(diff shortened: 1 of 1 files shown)


pandas/_libs/tslibs/period.pyx
```diff
@@ -1246,6 +1246,37 @@ cdef class _Period(object):
 
     @property
     def dayofweek(self):
+        """
+        Return the day of the week.
+
+        This attribute returns the day of the week on which the particular
+        date occurs with Monady=0, Sunday=6.
```
dukebody: Monday

pandas/_libs/tslibs/period.pyx
```diff
@@ -1246,6 +1246,37 @@ cdef class _Period(object):
...
+        See also
+        --------
+        Period.dayofyear
+            Return the day of year.
+        Period.daysinmonth
+            Return the days in that month.
+
+        Examples
+        --------
+        >>> p=pd.Period('2012-1-1 19:00', freq='H')
```
dukebody: 1) Add spaces around `p` to follow flake8 also in docstrings.
2) Perhaps better to use `period` for the variable name, it's short enough.

pandas/_libs/tslibs/period.pyx
```diff
@@ -1246,6 +1246,37 @@ cdef class _Period(object):
 
     @property
     def dayofweek(self):
+        """
+        Return the day of the week.
+
+        This attribute returns the day of the week on which the particular
```
dukebody: If I'm not mistaken, this will return the day of the week on which the particular *starting* date for the given period occurs. I think this is important to mention, because at first I was confused thinking "how a period of one month can have a day of the week?".

pandas/_libs/tslibs/period.pyx
```diff
@@ -1246,6 +1246,37 @@ cdef class _Period(object):
...
+        date occurs with Monady=0, Sunday=6.
+
+        Returns
+        -------
+        Int:Range of 0 to 6
```
dukebody: Not following the style guide, see https://python-sprints.github.io/pandas/guide/pandas_docstring.html#section-4-returns-or-yields

dukebody (changes requested):
Can you fix the highlighted issues?

pandas/_libs/tslibs/period.pyx
```diff
@@ -1246,6 +1246,38 @@ cdef class _Period(object):
...
+        """
+        Return the day of the week.
+
+        This attribute returns the day of the week on which the particular
+        starting date for the given period occurs with Monday=0, Sunday=6.
+
+        Returns
+        -------
+        Int
+            Range of 0 to 6
```
dukebody: - I'd say correct English should be "Range from 0 to 6". So "from", not "of". But I'm not a native English speaker so please double check.
- Perhaps add (included) to make clear 6 is also included. Since `range(0, 6)` in Python doesn't include the 6. :)
- End line with period.

pandas/_libs/tslibs/period.pyx
```diff
@@ -1246,6 +1246,38 @@ cdef class _Period(object):
...
+        """
+        Return the day of the week.
+
+        This attribute returns the day of the week on which the particular
+        starting date for the given period occurs with Monday=0, Sunday=6.
+
+        Returns
+        -------
+        Int
+            Range from 0 to 6(included).
```
jorisvandenbossche: space of 6

pandas/_libs/tslibs/period.pyx
jorisvandenbossche: Yes, we have this discussion in another PR as well (https://github.com/pandas-dev/pandas/pull/20277/files#r173651389). Only, it is not necessarily the starting date, it depends on the `freq`.

pandas/_libs/tslibs/period.pyx
```diff
@@ -1246,6 +1246,38 @@ cdef class _Period(object):
...
+
+        See also
+        --------
+        Period.dayofyear
+            Return the day of year.
```
jorisvandenbossche: Can you do this in form of `Period.dayofyear : Return the day of year.` ? (same for one below)

pandas/_libs/tslibs/period.pyx
jorisvandenbossche: Issue is here: https://github.com/pandas-dev/pandas/issues/20324.

jorisvandenbossche (approved):

jorisvandenbossche:
@hammadmashkoor Thanks a lot!

hammadmashkoor:
Thanks @jorisvandenbossche

PR #18462, opened 2017-11-24
Propogating NaN values when using str.split (#18450)

WillAyd:
- [X] closes #18450
- [X] tests added / passed
- [X] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

PR #10085, opened 2015-05-08
DEPR: Deprecate str.split return_type

sinhrks:
Closes #9847, Closes #9870. Default value is `expand=False` to be compat with`return_type='series'`. May better to change the default to True in future (and show warning about it)?

CC @jreback @jorisvandenbossche @sreejata

PR #20282, opened 2018-03-11
DOC: update the pandas.Series.str.split docstring

mananpal1997:
an unrelated PR):

```
##################### Docstring (pandas.Series.str.split)  #####################

Split strings around given separator/delimiter.

Split each str in the caller's values by given
pattern, propagating NaN values. Equivalent to :meth:`str.split`.

Parameters
----------
pat : string, default None
    String or regular expression to split on.
    If `None`, split on whitespace.
n : int, default -1 (all)
    Vary dimensionality of output.

    * `None`, 0 and -1 will be interpreted as return all splits
expand : bool, default False
    Expand the split strings into separate columns.

    * If `True`, return DataFrame/MultiIndex expanding dimensionality.
    * If `False`, return Series/Index.

Returns
-------
Type matches caller unless `expand=True` (return type is `DataFrame`)
split : Series/Index or DataFrame/MultiIndex of objects

Notes
-----
If `expand` parameter is `True` and:
  - If n >= default splits, makes all splits
  - If n < default splits, makes first n splits only
  - Appends `None` for padding.

Examples
--------
>>> s = pd.Series(["this is good text", "but this is even better"])

By default, split will return an object of the same size
having lists containing the split elements

>>> s.str.split()
0           [this, is, good, text]
1    [but, this, is, even, better]
dtype: object
>>> s.str.split("random")
0          [this is good text]
1    [but this is even better]
dtype: object

When using `expand=True`, the split elements will
expand out into separate columns.

>>> s.str.split(expand=True)
      0     1     2     3       4
0  this    is  good  text    None
1   but  this    is  even  better
>>> s.str.split(" is ", expand=True)
          0            1
0      this    good text
1  but this  even better

Parameter `n` can be used to limit the number of columns in
expansion of output.

>>> s.str.split("is", n=1, expand=True)
        0                1
0      th     is good text
1  but th   is even better

If NaN is present, it is propagated throughout the columns
during the split.

>>> s = pd.Series(["this is good text", "but this is even better", np.nan])
>>> s.str.split(n=3, expand=True)
      0     1     2            3
0  this    is  good         text
1   but  this    is  even better
2   NaN   NaN   NaN          NaN

Errors found:
	Errors in parameters section
		Parameter "n" description should finish with "."
	See Also section not found
```

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1122,9 +1122,12 @@ def str_split(arr, pat=None, n=None):
 
     Notes
     -----
-    - If n >= default splits, makes all splits
-    - If n < default splits, makes first n splits only
-    - Appends `None` for padding if ``expand=True``
+    The handling of the `n` keyword depends on the number of found splits:
+
+    - If found splits > `n`,  make first `n` splits only
+    - If found splits <= `n`, make all splits
+    - If for a certain row the number of found splits < `n`,
+      append `None` for padding up to `n` if ``expand=True``
 
     Examples
     --------
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1116,9 +1116,9 @@ def str_split(arr, pat=None, n=None):
 
     Returns
     -------
-    Type matches caller unless `expand=True` (return type is `DataFrame` or
-    `MultiIndex`)
     split : Series/Index or DataFrame/MultiIndex of objects
+        Type matches caller unless `expand=True` (return type is `DataFrame` or
+    `MultiIndex`)
 
     Notes
     -----
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1095,24 +1095,85 @@ def str_pad(arr, width, side='left', fillchar=' '):
 
 def str_split(arr, pat=None, n=None):
     """
-    Split each string (a la re.split) in the Series/Index by given
-    pattern, propagating NA values. Equivalent to :meth:`str.split`.
+    Split strings around given separator/delimiter.
+
+    Split each string in the caller's values by given
+    pattern, propagating NaN values. Equivalent to :meth:`str.split`.
 
     Parameters
     ----------
-    pat : string, default None
-        String or regular expression to split on. If None, splits on whitespace
+    pat : str, optional
+        String or regular expression to split on.
+        If `None`, split on whitespace.
     n : int, default -1 (all)
-        None, 0 and -1 will be interpreted as return all splits
+        Limit number of splits in output.
+        `None`, 0 and -1 will be interpreted as return all splits.
     expand : bool, default False
-        * If True, return DataFrame/MultiIndex expanding dimensionality.
-        * If False, return Series/Index.
+        Expand the splitted strings into separate columns.
 
-    return_type : deprecated, use `expand`
+        * If `True`, return DataFrame/MultiIndex expanding dimensionality.
+        * If `False`, return Series/Index, containing lists of strings.
 
     Returns
     -------
+    Type matches caller unless `expand=True` (return type is `DataFrame` or
+    `MultiIndex`)
     split : Series/Index or DataFrame/MultiIndex of objects
+
+    Notes
+    -----
+    - If n >= default splits, makes all splits
+    - If n < default splits, makes first n splits only
+    - Appends `None` for padding if `expand=True`
+
+    Examples
+    --------
+    >>> s = pd.Series(["this is good text", "but this is even better"])
+
+    By default, split will return an object of the same size
+    having lists containing the split elements
+
+    >>> s.str.split()
+    0           [this, is, good, text]
+    1    [but, this, is, even, better]
+    dtype: object
+    >>> s.str.split("random")
+    0          [this is good text]
+    1    [but this is even better]
+    dtype: object
+
+    When using `expand=True`, the split elements will
+    expand out into separate columns.
+
+    >>> s.str.split(expand=True)
+          0     1     2     3       4
+    0  this    is  good  text    None
+    1   but  this    is  even  better
+    >>> s.str.split(" is ", expand=True)
+              0            1
+    0      this    good text
+    1  but this  even better
+
+    Parameter `n` can be used to limit the number of splits in the output.
+
+    >>> s.str.split("is", n=1)
+    0          [th,  is good text]
+    1    [but th,  is even better]
+    dtype: object
+    >>> s.str.split("is", n=1, expand=True)
+            0                1
+    0      th     is good text
+    1  but th   is even better
+
+    If NaN is present, it is propagated throughout the columns
+    during the split.
+
+    >>> s = pd.Series(["this is good text", "but this is even better", np.nan])
+    >>> s.str.split(n=3, expand=True)
+          0     1     2            3
+    0  this    is  good         text
+    1   but  this    is  even better
+    2   NaN   NaN   NaN          NaN
     """
     if pat is None:
         if n is None or n == 0:
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1104,27 +1104,27 @@ def str_split(arr, pat=None, n=None):
     ----------
     pat : str, optional
         String or regular expression to split on.
-        If `None`, split on whitespace.
+        If not specified, split on whitespace.
     n : int, default -1 (all)
         Limit number of splits in output.
-        `None`, 0 and -1 will be interpreted as return all splits.
+        ``None``, 0 and -1 will be interpreted as return all splits.
     expand : bool, default False
         Expand the splitted strings into separate columns.
 
-        * If `True`, return DataFrame/MultiIndex expanding dimensionality.
-        * If `False`, return Series/Index, containing lists of strings.
+        * If ``True``, return DataFrame/MultiIndex expanding dimensionality.
+        * If ``False``, return Series/Index, containing lists of strings.
 
     Returns
     -------
     split : Series/Index or DataFrame/MultiIndex of objects
-        Type matches caller unless `expand=True` (return type is `DataFrame` or
-    `MultiIndex`)
+        Type matches caller unless ``expand=True`` (return type is DataFrame or
+    MultiIndex)
 
     Notes
     -----
     - If n >= default splits, makes all splits
     - If n < default splits, makes first n splits only
-    - Appends `None` for padding if `expand=True`
+    - Appends `None` for padding if ``expand=True``
 
     Examples
     --------
@@ -1142,7 +1142,7 @@ def str_split(arr, pat=None, n=None):
     1    [but this is even better]
     dtype: object
 
-    When using `expand=True`, the split elements will
+    When using ``expand=True``, the split elements will
     expand out into separate columns.
 
     >>> s.str.split(expand=True)


pandas/core/strings.py
```diff
@@ -1095,24 +1095,48 @@ def str_pad(arr, width, side='left', fillchar=' '):
 
 def str_split(arr, pat=None, n=None):
     """
+    Split strings around given separator/delimiter.
+
     Split each string (a la re.split) in the Series/Index by given
```
WillAyd: Can put reference to `re` module in backticks, so \`re.split\`

pandas/core/strings.py
```diff
@@ -1095,24 +1095,48 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
     pattern, propagating NA values. Equivalent to :meth:`str.split`.
 
     Parameters
     ----------
     pat : string, default None
-        String or regular expression to split on. If None, splits on whitespace
+        String or regular expression to split on. If None, split on whitespace.
```
WillAyd: \`None\`

pandas/core/strings.py
```diff
@@ -1095,24 +1095,48 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
     n : int, default -1 (all)
-        None, 0 and -1 will be interpreted as return all splits
+        * None, 0 and -1 will be interpreted as return all splits.
```
WillAyd: Shouldn't need periods at the end of bullet points

pandas/core/strings.py
```diff
@@ -1095,24 +1095,48 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
+        * Vary output dimensionality if `expand` is True:
```
WillAyd: \`True\`

pandas/core/strings.py
```diff
@@ -1095,24 +1095,48 @@ def str_pad(arr, width, side='left', fillchar=' '):
 
 def str_split(arr, pat=None, n=None):
     """
+    Split strings around given separator/delimiter.
+
     Split each string (a la re.split) in the Series/Index by given
     pattern, propagating NA values. Equivalent to :meth:`str.split`.
```
WillAyd: \`NaN\` instead of NA would be better

pandas/core/strings.py
```diff
@@ -1095,24 +1095,48 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
+         Appends None for padding.
         * If False, return Series/Index.
 
-    return_type : deprecated, use `expand`
-
     Returns
     -------
     split : Series/Index or DataFrame/MultiIndex of objects
+
+    Examples
```
WillAyd: The examples are good but I think the section can use a short sentence or two to introduce the examples and clue the reader in on what they should be looking at. Also, is it possible to show an example that deals with missing values?

pandas/core/strings.py
```diff
@@ -1095,24 +1095,48 @@ def str_pad(arr, width, side='left', fillchar=' '):
 
 def str_split(arr, pat=None, n=None):
     """
+    Split strings around given separator/delimiter.
+
     Split each string (a la re.split) in the Series/Index by given
     pattern, propagating NA values. Equivalent to :meth:`str.split`.
 
     Parameters
```
WillAyd: General comment - rather than putting sub-bullets under the parameters would it be clearer to move those into a dedicated Notes section? They do explain some of the implementation details so wonder if they are better served there

pandas/core/strings.py
```diff
@@ -1095,24 +1095,48 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
-        * If True, return DataFrame/MultiIndex expanding dimensionality.
+        * If True, return DataFrame/MultiIndex expanding dimensionality.\
+         Appends None for padding.
         * If False, return Series/Index.
 
-    return_type : deprecated, use `expand`
-
     Returns
     -------
     split : Series/Index or DataFrame/MultiIndex of objects
```
WillAyd: I find this return type rather confusing - how does the Index play a part? I think this could be clarified better

WillAyd (changes requested):

pandas/core/strings.py
mananpal1997: It's already in the extended summary.
```
Split each string (a la re.split) in the Series/Index by given pattern, propagating NA values. Equivalent to :meth:`str.split`.
```

pandas/core/strings.py
mananpal1997: That would be better. I'll update it.

pandas/core/strings.py
mananpal1997: Yes, I was confused about that as well. Should I keep it in the doc?

pandas/core/strings.py
WillAyd: Right I just mean actually surrounding the reference with backticks so it renders as inline code. So simply change any occurrence of re.split to \`re.split\`

pandas/core/strings.py
WillAyd: Add a reference to this in a See Also section

pandas/core/strings.py
```diff
@@ -1095,24 +1095,48 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
     pat : string, default None
-        String or regular expression to split on. If None, splits on whitespace
+        String or regular expression to split on. If None, split on whitespace.
     n : int, default -1 (all)
-        None, 0 and -1 will be interpreted as return all splits
+        * None, 0 and -1 will be interpreted as return all splits.
+        * Vary output dimensionality if `expand` is True:
+            - If n >= default splits, return all splits.
+            - If n < default splits, makes first n splits only.
     expand : bool, default False
```
WillAyd: Looking at this in more detail, I see why you have this documented here in spite of the fact that `str_split` doesn't actually accept this parameter (it's docstring is copied to `split`). Ref some of the work in https://github.com/pandas-dev/pandas/pull/10085

@jreback is there any reason why we wouldn't want to deprecate items like `str_split` and move to private internal methods like `_str_split` that the front-end facing API uses instead? If so may be a logical follow up to this docstring update

pandas/core/strings.py
```diff
@@ -1095,24 +1095,48 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
-    return_type : deprecated, use `expand`
-
     Returns
     -------
     split : Series/Index or DataFrame/MultiIndex of objects
+
+    Examples
+    --------
+    >>> s = pd.Series(["this is good text", "but this is even better"])
+    >>> s.str.split()
```
WillAyd: xref my other comment - kind of strange that the method here is actually `str_split` but we are showing how to use `str.split`. While technically the same, from an API perspective I feel like this docstring belongs under the `split` method and we should make this method private internal.

pandas/core/strings.py
WillAyd: For returns make sure the first line is just the type (unless returning more than one, which isn't the case here). I think that would best be `Series or Index` with a description that says something like "Return type matches caller unless `expand=True` (always returns a `DataFrame`)"

WillAyd (changes requested):

pandas/core/strings.py
mananpal1997: reference to re.split?

pandas/core/strings.py
mananpal1997: same is the case for `str.rsplit`

WillAyd: Yep thanks - might be a few other instances in the module as well. I don't want it to hold up what you've done here but could be a good follow up for you to clean up the actual functions

pandas/core/strings.py
WillAyd: Sorry you can ignore this. I was thinking of linking to `str.split` since you mention it as equivalent here, but it's the same exact docstring as noted in other comments

pandas/core/strings.py
mananpal1997: Should I work on it in this same pr, or should I make a separate issue for this and work on that?

WillAyd: Separate issue

mananpal1997 (commented):
Does everything seem fine now?

pandas/core/strings.py
```diff
@@ -1095,24 +1095,61 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
-    Split each string (a la re.split) in the Series/Index by given
-    pattern, propagating NA values. Equivalent to :meth:`str.split`.
+    Split strings around given separator/delimiter.
+
+    Split each string in the Series/Index by given
+    pattern, propagating NaN values. Equivalent to :meth:`str.split`.
 
     Parameters
     ----------
     pat : string, default None
```
WillAyd: Use `str` instead of `string`

pandas/core/strings.py
```diff
@@ -1095,24 +1095,61 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
-        String or regular expression to split on. If None, splits on whitespace
+        String or regular expression to split on.\
```
WillAyd: Hmm does this render the backslash in the output?

pandas/core/strings.py
```diff
@@ -1095,24 +1095,61 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
     split : Series/Index or DataFrame/MultiIndex of objects
+
+    Notes
+    -----
+    If `expand` parameter is `True` and:
+      - If n >= default splits, makes all splits
+      - If n < default splits, makes first n splits only
+      - Appends `None` for padding.
+
+    Examples
```
WillAyd: Might have missed mentioning this but while I think the examples are good you should add a sentence (or a few) to call out what users should be looking at with the examples. It would also be nice to show one or two with missing data to illustrate how `NaN` gets propagated

pandas/core/strings.py
```diff
@@ -1095,24 +1095,61 @@ def str_pad(arr, width, side='left', fillchar=' '):
 
 def str_split(arr, pat=None, n=None):
     """
-    Split each string (a la re.split) in the Series/Index by given
-    pattern, propagating NA values. Equivalent to :meth:`str.split`.
+    Split strings around given separator/delimiter.
+
+    Split each string in the Series/Index by given
```
WillAyd: I think generally writing `Series/Index` and `DataFrame/MultiIndex` is not very clear. I'd suggest saying "in the caller's values"

WillAyd (changes requested):
Still a few minor things but it's getting there

pandas/core/strings.py
WillAyd: Oh OK good to know. That said, I don't think the line real is required or part of the standard (see the **kwargs example in first instance of `class Series` in the sprint documentation <https://python-sprints.github.io/pandas/guide/pandas_docstring.html>)

Should be fine just to place on separate line and ensure proper indentation

pandas/core/strings.py
mananpal1997: Like this?
```
>>> s.str.split("is", n=1, expand=True)
        0                1
0      th     is good text
1  but th   is even better
see notes about expand=True
```
Also, I wrote an example to show NaN propagation but None is being propagated instead of NaN.
```
>>> s = pd.Series(["this is good text", "but this is even better", np.nan])
>>> s.str.split(expand=True)
      0     1     2     3       4
0  this    is  good  text    None
1   but  this    is  even  better
2   NaN  None  None  None    None
```

WillAyd: Put your comments before the example and just highlight what the user should look at. So for your first example say something like "By default, split will return an object of the same size containing lists to hold the split elements" and then introduce the second with something like "By contrast, when using `expand=True` the split elements will expand out into separate columns." Doesn't need to be exactly those words but something along those lines - make sense?

As far as your example is concerned, make sure you run everything on the master branch. My guess is you are using an older version of pandas as the fix to propagate `NaN` was released in v0.21.1 (see https://github.com/pandas-dev/pandas/pull/18462)

mananpal1997: ah, right.
thanks!

pandas/core/strings.py
jorisvandenbossche: @WillAyd small feedback: it's either no backticks or double backticks for small code snippets, single backticks are for parameter names. 
The question for such short things like None if we see them as 'code', but I think typically we do if it references a value that is passed.

(we should probably make a better overview of this in the docstring guidelines)

pandas/core/strings.py
jorisvandenbossche: @WillAyd the `str_split` methods are not considered public, it's just how it is implemented right now. I agree this one is a bit confusing as some keyword are used in `str_split` and some only in the method, but generally in the file, the docstrings are located at the functions (so `str_split`), although they are written as if they were documenting the method.

Feel free to open an issue if you have a proposal how this could be improved.

pandas/core/strings.py
```diff
@@ -1095,24 +1095,83 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
     ----------
     pat : string, default None
-        String or regular expression to split on. If None, splits on whitespace
+        String or regular expression to split on.
+        If `None`, split on whitespace.
     n : int, default -1 (all)
-        None, 0 and -1 will be interpreted as return all splits
+        Vary dimensionality of output.
+
+        * `None`, 0 and -1 will be interpreted as return all splits
```
jorisvandenbossche: I would not put this in a bullet point, as this breaks the flow (i.e. need to have a blank line for the list to render..). Lists with one bullet point are a bit strange anyhow

pandas/core/strings.py
```diff
@@ -1095,24 +1095,83 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
     expand : bool, default False
-        * If True, return DataFrame/MultiIndex expanding dimensionality.
-        * If False, return Series/Index.
+        Expand the split strings into separate columns.
```
jorisvandenbossche: split -> splitted

pandas/core/strings.py
```diff
@@ -1095,24 +1095,83 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
 
-    return_type : deprecated, use `expand`
+        * If `True`, return DataFrame/MultiIndex expanding dimensionality.
+        * If `False`, return Series/Index.
```
jorisvandenbossche: maybe add "containing lists of strings"

pandas/core/strings.py
```diff
@@ -1095,24 +1095,83 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
 
     Returns
     -------
+    Type matches caller unless `expand=True` (return type is `DataFrame`)
```
jorisvandenbossche: return type for expand=True can also be MultiIndex I think

pandas/core/strings.py
jorisvandenbossche: and ", default None" -> ", optional"

pandas/core/strings.py
```diff
@@ -1095,24 +1095,83 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
+          0     1     2     3       4
+    0  this    is  good  text    None
+    1   but  this    is  even  better
+    >>> s.str.split(" is ", expand=True)
+              0            1
+    0      this    good text
+    1  but this  even better
+
+    Parameter `n` can be used to limit the number of columns in
+    expansion of output.
```
jorisvandenbossche: I would say here to "limit the number of splits". Of course that directly maps to the number of columns, but for `n=1` you actually have two columns, which might be confusing

pandas/core/strings.py
```diff
@@ -1095,24 +1095,83 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
 
     Returns
     -------
+    Type matches caller unless `expand=True` (return type is `DataFrame`)
     split : Series/Index or DataFrame/MultiIndex of objects
+
+    Notes
+    -----
+    If `expand` parameter is `True` and:
+      - If n >= default splits, makes all splits
```
jorisvandenbossche: I think this is also true when expand=False ? (you just get less elements in the list)

jorisvandenbossche (commented):
Nice improvements to the docstring! Added some more comments

pandas/core/strings.py
mananpal1997: my bad. I'll correct that

jorisvandenbossche:
@mananpal1997 general comment for future reference: can you add commits when updating instead of squashing it into one commit? That makes it a bit easier to see what has changed after a review

pandas/core/strings.py
```diff
@@ -1095,24 +1095,85 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
+        Expand the splitted strings into separate columns.
 
-    return_type : deprecated, use `expand`
+        * If `True`, return DataFrame/MultiIndex expanding dimensionality.
+        * If `False`, return Series/Index, containing lists of strings.
 
     Returns
     -------
+    Type matches caller unless `expand=True` (return type is `DataFrame` or
+    `MultiIndex`)
```
jorisvandenbossche: Can you put this one below the line below?
Like

```
split : Series/Index or DataFrame/MultiIndex
    Type matches caller unless `expand=True` (return type is `DataFrame` or
    `MultiIndex`)
```

mananpal1997:
@jorisvandenbossche yeah sure :sweat_smile:

jorisvandenbossche:
@mananpal1997 I added a small commit with some edits regarding the usage of quotes (due to some in-clarities in the guidelines)

pandas/core/strings.py
```diff
@@ -1095,24 +1095,85 @@ def str_pad(arr, width, side='left', fillchar=' '):
...
     -------
     split : Series/Index or DataFrame/MultiIndex of objects
+        Type matches caller unless ``expand=True`` (return type is DataFrame or
+    MultiIndex)
+
+    Notes
+    -----
+    - If n >= default splits, makes all splits
+    - If n < default splits, makes first n splits only
+    - Appends `None` for padding if ``expand=True``
```
jorisvandenbossche: One final comment: I find this list not fully clear. What is 'n' and what is 'default splits' ?

I suppose it details how `n` is handles depending on whether the number of found splits is bigger/smaller than the specified value for `n` ?

Proposal (but not fully sure this is what you meant):

````
The handling of the `n` keyword depends on the number of found splits:

- If found splits > `n`,  make first `n` splits only
- If found splits <= `n`, make all splits
- If for a certain row the number of found splits < `n`, append `None` for padding up to `n` if ``expand=True``
```

mananpal1997: right!

jorisvandenbossche: Yes, nbsphinx is missing from the dev requirements, they are only included in the optional ones. Will open an issues about that.

jorisvandenbossche (approved):

jorisvandenbossche:
@mananpal1997 Thanks a lot!

pandas/core/strings.py
WillAyd: I think the wording around `Index` and `MultiIndex` is confusing here - how does this return a `MultiIndex`?

jorisvandenbossche: Ah, sorry @WillAyd, merged maybe a bit too quick

When you use this on an index, and do `expand=True`, you get a MultiIndex (similar as if Series -> DataFrame). 

Maybe it would also be good to add an example of this. 
@mananpal1997 Feel free to open a new PR to do this small follow-up change.

WillAyd: Ah OK makes sense. Yes I just find the text like "Series/Index" and "DataFrame/MultiIndex" to be rather confusing - part of me wants to interpret those as being bound together, so that a Series is always indexed normally and a DataFrame is indexed with a MultiIndex. Obviously that's not the case, but I think that we could more clearly delineate.

@mananpal1997 an example would certainly help for your next contribution!

mananpal1997: I'll add it :+1:

mananpal1997: @WillAyd @jorisvandenbossche 
Any edit suggestions to this change before I push it?
```
Parameter ``expand=True`` returns DataFrame and MultiIndex objects for
    Series and Index objects respectively.

    >>> i = pd.Index(["ba 100 001", "ba 101 002", "ba 102 003"])
    >>> type(s.str.split(expand=True))
    >>> type(i.str.split(expand=True))
    <class 'pandas.core.indexes.multi.MultiIndex'>
```

jorisvandenbossche: You can ping us both in the new PR, so let's further discss there. 
But I think I would just show the actual result, instead of the types, it will be clear from the result what the different types are

PR #20284, opened 2018-03-11
DOC/TST: Added NORMALIZE_WHITESPACE flag

TomAugspurger:
I think we want this to be always on. You can disable it with

```
doctest: -NORMALIZE_WHITESPACE
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,3 +32,4 @@ markers =
     slow: mark a test as slow
     network: mark a test as network
     high_memory: mark a test as a high-memory only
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
In the validation script we also set `IGNORE_EXCEPTION_DETAIL`, and have this in https://github.com/pandas-dev/pandas/pull/19952 as well (but give that's still some time before that will be merged, fine to add it already in this PR)

TomAugspurger:
Added `IGNORE_EXCEPTION_DETAIL` so we're consistent.

PR #20287, opened 2018-03-11
TST: adding join_types fixture

almaleksia:
follow up for PR #20059

Extracting join types to a fixture

- [ ] tests added / passed

### DIFF
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -89,3 +89,11 @@ def compression_no_zip(request):
 def datetime_tz_utc():
     from datetime import timezone
     return timezone.utc
+
+
+@pytest.fixture(params=['inner', 'outer', 'left', 'right'])
+def join_type(request):
+    """
+    Fixture for trying all types of join operations
+    """
+    return request.param
--- a/pandas/tests/indexes/test_multi.py
+++ b/pandas/tests/indexes/test_multi.py
@@ -450,11 +450,11 @@ class TestMultiIndex(Base):
 
         # Make sure label setting works too
         labels2 = [[0, 0, 0, 0, 0, 0], [0, 0, 0, 0, 0, 0]]
-        exp_values = np.empty((6, ), dtype=object)
+        exp_values = np.empty((6,), dtype=object)
         exp_values[:] = [(long(1), 'a')] * 6
 
         # Must be 1d array of tuples
-        assert exp_values.shape == (6, )
+        assert exp_values.shape == (6,)
         new_values = mi2.set_labels(labels2).values
 
         # Not inplace shouldn't change
@@ -583,7 +583,7 @@ class TestMultiIndex(Base):
 
     def test_constructor_no_levels(self):
         tm.assert_raises_regex(ValueError, "non-zero number "
-                               "of levels/labels",
+                                           "of levels/labels",
                                MultiIndex, levels=[], labels=[])
         both_re = re.compile('Must pass both levels and labels')
         with tm.assert_raises_regex(TypeError, both_re):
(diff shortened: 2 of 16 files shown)


pandas/conftest.py
```diff
@@ -99,3 +99,17 @@ def compression_no_zip(request):
 def datetime_tz_utc():
     from datetime import timezone
     return timezone.utc
+
+
+JOIN_TYPES = ['inner', 'outer', 'left', 'right']
+
+
+# workaround for parametrized tests
```
jreback: why is this a work-around?

pandas/tests/reshape/merge/test_join.py
```diff
@@ -297,7 +297,7 @@ def test_join_on_series_buglet(self):
                               'b': [2, 2]}, index=df.index)
         tm.assert_frame_equal(result, expected)
 
-    def test_join_index_mixed(self):
+    def test_join_index_mixed(self, join_types):
```
jreback: so I would still parameterize this with join_type (not the plural)

pandas/conftest.py
```diff
@@ -99,3 +99,17 @@ def compression_no_zip(request):
...
+@pytest.fixture(params=JOIN_TYPES)
+def join_type(request):
+    return request.param
+
+
+@pytest.fixture(scope='module')
```
jreback: I don't really want to have 2 fixtures like this, just join_type is fine.

pandas/conftest.py
jreback: can you give a nice doc-string to this, it shows up when we list fixtures

jreback (changes requested):

pandas/conftest.py
almaleksia: @jreback Sure.
Workaround because I cannot use join_types fixture in parametrize, there is a proposal for this in pytest but still not implemented. I will remove it to avoid confusion :)

pandas/tests/indexes/test_multi.py
```diff
@@ -2829,7 +2823,7 @@ def test_remove_unused_nan(self, level0, level1):
         result = mi.remove_unused_levels()
         tm.assert_index_equal(result, mi)
         for level in 0, 1:
-            assert('unused' not in result.levels[level])
+            assert ('unused' not in result.levels[level])
```
jreback: can you remove this paren

pandas/tests/indexes/test_multi.py
```diff
@@ -2435,7 +2429,7 @@ def check(nlevels, with_nulls):
     def test_duplicate_meta_data(self):
         # GH 10115
         index = MultiIndex(levels=[[0, 1], [0, 1, 2]], labels=[
-                           [0, 0, 0, 0, 1, 1, 1], [0, 1, 2, 0, 0, 1, 2]])
```
jreback: put labels on the next line then this will line up better

pandas/tests/reshape/merge/test_merge.py
```diff
@@ -431,11 +424,11 @@ def test_merge_nosort(self):
              "var2": np.random.randint(0, 10, size=10),
              "var3": [datetime(2012, 1, 12), datetime(2011, 2, 4),
                       datetime(
```
jreback: you can put *all* of the datetime on separate lines to avoid this kind of splitting

jreback (changes requested):
looks really good! just some minor whitespace / lint things. ping on green.

jreback:
thanks @almaleksia keep em coming!

if you are looking for things........want to make tz a top-level fixture (prob need to name it something different though as the name tz is pretty common), e.g.

https://github.com/pandas-dev/pandas/blob/master/pandas/tests/indexes/datetimes/test_timezones.py#L244 (and many other very similar lines of code).

almaleksia:
@jreback sure, thank you for the feedback. I'll make PR. I also wanted to take splitting pandas/tests/frame/test_indexing.py  :D

PR #20290, opened 2018-03-12
DOC: update the DataFrame.at[] docstring

akosel:
an unrelated PR):

```
####################### Docstring (pandas.DataFrame.at)  #######################

Access a single value for a row/column label pair.

Similar to ``loc``, in that both provide label-based lookups. Use
``at`` if you only need to get or set a single value in a DataFrame
or Series.

See Also
--------
DataFrame.iat : Access a single value for a row/column pair by integer position
DataFrame.loc : Access a group of rows and columns by label(s)
Series.at : Access a single value using a label

Examples
--------
>>> df = pd.DataFrame([[0, 2, 3], [0, 4, 1], [10, 20, 30]],
...                   index=[4, 5, 6], columns=['A', 'B', 'C'])
>>> df
    A   B   C
4   0   2   3
5   0   4   1
6  10  20  30

Get value at specified row/column pair

>>> df.at[4, 'B']
2

Set value at specified row/column pair

>>> df.at[4, 'B'] = 10
>>> df.at[4, 'B']
10

Get value within a series

>>> df.loc[5].at['B']
4

Raises
------
KeyError
    When label does not exist in DataFrame

Errors found:
	No returns section found

```

explicitly.

### DIFF
--- a/pandas/core/indexing.py
+++ b/pandas/core/indexing.py
@@ -1888,11 +1888,49 @@ class _ScalarAccessIndexer(_NDFrameIndexer):
 
 
 class _AtIndexer(_ScalarAccessIndexer):
-    """Fast label-based scalar accessor
+    """
+    Access a single value for a row/column label pair.
 
-    Similarly to ``loc``, ``at`` provides **label** based scalar lookups.
-    You can also set using these indexers.
+    Similar to ``loc``, in that both provide label-based lookups. Use
+    ``at`` if you only need to get or set a single value in a DataFrame
+    or Series.
 
+    See Also
+    --------
+    DataFrame.iat : Access a single value for a row/column pair by integer position
+    DataFrame.loc : Access a group of rows and columns by label(s)
+    Series.at : Access a single value using a label
+
+    Examples
+    --------
+    >>> df = pd.DataFrame([[0, 2, 3], [0, 4, 1], [10, 20, 30]],
+    ...                   index=[4, 5, 6], columns=['A', 'B', 'C'])
+    >>> df
+        A   B   C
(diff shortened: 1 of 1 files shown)


pandas/core/indexing.py
```diff
@@ -1888,11 +1888,52 @@ def __setitem__(self, key, value):
...
+    """
+    Access a single value for a row/column label pair.
+
+    Similar to ``loc``, in that both provide label-based lookups. Use
+    ``at`` if you only need to get or set a single value in a DataFrame
+    or Series.
+
+    See Also
+    --------
+    DataFrame.iat
```
WillAyd: For the See Also section the format should be `item : description` so move to the description up to the same line and just break where length exceeds 75 characters

pandas/core/indexing.py
```diff
@@ -1888,11 +1888,52 @@ def __setitem__(self, key, value):
...
+    >>> df.at[4, 'B']
+    2
+
+    Set value at specified row/column pair
+
+    >>> df.at[4, 'B'] = 10
+    >>> df.at[4, 'B']
+    10
+
+    Get value within a series
```
WillAyd: Capitalize Series

WillAyd (changes requested):
Very minor edits otherwise looks good

jorisvandenbossche (commented):
Apart from the comments of @WillAyd, looks good!

akosel:
Thank you for the comments! I believe I have addressed them all, but please let me know if there's anything else that can be improved.

jorisvandenbossche (approved):

PR #20293, opened 2018-03-12
Clean / Consolidate pandas/tests/io/test_html.py

WillAyd:
This is an **extremely** aggressive change to get all of the test cases unified between the LXML and BS4 parsers. On the plus side both parsers can share the same set of tests and functionality, but on the downside it gives lxml a little more power than it had previously, where it would quickly fall back to bs4 for malformed sites. 

Review / criticism appreciated


pandas/io/html.py
```diff
@@ -554,8 +553,7 @@ def _parse_td(self, row):
         return row.xpath('.//td|.//th')
 
     def _parse_tr(self, table):
-        expr = './/tr[normalize-space()]'
```
WillAyd: It wasn't clear to me why `normalize-space()` was added here. It is inconsistent with how bs4 parses `tr` elements and was actually causing a failure in `test_computer_sales_page`.

Should the need to strip trailing / leading whitespace come back up I think it would be better done in the base class than only implementing here in lxml

pandas/io/html.py
```diff
@@ -606,18 +604,20 @@ def _build_doc(self):
...
-
-        parser = HTMLParser(recover=False, encoding=self.encoding)
+        parser = HTMLParser(recover=True, encoding=self.encoding)
 
         try:
-            # try to parse the input in the simplest way
-            r = parse(self.io, parser=parser)
-
+            if _is_url(self.io):
+                with urlopen(self.io) as f:
```
WillAyd: This conditional is required for Py27 compat - in Python3 you can simply provide a call to urlopen on self.io directly as an argument to parse (i.e. without explicitly using the context manager)

pandas/io/html.py
```diff
@@ -627,17 +627,7 @@ def _build_doc(self):
                 except AttributeError:
                     pass
             else:
-                # not a url
-                scheme = parse_url(self.io).scheme
-                if scheme not in _valid_schemes:
-                    # lxml can't parse it
-                    msg = (('{invalid!r} is not a valid url scheme, valid '
```
WillAyd: Rather than creating a custom error message here I re-raised. This makes the behavior consistent between lxml and bs4, allowing the `test_bad_url_protocol` and `test_invalid_url` tests to pass

pandas/tests/io/data/banklist.html
```diff
@@ -340,6 +340,7 @@ <h1 class="page_title">Failed Bank List</h1>
 				<td class="closing">April 19, 2013</td>
 				<td class="updated">April 23, 2013</td>
 			</tr>
+			<tr>
```
WillAyd: I'm not sure if this element was intentionally removed from the test, but the two parsers did take a different path to parsing. bs4 would "insert" the leading `tr` tag while lxml would remove the trailing tag. 

On one hand this challenges if we really want to give lxml as much power in parsing here since most browsers (OK at least Safari and Chrome on my computer) matched the bs4 behavior, but on the other hand I'm not sure if it's generalizable to say that the magical insertion of this or like elements by bs4 would always be desired, and perhaps its just a risk that the user accepts when parsing malformed HTML?

pandas/io/html.py
```diff
@@ -606,18 +604,20 @@ def _build_doc(self):
         """
         from lxml.html import parse, fromstring, HTMLParser
         from lxml.etree import XMLSyntaxError
-
-        parser = HTMLParser(recover=False, encoding=self.encoding)
+        parser = HTMLParser(recover=True, encoding=self.encoding)
```
WillAyd: This is a change in behavior giving lxml a little more power to work through malformed HTML. May or may not be acceptable (see other comments)

pandas/io/html.py
```diff
@@ -657,12 +647,21 @@ def _parse_raw_thead(self, table):
         thead = table.xpath(expr)
         res = []
         if thead:
-            trs = self._parse_tr(thead[0])
-            for tr in trs:
-                cols = [_remove_whitespace(x.text_content()) for x in
-                        self._parse_td(tr)]
+            # Grab any directly descending table headers first
+            ths = thead[0].xpath('./th')
```
WillAyd: Because `_parse_td` with this parser doesn't really differentiate between `td` and `th` elements it was incorrectly parsing headers for things like `spam.html` where `td` and `th` elements are intermixed in the header. Hence to make the parsing more robust and pass the tests, I added an initial search for `th` elements before falling back to the existing behavior.

Even with that I'd argue it's confusing that `_parse_td` is implemented to return `td` and `th` elements and should be refactored to more clearly delineate, but I am trying to minimize behavior change with this PR

pandas/tests/io/test_html.py
```diff
@@ -815,69 +769,7 @@ def test_multiple_header_rows(self):
...
-
-class TestReadHtmlLxml(ReadHtmlMixin):
-    flavor = 'lxml'
-
-    @classmethod
-    def setup_class(cls):
-        _skip_if_no('lxml')
-
+    @pytest.mark.xfail
     def test_data_fail(self):
```
WillAyd: With the changes I made this no longer fails but instead parses. If we are comfortable with that this should probably be removed, though I put as `xfail` for visibility atm

pandas/tests/io/test_html.py
```diff
@@ -962,134 +832,99 @@ def test_displayed_only(self, displayed_only, exp0, exp1):
...
+            def seekable(self):
+                return False
 
-    while helper_thread1.is_alive() or helper_thread2.is_alive():
-        pass
-    assert None is helper_thread1.err is helper_thread2.err
+        bad = UnseekableStringIO('''
+            <table><tr><td>spam<foobr />eggs</td></tr></table>''')
 
+        assert self.read_html(bad)
```
WillAyd: I changed this test to remove a seek back to the start of the file that `lxml` was fine with. By definition of unseekable I'm not sure how that `seek` call was allowed...

pandas/tests/io/test_html.py
```diff
@@ -82,33 +44,56 @@ def assert_framelist_equal(list1, list2, *args, **kwargs):
         assert not frame_i.empty, 'frames are both empty'
 
 
+def _missing_bs4():
+    bs4 = td.safe_import('bs4')
+    if not bs4 or LooseVersion(bs4.__version__) == LooseVersion('4.2.0'):
```
jreback: note that would be ok requiring a later version of bs4 to avoid some of the older issues

WillAyd: Updated the min version. FWIW in `ci/requirements-2.7_COMPAT.pip` the requirement is hardcoded at `4.2.0` and `ci/requirements-2.7_LOCALE.pip` has `4.2.1`. I'm not sure the backstory to those but I assume that will cause conflict

WillAyd: May be rooted back in #4259?

pandas/tests/io/test_html.py
```diff
@@ -82,33 +44,53 @@ def assert_framelist_equal(list1, list2, *args, **kwargs):
         assert not frame_i.empty, 'frames are both empty'
 
 
-def test_bs4_version_fails():
-    _skip_if_none_of(('bs4', 'html5lib'))
+def _missing_bs4():
+    bs4 = td.safe_import('bs4')
+    if not bs4 or LooseVersion(bs4.__version__) == LooseVersion('4.2.0'):
```
jreback: 4.2.1

jreback (changes requested):
can you add a note indicating the new min for bs4, may need to update install.rst, does this have equiv tests to before (lots of remove code is good!)

WillAyd:
Missed your question before but the way this worked is that I combined all of the lxml and bs4 test classes into one, deleting tests where duplicated and also moving in some of the top level tests that weren't being tested for both (ex: `test_importcheck_thread_safety`)

The following were deleted:
-  `test_data_fail` as the changes to the lxml parser to make it more robust didn't trigger this failure
- ` test_bs4_finds_tables`as it wasn't something that could be shared, and also wasn't testing the parser
- `test_lxml_finds_tables` similar reason as above

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -358,13 +358,15 @@ Dependencies have increased minimum versions
...
-+-----------------+-----------------+----------+
-| Package         | Minimum Version | Required |
-+=================+=================+==========+
-| python-dateutil | 2.5.0           |    X     |
-+-----------------+-----------------+----------+
-| openpyxl        | 2.4.0           |          |
-+-----------------+-----------------+----------+
++-----------------+-----------------+----------+---------------+
+| Package         | Minimum Version | Required |     Issue     |
++=================+=================+==========+===============+
```
jreback: nice

ci/requirements-optional-conda.txt
```diff
@@ -1,4 +1,4 @@
-beautifulsoup4
+beautifulsoup4>=4.2.1
```
jreback: I think you need to change this:

```
ci/requirements-2.7_COMPAT.pip:beautifulsoup4==4.2.0
```

WillAyd: Figured as such. Without knowing the full impact, is that req file for a Travis build that tests the 4.2.0 incompatibility (ref #4259). If so do we even need this?

jreback: you can just change it to the new minimum version 4.2.1 (we have 1 other build which is pinned as well). always like to test 1 or 2 builds for the min.

WillAyd:
CircleCI failure was a timeout - don't believe it's related to this change

jreback (approved):

jreback:
thanks

note I saw a single skip on my local

```
pandas/tests/io/test_html.py::TestReadHtml::test_parse_failure_unseekable[lxml] SKIPPED                                                                                                                                            [ 93%]
```

jreback:
```
SKIP [1] /Users/jreback/pandas/pandas/tests/io/test_html.py:842: Not applicable for lxml
```

WillAyd:
Yep there was one test that was testing the failure message when trying to read an unseekable IO object twice. lxml actually didn't have any problem reading it twice, so I placed an imperative skip for that parser in the test

PR #20294, opened 2018-03-12
DOC: Update the Period.day docstring

IHackPy:
# Period.day Docstring

Return the number of days of particular month of a given Period.

This attribute returns the total number of days of given month on which
the particular date occurs.

Returns
-------
Int
Number of days till particular date

See also
--------
Period.dayofweek
    Return the day of the week

Period.dayofyear
    Return the day of the year

Examples
--------
>>> import pandas as pd
>>> p = pd.Period("2018-03-11", freq='H')
>>> p.day
11

Docstring for "pandas.Period.day" correct. :)

### DIFF
--- a/pandas/_libs/tslibs/period.pyx
+++ b/pandas/_libs/tslibs/period.pyx
@@ -1218,27 +1218,20 @@ cdef class _Period(object):
     @property
     def day(self):
         """
-        Return the number of days of particular month of a given Period.
+        Get day of the month that a Period falls on.
 
-        This attribute returns the total number of days of given month on which
-        the particular date occurs.
-        
         Returns
         -------
-        Int
-        Number of days till particular date
+        int
 
-        See also
+        See Also
         --------
-        Period.dayofweek
-            Return the day of the week
+        Period.dayofweek : Get the day of the week
 
-        Period.dayofyear
-            Return the day of the year
+        Period.dayofyear : Get the day of the year
 
         Examples
         --------
-        >>> import pandas as pd
         >>> p = pd.Period("2018-03-11", freq='H')
         >>> p.day
         11
--- a/pandas/_libs/tslibs/period.pyx
+++ b/pandas/_libs/tslibs/period.pyx
@@ -1217,6 +1217,32 @@ cdef class _Period(object):
 
     @property
     def day(self):
+        """
+        Return the number of days of particular month of a given Period.
+
+        This attribute returns the total number of days of given month on which
+        the particular date occurs.
+        
+        Returns
+        -------
+        Int
+        Number of days till particular date
+
+        See also
+        --------
+        Period.dayofweek
+            Return the day of the week
+
+        Period.dayofyear
+            Return the day of the year
+
+        Examples
+        --------
+        >>> import pandas as pd
+        >>> p = pd.Period("2018-03-11", freq='H')
+        >>> p.day
+        11
+        """
         base, mult = get_freq_code(self.freq)
         return pday(self.ordinal, base)
 


pandas/_libs/tslibs/period.pyx
```diff
@@ -1217,6 +1217,32 @@ cdef class _Period(object):
 
     @property
     def day(self):
+        """
+        Return the number of days of particular month of a given Period.
```
WillAyd: This description would make me think this gives the number of days of the month the period falls in, i.e. if the month was February it would return 28. That's obviously note the case - can you think of a way to reword and make clearer?

pandas/_libs/tslibs/period.pyx
```diff
@@ -1217,6 +1217,32 @@ cdef class _Period(object):
...
+
+        This attribute returns the total number of days of given month on which
```
WillAyd: This is a pretty simple function so depending on how concise the short summary is might be OK to not have an extended summary

pandas/_libs/tslibs/period.pyx
```diff
@@ -1217,6 +1217,32 @@ cdef class _Period(object):
...
+        the particular date occurs.
+        
+        Returns
+        -------
+        Int
```
WillAyd: lowercase this

pandas/_libs/tslibs/period.pyx
```diff
@@ -1217,6 +1217,32 @@ cdef class _Period(object):
...
+        Number of days till particular date
```
WillAyd: This should be indented. Might want to reword based off of what you do with short summary

pandas/_libs/tslibs/period.pyx
```diff
@@ -1217,6 +1217,32 @@ cdef class _Period(object):
...
+
+        See also
+        --------
+        Period.dayofweek
+            Return the day of the week
```
WillAyd: In this section the description should be on the same line as the referenced item separated by a " : "

pandas/_libs/tslibs/period.pyx
```diff
@@ -1217,6 +1217,32 @@ cdef class _Period(object):
...
+
+        Period.dayofyear
+            Return the day of the year
+
+        Examples
+        --------
+        >>> import pandas as pd
```
WillAyd: No need for `import pandas as pd` for any docstrings as it is implied (nor `import numpy as np`)

pandas/_libs/tslibs/period.pyx
```diff
@@ -1217,6 +1217,32 @@ cdef class _Period(object):
...
+
+        This attribute returns the total number of days of given month on which
+        the particular date occurs.
+        
+        Returns
+        -------
+        Int
+        Number of days till particular date
+
+        See also
```
WillAyd: Capitalize Also

WillAyd (changes requested):

pandas/_libs/tslibs/period.pyx
HimanshuAwasthi95: sure I try to make this more clear I'm still thinking that summary is not good sense .
I have query : if we give any period like 2018-03-28 it return 28 so how i explain because I think it count all days till given date ? 
If I think wrong correct me :)

pandas/_libs/tslibs/period.pyx
HimanshuAwasthi95: but I think when I remove extended summary validation_docstring.py gives error ???

pandas/_libs/tslibs/period.pyx
HimanshuAwasthi95: still have confusion  ?

pandas/_libs/tslibs/period.pyx
HimanshuAwasthi95: mean ===>>>>(Period.dayofweek: Return the day of the week)

pandas/_libs/tslibs/period.pyx
WillAyd: I would say "Get day of the month that a Period falls on"

pandas/_libs/tslibs/period.pyx
HimanshuAwasthi95: sorry my mistake

pandas/_libs/tslibs/period.pyx
WillAyd: You are right that the script would fail but I think this would pass as an exception

pandas/_libs/tslibs/period.pyx
WillAyd: Put a space before the colon and change the word "Return" to "Get" and yes you are correct

pandas/_libs/tslibs/period.pyx
WillAyd: Take a look at the contributing guide for the sprint. The very first example should show you how to format the Returns section

https://python-sprints.github.io/pandas/guide/pandas_docstring.html

HimanshuAwasthi95: can I write :  
int
   day of that month

HimanshuAwasthi95: or day of given month

WillAyd: Just write int here - this is simple enough that you don't need a description

For future reference though when you do put a description on the line below the type it needs to be indented (i.e. 4 spaces ahead of the line above)

HimanshuAwasthi95: sure

pandas/_libs/tslibs/period.pyx
HimanshuAwasthi95: I didn't find better sort explanation as you mention on above comment

jorisvandenbossche (approved):

jorisvandenbossche:
@HimanshuAwasthi95 Thanks for the PR, and @WillAyd thanks for the review!

HimanshuAwasthi95:
Thanks @jorisvandenbossche and @WillAyd

PR #20299, opened 2018-03-12
DOC: update the Index.sort_values docstring

DataOmbudsman:
an unrelated PR):

```
##################### Docstring (pandas.Index.sort_values) #####################

Return a sorted copy of the index.

Return a sorted copy of the index, and optionally return the indices
that sorted the index itself.

Parameters
----------
return_indexer : bool, default False
    Should the indices that would sort the index be returned.
ascending : bool, default True
    Should the index values be sorted in an ascending order.

Returns
-------
sorted_index : pandas.Index
    Sorted copy of the index.
_as : numpy.ndarray, optional
    The indices that the index itself was sorted by.

See Also
--------
pandas.Series.sort_values : Sort values of a Series.
pandas.DataFrame.sort_values : Sort values in a DataFrame.

Examples
--------
>>> idx = pd.Index([10, 100, 1, 1000])
>>> idx
Int64Index([10, 100, 1, 1000], dtype='int64')

Sort values in ascending order (default behavior).

>>> idx.sort_values()
Int64Index([1, 10, 100, 1000], dtype='int64')

Sort values in descending order, and also get the indices `idx` was
sorted by.

>>> idx.sort_values(ascending=False, return_indexer=True)
(Int64Index([1000, 100, 10, 1], dtype='int64'), array([3, 1, 0, 2]))

Docstring for "pandas.Index.sort_values" correct. :)

```

explicitly.

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -2311,7 +2311,46 @@ class Index(IndexOpsMixin, PandasObject):
 
     def sort_values(self, return_indexer=False, ascending=True):
         """
-        Return sorted copy of Index
+        Return a sorted copy of the index.
+
+        Return a sorted copy of the index, and optionally return the indices
+        that sorted the index itself.
+
+        Parameters
+        ----------
+        return_indexer : bool, default False
+            Should the indices that would sort the index be returned.
+        ascending : bool, default True
+            Should the index values be sorted in an ascending order.
+
+        Returns
+        -------
+        sorted_index : pandas.Index
+            Sorted copy of the index.
+        _as : numpy.ndarray, optional
+            The indices that the index itself was sorted by.
+
+        See Also
+        --------
+        pandas.Series.sort_values : Sort values of a Series.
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/base.py
```diff
@@ -2311,7 +2311,46 @@ def asof_locs(self, where, mask):
...
+        return_indexer : bool, default False
+            Should the indices that would sort the index be returned.
+        ascending : bool, default True
+            Should the index values be sorted in an ascending order.
+
+        Returns
+        -------
+        sorted_index : pandas.Index
+            Sorted copy of the index.
+        _as : numpy.ndarray, optional
```
TomAugspurger: I think numpy usually calls this kind of array an `indexer` or `index_array`.

TomAugspurger (commented):
Small change, LGTM otherwise.

pandas/core/indexes/base.py
DataOmbudsman: Sure, that's better

TomAugspurger:
💯 Thanks @DataOmbudsman!

PR #20313, opened 2018-03-12
DOC: update docstring of pandas.Series.add_prefix docstring

astrastefania:
an unrelated PR):

```
##################### Docstring (pandas.Series.add_prefix) #####################

Concatenate prefix string with panel items names.

Parameters
----------
prefix : str
    The string to add before each item name.

Returns
-------
Series
    Original Series with updated item names.

See Also
--------
pandas.Series.add_suffix: Add a suffix string to panel items names.

Examples
--------
>>> s = pd.Series([1,2,3,4])
>>> s
0    1
1    2
2    3
3    4
dtype: int64
>>> s.add_prefix('item_')
item_0    1
item_1    2
item_2    3
item_3    4
dtype: int64

Errors found:
	No extended summary found
```

explicitly.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -2963,21 +2963,23 @@ class NDFrame(PandasObject, SelectionMixin):
 
     def add_prefix(self, prefix):
         """
-        Concatenate prefix string with Series items names.
+        Prefix labels with string `prefix`.
+
+        For Series, the row labels are prefixed. For DataFrame, the column labels are prefixed.
 
         Parameters
         ----------
         prefix : str
-            The string to add before each item name.
+            The string to add before each label.
 
         Returns
         -------
-        Series
-            Original Series with updated item names.
+        Series or DataFrame
+            Original Series or DataFrame with updated labels.
 
         See Also
         --------
-        Series.add_suffix: Add a suffix string to Series items names.
+        Series.add_suffix: Suffix labels with string `suffix`.
 
         Examples
         --------
@@ -2994,6 +2996,20 @@ class NDFrame(PandasObject, SelectionMixin):
         item_2    3
         item_3    4
         dtype: int64
+
+        >>> df = pd.DataFrame({'A': [1, 2, 3, 4],  'B': [3, 4, 5, 6]})
+        >>> df
+           A  B
+        0  1  3
+        1  2  4
+        2  3  5
+        3  4  6
+        >>> df.add_suffix('_item')
+           A_item  B_item
+        0       1       3
+        1       2       4
+        2       3       5
+        3       4       6
         """
         new_data = self._data.add_prefix(prefix)
         return self._constructor(new_data).__finalize__(self)
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -2976,7 +2976,7 @@ class NDFrame(PandasObject, SelectionMixin):
         Returns
         -------
         Series or DataFrame
-            Original Series or DataFrame with updated labels.
+            New Series or DataFrame with updated labels.
 
         See Also
         --------
@@ -3005,8 +3005,8 @@ class NDFrame(PandasObject, SelectionMixin):
         1  2  4
         2  3  5
         3  4  6
-        >>> df.add_suffix('_item')
-           A_item  B_item
+        >>> df.add_prefix('col_')
+           col_A  col_B
         0       1       3
         1       2       4
         2       3       5
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -2977,7 +2977,7 @@ class NDFrame(PandasObject, SelectionMixin):
 
         See Also
         --------
-        Series.add_suffix: Add a suffix string to panel items names.
+        Series.add_suffix: Add a suffix string to Series items names.
 
         Examples
         --------
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -2965,7 +2965,7 @@ class NDFrame(PandasObject, SelectionMixin):
         """
         Prefix labels with string `prefix`.
 
-        For Series, the row labels are prefixed.  
+        For Series, the row labels are prefixed.
         For DataFrame, the column labels are prefixed.
 
         Parameters
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -2965,7 +2965,8 @@ class NDFrame(PandasObject, SelectionMixin):
         """
         Prefix labels with string `prefix`.
 
-        For Series, the row labels are prefixed. For DataFrame, the column labels are prefixed.
+        For Series, the row labels are prefixed.  
+        For DataFrame, the column labels are prefixed.
 
         Parameters
         ----------
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -2991,6 +2991,7 @@ class NDFrame(PandasObject, SelectionMixin):
         2    3
         3    4
         dtype: int64
+
         >>> s.add_prefix('item_')
         item_0    1
         item_1    2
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -2967,11 +2967,33 @@ def add_prefix(self, prefix):
...
+        Series
+            Original Series with updated item names.
+
+        See Also
+        --------
+        pandas.Series.add_suffix: Add a suffix string to panel items names.
+
+        Examples
+        --------
+        >>> s = pd.Series([1,2,3,4])
```
datapythonista: Missing spaces after commas to pass PEP-8

pandas/core/generic.py
```diff
@@ -2967,11 +2967,33 @@ def add_prefix(self, prefix):
...
 
         Returns
         -------
-        with_prefix : type of caller
+        Series
+            Original Series with updated item names.
+
+        See Also
+        --------
+        pandas.Series.add_suffix: Add a suffix string to panel items names.
```
datapythonista: I think the preferred option is to not user the `pandas.` prefix.

pandas/core/generic.py
```diff
@@ -2967,11 +2967,33 @@ def add_prefix(self, prefix):
```
datapythonista: `Panel` is deprecated, we may want to use `Series or DataFrame`, or something generic.

I'd try to find uses cases on the internet for this method, and add in the extended summary when this method can be useful, if possible.

datapythonista (commented):
Nice PR. Added couple of comments.

pandas/core/generic.py
astrastefania: Thanks! I used Series as it is a Series module in this case.

pandas/core/generic.py
astrastefania: Interesting, the PEP-8 passed with any comments, it could be updated with this case?

pandas/core/generic.py
astrastefania: I took it away. :)

astrastefania:
Docstring updated, thanks for the comments!

pandas/core/generic.py
TomAugspurger: For this and https://github.com/pandas-dev/pandas/pull/20315, these apply to both Series and DataFrame (and Panel, but we don't care about that). So perhaps

    Prefix row labels with a string `prefix`.

I want to avoid `items`, as (to me) I think `dict.items` so key-value pairs. But we're just touching the row labels here.

That summary is strange since we use Prefix as a verb and nound. Maybe "Prepend"  would be better?

pandas/core/generic.py
```diff
@@ -2963,15 +2963,37 @@ def _update_inplace(self, result, verify_is_copy=True):
...
         Parameters
         ----------
-        prefix : string
+        prefix : str
+            The string to add before each item name.
 
         Returns
         -------
-        with_prefix : type of caller
+        Series
```
TomAugspurger:
```
Series or DataFrame
    Same type as the calling object, with updated row labels.

File: pandas/core/generic.py [hunk 5]
```diff
@@ -2963,15 +2963,37 @@ def _update_inplace(self, result, verify_is_copy=True):
...
+            Original Series with updated item names.
+
+        See Also
+        --------
+        Series.add_suffix: Add a suffix string to Series items names.
```
TomAugspurger: "items names" -> "row labels"

pandas/core/generic.py
```diff
@@ -2963,15 +2963,37 @@ def _update_inplace(self, result, verify_is_copy=True):
...
+        1    2
+        2    3
+        3    4
+        dtype: int64
+        >>> s.add_prefix('item_')
+        item_0    1
+        item_1    2
+        item_2    3
+        item_3    4
+        dtype: int64
```
TomAugspurger: Add a blank line and then an example with `DataFrame`.

TomAugspurger (commented):
Thanks. Do you have time to update this and https://github.com/pandas-dev/pandas/pull/20315 with similar changes @astrastefania?

pandas/core/generic.py
astrastefania: I see, thank you, my doubt is that on DataFrames add_prefix() adds a prefix on columns names not on rows, for example: `df = pd.DataFrame({'A':[1,2,3,4], 'B':[3,4,5,6]}), df.add_suffix('_item')`. For this motivations I decided to leave on Series. 

Perhaps I could updated the docstring for pandas.DataFrame.add_prefix with the relative example to avoid confusion and let this as it is. 

Let me know what sound better. :)

pandas/core/generic.py
astrastefania: I see your point for "item names", please check the comment above related to include DataFrame or not (if not could make the change to "row labels" suggested here).

pandas/core/generic.py
TomAugspurger: Ahh, good catch! I would say splitting the docstring is more complexity that warranted. How about somehting like

"""
Prefix labels with string `prefix`.

For Series, the row labels are prefixed. For DataFrame, the column labels are prefixed

...
"""

pandas/core/generic.py
TomAugspurger: So anywhere we say "item names", let's say "labels." or "row labels" or "column labels".

pandas/core/generic.py
astrastefania: Yes, I do agree, thanks!

I didn't know about the issue of splitting docstring, so in this case is better have the same for both pandas.Series.add_prefix and pandas.DataFrame.add_prefix?

TomAugspurger: Yes, the docstring you've been editing in `core/generic.py` is used by both `Series.add_prefix` and `DataFrame.add_prefix`.

pandas/core/generic.py
```diff
@@ -2963,15 +2963,54 @@ def _update_inplace(self, result, verify_is_copy=True):
...
         ----------
-        prefix : string
+        prefix : str
+            The string to add before each label.
 
         Returns
         -------
-        with_prefix : type of caller
+        Series or DataFrame
+            Original Series or DataFrame with updated labels.
```
TomAugspurger: This make it sound a *bit* like the object is modified inplace. Could you instead say

```
Series or DataFrame
    New Series or DataFrame with updated labels.
```

pandas/core/generic.py
```diff
@@ -2963,15 +2963,54 @@ def _update_inplace(self, result, verify_is_copy=True):
...
+        dtype: int64
+
+        >>> df = pd.DataFrame({'A': [1, 2, 3, 4],  'B': [3, 4, 5, 6]})
+        >>> df
+           A  B
+        0  1  3
+        1  2  4
+        2  3  5
+        3  4  6
+        >>> df.add_suffix('_item')
```
TomAugspurger: This should be the `add_prefix` example I think :)

TomAugspurger:
LGTM other than those two minor issues.

astrastefania:
@TomAugspurger @datapythonista, if it's all fine in here now I'll proceed to modify #20315 (otherwise I'll update this one first)

pandas/core/generic.py
astrastefania: Yes!!! 😄

pandas/core/generic.py
astrastefania: Actually is modified inplace, should it be specified?

TomAugspurger: Hmm I don't think so

```python
In [6]: s = pd.Series(1, ['a', 'b'])

In [7]: s
Out[7]:
a    1
b    1
dtype: int64

In [8]: s.add_prefix('foo_')
Out[8]:
foo_a    1
foo_b    1
dtype: int64

In [9]: s
Out[9]:
a    1
b    1
dtype: int64
```

`s` is unmodified.

astrastefania: Sorry, I meant it's unmodified! I agree is **not** modified inplace.

TomAugspurger:
Looks great! Just added a tiny commit adjusting the spacing.

PR #20315, opened 2018-03-12
DOC: update the pandas.Series.add_suffix docstring

astrastefania:
an unrelated PR):

```
##################### Docstring (pandas.Series.add_suffix) #####################

Add a suffix string to panel items names.

Parameters
----------
suffix : str
    The string to add after each item name.

Returns
-------
Series
    Original Series with updated item names.

See Also
--------
pandas.Series.add_prefix: Concatenate prefix string with panel items names.

Examples
--------
>>> s = pd.Series([1,2,3,4])
>>> s
0    1
1    2
2    3
3    4
dtype: int64
>>> s.add_suffix('_item')
0_item    1
1_item    2
2_item    3
3_item    4
dtype: int64

Errors found:
	No extended summary found
```

explicitly.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -2978,15 +2978,37 @@ class NDFrame(PandasObject, SelectionMixin):
 
     def add_suffix(self, suffix):
         """
-        Concatenate suffix string with panel items names.
+        Add a suffix string to panel items names.
 
         Parameters
         ----------
-        suffix : string
+        suffix : str
+            The string to add after each item name.
 
         Returns
         -------
-        with_suffix : type of caller
+        Series
+            Original Series with updated item names.
+
+        See Also
+        --------
+        pandas.Series.add_prefix: Concatenate prefix string with panel items names.
+
+        Examples
+        --------
+        >>> s = pd.Series([1,2,3,4])
+        >>> s
+        0    1
+        1    2
+        2    3
+        3    4
+        dtype: int64
+        >>> s.add_suffix('_item')
+        0_item    1
+        1_item    2
+        2_item    3
+        3_item    4
+        dtype: int64
         """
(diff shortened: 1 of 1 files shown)


datapythonista:
Same comments as in #20313 :)

astrastefania:
Docstring fixed. :)

datapythonista (commented):
Didn't check, but as this is defined in `generic.py` it's likely to be reused by both `Series` and `DataFrame`. Can you build the whole documentation and check the page for `DataFrame.add_sufix` please? If it says that it returns a `Series`, and all the other stuff specific to `Series` it should be changed, so both pages are correct.

astrastefania:
@datapythonista, @TomAugspurger mentioned in #20313 this edited docstring in core/generic.py it will be used by both Series.add_suffix and DataFrame.add_suffix to avoid splitting the docstring.

If all agree both #20315 and #20313 should be fine now. 😸

pandas/core/generic.py
```diff
@@ -2978,15 +2978,54 @@ def add_prefix(self, prefix):
...
 
         Returns
         -------
-        with_suffix : type of caller
+        Series or DataFrame
+            New Series or DataFrame with updated labels.
+
+        See Also
+        --------
+        Series.add_prefix: Prefix labels with string `prefix`.
```
datapythonista: As it's for both `Series` and `DataFrame` would make sense to have the `.add_prefix` method of both classes. Or none of them, there is some discussion about whether to have them or not here: https://github.com/pandas-dev/pandas/pull/20289#discussion_r174212593

TomAugspurger: Good call. Added in the latest commit to both.

TomAugspurger (approved):

TomAugspurger:
Thanks again @astrastefania :)

PR #20316, opened 2018-03-12
DOC: update the day_name function docstring

NeehaK:
an unrelated PR):

- [ X] PR title is "DOC: update the <your-function-or-method> docstring"
- [ X] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [ X] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ X] It has been proofread on language by another sprint participant

```
Docstring for "pandas.DatetimeIndex.day_name" correct. :)
```

explicitly.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -2128,19 +2128,28 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
 
     def day_name(self, locale=None):
         """
-        Returns the day names.
+        Return the day names.
         
         Provides day names of the DateTimeIndex for the dataframe which consists of the time and date data with specified locale.
 
         Parameters
         ----------
         locale : string, default None (English locale)
-            locale determining the language in which to return the day name
-
+            The locale determines the language in which the day name should be returned.
+         
         Returns
         -------
         day_names : Index
             Index (instance) of day names
+            
+        See Also
+        --------
+        month_name : Return the month names
+        
+        Examples
+        --------
+        >>> day_name(numpyarr, locale=English) 
+        Index(['Monday', 'Wednesday,...,'Sunday'], dtype='object')
         """
         values = self.asi8
         if self.tz is not None:
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -2128,7 +2128,9 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
 
     def day_name(self, locale=None):
         """
-        Return the day names of the DateTimeIndex with specified locale.
+        Returns the day names.
+        
+        Provides day names of the DateTimeIndex for the dataframe which consists of the time and date data with specified locale.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/datetimes.py
```diff
@@ -2128,19 +2128,34 @@ def month_name(self, locale=None):
...
-            Index of day names
-
-        .. versionadded:: 0.23.0
+        day_names : Index
+            Index (instance) of day names
+            
+        See Also
+        --------
+        month_name : Return the month names
+
```
TomAugspurger: Add `locale` which will link to https://docs.python.org/3/library/locale.html

pandas/core/indexes/datetimes.py
```diff
@@ -2128,19 +2128,34 @@ def month_name(self, locale=None):
 
     def day_name(self, locale=None):
         """
-        Return the day names of the DateTimeIndex with specified locale.
+        Return the day names.
+        
+        Provides day names of the DateTimeIndex for the dataframe which consists of the time and date data with specified locale.
```
TomAugspurger: DatetimeIndex (lowercase t)

TomAugspurger: This section can probably be removed.

pandas/core/indexes/datetimes.py
```diff
@@ -2128,19 +2128,34 @@ def month_name(self, locale=None):
...
 
         Parameters
         ----------
         locale : string, default None (English locale)
-            locale determining the language in which to return the day name
-
+            The locale determines the language in which the day name should be returned.
```
TomAugspurger: This should go just below the `locale` line.

pandas/core/indexes/datetimes.py
```diff
@@ -2128,19 +2128,34 @@ def month_name(self, locale=None):
...
     def day_name(self, locale=None):
         """
-        Return the day names of the DateTimeIndex with specified locale.
+        Return the day names.
+        
+        Provides day names of the DateTimeIndex for the dataframe which consists of the time and date data with specified locale.
 
         Parameters
         ----------
         locale : string, default None (English locale)
```
TomAugspurger:
```
locale : string, optional
```

pandas/core/indexes/datetimes.py
TomAugspurger: Mention that the default is the current locale, which can be obtained with `locale.getlocale()`.

pandas/core/indexes/datetimes.py
```diff
@@ -2128,19 +2128,34 @@ def month_name(self, locale=None):
...
+        day_names : Index
+            Index (instance) of day names
+            
+        See Also
+        --------
+        month_name : Return the month names
+        
+        Examples
+        --------
+        #From the time stamps data based on it the day_name is returned.
```
TomAugspurger: Do you need help making examples? This should be like

```python
>>> idx = pd.date_range('2017', periods=2)
>>> idx
DatetimeIndex(['2017-01-01', '2017-01-02'],
              dtype='datetime64[ns]', freq='D')
>>> idx.day_name()
>>> idx.day_name()
Index(['Sunday', 'Monday'], dtype='object')
```

Then show an example with a different locale like

```python
In [37]: idx.day_name('zh_hk')
Out[37]: Index(['週日', '週一'], dtype='object')
```

mroeschke (commented):
You'll want to keep the `.. versionadded:: 0.23.0` note. This is a new method that will be available in the next release.

datapythonista:
The code for `day_name` changed, and this docstring seems to be defined somewhere else. Closing this PR. Thanks for the contribution @NeehaK.

PR #20319, opened 2018-03-13
DOC : Update the Period.second docstring

IHackPy:
```
# Period.second
############################################################################
####################### Docstring (pandas.Period.second) #######################

Get second of the minute component of the Period.

Returns
-------
int 
    The second as an integer, between 0 and 59.

See Also
--------
Period.hour : Get the hour component of the Period.
Period.minute : Get the minute component of the Period.

Examples
--------
>>> p = pd.Period("2018-03-11 13:03:12.050000")
>>> p.second
12

```

### DIFF
--- a/pandas/_libs/tslibs/period.pyx
+++ b/pandas/_libs/tslibs/period.pyx
@@ -1295,6 +1295,25 @@ cdef class _Period(object):
 
     @property
     def second(self):
+        """
+        Get second of the minute component of the Period.
+
+        Returns
+        -------
+        int 
+            The second as an integer, between 0 and 59.
+
+        See Also
+        --------
+        Period.hour : Get the hour component of the Period.
+        Period.minute : Get the minute component of the Period.
+
+        Examples
+        --------
+        >>> p = pd.Period("2018-03-11 13:03:12.050000")
+        >>> p.second
+        12
+        """
         base, mult = get_freq_code(self.freq)
         return psecond(self.ordinal, base)
 
--- a/pandas/_libs/tslibs/period.pyx
+++ b/pandas/_libs/tslibs/period.pyx
@@ -1296,12 +1296,12 @@ cdef class _Period(object):
     @property
     def second(self):
         """
-        Get second of the minute component of the Period.
+        Get the second component of the Period.
 
         Returns
         -------
         int 
-            The second as an integer, between 0 and 59.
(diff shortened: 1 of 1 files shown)


pandas/_libs/tslibs/period.pyx
```diff
@@ -1295,6 +1295,25 @@ cdef class _Period(object):
 
     @property
     def second(self):
+        """
+        Get second of the minute component of the Period.
```
WillAyd: Say the same thing you are saying in See Also but for second, so literally "Get the second component of the Period"

pandas/_libs/tslibs/period.pyx
```diff
@@ -1295,6 +1295,25 @@ cdef class _Period(object):
...
+
+        Returns
+        -------
+        int 
+            The second as an integer, between 0 and 59.
```
WillAyd: Already say second is an int above so no need to restate here. Maybe better worded as "The second of the Period (ranges from 0 to 59)"

WillAyd (changes requested):

pandas/_libs/tslibs/period.pyx
HimanshuAwasthi95: but period I'm bit confusesd because when I ren code it gives that particular second of that minute have

pandas/_libs/tslibs/period.pyx
HimanshuAwasthi95: yup sure :+1

pandas/_libs/tslibs/period.pyx
WillAyd: Not sure I understand your comment. The below will both return 12, so the minute is irrelevant:

```python
pd.Period("2018-03-11 13:03:12.050000").second
pd.Period("2018-03-11 13:04:12.050000").second
```

WillAyd:
lgtm

pandas/_libs/tslibs/period.pyx
HimanshuAwasthi95: yup I made changes :+1:  thanks for reveiw

HimanshuAwasthi95:
Thanks @WillAyd

TomAugspurger:
Thanks!

HimanshuAwasthi95:
thanks

PR #20329, opened 2018-03-13
DOC: Formatting for DtypeWarning Docstring [ci skip]

TomAugspurger:
xref : https://github.com/pandas-dev/pandas/pull/20208

fireshot capture 002 - pandas errors dtypewarning pandas 0_ - file____users_taugspurger_sandbox_

cc @jorisvandenbossche @hissashirocha

### DIFF
--- a/pandas/errors/__init__.py
+++ b/pandas/errors/__init__.py
@@ -68,8 +68,7 @@ class DtypeWarning(Warning):
     ...                    'b': ['b'] * 300000})
     >>> df.to_csv('test.csv', index=False)
     >>> df2 = pd.read_csv('test.csv')
-
-        DtypeWarning: Columns (0) have mixed types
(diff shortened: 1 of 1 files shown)


TomAugspurger:
Stiill have the issue with the warning going to stdout. Here's the tail of the validation script.

```
No warning was issued.

>>> import os
>>> os.remove('test.csv')
scripts/validate_docstrings.py:268: DtypeWarning: Columns (0) have mixed types. Specify dtype option on import or set low_memory=False.
  runner.run(test)

Errors found:
        No returns section found
```

jorisvandenbossche:
For the validation script I think it is not that much of a problem. But for pytest, I think we don't want to generally capture and silence the warnings, as there might be others raised in docstrings that we want to get notified of.

jorisvandenbossche (approved):

PR #20217, opened 2018-03-10
DOC: update the pandas.DataFrame.any docstring

ghost:
Checklist for the pandas documentation sprint (ignore this if you are doing
an unrelated PR):

- [+] PR title is "DOC: update the <your-function-or-method> docstring"
- [+] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [+] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [+] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [+] It has been proofread on language by another sprint participant

Please include the output of the validation script below between the "```" ticks:

```
################################################################################
####################### Docstring (pandas.DataFrame.any) #######################
################################################################################

Return whether any element is True over requested axis.

Unlike :meth:`DataFrame.all`, this performs an *or* operation. If any of the
values along the specified axis is True, this will return True.

Parameters
----------
axis : int, default 0
    Select the axis which can be 0 for indices and 1 for columns.
skipna : boolean, default True
    Exclude NA/null values. If an entire row/column is NA, the result
    will be NA.
level : int or level name, default None
    If the axis is a MultiIndex (hierarchical), count along a
    particular level, collapsing into a Series.
bool_only : boolean, default None
    Include only boolean columns. If None, will attempt to use everything,
    then use only boolean data. Not implemented for Series.
**kwargs : any, default None
    Additional keywords have no affect but might be accepted for
    compatibility with numpy.

Returns
-------
any : Series or DataFrame (if level specified)

See Also
--------
pandas.DataFrame.all : Return whether all elements are True.

Examples
--------
**Series**

For Series input, the output is a scalar indicating whether any element
is True.

>>> pd.Series([True, False]).any()
True

**DataFrame**

Whether each column contains at least one True element (the default).

>>> pd.DataFrame({
...     "A": [1, 2, 3],
...     "B": [4, 5, 6]
... }).any()
A    True
B    True
dtype: bool

Aggregating over the columns.

>>> pd.DataFrame({
...     "A": [True, False, True],
...     "B": [4, 5, 6]
... }).any(axis='columns')
0    True
1    True
2    True
dtype: bool

>>> pd.DataFrame({
...     "A": [True, False, True],
...     "B": [4, 0, 6]
... }).any(axis='columns')
0    True
1    False
2    True
dtype: bool

`any` for an empty DataFrame is an empty Series.

>>> pd.DataFrame([]).any()
Series([], dtype: bool)

[...]

PR #20336, opened 2018-03-13
DOC: update the pandas.DataFrame.cummax docstring

arminv:
an unrelated PR):

```
##################### Docstring (pandas.DataFrame.cummax)  #####################

Return cumulative maximum over a DataFrame or Series axis.

Returns a DataFrame or Series of the same size containing the cumulative
maximum.

Parameters
----------
    The index or the name of the axis. 0 is equivalent to None or 'index'.
    will be NA.
*args, **kwargs :
    Additional keywords have no effect but might be accepted for
    compatibility with NumPy.

Returns
-------
cummax : Series or DataFrame

Examples
--------
**Series**

>>> s = pd.Series([2, np.nan, 5, -1, 0])
>>> s
0    2.0
1    NaN
2    5.0
3   -1.0
4    0.0
dtype: float64

By default, NA values are ignored.

>>> s.cummax()
0    2.0
1    NaN
2    5.0
3    5.0
4    5.0
dtype: float64

To include NA values in the operation, use ``skipna=False``

>>> s.cummax(skipna=False)
0    2.0
1    NaN
2    NaN
3    NaN
4    NaN
dtype: float64

**DataFrame**

>>> df = pd.DataFrame([[2.0, 1.0],
...                    [3.0, np.nan],
...                    [1.0, 0.0]],
...                    columns=list('AB'))
>>> df
     A    B
0  2.0  1.0
1  3.0  NaN
2  1.0  0.0

By default, iterates over rows and finds the maximum
in each column. This is equivalent to ``axis=None`` or ``axis='index'``.

>>> df.cummax()
     A    B
0  2.0  1.0
1  3.0  NaN
2  3.0  1.0

To iterate over columns and find the maximum in each row,
use ``axis=1``

>>> df.cummax(axis=1)
     A    B
0  2.0  2.0
1  3.0  NaN
2  1.0  1.0

See also
--------
pandas.core.window.Expanding.max : Similar functionality
    but ignores ``NaN`` values.
DataFrame.max : Return the maximum over
    DataFrame axis.
DataFrame.cummax : Return cumulative maximum over DataFrame axis.
DataFrame.cummin : Return cumulative minimum over DataFrame axis.
DataFrame.cumsum : Return cumulative sum over DataFrame axis.
DataFrame.cumprod : Return cumulative product over DataFrame axis.

Errors found:
	Errors in parameters section
		Unknown parameters {'*args, **kwargs :'}
		Parameter "*args, **kwargs :" has no type


##################### Docstring (pandas.DataFrame.cummin)  #####################

Return cumulative minimum over a DataFrame or Series axis.

Returns a DataFrame or Series of the same size containing the cumulative
minimum.

Parameters
----------
    The index or the name of the axis. 0 is equivalent to None or 'index'.
    will be NA.
*args, **kwargs :
    Additional keywords have no effect but might be accepted for
    compatibility with NumPy.

Returns
-------
cummin : Series or DataFrame

Examples
--------
**Series**

>>> s = pd.Series([2, np.nan, 5, -1, 0])
>>> s
0    2.0
1    NaN
2    5.0
3   -1.0
4    0.0
dtype: float64

By default, NA values are ignored.

>>> s.cummin()
0    2.0
1    NaN
2    2.0
3   -1.0
4   -1.0
dtype: float64

To include NA values in the operation, use ``skipna=False``

>>> s.cummin(skipna=False)
0    2.0
1    NaN
2    NaN
3    NaN
4    NaN
dtype: float64

**DataFrame**

>>> df = pd.DataFrame([[2.0, 1.0],
...                    [3.0, np.nan],
...                    [1.0, 0.0]],
...                    columns=list('AB'))
>>> df
     A    B
0  2.0  1.0
1  3.0  NaN
2  1.0  0.0

By default, iterates over rows and finds the minimum
in each column. This is equivalent to ``axis=None`` or ``axis='index'``.

>>> df.cummin()
     A    B
0  2.0  1.0
1  2.0  NaN
2  1.0  0.0

To iterate over columns and find the minimum in each row,
use ``axis=1``

>>> df.cummin(axis=1)
     A    B
0  2.0  1.0
1  3.0  NaN
2  1.0  0.0

See also
--------
pandas.core.window.Expanding.min : Similar functionality
    but ignores ``NaN`` values.
DataFrame.min : Return the minimum over
    DataFrame axis.
DataFrame.cummax : Return cumulative maximum over DataFrame axis.
DataFrame.cummin : Return cumulative minimum over DataFrame axis.
DataFrame.cumsum : Return cumulative sum over DataFrame axis.
DataFrame.cumprod : Return cumulative product over DataFrame axis.

Errors found:
	Errors in parameters section
		Unknown parameters {'*args, **kwargs :'}
		Parameter "*args, **kwargs :" has no type


##################### Docstring (pandas.DataFrame.cumsum)  #####################

Return cumulative sum over a DataFrame or Series axis.

Returns a DataFrame or Series of the same size containing the cumulative
sum.

Parameters
----------
    The index or the name of the axis. 0 is equivalent to None or 'index'.
    will be NA.
*args, **kwargs :
    Additional keywords have no effect but might be accepted for
    compatibility with NumPy.

Returns
-------
cumsum : Series or DataFrame

Examples
--------
**Series**

>>> s = pd.Series([2, np.nan, 5, -1, 0])
>>> s
0    2.0
1    NaN
2    5.0
3   -1.0
4    0.0
dtype: float64

By default, NA values are ignored.

>>> s.cumsum()
0    2.0
1    NaN
2    7.0
3    6.0
4    6.0
dtype: float64

To include NA values in the operation, use ``skipna=False``

>>> s.cumsum(skipna=False)
0    2.0
1    NaN
2    NaN
3    NaN
4    NaN
dtype: float64

**DataFrame**

>>> df = pd.DataFrame([[2.0, 1.0],
...                    [3.0, np.nan],
...                    [1.0, 0.0]],
...                    columns=list('AB'))
>>> df
     A    B
0  2.0  1.0
1  3.0  NaN
2  1.0  0.0

By default, iterates over rows and finds the sum
in each column. This is equivalent to ``axis=None`` or ``axis='index'``.

>>> df.cumsum()
     A    B
0  2.0  1.0
1  5.0  NaN
2  6.0  1.0

To iterate over columns and find the sum in each row,
use ``axis=1``

>>> df.cumsum(axis=1)
     A    B
0  2.0  3.0
1  3.0  NaN
2  1.0  1.0

See also
--------
pandas.core.window.Expanding.sum : Similar functionality
    but ignores ``NaN`` values.
DataFrame.sum : Return the sum over
    DataFrame axis.
DataFrame.cummax : Return cumulative maximum over DataFrame axis.
DataFrame.cummin : Return cumulative minimum over DataFrame axis.
DataFrame.cumsum : Return cumulative sum over DataFrame axis.
DataFrame.cumprod : Return cumulative product over DataFrame axis.

Errors found:
	Errors in parameters section
		Unknown parameters {'*args, **kwargs :'}
		Parameter "*args, **kwargs :" has no type

##################### Docstring (pandas.DataFrame.cumprod) #####################

Return cumulative product over a DataFrame or Series axis.

Returns a DataFrame or Series of the same size containing the cumulative
product.

Parameters
----------
    The index or the name of the axis. 0 is equivalent to None or 'index'.
    will be NA.
*args, **kwargs :
    Additional keywords have no effect but might be accepted for
    compatibility with NumPy.

Returns
-------
cumprod : Series or DataFrame

Examples
--------
**Series**

>>> s = pd.Series([2, np.nan, 5, -1, 0])
>>> s
0    2.0
1    NaN
2    5.0
3   -1.0
4    0.0
dtype: float64

By default, NA values are ignored.

>>> s.cumprod()
0     2.0
1     NaN
2    10.0
3   -10.0
4    -0.0
dtype: float64

To include NA values in the operation, use ``skipna=False``

>>> s.cumprod(skipna=False)
0    2.0
1    NaN
2    NaN
3    NaN
4    NaN
dtype: float64

**DataFrame**

>>> df = pd.DataFrame([[2.0, 1.0],
...                    [3.0, np.nan],
...                    [1.0, 0.0]],
...                    columns=list('AB'))
>>> df
     A    B
0  2.0  1.0
1  3.0  NaN
2  1.0  0.0

By default, iterates over rows and finds the product
in each column. This is equivalent to ``axis=None`` or ``axis='index'``.

>>> df.cumprod()
     A    B
0  2.0  1.0
1  6.0  NaN
2  6.0  0.0

To iterate over columns and find the product in each row,
use ``axis=1``

>>> df.cumprod(axis=1)
     A    B
0  2.0  2.0
1  3.0  NaN
2  1.0  0.0

See also
--------
pandas.core.window.Expanding.prod : Similar functionality
    but ignores ``NaN`` values.
DataFrame.prod : Return the product over
    DataFrame axis.
DataFrame.cummax : Return cumulative maximum over DataFrame axis.
DataFrame.cummin : Return cumulative minimum over DataFrame axis.
DataFrame.cumsum : Return cumulative sum over DataFrame axis.
DataFrame.cumprod : Return cumulative product over DataFrame axis.

Errors found:
	Errors in parameters section
		Unknown parameters {'*args, **kwargs :'}
		Parameter "*args, **kwargs :" has no type
```

explicitly.

The *args/**kwargs error is a bug in the parameter evaluation


pandas/core/generic.py
```diff
@@ -8327,24 +8327,95 @@ def _doc_parms(cls):
...
 
 Parameters
 ----------
 axis : %(axis_descr)s
 skipna : boolean, default True
     Exclude NA/null values. If an entire row/column is NA, the result
-    will be NA
+    will be NA.
+*args : any, default None
+**kwargs : any, default None
```
datapythonista: Can you have `*args, **kwargs` without the type

pandas/core/generic.py
```diff
@@ -8327,24 +8327,95 @@ def _doc_parms(cls):
...
+%(examples)s
 See also
 --------
 pandas.core.window.Expanding.%(accum_func_name)s : Similar functionality
     but ignores ``NaN`` values.
+pandas.Series.%(outname)s : Return %(desc)s over Series axis.
+pandas.DataFrame.cummax : Return cumulative maximum over DataFrame axis.
+pandas.DataFrame.cummin : Return cumulative minimum over DataFrame axis.
+pandas.DataFrame.cumsum : Return cumulative sum over DataFrame axis.
+pandas.DataFrame.cumprod : Return cumulative product over DataFrame axis.
```
datapythonista: Can you get rid of the `pandas.` prefix

pandas/core/generic.py
```diff
@@ -8327,24 +8327,95 @@ def _doc_parms(cls):
 """
 
 _cnum_doc = """
+Return %(desc)s over a DataFrame or Series axis.
+
+Returns a DataFrame or Series of the same size containing the %(desc)s.
 
 Parameters
 ----------
 axis : %(axis_descr)s
```
datapythonista: Not sure were `axis_descr` is defined, but the format it `{0 or 'index', 1 or 'columns'}` if I'm not wrong. You can check recent merge PRs to be sure.

pandas/core/generic.py
```diff
@@ -8327,24 +8327,95 @@ def _doc_parms(cls):
...
+pandas.DataFrame.cumsum : Return cumulative sum over DataFrame axis.
+pandas.DataFrame.cumprod : Return cumulative product over DataFrame axis.
+"""
 
+_cummax_examples = """\
+Examples
+--------
+**DataFrame**
+
+Create a DataFrame:
```
datapythonista: We probably can get rid of this, I'm sure users will find out :)

pandas/core/generic.py
```diff
@@ -8327,24 +8327,95 @@ def _doc_parms(cls):
...
+
+axis=None : Iterates over rows and finds the maximum value in each column.
+If value is larger than the previous maximum, updates it:
+
+>>> df.cummax(axis=None)
+   A  B  C  D
+0  9  7  9  7
+1  9  7  9  7
+2  9  7  9  7
+3  9  7  9  7
```
datapythonista: If this docstring is reused, and we want to keep it this way, I think we should add examples for all methods.

pandas/core/generic.py
```diff
@@ -8327,24 +8327,95 @@ def _doc_parms(cls):
...
+--------
+**DataFrame**
+
+Create a DataFrame:
+
+>>> df = pd.DataFrame([[9, 7, 9, 7],
+...                    [7, 5, 2, 7],
+...                    [3, 5, 2, 2],
+...                    [8, 0, 9, 0]],
+...                    columns=list('ABCD'))
```
datapythonista: I'd use a much smaller dataset, probably 2 columns and 3 or 4 rows. Smaller numbers would make it easier for users to see what's being added, specially if we reuse same dataframe for `cumsum`...

Also, you can add a `NaN`, so you can show an example with `skipna`.

datapythonista (commented):
I think we can do better with the examples, but good work.

arminv:
@datapythonista Thank you for the helpful comments :) Please let me know if I need to change anything else.

datapythonista (commented):
Looking better. Just one comment I think can be useful... When you do more changes, generate the html version of one of the pages, and imagine that you're a user that never used these methods and wants to know what are they about. Is the documentation useful? May be too long? May be a bit confusing having all the methods shown together? Or does it help?

If as a user you think this documentation is really helpful and presents things in the most efficient and clear way, the PR should be mostly all right. The point is that, it's usually a bad practice to follow comments in the review blindly, feel free to disagree, and make sure you're happy with your changes and the final result.

pandas/core/generic.py
```diff
@@ -8732,11 +8906,11 @@ def stat_func(self, axis=None, skipna=None, level=None, ddof=1,
...
 def _make_cum_function(cls, name, name1, name2, axis_descr, desc,
-                       accum_func, accum_func_name, mask_a, mask_b):
+                       accum_func, accum_func_name, mask_a, mask_b, examples):
     @Substitution(outname=name, desc=desc, name1=name1, name2=name2,
-                  axis_descr=axis_descr, accum_func_name=accum_func_name)
-    @Appender("Return {0} over requested axis.".format(desc) +
-              _cnum_doc)
+                  axis_descr=axis_descr, accum_func_name=accum_func_name,
+                  examples=examples)
+    @Appender(_cnum_doc)
```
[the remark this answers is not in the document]
arminv: I think having separate string examples for each method makes everything clearer, especially when showing examples for use of `skipna` & `axis`. It also helps with keeping the docstring concise. For instance, now we can have a `Series` example for each method.

The disadvantage is user will only see examples for the method they’re checking, but I think this is ok because we are referencing all methods in the ‘See also’ section, which comes before 'Examples'.

In these PRs #20216 and #20217 examples for `DataFrame.all` and `DataFrame.any` are separate even though they are similar methods.

pandas/core/generic.py
```diff
@@ -8518,24 +8518,198 @@ def _doc_parms(cls):
...
 
 _cnum_doc = """
+Return %(desc)s over a DataFrame or Series axis.
+
+Returns a DataFrame or Series of the same size containing the %(desc)s.
 
 Parameters
 ----------
-axis : %(axis_descr)s
+axis : {0 or 'index', 1 or 'columns'}, default 0
```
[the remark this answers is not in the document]
arminv: This is right, `cum_func` (i.e. function corresponding to all cumulative methods) is defined with `axis=None` as default argument.

I also found #20289 as a PR with the correct axis parameter.

pandas/core/generic.py
```diff
@@ -8518,24 +8518,198 @@ def _doc_parms(cls):
...
+1  NaN NaN
+2  NaN NaN
+>>> df.cumprod(skipna=False)
+     A   B
+0  7.0 NaN
+1  NaN NaN
+2  NaN NaN
+
+**Series**
+
```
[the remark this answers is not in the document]
arminv: If we decide to have separate string examples for each method, we can keep the examples for `Series`.

pandas/core/generic.py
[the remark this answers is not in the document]
jorisvandenbossche: Although it is technically None, in practice it is 0 for Series/DataFrame, so I would keep the documentation like this. 
The technical reason is because for Panel it is 1, but Panel is deprecated and I think we should not care about them in the documentation.

pandas/core/generic.py
```diff
@@ -8518,24 +8518,198 @@ def _doc_parms(cls):
...
 -------
-%(outname)s : %(name1)s\n
-
-
+%(outname)s : %(name1)s or %(name2)s\n
+%(examples)s
 See also
 --------
-pandas.core.window.Expanding.%(accum_func_name)s : Similar functionality
+core.window.Expanding.%(accum_func_name)s : Similar functionality
```
jorisvandenbossche: sorry to go back and forth, but for this one we want to keep the `pandas.` (but for the others below the change is perfect!)

pandas/core/generic.py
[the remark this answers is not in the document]
jorisvandenbossche: > If we decide to have separate string examples for each method, we can keep the examples for Series.

\+ 1

Another suggestion would be to start with Series to just illustrate the concept of "cumulative max", as this will make the examples a little bit easier, and show the effect of NaNs. And then show DataFrame, saying that by default the same happens for each column of the DataFrame, and optionally use `axis=1` to take cumulative max for each row.

pandas/core/generic.py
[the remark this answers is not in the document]
jorisvandenbossche: Yes, I am also in favor of splitting up the examples.

jorisvandenbossche (commented):
You are currently only adding the example to the `cummax` method, but your example section has examples for both cummin/cumsum/cumprod/cummax. 
So we need to make a choice here. 

I would personally only include in the docstring of `cummax` examples of `cummax`. But no need to throw them away, what you could do is split your existing `_cummax_examples` in multiple strings, with the examples separated for each method.

pandas/core/generic.py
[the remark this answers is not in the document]
arminv: Thanks for the suggestion. I agree, it is definitely easier to see what is going on with NaNs if we use a Series example instead of DataFrame. I will change it this way.

pandas/core/generic.py
```diff
@@ -8518,24 +8518,198 @@ def _doc_parms(cls):
...
     but ignores ``NaN`` values.
+Series.%(outname)s : Return %(desc)s over Series axis.
+DataFrame.cummax : Return cumulative maximum over DataFrame axis.
```
arminv: Is it a good idea to also add: `DataFrame.sum` as a relevant method in the 'See also' section?

jorisvandenbossche: Yes, that might be a good idea, if you can automatically make it link in `DataFrame.cumsum` to `DataFrame.sum`, in `DataFrame.cummin` to `DataFrame.min`, etc (using the templating)

pandas/core/generic.py
```diff
@@ -8680,24 +8682,304 @@ def _doc_parms(cls):
...
+DataFrame.cumsum : Return cumulative sum over DataFrame axis.
+DataFrame.cumprod : Return cumulative product over DataFrame axis.
+"""
+
+_cummin_examples = """\
+Examples
+--------
+**Series**
+
+>>> s = pd.Series([2,np.nan,5,-1,0])
```
TomAugspurger: PEP8, spaces after `,`

pandas/core/generic.py
```diff
@@ -8680,24 +8682,304 @@ def _doc_parms(cls):
...
+>>> s
+0    2.0
+1    NaN
+2    5.0
+3   -1.0
+4    0.0
+dtype: float64
+
+skipna=True : Default value, ignores NaN values during operation:
```
TomAugspurger: I think we mostly use prose for the intermittent text in examples. Perhaps, say

```
By default, NA values are ignored.

>>> s.cummin()
...

To include NA values in the operation, use ``skipna=False``

>>> s.cummin(skipna=False) 
```

pandas/core/generic.py
```diff
@@ -8680,24 +8682,304 @@ def _doc_parms(cls):
...
+...                    [3.0, np.nan],
+...                    [1.0, 0.0]],
+...                    columns=list('AB'))
+>>> df
+     A    B
+0  2.0  1.0
+1  3.0  NaN
+2  1.0  0.0
+
+skipna : Works in the same way as for Series.
```
TomAugspurger: Can probably be removed.

pandas/core/generic.py
```diff
@@ -8680,24 +8682,304 @@ def _doc_parms(cls):
...
+
+axis=0 : Default value, equivalent to axis=None or axis='index'.
```
TomAugspurger: Same comment about the prose.

pandas/core/generic.py
TomAugspurger: To be clear, what you have is great and if the others are OK with it I'm +1 as well. Maybe wait to hear feedback from @datapythonista and @jorisvandenbossche before updating.

pandas/core/generic.py
```diff
@@ -8680,24 +8682,304 @@ def _doc_parms(cls):
...
-
+%(outname)s : %(name1)s or %(name2)s\n
+%(examples)s
 See also
 --------
 pandas.core.window.Expanding.%(accum_func_name)s : Similar functionality
     but ignores ``NaN`` values.
+Series.%(outname)s : Return %(desc)s over Series axis.
+DataFrame.%(accum_func_name)s : Return the %(accum_func_name)s over
+    DataFrame axis.
```
arminv: @jorisvandenbossche I used templating and `%(accum_func_name)s` to add `DataFrame.prod` to 'See also' ection of `DataFrame.cumprod`, etc. All methods look good except `DataFrame.prod` where it reads: 'Return the prod over DataFrame axis' instead of 'Return the product over...'. Any suggestions on this? Should we even worry about this?

pandas/core/generic.py
```diff
@@ -8680,24 +8682,304 @@ def _doc_parms(cls):
...
 Parameters
 ----------
-axis : %(axis_descr)s
+axis : {0 or 'index', 1 or 'columns'}, default 0
+    The index or the name of the axis. 0 is equivalent to None or 'index'.
 skipna : boolean, default True
     Exclude NA/null values. If an entire row/column is NA, the result
-    will be NA
+    will be NA.
+*args, **kwargs
```
jorisvandenbossche: Can you add here an explanation for args/kwargs: "Additional arguments have no effect but might be accepted for compatibility with NumPy."

pandas/core/generic.py
jorisvandenbossche: One relatively easy solution would be to change the ``desc`` param to `_make_cum_function` from "cumulative produce" to "product", because we can put the "cumulative" (which is the same for each of the 4 functions" in the template itself. Then you can use that name in the see also description.

pandas/core/generic.py
jorisvandenbossche: I think indeed your prose alternative reads a bit nicer.

jorisvandenbossche (commented):
BTW, you are doing really great work here!

arminv:
I also checked HTML rendering of the corresponding `Series` cumulative methods (`pandas.Series.cummax` ,etc.) and they appear to be ok. There is a duplication in their 'See also' section but I couldn't think of a simple way to fix this.

pandas/core/generic.py
```diff
@@ -8761,24 +8763,297 @@ def _doc_parms(cls):
...
-%(outname)s : %(name1)s\n
-
-
+%(outname)s : %(name1)s or %(name2)s\n
+%(examples)s
 See also
 --------
 pandas.core.window.Expanding.%(accum_func_name)s : Similar functionality
     but ignores ``NaN`` values.
+Series.%(outname)s : Return cumulative %(desc)s over Series axis.
```
jorisvandenbossche: I think this one can be left out, as it will already be included in one of the 4 last ones.

pandas/core/generic.py
```diff
@@ -8761,24 +8763,297 @@ def _doc_parms(cls):
...
+%(name2)s.%(accum_func_name)s : Return the %(desc)s over
+    %(name2)s axis.
+DataFrame.cummax : Return cumulative maximum over DataFrame axis.
```
jorisvandenbossche: This also `%(name2)s` like below ?

jorisvandenbossche (commented):
Looks good! 
Small comment about the "see also"

pandas/core/generic.py
arminv: If I implement these changes, we won't get a reference to any corresponding `Series` in 'See also' of `DataFrame` methods and vice versa. That's why I left them like this (at the expense of repeating one of them). 

I agree that we don't really need them because we have examples of both `Series` & `DataFrame` in all docstrings and this should be informative enough.

Just out of curiosity, is there a simple way to determine whether a `Series` or `DataFrame` method is being accessed at any one time? For example, if we were generating the doc for `Series.cummax`, we would determine it is a `Series` method. Based on this, we know that we need to show cummax but for a `DataFrame`.

jorisvandenbossche: The thing is that the docstring for both Series and DataFrame (of the same method) will be exactly the same (apart from some of the links here in see also). So linking to the same method but on the other object, is not that important I think, as the other one does not give you more information

> Just out of curiosity, is there a simple way to determine whether a Series or DataFrame method is being accessed at any one time? For example, if we were generating the doc for Series.cummax, we would determine it is a Series method. Based on this, we know that we need to show cummax but for a DataFrame.

Yes, the `cls` that is passed to `_make_cum_function` is either the Series or DataFrame class. So you can pass `cls.__name__` as a variable to the template, but I think this is already done as `name2` ? Based on that you can know the other.

jorisvandenbossche (approved):

jorisvandenbossche:
@arminv Thanks a lot for this PR!

arminv:
Thank you guys for being so helpful. I look forward to contributing again.

PR #20337, opened 2018-03-13
TST: Parameterize some compression tests

### LINKED ISSUE #19226: TST: clean up all of the compression tests
we have tests for compression kwargs all over the code base, e.g. ``grep -r compression pandas/tests/``

I think *only* the json and parser network tests parametrize these. The others have copy-pasted code all over the place.

Needs to be consolidated, and the fixtures that are used in json/parser should be pushed to the very top level conftest.py

would accept incremental (multiple) PR's for this.

reidy-p:
closes #19226 

Parameterizing some tests in ``pandas/tests/io/parser/compression.py``. I left the zip test alone because it was quite different to the others and I couldn't see an easy way to incorporate it in the parameterizing.

### DIFF
--- a/pandas/tests/io/parser/compression.py
+++ b/pandas/tests/io/parser/compression.py
@@ -14,7 +14,10 @@ import pandas.util._test_decorators as td
 
 import gzip
 import bz2
-lzma = compat.import_lzma()
+try:
+    lzma = compat.import_lzma()
+except ImportError:
+    lzma = None
 
 
 class CompressionTests(object):
@@ -71,7 +74,7 @@ class CompressionTests(object):
     @pytest.mark.parametrize('compress_type, compress_method, ext', [
         ('gzip', gzip.GzipFile, 'gz'),
         ('bz2', bz2.BZ2File, 'bz2'),
-        pytest.param('xz', lzma.LZMAFile, 'xz',
+        pytest.param('xz', getattr(lzma, 'LZMAFile', None), 'xz',
                      marks=td.skip_if_no_lzma)
     ])
     def test_other_compression(self, compress_type, compress_method, ext):
--- a/pandas/tests/io/parser/compression.py
+++ b/pandas/tests/io/parser/compression.py
@@ -12,6 +12,10 @@ import pandas.compat as compat
 import pandas.util.testing as tm
 import pandas.util._test_decorators as td
 
+import gzip
+import bz2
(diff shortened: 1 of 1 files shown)


pandas/tests/io/parser/compression.py
```diff
@@ -64,83 +68,36 @@ def test_zip(self):
...
 
-            result = self.read_csv(path, compression='bz2')
+            result = self.read_csv(path, compression=compress_type)
             tm.assert_frame_equal(result, expected)
 
-            pytest.raises(ValueError, self.read_csv,
-                          path, compression='bz3')
+            if compress_type == 'bz2':
+                pytest.raises(ValueError, self.read_csv,
+                              path, compression='bz3')
```
jreback: typo

reidy-p: But the reason it's 'bz3' and not 'bz2' is that it's supposed to raise a ValueError

jreback: ahh ok

jreback: hmm i would put that testing of the error co diction in a separate test

jreback: actually on 2nd thought, this is ok.

jreback:
ping on green.

pandas/tests/io/parser/compression.py
```diff
@@ -12,6 +12,10 @@
 import pandas.util.testing as tm
 import pandas.util._test_decorators as td
 
+import gzip
+import bz2
+lzma = compat.import_lzma()
```
jreback: this is slightly annoying you have to skip that test if can't import lzma but you need the file reference
```
Traceback:
pandas/tests/io/parser/test_parsers.py:23: in <module>
    from .compression import CompressionTests
pandas/tests/io/parser/compression.py:17: in <module>
    lzma = compat.import_lzma()
pandas/compat/__init__.py:346: in import_lzma
    from backports import lzma
E   ImportError: cannot import name lzma
```

you might be able to do a 
```
try:
    lzma = compat.import_lzma()
except ImportError:
    lzma = None
```

then change the below to
```
getattr(lzma, 'LZMAFile', None)
```

any other ideas @WillAyd

reidy-p:
@jreback your suggestion worked so I think this is ready. Thanks!

jreback:
thanks!

PR #20344, opened 2018-03-14
DOC: update the pd.Index.asof docstring

DataOmbudsman:
an unrelated PR):

```
######################## Docstring (pandas.Index.asof)  ########################

Return the latest index label up to and including the passed label.

For sorted indexes, return the index label that is the latest among
the labels that are not later than the passed index label.

Parameters
----------
label : object
    The label up to which the method returns the latest index label.

Returns
-------
object : The index label that is the latest as of the passed label,
    or NaN if there is no such label.

See also
--------
Index.get_loc : `asof` is a thin wrapper around `get_loc`
    with method='pad'.

Examples
--------
13 is the latest index label up to 14.

>>> pd.Index([13, 18, 20]).asof(14)
13

If the label is in the index, the method returns the passed label.

>>> pd.Index([13, 18, 20]).asof(18)
18

If all of the labels in the index are later than the passed label,
NaN is returned.

>>> pd.Index([13, 18, 20]).asof(1)
nan

If the index is not sorted, an error is raised.

>>> pd.Index([13, 20, 18]).asof(18)
ValueError: index must be monotonic increasing or decreasing

Docstring for "pandas.Index.asof" correct. :)

```

explicitly.

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -2397,32 +2397,37 @@ class Index(IndexOpsMixin, PandasObject):
 
         See Also
         --------
+        Series.asof : Return the latest value in a Series up to the
+            passed index.
+        merge_asof : Perform an asof merge (similar to left join but it
+            matches on nearest key rather than equal key).
         Index.get_loc : `asof` is a thin wrapper around `get_loc`
             with method='pad'.
 
         Examples
         --------
-        The method returns the latest index label up to the passed label.
+        `Index.asof` returns the latest index label up to the passed label.
 
-        >>> idx = pd.Index([13, 18, 20])
-        >>> idx.asof(14)
-        13
+        >>> idx = pd.Index(['2013-12-31', '2014-01-02', '2014-01-03'])
+        >>> idx.asof('2014-01-01')
+        '2013-12-31'
 
         If the label is in the index, the method returns the passed label.
 
-        >>> idx.asof(18)
-        18
+        >>> idx.asof('2014-01-02')
+        '2014-01-02'
 
         If all of the labels in the index are later than the passed label,
         NaN is returned.
 
-        >>> idx.asof(1)
+        >>> idx.asof('1999-01-02')
         nan
 
         If the index is not sorted, an error is raised.
 
-        >>> idx_not_sorted = pd.Index([13, 20, 18])
-        >>> idx_not_sorted.asof(18)
+        >>> idx_not_sorted = pd.Index(['2013-12-31', '2015-01-02',
+        ...                            '2014-01-03'])
+        >>> idx_not_sorted.asof('2013-12-31')
         Traceback (most recent call last):
         ValueError: index must be monotonic increasing or decreasing
         """
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -2378,10 +2378,11 @@ class Index(IndexOpsMixin, PandasObject):
 
     def asof(self, label):
         """
-        Return the latest index label up to and including the passed label.
+        Return the label from the index, or, if not present, the previous one.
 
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/base.py
```diff
@@ -2378,12 +2378,49 @@ def identical(self, other):
 
     def asof(self, label):
         """
-        For a sorted index, return the most recent label up to and including
-        the passed label. Return NaN if not found.
+        Return the latest index label up to and including the passed label.
+
+        For sorted indexes, return the index label that is the latest among
+        the labels that are not later than the passed index label.
```
datapythonista: I find this whole docstring a bit confusing to be honest. I think it's a complex method to document, but I think we can do better.

For the examples, I find this much clearer:
```
import pandas

idx = pandas.Index(['a', 'b', 'd', 'e'])
idx.asof('c')
```

For the description, it's difficult, but something like "Return the label from the index, or the previous if not found" would be something easier to understand for me.

For the rest looks quite good.

DataOmbudsman: It's hard to be completely understandable within a single line. But I'll try again using your suggestions. Thanks!

DataOmbudsman:
Hello @datapythonista I've updated the documentation based on your suggestions. Do you think it's okay now to be merged?

pandas/core/indexes/base.py
```diff
@@ -2378,12 +2378,53 @@ def identical(self, other):
...
+
+        >>> idx.asof(1)
+        nan
+
+        If the index is not sorted, an error is raised.
+
+        >>> idx_not_sorted = pd.Index([13, 20, 18])
+        >>> idx_not_sorted.asof(18)
+        Traceback (most recent call last):
+        ValueError: index must be monotonic increasing or decreasing
```
datapythonista: Looks great. The only think that comes to my mind is if it'd make sense to use dates in the examples. I think the example with integers is simpler, and easier to understand. But I'd say in practice the method is only used with dates.

Like, if you want to know the price of a stock, and you say, give me the price `asof` 1st of January. The 1st of January the markets are closed, so the price returned would be the one of the 31st of December or the last day the markets were open.

I'm not quite sure what would be the best. If you think you can improve the docstring with the user case I just mentioned, that would be great. Otherwise looks good to me.

And may be we could also add `Series.asof` and `pandas.merge_asof` to the See Also section.

Thanks for the good work.

DataOmbudsman: Thanks for the feedback! I've updated the docs using your suggestions.

pandas/core/indexes/base.py
```diff
@@ -2378,12 +2378,58 @@ def identical(self, other):
...
+        Parameters
+        ----------
+        label : object
+            The label up to which the method returns the latest index label.
+
+        Returns
+        -------
+        object : The passed label if it is in the index. The previous label
+            if the passed label is not in the sorted index, or NaN if there
+            is no such label.
```
datapythonista: Looks great, just one minor formatting thing. The description goes to the next line, and the colon is not required. Also you can add backticks around the `NaN`. For the rest is perfect, really like this final result.

datapythonista (approved):
lgtm. Great job!

datapythonista:
@DataOmbudsman  no need for that. Project maintainers and core developers are in most cases volunteers, that use their spare time to work on the projects. They will take a look at this when they have the time. In many cases that will be in few hours or couple of days. In other cases it can take weeks. You just need to wait. If you check it, you'll see that there are 165 pull requests like yours open. Be patient, and work in something else if you've got the time. :)

DataOmbudsman:
@datapythonista All right, thanks for the clarification. :)

mroeschke:
Thanks @DataOmbudsman! Sorry for the delay on our end. Great docstring!

Issue #20328, opened 2018-03-13
Parameterize NA sentinel for hashtables


TomAugspurger:
xref https://github.com/pandas-dev/pandas/pull/19938. There's I'm abusing the fact that the Int64HashTable class treats iNaT as a missing value. It'd be cleaner to just pass the Categorical's codes with a flag saying that `-1` is the missing value marker.

IIUC, the missing value condition is baked into the class definition.

```
{{py:

# name, dtype, null_condition, float_group
dtypes = [('Float64', 'float64', 'val != val', True),
          ('UInt64', 'uint64', 'False', False),
          ('Int64', 'int64', 'val == iNaT', False)]
```

I'm not familiar enough with Cython to know how that could be made a parameter. Possibly an optional scalar value. If that value is not None check if `val == na_sentinel`. Else, use the condition from the class definition?

WillAyd:
Not 100% certain but I believe you can define an optional argument for your sentinel in the `__cinit__`

https://github.com/pandas-dev/pandas/blob/3783ccc694ef93c0cf159ca1ad8e69a67d828bf0/pandas/_libs/hashtable_class_helper.pxi.in#L311

Then adjust the existing comparison to do something like `if if check_null and (val == self.sentinel or {{null_condition}}):`

https://github.com/pandas-dev/pandas/blob/3783ccc694ef93c0cf159ca1ad8e69a67d828bf0/pandas/_libs/hashtable_class_helper.pxi.in#L428

jreback:
yeah *maybe* we could just make a parameter for this, rather than ``check_null``, and default it in the template (so allow it to be overriden)

TomAugspurger:
Just doing `if val == self.na_value` doesn't quite work, since `na_value` is a (potentially?) a Python object, and this is all in a `no_gil` block.

WillAyd:
I don't think an object is possible to hit that branch because it's not explicitly listed as one of the dtypes - do you have an example of where that is happening?

https://github.com/pandas-dev/pandas/blob/3783ccc694ef93c0cf159ca1ad8e69a67d828bf0/pandas/_libs/hashtable_class_helper.pxi.in#L254

WillAyd:
If you need to add support for objects in your template you could replace the `nogil` block with `if True:` - I've done this a few times in `groupby_helper.pxi.in`:

[...]

PR #19938, opened 2018-02-28
REF/BUG/API: factorizing categorical data

TomAugspurger:
This changes / fixes how Categorical data are factorized. The return value of a
factorized categorical is now `Tuple[ndarray[int], Categorical]`.

Before

```python
In [2]: l, u = pd.factorize(pd.Categorical(['a', 'a', 'b']))

In [3]: l
Out[3]: array([0, 0, 1])

In [4]: u
Out[4]: array([0, 1])
```

after

```python
In [2]: l, u = pd.factorize(pd.Categorical(['a', 'a', 'b']))

In [3]: l
Out[3]: array([0, 0, 1])

In [4]: u
Out[4]:
[a, b]
Categories (2, object): [a, b]
```

The implementation is similar to `.unique`.

1. The algo (`pd.factorize`, `pd.unique`) handles unboxing / dtype coercion
2. The algo dispatches the actual array factorization for extension types
3. The algo boxes the output if necessary, depending on the input.

I've implemented this as a new public method on ``Categorical``, mainly since
this is what we do for unique, and I think it's a useful method to have.

This fixes a bug in factorizing categoricals with missing values. Previously, we
included -1 in the uniques.

Before

```python
In [2]: l, u = pd.factorize(pd.Categorical(['a', 'a', 'b', None]))

In [3]: u
Out[3]: array([ 0,  1, -1])
```

After

```python
In [2]: l, u = pd.factorize(pd.Categorical(['a', 'a', 'b', None]))

In [3]: u
Out[3]:
[a, b]
Categories (2, object): [a, b]
```

xref #19721. This doesn't close it, as I haven't addressed the issue of what `factorize(Series[EA])` should return. Just fixing the categorical bugs.

PR #20361, opened 2018-03-15
ENH/API: ExtensionArray.factorize

TomAugspurger:
Adds factorize to the interface for ExtensionArray, with a default
implementation.

This is a stepping stone to groupby.

Categorical already has a custom implementation.

Decimal reuses the default.

JSON can't use the default since an array of dictionaries isn't hashable, so we use a custom implementation.


pandas/tests/extension/json/array.py
```diff
@@ -104,6 +105,21 @@ def _concat_same_type(cls, to_concat):
         data = list(itertools.chain.from_iterable([x.data for x in to_concat]))
         return cls(data)
 
+    def factorize(self, na_sentinel=-1):
+        frozen = tuple(tuple(x.items()) for x in self)
+        labels, uniques = pd.factorize(frozen)
+
+        # fixup NA
```
TomAugspurger: This type of thing is going to be error prone. We expect that `[B, NA, C, B]` is coded as `[0, -1, 1, 0]`, not `[0, 1, 2, 0]` (I ran into this with cyberpandas as well).

If library authors are using our tests, this should be caught. Otherwise, I'm not sure how to handle it. We could design a method like around `def refactorize(labels, mask, uniques)` that does something like this, but I'm not sure.

TomAugspurger: Hmm apparently I messed this up, so the tests will fail. Fixing now.

TomAugspurger: Fixed in 434df7dc9c927ac52b881e748ebee12c1389cf15

TomAugspurger:
Marking this as a WIP for now. For JSONArray, the `value_counts` test is failing. I'd like to see if we can reuse `factorize` for `value_counts`. In principal, a counting the labels and doing a `uniques.take` on the index of the counts should work.

TomAugspurger:
The failures are caused by a "bug" in is_extension_dtype. https://github.com/pandas-dev/pandas/pull/20363

pandas/core/algorithms.py
```diff
@@ -502,9 +501,9 @@ def factorize(values, sort=False, order=None, na_sentinel=-1, size_hint=None):
     values = _ensure_arraylike(values)
     original = values
 
-    if is_categorical_dtype(values):
+    if is_extension_array_dtype(values):
         values = getattr(values, '_values', values)
-        labels, uniques = values.factorize()
+        labels, uniques = values.factorize(na_sentinel=na_sentinel)
```
TomAugspurger: Note: this was a bug in #19938 where I forgot to pass this through. It's covered by our extension tests.

TomAugspurger:
This should pass now.

pandas/core/arrays/base.py
```diff
@@ -248,6 +248,41 @@ def unique(self):
...
+
+        See Also
+        --------
+        pandas.factorize : top-level factorize method that dispatches here.
+
+        Notes
+        -----
+        :meth:`pandas.factorize` offers a `sort` keyword as well.
+        """
+        from pandas.core.algorithms import _factorize_array
```
jorisvandenbossche: We're OK with using this private API here? 

Because an extension authors might want to copy paste this method and change the `arr = self.astype(object)` line? (how do you implement `factorize` in cyberpandas?)

pandas/tests/extension/base/methods.py
```diff
@@ -42,3 +43,22 @@ def test_unique(self, data, box, method):
...
+                                   na_sentinel, 1, 1, 0, 2],
+                                   dtype='int64')
+        expected_uniques = data_for_grouping.take([0, 4, 7])
+
+        tm.assert_numpy_array_equal(labels, expected_labels)
+        self.assert_extension_array_equal(uniques, expected_uniques)
+
+    def test_factorize_equivalence(self, data_for_grouping):
+        l1, u1 = pd.factorize(data_for_grouping)
+        l2, u2 = pd.factorize(data_for_grouping)
```
jorisvandenbossche: What is this test doing?

pandas/tests/extension/json/array.py
jorisvandenbossche: If we require this, I think we should put a stronger comment (maybe as a comment below the docstring) about that we expect that missing values are not *in* the uniques.

What would go wrong if `.factorize()` on an extension array would include its missing value in the uniques?

pandas/util/testing.py
```diff
@@ -1083,6 +1084,32 @@ def _raise(left, right, err_msg):
...
+    ----------
+    left, right : ExtensionArray
+        The two arrays to compare
+
+    Notes
+    -----
+    Missing values are checked separately from valid values.
+    A mask of missing values is computed for each and checked to match.
+    The remaining all-valid values are cast to object dtype and checked.
+    """
```
jorisvandenbossche: Why was this needed only now? (wasn't the missing values the reason you added `assert_frame_equal` et al as overridable class methods?)

pandas/tests/extension/base/methods.py
TomAugspurger: Ha, I think I meant for one to be `data_for_grouping.factorize()`

pandas/tests/extension/json/array.py
TomAugspurger: Hmm probably not the end of the world, but it would violate our factorize (and eventually groupby) semantics. Until recently, we messed this up for Categorical. That didn't have downstream effects for groupby, since everything in groupby is special for categorical.

I'll make this clearer.

pandas/util/testing.py
TomAugspurger: `factorize` is I think the first case where we have a public method returning an ExtensionArray (`unique`).

I'll see if any old tests can make use of this.

TomAugspurger: I don't see any other cases where we could apply `assert_extension_array_equal`.

The closes would be the `__getitem__` tests, but in that case we don't have an expected array to compare to, since we're testing getitem / take / etc.

pandas/core/arrays/base.py
TomAugspurger: Quite similar to this. https://github.com/ContinuumIO/cyberpandas/blob/468644bcbdc9320a1a33b0df393d4fa4bef57dd7/cyberpandas/base.py#L72

In that case I think going to object dtypes is unavoidable, since there's no easy way to factorize a 2-D array, and I didn't want to write a new hashtable implementation :)

TomAugspurger: w.r.t. using factorize array, I don't think it's avoidable.

We might consider making it public / semi-public (keep the `_` to not confuse users, but point EA authors to it).

pandas/tests/extension/json/array.py
TomAugspurger: Well adding that simple note turned into a rabit hole. I decided to merge the 3 docstrings

- pandas.factorize
- pandas.core.base.IndexOpsMixin.factorize
- pandas.Categorical.factorize

pandas/core/arrays/base.py
```diff
@@ -248,6 +248,48 @@ def unique(self):
         uniques = unique(self.astype(object))
         return self._constructor_from_sequence(uniques)
 
+    def factorize(self, na_sentinel=-1):
+        """Encode the extension array as an enumerated type.
```
TomAugspurger: Ah, need to merge this one too maybe. Will have to check on import order...

TomAugspurger:
Sorry about the tangentially related changes refactoring the docstring. I can revert if needed, but I don't think it complicates review too much since the only non-doc change to `algos.py` is replacing an `is_categorical` with an `is_extension_array_dtype`.

Tracking down a py-2 failure now.

TomAugspurger:
Here's the new docstring for pandas.factorize

fireshot capture 005 - pandas factorize pandas 0 23 0 dev0_ - file____users_taugspurger_sandbox_

jreback:
hmm, ``order`` is still there I thought we removed that? (or maybe we just deprecated it, can you add back as a deprecated parm)?

also, as a public interface, may not really need ``size_hint`` (as now have ``_factorize_array`` which can be an internal interface, but this is small potatoes).

jreback:
maybe link to ``pd.unique`` as well

pandas/core/algorithms.py
```diff
@@ -464,32 +468,117 @@ def _factorize_array(values, check_nulls, na_sentinel=-1, size_hint=None):
...
+        `Index` is returned. Otherwise, a 1-D ndarray is returned.
+
+        .. note ::
+
+           Even if there's a missing value in `values`, `uniques` will
+           *not* contain an entry for it.
+
+    See Also
+    --------
+    pandas.cut : Discretize continuous-valued array.
```
jreback: add ``pandas.unique`` here

TomAugspurger:
> can you add back as a deprecated parm

yeah, lost that in the shuffle.

> also, as a public interface, may not really need size_hint

Agreed.

> maybe link to pd.unique as well

👍

pandas/core/algorithms.py
```diff
@@ -464,32 +468,117 @@ def _factorize_array(values, check_nulls, na_sentinel=-1, size_hint=None):
...
     ----------
-    values : Sequence
-        ndarrays must be 1-D. Sequences that aren't pandas objects are
-        coereced to ndarrays before factorization.
-    sort : boolean, default False
-        Sort by values
+    %(values)s%(sort)s
     na_sentinel : int, default -1
-        Value to mark "not found"
-    size_hint : hint to the hashtable sizer
```
jreback: need to show ``order`` as deprecated

jreback (changes requested):

jreback:
needs a rebase

pandas/core/algorithms.py
```diff
@@ -463,32 +468,117 @@ def _factorize_array(values, check_nulls, na_sentinel=-1, size_hint=None):
     return labels, uniques
 
 
-@deprecate_kwarg(old_arg_name='order', new_arg_name=None)
-def factorize(values, sort=False, order=None, na_sentinel=-1, size_hint=None):
-    """
-    Encode input values as an enumerated type or categorical variable
+_shared_docs['factorize'] = """
```
jorisvandenbossche: If it is only a single entry, can we make this a variable (eg `_shared_docstring_factorize`) ?

pandas/core/algorithms.py
```diff
@@ -463,32 +468,117 @@ def _factorize_array(values, check_nulls, na_sentinel=-1, size_hint=None):
...
+    For all other pandas objects, an Index of the appropriate type is
+    returned.
 
-    note: an array of Periods will ignore sort as it returns an always sorted
-    PeriodIndex.
+    >>> cat = pd.Series(['a', 'a', 'c'])
+    >>> labels, uniques = pd.factorize(cat)
+    >>> labels
+    array([0, 0, 1])
+
```
jorisvandenbossche: remove blank line here

pandas/core/algorithms.py
```diff
@@ -463,32 +468,117 @@ def _factorize_array(values, check_nulls, na_sentinel=-1, size_hint=None):
...
+    >>> uniques
+    Index(['a', 'c'], dtype='object')
     """
+
+
+@Substitution(
+    values=dedent("""\
+    values : Sequence
```
jorisvandenbossche: I would personally use something else as "Sequence", I don't think we use that somewhere else (or at least lower case "sequence"). It's "array-like" ?

pandas/core/algorithms.py
```diff
@@ -463,32 +468,117 @@ def _factorize_array(values, check_nulls, na_sentinel=-1, size_hint=None):
...
+        A 1-D seqeunce. Sequences that aren't pandas objects are
+        coereced to ndarrays before factorization.
+    """),
+    sort=dedent("""\
+    sort : boolean, default False
```
jorisvandenbossche: boolean -> bool

pandas/core/algorithms.py
TomAugspurger: Sorry, was thinking in `typing` terms.

I'm trying to use `array-like` to mean sequence + dtype (which is what `is_array_like` checks).

TomAugspurger:
Do we want to have `ExtensionArray.factorize` use the shared docstring as well? There's an import order issue, I've tried to keep `arrays/base.py` from importing much from pandas (just `AbstractMethodError` so far. If we do a shared doc I would have to move the docstring to a file on it's own, and then import it from `algorithms.py`, `arrays/base.py`, `core/base.py` and `categorical.py`.

I also haven't really thought through if that would break anything for EA implementers.

pandas/core/arrays/base.py
```diff
@@ -301,6 +319,60 @@ def unique(self):
...
+        Returns
+        -------
+        ndarray
+            An array suitable for factoraization. This should maintain order
+            and be a supported dtype.
+
+        """
+        return self.astype(object)
+
+    def factorize(self, na_sentinel=-1):
```
jreback: you need a sort=False arg here

pandas/core/arrays/base.py
```diff
@@ -301,6 +319,60 @@ def unique(self):
...
+        -----
+        :meth:`pandas.factorize` offers a `sort` keyword as well.
+        """
+        # Implementor notes: There are two options for overriding the
+        # behavior of `factorize`: here and `_values_for_factorize`.
+        from pandas.core.algorithms import _factorize_array
+
+        mask = self.isna()
+        arr = self._values_for_factorize()
+        arr[mask] = np.nan
```
jreback: shouldn't this be the ``self._na_value`` ?

pandas/core/arrays/categorical.py
```diff
@@ -2043,47 +2044,9 @@ def unique(self):
...
-        Missing values are handled
-
-        >>> labels, uniques = pd.factorize(pd.Categorical(['a', 'b', None]))
-        >>> labels
-        array([ 0,  1, -1])
-        >>> uniques
-        [a, b]
-        Categories (2, object): [a, b]
-        """
         from pandas.core.algorithms import _factorize_array
```
jreback: shouldn't the impl change here? (to use the newly created EA facilities)

jreback (changes requested):

pandas/core/arrays/base.py
TomAugspurger: Do we want / need that? It complicates the implementation a bit. Any idea what it's actually used for?

pandas/core/arrays/base.py
TomAugspurger: We removed that from the API. Anyway, that would be the type-specific NA value, we need an NA value that pandas knows about, which is just NaN or NaT.

Speaking of which, I don't really like how this is done right now. If your values_for_factorize are integers, this will force them to be floating. Will thing more about it.

TomAugspurger:
Ugh I mixed up my branches and merged https://github.com/pandas-dev/pandas/pull/19957 into here... Fixing now.

pandas/core/arrays/base.py
```diff
@@ -301,6 +319,64 @@ def unique(self):
...
+    def _values_for_factorize(self):
+        """Return an array and missing value suitable for factorization.
+
+        Returns
+        -------
+        values : ndarray
+            An array suitable for factoraization. This should maintain order
+            and be a supported dtype. By default, the extension array is cast
+            to object dtype.
+        na_value : scalar
```
TomAugspurger: OK, here's the meaningful change from my last round of commits. We have the user specify an ndarray of values to be factorized, and the na value *going into* the factorization routine.

Previously, `Categorical.factorize` had to be implemented, since it used integers and `iNaT` for the array and missing value in `_factorize_array`. So from a code reuse and interface flexibility point of view, this is a good thing.

I dislike how this opens up the internals of pandas' factorization algorithms to the public more. Is it publicly documented anywhere that iNaT is the minimum integer? How would an EA author know to do something like what we do for Categorical (and PeriodArray when we get there)?

jreback:
this looks fine. I am still concerned about the API expansion for implementors here, IOW you have both ``_values_for_factorize`` and now ``_from_factorize``.

jreback (approved):

jorisvandenbossche:
> I dislike how this opens up the internals of pandas' factorization algorithms to the public more. Is it publicly documented anywhere that iNaT is the minimum integer? How would an EA author know to do something like what we do for Categorical (and PeriodArray when we get there)?

But is it needed that the extension author specifies this value? 
Because what are the options? If the array has integer-like dtype, it is iNaT, otherwise it is np.nan ? 
(if that is the case, I think we could deduce that?)


In general, I am wondering if we couldn't make this simpler. If we could somehow extend `factorize` to specify a mask or a na_value, then the `factorize` method on EA could be something like:

```
def factorize(self, na_sentinel=-1):
    mask = self.isna()
    values = ... # this is the part to fill in by the extension array author (equivalent _values_for_factorize)
    labels, uniques = pd.factorize(values, na_sentinel=na_sentinel, mask=mask)
    uniques = ... # coerce back to extension array (_from_factorized)
    return labels, uniques
```

This is of course *very* similar to the current `factorize` implementation, but with the difference that it does not use internal APIs. And actually, overriding this does not take more lines of code than overriding both `_values_for_factorize` and `_from_factorized`. So I am not fully sure of the advantage of providing both ways to do it.

TomAugspurger:
> So I am not fully sure of the advantage of providing both ways to do it.

"both" being

1. `EA.factorize`
2. `EA._values_for_factorize`, and `EA._from_factorized`

correct?

Let me see if I can make `mask` as an argument to `factorize` work. That would indeed be cleaner.

jorisvandenbossche:
> "both" being
>
>   EA.factorize
>   EA._values_for_factorize, and EA._from_factorized

Yes. Looking at the diff now, it seems to me that overwriting `EA._values_for_factorize` and `EA._from_factorized` will actually need as much as custom code as just overwriting `EA.factorize`. So then the argument of providing `_values_for_factorize` to reduce duplication of the actual `factorize` implementation (like we had for `argsort`) does not fully hold in this case.

> Let me see if I can make mask as an argument to factorize work. That would indeed be cleaner.

Then of course we need to be OK with exposing this in the public `factorize` API. But from an extension author point of view this seems rather clean. 
An alternative would be to specify a `na_value` and then calculate the mask, but I am not sure if that has any advantage over a mask.

@jreback what are your thoughts on such a `mask` keyword for `pd.factorize`

TomAugspurger:
I'm having some trouble making the `mask` approach work. It seems like pandas' factorization methods do the right thing for some missing values, like `Decimal(NaN)`.


```python
In [2]: from decimal import Decimal

In [3]: pd.factorize([Decimal('0.0'), Decimal('NaN'), Decimal('0.0')])
Out[3]: (array([ 0, -1,  0]), array([Decimal('0.0')], dtype=object))
```

So in that situation we would have to choose to ignore the mask argument (side-note, `pd.isna` does not consider `Decimal('NaN')` to be NA, so detecting this seems fragile).

jorisvandenbossche:
But for those object arrays, we can always use `np.nan` or `None` I think if a mask is specified? That will always work, regardless of whether the actual NA value is recognized by `factorize` ?

jorisvandenbossche:
Ah, yes, I was thinking of the "change `values` before passing to `factorize`". 
The problem with that is of course that we then should take a copy of `values`, which might be an unnecessary copy in case the `values` were constructed specifically for `factorize`

TomAugspurger:
Here are the cases I see. I put quotes around the ExtensionArray's definition of missing, which may differ from pandas'.

1. float/object values with "NA" values as NA (Decimal)
2. float/object values with "NA" values as something else (IPArray)
3. integer values with "NA" values as iNaT (DatetimeTZ)
4. integer values with "NA" values as something else (Categorical)

You could add a 5/6 for our uint64 hashtable. I don't think that changes anything.

So for cases where the NA values align (1, 3), things will just work, no `mask` required.

For cases where they don't align (2, 4), we'll need to either modify the values going into the hashtable, so that `values[mask] = pandas_na_for_hashtable`, or modify that labels / uniques coming out.

TomAugspurger:
Though it occurs to me that a mask with NA values goes against https://github.com/pandas-dev/pandas/issues/20328 a bit... Implementing https://github.com/pandas-dev/pandas/issues/20328 would also make this cleaner, as the EX author need only pass through an `na_value`, and everything should be OK. I'll see if doing that is simple.

jorisvandenbossche:
@TomAugspurger Alternative is to just document that if you have float/object data, we expect np.nan, if you have integer data, we expect iNaT as missing value. And that extension authors then apply the mask themselves. 
That brings us back to the original implementation of the PR I think ..

TomAugspurger:
#20328 doesn't look trivial, so I've abandoned it for now.

> Alternative is to just document that if you have float/object data, we expect np.nan, if you have integer data, we expect iNaT as missing value.

Yes, we'll need to do this regardless, as extension-array authors may have to take precautions if they use iNaT (int min) as a valid value. But I think we'll have to make it clear that this is an implementation detail at the moment.

>  That brings us back to the original implementation of the PR I think ..

Close. The main difference is that in c288d67, we mask the values going *into* the hashtable.

If this approach looks OK, I will add sufficiently scary warnings that the default implementation is using features private to pandas (`_factorize_array`, the missing value markers for each type of hashtable, etc.).

Tangent, I noticed that the UInt64Hashtable doesn't have a missing value marker. Was that intentional (ping @gfyoung I think).

```python
In [9]: from pandas.core.algorithms import _factorize_array

In [10]: _factorize_array(arr, check_nulls=True)
Out[10]:
(array([0, 0, 1, 2]),
 array([                   0,                    1, 18446744073709551615],
       dtype=uint64))
```

In hashtable_helper, the null condition seems to be the string `'False'`.

```
# name, dtype, null_condition, float_group
dtypes = [('Float64', 'float64', 'val != val', True),
          ('UInt64', 'uint64', 'False', False),
          ('Int64', 'int64', 'val == iNaT', False)]
```

Anyway, changing this probably isn't desired, but thought I'd mention it.

gfyoung:
> Tangent, I noticed that the UInt64Hashtable doesn't have a missing value marker.

@TomAugspurger : Unfortunately, there is no well-defined missing value marker for `uint` AFAIK.

pandas/core/arrays/base.py
```diff
@@ -353,6 +371,66 @@ def unique(self):
...
+        arr = self._values_for_factorize()
+
+        # Mask values going into the hash table with the appropriate
+        # NA type.
+        if com.is_signed_integer_dtype(arr):
+            arr[mask] = iNaT
+        elif com.is_float_dtype(arr) or com.is_object_dtype(arr):
+            arr[mask] = np.nan
+
+        labels, uniques = _factorize_array(arr, check_nulls=True,
```
jorisvandenbossche: we could also use `pd.factorize` here if we want to limit the use of internal APIs

pandas/core/arrays/base.py
```diff
@@ -353,6 +371,66 @@ def unique(self):
...
+
+        mask = self.isna()
+        arr = self._values_for_factorize()
+
+        # Mask values going into the hash table with the appropriate
+        # NA type.
+        if com.is_signed_integer_dtype(arr):
+            arr[mask] = iNaT
+        elif com.is_float_dtype(arr) or com.is_object_dtype(arr):
+            arr[mask] = np.nan
```
jorisvandenbossche: My original idea with a `mask` keyword for `factorize`, is that this if/elif code block would do this in `factorize` itself

pandas/core/arrays/base.py
TomAugspurger: I was just trying that. Unfortunately, I have to pass `check_nulls=True` for categorical to work properly. It's typically False for integers. But perhaps that's a sign we're doing something improper :/

pandas/core/arrays/base.py
```diff
@@ -353,6 +371,66 @@ def unique(self):
...
+        """Return an array and missing value suitable for factorization.
+
+        Returns
+        -------
+        values : ndarray
+            An array suitable for factoraization. This should maintain order
+            and be a supported dtype. By default, the extension array is cast
+            to object dtype.
+        """
+        return self.astype(object)
```
jorisvandenbossche: We should add a comment here that those values might be modified (in the current state of the PR). So that you eg don't just return the ordinals / codes for Period (which is not the case for Categorical right now, because it does `astype(int64)` which always copies)

pandas/core/arrays/base.py
jorisvandenbossche: Ah, yes, because there we still know what the original dtype was ... 
Hmm, that maybe shows that we rather need a `na_value` keyword (just from a perspective of making `pd.factorize` generally usable, not necessarily in this PR). 
In general, the default of `na_value=None` could be the `val != val` idiom, and you can override that with `val == na_value` with this keyword (e.g. to specify iNaT or -1 for integer data).

TomAugspurger: OK, this issue and the fact that UInt64Hashtable doesn't have a null value are making me come back around to the mask approach.

That seems like the only way we can ensure that we'll actually get things correct.

TomAugspurger:
@jreback you can ignore all the back and forth between Joris and I yesterday about `mask`. I think https://github.com/pandas-dev/pandas/pull/20473 will make things nice for EA authors. They'll just need to specify

- The array to be used for factorization
- The scalar in their array to be considered NA.

- defualt: `self.astype(object), np.nan`
- Categorical: `self.codes.astype(int64), -1`
- MACArray: `self.data, 0` (this is in cyberpandas. A uint64 array, where 0 is NA)

jorisvandenbossche:
> @jorisvandenbossche do you think this will work well for geopandas? Can you have missing geometries?

Yes, that should be fine. I am not fully sure yet, but I think I will use a binary blob for each geometry ("well known binary" format), which will give an object array of bytes, and with eg None as missing value. So this should work (that should already work with `pd.factorize` itself without the new infrastructure I think). 
I will also need the `original` to add back metadata about the coordinate reference system.

TomAugspurger:
From Python your array of pointers look like Int64s, right? Could those be factorized? I *think* it should, though I haven't thought it through fully.

jorisvandenbossche:
No, because two separate Points (separate objects) but with same coordinates, will have different pointer. So the pointer information is not necessarily relevant for factorize.

TomAugspurger:
Understood.

CI is all green.

jreback:
thanks @TomAugspurger

PR #20363, opened 2018-03-15
REF: Changed ExtensionDtype inheritance

TomAugspurger:
`is_extension_array_dtype(dtype)` was incorrect for dtypes that haven't
implemented the new interface yet. This is because they indirectly subclassed
ExtensionDtype.

This PR changes the hierarchy so that PandasExtensionDtype doesn't subclass
ExtensionDtype. As we implement the interface, like Categorical, we'll add
ExtensionDtype as a base class.

Before:

```
DatetimeTZDtype <- PandasExtensionDtype <- ExtensionDtype (wrong)
CategoricalDtype <- PandasExtensionDtype <- ExtensionDtype (right)

After:

DatetimeTZDtype <- PandasExtensionDtype
                                        \
                                         - _DtypeOpsMixin
                                        /
                   ExtensionDtype ------

CategoricalDtype - PandasExtensionDtype -
                \                        \
                 \                        -_DtypeOpsMixin
                  \                      /
                   ExtensionDtype -------

```

Once all our extension dtypes have implemented the interface we can go back
to the simple, linear inheritance structure.

### DIFF
--- a/pandas/core/dtypes/base.py
+++ b/pandas/core/dtypes/base.py
@@ -5,26 +5,16 @@ from pandas import compat
 from pandas.errors import AbstractMethodError
 
 
-class ExtensionDtype(object):
-    """A custom data type, to be paired with an ExtensionArray.
-
-    Notes
-    -----
-    The interface includes the following abstract methods that must
-    be implemented by subclasses:
-
-    * type
-    * name
-    * construct_from_string
-
-    This class does not inherit from 'abc.ABCMeta' for performance reasons.
-    Methods and properties required by the interface raise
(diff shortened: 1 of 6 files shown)


jorisvandenbossche (approved):

jreback (changes requested):
can you add a test that asserts is_extension_array (or not) for each of the current pandas extension types (and so should only be True for CategoricalDtype, but false for Interval, Period, Datetimewtz)

TomAugspurger:
Added tests for each extension dtype. Will have to move from `not` to `is` as we migrate them.

Issue #17978, opened 2017-10-25
Investigate using Hypothesis for some tests


TomAugspurger:
Hypothesis recently added support for generating pandas objects: http://hypothesis.readthedocs.io/en/latest/numpy.html#pandas

This could be useful for testing some of the higher-level methods like `melt`, `merge`, etc.

Are there any objections to adding some hypothesis-based tests to our test suite? Is anyone familiar with hypothesis and interested in putting together examples of how we could use it?

TomAugspurger:
As another example, I could see Hypothesis being *very* useful for something like https://github.com/pandas-dev/pandas/issues/13432 (buggy index set ops for duplicates, non monotonic, categorical)

shoyer:
This sounds like it's definitely worth trying! Is there a guarantee that the test cases it produces will always be the same? (Non-reproducible tests could be pretty painful.)

TomAugspurger:
I believe that once a failing test case is found it's cached (somehow, not
sure on the details.)

jbrockmendel:
A test database is kept in a .hypothesis directory.  Not sure if that plays nicely with CI.

jbrockmendel:
@TomAugspurger did this idea ever go anywhere?  I'm thinking this might be a useful for testing offsets.

TomAugspurger:
Don't think so, go for it!

sushobhit27:
Hi, Can I work on this issue. I have recently implemented some unit test cases using hypothesis and therefore have lil bit of exp which I think can be used here.
Infact I have zeroed in some files, where it can be used easily like pandas/tests/tools/test_numeric.py, e.g

![image](https://user-images.githubusercontent.com/31987769/37984121-0dc4695e-3213-11e8-973b-6834195bffd2.png)

TomAugspurger:
That seems reasonable. I think the main thing is ensuring this doesn't
increase our test times too much.

sushobhit27:
@TomAugspurger, I created a pull request https://github.com/pandas-dev/pandas/pull/20590 for this issue by adding hypothesis based test cases in just one file. However travis CI build failed with error "ImportError: No module named 'hypothesis'". Could you please check how it (hypothesis module) can be available, so that tests can be run.

Zac-HD:
Hi all - I'm a maintainer of Hypothesis and happy user of Pandas, so hopefully I can help with this :smile: 

[...]

PR #20365, opened 2018-03-15
TST: fixtures for timezones

almaleksia:
This is a follow up for #20287 

Different tests use different lists of timezones and it's kind of hard to unify all of them. In this PR I tried to create a fixture that covers most frequent cases. Some tests use naive timezone, while others are designed to use only explicit timezones (and fail in case tz is not specified), that's why I created two fixtures to cover these two group of tests. List of timezones also includes one dateutil-timezone. 
 I'm not sure if described approach is correct though so I would appreciate any feedback.

- [ ] tests added / passed

### DIFF
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -87,3 +87,24 @@ def join_type(request):
     Fixture for trying all types of join operations
     """
     return request.param
+
+
+TIMEZONES = [None, 'UTC', 'US/Eastern', 'Asia/Tokyo', 'dateutil/US/Pacific']
+
+
+@td.parametrize_fixture_doc(str(TIMEZONES))
+@pytest.fixture(params=TIMEZONES)
+def tz_naive_fixture(request):
+    """
+    Fixture for trying timezones including default (None): {0}
+    """
+    return request.param
+
+
+@td.parametrize_fixture_doc(str(TIMEZONES[1:]))
+@pytest.fixture(params=TIMEZONES[1:])
+def tz_aware_fixture(request):
+    """
+    Fixture for trying explicit timezones: {0}
+    """
+    return request.param
--- a/pandas/tests/indexes/datetimes/test_ops.py
+++ b/pandas/tests/indexes/datetimes/test_ops.py
@@ -13,12 +13,17 @@ from pandas.tseries.offsets import BMonthEnd, CDay, BDay
 from pandas.tests.test_base import Ops
 
 
+@pytest.fixture(params=[None, 'UTC', 'Asia/Tokyo', 'US/Eastern',
+                        'dateutil/Asia/Singapore',
+                        'dateutil/US/Pacific'])
+def tz_fixture(request):
+    return request.param
+
+
 START, END = datetime(2009, 1, 1), datetime(2010, 1, 1)
 
 
 class TestDatetimeIndexOps(Ops):
-    tz = [None, 'UTC', 'Asia/Tokyo', 'US/Eastern', 'dateutil/Asia/Singapore',
-          'dateutil/US/Pacific']
 
     def setup_method(self, method):
         super(TestDatetimeIndexOps, self).setup_method(method)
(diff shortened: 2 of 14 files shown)


pandas/conftest.py
```diff
@@ -97,3 +97,22 @@ def join_type(request):
...
     return request.param
+
+
+TIMEZONES = [None, 'UTC', 'US/Eastern', 'Asia/Tokyo', 'dateutil/US/Pacific']
+
+
+@pytest.fixture(params=TIMEZONES)
+def tz_naive(request):
+    """
+    Fixture for trying timezones including default (None)
```
jreback: can you make this ore explicit (e.g. list the timezones here); you can programatically construct the doc string I think (to make this DRY)

pandas/conftest.py
```diff
@@ -97,3 +97,22 @@ def join_type(request):
...
+    """
+    return request.param
+
+
+@pytest.fixture(params=TIMEZONES[1::])
+def tz_aware(request):
+    """
+    Fixture for trying explicit timezones
```
jreback: same

pandas/conftest.py
```diff
@@ -97,3 +97,22 @@ def join_type(request):
     Fixture for trying all types of join operations
     """
     return request.param
+
+
+TIMEZONES = [None, 'UTC', 'US/Eastern', 'Asia/Tokyo', 'dateutil/US/Pacific']
+
+
+@pytest.fixture(params=TIMEZONES)
+def tz_naive(request):
```
jreback: I wouldn't be averse to making the name of this slightly more explict

maybe ``tz_naive_fixture`` (maybe we should do this with the other fixtures as well, can be separate PR for that change)

pandas/tests/indexes/datetimes/test_ops.py
```diff
@@ -13,12 +13,17 @@
 from pandas.tests.test_base import Ops
 
 
+@pytest.fixture(params=[None, 'UTC', 'Asia/Tokyo', 'US/Eastern',
+                        'dateutil/Asia/Singapore',
+                        'dateutil/US/Pacific'])
+def tz(request):
```
jreback: maybe ``tz_fixture``

pandas/tests/indexes/datetimes/test_ops.py
```diff
@@ -209,86 +210,85 @@ def test_nonunique_contains(self):
...
             assert idx[0] in idx
 
-    def test_order(self):
-        # with freq
-        idx1 = DatetimeIndex(['2011-01-01', '2011-01-02',
-                              '2011-01-03'], freq='D', name='idx')
-        idx2 = DatetimeIndex(['2011-01-01 09:00', '2011-01-01 10:00',
-                              '2011-01-01 11:00'], freq='H',
-                             tz='Asia/Tokyo', name='tzidx')
+    @pytest.mark.parametrize('idx',
```
jreback: you can format this more like
```
@pytest.mark.parameterize(
   'idx',
       [
         DatetimeIndex(......),
         DatetimeIndex(....)')
```

e.g. to fit more on a line

jreback (changes requested):

jreback:
@jbrockmendel ?

pandas/conftest.py
```diff
@@ -97,3 +97,22 @@ def join_type(request):
...
+    """
+    Fixture for trying timezones including default (None)
+    """
+    return request.param
+
+
+@pytest.fixture(params=TIMEZONES[1::])
```
jbrockmendel: Is a single-colon OK here?

almaleksia:
@jreback @jbrockmendel 
thank you for the feedback!

> I'm a bit concerned about always using the same ~4 timezones. Is there a risk of some special case weirdness in other timezones that we're missing?

Actually I had the same thought. As for this change, it affects only tests that already share similar list of timezones. I made sure that I didn't shorten or replace timezone lists used in the existing tests. Let's say test is using some timezone not covered by the fixture then list is not replaced. Several tests though got extended by one timezone (dateutil/US/Pacific) after being moved to the fixture. I thought about adding more timezones to the list but for some tests it looked a bit redundant, so I decided to save some build time.

> Would there be a use for separate fixtures for tzstr for tzinfo objects?

do you mean tests like in test_timezones parametrized with tzstr?

TomAugspurger:
> I thought about adding more timezones to the list but for some tests it looked a bit redundant, so I decided to save some build time.

I think that's the right choice. This would be a decent case for using hypothesis: https://github.com/pandas-dev/pandas/issues/17978

pandas/conftest.py
almaleksia: I added decorator for this

pandas/conftest.py
almaleksia: Done :)

pandas/util/_test_decorators.py
```diff
@@ -187,3 +187,10 @@ def decorated_func(func):
                                    "installed->{installed}".format(
                                        enabled=_USE_NUMEXPR,
                                        installed=_NUMEXPR_INSTALLED))
+
+
+def parametrize_fixture_doc(*args):
```
jreback: this is nice, in a future PR can you document this itself?

jreback (approved):

jreback:
can you rebase this just to make sure

pandas/util/_test_decorators.py
jreback: can you add docs on this here

almaleksia: @jreback done
sorry didn't see this comment at first, so I did two updates (one for rebase and one for doc), I think previous builds can be cancelled

TomAugspurger:
Thanks @almaleksia!

PR #20366, opened 2018-03-15
DOC: Update cheat sheet and create link in docs

Dr-Irv:
Update pandas cheat sheet to use `drop(columns=['colname'])`.  

Provide link to cheat sheet from the tutorials page.

Based on discussion a while back on gitter with @jorisvandenbossche , maintainers may want the link to the cheat sheet in tutorials.rst to not point to the GitHub repository as I've edited it here, but to a place within the `http://pandas.pydata.org/pandas-docs/stable/` tree, which means when you publish the docs, you have to copy the PDF version of the cheat sheet to the server hosting the docs.


TomAugspurger:
@Dr-Irv I uploaded the PPF and PPT from this PR to our webserver

- http://pandas.pydata.org/Pandas_Cheat_Sheet.pdf
- http://pandas.pydata.org/Pandas_Cheat_Sheet.pptx

Would you mind updating whatever links you want to point to those?

@jorisvandenbossche do we care about maintaining one of these per release, or just the latest? I think just the latest.

Dr-Irv:
@TomAugspurger OK, I just pushed a change to point to the PDF.  I don't think the PPT should be on the site - it's the "source code". 

Will you set something up in the scripts that do releases that copies the cheat sheet from the source directory to the proper location on the server?  So then if other changes are made as part of the development process, the final version would get published automatically upon a new release?

TomAugspurger:
https://github.com/pandas-dev/pandas/wiki/Release-Checklist/_compare/93f266d34866e92db81e9c61b1b153196119f536...0cacf4f3ad1bf73244507f128a9f87a00b00a0dd

need to automate that, but yes.

jorisvandenbossche:
The latest is fine I suppose. 
In principle, if we have a scripts that automates this, we could also use pandas.pydata.org/pandas-docs/stable/... as the url, which points to ../version/.. that gets updated at each release, so we keep the older ones available. 
But for now just the latest is certainly fine!

TomAugspurger:
OK, let's just keep latest at the root.

PR #20370, opened 2018-03-15
TST: Remove invalid dtype test

TomAugspurger:
This test was not valid. Not every dtype kind is a valid dtype.
For example, the dtype may be 'uint64', but the kind is 'u'.


jorisvandenbossche (approved):

jreback:
sure

PR #20371, opened 2018-03-15
API: Remove integer position args from xy for plotting

### LINKED ISSUE #20056: BUG: df.plot fails when given x,y args as positions

```python
from pandas import DataFrame

df = DataFrame({"A": [1, 2], 'B': [3, 4]})
df.plot(x=0, y=1)
```
#### Problem description

```
IndexError: index 1 is out of bounds for axis 0 with size 1
```

I came across this when writing test cases for my PR #20000 - looks like it happens due to calling `set_index` on the data rather than on a copy. I can potentially fix it there or open a separate PR.

masongallo:
- [x] closes #20056 
- [x] tests added / passed
- [?] whatsnew entry

This is based on discussion from #20000 where we decided it was confusing to allow `df.plot` to support both labels AND positions. I wasn't sure where to put this in whatsnew, so lmk and I'll update it.

### DIFF
--- a/pandas/tests/plotting/test_frame.py
+++ b/pandas/tests/plotting/test_frame.py
@@ -205,9 +205,6 @@ class TestDataFramePlots(TestPlotBase):
     def test_plot_xy(self):
         # columns.inferred_type == 'string'
         df = self.tdf
-        self._check_data(df.plot(x=0, y=1), df.set_index('A')['B'].plot())
-        self._check_data(df.plot(x=0), df.set_index('A').plot())
-        self._check_data(df.plot(y=0), df.B.plot())
         self._check_data(df.plot(x='A', y='B'), df.set_index('A').B.plot())
         self._check_data(df.plot(x='A'), df.set_index('A').plot())
         self._check_data(df.plot(y='B'), df.B.plot())
@@ -1019,7 +1016,6 @@ class TestDataFramePlots(TestPlotBase):
                        columns=['x', 'y', 'z', 'four'])
 
         _check_plot_works(df.plot.scatter, x='x', y='y')
-        _check_plot_works(df.plot.scatter, x=1, y=2)
 
         with pytest.raises(TypeError):
             df.plot.scatter(x='x')
@@ -1054,17 +1050,15 @@ class TestDataFramePlots(TestPlotBase):
                        index=list(string.ascii_letters[:6]),
                        columns=['x', 'y', 'z', 'four'])
 
-        axes = [df.plot.scatter(x='x', y='y', c='z'),
-                df.plot.scatter(x=0, y=1, c=2)]
-        for ax in axes:
-            # default to Greys
-            assert ax.collections[0].cmap.name == 'Greys'
+        axes = df.plot.scatter(x='x', y='y', c='z')
 
-            if self.mpl_ge_1_3_1:
+        # default to Greys
+        assert axes.collections[0].cmap.name == 'Greys'
 
-                # n.b. there appears to be no public method to get the colorbar
-                # label
-                assert ax.collections[0].colorbar._label == 'z'
+        if self.mpl_ge_1_3_1:
+            # n.b. there appears to be no public method to get the colorbar
+            # label
+            assert axes.collections[0].colorbar._label == 'z'
 
         cm = 'cubehelix'
         ax = df.plot.scatter(x='x', y='y', c='z', colormap=cm)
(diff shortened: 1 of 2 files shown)


pandas/tests/plotting/test_frame.py
```diff
@@ -1054,17 +1050,15 @@ def test_plot_scatter_with_c(self):
...
+        axes = df.plot.scatter(x='x', y='y', c='z')
 
-            if self.mpl_ge_1_3_1:
+        # default to Greys
+        assert axes.collections[0].cmap.name == 'Greys'
 
-                # n.b. there appears to be no public method to get the colorbar
-                # label
-                assert ax.collections[0].colorbar._label == 'z'
+        if self.mpl_ge_1_3_1:
```
jreback: @TomAugspurger we should be able to blow this code away as we > 1.4.3, yes (separate PR)

pandas/tests/plotting/test_frame.py
```diff
@@ -2172,14 +2166,27 @@ def test_invalid_kind(self):
...
+        (0, 1)
     ])
     def test_invalid_xy_args(self, x, y):
-        # GH 18671
+        # GH 18671 and # GH 20056
         df = DataFrame({"A": [1, 2], 'B': [3, 4], 'C': [5, 6]})
         with pytest.raises(ValueError):
             df.plot(x=x, y=y)
 
+    @pytest.mark.parametrize("x,y,colnames", [
```
jreback: can you add a test that checks the raise when passing ints and a non-int index

jreback (changes requested):
can you add a sub-section in the whatsnew about this in api-breaking changes. Note we *may* want to consider deprecating this and removing in a future version @jorisvandenbossche @TomAugspurger . though I am fine with just changing it.

jorisvandenbossche:
> Note we may want to consider deprecating this and removing in a future version @jorisvandenbossche @TomAugspurger . though I am fine with just changing it.

If it worked, I think we should deprecate it. But did it actually work? Because https://github.com/pandas-dev/pandas/issues/20056 seems to indicate not? (if it didn't work, of course fine to just remove)

jorisvandenbossche:
So it seems that for `line`, `bar`, `area` it doesn't work, but for `scatter`, `pie` and `hexbin` it does?

TomAugspurger:
I don't think we should remove this. I know positional plotting gets some use, especially for exploratory analysis when you have long column names.

I'll look more at why the original issue was failiing.

MasonGallo:
> So it seems that for line, bar, area it doesn't work, but for scatter, pie and hexbin it does?

@jorisvandenbossche yes I believe this is due to the structure of the code - the index setting I mentioned in #20056 only happens to `kind` not from that set - called `_dataframe_kinds` and `_series_kinds`.

> I don't think we should remove this. I know positional plotting gets some use, especially for exploratory analysis when you have long column names.

@TomAugspurger why not ask the user to supply `data.columns[ind]` instead of us making assumptions and calculating it for them? IMO that would make the API less complex.

TomAugspurger:
IME, DataFrame.plot is most often used in exploratory analysis. For this,
speed typing becomes an API consideration.
If you have a long column name, but want to quickly plot it, `df.plot(y=0)`
is nicer than `df.plot(y='percent_return_for_foo_over_bar')`.

> instead of us making assumptions

I don't think we ever make assumptions? If there's ever an integer in the
column we don't fall back to positional indexing.
IIUC there was just a bug in how we computed the positions when both `x`
and `y` are specified.

MasonGallo:
> IME, DataFrame.plot is most often used in exploratory analysis. For this,
speed typing becomes an API consideration.

That's fair - you have much more context than I do on usage.

> I don't think we ever make assumptions?

We make an assumption by allowing mixed types in column names - i.e. we check with `holds_integer` and then assume a given int is meant as a column name.

Seems like we don't have consensus on whether this should be removed from API. How should we proceed? I can close this and fix the integer indexing bug in #20000 ?

TomAugspurger:
> We make an assumption by allowing mixed types in column names - i.e. we check with holds_integer and then assume a given int is meant as a column name.

Perhaps it's just semantics then, but I wouldn't call that assuming :) To me the behavior is clear (but could be better documented). Specifying `x` and `y` as positions is allowed if and only if the columns holds no integers.

> I can close this and fix the integer indexing bug in #20000 ?

If you're comfortable with maybe wasting some effort, I would work around the bug in #20000. We can wait to hear what the other maintainers say before closing or working further on this issue, but I would be sad to see this go.

jorisvandenbossche:
> Perhaps it's just semantics then, but I wouldn't call that assuming :) To me the behavior is clear (but could be better documented). Specifying x and y as positions is allowed if and only if the columns holds no integers.

Although there might be some "guessing" cases (mixed integer columns), this is also behaviour that used a lot throughout the code base (eg `level` can often be specified both as integer or as name).

--

I don't have a strong feeling about this. If it is not too difficult to fix this to actually work, I am certainly fine with keeping it.

MasonGallo:
FYI: will be closing this in favor of #20000

Issue #15079, opened 2017-01-07
DataFrame.plot.box ignores by argument


ischurov:
#### Code Sample, a copy-pastable example if possible

```python
df = pd.DataFrame({'a': ['one', 'one', 'two', 'one', 'two', 'two'], 
                   'b': [1, 2, -1, 2, -2, -3]})
df.boxplot(by='a')
# two boxplots, one for 'one' and the other for 'two'
# this is an expected behaviour
```
![boxplot-ok](https://cloud.githubusercontent.com/assets/2717321/21741924/dfb41efa-d4f4-11e6-991e-65a4009b83c2.png)

```python
df.plot(kind='box', by='a')
# one boxplot, grouping ignored
```
![boxplot-kind](https://cloud.githubusercontent.com/assets/2717321/21741928/e94131a6-d4f4-11e6-9e39-645c033483b1.png)

```python
df.plot.box(by='a')
# one boxplot, grouping ignored
```
![boxplot-plot-box](https://cloud.githubusercontent.com/assets/2717321/21741933/0730a444-d4f5-11e6-94a2-157963f12a13.png)

#### Problem description

It seems that `by` argument is silently ignored and no grouping occurs. Probably, it is related to #9274.

#### Expected Output

I expect that three figures will be the same as the first one.

#### Output of ``pd.show_versions()``

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.2.final.0
python-bits: 64
OS: Darwin
OS-release: 16.3.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.19.0+311.gb895968
nose: 1.3.7
pip: 9.0.1
setuptools: 27.2.0
Cython: 0.25.1
numpy: 1.11.2
scipy: None
statsmodels: None
xarray: None
IPython: 5.1.0
sphinx: 1.5.1
patsy: None
dateutil: 2.6.0
pytz: 2016.10
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 1.5.3
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: 0.9999999
httplib2: None
apiclient: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.8
s3fs: None
pandas_datareader: None
</details>

[...]

PR #20152, opened 2018-03-10
DOC: update pandas.DataFrame.boxplot docstring. Fixes #8847 

mabelvj:
- [x] PR title is "DOC: update the <your-function-or-method> docstring"
- [x] The validation script passes: `scripts/validate_docstrings.py pandas.DataFrame.boxplot`
- [x] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] The html version looks good: `python doc/make.py --single pandas.DataFrame.boxplot`
- [x] It has been proofread on language by another sprint participant

```
################################################################################
##################### Docstring (pandas.DataFrame.boxplot) #####################
################################################################################

Make a box plot from DataFrame columns.

Make a box-and-whisker plot from DataFrame columns optionally grouped
by some other columns. A box plot is a method for graphically depicting
groups of numerical data through their quartiles.
The box extends from the Q1 to Q3 quartile values of the data,
with a line at the median (Q2).The whiskers extend from the edges
of box to show the range of the data. The position of the whiskers
is set by default to 1.5*IQR (IQR = Q3 - Q1) from the edges of the box.
Outlier points are those past the end of the whiskers.

For further details see
Wikipedia's entry for `boxplot <https://en.wikipedia.org/wiki/Box_plot>`_.

Parameters
----------
column : str or list of str, optional
    Column name or list of names, or vector.
    Can be any valid input to groupby.
by : str or array-like
    Column in the DataFrame to groupby.
ax : object of class matplotlib.axes.Axes, default `None`
    The matplotlib axes to be used by boxplot.
fontsize : float or str
    Tick label font size in points or as a string (e.g., ‘large’)
    (see `matplotlib.axes.Axes.tick_params
    <https://matplotlib.org/api/_as_gen/
    matplotlib.axes.Axes.tick_params.html>`_).
rot : int or float, default 0
    The rotation angle of labels (in degrees)
    with respect to the screen coordinate sytem.
grid : boolean, default `True`
    Setting this to True will show the grid.
figsize : A tuple (width, height) in inches
    The size of the figure to create in matplotlib.
layout : tuple (rows, columns) (optional)
    For example, (3, 5) will display the subplots
    using 3 columns and 5 rows, starting from the top-left.
return_type : {None, 'axes', 'dict', 'both'}, default 'axes'
    The kind of object to return. The default is ``axes``.

    * 'axes' returns the matplotlib axes the boxplot is drawn on.
[...]

PR #20373, opened 2018-03-15
DOC: update the pandas.DataFrame.plot.box docstring

4kxz:
A bit late to the party, but I did it.

```
#################### Docstring (pandas.DataFrame.plot.box)  ####################

Make a box plot of the DataFrame columns.

A box plot is a method for graphically depicting
groups of numerical data through their quartiles.
The box extends from the Q1 to Q3 quartile values of the data,
with a line at the median (Q2). The whiskers extend from the edges
of box to show the range of the data. The position of the whiskers
is set by default to 1.5*IQR (IQR = Q3 - Q1) from the edges of the
box. Outlier points are those past the end of the whiskers.

For further details see Wikipedia's
entry for `boxplot <https://en.wikipedia.org/wiki/Box_plot>`_.

A consideration when using this chart is that the box and the whiskers
can overlap, which is very common when plotting small sets of data.

Parameters
----------
by : string or sequence
    Column in the DataFrame to group by.
**kwds : optional
    Additional keywords are documented in

Returns
-------
axes : :class:`matplotlib.axes.Axes` or numpy.ndarray of them

See Also
--------
pandas.DataFrame.boxplot: Another method to draw a box plot.
pandas.Series.plot.box: Draw a box plot from a Series object.
matplotlib.pyplot.boxplot: Draw a box plot in matplotlib.

Examples
--------
Draw a box plot from a DataFrame with four columns of randomly
generated data.

.. plot::
    :context: close-figs

    >>> data = np.random.randn(25, 4)
    >>> df = pd.DataFrame(data, columns=list('ABCD'))
    >>> plot = df.plot.box()

Errors found:
	Errors in parameters section
		Unknown parameters {'**kwds'}
```

Due to issue #15079 I wasn't sure how to document the `by` argument. I will follow up to add it later.

Changed kwds description according to #20348.

### DIFF
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -3033,8 +3033,8 @@ class FramePlotMethods(BasePlotMethods):
         r"""
         Make a box plot of the DataFrame columns.
 
-        A box plot is a method for graphically depicting
-        groups of numerical data through their quartiles.
+        A box plot is a method for graphically depicting groups of numerical
+        data through their quartiles.
         The box extends from the Q1 to Q3 quartile values of the data,
         with a line at the median (Q2). The whiskers extend from the edges
         of box to show the range of the data. The position of the whiskers
@@ -3042,7 +3042,7 @@ class FramePlotMethods(BasePlotMethods):
         box. Outlier points are those past the end of the whiskers.
 
         For further details see Wikipedia's
-        entry for `boxplot <https://en.wikipedia.org/wiki/Box_plot>`_.
+        entry for `boxplot <https://en.wikipedia.org/wiki/Box_plot>`__.
 
         A consideration when using this chart is that the box and the whiskers
         can overlap, which is very common when plotting small sets of data.
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -3033,11 +3033,16 @@ class FramePlotMethods(BasePlotMethods):
         r"""
         Make a box plot of the DataFrame columns.
 
-        This method draws a box and whisker for each series in the DataFrame
-        in a single chart. The boxes share the vertical axis, which
-        makes makes it easier to compare if they are similar in scale.
-        Box plots are useful because they clearly show the median,
-        quartiles and flier points in the data.
+        A box plot is a method for graphically depicting
+        groups of numerical data through their quartiles.
+        The box extends from the Q1 to Q3 quartile values of the data,
+        with a line at the median (Q2). The whiskers extend from the edges
+        of box to show the range of the data. The position of the whiskers
+        is set by default to 1.5*IQR (IQR = Q3 - Q1) from the edges of the
+        box. Outlier points are those past the end of the whiskers.
+
+        For further details see Wikipedia's
+        entry for `boxplot <https://en.wikipedia.org/wiki/Box_plot>`_.
 
         A consideration when using this chart is that the box and the whiskers
         can overlap, which is very common when plotting small sets of data.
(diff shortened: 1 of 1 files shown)


pandas/plotting/_core.py
```diff
@@ -3031,19 +3031,46 @@ def barh(self, x=None, y=None, **kwds):
 
     def box(self, by=None, **kwds):
         r"""
-        Boxplot
+        Make a box plot of the DataFrame columns.
+
+        This method draws a box and whisker for each series in the DataFrame
+        in a single chart. The boxes share the vertical axis, which
+        makes makes it easier to compare if they are similar in scale.
```
dukebody: "makes makes" → "makes"

dukebody (approved):
Apart from the typo in the extended description, LGTM!

pandas/plotting/_core.py
```diff
@@ -3031,19 +3031,46 @@ def barh(self, x=None, y=None, **kwds):
...
+
+        Examples
+        --------
+        Draw a box plot from a DataFrame with four columns of randomly
+        generated data.
+
+        .. plot::
+            :context: close-figs
+
+            >>> data = np.random.randint(0, 10, size=(25, 4))
```
jorisvandenbossche: I think it is better to use normal distribution here (`np.random.randn(25, 4)`), it gives a bit more natural boxplots (eg with outliers) than a uniform distribution.

pandas/plotting/_core.py
```diff
@@ -3031,19 +3031,46 @@ def barh(self, x=None, y=None, **kwds):
...
 
         Returns
         -------
         axes : :class:`matplotlib.axes.Axes` or numpy.ndarray of them
+
+        See Also
+        --------
+        :meth:`pandas.DataFrame.boxplot` : Another method to draw a box plot.
+        :meth:`pandas.Series.plot.box` : Draw a box plot from a Series object.
+        :func:`matplotlib.pyplot.boxplot`: Draw a box plot in matplotlib.
```
jorisvandenbossche: The ``:meth:`.. `` part is not needed here, that is done automatically by sphinx.

pandas/plotting/_core.py
```diff
@@ -3031,19 +3031,46 @@ def barh(self, x=None, y=None, **kwds):
...
     def box(self, by=None, **kwds):
         r"""
-        Boxplot
+        Make a box plot of the DataFrame columns.
+
+        This method draws a box and whisker for each series in the DataFrame
+        in a single chart. The boxes share the vertical axis, which
+        makes it easier to compare if they are similar in scale.
+        Box plots are useful because they clearly show the median,
+        quartiles and flier points in the data.
```
jorisvandenbossche: There is also the PR on the `boxplot` method: https://github.com/pandas-dev/pandas/pull/20152. It has also extended the summary, you can check if there are some additions over there you could also use here (eg I think the wikipedia link is interesting) 

(ideally we would re-use this extended summary for both, but that is something we can fix later)

jorisvandenbossche (commented):

pandas/plotting/_core.py
Lkxz: I've tried to fix the issues that you commented.

The summary from #20152 is more explanatory so I've copied most of it here. I removed the mention to the `by` parameter which, unless I'm missing something, is not yet implemented for plot.box. Is someone working on that?

dukebody (approved):
IMO this is good enough. We can consider creating some shared doc parts for both `DataFrame.plot.box` and `DataFrame.boxplot` later.

jorisvandenbossche (approved):

jorisvandenbossche:
@Lkxz Thanks for the updates, and for the PR!

PR #20382, opened 2018-03-16
BUG: Handle all-NA blocks in concat

TomAugspurger:
Previously we special cased all-na blocks. We should only do that for
non-extension dtypes.


pandas/tests/extension/base/reshaping.py
```diff
@@ -25,6 +25,21 @@ def test_concat(self, data, in_frame):
         assert dtype == data.dtype
         assert isinstance(result._data.blocks[0], ExtensionBlock)
 
+    @pytest.mark.parametrize('in_frame', [True])
```
jschendel: Should this be `[True, False]`?  If `False` isn't supported yet, might be nice to write a quick comment, as separate code paths for `False` exist in the test.

TomAugspurger: Yeah, was debugging that case. Thanks.

jreback:
lgtm.

TomAugspurger:
The linting error was fixed on master: `pandas/_libs/tslibs/period.pyx:1358:12: W291 trailing whitespace`

PR #18202, opened 2017-11-09
DOC: document subclasses in API docs with selection of specific methods/attributes

TomAugspurger:
xref https://github.com/pandas-dev/pandas/issues/18147

OK, https://github.com/pandas-dev/pandas/commit/500ab71bda109327cf191a11fcf514eaffd70167 removes our hack in the vendored `numpydoc/numpydoc.py` to exclude autosummary generation for certain methods. Instead, we'll explicitly list the relevant methods in a new `Methods` section in those docs. This let's us include things like `IntervalIndex` in the API docs, without having all the regular index methods in the table, and without auto-gen making pages for all of those.

The downside is that we have to list the methods in `Methods` in the docstring and under a toctree in `api.rst`. But that's not so bad.

https://github.com/pandas-dev/pandas/commit/257fc8a0af7de854d0f5e2f0bbc9380320760c69 is removing all the warnings about references a non-existent document / this document isn't included in a toctree. But it's a bit tedious so I'll come back later (~600 warnings remaining :/ )

PR #20383, opened 2018-03-16
 DOC: update vendored numpydoc version 

jorisvandenbossche:
The first commit only updates our vendored version with the upstream one. 

Additional commits make it work with our docs:

- edit numpydoc to allow member listing for attributes (this is a change (and the only) to the source code of numpydoc that is not upstream: discussed in https://github.com/numpy/numpydoc/pull/106)
- numpydoc settings: use `numpydoc_use_blockquotes=False` for now for compatibility (it is deprecated). To fix this, we need to fix: css styling of description lists + warnings generated due to `**kwargs`

It is probably worth to look at the separate commits (and would also like to keep them when merging)


doc/sphinxext/numpydoc/docscrape_sphinx.py
```diff
@@ -33,6 +33,7 @@ def load_config(self, config):
         self.use_plots = config.get('use_plots', False)
         self.use_blockquotes = config.get('use_blockquotes', False)
         self.class_members_toctree = config.get('class_members_toctree', True)
+        self.attributes_as_param_list = config.get('attributes_as_param_list', True)
```
TomAugspurger: Will this ensure that Attributes is formatted like a table?

jorisvandenbossche: Yes (updated the top post). For now, that is the only change I made to the numpydoc source (it's in a separate commit).

But, if I am the only one that finds that better, I am OK to leave that.

jorisvandenbossche: Well, there isn't ;-) That's something I added here (but with only a few lines: https://github.com/pandas-dev/pandas/pull/20383/commits/075438f4b5ba14df925e79f00854c2d4b61e0cae), and that's what I proposed to do in https://github.com/numpy/numpydoc/pull/106, so it would be nice to have this in upstream)

jorisvandenbossche:
We still have the problem we fixed with this hack: https://github.com/pandas-dev/pandas/commit/70c9d318ed4ef32e1c294a4dbefcb783cda19dba, related to the property's that are None.

However, due to recent work, I think only `.index` and `.columns` are left for this (there are others there as well that don't show up properly, but that is because of the `if pydoc.getdoc(param_obj)` check and they have no docstring .. so that is easily fixable).

jacko8944:
Lundy06032013@gmail.com

jorisvandenbossche:
`.index` and `.columns` should be fixed with https://github.com/pandas-dev/pandas/pull/20385

jorisvandenbossche:
Updated this. 

The remaining related warnings I think are:

/tmp/doc/source/generated/pandas.Index.holds_integer.rst: WARNING: document isn't included in any toctree
/tmp/doc/source/generated/pandas.Index.is_type_compatible.rst: WARNING: document isn't included in any toctree
/tmp/doc/source/generated/pandas.Series.imag.rst: WARNING: document isn't included in any toctree
/tmp/doc/source/generated/pandas.Series.real.rst: WARNING: document isn't included in any toctree

this seems to be because they are not listed manually in api.rst, and are also not linked in the class docstring page (because they have no docstring they appear as plain table entry, not link to the docstring page. With as a result there is nowhere a link to that page).

TomAugspurger:
Would adding them to the `hidden` toctree at the bottom of `api.rst` be OK? It'd fix the warnings. Not sure what happens to the links...

jorisvandenbossche:
> Would adding them to the hidden toctree at the bottom of api.rst be OK? It'd fix the warnings. Not sure what happens to the links...

So now I need to remember again why we had those entries there at the bottom of api.rst ... Which you added in https://github.com/pandas-dev/pandas/pull/18202. Because this are attributes that were not yet included in api.rst before, and it is only now because they were no longer included in their class docstring page, there was a warning. 
But actually just giving them a proper docstring, should put them back in their class docstring page, and so fix the warning. So that is the more proper fix I think.

jorisvandenbossche:
OK, so it seems that as long as they are properly included in the class docstring page toctree, that is also fine, so in principle then we can get rid of the full list of hidden items in api.rst (but for that, we need to add docstrings to some of those methods). That's the proper fix for the warnings, but will leave that for another PR (now added them to the hidden list in api.rst)

TomAugspurger (approved):

jorisvandenbossche:
Travis doc build looks good

PR #18196, opened 2017-11-09
Allow accessing AxisProperties on classes

TomAugspurger:
xref https://github.com/pandas-dev/pandas/issues/18147

This will remove the doc warnings about DataFrame.index, Series.index, ...

Basically, before if you did `pd.DataFrame.index` you got an exception, about `None._data`, since there was no instance.

cc @jreback if you have any concerns.

PR #20385, opened 2018-03-16
DOC: enable docstring on DataFrame.columns/index

jorisvandenbossche:
Follow-up on https://github.com/pandas-dev/pandas/pull/18196, similar to https://github.com/pandas-dev/pandas/pull/19991

Didn't add an actual full docstring yet, but this already gives:

```
In [1]: pd.DataFrame.columns?
Type:        AxisProperty
String form: <pandas._libs.properties.AxisProperty object at 0x7fa975d74da0>
File:        ~/scipy/pandas/pandas/_libs/properties.cpython-35m-x86_64-linux-gnu.so
Docstring:   columns
```

instead of (on master, on 0.22 this raises an exception):

```
In [1]: pd.DataFrame.columns?
Type:        NoneType
String form: None
Docstring:   <no docstring>
```

### DIFF
--- a/pandas/_libs/properties.pyx
+++ b/pandas/_libs/properties.pyx
@@ -45,8 +45,6 @@ cdef class AxisProperty(object):
 
     cdef readonly:
         Py_ssize_t axis
-
-    cdef readonly:
         object __doc__
 
     def __init__(self, axis=0, doc=""):
--- a/pandas/_libs/properties.pyx
+++ b/pandas/_libs/properties.pyx
@@ -42,11 +42,16 @@ cache_readonly = CachedProperty
 
 
 cdef class AxisProperty(object):
-    cdef:
+
+    cdef readonly:
         Py_ssize_t axis
 
-    def __init__(self, axis=0):
+    cdef readonly:
+        object __doc__
+
+    def __init__(self, axis=0, doc=""):
         self.axis = axis
+        self.__doc__ = doc
 
     def __get__(self, obj, type):
         cdef:
@@ -54,7 +59,7 @@ cdef class AxisProperty(object):
 
         if obj is None:
             # Only instances have _data, not classes
-            return None
+            return self
         else:
             axes = obj._data.axes
         return axes[self.axis]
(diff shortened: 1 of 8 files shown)


pandas/_libs/properties.pyx
```diff
@@ -42,19 +42,24 @@ cache_readonly = CachedProperty
...
 
 cdef class AxisProperty(object):
-    cdef:
+
+    cdef readonly:
         Py_ssize_t axis
 
-    def __init__(self, axis=0):
+    cdef readonly:
+        object __doc__
```
TomAugspurger: Can this go under the same `def readonly` block as axis?

And do we want `object` or `str` for the type?

jorisvandenbossche: whoops, the above `cdef readonly` is a typo, I meant to keep it as `readonly`. 
I am not fully sure if it is actually needed for `axis`, or if we can also make it `readonly` (I would think it is never modified internally, bot not sure)

For the `object`, it is what is done for `CachedProperties` above as well. My cython is not good enough to really know the consequences. I think we usually use `object` for strings in our cython code.

jorisvandenbossche: If the test pass with using `readonly` for `axis`, I will combine them in one `cdef readonly`

jbrockmendel: `object` vs `str` is pretty harmless in this case.  I'd advocate sticking with `object` for now b/c then we don't need to worry about potential py2/py3 bytes/unicode corner cases.

pandas/core/generic.py
```diff
@@ -278,7 +278,7 @@ def _setup_axes(cls, axes, info_axis=None, stat_axis=None, aliases=None,
         if build_axes:
 
             def set_axis(a, i):
-                setattr(cls, a, properties.AxisProperty(i))
+                setattr(cls, a, properties.AxisProperty(i, a))
```
jbrockmendel: This `setattr` has always bothered me.  It kind of makes since in a general-case world with `PanelND`, but if we're only interested in `Series/DataFrame/[barely-]Panel` wouldn't it be clearer just to write:

```
class DataFrame(NDFrame):
    index = properties.AxisProperty(0)
    columns = properties.AxisProperty(1)
```
?

jorisvandenbossche: Yep, fully agree that if we only have 1D/2D this makes much more sense. But maybe we should only clean that up when Panel is removed?

jorisvandenbossche: Yep, fully agree that if we only have 1D/2D this makes much more sense. But maybe we should only clean that up when Panel is removed?

TomAugspurger (approved):

Issue #17778, opened 2017-10-04
DataFrame.to_pickle() fails for .zip format on MacOS and pandas 0.20.3


normanius:
#### Code Sample, a copy-pastable example if possible

```python
import pandas as pd
import numpy as np
df = pd.DataFrame(np.random.randn(3, 4), index=['A', 'B', 'C'])
df.to_pickle('out.zip')
#pd.read_pickle('out.zip')
```
#### Problem description
The below exception occurs. I do have writing permissions in the working directory. The code was working for pandas 0.19.0.

No problems observed for _bz2_ and _gzip_ compression (_xz_ I haven't tested).

```
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pandas/core/generic.py", line 1378, in to_pickle
    df.to_pickle('tmp.zip')
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pandas/io/pickle.py", line 27, in to_pickle
    is_text=False)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/site-packages/pandas/io/common.py", line 352, in _get_handle
    zip_file = zipfile.ZipFile(path_or_buf)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/zipfile.py", line 756, in __init__
    self.fp = open(file, modeDict[mode])
IOError: [Errno 2] No such file or directory: 'out.zip'
```
#### Expected Output
A zip file that one can re-read with `pandas.read_pickle()`.

#### Output of ``pd.show_versions()``

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 2.7.12.final.0
python-bits: 64
OS: Darwin
OS-release: 15.6.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.20.3
pytest: None
pip: 9.0.1
setuptools: 36.2.7
Cython: 0.26
numpy: 1.14.0.dev0+029863e
scipy: 0.18.1
xarray: None
IPython: 5.4.1
sphinx: None
patsy: None
dateutil: 2.6.1
pytz: 2017.2
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 2.0.2
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: 0.999999999
[...]

PR #20394, opened 2018-03-17
EHN: allow zip compression in `to_pickle`, `to_json`, `to_csv`

minggli:
- [x] closes https://github.com/pandas-dev/pandas/issues/17778
- [x] tests added / passed
- [x] whatsnew entry

Currently, `to_pickle` doesn't support compression='zip' whereas `read_pickle` is able to unpickle zip compressed file. The same applies to `to_json`, `to_csv` methods under generic, frame, series. 

Standard library ZipFile class default write method requires a filename but pickle.dump(obj, f, protocol) requires f being a file-like object (i.e. file handle) which offers writing bytes. Create a new BytesZipFile class that allows pickle.dump to write bytes into zip file handle. 

Now zip compressed objects (pickle, json, csv) are roundtripable with `read_pickle`, `read_json`, `read_csv`.

Need suggestion as to where to put BytesZipFile class which overrides `write` method with `writestr`. Other comments welcome.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -1654,9 +1654,9 @@ class DataFrame(NDFrame):
             A string representing the encoding to use in the output file,
             defaults to 'ascii' on Python 2 and 'utf-8' on Python 3.
         compression : string, optional
-            a string representing the compression to use in the output file,
-            allowed values are 'gzip', 'bz2', 'zip', 'xz', only used when the
-            first argument is a filename.
+            A string representing the compression to use in the output file.
+            Allowed values are 'gzip', 'bz2', 'zip', 'xz'. This input is only
+            used when the first argument is a filename.
         line_terminator : string, default ``'\n'``
             The newline character or character sequence to use in the output
             file
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -1655,8 +1655,9 @@ class DataFrame(NDFrame):
             defaults to 'ascii' on Python 2 and 'utf-8' on Python 3.
         compression : string, optional
             a string representing the compression to use in the output file,
-            allowed values are 'gzip', 'bz2', 'zip', 'xz',
-            only used when the first argument is a filename
+            allowed values are 'gzip', 'bz2', 'zip', 'xz', 'zip' only
+            supported with Python>=3.0, only used when the first argument is a
+            filename.
         line_terminator : string, default ``'\n'``
             The newline character or character sequence to use in the output
             file
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -1655,7 +1655,7 @@ class DataFrame(NDFrame):
             defaults to 'ascii' on Python 2 and 'utf-8' on Python 3.
         compression : string, optional
             a string representing the compression to use in the output file,
-            allowed values are 'gzip', 'bz2', 'xz',
+            allowed values are 'gzip', 'bz2', 'zip', 'xz',
             only used when the first argument is a filename
         line_terminator : string, default ``'\n'``
             The newline character or character sequence to use in the output
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -1655,9 +1655,8 @@ class DataFrame(NDFrame):
             defaults to 'ascii' on Python 2 and 'utf-8' on Python 3.
         compression : string, optional
             a string representing the compression to use in the output file,
-            allowed values are 'gzip', 'bz2', 'zip', 'xz', 'zip' only
-            supported with Python>=3.0, only used when the first argument is a
-            filename.
+            allowed values are 'gzip', 'bz2', 'zip', 'xz', only used when the
+            first argument is a filename.
         line_terminator : string, default ``'\n'``
             The newline character or character sequence to use in the output
             file
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -3633,8 +3633,8 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
             non-ascii, for python versions prior to 3
         compression : string, optional
             a string representing the compression to use in the output file,
-            allowed values are 'gzip', 'bz2', 'xz', only used when the first
-            argument is a filename
+            allowed values are 'gzip', 'bz2', 'zip', 'xz', only used when the
+            first argument is a filename
         date_format: string, default None
             Format string for datetime objects.
         decimal: string, default '.'
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -3632,9 +3632,9 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
             a string representing the encoding to use if the contents are
             non-ascii, for python versions prior to 3
         compression : string, optional
-            a string representing the compression to use in the output file,
-            allowed values are 'gzip', 'bz2', 'zip', 'xz', only used when the
-            first argument is a filename
+            A string representing the compression to use in the output file.
+            Allowed values are 'gzip', 'bz2', 'zip', 'xz'. This input is only
+            used when the first argument is a filename.
         date_format: string, default None
             Format string for datetime objects.
         decimal: string, default '.'
--- a/pandas/io/common.py
+++ b/pandas/io/common.py
@@ -363,20 +363,27 @@ def _get_handle(path_or_buf, mode, encoding=None, compression=None,
 
         # ZIP Compression
         elif compression == 'zip':
+            import io
             from zipfile import ZipFile
             # GH 17778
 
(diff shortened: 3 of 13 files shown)


pandas/io/common.py
```diff
@@ -425,6 +428,24 @@ def _get_handle(path_or_buf, mode, encoding=None, compression=None,
     return f, handles
 
 
+class BytesZipFile(ZipFile, BytesIO):
```
gfyoung: I personally like this location.

minggli:
ensure_clean() somehow produces same filename on one of the test but produces random filename as supposed to do locally and in other tests. strange.

minggli:
zip compression works for pickle in Python 2 now as well as Python 3, but csv zip compression doesn't in Python 2.

pandas/tests/frame/test_to_csv.py
```diff
@@ -919,30 +919,35 @@ def test_to_csv_path_is_none(self):
...
 
-    def test_to_csv_compression(self, compression_no_zip):
+    def test_to_csv_compression(self, compression):
 
         df = DataFrame([[0.123456, 0.234567, 0.567567],
                         [12.32112, 123123.2, 321321.2]],
                        index=['A', 'B'], columns=['X', 'Y', 'Z'])
 
+        if PY2 and compression == 'zip':
+            pytest.xfail(reason='zip compression for csv not suppported in'
```
jreback: this should be a skip

pandas/tests/frame/test_to_csv.py
```diff
@@ -919,30 +919,35 @@ def test_to_csv_path_is_none(self):
...
+            with tm.decompress_file(filename, compression) as fh:
                 text = fh.read().decode('utf8')
                 for col in df.columns:
                     assert col in text
 
-            with tm.decompress_file(filename, compression_no_zip) as fh:
+            with tm.decompress_file(filename, compression) as fh:
                 assert_frame_equal(df, read_csv(fh, index_col=0))
 
+    @pytest.mark.xfail(reason='zip compression is now supported for csv.')
```
jreback: why are you xfailing this?

pandas/tests/io/json/test_compression.py
```diff
@@ -5,22 +5,23 @@
...
+                                            compression=compression))
 
         # explicitly ensure file was compressed.
-        with tm.decompress_file(path, compression_no_zip) as fh:
+        with tm.decompress_file(path, compression) as fh:
             result = fh.read().decode('utf8')
         assert_frame_equal(df, pd.read_json(result))
 
 
+@pytest.mark.xfail(reason='zip compression is now supported for json.')
```
jreback: why are you xfailing this?

pandas/tests/series/test_io.py
```diff
@@ -138,26 +138,30 @@ def test_to_csv_path_is_none(self):
...
         assert isinstance(csv_str, str)
 
-    def test_to_csv_compression(self, compression_no_zip):
+    def test_to_csv_compression(self, compression):
 
         s = Series([0.123456, 0.234567, 0.567567], index=['A', 'B', 'C'],
                    name='X')
 
+        if PY2 and compression == 'zip':
+            pytest.xfail(reason='zip compression for csv not suppported in'
```
jreback: skip

pandas/tests/series/test_io.py
```diff
@@ -138,26 +138,30 @@ def test_to_csv_path_is_none(self):
...
+            s.to_csv(filename, compression=compression, header=True)
 
             # test the round trip - to_csv -> read_csv
-            rs = pd.read_csv(filename, compression=compression_no_zip,
+            rs = pd.read_csv(filename, compression=compression,
                              index_col=0, squeeze=True)
             assert_series_equal(s, rs)
 
             # explicitly ensure file was compressed
-            with tm.decompress_file(filename, compression_no_zip) as fh:
```
jreback: are there *any* uses of the compression_no_zip fixture left?

pandas/io/common.py
```diff
@@ -425,6 +428,18 @@ def _get_handle(path_or_buf, mode, encoding=None, compression=None,
...
```
jreback: can you add a little bit more to this class doc-strings. e.g. why its needed.

jreback (changes requested):

pandas/tests/series/test_io.py
minggli: I don't think so, the compression_no_zip fixture is solely for excluding zip compression in tests because writing zip compression had not been implemented.

pandas/tests/frame/test_to_csv.py
minggli: thanks for the comment. now it should not have to skip or xfail test_to_csv.

pandas/tests/frame/test_to_csv.py
minggli: this is an old test case that assert raising a BadZipFile exception when zip compression was not supported. so it will now fail the test because it doesn't no longer raise that exception. this test case is now redundant and removed in https://github.com/pandas-dev/pandas/pull/20394/commits/04886e9697cf7d6adf79ffaa939dec20983f6e9d

pandas/io/formats/csvs.py
```diff
@@ -150,6 +150,16 @@ def save(self):
 
             self._save()
 
+            # GH 17778 handles compression for byte strings.
```
jreback: why are you handling this here and not in the finally?

pandas/io/pickle.py
```diff
@@ -62,7 +62,6 @@ def to_pickle(obj, path, compression='infer', protocol=pkl.HIGHEST_PROTOCOL):
     2    2    7
     3    3    8
     4    4    9
-
```
jreback: can you add back the blank lines you removed

jreback (changes requested):

pandas/io/formats/csvs.py
minggli: moved to finally.

pandas/io/pickle.py
minggli: added back.

pandas/io/common.py
minggli: added. we currently don't have ability to write zip compressed pickle, json, csv, only read them. standard library ZipFile isn't designed exactly to produce a writable file handle.

pandas/tests/series/test_io.py
minggli: this skip or xfail is no longer needed to handle zip compression (write) in Python 2.

pandas/tests/io/json/test_compression.py
minggli: same as above.

minggli:
zip compression now should work both read/write for pickle, json, csv in Python 2/3.

minggli:
@jreback @jorisvandenbossche any comments on this PR?

pandas/core/frame.py
```diff
@@ -1655,8 +1655,8 @@ def to_csv(self, path_or_buf=None, sep=",", na_rep='', float_format=None,
             defaults to 'ascii' on Python 2 and 'utf-8' on Python 3.
         compression : string, optional
             a string representing the compression to use in the output file,
-            allowed values are 'gzip', 'bz2', 'xz',
-            only used when the first argument is a filename
+            allowed values are 'gzip', 'bz2', 'zip', 'xz', only used when the
+            first argument is a filename.
```
gfyoung: Let's fix this parameter description a bit:
~~~
a string representing the compression to use in the output file.
Allow values are 'gzip', 'bz2', 'zip', 'xz'.  This input is only used
when the first argument is a filename.
~~~

pandas/core/series.py
```diff
@@ -3633,8 +3633,8 @@ def to_csv(self, path=None, index=True, sep=",", na_rep='',
             non-ascii, for python versions prior to 3
         compression : string, optional
             a string representing the compression to use in the output file,
-            allowed values are 'gzip', 'bz2', 'xz', only used when the first
-            argument is a filename
+            allowed values are 'gzip', 'bz2', 'zip', 'xz', only used when the
+            first argument is a filename
```
gfyoung: Let's the fix the docstring here as I suggested for `frame.py`.

pandas/core/frame.py
minggli: done.

pandas/core/series.py
minggli: done.

minggli:
@gfyoung @jreback any comments?

jreback (approved):

jreback:
thanks @minggli keep em coming!

PR #20402, opened 2018-03-18
DOC: Only use ~ in class links to hide prefixes.

dukebody:
This is ported from https://github.com/python-sprints/python-sprints.github.io/pull/78/, as suggested in a comment.


jreback:
lgtm. @jorisvandenbossche

jorisvandenbossche:
@dukebody Thanks!

PR #20403, opened 2018-03-18
Named objs in asserts messages.

AutomateAaron:
### Feature
---

Ability to name objects being passed into `pandas.util.testing.assert_<type>_equal()` statements to allow for more helpful error messages Via the `obj` parameter.

#### Example:

```python
a = pd.Index([1, 2, 3, 4])
b = pd.Index([1, 2, 3, 5])

testing.assert_index_equal(a, b, obj=('Red', 'Blue'))
```
#### Output:
```diff
- AssertionError: Red and Blue are different
- 
- Red and Blue values are different (25.0 %)
- [Red]:  Int64Index([1, 2, 3, 4], dtype='int64')
- [Blue]: Int64Index([1, 2, 3, 5], dtype='int64')
```
---
However, messages & are still kept in tact.  For instance if I do a equal on two Dataframes with different values & pass in the names it will still append `.iloc[:, <idx>]` as shown:

#### Example:
```python
df1 = pd.DataFrame({'A': [1, 2, 3], 'B': [4, 5, 6]}, index=['a', 'b', 'c'])
df2 = pd.DataFrame({'A': [1, 2, 3], 'B': [6, 5, 4]}, index=['a', 'b', 'c'])

testing.assert_frame_equal(df1, df2, obj=('df1', 'df2'))
```
#### Output:
```diff
- AssertionError: df1.iloc[:, 1] and df2.iloc[:, 1] are different
- 
- df1.iloc[:, 1] and df2.iloc[:, 1] values are different (66.66667 %)
- [df1.iloc[:, 1]]: [4, 5, 6]
- [df2.iloc[:, 1]]: [6, 5, 4]
```
---
But, if this is undesired or unneeded, the default names, `('left', 'right')`, still persists.  As shown:

#### Example:
```python
testing.assert_index_equal(a, b)
```
#### Output:
```diff
- AssertionError: Index are different
- 
- Index values are different (25.0 %)
- [left]:  Int64Index([1, 2, 3, 4], dtype='int64')
- [right]: Int64Index([1, 2, 3, 5], dtype='int64')
```

It should be noted where were a _few_ cases where I had to change the default messages due to some hard-coded messages, I've updated the tests for those specific cases.  **but** if for some reason someone has something that relies on those messages being exactly what they were they will now break.

---
Checklist for other PRs:

- [x] closes #xxxx (N/A)
- [x] tests added / passed*
- [ ] whatsnew entry

\* There were 4 unrelated tests that were failing on upstream/master & those 4 still fail, but all the tests related to the files I edit still pass.

### DIFF
--- a/pandas/util/testing.py
+++ b/pandas/util/testing.py
@@ -1305,7 +1305,7 @@ def assert_frame_equal(left, right, check_dtype=True,
     # shape comparison
     if left.shape != right.shape:
         raise_assert_detail(obj,
-                            'DataFrame shape mismatch',
+                            '{obj} shape mismatch'.format(obj=join_obj(obj)),
                             '{shape!r}'.format(shape=left.shape),
                             '{shape!r}'.format(shape=right.shape))
 
--- a/pandas/util/testing.py
+++ b/pandas/util/testing.py
@@ -775,6 +775,8 @@ def assert_index_equal(left, right, exact='equiv', check_names=True,
         assertion message
     """
 
+    # obj = com._maybe_make_list(obj)
+
     def _check_types(l, r, obj='Index'):
         if exact:
             assert_class_equal(left, right, exact=exact, obj=obj)
@@ -801,14 +803,14 @@ def assert_index_equal(left, right, exact='equiv', check_names=True,
 
     # level comparison
     if left.nlevels != right.nlevels:
-        msg1 = '{obj} levels are different'.format(obj=obj)
+        msg1 = '{obj} levels are different'.format(obj=join_obj(obj))
         msg2 = '{nlevels}, {left}'.format(nlevels=left.nlevels, left=left)
         msg3 = '{nlevels}, {right}'.format(nlevels=right.nlevels, right=right)
         raise_assert_detail(obj, msg1, msg2, msg3)
 
     # length comparison
     if len(left) != len(right):
-        msg1 = '{obj} length are different'.format(obj=obj)
+        msg1 = '{obj} length are different'.format(obj=join_obj(obj))
         msg2 = '{length}, {left}'.format(length=len(left), left=left)
         msg3 = '{length}, {right}'.format(length=len(right), right=right)
         raise_assert_detail(obj, msg1, msg2, msg3)
@@ -833,7 +835,7 @@ def assert_index_equal(left, right, exact='equiv', check_names=True,
             diff = np.sum((left.values != right.values)
                           .astype(int)) * 100.0 / len(left)
             msg = '{obj} values are different ({pct} %)'.format(
-                obj=obj, pct=np.round(diff, 5))
+                obj=join_obj(obj), pct=np.round(diff, 5))
             raise_assert_detail(obj, msg, left, right)
     else:
         _testing.assert_almost_equal(left.values, right.values,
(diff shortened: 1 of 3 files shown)


pandas/util/testing.py
```diff
@@ -853,12 +858,25 @@ def _get_ilevel_values(index, level):
...
+                                     obj=map_obj('{obj} category', obj))
+
+
+def join_obj(obj):
+    obj = com._maybe_make_list(obj)
+    return ' and '.join(obj)
+
+
+def map_obj(form, objs, **kwargs):
+    assert isinstance(form, str)
```
jreback: don't put these here, rather in testing.pyx. I also don't particularly like these names.

pandas/util/testing.py
```diff
@@ -853,12 +858,25 @@ def _get_ilevel_values(index, level):
...
+    objs = com._maybe_make_list(objs)
+    return list(map(lambda obj: form.format(obj=obj, **kwargs), objs))
```
jreback: don't use ``_maybe_make_list``. we need to blow that away anyhow, or use it everywhere.

pandas/util/testing.py
```diff
@@ -874,12 +892,13 @@ def repr_class(x):
             # allow equivalence of Int64Index/RangeIndex
             types = set([type(left).__name__, type(right).__name__])
             if len(types - set(['Int64Index', 'RangeIndex'])):
-                msg = '{obj} classes are not equivalent'.format(obj=obj)
+                msg = '{obj} classes are not equivalent'.format(
```
jreback: not really sure I like having to handle this formatting everywhere.

pandas/util/testing.py
```diff
@@ -893,11 +912,13 @@ def assert_attr_equal(attr, left, right, obj='Attributes'):
         Attribute name being compared.
     left : object
     right : object
-    obj : str, default 'Attributes'
-        Specify object name being compared, internally used to show appropriate
+    obj : str or 2-length sequence (tuple, list, ...), default 'Attributes'
```
jreback: this is not correct, in any event you can just accept a list here (only), this is an internal routine

pandas/util/testing.py
```diff
@@ -958,35 +980,52 @@ def assert_categorical_equal(left, right, check_dtype=True,
...
+        Specify object(s) name(s) being compared, used to show appropriate
         assertion message
     check_category_order : bool, default True
         Whether the order of the categories should be compared, which
         implies identical integer codes.  If False, only the resulting
         values are compared.  The ordered attribute is
         checked regardless.
     """
+
+    # obj = com._maybe_make_list(obj)
```
jreback: don't comment things

pandas/util/testing.py
```diff
@@ -958,35 +980,52 @@ def assert_categorical_equal(left, right, check_dtype=True,
...
+
     _check_isinstance(left, right, Categorical)
 
+    obj = com._maybe_make_list(obj)
+
     if check_category_order:
         assert_index_equal(left.categories, right.categories,
-                           obj='{obj}.categories'.format(obj=obj))
+                           obj=map_obj('{obj}.categories', obj))
```
jreback: might be easier to push the formatting down and pass this like
``obj=obj, message='categories'`` (not sure if this makes it easier to read / better)

jreback (changes requested):
not really big on expanding this as it adds some complexity.

TomAugspurger:
I'm curious: what's the motivation for this?

Internal to pandas, we always use `result, expected` for the ordering, so that should be understood.

Are you using this elsewhere?

pandas/util/testing.py
AaronNBrock: As for right now, is there something better to use?

AaronNBrock:
@TomAugspurger, for a different project I had to write test cases that .cvs files & .json files matched but since I couldn't pass in names when the errors printed out it wasn't immediately easy to tell which file was wrong so I ended up having to Jerry rig something like this.  I figured someone else might also find it useful as well.

TomAugspurger:
Fair enough, just wanted to make sure since this seems like a fairly
pervasive change.

AaronNBrock:
At it's core, it's just changing "obj" to optionally be a list/tuple then adding some extra formatting at the end, but to do that requires to change each method that takes in obj to support it, so it's more of a refactoring change than a wide spread functionality change.

That being said, even now I do think the assertion messages could still be better & have more options.  For instance, it would be nice if instead of printing out _all_ of `left` & `right` there was an option to print out only the places where they are different.  (if, for instance, `left` & `right` are 100+ elements printing out the entire lists is kinda unhelpful).

Also, it seems a bit redundant to both print out `df1.iloc[:, 1] and df2.iloc[:, 1] are different` and `df1.iloc[:, 1] and df2.iloc[:, 1] values are different (66.66667 %)` it the same message since from the second statement the first is implied.

But those are projects for another weekend.

jreback:
I understand where you are going with this. but it adds quite a bit of complexity. ok with a simpler soln.

PR #20404, opened 2018-03-19
DEPR: deprecate get_ftype_counts (GH18243)

### LINKED ISSUE #18243: DEPR: deprecate get_ftype_counts
These are only to provide some inside into a sparse dtype. These should in theory be encoded in the dtypes (in pandas2), but in reality these are just cluttering the API and actually an implementation detail.

GGordonGordon:
Deprecate NDFrame.get_ftype_counts()

- [X] closes #18243
- [X] tests added / passed
- [X] whatsnew entry

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4677,6 +4677,8 @@ class NDFrame(PandasObject, SelectionMixin):
         """
         Return counts of unique ftypes in this object.
 
+        .. deprecated:: 0.23.0
+
         This is useful for SparseDataFrame or for DataFrames containing
         sparse arrays.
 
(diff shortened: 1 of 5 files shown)


pandas/core/generic.py
```diff
@@ -4707,6 +4709,45 @@ def get_ftype_counts(self):
         object:dense     1
         dtype: int64
         """
+        warnings.warn("get_ftype_counts is deprecated and will "
```
jreback: no need to create another function here, this is purely user facing (IOW don't create a _get_ftype_counts). keep the doc-string though.

pandas/tests/series/test_dtypes.py
```diff
@@ -64,7 +64,7 @@ def test_dtype(self):
         assert self.ts.ftypes == 'float64:dense'
         tm.assert_series_equal(self.ts.get_dtype_counts(),
                                Series(1, ['float64']))
-        tm.assert_series_equal(self.ts.get_ftype_counts(),
```
jreback: just assert the warning here

jreback (changes requested):

pandas/core/indexes/base.py
```diff
@@ -3396,8 +3396,11 @@ def map(self, mapper, na_action=None):
 
     def isin(self, values, level=None):
         """
+        Return a boolean array where the index values are in `values`.
```
TomAugspurger: Do you know where these changes are coming from? Ideally this PR will just do the deprecation.

TomAugspurger (changes requested):
There's a large file called `pd` in your PR that'll need to be deleted (our fault).

pandas/core/indexes/base.py
GGordonGordon: @TomAugspurger - Appears they snuck in with that large pd file under commit 4a43815d97811e1d4da78f495df3180e09eae7aa (GH20249)

TomAugspurger:
Could you try merging master into your branch? We had to do some history rewriting.

pandas/core/indexes/base.py
```diff
@@ -3396,11 +3396,8 @@ def map(self, mapper, na_action=None):
 
     def isin(self, values, level=None):
         """
-        Return a boolean array where the index values are in `values`.
```
TomAugspurger: Can you revert the changes here? Let me know if you need help with that.

GGordonGordon: @TomAugspurger - I had removed it as you had requested to only have the deprecation code in the branch / PR. So just to clarify you do want this code back in this branch but you don't want that large pr file back in the branch?

GGordonGordon:
@TomAugspurger - I've added the bases.py changes back in but left the large pd file out per your request.

TomAugspurger (approved):

TomAugspurger:
Thanks @GGordonGordon !

jorisvandenbossche:
@TomAugspurger the problem with the big .pd file, was that due to our rewriting of the history? I mean, will each PR that gets updated on master have this problem? (I don't see how you would get it from merging master or rebasing on master)

TomAugspurger:
I think it was because @GGordonGordon forked this branch from a master that included the big pd file. So anyone who created a branch during that ~7 hours might have the file in their PRs.

jorisvandenbossche:
OK, that should be not too many PRs then :)

PR #20405, opened 2018-03-19
Cythonized GroupBy Quantile

WillAyd:
I started this change with the intention of fully Cythonizing the GroupBy `describe` method, but along the way realized it was worth implementing a Cythonized GroupBy `quantile` function first. In theory we could concat together `count`, `mean`, `std`, `min`, `median`, `max`, and two `quantile` calls (one for 25% and the other for 75%) to get `describe`.

I decided to push this up as is because this alone is a decent amount of work and I think is worth getting reviewed before going further.

- [ ] closes #xxxx
- [X] tests added / passed
- [ ] whatsnew entry

Relevant ASVs provided below:

```bash
     [4a43815d]       [d7aec3fa]
-         236±6ms         421±40μs     0.00  groupby.GroupByMethods.time_dtype_as_field('float', 'quantile', 'transformation')
-         239±4ms         380±20μs     0.00  groupby.GroupByMethods.time_dtype_as_field('float', 'quantile', 'direct')
-         237±1ms          352±9μs     0.00  groupby.GroupByMethods.time_dtype_as_field('int', 'quantile', 'direct')
-         240±2ms         349±10μs     0.00  groupby.GroupByMethods.time_dtype_as_field('int', 'quantile', 'transformation')
-         348±6ms          372±9μs     0.00  groupby.GroupByMethods.time_dtype_as_group('int', 'quantile', 'transformation')
-         270±4ms         276±20μs     0.00  groupby.GroupByMethods.time_dtype_as_field('datetime', 'quantile', 'transformation')
-         357±6ms        347±0.9μs     0.00  groupby.GroupByMethods.time_dtype_as_group('int', 'quantile', 'direct')
-         271±4ms        263±0.6μs     0.00  groupby.GroupByMethods.time_dtype_as_field('datetime', 'quantile', 'direct')
-         515±9ms          351±2μs     0.00  groupby.GroupByMethods.time_dtype_as_group('datetime', 'quantile', 'transformation')
-         522±6ms          350±5μs     0.00  groupby.GroupByMethods.time_dtype_as_group('datetime', 'quantile', 'direct')
-         539±6ms          354±6μs     0.00  groupby.GroupByMethods.time_dtype_as_group('float', 'quantile', 'direct')
-         548±1ms          354±7μs     0.00  groupby.GroupByMethods.time_dtype_as_group('float', 'quantile', 'transformation')
```


pandas/tests/groupby/test_groupby.py
```diff
@@ -1898,6 +1871,82 @@ def test_intercept_builtin_sum(self):
...
+        # Timestamps
+        ([x for x in pd.date_range('1/1/18', freq='D', periods=5)],
+         [x for x in pd.date_range('1/1/18', freq='D', periods=5)[::-1]])
+    ])
+    @pytest.mark.parametrize('q', [0, .25, .5, .75, 1])
+    def test_quantile(self, interpolation, bar_vals, foo_vals, q):
+        # Fringe test case was not working as expected?
+        if (interpolation == 'nearest' and q == 0.5 and foo_vals == [
+                4, 3, 2, 1]):
+            pytest.skip("Unclear numpy expectation for nearest result with "
```
WillAyd: This is extremely nuanced but I was having trouble figuring out how NumPy would consistently handle the `nearest` tiebreaker when the provided quantile falls directly in between two values.

For instance:

```
In [2]: np.percentile([1,2,3,4], 50, interpolation='nearest')
Out[2]: 3

In [3]: np.percentile([2,4], 50, interpolation='nearest')
Out[3]: 2
```

It wasn't apparent why the first example above was rounding up and the second was rounding down. I glanced at the NumPy implementation which appears to use `np.around`. In that functions' documentation it says

```
For values exactly halfway between rounded decimal values, NumPy rounds to the nearest even value
```

But again that doesn't explain the behavior above. I'm wondering if the behavior here would even be well defined across different platforms?

pandas/core/groupby.py
```diff
@@ -1732,6 +1732,70 @@ def nth(self, n, dropna=None):
...
+
+        Returns
+        -------
+        Series or DataFrame
+            Return type determined by caller of GroupBy object.
+
+        See Also
+        --------
+        Series.quantile : Similar method for Series
+        DataFrame.quantile : Similar method for DataFrame
```
jreback: can you add numpy.percentile

pandas/core/groupby.py
```diff
@@ -1732,6 +1732,70 @@ def nth(self, n, dropna=None):
...
+        ...     [1, 2, 3, 4, 5, 5, 4, 3, 2, 1]],
+        ...    columns=['key', 'val'])
+        >>> df
+        """
+
+        is_dt = False
+        is_int = False
+
+        def pre_processor(vals):
+            if vals.dtype == np.object:
```
jreback: we really really need to clean this up and simply put this in a class. I would be really happy to do this *before* this PR.

pandas/core/groupby.py
```diff
@@ -1732,6 +1732,70 @@ def nth(self, n, dropna=None):
 
         return result
 
+    def quantile(self, q=0.5, interpolation='linear'):
+        """
+        Return group values at the given quantile, a la numpy.percentile.
+
+        Parameters
+        ----------
+        q : float or array-like, default 0.5 (50% quantile)
```
jreback: checking this should share code with Series.quantile.  (or be in cython is ok)

pandas/core/groupby.py
```diff
@@ -1732,6 +1732,70 @@ def nth(self, n, dropna=None):
...
+            0 <= q <= 1, the quantile(s) to compute
+        interpolation : str
```
jreback: can you list the methods here

jreback (changes requested):
before we change anything else. need to strip out all of this complexity in groupby with passing to cython.

jreback:
@WillAyd status?

WillAyd:
Will rebase against master in the next few days and see what this looks like again

pandas/tests/groupby/test_groupby.py
```diff
@@ -223,13 +223,13 @@ def f(x, q=None, axis=0):
     agg_result = df_grouped.agg(np.percentile, 80, axis=0)
     apply_result = df_grouped.apply(DataFrame.quantile, .8)
     expected = df_grouped.quantile(.8)
-    assert_frame_equal(apply_result, expected)
+    assert_frame_equal(apply_result, expected, check_names=False)
```
WillAyd: There is some inconsistency between the `apply_result` and the standard / agg counterparts, whereby the former is adding `.8` as the name. I don't think that's necessarily desirable but orthogonal to this change as the root cause there is somewhere within `.apply`

jbrockmendel:
@WillAyd LMK if I can be helpful in pushing this or #20024 over the finish line.

WillAyd:
Hmm actually @jbrockmendel any chance you can reproduce the errors showing up on Travis? I ran the test suite of Python 3.5 macOS locally without issue so not sure why they are failing on CI

jbrockmendel:
I can't reproduce this on OSX or Linux.  Maybe rebase?

pandas/_libs/groupby.pyx
```diff
@@ -380,5 +380,106 @@ def group_any_all(uint8_t[:] out,
...
+    cdef:
+        Py_ssize_t i, N=len(labels)
+        int64_t lab, ngroups, grp_sz, non_na_sz, grp_start=0, idx=0
+        uint8_t interp, offset
+        numeric val, next_val
+        float64_t q_idx, frac
+        ndarray[int64_t] counts, non_na_counts
+        ndarray[int64_t] sort_arr
+
+    assert <Py_ssize_t>len(values) == N
```
WillAyd: This check was throwing a build warning as the lhs was returning size_t whereas `len(labels)` was returning `Py_ssize_t` (eventually replaced with N since that was duplicative).

Not sure why `len` would have returned two different types though. May be a difference between `ndarray` and `numeric`

@jbrockmendel

jbrockmendel: A difference between ndarray and memoryview is plausible.  IIRC cython suggests using `arr.shape[0]` instead of `len(arr)` for memoryviews because the former is done wholly in C-space while the latter makes a python call.  I could imagine that being related to this, but am really just guessing.

WillAyd: Good call - using shape had the same effect as the cast though seems more idiomatic. Looks like I have some windows failures to look at but will push that up along with it. Good to know for future reference as well

WillAyd:
At a loss for windows failures here - can anyone help point out where the int/float mismatch is popping up in the test case? Maybe @chris-b1 ?

pandas/core/groupby/groupby.py
```diff
@@ -1670,6 +1670,72 @@ def nth(self, n, dropna=None):
...
+        inferences = {  # TODO (py27): replace with nonlocal
+            'is_dt': False,
+            'is_int': False
+        }
+
+        def pre_processor(vals):
+            if vals.dtype == np.object:
+                raise TypeError("'quantile' cannot be performed against "
+                                "'object' dtypes!")
+            elif vals.dtype == np.int:
```
chris-b1: I think your issue is this check - on windows np.int is 32 bit, and is probably too strict anyways.  Use [`is_integer_dtype`](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.api.types.is_integer_dtype.html#pandas-api-types-is-integer-dtype) instead

pandas/core/groupby/groupby.py
```diff
@@ -1670,6 +1671,72 @@ def nth(self, n, dropna=None):
...
+                inferences['is_dt'] = True
+
+            return vals
+
+        def post_processor(vals):
+            if inferences['is_dt']:
+                vals = vals.astype('datetime64[ns]')
+            elif inferences['is_int'] and interpolation in [
+                    'lower', 'higher', 'nearest']:
+                vals = vals.astype(np.int64)
```
WillAyd: @chris-b1 you were spot on with the issue - thanks! Is this the idiomatic way of casting back? Saw it elsewhere in the code base just not sure what implications may carry for windows

chris-b1:
Yep, I think that's the right way to do it. Any of the sized dtypes (int32,
int64, etc) won't be different on Windows, just the generic int.

pandas/core/groupby/groupby.py
```diff
@@ -1670,6 +1671,72 @@ def nth(self, n, dropna=None):
...
+        >>> df
+        """
+
+        inferences = {  # TODO (py27): replace with nonlocal
+            'is_dt': False,
+            'is_int': False
+        }
+
+        def pre_processor(vals):
+            if vals.dtype == np.object:
```
jreback: use is_object_dtype

pandas/tests/groupby/test_groupby.py
jreback: hmm, no this is definitily an api change you are making here, and not a good one. you *do* need to set the group name on the resulting index. we do this on purpose.

jreback (changes requested):

pandas/tests/groupby/test_groupby.py
WillAyd: What is the purpose of the API to set the index name to the value of the `q` argument via `apply` but not `agg`? 

Note that the `check_names=False` is applied to `agg_result` but I would think if anything the expectation is backwards on these, as I'm not aware of any other function that would do this

pandas/_libs/groupby.pxd
```diff
@@ -0,0 +1,6 @@
+cdef enum InterpolationEnumType:
```
jreback: IIRC we reverted using enum in other versions (because this would require an additional install in py2)?

jreback (changes requested):

pandas/_libs/groupby.pxd
WillAyd: Hmm we have something similar in algos:

https://github.com/pandas-dev/pandas/blob/decc8ce5601c32dc2313f0d2b1d611164262cae9/pandas/_libs/algos.pxd#L15

Do you know where that issue would have popped up?

jreback: ok great.

pandas/core/groupby/groupby.py
```diff
@@ -1688,6 +1690,72 @@ def nth(self, n, dropna=None):
...
+        Examples
+        --------
+        >>> df = pd.DataFrame(
+        ...    [['foo'] * 5 + ['bar'] * 5,
+        ...     [1, 2, 3, 4, 5, 5, 4, 3, 2, 1]],
+        ...    columns=['key', 'val'])
+        >>> df
+        """
+
+        inferences = {  # TODO (py27): replace with nonlocal
```
jreback: this is pretty janky I would change to do this

```
def pre_processor(vals):
     ....
     return vals, inferences

def post_processor(vals, inferences):
     ....
```
basically returning the state (or an empty dict) in the pre and passing it to the post

jreback: you will have to adjust the other callers of this but shouldn't be a big deal

jreback (changes requested):
lgtm. can add to 0.24.0 a whatsnew note

pandas/core/groupby/groupby.py
```diff
@@ -1924,10 +1992,15 @@ def _get_cythonized_result(self, how, grouper, aggregate=False,
...
-            Function to be applied to `values` prior to passing to Cython
-            Raises if `needs_values` is False
+            Function to be applied to `values` prior to passing to Cython.
+            Function should return a tuple where the first element is the
+            values to be passed to Cython and the second element is an
+            optional dictionary containing type inferences to be applied
+            after the Cython call. Raises if `needs_values` is False
         post_processing : function, default None
-            Function to be applied to result of Cython function
+            Function to be applied to result of Cython function. Should accept
```
jreback: maybe you want to show the signature as a comment of the Callable[object, dict]

jreback (changes requested):
lgtm. minor comment.

jreback:
is there an issue about this ? (don't create one if not)

WillAyd:
Hmm sorry got my issue numbers mixed up since I started this. There isn't one directly for this - it's progress towards https://github.com/pandas-dev/pandas/issues/19165 since this in theory could enable full Cythonization of `describe` (albeit in a separate PR) which would be the last thing on the list in that issue that I think we want to tackle

pandas/core/groupby/groupby.py
WillAyd: Are you asking to do that in this docstring or as a type hint in the implementations themselves?

jreback: both

pandas/core/groupby/groupby.py
```diff
@@ -1982,7 +2055,7 @@ def _get_cythonized_result(self, how, grouper, aggregate=False,
                 result = algorithms.take_nd(obj.values, result)
 
             if post_processing:
-                result = post_processing(result)
+                result = post_processing(result, inferences)
```
WillAyd: FYI this will fail if we try to pass in `post_processing` with `pre_processing` as `inferences` will be undefined in that case. Since it's all internal I suppose not a big deal for now, but could also add an explicit check at the start of the function if it doesn't add too much bloat.

WillAyd (commented):
Fine to add type comments and think they'd be very useful since the call pattern here is rather complicated, though wondering if we want to wait until after dropping Py2 support then so we don't have to use the compatibility signature and since we already cut the 0.24 RC.

Rather indifferent either way just offering that up; no rush on this PR

pandas/core/groupby/groupby.py
```diff
@@ -1966,7 +2041,7 @@ def _get_cythonized_result(self, how, grouper, aggregate=False,
             if needs_values:
```
jreback: just define ``inference = None`` at the top then you wont ever have a NameError

pandas/core/groupby/groupby.py
```diff
@@ -1924,10 +1993,16 @@ def _get_cythonized_result(self, how, grouper, aggregate=False,
...
+            Function should return a tuple where the first element is the
+            values to be passed to Cython and the second element is an optional
+            type which the values should be converted to after being returned
+            by the Cython operation. Raises if `needs_values` is False.
         post_processing : function, default None
-            Function to be applied to result of Cython function
+            Function to be applied to result of Cython function. Should accept
+            an array of values as the first argument and type inferences as its
+            second argument, i.e. the signature should be
+            (ndarray, typing.Type).
```
jreback: not really what I meant, this is a (ndarray, dict)

jreback (changes requested):

pandas/core/groupby/groupby.py
```diff
@@ -1688,6 +1692,71 @@ def nth(self, n, dropna=None):
...
+        DataFrame.quantile : Similar method for DataFrame
+        numpy.percentile : NumPy method to compute qth percentile
+
+        Examples
+        --------
+        >>> df = pd.DataFrame(
+        ...    [['foo'] * 5 + ['bar'] * 5,
+        ...     [1, 2, 3, 4, 5, 5, 4, 3, 2, 1]],
+        ...    columns=['key', 'val'])
+        >>> df
```
topper-123: This example misses the groupby and quantile.

doc/source/whatsnew/v0.24.0.rst
```diff
@@ -424,6 +424,7 @@ Other Enhancements
 - :func:`pandas.DataFrame.to_sql` has gained the ``method`` argument to control SQL insertion clause. See the :ref:`insertion method <io.sql.method>` section in the documentation. (:issue:`8953`)
 - :meth:`DataFrame.corrwith` now supports Spearman's rank correlation, Kendall's tau as well as callable correlation methods. (:issue:`21925`)
 - :meth:`DataFrame.to_json`, :meth:`DataFrame.to_csv`, :meth:`DataFrame.to_pickle`, and :meth:`DataFrame.to_XXX` etc. now support tilde(~) in path argument. (:issue:`23473`)
+- :class:`pandas.core.groupby.GroupBy` now has a built-in implementation of :meth:`pandas.core.groupby.GroupBy.quantile`. (:issue:`20405`)
```
jreback: this is not necessary here
instead put in a performance note

pandas/tests/groupby/test_function.py
```diff
@@ -1060,6 +1060,55 @@ def test_size(df):
     tm.assert_series_equal(df.groupby('A').size(), out)
 
 
+# quantile
+# --------------------------------
```
jreback: side note this file is getting pretty big, maybe should split it up a bit (later)

jreback:
lgtm. pls merge master and ping on green.

WillAyd:
Hmm I think green for the actual change. I think the failures are for any PRs with changes to the benchmarks so unrelated?

jreback (approved):

jreback:
thanks @WillAyd

Issue #19825, opened 2018-02-21
Ensure S3 tests are mocked


TomAugspurger:
Anaconda owns the `pandas-test` S3 bucket that we use for some of our tests. They're doing some spring cleaning, and would like to remove this bucket sometime next week.

We'll need to make sure that all of our tests are mocked.

From what I can tell, `pandas/tests/io/parser/test_network::TestS3` is not mocked.

We also have an ASV benchmark that hits S3, whose goal is to ensure that we don't read the whole file when doing `read_csv(..., nrows=10)`. I think we could remove that and replace it with a mocked fixture that asserts something about the number of bytes read.

jreback:
we could also sign up for an aws account with NuMFOCUS. would be minimal cost I think.

TomAugspurger:
Yeah, it'd be good to have one non-mocked test under a bucket we have
control over.

TomAugspurger:
We've temporarily set the pandas-test bucket to be unreadable, and it'll be going away entirely soon. Tests still pass for me locally, so hopefully things are OK.

There's one reference to `s3://pandas-test` in the docs, but that's in a non-executed code-block.

PR #20409, opened 2018-03-19
REF: Mock all S3 Tests

TomAugspurger:
Closes https://github.com/pandas-dev/pandas/issues/19825

### DIFF
--- a/pandas/tests/io/parser/test_network.py
+++ b/pandas/tests/io/parser/test_network.py
@@ -184,7 +184,7 @@ class TestS3(object):
         tm.assert_frame_equal(result, expected)
 
     def test_read_csv_chunked_download(self, s3_resource, caplog):
-        # 8 MB, S3FS usees ~5MB chunks
+        # 8 MB, S3FS usees 5MB chunks
         df = DataFrame(np.random.randn(100000, 4), columns=list('abcd'))
         buf = BytesIO()
         str_buf = StringIO()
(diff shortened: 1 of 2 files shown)


pandas/tests/io/parser/test_network.py
```diff
@@ -188,3 +182,22 @@ def test_read_csv_handles_boto_s3_object(self,
...
+
+        df.to_csv(str_buf)
+
+        buf = BytesIO(str_buf.getvalue().encode('utf-8'))
+
+        s3_resource.Bucket("pandas-test").put_object(
+            Key="large-file.csv",
+            Body=buf)
+
+        with caplog.at_level(logging.DEBUG, logger='s3fs.core'):
```
TomAugspurger: @martindurant does this seem like a reasonable way to test that `read_csv(key, nrows=5)` only triggers S3FS reading part of the object? Do you know of a better way, that's perhaps less reliant on the internals of S3FS?

martindurant: I'm afraid I don't have a better method for you, s3fs doesn't keep a log of transactions in any data structure you could access, and the s3file used for the download will have been cleaned up as soon as read_csv is done with it.

TomAugspurger: Hmm OK. Do you know if moto keeps a record anywhere?

I dislike this test since s3fs adding an additional `logger.debug` anywhere, or changing the log message, default bytes size, etc. will break it.

boto also has a callback mechanism on `download_file`, but I don't see that option for `get_object`. If I can't figure out a way to get that working, I'll try to make the test using the logger a bit less fragile.

martindurant: Certainly you could look through all log messages captured, not just the last one. Note that you *do* have access to the exact s3filesystem in S3FileSystem._singleton[0], but I don't see that that helps you in this case.

You could maybe patch `S3File.__exit__` to store the values of `self.loc` and `self.end`?

jreback:
lgtm. @TomAugspurger merge when you are ok with this.

PR #20412, opened 2018-03-19
BUG: ExtensionArray.fillna for scalar values

### LINKED ISSUE #20411: ExtensionArray.__fillna__ should not use lib.is_scalar
`lib.is_scalar` is a bit too strict for the check we do in https://github.com/pandas-dev/pandas/blob/7273ea0709590e6264607f227bb8def0ef656c50/pandas/core/arrays/base.py#L272. For example, it doesn't condisder `ipaddress.IPv4Address` as a scalar.

```python
In [1]: from pandas._libs import lib

In [2]: import ipaddress

In [3]: lib.is_scalar(ipaddress.IPv4Address(1))
Out[3]: False
```

Replacing that check with a `hasattr(value, '__len__')` should be sufficient.

TomAugspurger:
Closes #20411

### DIFF
--- a/pandas/core/arrays/base.py
+++ b/pandas/core/arrays/base.py
@@ -261,7 +261,7 @@ class ExtensionArray(object):
         -------
         filled : ExtensionArray with NA/NaN filled
         """
-        from pandas.api.types import is_scalar
+        from pandas.api.types import is_array_like
         from pandas.util._validators import validate_fillna_kwargs
         from pandas.core.missing import pad_1d, backfill_1d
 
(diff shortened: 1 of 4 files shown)


jreback:
thanks @TomAugspurger 

may want to think about someway of augmenting our is_scalar check that works on EA's.

jorisvandenbossche:
> may want to think about someway of augmenting our is_scalar check that works on EA's

I am not sure that will be possible (or desirable), because in principle such scalars can be anything, including lists and dicts

PR #20291, opened 2018-03-12
DOC: update `pandas/core/ops.py` docstring with examples for pandas.DataFrame.gt and .ge methods

ParfaitG:
Checklist for the pandas documentation sprint:

- [X] PR title is "DOC: update the <your-function-or-method> docstring"
- [ ] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [X] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [X] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [X] It has been proofread on language by another sprint participant

Please include the output of the validation script below between the "```" ticks:

```
################################################################################
####################### Docstring (pandas.DataFrame.gt)  #######################
################################################################################

Wrapper for flexible comparison methods gt

################################################################################
################################## Validation ##################################
################################################################################

Errors found:
	Docstring text (summary) should start in the line immediately after the opening quotes (not in the same line, or leaving a blank line in between)
	Closing quotes should be placed in the line after the last text in the docstring (do not close the quotes in the same line as the text, or leave a blank line between the last text and the quotes)
	Summary does not end with dot
	No extended summary found
	Errors in parameters section
		Parameters {'axis', 'level', 'other'} not documented
	No returns section found
	See Also section not found
	No examples section found

################################################################################
####################### Docstring (pandas.DataFrame.ge)  #######################
################################################################################

Wrapper for flexible comparison methods ge

################################################################################
################################## Validation ##################################
################################################################################

Errors found:
	Docstring text (summary) should start in the line immediately after the opening quotes (not in the same line, or leaving a blank line in between)
	Closing quotes should be placed in the line after the last text in the docstring (do not close the quotes in the same line as the text, or leave a blank line between the last text and the quotes)
	Summary does not end with dot
	No extended summary found
	Errors in parameters section
		Parameters {'level', 'other', 'axis'} not documented
	No returns section found
	See Also section not found
	No examples section found

```

If the validation script still gives errors, but you think there is a good reason
to deviate in this case (and there are certainly such cases), please state this
explicitly.

Comparison methods in `ops.py` including `.gt()` and `.ge()` do not have standard docstring like other methods. They use a generalized wrapper summary line. This pull request simply adds examples in line with recent merge for `.add()` method: https://github.com/pandas-dev/pandas/pull/20246/files

PR #20415, opened 2018-03-20
DOC: Updating operators docstrings

ParfaitG:
an unrelated PR):

```
####################### Docstring (pandas.DataFrame.gt)  #######################

Wrapper for flexible comparison methods gt

Examples
--------

>>> df1 = pd.DataFrame({'num1': range(1,6),
...                     'num2': range(2,11,2),
...                     'num3': range(1,20,4)})
>>> df1
   num1  num2  num3
0     1     2     1
1     2     4     5
2     3     6     9
3     4     8    13
4     5    10    17
>>> df2 = pd.DataFrame({'num1': range(6,11),
...                     'num2': range(1,10,2),
...                     'num3': range(1,20,4)})
>>> df2
   num1  num2  num3
0     6     1     1
1     7     3     5
2     8     5     9
3     9     7    13
4    10     9    17
>>> df1.gt(df2)
    num1  num2   num3
0  False  True  False
1  False  True  False
2  False  True  False
3  False  True  False
4  False  True  False

Errors found:
	Summary does not end with dot
	No extended summary found
	Errors in parameters section
		Parameters {'other', 'level', 'axis'} not documented
	No returns section found
	See Also section not found

####################### Docstring (pandas.DataFrame.ge)  #######################

Wrapper for flexible comparison methods ge

Examples
--------

>>> df1 = pd.DataFrame({'num1': range(1,6),
...                     'num2': range(2,11,2),
...                     'num3': range(1,20,4)})
>>> df1
   num1  num2  num3
0     1     2     1
1     2     4     5
2     3     6     9
3     4     8    13
4     5    10    17
>>> df2 = pd.DataFrame({'num1': range(6,11),
...                     'num2': range(1,10,2),
...                     'num3': range(1,20,4)})
>>> df2
   num1  num2  num3
0     6     1     1
1     7     3     5
2     8     5     9
3     9     7    13
4    10     9    17
>>> df1.ge(df2)
    num1  num2   num3
0  False  True   True
1  False  True   True
2  False  True   True
3  False  True   True
4  False  True   True

Errors found:
	Summary does not end with dot
	No extended summary found
	Errors in parameters section
		Parameters {'other', 'axis', 'level'} not documented
	No returns section found
	See Also section not found



```
If the validation script still gives errors, but you think there is a good reason to deviate in this case (and there are certainly such cases), please state this explicitly.

This PR follows up on previous attempt: https://github.com/pandas-dev/pandas/pull/20291. Please see discussion.

Specifically, comparison methods in ops.py including `.gt()` and `.ge()` do not have standard docstring like other methods. They use a generalized wrapper summary line. This pull request simply adds examples by conditionally building docstring with examples. General changes include: adding global string variable, __flex_comp_doc_FRAME_ and `if` logic to the method, *_flex_comp_method_FRAME*.


pandas/core/ops.py
```diff
@@ -582,6 +642,14 @@ def _get_op_name(op, special):
 DataFrame.{reverse}
 """
 
+_flex_comp_doc_FRAME = """
+Wrapper for flexible comparison methods {name}
```
jbrockmendel: This could use a description of what “flexible” means in this context.

pandas/core/ops.py
```diff
@@ -452,11 +512,11 @@ def _get_op_name(op, special):
     'gt': {'op': '>',
            'desc': 'Greater than',
            'reverse': None,
-           'df_examples': None},
+           'df_examples': _gt_example_FRAME},
     'ge': {'op': '>=',
            'desc': 'Greater than or equal to',
            'reverse': None,
-           'df_examples': None}}
+           'df_examples': _ge_example_FRAME}}
```
jbrockmendel: Any reason why you chose these two?

pandas/core/ops.py
```diff
@@ -397,6 +397,66 @@ def _get_op_name(op, special):
...
+>>> df1
+   num1  num2  num3
+0     1     2     1
+1     2     4     5
+2     3     6     9
+3     4     8    13
+4     5    10    17
+>>> df2 = pd.DataFrame({'num1': range(6,11),
+...                     'num2': range(1,10,2),
+...                     'num3': range(1,20,4)})
```
jbrockmendel: In case the doc linter cares,about it, add spaces after commas here

jbrockmendel:
This looks close to ready.  Pls rebase.

jbrockmendel (changes requested):
See comments inline.

pandas/core/ops.py
ParfaitG: We maintained original text and only added examples. Please advise on text changes.

pandas/core/ops.py
ParfaitG: This was part of the Pandas Doc Sprint some months ago. I was tasked with ge/gt functions with teammates handling others. But we can incorporate le/lt and eq/ne pairs with similar set of examples.

pandas/core/ops.py
jbrockmendel: Pls add a space after each of the commas in the `range` calls.

pandas/core/ops.py
```diff
@@ -1546,8 +1614,18 @@ def na_op(x, y):
...
+    doc = ('Wrapper for flexible comparison methods {name}'
+           .format(name=op_name))
+
+    if op_name in _op_descriptions:
+        op_desc = _op_descriptions[op_name]
+
+        if op_desc['df_examples'] is not None:
+            base_doc = _flex_comp_doc_FRAME
+            doc = base_doc.format(name=op_name,
+                                  df_examples=op_desc['df_examples'])
```
jbrockmendel: Pls add a `.strip()` to `op_desc['df_examples']`

jbrockmendel:
Needs rebase.  Aside from that and two small requested edits, this is ready pending OK from @datapythonista .

datapythonista:
Thanks for the contribution @ParfaitG.

Your examples look good, but if you have time it'd be cool to add some extra stuff to this PR (sorry we couldn't provide more clear information during the sprint).

I'd personally reuse a single docstring for `eq`, `ne`, `gt`, `gte`, `lt`, `lte`. I think we could provide a bit longer description, mainly for new users, but also explain how these methods are more flexible than the equivalent operators (`==`, `!=`, `>`...) as the axis and levels can be specified.

Then, it'd be good to have the `Parameters` section.

Also, the `See Also` section can be included to link to all these methods sharing the docstring.

For the examples, what you did looks good, but it could help the users if the dataframes are a bit smaller (something like 2 cols and 3 rows), and the example is a bit more meaningful (e.g. instead of columns `num1`... a value that users can understand and quickly compare between dataframes).

How does this sound?

ParfaitG:
Understood @datapythonista ... please see updated PR with a more comprehensive revision on the comparison operators.

pandas/core/ops.py
```diff
@@ -582,6 +751,35 @@ def _get_op_name(op, special):
...
+other : DataFrame
+axis : {{0, 1, 'columns', 'rows'}}
+level : int or name
+    Broadcast across a level, matching Index values on the
+    passed MultiIndex level
+
+Returns
+-------
+result : DataFrame
+    Consisting of boolean values
```
datapythonista: it's a bit redundant, but a bit more explanation, like "Result of the comparison.". Note the period at the end.

pandas/core/ops.py
```diff
@@ -397,6 +397,169 @@ def _get_op_name(op, special):
...
+DataFrame.lt : Return DataFrame of boolean values strictly less
+    than the elementwise rows or columns of one DataFrame to
+    another
+DataFrame.ge : Return DataFrame of boolean values greater than or
+    equal to the elementwise rows or columns of one DataFrame to
+    another
+DataFrame.gt : Return DataFrame of boolean values strictly greater
+    than to the elementwise rows or columns of one DataFrame to
+    another
+"""
```
datapythonista: I'm not sure if I'm missing something, but is this always the same for all methods? It could go directly to the doctring and not  in a separate variable if that's the case. Besides good docstrings, it's good if we keep the code as simple as possible. So we usually leave the same `See Also` for all methods in this case, even if the method being documented is self referencing.

Also, do you think we can find something less verbose for the descriptions. Something like `Compare DataFrames for equality elementwise` may be?

pandas/core/ops.py
```diff
@@ -582,6 +751,35 @@ def _get_op_name(op, special):
...
+result : DataFrame
+    Consisting of boolean values
+
+Examples
+--------
+{df_examples}
+
+See also
+--------
+{reverse}
```
datapythonista: `See Also` should go before examples. Also, note the capital `A`.

Can you run `./scripts/validate_docstring.py pandas.DataFrame.eq` after the next set of changes to validate these details?

datapythonista (commented):
Thanks for the changes @ParfaitG, looking much better now. I'd still change some things, mainly to simplify the code.

pandas/core/ops.py
```diff
@@ -397,6 +397,169 @@ def _get_op_name(op, special):
 e  NaN  -2.0
 """
 
+_eq_example_FRAME = """
+>>> df1 = pd.DataFrame({'tool': ['python', 'r', 'julia'],
+...                     'score': [100, 200, 300]},
+...                     columns=['tool', 'score'])
```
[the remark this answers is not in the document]
ParfaitG: Columns specify order here as score will precede tool due to alphabetical order. But I will adjust with new examples.

pandas/core/ops.py
```diff
@@ -397,6 +397,169 @@ def _get_op_name(op, special):
...
+>>> df1
+     tool  score
+0  python    100
+1       r    200
+2   julia    300
```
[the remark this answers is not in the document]
ParfaitG: Understood. Will add a new example. Just thought in the spirit of open source!

pandas/core/ops.py
```diff
@@ -397,6 +397,169 @@ def _get_op_name(op, special):
...
+>>> df2
+     tool  score
+0  python    300
+1       r    200
+2   julia    100
+>>> df1.ge(df2)
+   tool  score
+0  True  False
+1  True   True
+2  True   True
```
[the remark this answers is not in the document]
ParfaitG: All examples uses the same *df1* and *df2*. But I will work on showing the use of *rows* and *columns* axis args.

pandas/core/ops.py
datapythonista: No, what I meant is that if it's possible, it'd be good that each of these descriptions is shorter. Just making them more concise if you find the way.

pandas/core/ops.py
```diff
@@ -582,6 +751,35 @@ def _get_op_name(op, special):
...
+Flexible wrappers to comparison operators (specifically ``{name}``).
+
+Wrappers (``eq``, ``ne``, ``le``, ``lt``, ``ge``, ``gt``) are equivalent to
+operators (``==``, ``=!``, ``<=``, ``<``, ``>=``, ``>``) with support to choose
+axis (rows or columns) for comparison.
+
+Parameters
+----------
+other : DataFrame
+axis : {{0, 1, 'columns', 'rows'}}
```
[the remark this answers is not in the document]
datapythonista: yep, still plenty of work in many docstrings, but we're getting closer. :)

pandas/core/ops.py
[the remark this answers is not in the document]
datapythonista: Good point, in that case I'd use a list of tuples, with the `columns` argument. But just a personal opinion, whatever you find clearer.

pandas/core/ops.py
datapythonista: We still need to add several validations to that script. The order of the sections is one of them.

datapythonista:
Thanks for the update @ParfaitG. It looks better, but as we discussed before, I don't think it is as useful to have the example of each comparison in each page, than to show well why the methods exists, how they are different from the operators, and how you can use them on different axis.

Yesterday I was working on a tutorial that among other things covers this. You can take a look at the `pandas DataFrame` section to get the idea of what I mean: https://github.com/datapythonista/pandas-tutorials/blob/master/02_Data_structures.ipynb

What I think we should do: Get rid of all the `_eq_example_FRAME`... variables, and move the example to the main docstring (just one method to start with). Same for `_comp_others`. If I'm not missing anything, the same exact content is being using in every docstring anyway (so, there is no value on keeping it in a separate variable).

Then, the example should show:
- Comparing with a scalar (`other` parameter) and the rest of parameters by default.
- Comparing with another `DataFrame`.
- Comparing with `Series`, show how the `axis` parameter changes the behavior.
- Comparing with a `MultiIndex`, show how the `level` parameter changes the behavior.
- Showing the cases where the operator is equivalent to the method

As there are many cases to show, you can keep using the different methods in each of them, so more or less all them are covered. Meaning that when you compare with a scalar you can use `eq`, when you compare with a `DataFrame` you can use `gt`, and so on.

Let me know how does this sound, and if there is anything I can help with.

pandas/core/ops.py
```diff
@@ -436,27 +436,33 @@ def _get_op_name(op, special):
     'eq': {'op': '==',
            'desc': 'Equal to',
            'reverse': None,
-           'df_examples': None},
+           'df_examples': None,
+           'others': None},
```
datapythonista: If we don't use `reverse`, `df_examples` and `others`, can we get rid of them instead of having all them to `None`?

pandas/core/ops.py
```diff
@@ -582,6 +588,150 @@ def _get_op_name(op, special):
 DataFrame.{reverse}
 """
 
+_flex_comp_doc_FRAME = """
+Flexible wrappers to comparison operators (`eq`, `ne`, `le`, `lt`, `ge`, `gt`).
+
+Equivalent to `==`, `=!`, `<=`, `<`, `>=`, `>` with support to choose
```
datapythonista: I'm a bit confused... In the previous dictionaries, you define `op`, `desc`... and you also have `name` which is the key. Shouldn't we use them here? So, we explain in each page the documented method and not all them?

pandas/core/ops.py
```diff
@@ -582,6 +588,150 @@ def _get_op_name(op, special):
...
+axis (rows or columns) and level for comparison.
+
+Parameters
+----------
+other : constant, list, tuple, ndarray, Series, or DataFrame
```
datapythonista: I think `scalar` is preferred over `constant`. For `list`, `tuple` and `ndarray` I think you can abbreviate them as `sequence`.

pandas/core/ops.py
```diff
@@ -582,6 +588,150 @@ def _get_op_name(op, special):
...
+    Any single or multiple element data structure, or list-like object.
+axis : int or str, optional
+    Axis to target. Can be either the axis name ('index', 'rows',
+    'columns') or number (0, 1).
+level : int or name
```
datapythonista: Better to use `object` than `name`, as in this line we document the type.

pandas/core/ops.py
```diff
@@ -582,6 +588,150 @@ def _get_op_name(op, special):
...
+    Broadcast across a level, matching Index values on the passed
+    MultiIndex level.
+
+Returns
+-------
+result : DataFrame of bool
```
datapythonista: We don't need `result` here, you can just leave the type `DataFrame of bool`

pandas/core/ops.py
```diff
@@ -582,6 +588,150 @@ def _get_op_name(op, special):
...
+    Result of the comparison.
+
+See also
```
datapythonista: I think the `A` should be upper case

pandas/core/ops.py
```diff
@@ -582,6 +588,150 @@ def _get_op_name(op, special):
...
+>>> df3
+      cost  revenue
+Q1 A   250      100
+   B   150      250
+   C   100      300
+Q2 A   150      200
+   B   300      175
+   C   220      225
+
+Compare to a constant and operator version
```
datapythonista: Could you explain that both are equivalent in this case? And may be finish the sentence with a period?

pandas/core/ops.py
```diff
@@ -582,6 +588,150 @@ def _get_op_name(op, special):
...
+1       B   150      250
+2       C   100      300
+>>> df2 = pd.DataFrame({'company': ['A', 'B', 'C', 'D'],
+...                     'revenue': [300, 250, 100, 150]})
+>>> df2
+  company  revenue
+0       A      300
+1       B      250
+2       C      100
+3       D      150
```
datapythonista: Instead if defining all the DataFrames first, I'd define the first, show the examples that don't need the others, and define the others immediately before they are used.

For the naming, I'd name `df` the first, `other` the second used to be compared. And something like `df_multiindex` the last one. I think that will make things easier for the users.

I also think that would make things easier for the user having smaller DataFrames (like 2x2). These big `True`/`False` DataFrame results are not immediate to understand.

pandas/core/ops.py
```diff
@@ -582,6 +588,150 @@ def _get_op_name(op, special):
...
+2    False   True    False
+
+Compare to a Series by axis and operator version
+
+>>> df1 != [100, 250, 300]
+   company  cost  revenue
+0     True  True    False
+1     True  True    False
+2     True  True    False
+>>> df1.ne([100, 250, 300], axis=0)
```
datapythonista: I prefer `index` and `columns` for the `axis` values, it's more explicit and less confusing.

datapythonista (changes requested):
Looks much better now. Added some comments, but I think this new version adds much more value for users.

pandas/core/ops.py
ParfaitG: `df_examples` was kept from original docs and not removed for code consistency if used elsewhere. Can remove and test docstrings.

pandas/core/ops.py
ParfaitG: To be consistent with examples that run through several of these sibling methods, the intro lists all of them. It seems counterintuitive to have an intro that describes **one** method with examples from several methods.

pandas/core/ops.py
ParfaitG: Will do.

pandas/core/ops.py
ParfaitG: Will do.

pandas/core/ops.py
ParfaitG: Will do.

pandas/core/ops.py
ParfaitG: I see no current docs page that has `Also` as title case.

pandas/core/ops.py
ParfaitG: Will do.

pandas/core/ops.py
ParfaitG: Understood on names. As for lengths, df1 and df2 have different number of rows and columns to show results of across different lengths. And it is just one more row and one more column. Surely not that *big*. I can add a note to mention results of different lengths.

pandas/core/ops.py
ParfaitG: Will do.

pandas/core/ops.py
ParfaitG: `reverse` is needed for compatibility with arithmetic operators as they are included in same dictionary, `_op_descriptions`

pandas/core/ops.py
```diff
@@ -582,6 +577,155 @@ def _get_op_name(op, special):
...
+Equivalent to `==`, `=!`, `<=`, `<`, `>=`, `>` with support to choose axis
+(rows or columns) and level for comparison.
+
+Parameters
+----------
+other : scalar, sequence, Series, or DataFrame
+    Any single or multiple element data structure, or list-like object.
+axis : int or str, optional
+    Axis to target. Can be either the axis name ('index', 'rows',
+    'columns') or number (0, 1).
```
datapythonista: To be consistent, I'd document the axis as we usually do. See this docstring: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.drop.html

pandas/core/ops.py
```diff
@@ -582,6 +577,155 @@ def _get_op_name(op, special):
...
+...                    'cost': [250, 150, 100],
+...                    'revenue': [100, 250, 300]}})
+>>> df
+  company  cost  revenue
+0       A   250      100
+1       B   150      250
+2       C   100      300
+
+Compare to a scalar and operator version which return equivalent
+results.
```
datapythonista: I found `equivalent` a bit confusing. Would it make sense to say that they return the same results?

pandas/core/ops.py
```diff
@@ -582,6 +577,155 @@ def _get_op_name(op, special):
...
+  company  revenue
+0       A      300
+1       B      250
+2       C      100
+3       D      150
+>>> df.reindex(['company', 'revenue'], axis='columns') > other.iloc[:-1]
+   company  revenue
+0    False    False
+1    False    False
+2    False     True
```
datapythonista: It's not obvious to me what we're showing here.

pandas/core/ops.py
```diff
@@ -582,6 +577,155 @@ def _get_op_name(op, special):
...
+0    False  False    False
+1    False  False    False
+2    False  False     True
+3    False  False    False
+>>> df.gt(other, axis=1)
+   company   cost  revenue
+0    False  False    False
+1    False  False    False
+2    False  False     True
+3    False  False    False
```
datapythonista: I'd use `axis='index'` and `columns` instead of 0 and 1. Or may be even better add in the text explaining the example that the parameter `axis` is not used when comparing two `DataFrame` objects.

pandas/core/ops.py
```diff
@@ -582,6 +577,155 @@ def _get_op_name(op, special):
...
+Q2 A   150      200
+   B   300      175
+   C   220      225
+
+>>> df.set_index('company') <= df_multindex.loc['Q1']
+         cost  revenue
+company
+A        True     True
+B        True     True
+C        True     True
```
datapythonista: I think here we're comparing two `DataFrame` objects. This is more useful to show how `.loc` works with a multiindex than how operators behave. I'd get rid of it.

pandas/core/ops.py
```diff
@@ -582,6 +577,155 @@ def _get_op_name(op, special):
...
+>>> df.set_index('company').le(df_multindex, level=1)
+       cost  revenue
+Q1 A   True     True
+   B   True     True
+   C   True     True
+Q2 A  False     True
+   B   True    False
+   C   True    False
+"""
+
```
datapythonista: I think we need to explain what's going on here. And I'd probably set `company` as the index when creating the `DataFrame`, to avoid adding extra complexity here. I don't think it's a problem for other examples.

The blank line after the docstring is not required.

datapythonista (commented):
Looking good, I think there are few things in the examples that can be more clear, but it's looking great.

datapythonista (changes requested):
Looks great. Just few last comments, mainly of failing examples. Please run the validation script to make sure the examples work, after you make these last changes. Cheers!

pandas/core/ops.py
```diff
@@ -640,6 +635,139 @@ def _get_op_name(op, special):
...
+operators.
+
+Equivalent to `==`, `=!`, `<=`, `<`, `>=`, `>` with support to choose axis
+(rows or columns) and level for comparison.
+
+Parameters
+----------
+other : scalar, sequence, Series, or DataFrame
+    Any single or multiple element data structure, or list-like object.
+axis :  {{0 or 'index', 1 or 'columns'}}, default 'columns'
```
datapythonista: You also have an extra pair of brackets here. With this change and the pending ones from the previous review, lgtm

ParfaitG:
@datapythonista ... Did you get my note above regarding the need for double curly braces? The entire docstring is being run with `str.format()` and we need to escape the actuals from format's placeholders. Please advise.

datapythonista:
Oh, I see, didn't see that comment, sorry. As the number of variables is just two, and the number of curly brackets in the docstring is much larger, I think it'd be simpler to use `docstring % dict(var=...` than `docstring.format(...`. What do you think? Is there any advantage in `format`?

ParfaitG:
To stay consistent, most string formatting in `ops.py` uses `format`. And from my understanding, the modulo operator is an outmoded, less preferred way for string interpolation in Python. Nonetheless, I did run validation script and these methods passed. Next week I will report the validation result for a few for your reference.

datapythonista (approved):
Ok, fair enough.

lgtm

pandas/core/ops.py
```diff
@@ -1764,8 +1892,14 @@ def na_op(x, y):
...
-    @Appender('Wrapper for flexible comparison methods {name}'
-              .format(name=op_name))
+    if op_name in _op_descriptions:
+        op_desc = _op_descriptions[op_name]
+        doc = _flex_comp_doc_FRAME.format(desc=op_desc['desc'],
+                                          op_name=op_name)
+    else:
+        doc = "Flexible wrappers to comparison methods"
+
+    @Appender(doc)
```
datapythonista: I was taking a look before merging, and there is no case when `op_name` is not in `_op_descriptions`. Any reason why this if was added?

Can we simply have?

```
@Appender(_flex_comp_doc_FRAME.format(op_name=op_name,
                                      desc=_op_descriptions[op_name]['desc'])
```

Or add an assert `op_name in _op_descriptions` if you think its useful, but I don't see this `if` adding value.

datapythonista (changes requested):
Added a comment about the if to check whether the operator is in the operators dict. Can you address it before merging?

datapythonista (approved):
lgtm, thanks a lot for the hard work on this @ParfaitG 

@jbrockmendel you requested some changes that I think have been addressed. Can you take a look before we merge?

pandas/core/ops.py
```diff
@@ -640,6 +635,139 @@ def _get_op_name(op, special):
...
+Mismatched indices will be unioned together.
+`NaN` values are considered different (i.e. `NaN` != `NaN`).
+
+Examples
+--------
+>>> df = pd.DataFrame({{'cost': [250, 150, 100],
+...                    'revenue': [100, 250, 300]}},
+...                    index = ['A', 'B', 'C'])
+>>> df
+   cost  revenue
```
jreback: can we turn on doc-tests for this?

pandas/core/ops.py
```diff
@@ -640,6 +635,139 @@ def _get_op_name(op, special):
...
+A  True    False
+B  True    False
+C  True    False
+>>> df.ne([100, 250, 300], axis='index')
+  cost  revenue
+A  True    False
+B  True    False
+C  True    False
+>>> df != pd.Series([100, 250, 300])
```
jreback: put a blank line between cases

ParfaitG:
@jreback...Forgive me, I am not aware of doc-tests but I did run the validation string script on all methods (`eq`, `ne`,  etc.). Re blank line between examples, will do in several hours.

@datapythonista ... Thanks for your help. Curious, should we adjust arithmetic operator doc strings in same manner as comparison operators with a different PR?

datapythonista:
@ParfaitG, what @jreback is talking about is #22459. But we still don't have `pandas/core/ops.py` in the CI, so I'd leave adding this docstring to a separate PR where we can add some more files.

Regarding the arithmetic operator docstring, I didn't checked it, but we have more than 1,000 than need improvements, so I guess it'll be one of them, and it'd be really great if you can work on it.

pandas/core/ops.py
datapythonista: Also, when you separate these two blocks of examples (list vs Series), can you a short comment explaining that for Series `axis` is by default 'columns', but for list is `index`, as the example shows?

jbrockmendel:
> you requested some changes that I think have been addressed. Can you take a look before we merge?

@datapythonista I think all my concerns have been addressed.

@ParfaitG Thanks for pushing this through.  It took a little longer than expected, but you really nailed it.

ParfaitG:
@datapythonista, Please see update with arithmetic operators docstring revamped to resemble comparison operators. I removed the need of separate `df_examples` as previously set up.

datapythonista (changes requested):
Looks great. I added few comments about typos and formatting changes (some in the original docstring).

Besides those comments, it'd be nice to have an example of reverse operator (e.g. `df.div(10)` vs `df.rdiv(10)`).

And for the comparisons it wasn't that relevant, but I think for the arithmetic I'd strongly prefer to have an example with smaller amounts. It takes me some time to see that `120 - 250 = -130` but it'd be immediate to see that `4 - 1 = 3`. I'd prefer those simple examples so it's immediate to understand what's going on.

ParfaitG:
@datapythonista, I have updated the DataFrame arithmetic operator docstring. And while I was at it, I updated Series arithmetic docstring. I also come to find out there was no Series comparison docstring. It just reuses the arithmetic docstring. 

So, I added a brand new docstring for Series comparison operators which also involved adding a new `def` method in *Series* section at bottom. The `ops.py` does compile and all `validation_string` methods with `doc_test` do pass. Please advise on this extensive update. I hope your head doesn't spin for the confusion!

ParfaitG:
Yes, but both are being called separately above. And docstrings are assigned differently. But you make a good point we can combine and have callers pass args for conditional returns.

datapythonista:
@ParfaitG is it a big problem to address the new changes in a separate PR? The new changes are more substantial and will require some discussions and review, and having the previous changes which were ready to be merged, makes it more complex and time consuming.

I think you should be able to do a `git reset --soft <VERSION_WITH_PREVIOUS_CHANGES>`, and then create a new branch and commit the changes there.

ParfaitG:
@datapythonista, Absolutely makes sense. I will rollback in this PR and leave Series methods docstrings unchanged and propose an adjustment after merge.

ParfaitG:
@datapythonista, please note I have rolled back and updated the DataFrame arithmetic operator docstring. Let me know your thoughts and any next steps.

pandas/core/ops.py
```diff
@@ -604,40 +496,298 @@ def _get_op_name(op, special):
...
+DataFrame.floordiv : Divide DataFrames (integer division).
+DataFrame.mod : Calculate modulo (remainder after division) of
+    DataFrames.
+DataFrame.pow : Calculate exponential power of DataFrames.
 
 Examples
 --------
-{df_examples}
+>>> df = pd.DataFrame({{'A': [4, 6, 8],
+...                    'B': [3, 5, 9]}})
```
datapythonista: The indentation here doesn't seem right.

Also, I think it'd be better to have a simple real-world example, so it's easier to follow. For example, in the next example is quite obvious to understand what the operator is doing, even without checking the original DataFrame. With "random" data things are more complex for the user.
```
>>> df + 1
           num_legs  num_wings
cat          5        1
ostrich      3        3
tuna         1        1
```
Do you mind changing it?

pandas/core/ops.py
```diff
@@ -604,40 +496,298 @@ def _get_op_name(op, special):
...
+--------
+DataFrame.add : Add DataFrames.
+DataFrame.sub : Subtract DataFrames.
+DataFrame.mul : Multiply DataFrames.
+DataFrame.div : Divide DataFrames (float division).
+DataFrame.truediv : Divide DataFrames (float division).
+DataFrame.floordiv : Divide DataFrames (integer division).
+DataFrame.mod : Calculate modulo (remainder after division) of
+    DataFrames.
+DataFrame.pow : Calculate exponential power of DataFrames.
```
datapythonista: I don't think we need the "of DataFrames" in the last two descriptions.

datapythonista (commented):
Looking good, added couple of comments

datapythonista (commented):
Looks good.

Can you add the doctests for this in `ci/doctests.sh`?

Something like the existing:
```
    pytest --doctest-modules -v pandas/core/generic.py \
        -k"-_set_axis_name -_xs -describe -droplevel -groupby -interpolate -pct_change -pipe -reindex -reindex_axis -resample -sample -to_json -transpose -values -xs"

    if [ $? -ne "0" ]; then
        RET=1
    fi
```

But for `pandas/core/ops.py`. The `-k"-foo"` means that the doctest for the function `foo` will be skipped. With some luck there is nothing to skip in this file. Your docstrings should pass the tests, but if there is any other in the file that doesn't, feel free to add to the skip list, and we'll take care of them later.

Thanks for the hard work on this PR @ParfaitG

datapythonista:
Also, @ParfaitG, I was taking a look at the CI, and seems like we've got tests that validate the operations docstrings, that's why the checks are failing. Can you run the tests, and update the ones that are failing please?

datapythonista:
Sorry for the lack of context @ParfaitG. The validations you're doing are perfect, and I don't think your changes have any mistake.

What we would like is that every time someone sends a PR, validate whether all examples in the docstrings are correct. Correct in the way that the shown output is the actual result a user would get if they run the examples. For that, we've got the file `ci/doctests.sh`. While what we'd like is to validate every single docstring in pandas, so far we just added the files `core/series.py`, `core/frame.py` and `core/generic.py`. And even in those, we still have some broken examples, that at the moment we skip, and that we're slowly fixing.

What we think it would be good is to add `core/ops.py` to that validation, as you fix all (or most) of the docstrings in that file. So, not only your changes are correct, but any future change to thoses docstrings will necessarily have to pass the doctests.

So, what you can do to make this happen is to edit the file `ci/doctests.sh` and add another validation like the one I sent for `core/generic.py` but for `core/ops.py`. Remove to start with all the stuff in the `-k` parameter (including the parameter itself). Then run the file, and see if there is any broken docstring. If there is any method broken unrelated to the PR, you add its name to the `-k` option. If it's one of the ones you changes (it won't) you fix it. And with some luck, nothing will fail, and we'll start validating all the docstrings on that file since we merge your PR.

Unrelated to this, if you run the pandas "normal" tests, you'll see that some tests are broken. Seems like there are tests validating the content of the docstrings you updated. I assume we wanted to check that when we "inject" the docstrings, every method (`add`, `mul`...) gets its docstring, and not anonther. You'll need to update these tests. You can also see in the checks in this conversation, how they are in red. After fixing them they'll be in green, and we'll be ready to merge your changes.

Sorry this is getting a bit too complicated. Let me know if you need help.

ParfaitG:
@datapythonista ... Please advise on why tests failed. I updated `ci/doctests` per your instruction, finding only one method is defined and failed for `ops.py`: `_gen_eval_kwargs`.  I did notice a few others failed for `frame.py`, `series.py`, and `generic.py` and added those in the skip `-k` argument. Let me know if my implementation was off. Also, per this [PR](https://github.com/pandas-dev/pandas/pull/22701) I had to  remove the `addopts = --strict-data-files` from my `setup.cfg` to get this to work or it kept raising _unrecognized argument_.

Also, this time, `validate_string` script failed on the arithmetic methods with something like *summary should be one line*, so I broke the first line up to accommodate, moving *With reverse version, `{reverse}`.* to next line. I think this relates to one of the test fails here. Btw - we do not cover all docstrings for `ops.py`. Recall my earlier attempt that I rolled back to update the `Series` versions of these same comparison and arithmetic operators. We only touch the `DataFrame` versions here.

WillAyd:
The failure occurs because you are no longer substituting `equiv` so you aren't getting something like `dataframe + other` to show up in the add doctoring, which the test is explicitly checking for.

ci/df_info.txt
```diff
@@ -0,0 +1,8 @@
+<class 'pandas.core.frame.DataFrame'>
+RangeIndex: 5 entries, 0 to 4
+Data columns (total 3 columns):
+int_col      5 non-null int64
+text_col     5 non-null object
+float_col    5 non-null float64
+dtypes: float64(1), int64(1), object(1)
+memory usage: 200.0+ bytes
```
datapythonista: Is this file needed to pass the `to_excel` doctest? I think we want to think better where to place those files, or what to do in these cases. But in any case, let's leave that for a separate PR. We try to keep the PRs focused on one fix, and not mix different things. This is useful for example if we have to revert some changes because something is wrong. In that case, we don't want to revert two independent changes, because one is likely to be ok.

Other than this, the PR looks good to me. Thanks!

datapythonista (approved):
lgtm, happy to merge on green. Thanks @ParfaitG, great changes.

datapythonista:
@ParfaitG I see that the doctest that fails in `ops.py` is just because of quotes:
```
Expected:
    {"reversed": True, "truediv": True}
Got:
    {'reversed': True, 'truediv': True}
```
Being such a small change, if you want to change it in this PR, instead of skipping the `_gen_eval_kwargs` test, that would also be good.

datapythonista:
@ParfaitG can you please take care of the conflicts. Merging after that, and when it's all green. Sorry about the delay.

datapythonista:
thanks for the update @ParfaitG 

@jbrockmendel I think you reviewed this before. Do you mind taking another look and merge it if you're happy?

pandas/core/ops.py
```diff
@@ -640,6 +635,145 @@ def _get_op_name(op, special):
...
+    inequality elementwise
+
+Notes
+--------
+Mismatched indices will be unioned together.
+
+Examples
+--------
+>>> df = pd.DataFrame({{'cost': [250, 150, 100],
+...                    'revenue': [100, 250, 300]}},
```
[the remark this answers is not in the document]
jbrockmendel: Has this been resolved?

pandas/core/ops.py
jbrockmendel: Has this been resolved?

pandas/core/ops.py
```diff
@@ -640,6 +635,145 @@ def _get_op_name(op, special):
...
+   cost  revenue     0     1     2
+A  True     True  True  True  True
+B  True     True  True  True  True
+C  True     True  True  True  True
+
+Compare to a DataFrame of different shape by axis where both 'index' and
+'columns' return same results.
+
+>>> other = pd.DataFrame({{'revenue': [300, 250, 100, 150]}},
+...                       index = ['A', 'B', 'C', 'D'])
```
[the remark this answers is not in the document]
jbrockmendel: Usually we would use `index=` instead of `index = `.  Is that different here?

pandas/core/ops.py
```diff
@@ -637,38 +529,295 @@ def _get_op_name(op, special):
...
+A  False    False
+B  False    False
+C  False     True
+D  False    False
+
+Compare to a MultiIndex by level.
+
+>>> df_multindex = pd.DataFrame({{'cost': [250, 150, 100, 150, 300, 220],
+...                              'revenue': [100, 250, 300, 200, 175, 225]}},
+...                              index = [['Q1', 'Q1', 'Q1', 'Q2', 'Q2', 'Q2'],
```
jbrockmendel: `index=`?

pandas/core/ops.py
jbrockmendel: Has this been resolved?

jbrockmendel:
@datapythonista a handful of questions checking that your comments have been resolved.  If those are all OK, then LGTM.

ParfaitG:
@datapythonista Your suggested `flake 8` call raised no error. What is the issue with `index`? Do we need to remove space before and after equals sign, `=`?

datapythonista:
In PEP-8, this is the valid syntax:
```
some_variable = some_value
some_callable(some_param=the_provided_value)
```

So yes, the spaces around the `=` are wrong, as it's not a variable assignment, but setting a parameter value.

The call to `flake8` should work if made in a pandas environment with all the dependencies, and run in the root project directory. If after some research on the error you don't see the problem, share the error here to see if we can help.

ParfaitG:
Hmmm...the call `flake8 --doctests pandas/core/ops.py` returns nothing: no error or line number. To be sure I removed then recreated and activated a new anaconda pandas dev environment and installed dependendies using the [Python Sprint Pandas setup](https://python-sprints.github.io/pandas/guide/pandas_setup.html) document. Then I retried `flake8` at the pandas project root with same result as before of no result or error. This suggests no issue with the `ops.py` document.

Aside, with every docstirng change I run `git diff upstream/master -u -- "*.py" | flake8 --diff` which _will_ return line numbers of issues to be resolved and if no issue nothing is raised (like the above and I reran this line too with nothing raised). I run this line according to the [Python Sprint Submitting Changes](https://python-sprints.github.io/pandas/guide/pandas_pr.html) document. Do note: this document does not include a `flake8 --doctests` call. I even ran `git diff upstream/master -u -- "*.py" | flake8 --doctests pandas/core/ops.py` with same result of no error or line number result.

Please advise how to find these PEP-8 syntax issues. Otherwise, I can manually check and hope they pass on your end.

datapythonista:
Sorry, you're right, seems like `flake8 --doctests` is not working. Not sure why, and I would say that I've seen it working, but I'm not even sure. May be it never reported any error and I thought everything was all right.

I created #23154, that is anyway the way I think we should validate it. For now, if you can fix the reported errors, and check manually that nothing else is pep-8 invalid, that would be great.

ParfaitG:
Understood. I am noticing some items outside of comparison and arithmetic docstrings that do not adhere to PEP 8 such as the recurring assignment to lambda expressions. According to [PEP 8 - Style Guide](https://www.python.org/dev/peps/pep-0008/):

> Always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier

Unfortunately, it is not easy to switch as some `lambda` calls are inside `if` statements where `def` calls are not recommended. I began refactoring code but this may be substantial changes that require discussion and outside of scope for this PR. So I did not commit these changes. One example includes:

```python
    if special:
        dunderize = lambda x: '__{name}__'.format(name=x.strip('_'))
    else:
        dunderize = lambda x: x
    new_methods = {dunderize(k): v for k, v in new_methods.items()}
```

To be replaced as:

```python
    def dunderize(x):
        if special:
            return '__{name}__'.format(name=x.strip('_'))
        else:
            return x

    new_methods = {dunderize(k): v for k, v in new_methods.items()}
```

datapythonista:
if you check `setup.cfg` you'll see that the `lambda` thing is one of the few pep8 errors that we ignore.

pandas/core/ops.py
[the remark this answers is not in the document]
datapythonista: the extra pair of brackets is needed, as single curly brackets are interpreted as a variable like in `'{foo}'.format(foo=123)`

@ParfaitG did you check that the indentation is correct in the resulting docstring?

pandas/core/ops.py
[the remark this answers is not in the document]
datapythonista: @ParfaitG can you fix the `index = [...]` spaces (in all the cases). I think it's the only missing thing to merge this.

pandas/core/ops.py
datapythonista: Can you also add this explanation? So users can understand in an easier way what we are showing here.

WillAyd:
@ParfaitG can you merge master?

datapythonista:
@ParfaitG can you merge master and make sure the CI is green, so we can merge.

@jbrockmendel can you take another look and see if you're happy to merge this?

pandas/core/ops.py
```diff
@@ -635,38 +527,295 @@ def _get_op_name(op, special):
...
+DataFrame.truediv : Divide DataFrames (float division).
+DataFrame.floordiv : Divide DataFrames (integer division).
+DataFrame.mod : Calculate modulo (remainder after division).
+DataFrame.pow : Calculate exponential power.
 
 Examples
 --------
-{df_examples}
+>>> df = pd.DataFrame({{'angles': [0, 3, 4],
+...                    'degrees': [360, 180, 360]}},
```
jbrockmendel: one more space here?

jbrockmendel: nope, my mistake

jbrockmendel:
LGTM

jbrockmendel (approved):

Issue #8212, opened 2014-09-08
ENH: read_excel dtypes and converts


iosonofabio:
Dear all,

Here from the mailing list: https://groups.google.com/forum/#!topic/pydata/jKiPOvYUQ1c

I have an excel table about family ages like this

| Family | People | Mean size [cm] |
| --- | --- | --- |
| Foo | 5 | 173.0 |
| Bar | 3 | 189.0 |

and I would like to use read_excel to parse it into Python. I would like "People" to be read as an integer, "Mean size [cm]" as a float. (And "Family" as a string, but that might be a different issue.) Now:
- if I set convert_float=True, the last column reads as int
- if I set convert_float=False, the second column reads as float

Neither one is correct, for a stupid reason: there happen to be those .0 in all sizes! So I would like to specify something like:

``` python
convert_float = ['People']
```

so only that column gets converted. An even better solution would be to be explicit about types of some columns, letting pandas perform the automagic for the others, such as:

``` python
read_excel('foo.xlsx', types={'People': np.uint8, 'Family': 'S3'})
```

but this changes the signature of the function more significantly.

Are you folks in favour of any of this? If yes, I can get a look and try to code it in.

bolshoibooze:
Thanks for the clear explanation @iosonofabio :-)

PR #20429, opened 2018-03-20
read_excel with dtype=str converts empty cells to np.nan

### LINKED ISSUE #20377: read_excel with dtype=str converts empty cells to the string 'nan'

```python
In [10]: df = pd.DataFrame({'a': ['x', 'y', '', 'z']})

In [11]: df
Out[11]: 
   a
0  x
1  y
2   
3  z

In [12]: df.to_excel('temp.xlsx')

In [13]: df = pd.read_excel('temp.xlsx', dtype=str)

In [14]: df
Out[14]: 
     a
0    x
1    y
2  nan
3    z

In [15]: df.loc[2, 'a']
Out[15]: 'nan'

In [16]: type(df.loc[2, 'a'])
Out[16]: str

```

#### Problem description
The empty string of the original dataframe becomes the string 'nan', instead of numpy.nan.

#### Expected Output

```python
In [14]: df
Out[14]: 
     a
0    x
1    y
2  NaN
3    z

In [15]: df.loc[2, 'a']
Out[15]: nan

In [16]: type(df.loc[2, 'a'])
Out[16]: float
```

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.5.2.final.0
python-bits: 64
OS: Linux
OS-release: 4.13.0-36-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: ca_ES.UTF-8
LOCALE: ca_ES.UTF-8

pandas: 0.22.0
pytest: None
pip: 9.0.1
setuptools: 38.5.2
Cython: 0.27.3
numpy: 1.14.1
scipy: 1.0.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: None
patsy: 0.5.0
dateutil: 2.7.0
pytz: 2018.3
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 2.2.0
openpyxl: 2.5.1
xlrd: 1.1.0
xlwt: None
xlsxwriter: 0.7.3
lxml: None
bs4: None
html5lib: 1.0.1
sqlalchemy: 1.2.5
pymysql: 0.8.0
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

nikoskaragiannakis:

- [x] closes #20377
- [x] tests added / passed
- [x] whatsnew entry


nikoskaragiannakis:
No idea what's going on with this conflict. Any help?

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -981,6 +981,8 @@ I/O
 - :class:`Timedelta` now supported in :func:`DataFrame.to_excel` for all Excel file types (:issue:`19242`, :issue:`9155`, :issue:`19900`)
 - Bug in :meth:`pandas.io.stata.StataReader.value_labels` raising an ``AttributeError`` when called on very old files. Now returns an empty dict (:issue:`19417`)
 - Bug in :func:`read_pickle` when unpickling objects with :class:`TimedeltaIndex` or :class:`Float64Index` created with pandas prior to version 0.20 (:issue:`19939`)
+- Bug in :meth:`pandas.io.json.json_normalize` where subrecords are not properly normalized if any subrecords values are NoneType (:issue:`20030`)
+- Bug in :`read_excel` where it transforms np.nan to 'nan' if dtype=str is chosen. Now keeps np.nan as they are. (:issue:`20377`)
```
cbertinato: Should be :func:\`read_excel\`

cbertinato:
Try rebasing and then resolving the conflict?

jreback (changes requested):
@gfyoung do we handle this correctly in read_csv itself?

gfyoung:
> @gfyoung do we handle this correctly in read_csv itself?

@jreback : unfortunately, no.  Too bad I didn't see this PR earlier.  I would have actually suggested to fix it in `read_csv` and then check if the bug still persists in `read_excel` (since they share parsing engines).

pandas/io/excel.py
```diff
@@ -679,6 +679,11 @@ def _parse_cell(cell_contents, cell_typ):
                                     **kwds)
 
                 output[asheetname] = parser.read(nrows=nrows)
+                dtypes = output[asheetname].dtypes
+                output[asheetname].replace('nan', np.nan, inplace=True)
+                output[asheetname] = output[asheetname].astype(dtypes,
+                                                               copy=False)
```
gfyoung: I worry about this patch being a performance hit against `read_excel`.  The Python parser processes each of the Excel elements before placing it into a `DataFrame`.  I would look there for the fix.

cbertinato: The result from `read_csv` is not the same as the one from `read_excel` with `dtype=str`.  In the former case, empties are read in as `np.nan`, whereas in the latter they are read in as the string 'nan'.

gfyoung: True, though that doesn't change my opinion.   The problematic part still likely stems in the engine parsing, which would also effect `read_csv` (the more popular of the two IMO).  Thus, if we can kill two birds with one stone, that would be even better.

arnau126:
@jreback, @gfyoung , please read #20377, where we discussed about this issue.

`read_csv` works correctly (empty cells turn to `np.nan`)

In `read_excel` empty cells turn to the string `'nan'`, because it uses `PythonParser`. This parser has a function called `self._cast_types`, which uses `astype_nansafe`, which uses `astype_unicode` and `astype_str` (https://github.com/pandas-dev/pandas/blob/master/pandas/_libs/lib.pyx#L460).

The later functions are only used by `read_excel` (not `read_csv`), so I think that we should fix them by using `checknull` as `astype_intsafe` does (https://github.com/pandas-dev/pandas/blob/master/pandas/_libs/lib.pyx#L437).

jreback:
this is my point - the fix needs to be in read_csv not here

arnau126:
I think it's not a matter of `read_csv` or `read_excel`.
The fix needs to be in `PythonParser` which can be used in both `read_csv` and `read_excel`.

gfyoung:
I would go with empty as per the issue.

nikoskaragiannakis:
In the ticket I mentioned this https://github.com/pandas-dev/pandas/issues/20377#issuecomment-374693647.
@cbertinato gave an answer there but, since we're discussing it, I'd like some more opinions.

cbertinato:
> I would go with empty string as per the issue.

Should the C parser also be brought in line?

@nikoskaragiannakis: @gfyoung suggests an empty string. That sounds like a good idea as it is consistent with the originating DataFrame.

gfyoung:
> Should the C parser also be brought in line?

Absolutely!  Both parsers should be patched.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -981,6 +981,8 @@ I/O
 - :class:`Timedelta` now supported in :func:`DataFrame.to_excel` for all Excel file types (:issue:`19242`, :issue:`9155`, :issue:`19900`)
 - Bug in :meth:`pandas.io.stata.StataReader.value_labels` raising an ``AttributeError`` when called on very old files. Now returns an empty dict (:issue:`19417`)
 - Bug in :func:`read_pickle` when unpickling objects with :class:`TimedeltaIndex` or :class:`Float64Index` created with pandas prior to version 0.20 (:issue:`19939`)
+- Bug in :meth:`pandas.io.json.json_normalize` where subrecords are not properly normalized if any subrecords values are NoneType (:issue:`20030`)
```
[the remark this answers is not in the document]
nikoskaragiannakis: it's not mine. I deleted it by mistake and added it back.
You can check master here https://github.com/pandas-dev/pandas/blob/master/doc/source/whatsnew/v0.23.0.txt#L985
However, even after rebasing, I keep getting this conflict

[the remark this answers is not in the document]
cbertinato: If you rebased off master and resolved the conflicts in the rebase then it should be ok. Did you fetch the current master before rebasing?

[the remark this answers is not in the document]
nikoskaragiannakis: yep

nikoskaragiannakis:
Any comments for the new changes?

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -984,6 +984,8 @@ I/O
 - Bug in :meth:`pandas.io.stata.StataReader.value_labels` raising an ``AttributeError`` when called on very old files. Now returns an empty dict (:issue:`19417`)
 - Bug in :func:`read_pickle` when unpickling objects with :class:`TimedeltaIndex` or :class:`Float64Index` created with pandas prior to version 0.20 (:issue:`19939`)
 - Bug in :meth:`pandas.io.json.json_normalize` where subrecords are not properly normalized if any subrecords values are NoneType (:issue:`20030`)
+- Bug in :func:`read_excel` and :class:`TextReader` now turn np.nan to empty string when dtype=str. They used to turn np.nan to 'nan' (:issue `20377`)
```
TomAugspurger: Perhaps phrase this as "Bug in read_excel, old behavior, new behavior". Right now it reads a bit like turning NaN to the empty string was the bug.

TextReader isn't part of the public API so this link won't work (and it'll cause a warning in the doc build.)

Double backticks around `np.nan`.

Double backticsk around `dtype=str`.

Double backticks around `'nan'` (keep the quotes though).

TomAugspurger (commented):
It'd be good to run the benchmarks: http://pandas.pydata.org/pandas-docs/stable/contributing.html#running-the-performance-test-suite for the CSV parser.

doc/source/whatsnew/v0.23.0.txt
TomAugspurger: Should also mention that this affected parsing CSVs with `dtype=str`, right?:

arnau126:
I'm good with this PR because after using `read_csv` I always do `.fillna('')` for all `str` columns.
However, other operations like `.fillna('-')` or `.dropna()` won't work anymore because the empty string `''` is not `np.nan`.

pandas/tests/io/test_excel.py
```diff
@@ -361,6 +361,35 @@ def test_reader_dtype(self, ext):
...
+                                         'c': str,
+                                         'd': str})
+
+        expected['a'] = expected['a'].astype('float64')
+        expected['b'] = expected['b'].astype('float32')
+        expected['c'] = ['001', '002', '003', '004']
+        expected['d'] = ['1', '2', '', '4']
+        tm.assert_frame_equal(actual, expected)
+
+        with pytest.raises(ValueError):
```
gfyoung: What's the error message?

pandas/tests/series/test_io.py
```diff
@@ -162,6 +162,16 @@ def test_to_csv_compression(self, compression):
                                                    index_col=0,
                                                    squeeze=True))
 
+    def test_from_csv_dtype_str(self):
```
gfyoung: * This test should really belong in the `tests/io/parser/na_values` tests
* Do we really need to round-trip to surface this bug?  Can we try to just use `StringIO`?

doc/source/whatsnew/v0.23.0.txt
gfyoung: > Should also mention that this affected parsing CSVs with dtype=str, right?

@TomAugspurger : yes

jreback:
so this needs a test for csv reading. The intent here is NOT to use '' as the missing value, simply to remove the case where an existing ``np.nan`` is converted to the string ``nan``. You would need a test to prove this as its not obvious from the implementation (which actually looks quite strange).

pandas/tests/io/test_excel.py
nikoskaragiannakis:
```
ValueError: Unable to convert column d to type int64
```
which is the same as in the `test_reader_dtype`, which I copied it from. No reason to have this here I guess.

gfyoung: > No reason to have this here I guess.

What do you mean?

nikoskaragiannakis:
> However, other operations like .fillna('-') or .dropna() won't work anymore because an empty string '' is not np.nan.

I think what @arnau126 says here makes total sense.

> The intent here is NOT to use '' as the missing value, simply to remove the case where an existing np.nan is converted to the string nan.

@jreback what you're saying here agrees with the comment above, but that contradicts something @cbertinato said some days ago:
> I would go with empty string as per the issue.

Sorry if I misunderstood something, but I want to be sure there is a clear decision here.

cbertinato:
Sorry to have added confusion. I misinterpreted an earlier comment. I believe @jreback’s suggested changes are the best way to go.

nikoskaragiannakis:
So, what needs to be done here is to make sure that an existing `np.nan` is **not** converted to a string `'nan'`, right? Does this mean that it should remain as a `np.nan`?

nikoskaragiannakis:
I made the changes so now both `read_csv` and `read_excel`
* turn empty values to `np.nan` when `dtype=str` and `na_filter=True`
* turn empty values to empty string when `dtype=str` and `na_filter=False`

I also ran the performance tests for csv, excel, and series:
```
io.csv
======
[  0.00%] · For pandas commit hash 7d5f6b20:
[  0.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[  0.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[  3.33%] ··· Running io.csv.ReadCSVCategorical.time_convert_direct                                                                                                                                      44.4±0.2ms
[  6.67%] ··· Running io.csv.ReadCSVCategorical.time_convert_post                                                                                                                                        70.3±0.1ms
[ 10.00%] ··· Running io.csv.ReadCSVComment.time_comment                                                                                                                                                2.01±0.01ms
[ 13.33%] ··· Running io.csv.ReadCSVDInferDatetimeFormat.time_read_csv                                                                                                                              1.15±0.01ms;...
[ 16.67%] ··· Running io.csv.ReadCSVFloatPrecision.time_read_csv                                                                                                                                    1.64±0.01ms;...
[ 20.00%] ··· Running io.csv.ReadCSVFloatPrecision.time_read_csv_python_engine                                                                                                                      1.81±0.01ms;...
[ 23.33%] ··· Running io.csv.ReadCSVParseDates.time_baseline                                                                                                                                               3.15±0ms
[ 26.67%] ··· Running io.csv.ReadCSVParseDates.time_multiple_date                                                                                                                                          3.10±0ms
[ 30.00%] ··· Running io.csv.ReadCSVSkipRows.time_skipprows                                                                                                                                         21.5±0.04ms;...
[ 33.33%] ··· Running io.csv.ReadCSVThousands.time_thousands                                                                                                                                        18.0±0.02ms;...
[ 36.67%] ··· Running io.csv.ReadUint64Integers.time_read_uint64                                                                                                                                        1.13±0.01ms
[ 40.00%] ··· Running io.csv.ReadUint64Integers.time_read_uint64_na_values                                                                                                                              1.22±0.01ms
[ 43.33%] ··· Running io.csv.ReadUint64Integers.time_read_uint64_neg_values                                                                                                                                1.15±0ms
[ 46.67%] ··· Running io.csv.ToCSV.time_frame                                                                                                                                                         156±0.7ms;...
[ 50.00%] ··· Running io.csv.ToCSVDatetime.time_frame_date_formatting                                                                                                                                   11.6±0.05ms
[ 50.00%] · For pandas commit hash bd8a3cff:
[ 50.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt..............................................................................
[ 50.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[ 53.33%] ··· Running io.csv.ReadCSVCategorical.time_convert_direct                                                                                                                                      43.0±0.2ms
[ 56.67%] ··· Running io.csv.ReadCSVCategorical.time_convert_post                                                                                                                                        66.9±0.2ms
[ 60.00%] ··· Running io.csv.ReadCSVComment.time_comment                                                                                                                                                2.04±0.02ms
[ 63.33%] ··· Running io.csv.ReadCSVDInferDatetimeFormat.time_read_csv                                                                                                                              1.17±0.01ms;...
[ 66.67%] ··· Running io.csv.ReadCSVFloatPrecision.time_read_csv                                                                                                                                    1.69±0.01ms;...
[ 70.00%] ··· Running io.csv.ReadCSVFloatPrecision.time_read_csv_python_engine                                                                                                                      1.84±0.01ms;...
[ 73.33%] ··· Running io.csv.ReadCSVParseDates.time_baseline                                                                                                                                            3.07±0.02ms
[ 76.67%] ··· Running io.csv.ReadCSVParseDates.time_multiple_date                                                                                                                                       3.10±0.02ms
[ 80.00%] ··· Running io.csv.ReadCSVSkipRows.time_skipprows                                                                                                                                         21.4±0.06ms;...
[ 83.33%] ··· Running io.csv.ReadCSVThousands.time_thousands                                                                                                                                        17.9±0.03ms;...
[ 86.67%] ··· Running io.csv.ReadUint64Integers.time_read_uint64                                                                                                                                        1.13±0.07ms
[ 90.00%] ··· Running io.csv.ReadUint64Integers.time_read_uint64_na_values                                                                                                                              1.25±0.02ms
[ 93.33%] ··· Running io.csv.ReadUint64Integers.time_read_uint64_neg_values                                                                                                                             1.17±0.01ms
[ 96.67%] ··· Running io.csv.ToCSV.time_frame                                                                                                                                                         156±0.3ms;...
[100.00%] ··· Running io.csv.ToCSVDatetime.time_frame_date_formatting                                                                                                                                   11.4±0.03ms
BENCHMARKS NOT SIGNIFICANTLY CHANGED.



io.excel
========
[  0.00%] · For pandas commit hash 7d5f6b20:
[  0.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[  0.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[ 25.00%] ··· Running io.excel.Excel.time_read_excel                                                                                                                                                  154±0.2ms;...
[ 50.00%] ··· Running io.excel.Excel.time_write_excel                                                                                                                                                     641ms;...
[ 50.00%] · For pandas commit hash bd8a3cff:
[ 50.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[ 50.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[ 75.00%] ··· Running io.excel.Excel.time_read_excel                                                                                                                                                  154±0.1ms;...
[100.00%] ··· Running io.excel.Excel.time_write_excel                                                                                                                                                     641ms;...
BENCHMARKS NOT SIGNIFICANTLY CHANGED.



series_methods
==============
[  0.00%] · For pandas commit hash 7d5f6b20:
[  0.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[  0.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[  5.56%] ··· Running series_methods.Clip.time_clip                                                                                                                                                       125±0.5μs
[ 11.11%] ··· Running series_methods.Dir.time_dir_strings                                                                                                                                               1.74±0.01ms
[ 16.67%] ··· Running series_methods.Dropna.time_dropna                                                                                                                                                882±20μs;...
[ 22.22%] ··· Running series_methods.IsIn.time_isin                                                                                                                                                 1.63±0.02ms;...
[ 27.78%] ··· Running series_methods.Map.time_map                                                                                                                                                       948±6μs;...
[ 33.33%] ··· Running series_methods.NSort.time_nlargest                                                                                                                                             2.78±0.1ms;...
[ 38.89%] ··· Running series_methods.NSort.time_nsmallest                                                                                                                                            1.96±0.3ms;...
[ 44.44%] ··· Running series_methods.SeriesConstructor.time_constructor                                                                                                                                 314±1ms;...
[ 50.00%] ··· Running series_methods.ValueCounts.time_value_counts                                                                                                                                  2.14±0.03ms;...
[ 50.00%] · For pandas commit hash bd8a3cff:
[ 50.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[ 50.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[ 55.56%] ··· Running series_methods.Clip.time_clip                                                                                                                                                      121±0.05μs
[ 61.11%] ··· Running series_methods.Dir.time_dir_strings                                                                                                                                               1.75±0.01ms
[ 66.67%] ··· Running series_methods.Dropna.time_dropna                                                                                                                                                874±10μs;...
[ 72.22%] ··· Running series_methods.IsIn.time_isin                                                                                                                                                 1.63±0.01ms;...
[ 77.78%] ··· Running series_methods.Map.time_map                                                                                                                                                       950±8μs;...
[ 83.33%] ··· Running series_methods.NSort.time_nlargest                                                                                                                                             2.80±0.1ms;...
[ 88.89%] ··· Running series_methods.NSort.time_nsmallest                                                                                                                                            1.96±0.3ms;...
[ 94.44%] ··· Running series_methods.SeriesConstructor.time_constructor                                                                                                                                 320±4ms;...
[100.00%] ··· Running series_methods.ValueCounts.time_value_counts                                                                                                                                  2.08±0.08ms;...
BENCHMARKS NOT SIGNIFICANTLY CHANGED.
```
Looks like performance is not significantly affected.

I look forward for your feedback.

TomAugspurger:
Fixed a merge conflict.

Any concerns with this @gfyoung?

jreback:
don’t merge tom - need to look

pandas/_libs/lib.pyx
```diff
@@ -465,7 +465,11 @@ cpdef ndarray[object] astype_unicode(ndarray arr):
     for i in range(n):
         # we can use the unsafe version because we know `result` is mutable
         # since it was created from `np.empty`
-        util.set_value_at_unsafe(result, i, unicode(arr[i]))
+        arr_i = arr[i]
+        util.set_value_at_unsafe(
+            result,
+            i,
+            unicode(arr_i) if arr_i is not np.nan else np.nan)
```
gfyoung: Interesting spacing...maybe we should do this instead:

~~~
uni_arr_i = unicode(arr_i) if arr_i is not np.nan else np.nan
util.set_value_at_unsafe(result, i, uni_arr_i)
~~~

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1098,6 +1098,7 @@ I/O
 - Bug in :func:`read_pickle` when unpickling objects with :class:`TimedeltaIndex` or :class:`Float64Index` created with pandas prior to version 0.20 (:issue:`19939`)
 - Bug in :meth:`pandas.io.json.json_normalize` where subrecords are not properly normalized if any subrecords values are NoneType (:issue:`20030`)
 - Bug in ``usecols`` parameter in :func:`pandas.io.read_csv` and :func:`pandas.io.read_table` where error is not raised correctly when passing a string. (:issue:`20529`)
+- Bug in :func:`read_excel` and :func:`read_csv` where missing values turned to ``'nan'`` with ``dtype=str`` and ``na_filter=True``. Now, they turn to ``np.nan``. (:issue `20377`)
```
jreback: can you make the last part a bit more clear. These missing values are converted to the string missing indicator, ``np.nan``

pandas/_libs/lib.pyx
```diff
@@ -465,7 +465,11 @@ cpdef ndarray[object] astype_unicode(ndarray arr):
     for i in range(n):
         # we can use the unsafe version because we know `result` is mutable
         # since it was created from `np.empty`
-        util.set_value_at_unsafe(result, i, unicode(arr[i]))
+        arr_i = arr[i]
```
jreback: is arr_i in the cdef?

pandas/_libs/lib.pyx
jreback: use ``np.isnan`` here

pandas/_libs/lib.pyx
```diff
@@ -478,7 +482,11 @@ cpdef ndarray[object] astype_str(ndarray arr):
     for i in range(n):
         # we can use the unsafe version because we know `result` is mutable
         # since it was created from `np.empty`
-        util.set_value_at_unsafe(result, i, str(arr[i]))
+        arr_i = arr[i]
+        util.set_value_at_unsafe(
+            result,
+            i,
+            str(arr_i) if arr_i is not np.nan else np.nan)
```
jreback: same

pandas/tests/series/test_dtypes.py
```diff
@@ -149,6 +149,7 @@ def test_astype_str_map(self, dtype, series):
         # see gh-4405
         result = series.astype(dtype)
         expected = series.map(compat.text_type)
+        expected.replace('nan', np.nan, inplace=True)  # see gh-20377
```
jreback: don't use inplace

jreback (changes requested):

pandas/_libs/lib.pyx
nikoskaragiannakis: d'oh!

pandas/_libs/lib.pyx
nikoskaragiannakis: Using `np.nan` here raises:
```
TypeError: ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

nikoskaragiannakis: When i use that, all hell breaks loose. I get errors in tests like this one https://github.com/pandas-dev/pandas/blob/master/pandas/tests/frame/test_dtypes.py#L533

Is it because they use `np.NaN`? It looks like `checknull` checks both `np.NaN` and `np.nan`, while before the change I used to check only `np.nan`.
If that's the case, then I have to modify more tests.

nikoskaragiannakis: @gfyoung are you sure this indentation is a big problem? Because if I do what you suggest, then how should I declare uni_arr_i (and str_arr_i) in the cdef? 
Would it be ok if I changed it to sth like
```python
util.set_value_at_unsafe(
    ...
)
```
(moved the close bracket in the next line)?

gfyoung: That would work as well.

jreback: the nans are the same; iow they point to the same object
go ahead and change tests if need be and i will have a look

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1099,6 +1099,7 @@ I/O
 - Bug in :meth:`pandas.io.json.json_normalize` where subrecords are not properly normalized if any subrecords values are NoneType (:issue:`20030`)
 - Bug in ``usecols`` parameter in :func:`pandas.io.read_csv` and :func:`pandas.io.read_table` where error is not raised correctly when passing a string. (:issue:`20529`)
 - Bug in :func:`HDFStore.keys` when reading a file with a softlink causes exception (:issue:`20523`)
+- Bug in :func:`read_excel` and :func:`read_csv` where missing values turned to ``'nan'`` with ``dtype=str`` and ``na_filter=True``. Now, these missing values are converted to the string missing indicator, ``np.nan``. (:issue `20377`)
```
jreback: i mean make this a separate sub-section showing the previous and the new

pandas/core/series.py
```diff
@@ -4153,4 +4153,8 @@ def _try_cast(arr, take_fast_path):
                 data = np.array(data, dtype=dtype, copy=False)
             subarr = np.array(data, dtype=object, copy=copy)
 
+            # GH 20377
```
jreback: huh? this should *not* be necessary (not to mention non-performant)

pandas/tests/io/parser/na_values.py
```diff
@@ -369,3 +369,27 @@ def test_no_na_filter_on_index(self):
         expected = DataFrame({"a": [1, 4], "c": [3, 6]},
                              index=Index([np.nan, 5.0], name="b"))
         tm.assert_frame_equal(out, expected)
+
+    def test_na_values_with_dtype_str_and_na_filter_true(self):
```
jreback: can you parameterize this on na_filter (you will need to provide the nan_value as well in the parameterize as they are different)

pandas/tests/io/test_excel.py
```diff
@@ -360,6 +360,33 @@ def test_reader_dtype(self, ext):
...
+
+        tm.assert_frame_equal(actual, expected)
+
+        actual = self.get_exceldf(basename, ext,
+                                  dtype={'a': 'float64',
+                                         'b': 'float32',
+                                         'c': str,
+                                         'd': str})
+
+        expected['a'] = expected['a'].astype('float64')
```
jreback: move this higher (by the expected), you can simply construct things directly by using e.g. ``Series(...., dtype='float32')`` rather than a list

pandas/tests/io/test_excel.py
```diff
@@ -360,6 +360,33 @@ def test_reader_dtype(self, ext):
...
+        # GH 20377
+        basename = 'testdtype'
+        actual = self.get_exceldf(basename, ext)
+
+        expected = DataFrame({
+            'a': [1, 2, 3, 4],
+            'b': [2.5, 3.5, 4.5, 5.5],
+            'c': [1, 2, 3, 4],
+            'd': [1.0, 2.0, np.nan, 4.0]}).reindex(
+                columns=['a', 'b', 'c', 'd'])
```
jreback: just specify ``columns=list('abcd')`` rather than reindex

jreback (changes requested):

pandas/core/series.py
nikoskaragiannakis: Well, this fixes an error I got in this test https://github.com/pandas-dev/pandas/blob/master/pandas/tests/frame/test_dtypes.py#L533

This line https://github.com/nikoskaragiannakis/pandas/blob/7341cd17e11461728969afa159250e882b32dee0/pandas/core/series.py#L4154 does the damage by turning a `np.nan` to `'nan'`. But this comes directly from numpy, so I cannot change it.

Am I missing something?

pandas/tests/io/test_excel.py
nikoskaragiannakis: > move this higher (by the expected)

I'm not sure what you mean here.

> you can simply construct things directly by using e.g. Series(...., dtype='float32') rather than a list

First of all, this is copy-paste from the previous test, which was added for https://github.com/pandas-dev/pandas/issues/8212

Do you mean to do
```python
expected = DataFrame({'a': Series([1,2,3,4], dtype='float64'),
                      'b': Series([2.5,3.5,4.5,5.5], dtype='float32'),
                      ...})
```
?
If i use, for example, for the `'c'` column: `Series([1, 2, 3, 4], dtype=str)`, then it will give me `['1', '2', '3', '4']` instead of the expected `['001', '002', '003', '004']`.

pandas/core/series.py
jreback: how's that an error? that test is correct.

pandas/tests/frame/test_dtypes.py
```diff
@@ -529,7 +529,7 @@ def test_astype_str(self):
         # consistency in astype(str)
         for tt in set([str, compat.text_type]):
             result = DataFrame([np.NaN]).astype(tt)
-            expected = DataFrame(['nan'])
+            expected = DataFrame([np.NaN], dtype=object)
```
jreback: no leave this alone. turning a float nan into string should still work

pandas/tests/io/test_excel.py
jreback: yes exactly. If you need things like '001', then just do it that way, e.g. ``Series(['001', '002'....])``

pandas/tests/series/test_dtypes.py
```diff
@@ -149,6 +149,7 @@ def test_astype_str_map(self, dtype, series):
         # see gh-4405
         result = series.astype(dtype)
         expected = series.map(compat.text_type)
+        expected = expected.replace('nan', np.nan)  # see gh-20377
```
jreback: remove this, this is equiv to ``.astype(str)`` its mapping to a string and is correct

jreback (changes requested):

pandas/core/series.py
nikoskaragiannakis: I think that with the changes in the .pyx files, `np.nan` stays a float, even if you try to cast it to `str`.
For example
```
pd.DataFrame([np.NaN]).astype(str)
```
leaves `np.NaN` a float, while before the changes you used to get a `'nan'`.

So I have to ask: do we want calls like the one above to turn `np.NaN` into `nan` or not?

nikoskaragiannakis: @jreback ^

jreback:
closing as stale, if you want to continue working, pls ping and we can re-open. you will need to merge master.

Issue #19779, opened 2018-02-20
ASV: occasional asv failures on xlwt


jreback:
https://travis-ci.org/pandas-dev/pandas-ci/jobs/343616367
this is on current pandas master (this is just the CI job running it).

excel asv's

```
The following asvs benchmarks (if any) failed.
[ 43.02%] ··· Running io.excel.Excel.time_read_excel                  1/3 failed
                   xlwt      failed 
DONE displaying failed asvs benchmarks.
```

I have seen this work as well. maybe a race-condition?

jreback:
from ``(pandas) bash-3.2$ more asv_bench/benchmarks/io/excel.py ``
```
def setup(...):
      ....
        self.bio_write = BytesIO()
        self.bio_write.seek(0)
        self.writer_write = ExcelWriter(self.bio_write, engine=engine)

    def time_read_excel(self, engine):
        read_excel(self.bio_read)

    def time_write_excel(self, engine):
        self.df.to_excel(self.writer_write, sheet_name='Sheet1')
        self.writer_write.save()
```

I think the ``time_write_excel`` should have the ``.writer_write`` setup I think.

jorisvandenbossche:
If it is defined in the setup function (as it is now), it should be available in the benchmark function. It would be strange that this solves it.

jreback:
its passing 1 our of n times: https://travis-ci.org/pandas-dev/pandas-ci/jobs/345996588

jreback:
cc @mroeschke @WillAyd if you have any ideas

WillAyd:
Perhaps we should be explicitly closing the `BytesIO` objects that are getting created in a `teardown`? Due to the intermittency of it I'm wondering if the GC is taking an pass at closing those for us, but getting tripped up with the parallelized execution that asv provides

mroeschke:
Looks to be specifically a problem with the `time_read_excel` benchmark.

```
File "/home/travis/build/pandas-dev/pandas-ci/pandas/asv_bench/benchmarks/io/excel.py", line 29, in time_read_excel
                    read_excel(self.bio_read)
File "/home/travis/miniconda3/envs/pandas/lib/python3.6/site-packages/pandas-0.23.0.dev0+381.g8f1dfa74e-py3.6-linux-x86_64.egg/pandas/util/_decorators.py", line 172, in wrapper
                    return func(*args, **kwargs)
[...]

PR #19926, opened 2018-02-27
Added seek to buffer to fix xlwt asv failure

WillAyd:
- [X] closes #19779
- [X] tests added / passed
- [X] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] whatsnew entry

Sample ASV results from this as follows. Across 5 runs I got the upstream/master error every time and none on HEAD

```bash

(pandas_dev) williams-imac:asv_bench williamayd$ asv continuous -f 1.1 upstream/master HEAD -b io.excel.Excel.time_read_excel --show-stderr
· Creating environments
· Discovering benchmarks
·· Uninstalling from conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt.
·· Installing into conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt..
· Running 2 total benchmarks (2 commits * 1 environments * 1 benchmarks)
[  0.00%] · For pandas commit hash 15cd9d2d:
[  0.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[  0.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[ 50.00%] ··· Running io.excel.Excel.time_read_excel                                                                                         ok
[ 50.00%] ···· 
               ============ =============
                  engine                 
               ------------ -------------
                 openpyxl      133±4ms   
                xlsxwriter    118±0.7ms  
                   xlwt      54.2±0.05ms 
               ============ =============

[ 50.00%] · For pandas commit hash f4c9d966:
[ 50.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[ 50.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[...]

PR #20437, opened 2018-03-21
BUG: Patch for skipping seek() when loading Excel files from url

### LINKED ISSUE #20434: UnsupportedOperation 'seek' when loading excel files from url

```python
url = 'https://raw.github.com/pandas-dev/pandas/master/pandas/tests/io/data/test1.xls'
pandas.read_excel(url)
```
#### Problem description

In my version `0.23.0.dev0+657.g01882ba` I get an `UnsupportedOperation`:

```
UnsupportedOperation                      Traceback (most recent call last)
<ipython-input-5-71715bfd4345> in <module>()
----> 1 pandas.read_excel(url)

~/prj/pandas-mcrot/pandas/util/_decorators.py in wrapper(*args, **kwargs)
    170                 else:
    171                     kwargs[new_arg_name] = new_arg_value
--> 172             return func(*args, **kwargs)
    173         return wrapper
    174     return _deprecate_kwarg

~/prj/pandas-mcrot/pandas/util/_decorators.py in wrapper(*args, **kwargs)
    170                 else:
    171                     kwargs[new_arg_name] = new_arg_value
--> 172             return func(*args, **kwargs)
    173         return wrapper
    174     return _deprecate_kwarg

~/prj/pandas-mcrot/pandas/io/excel.py in read_excel(io, sheet_name, header, names, index_col, usecols, squeeze, dtype, engine, converters, true_values, false_values, skiprows, nrows, na_values, parse_dates, date_parser, thousands, comment, skipfooter, convert_float, **kwds)
    313 
    314     if not isinstance(io, ExcelFile):
--> 315         io = ExcelFile(io, engine=engine)
    316 
    317     return io._parse_excel(

~/prj/pandas-mcrot/pandas/io/excel.py in __init__(self, io, **kwds)
    390             if hasattr(io, 'seek'):
    391                 # GH 19779
--> 392                 io.seek(0)
    393 
    394             data = io.read()

UnsupportedOperation: seek
```
The PR #19926 was made in order to fix #19779. It introduced a call of seek() method
only for objects having a that method. In case of giving a URL to `pandas.read_excel()`, 
`seek()` is called on a `HTTPResponse` instance and it seems like that it does not support seeking, although the method `seek()` is available.

This issue is already covered by a test. When running 
```bash
pytest pandas/tests/io/test_excel.py
```
the test method `TestXlrdReader.test_read_from_http_url` fails for the same reason.

#### Expected Output

In version 0.22 this code returns the data as expected:
```
Out[4]: 
                   A         B         C         D
2000-01-03  0.980269  3.685731 -0.364217 -1.159738
2000-01-04  1.047916 -0.041232 -0.161812  0.212549
2000-01-05  0.498581  0.731168 -0.537677  1.346270
2000-01-06  1.120202  1.567621  0.003641  0.675253
2000-01-07 -0.487094  0.571455 -1.611639  0.103469
2000-01-10  0.836649  0.246462  0.588543  1.062782
2000-01-11 -0.157161  1.340307  1.195778 -1.097007
```

<details>
INSTALLED VERSIONS
------------------
commit: 01882ba5b4c21b0caf2e6b9279fb01967aa5d650
python: 3.6.4.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.0-116-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: de_DE.UTF-8
LOCALE: de_DE.UTF-8

pandas: 0.23.0.dev0+657.g01882ba
pytest: 3.4.2
pip: 9.0.1
setuptools: 38.5.1
Cython: 0.27.3
numpy: 1.14.2
scipy: 1.0.0
pyarrow: 0.8.0
xarray: 0.10.2
IPython: 6.2.1
sphinx: 1.7.1
patsy: 0.5.0
dateutil: 2.7.0
pytz: 2018.3
blosc: None
bottleneck: 1.2.1
tables: 3.4.2
numexpr: 2.6.4
feather: 0.4.0
matplotlib: 2.2.2
openpyxl: 2.5.0
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.2
lxml: 4.1.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.5
pymysql: 0.8.0
psycopg2: None
jinja2: 2.10
s3fs: 0.1.3
fastparquet: 0.1.4
pandas_gbq: None
pandas_datareader: None
</details>

mcrot:
- [x] closes #20434
- [x] tests added / passed
- [x] whatsnew entry

This is my first PR and I tried to follow the guideline. Sorry, if I overlooked something.
I don't have to add anything in whatsnew, because the bug isn't present in the current stable version 0.22, correct?

### DIFF
--- a/pandas/io/excel.py
+++ b/pandas/io/excel.py
@@ -10,6 +10,7 @@ import os
 import abc
 import warnings
 import numpy as np
+from io import UnsupportedOperation
 
 from pandas.core.dtypes.common import (
     is_integer, is_float,
@@ -388,8 +389,13 @@ class ExcelFile(object):
         elif not isinstance(io, xlrd.Book) and hasattr(io, "read"):
             # N.B. xlrd.Book has a read attribute too
             if hasattr(io, 'seek'):
-                # GH 19779
-                io.seek(0)
+                try:
+                    # GH 19779
+                    io.seek(0)
+                except UnsupportedOperation:
+                    # HTTPResponse does not support seek()
+                    # GH 20434
+                    pass
 
             data = io.read()
             self.book = xlrd.open_workbook(file_contents=data)


TomAugspurger:
cc @WillAyd

WillAyd:
Pretty simple change so lgtm - fine with it if you are Tom. Thanks for the contribution @mcrot

pandas/io/excel.py
```diff
@@ -10,6 +10,7 @@
 import abc
 import warnings
 import numpy as np
+from http.client import HTTPResponse
```
TomAugspurger: Did this exist in py2? May have to import from `pandas.compat`.

pandas/io/excel.py
```diff
@@ -387,7 +388,10 @@ def __init__(self, io, **kwds):
             self.book = io
         elif not isinstance(io, xlrd.Book) and hasattr(io, "read"):
             # N.B. xlrd.Book has a read attribute too
-            if hasattr(io, 'seek'):
+            #
+            # Excluding the case http.client.HTTPResponse.seek() here
+            # because it raises an UnsupportedOperation exception
+            if not isinstance(io, HTTPResponse) and hasattr(io, 'seek'):
```
TomAugspurger: I worry a bit about this hardcoding `HTTPResponse` here.

IIUC, `io` could be anything that implements the `__fspath__` protocol. Not all of those objects will implement `seek`.

What's the harm in cactching `(OSError, ValueError)` here? (I don't really know why we seek in the first place)

WillAyd: Back in #19779 we were getting very intermittent failures on ASV tests due to a `TypeError` being thrown. tbh I'm still not entirely clear on *why* it was happening but I believe that intermittently that buffer was being read before this code was being executed. Adding in the conditional seek back to the beginning it prevented the `TypeError`.

So it's certainly a fair question on your end to which I don't have a great answer. try...catch could be an option, but then we'd probably have to seek back to the start in the catch block and maybe still have the conditional (to distinguish from other `TypeError`s) so not sure there is a ton to gain there

pandas/io/excel.py
mcrot: Thanks for your feedback, Tom. You're right, `http.client` was named `httplib` in py2, I forgot to check py2 compatibility, sorry. So maybe `import pandas.compat.httplib as httplib` and using `httplib.HTTPResponse` would work, but I will think about your and @WillAyd's comment first.

WillAyd: Shouldn't this have failed on the AppVeyor 2.7 build?

mcrot: Yes, but it worked .. I'm just trying to figure out why.

mcrot: Sorry, I've no idea why the code worked for Python 2.7 - not only in AppVeyor, but also in TravisCI and CircleCI. In all of them pytest reports to use py 2.7.14. And the [docs](https://docs.python.org/2/library/httplib.html?highlight=httplib#module-httplib) clearly state that `http.client` has come up with py3. As example, `http.client` is not available in my local py27 environment:
```
Python 2.7.14 | packaged by conda-forge | (default, Dec 25 2017, 01:16:05) 
[GCC 4.8.2 20140120 (Red Hat 4.8.2-15)] on linux2
>>> from http.client import HTTPResponse
ImportError: No module named http.client
```
How is this possible, is there some CI magic I haven't understood yet?
I've also searched for some other compatibility libs which may have used in the pandas code base in order to achieve this, but haven't found any.

I think it would be good for me to understand this before implementing a py2 compatible fix here.

TomAugspurger: Are you able to set up a python2 dev environment and reproduce locally? That may be easier than going through the CI logs.

TomAugspurger: Just tested on your branch in my py2 env

```python

In [1]: import pandas as pd
ImportError                               Traceback (most recent call last)
<ipython-input-1-af55e7023913> in <module>()
----> 1 import pandas as pd

/Users/taugspurger/sandbox/pandas-ip-2/pandas/pandas/__init__.pyc in <module>()
     55
     56 from pandas.util._print_versions import show_versions
---> 57 from pandas.io.api import *
     58 from pandas.util._tester import test
     59 import pandas.testing

/Users/taugspurger/sandbox/pandas-ip-2/pandas/pandas/io/api.py in <module>()
      7 from pandas.io.parsers import read_csv, read_table, read_fwf
      8 from pandas.io.clipboards import read_clipboard
----> 9 from pandas.io.excel import ExcelFile, ExcelWriter, read_excel
     10 from pandas.io.pytables import HDFStore, get_store, read_hdf
     11 from pandas.io.json import read_json

/Users/taugspurger/sandbox/pandas-ip-2/pandas/pandas/io/excel.py in <module>()
     11 import warnings
     12 import numpy as np
---> 13 from http.client import HTTPResponse
     14
     15 from pandas.core.dtypes.common import (

ImportError: No module named http.client


```

So something strange is going on.

TomAugspurger: I wonder if one of our dependencies installs an `http` backport?

mcrot: I've tried to use `tox -e py27` in order to test with python 2.7 - that did not work, because the `bigquery` dependency could not be satisfied. I'll try to create a separate directory and build with python 2.7, that's maybe a better idea.

I'll also have a look if there are dependecies to a `http` backport...

mcrot: Hm, I've built a local development environment mit py27 and the tests fail as expected, because `http.client` cannot be imported. So it seems this is only related to the CI mechanism..

In order to check the CI dependencies, I've made a small py2 script which collects all dependencies from the files `ci/requirements-2.7(_[A-Z]+)?\.(build|pip|run)`, imports them one by one (assuming the correct version is installed) and tries to import `http.client` after each import. Haven't found a backport of `http` by doing this. 

I've added the script and my output here for reference. I had to manually install some of the collected dependencies before (couldn't find 'nomkl' btw). Not sure whether this helps.

```python

import os
import re

import logging
logging.basicConfig(level=logging.INFO)
    
if __name__=='__main__':
    DEPS_DIR = '/home/mcrot/prj/pandas-mcrot-py2/ci'
    DEPS_REGEX = r'^requirements-2.7(_[A-Z]+)?\.(build|pip|run)$'

    DEPS_BLACKLIST = ['python', 'ipython']

    DEPS_IMPORT_NAMES = { 'PyCrypto': 'Crypto',
                          'pytables': 'tables',
                          'python-dateutil': 'dateutil',
                          'beautifulsoup4': 'bs4',
                          'pandas-gbq': 'pandas_gbq'}
        
    deps = set()
    
    for fn in os.listdir(DEPS_DIR):
        if re.match(DEPS_REGEX, fn):
            logging.info("Loading deps from file '%s'..", fn)

            f = open(os.path.join(DEPS_DIR, fn))

            new_deps = []
            for s in f.readlines():
                dep = s.split('=')[0].strip()

                if dep in DEPS_BLACKLIST:
                    continue
                
                # rename if possible
                try:
                    dep = DEPS_IMPORT_NAMES[dep]
                except KeyError:
                    pass

                new_deps.append(dep)
            
            deps = deps.union(new_deps)

    deps = sorted(d for d in deps)

    logging.info("Checking these dependencies: %s", deps)
                
    for dep in deps:

        try:
            mod = __import__(dep)
        except ImportError:
            logging.warning("Cannot import module '%s'", dep)
            continue
            
        try:
            from http.client import HTTPResponse
            logging.error("Module '%s' DOES IMPORT http.", dep)
            raise Exception("Backport of http detected in module '%s'" % str(mod))
        except ImportError as exc:
            # as expected for py2
            logging.info("Module '%s' does not import http.", mod)
```

[check-backport.txt](https://github.com/pandas-dev/pandas/files/1834240/check-backport.txt)

So I guess, I need a local CI in order to find out why `http.client` can be imported?

TomAugspurger:
That sounds difficult. You might try adding a `conda list` to the end of
`ci/install_travis.sh` to print out the final version of all the packages.

My also throw a `python -c "import http.client"` there too, to ensure that
our hypothesis about a backport is correct.

jreback:
this might simpler, in ``pandas/compat/__init__.py`` we *already* do

```
try:
    import __builtin__ as builtins
    # not writeable when instantiated with string, doesn't handle unicode well
    from cStringIO import StringIO as cStringIO
    # always writeable
    from StringIO import StringIO
    BytesIO = StringIO
    import cPickle
    import httplib
except ImportError:
    import builtins
    from io import StringIO, BytesIO
    cStringIO = StringIO
    import pickle as cPickle
    import http.client as httplib
```

so you are likely getting:https://docs.python.org/2/library/httplib.html in python 2 (assuming we are doing the import from compat, which we *should* be doing)

TomAugspurger:
> which we should be doing

Right, but the latest CI run did an unconditional `import http.client`, which should have failed on py2 (and does for us locally, but not on any of the CI runs).

jreback:
its possible this test is skipped on a 2.7 run?

TomAugspurger:
Possibly, but I *think* that the import is hit by an `import pandas`, so that shouldn't matter. The entire test suite should fail (unless I'm misunderstanding something).

mcrot:
The test is not skipped and the `import http.client` statement is on top of `pandas.io.excel` and unconditional, so it is executed on every test which imports the `excel` modul. 
I've created a new branch in my repo, activated TravisCI and added the suggested commands at the end of the install script. Let's see..

mcrot:
Ok, the TravisCI run my branch [`why-does-import-http-client-work-in-py2`](https://github.com/mcrot/pandas/tree/why-does-import-http-client-work-in-py2) failed already  on the first `import pandas` as @TomAugspurger expected, see [here](https://travis-ci.org/mcrot/pandas). I don't understand why the pandas-dev TravisCI runs pass without problems - and also for the other two CI environments.

I've had a look at of a diff of TravisCI logs of both instances and haven't found a significant difference yet. The TravisCI on my repo had to built some wheels first and couldn't resuse Cython cache as the pandas-dev instance did. That shouldn't matter?!

However I'm going to temporarily add the extra output of `conda list` at the end of `travis_install.sh` also to this PR in order to see its output on pandas-dev TravisCI instance. Just to be sure we have the same package versions everywhere.

Concerning this PR here, maybe this py2 CI issue is slowly getting a little off-topic.. maybe I should be handled elsewhere in the long run, I'm not sure how you normally deal with CI issues like that. Or there is an easy explanation I cannot see, then sorry for the noise.

TomAugspurger:
> Concerning this PR here, maybe this py2 CI issue is slowly getting a little off-topic.. maybe I should be handled elsewhere in the long run

Yes, feel free to just make the changes to fix the excel bug (ideally tested locally with Py2), and we'll fix the CI in a followup.

mcrot:
Ok, thanks .. it's a little weird, because this time the tests here on pandas-dev also fail for py27 as expected.

However I've continued with the PR and just catched `OSError` and `ValueError` as suggested by @TomAugspurger. Btw. would catching `io.UnsupportedOperation` be too specific? It is available in py2 and py3. 

Locally the tests run both for py27 and py36, I will push now. I hope this PR does not affect negatively affect ASV tests as mentioned by @WillAyd.

TomAugspurger:
> Btw. would catching io.UnsupportedOperation be too specific? It is available in py2 and py3.

That should be equivalent, right? I just assumed it wasn't around in python2, but if so the use that.

TomAugspurger:
Great.

Could you open an issue with a small summary to what we found here about
the CI issue?

mcrot:
Done, see #20453.

jreback:
can you add the test from the issue?

mcrot:
The test which reveals #20434 is already included in the code base. I openend #20434, because
I'd just setup a working environment following the guide [Contributing to pandas](https://pandas-docs.github.io/pandas-docs-travis/contributing.html) and afterwards, when running the tests, especially
```
pytest pandas/tests/io/test_excel.py
```
I got failures for the test method `TestXlrdReader.test_read_from_http_url`. The failure is described in #20434.

I understand now that is strange that my tests didn't run, because #19926 introduced the seek() method and the tests should have failed already in the past for that PR. Probably the issue here is more related to my setup than to the code base or the CI, but I'm not sure. Now I think, a better approach would have been to wait for somebody else to confirm the issue #20434 before creating this PR here, right?

jreback:
if the test doesn't fail on CI, then either your fix doesn't actually fix anything, or the CI is not running it (or properly running it). so need to have a failure before it can get fixed.

TomAugspurger:
@mcrot @jreback I suspect this is because we don't run slow tests on py3. So we can do a couple things

- mock the test at the network layer and remove the slow marker
- ensure that at least one of our slow jobs in running on 3.6

We should make sure that we run the slow test against 3.6 regardless of 2.7. I'm planning some cleanup of our CI setup, to make these easier to spot.

Regarding this PR, I'm fine with merging as is. I'm not sure how difficult mocking would be, but I'd like to get our tests passing again.

jreback:
https://github.com/pandas-dev/pandas/pull/20559

let's see if this fails

TomAugspurger:
Merged in master and re-ran. @mcrot ping on green.

mcrot:
Green from my point of view. Thanks @jreback @TomAugspurger. If you still need me to do a rebase (not sure), I'm back in my office in about 15h.

PR #20443, opened 2018-03-21
DOC: update the Series.between docstring

mrw34:
an unrelated PR):

```
###################### Docstring (pandas.Series.between)  ######################

Return boolean Series equivalent to left <= series <= right.

This function returns a boolean vector containing `True` wherever the
corresponding Series element is between the boundary values `left` and
`right`. NA values are treated as `False`.

Parameters
----------
left : scalar
    Left boundary.
right : scalar
    Right boundary.
inclusive : bool, default True
    Include boundaries.

Returns
-------
is_between : Series

Examples
--------

>>> s = pd.Series([2, 0, 4, 8, np.nan])

Boundary values are included by default:

>>> s.between(1, 4)
0     True
1    False
2     True
3    False
4    False
dtype: bool

With `inclusive` set to `False` boundary values are excluded:

>>> s.between(1, 4, inclusive=False)
0     True
1    False
2    False
3    False
4    False
dtype: bool

`left` and `right` can be any scalar value:

>>> s = pd.Series(['Alice', 'Bob', 'Carol', 'Eve'])
>>> s.between('Anna', 'Daniel')
0    False
1     True
2     True
3    False
dtype: bool

See Also
--------
DataFrame.query : Query the columns of a frame with a boolean
    expression.

Docstring for "pandas.Series.between" correct. :)
```

### DIFF
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -3528,11 +3528,19 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         Returns
         -------
-        is_between : Series
+        Series of bool
+
+        Notes
+        -----
+        This function is equivalent to `(left <= series) & (series <= right)`
+
+        See Also
+        --------
+        pandas.Series.gt : Greater than of series and other
+        pandas.Series.lt : Less than of series and other
 
         Examples
         --------
-
         >>> s = pd.Series([2, 0, 4, 8, np.nan])
 
         Boundary values are included by default:
@@ -3564,11 +3572,6 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         2     True
         3    False
         dtype: bool
-
-        See Also
-        --------
-        DataFrame.query : Query the columns of a frame with a boolean
-            expression.
         """
         if inclusive:
             lmask = self >= left
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -3528,11 +3528,12 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         Returns
         -------
-        Series of bool
+        Series
+            Each element will be a boolean.
 
         Notes
         -----
-        This function is equivalent to `(left <= series) & (series <= right)`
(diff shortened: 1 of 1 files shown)


pandas/core/series.py
```diff
@@ -3511,19 +3511,64 @@ def isin(self, values):
...
+        0    False
+        1     True
+        2     True
+        3    False
+        dtype: bool
+
+        See Also
+        --------
+        DataFrame.query : Query the columns of a frame with a boolean
+            expression.
```
datapythonista: Personally I don't find `DataFrame.query` so related. But if we list it, I'd explain it a bit better than it can be used to perform comparisons, so users understand why it's related.

The ones that I do find related are `Series.gt` and `Series.lt`.

The `See Also` section should be placed before the examples, not after.

pandas/core/series.py
```diff
@@ -3511,19 +3511,64 @@ def isin(self, values):
...
+        inclusive : bool, default True
+            Include boundaries.
 
         Returns
         -------
         is_between : Series
+
+        Examples
+        --------
+
```
datapythonista: I think we don't need this blank line.

pandas/core/series.py
```diff
@@ -3511,19 +3511,64 @@ def isin(self, values):
...
+            Left boundary.
         right : scalar
-            Right boundary
+            Right boundary.
+        inclusive : bool, default True
+            Include boundaries.
 
         Returns
         -------
         is_between : Series
```
datapythonista: You can leave just `Series`, or `Series of bool` to be more explicit. The returned name `is_between` is not needed.

datapythonista (commented):
Looks great, added couple of comments.

mrw34:
@datapythonista Thanks Marc! Hopefully looking better now...

datapythonista (approved):
lgtm

TomAugspurger:
Thanks @mrw34

PR #20450, opened 2018-03-22
DOC: update the pandas.Series.map docstring

kaoboyandy:
an unrelated PR):

```
######################## Docstring (pandas.Series.map)  ########################

Map values of Series according to input correspondence.

Used for substituting each value in a Series with another value,
that may be derived from a function, a ``dict`` or
a :class:`pandas.Series`.

Parameters
----------
arg : function, dict, or Seriess
    Mapping correspondence.
na_action : {None, 'ignore'}
    If 'ignore', propagate NA values, without passing them to the
    mapping correspondence.

Returns
-------
y : Series
    Same index as caller.

Examples
--------

Map inputs to outputs (both of type :class:`pandas.Series`):

>>> x = pd.Series([1,2,3], index=['one', 'two', 'three'])
>>> x
one      1
two      2
three    3
dtype: int64

>>> y = pd.Series(['foo', 'bar', 'baz'], index=[1,2,3])
>>> y
1    foo
2    bar
3    baz
dtype: object

>>> x.map(y)
one   foo
two   bar
three baz
dtype: object

Map a function to a :class:`pandas.Series`.

>>> x.map(lambda x: x**2)
one      1
two      4
three    9
dtype: int64

If ``arg`` is a dictionary, return a new :class:`pandas.Series` with
values converted according to the dictionary's mapping:

>>> z = {1: 'A', 2: 'B', 3: 'C'}

>>> x.map(z)
one   A
two   B
three C
dtype: object

Use ``na_action`` to control whether NA values are affected by the
mapping function.

>>> s = pd.Series([1, 2, 3, np.nan])

>>> s.map('this is a string {}'.format, na_action=None)
0    this is a string 1.0
1    this is a string 2.0
2    this is a string 3.0
3    this is a string nan
dtype: object

>>> s.map('this is a string {}'.format, na_action='ignore')
0    this is a string 1.0
1    this is a string 2.0
2    this is a string 3.0
3                     NaN
dtype: object

See Also
--------
Series.apply : For applying more complex functions on a Series.
DataFrame.apply : Apply a function row-/column-wise.
DataFrame.applymap : Apply a function elementwise on a whole DataFrame.

Notes
-----
When `arg` is a dictionary, values in Series that are not in the
dictionary (as keys) are converted to ``NaN``. However, if the
dictionary is a ``dict`` subclass that defines ``__missing__`` (i.e.
provides a method for default values), then this default is used
rather than ``NaN``:

>>> from collections import Counter
>>> counter = Counter()
>>> counter['bar'] += 1
>>> y.map(counter)
1    0
2    1
3    0
dtype: int64

Docstring for "pandas.Series.map" correct. :)
```

explicitly.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


pandas/core/series.py
```diff
@@ -2850,7 +2853,7 @@ def map(self, arg, na_action=None):
         Examples
         --------
 
-        Map inputs to outputs (both of type `Series`):
+        Map inputs to outputs (both of type :class:`pandas.Series`):
 
         >>> x = pd.Series([1,2,3], index=['one', 'two', 'three'])
```
datapythonista: Missing spaces after comma for PEP-8.

pandas/core/series.py
```diff
@@ -2864,35 +2867,46 @@ def map(self, arg, na_action=None):
...
+
+        Map a function to a :class:`pandas.Series`.
 
-        If `arg` is a dictionary, return a new Series with values converted
-        according to the dictionary's mapping:
+        >>> x.map(lambda x: x**2)
+        one      1
+        two      4
+        three    9
+        dtype: int64
```
datapythonista: I would start by the example using a function as a parameter, which I think it's a more common use than with a series.

pandas/core/series.py
```diff
@@ -2864,35 +2867,46 @@ def map(self, arg, na_action=None):
...
+
+        If ``arg`` is a dictionary, return a new :class:`pandas.Series` with
+        values converted according to the dictionary's mapping:
 
         >>> z = {1: 'A', 2: 'B', 3: 'C'}
 
         >>> x.map(z)
```
datapythonista: As `z` is not very descriptive, I'd use the dictionary directly as a parameter, and avoid saving it to a variable.

pandas/core/series.py
```diff
@@ -2850,7 +2853,7 @@ def map(self, arg, na_action=None):
         Examples
```
datapythonista: Can you add a `See Also` section? I think at least `apply` and `applymap` are related. May be the Python `map` built-in could be linked too?

pandas/core/series.py
```diff
@@ -2864,35 +2867,46 @@ def map(self, arg, na_action=None):
         1    foo
         2    bar
         3    baz
+        dtype: object
```
datapythonista: I think these examples would be clearer if instead of `foo`, `bar`, `one`, `two`... we could find a more real-world example.

Also, I think it should be useful to show the behaviour of `map` with a `dict` or `Series` that doesn't have all the values in the `Series`.

TomAugspurger:
@kaoboyandy will you have time to update based on @datapythonista's comments?

datapythonista (approved):
@WillAyd do you mind taking a quick look (I made some last changes), and if you're happy merge it?

PR #20458, opened 2018-03-22
DOC: update the pandas.Series.str.startswith docstring

dcreekp:
an unrelated PR):

```
################### Docstring (pandas.Series.str.startswith) ###################

Test if the start of each string element matches a pattern.

Return a Series of booleans indicating whether the given pattern matches
the start of each string element.
Equivalent to :meth: `str.startswith`.

Parameters
----------
pat : string
    Character sequence.
na : object, default NaN
    Character sequence shown if element tested is not a string.

Returns
-------
startswith : Series/array of boolean values

Examples
--------
>>> s = pd.Series(['bat', 'bear', 'cat'])
>>> s.str.startswith('b')
0     True
1     True
2    False
dtype: bool
>>> s = pd.Series(['bat', 'bear', 'cat', np.nan])
>>> s.str.startswith('b', na='not_a_string')
0            True
1            True
2           False
3    not_a_string
dtype: object

See Also
--------
endswith : same as startswith, but tests the end of string

```

explicitly.

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -330,40 +330,41 @@ def str_startswith(arr, pat, na=np.nan):
     """
     Test if the start of each string element matches a pattern.
 
-    Return a Series of booleans indicating whether the given pattern matches
-    the start of each string element.
-    Equivalent to :meth: `str.startswith`.
+    Equivalent to :meth:`str.startswith`.
 
     Parameters
     ----------
-    pat : string
-        Character sequence.
+    pat : str
+        Character sequence. Regular expressions are not accepted.
     na : object, default NaN
-        Character sequence shown if element tested is not a string.
+        Object shown if element tested is not a string.
 
     Returns
     -------
-    startswith : Series/array of boolean values
-
-    Examples
-    --------
-    >>> s = pd.Series(['bat', 'bear', 'cat'])
-    >>> s.str.startswith('b')
-    0     True
-    1     True
-    2    False
-    dtype: bool
-    >>> s = pd.Series(['bat', 'bear', 'cat', np.nan])
-    >>> s.str.startswith('b', na='not_a_string')
-    0            True
-    1            True
-    2           False
-    3    not_a_string
-    dtype: object
+    startswith : Series or array-like of bool
+        A Series of booleans indicating whether the given pattern matches
+        the start of each string element.
 
     See Also
     --------
-    endswith : same as startswith, but tests the end of string
+    str_endswith : Same as startswith, but tests the end of string.
+    str.startswith : Python standard library string method.
+
+    Examples
+    --------
+    >>> s = pd.Series(['bat', 'Bear', 'cat', np.nan])
+    >>> s.str.startswith('b')
+    0     True
+    1    False
+    2    False
+    3      NaN
+    dtype: object
+    >>> s.str.startswith('b', na=False)
+    0     True
+    1    False
+    2    False
+    3    False
+    dtype: bool
     """
     f = lambda x: x.startswith(pat)
     return _na_map(f, arr, na, dtype=bool)
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -341,7 +341,7 @@ def str_startswith(arr, pat, na=np.nan):
 
     Returns
     -------
-    startswith : Series or array-like of bool
+    Series or Index of bool
         A Series of booleans indicating whether the given pattern matches
         the start of each string element.
 
@@ -360,6 +360,10 @@ def str_startswith(arr, pat, na=np.nan):
     2    False
     3      NaN
     dtype: object
+
+    Specifying `na` to be `False` instead of `NaN`.
+
+    >>> s = pd.Series(['bat', 'Bear', 'cat', np.nan])
     >>> s.str.startswith('b', na=False)
     0     True
     1    False
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -328,19 +328,42 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
 
 def str_startswith(arr, pat, na=np.nan):
     """
-    Return boolean Series/``array`` indicating whether each string in the
-    Series/Index starts with passed pattern. Equivalent to
-    :meth:`str.startswith`.
+    Test if the start of each string element matches a pattern.
+
+    Return a Series of booleans indicating whether the given pattern matches
+    the start of each string element.
+    Equivalent to :meth: `str.startswith`.
 
     Parameters
     ----------
     pat : string
-        Character sequence
-    na : bool, default NaN
+        Character sequence.
+    na : object, default NaN
+        Character sequence shown if element tested is not a string.
 
     Returns
     -------
     startswith : Series/array of boolean values
+
+    Examples
+    --------
+    >>> s = pd.Series(['bat', 'bear', 'cat'])
+    >>> s.str.startswith('b')
+    0     True
+    1     True
+    2    False
+    dtype: bool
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -328,19 +328,42 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
...
+    dtype: bool
+    >>> s = pd.Series(['bat', 'bear', 'cat', np.nan])
+    >>> s.str.startswith('b', na='not_a_string')
+    0            True
+    1            True
+    2           False
+    3    not_a_string
+    dtype: object
+
+    See Also
```
TomAugspurger: Can you move the See Also before the Examples.

pandas/core/strings.py
```diff
@@ -328,19 +328,42 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
...
+    --------
+    endswith : same as startswith, but tests the end of string
```
TomAugspurger: Add `str.startswith`.

pandas/core/strings.py
```diff
@@ -328,19 +328,42 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
...
+    Return a Series of booleans indicating whether the given pattern matches
+    the start of each string element.
+    Equivalent to :meth: `str.startswith`.
 
     Parameters
     ----------
     pat : string
-        Character sequence
-    na : bool, default NaN
+        Character sequence.
```
TomAugspurger: Can `pat` be a regex? I don't think it can. Best to state that explicitly.

pandas/core/strings.py
datapythonista: If I'm not wrong, I think your `endswith` is linked by Sphinx to `pandas.endswith` which doesn't exist. I think the right way would be `Series.str.endswith`. Similar examples link to the Python standard library version, `str.startswith` (the one @TomAugspurger says).

If I'm not wrong, the descriptions in the `See Also` section should start with a capital letter and finish with a period.

dcreekp: I can't seem to get a link, have tried:
str_endswith
Series.str_endswith
Series.str.endswith
Series.string.endswith
endswith
Series.str.str_endswith

dcreekp:
################### Docstring (pandas.Series.str.startswith) ###################

Test if the start of each string element matches a pattern.

Equivalent to :meth:`str.startswith`.

Parameters
----------
pat : str
    Character sequence. Regular expressions are not accepted.
na : object, default NaN
    Object shown if element tested is not a string.

Returns
-------
startswith : Series or array-like of bool
    A Series of booleans indicating whether the given pattern matches
    the start of each string element.

See Also
--------
str_endswith : Same as startswith, but tests the end of string.
str.startswith : Python standard library string method.

Examples
--------
>>> s = pd.Series(['bat', 'Bear', 'cat', np.nan])
>>> s.str.startswith('b')
0     True
1    False
2    False
3      NaN
dtype: object
>>> s.str.startswith('b', na=False)
0     True
1    False
2    False
3    False
dtype: bool

Docstring for "pandas.Series.str.startswith" correct. :)
@TomAugspurger @datapythonista 
incorporated the feedback, I left a comment on the one I couldn't work out

pandas/core/strings.py
```diff
@@ -328,19 +328,43 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
...
     -------
-    startswith : Series/array of boolean values
+    startswith : Series or array-like of bool
+        A Series of booleans indicating whether the given pattern matches
+        the start of each string element.
+
+    See Also
+    --------
+    str_endswith : Same as startswith, but tests the end of string.
+    str.startswith : Python standard library string method.
```
datapythonista: I think there is some bug preventing the links to be generated when you build the html with the `--single` option. If you build the whole documenation `doc/make.py html`, it will take something like 5 minutes to complete, but I think you should have the links working.

Replacing `str_endswith` with `Series.str.endswith` is the right way.

Also, I'd personally add `Series.str.contains`, which also looks for the pattern, but in any position. So, I think it could be useful for some users visiting the page to know about it too.

dcreekp: yep thanks.
The entire build one shows the links properly.
Added contains

pandas/core/strings.py
```diff
@@ -328,19 +328,44 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
...
-    na : bool, default NaN
+    pat : str
+        Character sequence. Regular expressions are not accepted.
+    na : object, default NaN
+        Object shown if element tested is not a string.
 
     Returns
     -------
-    startswith : Series/array of boolean values
+    startswith : Series or array-like of bool
```
datapythonista: I was checking the other methods of the `.str.` accessor, and seems like we're a bit inconsistent with the type. In some cases we use `Series/array` or `Series or array-like`, and in some others `Series or Index of whatever`. I think the latter is more specific, as I don't think currently this method can return a numpy array or a Python list.

We can change it at a later point, as some unifying would be nice. But if you want to change it now to `Series or Index of bool` we'd have this right already.

Also, the `startswith : ` here is something that we don't want to use anymore, unless the function/method returns more than one value. So you can leave just the type.

pandas/core/strings.py
```diff
@@ -328,19 +328,44 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
...
+    1    False
+    2    False
+    3      NaN
+    dtype: object
+    >>> s.str.startswith('b', na=False)
+    0     True
+    1    False
+    2    False
+    3    False
+    dtype: bool
```
datapythonista: I think in more cases we show `>>> s` after defining it, and we leave a blank line between the different examples, so they are in different boxes in the html. Also, for the last example a short explanation of what you are doing could be useful for users.

You can see an example of what I mean here: https://github.com/dcreekp/pandas/blob/39f76413374109d8c34021a6b61d121d3d05c9a0/pandas/core/strings.py#L1347 probably we don't need many explanations in this case, as the example is quite obvious, but a short sentence for the last case could help users see what's going on faster.

dcreekp:
@datapythonista changed as per suggestions:)

pandas/core/strings.py
```diff
@@ -328,19 +328,48 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
...
+    See Also
+    --------
+    str.startswith : Python standard library string method.
+    Series.str.endswith : Same as startswith, but tests the end of string.
+    Series.str.contains : Tests if string element contains a pattern.
+
+    Examples
+    --------
+    >>> s = pd.Series(['bat', 'Bear', 'cat', np.nan])
+    >>> s.str.startswith('b')
```
datapythonista: Sorry, I think I wasn't clear in part of my last comment. The explanation you added looks great, that part is perfect.

But what I meant with adding `>>> s` is:

```
>>> s = pd.Series(['bat', 'Bear', 'cat', np.nan])
>>> s
(the user can see the series here)

```

So, you'd have a first block (ended with a blank line to make it a different box), where the user can see the data you'll be using in both examples.

Then you show the basic usage (line 357 currently), then the explanation of the second example you added. And for the second example you don't need to create the series again, as you had before.

So it'd be simply adding `>>> s`, its output and a blank linne after L356. And removing L366. That would be the standard way use in most examples, and IMO the clearest.

Sorry my previous comment was confusing.

dcreekp:
@datapythonista ok please check

datapythonista (approved):
Looks perfect to me, really good job. Thanks!

TomAugspurger:
Thanks!

dcreekp:
Cool. Thanks for all the feedback!

PR #20463, opened 2018-03-22
DOC: update the Series.str.join docstring

fdroessler:
an unrelated PR):

```
###################### Docstring (pandas.Series.str.join) ######################

Join lists contained as elements in the Series/Index with passed delimiter.

If the elements of a Series are lists themselves, join the content of these
lists using the delimiter passed to the function.
This function is an equivalent to :meth:`str.join`.

Parameters
----------
sep : str
    Delimiter to use between list entries.

Returns
-------
Series/Index of objects

Notes
-----
If any of the lists does not contain string objects the result of the join
will be `NaN`.

See Also
--------
str.join : Standard library version of this method.
Series.str.split : Split strings around given separator/delimiter.

Examples
--------

Example with a list that contains non-string elements.

>>> s = pd.Series([['lion', 'elephant', 'zebra'],
...                [1.1, 2.2, 3.3],
...                ["cat", np.nan, "dog"],
...                ["cow", 4.5, "goat"]])
>>> s
0    [lion, elephant, zebra]
1            [1.1, 2.2, 3.3]
2            [cat, nan, dog]
3           [cow, 4.5, goat]
dtype: object

Join all lists using an '-', the list of floats will become a NaN.

>>> s.str.join('-')
0    lion-elephant-zebra
1                    NaN
2                    NaN
3                    NaN
dtype: object

Docstring for "pandas.Series.str.join" correct. :)

```

explicitly.

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -969,40 +969,23 @@ def str_join(arr, sep):
     Examples
     --------
 
-    >>> df = pd.Series({1: ['dog', 'cat', 'fish'],
-    ...                 2: ['lion', 'elephant', 'zebra'],
-    ...                 3: ['cow', 'pig', 'chicken']})
-    >>> df
-    1           [dog, cat, fish]
-    2    [lion, elephant, zebra]
-    3        [cow, pig, chicken]
-    dtype: object
-
-    Join all lists using '_'.
-
-    >>> df.str.join('_')
-    1           dog_cat_fish
-    2    lion_elephant_zebra
-    3        cow_pig_chicken
-    dtype: object
-
     Example with a list that contains non-string elements.
 
-    >>> df = pd.Series({1: [1.1, 2.2, 3.3],
-    ...                 2: ['lion', 'elephant', 'zebra'],
-    ...                 3: ['cow', 'pig', 'chicken']})
-    >>> df
-    1            [1.1, 2.2, 3.3]
-    2    [lion, elephant, zebra]
-    3        [cow, pig, chicken]
+    >>> s = pd.Series({1: ['lion', 'elephant', 'zebra'],
+    ...                2: [1.1, 2.2, 3.3],
+    ...                3: [np.nan, np.nan, np.nan]})
+    >>> s
+    1    [lion, elephant, zebra]
+    2            [1.1, 2.2, 3.3]
+    3            [nan, nan, nan]
     dtype: object
 
     Join all lists using an '-', the list of floats will become a NaN.
 
-    >>> df.str.join('-')
-    1                    NaN
-    2    lion-elephant-zebra
-    3        cow-pig-chicken
+    >>> s.str.join('-')
+    1    lion-elephant-zebra
+    2                    NaN
+    3                    NaN
     dtype: object
     """
     return _na_map(sep.join, arr)
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -941,13 +941,14 @@ def str_get_dummies(arr, sep='|'):
 
 def str_join(arr, sep):
     """
-    Join lists contained as elements in the Series/Index with
-    passed delimiter. Equivalent to :meth:`str.join`.
+    Join lists contained as elements in the Series/Index with passed delimiter.
+
+    This function is an equivalent to :meth:`str.join`.
 
     Parameters
     ----------
     sep : string
-        Delimiter
+        Delimiter to use between list entries.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -941,17 +941,69 @@ def str_get_dummies(arr, sep='|'):
...
+
+    See Also
+    --------
+    str.join : Standard library version of this method.
+    Series.str.split : Split strings around given separator/delimiter.
+
+    Examples
+    --------
+
+    >>> df = pd.Series({1: ['dog', 'cat', 'fish'],
```
datapythonista: can you rename `df` to `s`. `df` is used for `DataFrame`, and makes it a bit confusing.

pandas/core/strings.py
```diff
@@ -941,17 +941,69 @@ def str_get_dummies(arr, sep='|'):
...
+
+    >>> df.str.join('_')
+    1           dog_cat_fish
+    2    lion_elephant_zebra
+    3        cow_pig_chicken
+    dtype: object
+
+    Example with a list that contains non-string elements.
+
+    >>> df = pd.Series({1: [1.1, 2.2, 3.3],
```
datapythonista: Same as before.

pandas/core/strings.py
```diff
@@ -941,17 +941,69 @@ def str_get_dummies(arr, sep='|'):
...
+    3        [cow, pig, chicken]
+    dtype: object
+
+    Join all lists using an '-', the list of floats will become a NaN.
+
+    >>> df.str.join('-')
+    1                    NaN
+    2    lion-elephant-zebra
+    3        cow-pig-chicken
+    dtype: object
```
datapythonista: I think that this second example, the first one is actually a bit redundant. As this exemplifies both cases, strings and not strings.

And I think we could even show in one of the rows the floats, another the strings (both as you did), and use the third one to illustrate a list with a `NaN`, which I assume it returns a `NaN`, but it may be not obvious for all users.

datapythonista (commented):
Looks great, just a comment about naming conventions, and an idea to make examples more compact and more illustrative.

fdroessler:
Thanks for the feedback @datapythonista have updated the PR

pandas/core/strings.py
```diff
@@ -941,17 +941,52 @@ def str_get_dummies(arr, sep='|'):
...
+    Series.str.split : Split strings around given separator/delimiter.
+
+    Examples
+    --------
+
+    Example with a list that contains non-string elements.
+
+    >>> s = pd.Series({1: ['lion', 'elephant', 'zebra'],
+    ...                2: [1.1, 2.2, 3.3],
+    ...                3: [np.nan, np.nan, np.nan]})
```
datapythonista: Minor comment, but I think it'd be more useful for users to see that joining `['cat', np.nan, 'dog']` becomes `NaN`, than that joining some `NaN`s become `NaN`. That could also apply in the number example `['cat', 23, 'dog']`. Feel free to disagree if you think it's more clear with unique types.

Also, it's something very subtle, but I find slightly distracting using a specific index in the example, that is not used. It surely doesn't make a big difference, but I'd construct the `Series` with a list of lists, and use the default index, so nobody thinks the index has an impact on joining the elements.

pandas/core/strings.py
```diff
@@ -941,17 +941,52 @@ def str_get_dummies(arr, sep='|'):
...
+    >>> s
+    1    [lion, elephant, zebra]
+    2            [1.1, 2.2, 3.3]
+    3            [nan, nan, nan]
+    dtype: object
+
+    Join all lists using an '-', the list of floats will become a NaN.
```
datapythonista: I think this comment needs to be updated after adding the `NaN`?

pandas/core/strings.py
```diff
@@ -941,17 +941,52 @@ def str_get_dummies(arr, sep='|'):
...
     ----------
-    sep : string
-        Delimiter
+    sep : str
+        Delimiter to use between list entries.
 
     Returns
     -------
-    joined : Series/Index of objects
+    Series/Index of objects
```
datapythonista: I'd use `object` instead of `objects`, as this is more a type definition than an explanations (may be one day we can use these types as annotations?)

datapythonista (commented):
Looks great, added couple of ideas.

fdroessler:
good points, thanks a lot 👍

datapythonista (approved):
lgtm

TomAugspurger:
Thanks @fdroessler !

PR #20465, opened 2018-03-23
DOC: update the pandas.DatetimeIndex docstring

robinkiplangat:
an unrelated PR):

- [ ] PR title is "DOC: update the pandas.DatetimeIndex docstring"
- [ ] The validation script passes: `scripts/validate_docstrings.py pandas.DatetimeIndex`
- [ ] The html version looks good: `python doc/make.py --single pandas.DatetimeIndex`

```
####################### Docstring (pandas.DatetimeIndex) #######################

Store Datetime index using NumPy's datetime64 dtype.

Immutable ndarray of datetime64 data, represented internally as int64, and
which can be boxed to Timestamp objects that are subclasses of datetime and
carry metadata such as frequency information.

Parameters
----------
data : array-like (1-dimensional), optional
    Optional datetime-like data to construct index with.
freq : str or pandas offset object, optional
    One of pandas date offset strings or corresponding objects.
start : datetime-like, optional
    Shapes the starting value and if data is None,
    start is used as the start point in generating regular timestamp data.
end : datetime-like, optional
    Shapes the end time, if periods is none, generated index will extend
    to first conforming time on or just past end argument.
copy : bool
    Make a copy of input ndarray.
name : object
    Name to be stored in the index.
tz : pytz.timezone or dateutil.tz.tzfile.
verify_integrity : bool, default False
    Check new index for duplicates, otherwise defer until necessary.
normalize : bool
    Return DatetimeIndex with times to midnight.
closed : str or None, default None
    Make the interval closed with respect to the given frequency to
    the 'left', 'right', or both sides (None).
ambiguous : 'infer', bool-ndarray, 'NaT', default 'raise'
    - 'infer' will attempt to infer fall dst-transition hours based on
      order.
    - bool-ndarray where True signifies a DST time, False signifies a
      non-DST time (note that this flag is only applicable for ambiguous
      times).
    - 'NaT' will return NaT where there are ambiguous times.
    - 'raise' will raise an AmbiguousTimeError if there are ambiguous times
dtype : object
    Specify the datatype to DataFrame constructor.
**kwargs
    For compatibility. Has no effect on the result.

Attributes
----------
year
month
day
hour
minute
second
microsecond
nanosecond
date
time
dayofyear
weekofyear
week
dayofweek
weekday
quarter
tz
freq
freqstr
is_month_start
is_month_end
is_quarter_start
is_quarter_end
is_year_start
is_year_end
is_leap_year
inferred_freq

Methods
-------
normalize
strftime
snap
tz_convert
tz_localize
round
floor
ceil
to_period
to_perioddelta
to_pydatetime
to_series
to_frame
month_name
day_name

Notes
-----
To learn more about the frequency strings, please see `this link
<http://pandas.pydata.org/pandas-docs/stable/timeseries.html#offset-aliases>`__.

See Also
---------
Index : The base pandas Index type
TimedeltaIndex : Index of timedelta64 data
PeriodIndex : Index of Period data

Examples
--------
>>> # Generate a date range with month end frequency
>>> pd.DatetimeIndex(start='2018-3-10', freq='M', periods=4)
DatetimeIndex(['2018-03-31', '2018-04-30',
               '2018-05-31', '2018-06-30'],
              dtype='datetime64[ns]', freq='M')

>>> # Hourly time range with end conforming the period
>>> pd.DatetimeIndex(start='2015-03-01 21:00:00', end='2015-03-02', freq='H')
DatetimeIndex(['2015-03-01 21:00:00', '2015-03-01 22:00:00',
               '2015-03-01 23:00:00', '2015-03-02 00:00:00'],
              dtype='datetime64[ns]', freq='H')

>>> # Generate time range with with 1 minute frequency
>>> pd.date_range('2018-03-10', periods=4, freq='1min')
DatetimeIndex(['2018-03-10 00:00:00', '2018-03-10 00:01:00',
               '2018-03-10 00:02:00', '2018-03-10 00:03:00'],
              dtype='datetime64[ns]', freq='T')

>>> # Weekly date range with closed argument to mark end date interval
>>> pd.date_range('2018-03-10', '2018-03-31', freq='W', closed='right')
DatetimeIndex(['2018-03-11', '2018-03-18', '2018-03-25'],
              dtype='datetime64[ns]', freq='W-SUN')

Errors found:
	Errors in parameters section
		Parameters {'periods', 'kwargs'} not documented
```

explicitly.


pandas/core/indexes/datetimes.py
```diff
@@ -262,6 +264,14 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
     month_name
     day_name
 
+    Examples
+    --------
+    >>> dates = pd.DatetimeIndex(freq='M', start='2018-3-10', periods=6)
+    >>> dates
+    DatetimeIndex(['2018-03-31', '2018-04-30', '2018-05-31',
+                   '2018-06-30', '2018-07-31', '2018-08-31'],
+                  dtype='datetime64[ns]', freq='M')
```
datapythonista: `Examples` should be the last section.

As a convention, we name indices with `idx`, I think this would be preferred over `dates`. In this case, as `dates` is not being used more than to display it, I wouldn't store the index in a variable but show it directly

The example looks great, but I think couple of more examples showing the contructor with `end`, `closed`, and may be a different frequency, could be useful.

pandas/core/indexes/datetimes.py
```diff
@@ -175,6 +175,8 @@ def _new_DatetimeIndex(cls, d):
 class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
                     Int64Index):
     """
+    Store timestamps using NumPy’s datetime64 dtype.
```
datapythonista: You can ignore the comment about starting with a verb in this case, as this is the docstring of a class, and not a method or function.

I think someting lke `Datetime index using NumPy's...` would be more accurate.

pandas/core/indexes/datetimes.py
```diff
@@ -175,6 +175,8 @@ def _new_DatetimeIndex(cls, d):
...
+
     Immutable ndarray of datetime64 data, represented internally as int64, and
     which can be boxed to Timestamp objects that are subclasses of datetime and
     carry metadata such as frequency information.
```
datapythonista: For the message in your validation error, I'd say the parameters are documented in the wrong order. It could be a problem in the script, but can you take a look and double check please?

pandas/core/indexes/datetimes.py
```diff
@@ -267,6 +269,33 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
...
+                  dtype='datetime64[ns]', freq='T')
+
+    >>> pd.date_range('2018-03-01', '2018-04-30', freq='W', closed='right')
+    DatetimeIndex(['2018-03-04', '2018-03-11', '2018-03-18',
+                   '2018-03-25', '2018-04-01', '2018-04-08',
+                   '2018-04-15', '2018-04-22', '2018-04-29'],
+                  dtype='datetime64[ns]', freq='W-SUN')
+
     See Also
     ---------
```
datapythonista: See also section should go before the examples, not after

datapythonista (changes requested):

pandas/core/indexes/datetimes.py
4bic: sure, I was to query as I looked over and found they were all in

4bic: how about `infer_dst` that is .. deprecated:: 0.15.0. ?

datapythonista: If it's not in the signature (I don't think it is), you can get rid of it.

datapythonista (commented):
Looking much better, added some comments.

jorisvandenbossche:
I added a small edit (added back the periods keyword explanation), and fixed the conflict with master (in the meantime, there had been done some changes to that docstring)

pandas/core/indexes/datetimes.py
```diff
@@ -274,6 +282,35 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
...
+    Hourly time range with end conforming the period
+
+    >>> pd.DatetimeIndex(start='2015-03-01 21:00:00', end='2015-03-02', freq='H')
+    DatetimeIndex(['2015-03-01 21:00:00', '2015-03-01 22:00:00',
+                   '2015-03-01 23:00:00', '2015-03-02 00:00:00'],
+                  dtype='datetime64[ns]', freq='H')
+
+    Generate time range with with 1 minute frequency
+
+    >>> pd.date_range('2018-03-10', periods=4, freq='1min')
```
jorisvandenbossche: You are actually using `date_range` here instead of `DatetimeIndex`

pandas/core/indexes/datetimes.py
```diff
@@ -274,6 +282,35 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
     TimedeltaIndex : Index of timedelta64 data
     PeriodIndex : Index of Period data
     pandas.to_datetime : Convert argument to datetime
+
+    Examples
+    --------
+    Generate a date range with month end frequency
+
+    >>> pd.DatetimeIndex(start='2018-3-10', freq='M', periods=4)
```
jorisvandenbossche: All the example are for generating range, but I think the more typical use case for DatetimeIndex is to create an index from an array-like. Can you add examples for that as well?

jorisvandenbossche: (and for creating ranges, I would actually just refer to `date_range`

jreback:
can you rebase and update for comments

jreback:

PR #20474, opened 2018-03-23
DOC: docstring to series.unique

### LINKED ISSUE #20075: DOC: Series.unique docstring can be moved to series.py
Note for the documentation sprint:

- The docstring of `Series.unique` is currently defined in `pandas/core/base.py`, but since the docstring is no longer shared but only used by `Series.unique`, the docstring can be moved to `pandas/core/series.py` as a normal docstring of the Series.unique definition

minggli:
- [x] closes #20075 
- [x] tests added / passed
- [ ] whatsnew entry

seems that no one has picked this up. moving _shared_doc['unique'] used by Series.unique() only to Series.unique.
```
####################### Docstring (pandas.Series.unique) #######################

Return unique values of Series object.

Uniques are returned in order of appearance. Hash table-based unique,
therefore does NOT sort.

Returns
-------
unique values.
  - If the input is an Index, the return is an Index
  - If the input is a Categorical dtype, the return is a Categorical
  - If the input is a Series/ndarray, the return will be an ndarray

See Also
--------
unique : return unique values of 1d array-like objects.
Index.unique : return Index with unique values from an Index object.

Examples
--------
>>> pd.Series([2, 1, 3, 3], name='A').unique()
array([2, 1, 3])

>>> pd.Series([2] + [1] * 5).unique()
array([2, 1])

>>> pd.Series([pd.Timestamp('20160101') for _ in range(3)]).unique()
array(['2016-01-01T00:00:00.000000000'], dtype='datetime64[ns]')

>>> pd.Series([pd.Timestamp('20160101', tz='US/Eastern')                        for _ in range(3)]).unique()
array([Timestamp('2016-01-01 00:00:00-0500', tz='US/Eastern')],
      dtype=object)

An unordered Categorical will return categories in the order of
appearance.

>>> pd.Series(pd.Categorical(list('baabc'))).unique()
[b, a, c]
Categories (3, object): [b, a, c]

An ordered Categorical preserves the category ordering.

>>> pd.Series(pd.Categorical(list('baabc'), categories=list('abc'),                                      ordered=True)).unique()
[b, a, c]
Categories (3, object): [a < b < c]

Docstring for "pandas.Series.unique" correct. :)
```

### DIFF
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -1438,7 +1438,7 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         Returns
         -------
-        unique values : ndarray or Categorical
+        ndarray or Categorical
             The unique values returned as a NumPy array. In case of categorical
             data type, returned as a Categorical.
 
@@ -1452,13 +1452,10 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         >>> pd.Series([2, 1, 3, 3], name='A').unique()
         array([2, 1, 3])
 
-        >>> pd.Series([2] + [1] * 5).unique()
-        array([2, 1])
-
-        >>> pd.Series([pd.Timestamp('20160101') for _ in range(3)]).unique()
+        >>> pd.Series([pd.Timestamp('2016-01-01') for _ in range(3)]).unique()
         array(['2016-01-01T00:00:00.000000000'], dtype='datetime64[ns]')
 
-        >>> pd.Series([pd.Timestamp('20160101', tz='US/Eastern')
+        >>> pd.Series([pd.Timestamp('2016-01-01', tz='US/Eastern')
         ...            for _ in range(3)]).unique()
         array([Timestamp('2016-01-01 00:00:00-0500', tz='US/Eastern')],
               dtype=object)
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -1459,8 +1459,8 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         >>> pd.Series([pd.Timestamp('20160101') for _ in range(3)]).unique()
         array(['2016-01-01T00:00:00.000000000'], dtype='datetime64[ns]')
 
-        >>> pd.Series([pd.Timestamp('20160101', tz='US/Eastern') \
-                       for _ in range(3)]).unique()
+        >>> pd.Series([pd.Timestamp('20160101', tz='US/Eastern')
+        ...            for _ in range(3)]).unique()
         array([Timestamp('2016-01-01 00:00:00-0500', tz='US/Eastern')],
               dtype=object)
 
@@ -1473,8 +1473,8 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         An ordered Categorical preserves the category ordering.
 
-        >>> pd.Series(pd.Categorical(list('baabc'), categories=list('abc'), \
-                                     ordered=True)).unique()
+        >>> pd.Series(pd.Categorical(list('baabc'), categories=list('abc'),
+        ...                          ordered=True)).unique()
         [b, a, c]
         Categories (3, object): [a < b < c]
         """
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -1445,8 +1445,8 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         See Also
         --------
-        unique
-        Index.unique
+        unique : return unique values of 1d array-like objects.
+        Index.unique : return Index with unique values from an Index object.
 
         Examples
         --------
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -1438,7 +1438,9 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         Returns
         -------
-        unique values : Series or Categorical
+        unique values : ndarray or Categorical
+            The unique values returned as a NumPy array. In case of categorical
+            data type, returned as a Categorical.
 
         See Also
         --------
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -1459,7 +1459,7 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         >>> pd.Series([pd.Timestamp('20160101') for _ in range(3)]).unique()
         array(['2016-01-01T00:00:00.000000000'], dtype='datetime64[ns]')
 
-        >>> pd.Series([pd.Timestamp('20160101', tz='US/Eastern')
+        >>> pd.Series([pd.Timestamp('20160101', tz='US/Eastern') \
                        for _ in range(3)]).unique()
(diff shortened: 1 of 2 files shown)


pandas/core/series.py
```diff
@@ -1429,8 +1429,28 @@ def mode(self):
         # TODO: Add option for bins like value_counts()
         return algorithms.mode(self)
 
-    @Appender(base._shared_docs['unique'] % _shared_doc_kwargs)
     def unique(self):
+        """
+        Return unique values in the object. Uniques are returned in order
```
TomAugspurger: This should be a single line. Can you simplify? You can use an extended summary for the bit about uniques.

pandas/core/series.py
```diff
@@ -1429,8 +1429,28 @@ def mode(self):
...
+        of appearance, this does NOT sort. Hash table-based unique.
```
TomAugspurger: I'm not sure hash table-based unique will be meaningful to many users.

pandas/core/series.py
```diff
@@ -1429,8 +1429,28 @@ def mode(self):
...
+          - If the input is an Index, the return is an Index
+          - If the input is a Categorical dtype, the return is a Categorical
+          - If the input is a Series/ndarray, the return will be an ndarray
+
+        See Also
+        --------
+        unique
+        Index.unique
+        Series.unique
+        """
```
TomAugspurger: Can you add examples?

pandas/core/series.py
minggli: agree but that's the same in pandas.unique. I think it makes sense to people who understand data structure, and for people who don't understand it there is enough clarification around it. It's important to let people to know it's hash table so no sorting.

pandas/core/series.py
minggli: aded. and removed self reference in See Also and added descriptions.

pandas/core/series.py
minggli: broken into two lines. and minor rewording.

minggli:
@TomAugspurger comments welcome 👍

jreback (changes requested):

pandas/core/series.py
```diff
@@ -1429,8 +1429,55 @@ def mode(self):
         # TODO: Add option for bins like value_counts()
         return algorithms.mode(self)
 
-    @Appender(base._shared_docs['unique'] % _shared_doc_kwargs)
     def unique(self):
+        """
```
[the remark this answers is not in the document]
minggli: good point but they have some differences:
* pd.unique takes param and Series.unique doesn't take.
* pd.unique handles 1d array-like objects including Index and Series.unique applies on self.
* pd.unique examples contain more and Series.unique only series examples

do we have pattern somewhere in regards to conditionally show docstring lines?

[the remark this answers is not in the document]
TomAugspurger: https://github.com/pandas-dev/pandas/pull/20361/files is doing something similar for factorize. It's somewhat complex, since we have `pd.unique`, `Series/Index.unique`, and `Categorical.unique`. I'd be OK with improving the docstring here, and merging theme later.

pandas/core/series.py
```diff
@@ -1429,8 +1429,55 @@ def mode(self):
...
+        Returns
+        -------
+        unique values.
+          - If the input is an Index, the return is an Index
+          - If the input is a Categorical dtype, the return is a Categorical
+          - If the input is a Series/ndarray, the return will be an ndarray
+
+        See Also
+        --------
+        unique : return unique values of 1d array-like objects.
```
TomAugspurger: Can you include `pandas.` for this one. Otherwise it's unclear what's being referenced. The explanation can say "Top-level unique method for any 1-d array-like object."

pandas/core/series.py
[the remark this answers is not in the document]
minggli: agree, this PR is mainly about improve docstring. and should be another PR synthesising docs of all unique methods.

pandas/core/series.py
minggli: done

pandas/core/series.py
```diff
@@ -1429,8 +1429,55 @@ def mode(self):
...
+        An unordered Categorical will return categories in the order of
+        appearance.
+
+        >>> pd.Series(pd.Categorical(list('baabc'))).unique()
+        [b, a, c]
+        Categories (3, object): [b, a, c]
+
+        An ordered Categorical preserves the category ordering.
+
+        >>> pd.Series(pd.Categorical(list('baabc'), categories=list('abc'), \
```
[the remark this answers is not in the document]
minggli: done.

pandas/core/series.py
```diff
@@ -1429,8 +1429,55 @@ def mode(self):
...
+        >>> pd.Series([2, 1, 3, 3], name='A').unique()
+        array([2, 1, 3])
+
+        >>> pd.Series([2] + [1] * 5).unique()
+        array([2, 1])
+
+        >>> pd.Series([pd.Timestamp('20160101') for _ in range(3)]).unique()
+        array(['2016-01-01T00:00:00.000000000'], dtype='datetime64[ns]')
+
+        >>> pd.Series([pd.Timestamp('20160101', tz='US/Eastern') \
```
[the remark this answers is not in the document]
minggli: done.

pandas/core/series.py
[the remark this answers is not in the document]
jreback: that's fine too

pandas/core/series.py
```diff
@@ -1429,8 +1429,55 @@ def mode(self):
...
+        """
+        Return unique values of Series object.
+
+        Uniques are returned in order of appearance. Hash table-based unique,
+        therefore does NOT sort.
+
+        Returns
+        -------
+        unique values.
+          - If the input is an Index, the return is an Index
```
jorisvandenbossche: This part can be simplified, since it is now only the docstring of Series.unique. 
So there is also no "input"

In this case it is always an `array`, except for categorical dtype.

minggli: done.

minggli:
@jorisvandenbossche does it look good?

pandas/core/series.py
```diff
@@ -1429,8 +1429,52 @@ def mode(self):
...
     def unique(self):
+        """
+        Return unique values of Series object.
+
+        Uniques are returned in order of appearance. Hash table-based unique,
+        therefore does NOT sort.
+
+        Returns
+        -------
+        unique values : Series or Categorical
```
jorisvandenbossche: It's never a Series, only an numpy array or Categorical. I would also include something about that like

```
ndarray or Categorical
    The unique values returned as a NumPy array. In case of categorical data type, returned as a Categorical
```

jorisvandenbossche (approved):

jorisvandenbossche:
@minggli Thanks a lot!

minggli:
Thanks @jorisvandenbossche

PR #17774, opened 2017-10-04
DEPR: Deprecate parse_cols in read_excel

gfyoung:
Will now use "usecols" just like in read_csv.
closes #4988.

PR #20480, opened 2018-03-25
BUG: read_excel return empty dataframe when using usecols

### LINKED ISSUE #18273: read_excel return empty dataframe when using usecols
```python
In [3]: data = pd.read_excel("A.xlsx")

In [4]: data
Out[4]:
   A  B
0  1  2
1  3  4

In [5]: data1 = pd.read_excel("A.xlsx",usecols=['B'])

In [6]: data1
Out[6]:
Empty DataFrame
Columns: []
Index: []

In [7]: pd.__version__
Out[7]: '0.21.0'

```
#### Problem description
Having a excel file name A.xlsx(or A.xls) with column A,B
read_excel return empty dataframe if usecols used

jacksonjos:
- [x] closes #18273
- [x] tests added / passed
- [x] passes git diff master --name-only -- "*.py" | grep "pandas/" | xargs -r flake8
- [x] whatsnew entry

As mentioned read_excel returns an empty DataFrame when usecols argument is a list of strings.
Now lists of strings are correctly interpreted by read_excel function.

### DIFF
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -2852,23 +2852,46 @@ Parsing Specific Columns
 
 It is often the case that users will insert columns to do temporary computations
 in Excel and you may not want to read in those columns. ``read_excel`` takes
-a ``usecols`` keyword to allow you to specify a subset of columns to parse.
+either a ``usecols`` or ``usecols_excel`` keyword to allow you to specify a
+subset of columns to parse. Note that you can not use both ``usecols`` and
+``usecols_excel`` named arguments at the same time.
 
-If ``usecols`` is an integer, then it is assumed to indicate the last column
-to be parsed.
-
-.. code-block:: python
-
-   read_excel('path_to_file.xls', 'Sheet1', usecols=2)
-
-If `usecols` is a list of integers, then it is assumed to be the file column
-indices to be parsed.
+If ``usecols_excel`` is supplied, then it is assumed that indicates a comma
+separated list of Excel column letters and column ranges to be parsed.
 
 .. code-block:: python
 
-   read_excel('path_to_file.xls', 'Sheet1', usecols=[0, 2, 3])
+   read_excel('path_to_file.xls', 'Sheet1', usecols_excel='A:E')
+   read_excel('path_to_file.xls', 'Sheet1', usecols_excel='A,C,E:F')
 
-Element order is ignored, so ``usecols=[0, 1]`` is the same as ``[1, 0]``.
+If ``usecols`` is a list of integers, then it is assumed to be the file
+column indices to be parsed.
+
+.. code-block:: python
+
+   read_excel('path_to_file.xls', 'Sheet1', usecols=[1, 3, 5])
+
+Element order is ignored, so ``usecols_excel=[0, 1]`` is the same as ``[1, 0]``.
+
+If ``usecols`` is a list of strings, then it is assumed that each string
+correspond to column names provided either by the user in `names` or
+inferred from the document header row(s) and those strings define which columns
+will be parsed.
+
+.. code-block:: python
+
+   read_excel('path_to_file.xls', 'Sheet1', usecols=['foo', 'bar'])
+
+Element order is ignored, so ``usecols=['baz', 'joe']`` is the same as
+``['joe', 'baz']``.
+
+If ``usecols`` is callable, the callable function will be evaluated against the
+column names, returning names where the callable function evaluates to True.
+
+.. code-block:: python
+
+   read_excel('path_to_file.xls', 'Sheet1', usecols=lambda x: x.isalpha())
+
 
 Parsing Dates
 +++++++++++++
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -36,7 +36,7 @@ Datetimelike API Changes
 Other API Changes
 ^^^^^^^^^^^^^^^^^
 
--
+- :func:`read_excel` has gained the keyword argument ``usecols_excel`` that receives a string containing comma separated Excel ranges and columns. The ``usecols`` keyword argument at :func:`read_excel` had removed support for a string containing comma separated Excel ranges and columns and for an int indicating the first j columns to be read in a ``DataFrame``. Also, the ``usecols`` keyword argument at :func:`read_excel` had added support for receiving a list of strings containing column labels and a callable. (:issue:`18273`)
 -
 -
 
@@ -148,7 +148,7 @@ I/O
 ^^^
 
 -
--
+- Bug in :func:`read_excel` where ``usecols`` keyword argument as a list of strings were returning a empty ``DataFrame`` (:issue:`18273`)
 -
 
 Plotting
--- a/pandas/io/excel.py
+++ b/pandas/io/excel.py
@@ -10,6 +10,8 @@ import os
 import abc
 import warnings
 import numpy as np
+import string
+import re
 from io import UnsupportedOperation
 
 from pandas.core.dtypes.common import (
@@ -85,20 +87,42 @@ index_col : int, list of ints, default None
     Column (0-indexed) to use as the row labels of the DataFrame.
     Pass None if there is no such column.  If a list is passed,
     those columns will be combined into a ``MultiIndex``.  If a
-    subset of data is selected with ``usecols``, index_col
-    is based on the subset.
+    subset of data is selected with ``usecols_excel`` or ``usecols``,
+    index_col is based on the subset.
 parse_cols : int or list, default None
 
     .. deprecated:: 0.21.0
        Pass in `usecols` instead.
 
-usecols : int or list, default None
+usecols : list-like or callable, default None
+    Return a subset of the columns. If list-like, all elements must either
+    be positional (i.e. integer indices into the document columns) or string
+    that correspond to column names provided either by the user in `names` or
+    inferred from the document header row(s). For example, a valid list-like
+    `usecols` parameter would be [0, 1, 2] or ['foo', 'bar', 'baz']. Note that
+    you can not give both ``usecols`` and ``usecols_excel`` keyword arguments
+    at the same time.
+
+    If callable, the callable function will be evaluated against the column
+    names, returning names where the callable function evaluates to True. An
+    example of a valid callable argument would be ``lambda x: x.upper() in
+    ['AAA', 'BBB', 'DDD']``.
+
+    .. versionadded:: 0.24.0
+    Added support to column labels and now `usecols_excel` is the keyword that
+    receives separated comma list of excel columns and ranges.
+usecols_excel : string, default None
+    Return a subset of the columns from a spreadsheet specified as Excel column
+    ranges and columns. Note that you can not use both ``usecols`` and
+    ``usecols_excel`` keyword arguments at the same time.
+
     * If None then parse all columns,
-    * If int then indicates last column to be parsed
-    * If list of ints then indicates list of column numbers to be parsed
     * If string then indicates comma separated list of Excel column letters and
-      column ranges (e.g. "A:E" or "A,C,E:F").  Ranges are inclusive of
-      both sides.
+      column ranges (e.g. "A:E" or "A,C,E:F") to be parsed. Ranges are
+      inclusive of both sides.
+
+    .. versionadded:: 0.24.0
+
 squeeze : boolean, default False
     If the parsed data only contains one column then return a Series
 dtype : Type name or dict of column -> type, default None
@@ -269,6 +293,19 @@ def _get_default_writer(ext):
     return _default_writers[ext]
 
 
+def _is_excel_columns_notation(columns):
+    """
+    Receives a string and check if the string is a comma separated list of
+    Excel index columns and index ranges. An Excel range is a string with two
+    column indexes separated by ':').
+    """
+    if isinstance(columns, compat.string_types) and all(
+       (x in string.ascii_letters) for x in re.split(r',|:', columns)):
+        return True
+
+    return False
+
+
 def get_writer(engine_name):
     try:
         return _writers[engine_name]
@@ -286,6 +323,7 @@ def read_excel(io,
                names=None,
                index_col=None,
                usecols=None,
+               usecols_excel=None,
                squeeze=False,
                dtype=None,
                engine=None,
@@ -311,6 +349,7 @@ def read_excel(io,
         header=header,
         names=names,
         index_col=index_col,
+        usecols_excel=usecols_excel,
         usecols=usecols,
         squeeze=squeeze,
         dtype=dtype,
@@ -405,6 +444,7 @@ class ExcelFile(object):
               names=None,
               index_col=None,
               usecols=None,
+              usecols_excel=None,
               squeeze=False,
               converters=None,
               true_values=None,
@@ -439,6 +479,7 @@ class ExcelFile(object):
                                  header=header,
                                  names=names,
                                  index_col=index_col,
+                                 usecols_excel=usecols_excel,
                                  usecols=usecols,
                                  squeeze=squeeze,
                                  converters=converters,
@@ -455,7 +496,7 @@ class ExcelFile(object):
                                  convert_float=convert_float,
                                  **kwds)
 
-    def _should_parse(self, i, usecols):
+    def _should_parse(self, i, usecols_excel, usecols):
 
         def _range2cols(areas):
             """
@@ -481,12 +522,12 @@ class ExcelFile(object):
                     cols.append(_excel2num(rng))
             return cols
 
-        if isinstance(usecols, int):
-            return i <= usecols
-        elif isinstance(usecols, compat.string_types):
-            return i in _range2cols(usecols)
-        else:
-            return i in usecols
+        # check if usecols_excel is a string that indicates a comma separated
+        # list of Excel column letters and column ranges
+        if isinstance(usecols_excel, compat.string_types):
+            return i in _range2cols(usecols_excel)
+
+        return True
 
     def _parse_excel(self,
                      sheet_name=0,
@@ -494,6 +535,7 @@ class ExcelFile(object):
                      names=None,
                      index_col=None,
                      usecols=None,
+                     usecols_excel=None,
                      squeeze=False,
                      dtype=None,
                      true_values=None,
@@ -511,6 +553,25 @@ class ExcelFile(object):
                      **kwds):
 
         _validate_header_arg(header)
+
+        if (usecols is not None) and (usecols_excel is not None):
+            raise ValueError("Cannot specify both `usecols` and "
+                             "`usecols_excel`. Choose one of them.")
+
+        # Check if some string in usecols may be interpreted as a Excel
+        # range or positional column
+        elif _is_excel_columns_notation(usecols):
+            warnings.warn("The `usecols` keyword argument used to refer to "
+                          "Excel ranges and columns as strings was "
+                          "renamed to `usecols_excel`.", UserWarning,
+                          stacklevel=3)
+            usecols_excel = usecols
+            usecols = None
+
+        elif (usecols_excel is not None) and not _is_excel_columns_notation(
+                usecols_excel):
+            raise TypeError("`usecols_excel` must be None or a string as a "
+                            "comma separeted Excel ranges and columns.")
 
         if 'chunksize' in kwds:
             raise NotImplementedError("chunksize keyword of read_excel "
@@ -615,10 +676,13 @@ class ExcelFile(object):
                 row = []
                 for j, (value, typ) in enumerate(zip(sheet.row_values(i),
                                                      sheet.row_types(i))):
-                    if usecols is not None and j not in should_parse:
-                        should_parse[j] = self._should_parse(j, usecols)
+                    if ((usecols is not None) or (usecols_excel is not None) or
+                            (j not in should_parse)):
+                        should_parse[j] = self._should_parse(j, usecols_excel,
+                                                             usecols)
 
-                    if usecols is None or should_parse[j]:
+                    if (((usecols_excel is None) and (usecols is None)) or
+                            should_parse[j]):
                         row.append(_parse_cell(value, typ))
                 data.append(row)
 
@@ -674,6 +738,7 @@ class ExcelFile(object):
                                     dtype=dtype,
                                     true_values=true_values,
                                     false_values=false_values,
+                                    usecols=usecols,
                                     skiprows=skiprows,
                                     nrows=nrows,
                                     na_values=na_values,
--- a/pandas/tests/io/test_excel.py
+++ b/pandas/tests/io/test_excel.py
@@ -107,54 +107,43 @@ class SharedItems(object):
 class ReadingTestsBase(SharedItems):
     # This is based on ExcelWriterBase
 
-    def test_usecols_int(self, ext):
-
-        dfref = self.get_csv_refdf('test1')
-        dfref = dfref.reindex(columns=['A', 'B', 'C'])
-        df1 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0, usecols=3)
-        df2 = self.get_exceldf('test1', ext, 'Sheet2', skiprows=[1],
-                               index_col=0, usecols=3)
-
-        with tm.assert_produces_warning(FutureWarning):
-            df3 = self.get_exceldf('test1', ext, 'Sheet2', skiprows=[1],
-                                   index_col=0, parse_cols=3)
-
-        # TODO add index to xls file)
-        tm.assert_frame_equal(df1, dfref, check_names=False)
-        tm.assert_frame_equal(df2, dfref, check_names=False)
-        tm.assert_frame_equal(df3, dfref, check_names=False)
-
     def test_usecols_list(self, ext):
 
         dfref = self.get_csv_refdf('test1')
         dfref = dfref.reindex(columns=['B', 'C'])
         df1 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
-                               usecols=[0, 2, 3])
+                               usecols=[1, 2])
         df2 = self.get_exceldf('test1', ext, 'Sheet2', skiprows=[1],
-                               index_col=0, usecols=[0, 2, 3])
+                               index_col=0, usecols=[1, 2])
 
         with tm.assert_produces_warning(FutureWarning):
             df3 = self.get_exceldf('test1', ext, 'Sheet2', skiprows=[1],
-                                   index_col=0, parse_cols=[0, 2, 3])
+                                   index_col=0, parse_cols=[1, 2])
 
         # TODO add index to xls file)
         tm.assert_frame_equal(df1, dfref, check_names=False)
         tm.assert_frame_equal(df2, dfref, check_names=False)
         tm.assert_frame_equal(df3, dfref, check_names=False)
 
-    def test_usecols_str(self, ext):
+    def test_usecols_excel_str(self, ext):
 
         dfref = self.get_csv_refdf('test1')
 
         df1 = dfref.reindex(columns=['A', 'B', 'C'])
         df2 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
-                               usecols='A:D')
+                               usecols_excel='A:D')
         df3 = self.get_exceldf('test1', ext, 'Sheet2', skiprows=[1],
-                               index_col=0, usecols='A:D')
+                               index_col=0, usecols_excel='A:D')
 
-        with tm.assert_produces_warning(FutureWarning):
+        # The following code receives two warnings because FutureWarning is
+        # thrown when parse_cols is passed in read_excel and UserWarning is
+        # thrown when parse_cols (usecols) receives an comma separated list of
+        # Excel indexes and ranges
+        with tm.assert_produces_warning() as w:
             df4 = self.get_exceldf('test1', ext, 'Sheet2', skiprows=[1],
                                    index_col=0, parse_cols='A:D')
+            assert issubclass(w[0].category, FutureWarning)
+            assert issubclass(w[1].category, UserWarning)
 
         # TODO add index to xls, read xls ignores index name ?
         tm.assert_frame_equal(df2, df1, check_names=False)
@@ -163,20 +152,106 @@ class ReadingTestsBase(SharedItems):
 
         df1 = dfref.reindex(columns=['B', 'C'])
         df2 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
-                               usecols='A,C,D')
+                               usecols_excel='A,C,D')
         df3 = self.get_exceldf('test1', ext, 'Sheet2', skiprows=[1],
-                               index_col=0, usecols='A,C,D')
+                               index_col=0, usecols_excel='A,C,D')
         # TODO add index to xls file
         tm.assert_frame_equal(df2, df1, check_names=False)
         tm.assert_frame_equal(df3, df1, check_names=False)
 
         df1 = dfref.reindex(columns=['B', 'C'])
         df2 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
-                               usecols='A,C:D')
+                               usecols_excel='A,C:D')
         df3 = self.get_exceldf('test1', ext, 'Sheet2', skiprows=[1],
-                               index_col=0, usecols='A,C:D')
+                               index_col=0, usecols_excel='A,C:D')
         tm.assert_frame_equal(df2, df1, check_names=False)
         tm.assert_frame_equal(df3, df1, check_names=False)
+
+    def test_usecols_diff_positional_int_columns_order(self, ext):
+
+        df1 = self.get_csv_refdf('test1')[['A', 'C']]
+
+        df2 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
+                               usecols=[0, 2])
+        df3 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
+                               usecols=[2, 0])
+
+        tm.assert_frame_equal(df2, df1, check_names=False)
+        tm.assert_frame_equal(df3, df2, check_names=False)
+
+    def test_usecols_diff_positional_str_columns_order(self, ext):
+
+        df1 = self.get_csv_refdf('test1')[['B', 'D']]
+
+        df2 = self.get_exceldf('test1', ext, 'Sheet1', usecols=['B', 'D'])
+        df3 = self.get_exceldf('test1', ext, 'Sheet1', usecols=['D', 'B'])
+
+        tm.assert_frame_equal(df2, df1, check_names=False)
+        tm.assert_frame_equal(df3, df1, check_names=False)
+
+    def test_read_excel_without_slicing(self, ext):
+
+        df1 = self.get_csv_refdf('test1')
+        df2 = self.get_exceldf('test1', ext, 'Sheet1')
+
+        tm.assert_frame_equal(df2, df1, check_names=False)
+
+    def test_pass_callable_argument(self, ext):
+
+        dfref = self.get_csv_refdf('test1')[['C', 'D']]
+
+        df1 = dfref.reindex(columns=['C', 'D'])
+        df2 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
+                               usecols=lambda x: x > 'B')
+        df3 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
+                               usecols_excel='A,D:E')
+
+        tm.assert_frame_equal(df2, df1, check_names=False)
+        tm.assert_frame_equal(df3, df1, check_names=False)
+
+    def test_usecols_deprecated_excel_range_str(self, ext):
+
+        dfref = self.get_csv_refdf('test1')[['B', 'C', 'D']]
+
+        df1 = dfref.reindex(columns=['C', 'D'])
+        df2 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
+                               usecols=['C', 'D'])
+        with tm.assert_produces_warning(UserWarning):
+            df3 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
+                                   usecols='A,D:E')
+
+        tm.assert_frame_equal(df2, df1, check_names=False)
+        tm.assert_frame_equal(df3, df1, check_names=False)
+
+    def test_index_col_label_error(self, ext):
+        msg = "list indices must be integers.*, not str"
+        with tm.assert_raises_regex(TypeError, msg):
+            self.get_exceldf('test1', ext, 'Sheet1', index_col=["A"],
+                             usecols=["", "A", "C"])
+
+    def test_pass_non_existent_column(self, ext):
+        msg = "Usecols do not match columns, columns expected but not found: "
+        "['E']"
+        with tm.assert_raises_regex(ValueError, msg):
+            self.get_exceldf('test1', ext, usecols=['E'])
+
+    def test_usecols_excel_wrong_type(self, ext):
+        msg = "`usecols_excel` must be None or a string as a comma separeted "
+        " Excel ranges and columns."
+        with tm.assert_raises_regex(TypeError, msg):
+            self.get_exceldf('test1', ext, usecols_excel=1)
+
+    def test_usecols_wrong_type(self, ext):
+        msg = "'usecols' must either be list-like of all strings, all unicode,"
+        " all integers or a callable."
+        with tm.assert_raises_regex(ValueError, msg):
+            self.get_exceldf('test1', ext, usecols='E1')
+
+    def test_usecols_and_usecols_excel_error(self, ext):
+        msg = "Cannot specify both `usecols` and `usecols_excel`. Choose one"
+        " of them."
+        with tm.assert_raises_regex(ValueError, msg):
+            self.get_exceldf('test1', ext, usecols=[0, 2], usecols_excel="A:C")
 
     def test_excel_stop_iterator(self, ext):
 
@@ -422,7 +497,8 @@ class ReadingTestsBase(SharedItems):
                 path,
                 'no_header',
                 usecols=[0],
-                header=None
+                header=None,
+                nrows=0
             )
 
             actual_header_zero = read_excel(
@@ -431,9 +507,10 @@ class ReadingTestsBase(SharedItems):
                 usecols=[0],
                 header=0
             )
-        expected = DataFrame()
-        tm.assert_frame_equal(actual_header_none, expected)
-        tm.assert_frame_equal(actual_header_zero, expected)
+        expected_header_none = DataFrame(columns=[0])
+        tm.assert_frame_equal(actual_header_none, expected_header_none)
+        expected_header_zero = DataFrame({1: [2, 3, 4]}, index=3 * [np.nan])
+        tm.assert_frame_equal(actual_header_zero, expected_header_zero)
 
     @td.skip_if_no('openpyxl')
     @td.skip_if_no('xlwt')
@@ -450,7 +527,8 @@ class ReadingTestsBase(SharedItems):
                 path,
                 'with_header',
                 usecols=[0],
-                header=None
+                header=None,
+                nrows=1
             )
 
             actual_header_zero = read_excel(
@@ -461,7 +539,7 @@ class ReadingTestsBase(SharedItems):
             )
         expected_header_none = DataFrame(pd.Series([0], dtype='int64'))
         tm.assert_frame_equal(actual_header_none, expected_header_none)
-        expected_header_zero = DataFrame(columns=[0])
+        expected_header_zero = DataFrame(pd.Series(4 * [np.nan]))
         tm.assert_frame_equal(actual_header_zero, expected_header_zero)
 
     @td.skip_if_no('openpyxl')
@@ -503,34 +581,19 @@ class ReadingTestsBase(SharedItems):
         # GH10559: Minor improvement: Change "sheet_name" to "sheetname"
         # GH10969: DOC: Consistent var names (sheetname vs sheet_name)
         # GH12604: CLN GH10559 Rename sheetname variable to sheet_name
-        # GH20920: ExcelFile.parse() and pd.read_xlsx() have different
-        #          behavior for "sheetname" argument
         dfref = self.get_csv_refdf('test1')
-        df1 = self.get_exceldf('test1', ext,
-                               sheet_name='Sheet1')  # doc
+        df1 = self.get_exceldf('test1', ext, sheet_name='Sheet1')    # doc
         with tm.assert_produces_warning(FutureWarning, check_stacklevel=False):
             df2 = self.get_exceldf('test1', ext,
                                    sheetname='Sheet1')  # bkwrd compat
 
-        excel = self.get_excelfile('test1', ext)
-        df1_parse = excel.parse(sheet_name='Sheet1')    # doc
-        with tm.assert_produces_warning(FutureWarning, check_stacklevel=False):
-            df2_parse = excel.parse(sheetname='Sheet1')  # bkwrd compat
-
         tm.assert_frame_equal(df1, dfref, check_names=False)
         tm.assert_frame_equal(df2, dfref, check_names=False)
-        tm.assert_frame_equal(df1_parse, dfref, check_names=False)
-        tm.assert_frame_equal(df2_parse, dfref, check_names=False)
 
     def test_sheet_name_both_raises(self, ext):
         with tm.assert_raises_regex(TypeError, "Cannot specify both"):
             self.get_exceldf('test1', ext, sheetname='Sheet1',
                              sheet_name='Sheet1')
-
-        excel = self.get_excelfile('test1', ext)
-        with tm.assert_raises_regex(TypeError, "Cannot specify both"):
-            excel.parse(sheetname='Sheet1',
-                        sheet_name='Sheet1')
 
 
 @pytest.mark.parametrize("ext", ['.xls', '.xlsx', '.xlsm'])


pandas/io/excel.py
```diff
@@ -479,6 +482,9 @@ def _excel2num(x):
             return i <= usecols
         elif isinstance(usecols, compat.string_types):
             return i in _range2cols(usecols)
+        elif all(isinstance(x, compat.string_types) for x in usecols) is True:
```
jreback: you don't need the is True

pandas/io/excel.py
```diff
@@ -479,6 +482,9 @@ def _excel2num(x):
...
+            usecols_str = ",".join(usecols)
```
jreback: can you add a 1-line comment on this case (and the case above to differentiate)

pandas/tests/io/test_excel.py
```diff
@@ -179,6 +179,42 @@ def test_usecols_str(self, ext):
         tm.assert_frame_equal(df2, df1, check_names=False)
         tm.assert_frame_equal(df3, df1, check_names=False)
 
+    def test_usecols_str_list(self, ext):
+
```
jreback: can you add the gh issue as a comment

pandas/tests/io/test_excel.py
```diff
@@ -179,6 +179,42 @@ def test_usecols_str(self, ext):
...
+        dfref = self.get_csv_refdf('test1')
+
```
jreback: can you parameterize this test

jreback (changes requested):

pandas/tests/io/test_excel.py
jacksonjos: Hi, thank you for yur feedback! =)

I'm not sure I've got it right. Do you want me to parametrize the objects passed as argument to parse_cols argument to each inside of test_usecols_str_list test because its tests are a copy of test_usecols_str or you want me to parametrize something else?

jreback: you are testing multiple cases that are pretty similar. the test becomes much simpler if you can parameterize over the possible cases.

jacksonjos: Ok, I got it. Do you want me to do the same to the test above (test_usecols_str)?

chris-b1:
This looks OK, but re-reading it, #18273 is actually two somewhat separate problems and we have a bit of an API tangle here

1.  passing a list of Excel column _letters_ doesn't work (this fixes that)
2.  passing a list column _names_ (e.g. column title `'foo'` in the spreadsheet) to `usecols` doesn't work anymore

Number 2 wasn't part of the `read_excel` documented args, but once worked, because this param used to be called `parse_cols` (#17774), and if you passed something to `usecols` it would get passed down to the TextParser logic and do the right thing
https://github.com/pandas-dev/pandas/blob/1915ffc53ea60494f24d83844bbff00efa392c82/pandas/io/excel.py#L529

Not sure what the solution is, there could be ambiguous cases (column titled`'A'` in spreadsheet column `B`), that would be easiest to handle with separate kwargs.  Or could just have usecols do everything and warn in cases of ambiguity.

Not saying you need to address item 2 in this PR, just need a new issue if not.

jacksonjos:
Hi @chris-b1, thank you for your comment.

I tried to test what you said, but I don't get it.

```markdown
| Day      | Number | Animal |
|----------|--------|--------|
| 10/12/18 | 5      | Tatu   |
| 10/13/18 | 4      | Paca   |
```
I tried the following:
```
df_not_ok = pd.read_excel("mock.xlsx", ["Day"])
>> XLRDError: No sheet named <'Day'>

df_not_ok = pd.read_excel("mock.xlsx", usecols=["Day"])
>> Empty DataFrame
>> Columns: []
>> Index: []
```

```
pd.show_versions()

INSTALLED VERSIONS
------------------
commit: None
python: 3.5.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.9.0-6-amd64
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.22.0
pytest: None
pip: 9.0.1
setuptools: 39.0.1
Cython: None
numpy: 1.14.2
scipy: None
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: None
patsy: None
dateutil: 2.7.1
pytz: 2018.3
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: 1.1.0
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: 0.999999999
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: None
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
```

If someone create feature or issue I may can work to make possible to pass col names as arguments to read just them.

chris-b1:
Right, that doesn't work on `0.22`, but it used to, which is what #18273 is about
```python
print(pd.__version__)
# 0.20.3
df = pd.DataFrame({'foo': [1, 2, 3], 'bar': [4, 5, 6]})
df.to_excel('tmp.xlsx', index=False)
pd.read_excel('tmp.xlsx', usecols=['bar'])
   bar
0    4
1    5
2    6
```

jacksonjos:
Ok, I got it, @chris-b1. It isn't better to create another issue to implement the possibility to choose columns you want to load by column name using another named argument? I think that might be confusing use `usecols` to a lot of different possibilities.

pandas/tests/io/test_excel.py
jacksonjos: Hi @jreback. I implemented all requested changes in the code.
I would like to know your opinion about a suggestion made by @chris-b1 at https://github.com/pandas-dev/pandas/pull/20480#issuecomment-375986768.

jacksonjos:
Is there anything wrong with the requested changes that I implemented to make the pull requested fulfill the requirements to be accepted? I'm asking because I don't understand why this pull requested it wasn't accepted yet.

pandas/io/excel.py
```diff
@@ -96,7 +96,10 @@
     * If int then indicates last column to be parsed
     * If list of ints then indicates list of column numbers to be parsed
     * If string then indicates comma separated list of Excel column letters and
-      column ranges (e.g. "A:E" or "A,C,E:F").  Ranges are inclusive of
+      column ranges (e.g. "A:E" or "A,C,E:F") to be parsed. Ranges are
+      inclusive of both sides.
+    * If list of strings each string shall be a Excel column letter or column
```
jreback: a -> an

jreback (changes requested):
waiting for @chris-b1 to have a look for this. I am concerned that usecols is now ambiguous, yes?

chris-b1:
Yeah, I've been a bit stuck on what to do here.  Current (0.22) behavior where `usecols` refers only to excel names for regions is bad, it broke with (untested/undocumented behavior but working!) 0.20.3 and is inconsistent with, e.g. `read_csv`.

I suppose the only reasonable solution is two keywords.  `usecols_labels` and `usecols_?`

- `usecols_excel`
- `usecols_a1`  (very Excel specific, but "A1 notation" is a somewhat common term)
- `excel_columns`

jacksonjos:
Hi,

I prefer to use `usecols_lables` and `usecols_excel` because it's possible refer columns by A1 notation and by numbers. If we decide to use `usecols_a1` the named argument would cause misunderstanding about the arguments can be passed to it.

What do you think? @chris-b1 and @jreback ?

jreback:
``usecols`` should refer to names just like it does in all other parsers. no problem adding ``usecols_a1`` (or similar).

chris-b1:
OK, @jacksonjos here's what I propose.  Most painful part will be backwards compat & testing.  If you don't want to take it all the way, just let me know and I'll finish, would like to get this in for 0.23.

1.  Add a new kwarg `usecols_excel` which replicates the current `usecols` + your fix here
2.  Since we're breaking compat, inspect `usecols` and if it is string AND parses to an excel region, issue a warning that the arg was renamed.  This could raise a false positive for a single letter (e.g. `'A'`) but don't see a way around it.
3.  `usecols` should get passed to the `TextParser` object, which will use it in the manner @jreback suggests.
https://github.com/pandas-dev/pandas/blob/1915ffc53ea60494f24d83844bbff00efa392c82/pandas/io/excel.py#L529

jacksonjos:
Hi @chris-b1 and @jreback.

Thank you for your answers. i'm already working on this issue and I'll
update the pull request in the coming days following your suggestions.
If I have some trouble I ask for help.

Best regards

2018-04-16 16:56 GMT-03:00 chris-b1 <notifications@github.com>:

jacksonjos:
Hi @jreback and @chris-b1, I'm about to finish the new version of this pull request and I would like to know your opinion about what to do when someone pass both arguments, usecols_excel and usecols, to read_excel function?

Should I raise an exception, print some warning or do nothing?

Thanks in advance.

pandas/io/excel.py
```diff
@@ -85,19 +85,41 @@
...
+    in ``['foo', 'bar']`` order or
+    ``pd.read_excel(data, usecols=['foo', 'bar'])[['bar', 'foo']]``
+    for ``['bar', 'foo']`` order.
+
+    If callable, the callable function will be evaluated against the column
+    names, returning names where the callable function evaluates to True. An
+    example of a valid callable argument would be ``lambda x: x.upper() in
+    ['AAA', 'BBB', 'DDD']``. Using this parameter results in much faster
+    parsing time and lower memory usage.
+usecols_excel : int or list, default None
```
chris-b1: string, int, or list

I'd also a bit of narrative before the bullets, something like:
Columns to parse from the spreadsheet, specified as Excel location references

pandas/io/excel.py
```diff
@@ -615,13 +649,27 @@ def _parse_cell(cell_contents, cell_typ):
...
+                        should_parse[j] = self._should_parse(j, usecols_excel)
 
-                    if usecols is None or should_parse[j]:
+                    if usecols_excel is None or should_parse[j]:
                         row.append(_parse_cell(value, typ))
                 data.append(row)
 
+            # Check if some string in usecols may be interpreted as a Excel
+            # positional column
+            if (usecols is not None) and (not callable(usecols)) and \
```
chris-b1: I'll try it out later, but I think this warning is over-specified.  Since `usecols` only used to work with excel ranges as strings, not lists of strings, I would limit the warning to that case.

pandas/tests/io/test_excel.py
```diff
@@ -163,20 +164,96 @@ def test_usecols_str(self, ext):
...
+        tm.assert_frame_equal(df2, df1, check_names=False)
+        tm.assert_frame_equal(df3, df2, check_names=False)
+
+    def test_usecols_diff_positional_str_columns_order(self, ext):
+
+        df1 = self.get_csv_refdf('test1')[['B', 'D']]
+
+        with tm.assert_produces_warning(UserWarning):
+            df2 = self.get_exceldf('test1', ext, 'Sheet1', usecols=['B', 'D'])
```
chris-b1: As state above I'm not sure this should warn, since it didn't previously work.  But I would like a test for the warning, something like `usecols='B,D'`

chris-b1 (changes requested):
looks pretty good will try it out tonight, can you also add a note to the excel section in `io.rst`

pandas/tests/io/test_excel.py
jacksonjos: I can't pass a string to usecols because if I do it throws an error. I already tested it.
So, does not make sense maintain the warning just to a string type parameter unless you do not want the error be thrown.

Do I should remove the warning, then?

The other changes you asked me are done. I'm just waiting to know what to do about this warning to commit again.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -856,6 +856,7 @@ Other API Changes
 - Constructing a Series from a list of length 1 no longer broadcasts this list when a longer index is specified (:issue:`19714`, :issue:`20391`).
 - :func:`DataFrame.to_dict` with ``orient='index'`` no longer casts int columns to float for a DataFrame with only int and float columns (:issue:`18580`)
 - A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousins, will now *always* be passed a ``Series``, rather than a ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
+- Changed the named argument `usecols` at :func:`read_excel` to `usecols_excel` that receives a list of index numbers or A1 index to select the columns that must be in the DataFrame, so the `usecols` argument can serve its purpose to select the columns that must be in the DataFrame using column labels (:issue:`18273`)
```
jreback: Please simplify this to make it more readable.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1166,6 +1167,7 @@ I/O
 - Bug in :func:`DataFrame.to_latex()` where a ``MultiIndex`` with an empty string as its name would result in incorrect output (:issue:`18669`)
 - Bug in :func:`read_json` where large numeric values were causing an ``OverflowError`` (:issue:`18842`)
 - Bug in :func:`DataFrame.to_parquet` where an exception was raised if the write destination is S3 (:issue:`19134`)
+- Bug in :func:`read_excel` where `usecols_excel` named argument as a list of strings were returning a empty DataFrame (:issue:`18273`)
```
jreback: how is this a bug? ``usecols_excel`` doesn't exist yet

pandas/io/excel.py
```diff
@@ -85,19 +85,41 @@
...
+
+usecols : list-like or callable, default None
+    Return a subset of the columns. If list-like, all elements must either
+    be positional (i.e. integer indices into the document columns) or string
+    that correspond to column names provided either by the user in `names` or
+    inferred from the document header row(s). For example, a valid list-like
+    `usecols` parameter would be [0, 1, 2] or ['foo', 'bar', 'baz']. Element
+    order is ignored, so ``usecols=[0, 1]`` is the same as ``[1, 0]`` and
+    ``usecols=['foo', 'bar']`` is the same as ``['bar', 'foo']``.
+    To instantiate a DataFrame from ``data`` with element order preserved use
```
jreback: this is not needed here, starting with keeping ordering (or you can put in the Notes section if you really want)

pandas/io/excel.py
```diff
@@ -85,19 +85,41 @@
...
+    for ``['bar', 'foo']`` order.
+
+    If callable, the callable function will be evaluated against the column
+    names, returning names where the callable function evaluates to True. An
+    example of a valid callable argument would be ``lambda x: x.upper() in
+    ['AAA', 'BBB', 'DDD']``. Using this parameter results in much faster
+    parsing time and lower memory usage.
+usecols_excel : int or list, default None
     * If None then parse all columns,
     * If int then indicates last column to be parsed
```
jreback: should``usecols_excel`` *only* accept column ranges?

jreback (changes requested):
I am confused here. IIUC usecols_excel was *only* going to accep the column ranges, and usecols for everything else?  these HAVE to be orthogonal parametes, otherwise this is very confusing.

jacksonjos:
Hi, @chris-b1, may you take a look at @jreback observations in the last changes I've made?
I would like to know if after his observations and the comments I wrote at https://github.com/pandas-dev/pandas/pull/20480#discussion_r185143157 you have something else to add. So, i can push all suggested changes at one time.

pandas/io/excel.py
jacksonjos: How about the case where ``usecols_excel`` may receive an int that indicates last column to be parsed?
Does this should be in usecols, too? If I do that it will be different from what ``usecols`` allows at read_csv function and it will be like a special case to be handle in ``usecols``.

jacksonjos:
Guys (@chris-b1 and @jreback), I've made a new push. You may have any questions about changes in the tests involving 'index_col'. If you don't understand something feel free o ask.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -962,6 +962,7 @@ Other API Changes
...
 - :func:`DataFrame.to_dict` with ``orient='index'`` no longer casts int columns to float for a DataFrame with only int and float columns (:issue:`18580`)
 - A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousins, will now *always* be passed a ``Series``, rather than a ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
+- Created the named argument `usecols_excel` that receives a string containing comma separated Excel ranges and columns or a list containing strings indicating Excel ranges and columns. The `usecols` named argument at :func:`read_excel` now receives a list of strings containing column labels or a list of integers representing column indexes or a callable or an int indicating the last column to be parsed. (:issue:`18273`)
```
jreback: reference ``:func:`read_excel` ``. ``usecols`` is not changing here, yes? it is the same as before

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1285,6 +1286,7 @@ I/O
 - Bug in :func:`DataFrame.to_latex()` where missing space characters caused wrong escaping and produced non-valid latex in some cases (:issue:`20859`)
 - Bug in :func:`read_json` where large numeric values were causing an ``OverflowError`` (:issue:`18842`)
 - Bug in :func:`DataFrame.to_parquet` where an exception was raised if the write destination is S3 (:issue:`19134`)
+- Bug in :func:`read_excel` where `usecols` named argument as a list of strings were returning a empty DataFrame (:issue:`18273`)
```
jreback: use double back ticks around ``usecols``

pandas/io/excel.py
```diff
@@ -85,20 +87,36 @@
     Column (0-indexed) to use as the row labels of the DataFrame.
     Pass None if there is no such column.  If a list is passed,
     those columns will be combined into a ``MultiIndex``.  If a
-    subset of data is selected with ``usecols``, index_col
+    subset of data is selected with ``usecols_excel``, index_col
```
jreback: list ``usecols`` as well

pandas/io/excel.py
```diff
@@ -85,20 +87,36 @@
...
     is based on the subset.
 parse_cols : int or list, default None
 
     .. deprecated:: 0.21.0
        Pass in `usecols` instead.
 
-usecols : int or list, default None
+usecols : list-like or callable, default None
```
jreback: was a single int allowed before?

pandas/io/excel.py
```diff
@@ -85,20 +87,36 @@
...
+    inferred from the document header row(s). For example, a valid list-like
+    `usecols` parameter would be [0, 1, 2] or ['foo', 'bar', 'baz'].
+
+    If callable, the callable function will be evaluated against the column
+    names, returning names where the callable function evaluates to True. An
+    example of a valid callable argument would be ``lambda x: x.upper() in
+    ['AAA', 'BBB', 'DDD']``. Using this parameter results in much faster
+    parsing time and lower memory usage.
+usecols_excel : string or list, default None
+    Columns to parse from a spreadsheet specified as Excel column references or
```
jreback: can you add a link to column ranges here

pandas/io/excel.py
```diff
@@ -85,20 +87,36 @@
...
+    column indexes starting from 0.
+
     * If None then parse all columns,
```
jreback: rather than a list, just show what this is doing. In particular mention that both this argument and ``usecols`` cannot be used together (note this in both doc-strings)

pandas/io/excel.py
```diff
@@ -285,7 +311,7 @@ def read_excel(io,
                header=0,
                names=None,
                index_col=None,
```
jreback: what happened to ``usecols``??

pandas/io/excel.py
```diff
@@ -512,6 +541,36 @@ def _parse_excel(self,
 
         _validate_header_arg(header)
 
+        if (usecols is not None) and (usecols_excel is not None):
+            raise TypeError("Cannot specify both `usecols` and `usecols_excel`"
```
jreback: ValueError

pandas/io/excel.py
```diff
@@ -512,6 +541,36 @@ def _parse_excel(self,
...
+                            ". Choose one of them.")
+        # Check if some string in usecols may be interpreted as a Excel
```
jreback: blank lines between cases

pandas/io/excel.py
```diff
@@ -85,20 +87,36 @@
...
+
     * If None then parse all columns,
-    * If int then indicates last column to be parsed
-    * If list of ints then indicates list of column numbers to be parsed
     * If string then indicates comma separated list of Excel column letters and
-      column ranges (e.g. "A:E" or "A,C,E:F").  Ranges are inclusive of
-      both sides.
+      column ranges (e.g. "A:E" or "A,C,E:F") to be parsed. Ranges are
+      inclusive of both sides.
+    * If list of strings each string shall be an Excel column letter or column
```
jreback: why are you accepting a list of the excel ranges? is that actually useful? its making the code more complicated

pandas/io/excel.py
```diff
@@ -512,6 +541,36 @@ def _parse_excel(self,
...
+                          "Excel ranges and columns as strings was "
+                          "renamed to `usecols_excel`.", UserWarning,
+                          stacklevel=3)
+            usecols_excel = usecols
+            usecols = None
+        # Check if usecols_excel is a list of Excel ranges and columns and if
+        # it is usecols_excel is transformed in a string containing
+        # Excel ranges and columns
+        elif isinstance(usecols_excel, list) and all(
+            isinstance(elem, compat.string_types) and
```
jreback: see my comment about, if you actually want to accept this, then you are duplicating the above check

pandas/io/excel.py
```diff
@@ -512,6 +541,36 @@ def _parse_excel(self,
...
+            all((x is not "") and (x in string.ascii_letters)
+                for x in re.split(r',|:', elem)) for elem in usecols_excel):
+            usecols_excel = ','.join(usecols_excel)
+
+        if (usecols_excel is not None) and not (
```
jreback: huh? this is the 3rd repeat of this same check

pandas/io/excel.py
```diff
@@ -269,6 +287,14 @@ def _get_default_writer(ext):
     return _default_writers[ext]
 
 
+def _is_int_or_intlist(usecols):
```
jreback: what is this for?

jreback (changes requested):
pls  simpify this. in particular allowing a list of excel ranges is a bit odd to do here

pandas/io/excel.py
jacksonjos: What do you mean by a link? You mean a link to a website? If yes, why if there is already examples of column indexes and ranges in the named argument documentation?

doc/source/whatsnew/v0.23.0.txt
jacksonjos: Not true. `usecols` didn't receive column labels in pandas 0.22, what it did receive before, in pandas 0.19, I think, but in `parse_cols`. Also, `usecols` didn't receive callable, this was possible only for `read_csv` method.

doc/source/whatsnew/v0.23.0.txt
jacksonjos: Done!

pandas/io/excel.py
jacksonjos: Done!

pandas/io/excel.py
jacksonjos: Yes, it was. Look: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_excel.html

pandas/io/excel.py
jacksonjos: Done! I already wrote this in the io.rst file.

pandas/io/excel.py
jacksonjos: Actually, it was never there. I forgot to add `usecols=None` in some functions because the code worked without it. =p

pandas/io/excel.py
jacksonjos: Done!

pandas/io/excel.py
jacksonjos: Done!

pandas/io/excel.py
jacksonjos: Actually, this is what originated this issue[1]. Now it's way bigger than that, but it's what was requested and agreed as you can see at [1]. 

[1] https://github.com/pandas-dev/pandas/issues/18273#issue-273644750

pandas/io/excel.py
jacksonjos: Are you sure? I have to check if `usecols_excel` is a string like `"A:E,W"` and if it's a list like `["A:E", "W"]`. First I check if `usecols_excel` is the second one and if it's I transform it in the first one. After that I check if it's the first one.

The thing is that `usecols_excel` may not be that list above (second case) and if it's not I also have to check if it's a A1 string. What I can do to avoid the duplicated check in the  special case you cited is place the if for the string in the if else where is the check for `usecols_excel` as a list.

pandas/io/excel.py
jacksonjos: It's a function I wrote to check if a function must parse inside `_read_excel` function the columns in `usecols` when `usecols` is a list of ints or a int indicating the fist n columns to be parsed. I did this way because if `usecols` if one of those I have to make it None in line 700 after it's used because if I don't the `index_col` named argument causes a bug after passed in TextParser class because it behaves differently the comparing to its behavior inside `read_excel` function.

jacksonjos:
@jreback and @chris-b1 Is there some problem with the code, yet?

jacksonjos:
@jreback and @chris-b1 Is there some problem with the code, yet?

doc/source/io.rst
```diff
@@ -2851,23 +2851,62 @@ Parsing Specific Columns
 
 It is often the case that users will insert columns to do temporary computations
 in Excel and you may not want to read in those columns. ``read_excel`` takes
-a ``usecols`` keyword to allow you to specify a subset of columns to parse.
+either a ``usecols`` or ``usecols_excel`` keyword to allow you to specify a
+subset of columns to parse. Note that you can't use both ``usecols`` and
+``usecols_excel`` named arguments at the same time.
+
+If ``usecols_excel`` is a string, then it is assumed that indicates a comma
```
jreback: that it indicates

jreback: If ``usecols_excel`` is supplied

doc/source/io.rst
```diff
@@ -2851,23 +2851,62 @@ Parsing Specific Columns
 
 It is often the case that users will insert columns to do temporary computations
 in Excel and you may not want to read in those columns. ``read_excel`` takes
-a ``usecols`` keyword to allow you to specify a subset of columns to parse.
+either a ``usecols`` or ``usecols_excel`` keyword to allow you to specify a
+subset of columns to parse. Note that you can't use both ``usecols`` and
```
jreback: can't -> can not

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -966,6 +966,7 @@ Other API Changes
...
 - :func:`DataFrame.to_dict` with ``orient='index'`` no longer casts int columns to float for a DataFrame with only int and float columns (:issue:`18580`)
 - A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousins, will now *always* be passed a ``Series``, rather than a ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
+- Created the named argument `usecols_excel` that receives a string containing comma separated Excel ranges and columns or a list containing strings indicating Excel ranges and columns. The `usecols` named argument at :func:`read_excel` now receives a list of strings containing column labels or a list of integers representing column indexes or a callable or an int indicating the last column to be parsed. (:issue:`18273`)
```
jreback: use double-backticks around the args (``usecols_excel``). 

start like this:

`` :func:`read_excel` has gained the keyword argument ``usecols_excel`` ``

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1289,6 +1290,7 @@ I/O
 - Bug in :func:`DataFrame.to_latex()` where missing space characters caused wrong escaping and produced non-valid latex in some cases (:issue:`20859`)
 - Bug in :func:`read_json` where large numeric values were causing an ``OverflowError`` (:issue:`18842`)
 - Bug in :func:`DataFrame.to_parquet` where an exception was raised if the write destination is S3 (:issue:`19134`)
+- Bug in :func:`read_excel` where ``usecols`` named argument as a list of strings were returning a empty DataFrame (:issue:`18273`)
```
jreback: double backticks around DataFrame

pandas/io/excel.py
```diff
@@ -85,20 +87,39 @@
...
+    that correspond to column names provided either by the user in `names` or
+    inferred from the document header row(s). For example, a valid list-like
+    `usecols` parameter would be [0, 1, 2] or ['foo', 'bar', 'baz']. Note that
+    you can't give both ``usecols`` and ``usecols_excel`` named arguments at
+    the same time.
+
+    If callable, the callable function will be evaluated against the column
+    names, returning names where the callable function evaluates to True. An
+    example of a valid callable argument would be ``lambda x: x.upper() in
+    ['AAA', 'BBB', 'DDD']``. Using this parameter results in much faster
```
jreback: you can remove this last sentence (about perf)

pandas/io/excel.py
```diff
@@ -85,20 +87,39 @@
...
+    parsing time and lower memory usage.
+usecols_excel : string or list, default None
+    Return a subset of the columns from a spreadsheet specified as Excel column
+    ranges and columns. Note that you can't use both ``usecols`` and
```
jreback: can't -> can not

pandas/io/excel.py
```diff
@@ -269,6 +290,22 @@ def _get_default_writer(ext):
...
+def _is_int_or_intlist(usecols):
+    if isinstance(usecols, int) or (isinstance(usecols, list) and
+                                    all(isinstance(v, int) for v in usecols)):
+        return True
+
+    return False
+
+
+def _is_excel_columns_notation(columns):
+    if isinstance(columns, compat.string_types) and all((x is not "") and
```
jreback: why are you comparing against an empty string? that is an error. make this as simple as possible. IN fact I would *only* to accept a string here, and not a list of excel notations.

pandas/io/excel.py
```diff
@@ -285,6 +322,7 @@ def read_excel(io,
                header=0,
                names=None,
                index_col=None,
+               usecols_excel=None,
```
jreback: put this arg after usecols

pandas/io/excel.py
```diff
@@ -512,6 +556,33 @@ def _parse_excel(self,
...
+                          "Excel ranges and columns as strings was "
+                          "renamed to `usecols_excel`.", UserWarning,
+                          stacklevel=3)
+            usecols_excel = usecols
+            usecols = None
+
+        # Check if usecols_excel is a list of Excel ranges and columns and if
+        # it is usecols_excel is transformed in a string containing
+        # Excel ranges and columns
+        elif isinstance(usecols_excel, list) and all(
```
jreback: see my comment above, let's simplfiy this

pandas/io/excel.py
```diff
@@ -512,6 +556,33 @@ def _parse_excel(self,
...
+                _is_excel_columns_notation(elem) for elem in usecols_excel):
+            usecols_excel = ','.join(usecols_excel)
+
+        elif (usecols_excel is not None) and not _is_excel_columns_notation(
```
jreback: why is this not caught above?

pandas/io/excel.py
```diff
@@ -615,13 +686,19 @@ def _parse_cell(cell_contents, cell_typ):
                 row = []
                 for j, (value, typ) in enumerate(zip(sheet.row_values(i),
                                                      sheet.row_types(i))):
-                    if usecols is not None and j not in should_parse:
-                        should_parse[j] = self._should_parse(j, usecols)
+                    if (j not in should_parse) and (_is_int_or_intlist(usecols)
```
jreback: rather than checking with a bespoke function, this can just fall thru to the TextParser which validates all of this

jreback (changes requested):
pls make the validation much much simpler. don't accept a list of strings for usecols_excel. Let the TextParser handle the int/ string validation of usecols. It already does this.

Adding this extra logic just makes read_excel way more complex.

pandas/io/excel.py
jacksonjos: Because `usecols` can not be an int in TextParser and because `index_col` is handled differently in TextParser so that.

In `read_excel` the keyword argument `index_col` make index of the `DataFrame` the _mth_ column of the _filtered columns_ while in TextParser the `index_col` make as an index of the `DataFrame` the _mth_ column of the `DataFrame` whether is in the selected columns in `usecols` or not. Do you understand why I did this?

Then, after check the cases when `usecols` is an int or a list of ints I make the code read the columns indicated by `usecols` and after that I make `usecols` None, so nothing gets wrong in TextParser.

jacksonjos:
@chris-b1 and @jreback, please take a look to the changes I've made and have in mind what I wrote in https://github.com/pandas-dev/pandas/pull/20480#discussion_r187803587

pandas/io/excel.py
chris-b1: This is pretty hairy, but I do see what you're trying to do to preserve compatibility.  Long term probably better off if we sync on usecols behavior with `read_csv`

jacksonjos:
Please, can anyone review the updates and check if there is something else I have to do?

@chris-b1, @jreback

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -996,6 +996,7 @@ Other API Changes
 - :func:`DataFrame.to_dict` with ``orient='index'`` no longer casts int columns to float for a DataFrame with only int and float columns (:issue:`18580`)
 - A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousins, will now *always* be passed a ``Series``, rather than a ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
 - Rolling and Expanding types raise ``NotImplementedError`` upon iteration (:issue:`11704`).
+- :func:`read_excel` has gained the keyword argument ``usecols_excel`` that receives a string containing comma separated Excel ranges and columns. The `usecols` keyword argument at :func:`read_excel` now receives a list of strings containing column labels or a list of integers representing column indexes or a callable or an int indicating the last column to be parsed. (:issue:`18273`)
```
jreback: move to 0.24.0

jreback: use double-backticks around usecols. What is actually changing from current?

pandas/io/excel.py
jreback: needs a versionadded tag

pandas/io/excel.py
```diff
@@ -85,20 +87,38 @@
...
+    that correspond to column names provided either by the user in `names` or
+    inferred from the document header row(s). For example, a valid list-like
+    `usecols` parameter would be [0, 1, 2] or ['foo', 'bar', 'baz']. Note that
+    you can not give both ``usecols`` and ``usecols_excel`` keyword arguments
+    at the same time.
+
+    If callable, the callable function will be evaluated against the column
+    names, returning names where the callable function evaluates to True. An
+    example of a valid callable argument would be ``lambda x: x.upper() in
+    ['AAA', 'BBB', 'DDD']``.
```
jreback: needs a versionchanged tag, mention what is changing

pandas/io/excel.py
```diff
@@ -269,6 +289,14 @@ def _get_default_writer(ext):
     return _default_writers[ext]
 
 
+def _is_excel_columns_notation(columns):
+    if isinstance(columns, compat.string_types) and all(
```
jreback: add a doc-string

pandas/io/excel.py
```diff
@@ -615,13 +669,22 @@ def _parse_cell(cell_contents, cell_typ):
...
-                    if usecols is None or should_parse[j]:
+                    if (((usecols_excel is None) and (usecols is None)) or
+                            should_parse[j]):
                         row.append(_parse_cell(value, typ))
                 data.append(row)
 
+            # In 'TextParser' class below 'usecols' can not be an int and at
+            # this point 'usecols' has already been used when it is an int.
+            # So, it receives None so the rest of the flow behaves as expected.
+            if isinstance(usecols, int):
```
jreback: this is very very odd, what are you trying to do here?

pandas/tests/io/test_excel.py
```diff
@@ -129,32 +129,34 @@ def test_usecols_list(self, ext):
         dfref = self.get_csv_refdf('test1')
         dfref = dfref.reindex(columns=['B', 'C'])
         df1 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
-                               usecols=[0, 2, 3])
```
jreback: why are all of these changing?

pandas/tests/io/test_excel.py
```diff
@@ -129,32 +129,34 @@ def test_usecols_list(self, ext):
...
         df2 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0,
-                               usecols='A:D')
+                               usecols_excel='A:D')
         df3 = self.get_exceldf('test1', ext, 'Sheet2', skiprows=[1],
-                               index_col=0, usecols='A:D')
+                               index_col=0, usecols_excel='A:D')
 
-        with tm.assert_produces_warning(FutureWarning):
+        with tm.assert_produces_warning() as w:
             df4 = self.get_exceldf('test1', ext, 'Sheet2', skiprows=[1],
```
jreback: why 2 warnings? add a comment

jreback (changes requested):
looks pretty good. some questions. @chris-b1 if you can look.

pandas/tests/io/test_excel.py
jacksonjos: Because the last changes you required in the code implied in removing backward compatibility for the `index_col` keyword behavior of `read_excel`.

@jreback, take a look here to understand better what I'm meaning: https://github.com/pandas-dev/pandas/pull/20480#discussion_r187773244

pandas/io/excel.py
jacksonjos: I can't see how the parser accepts it.
@jreback take a look to check I'm doing something wrong here.

I've commented the following
```if isinstance(usecols, int):
        usecols = None
``` 
and ran the following test
`pytest pandas/tests/io/test_excel.py -k test_usecols_int`

and the result is:

```
____________________ TestXlrdReader.test_usecols_int[.xlsm] ____________________

self = <pandas.tests.io.test_excel.TestXlrdReader object at 0x7ffa4fe5a278>
ext = '.xlsm'

    def test_usecols_int(self, ext):
    
        dfref = self.get_csv_refdf('test1')
        dfref = dfref.reindex(columns=['A', 'B', 'C'])
>       df1 = self.get_exceldf('test1', ext, 'Sheet1', index_col=0, usecols=3)

pandas/tests/io/test_excel.py:114: 
pandas/tests/io/test_excel.py:104: in get_exceldf
    return read_excel(pth, *args, **kwds)
pandas/util/_decorators.py:177: in wrapper
    return func(*args, **kwargs)
pandas/util/_decorators.py:177: in wrapper
    return func(*args, **kwargs)
pandas/io/excel.py:369: in read_excel
    **kwds)
pandas/io/excel.py:498: in parse
    **kwds)
pandas/io/excel.py:759: in _parse_excel
    **kwds)
pandas/io/parsers.py:2008: in TextParser
    return TextFileReader(*args, **kwds)
pandas/io/parsers.py:787: in __init__
    self._make_engine(self.engine)
pandas/io/parsers.py:1024: in _make_engine
    self._engine = klass(self.f, **self.options)
pandas/io/parsers.py:2052: in __init__
    self.usecols, _ = _validate_usecols_arg(kwds['usecols'])

usecols = 3

    def _validate_usecols_arg(usecols):
        """
        Validate the 'usecols' parameter.
    
        Checks whether or not the 'usecols' parameter contains all integers
        (column selection by index), strings (column by name) or is a callable.
        Raises a ValueError if that is not the case.
    
        Parameters
        ----------
        usecols : list-like, callable, or None
            List of columns to use when parsing or a callable that can be used
            to filter a list of table columns.
    
        Returns
        -------
        usecols_tuple : tuple
            A tuple of (verified_usecols, usecols_dtype).
    
            'verified_usecols' is either a set if an array-like is passed in or
            'usecols' if a callable or None is passed in.
    
            'usecols_dtype` is the inferred dtype of 'usecols' if an array-like
            is passed in or None if a callable or None is passed in.
        """
        msg = ("'usecols' must either be list-like of all strings, all unicode, "
               "all integers or a callable.")
        if usecols is not None:
            if callable(usecols):
                return usecols, None
            # GH20529, ensure is iterable container but not string.
            elif not is_list_like(usecols):
>               raise ValueError(msg)
E               ValueError: 'usecols' must either be list-like of all strings, all unicode, all integers or a callable.

pandas/io/parsers.py:1203: ValueError

```

jacksonjos: @chris-b1 may you help me here, too, please?

jreback: it must be a list

jacksonjos: So, just to be sure. @jreback, you want me to remove support to receives and int in `usecols`?
Is that correct?

jreback: a single int is not allowed
you should not be special casing anything here
TextParser is the ground truth

jacksonjos:
@jreback and @chris-b1 I've just updated the commit. Take another look, please.

jacksonjos:
@jreback and @chris-b1, may you take a look ate the updated commit?

chris-b1:
@jacksonjos, sorry I'm having a bit of trouble wrapping my head around exactly how the changes to `index_col` / `usecols` interact - could you show a before and after example of where behavior will be different?  If we are going to change would want that in the docs anyways - otherwise this looks good to me.

jacksonjos:
The issue  is related to spreadsheets whose first column does not have a column name. So, when reading the `DataFrame` from `read_excel` without selecting it to be read, it is interpreted as the index in the current behavior of `read_excel`.

There is an example that is used in some tests of pandas whose spreadsheet can be seen at `pandas/tests/io/data/test1.xlsx`.

|                | A              | B               | C                | D                |
|----------------|-----------------|------------------|------------------|----------------|
| 1/3/2000 0:00  | 0.980268513777  | 3.68573087906    | -0.364216805298  | -1.15973806169 |
| 1/4/2000 0:00  | 1.04791624281   | -0.0412318367011 | -0.16181208307   | 0.212549316967 |
| 1/5/2000 0:00  | 0.498580885705  | 0.731167677815   | -0.537677223318  | 1.34627041952  |
| 1/6/2000 0:00  | 1.12020151869   | 1.56762092543    | 0.00364077397681 | 0.67525259227  |
| 1/7/2000 0:00  | -0.487094399463 | 0.571454623474   | -1.6116394093    | 0.103468562917 |
| 1/10/2000 0:00 | 0.836648671666  | 0.246461918642   | 0.588542635376   | 1.0627820613   |
| 1/11/2000 0:00 | -0.157160753327 | 1.34030689438    | 1.19577795622    | -1.09700699751 |

So, below there is a comparison of the change from the current behavior of `index_col` to the new one I have written.

```
> df_current = pd.read_excel("pandas/tests/io/data/test1.xlsx", usecols=[0, 2, 3], index_col=0)

> df_current

B         C
2000-01-03  3.685731 -0.364217
2000-01-04 -0.041232 -0.161812
2000-01-05  0.731168 -0.537677
2000-01-06  1.567621  0.003641
2000-01-07  0.571455 -1.611639
2000-01-10  0.246462  0.588543
2000-01-11  1.340307  1.195778

> df_new_one = pd.read_excel("/pandas/tests/io/data/test1.xlsx", usecols=[1, 2], index_col=0)

> df_new_one

B         C
2000-01-03  3.685731 -0.364217
2000-01-04 -0.041232 -0.161812
2000-01-05  0.731168 -0.537677
2000-01-06  1.567621  0.003641
2000-01-07  0.571455 -1.611639
2000-01-10  0.246462  0.588543
2000-01-11  1.340307  1.195778
```

doc/source/io.rst
```diff
@@ -2852,23 +2852,53 @@ Parsing Specific Columns
...
+   read_excel('path_to_file.xls', 'Sheet1', usecols_excel='A:E')
+   read_excel('path_to_file.xls', 'Sheet1', usecols_excel='A,C,E:F')
 
 If ``usecols`` is an integer, then it is assumed to indicate the last column
 to be parsed.
 
 .. code-block:: python
 
-   read_excel('path_to_file.xls', 'Sheet1', usecols=2)
+   read_excel('path_to_file.xls', 'Sheet1', usecols_excel=2)
```
WillAyd: What's the point of this addition? IMO this is going to be pretty confusing and `usecols_excel` should be limited to A1 notation. 

Would be much more consistent using the (admittedly more verbose) usecols alternative and raising for non-A1 input:

```python
read_excel('path_to_file.xls', 'Sheet1', usecols=range(2))
```

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1325,6 +1325,7 @@ I/O
 - Bug in :func:`DataFrame.to_latex()` where missing space characters caused wrong escaping and produced non-valid latex in some cases (:issue:`20859`)
 - Bug in :func:`read_json` where large numeric values were causing an ``OverflowError`` (:issue:`18842`)
 - Bug in :func:`DataFrame.to_parquet` where an exception was raised if the write destination is S3 (:issue:`19134`)
+- Bug in :func:`read_excel` where ``usecols`` keyword argument as a list of strings were returning a empty ``DataFrame`` (:issue:`18273`)
```
WillAyd: Don't need this note since you already have this covered in `v0.24.0`

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -35,7 +35,7 @@ Datetimelike API Changes
 Other API Changes
 ^^^^^^^^^^^^^^^^^
 
--
+- :func:`read_excel` has gained the keyword argument ``usecols_excel`` that receives a string containing comma separated Excel ranges and columns. The ``usecols`` keyword argument at :func:`read_excel` had removed support for a string containing comma separated Excel ranges and columns and for an int indicating the first j columns to be read in a ``DataFrame``. Also, the ``usecols`` keyword argument at :func:`read_excel` had added support for receiving a list of strings containing column labels and a callable. (:issue:`18273`)
```
WillAyd: Better written as:

```
- :func:`read_excel:" now has a keyword argument of ``usecols_excel`` which allows you to parse select columns via A1 notation (:issue:`18273`)
```

doc/source/io.rst
```diff
@@ -2852,23 +2852,53 @@ Parsing Specific Columns
...
+subset of columns to parse. Note that you can not use both ``usecols`` and
+``usecols_excel`` named arguments at the same time.
+
+If ``usecols_excel`` is supplied, then it is assumed that indicates a comma
+separated list of Excel column letters and column ranges to be parsed.
+
+.. code-block:: python
+
+   read_excel('path_to_file.xls', 'Sheet1', usecols_excel='A:E')
+   read_excel('path_to_file.xls', 'Sheet1', usecols_excel='A,C,E:F')
```
WillAyd: For consistency I think we should *only* allow A1 notation, which the second example here does not do. Instead I think this should be written as

```python
read_excel('path_to_file.xls', 'Sheet1', usecols_excel='A:A,C:C,E:F')
```

Raises for anything else.  See

https://docs.microsoft.com/en-us/previous-versions/office/developer/office-2007/bb211395(v=office.12)

pandas/io/excel.py
```diff
@@ -269,6 +296,17 @@ def _get_default_writer(ext):
     return _default_writers[ext]
 
 
+def _is_excel_columns_notation(columns):
+    """Receives a string and check if the string is a comma separated list of
```
WillAyd: Even though this is a private method would be preferable for the docstring to follow the pandas convention

WillAyd (changes requested):
Just adding my two cents on this change - I think there are some aspects of the API here that need to be cleaned up to make this really effective

doc/source/whatsnew/v0.24.0.txt
jacksonjos: @WillAyd, please, take a look at this: https://github.com/pandas-dev/pandas/pull/20480#discussion_r192605116

@jreback has asked to point what is changing.

@jreback, should I do what @WillAyd is asking or should I keep the message I have already written?

doc/source/io.rst
jacksonjos: I think that despite of the second example does not follow exactly the A1 notation it's simpler to write it.

@jreback, what do you think?

jacksonjos: @jreback and @chris-b1, can you review this, this https://github.com/pandas-dev/pandas/pull/20480#discussion_r194238427 and this https://github.com/pandas-dev/pandas/pull/20480#issuecomment-395975007, please?

jacksonjos: @jreback and @chris-b1, can you review this https://github.com/pandas-dev/pandas/pull/20480#discussion_r194238455, this https://github.com/pandas-dev/pandas/pull/20480#discussion_r194238427 and this https://github.com/pandas-dev/pandas/pull/20480#issuecomment-395975007, please?

chris-b1: We already support the single column `'A'` vs `'A:A'` so fine with this

chris-b1:
I re-read your explanation but I'm honestly still not following why the change to `usecols` with numerics is necessary, sorry I'm missing something.  `read_csv` is consistent with the old behavior and with `engine='python'` also uses the `TextParser` class, why can't we do the same here?

```python
In [6]: pd.read_excel('pandas/pandas/tests/io/data/test1.xlsx').to_csv('tmp.csv')

In [7]: !head tmp.csv
,A,B,C,D
2000-01-03,0.980268513777,3.68573087906,-0.364216805298,-1.15973806169
2000-01-04,1.04791624281,-0.0412318367011,-0.16181208307,0.212549316967
2000-01-05,0.498580885705,0.731167677815,-0.537677223318,1.34627041952
2000-01-06,1.12020151869,1.56762092543,0.00364077397681,0.67525259227
2000-01-07,-0.487094399463,0.571454623474,-1.6116394093,0.103468562917
2000-01-10,0.836648671666,0.246461918642,0.588542635376,1.0627820613
2000-01-11,-0.157160753327,1.34030689438,1.19577795622,-1.09700699751

In [8]: pd.read_csv('tmp.csv', usecols=[0, 2, 3], index_col=0)
Out[8]:
                   B         C
2000-01-03  3.685731 -0.364217
2000-01-04 -0.041232 -0.161812
2000-01-05  0.731168 -0.537677
2000-01-06  1.567621  0.003641
2000-01-07  0.571455 -1.611639
2000-01-10  0.246462  0.588543
2000-01-11  1.340307  1.195778

In [9]: pd.read_csv('tmp.csv', usecols=[0, 2, 3], index_col=0, engine='python')
Out[9]:
                   B         C
2000-01-03  3.685731 -0.364217
2000-01-04 -0.041232 -0.161812
2000-01-05  0.731168 -0.537677
2000-01-06  1.567621  0.003641
2000-01-07  0.571455 -1.611639
2000-01-10  0.246462  0.588543
2000-01-11  1.340307  1.195778
```

jreback:
@jacksonjos can you show the proposed right next to the existing. and use separators between so its clear.

jreback:
can you rebase / update

jacksonjos:
Hi @jreback,

Unfortunately, I don't have time to work on this project this semester. I
tried to hard to contribute to it last semester, but my efforts weren't
enough to finish the job.
So, I advise you to ask for someone else to continue this work.

Best regards

Le mar. 25 sept. 2018 à 13:41, Jeff Reback <notifications@github.com> a
écrit :

jreback:
@gfyoung if you wouldn't mind rebasing this

gfyoung:
@jreback : Before I do, would like to get a better understanding of this issue (see [here](https://github.com/pandas-dev/pandas/issues/18273#issuecomment-433751677)).

jreback:
closing in favor of #23544

PR #20481, opened 2018-03-25
removed not necessary bn switch decorator on nansum

asdf8601:
Removed not necessary bottleneck_switch decorator on nansum function like nanprod.

### DIFF
--- a/pandas/core/nanops.py
+++ b/pandas/core/nanops.py
@@ -326,7 +326,6 @@ def nanall(values, axis=None, skipna=True):
 
 
 @disallow('M8')
-@bottleneck_switch()
 def nansum(values, axis=None, skipna=True, min_count=0):
     values, mask, dtype, dtype_max = _get_values(values, skipna, 0)
     dtype_sum = dtype_max


jreback:
should also be on ``nansum``. how did you come to change this? e.g. what broke?

mmngreco:
oh, now I don't see it so clearly and I'm probably wrong. Nothing has been broken, I was checking the code and I noticed that `bottleneck_switch` seems to do nothing when the function is `nansum` or `nanprod` according to `_bn_ok_dtype`, I thought it would be an unnecessary call.

jreback:
@mmngreco no it IS unecessary since 0.22 (as we compute sum/prod inside pandas only). was just wondering how you detected it.

happy to have you remove the one on nansum as well. I don't think there is a test for this that would break, but if you have something would be great.

mmngreco:
I have not found any `bottleneck_switch`, I understand that this would be necessary for this case. I haven't written anything yet. I will try in the next few days as I am not sure how to do it.

TomAugspurger (approved):
I also don't think you'll find a test where this matters.

Merging later today.

TomAugspurger:
Thanks @mmngreco

PR #20485, opened 2018-03-26
POC for New GroupBy Dispatching Module

WillAyd:
This is nowhere near completion but looking for feedback on the direction.

@jreback 

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/util/_dispatching.py
+++ b/pandas/core/util/_dispatching.py
@@ -1,0 +1,186 @@
+import collections
+import inspect
+from functools import partial
+
+import numpy as np
+
+from pandas._libs import groupby as libgroupby
+from pandas.core.dtypes.missing import isnull
+
+
+class CythonDispatcher(object):
+
+    def __init__(self, groupby):
+        self.groupby = groupby
+        self.func_nm = None
+        self.obj = None
+
+    @property
+    def func_metadata(self):
+        """
+        Stores the metadata required to dispatch each function.
+
+        The format of the dict is as follows:
+
+        attr_name : {
+            'application': {'aggregate', 'transform'}
+            'cython_nm': ...  # Name of the Cython function to call
+            'extra_kwargs': {...}  # Extra kwargs to pass to Cython
+            'type_blacklist': [...]  # Dtypes for which func should raise
+            'result_type': ...  # dtype of result from Cython
+            'conversion_in': ...  # dtype or callable for conversion pre-Cython
+            'needs_values': ...  # Whether the obj values should pass to Cython
+            'needs_mask': ...   # Whether a mask of NA values should be passed
+            'conversion_out': ...  # dtype or callable for conv post-Cython
+        }
+        """
+        return {
+            'any': {
+                'application': 'aggregate',
+                'cython_nm' : 'group_any_all',
+                'extra_kwargs': {'val_test': 'any'},
+                'type_blacklist': [],
+                'result_type': np.uint8,
+                'conversion_in': self._any_all_convertor,
+                'needs_values': True,
+                'needs_mask': True,
+                'conversion_out': np.bool
+            },
+            'all': {
+                'application': 'aggregate',
+                'cython_nm' : 'group_any_all',
+                'extra_kwargs': {'val_test': 'all'},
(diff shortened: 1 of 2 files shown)


pandas/core/util/_dispatching.py
```diff
@@ -0,0 +1,186 @@
...
+            'type_blacklist': [...]  # Dtypes for which func should raise
+            'result_type': ...  # dtype of result from Cython
+            'conversion_in': ...  # dtype or callable for conversion pre-Cython
+            'needs_values': ...  # Whether the obj values should pass to Cython
+            'needs_mask': ...   # Whether a mask of NA values should be passed
+            'conversion_out': ...  # dtype or callable for conv post-Cython
+        }
+        """
+        return {
+            'any': {
```
WillAyd: Thought here is that a dict of dicts can most cleanly describe the metadata that each function may have to pass to Cython

pandas/core/util/_dispatching.py
```diff
@@ -0,0 +1,186 @@
...
+                'needs_mask': True,
+                'conversion_out': np.bool
+            }
+        }
+
+    @property
+    def application_type(self):
+        return self.func_metadata[self.func_nm]['application']
+
+    def _any_all_convertor(self, vals):
```
WillAyd: Hoping we don't have a ton of these, but any custom conversion routines can be defined in this module (maybe not even within this class) and simply reference in the metadata dict

pandas/core/util/_dispatching.py
```diff
@@ -0,0 +1,186 @@
...
+        Raises
+        ------
+        ``TypeError`` if the `grp_by` dtypes are not valid for `func_nm`.
+        """
+        if self.obj.values.dtype in self.func_metadata[
+                self.func_nm]['type_blacklist']:
+            raise TypeError("'{}' cannot be applied to a dtype of {}".format(
+                self.func_nm, self.obj.values.dtype))
+
+    def _get_result(self, **kwargs):
```
WillAyd: This func does all the heavy lifting; mostly matches the implementation of `_get_cythonized_result` currently in `groupby.py`

pandas/core/util/_dispatching.py
```diff
@@ -0,0 +1,186 @@
+import collections
+import inspect
```
jreback: rather than put this here and trying to make this too generic. let's do:

- move ``pandas.core.groupby`` to ``pandas.core.groupby.groupby`` (and add a shim to deprecate ``pandas.core.groupby``, simliar to what we do for ``pandas.core.categorical``), do as a pre-cursor PR, only doing this move
- put this module in ``pandas.core.groupby``, call it ``dispatch.py``

pandas/core/util/_dispatching.py
```diff
@@ -0,0 +1,186 @@
...
+        ----------
+        kwargs
+            Extra arguments to bind to the `func_nm` Cython signature.
+
+        Resolve function name in case of templating use.
+        """
+        # Since this func is called in a loop, the below might be better
+        # served outside of the loop and passed in?
+        labels, _, ngroups = self.groupby.grouper.group_info
+
```
jreback: see above these should be subclasses

jreback: I would make the things to dispatch just methods. Otherwise this has to have a lot of magic. Its much simpler to just call a method.

jreback (changes requested):

pandas/core/util/_dispatching.py
WillAyd: Struggling to conceptualize this - so do you think it would be best to have a `BaseDispatcher` and then have the various groupings of applications as the subclasses (so here `AnyAllDispatcher`)?

jreback: see my comments above 

first order of business is to move groupby to a sub module 
so can easily create more modules

jreback:
if you can move underneath pandas/core/groupby then can have a look

WillAyd:
Closing for now - going to take an additional pass at a later date

PR #20487, opened 2018-03-26
BUG: raise error when setting cached properties

jorisvandenbossche:
A bug I introduced myself some time ago in https://github.com/pandas-dev/pandas/pull/19991. Apparently the object needs to explicitly raise in the `__set__` method, otherwise the property is assumed to be settable:

```
In [6]: idx = pd.Index([1, 2, 3])

In [7]: idx.is_unique
Out[7]: True

In [8]: idx.is_unique = False

In [9]: idx.is_unique
Out[9]: False

In [10]: idx.is_monotonic
Out[10]: True

In [11]: idx.is_monotonic = False
AttributeError                            Traceback (most recent call last)
<ipython-input-11-b2a74ac7466c> in <module>()
----> 1 idx.is_monotonic = False

AttributeError: can't set attribute
```

`is_unique` is a cached property, didn't raise anymore on master (see above), in contrast to normal property `is_monotonic`.

This patch ensures also cached properties have the same error.

### DIFF
--- a/pandas/_libs/properties.pyx
+++ b/pandas/_libs/properties.pyx
@@ -37,6 +37,9 @@ cdef class CachedProperty(object):
             PyDict_SetItem(cache, self.name, val)
         return val
 
+    def __set__(self, obj, value):
+        raise AttributeError("Can't set attribute")
+
 
 cache_readonly = CachedProperty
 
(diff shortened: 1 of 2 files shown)


jreback:
lgtm.

PR #16247, opened 2017-05-05
BUG: rolling.quantile does not return an interpolated result

guillemborrell:
Now computing the quantile of a rolling window matches the default of the rolling method in Series, DataFrame and np.percentile.

Fixes bugs #9413 and #16211

 - [x] closes #9413
 - [x] tests added / passed
 - [x] passes ``git diff upstream/master --name-only -- '*.py' | flake8 --diff``
 - [x] whatsnew entry

PR #20497, opened 2018-03-27
Add interpolation options to rolling quantile

kornilova203:
It version 0.21.0 rolling quantile started to use linear interpolation, it broke backward compatibility.

Regular (not rolling) quantile supports these interpolation options: `linear`, `lower`, `higher`, `nearest` and `midpoint`.
This commit adds the same options to moving quantile.

Performance issues of this commit (note: I re-run benchmarks, see message below) 
This code has 15% worse performance on benchmarks with small values of window (`window=10`). This is because loop inside `roll_quantile` now contains switch.
I tried to replace switch with callback but it led to even worse performance. Even if I move some of the code to new function (without any change in logic) it still makes performance much worse.
How bad is it? Could you please give me an advice on how to arrange the code such that it has the same performance?

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/asv_bench/benchmarks/rolling.py
+++ b/asv_bench/benchmarks/rolling.py
@@ -65,25 +65,15 @@ class Quantile(object):
     params = (['DataFrame', 'Series'],
               [10, 1000],
               ['int', 'float'],
-              [0, 0.5, 1])
+              [0, 0.5, 1],
+              ['linear', 'nearest', 'lower', 'higher', 'midpoint'])
     param_names = ['constructor', 'window', 'dtype', 'percentile']
 
-    def setup(self, constructor, window, dtype, percentile):
+    def setup(self, constructor, window, dtype, percentile, interpolation):
         N = 10 ** 5
         arr = np.random.random(N).astype(dtype)
         self.roll = getattr(pd, constructor)(arr).rolling(window)
 
-    def time_quantile(self, constructor, window, dtype, percentile):
-        self.roll.quantile(percentile)
-
-    def time_quantile_nearest(self, constructor, window, dtype, percentile):
-        self.roll.quantile(percentile, interpolation='nearest')
-
-    def time_quantile_lower(self, constructor, window, dtype, percentile):
-        self.roll.quantile(percentile, interpolation='lower')
-
-    def time_quantile_higher(self, constructor, window, dtype, percentile):
-        self.roll.quantile(percentile, interpolation='higher')
-
-    def time_quantile_midpoint(self, constructor, window, dtype, percentile):
-        self.roll.quantile(percentile, interpolation='midpoint')
+    def time_quantile(self, constructor, window, dtype, percentile,
+                      interpolation):
+        self.roll.quantile(percentile, interpolation=interpolation)
--- a/asv_bench/benchmarks/rolling.py
+++ b/asv_bench/benchmarks/rolling.py
@@ -22,6 +22,7 @@ class Methods(object):
     def time_rolling(self, constructor, window, dtype, method):
         getattr(self.roll, method)()
 
+
 class VariableWindowMethods(Methods):
     sample_time = 0.2
     params = (['DataFrame', 'Series'],
@@ -36,6 +37,7 @@ class VariableWindowMethods(Methods):
         arr = (100 * np.random.random(N)).astype(dtype)
         index = pd.date_range('2017-01-01', periods=N, freq='5s')
         self.roll = getattr(pd, constructor)(arr, index=index).rolling(window)
+
 
 class Pairwise(object):
 
@@ -59,7 +61,6 @@ class Pairwise(object):
 
 
 class Quantile(object):
-
     sample_time = 0.2
     params = (['DataFrame', 'Series'],
               [10, 1000],
@@ -68,9 +69,21 @@ class Quantile(object):
     param_names = ['constructor', 'window', 'dtype', 'percentile']
 
     def setup(self, constructor, window, dtype, percentile):
-        N = 10**5
+        N = 10 ** 5
         arr = np.random.random(N).astype(dtype)
         self.roll = getattr(pd, constructor)(arr).rolling(window)
 
     def time_quantile(self, constructor, window, dtype, percentile):
         self.roll.quantile(percentile)
+
+    def time_quantile_nearest(self, constructor, window, dtype, percentile):
+        self.roll.quantile(percentile, interpolation='nearest')
+
+    def time_quantile_lower(self, constructor, window, dtype, percentile):
+        self.roll.quantile(percentile, interpolation='lower')
+
+    def time_quantile_higher(self, constructor, window, dtype, percentile):
+        self.roll.quantile(percentile, interpolation='higher')
+
+    def time_quantile_midpoint(self, constructor, window, dtype, percentile):
+        self.roll.quantile(percentile, interpolation='midpoint')
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -443,6 +443,7 @@ Other Enhancements
 - :meth:`DataFrame.to_sql` now performs a multivalue insert if the underlying connection supports itk rather than inserting row by row.
   ``SQLAlchemy`` dialects supporting multivalue inserts include: ``mysql``, ``postgresql``, ``sqlite`` and any dialect with ``supports_multivalues_insert``. (:issue:`14315`, :issue:`8953`)
 - :func:`read_html` now accepts a ``displayed_only`` keyword argument to controls whether or not hidden elements are parsed (``True`` by default) (:issue:`20027`)
+- :meth:`Rolling.quantile` and :meth:`Expanding.quantile` now accept ``interpolation`` keyword (``linear`` by default) (:issue:`20497`)
 - zip compression is supported via ``compression=zip`` in :func:`DataFrame.to_pickle`, :func:`Series.to_pickle`, :func:`DataFrame.to_csv`, :func:`Series.to_csv`, :func:`DataFrame.to_json`, :func:`Series.to_json`. (:issue:`17778`)
 - :class:`WeekOfMonth` constructor now supports ``n=0`` (:issue:`20517`).
 - :class:`DataFrame` and :class:`Series` now support matrix multiplication (```@```) operator (:issue:`10259`) for Python>=3.5
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -449,6 +449,7 @@ Other Enhancements
 - Updated ``to_gbq`` and ``read_gbq`` signature and documentation to reflect changes from
   the Pandas-GBQ library version 0.4.0. Adds intersphinx mapping to Pandas-GBQ
   library. (:issue:`20564`)
+- :meth:`Rolling.quantile` and :meth:`Expanding.quantile` now accept ``interpolation`` keyword
 
 .. _whatsnew_0230.api_breaking:
 
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -443,14 +443,13 @@ Other Enhancements
 - :meth:`DataFrame.to_sql` now performs a multivalue insert if the underlying connection supports itk rather than inserting row by row.
   ``SQLAlchemy`` dialects supporting multivalue inserts include: ``mysql``, ``postgresql``, ``sqlite`` and any dialect with ``supports_multivalues_insert``. (:issue:`14315`, :issue:`8953`)
 - :func:`read_html` now accepts a ``displayed_only`` keyword argument to controls whether or not hidden elements are parsed (``True`` by default) (:issue:`20027`)
-- :meth:`Rolling.quantile` and :meth:`Expanding.quantile` now accept ``interpolation`` keyword (``linear`` by default) (:issue:`20497`)
+- :meth:`Rolling.quantile` and :meth:`Expanding.quantile` now accept the ``interpolation`` keyword, ``linear`` by default (:issue:`20497`)
 - zip compression is supported via ``compression=zip`` in :func:`DataFrame.to_pickle`, :func:`Series.to_pickle`, :func:`DataFrame.to_csv`, :func:`Series.to_csv`, :func:`DataFrame.to_json`, :func:`Series.to_json`. (:issue:`17778`)
 - :class:`WeekOfMonth` constructor now supports ``n=0`` (:issue:`20517`).
 - :class:`DataFrame` and :class:`Series` now support matrix multiplication (```@```) operator (:issue:`10259`) for Python>=3.5
 - Updated ``to_gbq`` and ``read_gbq`` signature and documentation to reflect changes from
   the Pandas-GBQ library version 0.4.0. Adds intersphinx mapping to Pandas-GBQ
   library. (:issue:`20564`)
-- :meth:`Rolling.quantile` and :meth:`Expanding.quantile` now accept ``interpolation`` keyword
 
 .. _whatsnew_0230.api_breaking:
 
--- a/pandas/_libs/window.pyx
+++ b/pandas/_libs/window.pyx
@@ -1357,27 +1357,21 @@ cdef _roll_min_max(ndarray[numeric] input, int64_t win, int64_t minp,
     return output
 
 
-def _get_interpolation_id(str interpolation):
-    """
-    Converts string to interpolation id
+cdef enum InterpolationType:
+    LINEAR,
+    LOWER,
+    HIGHER,
+    NEAREST,
+    MIDPOINT
 
-    Parameters
-    ----------
-    interpolation: 'linear', 'lower', 'higher', 'nearest', 'midpoint'
-    """
-    if interpolation == 'linear':
-        return 0
-    elif interpolation == 'lower':
-        return 1
-    elif interpolation == 'higher':
-        return 2
-    elif interpolation == 'nearest':
-        return 3
-    elif interpolation == 'midpoint':
-        return 4
-    else:
-        raise ValueError("Interpolation {} is not supported"
-                         .format(interpolation))
+
+interpolation_types = {
+    'linear': LINEAR,
+    'lower': LOWER,
+    'higher': HIGHER,
+    'nearest': NEAREST,
+    'midpoint': MIDPOINT,
+}
 
 
 def roll_quantile(ndarray[float64_t, cast=True] input, int64_t win,
@@ -1395,14 +1389,16 @@ def roll_quantile(ndarray[float64_t, cast=True] input, int64_t win,
         ndarray[int64_t] start, end
         ndarray[double_t] output
         double vlow, vhigh
-        int interpolation_id
+        InterpolationType interpolation_type
 
     if quantile <= 0.0 or quantile >= 1.0:
         raise ValueError("quantile value {0} not in [0, 1]".format(quantile))
 
-    # interpolation_id is needed to avoid string comparisons inside the loop
-    # I tried to use callback but it resulted in worse performance
-    interpolation_id = _get_interpolation_id(interpolation)
+    try:
+        interpolation_type = interpolation_types[interpolation]
+    except ValueError:
+        raise ValueError("Interpolation {} is not supported"
+                         .format(interpolation))
 
     # we use the Fixed/Variable Indexer here as the
     # actual skiplist ops outweigh any window computation costs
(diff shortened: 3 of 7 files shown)


jreback:
what issue are you addressing here? if its about performance, pls show representative benchmarks (or run the asv)

> it version 0.21.0 rolling quantile started to use linear interpolation, it broke backward compatibility.
if you read this: https://github.com/pandas-dev/pandas/pull/16247/files not sure what to say here, it had different defaults.

jreback:
re-reading your changes I see that you are trying to fix perf. can you run the asv's here and show the results. (you *may* need to add one if the case you are addressing is not covered).

kornilova-l:
@jreback, hi,

There is no issue related to this pull request. Do I need to create an issue first?

I am aware about the BUG that you referenced. Company at which I work uses rolling quantile without interpolation, when we updated pandas version, we discovered that behavior of rolling quantile has changed and there is no way to get non-interpolated results.

I re-ran asv benchmarks for rolling quantile (default version uses linear interpolation) and it did not show significant changes, sorry for confusing you.
```
BENCHMARKS NOT SIGNIFICANTLY CHANGED
```
I added new benchmarks for different interpolation options. I cannot compare them with previous results because interpolation options were not supported (supporting them is the purpose of this commit). Here are new benchmarks:
```
[ 20.00%] ··· Running rolling.Quantile.time_quantile             1.37±0.2ms;...
[ 40.00%] ··· Running rolling.Quantile.time_quantile_higher      1.38±0.2ms;...
[ 60.00%] ··· Running rolling.Quantile.time_quantile_lower       1.37±0.2ms;...
[ 80.00%] ··· Running rolling.Quantile.time_quantile_midpoint    1.38±0.2ms;...
[100.00%] ··· Running rolling.Quantile.time_quantile_nearest     1.37±0.2ms;...
```
I am not sure if I should add this new benchmarks to the commit. Should all the code be covered by benchmarks?

TomAugspurger:
@kornilova-l some linting issues: https://travis-ci.org/pandas-dev/pandas/jobs/358799956#L3014

jreback:
@kornilova-l 

> There is no issue related to this pull request. Do I need to create an issue first?

Generally create an issue first. Since you already created a PR no need then.

Ok so perf is basically unchanged on small series. If you want to add an asv for that is ok.

pandas/_libs/window.pyx
```diff
@@ -1357,25 +1357,53 @@ cdef _roll_min_max(ndarray[numeric] input, int64_t win, int64_t minp,
     return output
 
 
+def _get_interpolation_id(str interpolation):
```
jreback: look in ``algos.pxd`` for how we handle ``cdef enum TiebreakEnumType:``

e.g you create an enum and a dict to map to that enum. This becomes a bit simpler then.

pandas/_libs/window.pyx
```diff
@@ -1414,18 +1442,31 @@ def roll_quantile(ndarray[float64_t, cast=True] input, int64_t win,
...
-            # Interpolated quantile
             else:
-                vlow = skiplist.get(idx)
-                vhigh = skiplist.get(idx + 1)
-                output[i] = ((vlow + (vhigh - vlow) *
-                             (quantile * (nobs - 1) - idx)))
+                idx_with_fraction = quantile * <double> (nobs - 1)
+                idx = int(idx_with_fraction)
+
+                if interpolation_id == 0:  # linear
```
jreback: so you would use the enums here

pandas/_libs/window.pyx
```diff
@@ -1357,25 +1357,53 @@ cdef _roll_min_max(ndarray[numeric] input, int64_t win, int64_t minp,
...
         bint is_variable
         ndarray[int64_t] start, end
         ndarray[double_t] output
         double vlow, vhigh
+        int interpolation_id
 
     if quantile <= 0.0 or quantile >= 1.0:
         raise ValueError("quantile value {0} not in [0, 1]".format(quantile))
 
+    # interpolation_id is needed to avoid string comparisons inside the loop
```
jreback: don't need the comments here

pandas/core/window.py
```diff
@@ -1246,9 +1246,21 @@ def kurt(self, **kwargs):
...
-    def quantile(self, quantile, **kwargs):
+        This optional parameter specifies the interpolation method to use,
+        when the desired quantile lies between two data points `i` and `j`:
+
+            * linear: `i + (j - i) * fraction`, where `fraction` is the
+              fractional part of the index surrounded by `i` and `j`.
+            * lower: `i`.
+            * higher: `j`.
+            * nearest: `i` or `j` whichever is nearest.
+            * midpoint: (`i` + `j`) / 2.""")
```
jreback: can you add a Returns, Examples and a See Also section (look at the Series.quantile doc-string for inspiration). Also add a reference from the Series.quantile doc-string to here

pandas/tests/test_window.py
```diff
@@ -1135,7 +1135,22 @@ def test_rolling_quantile_series(self):
         s = Series(arr)
         q1 = s.quantile(0.1)
         q2 = s.rolling(100).quantile(0.1).iloc[-1]
+        tm.assert_almost_equal(q1, q2)
```
jreback: can you move to a separate test and parameterize on the interpolation (inlcude linear as well)

jreback: these are all coming to the same result value, is that right?

jreback (changes requested):
looking good. comments and please add a line in the Other Enhancements section.

pandas/core/window.py
```diff
@@ -1246,9 +1246,50 @@ def kurt(self, **kwargs):
...
+    Examples
+    --------
+    >>> s = Series([1, 2, 3, 4])
+    >>> s.rolling(2).quantile(.4, interpolation='lower')
+    0    NaN
+    1    1.0
+    2    2.0
+    3    3.0
+    dtype: float64
+    >>> s.rolling(2).quantile(.4, interpolation='midpoint')
```
jreback: can you add a blank line here

jreback (changes requested):
can you add a whatsnew note (other enhancements section)

jreback:
linting issue:

```pandas/core/window.py:1289:1: W293 blank line contains whitespace```

you can run
``LINT=TRUE ci/lint.sh`` to see

**Comment by jreback** [comment]
@TomAugspurger comments?

**Comment by jreback** [comment]
@WillAyd if you'd have a look

File: pandas/_libs/window.pyx [hunk 7]
```diff
@@ -1357,25 +1357,49 @@ cdef _roll_min_max(ndarray[numeric] input, int64_t win, int64_t minp,
...
         ndarray[int64_t] start, end
         ndarray[double_t] output
         double vlow, vhigh
+        InterpolationType interpolation_type
 
     if quantile <= 0.0 or quantile >= 1.0:
         raise ValueError("quantile value {0} not in [0, 1]".format(quantile))
 
+    try:
+        interpolation_type = interpolation_types[interpolation]
```
WillAyd: Wouldn't this raise a `KeyError` not a `ValueError`?

pandas/_libs/window.pyx
```diff
@@ -1414,18 +1438,31 @@ def roll_quantile(ndarray[float64_t, cast=True] input, int64_t win,
...
                 output[i] = skiplist.get(0)
-
-            # Interpolated quantile
             else:
-                vlow = skiplist.get(idx)
-                vhigh = skiplist.get(idx + 1)
-                output[i] = ((vlow + (vhigh - vlow) *
-                             (quantile * (nobs - 1) - idx)))
+                idx_with_fraction = quantile * <double> (nobs - 1)
+                idx = int(idx_with_fraction)
```
WillAyd: Given everything else here is done with C syntax, can we use that casting convention here?

pandas/tests/test_window.py
```diff
@@ -1138,6 +1138,20 @@ def test_rolling_quantile_series(self):
 
         tm.assert_almost_equal(q1, q2)
 
+    @pytest.mark.parametrize('quantile', [0.0, 0.1, 0.45, 0.5, 1])
```
WillAyd: Can we add examples with NA data?

WillAyd (changes requested):
Couple comments on my end - is it also possible to use nogil? May help with some of the performance regression?

kornilova-l:
@WillAyd I tried to replace `range` with `prange` but it does not compile because calling `skiplist.get` requires gil: `Calling gil-requiring function not allowed without gil` did I miss something?

jreback:
ok we don't / can't use ``prange`` anywhere in the codebase. only you can use ``nogil``

kornilova-l:
@jreback but still using `nogil` is not possible because there are `gil-requiring` function calls

WillAyd:
The only gil-requiring function call at the time of that comment was `int` which is why I asked if it was possible to use the C-style cast instead. I think if you did those in combination it would work. I see that you since added the `round` built-in so to get `nogil` to work you'd have to use a C-style version of that as well. 

I did something similar in https://github.com/pandas-dev/pandas/pull/20405/files which you may be able to reference

pandas/tests/test_window.py
```diff
@@ -1130,14 +1130,41 @@ def test_rolling_quantile_np_percentile(self):
...
+    @pytest.mark.parametrize('na_probability', [0.0, 0.3])
+    @pytest.mark.parametrize('interpolation', ['linear', 'lower', 'higher',
+                                               'nearest', 'midpoint'])
+    def test_rolling_quantile_interpolation_options(self, quantile,
+                                                    na_probability,
+                                                    interpolation):
+        # Tests that rolling window's quantile behavior is analogous to
+        # Series' quantile for each interpolation option
+        size = 100
+        s = Series(np.random.rand(size))
```
WillAyd: I'd advise against using random data in a test case as it could cause intermittent failures and make bugs harder to find. To inject missing data, could you not just parametrize the function with a second array of say `[0., np.nan, 0.2, np.nan, 0.4]` to match what you had earlier?

pandas/_libs/window.pyx
```diff
@@ -1357,25 +1357,49 @@ cdef _roll_min_max(ndarray[numeric] input, int64_t win, int64_t minp,
...
         ndarray[double_t] output
         double vlow, vhigh
+        InterpolationType interpolation_type
 
     if quantile <= 0.0 or quantile >= 1.0:
         raise ValueError("quantile value {0} not in [0, 1]".format(quantile))
 
+    try:
+        interpolation_type = interpolation_types[interpolation]
+    except KeyError:
```
WillAyd: Is there a test case to cover that this raises the expected error message when passing an invalid argument? If not can you add?

pandas/tests/test_window.py
```diff
@@ -1130,14 +1130,41 @@ def test_rolling_quantile_np_percentile(self):
...
+            index = np.random.randint(0, size)
+            if not np.isnan(s[index]):
+                s[index] = np.NaN
+                na_count += 1
+
+        q1 = s.quantile(quantile, interpolation)
+        q2 = s.rolling(size, min_periods=1).quantile(
+            quantile, interpolation).iloc[-1]
+
+        tm.assert_almost_equal(q1, q2)
```
WillAyd: Is almost equal required here or is it possible to use input data so that float precision would not be an issue? I'm trying to be extra cautious that `almost_equal`doesn't allow bugs to silently pass through some interpolation options

WillAyd (changes requested):

kornilova-l:
About `nogil`: I changed `int()` to `<int>` and also removed `round()` function, but after adding `nogil` compiler still complains about `skiplist` methods (`skiplist.insert`, `skiplist.remove`, `skiplist.get`).

One of compiler messages:
```
Error compiling Cython file:
------------------------------------------------------------
...
                        elif idx_with_fraction - idx < 0.5:
                            output[i] = skiplist.get(idx)
                        else:
                            output[i] = skiplist.get(idx + 1)
                    elif interpolation_type == MIDPOINT:
                        vlow = skiplist.get(idx)
                                          ^
------------------------------------------------------------

pandas/_libs/window.pyx:1474:43: Calling gil-requiring function not allowed without gil
```

pandas/_libs/window.pyx
```diff
@@ -1414,18 +1438,42 @@ def roll_quantile(ndarray[float64_t, cast=True] input, int64_t win,
...
+                        if idx % 2 == 0:
+                            output[i] = skiplist.get(idx)
+                        else:
+                            output[i] = skiplist.get(idx + 1)
+                    elif idx_with_fraction - idx < 0.5:
+                        output[i] = skiplist.get(idx)
+                    else:
+                        output[i] = skiplist.get(idx + 1)
+                elif interpolation_type == MIDPOINT:
+                    vlow = skiplist.get(idx)
```
jreback: to make this work w/o the gil,  you need to call ``skiplist_get(skiplist, idx)`` (there are examples in the file). ``skiplist.get`` invokes a python function call and is not allowed

kornilova-l:
Finally code works with `nogil`

There is another problem:
Some tests in python 2.7 environment fail. All failed tests use `midpoint` interpolation. I am almost sure that it is related to https://github.com/numpy/numpy/issues/7163 
Tests compare results of `quantile` and `rolling_quantile`. `quantile` uses `np.percentile` which used to have a bug in 'midpoint' interpolation.

What should I do with failed tests? Is it okay to test `midpoint` separately and not compare it to `quantile`?

jreback:
just skip if < numpy 1.12 on those tests

pandas/tests/test_window.py
```diff
@@ -6,6 +6,7 @@
 from datetime import datetime, timedelta
 from numpy.random import randn
 import numpy as np
+from scipy._lib._version import NumpyVersion
```
jreback: use ``from pandas import _np_version_under1p12``

jreback (changes requested):

pandas/tests/test_window.py
```diff
@@ -1130,14 +1131,49 @@ def test_rolling_quantile_np_percentile(self):
...
+                                                    interpolation, data):
+        # Tests that rolling window's quantile behavior is analogous to
+        # Series' quantile for each interpolation option
+        s = Series(data)
+
+        q1 = s.quantile(quantile, interpolation)
+        q2 = s.rolling(len(data), min_periods=1).quantile(
+            quantile, interpolation).iloc[-1]
+
+        if np.isnan(q1):
```
WillAyd: Is `tm.assert_series_equal` not an option here?

pandas/tests/test_window.py
```diff
@@ -1130,14 +1131,49 @@ def test_rolling_quantile_np_percentile(self):
...
+    @pytest.mark.parametrize('quantile', [0.0, 0.1, 0.45, 0.5, 1])
+    @pytest.mark.parametrize('interpolation', ['linear', 'lower', 'higher',
+                                               'nearest', 'midpoint'])
+    @pytest.mark.parametrize('data', [[1., 2., 3., 4., 5., 6., 7.],
+                                      [8., 1., 3., 4., 5., 2., 6., 7.],
+                                      [0., np.nan, 0.2, np.nan, 0.4],
+                                      [np.nan, np.nan, np.nan, np.nan],
+                                      [np.nan, 0.1, np.nan, 0.3, 0.4, 0.5],
+                                      [0.5], [np.nan, 0.7, 0.5]])
+    def test_rolling_quantile_interpolation_options(self, quantile,
```
WillAyd: I think this and the above test cover practically the same use case. If that's the case I'd get rid of the test above

pandas/_libs/window.pyx
WillAyd: Minor nit but can you place the name of the passed interpolation in single quotes? Helps distinguish it from the rest of the text in the error message (will need to update test as well)

WillAyd (changes requested):

pandas/tests/test_window.py
kornilova-l: @WillAyd, it is not series type. I edited test data a bit and was able to get rid of `round`

pandas/core/window.py
```diff
@@ -1246,9 +1246,52 @@ def kurt(self, **kwargs):
...
-    def quantile(self, quantile, **kwargs):
+        This optional parameter specifies the interpolation method to use,
+        when the desired quantile lies between two data points `i` and `j`:
+
+            * linear: `i + (j - i) * fraction`, where `fraction` is the
+              fractional part of the index surrounded by `i` and `j`.
+            * lower: `i`.
+            * higher: `j`.
+            * nearest: `i` or `j` whichever is nearest. Implementation uses
+              round() built-in.
```
WillAyd: Can remove this comment about the round() built-in

pandas/core/window.py
```diff
@@ -1246,9 +1246,52 @@ def kurt(self, **kwargs):
...
+            * midpoint: (`i` + `j`) / 2.
+
+    Returns
+    -------
+    Series or DataFrame
+        Returned object type is determined by the caller of the %(name)s
+        calculation
```
WillAyd: *Very minor* but description should end with a period

pandas/core/window.py
```diff
@@ -1246,9 +1246,52 @@ def kurt(self, **kwargs):
...
+    >>> s.rolling(2).quantile(.4, interpolation='midpoint')
+    0    NaN
+    1    1.5
+    2    2.5
+    3    3.5
+    dtype: float64
+
+    See Also
+    --------
+    pandas.Series.quantile
```
WillAyd: Just add some quick descriptions after these (separated by colon). Can reference docstring guide:

https://python-sprints.github.io/pandas/guide/pandas_docstring.html#section-5-see-also

pandas/tests/test_window.py
```diff
@@ -1128,15 +1129,40 @@ def test_rolling_quantile_np_percentile(self):
...
+                                      [np.nan, 0.1, np.nan, 0.3, 0.4, 0.5],
+                                      [0.5], [np.nan, 0.7, 0.6]])
+    def test_rolling_quantile_interpolation_options(self, quantile,
+                                                    interpolation, data):
+        # Tests that rolling window's quantile behavior is analogous to
+        # Series' quantile for each interpolation option
+        s = Series(data)
+
+        q1 = s.quantile(quantile, interpolation)
+        q2 = s.rolling(len(data), min_periods=1).quantile(
```
WillAyd: I suppose this is just `expanding` instead of `rolling` so should use the former to stay idiomatic

WillAyd (changes requested):
Couple minor edits otherwise lgtm. Can you post updated ASVs?

asv_bench/benchmarks/rolling.py
```diff
@@ -68,9 +69,21 @@ class Quantile(object):
...
     def setup(self, constructor, window, dtype, percentile):
-        N = 10**5
+        N = 10 ** 5
         arr = np.random.random(N).astype(dtype)
         self.roll = getattr(pd, constructor)(arr).rolling(window)
 
     def time_quantile(self, constructor, window, dtype, percentile):
         self.roll.quantile(percentile)
+
+    def time_quantile_nearest(self, constructor, window, dtype, percentile):
```
WillAyd: Ha when I asked for updated ASVs I was just asking for a refresh of what you posted earlier in the thread but I suppose this is better. Is it possible to just add interpolation as an argument to `setup`? Would be more concise.

Here's an example from GroupBy that you could use for reference, though the existing Quantile class could clue you in on how to do this as well:

https://github.com/pandas-dev/pandas/blob/78fee04e95e3c53c83c938285580c39e7761ddc8/asv_bench/benchmarks/groupby.py#L342

WillAyd: Please also post results to a comment when done. I'm assuming that these will fail on master but would still be good to know how they baseline against the current implementation that you posted earlier in thread

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -402,6 +402,7 @@ Other Enhancements
 - :meth:`DataFrame.to_sql` now performs a multivalue insert if the underlying connection supports itk rather than inserting row by row.
   ``SQLAlchemy`` dialects supporting multivalue inserts include: ``mysql``, ``postgresql``, ``sqlite`` and any dialect with ``supports_multivalues_insert``. (:issue:`14315`, :issue:`8953`)
 - :func:`read_html` now accepts a ``displayed_only`` keyword argument to controls whether or not hidden elements are parsed (``True`` by default) (:issue:`20027`)
+- :meth:`Rolling.quantile` and :meth:`Expanding.quantile` now accept ``interpolation`` keyword
```
jreback: @jorisvandenbossche do these refs work?

jreback: can you add (this) issue number here

jreback (changes requested):
can you rebase and address small comments by @WillAyd, otherwise lgtm.

doc/source/whatsnew/v0.23.0.txt
kornilova-l: @jreback, how should they work? I built documentation and all references in the whatsnew.html are rendered as plain bold text.

kornilova-l:
Now there are lots of commits from master :( hope I did rebasing correctly

WillAyd:
Shouldn't be including the commits of other contributors - how did you rebranch this? Need to figure out how to undo that and only replay your commits on top of master

kornilova-l:
Finally I did rebasing correctly (I hope)
Should I squash these 13 commits into one?

jreback (approved):

jreback:
thanks @kornilova-l nice patch!

generally you don't need to squash, you can do it if you want to make things more readable / easier if you want. but not necessary to merge.

PR #20502, opened 2018-03-27
ENH: Support ExtensionArray in Groupby

TomAugspurger:
```python
In [1]: import pandas as pd

In [2]: from cyberpandas import IPArray

In [3]: df = pd.DataFrame({"A": IPArray([0, 0, 1, 2, 2]), "B": [1, 5, 1, 1, 3]})

In [4]: df
Out[4]:
         A  B
0  0.0.0.0  1
1  0.0.0.0  5
2  0.0.0.1  1
3  0.0.0.2  1
4  0.0.0.2  3

In [5]: df.groupby("A").B.mean()
Out[5]:
A
0.0.0.1    1
0.0.0.2    2
Name: B, dtype: int64
```

Note that right now `Out[5].index` just just an `Index` with object dtype. In the future, we could tie an Index type to an ExtensionArray type, and ensure that the extension type propagates through.

### DIFF
--- a/pandas/core/groupby.py
+++ b/pandas/core/groupby.py
@@ -44,7 +44,7 @@ from pandas.core.base import (PandasObject, SelectionMixin, GroupByError,
                               DataError, SpecificationError)
 from pandas.core.index import (Index, MultiIndex,
                                CategoricalIndex, _ensure_index)
-from pandas.core.arrays import Categorical
+from pandas.core.arrays import ExtensionArray, Categorical
 from pandas.core.frame import DataFrame
 from pandas.core.generic import NDFrame, _shared_docs
 from pandas.core.internals import BlockManager, make_block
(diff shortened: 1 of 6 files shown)


TomAugspurger:
What I have so far is relatively straightforward (surprisingly). But I'm probably missing things. Are there edge cases or other operations we should test?

jorisvandenbossche:
If you use `as_index=False`, can you in that way ensure it keeps the correct extension dtype?

TomAugspurger:
Not easily. By the time we're wrapping up the output, we've long since converted the input to an Index.

That said, once we have ExtensionIndexes, it should be a one-line change: 
https://github.com/pandas-dev/pandas/blob/cdfce2b0ad99f7faad57cc5247cf33aab5725bed/pandas/core/groupby.py#L3037

jreback:
lgtm. and another reason to make EA a base class for ``Index``.

PR #20506, opened 2018-03-27
Move GroupBy to Submodule and Add FutureWarning

WillAyd:
ref #20485 @jreback


pandas/tests/api/test_api.py
```diff
@@ -251,3 +251,16 @@ def test_categorical_move(self):
 
         with tm.assert_produces_warning(FutureWarning):
             from pandas.core.categorical import CategoricalDtype  # noqa
+
+    @pytest.mark.parametrize("attr", [
+        "Grouper", "groupby", "BinGrouper", "Grouper", "_GroupBy", "GroupBy",
+        "SeriesGroupBy", "_pipe_template", "PanelGroupBy", "Grouping",
+        "SpecificationError", "DataError", "generate_bins_generic"])
+    def test_groupby_move(self, attr):
```
WillAyd: This would be straightforward to combine with the method above if we wanted one test accepting a module / list of objects to test the deprecation for

WillAyd:
There was a failure in one of the "Allowed Failures" as a result of a `ResourceWarning` for an open TextIOBuffer (?). I don't believe that is related to this change and it didn't appear in any of the previous CI runs

jreback:
> There was a failure in one of the "Allowed Failures" as a result of a ResourceWarning for an open TextIOBuffer (?). I don't believe that is related to this change and it didn't appear in any of the previous CI runs

FYI there are NO allowed failures, meaning that yes these will allow the build to complete earlier, but generally we don't merge things if they are causing issues.

pandas/core/groupby/__init__.py
```diff
@@ -0,0 +1,20 @@
+import warnings
+
+# TODO: Remove after 0.XX.x
+warnings.warn("'pandas.core.groupby' has moved. Use "
+              "'pandas.core.groupby.groupby' instead", FutureWarning,
+              stacklevel=2)
+
+from pandas.core.groupby.groupby import Grouper  # noqa: F401
+from pandas.core.groupby.groupby import groupby  # noqa: F401
+from pandas.core.groupby.groupby import BinGrouper  # noqa: F401
```
jreback: only put public functions here. If something is needed internally then change the reference to it.

jreback (changes requested):

TomAugspurger:
FYI, merge conflict with master.

WillAyd:
Hmm so I'm not sure that the FutureWarning being in the `__init__.py` of the groupby package is ideal here because the internal references point to `pandas.core.groupby.groupby`. I think warning in that init will always get thrown, because the new module is part of the package where that `__init__.py` exists, so it would always get loaded?

`pandas.core.categorical` didn't have this problem because the updated reference was higher in the package hierarchy. The deprecated module doesn't get loaded when using the newer reference. Thinking through what our options here are but open to ideas

jorisvandenbossche:
Can we just import what is needed in `__init__.py`? Then no deprecation warning is needed (python doesn't care if it is a single file module or a package directory, so we can make use of that).
Personally I would even also use those imports internally. Then you don't need to change all the imports in the rest of pandas, if you reorganize things internally within the ``groupby`` subpackage.

WillAyd:
@jorisvandenbossche I already updated all the internal refs per Jeff's previous comment, but could undo if required. That said, maybe we just shouldn't put the warning on the __init__ given the change could be transparent from an API perspective since all we are doing is moving from a single module to a like-named package?

pandas/core/groupby/__init__.py
```diff
@@ -0,0 +1,9 @@
+import warnings
+
+# TODO: Remove after 0.XX.x
+warnings.warn("'pandas.core.groupby' has moved. Use "
+              "'pandas.core.groupby.groupby' instead", FutureWarning,
+              stacklevel=2)
+
+from pandas.core.groupby.groupby import (Grouper, GroupBy,  # noqa: F401
```
WillAyd: GroupBy, SeriesGroupBy and DataFrameGroupBy were all added here as they were causing downstream failures for Dask, even though they aren't explicitly referenced internally. Once we settle on a go-forward here we'll have to follow up with the Dask team to let them know of impacts

jorisvandenbossche:
> That said, maybe we just shouldn't put the warning on the init given the change could be transparent from an API perspective since all we are doing is moving from a single module to a like-named package?

Yes, that's exactly what I tried to say :-) 

IMO, this is much cleaner (also internally, but certainly for external packages)

pandas/core/groupby/__init__.py
jreback: yes this is fine. These are basically downstream dev dependencies (e.g. you may want to do something based on the object if its a SeriesGroupBy or a DataFrameGroupBy) so ok with this

pandas/core/resample.py
```diff
@@ -7,9 +7,9 @@
 import pandas as pd
 from pandas.core.base import GroupByMixin
 
-from pandas.core.groupby import (BinGrouper, Grouper, _GroupBy, GroupBy,
-                                 SeriesGroupBy, groupby, PanelGroupBy,
-                                 _pipe_template)
+from pandas.core.groupby.groupby import (BinGrouper, Grouper, _GroupBy,
```
jreback: style nit, I like to see a newline after the parens so these are wrapped a bit

jreback (approved):

jreback:
very tiny style issue, ping on green.

jorisvandenbossche:
@jreback did you see our discussion above?

So -1 on merge on green. There is no need to deprecate this (and we can still move to a subpackage)

jreback:
@jorisvandenbossche sorry, I did see that. yeah we can not deprecate is fine.

WillAyd:
Does anyone develop on Windows? Trying to figure out why CircleCI is failing but can't reproduce the issue locally. 

An easy way to solve it would be to add BinGrouper back to the API in the `__init__.py` but that doesn't seem required for anything else internal, so figure it's best to figure out what is different about the Windows import mechanism here

jorisvandenbossche:
CircleCI is also linux, not windows. But why it is failing there and not in the others, I don't know. 
But you can always import `from pandas. .... import BinGrouper` and then test against that, that should certainly work.

I would also like to advocate to keep the internal imports the same as before (so import into `__init__.py` everything that is "public" for the rest of pandas)

WillAyd:
@jreback what do you think about Joris' comment? Should I put everything back into the `__init__.py` and undo the changes to internal references or simply update this last failing test case?

jreback:
i would move non public this out of the __init__

jreback:
thanks @WillAyd

PR #20510, opened 2018-03-28
API/BUG: Enforce "normalized" pytz timezones for DatetimeIndex

### LINKED ISSUE #3746: ERR: disallow direct tz setting
related to #3714
tz should not be settable

```
df=pd.DataFrame({"events":np.ones(len(dates))},index=motor["timestamps"].astype("datetime64[s]"))
df.describe()
```

|  | events |
| --- | --- |
| count | 2393 |
| mean | 1 |
| std | 0 |
| min | 1 |
| 25% | 1 |
| 50% | 1 |
| 75% | 1 |
| max | 1 |

```
df

DatetimeIndex: 2393 entries, 2013-05-30 11:58:36 to 2013-06-03 13:30:43
Data columns (total 1 columns):
events    2393  non-null values
dtypes: float64(1)


df.index.tz='Asia/Kolkata'
df.index

<class 'pandas.tseries.index.DatetimeIndex'>
[2013-05-30 17:28:36, ..., 2013-06-03 19:00:43]
Length: 2393, Freq: None, Timezone: Asia/Kolkata

df2=df[df.index.day==3]

AttributeError                            Traceback (most recent call last)
<ipython-input-39-23de8e1402be> in <module>()
----> 1 df2=df[df.index.day==3]

/usr/local/lib/python2.7/dist-packages/pandas/tseries/index.pyc in f(self)
 39             utc = _utc()
 40             if self.tz is not utc:
---> 41                 values = self._local_timestamps()
 42         return tslib.get_date_field(values, field)
 43     f.__name__ = name

/usr/local/lib/python2.7/dist-packages/pandas/tseries/index.pyc in     _local_timestamps(self)
399             values = self.asi8
400             indexer = values.argsort()
--> 401             result = tslib.tz_convert(values.take(indexer), utc, self.tz)
402 
403             n = len(indexer)

/usr/local/lib/python2.7/dist-packages/pandas/tslib.so in pandas.tslib.tz_convert     (pandas/tslib.c:20949)()

/usr/local/lib/python2.7/dist-packages/pandas/tslib.so in pandas.tslib._get_deltas (pandas/tslib.c:22754)()

/usr/local/lib/python2.7/dist-packages/pandas/tslib.so in pandas.tslib._get_utcoffset (pandas/tslib.c:11327)()

AttributeError: 'str' object has no attribute 'utcoffset'
```

A temporary fix which i did was to add seconds to data manually while creating the index.

### LINKED ISSUE #13238: Resample discards timezone information
It appears that resample is now dropping timezone information on the index. Is this expected?

```
a = pp.Series(index=pp.DatetimeIndex(['2013-01-01 06:00', '2013-01-01 07:00', '2013-01-02 06:00'],
                                     tz=pytz.timezone('America/Los_Angeles')),
              data=[1, 2, 3])
b = a.resample('D', how='max')
print repr(a.index.tz)
print repr(b.index.tz)
print a.index.tz == b.index.tz
```

produces the following in 0.16.1. Notably the timezones for the original series and the resampled series are the same.

```
<DstTzInfo 'America/Los_Angeles' LMT-1 day, 16:07:00 STD>
<DstTzInfo 'America/Los_Angeles' LMT-1 day, 16:07:00 STD>
True
```

However that's not the case for 0.18.1. Caveat is that we're using the new resample syntax: `b = a.resample('D').max()`

```
<DstTzInfo 'America/Los_Angeles' LMT-1 day, 16:07:00 STD>
<DstTzInfo 'America/Los_Angeles' PST-1 day, 16:00:00 STD>
False
```

### LINKED ISSUE #18595: Add notion of pytz time zone normalization
Though part of the solution to #18523 will be to introduce a new time zone comparison operation, there's another, less pressing issue that's also brought up there:

```python
import pandas as pd

idx1 = pd.date_range('2011-01-01', periods=3, freq='H', tz='Europe/Paris')
idx2 = pd.date_range(start=idx1[0], end=idx1[-1], freq='H')

print(idx1.tz)    # <DstTzInfo 'Europe/Paris' LMT+0:09:00 STD>
print(idx2.tz)   # <DstTzInfo 'Europe/Paris' CET+1:00:00 STD>
```

The problem comes from the fact that `pytz.localize` spawns concrete offset-specific new `tzinfo` objects for each date that don't compare equal to one another (and aren't the same instance as one another). As such, when you pull `idx1[0]` and `idx1[-1]`, you get concrete timestamps, each of which have the `<DstTzInfo 'Europe/Paris' CET+1:00:00 STD>` zone attached and so the `DateTimeIndex` gets the concrete example, which is not actually consistent with the abstract notion of time zone from `idx1`. At the series level, it makes sense to use a "normalized" version of `pytz` zones, so I recommend adding a function like this:

```python
def tz_normalize(tzi):
    if is_pytz_zone(tzi):
        return pytz.timezone(str(tzi))
    else:
        return tzi
```

You can solve the above problem by either normalizing the time zone on construction *or* you can change `.tz` from being an attribute to being a property:

```python
@property
def tz(self):
    return tz_normalize(self._tz)

@tz.setter
def tz(self, value):
    msg = ('Directly setting DatetimeIndex is deprecated, instead, you should '
           'use tz_localize() or tz_convert() as appropriate')
    raise_deprecation_warning(msg)
    self._tz = value
```

You'll note that in my property implementation, I've demonstrated that this will also allow you to enforce that people use the proper interface for changing index time zones.

Issue raised per discussion with @jreback and @mriduls at PyData NYC sprints.

mroeschke:
closes #3746
closes #18595
closes #13238
- [x] tests added / passed
- [x] whatsnew entry

Addressing 3 birds with 2 stones here. Using @pganssle suggested implementation for a tz property and directly raising an error per #3746 (could depreciate and error in a future version as well, open to feedback on the prefered path of API change)

Additionally, solves the issue of resampling a DataFrame/Series with a DatetimeIndex that retained a local timezone instead of the "LMT" version.


doc/source/whatsnew/v0.23.0.txt
```diff
@@ -916,6 +918,7 @@ Timezones
 - Bug in :func:`DataFrame.diff` that raised an ``IndexError`` with tz-aware values (:issue:`18578`)
 - Bug in :func:`melt` that converted tz-aware dtypes to tz-naive (:issue:`15785`)
 - Bug in :func:`Dataframe.count` that raised an ``ValueError`` if .dropna() method is invoked for single column timezone-aware values. (:issue:`13407`)
+- Bug in :func:`Dataframe.resample` that dropped timezone information (:issue:`13238`)
```
jreback: can you move to resample section

pandas/_libs/tslibs/timezones.pyx
```diff
@@ -314,3 +314,21 @@ cpdef bint tz_compare(object start, object end):
     """
     # GH 18523
     return get_timezone(start) == get_timezone(end)
+
+
+cpdef tz_normalize(object tz):
```
jreback: this is not a great name, normalize has a specific meaning to timestamps. maybe ``tz_standardize``?

pandas/_libs/tslibs/timezones.pyx
```diff
@@ -314,3 +314,21 @@ cpdef bint tz_compare(object start, object end):
...
+    """
+    If the passed tz is a pytz timezone object, "normalize" it to the LMT
+    version
+
+    Parameters
+    ----------
```
jreback: can you add some examples

pandas/tests/indexes/datetimes/test_construction.py
```diff
@@ -441,6 +441,34 @@ def test_000constructor_resolution(self):
...
         assert idx.nanosecond[0] == t1.nanosecond
 
+    def test_disallow_setting_tz(self):
+        # GH 3746
+        dti = DatetimeIndex(['2010'], tz='UTC')
+        with pytest.raises(ValueError):
+            dti.tz = pytz.timezone('US/Pacific')
+
+    @pytest.mark.parametrize('tz', [
+        None, 'America/Los_Angeles', pytz.timezone('America/Los_Angeles'),
```
jreback: can you add a similar battery of these tests to Timestamp as well (disabled setting already is disabled, but see if we have a test for it).

jreback (changes requested):
looks good! some comments

pandas/core/indexes/datetimelike.py
```diff
@@ -1005,7 +1005,7 @@ def shift(self, n, freq=None):
             result = self + offset
 
             if hasattr(self, 'tz'):
```
pganssle: I generally [avoid `hasattr`](https://hynek.me/articles/hasattr/) in projects that support Python 2.

pganssle:
I'm not sure I understand the purpose of making it so you can't set `tz`.

It's also *not* certain that you will get `LMT` for any given zone (though it is likely for certain zones). In general I think [people should not use pytz at all](https://blog.ganssle.io/articles/2018/03/pytz-fastest-footgun.html), but I'm aware that that might be a pretty big API change. Still, I think any sort of datetime implementation should try to *hide* `pytz`'s implementation details, not expose them directly.

jreback:
> I'm not sure I understand the purpose of making it so you can't set tz.

Timestamp and DTI are immutable objects, so setting after creation doesn't make sense.

jreback:
> It's also not certain that you will get LMT for any given zone (though it is likely for certain zones). In general I think people should not use pytz at all, but I'm aware that that might be a pretty big API change. Still, I think any sort of datetime implementation should try to hide pytz's implementation details, not expose them directly.

this is generally true. The actual tz is an implementation detail, which we don't really care / want to expose to the user directly (of course we want them to be able to construct using there favorite tz library & ISO). Users mostly just care about the iso repr anyhow (US/Eastern). After this PR things will more closely follow this pattern.

pganssle:
@jreback Looking more closely, I think I was slightly confused because the comments and documentation are misleading. None of these are testing or ensuring that the tz is "LMT", they are just ensuring that it is a consistent `tzinfo` object. I think it is best to scrub all mentions of LMT from the comments and documentation.

pandas/tests/indexes/datetimes/test_construction.py
```diff
@@ -441,6 +441,34 @@ def test_000constructor_resolution(self):
...
+        tm.assert_index_equal(result, expected)
+        # Especially assert that the timezone is LMT for pytz
+        assert pytz.timezone('America/Los_Angeles') == result.tz
+
+    @pytest.mark.parametrize('tz', ['US/Pacific', 'US/Eastern', 'Asia/Tokyo'])
+    def test_constructor_with_non_normalized_pytz(self, tz):
+        # GH 18595
+        non_norm_tz = Timestamp('2010', tz=tz).tz
+        result = DatetimeIndex(['2010'], tz=non_norm_tz)
+        assert pytz.timezone(tz) == result.tz
```
pganssle: Timezone comparison should use `is`, not `==`. See [this post about aware datetime comparison](https://blog.ganssle.io/articles/2018/02/a-curious-case-datetimes.html) (or [this one about aware datetime semantics in general](https://blog.ganssle.io/articles/2018/02/aware-datetime-arithmetic.html)).

mroeschke:
@pganssle Yeah I was mainly using `LMT` as an alias (and consistent pytz basis) for "the first tzinfo object returned from `pytz.timezone(...)`" since I thought it would always be `LMT`. Since that may not be the case, agreed that I should remove mention of `LMT` to avoid confusion

pganssle:
@mroeschke I don't think "the first tzinfo object" is really the contract either. The point is that `tz` should represent the *canonical* time zone object, to the extent that there is one (any one of them can be chosen as a representative, so long as it's always the same one).

That said, exposing that information anywhere in the public-facing documentation is probably a bad idea, because it's very much a `pytz` concept, and specifically one of the weird ones that violates the `tzinfo` interface. I think it would be better if pandas started moving away from any overt sign that they use pytz and try to move away from using pytz-specific concepts.

pandas/_libs/tslibs/timezones.pyx
```diff
@@ -316,10 +316,10 @@ cpdef bint tz_compare(object start, object end):
     return get_timezone(start) == get_timezone(end)
 
 
-cpdef tz_normalize(object tz):
+cpdef tz_standardize(object tz):
```
pganssle: Does this need to be a `cpdef`? Can it just be `cdef`? I don't know why end users would need to be able to do this.

mroeschke: This function is called within a python file (`pandas/core/indexes/datetimelike.py`), so it can't just be a `cdef`

mroeschke:
Address your comments @jreback.

pandas/core/indexes/datetimelike.py
jreback: yeah this impl is shared by DTI and TDI so this could be restructured a bit. Please file an issue .

pandas/tests/frame/test_alter_axes.py
```diff
@@ -250,7 +250,7 @@ def test_set_index_cast_datetimeindex(self):
         df['C'] = i.to_series().reset_index(drop=True)
         result = df['C']
         comp = pd.DatetimeIndex(expected.values).copy()
```
jreback: can remove the .copy() here

jreback (changes requested):
small changes.

jorisvandenbossche:
> I think a property alone is enough to hide the implementation detail that the time zone object may be different for different datetime objects within the DateTimeIndex.

@pganssle what do you mean exactly here?

@mroeschke great work!

pandas/core/indexes/datetimes.py
```diff
@@ -684,6 +678,17 @@ def _values(self):
...
             return self.values
 
+    @property
+    def tz(self):
+        # GH 18595
+        return timezones.tz_standardize(self._tz)
+
+    @tz.setter
+    def tz(self, value):
+        # GH 3746
```
jreback: can you add the same to Timestamp itself.

jreback (changes requested):

pganssle:
@jorisvandenbossche Not sure what part you are confused about, but my intention was that `tz_standardize ` should be as private as possible, not part of the public interface. Having time zone accessed exclusively through a `tz` property should be sufficient to achieve this, and thus end users should never actually have to call `tz_standardize`.

Ideally, the `tz` property *itself* would just be implemented as a `cpdef` and `tz_standardize` would be a `cdef` not accessible by end users.

jreback:
> Ideally, the tz property itself would just be implemented as a cpdef and tz_standardize would be a cdef not accessible by end users.

@pganssle ALL of this is explicitly private already.

pganssle:
@jreback I think I was at least partially fooled by the fact that it has documentation - it's not always clear to me what ends up getting included in the docs.

Is it "explicitly private" in the sense that it's actually impossible for users to access from the Python API, or in the sense that it's under `_libs`? If the latter, then Hyrum's law indicates that someone will start using it eventually. I'm pretty sure there's a way to implement this as truly private.

That said, if it's a pain don't bother. I wasn't planning on harping on it, just explaining my comment.

jreback:
@pganssle if completely private as its under ``_libs``. we can and do change these things w/o warning. If someone uses it, then tough luck. We have enough issues with the public API.

jreback:
note that ALL functions really should have documentation. private ones as well as public ones. We have been making efforts on this recently. It is just useful for future readers.

pganssle:
Sure, but the same could be said for `dateutil.parser._timelex`, the undocumented private class not in `__all__`, but you were the one who persuaded me to add a deprecation warning to that. I'm suggesting that these things go much easier when the interpreter literally won't let you access the symbols (though this is not always possible).

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -700,6 +700,12 @@ class Timestamp(_Timestamp):
         """
         return self.tzinfo
 
+    @tz.setter
+    def tz(self, value):
+        # GH 3746
```
jorisvandenbossche: Can you put here a more informative comment instead of referring to the issue (if it is needed of course)? 
I think the only reason to have this is to have a more informative error message?

pandas/core/indexes/datetimes.py
```diff
@@ -684,6 +678,17 @@ def _values(self):
         else:
             return self.values
 
+    @property
+    def tz(self):
+        # GH 18595
```
jorisvandenbossche: same here

jorisvandenbossche:
> things go much easier when the interpreter literally won't let you access the symbols (though this is not always possible)

that's not possible in this case, as we need to use `tz_standardize` (internally) in a python file

> Not sure what part you are confused about, but my intention was that tz_standardize should be as private as possible, not part of the public interface. Having time zone accessed exclusively through a tz property should be sufficient to achieve this

Confusion solved :) My starting point was that only `.tz` was already the only public exposure.

That said,  `.tz` *is* public, and returns a pytz instance:

> I think it would be better if pandas started moving away from any overt sign that they use pytz and try to move away from using pytz-specific concepts.

I suppose this would only be solved by no longer returning a pytz instance? This could potentially break peoples codes that use a pytz-specific method of the `tz` object?

mroeschke:
@jorisvandenbossche address your comment and all green.

jreback (changes requested):
lgtm. can u run a quick asv  for this
should be ok but doing a bit more work when returning tz

mroeschke:
I'll try to get those ASV sometime tomorrow; my asv setup still doesn't work and need to look into it.

mroeschke:
I wasn't able to resolve my asv issues, but here's a timeit in the meantime. There appears to be a noticeable slowdown. 

```
In [1]: import pandas as pd

In [2]: dti = pd.DatetimeIndex([pd.Timestamp('2010', tz='US/Pacific')])

# This branch
In [3]: %timeit dti.tz
2.21 µs ± 25.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

#Master
In [4]: %timeit dti.tz
52.8 ns ± 1.37 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

pganssle:
@mroeschke Have you tried using the mechanism I mentioned [in this comment](https://github.com/pandas-dev/pandas/pull/20510#issuecomment-377802224), using `tz_standardize` as a `cdef`? That may end up being faster because it saves on Python function call overhead, but I'm not sure how much faster.

That said, it seems unlikely that anyone is accessing the `tz` property in a tight loop. It's an operation that happens once per index. I suspect spending *too* much time optimizing it might be a waste of time.

pganssle:
Actually, just tried it, a lot of the "slow" part is converting to the "canonical" `pytz` zone. It's much faster if you `tz` refers to a `dateutil` time zone.

Using a `pytz` zone:
```python
# master branch
In [4]: %timeit dti.tz
69.8 ns ± 0.826 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

# cdef tz_standardize
#master
In [8]: %timeit dti.tz
69.5 ns ± 1.25 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [3]: %timeit dti.tz
2.41 µs ± 11.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

# cpdef tz_standardize
In [6]: %timeit dti.tz
2.79 µs ± 113 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

And for a `dateutil` zone:
```python
# cdef tz_standardize
In [25]: %timeit dti.tz
478 ns ± 31.7 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

# cpdef tz_standardize
In [4]: %timeit dti.tz
691 ns ± 4.62 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

Another 190 ns is explained by using a property at all (and thus using the descriptor protocol). When `tz` is set to `return self._tz`, we get:

```python
In [4]: %timeit dti.tz
190 ns ± 1.11 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

I think the remaining timing is just the time it takes to determine if something is "pytz-like". That said, there's not much you can do about that, particularly since most people will be using `pytz` until and unless `pandas` starts to move towards using `dateutil` zones.

mroeschke:
Thanks for the investigation @pganssle! I am getting similar results for the `get_tz` property implementation as the one on this branch.

```
In [1]: import pandas as pd

In [2]:  dti = pd.DatetimeIndex([pd.Timestamp('2010', tz='US/Pacific')])

# tz = property(get_tz)
In [3]: %timeit dti.tz
2.43 µs ± 153 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

For a dateutil tz, this should just some overhead with a `hasattr` check, but you're probably correct that for pytz, a good amount of time is spent constructing the new pytz.

But overall I agree with you that I do not suspect many accessing `.tz` in a tight loop.

pganssle:
> For a dateutil tz, this should just some overhead with a hasattr check, but you're probably correct that for pytz, a good amount of time is spent constructing the new pytz.

It's more than just that, it's also the function call overhead. To the extent that `%timeit` is reliable, inlining the extra call to `tz_standardize` does provide a reasonable (looks like it could be about 30%) speedup when using `dateutil` zones.

jreback:
so what u need to do here is use the cache_only decorator
since these are immutable once constructed they r good

pandas/core/indexes/datetimes.py
```diff
@@ -684,6 +678,11 @@ def _values(self):
         else:
             return self.values
 
+    @cache_readonly
+    def tz(self):
+        # GH 18595
```
jreback: i believe we now have this in setter / getter versions

mroeschke: Hmm, the only caching decorator I could find was implemented here:
https://github.com/pandas-dev/pandas/blob/c7af4ae94092110b1872750d89ce55bff131b39f/pandas/_libs/properties.pyx#L9-L44

jreback:
will this PR still work if instead of calling ``tz_standarize`` on the ``.tz``, you instead set the ``._tz`` with the standardized value? this sidesteps the caching as its not longer needed then.

mroeschke:
Good call @jreback. Additionally, standardizing `._tz` directly allows us to provide a better error message when `.tz` is set.

jreback (approved):

jreback:
very nice @mroeschke ! keep em coming!

PR #17023, opened 2017-07-19
Set pd.options.display.max_columns=0 by default

cbrnr:
Update: Remove everything related to `max_rows` and only deal with `max_columns` in this PR.

Changed `max_columns` to `0` (automatically adapt the number of displayed columns to the actual terminal width) when run in a terminal ~~and `max_rows` to `20` (because I'd like to see the "whole" data frame at a glance like in R's tibble)~~.

 - [x] closes #16579
 - [x] tests added / passed
 - [x] passes ``git diff upstream/master -u -- "*.py" | flake8 --diff``
 - [x] whatsnew entry

Issue #4907, opened 2013-09-20
Add support for a pandasrc


jtratner:
Just using the existing configuration framework, but with a file format like matplotlib uses... See how they do it here: http://matplotlib.org/users/customizing.html.

Plus, we can document all the config options in a single file.

related #2452, #3046

cbrnr:
I'd like to revive this issue. A `pandasrc` file would be very useful for people that do not want to `import pandas` in their IPython startup script.

t-makaro:
I'd like to have a configuration file for pandas. I hate having to always place:
```python
pd.set_option('display.latex.repr', True)
pd.set_option('display.latex.longtable', True)
```
at the start of my notebook especially since I always have to look it up.

Using an ipython startup file is not a solution since I don't want pandas to always import (I also don't want to hide the import. I just want to hide the settings that I use to export to pdf).

benpayne:
I am looking into this implimenting this feature.  I've reviewed the mentioned guidelines for implimenting based on matplotlibrc.  It searches a few spots for the rc file that mostly make sense, but with a few exceptions.  

Steps to find the config file.  
1. check local directory: Seems like a good idea for a projec to override settings
2. checks a env varaible (MATPLOTLIBRC) and if it exists looks for a file at that path $MATPLOTLIBRC/matplotlibrc:  Seems redundant to look for a file at that path.  Why not just have the env variable point to the file.  This would allow you to have several RC files in the same directroy and just change your env to get different behavior.  
3. Look at the users home dir and find a rc file there: No changes to this, makes sense.
4. Check an install file location that will be over written every time the package is installed: This seems like a bad idea.  While I think a example RC file in this location makes sense, something poeple can copy and modify for themselves.  Also something that has comments documenting all the options.  Making this file that has to be parsed for every user that imports pandas seems to be excesive.  Furthermore if developers want to change some defaults of various options, change that in the code, not by a global config file.  So I am planning to drop this step.  

Please let me know if I am missing anything in this analysis.   

Another feature that comes to mind when looking at this is if we should search for the first RC file and stop or this should be a bottom up stack approach?  Basically would a user like to have global settings in there home dir for all projects and then have the ability for a local project to override some settings without replacing the global settings.  Or would this get confusing?  I'd impliment that by parsing every RC we find (3,2,1) and then calling pd.set_option for each setting in the files.  That way if the same setting was set in two files, the last file parsed would be the setting we run with.  I'd like to hear others thoughts on this.

[...]

PR #20514, opened 2018-03-28
Set pd.options.display.max_rows = 20 by default

cbrnr:
As a follow-up to #17023, I propose to change the default number of displayed rows to 20.

- [ ] closes #xxxx
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/options.rst
+++ b/doc/source/options.rst
@@ -352,7 +352,7 @@ display.max_info_rows                   1690785      df.info() will usually show
                                                      and max_info_cols limit this null
                                                      check only to frames with smaller
                                                      dimensions then specified.
-display.max_rows                        60           This sets the maximum number of rows
+display.max_rows                        20           This sets the maximum number of rows
                                                      pandas should output when printing
                                                      out various output. For example,
                                                      this value determines whether the
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -687,6 +687,11 @@ yourself. To revert to the old setting, you can run this line:
 
   pd.options.display.max_columns = 20
 
+Furthermore, the default value for the maximum number of displayed rows is now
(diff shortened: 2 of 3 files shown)


jreback:
can you add a whatsnew section about this

jorisvandenbossche:
See the mail I sent about this on the mailing list: https://mail.python.org/pipermail/pandas-dev/2018-March/000732.html

cbrnr:
Did you get any replies?

I like the idea of two thresholds, but the main idea behind reducing the number of displayed rows is to avoid having to scroll. R (tibble) sets these thresholds to 20 (show all rows up to) and 10 (truncated repr), which always result in reprs that fit within the typical terminal height.

However, RStudio has a `View` function, which opens the full dataframe in a spreadsheet-like view. Pandas doesn't have this, so I understand if people want to see more rows by default. If there were a way to show the desired number of rows somehow - e.g. a new method similar to `head` that doesn't return a dataframe, but just prints out the specified number of rows?

jorisvandenbossche:
Sorry for the late reply. I understand the reason of wanting to fit the output on the screen, but the balance with being able to actually see data is a difficult one. 
What you mention about the RStudio View, in principle this would be possible in eg the notebook (not terminal). There are already nice projects like qgrid.

Would you be interested in implementing the "two thresholds" approach?

cbrnr:
I agree that there is no one size fits all solution. I'm not sure the two thresholds approach is a real improvement over the current situation. I think all of these settings related to personal preferences would be solved if Pandas had a configuration file that is read during import, so I'd rather help in getting #4907 done.

jreback:
@jorisvandenbossche what is the status of this?

datapythonista:
Based on @jorisvandenbossche comments I understand we're not merging this PR, and we'll have a more complex solution. Closing, feel free to reopen if this is not the case.

PR #20522, opened 2018-03-28
PERF: GH2003 Series.isin for categorical dtypes

### LINKED ISSUE #20003: PERF: isin() is slower for categorical data than for integers
#### Problem description

For long series and many categories 'Series.isin()' is slower for `categorical` data rather than for `int64`. If categories are built from strings, then the degradation of the performance is even larger.

```python
import pandas as pd
import numpy as np

N = 3000000
Ncats = 100

cats = pd.Series(['abcdef%d'%_ for _ in range(Ncats)])

df = pd.DataFrame({'A': np.random.randn(N),
                   'B': np.random.randn(N),
                   'C': np.random.randint(0, Ncats, N),
                  })
df['D'] = cats.loc[df['C'].values].values
df['E'] = df['C'].astype('category')
df['F'] = df['D'].astype('category')

sel_codes = [1,2]
sel_cats = cats.loc[sel_codes].values

%timeit inds = df.C.isin(sel_codes)  # int64
%timeit inds = df.E.isin(sel_codes)  # category based on int64
%timeit inds = df.D.isin(sel_cats)  # object / string
%timeit inds = df.F.isin(sel_cats)  # category based on string
```
On my machine:
```
6.25 ms ± 412 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
28.7 ms ± 2.37 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
104 ms ± 4.67 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
142 ms ± 6.88 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Interestingly, if there're many categories to compare with, categorical data is faster, e.g. for
```python
sel_codes = range(90)
sel_cats = cats.loc[sel_codes].values

%timeit inds = df.C.isin(sel_codes)  # int64
%timeit inds = df.E.isin(sel_codes)  # category based on int64
%timeit inds = df.D.isin(sel_cats)  # object / string
%timeit inds = df.F.isin(sel_cats)  # category based on string
```
the timings are:
```
441 ms ± 61 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
422 ms ± 68.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
147 ms ± 7.95 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
171 ms ± 2.39 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

p.s. I'm not sure if such performance issues are worth filing.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Darwin
OS-release: 17.4.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.22.0
pytest: None
pip: 9.0.1
setuptools: 28.8.0
Cython: 0.27.3
numpy: 1.14.0
scipy: 1.0.0
pyarrow: 0.7.1
xarray: None
IPython: 6.2.1
sphinx: None
patsy: 0.4.1
dateutil: 2.6.1
pytz: 2017.3
blosc: None
bottleneck: 1.2.1
tables: 3.4.2
numexpr: 2.6.4
feather: 0.4.0
matplotlib: 2.1.0
openpyxl: None
xlrd: None
xlwt: 1.3.0
xlsxwriter: None
lxml: None
bs4: 4.6.0
html5lib: 1.0b10
sqlalchemy: 1.1.15
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

bourbaki:
I have added a branching for the categorical case in `Series.isin` function.
I have also added a test for the most crucial cases (nans).
closes #20003


TomAugspurger (commented):
I'd prefer to move your categorical isin logic to a new method `Categorical.isin`.

Then in `algorithms.py` you can do

```python
if is_categorical_dtype(values):
    values = getattr(values, '_values', values)  # extract Categorical from Series/Index
    # your code
```

Can you add an ASV for this?

pandas/core/series.py
```diff
@@ -3507,7 +3507,11 @@ def isin(self, values):
         5    False
         Name: animal, dtype: bool
         """
-        result = algorithms.isin(com._values_from_object(self), values)
```
TomAugspurger: I wonder if the `_values_from_object` can be moved to `algorithms.isin`? Then this could just be `result = algorithms.isin(self, values)`

pandas/tests/series/test_analytics.py
```diff
@@ -1255,6 +1255,17 @@ def test_isin_empty(self, empty):
         result = s.isin(empty)
         tm.assert_series_equal(expected, result)
 
+    def test_isin_cats(self):
```
TomAugspurger: This can go in `pandas/tests/category/test_algos.py`.

TomAugspurger:
> I'd prefer to move your categorical isin logic to a new method Categorical.isin.

To be a bit clearer about this. I'd imagine the `Categorical.isin` method doing the codes and categories stuff, to see which codes the items in `values` would map to. Then calling `algorithms.value_counts(codes, code_values)`, where `code_values` is the code that each value maps to (if any).

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -345,6 +345,7 @@ Other Enhancements
   ``SQLAlchemy`` dialects supporting multivalue inserts include: ``mysql``, ``postgresql``, ``sqlite`` and any dialect with ``supports_multivalues_insert``. (:issue:`14315`, :issue:`8953`)
 - :func:`read_html` now accepts a ``displayed_only`` keyword argument to controls whether or not hidden elements are parsed (``True`` by default) (:issue:`20027`)
 - zip compression is supported via ``compression=zip`` in :func:`DataFrame.to_pickle`, :func:`Series.to_pickle`, :func:`DataFrame.to_csv`, :func:`Series.to_csv`, :func:`DataFrame.to_json`, :func:`Series.to_json`. (:issue:`17778`)
+- Performance enhancement for :func:`Series.isin` in the case of categorical dtypes (:issue:`20003`)
```
jschendel: Can you move this to the "Performance Improvements" section? (starts around line 783).

ma3axaka: sure

pandas/tests/series/test_analytics.py
ma3axaka: OK.

ma3axaka:
@TomAugspurger Why do I need to call `algorithms.value_counts(codes, code_values)` in `Categorical.isin`

TomAugspurger:
Sorry, I meant `isin` :)

pandas/core/algorithms.py
```diff
@@ -403,8 +403,15 @@ def isin(comps, values):
     if not isinstance(values, (ABCIndex, ABCSeries, np.ndarray)):
         values = construct_1d_object_array_from_listlike(list(values))
 
-    comps, dtype, _ = _ensure_data(comps)
-    values, _, _ = _ensure_data(values, dtype=dtype)
+    if not is_categorical_dtype(comps):
```
jreback: reverse this logic here

jreback (changes requested):
does this have an associated issue? this would need asv benchmarks

pandas/core/algorithms.py
ma3axaka: Yeah. I am working on asv benchmark

ma3axaka:
@TomAugspurger I have moved the logic to `Categorical.isin`. But I still need to do more careful handling of `values` argument of that function. What is the current possible argument types that can be used in `Series.isin` for example?

TomAugspurger:
Merged in master. That'll hopefully fix the circleCI failure.

TomAugspurger:
@ma3axaka did you run the ASV benchmarks? Or just a simple `%timeit` before / after? How does the performance look?

TomAugspurger:
> What are the possible argument types that can be used in Series.isin for example?

It looks like it can be just about anything.  Did you run into issues without using `_sanitize_array`?

ma3axaka:
Yep. I have run the `asv` benchmark. I got the positive result. Do I need to post them here?

TomAugspurger:
Sure (just the new one). I'm curious to see the performance improvement.

TomAugspurger:
Can you also add a release note (doc/source/whatsnew/v0.23.0.txt) noting the performance improvement? Make sure to pull my commit first.

ma3axaka:
I've already added a line about the performance improvement to `whatsnew`. Do I need to add another one?

TomAugspurger:
Sorry, missed that.

ma3axaka:
That's what I got on my laptop:
```
     [fac2ef1b]       [d6c39534]
-         230±6ms      17.6±0.01ms     0.08  categoricals.IsIn.time_isin_categorical_strings
```

jreback (changes requested):

pandas/core/series.py
```diff
@@ -3564,7 +3564,10 @@ def isin(self, values):
...
+        if is_categorical_dtype(self):
+            result = self._values.isin(values)
```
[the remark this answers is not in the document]
ma3axaka: @jreback What's the purpose? Is it better for performance?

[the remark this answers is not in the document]
jreback: its more generic. you don't have to *know* its a categorical here.

ma3axaka:
@jreback Changed the condition to `hasattr`

TomAugspurger:
Sorry, I just noticed an issue. You've made the changes in `Series.isin`, so that the performance boost won't apply to other cases, most notably `CategoricalIndex.isin`.

We have a couple ways to proceed. IMO, I think it's best for the container classes `Series.isin` and `Index.isin` to be as simple as possible. Really just calling `algos.isin(self, values)` *without coercing the data.*

In `algos.isin` we'll unbox the series / index to get the underlying array (numpy array or categorical).

If we find a Categorical, we dispatch to `Categorical.isin`, which extracts the codes, and calls algos.isin again with the codes. Just like you've done.

So the order would be Series.isin -> algos.isin(categorical) -> Categorical.isin -> algos.isin(categorical).

In terms of code changes, it'll be

1. Ensure that `algos.isin` dispatches to `Categorical.isin`
2. Ensure that `algos.isin` does the `_values_from_object` stuff that's currently in `Series.isin` and `Index.isin`. (I think it's already doing that in `ensure_data`, so we may have been doing it twice originally.


Does that make sense?

ma3axaka:
Yes. It makes sense. I will try to refactor it the way you've proposed.

pandas/core/series.py
jreback: yes let's try to do this, @ma3axaka can you make this change

pandas/core/arrays/categorical.py
```diff
@@ -2216,6 +2218,15 @@ def _concat_same_type(self, to_concat):
     def _formatting_values(self):
         return self
 
+    def isin(self, values):
+        from pandas.core.series import _sanitize_array
```
jreback: can you add a doc-string here

pandas/core/arrays/categorical.py
```diff
@@ -2216,6 +2218,15 @@ def _concat_same_type(self, to_concat):
...
+        values = _sanitize_array(values, None, None)
+        null_mask = isna(values)
+        code_values = self.categories.get_indexer(values)
```
jreback: @TomAugspurger is there an EA version of this? (meaning API)  (maybe should raise NotImplementedError which could be caught with a default implementation)

jreback (changes requested):

pandas/core/arrays/categorical.py
ma3axaka: Sure

pandas/core/arrays/categorical.py
TomAugspurger: There isn't yet, but I was thinking about add it. I actually define it on IPArray, since an IP address can be "in" a network.

I'll open an issue (shouldn't delay the changes here).

TomAugspurger (commented):
This is good once Categorical.isin has a docstring. Thanks @ma3axaka.

TomAugspurger:
If you're feeling ambitious, you can make a shared doc between algorithms.isin & Categorical.isin.

ma3axaka:
@TomAugspurger I cannot reproduce CI errors locally. What is the best way to proceed here?

TomAugspurger:
Something strange with Py2.

```python
(Pdb) code_values[null_mask | (code_values >= 0)]
array([1, 1])
```

The reason is that `null_mask` is boolean series, since `values` was a `Categorical` (same would happen with a Series).

```python
(Pdb) null_mask
0    False
1    False
dtype: bool
```

Casting it to an array seems to help

```python
(Pdb) code_values[np.asarray(null_mask) | (code_values >= 0)]
array([0, 1])
```

ma3axaka:
I would prefer to convert `values` to `ndarray` before this logic. Is there any function I could use for it?

jreback:
use np.asarray as
tom suggests

ma3axaka:
@TomAugspurger @jreback I have added the docs and fixed the problem with Python 2 (thanks to @TomAugspurger). Is there something else to be done?

pandas/tests/categorical/test_algos.py
```diff
@@ -47,3 +47,24 @@ def test_factorized_sort_ordered():
 
     tm.assert_numpy_array_equal(labels, expected_labels)
     tm.assert_categorical_equal(uniques, expected_uniques)
+
+
+def test_isin_cats():
+    cat = pd.Categorical(["a", "b", np.nan])
+
```
jreback: can u add the issue number here as a comment

asv_bench/benchmarks/categoricals.py
```diff
@@ -148,3 +148,18 @@ def time_rank_int_cat(self):
 
     def time_rank_int_cat_ordered(self):
         self.s_int_cat_ordered.rank()
+
+
+class IsIn(object):
```
jreback: can u make Isin

pandas/core/algorithms.py
```diff
@@ -407,6 +407,12 @@ def isin(comps, values):
     if not isinstance(values, (ABCIndex, ABCSeries, np.ndarray)):
         values = construct_1d_object_array_from_listlike(list(values))
 
+    if is_categorical_dtype(comps):
```
jreback: if u change this to is_extension_type does everything still work?

jreback (changes requested):

ma3axaka:
@jreback I have renamed the benchmark class and a reference to the issue. Switching to `is_extension_type` doesn't work. It fails on `SparseArray` for example the class doesn't have `isin` method. I am assuming implementing these methods is the goal of #20617?

asv_bench/benchmarks/categoricals.py
```diff
@@ -148,3 +148,18 @@ def time_rank_int_cat(self):
...
+    goal_time = 0.2
+
+    def setup(self):
+        n = 5 * 10**5
+        sample_size = 100
+        arr = ['s%04d' % i for i in np.random.randint(0, n // 10, size=n)]
+        self.sample = np.random.choice(arr, sample_size)
+        self.ts = pd.Series(arr).astype('category')
+
+    def time_isin_categorical_strings(self):
```
jreback: there are 4 cases in the original issue can you cover them

pandas/core/algorithms.py
jreback: ok, can you add a note: TODO(extension) here

pandas/core/arrays/categorical.py
```diff
@@ -2216,6 +2218,60 @@ def _concat_same_type(self, to_concat):
...
+        a list of one element instead:
+
+        >>> s.isin(['lama'])
+        array([ True, False,  True, False,  True, False])
+        """
+        if not is_list_like(values):
+            raise TypeError("only list-like objects are allowed to be passed"
+                            " to isin(), you passed a [{values_type}]"
+                            .format(values_type=type(values).__name__))
+        from pandas.core.series import _sanitize_array
```
jreback: can you move the import to top of function

pandas/tests/categorical/test_algos.py
```diff
@@ -47,3 +47,25 @@ def test_factorized_sort_ordered():
...
+def test_isin_cats():
+    # GH2003
+    cat = pd.Categorical(["a", "b", np.nan])
+
+    result = cat.isin(["a", np.nan])
+    expected = np.array([True, False, True], dtype=bool)
+    tm.assert_numpy_array_equal(expected, result)
+
+    result = cat.isin(["a", "c"])
+    expected = np.array([True, False, False], dtype=bool)
```
jreback: there are a couple of tests in pandas/tests/test_algos.py that test cats for isin, can you move here

jreback (changes requested):

ma3axaka:
@jreback I've made requested changes except moving tests to a different file (see comments). But the Travis CI build is failing with S3 error.

ma3axaka:
@jreback There are still errors with S3. Is this OK?

TomAugspurger:
I'm looking into the s3 / moto issues today. Things should be OK on your
end.

ma3axaka:
@TomAugspurger Hi. Is the issue with s3 fixed? Is there something left to do before merging the. ranch?

jreback:
@ma3axaka can you rebase

jreback:
rebased,  so let's merge on green.

jreback (approved):

jreback:
thanks @ma3axaka what looked like an easy issue morphed into code refactoring! thanks for the patch and patience! keep em coming!

PR #20526, opened 2018-03-29
BUG: #19497 FIX. Add tupleize_cols option to internals._transform_index()

### LINKED ISSUE #19497: Bug: rename incapable of accepting tuples as new name
Pandas is incapable of renaming a pandas.Index object with tuples as the new value. Providing a tuple as ``new_name`` in ``pandas.DataFrame.rename({old_name: new_name}, axis="index")`` returns a ``pandas.MultiIndex`` object, and providing it within a singleton tuple returns an undesirable result. See code below (work-around at bottom):...

```python
import pandas as pd
import numpy as np
df = pd.DataFrame(data = np.arange(5), index=[(x, x) for x in range(5)], columns=["Value"])
print(df) # Note that df.index is a pd.Index object of 2-length tuples

# Wish to rename axis label, but keep the same style
df2 = df.rename({(1,1):(1,5)}, axis="index") 

print(df2)  # Woah! - df2.index is of MultiIndex type
print(df2.index) # ... and here's proof

# Maybe I can get around this by passing it as a singleton tuple...
df3 = df.rename({(1,1):((1,5),)}, axis="index") 
print(df3) # ... apparently not
```
Will produce the output:
```
        Value
(0, 0)      0
(1, 1)      1
(2, 2)      2
(3, 3)      3
(4, 4)      4

     Value
0 0      0
1 5      1
2 2      2
3 3      3
4 4      4
MultiIndex(levels=[[0, 1, 2, 3, 4], [0, 2, 3, 4, 5]],
           labels=[[0, 1, 2, 3, 4], [0, 4, 1, 2, 3]])

           Value
(0, 0)         0
((1, 5),)      1
(2, 2)         2
(3, 3)         3
(4, 4)         4
```
Desired/Expected output:
```
        Value
(0, 0)      0
(1, 5)      1
(2, 2)      2
(3, 3)      3
(4, 4)      4
```

#### Problem description

The current behaviour is a problem for two reasons:
1. It is un-intuitive - I can't see why a user would expect renaming an index to change the index's type.
2. There is no way rename Index objects with tuples

I have checked for similar issues by search of the word ``rename``, and at time of writing, pandas 0.22.0 is the latest released version.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.2.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.0-112-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.22.0
pytest: 3.0.3
pip: 9.0.1
setuptools: 28.8.0
Cython: 0.25.1
numpy: 1.11.2
scipy: 0.18.1
pyarrow: None
xarray: None
IPython: 5.1.0
sphinx: 1.4.8
patsy: 0.4.1
dateutil: 2.6.1
pytz: 2016.7
blosc: None
bottleneck: 1.1.0
tables: 3.3.0
numexpr: 2.6.1
feather: None
matplotlib: 1.5.3
openpyxl: 2.4.9
xlrd: 1.0.0
xlwt: 1.1.2
xlsxwriter: 0.9.3
lxml: 3.8.0
bs4: 4.5.1
html5lib: 1.0b10
sqlalchemy: 1.1.3
pymysql: None
psycopg2: None
jinja2: 2.8
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

#### Workaround
The workaround below uses ``set_value`` function which the documentation tells the user to avoid using (unless you really know what you're doing):
```
df.index.set_value(df.index.get_values(), (1,1), (1, 5)) 
df.reset_index(inplace=True)
df.set_index("index", inplace=True)
df.index.name = None # Arguably not necessary...
print(df)
```
Produces the output:
```
        Value
(0, 0)      0
(1, 5)      1
(2, 2)      2
(3, 3)      3
(4, 4)      4
```

charlie0389:
- Closes #19497
- NO TESTS have been added/passed. Currently experiencing issues getting pandas conda environment installed so unable to do this myself. Very simple alteration however...
- Code passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- Implements tupleize_cols option for pandas.core.internals_transform_index(), and provides the capability to manipulate this option using pandas.DataFrame.rename(). 
**NOTE**: for renaming, the keyword argument default is False, because as suggested [here](https://github.com/pandas-dev/pandas/issues/19497), this is unintuitive and unexepected. The keyword argument for the _transform_index() is True however, which is consistent with default pandas.Index() creation behaviour.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3642,6 +3642,10 @@ class DataFrame(NDFrame):
         level : int or level name, default None
             In case of a MultiIndex, only rename labels in the specified
             level.
+        tupleize_cols : boolean, default False
+            In case of an Index, when True, create MultiIndex if possible.
+            False ensures that an Index will not be converted to a
+            MultiIndex if labels are 'rename'd.
 
         Returns
         -------
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -933,6 +933,7 @@ class NDFrame(PandasObject, SelectionMixin):
         inplace = kwargs.pop('inplace', False)
         level = kwargs.pop('level', None)
         axis = kwargs.pop('axis', None)
+        tupleize_cols = kwargs.pop('tupleize_cols', False)
         if axis is not None:
             axis = self._get_axis_number(axis)
 
@@ -971,7 +972,7 @@ class NDFrame(PandasObject, SelectionMixin):
             if level is not None:
                 level = self.axes[axis]._get_level_number(level)
             result._data = result._data.rename_axis(f, axis=baxis, copy=copy,
-                                                    level=level)
+                                                    level=level, tupleize_cols=tupleize_cols)
             result._clear_item_cache()
 
         if inplace:
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -971,8 +971,10 @@ class NDFrame(PandasObject, SelectionMixin):
             baxis = self._get_block_manager_axis(axis)
             if level is not None:
                 level = self.axes[axis]._get_level_number(level)
-            result._data = result._data.rename_axis(f, axis=baxis, copy=copy,
-                                                    level=level, tupleize_cols=tupleize_cols)
+            result._data = \
+                result._data.rename_axis(f, axis=baxis, copy=copy,
+                                         level=level,
+                                         tupleize_cols=tupleize_cols)
             result._clear_item_cache()
 
         if inplace:
--- a/pandas/core/internals.py
+++ b/pandas/core/internals.py
@@ -3294,7 +3294,8 @@ class BlockManager(PandasObject):
 
         """
         obj = self.copy(deep=copy)
-        obj.set_axis(axis, _transform_index(self.axes[axis], mapper, level, tupleize_cols=tupleize_cols))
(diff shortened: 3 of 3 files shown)


pandas/core/frame.py
```diff
@@ -3642,6 +3642,10 @@ def rename(self, *args, **kwargs):
         level : int or level name, default None
             In case of a MultiIndex, only rename labels in the specified
             level.
+        tupleize_cols : boolean, default False
```
TomAugspurger: I'd prefer not to add a new keyword argument here. I don't think that `.rename` should be able to change between Index and MultiIndex, so it shouldn't be necessary.

charlie0389:
Ok, so from what I gather, the only change should be:

return Index(items, name=index.name, tupleize_cols=False)

(in internals._transform_index()). If that's agreed, I can make the change
on the other side of the weekend.

jreback:
this PR is not likely to be accepted, esp as there are no tests. We are not going to be adding a specific argument to ``.rename`` to special handle a very non-idiomatic special case here. If you want to propose something, do so on the original issue.

PR #20534, opened 2018-03-29
ENH20521 Added metadata argument to DataFrame.to_parquet 

### LINKED ISSUE #20521: ENH

Please comsider merging

https://github.com/JacekPliszka/pandas/commit/9459bdfabf3eff38da77718b9ec77e86b0ede54a

#### Problem description

Currently pandas can not add custom metadata to parquet file.

This patch add metadata argument to DataFrame.to_parquet that allows for that.
Warning is issued when pandas key is present in the dictionary passed.

JacekPliszka:
The argument allows for custom file metadata updating the default one.
Closes #20521 

- [x] closes #xxxx
- [x] tests added / passed
- [x] whatsnew entry

Checklist from comments:

- [ ] add example in to_parquet docstring
- [ ] add API docs
- [ ] investigate fastparquet
- [ ] correct kwargs
- [ ] add test for warning
- [ ] address other comments

### DIFF
--- a/pandas/tests/frame/conftest.py
+++ b/pandas/tests/frame/conftest.py
@@ -18,6 +18,20 @@ def float_frame():
 
 
 @pytest.fixture
+def float_frame_with_na():
+    """
+    Fixture for DataFrame of floats with index of unique strings
+
+    Columns are ['A', 'B', 'C', 'D']; some entries are missing
+    """
+    df = DataFrame(tm.getSeriesData())
+    # set some NAs
+    df.loc[5:10] = np.nan
+    df.loc[15:20, -2:] = np.nan
+    return df
+
+
+@pytest.fixture
 def float_frame2():
     """
     Fixture for DataFrame of floats with index of unique strings
@@ -25,6 +39,21 @@ def float_frame2():
     Columns are ['D', 'C', 'B', 'A']
     """
     return DataFrame(tm.getSeriesData(), columns=['D', 'C', 'B', 'A'])
+
+
+@pytest.fixture
+def bool_frame_with_na():
+    """
+    Fixture for DataFrame of booleans with index of unique strings
+
+    Columns are ['A', 'B', 'C', 'D']; some entries are missing
+    """
+    df = DataFrame(tm.getSeriesData()) > 0
+    df = df.astype(object)
+    # set some NAs
+    df.loc[5:10] = np.nan
+    df.loc[15:20, -2:] = np.nan
+    return df
 
 
 @pytest.fixture
--- a/pandas/tests/frame/test_analytics.py
+++ b/pandas/tests/frame/test_analytics.py
@@ -23,54 +23,188 @@ import pandas.core.algorithms as algorithms
 
 import pandas.util.testing as tm
 import pandas.util._test_decorators as td
-from pandas.tests.frame.common import TestData
 
 
-class TestDataFrameAnalytics(TestData):
+def _check_stat_op(name, alternative, main_frame, float_frame,
+                   float_string_frame, has_skipna=True,
+                   has_numeric_only=False, check_dtype=True,
+                   check_dates=False, check_less_precise=False,
+                   skipna_alternative=None):
+
+    f = getattr(main_frame, name)
+
+    if check_dates:
+        df = DataFrame({'b': date_range('1/1/2001', periods=2)})
+        _f = getattr(df, name)
+        result = _f()
+        assert isinstance(result, Series)
+
+        df['a'] = lrange(len(df))
+        result = getattr(df, name)()
+        assert isinstance(result, Series)
+        assert len(result)
+
+    if has_skipna:
+        def wrapper(x):
+            return alternative(x.values)
+
+        skipna_wrapper = tm._make_skipna_wrapper(alternative,
+                                                 skipna_alternative)
+        result0 = f(axis=0, skipna=False)
+        result1 = f(axis=1, skipna=False)
+        tm.assert_series_equal(result0, main_frame.apply(wrapper),
+                               check_dtype=check_dtype,
+                               check_less_precise=check_less_precise)
+        # HACK: win32
+        tm.assert_series_equal(result1, main_frame.apply(wrapper, axis=1),
+                               check_dtype=False,
+                               check_less_precise=check_less_precise)
+    else:
+        skipna_wrapper = alternative
+
+    result0 = f(axis=0)
+    result1 = f(axis=1)
+    tm.assert_series_equal(result0, main_frame.apply(skipna_wrapper),
+                           check_dtype=check_dtype,
+                           check_less_precise=check_less_precise)
+    if name in ['sum', 'prod']:
+        expected = main_frame.apply(skipna_wrapper, axis=1)
+        tm.assert_series_equal(result1, expected, check_dtype=False,
+                               check_less_precise=check_less_precise)
+
+    # check dtypes
+    if check_dtype:
+        lcd_dtype = main_frame.values.dtype
+        assert lcd_dtype == result0.dtype
+        assert lcd_dtype == result1.dtype
+
+    # bad axis
+    tm.assert_raises_regex(ValueError, 'No axis named 2', f, axis=2)
+    # make sure works on mixed-type frame
+    getattr(float_string_frame, name)(axis=0)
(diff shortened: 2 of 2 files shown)


pandas/io/parquet.py
```diff
@@ -113,9 +113,22 @@ def write(self, df, path, compression='snappy',
             table = self.api.Table.from_pandas(df, timestamps_to_ms=True)
             self.api.parquet.write_table(
                 table, path, compression=compression, **kwargs)
-
         else:
             table = self.api.Table.from_pandas(df)
+            custom_metadata = kwargs.pop('metadata', {})
+            if custom_metadata:
+                if 'pandas' in custom_metadata:
+                    warn(
```
WillAyd: There should also be a test that asserts this warning actually happens

WillAyd (changes requested):
Not being overly familiar with this format I'll admit I don't quite fully understand the purpose of this...and given we just went through a huge documentation sprint it seems like this is something that could be documented with an example in the `to_parquet` docstring

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -345,6 +345,7 @@ Other Enhancements
 - :meth:`DataFrame.to_sql` now performs a multivalue insert if the underlying connection supports itk rather than inserting row by row.
   ``SQLAlchemy`` dialects supporting multivalue inserts include: ``mysql``, ``postgresql``, ``sqlite`` and any dialect with ``supports_multivalues_insert``. (:issue:`14315`, :issue:`8953`)
 - :func:`read_html` now accepts a ``displayed_only`` keyword argument to controls whether or not hidden elements are parsed (``True`` by default) (:issue:`20027`)
+- :func:`DataFrame.to_parquet` now accepts a ``metadata`` keyword argument - the object passed updates key value file metadata generated by pandas. If pandas key is present - default pandas value is overwritten and warning is issued. Default ``None`` value means standard pandas metadata is used. (``None`` by default) (:issue:`20521`)
```
TomAugspurger: This should be a short highlight with a link to the API docs or a new section in `io.rst` if necessary.

pandas/tests/io/test_parquet.py
```diff
@@ -437,6 +437,27 @@ def test_s3_roundtrip(self, df_compat, s3_resource, pa):
         check_round_trip(df_compat, pa,
                          path='s3://pandas-test/pyarrow.parquet')
 
+    @pytest.mark.xfail(
+        is_platform_windows() or is_platform_mac(),
+        reason="reading pa metadata failing on Windows/mac"
+    )
+    def test_custom_metadata(self, pa_ge_070, df_full):
```
TomAugspurger: Add a comment about why pyarrow>=0.7.0 is required (is it?)

TomAugspurger (commented):
Could you add API docs?

Could you add a fastparquet implementation?

pandas/io/parquet.py
JacekPliszka: OK

pandas/io/parquet.py
```diff
@@ -113,9 +113,22 @@ def write(self, df, path, compression='snappy',
             table = self.api.Table.from_pandas(df, timestamps_to_ms=True)
             self.api.parquet.write_table(
                 table, path, compression=compression, **kwargs)
-
         else:
             table = self.api.Table.from_pandas(df)
+            custom_metadata = kwargs.pop('metadata', {})
```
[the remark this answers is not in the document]
JacekPliszka: OK - I will change.

pandas/tests/io/test_parquet.py
```diff
@@ -437,6 +437,27 @@ def test_s3_roundtrip(self, df_compat, s3_resource, pa):
         check_round_trip(df_compat, pa,
                          path='s3://pandas-test/pyarrow.parquet')
 
+    @pytest.mark.xfail(
+        is_platform_windows() or is_platform_mac(),
```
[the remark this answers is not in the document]
JacekPliszka: Not really. This feature is useful mostly when you have hundreds of large files.
Such deployments are almost 100% Linux. I've never seen such deployment on Windows or Mac.

I put it because couple lines above you have
```
@pytest.mark.xfail(is_platform_windows() or is_platform_mac(),
                   reason="reading pa metadata failing on Windows/mac")
def test_cross_engine_pa_fp(df_cross_compat, pa, fp):
```
The above comment suggests the features are less mature on Windows/Mac.
I do not have Windows or Mac so I can not test it/check it.

Anyone who needs it on Windows/Mac could validate it and correct.

doc/source/whatsnew/v0.23.0.txt
JacekPliszka: Do you mean this should be shorter and sentence moved to API docs?

pandas/tests/io/test_parquet.py
JacekPliszka: Not sure. It might work with 0.5.0.

Do you really think it is worth testing?

pyarrow is already 0.9.0 and people should rather not use versions older than 0.7.0.
In my opinion the newer versions of pandas should require at least 0.7.0
the older ones are not worth testing.

pandas/tests/io/test_parquet.py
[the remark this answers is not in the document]
TomAugspurger: Our CI runs on Mac, Linux, and Windows.

doc/source/whatsnew/v0.23.0.txt
TomAugspurger: Yeah, a link to the API documentation is probably best here.

pandas/tests/io/test_parquet.py
TomAugspurger: It's worth seeing if it works.

pandas/io/parquet.py
```diff
@@ -113,9 +113,22 @@ def write(self, df, path, compression='snappy',
...
+                    warn(
+                        'pandas key present in metadata argument - will '
+                        'overwrite default value needed for compatibility '
+                        'reasons, better use another key unless you know what '
+                        'you are doing.'
+                    )
+                custom_metadata = dict(
+                    table.schema.metadata or {},
+                    **custom_metadata
+                )
```
jreback: why do you want to do this in pandas? is this not a feature request in pyarrow itself? e.g. to process a ``metadata`` kwarg?

cc @cpcloud @wesm

jreback (changes requested):

pandas/tests/io/test_parquet.py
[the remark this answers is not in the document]
jreback: @JacekPliszka we have a test of cross compatibility of fastparquet and pyarrow. this is a very specific instance. this is not applicable to this PR

JacekPliszka:
Summing up:  pyarrow/pandas handling of file level parquet metadata is ugly and at some point of time should be rewritten but looks like it is a larger task than I can afford to do.

The above simple hack can be implement by anyone interested on its own just instead of df.to_parquet
use function like this:
```
import pyarrow.parquet
def to_parquet_with_updated_metadata(filename, df, metadata_update):
    table = pyarrow.Table.from_pandas(df)
    updated_metadata = dict(table.schema.metadata or {}, **metadata_update)
    table = table.replace_schema_metadata(updated_metadata)
    pyarrow.parquet.write_table(table, filename)

to_parquet_with_updated_metadata('test.parquet', df, {'test key ': 'test value'})
```
Should be usable enough until proper solution with access to all metadatas in parquet file is developed.
This is why I closed this PR

jorisvandenbossche:
@JacekPliszka coming back to this: can you be a bit more specific in what is missing in pyarrow (or pandas) for properly handling of parquet metadata? 
Do you know if there is an issue for that on the pyarrow side? (or otherwise want to open one?)

jorisvandenbossche:
Is what you are looking for the ability to specify additional metadata in the `pyarrow.parquet.write_table` function?

JacekPliszka:
It's been couple months since I looked at it but parquet allows metadata at several levels - full file, column and page header.

It would be nice to have access to read and write at least the first two.

Firstly some kind of convention/format needs to be set how to separate library/pandas metadata from user controlled one.

Also there is issue of parquet datasets - individual files has each its own metadata while they are read as single pandas dataframe

PR #20545, opened 2018-03-30
CLN: Use pandas.compat instead of sys.version_info for Python version checks

jschendel:
Hopefully this will make code changes related to dropping support for specific versions of Python slightly easier to identify.

Summary:
  - Replaced instances `sys.version_info` with equivalent checks from `pandas.compat`
  - Removed code blocks specific to unsupported versions of Python (<2.7, 3.0-3.4)

<br />
There were some very specific examples that I left as-is:

https://github.com/pandas-dev/pandas/blob/c4b4a81f56205082ec7f12bf77766e3b74d27c37/pandas/tests/io/formats/test_to_csv.py#L13

Didn't change any Cython related code, as it appears that we want to avoid imports from outside `_libs`:
https://github.com/pandas-dev/pandas/blob/c4b4a81f56205082ec7f12bf77766e3b74d27c37/pandas/_libs/tslibs/parsing.pyx#L25-L26

### DIFF
--- a/pandas/tests/io/parser/common.py
+++ b/pandas/tests/io/parser/common.py
@@ -1275,9 +1275,8 @@ eight,1,2,3"""
         else:  # Python engine
             assert output == 'Filled 1 NA values in column a\n'
 
+    @pytest.mark.skipif(PY3, reason="won't work in Python 3")
     def test_iteration_open_handle(self):
-        if PY3:
-            pytest.skip("won't work in Python 3")
 
         with tm.ensure_clean() as path:
             with open(path, 'wb') as f:
--- a/pandas/tests/io/parser/common.py
+++ b/pandas/tests/io/parser/common.py
@@ -1277,8 +1277,7 @@ eight,1,2,3"""
 
     def test_iteration_open_handle(self):
         if PY3:
-            pytest.skip(
-                "won't work in Python 3 {0}".format(sys.version_info))
+            pytest.skip("won't work in Python 3")
 
         with tm.ensure_clean() as path:
             with open(path, 'wb') as f:
--- a/pandas/tests/scalar/timestamp/test_comparisons.py
+++ b/pandas/tests/scalar/timestamp/test_comparisons.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import sys
 from datetime import datetime
 import operator
 
@@ -9,7 +8,7 @@ import numpy as np
 from dateutil.tz import tzutc
 from pytz import utc
 
-from pandas.compat import long
+from pandas.compat import long, PY2
 from pandas import Timestamp
 
 
@@ -104,7 +103,7 @@ class TestTimestampComparison(object):
         pytest.raises(Exception, b.__lt__, a)
         pytest.raises(Exception, b.__gt__, a)
 
-        if sys.version_info < (3, 3):
+        if PY2:
             pytest.raises(Exception, a.__eq__, b.to_pydatetime())
             pytest.raises(Exception, a.to_pydatetime().__eq__, b)
         else:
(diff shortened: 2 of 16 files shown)


pandas/util/testing.py
```diff
@@ -2088,7 +2088,7 @@ def dec(f):
 # and conditionally raise on these exception types
 _network_error_classes = (IOError, httplib.HTTPException)
 
-if sys.version_info >= (3, 3):
+if PY3:
```
jschendel: Since 3.0 - 3.4 are no longer supported, seems fine to use `PY3`.  Could use `PY35` if being explicit would be preferred here (or maybe just leave as-is).

pandas/tests/scalar/timestamp/test_comparisons.py
```diff
@@ -104,7 +103,7 @@ def test_cant_compare_tz_naive_w_aware(self):
         pytest.raises(Exception, b.__lt__, a)
         pytest.raises(Exception, b.__gt__, a)
 
-        if sys.version_info < (3, 3):
+        if PY2:
```
jschendel: Likewise, <3.3 implies `PY2` based on the currently supported versions.  Could be more explicit if desired though.

jorisvandenbossche (commented):
Looks good to me. 
Thanks for the clean-up!

jreback:
lgtm. the only thing would add is if you can fail ``lint.sh`` on a grep for ``sys.version_info`` would be great.

jschendel:
I can add a check for this to `lint.sh`.  There are a situations where `sys.version_info` is still valid though, as there are some very specific checks that probably don't make sense to encode in `pandas.compat`, eg:
https://github.com/pandas-dev/pandas/blob/c4b4a81f56205082ec7f12bf77766e3b74d27c37/pandas/tests/io/formats/test_to_csv.py#L13

This would get caught by `lint.sh`, but I'm guessing we'd rather be more aggressive here, and selectively allow failing cases to be merged when appropriate?  I'm assuming (possibly mistakenly) that `lint.sh` only runs over newly added code?  If it hits the entire codebase, we might not want to do this since the corner cases will fail every time.  I suppose I could try checking for specific examples that are clearly invalid if this is the case, though it seems like that could get out of sync as support for versions are added/dropped.

jreback:
we have a PY35 already. The lint check just enforces that the code all uses the same pattern, don't like to have people making up ways to do very standard things (and docs on how to do things like this, while helpful, get out of date), better to have an actual check.

jschendel:
The example I showed is actually different than `PY35`.  It's only skipping for 3.5-3.6.4; the test should pass for 2.7 and 3.6.5+.  Something like `PY2 or PY365` would work, so it'd be a matter of adding `PY365`.

That actually looks like the only oddly specific case of a version check.  The only other non-`pandas.compat` instance of `sys.version_info` that I can't get rid of is in `util/_print_versions.py`:
https://github.com/pandas-dev/pandas/blob/0a003651a67854b9776c46469a18e20ffd61f141/pandas/util/_print_versions.py#L41

Could either skip that file or create a variable like `PYVERSION = sys.version_info` in `pandas.compat` and import.

pandas/tests/scalar/timestamp/test_comparisons.py
gfyoung: That's reasonable.  2.7, 3.5 - 3.6 is the way to go.

pandas/util/testing.py
gfyoung: I'm fine with this change.  We should be encouraging code that relies solely on the supported versions of Python and not to the C-esque thing of making it super-backwards-compatible 😄

gfyoung (approved):
Nice fixes!

@jreback

jreback:
yeah have special cases is a bit annoying. how about you just remove the remaining ones that are possible? and leave lint.sh alone is ok.

jschendel:
Added some small fixes for the remaining easily removable instances of `sys.version_info`.

jreback:
great. can you rebase. just merged a patch which conflicts.

pandas/tests/io/parser/common.py
```diff
@@ -1277,8 +1277,7 @@ def test_verbose_import(self):
 
     def test_iteration_open_handle(self):
```
jreback: can you change this to a pytest.mark.skip decorator

jschendel:
rebased and made the requested change

jreback:
thanks!

Issue #20527, opened 2018-03-29
ERR: disallow non-hashables in Index construction & rename


hodossy:
```python
from pandas import Index, DataFrame

df = DataFrame(index=Index([1, 2, 3, 4]), columns=['A', 'B'])
df.index.rename(['foo'], inplace=True)
df.reset_index()
```

> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> <ipython-input-23-20da24c655af> in <module>()
> ----> 5 df.reset_index()
> 
> c:\program files\python36\lib\site-packages\pandas\core\frame.py in reset_index(self, level, drop, inplace, col_level, col_fill)
>    3377                 # to ndarray and maybe infer different dtype
>    3378                 level_values = _maybe_casted_values(lev, lab)
> -> 3379                 new_obj.insert(0, name, level_values)
>    3380 
>    3381         new_obj.index = new_index
> 
> c:\program files\python36\lib\site-packages\pandas\core\frame.py in insert(self, loc, column, value, allow_duplicates)
>    2611         value = self._sanitize_column(column, value, broadcast=False)
>    2612         self._data.insert(loc, column, value,
> -> 2613                           allow_duplicates=allow_duplicates)
>    2614 
>    2615     def assign(self, **kwargs):
> 
> c:\program files\python36\lib\site-packages\pandas\core\internals.py in insert(self, loc, item, value, allow_duplicates)
>    4059 
>    4060         """
> -> 4061         if not allow_duplicates and item in self.items:
>    4062             # Should this be a different kind of error??
>    4063             raise ValueError('cannot insert {}, already exists'.format(item))
> 
> c:\program files\python36\lib\site-packages\pandas\core\indexes\base.py in __contains__(self, key)
>    1692     @Appender(_index_shared_docs['__contains__'] % _index_doc_kwargs)
>    1693     def __contains__(self, key):
> -> 1694         hash(key)
>    1695         try:
>    1696             return key in self._engine
> 
> TypeError: unhashable type: 'list'

#### Problem description

It think it is reasonable to expect Index([1, 2, 3, 4]).rename(['foo']) and Index([1, 2, 3, 4]).rename('foo') resulting in the same, however this is not the case.

Probably this line be modified in Index.rename:

    - return self.set_names([name], inplace=inplace)
    + return self.set_names(name, inplace=inplace)

#### Expected Output

```python
>>>Index([1, 2, 3, 4]).rename(['foo'])
[...]

Issue #19295, opened 2018-01-18
Deprecate pd.*Index*(names='')


toobaz:
- ``pd.Index`` accepts ``name`` - great
- ``pd.MultiIndex`` accepts ``names`` - after all, they are multiple ``name``s
- ``pd.Index`` accepts ``names`` for compatibility with ``pd.MultiIndex`` - well, OK, after all, that constructor _can_ result in a ``MultiIndex``
- ``pd.Index`` should accept ( #19082 ) ``names`` even when it results in a flat index - well, OK, still for compatibility
- ``pd.MultiIndex`` accepts ``name`` for compatibility  - wait, wasn't ``names`` already provided for compatibility?! OK, forget it, go for ``name``.
- ``pd.Index`` accepts ``name`` even for multiple levels' names, for compatibility - with ``MultiIndex``, which accepts it for compatibility with ``pd.Index`` - aaaaalright
- All ``pd.Index`` subclasses (which will _never_ result in multiple levels, and which currently in some cases discard ``names``, in other cases raise an error) [should accept](https://github.com/pandas-dev/pandas/pull/19168/files#r161731210) ``names`` for compatibility - no, wait.

####  Proposal 1

1. There is _one_ way to have compatibility across any kind of ``Index`` (sub)class constructor, and it is ``name``. When the constructor results in a ``MultiIndex`` (which can happen with ``pd.Index`` or ``pd.MultiIndex``), then ``name`` should be list-like, and each level will "receive" the respective component

2. in those cases - and only in those cases - in which a constructor _actually_ results in a ``MultiIndex``, ``names`` can be used as an alias for ``name``. In all other cases, it is not accepted.

or alternatively:

2. (b) (my preference) ``names`` is deprecated (in all cases in which it is currently supported, and remains unsupported in other constructors/cases)

3. (c) (tradeoff between mental health and backward compatibility) ``names`` is supported in ``pd.MultiIndex``, still supported but deprecated in ``pd.Index`` when resulting in ``MultiIndex`` (and remains unsupported in other constructors/cases)

Corollary:

3. ``names`` will _not_ be supported by any constructor that is not ``pd.Index`` or ``pd.MultiIndex``.

Notice that a 1-level ``MultiIndex`` is still a ``MultiIndex``. That is,

- ``pd.Index([('a',), ('b',)], name=('only_one_name',))`` will still work
[...]

PR #20548, opened 2018-03-30
ERR: disallow non-hashables in Index/MultiIndex construction & rename

arminv:
Index & MultiIndex names need to be hashable. Both constructing and renaming without a hashable name raise TypeError exceptions now.

**Examples:**

- Index:

```
In [2]: pd.Index([1, 2, 3], name=['foo'])
>>> Int64Index([1, 2, 3], dtype='int64', name=['foo'])
```

```
In [3]: pd.Index([1, 2, 3], name='foo').rename(['bar'])
<ipython-input-3-f3327eccf0fc> in <module>()
----> 1 pd.Index([1, 2, 3], name='foo').rename(['bar'])

~/Documents/GitHub/pandas/pandas/core/indexes/base.py in rename(self, name, inplace)
   1406         new index (of same type and class...etc) [if inplace, returns None]
   1407         """
-> 1408         return self.set_names([name], inplace=inplace)
   1409 
   1410     @property

~/Documents/GitHub/pandas/pandas/core/indexes/base.py in set_names(self, names, level, inplace)
   1387         else:
   1388             idx = self._shallow_copy()
-> 1389         idx._set_names(names, level=level)
   1390         if not inplace:
   1391             return idx

~/Documents/GitHub/pandas/pandas/core/indexes/base.py in _set_names(self, values, level)
   1323                 if not is_hashable(name):
   1324                     raise TypeError('{}.name must be a hashable type'
-> 1325                                     .format(self.__class__.__name__))
   1326         if len(values) != 1:
   1327             raise ValueError('Length of new names must be 1, got %d' %

TypeError: Int64Index.name must be a hashable type
```

- MultiIndex:

```
In [4]: pd.MultiIndex(levels=[[1, 2], [u'one', u'two']],
                        labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
                        names=((['foo'], ['bar'])))
                        
<ipython-input-4-d211526eaa3d> in <module>()
      1 pd.MultiIndex(levels=[[1, 2], [u'one', u'two']],
      2                     labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
----> 3                     names=((['foo'], ['bar'])))
      4 

~/Documents/GitHub/pandas/pandas/core/indexes/multi.py in __new__(cls, levels, labels, sortorder, names, dtype, copy, name, verify_integrity, _set_identity)
    230         if names is not None:
    231             # handles name validation
--> 232             result._set_names(names)
    233 
    234         if sortorder is not None:

~/Documents/GitHub/pandas/pandas/core/indexes/multi.py in _set_names(self, names, level, validate)
    646             if not is_hashable(name):
    647                 raise TypeError('{}.name must be a hashable type'
--> 648                                 .format(self.__class__.__name__))
    649 
    650         # GH 15110

TypeError: MultiIndex.name must be a hashable type
```
```
In [10]: pd.MultiIndex(levels=[[1, 2], [u'one', u'two']],
                       labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
                       names=('foo', 'bar')).rename(([1], [2]))
<ipython-input-5-ff74dfc48455> in <module>()
      1 pd.MultiIndex(levels=[[1, 2], [u'one', u'two']],
      2                      labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
----> 3                      names=('foo', 'bar')).rename(([1], [2]))
      4 

~/Documents/GitHub/pandas/pandas/core/indexes/base.py in set_names(self, names, level, inplace)
   1387         else:
   1388             idx = self._shallow_copy()
-> 1389         idx._set_names(names, level=level)
   1390         if not inplace:
   1391             return idx

~/Documents/GitHub/pandas/pandas/core/indexes/multi.py in _set_names(self, names, level, validate)
    646             if not is_hashable(name):
    647                 raise TypeError('{}.name must be a hashable type'
--> 648                                 .format(self.__class__.__name__))
    649 
    650         # GH 15110

TypeError: MultiIndex.name must be a hashable type
```

- [x] closes [#20527](https://github.com/pandas-dev/pandas/issues/20527)
- [x] tests added / passed
- [x] whatsnew entry


jreback:
this already works on Serie

the issue is about Index

arminv:
Sorry I got confused. I will update it for Index.

pandas/core/indexes/base.py
```diff
@@ -1392,7 +1396,10 @@ def rename(self, name, inplace=False):
         -------
         new index (of same type and class...etc) [if inplace, returns None]
         """
-        return self.set_names([name], inplace=inplace)
+        if name is not None and not is_hashable(name):
+            raise TypeError('Index.name must be a hashable type')
+        else:
```
jreback: rather do this in set_names

jreback (changes requested):
can you add a few tests. and a whatsnew new (other API changes).

jreback (changes requested):
- can you add a note in other api changes section
- can you add tests on construction & for rename (these should use our current infrastructure to exercise all subclasses)

pandas/core/indexes/base.py
```diff
@@ -251,6 +252,9 @@ def __new__(cls, data=None, dtype=None, copy=False, name=None,
         if name is None and hasattr(data, 'name'):
             name = data.name
 
+        if name is not None and not is_hashable(name):
+            raise TypeError('Index.name must be a hashable type')
+
```
jreback: this very likely also needs checking for MultiIndex (as that's a different path in some cases).

pandas/core/indexes/base.py
```diff
@@ -251,6 +252,9 @@ def __new__(cls, data=None, dtype=None, copy=False, name=None,
         if name is None and hasattr(data, 'name'):
             name = data.name
 
+        if name is not None and not is_hashable(name):
+            raise TypeError('Index.name must be a hashable type')
```
jreback: use ``self.__class__.__name__`` rather than ``Index`` here

jreback (changes requested):

pandas/core/indexes/base.py
arminv: Do we allow non-hashable names for MultiIndex?

jreback: no

pandas/core/indexes/base.py
arminv: I did this in set_names and a lot of tests failed. Is there a particular reason we can’t keep it here?

pandas/core/indexes/base.py
arminv: For a MultiIndex, it seems that `names` is converted into FrozenList after creation. I found [this answer](https://stackoverflow.com/questions/25676107/why-is-frozenlist-different-from-tuple) from you on StackOverflow about hashability of a FrozenList.

Right now, if `names` can’t be converted to a FrozenList, it throws an exception. For example:
```
In [1]: pd.MultiIndex(levels=[[1, 2], [u'one', u'two']],
   ...:                    labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
   ...:                    names=(['foo'], ['bar']))
   ...:           

TypeError: unhashable type: 'list'
```
while this passes:
```
In [2]: pd.MultiIndex(levels=[[1, 2], [u'one', u'two']],
    ...:                    labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
    ...:                    names=[('foo'), ('bar')])
```
Do we need to change anything here?

jreback: no you just need to check that each name is hashable, not the frozen list itself. that's why .set_names is the best place for this

pandas/core/indexes/base.py
jreback: you are going to need it in set_names as that is the canonical way to do this. that's where it should validate. if we have tests that are clearly in error they should be changed.

TomAugspurger:
I'd prefer to make this a FutureWarning for a release, and then change it to raise. I don't think there's anything preventing that, correct?

arminv:
@TomAugspurger It works fine for an Index:

```
In [2]: pd.Index([1, 2, 3], name=['foo'])
~/Documents/GitHub/pandas/pandas/core/indexes/base.py:260: FutureWarning: Index.name must be a hashable type. In future this will raise a TypeError.
  warnings.warn(cls.__name__ + msg, FutureWarning)
Out[2]: Int64Index([1, 2, 3], dtype='int64', name=['foo'])
``` 

But MultiIndex already raises `TypeError` somewhere else, so the user gets an exception (after the warning):
```
In [6]: pd.MultiIndex(levels=[[1, 2], [u'one', u'two']],
   ...:    ...:                     labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
   ...:    ...:                     names=((['foo'], ['bar'])))
   ...:                     
~/Documents/GitHub/pandas/pandas/core/indexes/multi.py:652: FutureWarning: MultiIndex.name must be a hashable type. In future this will raise a TypeError.
  msg, FutureWarning)
--------------------------------------------------------------------------
TypeError                                Traceback (most recent call last)
<ipython-input-6-d211526eaa3d> in <module>()
      1 pd.MultiIndex(levels=[[1, 2], [u'one', u'two']],
      2                     labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
----> 3                     names=((['foo'], ['bar'])))
      4 

~/Documents/GitHub/pandas/pandas/core/indexes/multi.py in __new__(cls, levels, labels, sortorder, names, dtype, copy, name, verify_integrity, _set_identity)
    230         if names is not None:
    231             # handles name validation
--> 232             result._set_names(names)
    233 
    234         if sortorder is not None:

~/Documents/GitHub/pandas/pandas/core/indexes/multi.py in _set_names(self, names, level, validate)
    676         # set the name
    677         for l, name in zip(level, names):
--> 678             if name is not None and name in used:
    679                 raise ValueError('Duplicated level name: "{}", assigned to '
    680                                  'level {}, is already used for level '

TypeError: unhashable type: 'list'
```
Is it a good idea to show FutureWarning (for both Index & MultiIndex) and change the exception for MultiIndex?

jreback:
i don’t think this is a big deal to simply change this and raise an error

TomAugspurger:
Hmm, I didn't realize that MI already raised, though that's pretty clearly
a result of the implementation.

I would still prefer a deprecation cycle, but I'm not -1 on just breaking.

pandas/core/indexes/base.py
```diff
@@ -1354,6 +1357,19 @@ def set_names(self, names, level=None, inplace=False):
                    labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
                    names=[u'baz', u'bar'])
         """
+        # GH 20527
+        # All items in 'names' need to be hashable:
+        if names is not None:
+            for name in names:
+                if is_hashable(name):
```
jreback: use ``not is_hashable`` here as you are only checking the negative condition

pandas/core/indexes/base.py
```diff
@@ -1354,6 +1357,19 @@ def set_names(self, names, level=None, inplace=False):
...
+                    pass
+                else:
+                    if self.nlevels == 1:
```
jreback: this only needs to deal with regular indices. MI has its own version

pandas/core/indexes/multi.py
```diff
@@ -639,6 +640,14 @@ def _set_names(self, names, level=None, validate=True):
         Note that you generally want to set this *after* changing levels, so
         that it only acts on copies
         """
+        # GH 20527
+        # All items in 'names' need to be hashable:
+        for name in names:
+            if is_hashable(name):
```
jreback: like this
```
for level, name in enumerate(names):
     if not is_hashable(name):
       ...

```

show the value and level as well

pandas/tests/indexes/test_base.py
```diff
@@ -436,6 +436,20 @@ def test_constructor_empty(self):
         assert isinstance(empty, MultiIndex)
         assert not len(empty)
 
+    def test_constructor_nonhashable_name(self):
```
jreback: use ``self.create_index()`` here (you can pass a name)

pandas/tests/indexes/test_multi.py
```diff
@@ -615,8 +615,26 @@ def test_constructor_mismatched_label_levels(self):
...
+                               MultiIndex, levels=levels,
+                               labels=labels, names=names)
+
+        # With .rename()
+        mi = MultiIndex(levels=[[1, 2], [u'one', u'two']],
+                        labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
+                        names=('foo', 'bar'))
+        assert isinstance(mi, MultiIndex)
+        renamed = [['foor'], ['barr']]
+        tm.assert_raises_regex(TypeError, message, mi.rename, names=renamed)
```
jreback: also check set_names

pandas/tests/indexes/test_base.py
```diff
@@ -436,6 +436,20 @@ def test_constructor_empty(self):
...
+        name = ['0']
+        message = "Index.name must be a hashable type"
+        tm.assert_raises_regex(TypeError, message, data=data, name=name)
+
+        # With .rename()
+        idx = Index([1, 2, 3], name='0')
+        assert isinstance(idx, Index)
+        renamed = ['1']
+        tm.assert_raises_regex(TypeError, message,
+                               idx.rename, name=renamed)
```
jreback: also check set_names

jreback (changes requested):

jreback:
see https://github.com/pandas-dev/pandas/issues/19295

does need to be fixed / it’s orthogonal (somewhat) from this one though

pandas/core/indexes/base.py
```diff
@@ -251,6 +252,10 @@ def __new__(cls, data=None, dtype=None, copy=False, name=None,
         if name is None and hasattr(data, 'name'):
             name = data.name
 
+        if name is not None and not is_hashable(name):
+            raise TypeError(cls.__name__ +
```
jreback: can you use ``.format()`` here instead (and whenever you are composing an error string)

pandas/core/indexes/multi.py
```diff
@@ -639,6 +640,12 @@ def _set_names(self, names, level=None, validate=True):
         Note that you generally want to set this *after* changing levels, so
         that it only acts on copies
         """
+        # GH 20527
+        # All items in 'names' need to be hashable:
+        for levels, name in enumerate(names):
```
jreback: same

pandas/core/indexes/base.py
```diff
@@ -1354,6 +1359,14 @@ def set_names(self, names, level=None, inplace=False):
                    labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
                    names=[u'baz', u'bar'])
         """
+        # GH 20527
+        # All items in 'names' need to be hashable:
+        if names is not None:
+            for name in names:
+                if not is_hashable(name):
+                    raise TypeError(self.__class__.__name__ +
```
jreback: same

pandas/tests/indexes/test_base.py
```diff
@@ -436,6 +436,22 @@ def test_constructor_empty(self):
...
 
+    def test_constructor_nonhashable_name(self):
+        # GH 20527
+        idx = self.create_index()
+        name = ['0']
+        message = "Index.name must be a hashable type"
+        tm.assert_raises_regex(TypeError, message, name=name)
+
+        # With .rename()
+        assert isinstance(idx, Index)
```
jreback: you don't need this assert

pandas/tests/indexes/test_multi.py
```diff
@@ -615,8 +615,28 @@ def test_constructor_mismatched_label_levels(self):
...
+        message = "MultiIndex.name must be a hashable type"
+        tm.assert_raises_regex(TypeError, message,
+                               MultiIndex, levels=levels,
+                               labels=labels, names=names)
+
+        # With .rename()
+        mi = MultiIndex(levels=[[1, 2], [u'one', u'two']],
+                        labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
+                        names=('foo', 'bar'))
+        assert isinstance(mi, MultiIndex)
```
jreback: you dont' need the assert

jreback (changes requested):
not sure how ``set_names`` is actually working. I think the checks need to be in ``._set_names`` itself, no?

arminv:
@jreback 

`MultiIndex`: checks are already done in `_set_names`, as `set_names` is not defined there.

`Index`: `rename` calls `set_names` which in turn calls `_set_names`. So I agree that we should do checks in `_set_names` instead. This would also make it consistent with how we do `MultiIndex` (above).

So we'll have something like this after:

```
In [3]: pd.Index([1, 2, 3], name='0').rename(['1'])
<ipython-input-3-c2fc352ddf8e> in <module>()
----> 1 pd.Index([1, 2, 3], name='0').rename(['1'])

~/Documents/GitHub/pandas/pandas/core/indexes/base.py in rename(self, name, inplace)
   1411         new index (of same type and class...etc) [if inplace, returns None]
   1412         """
-> 1413         return self.set_names([name], inplace=inplace)
   1414 
   1415     @property

~/Documents/GitHub/pandas/pandas/core/indexes/base.py in set_names(self, names, level, inplace)
   1392         else:
   1393             idx = self._shallow_copy()
-> 1394         idx._set_names(names, level=level)
   1395         if not inplace:
   1396             return idx

~/Documents/GitHub/pandas/pandas/core/indexes/base.py in _set_names(self, values, level)
   1321                 if not is_hashable(name):
   1322                     raise TypeError('{}.name must be a hashable type'
-> 1323                                     .format(self.__class__.__name__))
   1324         if len(values) != 1:
   1325             raise ValueError('Length of new names must be 1, got %d' %

TypeError: Int64Index.name must be a hashable type```

File: pandas/core/indexes/base.py [hunk 15]
```diff
@@ -251,6 +252,10 @@ def __new__(cls, data=None, dtype=None, copy=False, name=None,
         if name is None and hasattr(data, 'name'):
             name = data.name
 
+        if name is not None and not is_hashable(name):
```
jreback: I don't actually think you need this here, as ``_set_name`` gets called from ``_simple_new`` (when ``.name`` is set). so rather more logical to put this in ``_set_name`` (like you do for MI)

pandas/core/indexes/base.py
```diff
@@ -1311,6 +1316,13 @@ def _get_names(self):
         return FrozenList((self.name, ))
 
     def _set_names(self, values, level=None):
+        # GH 20527
```
jreback: can you add a doc-string here

pandas/core/indexes/multi.py
```diff
@@ -639,6 +640,12 @@ def _set_names(self, names, level=None, validate=True):
         Note that you generally want to set this *after* changing levels, so
         that it only acts on copies
```
jreback: looks good here. can you update the doc-string with the comment below

pandas/tests/indexes/test_base.py
```diff
@@ -436,6 +436,21 @@ def test_constructor_empty(self):
         assert isinstance(empty, MultiIndex)
         assert not len(empty)
 
+    def test_constructor_nonhashable_name(self):
+        # GH 20527
+        idx = self.create_index()
+        name = ['0']
+        message = "Index.name must be a hashable type"
+        tm.assert_raises_regex(TypeError, message, name=name)
+
```
jreback: add this - it will test with all of the indices types (this is confusing atm as we have multiple way of running index creation: fixture, a function, and as a class property); other's PR's are cleaning this up.

```
diff --git a/pandas/tests/indexes/test_base.py b/pandas/tests/indexes/test_base.py
index 4265c4b56..ad7f0d7aa 100644
--- a/pandas/tests/indexes/test_base.py
+++ b/pandas/tests/indexes/test_base.py
@@ -436,9 +436,12 @@ class TestIndex(Base):
         assert isinstance(empty, MultiIndex)
         assert not len(empty)
 
-    def test_constructor_nonhashable_name(self):
+    def test_constructor_nonhashable_name(self, indices):
         # GH 20527
-        idx = self.create_index()
+
+        if isinstance(indices, MultiIndex):
+            pytest.skip("multiindex handled in test_multiindex.py")
+
         name = ['0']
         message = "Index.name must be a hashable type"
         tm.assert_raises_regex(TypeError, message, name=name)
@@ -446,10 +449,10 @@ class TestIndex(Base):
         # With .rename()
         renamed = [['1']]
         tm.assert_raises_regex(TypeError, message,
-                               idx.rename, name=renamed)
+                               indices.rename, name=renamed)
         # With .set_names()
         tm.assert_raises_regex(TypeError, message,
-                               idx.set_names, names=renamed)
+                               indices.set_names, names=renamed)
 
     def test_view_with_args(self):
```

jreback (changes requested):

pandas/tests/indexes/test_base.py
jreback: I pushed this to your branch

arminv: @jreback Thanks. Can we do the same for MI (i.e. through fixture) ?

pandas/core/indexes/base.py
arminv: It seems like `_simple_new` calls Index's `__new__` through `result = object.__new__(cls)`. So if we delete this we won't have a hashability check on Index **construction**? 

For example `pd.Index([1, 2, 3], name=['foo'])` won't raise anymore. Is this ok?

Also note we are already doing checks in `_set_names` (for both Index & MI).

arminv: Is there any way we can prevent the user from setting a non-hashable name **after** Index is created and if they do it **directly**  (i.e. not using `rename`, `set_names`,etc)? So that this will raise:
```
I1 = pd.Index([1, 2, 3], name='0')
I1.name = [[0]]

I1
>>> Int64Index([1, 2, 3], dtype='int64', name=[[0]])
```

arminv:
Is there any way we can prevent the user from setting a non-hashable name **after** Index is created and if they do it **directly**  (i.e. not using `rename`, `set_names`,etc)? So that this will raise:
```
I1 = pd.Index([1, 2, 3], name='0')

I1.name = [[0]]  # This shouldn't be ok!

I1
>>> Int64Index([1, 2, 3], dtype='int64', name=[[0]])
```

pandas/core/indexes/base.py
arminv: For MI, the name validation is delegated to `_set_names`:

https://github.com/pandas-dev/pandas/blob/2794474334cbd53315c248b605953579d010c693/pandas/core/indexes/multi.py#L223-L231

Can we do the same for Index here? Maybe something like this added in Index’s ` __new__()`:

```
_result = object.__new__(Index)

if name is not None:
    # handles name validation
    _result._set_names([name])
```

This way `pd.Index([1, 2, 3], name=['foo'])` would still raise but we do the check in a more logical place.

pandas/core/indexes/base.py
```diff
@@ -473,7 +474,7 @@ def _simple_new(cls, values, name=None, dtype=None, **kwargs):
 
         result = object.__new__(cls)
         result._data = values
-        result.name = name
+        result._set_names([name])
```
arminv: @jreback I wasn't sure if `_set_names` was getting called from `_simple_new`, so I made it explicit. Is this ok?

Also, we are not checking in `__new__` anymore (as you suggested).

pandas/core/indexes/base.py
```diff
@@ -1311,6 +1312,52 @@ def _get_names(self):
...
+
+        See Also
+        --------
+        Index.set_names : Set new names on index. Defaults to returning
+        new index.
+        Index.rename : Set new names on index. Defaults to returning new index.
+
+        Notes
+        -----
+        Both `set_names` and `rename` call this function to set name.
```
arminv: I'm not sure if this is needed?

pandas/core/indexes/base.py
jreback: you shouldn't need to do this, and just leave the original code

setting ``.name`` name is a property that calls ``_set_names``

pandas/core/indexes/base.py
```diff
@@ -1311,6 +1312,52 @@ def _get_names(self):
...
+
+        Examples
+        --------
+        on an index with no names:
```
jreback: you don't need the fulll doc string here (e.g. examples and such, leave Parameters and such), only on ``.set_names``

jreback (changes requested):
lgtm. some doc-comments. ping on green.

jreback:
can you update. ping on green.

pandas/core/indexes/base.py
```diff
@@ -1311,6 +1312,28 @@ def _get_names(self):
...
+
+        Parameters
+        ----------
+        values : str or sequence
+            name(s) to set
+        level : int, level name, or sequence of int/level names (default None)
+            If the index is a MultiIndex (hierarchical), level(s) to set (None
+            for all levels).  Otherwise level must be None
+
+        Returns
```
jreback: this should be Raises (and its a TypeError)

pandas/core/indexes/base.py
```diff
@@ -1311,6 +1312,28 @@ def _get_names(self):
         return FrozenList((self.name, ))
 
     def _set_names(self, values, level=None):
+        """
```
jreback: can you also add a mention on ``set_names`` itself that the names must be hashable (and examples if you want)

jreback (changes requested):

jreback (approved):

jreback:
moved the logic slightly. will merge on green.

pandas/tests/indexes/test_multi.py
```diff
@@ -615,8 +615,27 @@ def test_constructor_mismatched_label_levels(self):
...
+        mi = MultiIndex(levels=[[1, 2], [u'one', u'two']],
+                        labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
+                        names=('foo', 'bar'))
+        renamed = [['foor'], ['barr']]
+        tm.assert_raises_regex(TypeError, message, mi.rename, names=renamed)
+        # With .set_names()
+        tm.assert_raises_regex(TypeError, message, mi.set_names, names=renamed)
+
+    @pytest.mark.parametrize('names', [['a', 'b', 'a'], ['1', '1', '2'],
+                                       ['1', 'a', '1']])
```
jorisvandenbossche: @arminv Is there a reason that you changed those parametrize values to all strings? (I suppose by accident?) 
I am reworking the test in https://github.com/pandas-dev/pandas/pull/21423, so will revert there if this was by accident

arminv: @jorisvandenbossche IIRC  I changed it  [(in this commit)](https://github.com/pandas-dev/pandas/pull/20548/commits/b554bb3f050fd9e1aba4cbf02322ddc9fd4d4f41) because the test was failing, but implementation changed a lot after that commit so I'm not sure if reverting this would cause a problem now

jorisvandenbossche: Seems to be passing there!

PR #20563, opened 2018-03-30
CLN: Use new-style classes instead of old-style classes

jschendel:
Noticed some [lgtm.com alerts](https://lgtm.com/projects/g/pydata/pandas/snapshot/f85a69a8fdaed49747c4db61f5177a296cf290eb/files/scripts/validate_docstrings.py?sort=name&dir=ASC&mode=heatmap&showExcluded=false#L108) about the [`@property` decorator not working in old-style classes](https://lgtm.com/rules/10030086/) in `scripts/validate_docstrings.py`.  I don't think this actually causes any undesired behavior in that script though.

Figured this would be a good time to enforce new-style classes throughout the codebase:
  - Converted old-style classes to new-style classes
      - Basically just inserted `(object)`.
  - Added a check to `lint.sh` for old-style classes, since `flake8` doesn't look to be catching it.
      - Maybe there's a way to enforce this in `flake8` instead?
  - Didn't hit the Cython code; not exactly sure what the new-style vs. old-style conventions are there.

### DIFF
--- a/ci/lint.sh
+++ b/ci/lint.sh
@@ -165,6 +165,14 @@ if [ "$LINT" ]; then
         RET=1
     fi
     echo "Check for deprecated messages without sphinx directive DONE"
+
+    echo "Check for old-style classes"
+    grep -R --include="*.py" -E "class\s\S*[^)]:" pandas scripts
+
+    if [ $? = "0" ]; then
+        RET=1
+    fi
+    echo "Check for old-style classes DONE"
     
 else
     echo "NOT Linting"
--- a/pandas/io/sas/sas_constants.py
+++ b/pandas/io/sas/sas_constants.py
@@ -102,7 +102,7 @@ encoding_names = {29: "latin1", 20: "utf-8", 33: "cyrillic", 60: "wlatin2",
                   61: "wcyrillic", 62: "wlatin1", 90: "ebcdic870"}
 
 
-class index:
+class index(object):
     rowSizeIndex = 0
     columnSizeIndex = 1
     subheaderCountsIndex = 2
--- a/pandas/io/sas/sas_constants.py
+++ b/pandas/io/sas/sas_constants.py
@@ -102,49 +102,49 @@ encoding_names = {29: "latin1", 20: "utf-8", 33: "cyrillic", 60: "wlatin2",
                   61: "wcyrillic", 62: "wlatin1", 90: "ebcdic870"}
 
 
-class index(object):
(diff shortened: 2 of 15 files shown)


ci/lint.sh
```diff
@@ -165,6 +165,14 @@ if [ "$LINT" ]; then
         RET=1
     fi
     echo "Check for deprecated messages without sphinx directive DONE"
+
+    echo "Check for old-style classes"
+    grep -R --include="*.py" -E "class\s\S*[^)]:" pandas scripts
```
jschendel: not a regex wizard, so an extra careful look here would be appreciated

jreback:
maybe we should add some extra flake8 plugins
https://pypi.python.org/pypi?%3Aaction=search&term=flake8&submit=search

would need to add to the requirements_dev and prob modify the ci slightly to install

can u make a separate issue?

mroeschke:
There's a Python package called [hacking](https://github.com/openstack-dev/hacking) that has flake8 extensions that seems to have a lint check for this:

    H238 = hacking.checks.python23:hacking_no_old_style_class

pandas/io/sas/sas_constants.py
```diff
@@ -102,7 +102,7 @@
                   61: "wcyrillic", 62: "wlatin1", 90: "ebcdic870"}
 
 
-class index:
+class index(object):
```
jreback: can you rename this class (capitalize, but needs a differnt name then)

pandas/tests/io/json/test_ujson.py
```diff
@@ -800,7 +800,7 @@ def test_loadFile(self):
             np.array([1, 2, 3, 4]), ujson.load(f, numpy=True))
 
     def test_loadFileLikeObject(self):
-        class filelike:
+        class filelike(object):
```
jreback: can you rename this

pandas/tests/test_resample.py
```diff
@@ -1082,7 +1082,7 @@ def test_resample_how_callables(self):
         def fn(x, a=1):
             return str(type(x))
 
-        class fn_class:
+        class fn_class(object):
```
jreback: can you rename this

jreback (changes requested):
lgtm. some renamings. ping on green.

jschendel:
@jreback : green

jorisvandenbossche:
@jschendel Thanks!

Issue #16827, opened 2017-07-05
Scatter plot with colour_by and size_by variables


nipunbatra:
#### Problem description

Use case: Say we have a df with 4 columns- a, b, c, d. We want to make a scatter plot, with x=a, y=b, color_by=c and size_by=d. Here, if c is a categorical, we get a discrete set of colours and corresponding legend, else a continuous scale. size_by decides the size of the marker.

Such cases are often needed as evidenced by questions on Stack Overflow.

Image below shows an example.
![](http://r-statistics.co/screenshots/ggplot_masterlist_1.png)

I wrote a [blog post](https://nipunbatra.github.io/blog/2017/50-ggplot-python-1.html)(hand-wavy at times- marker size legend) on how to generate such a plot in Pandas. The code below shows how to make a similar plot.

#### Code Sample, a copy-pastable example if possible

```python
import matplotlib.pyplot as plt
import pandas as pd
midwest= pd.read_csv("http://goo.gl/G1K41K") 
# Filtering
midwest= midwest[midwest.poptotal<50000]

fig, ax = plt.subplots()
groups = midwest.groupby('state')

# Tableau 20 Colors
tableau20 = [(31, 119, 180), (174, 199, 232), (255, 127, 14), (255, 187, 120),  
             (44, 160, 44), (152, 223, 138), (214, 39, 40), (255, 152, 150),  
             (148, 103, 189), (197, 176, 213), (140, 86, 75), (196, 156, 148),  
             (227, 119, 194), (247, 182, 210), (127, 127, 127), (199, 199, 199),  
             (188, 189, 34), (219, 219, 141), (23, 190, 207), (158, 218, 229)]
             

# Rescale to values between 0 and 1 
for i in range(len(tableau20)):  
    r, g, b = tableau20[i]  
    tableau20[i] = (r / 255., g / 255., b / 255.)

colors = tableau20[::2]

# Plotting each group 
for i, (name, group) in enumerate(groups):
    group.plot(kind='scatter', x='area', y='poptotal', ylim=((0, 50000)), xlim=((0., 0.1)),
               s=10+group['popdensity']*0.1, # hand-wavy :(
               label=name, ax=ax, color=colors[i])

# Legend for State colours
lgd = ax.legend(numpoints=1, loc=1, borderpad=1, 
            frameon=True, framealpha=0.9, title="state")
for handle in lgd.legendHandles:
    handle.set_sizes([100.0])

[...]

PR #17582, opened 2017-09-18
Feat/scatter by size

VincentAntoine:
- [y ] partial solution to (only size_by_variable is handled, not color_by_variable) #16827 
- [y] tests added: scatter_test.py shows an example of use
- [y] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [n] whatsnew entry

Let me know if any changes are needed.
Thanks!
Vincent

PR #20572, opened 2018-04-01
Feat/scatter by size

VincentAntoine:
- [x] closes part of #16827 : makes bubble plots easy with df.plot.scatter(x='col1', y='col2', s='col3') with nice automatic bubble sizing and bubble size legend
- [x] 2 tests added / passed


VincentAntoine:
I just rebased what I did a few months ago on top of master. Let me know!
Vincent

pandas/plotting/_core.py
```diff
@@ -829,11 +831,30 @@ def _post_plot_logic(self, ax, data):
 class ScatterPlot(PlanePlot):
     _kind = 'scatter'
 
-    def __init__(self, data, x, y, s=None, c=None, **kwargs):
+    def __init__(self, data, x, y, s=None, c=None, size_factor=1, **kwargs):
```
WillAyd: If we are adding a new keyword here then there should be some documentation updates to go along with it - can you take a look at where that needs to be made and bundle it in accordingly?

pandas/plotting/_core.py
```diff
@@ -829,11 +831,30 @@ def _post_plot_logic(self, ax, data):
...
         if s is None:
-            # hide the matplotlib default for size, in case we want to change
-            # the handling of this argument later
+            # Set default size if no argument is given.
             s = 20
+        elif is_hashable(s) and s in data.columns:
```
WillAyd: What does the hashable check do for us here?

pandas/plotting/_core.py
```diff
@@ -829,11 +831,30 @@ def _post_plot_logic(self, ax, data):
...
+            # Handle the case where s is a label of a column of the df.
+            # The data is normalized to 200 * size_factor.
+            size_data = data[s]
+            if is_categorical_dtype(size_data):
```
WillAyd: Was this a discussion in a issue elsewhere? Personally it seems strange to me to leverage the category codes for any kind of sizing as that's not what I would think is in the scope of their services but again that's just my opinion

pandas/plotting/_core.py
```diff
@@ -889,6 +909,68 @@ def _make_plot(self):
             ax.errorbar(data[x].values, data[y].values,
                         linestyle='none', **err_kwds)
 
+    def _sci_notation(self, num):
+        """
+        Returns mantissa and exponent of the number passed in argument.
+        Example:
+        >>> _sci_notation(89278.8924)
+        (8.9, 5.0)
```
WillAyd: This returns (8.9, 4.0) for me - minor typo?

pandas/plotting/_core.py
```diff
@@ -889,6 +909,68 @@ def _make_plot(self):
...
+        """
+        scientific_notation = '{:e}'.format(num)
+        expnt = float(re.search(r'e([+-]\d*)$',
```
WillAyd: Hmm I was hoping there would be a better way to do this than using `re` but couldn't find anything better on a Google search myself as I'm sure you already tried ha. With that said, any way to do this in one search and return the appropriately matched groups instead of doing two passes?

pandas/plotting/_core.py
```diff
@@ -889,6 +909,68 @@ def _make_plot(self):
...
+                                scientific_notation).groups()[0])
+        coef = float(re.search(r'^([+-]?\d\.\d)',
+                               scientific_notation).groups()[0])
+        return coef, expnt
+
+    def _legend_bubbles(self, s_data_max, size_factor, bubble_points):
+        """
+        Computes and returns appropriate bubble sizes and labels for the
+        legend of a  bubble plot. Creates 4 bubbles with round values for the
```
WillAyd: May just be my lack of knowledge, but why is it creating 4 bubbles? Is that true all of the time, even if they only have say three groups of data?

pandas/plotting/_core.py
```diff
@@ -889,6 +909,68 @@ def _make_plot(self):
...
+            (5.5, 7): [6, 3, 1.5, 0.5],
+            (4.5, 5.5): [5, 2, 1, 0.2],
+            (3.5, 4.5): [4, 2, 1, 0.2],
+            (2.5, 3.5): [3, 1, 0.5, 0.2],
+            (1.5, 2.5): [2, 1, 0.5, 0.2],
+            (0, 1.5): [1, 0.5, 0.25, 0.1]
+        }
+        for lower_bound, upper_bound in labels_catalog:
+            if (coef >= lower_bound) & (coef < upper_bound):
+                labels = 10**expnt * np.array(labels_catalog[lower_bound,
```
WillAyd: What happens if `expnt` is negative?

pandas/plotting/_core.py
```diff
@@ -889,6 +909,68 @@ def _make_plot(self):
...
+            (7, 9): [8, 4, 2, 0.5],
+            (5.5, 7): [6, 3, 1.5, 0.5],
+            (4.5, 5.5): [5, 2, 1, 0.2],
+            (3.5, 4.5): [4, 2, 1, 0.2],
+            (2.5, 3.5): [3, 1, 0.5, 0.2],
+            (1.5, 2.5): [2, 1, 0.5, 0.2],
+            (0, 1.5): [1, 0.5, 0.25, 0.1]
+        }
+        for lower_bound, upper_bound in labels_catalog:
+            if (coef >= lower_bound) & (coef < upper_bound):
```
WillAyd: Any reason you chose the bitwise operator instead of the logical `and` operator? While they get you the same place here the latter is more idiomatic

pandas/plotting/_core.py
```diff
@@ -889,6 +909,68 @@ def _make_plot(self):
...
+        return coef, expnt
+
+    def _legend_bubbles(self, s_data_max, size_factor, bubble_points):
+        """
+        Computes and returns appropriate bubble sizes and labels for the
+        legend of a  bubble plot. Creates 4 bubbles with round values for the
+        labels, the largest of which is close to the maximum of the data.
+        """
+        coef, expnt = self._sci_notation(s_data_max)
+        labels_catalog = {
```
WillAyd: Where were these values taken from? Some comments may be helpful

pandas/tests/plotting/test_frame.py
```diff
@@ -1118,6 +1118,40 @@ def test_scatter_colors(self):
...
+                       columns=['x', 'y', 'z'])
+        ax = df.plot.scatter(x='x', y='y', s='z', size_factor=4)
+        bubbles = ax.collections[0]
+        bubble_sizes = bubbles.get_sizes()
+        max_data = df['z'].max()
+        expected_sizes = 200 * 4 * df['z'].values / max_data
+        tm.assert_numpy_array_equal(bubble_sizes, expected_sizes)
+
+    @pytest.mark.slow
+    def test_plot_scatter_with_categorical_s(self):
```
WillAyd: Does missing data affect the categorical display? Perhaps not if the label for NA is -1 and you add 1 in the code above, but it would be good to have a test to explicitly make sure we are OK

WillAyd (changes requested):
I'll admit that I'm not terribly familiar with graphing so this is a little out of my area but figured I could regardless give some feedback.

Thanks for submitting - generally the biggest room for improvement in the documentation. It's not entirely clear how this new feature is to be used, so updating doc strings and providing sample usage will go a long way.

pandas/plotting/_core.py
VincentAntoine: The hashability check ensures backward compatibility: it is possible (in the current stable version of pandas) to pass an array or a Series of sizes to s. For instance you may do something like:

```python
df.plot.scatter(x='height', y='weight', s=df['price'])
```

Without hashability check, `if s in data.columns` throws an error if s is an array or Series. The hashability check ensures that `if s in data.columns` will only be run if s is a legitimate column label.
If there is a better way to handle this, let me know.

pandas/plotting/_core.py
VincentAntoine: This was a suggestion from @TomAugspurger here:
https://github.com/pandas-dev/pandas/pull/17582

But I have to admit that I cannot think of a use case where it would be appropriate to represent category codes by sizes.

pandas/plotting/_core.py
VincentAntoine: You're right.

pandas/plotting/_core.py
VincentAntoine: Indeed I did the same and could not find a better way.
I modified as you suggest to make only one regexp search.

WillAyd: I'd also suggest putting it as a nested function where it is actually used (i.e. in `_legend_bubbles`). Even though it's private I don't think this should be an instance method

pandas/plotting/_core.py
WillAyd: So what happens here if a series is passed into the method? I feel like there is something missing from the if...else logic but I could be wrong

pandas/plotting/_core.py
VincentAntoine: If bubble sizes represent categories, you're right, there should be as many bubbles as there categories. But as said above I can't think of a use case of this, so before coding this, it would be nice to make sure it's actually useful. @TomAugspurger do you remember what you had in mind ?

pandas/plotting/_core.py
VincentAntoine: It works, no problem.

pandas/plotting/_core.py
VincentAntoine: The values were defined by testing until I got something visually pleasing, similar to this graph with R in the original issue submitted here:
https://github.com/pandas-dev/pandas/issues/16827

WillAyd:
Supplemented by #22403

PR #20574, opened 2018-04-02
Adding test_map_missing_mixed to test_apply.py in pandas test suite series

### LINKED ISSUE #20495: Inconsistent behavior of .map()

```python
pd.Series(list('abc') + [np.nan]).map({np.nan: 'bar'})
```

Result:

```
0   NaN
1   NaN
2   NaN
3   NaN
dtype: float64
```

Compare:

```python
pd.Series(list('abc') + [np.nan]).map({'a': 'foo', np.nan: 'bar'})
```

Result:

```
0    foo
1    NaN
2    NaN
3    bar
dtype: object
```

#### Problem description

If only `np.nan` is present in the mapping, then the associated mapping is not carried out. If another value is present, then it is carried out. Forgive me if this is intended behavior for some reason.

#### Expected Output

I'm not sure what is appropriate here.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.2.final.0
python-bits: 32
OS: Linux
OS-release: 4.4.0-116-generic
machine: i686
processor: i686
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.22.0
pytest: None
pip: 8.1.1
setuptools: 20.7.0
Cython: None
numpy: 1.11.0
scipy: 0.17.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: None
patsy: 0.5.0
dateutil: 2.4.2
pytz: 2014.10
blosc: None
bottleneck: None
tables: 3.2.2
numexpr: 2.4.3
feather: None
matplotlib: 1.5.1
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: 4.4.1
html5lib: 0.999
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: 0.6.0
</details>

readyready15728:

- [x] closes #20495
- [x] tests added / passed
- [x] whatsnew entry (n/a)

### DIFF
--- a/pandas/tests/series/test_apply.py
+++ b/pandas/tests/series/test_apply.py
@@ -579,7 +579,7 @@ class TestSeriesMap(TestData):
 
     @pytest.mark.parametrize("mapping,exp", [
         ({np.nan: 'not NaN'}, pd.Series(['not NaN'])),
-        ({'string': 'another string' }, pd.Series(['another string'])),
+        ({'string': 'another string'}, pd.Series(['another string'])),
         ({42: 'the answer'}, pd.Series(['the answer']))])
     def test_map_missing_mixed(self, mapping, exp):
         s = pd.Series(list(mapping.keys())[0])
--- a/pandas/tests/series/test_apply.py
+++ b/pandas/tests/series/test_apply.py
@@ -582,6 +582,7 @@ class TestSeriesMap(TestData):
         (list('abc'), {'a': 'a letter'}, ['a letter'] + [np.nan] * 3),
         (list(range(3)), {0: 42}, [42] + [np.nan] * 3)])
     def test_map_missing_mixed(self, vals, mapping, exp):
+        # GH20495
         s = pd.Series(vals + [np.nan])
         result = s.map(mapping)
 
--- a/pandas/tests/series/test_apply.py
+++ b/pandas/tests/series/test_apply.py
@@ -578,11 +578,11 @@ class TestSeriesMap(TestData):
         tm.assert_series_equal(result, exp)
 
     @pytest.mark.parametrize("vals,mapping,exp", [
-        (list('abc'), {np.nan: 'not NaN'}, pd.Series(['not NaN'])),
-        (list('abc'), {'string': 'another string'}, pd.Series(['another string'])),
-        (list(range(3)), {42: 'the answer'}, pd.Series(['the answer']))])
+        (list('abc'), {np.nan: 'not NaN'}, ['not NaN']),
+        (list('abc'), {'string': 'another string'}, ['another string']),
+        (list(range(3)), {42: 'the answer'}, ['the answer'])])
     def test_map_missing_mixed(self, vals, mapping, exp):
         s = pd.Series(vals + [list(mapping.keys())[0]])
         result = s.map(mapping)
 
-        tm.assert_series_equal(result[-1:].reset_index(drop=True), exp)
+        tm.assert_series_equal(result[-1:].reset_index(drop=True), pd.Series(exp))
--- a/pandas/tests/series/test_apply.py
+++ b/pandas/tests/series/test_apply.py
@@ -577,10 +577,12 @@ class TestSeriesMap(TestData):
         exp = pd.Series(['Asia/Tokyo'] * 25, name='XX')
         tm.assert_series_equal(result, exp)
 
-    def test_map_missing_mixed(self):
-        s = pd.Series(list('abc') + [np.nan])
+    @pytest.mark.parametrize("mapping,exp", [
+        ({np.nan: 'not NaN'}, pd.Series(['not NaN'])),
+        ({'string': 'another string' }, pd.Series(['another string'])),
+        ({42: 'the answer'}, pd.Series(['the answer']))])
+    def test_map_missing_mixed(self, mapping, exp):
+        s = pd.Series(list(mapping.keys())[0])
+        result = s.map(mapping)
 
-        result = s.map({np.nan: 'not NaN'})
-        result2 = s.map({'a': 42, np.nan: 'not NaN'})
-
-        tm.assert_series_equal(result[-1:], result2[-1:])
+        tm.assert_series_equal(result, exp)
--- a/pandas/tests/series/test_apply.py
+++ b/pandas/tests/series/test_apply.py
@@ -577,12 +577,12 @@ class TestSeriesMap(TestData):
         exp = pd.Series(['Asia/Tokyo'] * 25, name='XX')
         tm.assert_series_equal(result, exp)
 
-    @pytest.mark.parametrize("mapping,exp", [
-        ({np.nan: 'not NaN'}, pd.Series(['not NaN'])),
-        ({'string': 'another string'}, pd.Series(['another string'])),
-        ({42: 'the answer'}, pd.Series(['the answer']))])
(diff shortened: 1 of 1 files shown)


WillAyd (changes requested):
Based off of the comments in the linked issue I think there are (at least) three scenarios we wanted to test:

1. Mapping the NA value to a new value
2. Mapping a str value to a new value AND
3. Mapping an int value to a new value

You covered the first two but left out the third so we should add that.

With that said, whenever you have tests that are repetitive in nature and varying just slightly you should be thinking about parametrizing them. In this case, I think you should parametrize the test with values of "mapping,exp" where the former is the dict you want to use in the `map` function and the latter is the result you expect (should be a series in each case). Below is a relatively close example of how this works - take a look at that and see if you can apply something similar here

https://github.com/pandas-dev/pandas/blob/4efb39f01f5880122fa38d91e12d217ef70fad9e/pandas/tests/groupby/test_groupby.py#L2063

readyready15728:
Alright, I think I have something here. Is there any convenient way for me to run just an individual test suite rather than all tests in their entirety?

WillAyd:
`pytest path.to.the.module::ClassName::TestName` would be the most granular, though you can strip off any of those elements to move up a level and run more tests

readyready15728:
The parametrized test checks out.

pandas/tests/series/test_apply.py
```diff
@@ -576,3 +576,13 @@ def f(x):
         result = s.map(f)
         exp = pd.Series(['Asia/Tokyo'] * 25, name='XX')
         tm.assert_series_equal(result, exp)
+
+    @pytest.mark.parametrize("mapping,exp", [
+        ({np.nan: 'not NaN'}, pd.Series(['not NaN'])),
+        ({'string': 'another string' }, pd.Series(['another string'])),
+        ({42: 'the answer'}, pd.Series(['the answer']))])
+    def test_map_missing_mixed(self, mapping, exp):
+        s = pd.Series(list(mapping.keys())[0])
```
WillAyd: Nice job on the parametrization but in the process of updating this line I think we are losing sight of what we are trying to test. Originally you were constructing this series from `list('abcd')` and appending an NA record. Now all you are doing is using the key of your dict as the series value, but that's not the same test.

Now that I'm thinking about it it's probably good to also add a "vals" parameter in front of the mapping. For the first two scenarios you can use `list('abcd')` as you had before and for the third use `range(4)`.  Then just make your first line `s = pd.Series(vals + [np.nan])`

WillAyd (changes requested):

pandas/tests/series/test_apply.py
readyready15728: OK, I'll give that a shot.

pandas/tests/series/test_apply.py
```diff
@@ -576,3 +576,13 @@ def f(x):
         result = s.map(f)
         exp = pd.Series(['Asia/Tokyo'] * 25, name='XX')
         tm.assert_series_equal(result, exp)
+
+    @pytest.mark.parametrize("vals,mapping,exp", [
+        (list('abc'), {np.nan: 'not NaN'}, ['not NaN']),
+        (list('abc'), {'string': 'another string'}, ['another string']),
```
WillAyd: Your mapping should contain one of the values in the series, so use 'a' instead of 'string'

pandas/tests/series/test_apply.py
```diff
@@ -576,3 +576,13 @@ def f(x):
...
+        (list(range(3)), {42: 'the answer'}, ['the answer'])])
```
WillAyd: Same comment as above, use 1 instead of 42. Just for consistency make the value numeric as well instead of 'the answer'

pandas/tests/series/test_apply.py
```diff
@@ -576,3 +576,13 @@ def f(x):
...
+    def test_map_missing_mixed(self, vals, mapping, exp):
+        s = pd.Series(vals + [list(mapping.keys())[0]])
```
WillAyd: Don't use the mapping keys here. `s = pd.Series(vals + [np.nan])` is all you need

pandas/tests/series/test_apply.py
```diff
@@ -576,3 +576,13 @@ def f(x):
...
+        result = s.map(mapping)
+
+        tm.assert_series_equal(result[-1:].reset_index(drop=True), pd.Series(exp))
```
WillAyd: Think through the exp values you are passing in. They should obviously match the shape of your input but replace with NA values where appropriate.

For your first example, if you did `list('abc')` as your val `{'a': 'foo'}` as your mapping then your exp would be `['foo', np.nan, np.nan, np.nan]`.

I'm not sure what you are trying to do with `result[-1:].reset_index(drop=True)` but that's getting way too complicated. If you follow all of the above steps you can just do `tm.assert_series_equal(result, pd.Series(exp))`

WillAyd (changes requested):
Make sure you get the tests to pass locally before pushing to GitHub. These have some logical errors that you need to work through

pandas/tests/series/test_apply.py
readyready15728: Riiight

readyready15728:
Alright, I just learned that supplying `-s` to pytest allows output from `print()` statements to be shown. With that knowledge I was able to check whether the values being generated made sense before committing this time. Sorry for jumping the gun.

pandas/tests/series/test_apply.py
```diff
@@ -576,3 +576,13 @@ def f(x):
         result = s.map(f)
         exp = pd.Series(['Asia/Tokyo'] * 25, name='XX')
         tm.assert_series_equal(result, exp)
+
+    @pytest.mark.parametrize("vals,mapping,exp", [
+        (list('abc'), {np.nan: 'not NaN'}, [np.nan] * 3 + ['not NaN']),
+        (list('abc'), {'a': 'a letter'}, ['a letter'] + [np.nan] * 3),
+        (list(range(3)), {0: 42}, [42] + [np.nan] * 3)])
+    def test_map_missing_mixed(self, vals, mapping, exp):
+        s = pd.Series(vals + [np.nan])
```
WillAyd: Can you add `# GH20495` as a comment on its own line right below the method definition?

WillAyd (changes requested):
Very minor edit but otherwise lgtm. CI should run and a core maintainer will be able to review thereafter for merge ability.

Thanks for trying your hand at your first PR!

pandas/tests/series/test_apply.py
readyready15728: Between `def ...` and `s ...`?

WillAyd: Yes - you can look at `test_with_nested_series` in the same module for reference

pandas/tests/series/test_apply.py
readyready15728: The tests *did* pass

Just not like they should have

TomAugspurger (approved):
Just need the comment referencing the issue, then we're good.

readyready15728:
All set

Issue #19342, opened 2018-01-22
Series constructor doesn't validate / reindex categorical


TomAugspurger:
```python
In [1]: import pandas as pd

In [2]: c = pd.Series(pd.Categorical(['a', 'a']), index=[0, 1, 2])

In [3]: c
Out[3]:
0    a
1    a
2
dtype: category
Categories (1, object): [a]

In [4]: len(c)
Out[4]: 3

In [5]: len(c.values)
Out[5]: 2
```

This should probably raise.

cbertinato:
I'm going to make a pull request for this one, but just wanted to check on the strategy, in case I'm missing something else. 

This appears to be an issue just for Categoricals in Series. Attempting to make a DataFrame of Categoricals with a mismatched index raises as expected. The issue is due to the fact that `CategoricalBlock` subclasses `NonConsolidatableMixIn` which sets `_validate_ndim = False`, so the check isn't done when constructing a Series with a Categorical.  Setting that value to true elsewhere to fix this, say in `CategoricalBlock`, breaks functionality elsewhere. So it seems that a shallow fix in Series would be most appropriate.

jschendel:
I think this should be closed as the related PR has been merged?

TomAugspurger:
Yep, thanks.

PR #20577, opened 2018-04-02
TST: remove skip for values/index length mismatch in ExtensionArray tests

jorisvandenbossche:
https://github.com/pandas-dev/pandas/issues/19342 is fixed in the meantime


TomAugspurger:
Thanks!

Issue #20576, opened 2018-04-02
ExtensionArray series.align(frame) not working


jorisvandenbossche:
Aligning a Series containing extension array data with a dataframe does not work if there is actually something to be aligned:

```
In [27]: from pandas.tests.extension.decimal.array import DecimalArray, make_data

In [28]: dec_arr = DecimalArray(make_data()[:3])

In [29]: s = pd.Series(dec_arr)

In [30]: s
Out[30]: 
0    0.29242561210243966929311909552779980003833770...
1    0.34798224977276304148432473084540106356143951...
2    0.04963128775050906771326708621927537024021148...
dtype: decimal

In [31]: frame = pd.DataFrame({'col': np.arange(3)})

In [32]: s.align(frame)
Out[32]: 
(0    0.29242561210243966929311909552779980003833770...
 1    0.34798224977276304148432473084540106356143951...
 2    0.04963128775050906771326708621927537024021148...
 dtype: decimal,    col
 0    0
 1    1
 2    2)

In [33]: frame = pd.DataFrame({'col': np.arange(4)})

In [34]: s.align(frame)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-34-94a43a65d337> in <module>()
----> 1 s.align(frame)

/home/joris/scipy/pandas/pandas/core/series.py in align(self, other, join, axis, level, copy, fill_value, method, limit, fill_axis, broadcast_axis)
   3231                                          fill_value=fill_value, method=method,
   3232                                          limit=limit, fill_axis=fill_axis,
-> 3233                                          broadcast_axis=broadcast_axis)
   3234 
   3235     def rename(self, index=None, **kwargs):

/home/joris/scipy/pandas/pandas/core/generic.py in align(self, other, join, axis, level, copy, fill_value, method, limit, fill_axis, broadcast_axis)
   7174                                      copy=copy, fill_value=fill_value,
   7175                                      method=method, limit=limit,
-> 7176                                      fill_axis=fill_axis)
   7177         elif isinstance(other, Series):
   7178             return self._align_series(other, join=join, axis=axis, level=level,

/home/joris/scipy/pandas/pandas/core/generic.py in _align_frame(self, other, join, axis, level, copy, fill_value, method, limit, fill_axis)
   7210         left = self._reindex_with_indexers(reindexers, copy=copy,
   7211                                            fill_value=fill_value,
-> 7212                                            allow_dups=True)
[...]

PR #20580, opened 2018-04-02
BUG: Fixed Series.align(frame) with ExtensionArray

TomAugspurger:
Closes https://github.com/pandas-dev/pandas/issues/20576

cc @jorisvandenbossche

### DIFF
--- a/pandas/tests/extension/base/reshaping.py
+++ b/pandas/tests/extension/base/reshaping.py
@@ -1,4 +1,5 @@
 import pytest
+import numpy as np
 
 import pandas as pd
 from pandas.core.internals import ExtensionBlock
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
That fixed it! Thanks

pandas/tests/extension/base/reshaping.py
```diff
@@ -64,6 +65,20 @@ def test_align_frame(self, data, na_value):
         self.assert_frame_equal(r1, e1)
         self.assert_frame_equal(r2, e2)
 
+    def test_align_series_frame(self, data, na_value):
+        # /Users/taugspurger/sandbox/pandas-ip/pandas/pandas/tests/extension/base/reshaping.py
```
jorisvandenbossche: suppose this is a left over?

pandas/tests/extension/category/test_categorical.py
```diff
@@ -75,6 +75,10 @@ def test_align(self, data, na_value):
     def test_align_frame(self, data, na_value):
         pass
 
+    @pytest.mark.skip(reason="Unobserved categories preseved in concat.")
+    def test_align_series_frame(self, data, na_value):
```
jorisvandenbossche: Isn't this only because the way the test is constructed? You can also do it with the full of `data` instead of only the first 3 elements. Then this is not an issue?

jorisvandenbossche (approved):

PR #20582, opened 2018-04-02
TST: add tests for take() on empty arrays

jorisvandenbossche:
Another noticed during geopandas testing: `ExtensionArray.take` needs to be able to handle the case where it is empty.

Added a example implementation for Decimal/JSONArray. 
And apparently, this was also failing for Categorical, so fixed that as well.


pandas/core/arrays/categorical.py
```diff
@@ -1740,6 +1740,17 @@ def take_nd(self, indexer, allow_fill=True, fill_value=None):
...
+        # take on empty array
+        if not len(self):
+            # only valid if result is an all-missing array
+            if (indexer == -1).all():
+                codes = [-1] * len(indexer)
+            else:
+                raise IndexError(
+                    "cannot do a non-empty take from an empty array.")
+
         codes = take_1d(self._codes, indexer, allow_fill=True, fill_value=-1)
```
jorisvandenbossche: For the internal use cases (Categorical at the moment, later maybe more arrays), I could also push this test to `take_1d`. But, that is also used by many other things, so not sure I can just do that.

pandas/tests/extension/base/getitem.py
```diff
@@ -120,3 +121,11 @@ def test_take_sequence(self, data):
...
         assert result.iloc[1] == data[1]
         assert result.iloc[2] == data[3]
+
+    def test_take_empty_missing(self, data, na_value, na_cmp):
+        empty = data[:0]
+        result = empty.take([-1])
+        na_cmp(result[0], na_value)
+
+        with tm.assert_raises_regex(IndexError, "cannot do a non-empty take"):
+            empty.take([0, 1])
```
jorisvandenbossche: @TomAugspurger I didn't find any existing tests for `take`, so not sure the indexing tests is the best place (maybe rather the `BaseMethodsTests`). 
And, I could also add tests for the actual use cases where you get this (eg `reindex` on an empty series)

pandas/core/arrays/categorical.py
jreback: this should not be here rather in take1d (i see you comment)

pandas/tests/extension/decimal/array.py
```diff
@@ -75,9 +75,19 @@ def isna(self):
         return np.array([x.is_nan() for x in self.values])
 
     def take(self, indexer, allow_fill=True, fill_value=None):
+
         indexer = np.asarray(indexer)
         mask = indexer == -1
 
+        # take on empty array
+        if not len(self):
```
jreback: same

jreback (changes requested):

pandas/tests/extension/decimal/array.py
jorisvandenbossche: DecimalArray.take does not use `take_1d`, so it's not possible to move it

pandas/tests/extension/base/getitem.py
TomAugspurger: This seems like the right place for testing a behavior specific to take.

pandas/tests/extension/decimal/array.py
```diff
@@ -62,7 +62,7 @@ def __len__(self):
         return len(self.values)
 
     def __repr__(self):
-        return repr(self.values)
+        return "DecimalArray: " + repr(self.values)
```
TomAugspurger: Thanks, meant to add this a while ago :)

TomAugspurger (commented):
Could you document this behavior in ExtensionArray.take, either the docstring or as a comment?

pandas/core/arrays/categorical.py
jorisvandenbossche: I simplified the check a bit here (as it only needed to check the empty and not-all -1 case, as when indexer is all -1 `take_1d` works as expected on empty arrays). 

@jreback I am not sure if I should move it to `take_1d`. It depends on the guarantees we want to give (ourselves) for `take_1d` (if it should do such bounds checking). It might be that currently all codes that uses `take_1d/nd` already does this bounds checking (I did not find a practical use case (apart from Categorical.take itself) where we run into this bug), and then doing another one in `take_nd` would give a performance penalty.

pandas/core/arrays/base.py
```diff
@@ -477,6 +486,16 @@ def take(self, indexer, allow_fill=True, fill_value=None):
...
                mask = indexer == -1
+
+               # take on empty array not handled as desired by numpy
+               if not len(self):
+                   # only valid if result is an all-missing array
+                   if mask.all():
+                       return type(self)([self._na_value] * len(indexer))
+                   else:
+                       raise IndexError(
+                           "cannot do a non-empty take from an empty array.")
```
jorisvandenbossche: @TomAugspurger Adding the full example is maybe a bit too much (it defeats a bit the purpose of the small illustrative example)? On the other hand, it is needed for a correct implementation.

TomAugspurger: Agreed... I think that just having a comment noting the correct thing to do for `.take` on an empty EA is fine.

pandas/core/arrays/base.py
```diff
@@ -477,6 +486,16 @@ def take(self, indexer, allow_fill=True, fill_value=None):
            def take(self, indexer, allow_fill=True, fill_value=None):
                indexer = np.asarray(indexer)
                mask = indexer == -1
+
+               # take on empty array not handled as desired by numpy
+               if not len(self):
+                   # only valid if result is an all-missing array
+                   if mask.all():
+                       return type(self)([self._na_value] * len(indexer))
+                   else:
```
jreback: you don't need the else, just raise

pandas/core/arrays/categorical.py
```diff
@@ -1740,6 +1740,12 @@ def take_nd(self, indexer, allow_fill=True, fill_value=None):
         # but is passed thru internally
         assert isna(fill_value)
 
+        indexer = np.asarray(indexer)
+
+        # take on empty array only valid if result is an all-missing array
+        if not len(self) and not (indexer == -1).all():
```
jreback: this should be in ``take_1d``

pandas/core/arrays/base.py
jreback: is this raise condition hit in a test?

pandas/tests/extension/decimal/array.py
```diff
@@ -78,6 +78,15 @@ def take(self, indexer, allow_fill=True, fill_value=None):
         indexer = np.asarray(indexer)
         mask = indexer == -1
 
+        # take on empty array
+        if not len(self):
+            # only valid if result is an all-missing array
+            if mask.all():
+                return type(self)([self._na_value] * len(indexer))
+            else:
```
jreback: no else needed. is the raise hit?

pandas/tests/extension/decimal/array.py
```diff
@@ -78,6 +78,15 @@ def take(self, indexer, allow_fill=True, fill_value=None):
         indexer = np.asarray(indexer)
         mask = indexer == -1
 
+        # take on empty array
+        if not len(self):
+            # only valid if result is an all-missing array
+            if mask.all():
```
jreback: writing take is a fair amount of effort on the extension writer. meaning additional complexity and have to handle edge cases. take is a pure indexing operation and so ideally should be implemetned as high up in the stack as possible (meaning in the base class for EA) if the EA writer wants to override then ok, but it should just work generally. All of the machinery is already there. The point of the EA is to make it easy, having to handle all of these edge cases is not easy.

jreback (changes requested):

pandas/tests/extension/decimal/array.py
jreback: then you need to ask yourself if this is the correct implementation. pandas implementst take_1d for exactly this reason.

jreback (changes requested):

pandas/tests/extension/decimal/array.py
TomAugspurger: > All of the machinery is already there.

Assuming you have an ndarray. Do we want to provide a default implementation that does an `astype(object)` to convert to an ndarray of scalars, `take` on that, and then construct a new EA?

I'm going to make a PR documenting which methods use `.astype(object)`, so that the EA author should consider overriding them if performance is a concern.

pandas/core/arrays/base.py
jorisvandenbossche: Yes, see the single test I added in this PR: it tests both the all missing case and this error

pandas/tests/extension/decimal/array.py
jorisvandenbossche: `take_1d` is not a public function. Do you want to expose it publicly?

pandas/tests/extension/decimal/array.py
jorisvandenbossche: > Assuming you have an ndarray. Do we want to provide a default implementation that does an astype(object) to convert to an ndarray of scalars, take on that, and then construct a new EA?

It will almost never be what you want I think. It could still serve as an example implementation of course, like we now have one in the docstring.

pandas/core/arrays/base.py
jorisvandenbossche: Yeah, the 'else' clause is indeed not essential. I adapted the error message slightly ("from empty axes" -> "from empty array") because I did not find it the best for ExtensionArrays, but given the minor change I can certainly leave it out for code simplicity.

jreback: pls make this change.

also combine the comments, this doesn't read very well

pandas/core/arrays/categorical.py
jreback: pls make this change

pandas/core/arrays/categorical.py
```diff
@@ -1740,6 +1740,12 @@ def take_nd(self, indexer, allow_fill=True, fill_value=None):
...
+            raise IndexError("cannot do a non-empty take from an empty array.")
```
jreback: this also needs a test

jreback (changes requested):

pandas/core/arrays/base.py
jorisvandenbossche: yep, did it already locally, didn't push yet

pandas/core/arrays/categorical.py
jorisvandenbossche: This is already tested in the extension array tests for categorical (do you want a duplicate test in the categorical tests?)

pandas/core/arrays/categorical.py
jorisvandenbossche: Did you see my answer/question the first time you asked this? 

I gave reasons not to do this, so at least please answer to them.

jorisvandenbossche:
Regarding the discussion whether this "indexer of all -1 in case of empty array"-check should be in `take_1d` or in the array's `take`. 
I agree ideally we should do this in `take_1d`. I just want to note that, currently, we *don't do any* bounds checking in `take_1d` (also not for eg too large indices). So it might be all methods calling `take_1d` already do this and we didn't add it to `take_1d` for performance reasons, there might be other reasons, .. I don't know. 

It's just a bit a bigger endeavour to fix that. Of course I can rather easily add this two-line check to `take_1d`, but then we get into a state where `take_1d` does only a partial incomplete bounds check. So ideally we should decide on how low or high level we see `take_1d` (or eg add another `take` method that is more strict in checking input, to not have to change existing internal code), and if we decide that `take_1d` is a high level function that should guarantee bounds checking, update all code that uses it.

pandas/core/arrays/categorical.py
jreback: I already responsded. ``take_1d`` can certainly be used in an implementation of this. This is the general implementation of virtually all take behavior, so naturally a check for no indexers should happen there, rather than in each higher level implementation (e.g. EA). The point is that the way you are advocating is simply more complex / buggy and adds more code. Virtually all EA methods should use API's that pandas already has. Should these be public. Sure if possible. But for something like this we *don't* want to make the actual implementation public as it has certain guarantees. However I don't see any problem with an internal implementation (like EA) using it. it fact I would say they *have* / *should* use it. otherwise how else would categorical/DTI/II be implemented at all?

jorisvandenbossche:
Putting @jreback's comment here to not have it lost in a hidden inline comment:

> I already responsded. `take_1d` can certainly be used in an implementation of this. This is the general implementation of virtually all take behavior, so naturally a check for no indexers should happen there, rather than in each higher level implementation (e.g. EA). The point is that the way you are advocating is simply more complex / buggy and adds more code. Virtually all EA methods should use API's that pandas already has. Should these be public. Sure if possible. But for something like this we don't want to make the actual implementation public as it has certain guarantees. However I don't see any problem with an internal implementation (like EA) using it. it fact I would say they have / should use it. otherwise how else would categorical/DTI/II be implemented at all?

If you can point me to an answer that is not just "this should be in take_1d", I don't find it :)

I removed the changes in this PR regarding Categorical (and opened a separate issue for that: https://github.com/pandas-dev/pandas/issues/20664), so only adding tests for the ExtensionArray and the test implementations. 
So we can keep bigger `take_1d` changes for another PR.

`take_1d` is internally used in many cases where the `indexer` passed to `take_1d` is the result of eg `get_indexer`, so this ensures that the indexer is correct, and for those no out-of-bounds checks are needed.

You advocate for that ExtensionArrays should use pandas take implementation, but at the same time say that you don't want to make it public? That seems to contradict each other.

If we want to expose our `take` implementation to extension array authors (and maybe we are that ourselves as well), I would personally create a new `take` functions which just calls the existing `take_1d`, but additionally does first bounds checking. 
That way, we keep on using the faster `take_1d` functions internally where we know bounds checks are not needed, and we can make the new `take` method public for external usage as well.

pandas/core/arrays/base.py
jorisvandenbossche: @TomAugspurger I added some explanation in the docstring, but I decided to keep the code example as well: it is now a bit shorter + it is actually really needed to get a correct implementation.

jorisvandenbossche:
@jreback I repeated some of my comments above about `take_1d` / fixing a `take` implementation itself so it can be used in extension arrays, in https://github.com/pandas-dev/pandas/issues/20640. 
So maybe we can continue the discussion on that topic there? And then here only further discuss the actual changes in this PR (I removed everything related to Categorical, I am only adding some generic `take` tests and fixing it for the test example arrays).

jreback (approved):

jreback:
lgtm. though seems to be failing the numpy master build?

jorisvandenbossche:
> though seems to be failing the numpy master build?

Seems to be unrelated clipboard failures

jorisvandenbossche:
Appveyor failure is unrelated and was succeeding before.

PR #13440, opened 2016-06-14
BUG: alignment of CategoricalIndex

toobaz:
- [x] closes #13365
- [x] tests added / passed
- [x] passes `git diff upstream/master | flake8 --diff`
- [ ] whatsnew entry

@jreback take this just as a wild guess of how to fix #13365 . If the approach makes sense, I will add tests, whatsnew and some more docs.

PR #20583, opened 2018-04-02
API: categorical grouping will no longer return the cartesian product

### LINKED ISSUE #8869: BUG: Multi grouper containing a categorical not dropped from index when using groupby with as_index=False
Hello,

The following example definitely seems like a bug. The grouper is not dropped from the index of the resulting DataFrame, even when `as_index = False`.

Actually, even the aggregation step completely fails, so there may be more to it, as shown in this example.

``` python
import pandas as pd
d = {'foo': [10, 8, 4, 8, 4, 1, 1], 'bar': [10, 20, 30, 40, 50, 60, 70],
     'baz': ['d', 'c', 'e', 'a', 'a', 'd', 'c']}
df = pd.DataFrame(d)
cat = pd.cut(df['foo'], np.linspace(0, 10, 3))
df['range'] = cat
groups = df.groupby(['range', 'baz'], as_index=False)
result = groups.agg('mean')
```

``` python
result
```

|  |  | range | baz | bar | foo |
| --- | --- | --- | --- | --- | --- |
| range | baz |  |  |  |  |
| (0, 5] | a | NaN | NaN | NaN | NaN |
|  | c | NaN | NaN | NaN | NaN |
|  | d | NaN | NaN | NaN | NaN |
|  | e | NaN | NaN | NaN | NaN |
| (5, 10] | a | NaN | NaN | NaN | NaN |
|  | c | NaN | NaN | NaN | NaN |
|  | d | NaN | NaN | NaN | NaN |
|  | e | NaN | NaN | NaN | NaN |

Compare to the expected result:

``` python
groups2 = df.groupby(['range', 'baz'], as_index=True)
expected = groups2.agg('mean').reset_index()
```

``` python
expected
```

|  | range | baz | bar | foo |
| --- | --- | --- | --- | --- |
| 0 | (0, 5] | a | 50 | 4 |
| 1 | (0, 5] | c | 70 | 1 |
| 2 | (0, 5] | d | 60 | 1 |
| 3 | (0, 5] | e | 30 | 4 |
| 4 | (5, 10] | a | 40 | 8 |
| 5 | (5, 10] | c | 20 | 8 |
| 6 | (5, 10] | d | 10 | 10 |
| 7 | (5, 10] | e | NaN |  |

``` python
pd.__version__
Out[181]: '0.15.1'
```

### LINKED ISSUE #14942: groupby with category column and two additional columns eats up all main memory

```python
import numpy as np
import pandas as pd

df = pd.DataFrame({
    'cat': np.random.randint(0, 255, size=3000000),
    'int_id': np.random.randint(0, 255, size=3000000),
    'other_id': np.random.randint(0, 10000, size=3000000),
    'foo': 0
})
df['cat'] = df.cat.astype(str).astype('category')

# killed after 6 minutes of 100% cpu and  90G maximum main memory usage
grouped = df.groupby(['cat', 'int_id', 'other_id']).count()
```
#### Problem description

The problem occurs only when I try to group by at least three columns. For two and one columns, it works.

If I replace the categorical column with an integer one, the groupby only takes about 2 seconds and does not use so much memory. This is also the workaround I use currently when I have to group by columns where one of the columns has the type category. But this is kind of ugly.

#### Expected Output

```python
from collections import defaultdict

class CategoryColGrouper:
    def __init__(self):
        self.lookups = defaultdict(dict)
        self.max_num = {}
        self.seen_cat_cols = set()

    def get_data_type(self, max_num):
        if max_num < 2 ** 8:
            return np.uint8
        elif max_num < 2 ** 16:
            return np.uint16
        elif max_num < 2 ** 32:
            return np.uint32
        else:
            return np.uint64

    def get_cat_cols(self, df):
        cat_cols = []
        for col in df.columns:
            if str(df[col].dtype) == 'category':
                cat_cols.append(col)
        return cat_cols

    def get_lookup_for_col(self, df, col):
        col_lookup = self.lookups[col]
        max_num = self.max_num.get(col, 0)

        missing = []
        for value in df[col].unique():
            if value not in col_lookup:
                missing.append(value)

        for num, value in enumerate(missing):
            new_num = num + max_num
            col_lookup[value] = new_num
            self.max_num[col] = new_num

        return col_lookup

    def make_groupable(self, df):
        cat_cols = self.get_cat_cols(df)
        self.seen_cat_cols = self.seen_cat_cols.union(cat_cols)
        for cat_col in cat_cols:
            col_lookup = self.get_lookup_for_col(df, cat_col)
            df[cat_col] = df[cat_col].apply(lambda x: col_lookup[x])
            data_type = self.get_data_type(self.max_num[cat_col])
            df[cat_col] = df[cat_col].astype(data_type)
        return df

    def restore_category_cols(self, df):
        for cat_col in self.seen_cat_cols:
            col_lookup = self.lookups[cat_col]
            rcol_lookup = {v: k for k, v in col_lookup.items()}
            df[cat_col] = df[cat_col].apply(lambda x: rcol_lookup[x])
            df[cat_col] = df[cat_col].astype('category')
        return df

ccg = CategoryColGrouper()
df = ccg.make_groupable(df)

grouped = df.groupby(['cat', 'int_id', 'other_id']).count()

df = ccg.restore_category_cols(df)
```

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.2.final.0
python-bits: 64
OS: Linux
OS-release: 3.2.0-4-amd64
machine: x86_64
processor: 
byteorder: little
LC_ALL: C
LANG: en_US.UTF-8
LOCALE: None.None

pandas: 0.19.1
nose: None
pip: 9.0.1
setuptools: 32.1.0.post20161217
Cython: 0.25.2
numpy: 1.11.3
scipy: 0.18.1
statsmodels: 0.6.1
xarray: None
IPython: 5.1.0
sphinx: None
patsy: 0.4.1
dateutil: 2.6.0
pytz: 2016.10
blosc: None
bottleneck: None
tables: None
numexpr: None
matplotlib: 2.0.0rc2
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
httplib2: None
apiclient: None
sqlalchemy: 1.1.4
pymysql: None
psycopg2: None
jinja2: 2.8
boto: None
pandas_datareader: None

</details>

### LINKED ISSUE #15217: Categorical Column GroupBy agg with as_index=False produces NaN rows 7.5X Slower with unexpected extra Cardinality
```python
import numpy as np
import pandas as pd

df = pd.DataFrame(np.random.randint(10,100, size=(200,6)), columns=['C'+str(i) for i in range(6)])
df['C0'] = ['A','B','C','D']*50
df['C1'] = ['E','F']*100
df['C2'] = ['H','I','J','K', 'L']*40

for col in df.columns[:3]:
    df[col] = df[col].astype('category')

df.info()

RangeIndex: 200 entries, 0 to 199
Data columns (total 6 columns):
C0    200 non-null category
C1    200 non-null category
C2    200 non-null category
C3    200 non-null int32
C4    200 non-null int32
C5    200 non-null int32
dtypes: category(3), int32(3)
memory usage: 3.1 KB

%time ix_true = df.groupby(df.columns.tolist()[:3], as_index=True)['C5'].max()
Wall time: 2 ms

ix_true.shape
(20,)

%time ix_false = df.groupby(df.columns.tolist()[:3], as_index=False)['C5'].max()
Wall time: 15 ms

ix_false.shape
(40, 4)

ix_true
C0  C1  C2
A   E   H     93
        I     99
        J     88
        K     91
        L     94
B   F   H     98
        I     89
        J     94
        K     92
        L     96
C   E   H     96
        I     96
        J     85
        K     88
        L     98
D   F   H     96
        I     84
        J     71
        K     96
        L     94
Name: C5, dtype: int32

ix_false
	C0	C1	C2	C5
0	A	E	H	93.0
1	A	E	I	99.0
2	A	E	J	88.0
3	A	E	K	91.0
4	A	E	L	94.0
5	A	F	H	NaN
6	A	F	I	NaN
7	A	F	J	NaN
8	A	F	K	NaN
9	A	F	L	NaN
10	B	E	H	NaN
11	B	E	I	NaN
12	B	E	J	NaN
13	B	E	K	NaN
14	B	E	L	NaN
15	B	F	H	98.0
16	B	F	I	89.0
17	B	F	J	94.0
18	B	F	K	92.0
19	B	F	L	96.0
20	C	E	H	96.0
21	C	E	I	96.0
22	C	E	J	85.0
23	C	E	K	88.0
24	C	E	L	98.0
25	C	F	H	NaN
26	C	F	I	NaN
27	C	F	J	NaN
28	C	F	K	NaN
29	C	F	L	NaN
30	D	E	H	NaN
31	D	E	I	NaN
32	D	E	J	NaN
33	D	E	K	NaN
34	D	E	L	NaN
35	D	F	H	96.0
36	D	F	I	84.0
37	D	F	J	71.0
38	D	F	K	96.0
39	D	F	L	94.0

```

#### Problem description

Using ```as_index=False``` in ```df.groupby(df.columns.tolist()[:3], as_index=False)['C5'].max()``` with categorical columns produces NaN output rows.

#### Expected Output

I expect that the output should not contain any extra Cardinality explosion and have the same number of rows. as the ```as_index=True``` case.

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.2.final.0
python-bits: 64
OS: Windows
OS-release: 7
machine: AMD64
processor: Intel64 Family 6 Model 63 Stepping 2, GenuineIntel
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.19.2
nose: 1.3.7
pip: 8.1.1
setuptools: 20.10.1
Cython: 0.24.1
numpy: 1.11.2
scipy: 0.18.1
statsmodels: 0.6.1
xarray: 0.7.1
IPython: 4.2.0
sphinx: 1.3.6
patsy: 0.4.0
dateutil: 2.5.0
pytz: 2015.7
blosc: None
bottleneck: 1.0.0
tables: 3.3.0
numexpr: 2.6.1
matplotlib: 1.5.1
openpyxl: 2.2.5
xlrd: 0.9.4
xlwt: 1.0.0
xlsxwriter: None
lxml: 3.6.4
bs4: 4.4.1
html5lib: None
httplib2: None
apiclient: None
sqlalchemy: 1.0.12
pymysql: None
psycopg2: None
jinja2: 2.8
boto: None
pandas_datareader: None

</details>

### LINKED ISSUE #17594: groupby with categorical type returns all combinations

```python
import pandas as pd                                                                                                                                                                                                                    
df = pd.DataFrame({'a': ['x','x','y'], 'b': [0,1,0], 'c': [7,8,9]})                                                                                                                                                                    
print(df.groupby(['a','b']).mean().reset_index())                                                                                                                                                                                      
df['a'] = df['a'].astype('category')                                                                                                                                                                                                   
print(df.groupby(['a','b']).mean().reset_index())
```
Returns two different results:

```
   a  b  c
0  x  0  7
1  x  1  8
2  y  0  9

   a  b    c
0  x  0  7.0
1  x  1  8.0
2  y  0  9.0
3  y  1  NaN
```
#### Problem description

Performing a groupby with a categorical type returns all combination of the groupby columns. This is a problem in my actual application as it results in a massive dataframe that is mostly filled with nans. I would also prefer not to move off of category dtype since it provides necessary memory savings.
#### Expected Output

```
   a  b  c
0  x  0  7
1  x  1  8
2  y  0  9

   a  b  c
0  x  0  7
1  x  1  8
2  y  0  9
```

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.1.final.0
python-bits: 64
OS: Linux
OS-release: 4.10.0-26-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.20.3
pytest: None
pip: 9.0.1
setuptools: 33.1.1
Cython: None
numpy: 1.13.1
scipy: 0.19.0
xarray: None
IPython: 6.1.0
sphinx: None
patsy: None
dateutil: 2.6.1
pytz: 2017.2
blosc: None
bottleneck: None
tables: 3.4.2
numexpr: 2.6.2
feather: None
matplotlib: 2.0.2
openpyxl: None
xlrd: 1.0.0
xlwt: None
xlsxwriter: 0.9.6
lxml: None
bs4: 4.5.3
html5lib: 0.999999999
sqlalchemy: None
pymysql: None
psycopg2: 2.7.1 (dt dec pq3 ext lo64)
jinja2: 2.9.6
s3fs: None
pandas_gbq: None
pandas_datareader: None
</details>

jreback:
closes #14942
closes #15217
closes #17594 
closes #8869

xref #8138

### DIFF
--- a/doc/source/groupby.rst
+++ b/doc/source/groupby.rst
@@ -991,8 +991,8 @@ is only interesting over one column (here ``colname``), it may be filtered
 
 .. _groupby.observed:
 
-observed hanlding
-~~~~~~~~~~~~~~~~~
+Handling of (un)observed Categorical values
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 When using a ``Categorical`` grouper (as a single or as part of multipler groupers), the ``observed`` keyword
 controls whether to return a cartesian product of all possible groupers values (``observed=False``) or only those
--- a/doc/source/groupby.rst
+++ b/doc/source/groupby.rst
@@ -996,19 +996,19 @@ Handling of (un)observed Categorical values
 
 When using a ``Categorical`` grouper (as a single or as part of multipler groupers), the ``observed`` keyword
 controls whether to return a cartesian product of all possible groupers values (``observed=False``) or only those
-that are observed groupers (``observed=True``). The ``observed`` keyword will default to ``True`` in the future.
-
-Show only the observed values:
-
-.. ipython:: python
-
-   pd.Series([1, 1, 1]).groupby(pd.Categorical(['a', 'a', 'a'], categories=['a', 'b']), observed=True).count()
+that are observed groupers (``observed=True``).
 
 Show all values:
 
 .. ipython:: python
 
    pd.Series([1, 1, 1]).groupby(pd.Categorical(['a', 'a', 'a'], categories=['a', 'b']), observed=False).count()
+
+Show only the observed values:
+
+.. ipython:: python
+
+   pd.Series([1, 1, 1]).groupby(pd.Categorical(['a', 'a', 'a'], categories=['a', 'b']), observed=True).count()
 
 The returned dtype of the grouped will *always* include *all* of the catergories that were grouped.
 
--- a/doc/source/groupby.rst
+++ b/doc/source/groupby.rst
@@ -91,10 +91,10 @@ The mapping can be specified many different ways:
   - A Python function, to be called on each of the axis labels.
   - A list or NumPy array of the same length as the selected axis.
   - A dict or ``Series``, providing a ``label -> group name`` mapping.
-  - For ``DataFrame`` objects, a string indicating a column to be used to group. 
+  - For ``DataFrame`` objects, a string indicating a column to be used to group.
     Of course ``df.groupby('A')`` is just syntactic sugar for
     ``df.groupby(df['A'])``, but it makes life simpler.
-  - For ``DataFrame`` objects, a string indicating an index level to be used to 
+  - For ``DataFrame`` objects, a string indicating an index level to be used to
     group.
   - A list of any of the above things.
 
@@ -120,7 +120,7 @@ consider the following ``DataFrame``:
                       'D' : np.random.randn(8)})
    df
 
-On a DataFrame, we obtain a GroupBy object by calling :meth:`~DataFrame.groupby`. 
+On a DataFrame, we obtain a GroupBy object by calling :meth:`~DataFrame.groupby`.
 We could naturally group by either the ``A`` or ``B`` columns, or both:
 
 .. ipython:: python
@@ -360,8 +360,8 @@ Index level names may be specified as keys directly to ``groupby``.
 DataFrame column selection in GroupBy
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Once you have created the GroupBy object from a DataFrame, you might want to do 
-something different for each of the columns. Thus, using ``[]`` similar to 
+Once you have created the GroupBy object from a DataFrame, you might want to do
+something different for each of the columns. Thus, using ``[]`` similar to
 getting a column from a DataFrame, you can do:
 
 .. ipython:: python
@@ -421,7 +421,7 @@ statement if you wish: ``for (k1, k2), group in grouped:``.
 Selecting a group
 -----------------
 
-A single group can be selected using 
+A single group can be selected using
 :meth:`~pandas.core.groupby.DataFrameGroupBy.get_group`:
 
 .. ipython:: python
@@ -444,8 +444,8 @@ perform a computation on the grouped data. These operations are similar to the
 :ref:`aggregating API <basics.aggregate>`, :ref:`window functions API <stats.aggregate>`,
 and :ref:`resample API <timeseries.aggregate>`.
 
-An obvious one is aggregation via the 
-:meth:`~pandas.core.groupby.DataFrameGroupBy.aggregate` or equivalently 
+An obvious one is aggregation via the
+:meth:`~pandas.core.groupby.DataFrameGroupBy.aggregate` or equivalently
 :meth:`~pandas.core.groupby.DataFrameGroupBy.agg` method:
 
 .. ipython:: python
@@ -517,12 +517,12 @@ Some common aggregating functions are tabulated below:
 	:meth:`~pd.core.groupby.DataFrameGroupBy.nth`;Take nth value, or a subset if n is a list
 	:meth:`~pd.core.groupby.DataFrameGroupBy.min`;Compute min of group values
 	:meth:`~pd.core.groupby.DataFrameGroupBy.max`;Compute max of group values
-	
 
-The aggregating functions above will exclude NA values. Any function which 
+
+The aggregating functions above will exclude NA values. Any function which
 reduces a :class:`Series` to a scalar value is an aggregation function and will work,
 a trivial example is ``df.groupby('A').agg(lambda ser: 1)``. Note that
-:meth:`~pd.core.groupby.DataFrameGroupBy.nth` can act as a reducer *or* a 
+:meth:`~pd.core.groupby.DataFrameGroupBy.nth` can act as a reducer *or* a
 filter, see :ref:`here <groupby.nth>`.
 
 .. _groupby.aggregate.multifunc:
@@ -732,7 +732,7 @@ and that the transformed data contains no NAs.
 .. note::
 
    Some functions will automatically transform the input when applied to a
-   GroupBy object, but returning an object of the same shape as the original. 
+   GroupBy object, but returning an object of the same shape as the original.
    Passing ``as_index=False`` will not affect these transformation methods.
 
    For example: ``fillna, ffill, bfill, shift.``.
@@ -926,7 +926,7 @@ The dimension of the returned result can also change:
 
     In [11]: grouped.apply(f)
 
-``apply`` on a Series can operate on a returned value from the applied function, 
+``apply`` on a Series can operate on a returned value from the applied function,
 that is itself a series, and possibly upcast the result to a DataFrame:
 
 .. ipython:: python
@@ -984,20 +984,48 @@ will be (silently) dropped. Thus, this does not pose any problems:
 
    df.groupby('A').std()
 
-Note that ``df.groupby('A').colname.std().`` is more efficient than 
+Note that ``df.groupby('A').colname.std().`` is more efficient than
 ``df.groupby('A').std().colname``, so if the result of an aggregation function
-is only interesting over one column (here ``colname``), it may be filtered 
+is only interesting over one column (here ``colname``), it may be filtered
 *before* applying the aggregation function.
+
+.. _groupby.observed:
+
+observed hanlding
+~~~~~~~~~~~~~~~~~
+
+When using a ``Categorical`` grouper (as a single or as part of multipler groupers), the ``observed`` keyword
+controls whether to return a cartesian product of all possible groupers values (``observed=False``) or only those
+that are observed groupers (``observed=True``). The ``observed`` keyword will default to ``True`` in the future.
+
+Show only the observed values:
+
+.. ipython:: python
+
+   pd.Series([1, 1, 1]).groupby(pd.Categorical(['a', 'a', 'a'], categories=['a', 'b']), observed=True).count()
+
+Show all values:
+
+.. ipython:: python
+
+   pd.Series([1, 1, 1]).groupby(pd.Categorical(['a', 'a', 'a'], categories=['a', 'b']), observed=False).count()
+
+The returned dtype of the grouped will *always* include *all* of the catergories that were grouped.
+
+.. ipython:: python
+
+   s = pd.Series([1, 1, 1]).groupby(pd.Categorical(['a', 'a', 'a'], categories=['a', 'b']), observed=False).count()
+   s.index.dtype
 
 .. _groupby.missing:
 
 NA and NaT group handling
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-If there are any NaN or NaT values in the grouping key, these will be 
-automatically excluded. In other words, there will never be an "NA group" or 
-"NaT group". This was not the case in older versions of pandas, but users were 
-generally discarding the NA group anyway (and supporting it was an 
+If there are any NaN or NaT values in the grouping key, these will be
+automatically excluded. In other words, there will never be an "NA group" or
+"NaT group". This was not the case in older versions of pandas, but users were
+generally discarding the NA group anyway (and supporting it was an
 implementation headache).
 
 Grouping with ordered factors
@@ -1084,8 +1112,8 @@ This shows the first or last n rows from each group.
 Taking the nth row of each group
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To select from a DataFrame or Series the nth item, use 
-:meth:`~pd.core.groupby.DataFrameGroupBy.nth`. This is a reduction method, and 
+To select from a DataFrame or Series the nth item, use
+:meth:`~pd.core.groupby.DataFrameGroupBy.nth`. This is a reduction method, and
 will return a single row (or no row) per group if you pass an int for n:
 
 .. ipython:: python
@@ -1153,7 +1181,7 @@ Enumerate groups
 .. versionadded:: 0.20.2
 
 To see the ordering of the groups (as opposed to the order of rows
-within a group given by ``cumcount``) you can use 
+within a group given by ``cumcount``) you can use
 :meth:`~pandas.core.groupby.DataFrameGroupBy.ngroup`.
 
 
@@ -1273,7 +1301,7 @@ Regroup columns of a DataFrame according to their sum, and sum the aggregated on
 Multi-column factorization
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-By using :meth:`~pandas.core.groupby.DataFrameGroupBy.ngroup`, we can extract 
+By using :meth:`~pandas.core.groupby.DataFrameGroupBy.ngroup`, we can extract
 information about the groups in a way similar to :func:`factorize` (as described
 further in the :ref:`reshaping API <reshaping.factorize>`) but which applies
 naturally to multiple columns of mixed type and different
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -527,6 +527,41 @@ If you wish to retain the old behavior while using Python >= 3.6, you can use
                'Taxes': -200,
                'Net result': 300}).sort_index()
 
+.. _whatsnew_0230.api_breaking.categorical_grouping:
+
+Categorical Groupers will now require passing the observed keyword
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In previous versions, grouping by 1 or more categorical columns would result in an index that was the cartesian product of all of the categories for
+each grouper, not just the observed values.``.groupby()`` has gained the ``observed`` keyword to toggle this behavior. The default remains backward
+compatible (generate a cartesian product). Pandas will show a ``FutureWarning`` if the ``observed`` keyword is not passed; the default will
+change to ``observed=True`` in the future. (:issue:`14942`, :issue:`8138`, :issue:`15217`, :issue:`17594`, :issue:`8669`, :issue:`20583`)
+
+
+.. ipython:: python
+
+   cat1 = pd.Categorical(["a", "a", "b", "b"],
+                         categories=["a", "b", "z"], ordered=True)
+   cat2 = pd.Categorical(["c", "d", "c", "d"],
+                         categories=["c", "d", "y"], ordered=True)
+   df = pd.DataFrame({"A": cat1, "B": cat2, "values": [1, 2, 3, 4]})
+   df['C'] = ['foo', 'bar'] * 2
+   df
+
+Previous Behavior (show all values):
+
+.. ipython:: python
+
+.. code-block:: python
+   df.groupby(['A', 'B', 'C'], observed=False).count()
+
+
+New Behavior (show only observed values):
+
+.. ipython:: python
+
+   df.groupby(['A', 'B', 'C'], observed=True).count()
+
 .. _whatsnew_0230.api_breaking.deprecate_panel:
 
 Deprecate Panel
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -548,19 +548,46 @@ change to ``observed=True`` in the future. (:issue:`14942`, :issue:`8138`, :issu
    df['C'] = ['foo', 'bar'] * 2
    df
 
-Previous Behavior (show all values):
+``observed`` must now be passed when grouping by categoricals, or a
+``FutureWarning`` will show:
+
+.. ipython:: python
+   :okwarning:
+
+   df.groupby(['A', 'B', 'C']).count()
+
+
+To suppress the warning, with previous Behavior (show all values):
 
 .. ipython:: python
 
-.. code-block:: python
    df.groupby(['A', 'B', 'C'], observed=False).count()
 
 
-New Behavior (show only observed values):
+Future Behavior (show only observed values):
 
 .. ipython:: python
 
    df.groupby(['A', 'B', 'C'], observed=True).count()
+
+For pivotting operations, this behavior is *already* controlled by the ``dropna`` keyword:
+
+.. ipython:: python
+
+   cat1 = pd.Categorical(["a", "a", "b", "b"],
+                         categories=["a", "b", "z"], ordered=True)
+   cat2 = pd.Categorical(["c", "d", "c", "d"],
+                         categories=["c", "d", "y"], ordered=True)
+   df = DataFrame({"A": cat1, "B": cat2, "values": [1, 2, 3, 4]})
+   df
+
+.. ipython:: python
+
+   pd.pivot_table(df, values='values', index=['A', 'B'],
+                  dropna=True)
+   pd.pivot_table(df, values='values', index=['A', 'B'],
+                  dropna=False)
+
 
 .. _whatsnew_0230.api_breaking.deprecate_panel:
 
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -396,6 +396,58 @@ documentation. If you build an extension array, publicize it on our
 
 .. _cyberpandas: https://cyberpandas.readthedocs.io/en/latest/
 
+.. _whatsnew_0230.enhancements.categorical_grouping:
+
+Categorical Groupers has gained an observed keyword
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In previous versions, grouping by 1 or more categorical columns would result in an index that was the cartesian product of all of the categories for
+each grouper, not just the observed values.``.groupby()`` has gained the ``observed`` keyword to toggle this behavior. The default remains backward
+compatible (generate a cartesian product). (:issue:`14942`, :issue:`8138`, :issue:`15217`, :issue:`17594`, :issue:`8669`, :issue:`20583`)
+
+
+.. ipython:: python
+
+   cat1 = pd.Categorical(["a", "a", "b", "b"],
+                         categories=["a", "b", "z"], ordered=True)
+   cat2 = pd.Categorical(["c", "d", "c", "d"],
+                         categories=["c", "d", "y"], ordered=True)
+   df = pd.DataFrame({"A": cat1, "B": cat2, "values": [1, 2, 3, 4]})
+   df['C'] = ['foo', 'bar'] * 2
+   df
+
+To show all values, the previous behavior:
+
+.. ipython:: python
+
+   df.groupby(['A', 'B', 'C'], observed=False).count()
+
+
+To show only observed values:
+
+.. ipython:: python
+
+   df.groupby(['A', 'B', 'C'], observed=True).count()
+
+For pivotting operations, this behavior is *already* controlled by the ``dropna`` keyword:
+
+.. ipython:: python
+
+   cat1 = pd.Categorical(["a", "a", "b", "b"],
+                         categories=["a", "b", "z"], ordered=True)
+   cat2 = pd.Categorical(["c", "d", "c", "d"],
+                         categories=["c", "d", "y"], ordered=True)
+   df = DataFrame({"A": cat1, "B": cat2, "values": [1, 2, 3, 4]})
+   df
+
+.. ipython:: python
+
+   pd.pivot_table(df, values='values', index=['A', 'B'],
+                  dropna=True)
+   pd.pivot_table(df, values='values', index=['A', 'B'],
+                  dropna=False)
+
+
 .. _whatsnew_0230.enhancements.other:
 
 Other Enhancements
@@ -526,68 +578,6 @@ If you wish to retain the old behavior while using Python >= 3.6, you can use
                'Expenses': -1500,
                'Taxes': -200,
                'Net result': 300}).sort_index()
-
-.. _whatsnew_0230.api_breaking.categorical_grouping:
-
-Categorical Groupers will now require passing the observed keyword
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-In previous versions, grouping by 1 or more categorical columns would result in an index that was the cartesian product of all of the categories for
-each grouper, not just the observed values.``.groupby()`` has gained the ``observed`` keyword to toggle this behavior. The default remains backward
-compatible (generate a cartesian product). Pandas will show a ``FutureWarning`` if the ``observed`` keyword is not passed; the default will
-change to ``observed=True`` in the future. (:issue:`14942`, :issue:`8138`, :issue:`15217`, :issue:`17594`, :issue:`8669`, :issue:`20583`)
-
-
-.. ipython:: python
-
-   cat1 = pd.Categorical(["a", "a", "b", "b"],
-                         categories=["a", "b", "z"], ordered=True)
-   cat2 = pd.Categorical(["c", "d", "c", "d"],
-                         categories=["c", "d", "y"], ordered=True)
-   df = pd.DataFrame({"A": cat1, "B": cat2, "values": [1, 2, 3, 4]})
-   df['C'] = ['foo', 'bar'] * 2
-   df
-
-``observed`` must now be passed when grouping by categoricals, or a
-``FutureWarning`` will show:
-
-.. ipython:: python
-   :okwarning:
-
-   df.groupby(['A', 'B', 'C']).count()
-
-
-To suppress the warning, with previous Behavior (show all values):
-
-.. ipython:: python
-
-   df.groupby(['A', 'B', 'C'], observed=False).count()
-
-
-Future Behavior (show only observed values):
-
-.. ipython:: python
-
-   df.groupby(['A', 'B', 'C'], observed=True).count()
-
-For pivotting operations, this behavior is *already* controlled by the ``dropna`` keyword:
-
-.. ipython:: python
-
-   cat1 = pd.Categorical(["a", "a", "b", "b"],
-                         categories=["a", "b", "z"], ordered=True)
-   cat2 = pd.Categorical(["c", "d", "c", "d"],
-                         categories=["c", "d", "y"], ordered=True)
-   df = DataFrame({"A": cat1, "B": cat2, "values": [1, 2, 3, 4]})
-   df
-
-.. ipython:: python
-
-   pd.pivot_table(df, values='values', index=['A', 'B'],
-                  dropna=True)
-   pd.pivot_table(df, values='values', index=['A', 'B'],
-                  dropna=False)
-
 
 .. _whatsnew_0230.api_breaking.deprecate_panel:
 
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -649,6 +649,11 @@ class Categorical(ExtensionArray, PandasObject):
 
     def _codes_for_groupby(self, sort, observed):
         """
+        Code the categories to ensure we can groupby for categoricals.
+
+        If observed=True, we return a new Categorical with the observed
+        categories only.
+
         If sort=False, return a copy of self, coded with categories as
         returned by .unique(), followed by any categories not appearing in
         the data. If sort=True, return self.
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -683,11 +683,15 @@ class Categorical(ExtensionArray, PandasObject):
                 take_codes = np.sort(take_codes)
 
             # we recode according to the uniques
-            cat._categories = self.categories.take(take_codes)
-            cat._codes = _recode_for_categories(self.codes,
-                                                self.categories,
-                                                cat._categories)
-            return cat
+            categories = self.categories.take(take_codes)
+            codes = _recode_for_categories(self.codes,
+                                           self.categories,
+                                           categories)
+
+            # return a new categorical that maps our new codes
+            # and categories
+            dtype = CategoricalDtype(categories, ordered=self.ordered)
+            return type(self)(codes, dtype=dtype, fastpath=True)
 
         # Already sorted according to self.categories; all is fine
         if sort:
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -647,7 +647,7 @@ class Categorical(ExtensionArray, PandasObject):
 
         self._dtype = new_dtype
 
-    def _codes_for_groupby(self, sort):
+    def _codes_for_groupby(self, sort, observed):
         """
         If sort=False, return a copy of self, coded with categories as
         returned by .unique(), followed by any categories not appearing in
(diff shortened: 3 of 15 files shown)


jreback:
this makes categorical groupers work like other groupers and should make things more performant and intuitive. It is somewhat walking back a change from when we first introduced categorical groupers. But the information is preserved (meaning the categories are still there), just removes the automatic re-indexing (which was causing memory to blow up).

still need some increased test coverage. note the first commit actually has almost all of the changes, the next are just cleaning up tests.

TomAugspurger:
Just making sure, this affects grouping by a single categorical as well?

jreback:
yes this affects just a single categorical column also

jreback:
so this is all ready to go if any comments. @TomAugspurger @jorisvandenbossche

jorisvandenbossche:
> yes this affects just a single categorical column also

That doesn't seem to be the case?

With this branch, and the example of the whatsnew docs:

```
In [14]: df.groupby(['A', 'B']).sum()
Out[14]: 
     values
A B        
a c       1
  d       2
b c       3
  d       4

In [15]: df.groupby('A').sum()
Out[15]: 
   values
A        
a       3
b       7
z       0               <---------- now unobserved category is included
```

Further a general comment (will try to do more detailed review later this week): I am not sure we can just change this. First, it is a API breaking change in several places, eg also pivot_table\*. And second, I think the current behaviour can actually useful in certain cases and it would be nice to have a way to keep this behaviour. 
I know this is very ugly, but it would be worth to have a keyword for this in `groupby` ?

\* but I agree we should look into it and try to make this more consistent. As, for example, `pivot` does not seem to include unobserved categories (already currently on master), while `pivot_table` does include them but apparently only for the index and not columns? (try `df.pivot_table('values', 'A', 'B')` with the example of whatsnew). On the other hand, `value_counts` does include them (and I think rightly so), but that also introduces an inconsistency with groupby.

jreback:
So I could also change this for a single grouper. That breaks a couple of tests. I am inclined to do this actually as then it makes the multi and single case consistent. 

We cannot support a full cartesian product for multi-groupers as this will blow up memory and kill performance. So either we go for:

-  an option to turn this on/off. 
- differing behavior for single vs multi
- change single grouping behavior as well

TomAugspurger:
I’d favor an option. I think the default should eventually be to exclude unobserved categories (after a deprecation cycle)

pandas/tests/groupby/test_categorical.py
```diff
@@ -183,536 +45,813 @@ def get_stats(group):
...
+    groups = df.groupby(['range', 'baz'], as_index=False, observed=observed)
+    result = groups.agg('mean')
+
+    groups2 = df.groupby(['range', 'baz'], as_index=True, observed=observed)
+    expected = groups2.agg('mean').reset_index()
+    tm.assert_frame_equal(result, expected)
+
+
+@pytest.mark.xfail(reason="failing with observed")
+def test_observed_failing(observed):
```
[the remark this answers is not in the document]
WillAyd: Haven't been able to follow all the way through yet but I think this is a regression with the below method:

https://github.com/pandas-dev/pandas/blob/41db52730b47ff1d707004decd1214fb87e5c533/pandas/core/sorting.py#L127

Using the failing test example this returns `[array(1, 1, 0, 0], ...` but on master that same object would return `[array(1, 1, 2, 2], ...`. As a result, I think the reconstructed group labels are getting swapped in `_wrap_agged_blocks` and causing the failure.

Will try to find time next day or so to walk through in more detail but sharing in case it helps anyone else reviewing the issue

jreback:
this is read for a look

@TomAugspurger @jorisvandenbossche

pandas/core/reshape/pivot.py
```diff
@@ -79,7 +79,7 @@ def pivot_table(data, values=None, index=None, columns=None, aggfunc='mean',
                 pass
         values = list(values)
 
-    grouped = data.groupby(keys)
+    grouped = data.groupby(keys, observed=dropna)
```
TomAugspurger: Need to think about this a bit. Are we all OK with "overloading" `dropna` to serve two purposes? I think it's ok...

pandas/core/reshape/pivot.py
```diff
@@ -241,10 +241,13 @@ def _all_key(key):
             return (key, margins_name) + ('',) * (len(cols) - 1)
 
         if len(rows) > 0:
-            margin = data[rows + values].groupby(rows).agg(aggfunc)
+            margin = data[rows + values].groupby(
```
TomAugspurger: `observed=True` for these changes are all backwards compatible?

pandas/tests/groupby/aggregate/test_other.py
```diff
@@ -488,12 +488,12 @@ def test_agg_structs_series(structure, expected):
 
 
 @pytest.mark.xfail(reason="GH-18869: agg func not called on empty groups.")
-def test_agg_category_nansum():
+def test_agg_category_nansum(observed):
```
TomAugspurger: Does this need to be xfailed for `observed=True`? I think it may be the right answer (not sure though).

pandas/tests/groupby/conftest.py
```diff
@@ -4,6 +4,11 @@
 from pandas.util import testing as tm
 
 
+@pytest.fixture(params=[True, False])
+def observed(request):
```
TomAugspurger: Docstring would be nice.

pandas/tests/reshape/test_pivot.py
```diff
@@ -93,23 +93,23 @@ def test_pivot_table_dropna(self):
...
         result = pd.pivot_table(df, values='values', index=['A', 'B'])
 
-        exp_index = pd.MultiIndex.from_product(
-            [Categorical(["a", "b", "z"], ordered=True),
-             Categorical(["c", "d", "y"], ordered=True)],
+        exp_index = pd.MultiIndex.from_arrays(
+            [cat1, cat2],
             names=['A', 'B'])
         expected = DataFrame(
-            {'values': [1, 2, np.nan, 3, 4, np.nan, np.nan, np.nan, np.nan]},
```
TomAugspurger: Ah, so those changes in pivot are API breaking? I'd prefer that this goes through the same deprecation cycle. On the other hand, this would mean an additional keyword, so that we can do the deprecation cycle properly...

Do we think `pivot(..., dropna=True, observed=False)` (i.e. the current default) is a useful combination? I could see it being desired.

pandas/tests/groupby/test_categorical.py
```diff
@@ -525,13 +666,34 @@ def test_groupby_categorical_no_compress():
...
+    result = data.groupby("b", observed=False).mean()
     result = result["a"].values
     exp = np.array([1, 2, 4, np.nan])
     tm.assert_numpy_array_equal(result, exp)
 
 
-def test_groupby_sort_categorical():
+def test_sort():
+
+    # http://stackoverflow.com/questions/23814368/sorting-pandas-categorical-labels-after-groupby
```
TomAugspurger: Is this line too long? https://stackoverflow.com/q/23814368/1889400 is a short link to the same q.

TomAugspurger (commented):
Looks good. Main thing is the change to `pivot_table`.

It'd be good to document whether the unobserved categories are present in the resulting index's *type*.

i.e. is

```python
In [7]: pd.Series([1, 1, 1]).groupby(pd.Categorical(['a', 'a', 'a'], categories=['a', 'b'])).count().index.dtype
```

```
Out[7]: CategoricalDtype(categories=['a', 'b'], ordered=False)
```

or

```python
Out[7]: CategoricalDtype(categories=['a'], ordered=False)
```

This could go in the whatsnew and either groupby.rst or categorical.rst probably. I think either behavior is fine.

Still going through the test changes.

pandas/tests/groupby/aggregate/test_other.py
jreback: fixed so that observed=True is XPASS (not sure how to xfail on a particular fixture value)

pandas/core/reshape/pivot.py
jreback: yes, but that's the meaning of the ``dropna`` now here anyhow

jorisvandenbossche:
I will try to give this a more detailed look tomorrow.

But general comment: now that we have the keyword to specify the behaviour (for now with `observed=False` for back compat), I am not really sure that we should change the default in the future.

It's difficult to judge, as I personally don't run much in such situations, but my gut feeling says that the original issue that motivated the change (the combinatorial explosion with multiple categorical groupers) is not the majority usage pattern of categoricals. And if you don't want to include unobserved categoricals in your analyses in general (groupby, pivot, value_counts, plotting, ..), you always have the easy functionality of `remove_unused_categories`.

cc @jankatins

jreback:
@TomAugspurger the pivot was a red herring, was not passing things thru. and ``dropna`` (in pivot) is exactly equivalent of the ``observed`` kwarg (to groupby). So maybe should just rename ``observed`` -> ``dropna``, and would be consistent across other functions (``value_counts``) as well.

@jorisvandenbossche I disagree. You almost *never* want a cartesian product of all of the groupers. It can easily blow you up and shouldn't be the default. (its also easy to create if you need). Not this has *nothing* to do with the actual categories that are returned, they are in BOTH cases indicated on the dtype of the level of that index, (observed or not), its the groupers that are the issue.

jorisvandenbossche:
> You almost *never* want a cartesian product of all of the groupers. It can easily blow you up and shouldn't be the default. 

But in most cases you have *no* cartesian product, you only have a single categorical key

> It can easily blow you up and shouldn't be the default. (its also easy to create if you need).

It's also easy to get the version with only the observed ones (certainly now there is the keyword), so that is not really an argument IMO

> Not this has nothing to do with the actual categories that are returned, they are in BOTH cases indicated on the dtype of the level of that index

Yes, and that is certainly a good thing. But for me it is still the question what we want the *visible* output to be. 
For example a pivot table is often used as a kind of a summary table. In such a case, I often do want to know that a certain category I care about (otherwise it would not be in the categories) has no values. 
And it is also about consistency: `value_counts` does include the unobserved one (and rightly so, IMO).

jankatins:
My original motivation to work on categoricals were stuff like lickert scales ("completly agree ... completly disagree", 5 to 7 values). For that seing unobserved categoricals in group bys are a good thing.

Since then a lot of changes have been made to make categoricals usefull in other situations (like as a memory efficient string replacement or to analyse genes?).

This change feels like it is a change to make working with the latter easier but will make the former harder.

toobaz:
> And it is also about consistency: value_counts does include the unobserved one (and rightly so, IMO).

I gave it as granted that if we change ``groupby`` we will also change ``value_counts`` - otherwise I agree it doesn't make any sense.

> This change feels like it is a change to make working with the latter easier but will make the former harder.

True... I just think your use case is more rare (but I have no data to support this statement, other than personal experience). Anyway, with the new argument both things will be pretty easy anyway, it is mainly a matter of which use case should require awareness from the user.

jreback:
agree for consistency should change:

```
In [1]:  pd.Series(list('aabc')).astype(pd.api.types.CategoricalDtype(list('abcd'))).value_counts(dropna=True)
Out[1]: 
a    2
c    1
b    1
d    0
dtype: int64

In [2]:  pd.Series(list('aabc')).astype(pd.api.types.CategoricalDtype(list('abcd'))).value_counts(dropna=False)
Out[2]: 
a    2
c    1
b    1
d    0
dtype: int64
```

so that ``True`` would remove the unobserved.

TomAugspurger:
I think in the long run I'd prefer unobserved categories to be excluded from groupby, primarily because the cartesian product issue seems like such a difficult gotcha. Though if / when we have an "interned string" array type (categorical, but with different semantics), would we want to go back on that change?

---

`value_counts` highlights my discomfort with overloading `dropna`. It will require an `observed` keyword to handle cases like

```python
In [21]: c1 = pd.Categorical(['a', 'a', None], categories=['a'])

In [22]: c2 = pd.Categorical(['a', 'a', None], categories=['a', 'b'])
```

I don't think `c1.value_counts(dropna=True)`  should be the same as `c2.value_counts(dropna=True)`. `c2` has the additional category. So either recommend `c2.remove_unused_categories.value_counts(dropna=True)` or `c2.value_counts(dropna=True, observed=True)`.

toobaz:
Might be a crazy idea... but just in case it's not, I'll rather propose it now than later: what if showing empty categories (or not) was a property of the _dtype_, preserved across transformations, just like the (possible) ordering?

That way, @jankatins would need to specify only once, when creating the categorical for the first time, that his categories (the Likert scale values) are "all important". Otherwise, the categories would be considered as merely "functional".

We could even think that a categorical created by passing the categories has "important" categories by default, while a categorical which creates its own categories automatically has "functional" categories by default (clearly with a parameter allowing to change this). (But this might be too much magic.)

TomAugspurger:
In a pandas with an interned string array, would we want the `observed` keyword? I think so, but the default would stay `False`, as it is now.

pandas/core/groupby/groupby.py
```diff
@@ -2953,17 +2962,35 @@ def __init__(self, index, grouper=None, obj=None, name=None, level=None,
             # a passed Categorical
             elif is_categorical_dtype(self.grouper):
 
-                self.grouper = self.grouper._codes_for_groupby(self.sort)
+                # Use the observed values of the grouper if inidcated
+                observed = self.observed
+                if observed is None:
+                    msg = ("pass observed=True to ensure that a "
+                           "categorical grouper only returns the "
+                           "observed groupers, or\n"
```
jorisvandenbossche: "groupers" -> "categories" or "values"

pandas/core/arrays/categorical.py
```diff
@@ -659,6 +661,22 @@ def _codes_for_groupby(self, sort):
...
+        if observed:
+            unique_codes = unique1d(self.codes)
+            cat = self.copy()
+
+            take_codes = unique_codes[unique_codes != -1]
+            if self.ordered:
+                take_codes = np.sort(take_codes)
+
+            # we recode according to the uniques
+            cat._categories = self.categories.take(take_codes)
```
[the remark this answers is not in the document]
jorisvandenbossche: To explain my confusion: what this returns is "strange", as it does returns a modified categorical that does not match the data any longer. Eg for

```
cat = pd.Categorical(['a', 'c', 'a'], categories=['a', 'b', 'c'])
```

this returns a categorical with values ['a', 'b', 'a'], because you recode the codes (based on observed categories) without changing the categories. 

This makes that the `Grouping._group_index` is "incorrect", but apparently this is fine for the actual groupby implementation because the groupby results are correct (but, it makes the above function somewhat confusing). But with playing around with it, I found a place where it is actually propagated:

```
In [52]: cat = pd.Categorical(['a', 'c', 'a'], categories=['a', 'b', 'c'])

In [53]: df = pd.DataFrame({'cat': cat, 'vals': [1, 2, 3]})

In [54]: g = df.groupby('cat', observed=True)

In [55]: g.sum()   <--- correct
Out[55]: 
     vals
cat      
a       4
c       2

In [56]: g.groups    <--- incorrect
Out[56]: 
{'a': Int64Index([0, 2], dtype='int64'),
 'b': Int64Index([1], dtype='int64'),
 'c': Int64Index([], dtype='int64')}
```

(same for `g.indices`)

jorisvandenbossche (changes requested):
General question (in case we keep the deprecation warning): do we want to show it always? Or only in the case where you actually *have* unobserved values?

Didn't look at the tests yet

toobaz:
> I would rather make a "dict-encoded / interned string" type than overloading Categorical with this

While I have no objection to relying on a different (as in "differently named") type, which in turn might benefit from convenience functions to categorize strings (thanks for the link @jorisvandenbossche ), it seems to me that 1) there is no need to make that different dtype specifically string-ish 2) the implementation would/could be 99% identical to ``Categorical`` 3) it would be useful to be able to switch from one type to the other without re-encoding, and 4) anyway, the fact that we don't allow categories to change is a limitation of implementation, not a virtue, hopefully to be lifted some day regardless of this issue (am I wrong?)

In particular, if 3) was true (which would most certailny be the case if 2) was true), then the use of the ``observed`` keyword would be quite redundant ( @TomAugspurger ). (Which is not to say it is necessarily useless - it is mostly a matter of which is the most convenient API: what do you think @jankatins ?)

jorisvandenbossche:
> 4) anyway, the fact that we don't allow categories to change is a limitation of implementation, not a virtue, hopefully to be lifted some day regardless of this issue (am I wrong?)

As far as I know, this is by design (to preserve the integrity of the categorical dtype). We have eg a `union_categoricals` helper function if you want that when merging data: https://pandas.pydata.org/pandas-docs/stable/categorical.html#unioning (I seem to recall we had quite some discussion if we should use that by default in `concat` or not)

> In a pandas with an interned string array, would we want the observed keyword? I think so, but the default would stay False, as it is now.

That's a bit what I meant above with that I think we should design the Categorical default behaviour for the "categorical" use case, and not for the "memory efficient strings" usecase for which it is also used in absence of a better string dtype.

jreback:
> General question (in case we keep the deprecation warning): do we want to show it always? Or only in the case where you actually have unobserved values?

we *only* show it if you have a categorical in the groupers. I think this is a fine case to surface to the user to make them be explicit. Otherwise you end up showing it depending on the data, generally not a great idea.

jreback:
updated, but didn't address @jorisvandenbossche comments yet.

TomAugspurger:
@jreback what are your thoughts on the likelihood of an "interned string" extension array (for 0.24)? It'll be able to share much of the implementation with Categorical, but would have different semantics around groupby, value_counts, pivot_table, etc.

Assuming we had that already, would `observed=True` be a good default *for Categorical* (it wouldn't apply to the interrned string EA)?

Since we don't have that extension array today, I think this PR is still great to have for 0.23. I'm just worried about the default being `observed=True`.

jreback:
well I think we *could* do a String dtype, ideally would be based on ``pyarrow`` but don't think that will be ready, we *could* store as a fixed length numpy array with offsets so that might work. I think this will be some work to do it though (was planning on int NA better)

jreback:
> Since we don't have that extension array today, I think this PR is still great to have for 0.23. I'm just worried about the default being observed=True.

I think even IF we have a string type (EA), this should still be ``observed=True``. The reason is that this can blow up a grouper very easily (or work, yet be very very slow).

TomAugspurger:
(agreed about int NA being more useful, but let's keep that separate).

I could see a use-case for an general String EA, and an InternedString EA
(or DictEncodedString). What you're describing sounds like the general
String type.

InternedString would have the same impl as Categorical: a set of codes
indexing into a Index of strings (instead of categories). But things like
`_codes_for_groupby` would be different, since it would exclude unobserved
categories.

TomAugspurger:
> The reason is that this can blow up a grouper very easily (or work, yet be very very slow).

That's fair, it's a definite gotcha if you have a large number of categories and are doing a groupby across multiple categoricals. I wish we had more information on how people actually use this :/ In the linked issues, most people are saying they're using categoricals for memory efficiency, so they'd be well served by the Dictionary-encoded EA (doesn't have to be strings I guess). But I *assume* there are people out there who expect grouping by a Categorical to include everything, and would be confused about why we're changing the default.

jreback:
@TomAugspurger I c. so pretty much a memory-friendly string type. Actually we do that as the default string type and it would just work. 

But even still I would have ``observed=True`` as the default. I could see ``observed=False`` being the default for a single grouper, but for multiple it can quickly blow up (that's the case today). But having these different is odd, so have to pick.

jreback:
so is the proposal to add the observed kw as ``False`` and have NO deprecation ? and then add a DictSting type (or whatever) soonish?

TomAugspurger:
> But having these different is odd, so have to pick.

Agreed, we shouldn't do that.

> so is the proposal to add the observed kw as False and have NO deprecation ? and then add a DictSting type (or whatever) soonish?

That sounds right. @jorisvandenbossche? (to be clear, I'm on the fence. I think I prefer `observed=False`, with no deprecation, but not 100% sure).

jreback:
an alternative is I could calculate the size of the result & emit a PerformanceWarning I think?

TomAugspurger:
Hmm, yes that might be good to do regardless of the default. I imagine
people will continue to use Categorical as a memory-efficient string for a
time going forward. They would want to be informed that they're blowing up
memory.

toobaz:
Considering that the memory-oriented alternative to ``Categorical`` is not there already, that we might end up focusing on the case of strings (which would certainly be the most typical application), that in some cases once you emit a``PerformanceWarning`` it could be already too late, and that the ``observed=False`` se case still seems to me an "advanced" mode (as in: "what people happen to be puzzled by on StackOverflow"),  I would still prefer to aim for ``observed=True``.

(I admit I'm biased by the fact that having ``observed`` as a property of a ``Categorical`` dtype still looks to me like a clean - and non exclusive - solution to make the ``observed=False`` use case more comfortable)

jorisvandenbossche:
> (I admit I'm biased by the fact that having observed as a property of a Categorical dtype still looks to me like a clean - and non exclusive - solution to make the observed=False use case more comfortable)
Merge state

@toobaz but also for having it as metadata on the Categorical, you need to pick a default :-)

---

To be clear, I also don't have a good view on typical usage, so my preference for keeping `observed=False` as the default and doing no deprecation is not that strong. But since that is the current default, I would personally be conservative and keep it on that.

In any case, what this PR adds, even without the deprecation is, is very useful and will at least give people that have the problem of the blowing memory the option to fix it. 
And adding the observed keyword now, does not exclude that we might later deprecate/change the default if we get more feedback.

jreback:
ok revised to make ``observed=False`` the default and removed the deprecation warning.

jankatins:
If you make observed a property of the dtype, then refactoring this into a `BaseCategorical` with both a `Categorical` and `IntEncodedString` is probably mostly only touching the categorical.py code but not anymore the internals (which were mostly about making sure that all cats are reported instead of only observed values).

toobaz:
@jankatins maybe I misunderstand you, but clearly my idea would be to touch the ``groupby`` (and related) code more or less as follows:
```
# Assuming observed=None by default
if observed is None:
    # Assuming Categorical.observed=True by default
    observed = the_categorical.observed
# ... show all categories iff observed=True
```
(this would have the added benefit of being able to specify that one grouper should include empty categories, while another shouldn't... not that I would consider this a killer feature)

@jreback @jorisvandenbossche @TomAugspurger ... which reminds me of the fact that, while my proposal is maybe getting OT here (introducing the parameter is more urgent than selecting the default behavior), maybe a more future-proof approach is to have ``observed=None`` by default? This would allow us to understand whether to prioritize the parameter or the grouper property (regardless of whether the grouper is an enhanced categorical or a "dict-encoded / interned string")

jankatins:
@toobaz The main difference between a Categorical and a IntEncodedString is IMO:

* Grouper/valuecount/etc show observed (string) or also not observed categoricals (Lickert like Categorical) -> taken care of here if observed is a property of the dtype
* Sorting is done by defined categorical order (cat) or value (string)
* You can insert new values / merge with other values / compare with unknown values (string) or not (cat)

toobaz:
> Grouper/valuecount/etc show only observed (string) or also not observed categoricals (Lickert like Categorical) -> taken care of here if observed is a property of the dtype -> you would add a StringDtype which always observed == True
> Sorting is done by defined categorical order (cat) or value (string)

I see these two points as orthogonal. I'm sure I will mostly use ``observed=True``, and this doesn't mean I attribute any meaning to the "natural" (alphabetic) order of categories (which, by the way, won't necessarily be strings).

I am currently working on administrative data in which a variable can take ~30 values (types), but most of them are extremely rare, or anyway uninteresting for the analysis. I mostly work on the 5-10 most frequent types, and in those cases I will always use ``observed=False`` (I currently often drop zeros manually). The types have their own ordering, while the alphabetic one is meaningless.

> You can insert new values / merge with other values / compare with unknown values (string) or not (cat)

I know I'm repeating myself, but I was never able to consider this limitation as a feature, regardless of order, dtype, and ``observed=``. If my data provider introduces a new category, I don't see why it is should be good that my code breaks when appending the new data to the old data. I totally accept it as a limitation (with annoying consequences e.g. for roundtrips, and [compatibility](https://github.com/pandas-dev/pandas/pull/13440#issuecomment-228282334) with other types), just not as a feature.

jreback:
I changed the default back to ``observed=None`` to preserve options. I think we can discuss wether / if to change this in the future. But for now allows folks to deal with grouping in a better way now.

jankatins:
> I don't see why it is should be good that my code breaks when appending the new data to the old data.

If you have a defined order, where would you put the new categorical into the orders? E.g.  `good - middle - bad`, where would you put 'extreme' -> if you care about order, then you also care about not adding new stuff

> I mostly work on the 5-10 most frequent types, and in those cases I will always use observed=False

This usecase might also be satisfied by a `collapse` or `lump` method (e.g. see here: http://r4ds.had.co.nz/factors.html#modifying-factor-levels)

TomAugspurger:
FYI, discussion for a new dict-encoded / interned values type can go at https://github.com/pandas-dev/pandas/issues/20899 so we don't lose it.

Reviewing this PR one more time, but I think people are all on board with having the default be `observed=False` (back compat) for now?

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -396,6 +396,58 @@ documentation. If you build an extension array, publicize it on our
 
 .. _cyberpandas: https://cyberpandas.readthedocs.io/en/latest/
 
+.. _whatsnew_0230.enhancements.categorical_grouping:
+
+Categorical Groupers has gained an observed keyword
```
TomAugspurger: has -> have? Because "categorical Groupers" is plural right?

pandas/core/arrays/categorical.py
```diff
@@ -671,6 +678,26 @@ def _codes_for_groupby(self, sort):
             categories in the original order.
         """
 
+        # we only care about observed values
+        if observed:
+            unique_codes = unique1d(self.codes)
```
TomAugspurger: Haven't thought this through, but can this `if` block be replaced with `self.remove_unused_cateogories()._codes_for_groupby(sort=sort, observed=False)`?

TomAugspurger (approved):

doc/source/groupby.rst
```diff
@@ -984,20 +984,48 @@ will be (silently) dropped. Thus, this does not pose any problems:
...
-is only interesting over one column (here ``colname``), it may be filtered 
+is only interesting over one column (here ``colname``), it may be filtered
 *before* applying the aggregation function.
 
+.. _groupby.observed:
+
+Handling of (un)observed Categorical values
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When using a ``Categorical`` grouper (as a single or as part of multipler groupers), the ``observed`` keyword
```
jorisvandenbossche: we don't use "grouper" as terminology in our documentation (except for the `pd.Grouper` object), so I would write "groupby key" or "to group by"

also "multipler" -> "multiple"

doc/source/groupby.rst
```diff
@@ -984,20 +984,48 @@ will be (silently) dropped. Thus, this does not pose any problems:
...
+controls whether to return a cartesian product of all possible groupers values (``observed=False``) or only those
+that are observed groupers (``observed=True``).
```
jorisvandenbossche: "or only those that are observed groupers" -> "or only the observed categories"

doc/source/groupby.rst
```diff
@@ -984,20 +984,48 @@ will be (silently) dropped. Thus, this does not pose any problems:
...
+
+   pd.Series([1, 1, 1]).groupby(pd.Categorical(['a', 'a', 'a'], categories=['a', 'b']), observed=False).count()
+
+Show only the observed values:
+
+.. ipython:: python
+
+   pd.Series([1, 1, 1]).groupby(pd.Categorical(['a', 'a', 'a'], categories=['a', 'b']), observed=True).count()
+
+The returned dtype of the grouped will *always* include *all* of the catergories that were grouped.
```
jorisvandenbossche: catergories -> categories

doc/source/groupby.rst
```diff
@@ -984,20 +984,48 @@ will be (silently) dropped. Thus, this does not pose any problems:
...
+
+When using a ``Categorical`` grouper (as a single or as part of multipler groupers), the ``observed`` keyword
+controls whether to return a cartesian product of all possible groupers values (``observed=False``) or only those
+that are observed groupers (``observed=True``).
+
+Show all values:
+
+.. ipython:: python
+
+   pd.Series([1, 1, 1]).groupby(pd.Categorical(['a', 'a', 'a'], categories=['a', 'b']), observed=False).count()
```
jorisvandenbossche: I would maybe just create `s` and `cat` to avoid repeating this a few times

pandas/core/generic.py
```diff
@@ -6632,6 +6632,13 @@ def groupby(self, by=None, axis=0, level=None, as_index=True, sort=True,
         squeeze : boolean, default False
             reduce the dimensionality of the return type if possible,
             otherwise return a consistent type
+        observed : boolean, default None
+            if True: only show observed values for categorical groupers.
```
jorisvandenbossche: capital If (below as well)

pandas/core/generic.py
```diff
@@ -6632,6 +6632,13 @@ def groupby(self, by=None, axis=0, level=None, as_index=True, sort=True,
...
+            if False: show all values for categorical groupers.
+            if None: if any categorical groupers, show a FutureWarning,
+                default to False.
```
jorisvandenbossche: no identation for rst formatting

pandas/core/groupby/groupby.py
```diff
@@ -2898,14 +2907,16 @@ class Grouping(object):
     """
 
     def __init__(self, index, grouper=None, obj=None, name=None, level=None,
-                 sort=True, in_axis=False):
+                 sort=True, observed=None, in_axis=False):
```
jorisvandenbossche: Why not have it as `False` default? (if we want to deprecate in the future, we can then just use `None` ?)

pandas/core/generic.py
jorisvandenbossche: Also, can you start this explanation with noting this keyword is only when grouping by categorical values?

pandas/core/groupby/groupby.py
```diff
@@ -4696,6 +4741,14 @@ def _reindex_output(self, result):
 
         This can re-expand the output space
         """
+
+        # TODO(jreback): remove completely
+        # when observed parameter is defaulted to True
+        # gh-20583
```
jorisvandenbossche: should this comment be removed for now?

jorisvandenbossche (commented):
@jreback I would like to ask, again, can you *please* only add new commits when updating for reviews instead of amending different parts to different previous commits? 

I wanted to start complaining that you didn't update for the bug I pointed out in `_codes_for_groupby` because I didn't see it in the new commits, but I see you actually fixed it. 
But this way it is really hard to see that you updated it and added tests for it, and to see what you actually changed in that function compared to the previous time I reviewed.

jorisvandenbossche:
My remaining comments are mainly doc related, so I am fine with merging this now for 0.23rc, if @jreback does a follow-up PR.

jreback:
@jorisvandenbossche 

> I would like to ask, again, can you please only add new commits when updating for reviews instead of amending different parts to different previous commits?

> I wanted to start complaining that you didn't update for the bug I pointed out in _codes_for_groupby because I didn't see it in the new commits, but I see you actually fixed it.
But this way it is really hard to see that you updated it and added tests for it, and to see what you actually changed in that function compared to the previous time I reviewed.

and I did push new ones.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -396,6 +396,58 @@ documentation. If you build an extension array, publicize it on our
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In previous versions, grouping by 1 or more categorical columns would result in an index that was the cartesian product of all of the categories for
```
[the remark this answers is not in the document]
jreback: I didn't change this on purpose, this is more correct.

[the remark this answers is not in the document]
TomAugspurger: "Cartesian product" really only makes sense in the *2 or more* case, right? But you say "1 or more" above. I would phrase it at

"Grouping by a categorical includes the unobserved categories in the output. When grouping by multiple categories, this means you get the cartesian product of all the categories, including combinations where there are no observations, which can result in high memory usage."

jorisvandenbossche:
> and I did push new ones.

Well, if I look at the diff for only the commits you added the last day (https://github.com/pandas-dev/pandas/pull/20583/files/19c9cf7871847de8f0a8504e9f121ad1460512d0..bdf7525812ca670f9406ab8df333030d36d30947), there is no change in the `_codes_for_groupby` function, while you *did* update it according to my comments.

doc/source/whatsnew/v0.23.0.txt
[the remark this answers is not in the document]
jorisvandenbossche: Yep, the explanation of Tom is exactly what I meant.

@jreback I have no problem at all with that you don't agree with a comment (it would be strange otherwise :-)) and thus not update for it, but can you then answer to that comment noting that? Otherwise I cannot know that I should not repeat a comment (or that I shouldn't get annoyed with my comments being ignored :))

TomAugspurger:
@jreback opened https://github.com/pandas-dev/pandas/issues/20902

Will merge in ~1 hour.

jreback:
@TomAugspurger if you want to do the RC now, then i'll have to followup on comments later.

TomAugspurger:
If we're doing this for 0.23 then it should go in the RC I think. I can wait a bit longer if you plan to push more changes.

jreback:
changes will only be cosmetic so u can merge if u want now

TomAugspurger:
That's what I thought, thanks.

dragoljub:
Great work! Looking forward to trying this out. :+1

pandas/core/arrays/categorical.py
jreback: no, you actually need the uniques

abdullah-online:
I think there is a bug with observed=True. The below statement removed the unobserved categorical values, but the dataframe returned wrong counts.
The categorical values were not sorted and the counts were mismatched with the categories.

df.groupby('categorical_column', observed=True, as_index=False).count()

TomAugspurger:
Try with pandas 0.24.0 if you're not already on it. There were a couple bug
fixes in this area.

If it still persists, search for open issues / open a new one with a
minimal example.

PR #20584, opened 2018-04-02
API: rolling.apply will pass Series to function

### LINKED ISSUE #5071: ENH/API: rolling_apply to pass frames to the rolled function (rather than ndarrays)
Returning a Series: http://stackoverflow.com/questions/19121854/using-rolling-apply-on-a-dataframe-object

Returning a Scalar: http://stackoverflow.com/questions/21040766/python-pandas-rolling-apply-two-column-input-into-function/21045831#21045831

jreback:
closes #5071

### DIFF
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -844,7 +844,7 @@ Other API Changes
 - :func:`DatetimeIndex.strftime` and :func:`PeriodIndex.strftime` now return an ``Index`` instead of a numpy array to be consistent with similar accessors (:issue:`20127`)
 - Constructing a Series from a list of length 1 no longer broadcasts this list when a longer index is specified (:issue:`19714`, :issue:`20391`).
 - :func:`DataFrame.to_dict` with ``orient='index'`` no longer casts int columns to float for a DataFrame with only int and float columns (:issue:`18580`)
-- A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousines, will now *always* be passed a ``Series``, rather than an ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
+- A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousins, will now *always* be passed a ``Series``, rather than an ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
 
 .. _whatsnew_0230.deprecations:
 
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -74,7 +74,7 @@ Rolling/Expanding.apply() accepts a ``raw`` keyword to pass a ``Series`` to the 
 :func:`Series.expanding().apply() <pandas.core.window.Expanding.apply>`, and :func:`DataFrame.expanding().apply() <pandas.core.window.Expanding.apply>` have gained a ``raw=None`` parameter.
 This is similar to :func:`DataFame.apply`. This parameter, if ``True`` allows one to send a ``np.ndarray`` to the applied function. If ``False`` a ``Series`` will be passed. The
 default is ``None``, which preserves backward compatibility, so this will default to ``True``, sending an ``np.ndarray``.
-In a future version the default will be changed to ``False``, sending a ``Series``. (:issue:`5071`)
+In a future version the default will be changed to ``False``, sending a ``Series``. (:issue:`5071`, :issue:`20584`)
 
 .. ipython:: python
 
@@ -844,6 +844,7 @@ Other API Changes
 - :func:`DatetimeIndex.strftime` and :func:`PeriodIndex.strftime` now return an ``Index`` instead of a numpy array to be consistent with similar accessors (:issue:`20127`)
 - Constructing a Series from a list of length 1 no longer broadcasts this list when a longer index is specified (:issue:`19714`, :issue:`20391`).
 - :func:`DataFrame.to_dict` with ``orient='index'`` no longer casts int columns to float for a DataFrame with only int and float columns (:issue:`18580`)
+- A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousines, will now *always* be passed a ``Series``, rather than an ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
 
 .. _whatsnew_0230.deprecations:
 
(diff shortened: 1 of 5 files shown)


pandas/tests/test_window.py
```diff
@@ -2805,20 +2823,25 @@ def expanding_func(x, min_periods=1, center=False, axis=0):
...
 
         def expanding_mean(x, min_periods=1):
+
             exp = x.expanding(min_periods=min_periods)
-            return exp.apply(lambda x: x.mean())
+            result = exp.apply(lambda x: x.mean(), raw=raw)
+            return result
 
-        self._check_expanding(expanding_mean, np.mean)
+        # TODO(jreback), needed to add preserve_nan=False
```
jreback: if anyone is interested to see why this doesn't work, would be hepful, @WillAyd (you have some xp in this area)

TomAugspurger:
Thoughts on the default being ndarray with a warning that the default will change to `raw=False` in the future? Then this isn't a breaking change.l

pandas/tests/test_window.py
WillAyd: It looks like rows [20:39] of the series in use here contains NA values. `preserve_nan` is assuming that those will still be NA after the function application, but isn't it by design with expanding functions that the NA values do not get preserved?  Ex:

```python
>>> ser = pd.Series(range(5))
>>> ser[2] = np.nan
>>> ser.expanding().sum()
0    0.0
1    1.0
2    1.0
3    4.0
4    8.0
dtype: float64
```

So I think this keyword is necessary by design of the expanding function

jreback: yeah exactly - so wondering if the current impl (eg master) is actually wrong

jreback:
yeah i could make the keyword default (of None) and then warn that it’s going to change

that’s back compat

pandas/tests/test_window.py
WillAyd: I haven't looked at the implementation in detail but I'm assuming this ties back to the default argument of `min_periods=1` that the Expanding class uses - I think that would make it almost impossible for NA data to appear while expanding. 

I'd buy the argument that the default here should match the Rolling impl (i.e. 0) and Expanding should deal with "expected" NAs (i.e. forward looking values in the window that just haven't been reached yet) vs "unexpected" NA (data from a previous entry in the window) in  a more robust fashion. 

It would be a breaking change; happy to open the ticket if you agree

jreback: yeah this might be a bit of an unexplored corner

yeah if i can show a simple example would be great to open an issue

after all expanding is just rolling with the window len of the len of the obj

WillAyd: Maybe it shouldn't be though? I.e. should the window size of expanding be `i for i in range(N)` instead of `N for i in range(N)`? The difference can be illustrated looking at the third element below (elements 4 and 5 are expected to differ from window size)

```python
>>> ser = pd.Series([0, np.nan, 2, 3, 4])
In [127]: ser.rolling(window=2, min_periods=2).sum()
Out[127]: 
0    NaN
1    NaN
2    NaN  # Doesn't populate here; there aren't 2 non-NA entries at this point
3    5.0
4    7.0
dtype: float64

In [128]: ser.rolling(window=len(ser), min_periods=2).sum()
Out[128]: 
0    NaN
1    NaN
2    2.0  # Populates here, even though this is the first non-NA record and min_periods is 2
3    5.0
4    9.0
dtype: float64
```

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -479,6 +479,33 @@ If you wish to retain the old behavior while using Python >= 3.6, you can use
...
                'Net result': 300}).sort_index()
 
+.. _whatsnew_0230.api_breaking.window_raw:
+
+Rolling/Expanding.apply() will by default send a ``Series`` to the function
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The :func`Series.rolling`, :func:`DataFrame.rolling`, :func`Series.expanding`, :func:`DataFrame.expanding` methods when used with ``.apply()`` have gained a ``raw=False`` parameter.
+This is similar to :func:`DataFame.apply`. This parameter, ``False`` by default allows one to send a ``np.ndarray`` to the applied function, rather than the default of a ``Series``.
+This is a change from prior versions, when the applied function would *always* received an ndarray. This allow one to use pandas operations generically. (:issue:`5071`)
```
WillAyd: Minor typo on this line - either "would always receive" or "always received"

jreback:
this is all fixed up. now back-compat (with a warning). if anyone wants to look 

cc @WillAyd 
@TomAugspurger

WillAyd:
lgtm

jorisvandenbossche (commented):
This is really nice to get it more consistent with DataFrame.apply!

jorisvandenbossche:
While trying this out, I also noticed that rolling apply only seems to work on numeric (int/float) data. Is this a known limitation/bug? (didn't directly find an issue about it):

```python
In [69]: s = pd.Series(list('abcd'))

In [71]: s.rolling(2).apply(lambda x: x+1, raw=True)   
Out[71]: 
0    a         #<-------- for string data, just returns the original series, whathever the function is
1    b
2    c
3    d
dtype: object

In [72]: s = pd.Series(pd.Categorical(list('abcd')))

In [73]: s.rolling(2).apply(lambda x: x+1, raw=True)
Out[73]: 
[a, b, c, d]     #<-------- for categorical data, just returns the original Categorical (not even series)
Categories (4, object): [a, b, c, d]

In [74]: s = pd.Series(pd.timedelta_range("1 days", periods=4))

In [75]: s.rolling(2).apply(lambda x: x+1, raw=True)   #<----- for timedelta/datetime, raises informative error
...
NotImplementedError: ops for Rolling for this dtype timedelta64[ns] are not implemented
```

(this is a separate issue as this PR, it's already like this in released version)

pandas/core/window.py
```diff
@@ -314,7 +314,7 @@ def _center_window(self, result, window):
     def aggregate(self, arg, *args, **kwargs):
         result, how = self._aggregate(arg, *args, **kwargs)
         if result is None:
-            return self.apply(arg, args=args, kwargs=kwargs)
+            return self.apply(arg, raw=False, args=args, kwargs=kwargs)
```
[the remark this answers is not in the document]
jorisvandenbossche: It's not that a special case, it happens when the user passes a custom function to `r.agg(..)`, which eg is done in the documentation about it: http://pandas.pydata.org/pandas-docs/stable/computation.html#aggregation

So just as with `apply`, if the user has a function that behaves differently for a Series than an array, this will break.

jreback:
any comments.

TomAugspurger (approved):

jorisvandenbossche:
@jreback can you *please* work with additional commits, that makes a second review round really a lot easier for me

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -843,6 +872,9 @@ Deprecations
 - ``Index.summary()`` is deprecated and will be removed in a future version (:issue:`18217`)
 - ``NDFrame.get_ftype_counts()`` is deprecated and will be removed in a future version (:issue:`18243`)
 - The ``convert_datetime64`` parameter in :func:`DataFrame.to_records` has been deprecated and will be removed in a future version. The NumPy bug motivating this parameter has been resolved. The default value for this parameter has also changed from ``True`` to ``None`` (:issue:`18160`).
+- :func:`Series.rolling().apply() <pandas.core.window.Rolling.apply>`, :func:`DataFrame.rolling().apply() <pandas.core.window.Rolling.apply>`,
+  :func:`Series.expanding().apply() <pandas.core.window.Expanding.apply>`, and :func:`DataFrame.expanding().apply() <pandas.core.window.Expanding.apply>` will show a ``FutureWarning`` if the new
+  ``raw`` parameter is not explicity passed (:issue:`20584`)
```
jorisvandenbossche: I would rather say *what* has been deprecated: passing an array by default (and then indeed `raw` can be used to keep this behaviour or not)

pandas/core/window.py
```diff
@@ -954,23 +954,53 @@ def count(self):
...
+    raw : bool, default None
+        * ``False`` : passes each row or column as a Series to the
+          function.
+        * ``True`` or ``None`` : the passed function will receive ndarray
+          objects instead.
+          If you are just applying a NumPy reduction function this will
+          achieve much better performance.
+
+        The raw parameter is required and will show a FutureWarning if
+        not passed. In the futures raw will default to False.
```
jorisvandenbossche: futures -> future, raw -> `` `raw` ``

pandas/core/window.py
```diff
@@ -954,23 +954,53 @@ def count(self):
     Parameters
     ----------
     func : function
-        Must produce a single value from an ndarray input
-        \*args and \*\*kwargs are passed to the function""")
+        Must produce a single value from an ndarray input if raw=True
```
jorisvandenbossche: quote raw=True as a code snippet (``` ``raw=True`` ```), same on line below for raw=False

pandas/core/window.py
```diff
@@ -954,23 +954,53 @@ def count(self):
...
         _level = kwargs.pop('_level', None)  # noqa
         window = self._get_window()
         offset = _offset(window, self.center)
         index, indexi = self._get_index()
 
+        # TODO: default is for backward compat
+        # change to False in the future
+        if raw is None:
+            warnings.warn(
+                "raw defaults to False "
```
jorisvandenbossche: Can you put single quotes around raw? (like "'raw' defaults to False ...") And two lines below as well. I think that will make it read better (clearer that it is a keyword)

jorisvandenbossche: Also, I think it is the other way around? The warning says "default is False, in future will change to True", but isn't it the other way around?

jorisvandenbossche (commented):
I think we still need to decide about the `rolling().agg` case (https://github.com/pandas-dev/pandas/pull/20584/files#r180718011). It got (in certain cases of passing a custom function) a ndarray before, which is now changed to a Series. 
I don't think it is worth adding a `raw` keyword here as well (people who need it can use `apply`), and I am fine with deciding that it is also not worth to deprecate, but then we should do that explicitly (thus also document this API change).

Apart from that, looks good to me, added only some doc comments

jreback:
updated, I adjusted the ``.aggregate`` doc-string & put in an api-note. lmk if any further comments.

jreback:
bombs away

jorisvandenbossche:
Thanks!

H0bbitBaron:
@jreback hello! I googled insanely, but now I'm even more confused. Is there a way to pass 2D with few columns instead of 1D series to apply(function)? There was a function rolling_apply, but it's depricated. Ended up writing my own function with cycle inside, which emulates rolling window with some cross-column calculations, but it's rather slow. Thank you.

Issue #19209, opened 2018-01-12
IntervalIndex nbytes computation


TomAugspurger:
Currently, it looks like `IntervalIndex.nbytes` just uses `IndexOpsMixin.nbytes`

```python
@property
def nbytes(self):
    """ return the number of bytes in the underlying data """
    return self._values.nbytes
```

Should it instead be `self.left.nbytes + self.right.nbytes`? This will be faster, and I think more correct.

jschendel:
> Should it instead be self.left.nbytes + self.right.nbytes?

That makes more sense to me.  Not sure exactly what the use case of `nbytes` is though, so really just basing this off the docstring description.

PR #20600, opened 2018-04-04
BUG/PERF: Fixed IntervalIndex.nbytes and itemsize

TomAugspurger:
Avoid materializing the ndarray of Intervals. Just get from left and right
instead.

Closes https://github.com/pandas-dev/pandas/issues/19209

### DIFF
--- a/pandas/core/indexes/interval.py
+++ b/pandas/core/indexes/interval.py
@@ -729,6 +729,14 @@ class IntervalIndex(IntervalMixin, Index):
         return self.left.size
 
     @property
+    def nbytes(self):
+        return self.left.nbytes + self.right.nbytes
+
+    @property
+    def itemsize(self):
+        return self.left.itemsize + self.right.itemsize
+
+    @property
     def shape(self):
         # Avoid materializing self.values
         return self.left.shape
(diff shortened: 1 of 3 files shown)


pandas/core/indexes/interval.py
```diff
@@ -728,6 +728,14 @@ def size(self):
         # Avoid materializing self.values
         return self.left.size
 
+    @property
+    def nbytes(self):
+        return self.left.nbytes + self.right.nbytes
+
+    @property
+    def itemsize(self):
+        return self.left.itemsize + self.right.itemsize
```
jreback: this doesn’t really make sense to have 
why do u need this?

pandas/tests/indexes/interval/test_interval.py
```diff
@@ -24,6 +24,7 @@ def name(request):
 
 class TestIntervalIndex(Base):
     _holder = IntervalIndex
+    _compat_props = ['shape', 'ndim', 'size']
```
jreback: what is this for? we don’t use it anywhere

jreback (changes requested):
maybe add an asv?

TomAugspurger:
Sorry, was just using this to demo stuff for a talk on OSS I gave last night.

I think this will be solved by Interval being an ExtensionArray (should have that PR today).

TomAugspurger:
Closing in favor of #20611.

Issue #19696, opened 2018-02-14
ExtensionArray meta-issue


TomAugspurger:
Just so things don't get lost

- [x] Interface, ExtensionBlock, and Internals (#19268)
- [x] Series and DataFrame can hold extension arrays (#19520)
- [x] overridable `assert_*_equal` (https://github.com/pandas-dev/pandas/pull/19863)
- [x] unique (#19869)
- [x] fillna (#19909)
- [x] `__setitem__` works properly (https://github.com/pandas-dev/pandas/pull/19907)
- [x] Document everything: (#19936)
- [x] argsort, sort_values (https://github.com/pandas-dev/pandas/pull/19957)
- [x] factorize (https://github.com/pandas-dev/pandas/pull/20361)
- [x] groupby works properly (blocked by factorize)
- [x] IntervalArray: (https://github.com/TomAugspurger/pandas/tree/pandas-array-upstream%2Bfu1%2Binterval)
- [x] PeriodArray: index backed by PerioidArray, store in Series / Frame
- [x] Refactor arithmetic methods from DatetimeIndexOpsMixin/DatetimeIndex/TimedeltaIndex/PeriodIndex into array classes (#19902)
- [x] Refactor comparison methods from DTI/TDI/PI into array classes
- [ ] select_dtypes

jbrockmendel:
Should these have a `_typ` attribute?  If so, should `SparseArray._typ` be changed from "array" to something more specific?

TomAugspurger:
We're adding an ABCExtensionArray and `EA._type` with the value 'extension'
in https://github.com/pandas-dev/pandas/pull/19520.

SparseArray being _array isn't the perfect name, but I don't think that
matters, do you? The more important name is the ABC.

TomAugspurger:
Quick status update. The two big ones are in, and now work can mostly proceed in parallel (aside from https://github.com/pandas-dev/pandas/pull/19863 which a couple of my followups depend on).

I'm currently rebasing all my branches on master and will make PRs today.

I'll also go through all the TODOs we've added an make issues / checkboxes for those.

If anyone is interested in working on these, `PeriodArray` will likely be the most sizable PR that I don't really have a start on, aside from a now out of date branch.

jbrockmendel:
[...]

Issue #19371, opened 2018-01-24
API: Add closed to IntervalDtype


jschendel:
This was brought up by @shoyer in a different issue, and it makes sense to me

Adding `closed` to `IntervalDtype` makes sense to me, as in my mind `closed` is inherently part of the dtype.  Since differing `closed` makes two `IntervalIndex` incompatible, I'd expect their dtypes to not be equal, but they currently are, which seems a little strange:
```python
In [2]: ii1 = pd.interval_range(0, 3, closed='left')

In [3]: ii2 = pd.interval_range(0, 3, closed='right')

In [4]: ii1.dtype == ii2.dtype
Out[4]: True
```

There's also a larger discussion as to if adding `closed` to `IntervalDtype` would allow us to remove `closed` as a parameter to the `IntervalIndex` constructor.  My preference would be to keep the `closed` parameter for user convenience, similar to how `CategoricalIndex` accepts `categories` and `ordered` parameters despite also accepting `CategoricalDtype` (though maybe this is just for legacy reasons).  Willing to be convinced otherwise though.

xref https://github.com/pandas-dev/pandas/issues/19263#issuecomment-359986104 (original comment)
xref #19370 (might be rendered moot)

cc @shoyer

jreback:
+1 on this.

jbrockmendel:
+1; recent work in implementing validators this would simplify things

PR #20611, opened 2018-04-04
REF: IntervalIndex[IntervalArray]

### LINKED ISSUE #19209: IntervalIndex nbytes computation

### LINKED ISSUE #19453: API/ENH: Allow Interval to be a first-class type via an IntervalBlock internal type
This has been mentioned in passing in a few other issues, so making an official issue to document this.  In the same spirit as `PeriodBlock` (#7964).

I've started working on an interval accessor, so it seems likely that there will be more exposure to `Interval` objects in a `Series`/`DataFrame`, hence this becoming more relevant.

xref #19272

TomAugspurger:
Closes #19453
Closes #19209


pandas/core/indexes/base.py
```diff
@@ -270,7 +270,8 @@ def __new__(cls, data=None, dtype=None, copy=False, name=None,
                                     **kwargs)
 
         # interval
-        if is_interval_dtype(data) or is_interval_dtype(dtype):
+        if ((is_interval_dtype(data) or is_interval_dtype(dtype)) and
+                not is_object_dtype(dtype)):
```
TomAugspurger: This is so `Index(intervals, dtype=object)` works correctly (Index, not IntervalIndex)

pandas/core/util/hashing.py
```diff
@@ -205,7 +205,9 @@ def _hash_categorical(c, encoding, hash_key):
     -------
     ndarray of hashed values array, same size as len(c)
     """
-    hashed = hash_array(c.categories.values, encoding, hash_key,
+    # Convert ExtensionArrays to ndarrays
+    values = np.asarray(c.categories.values)
```
TomAugspurger: Same as above. `Categorica[ExtensionArray].values` will be an EA, and this has to be an ndarray.

pandas/io/packers.py
```diff
@@ -402,13 +403,17 @@ def encode(obj):
                     u'freq': u_safe(getattr(obj, 'freqstr', None)),
                     u'tz': tz,
                     u'compress': compressor}
-        elif isinstance(obj, IntervalIndex):
-            return {u'typ': u'interval_index',
+        elif isinstance(obj, (IntervalIndex, IntervalArray)):
```
TomAugspurger: Should talk about this more generally in https://github.com/pandas-dev/pandas/issues/20612

pandas/tests/dtypes/test_dtypes.py
```diff
@@ -546,10 +546,8 @@ def test_basic(self):
 
         s = Series(ii, name='A')
 
-        # dtypes
```
TomAugspurger: API change.

pandas/tests/indexes/common.py
```diff
@@ -886,7 +886,7 @@ def test_hasnans_isnans(self):
                 assert not idx.hasnans
 
                 idx = index.copy()
-                values = idx.values
+                values = np.asarray(idx.values)
```
TomAugspurger: Made this change since `IntervalArray` currently isn't mutable. Worth talking about that.

pandas/tests/indexes/interval/test_astype.py
```diff
@@ -51,7 +51,7 @@ def test_astype_category(self, index):
         'timedelta64', 'timedelta64[ns]', 'datetime64', 'datetime64[ns]',
         'datetime64[ns, US/Eastern]'])
     def test_astype_cannot_cast(self, index, dtype):
-        msg = 'Cannot cast IntervalIndex to dtype'
+        msg = 'Cannot cast IntervalArray to dtype'
```
TomAugspurger: Hmm shouldn't have changed this. Looking into it...

TomAugspurger (commented):
High-level summary:

- Still have IntervalIndex composed with IntervalArray. This seems the most sensible for me.
- IntervalArray currently isn't mutable. That could be changed but will require some care to ensure that `.left` and `.right` are modified inplace *atomically*

pandas/core/indexes/interval.py
```diff
@@ -189,161 +127,63 @@ class IntervalIndex(IntervalMixin, Index):
...
-    cut, qcut : Convert arrays of continuous data into Categoricals/Series of
-                Intervals
-    """
+))
+class IntervalIndex(IntervalMixin, Index):
     _typ = 'intervalindex'
     _comparables = ['name']
     _attributes = ['name', 'closed']
+    _allow_index_ops = True
+    _exception_rewrite = lambda: rewrite_exception('IntervalArray',
```
TomAugspurger: Not sure what people's thoughts are. Without this, the exception would be

```python
In [6]: pd.IntervalIndex(2)
```

```pytb
TypeError: IntervalArray(...) must be called with a collection of some kind, 2 was passed
```

This lets the message be `IntervalIndex(...)`.

pandas/core/arrays/__init__.py
```diff
@@ -1,2 +1,10 @@
-from .base import ExtensionArray  # noqa
-from .categorical import Categorical  # noqa
+from .base import ExtensionArray
+from .categorical import Categorical
+from .interval import IntervalArray
+
+
+__all__ = [
```
jreback: you don’t need the all

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,838 @@
...
+        return cls._simple_new(left, right, closed, copy=copy, dtype=dtype,
+                               verify_integrity=verify_integrity)
+
+    @classmethod
+    def _simple_new(cls, left, right, closed=None,
+                    copy=False, dtype=None, verify_integrity=True):
+        result = IntervalMixin.__new__(cls)
+
+        closed = closed or 'right'
+        left = _ensure_index(left, copy=copy)
```
jreback: this is SO duplicating II
have to do something about this - this is a technical debt nightmare

jreback (changes requested):
see comments -1 unless much more shared code

jreback:
you copied code from II this is completely he wrong direction to go
instead move it to a shared mixin

pandas/core/arrays/__init__.py
TomAugspurger: That's so the `#noqa` aren't needed.

pandas/core/arrays/interval.py
TomAugspurger: Did you see IntervalIndex._simple_new? There isn't any duplicated code between the two.

TomAugspurger:
>  instead move it to a shared mixin

Could you maybe demo a single method that you think can be moved to a mixin?

jreback (changes requested):

pandas/core/indexes/interval.py
```diff
@@ -412,272 +252,58 @@ def contains(self, key):
...
-        Return an Index of tuples of the form (left, right)
-
-        Parameters
-        ----------
-        na_tuple : boolean, default True
-            Returns NA as a tuple if True, ``(nan, nan)``, or just as the NA
-            value itself if False, ``nan``.
-
-            ..versionadded:: 0.23.0
-
```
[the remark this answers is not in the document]
TomAugspurger: https://github.com/pandas-dev/pandas/pull/20611/files/9e5fc500d0fb7abd3e270771b138cce905923430#r179213986

TomAugspurger:
> my point is evey method here should be shared
> and if not you have to have a really really good reason

Again, I really don't think there's any code duplication in the current implementation. We've been discussing Indexes subclassing EAs in https://github.com/pandas-dev/pandas/issues/19696, but the prevailing opinion there is that composition is better, specifically https://github.com/pandas-dev/pandas/issues/19696#issuecomment-376284210 and https://github.com/pandas-dev/pandas/issues/19696#issuecomment-377505412. If you want to weigh in there, then please do.

But assuming we do have `IntervalIndex` subclass `IntervalArray`, what does that actually gain? We can eliminate things like

```@property
def left(self):
    return self._data.left
```

Ok... I don't think that's worth the extra time it'll take to rewrite CategoricalIndex and IntervalIndex to use inheritance.

pandas/core/indexes/interval.py
```diff
@@ -1347,7 +926,7 @@ def _format_data(self, name=None):
                 tail = [formatter(x) for x in self]
                 summary = '[{tail}]'.format(tail=', '.join(tail))
 
-        return summary + self._format_space()
+        return summary + ',' + self._format_space()
```
TomAugspurger: note: This was missing a `,` at the end of the first line.

```python
In [2]: pd.IntervalIndex.from_breaks([1, 2])
Out[2]:
IntervalIndex([(1, 2]]  # <-- missing comma
              closed='right',
              dtype='interval[int64]')
```

jschendel:
I think also closes #19453, as it achieves the desired end result?  At the very least it looks to implement the behavior I wanted from that issue.  Will try taking a closer look at the changes here after work.

TomAugspurger:
Thanks, forgot about that issue. Adding to the closed list.

TomAugspurger:
Did a basic version of `IntervalArray.__setitem__` in https://github.com/pandas-dev/pandas/pull/20611/commits/de96a61c4d46cc7ca7e4249bd4a0e63bc68b549b.

I think that we have to copy both `left` and `right`, to ensure that we're never in a state where only `left.__setitem__` succeeds, while `right.__setitem__` fails.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+        right = self.right.copy(deep=True) if deep else self.right
+        closed = self.closed
+        # TODO: Could skip verify_integrity here.
+        return type(self).from_arrays(left, right, closed=closed)
+
+    def _formatting_values(self):
+        return self.values
+
+    def get_values(self):
+        return self.values
```
jorisvandenbossche: Is this method needed?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+        except ValueError:
+
+            # we need to coerce; migth have NA's in an
+            # integer dtype
+            new_left = taker(left.astype(float))
+            new_right = taker(right.astype(float))
+
+        return self._shallow_copy(new_left, new_right)
+
+    take_nd = take
```
jorisvandenbossche: Don't think this should be needed?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+                        (self.left[:-1] > self.right[1:]).all())
+
+        # non-strict inequality when closed != 'both'; at least one side is
+        # not included in the intervals, so equality does not imply overlapping
+        return bool((self.right[:-1] <= self.left[1:]).all() or
+                    (self.left[:-1] >= self.right[1:]).all())
+
+    # Conversion
+    @property
+    def values(self):
```
jorisvandenbossche: I would not call this `.values` (is this property required somewhere else?) 
This is basically `__array__` ?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+        examples='',
+    ))
+    def to_tuples(self, na_tuple=True):
+        tuples = _asarray_tuplesafe(zip(self.left, self.right))
+        if not na_tuple:
+            # GH 18756
+            tuples = np.where(~self.isna(), tuples, np.nan)
+        return tuples
+
+    def tolist(self):
```
jorisvandenbossche: Is this needed somewhere?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+        Return a list of Interval objects.
+
+        See Also
+        --------
+        numpy.ndarray.tolist
+        """
+        # TODO: think about putting this in a parent
+        return self.values.tolist()
+
+    def repeat(self, repeats):
```
jorisvandenbossche: idem, is this needed?

pandas/core/indexes/interval.py
```diff
@@ -691,61 +317,67 @@ def left(self):
...
-                   'e.g. Intervals with string endpoints')
-            raise TypeError(msg)
+        return self._data.length
 
     @property
     def size(self):
-        # Avoid materializing self.values
-        return self.left.size
+        # Avoid materializing ndarray[Interval]
+        return self._data.size
```
jorisvandenbossche: This could in principle be moved to the base Index class (same for shape, itemsize, dtype, ..)

jorisvandenbossche (commented):
I won't have time this week (offline the next days) to do a thorough review, but +1 on general approach. Thanks a lot for this!

For IntervalIndex, if we want to reduce the lines of code, we could rather easily add the interval-specific methods and attributes (left, right, closed, length, mid, is_non_overlapping_monotonic, ...) in an automatic way (like we add arithmetic methods on DataFrame etc). 
Basically the methods that just passthrough the `self._data` equivalent and could also inherit the docstring automatically.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -299,6 +299,41 @@ Supplying a ``CategoricalDtype`` will make the categories in each column consist
     df['A'].dtype
     df['B'].dtype
 
+.. _whatsnew_023.enhancements.interval:
```
jschendel: ` _whatsnew_023.` --> ` _whatsnew_0230.` and likewise on line 337.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+        left = self.left[value]
+        right = self.right[value]
+
+        # scalar
+        if not isinstance(left, Index):
+            return Interval(left, right, self.closed)
+
+        return self._shallow_copy(left, right)
+
+    def __setitem__(self, key, value):
```
jschendel: Should this support the case where `value` is `np.nan`?  Doesn't look like that currently works.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+        if not (is_interval_dtype(value) or isinstance(value, Interval)):
```
jschendel: Should `ABCInterval` be used for the `isinstance` check?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+    def fillna(self, value=None, method=None, limit=None):
+        if method is not None:
+            raise TypeError('Filling by method is not supported for '
+                            'IntervalArray.')
+        if limit is not None:
+            raise TypeError('limit is not supported for IntervalArray.')
+
+        if not isinstance(value, Interval):
+            msg = ("Interval.fillna only supports filling with scalar "
+                   "'value'. Got a '{}' instead of a 'pandas.Interval'."
```
jschendel: This message seems a little unclear/confusing.  Maybe something like: `"IntervalArray.fillna only supports filling with pandas.Interval, got a '{}' instead."`

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+        left = self.left.copy(deep=True)
+        right = self.right.copy(deep=True)
+
+        left.values[key] = value.left
+        right.values[key] = value.right
+
+        self._left = left
+        self._right = right
+
+    def fillna(self, value=None, method=None, limit=None):
```
jschendel: Should we do some extra checks on `value`?  Right now if I pass an `Interval` with a different `closed`, it silently gets changed to match the array's `closed`:

```python
In [2]: ia = pd.interval_range(0, 4).insert(0, np.nan).values

In [3]: ia
Out[3]:
IntervalArray([nan, (0.0, 1.0], (1.0, 2.0], (2.0, 3.0], (3.0, 4.0]],
              closed='right',
              dtype='interval[float64]')

In [4]: ia.fillna(pd.Interval(10, 20, closed='both'))
Out[4]:
IntervalArray([(10.0, 20.0], (0.0, 1.0], (1.0, 2.0], (2.0, 3.0], (3.0, 4.0]],
              closed='right',
              dtype='interval[float64]')
```
Could probably create some type of helper function/method for this, as I'd guess it'd be somewhat of a common thing to do.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+        # scalar
+        if not isinstance(left, Index):
+            return Interval(left, right, self.closed)
+
+        return self._shallow_copy(left, right)
+
+    def __setitem__(self, key, value):
+        if not (is_interval_dtype(value) or isinstance(value, Interval)):
+            msg = "'value' should be an interval type, got {} instead."
+            raise TypeError(msg.format(type(value)))
```
jschendel: Looks like generally `type(value).__name__` is used in messages instead of just `type(value)`.  Not super concerned with this distinction though, just a minor inconsistency I noticed.

jschendel (commented):
A few quick comments; will review more later.

pandas/core/arrays/interval.py
jreback: so EA takes a different take on construction that in Index. Where we use _simple_new and _shalllow_copy, but conceptually these are the same things. Would be ok with changing the spellings inside Index to conform this. This is my main objection generally. We spell things one way and a different way in EA, this is so confusing.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+
+        left = self.left.fillna(value=value.left)
+        right = self.right.fillna(value=value.right)
+        return self._shallow_copy(left, right)
+
+    @property
+    def dtype(self):
+        return IntervalDtype.construct_from_string(str(self.left.dtype))
+
+    def astype(self, dtype, copy=True):
```
jreback: doc-string.  also these should have an ``errors=`` parameters. We do everywhere else (should be pushed to EA generally)

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+        if len(closed) != 1:
+            raise ValueError("Intervals must all be closed on the same side.")
+        closed = closed.pop()
+
+        # TODO: avoid intermediate list
+        left = np.concatenate([interval.left for interval in to_concat])
+        right = np.concatenate([interval.right for interval in to_concat])
+        return cls._simple_new(left, right, closed=closed, copy=False)
+
+    def _shallow_copy(self, left=None, right=None):
```
jreback: is this a method in EA? it should be

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+        else:
+
+            # both left and right are values
+            pass
+
+        return self._simple_new(left, right, closed=self.closed,
+                                verify_integrity=False)
+
+    # TODO: doc
+    def copy(self, deep=False):
```
jreback: is this in EA?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+    @property
+    def nbytes(self):
+        return self.left.nbytes + self.right.nbytes
+
+    @property
+    def itemsize(self):
+        return self.left.itemsize + self.right.itemsize
+
+    def take(self, indices, axis=0, allow_fill=True, fill_value=None,
+             **kwargs):
```
jreback: doc. wonder how of this you can push to EA base clasess (I suspect almost all of it), this is a pure indexing operation.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+            new_left = taker(left.astype(float))
+            new_right = taker(right.astype(float))
+
+        return self._shallow_copy(new_left, new_right)
+
+    take_nd = take
+
+    # Formatting
+
+    def _format_data(self):
```
jreback: this is really a generic formatter Index. should this be in EA?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,856 @@
...
+                n = min(max_seq_items // 2, 10)
+                head = [formatter(x) for x in self[:n]]
+                tail = [formatter(x) for x in self[-n:]]
+                summary = '[{head} ... {tail}]'.format(
+                    head=', '.join(head), tail=', '.join(tail))
+            else:
+                head = []
+                tail = [formatter(x) for x in self]
+                summary = '[{tail}]'.format(tail=', '.join(tail))
+
```
jreback: so there is lots of copying of other code because you can't share the Index base things for this. You need prob push some of this code (e.g. formatting) to something like IndexOpsMixin (or the like). and use that here.

jreback (changes requested):
left some comments. But this move shows a lot of rough edges. you had to copy quite a bit of code that is *already* in the index by base class. I really dont' want to re-invent the wheel. Need to decide what goes to EA base class (but that can't be used in Index ATM), and what should be in a shared EA/Index Mixin. This is getting pretty messy.

pandas/core/indexes/interval.py
```diff
@@ -691,61 +317,67 @@ def left(self):
...
 
     @property
     def shape(self):
-        # Avoid materializing self.values
-        return self.left.shape
+        # Avoid materializing ndarray[Interval]
+        return self._data.shape
+
+    @property
```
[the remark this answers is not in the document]
TomAugspurger: > its a very odd thing to have generally for non-scalar based ndarray like things

A bit strange, yeah, since there are two representations: left & right arrays, or an array of Interval objects.

Anyway, on master this inherited from IndexOpsMixin, which reports the `ndarray[Interval]` version.
This is inconsistent with `Categorical`, which reports the itemsize of the codes. I'd also prefer that it's consistent with `nbytes`, which (should) report the memory used by the ExtensionArray, not the array of objects.

pandas/core/util/hashing.py
TomAugspurger: I don't think so, `asarray` won't copy an ndarray, so calling it on an ndarray should be fine.

`is_extension_array_dtype` takes ~ 2 microseconds, and `asarray` takes ~10 ns, so better to just always call asarray instead of checking.

pandas/io/packers.py
TomAugspurger: Mmm, that doesn't sound fun. I'll see what I can do to avoid that.

pandas/core/arrays/interval.py
TomAugspurger: Yes, it currently fails. I'll see what can be done.

pandas/core/arrays/interval.py
jorisvandenbossche: I am not sure it is needed to add the `errors` keyword to our arrays (Categorical also does not have this, and none of the indexes have it). For me it would be fine to keep this as an extra feature on `Series.astype`

pandas/core/arrays/interval.py
jorisvandenbossche: As long as this is only uses internally within the IntervalArray class, it is not necessarily needed to make this part of the official ExtensionArray interface. 

What are the advantages in doing so? (also the signature will be different for each array type)

pandas/core/arrays/interval.py
jorisvandenbossche: Yes, `copy` is part of the extension array interface

pandas/core/arrays/interval.py
jorisvandenbossche: I might be in favor of adding some basic implementation of `take` to EA, but in this case, `IntervalArray` would have to override it anyhow, as it needs to take both from `left` and `right`, which is something very specific to intervals.

pandas/core/arrays/interval.py
jorisvandenbossche: Or rather somewhere in the formatting code? (and then it could still be called from within a default EA repr)

pandas/core/arrays/interval.py
jorisvandenbossche: @jreback Can you elaborate on this comment (as you say this is your main objection, but I don't fully understand it).

So `IntervalArray._simple_new` construct from left/right, while `IntervalIndex._simple_new` now constructs from values (IntervalArray), while previously it constructed from left/right. So yes, this is different between array and index, but, for IntervalIndex it is now more consistent with the base Index._simple_new (which constructs from values), so I think that is a good thing.

pandas/core/arrays/interval.py
jschendel: Looks like there's also something strange going on with datetime data:
```python-traceback
In [2]: ia = pd.interval_range(pd.Timestamp('20180101'), periods=3).values

In [3]: ia
Out[3]:
IntervalArray([(2018-01-01, 2018-01-02], (2018-01-02, 2018-01-03], (2018-01-03, 2018-01-04]],
              closed='right',
              dtype='interval[datetime64[ns]]')

In [4]: new_iv = pd.Interval(pd.Timestamp('20180101'), pd.Timestamp('20180105'))

In [5]: new_iv
Out[5]: Interval('2018-01-01', '2018-01-05', closed='right')

In [6]: ia[0] = new_iv

In [7]: ia
Out[7]: ---------------------------------------------------------------------------
ValueError: left side of interval must be <= right side

In [8]: ia.left
Out[8]: DatetimeIndex(['2018-01-01', '2018-01-05', '2018-01-03'], dtype='datetime64[ns]', freq='D')

In [9]: ia.right
Out[9]: DatetimeIndex(['2018-01-05', '2018-01-03', '2018-01-04'], dtype='datetime64[ns]', freq='D')
```

TomAugspurger:
(Pulling the `__repr__` discussion out to the main thread.)

So, AFAICT, the only duplicate code between IntervalIndex and IntervalArray is in formatting. This is just like the duplicate code between CategoricalIndex and CategoricalArray. I can move the "array" formatting, which will be common to Index and pandas' extension arrays, to a new mixin class. I can do that as part of this PR, or a followup.

Still need to fixup `__setitem__` calls that modify the type of `left` & `right` and address a few other comments.

pandas/core/arrays/interval.py
```diff
@@ -516,10 +517,10 @@ def astype(self, dtype, copy=True):
             else:
                 return self
         elif is_categorical_dtype(dtype):
-            return Categorical(self.values)
+            return Categorical(np.asarray(self))
```
jorisvandenbossche: Not top priority, but this will result in conversion to interval objects and then in categorical back to IntervalIndex. So long term maybe Categorical should be able to handle arrays natively (so you could just pass `self` here) ?

jorisvandenbossche:
Other comment, related to the formatting: for Index we use eg `interval[int64]` as dtype repr to indicate the type of the intervals, but for Series repr it only uses 'interval' (not sure if it is needed there, `Series(..).dtype` still gives `interval[int64]`

doc/source/basics.rst
```diff
@@ -1925,11 +1925,23 @@ untouched. If the data is modified, it is because you did so explicitly.
...
+columns of a DataFrame. The main types allowed in pandas objects are ``float``,
+``int``, ``bool``, and ``datetime64[ns]`` (note that NumPy does not support
+timezone-aware datetimes).
+
+In addition to NumPy's types, pandas :ref:`extends <extending.extension-types>`
+NumPy's type-system for a few cases.
+
+* :ref:`Categorical <categorical>`
+* :ref:`Datetime with Timezone <timeseries.timezone_series>`
+* Interval
```
jorisvandenbossche: also Periods ?

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -299,7 +299,42 @@ Supplying a ``CategoricalDtype`` will make the categories in each column consist
...
+   ser = pd.Series(pd.interval_range(0, 5))
+   ser
+   ser.dtype
+
+Previously, these would be cast to a NumPy array of Interval objects. In general,
+this should result in better performance when storing an array of intervals in
+a Series.
+
+Note that the ``.values`` of a Series containing intervals is no longer a NumPy
+array. Rather, it's an ``ExtensionArray``, composed of two arrays ``left`` and
```
jorisvandenbossche: IntervalArray instead of ExtensionArray ?

pandas/compat/__init__.py
```diff
@@ -451,3 +451,11 @@ def is_platform_mac():
 
 def is_platform_32bit():
     return struct.calcsize("P") * 8 < 64
+
+
+class _WritableDoc(type):
```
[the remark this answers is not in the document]
jreback: actually we have a pandas.util._doctools, move there

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+                fastpath=False, verify_integrity=True):
+
+        if fastpath:
+            return cls._simple_new(data.left, data.right, closed,
+                                   copy=copy, dtype=dtype,
+                                   verify_integrity=False)
+
+        if isinstance(data, ABCSeries) and is_interval_dtype(data):
+            data = data.values
+        if isinstance(data, (cls, ABCIntervalIndex)):
```
jreback: blank line

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+            closed = data.closed
+        else:
+
+            # don't allow scalars
+            if is_scalar(data):
+                msg = ("{}(...) must be called with a collection of some kind,"
+                       " {} was passed")
+                raise TypeError(msg.format(cls.__name__, data))
+
+            data = maybe_convert_platform_interval(data)
```
jreback: can you comment this

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+            left, right, infer_closed = intervals_to_interval_bounds(data)
+
+            if (com._all_not_none(closed, infer_closed) and
```
jreback: don't you have a _check_closed function? maybe remove that from Interval (class) and make it a free function?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+
+        return cls._simple_new(left, right, closed, copy=copy, dtype=dtype,
+                               verify_integrity=verify_integrity)
+
+    @classmethod
+    def _simple_new(cls, left, right, closed=None,
+                    copy=False, dtype=None, verify_integrity=True):
+        result = IntervalMixin.__new__(cls)
+
+        closed = closed or 'right'
```
jreback: did we default this before here? (e.g. this should be filled in I think by now). maybe assert that it is not None here (and push the default higher up). IOW don't want to have in too many places.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+    """
+
+    @classmethod
+    @Appender(_interval_shared_docs['from_tuples'] % _shared_docs_kwargs)
+    def from_tuples(cls, data, closed='right', copy=False, dtype=None):
+        if len(data):
+            left, right = [], []
+        else:
+            left = right = data
+
```
jreback: can you add a comment here on what is going on

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+                    raise TypeError(msg)
+                lhs, rhs = d
+            left.append(lhs)
+            right.append(rhs)
+
+        return cls.from_arrays(left, right, closed, copy=False,
+                               dtype=dtype)
+
+    def _validate(self):
+        """Verify that the IntervalIndex is valid.
```
jreback: see my comment above about setting closed. I *think* you *always* validate before calling _simple_new (though maybe not in the main constructor)?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+    # Interface
+    # ---------
+    def __iter__(self):
+        return iter(np.asarray(self))
+
+    def __len__(self):
+        return len(self.left)
+
+    def __getitem__(self, value):
+        mask = self.isna()[value]
```
jreback: doc-string?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+        right = self.right.copy(deep=True)
+
+        left.values[key] = value.left
+        right.values[key] = value.right
+
+        self._left = left
+        self._right = right
+
+    def fillna(self, value=None, method=None, limit=None):
+        if method is not None:
```
jreback: doc-string?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+        # TODO: This try/except will be repeated.
+        try:
+            return np.asarray(self).astype(dtype, copy=copy)
+        except (TypeError, ValueError):
+            msg = 'Cannot cast {name} to dtype {dtype}'
+            raise TypeError(msg.format(name=type(self).__name__, dtype=dtype))
+
+    @classmethod
+    def _concat_same_type(cls, to_concat):
+        closed = set(interval.closed for interval in to_concat)
```
jreback: can you add a comment / doc-string here

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+        closed = closed.pop()
+
+        # TODO: avoid intermediate list
+        left = np.concatenate([interval.left for interval in to_concat])
+        right = np.concatenate([interval.right for interval in to_concat])
+        return cls._simple_new(left, right, closed=closed, copy=False)
+
+    def _shallow_copy(self, left=None, right=None):
+        if left is None:
+
```
jreback: can you add a doc-string.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+        return self.left.nbytes + self.right.nbytes
+
+    @property
+    def itemsize(self):
+        return self.left.itemsize + self.right.itemsize
+
+    def take(self, indices, axis=0, allow_fill=True, fill_value=None,
+             **kwargs):
+        nv.validate_take(tuple(), kwargs)
+        indices = _ensure_platform_int(indices)
```
jreback: doc-string

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+            new_right = taker(right.astype(float))
+
+        return self._shallow_copy(new_left, new_right)
+
+    # Formatting
+
+    def _format_data(self):
+
+        # TODO: integrate with categorical and make generic
+        # name argument is unused here; just for compat with base / categorical
```
jreback: yep

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,814 @@
...
+        """
+        Return the midpoint of each Interval in the IntervalIndex as an Index
+        """
+        try:
+            return 0.5 * (self.left + self.right)
+        except TypeError:
+            # datetime safe version
+            return self.left + 0.5 * self.length
+
+    @property
```
jreback: move these 2 up near nbytes

pandas/core/indexes/base.py
```diff
@@ -270,7 +270,8 @@ def __new__(cls, data=None, dtype=None, copy=False, name=None,
                                     **kwargs)
 
         # interval
-        if is_interval_dtype(data) or is_interval_dtype(dtype):
+        if ((is_interval_dtype(data) or is_interval_dtype(dtype)) and
```
jreback: this hit in the tests?

pandas/core/indexes/interval.py
```diff
@@ -189,161 +128,61 @@ class IntervalIndex(IntervalMixin, Index):
...
-
-        return cls._simple_new(left, right, closed, name, copy=copy,
-                               dtype=dtype, verify_integrity=verify_integrity)
+        return cls._simple_new(array, name)
 
     @classmethod
-    def _simple_new(cls, left, right, closed=None, name=None, copy=False,
-                    dtype=None, verify_integrity=True):
-        result = IntervalMixin.__new__(cls)
+    def _simple_new(cls, array, name, closed=None):
```
jreback: prob can consolidate this with base Index (why is closed passed?)

pandas/core/indexes/interval.py
```diff
@@ -1261,33 +862,10 @@ def _concat_same_dtype(self, to_concat, name):
...
-            new_right = taker(right)
-        except ValueError:
-
-            # we need to coerce; migth have NA's in an
-            # integer dtype
-            new_left = taker(left.astype(float))
-            new_right = taker(right.astype(float))
-
-        return self._shallow_copy(new_left, new_right)
+        result = self._data.take(indices, axis=axis, allow_fill=allow_fill,
```
jreback: this looks like the higher level (in base) maybe can just remove

pandas/util/exceptions.py
```diff
@@ -0,0 +1,16 @@
+import contextlib
+
+
```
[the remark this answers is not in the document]
jreback: you can leave, but reanme to _exceptions

jreback (changes requested):
lgtm overall.

jschendel:
Pushed some pre-rebase commits:
- Added NA support to `__setitem__`
- Implemented `IntervalArray.repeat` to fix some failing tests that were complaining about it not existing
- Addressed a number of review comments
   - Still a few that need to be addressed, but mostly docstring related

TomAugspurger (commented):
Overall looks good. I think you're good to merge in master anytime.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -299,7 +299,42 @@ Supplying a ``CategoricalDtype`` will make the categories in each column consist
     df['A'].dtype
     df['B'].dtype
 
-.. _whatsnew_023.enhancements.extension:
+.. _whatsnew_0230.enhancements.interval:
```
jreback: this will need changing during the rebase

jreback (approved):

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -66,6 +66,43 @@ Current Behavior:
 
     result
 
+
+.. _whatsnew_0240.enhancements.interval:
+
+Storing Interval Data in Series and DataFrame
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Interval data may now be stored in a Series or DataFrame, in addition to an
```
jreback: double backticks on Series/DataFrame.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -66,6 +66,43 @@ Current Behavior:
...
+:class:`IntervalIndex` like before (:issue:`19453`).
```
jreback: before -> previously.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -66,6 +66,43 @@ Current Behavior:
...
+
+.. ipython:: python
+
+   ser = pd.Series(pd.interval_range(0, 5))
+   ser
+   ser.dtype
+
+Previously, these would be cast to a NumPy array of Interval objects. In general,
```
jreback: double backticks around Interval, or use a class ref. & Series

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -66,6 +66,43 @@ Current Behavior:
...
+this should result in better performance when storing an array of intervals in
+a Series.
+
+Note that the ``.values`` of a Series containing intervals is no longer a NumPy
+array. Rather, it's an ``ExtensionArray``, composed of two arrays ``left`` and
```
jreback: I don't think you need to show the internal impl. (e.g. just the first sentence is ok)

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -66,6 +66,43 @@ Current Behavior:
...
+``right``.
+
+.. ipython:: python
+
+   ser.values
+
+To recover the NumPy array of Interval objects, use :func:`numpy.asarray`:
```
jreback: if you are going to have the below section I would not use this here (showing asarray) as its duplicative, just point to the ref for the below section)

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -90,6 +127,45 @@ Other Enhancements
...
+New Behavior:
+
+.. ipython:: python
+
+   idx = pd.interval_range(0, 4)
+   idx.values
+
+This mirrors ``CateogricalIndex.values``, which returns a ``Categorical``.
+
+For situations where you need an ``ndarray`` of Interval objects, use
```
jreback: double backticks around Interval

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -345,6 +421,7 @@ Interval
 ^^^^^^^^
 
 - Bug in the :class:`IntervalIndex` constructor where the ``closed`` parameter did not always override the inferred ``closed`` (:issue:`19370`)
+- Bug in the ``IntervalIndex`` repr missing a trailing comma at the end of the "data" section (:issue`20611`)
```
jreback: what does this mean? what is the 'data' section? that's an internal reference

pandas/core/arrays/categorical.py
```diff
@@ -1218,6 +1219,8 @@ def __array__(self, dtype=None):
         ret = take_1d(self.categories.values, self._codes)
         if dtype and not is_dtype_equal(dtype, self.categories.dtype):
             return np.asarray(ret, dtype)
+        if is_extension_array_dtype(ret):
```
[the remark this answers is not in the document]
jreback: I have an update on this section already in intna

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,1006 @@
...
+            raise ValueError("Intervals must all be closed on the same side.")
+        closed = closed.pop()
+
+        # TODO: avoid intermediate list
+        left = np.concatenate([interval.left for interval in to_concat])
+        right = np.concatenate([interval.right for interval in to_concat])
+        return cls._simple_new(left, right, closed=closed, copy=False)
+
+    def _shallow_copy(self, left=None, right=None, closed=None):
+        if left is None:
```
jreback: can you add a doc-string

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,1006 @@
...
+        Series.value_counts
+        """
+        # TODO: implement this is a non-naive way!
+        from pandas.core.algorithms import value_counts
+        return value_counts(np.asarray(self), dropna=dropna)
+
+    # Formatting
+
+    def _format_data(self):
+
```
jreback: can do in a followup (though prob easy here) to use
``format_object_summary`` from ``pandas.core.formats.printing`` as this is vastly simplified

jreback: its used generally for all Indexes now

pandas/core/util/hashing.py
jreback: do we still need this change? what is it for?

pandas/tests/extension/base/getitem.py
```diff
@@ -240,6 +240,7 @@ def test_reindex_non_na_fill_value(self, data_missing):
         na = data_missing[0]
 
         array = data_missing._from_sequence([na, valid])
+        print(array)
```
jreback: extraneous print

pandas/util/_exceptions.py
```diff
@@ -0,0 +1,16 @@
+import contextlib
+
+
+@contextlib.contextmanager
+def rewrite_exception(old_name, new_name):
```
jreback: what is this for?

jreback (changes requested):
looks pretty good.

doc/source/whatsnew/v0.24.0.txt
jschendel: Cleaned this up.  IIRC this was a side issue that @TomAugspurger found when he originally made the PR.  A comma was missing after the list of intervals in the repr.

pandas/core/indexes/base.py
jschendel: added a test for this that will appear in my next commit

pandas/core/util/hashing.py
jschendel: I tried removing it, and `test_hash_scalar` failed for `Interval(0, 1)` (I've also parametrized the test):
https://github.com/pandas-dev/pandas/blob/1033e8b1195d4071253889ada60523832285354c/pandas/tests/util/test_hashing.py#L91-L102

pandas/util/_exceptions.py
jschendel: This is being used to overwrite exception messages so they display the correct class name.  Looks like it's currently just used for `IntervalIndex` when it calls an underlying `IntervalArray` method that could raise some with an `IntervalArray` specific message, e.g. `'category, object, and string subtypes are not supported for IntervalArray'`.

I suppose we could try to either make these messages slightly more generic ("IntervalArray" --> "IntervalArray-like" or "Interval classes" maybe) or try implementing an additional common interface.

jschendel:
Addressed the latest set of review comments, aside from the `_format_data` one (will do in a followup pr or with  another set of review comments).

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,1024 @@
...
+        if method is not None:
+            raise TypeError('Filling by method is not supported for '
+                            'IntervalArray.')
+        if limit is not None:
+            raise TypeError('limit is not supported for IntervalArray.')
+
+        if not isinstance(value, ABCInterval):
+            msg = ("'IntervalArray.fillna' only supports filling with a "
+                   "'scalar pandas.Interval'. Got a '{}' instead."
+                   .format(type(value).__name__))
```
TomAugspurger: Single quote is in the wrong place. Should be `'pandas.Interval'`.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,1024 @@
...
+        if not isinstance(left, Index):
+            if isna(left):
+                return self._fill_value
+            return Interval(left, right, self.closed)
+
+        return self._shallow_copy(left, right)
+
+    def __setitem__(self, key, value):
+        # na value: need special casing to set directly on numpy arrays
+        _needs_float_conversion = False
```
TomAugspurger: _needs_float_conversion -> needs_float_conversion?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,1024 @@
...
+        Returns
+        -------
+        IntervalArray
+        """
+        closed = set(interval.closed for interval in to_concat)
+        if len(closed) != 1:
+            raise ValueError("Intervals must all be closed on the same side.")
+        closed = closed.pop()
+
+        # TODO: avoid intermediate list
```
TomAugspurger: This TODO (which I think I added) seems unavoidable, right?

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,1024 @@
...
+              closed='right',
+              dtype='interval[int64]')
+        >>>  index.set_closed('both')
+        %(klass)s([[0, 1], [1, 2], [2, 3]]
+              closed='both',
+              dtype='interval[int64]')
+        """
+
+    @Appender(_interval_shared_docs['set_closed'] % _shared_docs_kwargs)
+    def set_closed(self, closed):
```
TomAugspurger: Should we make `IntervalArray.closed` a setter as well?

pandas/io/packers.py
TomAugspurger: I wonder, should this be reverted? Do any msgpack tests fail if it is?

pandas/util/_exceptions.py
TomAugspurger: What about something like defining a custom exception type in arrays/interval.py and then

```
def _raise_exception(self):
    # We have access to the correct `type(self).__name__`
    pass

def method(self):
    try:
        do_thing()
    except _IntervalException:
        self._raise_exception()
```

does that make sense?

TomAugspurger: Or we could just use IntervalArray in the exception message. Doesn't really matter to me.

pandas/io/packers.py
jschendel: The test below fails on the last copied line for `Interval` data if reverted:

https://github.com/pandas-dev/pandas/blob/da6e26db079cd4050c1693939db6a52afb2978e1/pandas/tests/io/test_packers.py#L357-L360

pandas/core/arrays/interval.py
jschendel: I don't think this would be a problem with the current implementation, but #19371 makes me a bit hesitant since it'd make a `closed` setter a (limited) `dtype` setter.

pandas/core/arrays/interval.py
jschendel: Yeah, I don't think it can be avoided. Removed the TODO.  IIUC the intermediate list doesn't copy data, so not a big deal (`np.concatenate` does of course).

pandas/util/_exceptions.py
jschendel: Yeah, I don't really have a preference either.  I might end up trying a few things to see what's cleanest.

jreback (changes requested):

pandas/core/arrays/interval.py
jschendel: There are a couple corner cases: empty data and purely NA data.  In both cases, the inferred `closed` will be `None`, so if `closed` is also not passed to the constructor it will be `None` here.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,1023 @@
...
+            try:
+                array = IntervalArray(value)
+                value_left, value_right = array.left, array.right
+            except TypeError:
+                # wrong type: not interval or NA
+                msg = "'value' should be an interval type, got {} instead."
+                raise TypeError(msg.format(type(value)))
+
+        # Need to ensure that left and right are updated atomically, so we're
+        # forced to copy, update the copy, and swap in the new values.
```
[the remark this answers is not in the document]
jschendel: Without the forcing the copy I'm seeing `test_setitem_sequence` fail:

https://github.com/pandas-dev/pandas/blob/1dd05ccf56b163dc51786e4f5ce76b279cc0dec3/pandas/tests/extension/base/setitem.py#L17-L23

pandas/tests/extension/interval/test_interval.py
```diff
@@ -0,0 +1,193 @@
...
+    def test_non_scalar_raises(self, data_missing):
+        msg = "Got a 'list' instead."
+        with tm.assert_raises_regex(TypeError, msg):
+            data_missing.fillna([1, 1])
+
+
+class TestReshaping(BaseInterval, base.BaseReshapingTests):
+    pass
+
+
```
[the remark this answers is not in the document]
jschendel: I don't follow.  I didn't add any tests to the base class, so not sure what should be moved here?

TomAugspurger:
@jschendel how do you feel about this? IMO, things looked quite nice the last time I looked. Are the outstanding issues

1. Sharing printing between Index and IntervalArray
2. (Maybe) changing `rewrite_exception` to use a method to handle raising with the right name?

Anything else? Are we comfortable doing those as followups? (fine by me)

jschendel:
Yes, I believe those are the only outstanding issues, aside from the question I had about @jreback's last comment.  I'm fine with doing those items as followups.

TomAugspurger:
Cool, +1 from me then.

jreback:
@jschendel what comment? i am ok with merging and doing follow ups as well

jschendel:
@jreback : https://github.com/pandas-dev/pandas/pull/20611#discussion_r201291862

"I suppose should move some of these tests to `pandas/tests/arrays/interval`, maybe leave the extension API specific ones here"

Regardless, doesn't seem like something that would block this from being merged, and could be handled in a followup too.

pandas/core/arrays/interval.py
```diff
@@ -0,0 +1,1049 @@
...
+        # Avoid materializing self.values
+        return self.left.size
+
+    @property
+    def shape(self):
+        return self.left.shape
+
+    @property
+    def itemsize(self):
+        return self.left.itemsize + self.right.itemsize
```
jorisvandenbossche: @jschendel a bit after the fact, but I don't think that we should add this to `IntervalArray`. We deprecated `itemsize` some time ago for Series/Index (https://github.com/pandas-dev/pandas/pull/20721). 
Just noticed this because we get the warning during the test:

```
pandas/tests/indexes/interval/test_interval.py::TestIntervalIndex::()::test_itemsize
  /home/travis/build/pandas-dev/pandas/pandas/core/arrays/interval.py:694: FutureWarning: Int64Index.itemsize is deprecated and will be removed in a future version
    return self.left.itemsize + self.right.itemsize
```

jschendel: This was originally part of `IntervalIndex` and is currently what `IntervalIndex` calls under the hood, so if we remove it from `IntervalArray` we'd need to add it back `IntervalIndex`.  Can certainly do that, unless there's some objection based on it making the two inconsistent.  Removing it from `IntervalArray` seems like the best course of action though.

jorisvandenbossche: Yes, removing it from `IntervalArray` is the best way forward I think. 
Also, on `IntervalIndex` it should be deprecated (now it is deprecated on the base IndexOpsMixin class, but by overriding it in IntervalIndex, it lost that deprecation. But apparently IntervalIndex is the only index that is not tested in the base tests, so this 'regression' was not catched by the tests).

jorisvandenbossche: Opened https://github.com/pandas-dev/pandas/issues/22049 so we don't loose track of this discussion

qwhelan:
A run of `asv find` has identified this PR as resulting in a ~1000x slowdown in the `indexing.IntervalIndexing.time_loc_list` benchmark:
```
asv find 7e6e7^..HEAD indexing.IntervalIndexing.time_loc_list
[...]
·· Installing 365eac4d <master~642> into conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt..
·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
··· Setting up indexing.py:220                                                                                                                                                                         ok
··· Running (indexing.IntervalIndexing.time_loc_list--).
··· indexing.IntervalIndexing.time_loc_list                                                                                                                                                      2.27±0ms
· Testing -----------------------------------O>-----------------------
·· For pandas commit f3160bfa <master~639>:
·· Uninstalling from conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt..
·· Building f3160bfa <master~639> for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt............................................................................................................
·· Installing f3160bfa <master~639> into conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt..
·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
··· Setting up indexing.py:220                                                                                                                                                                         ok
··· Running (indexing.IntervalIndexing.time_loc_list--).
··· indexing.IntervalIndexing.time_loc_list                                                                                                                                                    2.90±0.03s
· Testing -----------------------------------O------------------------
·· For pandas commit 610a19a3 <master~641>:
·· Uninstalling from conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt..
·· Building 610a19a3 <master~641> for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt.................................................................................................................
·· Installing 610a19a3 <master~641> into conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt..
·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
··· Setting up indexing.py:220                                                                                                                                                                         ok
··· Running (indexing.IntervalIndexing.time_loc_list--).
··· indexing.IntervalIndexing.time_loc_list                                                                                                                                                    2.92±0.01s.
· Greatest regression found: 610a19a3 <master~641>
```
Bisection done on Arch Linux and reproducible on Ubuntu 18.04.

PR #20613, opened 2018-04-05
Stop concat from attempting to sort mismatched columns by default

### LINKED ISSUE #4588: BUG: concat unwantedly sorts DataFrame column names if they differ
When concat'ing DataFrames, the column names get alphanumerically sorted if there are any differences between them. If they're identical across DataFrames, they don't get sorted.
This sort is undocumented and unwanted. Certainly the default behavior should be no-sort.

``` python
df4a = DataFrame(columns=['C','B','D','A'], data=np.random.randn(3,4))
df4b = DataFrame(columns=['C','B','D','A'], data=np.random.randn(3,4))
df5  = DataFrame(columns=['C','B','E','D','A'], data=np.random.randn(3,5))

print "Cols unsorted:", concat([df4a,df4b])
# Cols unsorted:           C         B         D         A

print "Cols sorted", concat([df4a,df5])
# Cols sorted           A         B         C         D         E
``'
```

brycepg:
Preserve column order upon concatenation to obey
least astonishment principle.

Allow old behavior to be enabled by adding a boolean switch to
concat and DataFrame.append, mismatch_sort, which is by default disabled.  

- [x] Close #4588
- [x] tests added / passed
- [x] whatsnew entry


pandas/core/reshape/concat.py
```diff
@@ -20,7 +20,7 @@
 
 def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
            keys=None, levels=None, names=None, verify_integrity=False,
-           copy=True):
+           copy=True, mismatch_sort=False):
```
jreback: just call this sort 
put before copy

jreback (changes requested):
needs a few more tests
replicate any tests from the original issue
needs a whatsnew note
will comment on the impl later

jorisvandenbossche:
Thanks for working on this!

* We also need to check how this works in combination with other keywords (eg `join`)
* Do we want to ignore the keyword if all columns match? (what you currently do?)  I would also find that surprising that the keyword then does not work. And given that `sort=False` is the default, I think we can let `sort=True` sort always?

brycepg:
@jreback 

I've make the requested changes

@jorisvandenbossche 

* Yeah it does look like `concat` is used a lot internally. I'll have a look at each location its called.
* That sounds reasonable

jorisvandenbossche:
> Yeah it does look like concat is used a lot internally. I'll have a look at each location its called.

Ah, yes, that might be needed as well. But what I meant was other keywords of concat itself. Eg does the sort keyword work for both `join='inner'` and `join='outer'` ?

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1160,6 +1160,7 @@ Reshaping
 - Bug in :meth:`DataFrame.astype` where column metadata is lost when converting to categorical or a dictionary of dtypes (:issue:`19920`)
 - Bug in :func:`cut` and :func:`qcut` where timezone information was dropped (:issue:`19872`)
 - Bug in :class:`Series` constructor with a ``dtype=str``, previously raised in some cases (:issue:`19853`)
+- Stop :func:`concat` and ``Dataframe.append`` from sorting columns by default. Use ``sort=True`` to retain old behavior (:issue:`4588`)
```
TomAugspurger: This should be ``sort=True`` by default to preserve backwards compatibility, right?

Or rather, I think the eventual goal is to have ``sort=False`` be the default, so for now it should be

- ``sort=None`` is the default
- If the default is passed, use ``sort=True`` and warn that the default is changing in the futrue
- If True or False, no warning.

jreback: actually this needs a sub-section. this is a rather large change (even if its None by default). highliting it is best. pls show an example of previous and new

pandas/core/frame.py
```diff
@@ -5982,7 +5982,8 @@ def infer(x):
     # ----------------------------------------------------------------------
     # Merging / joining methods
 
-    def append(self, other, ignore_index=False, verify_integrity=False):
+    def append(self, other, ignore_index=False,
+               verify_integrity=False, sort=False):
```
jreback: sort before verify_integrity

pandas/core/frame.py
```diff
@@ -5995,6 +5996,8 @@ def append(self, other, ignore_index=False, verify_integrity=False):
             If True, do not use the index labels.
         verify_integrity : boolean, default False
             If True, raise ValueError on creating index with duplicates.
+        sort: boolean, default False
+            Sort columns if given object doesn't have the same columns
```
jreback: needs a versionadded.

use does not

pandas/core/reshape/concat.py
```diff
@@ -20,7 +20,7 @@
 
 def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
            keys=None, levels=None, names=None, verify_integrity=False,
-           copy=True):
```
jreback: actually move before verify_integrity

pandas/core/reshape/concat.py
```diff
@@ -60,6 +60,8 @@ def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
     verify_integrity : boolean, default False
         Check whether the new concatenated axis contains duplicates. This can
         be very expensive relative to the actual data concatenation
+    sort : boolean, default False
```
jreback: simiar to above

pandas/tests/reshape/test_concat.py
```diff
@@ -2155,3 +2155,24 @@ def test_concat_empty_and_non_empty_series_regression():
...
     tm.assert_series_equal(result, expected)
+
+
+def test_concat_preserve_column_order_differing_columns():
+    # GH 4588 regression test
+    # for new columns in concat
+    dfa = pd.DataFrame(columns=['C', 'A'], data=[[1, 2]])
+    dfb = pd.DataFrame(columns=['C', 'Z'], data=[[5, 6]])
+    result = pd.concat([dfa, dfb])
+    assert result.columns.tolist() == ['C', 'A', 'Z']
```
jreback: create an expected frame and use assert_frame_equal

pandas/tests/reshape/test_concat.py
```diff
@@ -2155,3 +2155,24 @@ def test_concat_empty_and_non_empty_series_regression():
...
+def test_concat_preserve_column_order_uneven_data():
+    # GH 4588 regression test
+    # add to column, concat with uneven data
+    df = pd.DataFrame()
+    df['b'] = [1, 2, 3]
+    df['c'] = [1, 2, 3]
+    df['a'] = [1, 2, 3]
+    df2 = pd.DataFrame({'a': [4, 5]})
+    df3 = pd.concat([df, df2])
+    assert df3.columns.tolist() == ['b', 'c', 'a']
```
jreback: same

pandas/tests/reshape/test_concat.py
```diff
@@ -2155,3 +2155,24 @@ def test_concat_empty_and_non_empty_series_regression():
...
+
+def test_concat_preserve_column_order_uneven_data():
+    # GH 4588 regression test
+    # add to column, concat with uneven data
+    df = pd.DataFrame()
+    df['b'] = [1, 2, 3]
+    df['c'] = [1, 2, 3]
+    df['a'] = [1, 2, 3]
+    df2 = pd.DataFrame({'a': [4, 5]})
+    df3 = pd.concat([df, df2])
```
jreback: use result =

jreback (changes requested):

doc/source/whatsnew/v0.23.0.txt
brycepg: @TomAugspurger
If I do what @jorisvandenbossche suggests then, `sort=True` will not be backwards compatible because it will sort the axes in question regardless of whether the columns are mismatched.
I could have `sort=None` be the default, give a warning and revert to old behavior. In future versions this behavior of only sorting the axes sometimes would not be available because it doesn't make sense and concat could default to `sort=False`

jorisvandenbossche: @brycepg I think we can do both (it might complicate the code a bit, but not too much I think, as in `_get_combined_index` those cases are already handled separately). As @TomAugspurger suggests, the default can be None for now, so we can raise a warning in the appropriate cases:

- when all axes are equal (current case when there is already no sorting): no warning should be raised, as the future default of `sort=False` will not change anything, but add the ability to also sort the index with `sort=True`
- when not all axes are equal (current case that the unwanted sorting happens): issue a warning that in the future this will no longer sort

brycepg:
Sure I’ll try to do it tonight

TomAugspurger:
@brycepg I pushed some changes to your branch.

1. I updated the API to keep this backwards compat, but you'll get a warning.
2. I updated the docs to explain that

Will you have time today to get go through and update the tests to pass `sort=True` where necessary? The build log will have a bunch of warnings when it finishes.

And I didn't address most of @jreback's comments yet.

I'd like to do a release candidate tomorrow, so if you can't get to it today let me know and I'll push more fixes here.

pandas/core/frame.py
TomAugspurger: @jreback why do you want sort before verify_integrity?

TomAugspurger:
Can someone proofread the warning text? Specifically, does the term "non-concatenation axis" make sense? I could also call it the "non-expanding axis", or something else entirely.

TomAugspurger:
Sorry, that was a bad example. join='inner' already preserves the column order of the intersection

```python

In [4]:     df1 = pd.DataFrame({"a": [1, 2], "b": [1, 2], "c": [1, 2]},
   ...:                        columns=['b', 'a', 'c'])
   ...:     df2 = pd.DataFrame({"a": [1, 2], 'b': [3, 4]}, index=[3, 4])
   ...:     result = pd.concat([df1, df2], join='inner')
   ...:
   ...:     result
   ...:
   ...:
Out[4]:
   b  a
0  1  1
1  2  2
3  3  1
4  4  2
```

I'll document that `sort` has no effect for `join='inner'`.

TomAugspurger:
> Do we want to ignore the keyword if all columns match?

This depends a bit on the eventual fate of the `sort` keyword. While we may not ever remove it, since we don't want to introduce a keyword and deprecate it shortly after, we can at least discourage its use once the default has been changed. I could imagine the docstring for `sort` eventually just being "for backwards compatibility. Has no effect."

If people want to sort, they should sort that axis and then reindex before or after the concat.

jorisvandenbossche:
> This depends a bit on the eventual fate of the sort keyword.

What's the problem with keeping the keyword, and let it mean to always sort? (sorry if this makes no sense, it's a while ago I looked at this)

TomAugspurger:
No problem, just not useful I think.

Though passing sort=True and not having it sort, for already aligned, would be strange. I’ll push an update for that later today.

doc/source/merging.rst
```diff
@@ -181,8 +181,14 @@ behavior:
           labels=['df1', 'df4'], vertical=False);
    plt.close('all');
 
-Note that the row indexes have been unioned and sorted. Here is the same thing
-with ``join='inner'``:
+.. versionchanged:: 0.23.0
```
jreback: make this a note (maybe a warning)

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -639,6 +639,34 @@ Returning a ``Series`` allows one to control the exact return structure and colu
 
     df.apply(lambda x: Series([1, 2, 3], index=['D', 'E', 'F']), axis=1)
```
jreback: need a ref here

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -639,6 +639,34 @@ Returning a ``Series`` allows one to control the exact return structure and colu
...
+
+   pd.concat([df1, df2], sort=True)
+
+To accept the future behavior (no sorting), pass ``sort=False``
+
+.. ipython
+
+   pd.concat([df1, df2], sort=False)
+
+Note that this change also applies to :meth:`DataFrame.append`, which has also received a `sort` keyword for controlling this behavior.
```
jreback: use double backticks around sort

pandas/core/indexes/api.py
```diff
@@ -89,13 +103,19 @@ def conv(i):
         index = indexes[0]
         for other in indexes[1:]:
             if not index.equals(other):
+
```
jreback: isn't this already passed in as ``sort=True``?

pandas/core/reshape/concat.py
jreback: pls do this

jreback (changes requested):
would prefer that you add a sort fixture to the top-level conftest and test most things you changed to sort=True to use that fixture. a bit of work, but often this could surface some bugs. Also are there any remaining warnings?

pandas/core/indexes/api.py
TomAugspurger: What do you mean?

doc/source/whatsnew/v0.23.0.txt
TomAugspurger: Single backticks for parameter names.

pandas/core/reshape/concat.py
TomAugspurger: Why make this an API breaking change?

TomAugspurger (commented):
One subtly that's probably unimportant. If you have an Index with incomparable objects (mix of ints and strs) then we'll silently fail to sort. That's consistent with how union works. I think it's uncommon enough that it's not a big deal.

doc/source/whatsnew/v0.23.0.txt
jreback: that doesn't render as nicely, and its inconsistent with the docs themselves. Unless you want to change all of them I would use double-backticks.

pandas/core/frame.py
```diff
@@ -6051,6 +6052,14 @@ def append(self, other, ignore_index=False, verify_integrity=False):
             If True, do not use the index labels.
         verify_integrity : boolean, default False
             If True, raise ValueError on creating index with duplicates.
+        sort : boolean, default None
```
jreback: put before verify_integrity

pandas/core/indexes/api.py
```diff
@@ -31,33 +44,40 @@
...
         for other in indexes[1:]:
             index = index.intersection(other)
-        return index
-    union = _union_indexes(indexes)
-    return _ensure_index(union)
+    else:
+        index = _union_indexes(indexes, sort=sort)
+        index = _ensure_index(index)
+
+    if sort and not index.is_monotonic_increasing:
```
jreback: this check is not needed (that's the first thing sort checks)

pandas/core/indexes/api.py
jreback: I mean this is deeply buried here, why is this showing a user facing warning? IOW what is the user call that triggers this?

pandas/core/reshape/concat.py
jreback: because it more logical.

pandas/core/reshape/concat.py
```diff
@@ -60,6 +60,19 @@ def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
     verify_integrity : boolean, default False
         Check whether the new concatenated axis contains duplicates. This can
         be very expensive relative to the actual data concatenation
+    sort : boolean, default None
```
jreback: same

pandas/tests/reshape/test_concat.py
```diff
@@ -793,7 +800,32 @@ def test_append_records(self):
...
+            # around to get the right stackleve.
+            result = df1.append(df2)
+
+        expected = pd.DataFrame({"b": [1, 2, None, None],
+                                 "a": [1, 2, 1, 2],
+                                 "c": [None, None, 3, 4]},
+                                columns=['a', 'b', 'c'])
+        tm.assert_frame_equal(result, expected)
+
+        # sort=True, the previous behavior
```
jreback: you could just use the fixture and just change the expected if sort is False (like you do elsewhere)

pandas/tests/reshape/test_concat.py
```diff
@@ -2249,3 +2295,109 @@ def test_concat_empty_and_non_empty_series_regression():
...
+def test_concat_aligned_sort():
+    # GH-4588
+    df = pd.DataFrame({"c": [1, 2], "b": [3, 4], 'a': [5, 6]},
+                      columns=['c', 'b', 'a'])
+    result = pd.concat([df, df], sort=True, ignore_index=True)
+    expected = pd.DataFrame({'a': [5, 6, 5, 6], 'b': [3, 4, 3, 4],
+                             'c': [1, 2, 1, 2]},
+                            columns=['a', 'b', 'c'])
+    tm.assert_frame_equal(result, expected)
+
```
jreback: not that I don't mind more tests, but are any of these duplicates of existing ones?

jreback (changes requested):

doc/source/whatsnew/v0.23.0.txt
jorisvandenbossche: Single backticks is the "rule" according to our docstring guidelines, but I agree we probably don't do it consistently following that rule in the narrative docs.

jreback: this is not a doc-string

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -639,6 +639,36 @@ Returning a ``Series`` allows one to control the exact return structure and colu
 
     df.apply(lambda x: Series([1, 2, 3], index=['D', 'E', 'F']), axis=1)
 
+.. _whatsnew_0230.api_breaking.concat:
+
+Concatenation will no longer sort
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In a future version of pandas :func:`pandas.concat` will no longer sort the non-concatenation axis when it is not already aligned.
+The current behavior is the same as the previous (sorting), but now a warning is issued (:issue:`4588`).
```
jorisvandenbossche: I would mention the warning is only issues if the labels are not exactly the same.

pandas/core/reshape/concat.py
TomAugspurger: How so?

I'm OK with breaking API when necessary, but this seems unnecessary.

pandas/core/indexes/api.py
TomAugspurger: > IOW what is the user call that triggers this?

`pd.concat` or `DataFrame.append` with unaligned objects and an unspecified `sort`.

Where else would it go? Before this point we don't know what the indexes to align are, and whether or not they're aligned.

pandas/tests/reshape/test_concat.py
TomAugspurger: Yeah, probably. I'll try to de-duplicate a bit.

TomAugspurger:
Found one more issue in crosstab. Fixing now.

pandas/core/base.py
```diff
@@ -507,7 +507,7 @@ def is_any_frame():
                            for r in compat.itervalues(result))
 
             if isinstance(result, list):
-                return concat(result, keys=keys, axis=1), True
```
jreback: thre are a bunch more concats in this same section. basically they control the resulting ordeing of the aggregation. I guess ``sort=True`` is fine here (for the other cases).

jreback: or maybe these should be ``sort=False``, not sure what is actually affected (as these might be aligned ops already)

pandas/core/groupby/groupby.py
```diff
@@ -1098,7 +1098,8 @@ def reset_identity(values):
                 group_names = self.grouper.names
 
                 result = concat(values, axis=self.axis, keys=group_keys,
-                                levels=group_levels, names=group_names)
+                                levels=group_levels, names=group_names,
```
jreback: same here, there are like 10 calls to concat. I think should be explicit about sort

pandas/core/indexes/api.py
```diff
@@ -89,13 +110,19 @@ def conv(i):
         index = indexes[0]
         for other in indexes[1:]:
             if not index.equals(other):
+
+                if sort is None:
```
jreback: I would move ``_unique_indices`` from a nested function to module level (e.g. same as ``_union_indices`` (and maybe conform the spelling indices / indexes), and simply add a ``sort=`` kwarg (which you are already passing into fast_unique_multiple_lists). Then you can do the warning there. just makes this whole function a bit simpler.

pandas/core/panel.py
```diff
@@ -1499,8 +1499,11 @@ def _extract_axis(self, data, axis=0, intersect=False):
                 raw_lengths.append(v.shape[axis])
 
         if have_frames:
+            # we want the "old" behavior here, of sorting only
+            # 1. we're doing a union (intersect=False)
+            # 2. the indices are not aligned.
             index = _get_objs_combined_axis(data.values(), axis=axis,
-                                            intersect=intersect)
```
jreback: so this always warns? shouldn't this be ``sort=False``?

jreback (changes requested):

TomAugspurger:
K, I'll try to go through all the internal calls to concat and set explicitly.

One thing that's made this difficult is that we don't really have the old option of *sort only if not aligned* anymore. So just setting `sort=True` or `sort=False` won't necessarily reproduce the old output. Maybe all of these will be OK though.

TomAugspurger:
I'm going to push this to 0.23.1. I'm not confident enough that we have complete test coverage for all the cases of concat used internally.

I can pick it up after PyCon.

TomAugspurger:
Yeah. The concern is about internal calls to concat.

For example, `Panel({b: df, a: df})` was just broken on this branch for a
while with certain conditions. None of the constructor tests caught that,
it was just in a pytables test that it was caught. Though perhaps panels
isn't the best example, not sure how coverage is on that in general.

jorisvandenbossche:
In what way was it broken? (there should for now mainly be a warning and new keyword, but not really change in default behaviour?)

>  I'm not confident enough that we have complete test coverage for all the cases of concat used internally.

I can't really say anything founded, as I didn't look enough into detail to what tests have broken throughout implementing this and what you needed to add, but since our tests are now passing, my gut feeling would say: let's use the rc period to get more real-world testing on that .. ;)

TomAugspurger:
> In what way was it broken? (there should for now mainly be a warning and new keyword, but not really change in default behaviour?)

I don't recall the exact circumstances, but a MultiIndex was (or wasn't?) being sorting.

If you're OK with merging as is, then I can commit to adding more tests (and fixing bugs) between now and Friday. I just don't think it should hold up the release.

jorisvandenbossche:
As I said, difficult to say myself, but I am OK with relying on your judgement here. 
(and I certainly approve the gist of the PR)

TomAugspurger:
Alright, let's do it.

https://github.com/pandas-dev/pandas/issues/20909 for the followup.

Issue #20592, opened 2018-04-03
assign None to Series got ValueError


hack4code:
```python
import pandas as pd

s = pd.Series({'aa': 1})
s['aa'] = None
```
#### Problem description

got exception:
ValueError: ['a' 'a'] not contained in the index

#### Output of ``pd.show_versions()``

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Darwin
OS-release: 17.5.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.21.0
pytest: None
pip: 9.0.1
setuptools: 28.8.0
Cython: None
numpy: 1.13.1
scipy: 0.19.1
pyarrow: None
xarray: None
IPython: 6.0.0
sphinx: None
patsy: None
dateutil: 2.6.1
pytz: 2017.3
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: 0.999999999
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

chris-b1:
Strange, somewhere in the indexing code 'aa' gets treated as a iterable.  PR would be welcome.  Note that it would be more typical to use `np.nan` as a missing value, which is handled correctly (`s['aa'] = np.nan`)

fredrik-1:
The problem is that `_set_with_engine` throws a typeError, ` _set_with` is used instead which for some reason change the key to a list with `key = list(key)` and list('aa')=['a','a']

So one solution might be to change `key=list(key)` with 
```
if isinstance(key, str):
     key=[key]
else:
     key = list(key)
```
I don't know if this works in both python 2 and 3

JustinZhengBC:
Solved in #23495

PR #20615, opened 2018-04-05
BUG: Avoid splitting string with list() (#20592)

fredrik-1:
Solves #20592 and I doubt that there are any cases when a key string should be split by `list()`

### DIFF
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -951,7 +951,10 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
             if not isinstance(key, (list, Series, np.ndarray, Series)):
                 try:
-                    key = list(key)
+                    if isinstance(key, compat.string_types):
+                        key = [key]
+                    else:
+                        key = list(key)
                 except Exception:
                     key = [key]
 


WillAyd:
This needs tests

TomAugspurger:
And a release note: http://pandas.pydata.org/pandas-docs/stable/contributing.html#documenting-your-code

pandas/core/series.py
```diff
@@ -951,7 +951,10 @@ def _set_with(self, key, value):
 
             if not isinstance(key, (list, Series, np.ndarray, Series)):
```
jreback: looks like extra Series listed here.

pandas/core/series.py
```diff
@@ -951,7 +951,10 @@ def _set_with(self, key, value):
...
                 try:
-                    key = list(key)
```
jreback: let's break this if up to specificy target a string e.g.

```
if isinstance(key, compat.string_types):
 ......
elif not isinstance(key, (list, Series, np.ndarray)):
    ....
```

jreback (changes requested):

jreback:
closing as stale, though would certainly take a PR here

PR #20618, opened 2018-04-05
DOC: update the str_replace docstring

gsganden:
Remove claim in `str_replace` docstring of equivalence to base Python `str.replace` -- the former does regex parsing while the latter does not.

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -439,8 +439,7 @@ def str_endswith(arr, pat, na=np.nan):
 def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
     r"""
     Replace occurrences of pattern/regex in the Series/Index with
-    some other string. Equivalent to :meth:`str.replace` or
-    :func:`re.sub`.
+    some other string. Equivalent to :func:`re.sub`.
 
     Parameters
     ----------


gfyoung:
Aren't they equivalent if `regex=False` ?

gsganden:
If nothing else, the relevance of `regex=False` should be explicit. E.g.

"""
Replace occurrences of pattern/regex in the Series/Index with some other string. Equivalent to :func:`re.sub` when regex=True, or :meth:`str.replace` when `regex=False`.
"""

pandas/core/strings.py
```diff
@@ -439,8 +439,7 @@ def str_endswith(arr, pat, na=np.nan):
 def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
     r"""
     Replace occurrences of pattern/regex in the Series/Index with
-    some other string. Equivalent to :meth:`str.replace` or
-    :func:`re.sub`.
+    some other string. Equivalent to :func:`re.sub`.
```
datapythonista: @gsganden, somehow unrelated to your change, but the docstring should start with a one-liner (see [1]). Do you think you can summarize what the method does, and then add the detailed explanation you're proposing in the extended summary?

Also, it'd be great if you can fix the rest of the docstring based on the standard defined in [1].


1. http://pandas-docs.github.io/pandas-docs-travis/contributing_docstring.html#section-1-short-summary

datapythonista:
@gsganden feel free to open a new PR with your last suggestion (being explicit about when it's equivalent to `str.replace` and when to `re.sub`). Closing this one, as I think it's helpful for users to know it's equivalent to `str.replace`.

PR #20623, opened 2018-04-06
BUG: excel index label can be an array

### LINKED ISSUE #20622: ValueError: The truth value of an array with more than one element is ambiguous.
```python
            if (self.index_label and
                    isinstance(self.index_label, (list, tuple, np.ndarray,
                                                  Index))):
```
#### Problem description

self.index_label can be an array, list, Index or tuple, but boolean values of arrays are ambiguous. Should be changed to 

            if (self.index_label is not None and
                    isinstance(self.index_label, (list, tuple, np.ndarray,
                                                  Index))):

#### Expected Output

Formatted excel cell

#### Output of ``pandas/io/formats/excel.get_formatted_cells``

<details>
 in get_formatted_cells
    yield cell
 in _format_hierarchical_rows
    if (self.index_label and isinstance(self.index_label, (list, tuple, np.ndarray, Index))):
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

</details>

perl-girl:
Fixes "ValueError: The truth value of an array with more than one element is ambiguous." where self.index_label could be a list, tuple np.ndarray, or Index type

closes #20622

### DIFF
--- a/pandas/io/formats/excel.py
+++ b/pandas/io/formats/excel.py
@@ -502,7 +502,8 @@ class ExcelFormatter(object):
                                                   Index))):
                 index_label = self.index_label[0]
             # if string good to go
-            elif self.index_label is not None and isinstance(self.index_label, str):
+            elif self.index_label is not None and 
+                    isinstance(self.index_label, str):
                 index_label = self.index_label
             else:
                 index_label = self.df.index.names[0]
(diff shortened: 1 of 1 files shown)


TomAugspurger:
Could you add tests (that fail without the change) and a release note? http://pandas.pydata.org/pandas-docs/stable/contributing.html

saxaholic:
I will try if I have time.

jreback:
closing as stale

PR #20624, opened 2018-04-06
TST: Parametrized index tests

WillAyd:
I came across this module on another change and noticed that a lot of the tests could really use refactoring. There's a ton more to be done with this module but submitting as is so it doesn't get too large.

Can either add other commits on top of this or have this merged (assuming looks OK) and continue down the module in additional PR(s)


pandas/tests/indexes/test_base.py
```diff
@@ -125,88 +126,75 @@ def test_construction_list_tuples_nan(self, na_value, vtype):
...
+            result = pd.Index(idx)
 
-    def test_constructor_from_series_timedelta(self):
-        idx = pd.timedelta_range('1 days', freq='D', periods=3)
-        result = pd.Index(pd.Series(idx))
         tm.assert_index_equal(result, idx)
-
-    def test_constructor_from_series_period(self):
-        idx = pd.period_range('2015-01-01', freq='D', periods=3)
+        if has_tz:
```
WillAyd: I suppose this could also leverage hasattr instead but I felt the explicitness of the parameter is more useful

pandas/tests/indexes/test_base.py
```diff
@@ -125,88 +126,75 @@ def test_construction_list_tuples_nan(self, na_value, vtype):
...
         s = Series([Timestamp('20110101'), Timestamp('20120101'),
                     Timestamp('20130101')])
-        result = Index(s)
-        tm.assert_index_equal(result, expected)
-        result = DatetimeIndex(s)
+        result = klass(s)
         tm.assert_index_equal(result, expected)
 
+    @pytest.mark.parametrize("klass", [pd.Series, pd.DataFrame])
+    def test_constructor_from_series_freq(self, klass):
```
WillAyd: This could arguably be split into two separate tests given the size of the conditional. Open to suggestions

pandas/tests/indexes/test_base.py
```diff
@@ -237,61 +223,63 @@ def test_constructor_int_dtype_float(self, dtype):
         result = Index([0., 1., 2., 3.], dtype=dtype)
         tm.assert_index_equal(result, expected)
 
-    def test_constructor_int_dtype_nan(self):
+    @pytest.mark.parametrize("dtype,klass_or_raises", [
```
WillAyd: Not sure how we feel about mixing types in `klass_or_raises` - could also be split into a boolean flag for raises and the klass, though the test would typically just use one or the other

pandas/tests/indexes/test_base.py
```diff
@@ -237,61 +223,63 @@ def test_constructor_int_dtype_float(self, dtype):
...
-        for data in [[np.nan, np.timedelta64('nat')],
-                     [np.timedelta64('nat'), np.nan],
-                     [pd.NaT, np.timedelta64('nat')],
-                     [np.timedelta64('nat'), pd.NaT]]:
-            tm.assert_index_equal(Index(data), exp)
-            tm.assert_index_equal(Index(np.array(data, dtype=object)), exp)
+        tm.assert_index_equal(Index(data), exp)
+        tm.assert_index_equal(Index(np.array(data, dtype=object)), exp)
 
+    @pytest.mark.parametrize("swap_objs", [True, False])
```
WillAyd: Kind of a strange parameter but just emulating the existing test - one assertion is done with the datetime at the initial position and another assertion is done with the timedelta at the initial position

pandas/tests/indexes/test_base.py
```diff
@@ -125,88 +126,75 @@ def test_construction_list_tuples_nan(self, na_value, vtype):
...
-        tm.assert_series_equal(df['date'], exp)
+        if klass is pd.Series:
+            s = Series(pd.to_datetime(dts))
+            result = DatetimeIndex(s, freq='MS')
+        else:
+            df = pd.DataFrame(np.random.rand(5, 3))
+            df['date'] = dts
+            result = DatetimeIndex(df['date'], freq='MS')
+            assert df['date'].dtype == object
+            expected.name = 'date'
```
jreback: put some blank lines in between text like this (so its easily readable)

pandas/tests/indexes/test_base.py
```diff
@@ -217,11 +205,9 @@ def __init__(self, array):
             def __array__(self, dtype=None):
                 return self.array
 
-        for array in [np.arange(5), np.array(['a', 'b', 'c']),
-                      date_range('2000-01-01', periods=3).values]:
-            expected = pd.Index(array)
-            result = pd.Index(ArrayLike(array))
-            tm.assert_index_equal(result, expected)
+        expected = pd.Index(array)
```
jreback: can you audit for whether we use ``pd.Index`` (and similar) or ``Index`` (we are not generally consistent), like to be consistent within in a module

pandas/tests/indexes/test_base.py
```diff
@@ -302,138 +290,138 @@ def test_index_ctor_infer_periodindex(self):
...
 
-        idx = Index([1.1, np.nan, 2.2, 3.0], name='float')
-        result = idx._simple_new(idx, 'float')
-        tm.assert_index_equal(result, idx)
+    @pytest.mark.parametrize("vals", [
+        [1, 2, 3], np.array([1, 2, 3]), np.array([1, 2, 3], dtype=int),
+        # below should coerce
+        [1., 2., 3.], np.array([1., 2., 3.], dtype=float)
+    ])
+    def test_constructor_dtypes_to_int64(self, vals):
```
WillAyd: This is similar to a lot of the tests below and could arguably be built with parameters of "vals,dtype,klass" but I felt it was cleaner and less repetition to just break the tests by dtype

pandas/tests/indexes/test_base.py
jreback: leave these as 2 separate. generally having parameterize where one case raises is not great (you can name the same however, with _errors on one of the test names)

pandas/tests/indexes/test_base.py
```diff
@@ -237,61 +223,63 @@ def test_constructor_int_dtype_float(self, dtype):
...
-
-        exp = pd.DatetimeIndex([pd.NaT, pd.NaT])
-        assert exp.dtype == 'datetime64[ns]'
-        tm.assert_index_equal(Index([pd.NaT, pd.NaT]), exp)
-        tm.assert_index_equal(Index(np.array([pd.NaT, pd.NaT])), exp)
-
+        na_list = [na_val, na_val]
+        exp = klass(na_list)
+        assert exp.dtype == dtype
+        tm.assert_index_equal(Index(na_list), exp)
```
jreback: try to use 
result =

pandas/tests/indexes/test_base.py
```diff
@@ -237,61 +223,63 @@ def test_constructor_int_dtype_float(self, dtype):
...
+        tm.assert_index_equal(Index(np.array(na_list)), exp)
+
+    @pytest.mark.parametrize("data", [
+        [pd.NaT, np.nan], [np.nan, pd.NaT], [np.nan, np.datetime64('nat')],
```
jreback: we should make this into a fixture (e.g. nulls_fixture) (if you want to add to conftest). note we should change globally (but can do that in another PR)

pandas/tests/indexes/test_base.py
```diff
@@ -302,138 +290,138 @@ def test_index_ctor_infer_periodindex(self):
...
+        [datetime(2011, 1, 1), datetime(2011, 1, 2)]
+    ])
+    def test_constructor_dtypes_to_categorical(self, vals):
+        idx = Index(vals, dtype='category')
+        assert isinstance(idx, CategoricalIndex)
 
-        for idx in [Index(np.array([np.timedelta64(1, 'D'), np.timedelta64(
-                1, 'D')])), Index([timedelta(1), timedelta(1)])]:
-            assert isinstance(idx, TimedeltaIndex)
+    @pytest.mark.parametrize("cast_idx", [True, False])
```
WillAyd: I've used this in a few tests and am not a huge fan of it, but I wasn't sure of a better way to emulate the existing behavior where `dtype=object` is in the `Index` constructor in half of the tests

pandas/tests/indexes/test_base.py
```diff
@@ -499,25 +483,25 @@ def test_insert(self):
         null_index = Index([])
         tm.assert_index_equal(Index(['a']), null_index.insert(0, 'a'))
 
+    @pytest.mark.parametrize("na_val", [np.nan, pd.NaT, None])
```
jreback: e.g. you can use the nulls_fixture from above

pandas/tests/indexes/test_base.py
```diff
@@ -302,138 +290,138 @@ def test_index_ctor_infer_periodindex(self):
...
-                        pd.TimedeltaIndex(list(values), dtype=dtype)]:
-                tm.assert_index_equal(res, idx)
+        res = klass(list(values), dtype=dtype)
+        tm.assert_index_equal(res, idx)
 
-    def test_constructor_empty(self):
+    def test_constructor_empty_gen(self):
         skip_index_keys = ["repeats", "periodIndex", "rangeIndex",
                            "tuples"]
         for key, idx in self.generate_index_types(skip_index_keys):
```
WillAyd: In general some of these magic class methods could probably be replaced with pytest functionality. I haven't looked in too much detail just yet but was planning on reviewing after giving the module a first pass at parametrization without changing class behavior

pandas/tests/indexes/test_base.py
```diff
@@ -499,25 +483,25 @@ def test_insert(self):
...
 
-        expected = Index(['a', 'b', 'c'], name='idx')
-        result = idx.delete(-1)
-        tm.assert_index_equal(result, expected)
-        assert result.name == expected.name
+    @pytest.mark.parametrize("pos,exp", [
+        (0, Index(['b', 'c', 'd'], name='idx')),
+        (-1, Index(['a', 'b', 'c'], name='idx'))
+    ])
+    def test_delete(self, pos, exp):
```
jreback: note that some of the tests here should really be parametreized on index types themselves, there is an open issue on that (eg. these set type ops, test_delete and such), are generally tested by the subclasses, but we need a more general cleanup on that

jreback (changes requested):
generally looks good.

pandas/tests/indexes/test_base.py
jreback: haha just made this comment

pandas/tests/indexes/test_base.py
jreback: yes see my comments below

pandas/tests/indexes/test_base.py
WillAyd:
```sh
grep " Index(" -r pandas/tests --include="*.py" | wc -l
     891
grep " pd.Index(" -r pandas/tests --include="*.py" | wc -l
     264
```

May be a more robust way of doing it but assuming directionally accurate first is more widely used. Particular to this module I see numbers of 175 and 48, respectively.

I personally prefer `pd.Index` for explicitness but am fine to change to `Index` for module-consistency. lmk

WillAyd (commented):
I've deferred making any changes around the `Index` vs `pd.Index` conversation. I'm assuming that we can either do that as one cleanup at the end for the entire module or bundle into a separate change somewhere else but can also include incrementally in this PR if you'd prefer

pandas/tests/indexes/test_base.py
```diff
@@ -240,58 +228,59 @@ def test_constructor_int_dtype_float(self, dtype):
...
-
-        exp = pd.DatetimeIndex([pd.NaT, pd.NaT])
-        assert exp.dtype == 'datetime64[ns]'
-        tm.assert_index_equal(Index([pd.NaT, pd.NaT]), exp)
-        tm.assert_index_equal(Index(np.array([pd.NaT, pd.NaT])), exp)
-
-        exp = pd.DatetimeIndex([pd.NaT, pd.NaT])
-        assert exp.dtype == 'datetime64[ns]'
-
-        for data in [[pd.NaT, np.nan], [np.nan, pd.NaT],
```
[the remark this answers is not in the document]
jreback: hmm, I don't think so, this is a construction with mixed types of nulls.

pandas/conftest.py
```diff
@@ -87,3 +87,11 @@ def join_type(request):
     Fixture for trying all types of join operations
     """
     return request.param
+
+
+@pytest.fixture(params=[numpy.nan, pandas.NaT])
```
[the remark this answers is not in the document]
jreback: can you use np and pd

pandas/tests/indexes/test_base.py
jreback: yes woudl do that

pandas/tests/indexes/test_base.py
```diff
@@ -125,88 +126,77 @@ def test_construction_list_tuples_nan(self, na_value, vtype):
...
-
-    def test_constructor_from_series_datetimetz(self):
-        idx = pd.date_range('2015-01-01 10:00', freq='D', periods=3,
-                            tz='US/Eastern')
-        result = pd.Index(pd.Series(idx))
-        tm.assert_index_equal(result, idx)
-        assert result.tz == idx.tz
+    @pytest.mark.parametrize("cast_as_obj", [True, False])
+    @pytest.mark.parametrize("idx,has_tz", [
+        (pd.date_range('2015-01-01 10:00', freq='D', periods=3,
```
jreback: would prob drop the has_tz, then can make the idx a fixture above (in the test conditional you can directly test if isnstance DTI & tz is not None); then you can add a DTI w/o a tz here as well.

pandas/tests/indexes/test_base.py
```diff
@@ -240,58 +228,59 @@ def test_constructor_int_dtype_float(self, dtype):
...
+        exp = klass(na_list)
+        assert exp.dtype == dtype
+
+        result = Index(na_list)
+        tm.assert_index_equal(result, exp)
+
+        result = Index(np.array(na_list))
+        tm.assert_index_equal(result, exp)
 
+    @pytest.mark.parametrize("pos", [0, 1])
```
jreback: these tests that are strictly for datetime like things can prob just move to a new file: ``test_datetimelike.py`` (IOW tests that test *all* of DTI,TDI,PI but all together). can be future PR (or here).

IOW test_base should be only non-datetimelike tests.

pandas/tests/indexes/test_base.py
```diff
@@ -302,138 +291,138 @@ def test_index_ctor_infer_periodindex(self):
...
+            assert isinstance(idx, Index)
+            assert idx.dtype == object
+        else:
+            idx = Index(vals)
+            assert isinstance(idx, TimedeltaIndex)
 
-        for tz in [None, 'UTC', 'US/Eastern', 'Asia/Tokyo']:
-            idx = pd.date_range('2011-01-01', periods=5, tz=tz)
-            dtype = idx.dtype
+    @pytest.mark.parametrize("tz", [
```
jreback: move to pandas/tests/indexes/datetimes/test_timezones.py (might be a duplicate as well)

pandas/tests/indexes/test_base.py
```diff
@@ -302,138 +291,138 @@ def test_index_ctor_infer_periodindex(self):
...
+        res = klass(list(values), tz=tz)
+        tm.assert_index_equal(res, idx)
 
-    def test_constructor_dtypes_timedelta(self):
+        res = klass(list(values), dtype=dtype)
+        tm.assert_index_equal(res, idx)
 
+    @pytest.mark.parametrize("attr", ['values', 'asi8'])
+    @pytest.mark.parametrize("klass", [pd.Index, pd.TimedeltaIndex])
+    def test_constructor_dtypes_timedelta(self, attr, klass):
```
jreback: move this to pandas/tests/indexes/timedeltas/test_construction (maybe duplicate)

pandas/tests/indexes/test_base.py
jreback: note there is already a fixture of ``indices`` which works, so have 3 methods of specifying things ATM:

- ``indices`` fixture
- ``generate_index_types(...)``
- ``self.indices``

needs to clean this and just make fixtures that we can use generally (in conftest) with docs

pandas/tests/indexes/test_base.py
```diff
@@ -302,138 +291,138 @@ def test_index_ctor_infer_periodindex(self):
...
-
-        empty = RangeIndex(step=1)
-        assert isinstance(empty, pd.RangeIndex)
-        assert not len(empty)
-
-        empty = MultiIndex(levels=[[1, 2], ['blue', 'red']],
-                           labels=[[], []])
-        assert isinstance(empty, MultiIndex)
+    @pytest.mark.parametrize("empty,klass", [
+        (PeriodIndex([], freq='B'), PeriodIndex),
```
jreback: might be duplicated elsewhere. am iffy on where to put things like this. maybe worth extracting things from this file and makign a ``test_construction`` for things like this

pandas/tests/indexes/test_base.py
jreback: I *think* should be including ``None`` because I think we *do* convert it for all but ``object`` dtypes (where it is left alone). so maybe need 2 fixtures. might be tricky to do this in a general way.

jreback (changes requested):
this is a messy file. prob need to do multiple PR's to get this really nice.

WillAyd:
Most but not all edits made. I avoided anything that requires moving things out of the module, if only because my suggested plan of attack would be:

 1. Parametrize the existing module where appropriate (this covers ~650 of ~2400 lines, so maybe 3 or 4 more similar PRs)
 2. Replace any class methods / attributes with fixtures where appropriate
 3. Move tests into new modules / sub-modules

Open to suggestions on approach.

pandas/conftest.py
```diff
@@ -89,6 +89,14 @@ def join_type(request):
     return request.param
 
 
+@pytest.fixture(params=[None, np.nan, pd.NaT])
```
jbrockmendel: This is a good idea.  Some more for the list: `np.timedelta64('NaT'), np.datetime64('NaT'), float('nan'), np.float('NaN')`.

Note that `np.float('NaN')` does not return `np.nan` (i.e. does not behave like `pd.NaT`), so this would catch any occurrences in pandas of e.g `if foo is np.nan`

jreback (approved):

jreback:
@WillAyd can you rebase.

can you expand the nulls this as well? to as much as you can while still having this pass?

WillAyd:
Added the suggestions from @jorisvandenbossche with the exception of `np.timedelta64('NaT')` and `np.datetime64('NaT')` as they caused a failure with the below test. 

https://github.com/WillAyd/pandas/blob/6e46b57bc99e70b96eb6de448ab12ff7930a2d10/pandas/tests/indexes/test_base.py#L272

The test is parametrized for a `DatetimeIndex` and a `TimedeltaIndex` and the failure occurs when you try to construct from a list that contains one instance of both (it returns an object-typed Index instead of one or the other). 

I think this is a corner case for this test so we could definitely still add to the fixture and have the test be responsible for skipping where appropriate, but for now I've just kept out of the fixture

jreback:
thanks @WillAyd 

if you can create an issue for extending the nulls fixture (and issue around that), and an issue about cleaning more testing things (e.g. comments above).

PR #20630, opened 2018-04-07
DOC: Updated the docstring of pandas.Series.str.get

pranavsuri:
an unrelated PR):

```
###################### Docstring (pandas.Series.str.get)  ######################

Extract element for each component.

Extract element from lists, tuples, or strings in each element in the
Series/Index.

Parameters
----------
i : int
    Integer index (location).

Returns
-------
items : Series/Index of objects

Examples
--------
>>> x = pd.Series(["String", (1, 2, 4), ["a", "b", "c"]])
>>> x
0       String
1    (1, 2, 4)
2    [a, b, c]
dtype: object

>>> x.str.get(1)
0      t
1      2
2      b
dtype: object

Errors found:
	See Also section not found
```

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1651,17 +1651,34 @@ def str_translate(arr, table, deletechars=None):
 
 def str_get(arr, i):
     """
+    Extract element for each component.
+
     Extract element from lists, tuples, or strings in each element in the
     Series/Index.
 
     Parameters
     ----------
     i : int
-        Integer index (location)
+        Integer index (location).
 
     Returns
     -------
     items : Series/Index of objects
+
+    Examples
+    --------
+    >>> x = pd.Series(["String", (1, 2, 4), ["a", "b", "c"]])
+    >>> x
+    0       String
+    1    (1, 2, 4)
+    2    [a, b, c]
+    dtype: object
+
+    >>> x.str.get(1)
+    0      t
+    1      2
+    2      b
+    dtype: object
     """
     f = lambda x: x[i] if len(x) > i >= -len(x) else np.nan
     return _na_map(f, arr)


pandas/core/strings.py
```diff
@@ -1651,17 +1651,34 @@ def str_translate(arr, table, deletechars=None):
...
-        Integer index (location)
+        Integer index (location).
 
     Returns
     -------
     items : Series/Index of objects
+
+    Examples
+    --------
+    >>> x = pd.Series(["String", (1, 2, 4), ["a", "b", "c"]])
```
datapythonista: The standard way is to name the Series `s` and not `x`.

pandas/core/strings.py
```diff
@@ -1651,17 +1651,34 @@ def str_translate(arr, table, deletechars=None):
...
+    >>> x
+    0       String
+    1    (1, 2, 4)
+    2    [a, b, c]
+    dtype: object
```
datapythonista: Can you add a numerical value, to show what happens.

pandas/core/strings.py
```diff
@@ -1651,17 +1651,34 @@ def str_translate(arr, table, deletechars=None):
...
+
+    >>> x.str.get(1)
+    0      t
+    1      2
+    2      b
+    dtype: object
```
datapythonista: May be a example with a negative value too?

pandas/core/strings.py
```diff
@@ -1651,17 +1651,34 @@ def str_translate(arr, table, deletechars=None):
...
+    Extract element for each component.
+
     Extract element from lists, tuples, or strings in each element in the
     Series/Index.
 
     Parameters
     ----------
     i : int
-        Integer index (location)
+        Integer index (location).
```
datapythonista: I find this not very clear. May be something like "Position of the element to extract"?

pandas/core/strings.py
```diff
@@ -1651,17 +1651,34 @@ def str_translate(arr, table, deletechars=None):
 
 def str_get(arr, i):
     """
+    Extract element for each component.
```
datapythonista: That is not very clear. May be "Extract element from the specified position" or "Extract element from the specified position in a list or string" if not too long.

pandas/core/strings.py
pranavsuri: I see. I'd change this.

pandas/core/strings.py
pranavsuri: Since `get` is a method for strings, I didn't add a numerical value (using one returns 'NaN'). However, it can be added it if you say.

pandas/core/strings.py
pranavsuri: Okay. Would make this change.

pandas/core/strings.py
pranavsuri: Okay. How about, "Extract element for each component from the specified position." or "Extract the element from the specified position for all elements in Series."; something along those lines?

pranavsuri:
Hello! I accidentally deleted the original fork from which I had created this pull-request. I've generated a new request at – https://github.com/pandas-dev/pandas/pull/20667 with the mentioned changes. I'm sorry for the trouble.

PR #20634, opened 2018-04-08
update the pandas.Series.str.repeat docstring

manojpandey:
an unrelated PR):

```
##################### Docstring (pandas.Series.str.repeat) #####################

Duplicate each string repeated by indicated number of times.

Parameters
----------
repeats : int or array
    Same value for all (int) or different value per (array).

Returns
-------
repeated : Series/Index of objects
    Same type as the original object

Examples
--------
>>> s = pd.Series(['a', 'b', 'c', 'd', 'e'])

Using same value for all:

>>> s.str.repeat(4)
0    aaaa
1    bbbb
2    cccc
3    dddd
4    eeee
dtype: object

Using different value per element:

>>> s.str.repeat([3, 2, 5, 1, 4])
0      aaa
1       bb
2    ccccc
3        d
4     eeee
dtype: object

Passing zero or negative integer will return an empty string

>>> s.str.repeat([0, 0, -2, -1, 0])
0
1
2
3
4
dtype: object

Notes
--------
A passed value of zero or negative integer will return an empty string.

See also
--------
numpy.ndarray.repeat: Repeat elements of an array.

Errors found:
	No extended summary found
```

explicitly.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -678,17 +678,45 @@ def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
 
 def str_repeat(arr, repeats):
     """
-    Duplicate each string in the Series/Index by indicated number
-    of times.
+    Duplicate each string repeated by indicated number of times.
+
+    Duplicate each string in the Series/Index by indicated number of times.
+    A passed value of zero or negative integer will return an empty string.
 
     Parameters
     ----------
-    repeats : int or array
-        Same value for all (int) or different value per (array)
+    repeats : int or array-like
+        Same value for all (int) or different value per (array).
 
     Returns
     -------
-    repeated : Series/Index of objects
+    Series or Index
+        Series or Index of repeated string objects specified by
+        input parameter repeats.
+
+    Examples
+    --------
+    >>> s = pd.Series(['a', 'b', 'c', 'd', 'e'])
+
+    Using same value for all:
+
+    >>> s.str.repeat(repeats=4)
+    0    aaaa
+    1    bbbb
+    2    cccc
+    3    dddd
+    4    eeee
+    dtype: object
+
+    Using different value per element:
+
+    >>> s.str.repeat(repeats=[-2, -1, 0, 1, 2])
+    0
+    1
+    2
+    3     d
+    4    ee
+    dtype: object
     """
     if is_scalar(repeats):
 


pandas/core/strings.py
```diff
@@ -594,17 +594,59 @@ def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
...
+    4
+    dtype: object
+
+    Notes
+    --------
+    A passed value of zero or negative integer will return an empty string.
+
+    See also
+    --------
+    numpy.ndarray.repeat: Repeat elements of an array.
```
datapythonista: `Notes` and `See Also` goes before examples. But in this case, I personally wouldn't have any. The comment in the notes could go in the extended summary, as well as an explanation that the value can be a literal or a list of the same size of the elements.

For the see also, I don't think `ndarray.repeat` is actually related, even if they repeat things. The user cases are unrelated. May be we could have `str.len` as it's slightly related. In that case `See Also` should have a capital `A`, and also should be placed before the examples.

pandas/core/strings.py
```diff
@@ -594,17 +594,59 @@ def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
...
+
+    Passing zero or negative integer will return an empty string
+
+    >>> s.str.repeat([0, 0, -2, -1, 0])
+    0
+    1
+    2
+    3
+    4
+    dtype: object
```
datapythonista: I think this example could be merged with the previous, if you have a list with the positives and negative together (e.g. `s.str.repeat([-3, 0, 1, 3, 5])`)

pandas/core/strings.py
```diff
@@ -594,17 +594,59 @@ def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
 
 def str_repeat(arr, repeats):
     """
-    Duplicate each string in the Series/Index by indicated number
-    of times.
+    Duplicate each string repeated by indicated number of times.
 
     Parameters
     ----------
     repeats : int or array
```
datapythonista: `int or array-like` is more standard

pandas/core/strings.py
```diff
@@ -594,17 +594,59 @@ def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
...
-        Same value for all (int) or different value per (array)
+        Same value for all (int) or different value per (array).
 
     Returns
     -------
     repeated : Series/Index of objects
```
datapythonista: You can get rid of `repeated` and simply leave `Series or Index`.

datapythonista (commented):

pandas/core/strings.py
manojpandey: Sure - I'll merge with the previous example 👍

datapythonista:
@manojpandey do you have time to make the changes based on the review?

manojpandey:
Hey @datapythonista - surely. I totally missed this. Thanks for reminding :) Let me make the changes!

manojpandey:
Yes, I'm updating this. Thanks for understanding. I should have done it wayy before :(

Let me send the PR this Mon/Tue!

manojpandey:
I just pushed all the changes requested. Please have a look when you get a moment. Thanks, and apologies again for the delay.

datapythonista:
Thanks for the update @manojpandey , seems like @JesperDramsch has been working on the same changes in #22571. Can the two of you discuss which PR makes more sense to keep? Sorry I've been encouraging too much to work on the `Series.str` methods, and we ended up with overlapping work.

JesperDramsch:
Haha, the sprint-prophecy has been fulfilled.

manojpandey:
Hi, I feel that this current PR has better examples than those of #22571 - and if no (or minimum) changes are required, can we move forward to merge this and close it? @datapythonista 

@JesperDramsch What are your thoughts?

Thanks!

JesperDramsch:
Hej @manojpandey, I think it's great you document the effect of repeat values <1. I wonder why you think these are significantly better, but personally, I don't have too many stakes in my PR, so it's fine if we close mine and you update yours.

Things I see that should be done here are:

- Mention the keyword in the call
- Update the Returns string
- Remove the word array and replace it with sequence to make it package agnostic

You have been working at this much longer, so feel free to take it to the end.

manojpandey:
@JesperDramsch Thanks for your feedback on this, really appreciated - and I've incorporated the changes. I updated the `Returns` string part from your PR, thanks for that!

@datapythonista can you see if everything is alright now?

pandas/core/strings.py
```diff
@@ -678,17 +678,45 @@ def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
 
 def str_repeat(arr, repeats):
     """
-    Duplicate each string in the Series/Index by indicated number
-    of times.
+    Duplicate each string repeated by indicated number of times.
+
+    Duplicate each string in the Series/Index by indicated number of times.
```
datapythonista: This seems redundant, can we get rid of it?

pandas/core/strings.py
```diff
@@ -678,17 +678,45 @@ def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
...
+    A passed value of zero or negative integer will return an empty string.
 
     Parameters
     ----------
-    repeats : int or array
-        Same value for all (int) or different value per (array)
+    repeats : int or array-like
```
datapythonista: can we use `sequence` instead of `array-like` (we use `array-like` to refer to objects that follow numpy ndarray api)

pandas/core/strings.py
```diff
@@ -678,17 +678,45 @@ def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
...
     Returns
     -------
-    repeated : Series/Index of objects
+    Series or Index
+        Series or Index of repeated string objects specified by
+        input parameter repeats.
+
+    Examples
+    --------
+    >>> s = pd.Series(['a', 'b', 'c', 'd', 'e'])
```
datapythonista: I think having 5 cases that illustrate exactly the same is not very useful. Can we have:
- A single letter (may be two or three, for the next example to show different number of repetitions)
- A word
- A `NaN`
- A number

Can you also show the value of `s`, so it's easier to compare with the result of `str.repeat`

datapythonista (commented):
Added some comments.

manojpandey:
Sure, updating with the changes requested.

datapythonista:
Closing in favor of #22571

Thanks @manojpandey for the work on this. Feel free to make improvements to the docstring in a new PR.

PR #20645, opened 2018-04-10
MAINT: Remove console encode

### LINKED ISSUE #20616: `console_encode` doesn't exist
https://github.com/pandas-dev/pandas/blob/6d610a4d9393c0d0335267dc3252ccabb9e51e43/pandas/io/formats/printing.py#L86

refers to a `console_encoding` util that encodes the current string into the appropriate encoding for the platform or display environment. This util does not exist however

gfyoung:
Closes #20616 

Nice to finally get back to some `pandas` work!


gfyoung:
I ran `grep` to double check, and there are no other instances left.

PR #20667, opened 2018-04-12
DOC: Updated the docstring of pandas.Series.str.get

pranavsuri:
an unrelated PR):

```
###################### Docstring (pandas.Series.str.get)  ######################

Extract the element from each component at the specified position.

Extract element from lists, tuples, or strings in each element in the
Series/Index.

Parameters
----------
i : int
    Position of the element to extract.

Returns
-------
items : Series/Index of objects

Examples
--------
>>> s = pd.Series(["String", (1, 2, 3), ["a", "b", "c"], 123])
>>> s
0       String
1    (1, 2, 3)
2    [a, b, c]
3          123
dtype: object

>>> s.str.get(1)
0      t
1      2
2      b
3    NaN
dtype: object

Errors found:
	See Also section not found

```

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1651,17 +1651,36 @@ def str_translate(arr, table, deletechars=None):
 
 def str_get(arr, i):
     """
+    Extract element from each component at specified position.
+
     Extract element from lists, tuples, or strings in each element in the
     Series/Index.
 
     Parameters
     ----------
     i : int
-        Integer index (location)
+        Position of element to extract.
 
     Returns
     -------
     items : Series/Index of objects
+
+    Examples
+    --------
+    >>> s = pd.Series(["String", (1, 2, 3), ["a", "b", "c"], 123])
+    >>> s
+    0       String
+    1    (1, 2, 3)
+    2    [a, b, c]
+    3          123
+    dtype: object
+
+    >>> s.str.get(1)
+    0      t
+    1      2
+    2      b
+    3    NaN
+    dtype: object
     """
     f = lambda x: x[i] if len(x) > i >= -len(x) else np.nan
     return _na_map(f, arr)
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -1651,17 +1651,36 @@ def str_translate(arr, table, deletechars=None):
...
+    2    [a, b, c]
+    3          123
+    dtype: object
+
+    >>> s.str.get(1)
+    0      t
+    1      2
+    2      b
+    3    NaN
+    dtype: object
```
datapythonista: Looks good. May be we could a dictionary to the `Series` too, and an example indexing by a negative number. I'm not sure, as the example could be too complex for the simplicity of the function. But it'd illustrate the different behaviors.

It'd also help to have an explanation before the example, saying something like "When the value is not an iterable, or a dict where the index is not in the keys, `str.get` returns `NaN`".

If you do it you may encounter this bug: #20671. I'd wait until it's fixed, otherwise you'd have to use an index that is a key in the dictionary, and not illustrate the case when the index is not a key, which raises a `KeyError` right now.

pranavsuri: Hello @datapythonista! I see that you've fixed the bug (which I tried fixing, but your solution is a lot better). Should I proceed writing the docstring with a dictionary & a negative value?

pranavsuri:
Hello @datapythonista! I see that you've fixed the bug (which I tried fixing, but your solution is a lot better). Should I proceed writing the docstring with a dictionary & a negative value?

datapythonista:
You can add it, yes. My changes are not merged, so when you run the validation script you'll get the doctest fail, but no problem with that, when the bug is fixed your doctest will pass and your PR could be merged.

datapythonista:
@pranavsuri, Can you fix the PEP-8 issues? Besides that lgtm. But it needs to wait until #20671 is merged.

jreback:
thanks @pranavsuri

PR #20670, opened 2018-04-12
ERR: Consistent errors for non-numeric ranking. (#19560)

### LINKED ISSUE #19560: Raise ValueError When Attempting to Rank Object Dtypes
Referencing the comments in #19481, right now rank operations performed against objects have a few issues, namely that they:

- Are inherently ambiguous, relying on lexical encoding AND
- Are not consistent across Series, DataFrame and GroupBy objects with various arguments

To illustrate the latter:

```python
In [1]: vals = ['apple', 'orange', 'banana']
In [2]  pd.Series(vals).rank()  # this will "work"
Out[8]: 
0    1.0
1    3.0
2    2.0
dtype: float64

In [3]: pd.Series(vals).rank(method='first')  # raises
ValueError: first not supported for non-numeric data

In [4]: pd.DataFrame({'key': ['foo'] * 3, 'vals': vals}).groupby('key').rank(method='first')  # should raise?
Out[4]: 
Empty DataFrame
Columns: []
Index: []
```

(see also #19482)

With this change I'd propose that we simply raise ValueError consistently for rank against ``object`` dtypes regardless of which type of object performs the transformation and regardless of arguments. 

One known caveat is that ``Categorical`` types currently use the ``rank_object`` methods in ``algos``. My assumption is that we would want to continue supporting ranking for ordered Categoricals but raise for unordered Categoricals.

mapehe:
- [ ] closes #19560 
- [x] tests added / passed
- [x] whatsnew entry

*This is only partial solution to issue #19560. 
*There were some errors with the tests, but I guess they are unrelated to these changes since I also
have them with master.
*I modified some tests so that they don't contradict with the update "don't allow objects to be ranked unless they are ordered categorials" that was suggested in #19560.

### DIFF
--- a/pandas/tests/series/test_rank.py
+++ b/pandas/tests/series/test_rank.py
@@ -134,22 +134,27 @@ class TestSeriesRank(TestData):
         assert_series_equal(ordered.rank(), exp)
         assert_series_equal(ordered.rank(ascending=False), exp_desc)
 
-        # Unordered categoricals should be ranked as objects
+        # See #19560
+        error_msg = ("pandas.core.algorithms.rank "
+                     "not supported for unordered "
+                     "non-numeric data")
+
+        # Ranking unordered categorials depreciated per #19560
         unordered = Series(['first', 'second', 'third', 'fourth',
                             'fifth', 'sixth']).astype(
             CategoricalDtype(categories=['first', 'second', 'third',
                                          'fourth', 'fifth', 'sixth'],
                              ordered=False))
-        exp_unordered = Series([2., 4., 6., 3., 1., 5.])
-        res = unordered.rank()
-        assert_series_equal(res, exp_unordered)
+
+        with tm.assert_raises_regex(ValueError, error_msg):
+            unordered.rank()
 
         unordered1 = Series(
             [1, 2, 3, 4, 5, 6],
         ).astype(CategoricalDtype([1, 2, 3, 4, 5, 6], False))
-        exp_unordered1 = Series([1., 2., 3., 4., 5., 6.])
-        res1 = unordered1.rank()
-        assert_series_equal(res1, exp_unordered1)
+
+        # Won't raise ValueError because entries not objects.
+        unordered1.rank()
 
         # Test na_option for rank data
         na_ser = Series(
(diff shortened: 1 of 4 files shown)


jreback:
it looks like another commit is included here. can you rebase on master.

mapehe:
Rebased and added a whatsnew entry @jreback.

jreback (changes requested):

pandas/tests/series/test_rank.py
```diff
@@ -134,22 +134,27 @@ def test_rank_categorical(self):
         assert_series_equal(ordered.rank(), exp)
         assert_series_equal(ordered.rank(ascending=False), exp_desc)
 
-        # Unordered categoricals should be ranked as objects
+        # See #19560
+        error_msg = ("pandas.core.algorithms.rank "
+                     "not supported for unordered "
+                     "non-numeric data")
+
+        # Ranking unordered categorials depreciated per #19560
```
WillAyd: I'd prefer to say "not supported" instead of "deprecated"

pandas/tests/series/test_rank.py
```diff
@@ -134,22 +134,27 @@ def test_rank_categorical(self):
...
 
         unordered1 = Series(
             [1, 2, 3, 4, 5, 6],
         ).astype(CategoricalDtype([1, 2, 3, 4, 5, 6], False))
-        exp_unordered1 = Series([1., 2., 3., 4., 5., 6.])
-        res1 = unordered1.rank()
-        assert_series_equal(res1, exp_unordered1)
+
+        # Won't raise ValueError because entries not objects.
+        unordered1.rank()
```
WillAyd: Is this right? Certainly the example passes the "eye test" but if the Categorical is not ordered what semantics are we using for ranking?

pandas/tests/series/test_rank.py
```diff
@@ -379,9 +375,15 @@ def test_rank_int(self):
     def test_rank_object_bug(self):
```
WillAyd: Name of this test can be changed to `test_rank_na_object_raises`

WillAyd (changes requested):

PR #20703, opened 2018-04-15
PERF: enhance MultiIndex.remove_unused_levels when no levels are unused

### LINKED ISSUE #19289: Unstack performance regression
http://pandas.pydata.org/speed/pandas/#reshape.SparseIndex.time_unstack

https://github.com/pandas-dev/pandas/pull/18460#issuecomment-358365042

cc @toobaz

toobaz:
- [x] closes #19289
- [x] tests passed
- [x] whatsnew entry

This uses ``np.bincount`` rather than ``algos.unique`` to check _if_ there are unused levels: since levels indexes are adjacent integers, the former is faster (~50% in my tests). It then fallbacks to ``algos.unique`` if there are unused levels.

The result is that asv tests on ``--bench reshape`` give
```
     [d5d5a718]       [499ae6bf]
-     1.43±0.01ms      1.19±0.01ms     0.83  reshape.SparseIndex.time_unstack
```

There is a penalty (of the same order of magnitude) for cases in which _there are_ unused levels, but I think this is worth because:
- there is room for cythonizing and bringing this penalty close to zero in the average case (see comment in code)
- I think the case of no unused level is much more frequent
- if people want to avoid this penalty they can just clean their index from unused levels once for all
- in any case, asv tests on ``--bench multiindex_object`` (where the index in ``Duplicates`` always has unused levels) do not find any change in performance

### DIFF
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -901,6 +901,7 @@ Performance Improvements
 - :func:`Series` / :func:`DataFrame` tab completion limits to 100 values, for better performance. (:issue:`18587`)
 - Improved performance of :func:`DataFrame.median` with ``axis=1`` when bottleneck is not installed (:issue:`16468`)
 - Improved performance of :func:`MultiIndex.get_loc` for large indexes, at the cost of a reduction in performance for small ones (:issue:`18519`)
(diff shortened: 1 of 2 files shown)


jreback:
thanks @toobaz I closed the issue, but I suppose can revisit a re-factorting of unstack and/or cythonize removing unused levels at some point.

PR #20718, opened 2018-04-17
PKG: remove pyproject.toml for now

jorisvandenbossche:
Since this gives problems, it might be better to remove this until support in pip is better? 
Eg I also have errors on the geopandas CI when installing pandas master with pip due to this.

Scipy did something similar: https://github.com/scipy/scipy/pull/8735

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,0 @@
-[build-system]
-requires = [
-    "wheel",
-    "setuptools",
-    "Cython",  # required for VCS build, optional for released source
-    "numpy==1.9.3; python_version=='3.5'",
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
That's still something else. It's also related to the release of pip 10, but because they changed some api's that multibuild relied upon: https://github.com/matthew-brett/multibuild/issues/159

jorisvandenbossche:
cc @TomAugspurger

jorisvandenbossche:
Going to merge this, as currently it is just broken. We can always add back before the 0.23.0 release if pip is updated by then.

PR #20721, opened 2018-04-17
DEPR: Series ndarray properties (strides, data, base, itemsize, flags)

### LINKED ISSUE #20419: Deprecate Series.strides, .base, .data, .itemsize, .flags (numpy) attributes ?
Related to https://github.com/pandas-dev/pandas/issues/18262, but since this is a very specific class of attributes, thought to open a specific issue about it.

We have a bunch of attributes on the `Series` class that stem from the time it was a numpy array subclass, and now just pass through the attribute of the underlying numpy array. It are typically attributes describing the data layout specific to the numpy array, which I don't think necessarily makes sense for a Series:

- `Series.base`
- `Series.data`
- `Series.strides`
- `Series.itemsize`
- `Series.flags`

and potentially also:

- `Series.real` and `Series.imag`

So deprecating those can potentially remove 7 entries from the Series namespace.

Are there good reasons to keep them? Is this somehow useful for "compatibility" (writing code that works for both series as numpy array) 
(I personally can't think of a usecase where you would want one of the above, unless you explicitly know will deal with numpy arrays)

One of the problems might be that if we refer users to `Series.values.<attribute>` that this will depend on the underlying array type if that will work or not (eg if `.values` starts giving an ExtensionArray, it will also not have those attributes)

cc @shoyer I don't think you kept those for "compatibility" in DataArray in xarray? 
And in dask I think `itemsize`, `real` and `imag` is provided.

jorisvandenbossche:
- [x] closes #20419
- [ ] tests added / passed
- [ ] whatsnew entry


pandas/core/base.py
```diff
@@ -737,11 +737,17 @@ def item(self):
     @property
     def data(self):
         """ return the data pointer of the underlying data """
+        warnings.warn("Series/Index.data is deprecated and will be "
```
jschendel: Might be nice to make the warning messages a bit more dynamic with something like: 
```python
"{obj}.data is deprecated...".format(obj=type(self).__name__)
```

jorisvandenbossche: Yes, thank you (somehow I was the docstring mindset thinking we needed to adapt this upon class definition, but of course this can be done dynamically on run time :-))

pandas/tests/groupby/aggregate/test_other.py
```diff
@@ -328,7 +328,7 @@ def test_series_agg_multi_pure_python():
          'F': np.random.randn(11)})
 
     def bad(x):
-        assert (len(x.base) > 0)
+        assert (len(x.values.base) > 0)
```
jorisvandenbossche: Not really sure what the purpose of this assert actually is (was introduced in https://github.com/pandas-dev/pandas/commit/71e9046c52246535d4db1f350e82c3a84d748f88)

pandas/core/indexes/datetimelike.py
```diff
@@ -209,6 +209,16 @@ def ceil(self, freq):
 class DatetimeIndexOpsMixin(object):
     """ common ops mixin to support a unified interface datetimelike Index """
 
+    @property
+    def base(self):
+        """ return the base object if the memory of the underlying data is
+        shared
+        """
+        # override deprecated property in IndexOpsMixin, as we still need
```
jreback: your explanation doesn't make sense, we are deprecating these no? you need to change the way its accessed in the code to remove the warnings.

jreback (changes requested):
pls run all tests and see if any warnings comes up

pandas/core/indexes/datetimelike.py
jorisvandenbossche: The explanation perfectly makes sense, but you don't like the way I solved it I suppose.

`.base` is used in two placed:

- `Block.is_view`: this I can override in `DatetimeTzBlock` to check `self.values.values.base`
- `concatenate_join_units`: https://github.com/pandas-dev/pandas/blob/ede11af8d02a4ac37ed866593839024b941a8086/pandas/core/internals.py#L5544  Not fully sure about this one what to do here.

pandas/core/algorithms.py
```diff
@@ -1491,6 +1491,8 @@ def take_nd(arr, indexer, axis=0, out=None, fill_value=np.nan, mask_info=None,
     if is_sparse(arr):
         arr = arr.get_values()
 
+    arr = np.array(arr, copy=False)
```
jreback: ``np.asarray`` slightly more idiomatic

pandas/core/indexes/datetimelike.py
jreback: my point is this method should not exist (as you are ddeprecateding), and would rather have you fix the usage

jreback (changes requested):

pandas/core/indexes/datetimelike.py
jorisvandenbossche: As I said, I agree with that. Do you have a suggestion for the second case? (inside `concatenate_join_units`) In https://github.com/pandas-dev/pandas/pull/20745 I need to touch the same code, and checked if the values have a `base` attribute, but that also feels hacky

pandas/core/algorithms.py
jorisvandenbossche: Ah, yes (for some reason I wanted the `copy=False` keyword, but for `asarray` that is of course just the default behaviour)

pandas/core/indexes/datetimelike.py
jreback: yes, see my comment there, only check base if its an ndarray

jreback:
can you update

jorisvandenbossche:
Now that #20745 is merged, updated this.

jreback:
lgtm, other than @TomAugspurger comments.

jreback (approved):

jreback:
lgtm. modulo any warnings fixes. note that I just pushed a CI fix for lint (on other PR)

pandas/core/indexes/base.py
```diff
@@ -548,6 +548,9 @@ def _deepcopy_if_needed(self, orig, copy=False):
         """
         if copy:
             # Retrieve the "base objects", i.e. the original memory allocations
+            if not isinstance(orig, np.ndarray):
+                # orig is a DatetimeIndex
+                orig = orig.values
```
jorisvandenbossche: @jreback This special case is because we sometimes construct DatetimeIndex objects where the `_data` attribute still is a DatetimeIndex in itself. 
To me this seems like something we should not do? (it doesn't hurt, but there is also no reason to have it I think, so better always store the ndarray to be consistent?)

jorisvandenbossche: Just opened an issue about it: https://github.com/pandas-dev/pandas/issues/20810 
It came up from fixing the warnings in the tests. The above case, where `orig` is not always a ndarray (hence my special case `if` statement), triggered it, because `orig` always comes from a `_data` attribute, and so I expected it to always be a ndarray.

TomAugspurger:
Thanks @jorisvandenbossche

naazneen:
I was doing scipy.fftpack.fft over a series and it showed error as
Attriute error: Series object has no attribute flags.

What should I use instead? And how to overcome it?

PR #20727, opened 2018-04-17
BUG:  Fix a bug in plotting when using color array. #20726

### LINKED ISSUE #20726: BUG: Extra color introduced with plot.
User reported in StackOverflow: https://stackoverflow.com/questions/49887076/matplotlib-error-while-coloring-barplot-with-list#49887076
```python
import numpy as np
import pandas as pd

df = pd.DataFrame(np.abs(np.random.randn(48, 4)), columns=list("ABCD"))

color_list = plt.cm.gnuplot(np.linspace(0,1,16))

df.A.plot.bar( figsize=(16,7), color=color_list)
plt.axvline(15.5, c="black", lw=1)
plt.axvline(31.5, c="black", lw=1)
plt.plot()
```
#### Problem description
This bug happens when single color is needed for plotting but an array is provided. 
In the end of `_get_standard_colors` function in plotting/_style.py. An extra first item will be added to the color array which caused the circular misbehaved in such case.
image

#### Expected Output
The bar 17 should have the same color as the bar 1.
####

wuhaochen:
- [x] closes #20726 
- [x] tests added / passed
- [x] passes `git diff master --name-only -- "*.py" | flake8`
- [x] whatsnew entry


jreback:
you would need a test for this & a whatsnew entry.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1159,6 +1159,7 @@ Plotting
 - Bug in formatting tick labels with ``datetime.time()`` and fractional seconds (:issue:`18478`).
 - :meth:`Series.plot.kde` has exposed the args ``ind`` and ``bw_method`` in the docstring (:issue:`18461`). The argument ``ind`` may now also be an integer (number of sample points).
 - :func:`DataFrame.plot` now supports multiple columns to the ``y`` argument (:issue:`19699`)
+- Bug in providing color array as color for a single plot (:issue:`20726`)
```
jreback: move to 0.24.0

jreback: can you make this slightly more verbose, a user won't really understand this

pandas/tests/plotting/test_misc.py
```diff
@@ -313,3 +313,14 @@ def test_get_standard_colors_random_seed(self):
         color1 = _get_standard_colors(1, color_type='random')
         color2 = _get_standard_colors(1, color_type='random')
         assert color1 == color2
+
+    def test_get_standard_colors_no_appending(self):
+        # GH20726
+
+        # Make sure not to add more colors so that matplotlib can cycle
+        # correctly.
```
jreback: ideally you can also test the example in the OP (as well)

jreback (changes requested):

pandas/tests/plotting/test_misc.py
wuhaochen: Added.

WillAyd:
Can you merge master for this?

datapythonista:
@jreback is anything else needed here, or can this be merged?

pandas/tests/plotting/test_misc.py
```diff
@@ -308,3 +308,26 @@ def test_get_standard_colors_random_seed(self):
...
+        from pandas.plotting._style import _get_standard_colors
+        from matplotlib import cm
+        color_before = cm.gnuplot(range(5))
+        color_after = _get_standard_colors(1, color=color_before)
+        assert len(color_after) == len(color_before)
+
+        # Original bug report example.
+        import numpy as np
```
WillAyd: These imports are already done at the top of the file; no need for them here

pandas/tests/plotting/test_misc.py
```diff
@@ -308,3 +308,26 @@ def test_get_standard_colors_random_seed(self):
...
+
+        # Make sure not to add more colors so that matplotlib can cycle
+        # correctly.
+        from pandas.plotting._style import _get_standard_colors
+        from matplotlib import cm
+        color_before = cm.gnuplot(range(5))
+        color_after = _get_standard_colors(1, color=color_before)
+        assert len(color_after) == len(color_before)
+
+        # Original bug report example.
```
WillAyd: Don't need this comment

pandas/plotting/_style.py
```diff
@@ -80,7 +80,7 @@ def _maybe_valid_colors(colors):
             # mpl will raise error any of them is invalid
             pass
 
-    if len(colors) != num_colors:
+    if len(colors) < num_colors:
```
WillAyd: Could you add a docstring to this function while updating? Personally not familiar with it would be helpful as you just reviewed

pandas/tests/plotting/test_misc.py
```diff
@@ -308,3 +308,26 @@ def test_get_standard_colors_random_seed(self):
         color1 = _get_standard_colors(1, color_type='random')
         color2 = _get_standard_colors(1, color_type='random')
         assert color1 == color2
+
+    def test_get_standard_colors_no_appending(self):
+        # GH20726
+
+        # Make sure not to add more colors so that matplotlib can cycle
+        # correctly.
+        from pandas.plotting._style import _get_standard_colors
```
WillAyd: Make these module level imports

pandas/tests/plotting/test_misc.py
```diff
@@ -308,3 +308,26 @@ def test_get_standard_colors_random_seed(self):
...
+        from matplotlib import cm
+        color_before = cm.gnuplot(range(5))
+        color_after = _get_standard_colors(1, color=color_before)
+        assert len(color_after) == len(color_before)
+
+        # Original bug report example.
+        import numpy as np
+        import pandas as pd
+
+        df = pd.DataFrame(np.abs(np.random.randn(48, 4)),
```
WillAyd: Do we need the call to `np.abs` for any particular reason?

WillAyd (changes requested):

pandas/tests/plotting/test_misc.py
wuhaochen: Nope. This is just a literal copy of original SO report. Removed.

pandas/tests/plotting/test_misc.py
wuhaochen: Done. However there are a lot of other local import in this file and many other tests, maybe it's time for some clean up?

pandas/plotting/_style.py
wuhaochen: This is a bit hacky. Though I added some comments about why this newer version is more correct, it doesn't actually solve the root cause of the bug which is to guess the number of colors correctly. It should be good enough combined with matplotlib's mechanism though.

mroeschke (commented):
A few comments otherwise LGTM.

jreback (approved):

jreback:
thanks @wuhaochen

Issue #19904, opened 2018-02-26
Unable to specify AWS profile when saving parquet file with s3 url


tomers:
I would like to be able to specify the [named profile](http://boto3.readthedocs.io/en/latest/guide/configuration.html#shared-credentials-file) to use when uploading a dataframe as parquet file to S3. This feature seems to be missing. The work-around is to specify the default account as the the account I would like to use, but this does not allow me to [specify programmatically the account name to use](https://stackoverflow.com/questions/33378422/how-to-chose-an-aws-profile-when-using-boto3-to-connect-to-cloudfront).

#### Code Sample

```bash
$ python
Python 3.6.2 (default, Jul 17 2017, 16:44:45) 
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy as np
>>> import pandas as pd
>>> 
>>> s3_url = 's3://my_bucket/foo/bar/example.parquet'
>>> df = pd.DataFrame({'one': [-1, np.nan, 2.5], 'two': ['foo', 'bar', 'baz'], 'three': [True, False, True]})
>>> 
>>> df.to_parquet(s3_url, engine='fastparquet', profile='production')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/user/mytool/env/lib/python3.6/site-packages/pandas/core/frame.py", line 1691, in to_parquet
    compression=compression, **kwargs)
  File "/Users/user/mytool/env/lib/python3.6/site-packages/pandas/io/parquet.py", line 248, in to_parquet
    return impl.write(df, path, compression=compression, **kwargs)
  File "/Users/user/mytool/env/lib/python3.6/site-packages/pandas/io/parquet.py", line 210, in write
    compression=compression, **kwargs)
TypeError: write() got an unexpected keyword argument 'profile'
>>> 
```

I also tried the `profile_name` argument. This feature seems to be missing.

<details>

[paste the output of ``pd.show_versions()`` here below this line]

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.2.final.0
python-bits: 64
OS: Darwin
OS-release: 16.7.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.23.0.dev0+273.gcd484cc52
[...]

PR #20729, opened 2018-04-18
[ENH] Add read support for Google Cloud Storage

### LINKED ISSUE #19454: ENH Adding read_* support for Google Cloud Storage gs:// URLs
Would be great if pandas were to support reading directly from Google Cloud Storage URLs (gs://), similar to the way it supports AWS s3 URLs (s3://)

Ideally, one could write:
```python
df_csv = pd.read_csv('gs://my-bucket/super-cool-dataset.csv')
df_excel = pd.read_excel('gs://my-bucket/super-cool-dataset.xlsx')

```

I took a quick stab at it in https://github.com/pandas-dev/pandas/commit/4c9196ace8e70f83f81d84d921a73db2581453a4. Let me know if this makes sense and I can add some documentation/tests around it and open a pull request.

bnaul:
- [x] closes #19454
- [x] tests added / passed
- [x] whatsnew entry

Couple of remaining issues:
- I had to do some weird logging magic to suppress the output of `gcsfs` when catching authentication exceptions
- Proper Parquet support seems to require more specialization so I didn't try
- I'm not sure how to properly test this functionality without a `pandas-test` bucket and/or a mock GCS library like `moto`

cc @martindurant who might have some thoughts on other things I am doing wrong

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -276,6 +276,7 @@ Optional Dependencies
 * `Jinja2 <http://jinja.pocoo.org/>`__: Template engine for conditional HTML formatting.
 * `s3fs <http://s3fs.readthedocs.io/>`__: necessary for Amazon S3 access (s3fs >= 0.0.7).
 * `blosc <https://pypi.org/project/blosc>`__: for msgpack compression using ``blosc``
+* `gcsfs <http://gcsfs.readthedocs.io/>`__: necessary for Google Cloud Storage access (gcsfs >= 0.1.0).
 * One of
   `qtpy  <https://github.com/spyder-ide/qtpy>`__ (requires PyQt or PySide),
   `PyQt5 <https://www.riverbankcomputing.com/software/pyqt/download5>`__,
--- a/pandas/io/common.py
+++ b/pandas/io/common.py
@@ -88,7 +88,7 @@ def _is_url(url):
     """
     try:
         return parse_url(url).scheme in _VALID_URLS
-    except:
+    except Exception:
         return False
 
 
@@ -165,7 +165,15 @@ def is_s3_url(url):
     """Check for an s3, s3n, or s3a url"""
     try:
         return parse_url(url).scheme in ['s3', 's3n', 's3a']
-    except:  # noqa
+    except Exception:
+        return False
+
+
+def is_gcs_url(url):
+    """Check for a gcs url"""
+    try:
+        return parse_url(url).scheme in ['gcs', 'gs']
+    except Exception:
         return False
 
 
@@ -207,6 +215,13 @@ def get_filepath_or_buffer(filepath_or_buffer, encoding=None,
                                          encoding=encoding,
                                          compression=compression,
                                          mode=mode)
+
+    if is_gcs_url(filepath_or_buffer):
+        from pandas.io import gcs
+        return gcs.get_filepath_or_buffer(filepath_or_buffer,
+                                          encoding=encoding,
+                                          compression=compression,
+                                          mode=mode)
 
     if isinstance(filepath_or_buffer, (compat.string_types,
                                        compat.binary_type,
--- a/pandas/io/gcs.py
+++ b/pandas/io/gcs.py
@@ -1,0 +1,16 @@
+""" GCS support for remote file interactivity """
+try:
+    import gcsfs
+except ImportError:
+    raise ImportError("The gcsfs library is required to handle GCS files")
+
+
+def get_filepath_or_buffer(filepath_or_buffer, encoding=None,
+                           compression=None, mode=None):
+
+    if mode is None:
+        mode = 'rb'
+
+    fs = gcsfs.GCSFileSystem()
+    filepath_or_buffer = fs.open(filepath_or_buffer, mode)
+    return filepath_or_buffer, None, compression, True
--- a/pandas/tests/io/test_gcs.py
+++ b/pandas/tests/io/test_gcs.py
@@ -1,0 +1,57 @@
+import numpy as np
+import pytest
+
+from pandas import DataFrame, date_range, read_csv
+from pandas.compat import StringIO
+from pandas.io.common import is_gcs_url
+from pandas.util import _test_decorators as td
+from pandas.util.testing import assert_frame_equal
+
+
+@pytest.fixture
+def mock_patch():
+    try:
+        from unittest.mock import patch
+    except ImportError:
+        from mock import patch
+
+    return patch
+
+
+def test_is_gcs_url():
+    assert is_gcs_url("gcs://pandas/somethingelse.com")
+    assert is_gcs_url("gs://pandas/somethingelse.com")
+    assert not is_gcs_url("s3://pandas/somethingelse.com")
+
+
+@td.skip_if_no('gcsfs')
+def test_read_csv_gcs(mock_patch):
+    df1 = DataFrame({'int': [1, 3], 'float': [2.0, np.nan], 'str': ['t', 's'],
+                     'dt': date_range('2018-06-18', periods=2)})
+    with mock_patch('gcsfs.GCSFileSystem') as MockFileSystem:
+        instance = MockFileSystem.return_value
+        instance.open.return_value = StringIO(df1.to_csv(index=False))
+        df2 = read_csv('gs://test/test.csv', parse_dates=['dt'])
+
+    assert_frame_equal(df1, df2)
+
+
+@td.skip_if_no('gcsfs')
+def test_gcs_get_filepath_or_buffer(mock_patch):
+    df1 = DataFrame({'int': [1, 3], 'float': [2.0, np.nan], 'str': ['t', 's'],
+                     'dt': date_range('2018-06-18', periods=2)})
+    with mock_patch('pandas.io.gcs.get_filepath_or_buffer') as MockGetFilepath:
+        MockGetFilepath.return_value = (StringIO(df1.to_csv(index=False)),
+                                        None, None, False)
+        df2 = read_csv('gs://test/test.csv', parse_dates=['dt'])
+
+    assert_frame_equal(df1, df2)
+    assert MockGetFilepath.called
+
+
+@pytest.mark.skipif(td.safe_import('gcsfs'),
+                    reason='Only check when gcsfs not installed')
+def test_gcs_not_present_exception():
+    with pytest.raises(ImportError) as e:
+        read_csv('gs://test/test.csv')
+        assert 'gcsfs library is required' in str(e.value)
--- a/pandas/tests/io/test_gcs.py
+++ b/pandas/tests/io/test_gcs.py
@@ -8,16 +8,6 @@ from pandas.util import _test_decorators as td
 from pandas.util.testing import assert_frame_equal
 
 
-@pytest.fixture
-def mock_patch():
-    try:
-        from unittest.mock import patch
-    except ImportError:
-        from mock import patch
-
-    return patch
-
-
 def test_is_gcs_url():
     assert is_gcs_url("gcs://pandas/somethingelse.com")
     assert is_gcs_url("gs://pandas/somethingelse.com")
@@ -25,10 +15,10 @@ def test_is_gcs_url():
 
 
 @td.skip_if_no('gcsfs')
-def test_read_csv_gcs(mock_patch):
+def test_read_csv_gcs(mock):
     df1 = DataFrame({'int': [1, 3], 'float': [2.0, np.nan], 'str': ['t', 's'],
                      'dt': date_range('2018-06-18', periods=2)})
-    with mock_patch('gcsfs.GCSFileSystem') as MockFileSystem:
+    with mock.patch('gcsfs.GCSFileSystem') as MockFileSystem:
         instance = MockFileSystem.return_value
         instance.open.return_value = StringIO(df1.to_csv(index=False))
         df2 = read_csv('gs://test/test.csv', parse_dates=['dt'])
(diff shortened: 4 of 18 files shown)


bnaul:
I'm also not sure what the all the various requirements files I touched are, I just mimicked the s3fs pattern; looks like Circle is unhappy with that, should I revert everything in `ci/` or is there some other step I'm missing?

doc/source/install.rst
```diff
@@ -275,6 +275,7 @@ Optional Dependencies
 
 * `Jinja2 <http://jinja.pocoo.org/>`__: Template engine for conditional HTML formatting.
 * `s3fs <http://s3fs.readthedocs.io/>`__: necessary for Amazon S3 access (s3fs >= 0.0.7).
+* `gcsfs <http://gcsfs.readthedocs.io/>`__: necessary for Google Cloud Storage access (gcsfs >= 0.6.0).
```
martindurant: 0.0.6?

martindurant:
Testing in gcsfs is done via the VCR package, which records all HTTP calls and their responses from the server, for playback during the running of a test. It is rather finicky/awkward to use.

Parquet is difficult because you need to supply the `open` function rather than file-like objects?

pandas/io/common.py
```diff
@@ -169,6 +169,14 @@ def is_s3_url(url):
         return False
 
 
+def is_gcs_url(url):
+    """Check for a gcs url"""
+    try:
+        return parse_url(url).scheme in ['gcs', 'gs']
+    except:  # noqa
```
TomAugspurger: Make this `except Exception`, or even better figure out what `parse_url` can raise, and catch those (plus `AttributeError`).

pandas/io/gcs.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def get_filepath_or_buffer(filepath_or_buffer, encoding=None,
+                           compression=None, mode=None):
+
+    if mode is None:
+        mode = 'rb'
+
+    gcsfs_logger_disabled = gcsfs.core.logger.disabled
+    gcsfs.core.logger.disabled = True
```
TomAugspurger: Why disable this? I'd rather leave that up to the user if possible.

pandas/io/gcs.py
```diff
@@ -0,0 +1,26 @@
...
+    fs = gcsfs.GCSFileSystem()
+    try:
+        filepath_or_buffer = fs.open(filepath_or_buffer, mode)
+    except (compat.FileNotFoundError, GoogleAuthError, gcsfs.utils.HtmlError):
+        fs = gcsfs.GCSFileSystem(token='anon')
```
TomAugspurger: note: this was originally a hack for not breaking backwards compat with pre-s3fs pandas. I'd like to eventually move to something like dask's `store_options` here, a dict of keywords that gets passed down to the backend (https://github.com/pandas-dev/pandas/issues/19904)

pandas/io/gcs.py
```diff
@@ -0,0 +1,26 @@
...
+        filepath_or_buffer = fs.open(filepath_or_buffer, mode)
+    finally:
+        gcsfs.core.logger.disabled = gcsfs_logger_disabled
```
TomAugspurger: If the anon version fails, this line is not hit, correct? It would have to be wrapped in a `try / finally` as well (or ideally the logger stuff would be removed).

TomAugspurger (commented):
Not sure about testing yet. VCR has sounded like more trouble than it's worth.

pandas/io/common.py
bnaul: From looking over `urlparse` it seems that it will also only raise `AttributeError` so I used that. For consistency I changed the other url parsers `except`s as well.

pandas/io/gcs.py
bnaul: This was probably an overly complicated solution; the problem I was facing was that `gcsfs` retries several times if it fails to authenticate, and each failure prints a long traceback. But if that's the behavior of the library then I guess there's no sense in using hacks to get around it. For now I removed all of the logging magic.

pandas/io/gcs.py
bnaul: I agree this would be the best way to do things. For now I simplified things and am just using `gcsfs`'s default authentication preferences. If allowing anonymous access to public files is important we could try to think of a better way to accomplish that, but arguably `gcsfs` should handle that logic?

martindurant: Note that gcsfs defaults to anon access if no other mechanisms worked - but this will not happen if one of the other mechanisms contained real credentials that turned out to be no longer valid.

bnaul: That's what I thought but it didn't seem to be the case when I tested it. In a `python:3.6.4` Docker image w/ `gcsfs` installed I get:
```
# default auth fails
In [2]: gcsfs.GCSFileSystem().open('gs://gcp-public-data-landsat/index.csv.gz').read(10)
# lots of error logs from failing to access /computeMetadata/v1/instance/service-accounts/default/?recursive=true
...
RefreshError: HTTPConnectionPool(host='metadata.google.internal', port=80): Max retries exceeded with url: /computeMetadata/v1/instance/service-accounts/default/?recursive=true (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fab36881518>: Failed to establish a new connection: [Errno -2] Name or service not known',))

# specifying token='anon' works
In [3]: gcsfs.GCSFileSystem(token='anon').open('gs://gcp-public-data-landsat/index.csv.gz').read(10)
Out[3]: b'\x1f\x8b\x08\x08\x89\xd8\xd6Z\x02\xff'
```

Should I create a separate `gcsfs` issue for this?

martindurant: Hm, OK, it should be an issue - but worth checking whether it's still the case with current master.

bnaul:
The issue with anonymous credentials in `gcsfs` appears to be resolved by https://github.com/dask/gcsfs/pull/100, so once that is released I think this PR should cover most use cases.

Any more thoughts re: testing? Would simply mocking out `gcsfs.GCSFileSystem.read` to return a simple CSV be sufficient?

martindurant:
This was not implemented in Dask (yet), but the [MemoryFileSystem](https://github.com/dask/dask/pull/2741) might be a nice and simple mock for GCSFileSystem.

martindurant:
gcsfs v0.0.7 being released: https://github.com/conda-forge/gcsfs-feedstock/pull/7

bnaul:
Any more thoughts about testing? Would a limited set of tests based on `mock.patch`ing `GCSFileSystem.open` be sufficient for now?

jreback:
I think we would need some tests, mocking would be ok

pandas/io/gcs.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def get_filepath_or_buffer(filepath_or_buffer, encoding=None,
+                           compression=None, mode=None):
+
+    if mode is None:
+        mode = 'rb'
+
+    fs = gcsfs.GCSFileSystem()
+    filepath_or_buffer = fs.open(filepath_or_buffer, mode)
```
TomAugspurger: @bnaul do you think this is an appropriate place to mock? `fs.open` could return a BytesIO object?

bnaul: Added a simple mock-based test here; any other methods I should include besides `read_csv`?

pandas/tests/io/test_gcs.py
```diff
@@ -0,0 +1,25 @@
...
+def test_read_csv_gcs():
+    try:
+        from unittest.mock import patch
+    except ImportError:
+        from mock import patch
+
+    with patch('gcsfs.GCSFileSystem') as MockFileSystem:
+        instance = MockFileSystem.return_value
+        instance.open.return_value = BytesIO(b'a,b\n1,2\n3,4')
+        df = read_csv('gs://test/test.csv')
```
TomAugspurger: Mock let's you assert that certain functions are called, right?

I guess, maybe we could have a second test that ensures `pandas.io.gcs.get_filepath_or_buffer` is called for `gs://` type `filepath_or_buffer`?

TomAugspurger (commented):
Could you also

- add a whatsnew 0.24 probably
- rerun `python scripts/convert_deps.py` so that gcsfs is in the optional pip deps.

bnaul:
@TomAugspurger sorry for the delay, was away for a while; made the requested changes!

TomAugspurger:
Changes look good, but the CI is failing on builds that don't have gcsfs.
Could you

1. Add a `td.skip_if_no('gcsfs')` decorator to the tests that use gcsfs
2. Add a test that is skipped if gcsfs *is* present, and checks that not
having gcsfs raises with the right error message.

bnaul:
Still waiting on Appveyor but CircleCI and Travis are happy now

bnaul:
💚  @TomAugspurger

ci/requirements_dev.txt
```diff
@@ -9,4 +9,4 @@ python-dateutil>=2.5.0
 pytz
 setuptools>=24.2.0
 sphinx
-sphinxcontrib-spelling
```
jreback: what changed here? can you revert this file

ci/travis-36-slow.yaml
```diff
@@ -5,6 +5,7 @@ channels:
 dependencies:
   - beautifulsoup4
   - cython
+  - gcsfs
```
jreback: so I would remove gcsfs from the *slow* ci files (not tested anyhow as these tests are not slow)

ci/circle-36-locale.yaml
```diff
@@ -5,6 +5,7 @@ channels:
 dependencies:
   - beautifulsoup4
   - cython
+  - gcsfs
```
jreback: let's remove from here to make sure the tests are skipped properly (as tested in 2.7 / 3.6 builds already)

pandas/tests/io/test_gcs.py
```diff
@@ -0,0 +1,55 @@
...
+
+def test_is_gcs_url():
+    assert is_gcs_url("gcs://pandas/somethingelse.com")
+    assert is_gcs_url("gs://pandas/somethingelse.com")
+    assert not is_gcs_url("s3://pandas/somethingelse.com")
+
+
+@td.skip_if_no('gcsfs')
+def test_read_csv_gcs():
+    try:
```
jreback: hmm we do install mock in py27, can you move this to pandas.util.testing (and then just import patch from there)

pandas/tests/io/test_gcs.py
```diff
@@ -0,0 +1,55 @@
...
+        from unittest.mock import patch
+    except ImportError:
+        from mock import patch
+
+    with patch('gcsfs.GCSFileSystem') as MockFileSystem:
+        instance = MockFileSystem.return_value
+        instance.open.return_value = BytesIO(b'a,b\n1,2\n3,4')
+        df = read_csv('gs://test/test.csv')
+
```
jreback: can you use ``tm.assert_from_equal`` so dtypes are tested

pandas/tests/io/test_gcs.py
```diff
@@ -0,0 +1,55 @@
...
+        from unittest.mock import patch
+    except ImportError:
+        from mock import patch
+
+    with patch('pandas.io.gcs.get_filepath_or_buffer') as MockGetFilepath:
+        MockGetFilepath.return_value = (BytesIO(b'a,b\n1,2\n3,4'), None, None,
+                                        False)
+        df = read_csv('gs://test/test.csv')
+
+    assert isinstance(df, DataFrame)
```
jreback: same

jreback: can you add other dtypes such as float, datetime, string, and some NaNs

jreback (changes requested):

pandas/tests/io/test_gcs.py
bnaul: Done

ci/travis-36-slow.yaml
bnaul: Done.

pandas/tests/io/test_gcs.py
TomAugspurger: I don't think we should make changes to `util/testing.py`. We don't want to require `mock` (for py27), and testing will be awkward if it's in `pandas/util`.

Better to make a little fixture or something for it.

bnaul: @TomAugspurger do you mean a fixture that does use mock, but in another directory that won't get imported outside of testing? Or a fixture that doesn't use mock at all?

TomAugspurger:
I meant a fixture that provides the imported module. Like pytest imporotskip, but with the try / except for the two locations.

bnaul:
Made a pytest fixture; if it belongs in somewhere more generic let me know, the various testing utils seem spread out over quite a few different places and it's a bit confusing to me

TomAugspurger:
Not quite what I had in mind. I'll push a change in a minute :)

bnaul:
Sounds good 🙃

TomAugspurger:
OK, pushed. Only real changes were

1. returning `mock` instead of `mock.patch`
2. moving to `conftest.py`, since it's a generally useful fixture
3. skipping if neither `unittest.mock` nor `mock` are available

TomAugspurger (approved):

TomAugspurger:
Fixed the linting error I introduced.

May have a random failure on PY37, will investigate more later.

bnaul:
Oops might have overwritten that w/ the same change but I assume it's fine

bnaul:
@jreback looking 🍏

TomAugspurger:
Thanks @bnaul!

PR #20736, opened 2018-04-18
DOC: add coverage href to README.md

wuhaochen:


jreback:
thanks!

jreback:
if you are interested in cleaning any of the other badges on the README would be great (I think we have some older refs)

PR #14832, opened 2016-12-08
BF(TST): use = (native) instead of < (little endian) for target data types

yarikoptic:
- [ ] closes #14830
 - [ ] tests added / passed
 - [ ] passes ``git diff upstream/master | flake8 --diff``
 - [ ] whatsnew entry

It is still WiP -- just thought to submit a PR to see how travis etc reacts on typical platforms

PR #20738, opened 2018-04-18
Fix more tests expecting little-endian

ginggs:
As in #14832, use = (native) instead of < (little-endian)

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


jreback:
thanks!

Issue #20743, opened 2018-04-19
BUG: pd.merge with ExtensionArray does not preserve extension dtype


jorisvandenbossche:
```
In [1]: from pandas.tests.extension.decimal.array import DecimalArray, make_data

In [5]: dec_arr = DecimalArray(make_data())

In [6]: df1 = pd.DataFrame({'int1': [1, 2, 3], 'key':[0, 1, 2], 'ext1': dec_arr[:3]})

In [7]: df2 = pd.DataFrame({'int2': [1, 2, 3, 4], 'key':[0, 0, 1, 3], 'ext2': dec_arr[3:7]})

In [8]: pd.merge(df1, df2)
Out[8]: 
                                                ext1  int1  key                                               ext2  int2
0  0.90013275661511904512934734157170169055461883...     1    0  0.67786011817398117429434023506473749876022338...     1
1  0.90013275661511904512934734157170169055461883...     1    0  0.94029656863099908559178174982662312686443328...     2
2  0.96839085663514357094072693143971264362335205...     2    1  0.12455159685855177187363551638554781675338745...     3

In [9]: pd.merge(df1, df2, how='outer')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-9-a573147da092> in <module>()
----> 1 pd.merge(df1, df2, how='outer')

/home/joris/scipy/pandas/pandas/core/reshape/merge.py in merge(left, right, how, on, left_on, right_on, left_index, right_index, sort, suffixes, copy, indicator, validate)
     59                          copy=copy, indicator=indicator,
     60                          validate=validate)
---> 61     return op.get_result()
     62 
     63 

/home/joris/scipy/pandas/pandas/core/reshape/merge.py in get_result(self)
    579             [(ldata, lindexers), (rdata, rindexers)],
    580             axes=[llabels.append(rlabels), join_index],
--> 581             concat_axis=0, copy=self.copy)
    582 
    583         typ = self.left._constructor

/home/joris/scipy/pandas/pandas/core/internals.py in concatenate_block_managers(mgrs_indexers, axes, concat_axis, copy)
   5407         else:
   5408             b = make_block(
-> 5409                 concatenate_join_units(join_units, concat_axis, copy=copy),
   5410                 placement=placement)
   5411         blocks.append(b)

/home/joris/scipy/pandas/pandas/core/internals.py in concatenate_join_units(join_units, concat_axis, copy)
[...]

PR #20745, opened 2018-04-19
Fix pd.merge to preserve ExtensionArrays dtypes

jorisvandenbossche:
Closes https://github.com/pandas-dev/pandas/issues/20743

### DIFF
--- a/pandas/core/internals.py
+++ b/pandas/core/internals.py
@@ -5547,7 +5547,7 @@ def concatenate_join_units(join_units, concat_axis, copy):
                 # in JoinUnit.get_reindexed_values
                 if concat_values.base is not None:
                     concat_values = concat_values.copy()
-            elif:
+            else:
                 concat_values = concat_values.copy()
     else:
         concat_values = _concat._concat_compat(to_concat, axis=concat_axis)
--- a/pandas/core/internals.py
+++ b/pandas/core/internals.py
@@ -5541,8 +5541,14 @@ def concatenate_join_units(join_units, concat_axis, copy):
     if len(to_concat) == 1:
         # Only one block, nothing to concatenate.
         concat_values = to_concat[0]
-        if copy and getattr(concat_values, 'base', 1) is not None:
-            concat_values = concat_values.copy()
+        if copy:
+            if isinstance(concat_values, np.ndarray):
+                # non-reindexed (=not yet copied) arrays are made into a view
+                # in JoinUnit.get_reindexed_values
+                if concat_values.base is not None:
+                    concat_values = concat_values.copy()
+            elif:
+                concat_values = concat_values.copy()
     else:
         concat_values = _concat._concat_compat(to_concat, axis=concat_axis)
 
--- a/pandas/core/internals.py
+++ b/pandas/core/internals.py
@@ -5541,7 +5541,7 @@ def concatenate_join_units(join_units, concat_axis, copy):
     if len(to_concat) == 1:
         # Only one block, nothing to concatenate.
         concat_values = to_concat[0]
-        if copy and concat_values.base is not None:
+        if copy and getattr(concat_values, 'base', 1) is not None:
             concat_values = concat_values.copy()
     else:
         concat_values = _concat._concat_compat(to_concat, axis=concat_axis)
@@ -5823,7 +5823,7 @@ class JoinUnit(object):
                 # External code requested filling/upcasting, bool values must
                 # be upcasted to object to avoid being upcasted to numeric.
                 values = self.block.astype(np.object_).values
-            elif self.block.is_categorical:
+            elif self.block.is_extension:
                 values = self.block.values
             else:
                 # No dtype upcasting is done here, it will be performed during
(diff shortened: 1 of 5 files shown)


pandas/core/internals.py
```diff
@@ -5541,7 +5541,7 @@ def concatenate_join_units(join_units, concat_axis, copy):
     if len(to_concat) == 1:
         # Only one block, nothing to concatenate.
         concat_values = to_concat[0]
-        if copy and concat_values.base is not None:
+        if copy and getattr(concat_values, 'base', 1) is not None:
```
jorisvandenbossche: Is this a hack, would like to have a better solution.

(related to discussion earlier this day in https://github.com/pandas-dev/pandas/pull/20721#discussion_r182716666 about deprecating Index.base)

jorisvandenbossche:
Note to self: 1 actual failing test (rest are network failures):

```
___________________________ TestReshaping.test_merge ___________________________
[gw1] linux -- Python 3.6.5 /home/travis/miniconda3/envs/pandas/bin/python
self = <pandas.tests.extension.json.test_json.TestReshaping object at 0x7fb92d208dd8>
data = JSONArary([{'D': 91, 'k': 84, 'L': 34, 'm': 63, 'C': 8, 'O': 1, 'R': 5}, {'T': 49, 'o': 46}, {'g': 81, 'O': 8, 'p': 54...': 32, 'p': 0, 'J': 89, 'd': 32, 's': 22}, {'W': 70, 'Y': 68, 'D': 87, 'k': 52}, {'F': 29, 'h': 10, 'S': 41, 'y': 24}])
na_value = {}
    def test_merge(self, data, na_value):
    
        df1 = pd.DataFrame({'int1': [1, 2, 3], 'key': [0, 1, 2],
                            'ext': data[:3]})
        df2 = pd.DataFrame({'int2': [1, 2, 3, 4], 'key': [0, 0, 1, 3]})
    
        res = pd.merge(df1, df2)
        exp = pd.DataFrame(
            {'int1': [1, 1, 2], 'int2': [1, 2, 3], 'key': [0, 0, 1],
             'ext': data._constructor_from_sequence(
                 [data[0], data[0], data[1]])})
>       self.assert_frame_equal(res, exp[['ext', 'int1', 'key', 'int2']])
pandas/tests/extension/base/reshaping.py:110: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

E       AssertionError: DataFrame.columns are different
E       
E       DataFrame.columns values are different (75.0 %)
E       [left]:  Index(['int1', 'key', 'ext', 'int2'], dtype='object')
E       [right]: Index(['ext', 'int1', 'key', 'int2'], dtype='object')
```

Not seeing this locally, and also not failing for Decimal, so the column order seems to depend on the extension dtype? (which is a bit strange)

pandas/core/internals.py
jreback: you should only check base if concat_values is an ndarray

pandas/tests/extension/base/reshaping.py
```diff
@@ -95,3 +95,24 @@ def test_set_frame_overwrite_object(self, data):
         df = pd.DataFrame({"A": [1] * len(data)}, dtype=object)
         df['A'] = data
         assert df.dtypes['A'] == data.dtype
+
+    def test_merge(self, data, na_value):
+
```
jreback: issue number

jreback (changes requested):

TomAugspurger:
I can reproduce your failure locally. Looking into it now.

TomAugspurger:
Ah, bug in `BaseDecimal.assert_frame_equal` not checking order of all the columns. It only considered extension & non extension columns independently.

TomAugspurger:
Pushed a commit fixing the Decimal tests base class and adjust the expected order. Hopefully this will pass things.

jorisvandenbossche:
Thanks for the fix! The order was indeed the culprit, but changed it again to be alphabetically in the original test frame, so it behaves the same for python < 3.6

pandas/core/internals.py
jorisvandenbossche: @jreback updated

jorisvandenbossche:
For Travis, the remaining failure is an s3 one, so this seems to be passing now.

TomAugspurger (approved):

pandas/tests/extension/base/reshaping.py
```diff
@@ -95,3 +95,24 @@ def test_set_frame_overwrite_object(self, data):
         df = pd.DataFrame({"A": [1] * len(data)}, dtype=object)
         df['A'] = data
         assert df.dtypes['A'] == data.dtype
+
+    def test_merge(self, data, na_value):
```
jreback: you prob should test with with the ``how=join_type`` fixture.

jreback (changes requested):

jreback (approved):

pandas/tests/extension/base/reshaping.py
jreback: ok, its worth doing way more tests than a single usecase. but ok here I guess.

PR #20746, opened 2018-04-19
TST: correct constructor in extension array tests

jorisvandenbossche:
Replace `type(data)(..)` with `data._constructor_from_sequence`. 

A bit annoying the name is that long :-), but this is the correct usage.

cc @TomAugspurger


TomAugspurger (approved):
Thanks.

I've grown to dislike the name `_constructor_from_sequence` even more in the interim. It sounds like the method is returning constructor, which would then be used to construct the array. Instead it's creating a new array from a sequence of scalars (so `_from_scalars`...)

Is it worth changing before it's too late?

jreback:
why dont we rename to ``_from_sequence``?

jorisvandenbossche:
I think we can certainly leave out the `constructor` part. For `_from_scalars` or `_from_sequence`, the doesn't indicate it is from a list/array of scalars, and the second does not indicate that it is from the dtype's scalar objects. 
I personally have a preference for `_from_scalars` because I find the 'scalar' the more important information of the two, but `_from_sequence` is certainly already an improvement compared to the current name.

TomAugspurger:
Either of those are fine. Slight preference for `_from_scalars`, for the
reasons Joris outlined. But from the context, I think it's clear enough
what `_from_sequence` is doing.

jreback:
let's do the rename to ``_from_sequence`` in this PR (and update docs etc)

TomAugspurger:
The Appveyor failure was an HTTP issue. @jorisvandenbossche you can merge whenever.

PR #20749, opened 2018-04-19
DOC: Clean up badges in README

wuhaochen:
Since we have pypi and conda covered in latest release and the links to them are no longer available. It probably make sense to remove these two rows.


jreback:
thanks!

Issue #20720, opened 2018-04-17
TST: failure on s3 mocking


jreback:
master seems to fail recently: https://travis-ci.org/pandas-dev/pandas/jobs/367585545

on ``pandas/tests/iotest_parquet.py -k s3``.

works locally for me. very odd as this is mocked and shouldn't ever have a connection error.

In particular xfailing:  https://github.com/pandas-dev/pandas/commit/336fba7c0191444c3328009e6d4f9f5d00ee224b
- [ ] ``test_read_csv_chunked_download`` in ``pandas.tests.io.test_network.py``
- [ ] ``test_s3_roundtrip`` for fp & pyarrow in ``pandas.testsio.test_parquet.py``

jreback:
@TomAugspurger any idea?

TomAugspurger:
cc @mdurant, have you been dealing with moto / requests issues recently?

martindurant:
The references issue causes anything based on requests to fail following the use of moto in some cases. In this particular case, though, the test **is** using moto, right?

The following fixed the problem with requests: https://github.com/dask/dask/blob/master/dask/bytes/tests/test_s3.py#L46

spulec:
Hi all; I'm one of the maintainers of Moto.

Our most recent version had a regex bug in catching certain S3 urls to patch. We are releasing a new version in the next couple minutes that should resolve any issues (`1.3.3`).

Feel free to ping me if these issues don't seem to go away.

jreback:
thanks @spulec!

TomAugspurger:
Thanks @spulec, I can confirm that upgrading to 1.3.3 fixed things for me locally!

We still have some failures on our CI, though I *think* that this is fully pandas problem now: https://travis-ci.org/TomAugspurger/pandas/jobs/369087870. Working on it now.

spulec:
Okay, let me know if it does end up pointing back to Moto.

csolamx:
Hi @spulec!  I installed 1.3.5 today and it seems that it's mocking any requests not belonging to AWS.  Are there other updates to consider or workarounds?

Thanks!

spulec:
@csolamx feel free to follow along at https://github.com/spulec/moto/issues/1567

PR #20767, opened 2018-04-20
Use conda-env for CI environments

TomAugspurger:
Closes https://github.com/pandas-dev/pandas/issues/20720

The old setup where things were installed in multiple places resulted in some strange environments, with multiple copies of the same package in `site-packages`.

This attempts to reduce the environment definition to a single file per CI job, and the environment installation to 3 steps

- create the environment
- uninstall pandas if it's present (installing a pandas dependency)
- build pandas from source & install

Travis only for now. Will move over the appveyor / circle ones too.


TomAugspurger:
Just one failure, on the OSX build. That's pinned to numpy 1.10.4, which forces fastparquet 0.1.3, which is not compatible with thrift 0.11. I could pin fastparquet & thrift to those versions, but I think fastparquet is adequately tested elsewhere, so I've removed it.

TomAugspurger:
Updated for circle-ci and appveyor.

AFAICT, the `requirements-3.5.*` files are unused. Does that look correct?

TomAugspurger:
OK, passing now.

cc @jreback @jorisvandenbossche.

I verified that that Travis tests all have the same number of skips / passes. The only difference is https://travis-ci.org/pandas-dev/pandas/jobs/369333014#L2982, which apparently isn't run on PRs.

ci/travis-36-numpydev.yaml
```diff
@@ -0,0 +1,18 @@
+name: pandas
+channels:
```
jreback: it seems that you added conda-forge to virtually all recipes. We did not have that before. See if you can remove it from a few (IOW just defaults)

ci/travis-36-numpydev.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - python=3.6*
+  - pytz
+  - Cython
+  # universal
+  - pytest
+  - pytest-xdist
+  - moto
+  - pip:
+    - "git+git://github.com/dateutil/dateutil.git"
+    - "-f https://7933911d6844c6c53a7d-47bd50c35cd79bd838daf386af554a83.ssl.cf2.rackcdn.com"
```
jreback: this actually works in this syntax?

jreback (changes requested):

ci/travis-36-numpydev.yaml
TomAugspurger: Will do. I needed it for moto, but it shouldn’t be required / used here, since s3fs isn’t installed.

ci/travis-36-numpydev.yaml
TomAugspurger: Surprisingly, yes! This is valid syntax for requirements.txt, which is all that conda is doing.

ci/travis-27.yaml
```diff
@@ -0,0 +1,49 @@
+name: pandas
+channels:
```
jreback: can we have one of the main builds NOT use conda-forge (maybe the 3.5 one)? in the original we had *only* the 3.6 build using it.

TomAugspurger:
conda-forge is only used for packages that are not found in the main
channel. The 3.5-osx env uses s3fs, moto, and feather which aren't in main.

jreback:
> conda-forge is only used for packages that are not found in the main
channel. The 3.5-osx env uses s3fs, moto, and feather which aren't in main.

I understand. my point is in the existing setup, we install from defaults ONLY for most packages. I am loath to give that up on at least 1 build. I would either go back to 2 steps on 1 or more builds, or remove these deps (e.g. you can install from pip instead these packages).

jreback:
I think you can completely remove the JOB= env variable as well here.

jreback:
also I don't think the issue number is correct here, #20720 ?

TomAugspurger:
> you can install from pip instead these packages

I'd like to avoid installing moto from pip if possible. It brings in a lot of dependencies, which aren't managed by conda.

This closes #20720. It looks like that was caused by the 2 / 3 stage installs, since we ended up with multiple versions of requests, boto, etc., in the same environment. Now we just have the one. I haven't seen any of the network failures on this branch.

ci/install_travis.sh
```diff
@@ -90,55 +80,10 @@ echo
...
-
-time conda install -n pandas pytest>=3.1.0
-time pip install -q pytest-xdist moto
-
-if [ "$LINT" ]; then
-   conda install flake8=3.4.1
-   pip install cpplint
-fi
-
-if [ "$COVERAGE" ]; then
```
jreback: where did coverage go?

jreback:
ok I recently added some xfails for #20720, can you remove in this PR?

ci/install_travis.sh
TomAugspurger: The coverage deps are run on the 3.6 build. `travis-36.yaml` has the additional coverage deps.

jorisvandenbossche:
Does the `conda-forge::packgage` syntax work for environment files? 
That might be another way to install some packages from conda-forge in an environment without adding conda-forge as a full channel (as then it might be more difficult to control which ones are installed from conda-forge / defaults)

jorisvandenbossche:
Small note: I personally liked the JOB tags to have a quick overview of the different build types. I know the ENV_FILE tag is similarly unique, but a bit less user friendly to read or to refer to ("check travis job ...")

TomAugspurger:
`Does the conda-forge::packgage syntax work for environment files?`

Not correctly unfortunately :( When I tried that, all the packages came from conda-forge. I suspect conda parses the deps for channels, and if found adds them to the start of the package list.  Opened https://github.com/conda/conda/issues/7202 for that.

TomAugspurger:
Reverted the `JOB` removal.

jreback:
> Small note: I personally liked the JOB tags to have a quick overview of the different build types. I know the ENV_FILE tag is similarly unique, but a bit less user friendly to read or to refer to ("check travis job ...")

I don't have a problem with the slightly longer ENV_FILE, maybe just rename this to JOB?
having the duplicated ones is a bit annoying

jorisvandenbossche:
Or maybe like before where the JOB tag is the variable part in the file name?

TomAugspurger:
By file name, do you mean the environment.yaml? Because if so, I think
that's the case.

I don't really care what we do here. In general, I like the idea of a small
tag that identifies the relevant bit for each build. Now that we aren't
using the `{.build,run,pip,sh}` files, we're free to put whatever we want
in `JOB`.

I removed the unused 3.5 files.

jorisvandenbossche:
> By file name, do you mean the environment.yaml? Because if so, I think that's the case.

Yes, that is what I meant. All environment files for travis are like `ci/travis-specifics.yml`, so JOB could just indicate 'specifics' (which would almost be identical to the current job tag, except for casing) to avoid duplication between JOB and ENV_FILE ?

TomAugspurger:
Re-pushed with more informative job tags. Now they include things like indicating which one is linting, running coverage, etc.

screen shot 2018-04-23 at 1 02 14 pm

Easier to tell at a glance which ones are special.

jreback:
lgtm

merge away

TomAugspurger:
FYI, the `channel::package` support in environment.yaml was recently added to conda in https://github.com/conda/conda/pull/7178

I tried it out with conda-canary and things seemed to work fine.

PR #15504, opened 2017-02-25
PERF: improve iloc list indexing

jorisvandenbossche:
Did a profile of `iloc` using a list indexer, and noticed a few parts that seemingly can be optimized rather easily:

- convert early to array, and keep as array
- do not validate the input for out of bounds (this validation is also (and much faster) done in `np.take`)
- do not validate kwargs (for numpy compat) (-> could define a `_take` method that does not do this for internal usage?)
- use fastpath for resulting Series creation

The failing tests are currently just the ones that checked for the kwarg validation (and one failing test due to the fastpath taken where object datetimes are not coerced to datetime64).

Compared to master, I get up to 40 to 60% improvement on some simple tests:

PR:
```
In [1]: s = pd.Series(np.random.randn(10000))

In [2]: s.iloc[10]
Out[2]: 0.91268983379322421

In [3]: %timeit s.iloc[[10]]
The slowest run took 14.18 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 47.3 µs per loop

In [4]: indexer = list(range(10, 30)) + list(range(50,80))

In [5]: %timeit s.iloc[indexer]
The slowest run took 5.92 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 49.8 µs per loop

In [6]: indexer = list(range(10, 30)) + list(range(50,800))

In [7]: %timeit s.iloc[indexer]
10000 loops, best of 3: 104 µs per loop
```

master:
```
In [1]: s = pd.Series(np.random.randn(10000))

In [2]: s.iloc[10]
Out[2]: -1.4152272792497846

In [3]: %timeit s.iloc[[10]]
The slowest run took 6.70 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 103 µs per loop

In [4]: indexer = list(range(10, 30)) + list(range(50,80))

In [5]: %timeit s.iloc[indexer]
The slowest run took 4.12 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 131 µs per loop

In [6]: indexer = list(range(10, 30)) + list(range(50,800))

In [7]: %timeit s.iloc[indexer]
The slowest run took 17.01 times longer than the fastest. This could mean that an intermediate result is being cached.
1000 loops, best of 3: 247 µs per loop
```

There are some details to work out, but already putting it up here for discussion (and need to run the benchmark suite).

[...]

PR #20770, opened 2018-04-20
Emit warning for missing labels in Multiindex.loc[[...]] (and more)

### LINKED ISSUE #17758: No warning is raised by MultiIndex when .loc is called with list containing missing keys

From https://github.com/pandas-dev/pandas/pull/17295#issuecomment-324014362
```python
In [2]: pd.DataFrame([[i] for i in range(3)], index=pd.MultiIndex.from_product([[1], [5,6,7]])).loc[[(1,5), (3,7)]]
Out[2]: 
       0
1 5  0.0
3 7  NaN
```
#### Problem description

#17295 introduced a warning when ``obj.loc[a_list]`` is called with ``a_list`` containing keys not in ``obj.index``, but only applies to flat ``Index``es. ``MultiIndex``es need an analogous PR.

#### Expected Output

The same, but with a ``FutureWarning``.

### LINKED ISSUE #20748: .loc[iterator] treats missing keys differently than .loc[list]

```python
In [2]: pd.Series([1,2,3]).loc[[i for i in (4,5)]]
KeyError                                  Traceback (most recent call last)
<ipython-input-2-aba9eb9eea33> in <module>()
----> 1 pd.Series([1,2,3]).loc[[i for i in (4,5)]]

/home/nobackup/repo/pandas/pandas/core/indexing.py in __getitem__(self, key)
   1370 
   1371             maybe_callable = com._apply_if_callable(key, self.obj)
-> 1372             return self._getitem_axis(maybe_callable, axis=axis)
   1373 
   1374     def _is_scalar_access(self, key):

/home/nobackup/repo/pandas/pandas/core/indexing.py in _getitem_axis(self, key, axis)
   1829                     raise ValueError('Cannot index with multidimensional key')
   1830 
-> 1831                 return self._getitem_iterable(key, axis=axis)
   1832 
   1833             # nested tuple slicing

/home/nobackup/repo/pandas/pandas/core/indexing.py in _getitem_iterable(self, key, axis)
   1109 
   1110         if self._should_validate_iterable(axis):
-> 1111             self._has_valid_type(key, axis)
   1112 
   1113         labels = self.obj._get_axis(axis)

/home/nobackup/repo/pandas/pandas/core/indexing.py in _has_valid_type(self, key, axis)
   1683                         raise KeyError(
   1684                             u"None of [{key}] are in the [{axis}]".format(
-> 1685                                 key=key, axis=self.obj._get_axis_name(axis)))
   1686                     else:
   1687 

KeyError: 'None of [[4, 5]] are in the [index]'

In [3]: pd.Series([1,2,3]).loc[(i for i in (4,5))]
Out[3]: 
4   NaN
5   NaN
dtype: float64


```
#### Problem description

Since we convert iterators to lists anyway...

https://github.com/pandas-dev/pandas/blob/1e4e04bf47417aadaf11c7d55c206508f2899fa5/pandas/core/indexing.py#L1124

... we might as well do the conversion as soon as possible (i.e., in ``__getitem__``), and simplify the code by only handling list-likes which have a length. I would also consider changing ``is_list_like`` to return ``False`` for iterators, or provide it with a ``has_len=False`` argument.

It would also solve this other, less important, difference:
```
In [2]: pd.Series([1,2,3]).loc[[i for i in (2,5)]]
/usr/bin/ipython3:1: FutureWarning: 
Passing list-likes to .loc or [] with any missing label will raise
KeyError in the future, you can use .reindex() as an alternative.

See the documentation here:
https://pandas.pydata.org/pandas-docs/stable/indexing.html#deprecate-loc-reindex-listlike
  #! /bin/sh
Out[2]: 
2    3.0
5    NaN
dtype: float64

In [3]: pd.Series([1,2,3]).loc[(i for i in (2,5))]
Out[3]: 
2    3.0
5    NaN
dtype: float64
```

... and probably others.

#### Expected Output

Exactly the same for lists and iterators.

<details>

INSTALLED VERSIONS
------------------
commit: d04b7464dcc20051ef38ac2acda580de854d3e01
python: 3.5.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.9.0-6-amd64
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: it_IT.UTF-8
LOCALE: it_IT.UTF-8

pandas: 0.23.0.dev0+754.gd04b7464d.dirty
pytest: 3.5.0
pip: 9.0.1
setuptools: 39.0.1
Cython: 0.25.2
numpy: 1.14.1
scipy: 0.19.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: 1.5.6
patsy: 0.5.0
dateutil: 2.7.0
pytz: 2017.2
blosc: None
bottleneck: 1.2.0dev
tables: 3.3.0
numexpr: 2.6.1
feather: 0.3.1
matplotlib: 2.0.0
openpyxl: 2.3.0
xlrd: 1.0.0
xlwt: 1.3.0
xlsxwriter: 0.9.6
lxml: 4.1.1
bs4: 4.5.3
html5lib: 0.999999999
sqlalchemy: 1.0.15
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: 0.2.1

</details>

### LINKED ISSUE #20753: Inconsistent behaviour of .ix between list and scalar key with (missing) ints

```python
In [2]: pd.Series(range(3)).ix[5]
/usr/bin/ipython3:1: DeprecationWarning: 
.ix is deprecated. Please use
.loc for label based indexing or
.iloc for positional indexing

See the documentation here:
http://pandas.pydata.org/pandas-docs/stable/indexing.html#ix-indexer-is-deprecated
  #! /bin/sh
KeyError                                  Traceback (most recent call last)
/home/nobackup/repo/pandas/pandas/core/indexes/base.py in get_loc(self, key, method, tolerance)
   3017             try:
-> 3018                 return self._engine.get_loc(key)
   3019             except KeyError:

/home/nobackup/repo/pandas/pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc (pandas/_libs/index.c:5729)()
    138 
--> 139     cpdef get_loc(self, object val):
    140         if is_definitely_invalid_key(val):

/home/nobackup/repo/pandas/pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc (pandas/_libs/index.c:5575)()
    160         try:
--> 161             return self.mapping.get_item(val)
    162         except (TypeError, ValueError):

/home/nobackup/repo/pandas/pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Int64HashTable.get_item (pandas/_libs/hashtable.c:15398)()
    957 
--> 958     cpdef get_item(self, int64_t val):
    959         cdef khiter_t k

/home/nobackup/repo/pandas/pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Int64HashTable.get_item (pandas/_libs/hashtable.c:15342)()
    963         else:
--> 964             raise KeyError(val)
    965 

KeyError: 5

KeyError                                  Traceback (most recent call last)
/home/nobackup/repo/pandas/pandas/core/indexing.py in _get_label(self, label, axis)
    131             try:
--> 132                 return self.obj._xs(label, axis=axis)
    133             except:

/home/nobackup/repo/pandas/pandas/core/generic.py in xs(self, key, axis, level, drop_level)
   2991         else:
-> 2992             loc = self.index.get_loc(key)
   2993 

/home/nobackup/repo/pandas/pandas/core/indexes/base.py in get_loc(self, key, method, tolerance)
   3019             except KeyError:
-> 3020                 return self._engine.get_loc(self._maybe_cast_indexer(key))
   3021 

/home/nobackup/repo/pandas/pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc (pandas/_libs/index.c:5729)()
    138 
--> 139     cpdef get_loc(self, object val):
    140         if is_definitely_invalid_key(val):

/home/nobackup/repo/pandas/pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc (pandas/_libs/index.c:5575)()
    160         try:
--> 161             return self.mapping.get_item(val)
    162         except (TypeError, ValueError):

/home/nobackup/repo/pandas/pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Int64HashTable.get_item (pandas/_libs/hashtable.c:15398)()
    957 
--> 958     cpdef get_item(self, int64_t val):
    959         cdef khiter_t k

/home/nobackup/repo/pandas/pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Int64HashTable.get_item (pandas/_libs/hashtable.c:15342)()
    963         else:
--> 964             raise KeyError(val)
    965 

KeyError: 5

KeyError                                  Traceback (most recent call last)
<ipython-input-2-bba00cd402c6> in <module>()
----> 1 pd.Series(range(3)).ix[5]

/home/nobackup/repo/pandas/pandas/core/indexing.py in __getitem__(self, key)
    118 
    119             key = com._apply_if_callable(key, self.obj)
--> 120             return self._getitem_axis(key, axis=axis)
    121 
    122     def _get_label(self, label, axis=None):

/home/nobackup/repo/pandas/pandas/core/indexing.py in _getitem_axis(self, key, axis)
   1104                     return self._get_loc(key, axis=axis)
   1105 
-> 1106             return self._get_label(key, axis=axis)
   1107 
   1108     def _getitem_iterable(self, key, axis=None):

/home/nobackup/repo/pandas/pandas/core/indexing.py in _get_label(self, label, axis)
    132                 return self.obj._xs(label, axis=axis)
    133             except:
--> 134                 return self.obj[label]
    135         elif isinstance(label, tuple) and isinstance(label[axis], slice):
    136             raise IndexingError('no slices here, handle elsewhere')

/home/nobackup/repo/pandas/pandas/core/series.py in __getitem__(self, key)
    753         key = com._apply_if_callable(key, self)
    754         try:
--> 755             result = self.index.get_value(self, key)
    756 
    757             if not is_scalar(result):

/home/nobackup/repo/pandas/pandas/core/indexes/base.py in get_value(self, series, key)
   3051         try:
   3052             return self._engine.get_value(s, k,
-> 3053                                           tz=getattr(series.dtype, 'tz', None))
   3054         except KeyError as e1:
   3055             if len(self) > 0 and self.inferred_type in ['integer', 'boolean']:

/home/nobackup/repo/pandas/pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_value (pandas/_libs/index.c:4856)()
    103         return val in self.mapping
    104 
--> 105     cpdef get_value(self, ndarray arr, object key, object tz=None):
    106         """
    107         arr : 1-dimensional ndarray

/home/nobackup/repo/pandas/pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_value (pandas/_libs/index.c:4539)()
    111             void* data_ptr
    112 
--> 113         loc = self.get_loc(key)
    114         if PySlice_Check(loc) or cnp.PyArray_Check(loc):
    115             return arr[loc]

/home/nobackup/repo/pandas/pandas/_libs/index.pyx in pandas._libs.index.IndexEngine.get_loc (pandas/_libs/index.c:5575)()
    159 
    160         try:
--> 161             return self.mapping.get_item(val)
    162         except (TypeError, ValueError):
    163             raise KeyError(val)

/home/nobackup/repo/pandas/pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Int64HashTable.get_item (pandas/_libs/hashtable.c:15398)()
    956                                        sizeof(uint32_t)) # flags
    957 
--> 958     cpdef get_item(self, int64_t val):
    959         cdef khiter_t k
    960         k = kh_get_int64(self.table, val)

/home/nobackup/repo/pandas/pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Int64HashTable.get_item (pandas/_libs/hashtable.c:15342)()
    962             return self.table.vals[k]
    963         else:
--> 964             raise KeyError(val)
    965 
    966     cpdef set_item(self, int64_t key, Py_ssize_t val):

KeyError: 5

In [3]: pd.Series(range(3)).ix[[5]]
/usr/bin/ipython3:1: DeprecationWarning: 
.ix is deprecated. Please use
.loc for label based indexing or
.iloc for positional indexing

See the documentation here:
http://pandas.pydata.org/pandas-docs/stable/indexing.html#ix-indexer-is-deprecated
  #! /bin/sh
Out[3]: 
5   NaN
dtype: float64

```

#### Problem description

Already fixed, opening the issue just for xref from the test/whatsnew.

#### Expected Output

An error

<details>

INSTALLED VERSIONS
------------------
commit: d04b7464dcc20051ef38ac2acda580de854d3e01
python: 3.5.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.9.0-6-amd64
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: it_IT.UTF-8
LOCALE: it_IT.UTF-8

pandas: 0.23.0.dev0+754.gd04b7464d
pytest: 3.5.0
pip: 9.0.1
setuptools: 39.0.1
Cython: 0.25.2
numpy: 1.14.1
scipy: 0.19.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: 1.5.6
patsy: 0.5.0
dateutil: 2.7.0
pytz: 2017.2
blosc: None
bottleneck: 1.2.0dev
tables: 3.3.0
numexpr: 2.6.1
feather: 0.3.1
matplotlib: 2.0.0
openpyxl: 2.3.0
xlrd: 1.0.0
xlwt: 1.3.0
xlsxwriter: 0.9.6
lxml: 4.1.1
bs4: 4.5.3
html5lib: 0.999999999
sqlalchemy: 1.0.15
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: 0.2.1

</details>

toobaz:
- [x] closes #17758
- [x] closes #20748
- [x] closes #20753
- [x] tests added / passed
- [x] whatsnew entry

Main changes:
- the presence of at least one key (and of all of them, for the temporary warning) is now checked _after_ building the indexer, using the indexer itself, in a unique place
- for this to be possible, ``obj.reindex`` was replaced with ``obj._reindex_with_indexers``
- ``_multi_take`` is uglier than before... but it's ugly anyway, and must be removed in a future refactoring (in which first indexers are built in all possible code paths, _then_ values are extracted)
- ``_has_valid_type`` was mostly used in "assertion mode" (without looking at its return value), so I changed it to ``_validate_key`` and now it is _only_ used in "assertion mode"

Asv (``--bench indexing``) give
```
     [3a2e9e6c]       [25711d3d]
+       162±0.2ms          273±5ms     1.68  indexing.NumericSeriesIndexing.time_getitem_array(<class 'pandas.core.indexes.numeric.Float64Index'>)
+         162±1ms        269±0.7ms     1.66  indexing.NumericSeriesIndexing.time_loc_array(<class 'pandas.core.indexes.numeric.Float64Index'>)
+       162±0.6ms          267±1ms     1.65  indexing.NumericSeriesIndexing.time_ix_list_like(<class 'pandas.core.indexes.numeric.Float64Index'>)
+         164±2ms          270±1ms     1.64  indexing.NumericSeriesIndexing.time_ix_array(<class 'pandas.core.indexes.numeric.Float64Index'>)
+       166±0.6ms        273±0.6ms     1.64  indexing.NumericSeriesIndexing.time_getitem_lists(<class 'pandas.core.indexes.numeric.Float64Index'>)
+       163±0.5ms        267±0.8ms     1.64  indexing.NumericSeriesIndexing.time_getitem_list_like(<class 'pandas.core.indexes.numeric.Float64Index'>)
+         163±3ms        267±0.4ms     1.64  indexing.NumericSeriesIndexing.time_loc_list_like(<class 'pandas.core.indexes.numeric.Float64Index'>)
+         412±3μs          664±4μs     1.61  indexing.NumericSeriesIndexing.time_ix_list_like(<class 'pandas.core.indexes.numeric.Int64Index'>)
+        1.08±0ms      1.42±0.05ms     1.32  indexing.NumericSeriesIndexing.time_ix_array(<class 'pandas.core.indexes.numeric.Int64Index'>)
+     1.14±0.01ms      1.31±0.02ms     1.15  indexing.PanelIndexing.time_subset
-       133±0.2ms        121±0.4ms     0.91  frame_methods.Iteration.time_iteritems_indexing
-         442±2ns          399±2ns     0.90  indexing.MethodLookup.time_lookup_iloc
-       145±0.8μs        130±0.8μs     0.90  indexing.IntervalIndexing.time_getitem_list
-         151±2μs          134±1μs     0.89  indexing.AssignTimeseriesIndex.time_frame_assign_timeseries_index
-      48.5±0.8μs       41.2±0.1μs     0.85  indexing.IntervalIndexing.time_getitem_scalar
-        44.3±1μs       35.6±0.1μs     0.80  indexing.NumericSeriesIndexing.time_iloc_list_like(<class 'pandas.core.indexes.numeric.Float64Index'>)
-     3.81±0.02ms      2.06±0.02ms     0.54  indexing.DataFrameNumericIndexing.time_loc_dups

```

notice that ``FloatIndex`` was _not_ being checked at all (i.e. #17758 applied to it too), now it is.

In any case, as I wrote above, there are several improvements still to be made, but I want to go gradually because the complexity of this PR is already pretty high for me.

### DIFF
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -873,6 +873,7 @@ Deprecations
 - The ``is_copy`` attribute is deprecated and will be removed in a future version (:issue:`18801`).
 - ``IntervalIndex.from_intervals`` is deprecated in favor of the :class:`IntervalIndex` constructor (:issue:`19263`)
 - ``DataFrame.from_items`` is deprecated. Use :func:`DataFrame.from_dict` instead, or ``DataFrame.from_dict(OrderedDict())`` if you wish to preserve the key order (:issue:`17320`, :issue:`17312`)
+- Indexing a ``MultiIndex`` or a ``FloatIndex`` with a list containing some missing keys will now show a ``FutureWarning``, consistently with other types of indexes (:issue:`17758`).
 
 - The ``broadcast`` parameter of ``.apply()`` is deprecated in favor of ``result_type='broadcast'`` (:issue:`18577`)
 - The ``reduce`` parameter of ``.apply()`` is deprecated in favor of ``result_type='reduce'`` (:issue:`18577`)
@@ -1107,6 +1108,8 @@ Indexing
 - :func:`Index.to_series` now accepts ``index`` and ``name`` kwargs (:issue:`18699`)
 - :func:`DatetimeIndex.to_series` now accepts ``index`` and ``name`` kwargs (:issue:`18699`)
 - Bug in indexing non-scalar value from ``Series`` having non-unique ``Index`` will return value flattened (:issue:`17610`)
+- Bug in indexing with iterator containing only missing keys, which raised no error (:issue:`20748`)
+- Fixed inconsistency in ``.ix`` between list and scalar keys when index has integer dtype and does not include desired key(s) (:issue:`20753`)
 - Bug in ``__setitem__`` when indexing a :class:`DataFrame` with a 2-d boolean ndarray (:issue:`18582`)
 - Bug in ``str.extractall`` when there were no matches empty :class:`Index` was returned instead of appropriate :class:`MultiIndex` (:issue:`19034`)
 - Bug in :class:`IntervalIndex` where empty and purely NA data was constructed inconsistently depending on the construction method (:issue:`18421`)
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -873,7 +873,7 @@ Deprecations
 - The ``is_copy`` attribute is deprecated and will be removed in a future version (:issue:`18801`).
 - ``IntervalIndex.from_intervals`` is deprecated in favor of the :class:`IntervalIndex` constructor (:issue:`19263`)
 - ``DataFrame.from_items`` is deprecated. Use :func:`DataFrame.from_dict` instead, or ``DataFrame.from_dict(OrderedDict())`` if you wish to preserve the key order (:issue:`17320`, :issue:`17312`)
-- Indexing a ``MultiIndex`` or a ``FloatIndex`` with a list containing some missing keys will now show a ``FutureWarning``, consistently with other types of indexes (:issue:`17758`).
+- Indexing a :class:`MultiIndex` or a :class:`FloatIndex` with a list containing some missing keys will now show a :class:`FutureWarning`, which is consistent with other types of indexes (:issue:`17758`).
 
 - The ``broadcast`` parameter of ``.apply()`` is deprecated in favor of ``result_type='broadcast'`` (:issue:`18577`)
 - The ``reduce`` parameter of ``.apply()`` is deprecated in favor of ``result_type='reduce'`` (:issue:`18577`)
@@ -1109,7 +1109,7 @@ Indexing
 - :func:`DatetimeIndex.to_series` now accepts ``index`` and ``name`` kwargs (:issue:`18699`)
 - Bug in indexing non-scalar value from ``Series`` having non-unique ``Index`` will return value flattened (:issue:`17610`)
 - Bug in indexing with iterator containing only missing keys, which raised no error (:issue:`20748`)
-- Fixed inconsistency in ``.ix`` between list and scalar keys when index has integer dtype and does not include desired key(s) (:issue:`20753`)
+- Fixed inconsistency in ``.ix`` between list and scalar keys when the index has integer dtype and does not include the desired keys (:issue:`20753`)
 - Bug in ``__setitem__`` when indexing a :class:`DataFrame` with a 2-d boolean ndarray (:issue:`18582`)
 - Bug in ``str.extractall`` when there were no matches empty :class:`Index` was returned instead of appropriate :class:`MultiIndex` (:issue:`19034`)
 - Bug in :class:`IntervalIndex` where empty and purely NA data was constructed inconsistently depending on the construction method (:issue:`18421`)
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -4881,6 +4881,9 @@ def _ensure_index(index_like, copy=False):
     if hasattr(index_like, 'name'):
         return Index(index_like, name=index_like.name, copy=copy)
 
+    if is_iterator(index_like):
+        index_like = list(index_like)
+
     # must check for exactly list here because of strict type
     # check in clean_index_list
     if isinstance(index_like, list):
--- a/pandas/core/indexing.py
+++ b/pandas/core/indexing.py
@@ -186,7 +186,7 @@ class _NDFrameIndexer(_NDFrameIndexerBase):
         indexer = self._get_setitem_indexer(key)
         self._setitem_with_indexer(indexer, value)
 
-    def _has_valid_type(self, k, axis):
+    def _validate_key(self, k, axis):
         raise NotImplementedError()
 
     def _has_valid_tuple(self, key):
@@ -194,24 +194,12 @@ class _NDFrameIndexer(_NDFrameIndexerBase):
         for i, k in enumerate(key):
             if i >= self.obj.ndim:
                 raise IndexingError('Too many indexers')
-            if not self._has_valid_type(k, i):
+            try:
+                self._validate_key(k, i)
+            except ValueError:
                 raise ValueError("Location based indexing can only have "
                                  "[{types}] types"
                                  .format(types=self._valid_types))
-
-    def _should_validate_iterable(self, axis=None):
-        """ return a boolean whether this axes needs validation for a passed
-        iterable
-        """
-        if axis is None:
-            axis = self.axis or 0
-        ax = self.obj._get_axis(axis)
-        if isinstance(ax, MultiIndex):
-            return False
-        elif ax.is_floating():
-            return False
-
-        return True
 
     def _is_nested_tuple_indexer(self, tup):
         if any(isinstance(ax, MultiIndex) for ax in self.obj.axes):
@@ -906,38 +894,31 @@ class _NDFrameIndexer(_NDFrameIndexerBase):
         """
         try:
             o = self.obj
-            d = {a: self._convert_for_reindex(t, axis=o._get_axis_number(a))
-                 for t, a in zip(tup, o._AXIS_ORDERS)}
-            return o.reindex(**d)
-        except(KeyError, IndexingError):
-            raise self._exception
+            d = {}
+            for key, axis in zip(tup, o._AXIS_ORDERS):
+                ax = o._get_axis(axis)
+                indexer, keyarr = ax._convert_listlike_indexer(key,
+                                                               kind=self.name)
+                if indexer is not None and (indexer != -1).all():
+                    self._validate_read_indexer(key, indexer, axis)
+                    d[axis] = (ax[indexer], indexer)
+                    continue
+
+                # If we are trying to get actual keys from empty Series, we
+                # patiently wait for a KeyError later on - otherwise, convert
+                if len(ax) or not len(key):
+                    key = self._convert_for_reindex(key, axis)
+                indexer = ax.get_indexer_for(key)
+                keyarr = ax.reindex(keyarr)[0]
+                self._validate_read_indexer(keyarr, indexer,
+                                            o._get_axis_number(axis))
+                d[axis] = (keyarr, indexer)
+            return o._reindex_with_indexers(d, copy=True, allow_dups=True)
+        except(KeyError, IndexingError) as detail:
+            raise self._exception(detail)
 
     def _convert_for_reindex(self, key, axis=None):
-        if axis is None:
-            axis = self.axis or 0
-        labels = self.obj._get_axis(axis)
-
-        if com.is_bool_indexer(key):
-            key = check_bool_indexer(labels, key)
-            return labels[key]
-        else:
-            if isinstance(key, Index):
-                keyarr = labels._convert_index_indexer(key)
-            else:
-                # asarray can be unsafe, NumPy strings are weird
-                keyarr = com._asarray_tuplesafe(key)
-
-            if is_integer_dtype(keyarr):
-                # Cast the indexer to uint64 if possible so
-                # that the values returned from indexing are
-                # also uint64.
-                keyarr = labels._convert_arr_indexer(keyarr)
-
-                if not labels.is_integer():
-                    keyarr = _ensure_platform_int(keyarr)
-                    return labels.take(keyarr)
-
-            return keyarr
+        return key
 
     def _handle_lowerdim_multi_index_axis0(self, tup):
         # we have an axis0 multi-index, handle or raise
@@ -1073,8 +1054,9 @@ class _NDFrameIndexer(_NDFrameIndexerBase):
         if axis is None:
             axis = self.axis or 0
 
-        if self._should_validate_iterable(axis):
-            self._has_valid_type(key, axis)
+        if is_iterator(key):
+            key = list(key)
+        self._validate_key(key, axis)
 
         labels = self.obj._get_axis(axis)
         if isinstance(key, slice):
@@ -1110,8 +1092,7 @@ class _NDFrameIndexer(_NDFrameIndexerBase):
         if axis is None:
             axis = self.axis or 0
 
-        if self._should_validate_iterable(axis):
-            self._has_valid_type(key, axis)
+        self._validate_key(key, axis)
 
         labels = self.obj._get_axis(axis)
 
@@ -1125,19 +1106,17 @@ class _NDFrameIndexer(_NDFrameIndexerBase):
             indexer, keyarr = labels._convert_listlike_indexer(
                 key, kind=self.name)
             if indexer is not None and (indexer != -1).all():
+                self._validate_read_indexer(key, indexer, axis)
                 return self.obj.take(indexer, axis=axis)
 
+            ax = self.obj._get_axis(axis)
             # existing labels are unique and indexer are unique
             if labels.is_unique and Index(keyarr).is_unique:
+                self._validate_read_indexer(key, ax.get_indexer_for(key), axis)
 
-                try:
-                    return self.obj.reindex(keyarr, axis=axis)
-                except AttributeError:
-
-                    # Series
-                    if axis != 0:
-                        raise AssertionError('axis must be 0')
-                    return self.obj.reindex(keyarr)
+                d = {axis: [ax.reindex(keyarr)[0], ax.get_indexer_for(key)]}
+                return self.obj._reindex_with_indexers(d, copy=True,
+                                                       allow_dups=True)
 
             # existing labels are non-unique
             else:
@@ -1154,14 +1133,52 @@ class _NDFrameIndexer(_NDFrameIndexerBase):
                     result = self.obj._take(indexer[indexer != -1], axis=axis,
                                             convert=False)
 
+                    self._validate_read_indexer(key, new_indexer, axis)
                     result = result._reindex_with_indexers(
                         {axis: [new_target, new_indexer]},
                         copy=True, allow_dups=True)
 
                 else:
+                    self._validate_read_indexer(key, indexer, axis)
                     result = self.obj._take(indexer, axis=axis)
 
                 return result
+
+    def _validate_read_indexer(self, key, indexer, axis):
+        """
+        Check that indexer is OK (e.g. at least one element was found, unless
+        the list of keys was actually empty).
+        """
+        ax = self.obj._get_axis(axis)
+        # True indicates missing values
+        if len(key) == 0:
+            return
+
+        missing = indexer < 0
+
+        if np.any(missing):
+            if np.all(missing):
+                raise KeyError(
+                    u"None of [{key}] are in the [{axis}]".format(
+                        key=key, axis=self.obj._get_axis_name(axis)))
+            else:
+
+                # we skip the warning on Categorical/Interval
+                # as this check is actually done (check for
+                # non-missing values), but a bit later in the
+                # code, so we want to avoid warning & then
+                # just raising
+
+                _missing_key_warning = textwrap.dedent("""
+                Passing list-likes to .loc or [] with any missing label will raise
+                KeyError in the future, you can use .reindex() as an alternative.
+
+                See the documentation here:
+                https://pandas.pydata.org/pandas-docs/stable/indexing.html#deprecate-loc-reindex-listlike""")  # noqa
+
+                if not (ax.is_categorical() or ax.is_interval()):
+                    warnings.warn(_missing_key_warning,
+                                  FutureWarning, stacklevel=5)
 
     def _convert_to_indexer(self, obj, axis=None, is_setter=False):
         """
@@ -1337,7 +1354,7 @@ class _IXIndexer(_NDFrameIndexer):
                       DeprecationWarning, stacklevel=2)
         super(_IXIndexer, self).__init__(name, obj)
 
-    def _has_valid_type(self, key, axis):
+    def _validate_key(self, key, axis):
         if isinstance(key, slice):
             return True
 
@@ -1352,6 +1369,33 @@ class _IXIndexer(_NDFrameIndexer):
             self._convert_scalar_indexer(key, axis)
 
         return True
+
+    def _convert_for_reindex(self, key, axis=None):
+        if axis is None:
+            axis = self.axis or 0
+        labels = self.obj._get_axis(axis)
+
+        if com.is_bool_indexer(key):
+            key = check_bool_indexer(labels, key)
+            return labels[key]
+        else:
+            if isinstance(key, Index):
+                keyarr = labels._convert_index_indexer(key)
+            else:
+                # asarray can be unsafe, NumPy strings are weird
+                keyarr = com._asarray_tuplesafe(key)
+
+            if is_integer_dtype(keyarr):
+                # Cast the indexer to uint64 if possible so
+                # that the values returned from indexing are
+                # also uint64.
+                keyarr = labels._convert_arr_indexer(keyarr)
+
+                if not labels.is_integer():
+                    keyarr = _ensure_platform_int(keyarr)
+                    return labels.take(keyarr)
+
+            return keyarr
 
 
 class _LocationIndexer(_NDFrameIndexer):
@@ -1656,7 +1700,7 @@ class _LocIndexer(_LocationIndexer):
                     "index is integers), listlike of labels, boolean")
     _exception = KeyError
 
-    def _has_valid_type(self, key, axis):
+    def _validate_key(self, key, axis):
         ax = self.obj._get_axis(axis)
 
         # valid for a label where all labels are in the index
@@ -1665,48 +1709,12 @@ class _LocIndexer(_LocationIndexer):
         # boolean
 
         if isinstance(key, slice):
-            return True
+            return
 
         elif com.is_bool_indexer(key):
-            return True
+            return
 
-        elif is_list_like_indexer(key):
-
-            # mi is just a passthru
-            if isinstance(key, tuple) and isinstance(ax, MultiIndex):
-                return True
-
-            if not is_iterator(key) and len(key):
-
-                # True indicates missing values
-                missing = ax.get_indexer_for(key) < 0
-
-                if np.any(missing):
-                    if len(key) == 1 or np.all(missing):
-                        raise KeyError(
-                            u"None of [{key}] are in the [{axis}]".format(
-                                key=key, axis=self.obj._get_axis_name(axis)))
-                    else:
-
-                        # we skip the warning on Categorical/Interval
-                        # as this check is actually done (check for
-                        # non-missing values), but a bit later in the
-                        # code, so we want to avoid warning & then
-                        # just raising
-                        _missing_key_warning = textwrap.dedent("""
-                        Passing list-likes to .loc or [] with any missing label will raise
-                        KeyError in the future, you can use .reindex() as an alternative.
-
-                        See the documentation here:
-                        https://pandas.pydata.org/pandas-docs/stable/indexing.html#deprecate-loc-reindex-listlike""")  # noqa
-
-                        if not (ax.is_categorical() or ax.is_interval()):
-                            warnings.warn(_missing_key_warning,
-                                          FutureWarning, stacklevel=5)
-
-            return True
-
-        else:
+        elif not is_list_like_indexer(key):
 
             def error():
                 if isna(key):
@@ -1728,8 +1736,6 @@ class _LocIndexer(_LocationIndexer):
                 raise
             except:
                 error()
-
-        return True
 
     def _is_scalar_access(self, key):
         # this is a shortcut accessor to both .loc and .iloc
@@ -1789,11 +1795,14 @@ class _LocIndexer(_LocationIndexer):
         if axis is None:
             axis = self.axis or 0
 
+        if is_iterator(key):
+            key = list(key)
+
         labels = self.obj._get_axis(axis)
         key = self._get_partial_string_timestamp_match_key(key, labels)
 
         if isinstance(key, slice):
-            self._has_valid_type(key, axis)
+            self._validate_key(key, axis)
             return self._get_slice_axis(key, axis=axis)
         elif com.is_bool_indexer(key):
             return self._getbool_axis(key, axis=axis)
@@ -1839,7 +1848,7 @@ class _LocIndexer(_LocationIndexer):
                 return self.obj.iloc[tuple(indexer)]
 
         # fall thru to straight lookup
-        self._has_valid_type(key, axis)
+        self._validate_key(key, axis)
         return self._get_label(key, axis=axis)
 
 
@@ -1871,7 +1880,7 @@ class _iLocIndexer(_LocationIndexer):
                     "point is EXCLUDED), listlike of integers, boolean array")
     _exception = IndexError
 
-    def _has_valid_type(self, key, axis):
+    def _validate_key(self, key, axis):
         if com.is_bool_indexer(key):
             if hasattr(key, 'index') and isinstance(key.index, Index):
                 if key.index.inferred_type == 'integer':
@@ -1880,15 +1889,17 @@ class _iLocIndexer(_LocationIndexer):
                                               "is not available")
                 raise ValueError("iLocation based boolean indexing cannot use "
                                  "an indexable as a mask")
-            return True
+            return
 
         if isinstance(key, slice):
-            return True
+            return
         elif is_integer(key):
-            return self._is_valid_integer(key, axis)
+            assert(self._is_valid_integer(key, axis))
         elif is_list_like_indexer(key):
-            return self._is_valid_list_like(key, axis)
-        return False
+            assert(self._is_valid_list_like(key, axis))
+        else:
+            raise ValueError("Can only index by location with "
+                             "a [{types}]".format(types=self._valid_types))
 
     def _has_valid_setitem_indexer(self, indexer):
         self._has_valid_positional_setitem_indexer(indexer)
@@ -2017,7 +2028,7 @@ class _iLocIndexer(_LocationIndexer):
             axis = self.axis or 0
 
         if isinstance(key, slice):
-            self._has_valid_type(key, axis)
+            self._validate_key(key, axis)
             return self._get_slice_axis(key, axis=axis)
 
         if isinstance(key, list):
@@ -2027,7 +2038,7 @@ class _iLocIndexer(_LocationIndexer):
                 pass
 
         if com.is_bool_indexer(key):
-            self._has_valid_type(key, axis)
+            self._validate_key(key, axis)
             return self._getbool_axis(key, axis=axis)
 
         # a list of integers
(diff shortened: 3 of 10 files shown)


toobaz:
@jreback @jorisvandenbossche There is some problem with warnings in Python 2.7.

On my machine,

```
 pytest pandas/tests/indexing/test_indexing.py
```

passes and

```
pytest pandas/tests/indexing/test_loc.py
```

passes, but
```
pytest pandas/tests/indexing/test_loc.py  pandas/tests/indexing/test_indexing.py
```

fails with ``Did not see expected warning`` (as in circleci and travis-ci).

Any suggestion (apart from skipping those on Python 2.7)?

pandas/tests/frame/test_indexing.py
```diff
@@ -396,8 +396,9 @@ def test_getitem_setitem_ix_negative_integers(self):
         assert (self.frame['D'] == 0).all()
 
         df = DataFrame(np.random.randn(8, 4))
-        with catch_warnings(record=True):
-            assert isna(df.ix[:, [-1]].values).all()
+        # ix does label-based indexing when having an integer index
+        with pytest.raises(KeyError):
```
jreback: can you also test on the row dim (maybe make these 2 a separate test), may already be an existing test

pandas/tests/indexing/test_indexing.py
```diff
@@ -192,15 +192,9 @@ def test_dups_fancy_indexing(self):
...
-        # inconsistent returns for unique/duplicate indices when values are
-        # missing
-        df = DataFrame(np.random.randn(4, 3), index=list('ABCD'))
-        expected = df.reindex(['E'])
-
         dfnu = DataFrame(np.random.randn(5, 3), index=list('AABCD'))
-        with catch_warnings(record=True):
-            result = dfnu.ix[['E']]
-        tm.assert_frame_equal(result, expected)
+        with pytest.raises(KeyError):
```
jreback: can you a comment here on what you are testing (maybe separate test)?

pandas/tests/series/indexing/test_numeric.py
```diff
@@ -59,12 +59,22 @@ def test_get_nan():
 
     # ensure that fixing the above hasn't broken get
     # with multiple elements
+    idx = [2, 30]
```
jreback: move into separate test

pandas/tests/indexing/test_loc.py
```diff
@@ -119,15 +119,15 @@ def test_loc_getitem_label_out_of_range(self):
                           typs=['ints', 'uints', 'labels', 'mixed', 'ts'],
                           fails=KeyError)
         self.check_result('label range', 'loc', 'f', 'ix', 'f',
-                          typs=['floats'], fails=TypeError)
+                          typs=['floats'], fails=KeyError)
```
jreback: is this reflected in the whatsnew?

pandas/core/indexing.py
```diff
@@ -905,38 +893,31 @@ def _multi_take(self, tup):
...
-                 for t, a in zip(tup, o._AXIS_ORDERS)}
-            return o.reindex(**d)
-        except(KeyError, IndexingError):
-            raise self._exception
+            d = {}
+            for key, axis in zip(tup, o._AXIS_ORDERS):
+                ax = o._get_axis(axis)
+                indexer, keyarr = ax._convert_listlike_indexer(key,
+                                                               kind=self.name)
+                if indexer is not None and (indexer != -1).all():
```
jreback: can you add a comment here

pandas/core/indexing.py
```diff
@@ -1153,15 +1133,53 @@ def _getitem_iterable(self, key, axis=None):
...
+                    self._validate_read_indexer(key, indexer, axis)
                     result = self.obj._take(indexer, axis=axis)
 
                 return result
 
+    def _validate_read_indexer(self, key, indexer, axis):
+        """
+        Check that indexer is OK (e.g. at least one element was found, unless
+        the list of keys was actually empty).
+        """
```
jreback: can you add a more complete doc-string

pandas/core/indexing.py
```diff
@@ -1153,15 +1133,53 @@ def _getitem_iterable(self, key, axis=None):
...
+        ax = self.obj._get_axis(axis)
+        # True indicates missing values
```
jreback: maybe a blank line would help here (not sure what this comment refers)

pandas/core/indexing.py
```diff
@@ -1153,15 +1133,53 @@ def _getitem_iterable(self, key, axis=None):
...
+            return
+
+        missing = indexer < 0
+
+        if np.any(missing):
+            if np.all(missing):
+                raise KeyError(
+                    u"None of [{key}] are in the [{axis}]".format(
+                        key=key, axis=self.obj._get_axis_name(axis)))
+            else:
```
jreback: you don't need the else here

pandas/core/indexing.py
```diff
@@ -1352,6 +1370,33 @@ def _has_valid_type(self, key, axis):
 
         return True
 
+    def _convert_for_reindex(self, key, axis=None):
+        if axis is None:
```
jreback: can you add a doc-string

pandas/core/indexing.py
```diff
@@ -1352,6 +1370,33 @@ def _has_valid_type(self, key, axis):
...
+            axis = self.axis or 0
+        labels = self.obj._get_axis(axis)
+
+        if com.is_bool_indexer(key):
+            key = check_bool_indexer(labels, key)
+            return labels[key]
+        else:
```
jreback: else not needed

pandas/core/indexing.py
```diff
@@ -1879,15 +1889,17 @@ def _has_valid_type(self, key, axis):
...
                                  "an indexable as a mask")
-            return True
+            return
 
         if isinstance(key, slice):
-            return True
+            return
         elif is_integer(key):
-            return self._is_valid_integer(key, axis)
+            assert(self._is_valid_integer(key, axis))
```
jreback: what is the purpose of this assert?

jreback (changes requested):
looks pretty good. need a few more comments. must be not-catching a warning (maybe only in py2).

pandas/core/indexing.py
toobaz: Basically, ``_validate_key`` should raise an error if the keys are not valid (and they aren't if they are an integer but not a valid integer). Better to test and eventually raise a ``ValueError``?

jreback: yes this should go thru ``_validate_key`` (the old ``_has_valid_type``). it should raise if indicated by that type of indexer, so yes test for that

pandas/tests/indexing/test_loc.py
toobaz: This is not a change in the API: I just fixed the test, which was checking nothing at all (because test objects for floats were missing).

pandas/tests/frame/test_indexing.py
toobaz: (done)

pandas/tests/indexing/test_indexing.py
toobaz: (done)

toobaz:
@jreback ready for me

toobaz:
> Are we doing this for 0.23?

I guess so!

jreback:
yes we should
need to review again - ok after rc

pandas/core/indexes/base.py
```diff
@@ -4881,6 +4881,9 @@ def _ensure_index(index_like, copy=False):
     if hasattr(index_like, 'name'):
         return Index(index_like, name=index_like.name, copy=copy)
 
+    if is_iterator(index_like):
```
jreback: I *think* this covers generators as well?

pandas/core/indexing.py
```diff
@@ -186,33 +186,21 @@ def __setitem__(self, key, value):
         indexer = self._get_setitem_indexer(key)
         self._setitem_with_indexer(indexer, value)
 
-    def _has_valid_type(self, k, axis):
+    def _validate_key(self, k, axis):
         raise NotImplementedError()
```
jreback: can you make this an AbstractClassError, and add a doc-string here

pandas/core/indexing.py
```diff
@@ -1337,7 +1377,7 @@ def __init__(self, name, obj):
                       DeprecationWarning, stacklevel=2)
         super(_IXIndexer, self).__init__(name, obj)
 
-    def _has_valid_type(self, key, axis):
+    def _validate_key(self, key, axis):
         if isinstance(key, slice):
             return True
```
jreback: can you add a doc-string to these

pandas/core/indexing.py
```diff
@@ -1656,7 +1739,7 @@ class _LocIndexer(_LocationIndexer):
                     "index is integers), listlike of labels, boolean")
     _exception = KeyError
 
-    def _has_valid_type(self, key, axis):
+    def _validate_key(self, key, axis):
         ax = self.obj._get_axis(axis)
```
jreback: same

pandas/core/indexing.py
```diff
@@ -1880,15 +1928,26 @@ def _has_valid_type(self, key, axis):
...
         if isinstance(key, slice):
-            return True
+            return
         elif is_integer(key):
-            return self._is_valid_integer(key, axis)
+            self._validate_integer(key, axis)
+        elif isinstance(key, tuple):
+            # a tuple should already have been caught by this point
+            # so don't treat a tuple as a valid indexer
+            raise IndexingError('Too many indexers')
```
jreback: this is hit in tests?

pandas/core/indexing.py
```diff
@@ -1880,15 +1928,26 @@ def _has_valid_type(self, key, axis):
...
         elif is_list_like_indexer(key):
-            return self._is_valid_list_like(key, axis)
-        return False
+            # check that the key does not exceed the maximum size of the index
+            arr = np.array(key)
+            l = len(self.obj._get_axis(axis))
+
+            if len(arr) and (arr.max() >= l or arr.min() < -l):
+                raise IndexError("positional indexers are out-of-bounds")
```
jreback: this is hit in tests?

pandas/core/indexing.py
```diff
@@ -1880,15 +1928,26 @@ def _has_valid_type(self, key, axis):
...
+        else:
```
jreback: can do w/o the else here

pandas/core/indexing.py
```diff
@@ -2017,17 +2056,13 @@ def _getitem_axis(self, key, axis=None):
             axis = self.axis or 0
 
         if isinstance(key, slice):
-            self._has_valid_type(key, axis)
             return self._get_slice_axis(key, axis=axis)
 
         if isinstance(key, list):
-            try:
-                key = np.asarray(key)
-            except TypeError:  # pragma: no cover
```
jreback: this is no longer needed?

pandas/tests/indexing/test_indexing.py
```diff
@@ -131,6 +132,8 @@ def test_setitem_dtype_upcast(self):
         assert is_float_dtype(left['foo'])
         assert is_float_dtype(left['baz'])
 
+    @pytest.mark.skipif(PY2, reason=("Catching warnings unreliable with "
```
jreback: ?? where does this come from? this means we are  NOT catching a warning. need to pin this down.

jreback (changes requested):

pandas/tests/indexing/test_indexing.py
toobaz: Yes, as I wrote: https://github.com/pandas-dev/pandas/pull/20770#pullrequestreview-115735997

A bug in testing code only, which arises with Python 2 only, and only when two test files are run _together_, is in the intersection of my "no idea where to start" and "not high priority anyway" categories.

I can open an issue once merged - as there's no simple way (as far as I know) to reproduce otherwise.

toobaz: As I said, see https://github.com/pandas-dev/pandas/pull/20770#issuecomment-383271735

toobaz: (the full suite case behaves like the case with two files)

jorisvandenbossche: None of the travis builds are showing any warnings.

I agree with @jreback we should figure out what is going on before merging. But given the tight timeline for 0.23, *if* this would be the only remaining issue, I would rather merge it before the rc and fix it before the final release.

Anyhow, will try to take a closer look at the full PR tomorrow. Quick general comment: I think this deserves a more elaborate whatsnew notice?

pandas/core/indexing.py
toobaz: I don't see how this was ever needed. I don't think there is _any_ way in which ``np.asarray([...])`` could raise a ``TypeError``. ``np.asarray([0, [1, 2]])`` raises ``ValueError`` (catched elsewhere).

pandas/core/indexes/base.py
toobaz: sure

pandas/core/indexing.py
toobaz: Sure: https://github.com/pandas-dev/pandas/blob/6cacdde5630c593999059833b516e1fec60aaf72/pandas/tests/frame/test_indexing.py#L1389

(this is old code I just moved around)

pandas/core/indexing.py
toobaz: Sure, moreover this is trivial refactoring.

pandas/core/indexing.py
toobaz: > can you make this an AbstractClassError,

Pointer?

pandas/tests/indexing/test_indexing.py
toobaz: > if this would be the only remaining issue

I'm afraid I won't have time for (inherited) docstrings and learning what is a ``AbstractClassError`` soon.

> I would make an exception in this case

To be honest I don't see _to what_ is an exception been made. This PR fixes 3 bugs in code, and the fact that it identifies an additional, already existing, one in tests is an added bonus.

jreback:
how’s that?

toobaz:
> how’s that?

The bug is clearly in the test suite (because whatever the pandas code does, tests shouldn't change behavior when they are called alone vs. together). So I'm just providing a reproducible example and making it evident... which is the opposite of ignoring.

pandas/core/indexing.py
jreback: AbstractMethoderror in pandas/core/base.py

pandas/core/indexing.py
jreback: right my point was that this was catching *something*, so let's see what it was (and if not, of course remove)

pandas/tests/indexing/test_indexing.py
jreback: @toobaz I am not questing the refactoring or that you fixed some bugs, just that something that was added is hiding a bug, maybe in an exception. This is much worse than known bugs and very very hard to find.

toobaz: Do we all agree that the bug is in the test suite? If yes, I guess I don't have much to add: as you correctly point out, it is a nasty bug, which will be hard to isolate - even harder if we don't merge this (because none of us knows how to reproduce otherwise).

jreback: but the point is that this refactor created the bug. i have no problem waiting to merge this until this is fixed. would rather have this correct, this is pretty tricky code.

pandas/core/indexing.py
toobaz: This was added in https://github.com/pandas-dev/pandas/pull/15504/files#diff-7489e290a23303c4db4f803011ecaf8eR1728 ... but I think it was an unnecessary precaution ( @jorisvandenbossche ?), to the best of my understanding it never catched anything.

pandas/core/indexing.py
toobaz: No, I would have to add a

```python
else:
    return
```
branch to the ``if`` just above, because when the indexer is valid it does not return/raise

pandas/tests/indexing/test_indexing.py
toobaz: > but the point is that this refactor created the bug

This "but" has no logical meaning in this discussion. Either you agree with my reasoning that the bug is in the test suite, or you think that a bug in the code can lead a test to pass only when called in isolation, and then I ask you to explain me how.  In the first case, this refactor didn't "create" any bug. In the second... I'm patiently waiting for an explanation (not for one more repetition of the same illogical claim).

Then, sure, there might be a bug in the code _and_ a bug in the tests suite. That's why I have manually tried the code in the failing tests, and it emits the warning just as it should - feel free to try.

The point is not just 0.23.0, it is that we have no idea of what the problem is, it is certainly not _caused_ by this branch, it will be most probably not fixed soon, this branch will soon become a rebasing nightmare (I know by experience), and the indexing code desperately needs cleanup (and it is already difficult enough).

And of course, the point is also that I still didn't see any logical argument for postponing. You know,  I like when discussions are based on actual arguments (i.e. when they are not just a waste of time).

jorisvandenbossche:
@toobaz the reason I mentioned the bigger whatsnew notice is because I think this is actually an important change, and one that will have an impact on people's code. Yes, sure, the change was already done for 0.21 for single level indexes, and you can certainly refer to that and keep it shorter as there, but IMO it still deserves a subsection in the API changes section.

jreback:
@toobaz you argument is not logical. master is passing, this branch is not. if you want to do a smaller PR to show the warning and not do a refactoring ok, otherwise, this PR will be held up.

jorisvandenbossche:
@toobaz to clarify the scope of the introduced warning: this is only for the case where you pass a list of exact labels? (so fully specified for all levels)

toobaz:
> you argument is not logical.

this is pure rethoric - which part of my argument is not logical?

> master is passing, this branch is not.

... depending on how you call the tests, which proves that it is a problem in the tests

> if you want to do a smaller PR to show the warning and not do a refactoring ok

Feel free to stratify more cruft to indexing code, there's already plenty of it. I won't. And by the way, what makes you think that the refactoring, and not the warning itself, triggers a problem _in the code that catches warnings_?!

toobaz:
> you could reproduce it locally on python 2.7 by running the two files together?

yes

> I can't

Interesting, let me investigate

toobaz:
> @toobaz to clarify the scope of the introduced warning: this is only for the case where you pass a list of exact labels? (so fully specified for all levels)

Yes, partial indexing is not affected, it requires a different approach.

jorisvandenbossche:
In the meantime I can reproduce it, I only removed one of the two skips

jorisvandenbossche:
In the meantime I can reproduce it, I only removed one of the two skips

toobaz:
> In the meantime I can reproduce it, I only removed one of the two skips

You are right - I have added the other because the test was failing, but it is not required.

jorisvandenbossche:
Found the test problem (but not yet solved it ;-)): it's in the catching of all xarnings for Panels in check_result

jorisvandenbossche:
> But, then I also needed to add an extra assert_warning, and didn't look into the actual test (the check_result stuff is rather difficult to understand ..) if this is correct or not.

To conclude myself: yes the added `with tm.assert_produces_warning(..)` is correct as it is testing a list with missing labels. The fact that the assert was not there was not logical (and a bug in the testing code), and should actually been have investigated when the asserts were added to the test in the first place.

Anyhow, the full test is also a bit bogus. As it only checks that at least one FutureWarning is raised, but the `self.check_result` tests several things at the same time .. (but that is not for this PR to fix :-))

toobaz:
@jorisvandenbossche thanks, I should have applied your fix.

toobaz:
@jorisvandenbossche let me know if you see any obvious mistake in implementation of your fix...

toobaz:
@jorisvandenbossche for sure your fix is good, but I'm afraid there's something deeper (also because your fix alone should in principle affect Python 2 and 3, and invocation of the entire test suite or of single files, in the same way)

toobaz:
> That said, I am totally fine with skipping those for now, given it is probably a bug in existing testing code.

Agree. So OK for 0.23.0?

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -873,6 +873,7 @@ Deprecations
 - The ``is_copy`` attribute is deprecated and will be removed in a future version (:issue:`18801`).
 - ``IntervalIndex.from_intervals`` is deprecated in favor of the :class:`IntervalIndex` constructor (:issue:`19263`)
 - ``DataFrame.from_items`` is deprecated. Use :func:`DataFrame.from_dict` instead, or ``DataFrame.from_dict(OrderedDict())`` if you wish to preserve the key order (:issue:`17320`, :issue:`17312`)
+- Indexing a ``MultiIndex`` or a ``FloatIndex`` with a list containing some missing keys will now show a ``FutureWarning``, consistently with other types of indexes (:issue:`17758`).
```
jreback: can you add ``:class:`` for these

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1107,6 +1108,8 @@ Indexing
 - :func:`Index.to_series` now accepts ``index`` and ``name`` kwargs (:issue:`18699`)
 - :func:`DatetimeIndex.to_series` now accepts ``index`` and ``name`` kwargs (:issue:`18699`)
 - Bug in indexing non-scalar value from ``Series`` having non-unique ``Index`` will return value flattened (:issue:`17610`)
+- Bug in indexing with iterator containing only missing keys, which raised no error (:issue:`20748`)
+- Fixed inconsistency in ``.ix`` between list and scalar keys when index has integer dtype and does not include desired key(s) (:issue:`20753`)
```
jreback: ... scalar keys when the index has an integer dtype and does not include the desired ...

pandas/core/indexing.py
```diff
@@ -186,33 +188,47 @@ def __setitem__(self, key, value):
...
+
+        Parameters
+        ----------
+        key : scalar, slice or list-like
+            The key requested
+
+        axis : int
+            Dimension on which the indexing is being made
+
+        Returns:
```
jreback: I think we dont' put this if it returns None, @TomAugspurger

pandas/core/indexing.py
```diff
@@ -906,38 +922,34 @@ def _multi_take(self, tup):
...
+                # patiently wait for a KeyError later on - otherwise, convert
+                if len(ax) or not len(key):
+                    key = self._convert_for_reindex(key, axis)
+                indexer = ax.get_indexer_for(key)
+                keyarr = ax.reindex(keyarr)[0]
+                self._validate_read_indexer(keyarr, indexer,
+                                            o._get_axis_number(axis))
+                d[axis] = (keyarr, indexer)
+            return o._reindex_with_indexers(d, copy=True, allow_dups=True)
+        except(KeyError, IndexingError) as detail:
```
jreback: not sure if its worth it, but is it possible to surround only the op that could raise the IndexingError? (with the try/except)

pandas/core/indexing.py
```diff
@@ -1073,8 +1085,9 @@ def _getitem_axis(self, key, axis=None):
         if axis is None:
             axis = self.axis or 0
 
-        if self._should_validate_iterable(axis):
-            self._has_valid_type(key, axis)
+        if is_iterator(key):
```
jreback: is this is_iterator needed here?

pandas/core/indexing.py
```diff
@@ -1154,15 +1165,72 @@ def _getitem_iterable(self, key, axis=None):
...
+        Parameters
+        ----------
+        key : list-like
+            Target labels (only used to show correct error message)
+        indexer: array-like of booleans
+            Indices corresponding to the key (with -1 indicating not found)
+        axis: int
+            Dimension on which the indexing is being made
+
+        Returns
```
jreback: remove this I think

pandas/tests/indexing/test_indexing.py
```diff
@@ -225,6 +221,10 @@ def test_dups_fancy_indexing(self):
             result = df.loc[['A', 'A', 'E']]
         tm.assert_frame_equal(result, expected)
 
+        if PY2:
```
jreback: can you split this test, and use the skip decorator instead (and add the reason)

jreback (changes requested):
minor changes, lgtm.

pandas/core/indexing.py
TomAugspurger: Typically we don't.

pandas/core/indexing.py
TomAugspurger: Didn't address this. I agree that'd be nice, but at a glance it looks like there are 3-4 places that could raise one of those errors. Seems tricky

pandas/core/indexing.py
TomAugspurger: Didn't check this.

TomAugspurger:
Fixed up all but two of the comments. I don't think the tighter exception catching is easy to do correctly.

No idea about checking the key for an iterator. Running the tests now to see if it's hit.

pandas/core/indexing.py
toobaz: It could be moved at a later stage... but I would see it as a loss (more general argument in #20748)

toobaz:
> I don't think the tighter exception catching is easy to do correctly.

... and in any case, I plan to soon merge that code in the "non-multi" branch.

toobaz:
> Is that going in this PR, or a different one?

Future (it's the "future refactoring" I mention in the [description](https://github.com/pandas-dev/pandas/pull/20770#issue-183134763))

TomAugspurger:
Great, thanks. Merging on green.

toobaz:
> do we need to keep track of the problem with the warning being catched by other tests (and some of the discussion / things we found out from this PR) in an issue?

Sure: #20914 



> I asked above about the scope of the warning

Indeed, the warning for the partial indexing (including indexing by levels, even where each level is specified) requires a separate approach: see #20916

PR #20772, opened 2018-04-20
BUG: Fix freq setter for DatetimeIndex/TimedeltaIndex and deprecate for PeriodIndex

### LINKED ISSUE #20678: BUG: freq setter for DatetimeIndex/TimedeltaIndex/PeriodIndex is poorly supported
**1. Setting to a frequency string alias fails for all datetimelike indexes (same error for all three):**
```python-traceback
In [1]: import pandas as pd

In [2]: dti = pd.DatetimeIndex(['20170101', '20170102', '20170103'])

In [3]: dti
Out[3]: DatetimeIndex(['2017-01-01', '2017-01-02', '2017-01-03'], dtype='datetime64[ns]', freq=None)

In [4]: dti.freq = 'D'

In [5]: dti
Out[5]: ---------------------------------------------------------------------------
AttributeError: 'str' object has no attribute 'freqstr'
```

**2. `DatetimeIndex` and `TimedeltaIndex` allow setting to invalid frequencies:**
```python
In [6]: dti = pd.DatetimeIndex(['20170101', '20170102', '20170103'])

In [7]: dti.freq = pd.offsets.Day(2)

In [8]: dti
Out[8]: DatetimeIndex(['2017-01-01', '2017-01-02', '2017-01-03'], dtype='datetime64[ns]', freq='2D')
```
Note that trying to construct this from scratch fails, as expected:
```python-traceback
In [9]: dti = pd.DatetimeIndex(['20170101', '20170102', '20170103'], freq='2D')
ValueError: Inferred frequency D from passed dates does not conform to passed frequency 2D
```

**3. `PeriodIndex` gives nonsensical output when setting with an offset:**
```python
In [10]: pi = pd.PeriodIndex(['2018Q1', '2018Q2', '2018Q3'], freq='Q')

In [11]: pi
Out[11]: PeriodIndex(['2018Q1', '2018Q2', '2018Q3'], dtype='period[Q-DEC]', freq='Q-DEC')

In [12]: pi.freq = pd.offsets.Day()

In [13]: pi
Out[13]: PeriodIndex(['1970-07-12', '1970-07-13', '1970-07-14'], dtype='period[Q-DEC]', freq='D')
```
In addition to the nonsensical dates, note the incompatibility in `[13]` between `dtype` and `freq`.

#### Expected Output
I don't know that setting `freq` for a `PeriodIndex` should be supported: there's some ambiguity in regards aligning to the start/end of a period, and there's also the `PeriodIndex.asfreq` method that gives more control over this.

If setting `freq` is disallowed for `PeriodIndex`, should setting also be disallowed for `DatetimeIndex` and `TimedeltaIndex` in order to remain consistent?  If so, should there then be an analogous `asfreq` for these two?

My thoughts on the three issues:
1) String aliases should be coerced as appropriate (be it via setting or a `asfreq` method).

2) This should raise with a similar message as the constructor (be it via setting or a `asfreq` method).

3) This should raise, potentially with a message indicating to use `asfreq`.  Or if we choose to allow setting, this should be consistent with the default `asfreq` options.

<details>

INSTALLED VERSIONS
------------------
commit: fa231e8766e02610ae5a45e4b2bc90b6c7e9ee6f
python: 3.6.1.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.23.0.dev0+740.gfa231e8
pytest: 3.1.2
pip: 9.0.1
setuptools: 39.0.1
Cython: 0.25.2
numpy: 1.13.3
scipy: 1.0.0
pyarrow: 0.6.0
xarray: 0.9.6
IPython: 6.1.0
sphinx: 1.5.6
patsy: 0.4.1
dateutil: 2.6.0
pytz: 2017.2
blosc: None
bottleneck: None
tables: 3.4.2
numexpr: 2.6.4
feather: 0.4.0
matplotlib: 2.0.2
openpyxl: 2.4.8
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 0.9.8
lxml: 3.8.0
bs4: None
html5lib: 0.999
sqlalchemy: 1.1.13
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
fastparquet: 0.1.0
pandas_gbq: None
pandas_datareader: None

</details>

jschendel:
- [X] closes #20678
- [X] tests added / passed
- [X] whatsnew entry

For `DatetimeIndex`/`TimedeltaIndex`:
  - Setting `freq` with a string alias now works
  - Frequencies are validated with invalid frequencies raising
     - Same behavior as when the constructor is called with an invalid frequency

For `PeriodIndex`:
  - Setting `freq` is deprecated in favor of `asfreq`
      - The behavior of the setter would be ambiguous without the extra `asfreq` params
  - Left the existing (buggy) `freq` setter behavior as-is
     - Didn't seem worthwhile to try fixing it
  - I don't think that the setter was widely used for `PeriodIndex` anyways
     - Didn't see any warnings in the tests


pandas/core/indexes/period.py
```diff
@@ -560,6 +560,19 @@ def is_full(self):
...
+    def freq(self):
+        """Return the frequency object if it is set, otherwise None"""
+        return self._freq
+
+    @freq.setter
+    def freq(self, value):
+        msg = ('Setting PeriodIndex.freq has been deprecated and will be '
+               'removed in a future version; use PeriodIndex.asfreq instead. '
+               'The PeriodIndex.freq setter is not guaranteed to work.')
+        warnings.warn(msg, FutureWarning, stacklevel=2)
```
jschendel: Not sure if we really need to deprecate setting this: it's user facing, but I don't think it really work unless you set to the existing frequency.  Could potentially remove the setter entirely, causing an `AttributeError` to be raised, or could raise a custom `AttributeError` within the setter to give a better message directing users to `.asfreq`.

jreback:
I would also deprecate settig for DTI/TDI, and direct to using ``.asfreq``. I don't like the idea of disallowing on PI but not the others.

jschendel:
Can implement `.asfreq` and deprecate the setters for DTI/TDI in the next day or so.

cc @jorisvandenbossche : Thoughts on this?  Your preference in the issue was to allow setting for DTI/TDI and not implement `.asfreq`.

jorisvandenbossche:
Yes, -1 on `asfreq`. As I tried to argue on the issue, it is something different as `PeriodIndex.asfreq`, so we should not call it the same. Directly setting the attribute is fine IMO if you need this.

pandas/core/indexes/datetimelike.py
```diff
@@ -205,6 +205,43 @@ def floor(self, freq):
...
+            The index on which to determine if the given frequency is valid
+        freq : DateOffset
+            The frequency to validate
+        """
+        inferred = index.inferred_freq
+        if index.empty or inferred == freq.freqstr:
+            return None
+
+        on_freq = cls._generate(
+            index[0], None, len(index), None, freq, **kwargs)
```
jorisvandenbossche: is generating the index needed to know if it a correct frequence? Checking the inferred frequency is not enough?

pandas/core/indexes/datetimes.py
```diff
@@ -454,15 +454,7 @@ def __new__(cls, data=None,
 
         if verify_integrity and len(subarr) > 0:
             if freq is not None and not freq_infer:
-                inferred = subarr.inferred_freq
-                if inferred != freq.freqstr:
-                    on_freq = cls._generate(subarr[0], None, len(subarr), None,
-                                            freq, tz=tz, ambiguous=ambiguous)
```
jorisvandenbossche: passing the `tz` in the new version is not needed?

pandas/core/indexes/datetimelike.py
jorisvandenbossche: OK, see it was also done like that below (still wondering though)

pandas/core/indexes/period.py
jorisvandenbossche: Yes, IMO it was buggy, and as you say, only if it was the same frequency, it was working. But given that corner case, it maybe does not hurt to do the deprecation I would say.

I think the custom AttributeError message is a good idea in the future.

jorisvandenbossche (commented):
Apart from the `tz` question, this looks good IMO

jorisvandenbossche:
> I don't like the idea of disallowing on PI but not the others.

I think this difference is acceptable, because `freq` also has a different notion for both: for PI the `freq` is required (you cannot have a PI without a `freq`), and the underlying int values are dependent on this freq. So changing the `freq` is either buggy because the values and the freq no longer match (current behaviour), or it converts the underlying int values to match the new `freq` (but this is what `asfreq` already does, and since this alters the data, I think should not be done by setting an attribute). 
On the other hand, for DTI and TDI, the `freq` is not required, and setting it allows you to add one. But this only changes the `freq` to the `inferred_freq` and does not alter the data (as otherwise an error is raised).

jorisvandenbossche:
(note: I tagged for 0.23, but it does not need to hold up the rc)

pandas/core/indexes/datetimelike.py
jschendel: I think it's necessary, at least to a certain extent.  There are cases where multiple frequencies are valid for a given set of dates, so you can have a valid frequency that's not the inferred frequency, e.g. `['2018-01-01', '2018-01-02', '2018-01-03']` could be calendar day or business day.  Users can also define custom frequencies that would need to be validated but would not be the inferred frequency.

There are cases where you don't need to generate the entire index to reject invalid frequencies, e.g. if the second generated value doesn't match.  But I don't immediately see how to get around generating the entire index to determine if a frequency is valid.  I suppose you could maybe optimize memory usage with huge indexes by doing some type of elementwise validation, only keeping the current generated element in memory.

jorisvandenbossche: Ah, yes, I see, that's for sure a valid case.

For me, improving the performance of the validation is not very high priority.

pandas/core/indexes/datetimes.py
jorisvandenbossche: Ah, yep, indeed, that sounds correct (and tested it with a small example)

jorisvandenbossche (approved):

jreback (changes requested):
this code change looks fine. I still think should deprecate setting the freq on DTI/TDI. I don't see much value there and the inconcistenty with that you *cannot* set PI is pretty jaring. Is there an actual good usecase for setting DTI/TDI? If there is we should actually do this via a function rather than an attribute to make it more explicit.

TomAugspurger:
> (note: I tagged for 0.23, but it does not need to hold up the rc)

If this is going in 0.23, then I'd like it to be in the RC. Ideally we only merge bugfixes for the RC to the 0.23 release branch.

jschendel:
I more or less have code ready that addresses @jreback's comments, but am a bit hesitant to push since there isn't agreement on if the freq setter should be deprecated, and if so what the name of the alternative method would be.

jschendel:
This commit would go on top of the current changes in this PR to address @jreback's comments, in case anyone care too look while things are still being discussed here: https://github.com/jschendel/pandas/commit/2fca7751205267bcc3fa1c07de9af19d646b5f6a

The CI services that are hooked to my repo all ran successfully, and I didn't see any warnings related to my changes in the logs.  Didn't want to mess around with this PR too much while there's still ongoing discussion.

jreback:
> So I have a clear -1 on re-using the name asfreq for this, as I explained before (asfreq on Period is doing something different, I haven't see anybody giving answers to that argument).

Can you repeat I can't find what you said.

I am fine the the PI change here. The issue comes down to whether we should allow one to change ``freq`` on the *immutable* DTI/TDI. I see no good reason for this. If you want to set the freq, simply create another one, or we make availble ``.asfreq on TDI/DTI (which I am little less excited about). I think having a method only on PI is fine. However the *setting* of an attribute is just plane wrong (yes we allow it for ``.name``, but recently removed it for ``.tz`` as that was buggy).

jorisvandenbossche:
I have said it here https://github.com/pandas-dev/pandas/pull/20772#issuecomment-385060944 and here https://github.com/pandas-dev/pandas/issues/20678#issuecomment-381050890. But it actually just boils down to "for PI `asfreq` *changes* the underlying data (it is for conversion of one frequency to another), for DTI/TDI setting the `freq` just updates an informative attribute that does not change the data at all"

>  However the *setting* of an attribute is just plane wrong 

There is nothing "plain wrong" about setting an attribute, it is perfectly valid and often even preferable python code (I think often having a lot of "set_.." methods is seen as "old-java-style" while just having a property with getter and setter under the hood is more pythonic).

jreback:
> However the setting of an attribute is just plane wrong

On an immutable object this is True. We have 1 and only 1 attribute that we allow to be changed on an Index and that is the name. All others give you back a new object. Why break this paradigm?

Further I don't buy your argument. Whether ``.asfreq`` happens to do work on a PI, and doesn't on a TDI/DTI is not a compelling argument.

Is there a compelling use-case to actually set freq (whether inplace or return a new object) on a DTI/TDI?

jschendel:
> Is there a compelling use-case to actually set freq (whether inplace or return a new object) on a DTI/TDI?

I made this PR after noticing a few SO questions that worked around the buggy setting behavior.  That doesn't necessarily there's actually a _valid_ use case for it, as there may be more direct workarounds for the issues people are having, but at the very least people are trying to use it.

I think one of the issues is that people generally default to using `pd.to_datetime` for parsing, which as far as I can tell does not have the ability to set a frequency on creation, and so they then want to set a frequency after creation.  This might be more of an argument for a `freq` parameter in `pd.to_datetime` though.  There also isn't much documentation for the `freq='infer'` option of the DTI constructor, so I also see people wanting to set the inferred frequency after creation, without actually knowing what that frequency is (again, more of a documentation issue).

I could also imagine a use case where one has messy data which needs to be cleaned first, prior to actually setting the `freq`.  So a workflow like parse -> clean -> set frequency.  Some cleaning operations may take care of the setting the `freq`, but there are cases where multiple frequencies are possible, so the correct one may not be set, or a custom frequency may be desired, or maybe no frequency is desired and a user wants to remove by setting `None`.


>Whether .asfreq happens to do work on a PI, and doesn't on a TDI/DTI is not a compelling argument.

To add another data point, `Series.asfreq` and `DataFrame.asfreq` will also do work on an underlying DTI (doesn't appear to work for TDI).

Not particularly invested in being for/against reusing the `asfreq` name though, as there are downsides to both without a clear winner in my mind (potential for confusion with reuse vs. API bloat for new method).

jreback:
@jschendel thanks for the color. But *why* are people actually setting the ``freq``. This doesn't actually do anything (I don't think). Anytime we need the freq we actually infer it (again for DTI/TDI)

jschendel:
Setting the `freq` enables a couple operations that fail without one:

```python
In [2]: dti = pd.DatetimeIndex(['20180101', '20180102', '20180103'])

In [3]: dti
Out[3]: DatetimeIndex(['2018-01-01', '2018-01-02', '2018-01-03'], dtype='datetime64[ns]', freq=None)

In [4]: dti.shift(1)
NullFrequencyError: Cannot shift with no freq

In [5]: dti + 4
NullFrequencyError: Cannot shift with no freq

In [6]: dti = pd.DatetimeIndex(['20180101', '20180102', '20180103'], freq='B')

In [7]: dti.shift(1)
Out[7]: DatetimeIndex(['2018-01-02', '2018-01-03', '2018-01-04'], dtype='datetime64[ns]', freq='B')

In [8]: dti + 4
Out[8]: DatetimeIndex(['2018-01-05', '2018-01-08', '2018-01-09'], dtype='datetime64[ns]', freq='B')
```

jreback:
right, we allow an integer to stand in an represent the freq. ok I would say these are ok usecases (if prob not very well documented). but that is not sufficient reason to actually allow setting of an attribute on an immutable object (rather than just having a function give you a new one). its just bad semantics.

I would be ok with

``dti.asfreq('D')``, makes it pretty explict. We have methods for setting things like tz's, why not freq.

jorisvandenbossche:
> On an immutable object this is True. We have 1 and only 1 attribute that we allow to be changed on an Index and that is the name. All others give you back a new object. Why break this paradigm?

Exactly because the `freq` does *not change* the index, so the index is still immutable, except for those informative attributes like `name` and `freq`. So that in sense, `freq` is very similar to `name` for a DTI/TDI.

> To add another data point, Series.asfreq and DataFrame.asfreq will also do work on an underlying DTI (doesn't appear to work for TDI).

For me that is another reason not to use `asfreq`, as for Series this is indeed a basic form of resample. So the name `asfreq` is already too much overloaded, no need to add yet another meaning.

jreback:
ok, so I would be on board with depreating ``.freq`` setting as in this PR, and adding a ``.set_freq()`` as an alternative. This would be much more in line with our handling of names in any event. The PI changes are fine.

Agree that ``.asfreq`` might be confusing here for DTI/TDI.

jorisvandenbossche:
@jreback Are you fine with merging this now as is? The changes it does are already good (fixing a clear bug in the current setting functionality + deprecation for PI), and the other changes you propose in your last comment can be done separately I think  (on top of this PR). 

(just as way to include this in the rc)

TomAugspurger:
Opened https://github.com/pandas-dev/pandas/issues/20886 for the rest.

TomAugspurger:
Thanks all.

PR #20780, opened 2018-04-21
Accessor registry

### LINKED ISSUE #19963: Classes should maintain a registry of accessors
Dask would like to inherit the accessors registered by `register_*_acessor`. This would be much easier if the `regiser_*_accessor` methods added the name of the accessor to a (private) class variable.

It seems like we have use of it in pandas as well: https://github.com/pandas-dev/pandas/pull/19960/files#r171815351

TomAugspurger:
Closes #19963 

cc @datapythonista I think this would have solved your issues w/ the validation script.

### DIFF
--- a/pandas/core/accessor.py
+++ b/pandas/core/accessor.py
@@ -148,6 +148,7 @@ def _register_accessor(name, cls):
                 UserWarning,
                 stacklevel=2)
         setattr(cls, name, CachedAccessor(name, accessor))
+        cls._accessors.add(name)
         return accessor
     return decorator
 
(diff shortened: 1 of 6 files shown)


TomAugspurger:
Do we want to document this as the official way to get a list of the registered accessors for an object?

jreback:
lgtm. ok to update docs / merge.

jorisvandenbossche:
> Do we want to document this as the official way to get a list of the registered accessors for an object?

Do you want to use this in eg Dask? 
If so, then maybe we should document that.

TomAugspurger:
I think the current preference is for authors to explicitly register their
accessors with Dask as well.

Let's leave it private for now, and if people need it later on we will
document it.

PR #20782, opened 2018-04-21
DEPR: remove v018 resample compatibilitiy

### LINKED ISSUE #20554: DEPR: remove resample compatibility
we have a bunch of resample compatibility code from the transition from an eager execution model to a groupby based; this should be removed for 0.23.0

(14cf67f L132, 14cf67f L685, 14cf67f L700)

jreback:
closes #20554

### DIFF
--- a/pandas/core/resample.py
+++ b/pandas/core/resample.py
@@ -62,20 +62,6 @@ class Resampler(_GroupBy):
     _attributes = ['freq', 'axis', 'closed', 'label', 'convention',
                    'loffset', 'base', 'kind']
 
-    # API compat of allowed attributes
-    _deprecated_valids = _attributes + ['__doc__', '_cache', '_attributes',
-                                        'binner', 'grouper', 'groupby',
-                                        'sort', 'kind', 'squeeze', 'keys',
-                                        'group_keys', 'as_index', 'exclusions',
-                                        '_groupby']
-
-    # don't raise deprecation warning on attributes starting with these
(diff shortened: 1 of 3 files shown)


jreback:
@jorisvandenbossche @TomAugspurger if any comments

jorisvandenbossche (commented):
Looks good in general, just a small note on something I don't fully understand

pandas/core/resample.py
```diff
@@ -62,20 +62,6 @@ class Resampler(_GroupBy):
     _attributes = ['freq', 'axis', 'closed', 'label', 'convention',
                    'loffset', 'base', 'kind']
 
-    # API compat of allowed attributes
-    _deprecated_valids = _attributes + ['__doc__', '_cache', '_attributes',
-                                        'binner', 'grouper', 'groupby',
```
jorisvandenbossche: whoops, didn't push the button probably :-)

My question was about the above `_deprecated_valids`: what was the purpose of them? At least before, those attributes did not raise any warning.

jorisvandenbossche: But you can probably ignore my comment, those attributes are "hided" from tab completion anyway, and didn't change for the rest. I suppose this was here before to prevent raising any warning on them?

jorisvandenbossche (approved):

pandas/core/resample.py
jreback: yes, this was gymnastics to make all of this work

Issue #20735, opened 2018-04-18
pd.concat() does not work generically for ExtensionArrays


bfollinprm:
#### Code Sample, a copy-pastable example if possible

Sadly, pseudocode because the underlying container is proprietary, but this is a generic problem. If I get a chance this weekend I will write up a mock open-source-compliant example for illustration and testing.

```python
class MyArray(ExtensionArray):
    def __init__(self, values, **kwargs):
        # All the things
        self.values = MyNotNumpyContainer(values)
    # All the other methods

df = pd.DataFrame({'mycolumn': MyArray(values)})

# Raises AttributeError: 'MyNotNumpyContainer' object has no attribute 'dtype'
```
#### Problem description

There is a check at the top of `_isna_ndarraylike()`:

```python
def _isna_ndarraylike(obj):
    values = getattr(obj, 'values', obj)
    dtype = values.dtype

    if is_extension_array_dtype(obj):
        if isinstance(obj, (ABCIndexClass, ABCSeries)):
            values = obj._values
        else:
            values = obj
        result = values.isna()
```
This fails for `ExtensionArray` objects which define a `values` attribute, but whose `values` attribute does not have a `dtype` attribute.

I noticed this call in `pd.concat` above, but no doubt it occurs elsewhere.

#### Expected Output

Since `dtype` is a required attribute of an `ExtensionArray`, `_isna_ndarraylike()` should at least get the dtype from the `ExtensionArray` class. Really, a check for whether we are dealing with an `ExtensionArray` should occur upstream somewhere, since there is no guarantee an ExtensionArray is backended by a numpy array-like object.

#### Output of ``pd.show_versions()``

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 2.7.13.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.0-1048-aws
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: None.None

pandas: 0.23.0.dev0+762.gbb095a6
pytest: 3.2.2
pip: 9.0.1
setuptools: 27.2.0
Cython: 0.28.2
numpy: 1.13.3
scipy: 0.19.1
pyarrow: None
xarray: None
IPython: 5.3.0
sphinx: 1.5.4
patsy: None
dateutil: 2.6.1
pytz: 2017.2
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 2.0.2
[...]

PR #20793, opened 2018-04-23
EA.values warning test

TomAugspurger:
xref #20735 (doesn't close)

Here's a local "test" for it.

```
pytest pandas/tests/extension/test_values.py -k test_no_values
==================================================================== test session starts ====================================================================
platform darwin -- Python 3.7.0b3, pytest-3.5.0, py-1.5.3, pluggy-0.6.0
rootdir: /Users/taugspurger/sandbox/pandas, inifile: setup.cfg
plugins: xdist-1.22.2, forked-0.2, cov-2.5.1
collected 9 items / 8 deselected

pandas/tests/extension/test_values.py F                                                                                                               [100%]

========================================================================= FAILURES ==========================================================================
______________________________________________________ TestBaseInterfaceTests.test_no_values_attribute ______________________________________________________

self = <pandas.tests.extension.test_values.TestBaseInterfaceTests object at 0x112ed4748>, data = JSONArary([{'A': [1, 2], 'B': [3, 4]}])

    def test_no_values_attribute(self, data):
        # GH-20735
        # Currently, pandas has places where we accepts Union[ndarray, EA]
        # but in practice expect an ndarray, since we get `data.values.dtype`.
        # This warns users to avoid using a `.values` attribute that is not
        # an ndarray-like
        if hasattr(data, 'values') and not hasattr(data.values, 'dtype'):
            msg = ("ExtensionArray contains a 'values' attribute that does "
                   "not have a dtype attribute. This may cause issues in "
                   "pandas' internals.")
>           raise ValueError(msg)
E           ValueError: ExtensionArray contains a 'values' attribute that does not have a dtype attribute. This may cause issues in pandas' internals.

pandas/tests/extension/base/interface.py:64: ValueError
========================================================== 1 failed, 8 deselected in 0.12 seconds ===========================================================
```

What, if anything, should we include in pandas to test this? I'm fine with the current PR (no tests), since it's just test code.

Edit: ignore what I said about testing. Added one that calls `super()` inside a `raises` context manager. It's not too bad.

### DIFF
--- a/pandas/core/arrays/base.py
+++ b/pandas/core/arrays/base.py
@@ -5,6 +5,19 @@
    This is an experimental API and subject to breaking changes
    without warning.
 """
+# Some notes, just for implementors.
+#
+# * Avoid a `.values` attribute
+#   There are corners in pandas that assume assume array.values is an ndarray,
+#   and does ndarray things like ge the dtype. If you have a non-ndarray
+#   `.values` attribute, things may break. We're working to fix these.
+# * Reference the default implementations, but avoid using private methods.
+#   Some default implementations (e.g. `factorize`) use pandas' internal
+#   methods. If possible, avoid using those methods when overriding the default
+#   implementation. If that isn't possible, raise an issue to discuss adding
+#   those to the public API.
+
+
 import numpy as np
 
 from pandas.errors import AbstractMethodError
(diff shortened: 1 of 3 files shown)


TomAugspurger:
Closing in favor of #20794

PR #20794, opened 2018-04-23
 TST: ExtensionArrays disallow .values attribute 

jorisvandenbossche:
xref https://github.com/pandas-dev/pandas/issues/20735. 
Adds a test to ensure EA authors don't use the `.values` or `._values` attribute, which can give problems with pandas internals depending on how the data is stored.

### DIFF
--- a/pandas/tests/extension/decimal/array.py
+++ b/pandas/tests/extension/decimal/array.py
@@ -30,10 +30,13 @@ class DecimalArray(ExtensionArray):
     def __init__(self, values):
         values = np.asarray(values, dtype=object)
 
-        self.values = values
+        self._data = values
         # Some aliases for common attribute names to ensure pandas supports
         # these
-        self._items = self._data = self.data = self.values
+        self._items = self.data = self._data
+        # those aliases are currently not working due to assumptions
+        # in internal code (GH-20735)
+        # self._values = self.values = self.data
 
     @classmethod
     def _constructor_from_sequence(cls, scalars):
@@ -45,13 +48,13 @@ class DecimalArray(ExtensionArray):
 
     def __getitem__(self, item):
         if isinstance(item, numbers.Integral):
-            return self.values[item]
+            return self._data[item]
         else:
-            return type(self)(self.values[item])
+            return type(self)(self._data[item])
 
     def copy(self, deep=False):
         if deep:
-            return type(self)(self.values.copy())
+            return type(self)(self._data.copy())
         return type(self)(self)
 
     def __setitem__(self, key, value):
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
This is a temporary "fix" to the interface. 
Long term solution would be to make sure we don't make such assumptions internally, but that would need to fix all places where we currently check for a `_values` or `values` attribute, which is much more work (as often it is actually a way to have code be generic for eg both Series and EA/ndarray, as `_values` on the Series then returns the underlying EA/ndarray).

jorisvandenbossche:
Ah, see you just did the same https://github.com/pandas-dev/pandas/pull/20793 :-)

TomAugspurger:
Do you want to copy-paste my test here? https://github.com/pandas-dev/pandas/pull/20793/files#diff-fac89eff36e78c735bd1046e704780c9

I prefer your check. I was duck-typing on `.values` without a `.dtype`, but just checking for `.values` more generically like you do is best.

pandas/tests/extension/decimal/array.py
```diff
@@ -30,10 +30,13 @@ class DecimalArray(ExtensionArray):
     def __init__(self, values):
         values = np.asarray(values, dtype=object)
 
-        self.values = values
+        self._data = values
         # Some aliases for common attribute names to ensure pandas supports
         # these
-        self._items = self._data = self.data = self.values
+        self._items = self.data = self._data
```
jreback: you have .data here? (no a big deal, but this is private yes?)

pandas/tests/extension/json/array.py
```diff
@@ -33,6 +33,13 @@ def __init__(self, values):
                 raise TypeError
         self.data = values
 
+        # Some aliases for common attribute names to ensure pandas supports
+        # these
+        self._items = self._data = self.data
```
jreback: same?

pandas/tests/extension/decimal/array.py
TomAugspurger: These are just in the tests.

jorisvandenbossche:
> Do you want to copy-paste my test here? https://github.com/pandas-dev/pandas/pull/20793/files#diff-fac89eff36e78c735bd1046e704780c9

which test do you mean?

pandas/tests/extension/decimal/array.py
jorisvandenbossche: What do you exactly mean? It's not really about private/public, that's up to the EA author. The purpose of this it to ensure pandas has no problems if an EA author would use `.data` to store the underlying data (in geopandas we are actually using a `data` attribute)

jreback: I understand you *can* have a ``.data`` items, it just doesn't seem a good pattern to advertise

The fact that these are in the tests is very odd itself.

And the comment makes these doubly confusing.

jreback: @jorisvandenbossche ok I re-read what you are testing, this is ok

jreback:
thanks!

PR #20795, opened 2018-04-23
DOC/API: expose ExtensionArray in public api

TomAugspurger:
cc @jorisvandenbossche  do you want to include these in the public API docs? It's nice to have a place to link to I think.

### DIFF
--- a/doc/source/extending.rst
+++ b/doc/source/extending.rst
@@ -61,7 +61,7 @@ Extension Types
 
 .. warning::
 
-   The ``ExtensionDtype`` and ``ExtensionArray`` APIs are new and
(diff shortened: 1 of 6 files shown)


doc/source/extending.rst
```diff
@@ -79,10 +79,10 @@ on :ref:`ecosystem.extensions`.
 
 The interface consists of two classes.
 
-``ExtensionDtype``
-^^^^^^^^^^^^^^^^^^
+:class:`pandas.api.extension.ExtensionDtype`
```
jorisvandenbossche: maybe add a `~` in the title link (to only show ExtensionDtype), or in the link in the text. (using the full link once is enough I think?)

jorisvandenbossche (approved):
Yes, we should indeed do this.

jorisvandenbossche:
Thanks!

Issue #19760, opened 2018-02-19
BUG: regression in DataFrame.transform('func')


jorisvandenbossche:
From a failing example in the doc build:

```python
tsdf = pd.DataFrame(np.random.randn(10, 3), columns=['A', 'B', 'C'],
                    index=pd.date_range('1/1/2000', periods=10))
tsdf.iloc[3:7] = np.nan
```

Using a numpy func directly is fine:
```
In [43]: tsdf.transform(np.abs)
Out[43]: 
                   A         B         C
2000-01-01  1.038786  0.380848  1.190139
2000-01-02  2.201640  0.004028  0.704541
2000-01-03  1.803344  0.774565  0.034836
2000-01-04       NaN       NaN       NaN
2000-01-05       NaN       NaN       NaN
2000-01-06       NaN       NaN       NaN
2000-01-07       NaN       NaN       NaN
2000-01-08  0.351827  0.265785  1.424491
2000-01-09  0.962042  1.291740  1.107973
2000-01-10  0.366717  0.318791  0.117215
```

But specifying the function with a string now breaks:
```
In [44]: tsdf.transform('abs')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-44-405a88736ee3> in <module>()
----> 1 tsdf.transform('abs')

/home/joris/scipy/pandas/pandas/core/generic.py in transform(self, func, *args, **kwargs)
   7694         @Appender(_shared_docs['transform'] % _shared_doc_kwargs)
   7695         def transform(self, func, *args, **kwargs):
-> 7696             result = self.agg(func, *args, **kwargs)
   7697             if is_scalar(result) or len(result) != len(self):
   7698                 raise ValueError("transforms cannot produce "

/home/joris/scipy/pandas/pandas/core/frame.py in aggregate(self, func, axis, *args, **kwargs)
   4830                 pass
   4831         if result is None:
-> 4832             return self.apply(func, axis=axis, args=args, **kwargs)
   4833         return result
   4834 

/home/joris/scipy/pandas/pandas/core/frame.py in apply(self, func, axis, broadcast, raw, reduce, result_type, args, **kwds)
   5020                          args=args,
   5021                          kwds=kwds)
-> 5022         return op.get_result()
   5023 
   5024     def applymap(self, func):

/home/joris/scipy/pandas/pandas/core/apply.py in get_result(self)
    309                                       *self.args, **self.kwds)
    310 
[...]

PR #20800, opened 2018-04-23
BUG: Fixed NDFrame.transform('abs')

TomAugspurger:
Closes https://github.com/pandas-dev/pandas/issues/19760

So this is a hacky workaround. I think we would ideally whitelist a set of valid NDFrame.transform methods, but I don't want to break things. What would that list include?

- abs
- pct_change
- shift
- tshift
- cum*

### DIFF
--- a/pandas/core/apply.py
+++ b/pandas/core/apply.py
@@ -111,6 +111,9 @@ class FrameApply(object):
 
         # string dispatch
         if isinstance(self.f, compat.string_types):
+            # Support for `frame.transform('method')`
+            # Some methods (shift, etc.) require the axis argument, others
+            # don't, so inspect and insert if nescessary.
             func = getattr(self.obj, self.f)
             sig = compat.signature(func)
             if 'axis' in sig.args:
--- a/pandas/core/apply.py
+++ b/pandas/core/apply.py
@@ -111,7 +111,9 @@ class FrameApply(object):
 
         # string dispatch
         if isinstance(self.f, compat.string_types):
-            self.kwds['axis'] = self.axis
+            if self.f not in {'abs'}:
+                # Not all transform functions take an axis keyword.
+                self.kwds['axis'] = self.axis
             return getattr(self.obj, self.f)(*self.args, **self.kwds)
 
         # ufunc
--- a/pandas/core/apply.py
+++ b/pandas/core/apply.py
@@ -111,10 +111,11 @@ class FrameApply(object):
 
         # string dispatch
         if isinstance(self.f, compat.string_types):
-            if self.f not in {'abs'}:
-                # Not all transform functions take an axis keyword.
+            func = getattr(self.obj, self.f)
+            sig = compat.signature(func)
+            if 'axis' in sig.args:
                 self.kwds['axis'] = self.axis
-            return getattr(self.obj, self.f)(*self.args, **self.kwds)
+            return func(*self.args, **self.kwds)
 
         # ufunc
         elif isinstance(self.f, np.ufunc):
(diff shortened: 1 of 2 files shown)


WillAyd:
I'd also add fill and rank to your list. FWIW there's probably a lot of overlap with the groupby module and what it considers to be a transformation vs an aggregation. Wonder if we should keep all of these functions and their "application type" in a shared module for both DataFrame and GroupBy ops

pandas/core/apply.py
```diff
@@ -111,7 +111,9 @@ def get_result(self):
 
         # string dispatch
         if isinstance(self.f, compat.string_types):
-            self.kwds['axis'] = self.axis
+            if self.f not in {'abs'}:
```
jreback: NO, anytime 'hacky' work-around you need to re-think the patch.

jreback: either you need signature introspection (in python 3, may not be possible in python 2), alternatively can catch this type of error, but should be much higher level (in .aggregate), and not here.

jreback (changes requested):

pandas/core/apply.py
TomAugspurger: Ugh, signature introspection seems hackier than hard-coding, no?

TomAugspurger:
This should be for 0.23. It's a regression and was included in the docs.

pandas/core/apply.py
TomAugspurger: I'll see how hard it is to get Py2 & 3 working by inspecting the signature.

pandas/core/apply.py
```diff
@@ -111,8 +111,11 @@ def get_result(self):
 
         # string dispatch
         if isinstance(self.f, compat.string_types):
-            self.kwds['axis'] = self.axis
-            return getattr(self.obj, self.f)(*self.args, **self.kwds)
+            func = getattr(self.obj, self.f)
```
jreback: much better, and now IIRC we actually do something like this in groupby apply (for the same reason). can you add some comments here on what is going on

jreback (changes requested):

pandas/tests/frame/test_apply.py
```diff
@@ -880,6 +880,13 @@ def f():
             with np.errstate(all='ignore'):
                 df.agg({'A': ['abs', 'sum'], 'B': ['mean', 'max']})
 
+    def test_transform_abs_name(self):
```
jreback: do we need more tests? (for other funcs)

jreback (changes requested):

pandas/tests/frame/test_apply.py
```diff
@@ -880,11 +881,14 @@ def f():
...
+    @pytest.mark.parametrize('method', [
+        'abs', 'shift', 'pct_change', 'cumsum', 'rank',
+    ])
+    def test_transform_method_name(self, method):
         # https://github.com/pandas-dev/pandas/issues/19760
         df = pd.DataFrame({"A": [-1, 2]})
-        result = df.transform('abs')
-        expected = pd.DataFrame({"A": [1, 2]})
+        result = df.transform(method)
+        expected = operator.methodcaller(method)(df)
```
WillAyd: Learned something new today. Just out of curiosity, is this any different from doing `getattr(df, method)()`?

pandas/tests/frame/test_apply.py
```diff
@@ -880,11 +881,14 @@ def f():
             with np.errstate(all='ignore'):
                 df.agg({'A': ['abs', 'sum'], 'B': ['mean', 'max']})
 
-    def test_transform_abs_name(self):
+    @pytest.mark.parametrize('method', [
+        'abs', 'shift', 'pct_change', 'cumsum', 'rank',
```
WillAyd: Could also add `ffill` and `bfill`

pandas/tests/frame/test_apply.py
TomAugspurger: in this case, no. methodcaller is nicer though when you need to bind parameters, but don't have the object yet (common in pytest fixtures).

PR #20801, opened 2018-04-23
ENH: ExtensionArray support for objects with _can_hold_na=False and relational operators

### LINKED ISSUE #20659: ExtensionArray support for comparisons to fundamental types and lists of those types
This is probably for @TomAugspurger to look at.

I am trying out an idea that uses the new `ExtensionArray` capability that is in master.  What I would like to do is override the relational operators `__le__`, `__ge__`, etc. I'm following the cyberpandas example that @TomAugspurger published.

Using cyberpandas, I ran the following code

```python
import cyberpandas as ip
import pandas as pd

v = ip.IPArray.from_pyints([1, 2, 3])
s = pd.Series(v)
try:
    res = (s <= pd.Series([4,1,3]))
except Exception as e:
    print("exception raised for comparison to Series ", e)
try:
    res = (s <= [4,1,3])
except Exception as e:
    print("exception raised for comparison to list ", e)

try:
    res = (s <= 2)
except Exception as e:
    print("exception raised for comparison to int ", e)
    
try:
    res = (s <= s[2])
    print("no exception raised for comparison to IPType")
except Exception as e:
    print("exception raised for comparison to IPType ", e)
```

The result is the following output:
```
exception raised for comparison to Series  invalid type comparison
exception raised for comparison to list  '<=' not supported between instances of 'IPv4Address' and 'int'
exception raised for comparison to int  '<=' not supported between instances of 'IPv4Address' and 'int'
no exception raised for comparison to IPType
```

#### Problem description

The first exception is expected, because `cyberpandas.ip_array.__le__` raises `NotImplemented`, which is then caught in `pandas.core.ops.na_op()` at line 1166.

The second two exceptions are a bit not expected, because I would want the `IPArray.__le__()` method to be called so that the `NotImplemented` exception is raised, but in this case, there is code in `pandas.core.ops.wrapper()` around line 1248 that converts the `Series` into a `numpy` array of objects, and then `pandas.core.ops._comp_method_OBJECT_ARRAY()` is called, which is trying to do an efficient comparison, which never calls the `IPArray.__le__`  method.

So the problem here is that for my application, I would like the `ExtensionArray.__le__()` implementation to be called when `other` is either a constant or a list.  I think the right fix is to place a test in `wrapper` that says if the `self.dtype` is an object, then call `na_op()` directly.  

But maybe that's not the best way to do it??

#### Expected Output
exception raised for comparison to Series  invalid type comparison
exception raised for comparison to list invalid type comparison
exception raised for comparison to int invalid type comparison
no exception raised for comparison to IPType

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.23.0.dev0+683.g402ad45da
pytest: 3.4.0
pip: 9.0.1
setuptools: 38.5.1
Cython: 0.25.1
numpy: 1.14.1
scipy: 1.0.0
pyarrow: 0.8.0
xarray: None
IPython: 6.2.1
sphinx: 1.7.1
patsy: 0.5.0
dateutil: 2.6.1
pytz: 2018.3
blosc: 1.5.1
bottleneck: 1.2.1
tables: 3.4.2
numexpr: 2.6.4
feather: None
matplotlib: 2.2.0
openpyxl: 2.5.0
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.2
lxml: 4.1.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.5
pymysql: 0.8.0
psycopg2: None
jinja2: 2.10
s3fs: 0.1.3
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

### LINKED ISSUE #20761: ExtensionArray._can_hold_na needs to be a class attribute
#### Problem description

`ExtensionArray._can_hold_na` is implemented as an instance property, but needs to be a class attribute to be consistent with its use throughout pandas.

(Note - I am working on a PR that will take care of this)

<details>
INSTALLED VERSIONS
------------------
commit: f463cf26b35ec11ae4fdf332635d6e89a9db2f17
python: 3.6.4.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.23.0.dev0+747.gf463cf26b.dirty
pytest: 3.4.0
pip: 9.0.1
setuptools: 38.5.1
Cython: 0.25.1
numpy: 1.14.1
scipy: 1.0.0
pyarrow: 0.8.0
xarray: None
IPython: 6.2.1
sphinx: 1.7.1
patsy: 0.5.0
dateutil: 2.6.1
pytz: 2018.3
blosc: 1.5.1
bottleneck: 1.2.1
tables: 3.4.2
numexpr: 2.6.4
feather: None
matplotlib: 2.2.0
openpyxl: 2.5.0
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.2
lxml: 4.1.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.5
pymysql: 0.8.0
psycopg2: None
jinja2: 2.10
s3fs: 0.1.3
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

Dr-Irv:
closes #20659
closes #20761
- [x] tests added / passed
  - New tests/extensions/relobject tests arrays that cannot hold NaN, and has relational operators defined
- [ ] whatsnew entry
  - Not added as this is part of the ExtensionArray support coming in the next release, which already has a whatsnew entry

NOTE: I expect that the pandas main developers will object to a lot of what is done here!  So let's be open on the discussion.

## Goals of this pull request:

1. If there is an `ExtensionArray`, allow the relational operators to delegate to the base type
2. Modify `ExtensionArray` tests to handle the cases of `_can_hold_na==False` as best as possible.
3. An eventual goal is that if I have a `RelObjectArray` that consists of objects that have operators such as `__le__` defined that return objects, and `a` and `b` are `Series` containing that type of array, then `a <= b` returns a `Series` of objects rather than booleans.

### Things I'm unsure of:

- In core/ops.py, I am using `inspect` to determine whether the user has implemented the relational operators for the subclassed `ExtensionArray` type as well as for the base type of elements in the array. I'm not sure if this is best practice or not.
- In my example that I'm separately working on, the relational operators (including `__eq__`) return objects as opposed to booleans, and some (intentionally) throw exceptions.  So I tried to catch these various cases.  Not having `__lt__` defined means that sorting is undefined, and not having `__eq__` return a boolean messes up some things related to `groupby`.  For my application, that is OK.


TomAugspurger:
> I am using inspect to determine whether the user has implemented the relational operators for the subclassed ExtensionArray type as well as for the base type of elements in the array. I'm not sure if this is best practice or not.

One issue here is for `__eq__` and `__neq__`. All EAs will implement those by virtue of inheriting from object, even if the author doesn't implement it "correctly". We could have EA authors opt into this by defining a class attributes like `_equatable`, `_orderable`, etc.

pandas/core/arrays/base.py
```diff
@@ -38,11 +38,21 @@ class ExtensionArray(object):
...
-    Some additional methods are available to satisfy pandas' internal, private
-    block API:
+    An additional method and attribute is available to satisfy pandas'
+    internal, private block API.
 
-    * _can_hold_na
     * _formatting_values
+    * _can_hold_na
+
+    Some methods require casting the ExtensionArray to an ndarray of Python
```
TomAugspurger: Merge snafu?

pandas/core/arrays/base.py
```diff
@@ -560,16 +571,13 @@ def _concat_same_type(cls, to_concat):
         """
         raise AbstractMethodError(cls)
 
-    @property
-    def _can_hold_na(self):
-        # type: () -> bool
-        """Whether your array can hold missing values. True by default.
+    _can_hold_na = True
+    """Whether your array can hold missing values. True by default.
```
TomAugspurger: Not sure what our recommended style is here. You can probably just change it to a comment.

pandas/tests/extension/base/getitem.py
```diff
@@ -82,8 +82,9 @@ def test_getitem_scalar(self, data):
         assert isinstance(result, data.dtype.type)
 
     def test_getitem_scalar_na(self, data_missing, na_cmp, na_value):
-        result = data_missing[0]
-        assert na_cmp(result, na_value)
+        if data_missing._can_hold_na:
```
TomAugspurger: All these changes are a bit unfortunate...

Could we instead have the `data_missing` fixture raise `pytest.skip` when necessary? I suspect this would have to be done by the EA author in their code.

pandas/tests/extension/relobject/test_relobject.py
```diff
@@ -0,0 +1,268 @@
...
+    return RelObjectArray([RelObj(10), RelObj(5)])
+
+
+@pytest.fixture
+def data_for_sorting():
+    return RelObjectArray([RelObj(4), RelObj(5), RelObj(3)])
+
+
+@pytest.fixture
+def data_missing_for_sorting():
```
TomAugspurger: Ah, I clearly didn't have arrays that don't support NA in mind when I wrote this :)

Could you add this note to the `data_missing_for_sorting` docstring (in `tests/extensions/conftest.py`).?

TomAugspurger (commented):
The NA stuff looks good. Small question about the test changes.

Taking a look at the ops stuff in a bit.

pandas/core/arrays/base.py
Dr-Irv: No, I reordered `_formatting_values` and `_can_hold_na`, since one is a method and the other is an attribute.

TomAugspurger: I meant this section specifically. This block is repeated starting on line 57. Or perhaps I'm missing something.

pandas/tests/extension/base/getitem.py
Dr-Irv: I could make that change to avoid any tests where `data_missing` is used, and the EA author would have to make that change if they wanted `_can_hold_na=False`.

pandas/core/arrays/base.py
Dr-Irv: This is really weird.  The copy on my machine is clean and fine, but the copy on GitHub has the repeat.  Maybe because I resolved conflicts with master using GitHub interface.  UGH.

TomAugspurger:
> this won't be for 0.23

Yeah, I think I agree given the deadline we set (RC was supposed to be yesterday). I could see this taking a bit of time to get right.

@Dr-Irv could you split the changes for `_can_hold_na` out to a separate PR? That part seems ready to go.l

Dr-Irv:
@TomAugspurger Yes, I can split the changes out for the `_can_hold_na` part.  And I see that the operator stuff I tried to do is part of a bigger discussion.  Should I create a new test case for the `_can_hold_na` part?

TomAugspurger:
Yeah, that'd be nice.

It can be a very simple example I think... Maybe make a subclass of
`DecimalArray` with `_can_hold_na=False` and see if the tests pass once you
apply your changes?

Issue #17261, opened 2017-08-15
API: Use object dtype for empty Series


TomAugspurger:
Currently an empty `Series` has a float dtype:

```python
In [15]: pd.Series().dtype
Out[15]: dtype('float64')
```

This is consistent with NumPy, but inconsistent with the rest of pandas:

```python
In [18]: pd.Index([]).dtype
Out[18]: dtype('O')

In [19]: pd.DataFrame(columns=['A']).dtypes
Out[19]:
A    object
dtype: object
```

I think it's best to just make all our containers treat empty as object dtype, unless explicitly told otherwise (excluding specialized containers like `Int64Index`).

jreback:
xref https://github.com/pandas-dev/pandas/issues/1783

TomAugspurger:
This will potentially fix https://github.com/pandas-dev/pandas/issues/16737 as well

jreback:
> This will potentially fix #16737 as well

not at all, that is completely different, but #16961 is a PR for that issue.

TomAugspurger:
Probably not getting to this for 0.24.

TomAugspurger:
Pushing again :/ Maybe forever at this point?

SaturnFromTitan:
I'm looking into this

PR #20802, opened 2018-04-23
DEPR: Warn on empty Series dtype inference

TomAugspurger:
We're changing `Series([])` to be `object` instead of `float`. This does the warning

- [x] xref https://github.com/pandas-dev/pandas/issues/17261 (leave open till we make the change)
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -862,6 +862,7 @@ Other API Changes
 Deprecations
 ~~~~~~~~~~~~
 
+- The inferred dtype for an empty Series will change from ``float`` to ``object`` (:issue:`17261`).
 - ``Series.from_array`` and ``SparseSeries.from_array`` are deprecated. Use the normal constructor ``Series(..)`` and ``SparseSeries(..)`` instead (:issue:`18213`).
 - ``DataFrame.as_matrix`` is deprecated. Use ``DataFrame.values`` instead (:issue:`18458`).
 - ``Series.asobject``, ``DatetimeIndex.asobject``, ``PeriodIndex.asobject`` and ``TimeDeltaIndex.asobject`` have been deprecated. Use ``.astype(object)`` instead (:issue:`18572`)
--- a/pandas/core/base.py
+++ b/pandas/core/base.py
@@ -954,7 +954,11 @@ class IndexOpsMixin(object):
                 # we specify the keys here to handle the
                 # possibility that they are tuples
                 from pandas import Series
-                mapper = Series(mapper)
+                if len(mapper) == 0:
+                    mapper_dtype = 'float'
+                else:
+                    mapper_dtype = None
+                mapper = Series(mapper, dtype=mapper_dtype)
 
         if isinstance(mapper, ABCSeries):
             # Since values were input this means we came from either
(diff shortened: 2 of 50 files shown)


TomAugspurger:
FYI, d80e1c7dea07303c2aacef32afbd6ccfc6fd47bc has the changes. Everything else is catching / fixing warning internal to pandas. This ended up being a decent-sized change.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -862,6 +862,7 @@ Other API Changes
 Deprecations
 ~~~~~~~~~~~~
 
+- The inferred dtype for an empty Series will change from ``float`` to ``object`` (:issue:`17261`).
```
jreback: double-backticks on Series

pandas/core/base.py
```diff
@@ -954,7 +954,11 @@ def _map_values(self, mapper, na_action=None):
                 # we specify the keys here to handle the
                 # possibility that they are tuples
                 from pandas import Series
-                mapper = Series(mapper)
+                if len(mapper) == 0:
```
jreback: I think this might be worthwhile to make a function in pandas.core.dtypes.cast.maybe_infer_1d_array_dtype

pandas/core/generic.py
```diff
@@ -5399,7 +5399,11 @@ def fillna(self, value=None, method=None, axis=None, inplace=False,
             if self.ndim == 1:
                 if isinstance(value, (dict, ABCSeries)):
                     from pandas import Series
-                    value = Series(value)
+                    if len(value) == 0:
+                        series_dtype = 'float'
```
jreback: see above

pandas/core/indexes/base.py
```diff
@@ -1313,33 +1312,9 @@ def _get_names(self):
...
-        """
-        Set new names on index. Each name has to be a hashable type.
-
-        Parameters
-        ----------
-        values : str or sequence
-            name(s) to set
-        level : int, level name, or sequence of int/level names (default None)
-            If the index is a MultiIndex (hierarchical), level(s) to set (None
-            for all levels).  Otherwise level must be None
```
jreback: looks like you have some extra commits?

pandas/tests/series/test_apply.py
```diff
@@ -388,10 +388,14 @@ def test_map(self):
 
     @pytest.mark.parametrize("index", tm.all_index_generator(10))
     def test_map_empty(self, index):
-        s = Series(index)
+        if len(index) == 0:
```
jreback: use function as above

jreback (changes requested):
generally looks good. looks like you have some extra commits (so rebase on master), and a couple of style changes.

TomAugspurger:
Sorry about the empty commits, will fix that up.

Still going through and squashing warnings. There are a few cases internally where we construct an empty series. e.g.

```python
In [1]: import pandas as pd

In [2]: pd.Series([], dtype='object').groupby([]).apply(lambda x: 1)
Out[2]: Series([], dtype: float64)
```

How should we handle that?

1. Leave as float64, specify the dtype to keep compat
2. Leave as float64, no dtype to generate a warning

The problem w/ 2 is that there's nothing the user can do to silence the warning.

jreback:
@TomAugspurger can you merge master

TomAugspurger:
There's a decent amount of work to do here to avoid the warnings
internally. Will try to make time.

jreback:
@TomAugspurger keep this one open?

TomAugspurger:
Probably not :/ I suppose we can just live with this. A little artifact of the time Series used to be an ndarray.

PR #20811, opened 2018-04-24
Fixed typo

mmxmb:
I believe it should be either `Then another Python operation ``dfmi_with_one['second']`` happens.` or `Then another Python operation ``dfmi_with_one['second']`` selects the series indexed by ``'second'``.` .
Not `Then another Python operation ``dfmi_with_one['second']`` selects the series indexed by ``'second'`` happens.`.


jreback:
thanks!

Issue #20812, opened 2018-04-24
TST: Cleanup tests/indexes/test_base.py


WillAyd:
This module could use some major refactoring. Approach outlined with checkboxes below - will add or modify as we go through:

- [X] Parametrize the existing module where appropriate
    - [X] Pass 1: #20624
    - [X] Pass 2: #20813
    - [X] Pass 3: #20931
    - [X] Pass 4: #20979
- [ ] Replace any class methods / attributes with fixtures where appropriate
- [ ] Move tests into new modules / sub-modules (ex: https://github.com/pandas-dev/pandas/pull/20624#discussion_r180617229, https://github.com/pandas-dev/pandas/pull/20624#discussion_r180617681, https://github.com/pandas-dev/pandas/pull/20931#discussion_r185675690)
- [ ] Add fixture for fill arguments (see https://github.com/pandas-dev/pandas/pull/20931#discussion_r185675215)

jbrockmendel:
All the PRs linked in the OP are merged, and I've done several passes on this file since then too.  Closing.

PR #20813, opened 2018-04-24
CLN: Continued indexes/test_base.py cleanup

WillAyd:
progress towards #20812

### DIFF
--- a/pandas/tests/indexes/test_base.py
+++ b/pandas/tests/indexes/test_base.py
@@ -1108,18 +1108,18 @@ class TestIndex(Base):
 
         self.strIndex[:0].format()
 
-    def test_format_missing(self, nulls_fixture):
+    @pytest.mark.parametrize("vals", [
+        [1, 2.0 + 3.0j, 4.], ['a', 'b', 'c']])
+    def test_format_missing(self, vals, nulls_fixture):
         # 2845
-        index = Index([1, 2.0 + 3.0j, nulls_fixture])
+        vals = list(vals)  # Copy for each iteration
+        vals.append(nulls_fixture)
+        index = Index(vals)
+
         formatted = index.format()
-        expected = [str(index[0]), str(index[1]), u('NaN')]
+        expected = [str(index[0]), str(index[1]), str(index[2]), u('NaN')]
+
         assert formatted == expected
-
-    def test_format_missing_obj(self, nulls_fixture):
-        values = ['a', 'b', 'c', nulls_fixture]
-
-        index = Index(values)
-        index.format()
         assert index[3] is nulls_fixture
 
     def test_format_with_name_time_info(self):
--- a/pandas/tests/indexes/test_base.py
+++ b/pandas/tests/indexes/test_base.py
@@ -1227,7 +1227,7 @@ class TestIndex(Base):
         index = Index(np.arange(10))
 
         actual = index.get_indexer([0.2, 1.8, 8.5], method='nearest',
-                                 tolerance=listtype(tolerance))
+                                   tolerance=listtype(tolerance))
         tm.assert_numpy_array_equal(actual, np.array(expected,
                                                      dtype=np.intp))
 
@@ -1238,7 +1238,7 @@ class TestIndex(Base):
 
         with pytest.raises(ValueError, match='tolerance size must match'):
             index.get_indexer([1, 0], method='nearest',
-                            tolerance=[1, 2, 3])
+                              tolerance=[1, 2, 3])
 
     @pytest.mark.parametrize("method,expected", [
         ('pad', [8, 7, 0]), ('backfill', [9, 8, 1]), ('nearest', [9, 7, 0])])
@@ -1271,7 +1271,7 @@ class TestIndex(Base):
 
         with pytest.raises(TypeError):
             index.get_indexer(['a', 'b', 'c', 'd'], method='pad',
-                            tolerance=[2, 2, 2, 2])
+                              tolerance=[2, 2, 2, 2])
 
     def test_get_indexer_numeric_index_boolean_target(self):
         # GH 16877
--- a/pandas/tests/indexes/test_base.py
+++ b/pandas/tests/indexes/test_base.py
@@ -187,18 +187,18 @@ class TestIndex(Base):
         df = pd.DataFrame(np.random.rand(5, 3))
         df['date'] = dts
         result = DatetimeIndex(df['date'], freq='MS')
-        assert df['date'].dtype == object
 
+        assert df['date'].dtype == object
         expected.name = 'date'
-        exp = pd.Series(dts, name='date')
-        tm.assert_series_equal(df['date'], exp)
+        tm.assert_index_equal(result, expected)
+
+        expected = pd.Series(dts, name='date')
+        tm.assert_series_equal(df['date'], expected)
 
         # GH 6274
         # infer freq of same
         freq = pd.infer_freq(df['date'])
         assert freq == 'MS'
-
-        tm.assert_index_equal(result, expected)
 
     @pytest.mark.parametrize("array", [
         np.arange(5), np.array(['a', 'b', 'c']), date_range(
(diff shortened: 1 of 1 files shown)


pandas/tests/indexes/test_base.py
```diff
@@ -678,61 +678,53 @@ def test_intersection(self):
...
-        expected = Index(['A', 'B', 'A'], dtype='object')
-        result = idx1.intersection(idx2)
-        tm.assert_index_equal(result, expected)
+        assert res.name == exp.name
+        tm.assert_index_equal(res, exp)
 
-        # preserve names
+    @pytest.mark.parametrize("first_nm,sec_nm,exp_nm", [
+        ('A', 'A', 'A'), ('A', 'B', None), (None, 'B', None)])
+    def test_intersection_name_preservation2(self, first_nm, sec_nm, exp_nm):
```
WillAyd: Kept as a separate test for now, but ideally should be combined with the test above as they cover the same topics. Waiting until I get around to replacing the index class attrs with fixtures to explore in more detail

pandas/tests/indexes/test_base.py
```diff
@@ -878,24 +845,23 @@ def test_map_with_tuples(self):
...
         #   returns an Index.
         first_level = ['foo', 'bar', 'baz']
         multi_index = MultiIndex.from_tuples(lzip(first_level, [1, 2, 3]))
         reduced_index = multi_index.map(lambda x: x[0])
         tm.assert_index_equal(reduced_index, Index(first_level))
 
-    def test_map_tseries_indices_return_index(self):
-        date_index = tm.makeDateIndex(10)
+    @pytest.mark.parametrize("attr", [
+        'makeDateIndex', 'makePeriodIndex', 'makeTimedeltaIndex'])
```
WillAyd: Will most likely replace with a fixture in the future, but made sense to parametrize for now

pandas/tests/indexes/test_base.py
```diff
@@ -996,30 +958,38 @@ def test_iadd_string(self):
         index += '_x'
         assert 'a_x' in index
 
-    def test_difference(self):
-
+    @pytest.mark.parametrize("second_nm,exp", [
+        (None, None), ('name', 'name')])
```
WillAyd: Might be over-engineering the parameters in this case, but figure if we wanted to expand to cover more scenarios this is a better framework to do so

pandas/tests/indexes/test_base.py
```diff
@@ -1092,25 +1058,23 @@ def test_intersection_difference(self):
...
-        assert not self.floatIndex.is_object()
-
-    def test_is_all_dates(self):
-        assert self.dateIndex.is_all_dates
-        assert not self.strIndex.is_all_dates
-        assert not self.intIndex.is_all_dates
+    @pytest.mark.parametrize("attr,exp", [
+        ('strIndex', False), ('boolIndex', False), ('catIndex', False),
+        ('intIndex', True), ('dateIndex', False), ('floatIndex', True)])
+    def test_is_numeric(self, attr, exp):
```
WillAyd: This and the two tests below it did not consistently test all of the index types so I've added where missing for better coverage. Should ultimately be replaced with fixtures but same concept will apply

pandas/tests/indexes/test_base.py
```diff
@@ -1142,19 +1106,21 @@ def test_format(self):
             expected = [str(index[0])]
             assert formatted == expected
 
+        self.strIndex[:0].format()
+
+    def test_format_missing(self, nulls_fixture):
```
WillAyd: Was only testing `np.nan` and `None` before but figured this was a good use case for the new nulls fixture

pandas/tests/indexes/test_base.py
```diff
@@ -1142,19 +1106,21 @@ def test_format(self):
...
         formatted = index.format()
         expected = [str(index[0]), str(index[1]), u('NaN')]
         assert formatted == expected
 
-        # is this really allowed?
-        index = Index([1, 2.0 + 3.0j, None])
-        formatted = index.format()
-        expected = [str(index[0]), str(index[1]), u('NaN')]
-        assert formatted == expected
+    def test_format_missing_obj(self, nulls_fixture):
```
WillAyd: Trying to balance too many changes but am leaning towards combining this with the above test. Both are pretty similar but leave strange gaps in what they test across the float and object types

pandas/tests/indexes/test_base.py
```diff
@@ -678,61 +678,53 @@ def test_intersection(self):
...
 
-        idx2 = Index(['B', 'D', 'A'])
-        expected = Index(['A', 'B', 'A'], dtype='object')
-        result = idx1.intersection(idx2)
-        tm.assert_index_equal(result, expected)
+        assert res.name == exp.name
+        tm.assert_index_equal(res, exp)
 
-        # preserve names
+    @pytest.mark.parametrize("first_nm,sec_nm,exp_nm", [
```
jreback: ideally can you spell things out a bit more, e.g. ``first_name, second_name, expected_name`` are just fine

pandas/tests/indexes/test_base.py
```diff
@@ -678,61 +678,53 @@ def test_intersection(self):
...
+        first.name = first_nm
+        second.name = sec_nm
         intersect = first.intersection(second)
-        assert intersect.name == 'A'
+        assert intersect.name == exp_nm
 
-        second.name = 'B'
-        intersect = first.intersection(second)
-        assert intersect.name is None
+    @pytest.mark.parametrize("idx2,keeps_nm", [
```
jreback: so ``keeps_nm`` -> ``keeps_name`` and normalize the input params, e.g. ``index`` (or index2 or other) as appropriate

pandas/tests/indexes/test_base.py
```diff
@@ -744,19 +736,31 @@ def test_intersect_str_dates(self):
...
     def test_union(self):
+        # TODO: Replace with fixtures
         first = self.strIndex[5:20]
         second = self.strIndex[:10]
         everything = self.strIndex[:20]
+
         union = first.union(second)
         assert tm.equalContents(union, everything)
 
+    @pytest.mark.parametrize("klass", [
```
jreback: this notion of klass we have all over the place, sometimes its also called box (not necessary to do here, just letting you know, prob best to update the issue you created)

pandas/tests/indexes/test_base.py
```diff
@@ -766,61 +770,23 @@ def test_union(self):
...
-
-        first = Index(list('ab'), name='A')
-        second = Index([])
-        union = first.union(second)
-        expected = Index(list('ab'), name='A')
-        tm.assert_index_equal(union, expected)
-
-        first = Index([], name='A')
-        second = Index(list('ab'))
+    @pytest.mark.parametrize("first,second,nm", [
```
jreback: maybe could replace this with 2 parameterize (for a cartesian product of the inputs)

pandas/tests/indexes/test_base.py
```diff
@@ -967,18 +933,14 @@ def test_append_multiple(self):
         result = index.append([])
         tm.assert_index_equal(result, index)
 
-    def test_append_empty_preserve_name(self):
+    @pytest.mark.parametrize("nm,exp", [
```
jreback: ``name, expected``

pandas/tests/indexes/test_base.py
jreback: ok by me

pandas/tests/indexes/test_base.py
jreback: sure

pandas/tests/indexes/test_base.py
```diff
@@ -678,61 +678,53 @@ def test_intersection(self):
         inter = first.intersection(first)
         assert inter is first
 
+    @pytest.mark.parametrize("idx2,keeps_nm", [
+        (Index([3, 4, 5, 6, 7], name="idx"), True),  # preserve same name
+        (Index([3, 4, 5, 6, 7], name="other"), False),  # drop diff names
+        (Index([3, 4, 5, 6, 7]), False)])
+    def test_intersection_name_preservation(self, idx2, keeps_nm):
```
jreback: side note, we prob need to split out test_base into multiple files its growing pretty big (do this as a separate step), these might be things like: test_indexing.py, test_missing.py, test_set_ops.py etc, in essence mirroring the various sub-files we do for series ops

jreback (changes requested):
looks good. if can fix the naming then looks good to merge

pandas/tests/indexes/test_base.py
WillAyd: Yep - I have this as the last TODO in the referenced issue so we don't lose sight of it

pandas/tests/indexes/test_base.py
WillAyd: Would you prefer generally using `result` and `expected` over `res` and `exp`? I've seen both and typically use the latter but if we have a preference we could start being explicit about those standards (maybe in TESTING_README)

jreback (approved):

jreback:
thanks @WillAyd

PR #20814, opened 2018-04-24
ExtensionArray.take default implementation

### LINKED ISSUE #20640: API: take interface for (Extension)Array-likes
Triggered by https://github.com/pandas-dev/pandas/pull/20582, I was looking at the `take` implementation in ExtensionArray and Categorical (which is already an ExtensionArray subclass) and in the rest of pandas:

- `ExtensionArray.take` currently uses the "internal pandas"-like behaviour for take: `-1` is an indicator for missing value (the behaviour we need for reindexing etc)
- `Series.take` actually uses the numpy behaviour, where negative values (including `-1`) start counting from the end of the array-like.

To illustrate the difference with a small example: 

```
In [9]: pd.Categorical(['a', 'b', 'c']).take([0, -1])
Out[9]: 
[a, NaN]
Categories (3, object): [a, b, c]

In [10]: pd.Series(['a', 'b', 'c']).take([0, -1])
Out[10]: 
0    a
1    c
dtype: object
```

This difference is a bit unfortunate IMO. If `ExtensionArray.take` is a public method (which it is right now), it would be nice if it has consistent behaviour with `Series.take`. 
If we agree on that, I was thinking about following options:

- make `ExtensionArray.take` private for now (eg require a `_take` method for the interface) and keep the "internal pandas"-like behaviour
- make `ExtensionArray.take` *default* behaviour consistent with `Series.take`, but still have the `allow_fill`/`fill_value` arguments so that *when* they are specified it has the "internal pandas"-like behavour (so that internal code that expects this behaviour which already passes those keywords keeps working)

TomAugspurger:
Implements a take interface that's compatible with NumPy and optionally pandas'
NA semantics.

```python
In [1]: import pandas as pd

In [2]: from pandas.tests.extension.decimal.array import *

In [3]: arr = DecimalArray(['1.1', '1.2', '1.3'])

In [4]: arr.take([0, 1, -1])
Out[4]: DecimalArray(array(['1.1', '1.2', '1.3'], dtype=object))

In [5]: arr.take([0, 1, -1], fill_value=float('nan'))
Out[5]: DecimalArray(array(['1.1', '1.2', Decimal('NaN')], dtype=object))
```

Closes https://github.com/pandas-dev/pandas/issues/20640

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


pandas/compat/__init__.py
```diff
@@ -451,3 +451,5 @@ def is_platform_mac():
 
 def is_platform_32bit():
     return struct.calcsize("P") * 8 < 64
+
+_default_fill_value = object()
```
TomAugspurger: This probably isn't the right place for this, but it has to be importable from `pandas/arrays/base.py` and `core/algorithms.py`

pandas/core/algorithms.py
```diff
@@ -1558,6 +1559,81 @@ def take_nd(arr, indexer, axis=0, out=None, fill_value=np.nan, mask_info=None,
 take_1d = take_nd
 
 
+def take_ea(arr, indexer, fill_value=_default_fill_value):
```
TomAugspurger: I threw this in `core/algorithms.py`. Does it make more sense in `core/arrays/base.py`, either as a standalone method or as a staticmethod on ExtensionArray?

pandas/core/algorithms.py
```diff
@@ -1558,6 +1559,81 @@ def take_nd(arr, indexer, axis=0, out=None, fill_value=np.nan, mask_info=None,
...
+    """Extension-array compatible take.
```
TomAugspurger: Just copy-pased the docstring for now. Could do a shared doc if moved.

pandas/core/arrays/base.py
```diff
@@ -462,22 +464,38 @@ def factorize(self, na_sentinel=-1):
...
-        # type: (Sequence[int], bool, Optional[Any]) -> ExtensionArray
+    def _values_for_take(self):
+        """Values to use for `take`.
+
+        Coerces to object dtype by default.
+
+        Returns
+        -------
+        array-like
+            Must satisify NumPy's `take` semantics.
```
TomAugspurger: n.b.: this doesn't have to be an ndarray. Anything that satisfies NumPy's `take` (and I suppose masking) semantics will work.

TomAugspurger (commented):
Marked as WIP for now, because something is going strange with JSONArray. NumPy doesn't like us trying to shove dictionaries into scalar spots of an ndarray I think. Still debugging.

jreback (changes requested):

pandas/core/algorithms.py
```diff
@@ -1558,6 +1559,81 @@ def take_nd(arr, indexer, axis=0, out=None, fill_value=np.nan, mask_info=None,
...
+    ----------
+    arr : array-like
+        Must satisify NumPy's take semantics.
+    indexer : sequence of integers
+        Indices to be taken. See Notes for how negative indicies
+        are handled.
+    fill_value : any, optional
+        Fill value to use for NA-indicies. This has a few behaviors.
+
+        * fill_value is not specified : triggers NumPy's semantics
```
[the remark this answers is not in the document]
TomAugspurger: Although, there are two things going on here:

1. `np.nan` being swapped for `ExtensionArray.dtype.na_value`.
2. `fill_value` being overloaded to switch the meaning of negative indexers.

Which bit is confusing? If I'm able to change things so that `fill_value` is simply passed through, would that be OK? Then someone specifying `ExtensionArray.take(indexer, fill_value=thing)` would mean `-1` is an NA marker. `ExtensionArray.take(indexer)` would be `-1` is the last element.

jorisvandenbossche:
Quick general comment: why make a EA specific take method instead of just a general one that works for both ndarrays and EAs? (and for EA it would just dispatch to the EA). 

I am also not sure if it is worth to add the `_values_for_take`. If we have a general purpose `take` function, an `ExtensionArray.take` implementation could simply be (assuming `self.data` is the underlying numpy array):

```
           def take(self, indexer, fill_value=None):
               ... somehow handle default fill_value ...
               result = take(self.data, indexer, fill_value=fill_value)
               return self.__class__(result)
```

I think it is not too much asked for an ExtensionArray author to implement this themselves.

There is also a problem with what a public `na_value` on ExtensionDtype/ExtensionArray means. Is it the value that is used "publicly"? Or is it the value that is used in the underlying array? Eg for geopandas the one is `None`, the other is `0` (but the 0 nobody should care about, as that is an implementation detail)

pandas/core/algorithms.py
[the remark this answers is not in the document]
jorisvandenbossche: Not overloading `fill_value` is how it is currently, with both `allow_fill` and `fill_value`. If we find that clearer, I am fine with going back to that (although it would be nice to only need one keyword for this).

I have more an objection on the np.nan swapping. It would be nice if we could prevent this, but didn't look in detail enough to assess whether this would be possible or not.

TomAugspurger:
> Quick general comment: why make a EA specific take method instead of just a general one that works for both ndarrays and EAs?


I think the way `take_ea` is written, it works on either.

```python
In [1]: from pandas.core.arrays.base import take_ea

In [2]: import numpy as np

In [3]: arr = np.array([1, 2, 3])

In [4]: take_ea(arr, [0, 1, -1])
Out[4]: array([1, 2, 3])
```

Suggestions for a better name?

> I think it is not too much asked for an ExtensionArray author to implement this themselves.

That seems fine.

> There is also a problem with what a public na_value on ExtensionDtype/ExtensionArray means.

This can be private.

jorisvandenbossche:
> I think the way take_ea is written, it works on either.

The `_from_sequence` in there is only for EAs

jorisvandenbossche:
How I had it in my head (but which doesn't mean it is better!), was that such a `take` method that you now wrote in algorithms.py would actually only work on numpy arrays (and on EAs as well, but for those we would dispatch), and then EA.take can call that function again but now with a numpy array of data.
Then the `take` method in algorithms.py would be "cleaner" in the sense that it doesn't need to be aware of EA's semantics. 
But trying to look now in more detail to what you did to better see advantages/disadvantages.

pandas/core/algorithms.py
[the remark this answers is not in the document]
jorisvandenbossche: Although, if we make sure to not pass through a default `fill_value`, you would in most cases only need to specify `allow_fill=True` to get pandas semantics, instead of needing to specify both `allow_fill` and `fill_value`. So maybe that is not that bad. 

Currenlty, eg `Series.reindex` calls under the hood (via Block.take_nd) `take_1d` on the Block values, by passing through the `Block.fill_value` as default `fill_value`. Could we make sure to not pass anything through in case of ExtensionBlocks, or to let the ExtensionBlock.fill_value be the appropriate value? (to avoid the swapping of `np.nan`)

pandas/core/algorithms.py
```diff
@@ -1558,6 +1559,81 @@ def take_nd(arr, indexer, axis=0, out=None, fill_value=np.nan, mask_info=None,
...
+                             "must be non-negative or -1. When "
+                             "'fill_value' is specified.")
+
+        # take on empty array not handled as desired by numpy
+        # in case of -1 (all missing take)
+        if not len(arr) and mask.all():
+            return arr._from_sequence([fill_value] * len(indexer))
+
+        result = arr.take(indexer)
+        result[mask] = fill_value
```
jorisvandenbossche: Is there a reason to not use the `take_1d` machinery here? (maybe there is, or maybe there is no reason to use it since the above is simpler and works fine, but was just wondering)

TomAugspurger:
We should be able to have a clean implementation of `take_ea` (or `take` or whatever).

The tricky part is fixing the callers like `Block.reindex_indexer` / `Block.get_reindexed_values` / `NDFrame.align`, to call things correctly. These need to call `algos.take(thing, fill_value=fill_value)` with the right fill value for that dtype. Things should work, but it's a bit tricky.

TomAugspurger:
Pushed another WIP if anyone has some thoughts on the general approach. The extension tests all pass but I'm still breaking things elsewhere, and I think things can be simplified.

The main idea is to have a sentinel `_default_na_value` that's used as the default for all reindexing operations (align, etc.). Then in internals, once we have a single-dtype block, we can figure out what default value to actually use based on the block type. In principle, this should work, but I'm struggling with various things, especially around upcasting. Hoping to clean it up today.

jorisvandenbossche:
Regarding `_default_fill_value`, I have the feeling (again, from only taking a quick look) that we are mixing two different concepts: for `take` it means "no filling", for `reindex` it means "default fill value"

jorisvandenbossche:
> That feeling is correct. Is that not what we want? I can re-add the allow_fill argument.

I am going a bit back and forth on it, not sure ...

Regarding what I said above about the dubious meaning of "na_value" (the physical value that is stored, or the public one), I think this is actually not really a problem. It should just be the public one, so the value that you get if you get an item of an EA that is missing (type of `arr[0]` if the first item is missing). So it is fine to keep this public.

jorisvandenbossche:
Trying to get my head around the current logic. Some pseudo code flow path in case of a Series reindex operation with EA values:

```
s.reindex(self, ..., fill_value=_default_fill_value)
-> ... -> BlockManager.reindex_indexer(..., fill_value=fill_value)
-> Block.take_nd(..., 
        fill_tuple=(fill_value if fill_value is not _default_fill_value else Block.fill_value,)
-> algos.take_nd(values, ... allow_fill=True, fill_value=fill_tuple[1])
-> [in case of EA] values.take(..., fill_value=fill_value)
   -> if fill_value == _default_fill_value -> no filling
   -> if fill_value == Block.fill_value -> default filling behaviour -> algos.take with passed fill_value
```

(but in case of a reindex/aligment operation, the final `fill_value` actively passed to `values.take(..)` will never be `_default_fill_value` I think?)

So what I find complex about this: a) as said above different meaning of `_default_fill_value` throughout the code base in reindex and take, b) the fact that Block is responsible of passing an appropriate filling value to EA.take, while the EA could perfectly know and do that itself and c) that an EA.take implementation needs to be aware of the ``_default_fill_value`` object (which means we need to make that public).

TomAugspurger:
> EA.take implementation needs to be aware of the _default_fill_value

I think if we use `allow_fill`, then we can just switch to `fill_value=None`. The issue with `fill_value=None` as the default is that an EA may want to use `None` as their NA value, and so wouldn't be able to tell if it's explicit or the default.

This should be able to be simplified greatly. Let's see...

jorisvandenbossche:
Just brainstorming on an alternative:

```
s.reindex(self, ..., fill_value=_default_fill_value)
-> ... -> BlockManager.reindex_indexer(..., fill_value=fill_value)
-> Block.take_nd(..., fill_tuple=(fill_value,)
-> algos.take_nd(values, ... allow_fill=True, fill_value=fill_tuple[1])
  -> if fill_value == _default_fill_value -> values.take(..., allow_fill=True) (let EA use its default)
  -> else -> values.take(..., allow_fill=True, fill_value=fill_value) (override EA default)
-> values.take(..., allow_fill=False, fill_value=None) # can use their own default for fill_value
   -> if allow_fill == False -> no filling
   -> if allow_fill == True -> filling behaviour 
      -> if fill_value == None -> fill_value = EA.dtype.na_value
```

(I don't know if this scheme is anywhere near clear for somebody that did not write it ...?)

This introduces again the second keyword, but eliminates double usecase of ``_default_fill_value``, and EA.take does not need to be aware of it (the EA author can still use a similar trick if they want to have None as a valid fill_value instead of denoting the default, but that is then up to the EA author).

But of course this is only a single code path, I don't know how it interacts with all other usages of the take machinery.

TomAugspurger:
> I think if we use allow_fill, then we can just switch to fill_value=None. The issue with fill_value=None as the default is that an EA may want to use None as their NA value, and so wouldn't be able to tell if it's explicit or the default.

Actually, I don't think that's true. There would still be the ambiguity. Anyway, we can leave that up to the EA author I think...

Your proposal looks sensible. I'll try to implement it.

TomAugspurger:
Getting there...

This will still fail on `DataFrame.merge(..., how='outer')` that requires reindexing the dataframes. Somewhere in the join units stuff we're determining that 

```python
(Pdb) join_units
[JoinUnit(ExtensionBlock: slice(0, 1, 1), 1 x 3, dtype: decimal, {1: array([ 0,  0,  1,  2, -1])})]
```

should have `np.nan` for it's `upcasted_na`.

TomAugspurger:
Alrighty, I think this is ready for another look. Overall, I think this is better. Thanks @jorisvandenbossche.

Some API questions:

- Name for the new `take` method?
- Remove `_values_for_take` and make `take` abstract again? With the idea that the new `take` method makes it straightforward?
- Are we OK with the names `fill_value` and `allow_fill` for the parameter in `EA.take` / `take`?

probably others.l

pandas/core/algorithms.py
```diff
@@ -1448,6 +1448,27 @@ def func(arr, indexer, out, fill_value=np.nan):
     return func
 
 
+def take(arr, indexer, fill_value=None, allow_fill=None):
```
jorisvandenbossche: reason to use `fill_value=None` instead of `fill_value=True` ?

pandas/core/arrays/base.py
```diff
@@ -462,22 +463,40 @@ def factorize(self, na_sentinel=-1):
...
+        """
+        return self.astype(object)
+
+    def take(self, indexer, fill_value=None, allow_fill=None):
```
jorisvandenbossche: Switch order of fill_value / allow_fill keywords? (and same comment about `allow_fill` default)

pandas/core/arrays/base.py
```diff
@@ -462,22 +463,40 @@ def factorize(self, na_sentinel=-1):
...
+            Fill value to use for NA-indicies when `allow_fill` is True.
+            This may be ``None``, in which case the default NA value for
+            the type, ``self.dtype.na_value``, is used.
+        allow_fill : bool, optional
+            How to handle negative values in `indexer`.
+
+            For False values (the default), NumPy's behavior is used. Negative
+            values in `indexer` mean slices from the right.
+
+            For True values, Pandas behavior is used. Indicies where `indexer`
```
jorisvandenbossche: in external documentation, I would not use "pandas behaviour", as this is only for our internal `take` algos. `Series.take` has also numpy behaviour

pandas/core/arrays/base.py
```diff
@@ -487,44 +506,29 @@ def take(self, indexer, allow_fill=True, fill_value=None):
...
-               # in case of -1 (all missing take)
-               if not len(self) and mask.all():
-                   return type(self)([np.nan] * len(indexer))
-
-               result = self.data.take(indexer)
-               result[mask] = np.nan  # NA for this type
-               return type(self)(result)
+        ExtensionArray.take is called by ``Series.__getitem__``, ``.loc``,
+        ``iloc``, when the indexer is a sequence of values. Additionally,
+        it's called by :meth:`Series.reindex` with a `fill_value`.
```
jorisvandenbossche: maybe add after reindex something like "and any operation that causes alignment"

pandas/core/dtypes/base.py
```diff
@@ -16,6 +16,10 @@ class _DtypeOpsMixin(object):
     # classes will inherit from this Mixin. Once everything is compatible, this
     # class's methods can be moved to ExtensionDtype and removed.
 
+    # na_value is the default NA value to use for this type. This is used in
+    # e.g. ExtensionArray.take.
+    na_value = np.nan  # TODO: change to _na_value
```
jorisvandenbossche: remove the TODO?

pandas/core/dtypes/cast.py
```diff
@@ -255,9 +255,12 @@ def changeit():
 
 
 def maybe_promote(dtype, fill_value=np.nan):
-
     # if we passed an array here, determine the fill value by dtype
-    if isinstance(fill_value, np.ndarray):
+    if is_extension_array_dtype(dtype):
+        # XXX: verify this change
+        fill_value = dtype.na_value
```
jorisvandenbossche: this should only be done when `fill_value` is NaN? (similar check for datetime-likes to overwrite it with iNaT) 

As when you fill with other values, it should still check it?

pandas/core/internals.py
```diff
@@ -5434,6 +5443,13 @@ def is_uniform_join_units(join_units):
         len(join_units) > 1)
 
 
+def is_uniform_reindex(join_units):
+    return (
+        # TODO: should this be ju.block.can_hold_na?
+        all(ju.block and ju.block.is_extension for ju in join_units) and
+        len(set(ju.block.dtype.name for ju in join_units)) == 1
```
jorisvandenbossche: Is this checking of only the name potentially dangerous for parameterized dtypes like categorical?

Dr-Irv:
@TomAugspurger I think the `ExtensionDtype.na_value` is needed and required so that the only NA values inside of an `ExtensionArray` are that specific value.  This relates to making it so that `_can_hold_na=True` for all `ExtensionArray` types (given the discussion at #20815

pandas/tests/extension/base/getitem.py
```diff
@@ -134,12 +134,29 @@ def test_take(self, data, na_value, na_cmp):
 
     def test_take_empty(self, data, na_value, na_cmp):
         empty = data[:0]
-        result = empty.take([-1])
-        na_cmp(result[0], na_value)
+        # result = empty.take([-1])
```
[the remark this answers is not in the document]
jorisvandenbossche: Yes, it should then raise (as the indexer is out of bounds?). Numpy and pandas do allow empty takes on empty values:

```
In [93]: np.array([]).take([])
Out[93]: array([], dtype=float64)

In [94]: pd.Series([]).take([])
Out[94]: Series([], dtype: float64)

In [95]: np.array([]).take([0])
...
IndexError: cannot do a non-empty take from an empty axes.
```

But we should keep a test like this for the case of `allow_fill=True`

pandas/tests/extension/base/reshaping.py
```diff
@@ -124,11 +124,11 @@ def test_merge(self, data, na_value):
...
-        res = pd.merge(df1, df2)
-        exp = pd.DataFrame(
-            {'int1': [1, 1, 2], 'int2': [1, 2, 3], 'key': [0, 0, 1],
-             'ext': data._from_sequence([data[0], data[0], data[1]])})
-        self.assert_frame_equal(res, exp[['ext', 'int1', 'key', 'int2']])
+        # res = pd.merge(df1, df2)
+        # exp = pd.DataFrame(
+        #     {'int1': [1, 1, 2], 'int2': [1, 2, 3], 'key': [0, 0, 1],
+        #      'ext': data._from_sequence([data[0], data[0], data[1]])})
+        # self.assert_frame_equal(res, exp[['ext', 'int1', 'key', 'int2']])
```
jorisvandenbossche: this failed?

pandas/tests/extension/decimal/test_decimal.py
```diff
@@ -108,7 +108,26 @@ class TestReshaping(BaseDecimal, base.BaseReshapingTests):
 
 
 class TestGetitem(BaseDecimal, base.BaseGetitemTests):
-    pass
+
+    def test_take_basic(self):
```
jorisvandenbossche: Move this to the generic base tests?

jorisvandenbossche (commented):
Some first comments, but looks indeed cleaner!

jorisvandenbossche:
> 2. Remove _values_for_take and make take abstract again? With the idea that the new take method makes it straightforward?

I would personally remove `_values_for_take` and just have a `take`, to keep the interface simpler. I think most actual ExtensionArrays will not be able to use the `_values_for_take` anyhow, and will need to override `take` itself. This will be the case for geopandas, and I think also for cyberpandas? (as you will want to fill the underlying array with a native numpy scalar, and not with the public `na_value` (ipaddress.IPv4Address(0))?) 
If that is the case, there is even less reason to have it I think.

pandas/core/algorithms.py
TomAugspurger: I assume you mean `allow_fill=True`? It should be a bool, not None.

What do we want the default to be? I thought `allow_fill=False`, to be compatible with ndarray makes sense.

jorisvandenbossche: yes, sorry, both correct :-) I meant `allow_fill`, and `False`

pandas/core/arrays/base.py
TomAugspurger: Why switch the order?

jorisvandenbossche: So we do it in this order currently in algos. But, I think I find the reverse more natural, because `fill_value` is dependent on `allow_fill`. 
If I would explain the behaviour of those keywords, I would first say: "depending on allow_fill an indexer of -1 is interpreted as .. or ...", and only then "if allow_fill is True, with fill_value you can specify another value than the default missing value to fill in the non-existing values"

Here I said "allow_fill" correctly :-)

pandas/core/internals.py
TomAugspurger: Yes, I meant to go back and check that :) I haven't figured out how to best handle it, but this the current version needs to be improved.

pandas/tests/extension/base/reshaping.py
TomAugspurger: No, was just debugging & forgot to uncomment.

pandas/tests/extension/decimal/test_decimal.py
TomAugspurger: Just added these for easier debugging. All of the are tested elsewhere.

I did notice a bug in the generic `test_take_non_na_fill_value`. I forgot to pass `allow_fill=True`, so we were doing NumPy style instead of pandas style. The tests happened to pass since `data_missing.take([-1, 1], fill_value)` happened to be the correct result.

pandas/core/arrays/base.py
TomAugspurger: I see. I agree with that the other order (`allow_fill`, `fill_value`) is more natural.

I did this order because it matches, but since `take_nd`, aren't in the public API, it doesn't matter so much.

TomAugspurger:
FYI, there's an occasional failure in `test_merge` for JSONArray. This happens when all of the arrays are the same length. 69e7fe76b540807bb973496ee5a2e3e636b90287 had some notable changes:

1. Reversed order of take keywords
2. Added to extensions API
3. Removed default implementation

jreback (commented):
haven’t reviewed fully but looks like a nice change

TomAugspurger:
Thanks. I think the biggest question is how to the change to `get_empty_dtype_and_na` at https://github.com/pandas-dev/pandas/pull/20814/files#diff-e705e723b2d6e7c0e2a0443f80916abfR5454 can fit in better with the rest of what's going on.

TomAugspurger:
Oh, another question: do we or NumPy have any code that'll do bounds checking for us? Or should I just do `if (indexer > len(arr)).any(): raise IndexError`

Dr-Irv:
@TomAugspurger So my (now very small) changes that remove `_can_hold_na` as something people would set in `ExtensionArray` are all green in #20819 , so if they pass muster, you might want to merge them in and use it here, since there is a bit of a relation to your PR for `take`

jreback:
@TomAugspurger see ``maybe_convert_indices`` (in generic.py/_take)

TomAugspurger:
I wasn't really able to share code with `maybe_convert_indices`. It has a couple differences

1. modifies `indicies` in place, we just want bounds checking
2. allows for indices < -1, we want to raise for anything other than that
3. handles `take([], [-1, -1])` differently (we want to allow it)

pandas/tests/extension/category/test_categorical.py
```diff
@@ -81,18 +81,32 @@ def test_merge(self, data, na_value):
 
 
 class TestGetitem(base.BaseGetitemTests):
+    skip_take = pytest.mark.skip(reason="GH-20664.")
```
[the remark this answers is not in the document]
TomAugspurger: Probably doing that in a followup PR...

TomAugspurger:
I think the failure is a random one due to https://github.com/pandas-dev/pandas/pull/20827. Will re-run once that's in.

pandas/core/arrays/base.py
```diff
@@ -487,42 +494,40 @@ def take(self, indexer, allow_fill=True, fill_value=None):
...
+        --------
+        numpy.take
+        pandas.api.extensions.take
 
         Examples
         --------
-        Suppose the extension array is backed by a NumPy array stored as
-        ``self.data``. Then ``take`` may be written as
+        Here's an example implementation, which relies on casting the
+        extension array to object dtype. This uses the helper method
```
jorisvandenbossche: I personally found the previous example that uses something like `self.data` more illustrating how it could look like (except the part on checking for empty, and that the masking would not be done manually anymore). But of course, the example that you did now has the advantage of being an actual working implementation (only one that you won't use in many cases).

pandas/core/dtypes/base.py
```diff
@@ -16,6 +16,10 @@ class _DtypeOpsMixin(object):
     # classes will inherit from this Mixin. Once everything is compatible, this
     # class's methods can be moved to ExtensionDtype and removed.
 
+    # na_value is the default NA value to use for this type. This is used in
+    # e.g. ExtensionArray.take.
```
jorisvandenbossche: I think the "used in EA.take" is misleading. In many cases, it will not be used in take, as this `na_value` is your "boxed" scalar NA type, not the underlying physical NA value.

pandas/core/indexing.py
```diff
@@ -2417,12 +2417,53 @@ def maybe_convert_indices(indices, n):
...
     if mask.any():
         indices[mask] += n
+
     mask = (indices >= n) | (indices < 0)
     if mask.any():
         raise IndexError("indices are out-of-bounds")
     return indices
 
 
+def validate_indices(indices, n):
```
jorisvandenbossche: There is similar functionality in this file (`_iLocIndexer._is_valid_integer/_is_valid_list_like`), but since that is for checking without the -1 specific behaviour, I don't think it is possible / easy to share the implementation (it would be two separate cases inside the function anyhow)

pandas/core/internals.py
```diff
@@ -5399,6 +5405,9 @@ def concatenate_block_managers(mgrs_indexers, axes, concat_axis, copy):
 
     for placement, join_units in concat_plan:
 
+        # The issue: we have a join unit (or maybe several) that needs to be
+        # reindexed.
```
jorisvandenbossche: What is the context of this comment?

pandas/core/internals.py
```diff
@@ -5440,6 +5449,14 @@ def is_uniform_join_units(join_units):
         len(join_units) > 1)
 
 
+def is_uniform_reindex(join_units):
+    return (
```
jorisvandenbossche: Can you add a docstring to indicate what you mean with "uniform reindex"

I understand it as "this are uniform blocks that will keep being uniform after reindexing", correct? 
I am only wondering, does that not depend on the `fill_value` used? Or is this code path only taken in case of default filling (I don't think you can specify the filling with things like concat and merge, as opposed to reindex itself)

pandas/core/internals.py
```diff
@@ -5440,6 +5449,14 @@ def is_uniform_join_units(join_units):
...
+        # TODO: should this be ju.block._can_hold_na?
+        all(ju.block and ju.block.is_extension for ju in join_units) and
+        len(set(ju.block.dtype.name for ju in join_units)) == 1
```
jorisvandenbossche: Did you check this for the Categorical case? (I think I asked about it before) It's fine to disallow the categorical case in this check for now (we still need to fix Categorical.take in general in a follow-up PR anyhow), but I think categoricals now pass this check? 
(and now with the current code, they would return object dtype I think)

pandas/tests/extension/base/getitem.py
```diff
@@ -134,12 +134,42 @@ def test_take(self, data, na_value, na_cmp):
```
jorisvandenbossche: The test above (can't comment on the exact line), should actually fail now, as it is still testing the old behaviour (-1 giving NaN). So need to edit the test to check for `data[len(data)-1]` instead of `nvalue`

pandas/tests/extension/base/getitem.py
```diff
@@ -134,12 +134,42 @@ def test_take(self, data, na_value, na_cmp):
...
+        with pytest.raises(ValueError):
+            data.take([0, -2], fill_value=na_value, allow_fill=True)
+
+    @pytest.mark.parametrize('allow_fill', [True, False])
+    def test_take_out_of_bounds_raises(self, data, allow_fill):
+        arr = data[:3]
+        with pytest.raises(IndexError):
+            arr.take(np.asarray([0, 3]), allow_fill=allow_fill)
+
     @pytest.mark.xfail(reason="Series.take with extension array buggy for -1")
```
jorisvandenbossche: We should be able to remove this xfail now

pandas/tests/extension/json/array.py
```diff
@@ -91,17 +96,41 @@ def nbytes(self):
...
+            except IndexError:
+                raise IndexError(msg)
+
         return self._from_sequence(output)
 
+    # def astype(self, dtype, copy=True):
+    #     # NumPy has issues when all the dicts are the same length.
+    #     # np.array([UserDict(...), UserDict(...)]) fails,
+    #     # but np.array([{...}, {...}]) works, so cast.
+    #     return np.array([dict(x) for x in self], dtype=dtype, copy=copy)
```
jorisvandenbossche: clean-up?

pandas/tests/extension/json/array.py
```diff
@@ -14,6 +14,11 @@
 class JSONDtype(ExtensionDtype):
     type = collections.Mapping
     name = 'json'
+    try:
+        na_value = collections.UserDict()
```
jorisvandenbossche: What is actually the reason you use `UserDict` here?

jorisvandenbossche (commented):
Thanks a lot for this! Did a detailed review and added a bunch of comments, will now try it out with geopandas.

Additional comment:

- In the `BaseGetitemTests.test_reindex` test, can you add there a test case with a different `fill_value` (not None or nan), to make sure it is passed through correctly to EA.take

TomAugspurger:
> can you add there a test case with a different fill_value (not None or nan), to make sure it is passed through correctly to EA.take

I don't think we can do that generically, since we don't know what a valid na_value is.

I'll add one in decimal though.

jorisvandenbossche:
> I don't think we can do that generically, since we don't know what a valid na_value is.

If you fill it with an existing scalar (like `pd.Series(data[:2]).reindex([1, 2], fill_value=data[2])`), then we can expect that to work I think.

But it is actually a good question what should happen when you pass a value that is not accepted by the ExtensionArray. I suppose it is up to the EA to decide on that? (raise an error, or coerce to object)

jorisvandenbossche:
One additional point of feedback from geopandas: 

From running the extension tests in geopandas, it seems `.take(..)` still receives `fill_value=np.nan` quite a few times, so the

```
if allow_fill and fill_value is None:
    fill_value = self.dtype.na_value (in my case: fill_value = 0 
```

snippet was not sufficient and I needed to add `or pd.isna(fill_value)` to make the tests pass. 

But I don't understand why it is not failing for DecimalArray, as you added an assert to the class constructor that all values passed are actually decimals (so also Decimal('NaN') instead of np.nan), which should catch this problem.

TomAugspurger:
Huh... I'll see what I can cook up. Maybe a dtype thing, since DecimalArray is fundamentally object dtype, but geopandas is integer?

For cyberpandas, I can't quite use this `take` with my structured array. Fixing that will be a decent bit of work though.

This does raise the question, should `fill_value` be the high-level NA value (`ipaddres.IPv4Address(0)`) or the low-level memory storage one `(0, 0)`. I think it'll have to be the low-level one.

jorisvandenbossche:
OK, found the reason: I forgot to set the `na_value` on the dtype. 
But now I did that, I have another failure in `pd.merge`, which casts my geometries to object. It might be this is because our `na_value` is `None`, and somewhere this is then seen as object. But will try to figure that out, you don't need to fix that here in this PR :-)

jorisvandenbossche:
> This does raise the question, should fill_value be the high-level NA value (ipaddres.IPv4Address(0)) or the low-level memory storage one (0, 0). I think it'll have to be the low-level one.

So that's the question I raised before (and therefore also questioning whether it should be public), but I have now come to a somewhat conclusion that I think this should be the user-facing scalar ipaddres.IPv4Address(0) and not the low-level (0, 0) (so in case of geopandas: None and not 0). Although it makes it not very useful for actual implementations, it's then more kind of an "informative attribute". 
That's also the reason I commented that the "na_value is used in EA.take" explanation in the `na_value` docstring comment is a bit misleading. As many actual EAs will not use this in the `take` implementation, they will rather pass the low-level fill_value to `take` (I suppose your case is mainly complicated because `take_1d` does not handle struct arrays?)

TomAugspurger:
> I think this should be the user-facing scalar and not the low-level

That's my conclusion to. I will document this more fully in ExtensionArray.take, as `algos.take` doesn't really need to worry about that.

> I suppose your case is mainly complicated because take_1d does not handle struct arrays?

Yeah. In my case, all of the fields are the same dtype, so I'm able reshape & view as a unit64 ndarray, do the take, and the convert back (without copies). I'll make sure everything works, but all that reshaping is a bit complicated :)

pandas/core/internals.py
TomAugspurger: > this are uniform blocks that will keep being uniform after reindexing

Exactly. That's the intent anyway...

It probably does depend on the `fill_value`. I need to better understand `get_empty_dtype_and_na`

TomAugspurger:
> > I don't think we can do that generically, since we don't know what a valid na_value is.

> If you fill it with an existing scalar (like pd.Series(data[:2]).reindex([1, 2], fill_value=data[2])), then we can expect that to work I think.

https://github.com/pandas-dev/pandas/pull/20814/files#diff-0267ba650641bdaf3930400c2478b084R203

pandas/core/algorithms.py
```diff
@@ -1504,7 +1506,7 @@ def take(arr, indexer, allow_fill=False, fill_value=None):
     >>> from pandas.api.extensions import take
 
     With the default ``allow_fill=False``, negative numbers indicate
-    slices from the right.
+    positional indicies from the right.
```
jorisvandenbossche: indicies -> indices

jorisvandenbossche (approved):

pandas/core/algorithms.py
```diff
@@ -1448,6 +1448,96 @@ def func(arr, indexer, out, fill_value=np.nan):
...
+    ...      fill_value=-10)
+    array([ 10,  10, -10])
+    """
+    from pandas.core.indexing import validate_indices
+
+    if not is_array_like(arr):
+        arr = np.asarray(arr)
+
+    # Do we require int64 or intp here?
+    indices = np.asarray(indices, dtype='int')
```
jreback: should be intp, which is what take accepts for indexers

pandas/core/arrays/base.py
```diff
@@ -462,22 +462,35 @@ def factorize(self, na_sentinel=-1):
...
-            indices to be taken. -1 is used to indicate values
-            that are missing.
-        allow_fill : bool, default True
-            If False, indexer is assumed to contain no -1 values so no filling
-            will be done. This short-circuits computation of a mask. Result is
-            undefined if allow_fill == False and -1 is present in indexer.
-        fill_value : any, default None
-            Fill value to replace -1 values with. If applicable, this should
-            use the sentinel missing value for this type.
+        indices : sequence of integers
```
jreback: can you share this doc?

pandas/core/indexing.py
jreback: why is this needed, as opposed to maybe_convert_indices? (maybe refactor that a bit?)

jreback (changes requested):

pandas/core/arrays/base.py
jorisvandenbossche: we want people to look at the source code of this (as we explicitly say the code of this base class is kind of the documentation), so I would certainly try to keep this one here (not sure if it would be easy to reuse it for the `take` function)

pandas/core/indexing.py
jorisvandenbossche: `maybe_convert_indices` *converts* indices, this one only checks bounds. For this use-case, it is not needed to convert negative into positive indices, as this is handles well by numpy (it's only when you want to use our own `take_1d` that you need to do such conversion)

jreback: then fix maybe_convert_indixes which duplicates this logic

jorisvandenbossche: As I said, `maybe_convert_indices` *converts* the indices first and then checks bounds 0-1, this function only checks bounds but different bounds (-1 to n), so there is no real duplicate code. Combining both will only make the single function more complex to understand. Both serve a different purpose, and are used in different places in the code base. There is IMO really no reason to combine them

TomAugspurger: FWIW, I initially started by changing `maybe_convert_indices` to handle both. But it required two new keywords to toggle checking for invalid negative indexers and whether to allow an all `-1` take from an empty index. The only thing it ended up sharing was the integer too large check, which wasn't worthwhile.

pandas/core/arrays/base.py
TomAugspurger: Any thoughts on the new example? I added a comment about how to proceed if your array is backed by an ndarray.

pandas/core/arrays/base.py
TomAugspurger: Sharing docs between these two are a bit difficult because of circular import issues. The docstring would have to go in a third module, which isn't desirable.

TomAugspurger:
Apparently the random failure in test_merge was *not* fixed, at least not for all numpy arrays. https://circleci.com/gh/pandas-dev/pandas/13593#tests/containers/2 had a failures.

Hopefully eb43fa4, which is ugly, fixes it for good. I ran test_merge 1000 times both before and after that commit and things pass, so who knows :/

pandas/core/arrays/base.py
jorisvandenbossche: Maybe add a similar comment to the `fill_value = self.dtype.na_value` line ? (that there you can pass the low-level value that is used in the backing non-object ndarray instead `self.dtype.na_value`

jorisvandenbossche: Ah, I see you added a note about that in the normal comments below the docstring (but maybe still add a short note like "pass here the appropriate NA value for the physical storage, see below")

pandas/core/dtypes/base.py
```diff
@@ -16,6 +16,11 @@ class _DtypeOpsMixin(object):
     # classes will inherit from this Mixin. Once everything is compatible, this
     # class's methods can be moved to ExtensionDtype and removed.
 
+    # na_value is the default NA value to use for this type. This is used in
+    # e.g. ExtensionArray.take. This should be the user-facing "boxed" version
+    # of the NA value, not the physical NA vaalue for storage.
+    na_value = np.nan
```
jreback: maybe even show an example (from JSON?)

pandas/core/indexing.py
jreback: so now we have 2 different code paths which do very similar type of validation, I would try to simplify ``Series._take`` and ``NDFrame._take`` to use these new helper functions, removing the need for ``maybe_convert_indices`` entirely.

jreback (changes requested):

pandas/core/indexing.py
TomAugspurger: They have different behaviors.

1. `maybe_convert_indices` allows for any negative numbers, while `validate_indices` allows for just `-1`
2. `validate_indices` allows for an array of `-1` taken from an empty array, while `maybe_convert_indices` doesn't.

You'd need two keywords to control those behaviors, and the only then that could be shared is a `max_idx > n`. Better to have two simpler functions here.

jorisvandenbossche:
@jreback do you have any other comments?

jreback:
I guess this is fine. I am sure this will spawn many more bugs and would like this to live in master for a while.

jreback (approved):

jorisvandenbossche:
> I am sure this will spawn many more bugs

To what are you referring?

pandas/core/internals.py
```diff
@@ -5457,6 +5471,12 @@ def get_empty_dtype_and_na(join_units):
         if blk is None:
             return np.float64, np.nan
 
+    if is_uniform_reindex(join_units):
+        # XXX: integrate property
```
jbrockmendel, line 5475: @TomAugspurger long shot any recollection what this comment is asking for?

TomAugspurger: Removing the need for that `is_uniform_reindex` check by including that case in the main `if / elif` block (which may be done on master now?)

pandas/core/algorithms.py
jbrockmendel: this ended up exposed in pd.api.extensions but it isn't clear to me why.  are we sure its needed there?

PR #20815, opened 2018-04-24
BUG: Make ExtensionArray:_can_hold_na a class attribute

### LINKED ISSUE #20761: ExtensionArray._can_hold_na needs to be a class attribute

Dr-Irv:
- [x] closes #20761 
- [x] tests added / passed
  - pandas/tests/extension/decimal/test_nona.py
    - This tests an example where `_can_hold_na==False`
- [ ] whatsnew entry
  - Not included as it is a fix to the new `ExtensionArray` interface

Supersedes #20801 by removing ops functionality from there.


pandas/tests/extension/base/getitem.py
```diff
@@ -134,8 +134,9 @@ def test_take(self, data, na_value, na_cmp):
 
     def test_take_empty(self, data, na_value, na_cmp):
         empty = data[:0]
-        result = empty.take([-1])
-        na_cmp(result[0], na_value)
+        if data._can_hold_na:
+            result = empty.take([-1])
+            na_cmp(result[0], na_value)
```
jorisvandenbossche: Question is what this should do if `_can_hold_na=False` ? And in general with -1 indexers (eg `reindex`, `concat` with realignment, ...) 
I suppose to raise an error? Or coerce to object?

pandas/tests/extension/decimal/test_nona.py
```diff
@@ -0,0 +1,135 @@
...
+    expected = pd.DataFrame({"A": arr})
+    tm.assert_frame_equal(result, expected)
+
+
+def test_dataframe_constructor_with_different_dtype_raises():
+    arr = DecimalNoNaArray([decimal.Decimal('10.0')])
+
+    xpr = "Cannot coerce extension array to dtype 'int64'. "
+    with tm.assert_raises_regex(ValueError, xpr):
+        pd.DataFrame({"A": arr}, dtype='int64')
```
jorisvandenbossche: Those constructor tests are not specific to the "nona", so I would move them to a separate file

pandas/tests/extension/decimal/test_nona.py
```diff
@@ -0,0 +1,135 @@
...
+    pass
+
+
+class TestNoNaGroupby(TestGroupby):
+    pass
+
+
+def test_series_constructor_with_same_dtype_ok():
+    arr = DecimalNoNaArray([decimal.Decimal('10.0')])
+    result = pd.Series(arr, dtype=DecimalDtype())
```
jorisvandenbossche: This is maybe something we need to discuss, as in the `take` PR Tom was adding an `_na_value` to the dtype itself. We should see how this interacts

pandas/tests/extension/base/getitem.py
jorisvandenbossche: Yes, maybe, but then I am just wondering what this design decision is (and we should also test that here). 

As far as I know, this is rather new terrain (in pandas we only have ints, and they are upcasted to float), so it might be we still need to discuss/decide on this.

jorisvandenbossche: We also have a BoolBlock that cannot hold NaNs, so a boolean Series gets converted to object Series when NaNs are introduced:

```
In [49]: pd.Series([True, False])
Out[49]: 
0     True
1    False
dtype: bool

In [50]: pd.Series([True, False]).reindex([1, 2])
Out[50]: 
1    False
2      NaN
dtype: object
```

So this is some prior art

Dr-Irv: So by the statement "we still need to discuss/decide on this", I presume that the core dev team will make that call.  Right? (FWIW, I think I may use `_can_hold_na=True`, and since I'm storing objects, I'll just use NaN to be the na_value that is then used in `reindex`, `shift`, etc.), and I'd want the type of the resulting array to remain in my subclass of `ExtensionArray`

jorisvandenbossche: > So by the statement "we still need to discuss/decide on this", I presume that the core dev team will make that call. Right?

No, this is an open discussion, and your voice in this discussion is important since you are actually one of the first people trying to use this (both Tom and I do not use `_can_hold_na=False` in our applications).

But if you would also not use this, we should maybe consider removing the feature for now (meaning this will always be True) until somebody comes up with a good use case.

Dr-Irv: Well, I've come full circle on this.  In my case, I thought that I would not want NA in my arrays. My objects stored in the arrays have no NA value that makes sense. And then when you brought up `reindex` as an example, I realized that I would use things like `shift` and would want the NA to be filled in, but keeping the `ExtensionArray` class as the holder, meaning that I need `_can_hold_na` as True.  I'd want `ExtensionArray` as holding it so that arithmetic would work, as coercing to an `object` array would mess up the arithmetic. So I'd have no problem if we dropped `_can_hold_na` from the `ExtensionArray` design, and tell people that they must define a _na_value(), which is then used as the "fill in" when we do things like `reindex` and `shift`.  And because there are pandas operations that need some NA value for fill in purposes, requiring it on `ExtensionArray` shouldn't be an issue (IMHO).

If we agree that getting rid of `_can_hold_na` is the thing to do, I'll modify this PR to totally get rid of it (and the associated tests).

jreback:
i would not add API unless we have a need for it
so yes would remove until / until is necessary

Dr-Irv:
Closing due to new PR #20819

Issue #20648, opened 2018-04-10
DOC: rpy2 examples started failing


jorisvandenbossche:
The rpy2 examples in the documentation started failing recently (http://pandas-docs.github.io/pandas-docs-travis/r_interface.html) with:

```
In [1]: from rpy2.robjects import r, pandas2ri
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-79b90b86f23f> in <module>()
----> 1 from rpy2.robjects import r, pandas2ri

~/miniconda3/envs/pandas/lib/python3.6/site-packages/rpy2/robjects/__init__.py in <module>()
     14 import itertools
     15 from datetime import datetime
---> 16 import rpy2.rinterface as rinterface
     17 import rpy2.rlike.container as rlc
     18 

~/miniconda3/envs/pandas/lib/python3.6/site-packages/rpy2/rinterface/__init__.py in <module>()
     48 del(os)
     49 
---> 50 from rpy2.rinterface._rinterface import (baseenv,
     51                                          emptyenv,
     52                                          endr,

ImportError: dlopen: cannot load any more object with static TLS
```

However, I don't see any change in version between a recent working build and the current failing ([this](https://travis-ci.org/pandas-dev/pandas/jobs/358061154) working one uses rpy2 2.9.1-py36mro343h14c3975_0, and [this](https://travis-ci.org/pandas-dev/pandas/jobs/364254345) failing one 2.9.1-py36mro343h14c3975_0, so the same)

Could this be caused by installing another library?

cc @lgautier

WillAyd:
Hmm this has been inconsistent right? Googling the error I saw a few other issues for other projects but none seemed to have a definite resolution. Here's something similar from Tensorflow:

https://github.com/tensorflow/models/issues/523

A lot of the answers seemed to be doing things that "fixed" the issue without any explanation (ex: changing import order) or just restarting the system...

datapythonista:
Looks like this is failing even more than when the issue was created. We don't even have the required modules in the dependencies we use to build the docs:

https://pandas.pydata.org/pandas-docs/stable/r_interface.html

As the functionality being documented is not on pandas but on `rpy2`, and nobody complained about the page giving errors everywhere, what I'd do is simply remove the page from the docs. One of the reasons I think it's better to delete the page instead of fixing the problem, is that fixing the problem implies installing more than 50 additional dependencies (required by rpy2) in the builds and in our local development environments.

[...]

PR #20830, opened 2018-04-26
[WIP] Debug R doc build failure

TomAugspurger:
#20648

### DIFF
--- a/ci/travis-36-doc.yaml
+++ b/ci/travis-36-doc.yaml
@@ -1,14 +1,14 @@
 name: pandas
 channels:
   - defaults
-  - conda-forge
+  # - conda-forge
   - r
 dependencies:
   - beautifulsoup4
   - bottleneck
   - cython
   - fastparquet
-  - feather-format
+  # - feather-format
   - html5lib
   - ipykernel
   - ipython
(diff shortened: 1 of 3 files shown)


TomAugspurger:
The packages built by anaconda include a compiler flag (`-fPIC`) that should make this error impossible. Seeing if (temporarily) excluding conda-forge packages fixes things.

TomAugspurger:
Hmm, still trying to get  a local reproducer, but https://travis-ci.org/pandas-dev/pandas/jobs/371582150#L2705 managed to import rpy2, I guess since it was imported early enough :/

https://stackoverflow.com/questions/19268293/matlab-error-cannot-open-with-static-tls/19468365#19468365

It sounds like either Travis' glibc is the bottleneck here. We won't be able to do much to work around it unfortunately.

jorisvandenbossche:
Thanks for looking into this. 
Temporary / definite workaround would be just remove that doc section, or to convert it to static code-blocks.

TomAugspurger:
Yeah. I'm able to build things locally, so I'm not too worried about it delaying the relase.

jreback:

PR #20840, opened 2018-04-27
BUG: concat of Series of EA and other dtype fails

### LINKED ISSUE #20832: concat with ExtensionArray and object upcasting fails
```python
In [1]: import pandas as pd

In [2]: from pandas.tests.extension.decimal.array import *

In [3]: result = pd.concat([pd.Series(DecimalArray([decimal.Decimal(1)])), pd.Series([1])])

In [4]: result.shape
Out[4]: (2, 1)
```

and the repr fails, since things are messed up.

jorisvandenbossche:
Closes #20832

### DIFF
--- a/pandas/core/dtypes/concat.py
+++ b/pandas/core/dtypes/concat.py
@@ -176,7 +176,7 @@ def _concat_compat(to_concat, axis=0):
 
     extensions = [is_extension_array_dtype(x) for x in to_concat]
     if any(extensions) and axis == 1:
-            to_concat = [np.atleast_2d(x.astype('object')) for x in to_concat]
+        to_concat = [np.atleast_2d(x.astype('object')) for x in to_concat]
 
     if not nonempty:
         # we have all empties, but may need to coerce the result dtype to
--- a/pandas/core/dtypes/concat.py
+++ b/pandas/core/dtypes/concat.py
@@ -175,8 +175,8 @@ def _concat_compat(to_concat, axis=0):
         return _concat_sparse(to_concat, axis=axis, typs=typs)
 
     extensions = [is_extension_array_dtype(x) for x in to_concat]
-    if any(extensions):
-        to_concat = [np.atleast_2d(x.astype('object')) for x in to_concat]
+    if any(extensions) and axis == 1:
+            to_concat = [np.atleast_2d(x.astype('object')) for x in to_concat]
 
     if not nonempty:
         # we have all empties, but may need to coerce the result dtype to
--- a/pandas/tests/extension/base/reshaping.py
+++ b/pandas/tests/extension/base/reshaping.py
@@ -64,6 +64,11 @@ class BaseReshapingTests(BaseExtensionTests):
         expected = pd.concat([df1.astype('object'), df2.astype('object')])
         self.assert_frame_equal(result, expected)
 
+        result = pd.concat([df1['A'], df2['A']])
+        expected = pd.concat([df1['A'].astype('object'),
+                              df2['A'].astype('object')])
+        self.assert_series_equal(result, expected)
+
     def test_align(self, data, na_value):
         a = data[:3]
         b = data[2:5]
--- a/pandas/tests/extension/base/reshaping.py
+++ b/pandas/tests/extension/base/reshaping.py
@@ -69,6 +69,26 @@ class BaseReshapingTests(BaseExtensionTests):
                               df2['A'].astype('object')])
         self.assert_series_equal(result, expected)
 
+    def test_concat_columns(self, data, na_value):
+        df1 = pd.DataFrame({'A': data[:3]})
+        df2 = pd.DataFrame({'B': [1, 2, 3]})
+
+        expected = pd.DataFrame({'A': data[:3], 'B': [1, 2, 3]})
+        result = pd.concat([df1, df2], axis=1)
+        self.assert_frame_equal(result, expected)
+        result = pd.concat([df1['A'], df2['B']], axis=1)
+        self.assert_frame_equal(result, expected)
+
+        # non-aligned
+        df2 = pd.DataFrame({'B': [1, 2, 3]}, index=[1, 2, 3])
+        expected = pd.DataFrame({
+            'A': data._from_sequence(list(data[:3]) + [na_value]),
+            'B': [np.nan, 1, 2, 3]})
+        result = pd.concat([df1, df2], axis=1)
+        self.assert_frame_equal(result, expected)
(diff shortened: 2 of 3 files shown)


pandas/core/dtypes/concat.py
```diff
@@ -175,8 +175,8 @@ def is_nonempty(x):
         return _concat_sparse(to_concat, axis=axis, typs=typs)
 
     extensions = [is_extension_array_dtype(x) for x in to_concat]
-    if any(extensions):
-        to_concat = [np.atleast_2d(x.astype('object')) for x in to_concat]
+    if any(extensions) and axis == 1:
```
jreback: so what happens on axis=0? coerced to``object``?

pandas/tests/extension/base/reshaping.py
```diff
@@ -64,6 +64,11 @@ def test_concat_mixed_dtypes(self, data):
         expected = pd.concat([df1.astype('object'), df2.astype('object')])
         self.assert_frame_equal(result, expected)
 
+        result = pd.concat([df1['A'], df2['A']])
```
jreback: is this just for axis=1? I would make this a separate test.

jreback (changes requested):

pandas/core/dtypes/concat.py
TomAugspurger: Right (everything is upcast).

pandas/core/dtypes/concat.py
```diff
@@ -175,8 +175,8 @@ def is_nonempty(x):
...
+            to_concat = [np.atleast_2d(x.astype('object')) for x in to_concat]
```
TomAugspurger: Any reason you added this extra indent?

pandas/tests/extension/base/reshaping.py
TomAugspurger: The bug is just with `axis=0`, the default, which this is testing.

For `axis=1` there's not any up-casting required.

Though I don't see any tests for `concat(..., axis='columns')` with the extension array. Can you add some Joris? concat EA and EA, mix of EA and non-EA, aligned and no aligned. Or should I?

jorisvandenbossche: Hmm, thought we already had some, but indeed, apparently not. Will add some.

TomAugspurger:
I may have missed some, only glanced briefly in extension/base/reshape.py

pandas/core/dtypes/concat.py
```diff
@@ -175,7 +175,7 @@ def is_nonempty(x):
         return _concat_sparse(to_concat, axis=axis, typs=typs)
 
     extensions = [is_extension_array_dtype(x) for x in to_concat]
-    if any(extensions):
+    if any(extensions) and axis == 1:
         to_concat = [np.atleast_2d(x.astype('object')) for x in to_concat]
```
jreback: I find this very strange that you need to add axis=1 here, I don't think this routine should be call at all if axis=1. This seems like it should be handled at a higher level

jreback (changes requested):

pandas/core/dtypes/concat.py
TomAugspurger: >  I find this very strange that you need to add axis=1 here, I don't think this routine should be call at all if axis=1

`axis=1` is used all over `_concat_compat`.  Why do you think it's strange?

For the specific case from #20832 it is `axis=0`, but we don't want to reshape to 2d.

TomAugspurger (approved):

pandas/core/dtypes/concat.py
jorisvandenbossche: Yes, the special casing non-consolidateble blocks with `axis=1` is indeed an existing pattern. 

For categoricals:

https://github.com/pandas-dev/pandas/blob/563a6ad108a79b0192a25d51d743812a18e21b15/pandas/core/dtypes/concat.py#L220-L221

for datetimetz:

https://github.com/pandas-dev/pandas/blob/563a6ad108a79b0192a25d51d743812a18e21b15/pandas/core/dtypes/concat.py#L453-L454

Note that this is all rather confusing, as `axis=1` in the "concatting code" actually means `axis=0` in user facing code (because we store the 1D data in 2D objects):

https://github.com/pandas-dev/pandas/blob/563a6ad108a79b0192a25d51d743812a18e21b15/pandas/core/reshape/concat.py#L312-L315

jreback: right this should be changed to use ``_get_block_manager_axis`` FYI (but can be in future).

jreback (approved):

pandas/core/dtypes/concat.py
jorisvandenbossche: > right this should be changed to use _get_block_manager_axis FYI (but can be in future).

I don't think so, `_get_block_manager_axis` is for a full Series/DataFrame, the logic here is dtype-dependent

jreback: what does your respsone have to do with my comment? this is exactly what _get_block_managed_axis does

jorisvandenbossche: OK, my response was not very well phrased, but still, then you will have to explain your comment better :-) 

I understood your comment as: "we should use `_get_block_manager_axis` here ", and AFAIK, `_get_block_manager_axis` lets you convert an axis of 0/1 to the appropriate number for blocks (so for DataFrame it switches 0 and 1). But here, I don't convert any `axis` argument, I reshape data depending on the `axis` value (and depending on the dtype). 
So I don't fully understand where in the above code lines I would use ``_get_block_manager_axis``

jreback:
```
diff --git a/pandas/core/reshape/concat.py b/pandas/core/reshape/concat.py
index 6e564975f..6570e7a83 100644
--- a/pandas/core/reshape/concat.py
+++ b/pandas/core/reshape/concat.py
@@ -312,7 +312,7 @@ class _Concatenator(object):
         # Need to flip BlockManager axis in the DataFrame special case
         self._is_frame = isinstance(sample, DataFrame)
         if self._is_frame:
-            axis = 1 if axis == 0 else 0
+            axis = sample._get_block_manager_axis(axis)
 
         self._is_series = isinstance(sample, Series)
         if not 0 <= axis <= sample.ndim:
```

is a more correct usage, though prob doesn't simplify code much. this is spagetti anyhow.

jorisvandenbossche: > this is spagetti anyhow.

On that we certainly agree! :-)

PR #20844, opened 2018-04-27
ENH: Add class to write dta format 117 files

### LINKED ISSUE #16450: to_stata: Fixed width strings in Stata .dta files are limited to 244 (or fewer)

```python
import pandas
frame = pandas.DataFrame({'A':['h'*250,'hi','hola']})
frame.to_excel("text.xlsx", index=False)
frame.to_stata("test.dta")

```
#### Problem description

Raises the following error: 
> ValueError:
> Fixed width strings in Stata .dta files are limited to 244 (or fewer)
> characters.  Column 'A' does not satisfy this restriction.

However this restriction seems to not exists in STATA, as the Excel file can be imported correctly
Open STATA, import the Excel file

```STATA
import excel "C:\data\tesi\software\text.xlsx", sheet("Sheet1") firstrow clear

```
Now we can get the type of data in column 'A', and as you can see, it's str250. So STATA can store string longer than 244 characters

```STATA
. describe A

```
A               str250  %250s                 A

#### Expected Output
File gets exported with the correct format and without problems

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.2.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
processor: Intel64 Family 6 Model 94 Stepping 3, GenuineIntel
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.19.2
nose: 1.3.7
pip: 9.0.1
setuptools: 23.0.0
Cython: 0.24
numpy: 1.11.2
scipy: 0.18.1
statsmodels: None
xarray: None
IPython: 4.2.0
sphinx: 1.5.5
patsy: None
dateutil: 2.5.3
pytz: 2016.6.1
blosc: None
bottleneck: None
tables: 3.2.3.1
numexpr: 2.6.1
matplotlib: 1.5.3
openpyxl: 2.4.7
xlrd: 1.0.0
xlwt: None
xlsxwriter: None
lxml: 3.7.3
bs4: 4.5.3
html5lib: 0.999999999
httplib2: None
apiclient: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.8
boto: None
pandas_datareader: 0.2.1
</details>

bashtage:
Add export for dta 117 files which add support for long strings
Refactor StataWriter to simplify new writer

closes #16450

- [x] closes #16450
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -1769,7 +1769,8 @@ class DataFrame(NDFrame):
 
     def to_stata(self, fname, convert_dates=None, write_index=True,
                  encoding="latin-1", byteorder=None, time_stamp=None,
-                 data_label=None, variable_labels=None):
+                 data_label=None, variable_labels=None, version=114,
+                 convert_strl=None):
         """
         A class for writing Stata binary dta files from array-like objects
 
@@ -1800,6 +1801,23 @@ class DataFrame(NDFrame):
             values. Each label must be 80 characters or smaller.
 
             .. versionadded:: 0.19.0
+
+        version : {114, 117}
+            dta version to use in the output file.  Version 114 can be used
+            read by Stata 10 and later.  Version 117 can be read by Stata 13
+            or later. Version 114 limits string variables to 244 characters or
+            fewer while 117 allows strings with lengths up to 2,000,000
+            characters.
+
+            .. versionadded:: 0.23.0
+
+        convert_strl : list, optional
+            List of column names to convert to string columns to Stata StrL
+            format. Only available if version is 117.  Storign strings in the
+            StrL format can produce smaller dta files if strings have more than
+            8 characters and values are repeated.
+
+            .. versionadded:: 0.23.0
 
         Raises
         ------
@@ -1832,12 +1850,23 @@ class DataFrame(NDFrame):
         >>> writer = StataWriter('./date_data_file.dta', data, {2 : 'tw'})
         >>> writer.write_file()
         """
-        from pandas.io.stata import StataWriter
-        writer = StataWriter(fname, self, convert_dates=convert_dates,
+        kwargs = {}
+        if version not in (114, 117):
+            raise ValueError('Only formats 114 and 117 supported.')
+        if version == 114:
+            if convert_strl is not None:
+                raise ValueError('strl support is only available when using '
+                                 'format 117')
+            from pandas.io.stata import StataWriter as statawriter
+        else:
+            from pandas.io.stata import StataWriter117 as statawriter
+            kwargs['convert_strl'] = convert_strl
+
+        writer = statawriter(fname, self, convert_dates=convert_dates,
                              encoding=encoding, byteorder=byteorder,
                              time_stamp=time_stamp, data_label=data_label,
                              write_index=write_index,
-                             variable_labels=variable_labels)
+                             variable_labels=variable_labels, **kwargs)
         writer.write_file()
 
     def to_feather(self, fname):
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -1772,25 +1772,25 @@ class DataFrame(NDFrame):
                  data_label=None, variable_labels=None, version=114,
                  convert_strl=None):
         """
-        A class for writing Stata binary dta files from array-like objects
+        Export Stata binary dta files.
 
         Parameters
         ----------
         fname : str or buffer
-            String path of file-like object
+            String path of file-like object.
         convert_dates : dict
             Dictionary mapping columns containing datetime types to stata
             internal format to use when writing the dates. Options are 'tc',
             'td', 'tm', 'tw', 'th', 'tq', 'ty'. Column can be either an integer
             or a name. Datetime columns that do not have a conversion type
             specified will be converted to 'tc'. Raises NotImplementedError if
-            a datetime column has timezone information
+            a datetime column has timezone information.
         write_index : bool
             Write the index to Stata dataset.
         encoding : str
-            Default is latin-1. Unicode is not supported
+            Default is latin-1. Unicode is not supported.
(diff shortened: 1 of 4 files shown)


bashtage:
Is it intentional that `stata.py` is not covered?  Seems a bit strange.

pandas/core/frame.py
```diff
@@ -1801,6 +1802,23 @@ def to_stata(self, fname, convert_dates=None, write_index=True,
...
+            read by Stata 10 and later.  Version 117 can be read by Stata 13
+            or later. Version 114 limits string variables to 244 characters or
+            fewer while 117 allows strings with lengths up to 2,000,000
+            characters.
+
+            .. versionadded:: 0.23.0
+
+        convert_strl : list, optional
+            List of column names to convert to string columns to Stata StrL
+            format. Only available if version is 117.  Storign strings in the
```
jreback: Storing

jreback:
> Is it intentional that stata.py is not covered? Seems a bit strange.

what does this mean? the coverage that prints on the issue is odd (and has always been that way), but you can click thru and see the actual reports.

pandas/core/frame.py
```diff
@@ -1801,6 +1802,23 @@ def to_stata(self, fname, convert_dates=None, write_index=True,
 
             .. versionadded:: 0.19.0
 
+        version : {114, 117}
+            dta version to use in the output file.  Version 114 can be used
+            read by Stata 10 and later.  Version 117 can be read by Stata 13
```
jreback: should we just support version 117 and forward? how old are these respective versions?

jreback (changes requested):

pandas/core/frame.py
jreback: how much simplfication do we get from only supporting new format?

pandas/io/stata.py
```diff
@@ -2307,3 +2384,553 @@ def _null_terminate(self, s, as_string=False):
...
+        """Write the file header"""
+        byteorder = self._byteorder
+        self._file.write(_bytes('<stata_dta>', 'utf-8'))
+        bio = BytesIO()
+        # ds_format - 117
+        bio.write(self._tag(_bytes('117', 'utf-8'), 'release'))
+        # byteorder
+        bio.write(self._tag(byteorder == ">" and "MSF" or "LSF", 'byteorder'))
+        # number of vars, 2 bytes
+        assert self.nvar < 2 ** 16
```
jreback: maybe add a blank line before comments, easier to read

pandas/tests/io/test_stata.py
```diff
@@ -5,21 +5,22 @@
 import os
 import struct
 import warnings
-from datetime import datetime
 from collections import OrderedDict
+from datetime import datetime
 
 import numpy as np
+import pytest
+from pandas._libs.tslib import NaT
```
jreback: import from pandas

jreback (changes requested):
minor comments. Might consiser renaming StataWriter to StataWriter114 (and essentially make it private). Otherwise lgtm.

bashtage:
> minor comments. Might consiser renaming StataWriter to StataWriter114 (and essentially make it private). Otherwise lgtm.

I think this would need a dep cycle since `StataWriter` has been public.  Can do this if you want.

pandas/core/frame.py
```diff
@@ -1801,6 +1802,23 @@ def to_stata(self, fname, convert_dates=None, write_index=True,
 
             .. versionadded:: 0.19.0
 
+        version : {114, 117}
```
TomAugspurger: Hmm, should these be strings? Is this exposed anywhere in stats itself? Do they use integers? (when I see version number, I think string).

TomAugspurger: Huh, https://www.stata.com/support/faqs/data-management/save-for-previous-version/ seems to suggest that stata uses integers? `version(13)`. OK then, let's follow that.

pandas/core/frame.py
```diff
@@ -1832,12 +1856,23 @@ def to_stata(self, fname, convert_dates=None, write_index=True,
         >>> writer = StataWriter('./date_data_file.dta', data, {2 : 'tw'})
         >>> writer.write_file()
         """
-        from pandas.io.stata import StataWriter
-        writer = StataWriter(fname, self, convert_dates=convert_dates,
+        kwargs = {}
+        if version not in (114, 117):
+            raise ValueError('Only formats 114 and 117 supported.')
```
TomAugspurger: Would be nice to include the user passed `version` in the error message.

pandas/tests/io/test_stata.py
```diff
@@ -1355,3 +1413,63 @@ def test_date_parsing_ignores_format_details(self, column):
...
+                                  original, check_index_type=False)
+            tm.assert_frame_equal(original, copy)
+
+    def test_convert_strl_name_swap(self):
+        original = DataFrame([['a' * 3000, 'A', 'apple'],
+                              ['b' * 1000, 'B', 'banana']],
+                             columns=['long1' * 10, 'long', 1])
+        original.index.name = 'index'
+
+        with warnings.catch_warnings(record=True) as w:  # noqa
```
TomAugspurger: What warnings are you catching here?

TomAugspurger (approved):
Couple small comments. OK to do as a followup (I'm tagging the RC momentarily.)

TomAugspurger:
@jreback are you OK with merging this as is? I'll fix the git conflict on merge.

jreback:
yep this is fine
if we need follow ups after RC that is fine

TomAugspurger:
K. @bashtage I'm going to fix the conflict and then merge. A followup PR would be welcome.

PR #20851, opened 2018-04-27
 BUG: allow None as upcasted_na value 

jorisvandenbossche:
In geopandas I use `dtype.na_value = None`, but this gives problems in `merge` (in `JoinUnit.get_reindexed_values`, `upcasted_na` of None is handled specifically).

The first commit works, but is not very nice, so in the second commit I attempted to just return `values` instead of `get_values`


TomAugspurger:
values vs. get_values will have an effect for

- SparseBlock: not densified
- ExtensionBlock / NonConsolidatableMixin (no reshaping to `(1, len(values))`

No effect for

- Block (dtype is None)
- DatetimeTZBlock

Does that sound right?

jorisvandenbossche:
From looking at the code, it seemed that `upcasted_na=None` only occurred in case of bool or numeric dtype, so for that block types, there is no difference between `values` and `get_values`. So therefore I tried to just change that.

The main problem with the current version is that this overloads the use of `None` there ("no upcasting -> different code path" vs "extensionarray with None as fill_value -> by "accident" the correct code" ). The first commit (if you look at it separately) is uglier but more "correct"

jreback:
do any tests actually hit this?

jorisvandenbossche:
Reviving this, as I hit this bug again with geopandas. 

For adding a test, I need an EA that uses None as missing value indicator. I could add a second version of the DecimalArray? (although that will duplicate quite a bit) Or I could adapt the JSON version to use None instead of `{}`, add a new ListArray that uses None as missing value indicator, or change the arrow-based one. Any preferences?

TomAugspurger:
> For adding a test, I need an EA that uses None as missing value indicator. I could add a second version of the DecimalArray?

Can you subclass DecimalArray and override the na_value? That may not be too much duplication.

jreback:
ListArray would also be nice

jreback:
@jorisvandenbossche can you merge master

jreback (changes requested):
tests for this?

PR #20628, opened 2018-04-07
DOC: update the pandas.Series.str.strip docstring

techedlaksh:
Checklist for the pandas documentation sprint (ignore this if you are doing
an unrelated PR):

- [x] PR title is "DOC: update the <your-function-or-method> docstring"
- [x] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [x] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [x] It has been proofread on language by another sprint participant

Please include the output of the validation script below between the "```" ticks:

```
################################################################################
##################### Docstring (pandas.Series.str.strip)  #####################
################################################################################

Strip whitespaces from string in Series.

Strip whitespace (including newlines) or given string from
each string in the Series/Index from left and right sides. Equivalent to
:meth:`str.strip`.

Parameters
----------
to_strip : str or unicode
    String or unicode to strip in the given string.

Examples
--------
>>> # strip method
>>> s = pd.Series(['   This is a Test 1  '])
>>> s
0    This is a Test 1
dtype: object
>>> s = s.str.strip()
>>> s
0    This is a Test 1
dtype: object
>>> # lstrip method
>>> s1 = pd.Series(['This is another Test'])
>>> s1
0    This is another Test
dtype: object
>>> s1 = s1.str.lstrip('This')
>>> s1
0     is another Test
dtype: object
>>> # rstrip method
>>> s2 = pd.Series(['This is the last test'])
>>> s2
0    This is the last test
dtype: object
>>> s2 = s2.str.rstrip('test')
>>> s2
0     This is the last
dtype: object

Returns
-------
stripped : Series/Index of objects

################################################################################
################################## Validation ##################################
################################################################################

Errors found:
	See Also section not found


```

PR #20863, opened 2018-04-29
DOC: update the pandas.Series.str.strip docstring

Cheukting:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


pandas/core/strings.py
```diff
@@ -2151,12 +2151,54 @@ def encode(self, encoding, errors="strict"):
...
-    Series/Index from %(side)s. Equivalent to :meth:`str.%(method)s`.
+    Remove leading and trailing characters.
+
+    Strip whitespaces (including newlines) or a set of specified characters
+    from each string in the Series/Index from %(side)s.
+    Equivalent to :meth:`str.%(method)s`.
+
+    Parameters
+    ----------
+    to_strip : str
```
WillAyd: `to_strip : str, optional`

pandas/core/strings.py
```diff
@@ -2151,12 +2151,54 @@ def encode(self, encoding, errors="strict"):
...
+        Specifying the set of characters to be removed.
```
WillAyd: Let's just simplify this and say "The set of characters to be removed". It matches the standard Python string method so no need to be too wordy

pandas/core/strings.py
```diff
@@ -2151,12 +2151,54 @@ def encode(self, encoding, errors="strict"):
...
+    --------
+    str.slice : Slice substrings from each element in the Series/Index
+
+    Examples
+    --------
+    Striping whitespaces for Series
+
+    >>> s = pd.Series([' ant', 'bee ', ' cat  '])
+
+    >>> s
```
WillAyd: This is simple enough that you don't need to print the Series as is

pandas/core/strings.py
```diff
@@ -2151,12 +2151,54 @@ def encode(self, encoding, errors="strict"):
...
     -------
     stripped : Series/Index of objects
+
+    See Also
+    --------
+    str.slice : Slice substrings from each element in the Series/Index
+
+    Examples
+    --------
+    Striping whitespaces for Series
```
WillAyd: "Striping" is a typo here and below, though these examples are simple enough that I don't think you need to explain them

pandas/core/strings.py
```diff
@@ -2151,12 +2151,54 @@ def encode(self, encoding, errors="strict"):
...
+
+    >>> s.str.strip()
+    0    ant
+    1    bee
+    2    cat
+    dtype: object
+
+    Striping a set of characters for Index
+
+    >>> df = pd.DataFrame(index=['1.ant ','2._bee__','3. cat_'])
```
WillAyd: Just create the index directly instead of creating a DataFrame

`index = pd.Index(['1.ant ']) ...`

pandas/core/strings.py
```diff
@@ -2151,12 +2151,54 @@ def encode(self, encoding, errors="strict"):
...
+
+    >>> df.index
```
WillAyd: Again simple enough you don't need to print

pandas/core/strings.py
```diff
@@ -2151,12 +2151,54 @@ def encode(self, encoding, errors="strict"):
...
+    2     cat  
+    dtype: object
+
+    >>> s.str.strip()
+    0    ant
+    1    bee
+    2    cat
+    dtype: object
+
+    Striping a set of characters for Index
```
WillAyd: Typo here, but as mentioned above these are basic enough that I personally don't think you need to explain in detail

WillAyd (changes requested):
Thanks for the PR. Can you post the results of the doc string validator?

Cheukting:
###############################################################################

################################## Validation #################################

###############################################################################


Docstring for "pandas.Series.str.strip" correct. :)

pandas/core/strings.py
```diff
@@ -2151,12 +2151,54 @@ def encode(self, encoding, errors="strict"):
...
+        All combinations of this set of characters will be stripped.
+        Default value is None, which means whaitspaces will be removed.
 
     Returns
     -------
     stripped : Series/Index of objects
+
+    See Also
+    --------
+    str.slice : Slice substrings from each element in the Series/Index
```
datapythonista: `slice` doesn't seem very related to strip.

Not sure if there is anything else related, but we should have here `Series.str.strip`, `Series.str.lstrip` and `Series.str.rstrip`, so they link each other.

datapythonista (commented):
Added some comments.

Also, #20628 was working on this too. I think your changes look better, but you can take a look at that PR and see if there is any idea that you can copy here.

pandas/core/strings.py
```diff
@@ -2159,46 +2159,52 @@ def encode(self, encoding, errors="strict"):
...
-    2    cat
+    0    1. Ant.
+    1    2. Bee!
+    2    3. Cat?
     dtype: object
 
-    Stripping a set of characters for Index
-
-    >>> idx = pd.Index(['1.ant ','2._bee__','3. cat_'])
+    >>> s.str.strip('123.!? \n\t')
```
datapythonista: Personally I find this example overcomplicated. I think something more realistic and simple could be:
- Illustrate the default `to_strip` parameter (whitespaces) with `rstrip` in a value like ` text    \n'
- `lstrip(to_strip='0')` for a value like `0010010`)
- And then you can show `strip(to_strip=' \n0')` with both values

alsobay:
Hi @Cheukting, great work! As @datapythonista pointed out in #20897, it seems that our wires crossed during the sprint and we ended up working on the same thing. 

Seeing as this PR got in earlier, please feel free to grab anything from our PR that might help address Marc's notes, and we'll close ours. If you don't plan to refine this PR, please let us know and we'll continue on #20897.

Cheukting:
Thanks @alsobay ! I really like your PR and I think I may borrow some ideas and put it here. Feel free to keep commenting and reviewing this PR as it is our collaboration effort :-)

pandas/core/strings.py
Cheukting: I  still wants to keep the 'Animal' theme in the example but I will simplify the `lstrip` and `rstrip` example

pandas/core/strings.py
```diff
@@ -2151,12 +2151,60 @@ def encode(self, encoding, errors="strict"):
...
-    Series/Index from %(side)s. Equivalent to :meth:`str.%(method)s`.
+    Remove leading and trailing characters.
+
+    Strip whitespaces (including newlines) or a set of specified characters
+    from each string in the Series/Index from %(side)s.
+    Equivalent to :meth:`str.%(method)s`.
+
+    Parameters
+    ----------
+    to_strip : str or None, default None (whitespaces are removed)
```
jreback: the coment about whitespaces are removed should be on the next line with a case, e.g.
If None then whitespaces are removed

pandas/core/strings.py
```diff
@@ -2151,12 +2151,60 @@ def encode(self, encoding, errors="strict"):
...
+
+    See Also
+    --------
+    Series.str.strip : Remove leading and trailing characters in Series/Index
+    Series.str.lstrip : Remove leading characters in Series/Index
+    Series.str.rstrip : Remove trailing characters in Series/Index
+
+    Examples
+    --------
+    >>> s = pd.Series(['1. Ant.  ', '2. Bee!\n', '3. Cat?\t'])
```
jreback: can you add a np.nan in the example (in place of an entry or in addition is ok)

jreback (changes requested):

mroeschke:
Looks great! Will merge on green.

jreback:
@mroeschke merge away

jreback (approved):

mroeschke:
Thanks @Cheukting!

PR #20864, opened 2018-04-29
DOC: Improve the docstring of String.str.zfill()

alkamid:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

Result of the Pandas sprint at PyData London 2018. Includes work by @owlas.


pandas/core/strings.py
```diff
@@ -2114,19 +2114,55 @@ def rjust(self, width, fillchar=' '):
...
-        Equivalent to :meth:`str.zfill`.
+        Pad strings in the Series/Index by prepending '0' characters.
+
+        Strings in the Series/Index are padded with prepending '0' characeters
+        (i.e. on the left of the string) to reach a total string length
+        of `width`.  in the Series/Index with length greater than `width`
+        are unchanged.
+
+        Note: Differs from :meth:`str.zfill` which has special handling
+        for '+'/'-' in the string.
```
datapythonista: There is a `Notes` section in the numpy docstring convention that we can use for this:

```
Notes
-------
```

https://python-sprints.github.io/pandas/guide/pandas_docstring.html#section-6-notes

pandas/core/strings.py
```diff
@@ -2114,19 +2114,55 @@ def rjust(self, width, fillchar=' '):
...
         ----------
         width : int
-            Minimum width of resulting string; additional characters will be
-            filled with 0
+            Minimum length of resulting string; strings with length less
+            than `width` be prepended with '0' characters.
 
         Returns
         -------
         filled : Series/Index of objects
```
datapythonista: you can get rid of the `filled` and just leave the type (it was usual to name the return, but it was decided to not name them anymore as it doesn't add much value).

pandas/core/strings.py
```diff
@@ -2114,19 +2114,55 @@ def rjust(self, width, fillchar=' '):
...
+        4    423523
+        dtype: object
+
+        >>> s = pd.Series([-2, 5], dtype=str)
+        >>> s.str.zfill(5)
+        0    000-2
+        1    00005
+        dtype: object
         """
+
```
datapythonista: no need to add this blank line here

pandas/core/strings.py
```diff
@@ -2114,19 +2114,55 @@ def rjust(self, width, fillchar=' '):
...
+        2     00010
+        3     00127
+        4    423523
+        dtype: object
+
+        >>> s = pd.Series([-2, 5], dtype=str)
+        >>> s.str.zfill(5)
+        0    000-2
+        1    00005
+        dtype: object
```
datapythonista: the example looks cool, but I think we can make it a bit more compact and clear by:
- Just creating a single `Series` (I'd get rid of one of `10` or `127` as they show the same user case, and I'd add a `NaN` and a number (as number, not as string).
- After creating the `Series` I'd show it without modification (i.e. `>>> s`). This makes the result easier to compare with the original data.
- We can get rid of the second example if we add the numbers in the first.

datapythonista (commented):
Great job, added some comments with typos and ideas.

alkamid:
@datapythonista thanks a lot for your valuable feedback! I've addressed all your comments, does it look better to you now?

alkamid:
@datapythonista thanks again, suggestions incorporated into PR. Let me know what you think now.

datapythonista (approved):
lgtm

PR #18496, opened 2017-11-26
Construction of Series from dict containing NaN as key

toobaz:
- [x] closes #18480
closes #18515

- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

This is also a prerequisite for fixing #18455 (which is a prerequisite for fixing #18460 ). The workaround to #18485 is annoying, but it is easy to remove it when the bug is fixed.

PR #20865, opened 2018-04-29
PERF: Fixed regression in Series(index=idx) constructor

TomAugspurger:
From https://github.com/pandas-dev/pandas/pull/18496/

Special cases empty series construction, since the reindex is not necessary.

xref https://github.com/pandas-dev/pandas/issues/18532#issuecomment-385190350

Setup

```python
data = None
idx = date_range(start=datetime(2015, 10, 26),
                              end=datetime(2016, 1, 1),
                              freq='50s')
dict_data = dict(zip(idx, range(len(idx))))
data = None if data is None else dict_data
```

benchmark

```
%timeit out = Series(data, index=idx)
```

HEAD:

```
113 µs ± 6.22 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

0.21.0

```
96.5 µs ± 6.05 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Master

```
320 ms ± 11.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -207,10 +207,20 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
                 else:
                     data = data.reindex(index, copy=copy)
                 data = data._data
-            elif isinstance(data, dict):
+            elif isinstance(data, dict) and (len(data) or index is None):
+                # Include the len(data) check here, since _init_dict contains
+                # a relatively expensive reindex. When called with
+                # Series(data=None, index=idx`, that is unnescessary. We know
+                # we're all NaN anyway, so we handle this in the next block.
+                # https://github.com/pandas-dev/pandas/pull/18496/
                 data, index = self._init_dict(data, index, dtype)
                 dtype = None
                 copy = False
+            elif isinstance(data, dict):
+                # Same as previous block, but special cased for data=None,
+                # for performance when creating empty arrays.
+                data = np.nan
+
             elif isinstance(data, SingleBlockManager):
                 if index is None:
                     index = data.index
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -318,7 +318,9 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         # Now we just make sure the order is respected, if any
         if data and index is not None:
             s = s.reindex(index, copy=False)
-        elif not PY36 and not isinstance(data, OrderedDict):
+        elif not PY36 and not isinstance(data, OrderedDict) and data:
+            # Need the `and data` to avoid sorting Series(None, index=[...])
+            # since that isn't really dict-like
             try:
                 s = s.sort_index()
             except TypeError:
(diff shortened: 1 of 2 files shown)


TomAugspurger:
cc @toobaz

pandas/core/series.py
```diff
@@ -207,10 +211,20 @@ def __init__(self, data=None, index=None, dtype=None, name=None,
                 else:
                     data = data.reindex(index, copy=copy)
                 data = data._data
-            elif isinstance(data, dict):
+            elif isinstance(data, dict) and (len(data) or index is None):
```
jreback: rather than doing this here, why not inside ``_init_dict`` itself? (you can simply return early if these conditions ar met). you can also return the ``type, copy`` parameter as well. ideally like to locate all of a single 'type' code in the same place (dict here)

jreback (changes requested):
can you add an asv & a whatsnew (needed?)

pandas/core/series.py
toobaz: I agree - in particular, ``_init_dict`` already checks whether ``data`` is empty

TomAugspurger:
Good call on the init_dict. For some reason I thought it had to be in the main Series.__init__, since _init_dict returns a BlockManager & index, but this seems to work.

```
In [4]: %timeit out = Series(data, index=idx)
140 µs ± 1.24 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Shouldn't need a whatsnew since https://github.com/pandas-dev/pandas/pull/18496 was in 0.23. I think that was the original slowdown.

And this was caught by the `series_methods.SeriesConstructor.time_constructor(None)` asv, so we're good there (that's the benchmark I've been running).

TomAugspurger:
Does d8b1312 change your opinion on putting this stuff in `_init_dict`? It's getting borderline too messy now.

`Series._init_dict` sorts the index (keys). This would break

```python
Series(index=['b', 'a'])
```

Since the user input isn't actually dict-like, it's just going down _init_dict.

toobaz:
I might be wrong (can't test now, sorry), but a simpler solution could be to replace

https://github.com/pandas-dev/pandas/blob/4afc7563895830d224ac949f571edef2f069c314/pandas/core/series.py#L180

with

``` python
if data in [None, {}]:
    data = []
```

TomAugspurger:
That won't quiet work since `data` may not be hashable, but I like the
thought. Will try a variant.

TomAugspurger:
Ah, no, doesn't quite work since the list isn't broadcast to the length of
index. We raise on Series([], index=[0]) since they're different lengths.

TomAugspurger:
Merging in 2 hours if there are no further comments. I think we're just choosing between the two least-bad outcomes here.

jreback:
let me look again

jreback (approved):

jreback:
lgtm. thanks (I believe you said this was after 0.22 so no whatsnew needed)

PR #20491, opened 2018-03-26
DOC: update the pandas.Series.str.endswith docstring

dcreekp:
Checklist for the pandas documentation sprint (ignore this if you are doing
an unrelated PR):

- [ X] PR title is "DOC: update the <your-function-or-method> docstring"
- [X ] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [ X] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [X ] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [ X ] It has been proofread on language by another sprint participant

Please include the output of the validation script below between the "```" ticks:

```
################################################################################
#################### Docstring (pandas.Series.str.endswith) ####################
################################################################################

Test if the end of each string element matches a pattern.

Equivalent to :meth:`str.endswith`.

Parameters
----------
pat : str
    Character sequence. Regular expressions are not accepted.
na : object, default NaN
    Object shown if element tested is not a string.

Returns
-------
Series or Index of bool
    A Series of booleans indicating whether the given pattern matches
    the end of each string element.

See Also
--------
str.endswith : Python standard library string method.
Series.str.startswith : Same as endswith, but tests the start of string.
Series.str.contains : Tests if string element contains a pattern.

Examples
--------
>>> s = pd.Series(['bat', 'bear', 'caT', np.nan])
>>> s
0     bat
1    bear
2     caT
3     NaN
dtype: object

>>> s.str.endswith('t')
0     True
1    False
2    False
3      NaN
dtype: object

Specifying `na` to be `False` instead of `NaN`.

>>> s.str.endswith('t', na=False)
0     True
1    False
2    False
3    False
dtype: bool

################################################################################
################################## Validation ##################################
################################################################################

Docstring for "pandas.Series.str.endswith" correct. :)
```

PR #20870, opened 2018-04-29
DOC: Improve the docstring of Str.contains()

Blair-Young:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

```############################################################################

Errors found:
	Errors in parameters section
		Parameter "flags" description should start with capital letter
```
`Achieved through PyData 2018 pandas sprint.`


Blair-Young:
Unsure best way to phrase `flags` parameter as description starts with the ```re``` module.

pandas/core/strings.py
```diff
@@ -295,20 +295,24 @@ def str_count(arr, pat, flags=0):
 
 def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
     """
-    Return boolean Series/``array`` whether given pattern/regex is
-    contained in each string in the Series/Index.
+    Test if pattern or regex is contained within each string of a Series.
```
WillAyd: This method is used by `Index` as well which was removed here - please be sure to add that back

pandas/core/strings.py
```diff
@@ -295,20 +295,24 @@ def str_count(arr, pat, flags=0):
...
+
+    Return boolean Series based on whether a given pattern or regex is
```
WillAyd: Related to comment above, this isn't always true

```python
index = pd.Index(['foo>>> index = pd.Index(['foo', 'bar', 'baz'])
>>> index.str.contains('a')
array([False,  True,  True])
```

pandas/core/strings.py
```diff
@@ -295,20 +295,24 @@ def str_count(arr, pat, flags=0):
...
-    na : default NaN, fill value for missing values.
+        re module flags, e.g. re.IGNORECASE.
+    na : default NaN
+        Fill value for missing values.
     regex : bool, default True
-        If True use re.search, otherwise use Python in operator
+        If True, assumes the passed-in pattern is a regular expression.\n
+        If False, treats the pattern as a literal string.
 
     Returns
```
WillAyd: Maybe use the Returns section from the below PR as reference

https://github.com/pandas-dev/pandas/pull/20491/files

pandas/core/strings.py
```diff
@@ -295,20 +295,24 @@ def str_count(arr, pat, flags=0):
...
+        If True, case sensitive.
     flags : int, default 0 (no flags)
-        re module flags, e.g. re.IGNORECASE
-    na : default NaN, fill value for missing values.
+        re module flags, e.g. re.IGNORECASE.
+    na : default NaN
+        Fill value for missing values.
     regex : bool, default True
-        If True use re.search, otherwise use Python in operator
+        If True, assumes the passed-in pattern is a regular expression.\n
```
WillAyd: Rogue newline?

WillAyd (changes requested):
Thanks for submitting this

pandas/core/strings.py
```diff
@@ -295,20 +295,24 @@ def str_count(arr, pat, flags=0):
...
-    case : boolean, default True
-        If True, case sensitive
+    pat : str
+        Character sequence or regular expression.
+    case : bool, default True
+        If True, case sensitive.
     flags : int, default 0 (no flags)
-        re module flags, e.g. re.IGNORECASE
-    na : default NaN, fill value for missing values.
+        re module flags, e.g. re.IGNORECASE.
```
WillAyd: To suppress the warning just rewrite this as "Flags to pass through to the re module, e.g. re.IGNORECASE." or something similar

pandas/core/strings.py
```diff
@@ -318,6 +322,48 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
...
+
+    >>> s.str.contains('og', regex=False)
+    0    False
+    1     True
+    2    False
+    3    False
+    4      NaN
+    dtype: object
+
+    >>> s.str.contains('og', na=False, regex=True)
```
WillAyd: These examples are good but some text to direct the user on what to look at preceding each would be helpful. For example here I would prepend something like:

```With the above examples, you'll note that NA values in the caller return `np.nan` by default and as a result cast the dtype of the returned object to `object`. You can control this behavior by specifying a an alternate `na` argument```

File: pandas/core/strings.py [hunk 7]
```diff
@@ -318,6 +322,48 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
     --------
     match : analogous, but stricter, relying on re.match instead of re.search
 
+    Examples
```
WillAyd: Can you add examples for the `case` and `flags` parameters?

WillAyd (changes requested):

Blair-Young:
Hey @WillAyd I've tried to fix your correction. What's the next step?

jreback (approved):
lgtm. @TomAugspurger @jorisvandenbossche

WillAyd (approved):

TomAugspurger:
Thanks @Blair-Young!

Blair-Young:
Thanks @TomAugspurger looking forward to more.

PR #20872, opened 2018-04-29
DOC: Link to new PyPI instead of legacy PyPI

jschendel:
The new PyPI site has [officially launched](https://pythoninsider.blogspot.ca/2018/04/new-pypi-launched-legacy-pypi-shutting.html), and they've recommended [linking to the new site](https://warehouse.readthedocs.io/api-reference/integration-guide/#migrating-to-the-new-pypi).

Links to the legacy site are currently redirecting to the new site, so the content of what users will actually see isn't different than with the existing links, but seems best to avoid the redirect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 <tr>
   <td>Latest Release</td>
   <td>
-    <a href="https://pypi.python.org/pypi/pandas/">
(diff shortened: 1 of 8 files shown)


jreback:
thanks for the update!

Issue #20862, opened 2018-04-29
Problem Compiling Docs using make.py 


ThiviyanThanapalasingam:
When there is a space in the name of a parent directory, make.py fails to run. 

```
$ python make.py html
```

This is the output of the command: 
```
usage: sphinx-build [OPTIONS] SOURCEDIR OUTPUTDIR [FILENAMES...]
sphinx-build: error: cannot find source directory (/Users/usr/Documents/Pandas Sprint/pandas/doc/Sprint/pandas/doc/build/doctrees)
```

However, when after removing the space from the directory name (i.e. change 'Pandas Sprint' to 'PandasSprint') the python script executed as expected. 

Hope this helps. 
Thiviyan

ThiviyanThanapalasingam:
Thank you @jorisvandenbossche. I seem to have solved the issue by modifying 3 lines of code. As this will be my first contribution towards the Pandas project I am struggling to write a git commit message that can be readily understood by the community. Would the following message be fine?

**DOC: Fix whitespace in pathnames issue by wrapping paths with quotes (#20862)**

jorisvandenbossche:
@ThiviyanThanapalasingam That sounds good!

jorisvandenbossche:
Closed by https://github.com/pandas-dev/pandas/pull/20880

PR #20880, opened 2018-04-30
DOC: Handle whitespace in pathnames

thiviyanT:
Fixed make.py - Now should be able to run make.py in paths with whitespaces (#20862)


jreback:
@datapythonista @jorisvandenbossche 

how does this look?

datapythonista:
I was taking a look at it yesterday, as I'm not sure in windows the single quotes will handle backslashes well. But unrelated to the ticket I'm getting a segmentation fault when running sphinx-built. Will comment on the ticket with feedback when I manage to test it.

jorisvandenbossche:
Can't test on Windows at the moment, but for linux this patch works nicely and fixes the problem for me.

datapythonista:
@ThiviyanThanapalasingam seems like in Windows the paths need to be escaped with double quotes (the backlash is interpreted as an escape character like in `\n`).

Besides that, I personally find it's not very readable to use backslashes to escape the quotes. Meaning that instead of `'\'{}\''` I think it'd be more clear to use `"'{}'"`. Or now that we need to change to double quotes, the opposite: `'"{}"'`.

Also, I see that there are several paths that are built like `source/*.rst` (instead of `os.path.join('source', '*.rst')`. I couldn't try, but I guess sphinx is failing to exclude these files like in unix systems. Do we want to open an issue for making the script work in Windows @jreback?

datapythonista:
@ThiviyanThanapalasingam do you have time to update the PR based on the feedback I provided?

ThiviyanThanapalasingam:
Yes, I can do that. @datapythonista Have you opened another issue for making the script work in Windows?

datapythonista:
No, I think it should be just the one you opened.

jreback:
@datapythonista ?

datapythonista:
@ThiviyanThanapalasingam can you update the PR based on the previous feedback?

datapythonista:
Made the updates to this PR. But seems like there is a new unrelated problem when the path contains whitespaces: #22836

I think that can be addressed in a separate PR. I would merge this on green, but happy to close this and take care of both problems in a new PR.

jorisvandenbossche:
Let's do that in a separate PR

@ThiviyanThanapalasingam thanks for the original PR!

Issue #20825, opened 2018-04-25
BUG: Series.combine() fails with ExtensionArray inside of Series


Dr-Irv:
#### Code Sample, a copy-pastable example if possible

```python
In [1]: import pandas as pd

In [2]: from pandas.tests.extension.decimal.array import DecimalArray, make_dat
   ...: a

In [3]: da1= make_data()
   ...: da2= make_data()
   ...:

In [4]: s1 = pd.Series(DecimalArray(da1))
   ...: s2 = pd.Series(DecimalArray(da2))
   ...:

In [5]: s1.head(), s2.head()
Out[5]:
(0    0.57581534881735985109685316274408251047134399...
 1    0.05647135567908745379384072293760254979133605...
 2    0.41049738961593973396446699553052894771099090...
 3    0.13724377491342376611527242857846431434154510...
 4    0.24154934068629707599740186196868307888507843...
 dtype: decimal, 0    0.40855027024154888515283801098121330142021179...
 1    0.21243084028671055385473209753399714827537536...
 2    0.15218065149055393092680787958670407533645629...
 3    0.87747422249812989658579454044229350984096527...
 4    0.53991488184898328572813852588296867907047271...
 dtype: decimal)

In [6]: s1.combine(s2, lambda x1, x2: x1 if x1 < x2 else x2)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-14abc20f0095> in <module>()
----> 1 s1.combine(s2, lambda x1, x2: x1 if x1 < x2 else x2)

C:\EclipseWorkspaces\LiClipseWorkspace\pandas-dev\pandas36\pandas\core\series.py in combine(self, other, func, fill_value)
   2220             new_index = self.index.union(other.index)
   2221             new_name = ops.get_op_result_name(self, other)
-> 2222             new_values = np.empty(len(new_index), dtype=self.dtype)
   2223             for i, idx in enumerate(new_index):
   2224                 lv = self.get(idx, fill_value)

TypeError: data type not understood

```
#### Problem description

The `Series.combine()` method uses `numpy.empty` with the `dtype` of the `ExtensionArray`, and `numpy` isn't happy with that.

Note:  This also happens with `Categorical` in v0.22 and in master:
```
In [3]: cat1 = pd.Categorical(values=["one","two","three","three","two","one"],
   ...:  categories=["one","two","three"], ordered=True)
[...]

Issue #19577, opened 2018-02-07
ExtensionArray Ops discussion


TomAugspurger:
Split from https://github.com/pandas-dev/pandas/pull/19520

How to handle things like `__eq__`, `__add__`, etc.

Neither the Python 2 or Python 3 object defaults are appropriate, so I think we should make it abstract or provide a default implementation for some / all of these.

We should be able to do a default implementation for the simple case of binop(extension_array, extension_array) by

- masking nulls in either self | other
- cast self / other to ndarrays
- compare and apply null mask

@jorisvandenbossche raises the point that for some extension arrays, casting to an object ndarray can be expensive. It'd be unfortunate to do the casting if we're just going to raise a `TypeError` when the `binop` is done. In this case the subclass will need to override all those ops (or we could add class attribute like `_comparable` and check that before doing `__eq__` or `__lt__`, etc).

How much coercion do we want to attempt? I would vote to stay simple and only attempt a comparison if

    other is an instance of the same ExtensionArray type
    other is a scalar of self.dtype.type.

Otherwise we return NotImplemented. Subclasses can of course choose more or less aggressive coercion rules.

When boxed in a `Series` or `Index`, I think the goal should be

1. dispatch to the underlying `.values` comparsion
2. Reconstruct the appropriate output (box in series / index with name).

jbrockmendel:
Migrated from #19520:

> A thorny issue: What to do with comparison methods

I've been planning on moving the Index/Series arithmetic/comparison ops into Array subclasses, so that `Series.__op__(other)` could just wrap `self._data.blocks[0].__op__(other)` and we get consistency between Series/DataFrame ops for free.

Most of the comparison ops are ready to make that jump; I can prioritize it if getting that done quickly will help you out.

One caveat I'm concerned about is that the existing implementations in the Index subclasses have gotten tangled up with a bunch of unrelated Index machinery. Ideally I'd like these operations to go into self-contained mixin classes that rely on constructor methods, but are independent of slicing/concat/reindex/dropna/... mentioned above.

jbrockmendel:
@TomAugspurger two quick namespace-gameplan questions.

Assuming the arith/comparison methods currently in DatetimeIndexOpsMixin/DTI/TDI/PI get moved into analogous array classes, do you envision these getting a) mixed into the appropriate Index/Block subclasses or b) accessed via composition?  If the latter, what name (`DatetimeIndex._??_values`) should these use to get at the pdarray implementation?

[...]

PR #20885, opened 2018-04-30
BUG: Fix Series.get() for ExtensionArray and Categorical

Dr-Irv:
- [x] closes #20882
- [x] tests added / passed
  - pandas/tests/extension/base/getitem.py::test_get()
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] whatsnew entry
  - Not created - assume will go in v0.23 as part of ExtensionArray support

PR #19972, opened 2018-03-02
Small ops cleanups

jbrockmendel:
Consistently use `op_name` and `res_name` and avoid `name`

Use find_common_type instead of np.find_common_type, moving towards de-duping Series/Frame code.

PR #20889, opened 2018-04-30
ENH: Support operators for ExtensionArray

### LINKED ISSUE #20659: ExtensionArray support for comparisons to fundamental types and lists of those types

Dr-Irv:
- [x] closes #20659, #19577, #20825
- [x] tests added / passed
  - pandas/tests/extension/decimal/test_decimal.py
    - Tests for operators using existing operator tests for fundamental types
  - pandas/tests/extension/json/test_json.py
    - Create an `__add__` method for json and test it, make sure others fail
  - pandas/tests/extension/category/test_categorical.py
    - Adds a `test_combine` method
- [x] whatsnew entry
  - Not included since I'm not sure if it is v0.23 or v0.24, and I wasn't about to start a v0.24 whatsnew

### Some Notes

- @jorisvandenbossche suggested that I put up an initial implementation for review, so I expect there will be lots of suggestions for doing it more, better, or different.
- There is a hook in there that is untested, where if an implementer (like me) wants to have the logical operators not return a `Series` with `bool`, they can do that.  I'm testing it with my library, but once we agree on this basic implementation, I will add tests for that specific use case.
- The idea is that you first see if the `ExtensionArray` class has implemented the operator.  If not, then you just go element by element.


jreback:
just an FYI, rather than a single PR, this needs to be a series of smaller, self-consistent PR's. much easier to review and revise, otherwise this will get bogged down.

Dr-Irv:
@jreback  Right now, I had to include some code from an early version of #20885.  Once that is approved, that will shorten this one a little bit. I can then first submit something that deals with #20825 . And once that is approved, I can submit the things related to ops.

Let me know if this plan works for you.

jbrockmendel:
> just an FYI, rather than a single PR, this needs to be a series of smaller, self-consistent PR's. much easier to review and revise, otherwise this will get bogged down.

@Dr-Irv There are a bunch of planned/discussed steps between the status quo and the functionality in this PR.  If you're up to help with the whole transition that'd be great.  If you're mostly interested in getting this specific feature in place ASAP, we can talk about workarounds that you can apply in the interim.

@jreback LMK if I'm overstepping by offering to shepherd this process.  It's up my alley.

TomAugspurger:
> @jreback LMK if I'm overstepping by offering to shepherd this process. It's up my alley.

I'd appreciate your expertise here @jbrockmendel.

FWIW, I'll have some availability this week, but very little next week.

jbrockmendel:
> then I'd like to see the workaround so I can get my use case working

The workaround I have in mind assumes you are mostly interested in patching Series arithmetic and logical operators.  Is that accurate?

> and since I have put a lot of thought into getting this PR to work, I can also help with the transition

@Dr-Irv sounds good.  The first thing I'd suggest is figuring out if there is anything in this PR that is not strictly necessary for the feature(s) you need or can otherwise be separated into independent PRs.  Based on a very quick glance it looks like the tests for `get` or `combine` might be independent of the rest of the PR.

Are the changes to boolean ops logically independent of the changes to arithmetic ops?  If so, I suggest focusing on the boolean ops first, since they aren't _nearly_ so wound up with Index vs Series shenanigans.  There might be something in #19795 worth resurrecting.

Dr-Irv:
@jbrockmendel 
> The workaround I have in mind assumes you are mostly interested in patching Series arithmetic and logical operators. Is that accurate?

Yes, that is correct.  I also need the results of arithmetic and logical operators to be Series of objects.

This PR has 3 things in it. 

1. Code from #20885 that was still in review mode when I submitted the present PR.  The tests for `get` that you mention fall into that category.  I can remove that to make this PR more up-to-date.
2. Fixes to support `combine` correctly.  This addresses #20825 .  I can break that out into a separate PR.  But that is needed for testing the operator stuff.
3. The code to support the arithmetic and logical operators for `ExtensionArray`.  It is the same code that does both.

So, if you want, I can create a new PR that only works on `combine`.  Then this one we are discussing here will only have the operator support, the meat of which is in the new internal method `dispatch_to_extension_ops`.

jreback:
FYI I have this almost all implemented for integer na support: https://github.com/jreback/pandas/commit/6fc19f9c6f01f0fd42d0551aa1defd6f20bc3fb9 (this is a bit stale), will be updated soon.

jorisvandenbossche:
I think it is still worthwhile to push forward this PR to implement this separately from the integer-array PR (which might take more discussion, and it also a very big one right now). 

@jreback can you give more concrete feedback on the way @Dr-Irv implemented things here, based on what you did in the other PR? 
(once the unrelated changes are removed (for `combine` and `get`), I don't think this PR needs to be splitted)

> There are a bunch of planned/discussed steps between the status quo and the functionality in this PR. 

@jbrockmendel can you elaborate on this a bit? (regarding this PR, what would you do instead)

pandas/core/arrays/base.py
```diff
@@ -53,6 +53,13 @@ class ExtensionArray(object):
     * factorize / _values_for_factorize
     * argsort / _values_for_argsort
 
+    For logical operators, the default is to return a Series of boolean.
+    However, if the underlying ExtensionDtype overrides the logical
+    operators, then the implementer may want to have an ExtensionArray
+    subclass contain the result.  This can be done by changing the property
+    _logical_result from its default value of None to the _from_sequence
```
jorisvandenbossche: Why is it needed to have this property? Can't we simply detect whether the result is a boolean numpy array or again an ExtensionArray ?

pandas/core/arrays/base.py
```diff
@@ -610,3 +620,14 @@ def _ndarray_values(self):
         used for interacting with our indexers.
         """
         return np.array(self)
+
+    # ------------------------------------------------------------------------
+    # Utilities for use by subclasses
+    # ------------------------------------------------------------------------
+    def is_sequence_of_dtype(self, seq):
```
jorisvandenbossche: For what is this needed?

pandas/core/ops.py
```diff
@@ -990,6 +991,93 @@ def _construct_divmod_result(left, result, index, name, dtype):
     )
 
 
+def dispatch_to_extension_op(left, right, op_name=None, is_logical=False):
```
jorisvandenbossche: the `dispatch_to_index_op` uses `op` instead of `op_name`. Is there a reason for this difference? (and I mean in the actual implementation here it assumes a method name, and not an operator function that can be called)

pandas/core/ops.py
```diff
@@ -990,6 +991,93 @@ def _construct_divmod_result(left, result, index, name, dtype):
...
+    if method is not None and inspect.ismethod(method):
+        rvalues = right
+        if is_extension_array_dtype(right) and isinstance(right, ABCSeries):
+            rvalues = right.values
+        try:
+            res = method(rvalues)
+        except TypeError:
+            pass
+        except Exception as e:
+            raise e
```
jorisvandenbossche: Why not doing `op(left.values, right/right.values)`? What does this manual checking/trying does that the former does not?

pandas/core/ops.py
```diff
@@ -990,6 +991,93 @@ def _construct_divmod_result(left, result, index, name, dtype):
...
+    """
+    Assume that left is a Series backed by an ExtensionArray,
+    apply the operator defined by op_name.
+    """
+
+    method = getattr(left.values, op_name, None)
+    deflen = len(left)
+    excons = type(left.values)._from_sequence
+    exclass = type(left.values)
+    testseq = left.values
```
TomAugspurger: I'm having trouble understanding these names. (method makes sense).

pandas/core/ops.py
```diff
@@ -990,6 +991,93 @@ def _construct_divmod_result(left, result, index, name, dtype):
...
+    def convert_values(parm):
+        if is_extension_array_dtype(parm):
+            ovalues = parm.values
+        elif is_list_like(parm):
+            ovalues = parm
+        else:  # Assume its an object
+            ovalues = [parm] * deflen
+        return ovalues
+
+    if res is NotImplemented:
```
TomAugspurger: Could you explain this fallback a bit more?

If the EA doesn't define ops, then I'm perfectly fine with raising `NotImplementedError` at the end.

pandas/core/arrays/base.py
jbrockmendel: I can't speak for the author, but my assumption was that this has to do with some of the spaghetti-code in `ops._bool_method_SERIES`, where sometimes a bool-dtype is returned and other times an int-dtype is returned (and datetimelike are currently all broken, see #19972, #19759).  Straightening out this mess independently of EA implementations is part of the plan referred to above.

jbrockmendel:
>> There are a bunch of planned/discussed steps between the status quo and the functionality in this PR.

> can you elaborate on this a bit? (regarding this PR, what would you do instead)

@jorisvandenbossche 
- I'm pretty sure there is some test code here that is orthogonal to the rest of the PR; I'd separate that out to make review easier.
- The place to define the element-by-element fallback (essentially object-dtype behavior) is in the downstream EA subclass.
- Eventually _all_ of the arith/comparison Series ops are intended to dispatch to their underlying EA objects (possibly via `self._data.blocks[0])`.  (This will render dispatch_to_index_op unnecessary)
- Some steps towards making that happen include:
  - Something like #21160 (and analogues for float, bool, object, datetimelike (#19902), ... dtypes)
  - Align remaining mismatches between Series vs Index arithmetic/comparison behavior #19322, #18824, #19369.  Note 19322 is the most relevant one for the Integer EA case.
  - Categorical EA (and possibly the base class) should implement the relevant ops, even if they just raise.  They also need to figure out some inconsistencies between Categorical vs CategoricalIndex behavior (#19513)
 - As mentioned in a comment above, the Series bool ops need some serious work, and I'd much prefer to have that squared away before trying to make this transition. #19792, #19759

jorisvandenbossche:
@jbrockmendel but is it needed to fix all of this (make sure all of Series ops are dispatched correctly etc, resolve duplication between index and series, ...) before getting basic ops working for ExtensionArrays? 
It's true there is a lot of work to do the full internal transition, but that doesn't mean we can already support basic ops for external extension arrays.

jreback:
echo the above comments, this has a lot of othrogonal changes going on. let's strip out things to the minimal changes.

pandas/core/arrays/base.py
Dr-Irv: Actually, in my use case, I need the boolean operators to return an object that represents the relation.  I'm using pandas on top of 2 different libraries (that functionally are the same) where the operators (x <= y), (x >= y) and (x == y) are not booleans, but objects representing the relations.

pandas/core/arrays/base.py
Dr-Irv: Let's suppose that you don't implement the `ExtensionArray` operators as methods of the subclass of `ExtensionArray`, but you let the underlying `ExtensionDtype` handle the operators for you.  (This is what I used for `Decimal`).  Some of the operators will return a sequence containing all objects of `ExtensionDtype`.  Some operators (e.g., logical ones), will not.  So internally, it's useful to have a test to know whether a sequence has objects of the corresponding `ExtensionDtype` so that you can then return an `ExtensionArray` as a result, otherwise, you just let things get coerced based on the type in the sequence.

pandas/core/ops.py
Dr-Irv: `excons` is the constructor used to construct a result from a sequence that has all its elements of type `ExtensionDtype`.  (so "ex" for "Extension" and "cons" for "constructor")  `exclass` is the underlying class of the `ExtensionArray` subclass.

pandas/core/ops.py
Dr-Irv: See above.  The idea here is that either the EA defines the ops, *or* the `ExtensionDtype` defines the ops. 

The idea here is that if you know that your underlying `ExtensionDtype` already has the ops defined, you don't have to implement each of the ops at the `ExtensionArray` level.

I used `DecimalArray` as an example.  The operators are already defined for `Decimal` so there is no reason to have to implement the operators for an array of `Decimal`.

Dr-Irv:
In response to @jbrockmendel:

> The place to define the element-by-element fallback (essentially object-dtype behavior) is in the downstream EA subclass.

I really disagree with this.  If you already have the operators defined for the `ExtensionDtype`, why should you have to implement the operators for `ExtensionArray`?  My implementation gives the person implementing an `ExtensionArray` subclass the option of defining the operators at the EA level, or letting the operators defined (if they are defined) for the `ExtensionDtype` do all of the work.

Dr-Irv:
@jreback Since you wanted something smaller, I have split off the code that fixes #20825 into #21183 .  This is to handle the issue with `combine`.  I suggest that people reviewing this PR look at #21183 first.  Once it is accepted, I will then update this PR to only have the ops functionality.

Note that the `combine` bug has to be fixed in order for the tests in this PR to work right.  So hopefully you will accept #21183, and then this PR for ops will end up being smaller.

jorisvandenbossche:
Why do the tests need the combine fix?

Dr-Irv:
@jorisvandenbossche To test the operator functionality, I am leveraging the code that tests operators for `Series` that is in `pandas/tests/series/test_operators.py:TestSeriesOperators`, and those tests use `combine` to verify the results.  By fixing the bug in `combine`, then the tests for operators are leveraging existing test code.

jbrockmendel:
> I really disagree with this. If you already have the operators defined for the ExtensionDtype, why should you have to implement the operators for ExtensionArray? My implementation gives the person implementing an ExtensionArray subclass the option of defining the operators at the EA level, or letting the operators defined (if they are defined) for the ExtensionDtype do all of the work.

Implementing it on the DType is fine, I guess.  It should be _really easy_ to go from there to implementing it on the EA subclass.  Once that is done, the changes to `core.ops` (and I hope/think `Series._binop`) can pretty much all be skipped and just call `dispatch_to_index_op` with the appropriate EA subclass (this is already done with pd.Categorical on L1193)

In most cases returning `NotImplemented` means "try the reversed operation".  I'd like to avoid giving that return value a double meaning if possible.

Dr-Irv:
@jbrockmendel You wrote:

> Implementing it on the DType is fine, I guess. It should be _really easy_ to go from there to implementing it on the EA subclass. 

Are you suggesting that for someone implementing their own EA subclass, it should be really easy for _them_ to implement the operators on the EA subclass?  Or are you suggesting that pandas provide default operators for `ExtensionArray` that are defined based on the operators on the `ExtensionDtype`? If the former, I don't see how we can do that.  If the latter, the implementation in this PR does that.

The changes in this PR that are in `Series._binop` are now moved to #21183 in order to make `combine` work.

I can investigate whether `dispatch_to_index_ops` can do what `dispatch_to_extension_ops` does, but I'm doubtful, because my goal was to allow supporting the operators whether they were defined by the user at the EA subclass level or at the `ExtensionDtype` subclass level.

jorisvandenbossche:
To try to take a step back from the discussion:

- We want that it is easy for ExtensionArray ops to just fall back to the scalar ops (the `ExtensionDtype.type` objects might already define ops, so use that). 
  - For this reason, @Dr-Irv implemented this fall back inside `ops.py` in case the ExtensionArray has no method for the specific `op`.
- On the other hand, we can also argue that it might be easier that the ops are always defined on the ExtensionArray. That would make the implementation inside `ops.py` a lot easier, as the fallback to the `ExtensionDtype.type` scalars could be removed.
- I would like to add that for my use case, I want that it is possible for the ExtensionArray to indicate that *no* such fallback should be tried (as this can be expensive). Defining the op and letting it raise might be enough for that (however, currently `TypeError` is catched and then still the fallback is tried. I think we should not do that).

Given the above, I think we can try to harmonize the goal of making this fallback both easy as explicit (1 and 3), and keep the ops defined on the ExtensionArray and `ops.py` simple (2). 
What if we **provide a mixin that defines all those basic fallbacks?** In that case, if the extension author decides it would like to enable the fallback, it can simply inherit from the mixin. Also, if then the behaviour of certain methods (like the booleans) are not exactly what he/she wants, just overwrite that specific method again in your array class.

pandas/core/ops.py
jorisvandenbossche: If the op is not defined on ExtensionArray, calling it directly (`op(left_values, right_values)`) will raise a TypeError that you can catch (which you already do), so I don't really see the difference

TomAugspurger:
Regarding the elementwise fallback to `ExtensionDtype.type`, isn't that relatively easy to define on the EA subclass?

```python
# in MyEASubclass
def __lt__(self, other):
    return np.array([a < b for a, b in zip(self, other)])
```

likewise for the other methods.

> I would like to add that for my use case, I want that it is possible for the ExtensionArray to indicate that no such fallback should be tried (as this can be expensive).

For cyberpandas, the scalar fall back would be incorrect. We use IPAddress(0) as the NA value, so comparisions with it should be NA to match pandas behavior.

```python
In [3]: arr = IPArray([0, 1, 2, 3])

In [4]: arr <= arr
Out[4]: array([False,  True,  True,  True])

In [5]: arr.astype(object) <= arr.astype(object)
Out[5]: array([ True,  True,  True,  True])
```

A mixin seems like a reasonable compromise.

jorisvandenbossche:
> Regarding the elementwise fallback to ExtensionDtype.type, isn't that relatively easy to define on the EA subclass?

It's indeed relatively straightforward, although a bit more complicated as you small example if you want to cover all cases (scalar vs array `other` etc). Therefore providing a default implementation available as mixin would make this easier.

Dr-Irv:
@TomAugspurger You wrote:

> Regarding the elementwise fallback to ExtensionDtype.type, isn't that relatively easy to define on the EA subclass?

Yes, but having to do it for *all* of the operators is painful.  What I implemented avoids that.  Take the example of `DecimalArray`.  You'd like to have to avoid implementing every single operator supported by `decimal.Decimal` in `DecimalArray`.

I like the idea of using a mixin as proposed by @jorisvandenbossche.  However, it seems that I ought to wait until @jreback finishes #21191 and #21160 and it is merged in, and then I can see where things are and make modifications as needed for my use case.

jbrockmendel:
@Dr-Irv 
> Are you suggesting that for someone implementing their own EA subclass, it should be really easy for them to implement the operators on the EA subclass?

Yes, this.

> Or are you suggesting that pandas provide default operators for ExtensionArray that are defined based on the operators on the ExtensionDtype?

The element-by-element implementation in the PR is effectively the object-dtype behavior, and I'd be fine with that being a default at the BaseArray level (@TomAugspurger thoughts?  It would also be reasonable for BaseArray to implement these ops to raise `com.AbtractMethodError`).

> The changes in this PR that are in Series._binop are now moved to #21183 in order to make combine work.

I'll take a closer look at #21883, thanks.

BTW, thanks for pushing forward on this.  You've clearly put some thought+effort into this and it is very much appreciated.

@jorisvandenbossche 
> but is it needed to fix all of this (make sure all of Series ops are dispatched correctly etc, resolve duplication between index and series, ...) before getting basic ops working for ExtensionArrays?

Not at all, the steps I mentioned were very much for the "what would you do instead" question.  I _do_ think that implementing ops at the array level will minimize duplicated effort in `core.ops` (a lot of effort went in to un-tangling older special-casing in core.ops, a process which I'm not eager to repeat).

> What if we provide a mixin that defines all those basic fallbacks? In that case, if the extension author decides it would like to enable the fallback, it can simply inherit from the mixin. Also, if then the behaviour of certain methods (like the booleans) are not exactly what he/she wants, just overwrite that specific method again in your array class.

+1.

Dr-Irv:
My plan right now is to look at the mixin approach on May 29 or later.

When I created this PR, the whole point was to inspire discussion on the design and implementation, so I think this discussion has been healthy in terms of us moving forward.

Dr-Irv:
Closing this, as there is a new proposed implementation in #21261 .

PR #20892, opened 2018-05-01
API: Implement set_freq for DTI/TDI, deprecate freq setter

### LINKED ISSUE #20886: API: DatetimeIndex.set_freq
leftover from https://github.com/pandas-dev/pandas/pull/20772

IIUC, the proposal is for `datetime_index.set_freq(new_freq)` to be an alternative to `datetimeindex.freq = new_freq` (please correct me if I'm wrong).

This would also apply to TimedeltaIndex.

jschendel:
- [X] closes #20886
- [X] tests added / passed
- [X] whatsnew entry

Summary:
- Follow-up to #20772
  - See that PR for relevant discussion
- Implements a `set_freq` method for `DatetimeIndex` and `TimedeltaIndex`
- Deprecates the `freq` setter for `DatetimeIndex` and `TimedeltaIndex`
   - Had to change some instances of the DTI/TDI implementation to directly set the `_freq` attribute
   - Had to modify some tests to use `set_freq` instead of the setter
   - Didn't see any related deprecation warnings in the logs from the CI runs hooked to my local repo

cc @jreback @jorisvandenbossche

### DIFF
--- a/pandas/core/indexes/datetimelike.py
+++ b/pandas/core/indexes/datetimelike.py
@@ -236,11 +236,43 @@ class TimelikeOps(object):
 
     @freq.setter
     def freq(self, value):
+        msg = ('Setting {obj}.freq has been deprecated and will be removed '
+               'in a future version; use {obj}.set_freq instead.'
+               ).format(obj=type(self).__name__)
+        warnings.warn(msg, FutureWarning, stacklevel=2)
         if value is not None:
             value = frequencies.to_offset(value)
             self._validate_frequency(self, value)
 
         self._freq = value
+
+    def set_freq(self, freq):
+        """
+        Set the frequency of the DatetimeIndex or TimedeltaIndex to the
+        specified frequency `freq`.
+
+        Parameters
+        ----------
+        freq: str or Offset
+            The frequency to set on the DatetimeIndex or TimedeltaIndex
+
+        Returns
+        -------
+        new: DatetimeIndex or TimedeltaIndex with the new frequency
+
+        Raises
+        ------
+        ValueError
+            If the values of the DatetimeIndex or TimedeltaIndex are not
+            compatible with the new frequency
+        """
+        if freq is not None:
+            freq = frequencies.to_offset(freq)
+            self._validate_frequency(self, freq)
+
+        new = self.copy()
+        new._freq = freq
+        return new
 
 
 class DatetimeIndexOpsMixin(object):
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -320,7 +320,7 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
     _other_ops = ['date', 'time']
     _datetimelike_ops = _field_ops + _object_ops + _bool_ops + _other_ops
     _datetimelike_methods = ['to_period', 'tz_localize',
-                             'tz_convert',
+                             'tz_convert', 'set_freq',
                              'normalize', 'strftime', 'round', 'floor',
                              'ceil', 'month_name', 'day_name']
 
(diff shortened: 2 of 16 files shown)


pandas/tests/indexes/datetimes/test_date_range.py
```diff
@@ -399,7 +400,6 @@ def test_daterange_bug_456(self):
         # GH #456
         rng1 = bdate_range('12/5/2011', '12/5/2011')
         rng2 = bdate_range('12/2/2011', '12/5/2011')
-        rng2.freq = BDay()
```
jschendel: This is redundant, as `bdate_range` should already set this as the frequency.  I couldn't find a test that verified this though, so I added a check to a test a few lines above.  Same deal with the `CDay()` changes below.

pandas/core/indexes/datetimes.py
```diff
@@ -1730,9 +1731,13 @@ def offset(self):
     def offset(self, value):
         """get/set the frequency of the Index"""
         msg = ('DatetimeIndex.offset has been deprecated and will be removed '
-               'in a future version; use DatetimeIndex.freq instead.')
+               'in a future version; use DatetimeIndex.set_freq instead.')
         warnings.warn(msg, FutureWarning, stacklevel=2)
-        self.freq = value
+        if value is not None:
+            value = to_offset(value)
```
jreback: call set_freq here

jreback (changes requested):
can u add to api.rst

pandas/core/indexes/datetimes.py
jschendel: This is for the deprecated `offset` setter which is just an alias for the `freq` setter, so I don't think `set_freq` should be used here since setting is to happen inplace (as was the previous convention).  This previously would call the `freq` setter, but since that's also being deprecated it was subsequently raising the `freq` setter warning too, so I just reused the `freq` setter code to avoid the additional warning.

jschendel:
Added `set_freq` to `api.rst`

pandas/core/indexes/datetimelike.py
```diff
@@ -236,12 +236,44 @@ def freq(self):
...
+               'in a future version; use {obj}.set_freq instead.'
+               ).format(obj=type(self).__name__)
+        warnings.warn(msg, FutureWarning, stacklevel=2)
         if value is not None:
             value = frequencies.to_offset(value)
             self._validate_frequency(self, value)
 
         self._freq = value
 
+    def set_freq(self, freq):
```
jreback: can you make this available on the .dt accessors as well

jreback (changes requested):

pandas/core/indexes/datetimelike.py
jschendel: This currently does nothing when used with the `dt` accessor due to #20937; the underlying frequency gets changed with `set_freq`, but upon `Series` construction the new frequency gets lost.

I've pushed a commit to add this to the `dt` accessor, and xfailed the corresponding tests I've added.  Can revert the commit if this is no longer deemed important.

jreback (approved):

jreback:
nice PR @jschendel , lgtm.

any comments @jorisvandenbossche @TomAugspurger

jreback:
@jorisvandenbossche you commented on the issue. but this PR looks pretty good to me, and it allows chained operations, something we like, not to mention ``.dt`` friendliness.

jorisvandenbossche:
As I commented on the issue, I still don't see the need or added value to add a new method for this (and yes, this is a nicely done PR, that's not my problem :-)). See https://github.com/pandas-dev/pandas/issues/20886#issuecomment-386710752

BTW, I don't feel strongly about this (as it is something I will never use), so I am fine to go with the majority if there is one, I just have the feeling we are adding a new method for something we don't need (and the bar to add new methods should be higher)

jreback:
ok, closing this. thanks for the PR @jschendel we can revisit in the future if more interest.

PR #20905, opened 2018-05-01
Pyside2 is arriving

stonebig:
see http://blog.qt.io/blog/2018/04/13/qt-for-python-is-coming-to-a-computer-near-you/

- [x] install.rst new option from Qt World


TomAugspurger:
Have we verified that `read_clipboard` actually works with pyside2?

stonebig:
have you an example, so I can check in 30 seconds ?

stonebig:
as "Qt for Python" is our last chance to get a strong and official support of the Python-Qt bridge, we should help them gaining visibility.

TomAugspurger:
Sure, but we'll want to make sure that it actually works for
`read_clipboard` before recommending people install it :)

stonebig:
ok, anyway QtConsole team says it's unwise to suggest:
- an http link (corrected)
- an alpha-branded version .... so, closing

PR #20912, opened 2018-05-01
BUG: DatetimeIndex._data should return an ndarray

### LINKED ISSUE #20810: BUG: inconsistent state of DatetimeIndex._data
Depending on how a DatetimeIndex is constructed, the underlying `._data` attribute is a DatetimeIndex or a datetime64 ndarray:

```
In [1]: idx1 = pd.DatetimeIndex(start="2012-01-01", periods=3, freq='D') # date_range kind of construction

In [2]: idx1._data
Out[2]: DatetimeIndex(['2012-01-01', '2012-01-02', '2012-01-03'], dtype='datetime64[ns]', freq=None)

In [3]: idx2 = pd.DatetimeIndex(idx1)

In [4]: idx2._data
Out[4]: 
array(['2012-01-01T00:00:00.000000000', '2012-01-02T00:00:00.000000000',
       '2012-01-03T00:00:00.000000000'], dtype='datetime64[ns]')
```

I think this should always be a numpy array? (it clearly doesn't hurt, but I don't see any reason to have it sometimes as a DatetimeIndex)

This came out of fixing warnings in https://github.com/pandas-dev/pandas/pull/20721, and is due to how `_generate_regular_range` and `_simple_new` on DatetimeIndex are implemented. From the code of `_simple_new`, I suspect that it assumes the input is always an ndarray and not DatetimeIndex, but in several places (like `_generate_regular_range`) an already constructed DatetimeIndex is passed to `_simple_new`.

reidy-p:
- [x] closes #20810
- [ ] tests added / passed
- [x] whatsnew entry

The change I made seems to fix the case in the original issue without breaking any tests. 

On my branch:
```
In [1]: idx1 = pd.DatetimeIndex(start="2012-01-01", periods=3, freq='D') # date_range kind of construction

In [2]: idx1._data
array(['2012-01-01T00:00:00.000000000', '2012-01-02T00:00:00.000000000',
       '2012-01-03T00:00:00.000000000'], dtype='datetime64[ns]')

In [3]: idx2 = pd.DatetimeIndex(idx1)

In [4]: idx2._data
Out[4]: 
array(['2012-01-01T00:00:00.000000000', '2012-01-02T00:00:00.000000000',
       '2012-01-03T00:00:00.000000000'], dtype='datetime64[ns]')
```

But is the solution too simple or is something more sophisticated required?

And do we need tests for this issue?

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -506,6 +506,9 @@ class Index(IndexOpsMixin, PandasObject):
         attributes.update(kwargs)
         if not len(values) and 'dtype' not in kwargs:
             attributes['dtype'] = self.dtype
+        from pandas import DatetimeIndex
+        if isinstance(values, DatetimeIndex):
+            values = values.values
         return self._simple_new(values, **attributes)
 
     def _shallow_copy_with_infer(self, values=None, **kwargs):
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -41,7 +41,6 @@ from pandas.core.dtypes.common import (
     is_signed_integer_dtype,
     is_unsigned_integer_dtype,
     is_integer_dtype, is_float_dtype,
-    is_datetime64_dtype,
     is_datetime64_any_dtype,
     is_datetime64tz_dtype,
     is_timedelta64_dtype,
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -41,6 +41,7 @@ from pandas.core.dtypes.common import (
     is_signed_integer_dtype,
     is_unsigned_integer_dtype,
     is_integer_dtype, is_float_dtype,
+    is_datetime64_dtype,
     is_datetime64_any_dtype,
     is_datetime64tz_dtype,
     is_timedelta64_dtype,
@@ -509,6 +510,7 @@ class Index(IndexOpsMixin, PandasObject):
 
         # _simple_new expects an ndarray
         values = getattr(values, 'values', values)
+
         return self._simple_new(values, **attributes)
 
     def _shallow_copy_with_infer(self, values=None, **kwargs):
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -506,9 +506,9 @@ class Index(IndexOpsMixin, PandasObject):
         attributes.update(kwargs)
         if not len(values) and 'dtype' not in kwargs:
             attributes['dtype'] = self.dtype
-        from pandas import DatetimeIndex
-        if isinstance(values, DatetimeIndex):
-            values = values.values
+
+        # _simple_new expects an ndarray
+        values = getattr(values, 'values', values)
         return self._simple_new(values, **attributes)
 
     def _shallow_copy_with_infer(self, values=None, **kwargs):
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -525,7 +525,6 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
                                           freq=freq, name=name)
             else:
                 index = _generate_regular_range(start, end, periods, freq)
-
         else:
 
             if tz is not None:
@@ -549,12 +548,13 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
                                               freq=freq, name=name)
                 else:
                     index = _generate_regular_range(start, end, periods, freq)
-
                 if tz is not None and getattr(index, 'tz', None) is None:
-                    index = conversion.tz_localize_to_utc(_ensure_int64(index),
-                                                          tz,
-                                                          ambiguous=ambiguous)
-                    index = index.view(_NS_DTYPE)
+                    arr = conversion.tz_localize_to_utc(_ensure_int64(index),
+                                                        tz,
+                                                        ambiguous=ambiguous)
+
+                    arr = arr.view(_NS_DTYPE)
+                    index = DatetimeIndex(arr)
 
                     # index is localized datetime64 array -> have to convert
                     # start/end as well to compare
@@ -575,7 +575,9 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
             index = index[1:]
         if not right_closed and len(index) and index[-1] == end:
             index = index[:-1]
-        index = cls._simple_new(index, name=name, freq=freq, tz=tz)
+
+        index = cls._simple_new(index.values, name=name, freq=freq, tz=tz)
+
         return index
 
     def _convert_for_op(self, value):
@@ -597,9 +599,6 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
         we require the we have a dtype compat for the values
         if we are passed a non-dtype compat, then coerce using the constructor
         """
-
-        if isinstance(values, DatetimeIndex):
-            values = values.values
 
         if getattr(values, 'dtype', None) is None:
             # empty, but with dtype compat
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -525,6 +525,7 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
                                           freq=freq, name=name)
             else:
                 index = _generate_regular_range(start, end, periods, freq)
+
         else:
 
             if tz is not None:
@@ -548,6 +549,7 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
                                               freq=freq, name=name)
                 else:
                     index = _generate_regular_range(start, end, periods, freq)
+
                 if tz is not None and getattr(index, 'tz', None) is None:
                     arr = conversion.tz_localize_to_utc(_ensure_int64(index),
                                                         tz,
@@ -607,6 +609,8 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
                 return cls(values, name=name, freq=freq, tz=tz,
                            dtype=dtype, **kwargs)
             values = np.array(values, copy=False)
+
+        assert isinstance(values, np.ndarray)
 
         if is_object_dtype(values):
             return cls(values, name=name, freq=freq, tz=tz,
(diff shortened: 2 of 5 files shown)


jreback:
this is a band aid
it shouldn’t be set in the first place like this

jreback (changes requested):
see comments

pandas/core/indexes/datetimes.py
```diff
@@ -573,12 +572,13 @@ def _generate(cls, start, end, periods, name, freq,
                                               freq=freq, name=name)
                 else:
                     index = _generate_regular_range(start, end, periods, freq)
-
                 if tz is not None and getattr(index, 'tz', None) is None:
-                    index = conversion.tz_localize_to_utc(_ensure_int64(index),
-                                                          tz,
-                                                          ambiguous=ambiguous)
-                    index = index.view(_NS_DTYPE)
+                    arr = conversion.tz_localize_to_utc(_ensure_int64(index),
```
reidy-p: ``tz_localize_to_utc`` returns an array and not a ``DatetimeIndex`` so I then convert this array to a ``DatetimeIndex`` called ``index`` so I can pass ``index.values`` to ``_simple_new`` below

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -111,3 +111,4 @@ Other
 
 - Tab completion on :class:`Index` in IPython no longer outputs deprecation warnings (:issue:`21125`)
 - Bug preventing pandas from being importable with -OO optimization (:issue:`21071`)
+- ``DatetimeIndex._data`` now returns a numpy array in all cases (:issue:`20810`)
```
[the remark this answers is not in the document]
reidy-p: Good point, thanks.

jreback:
happy to take a patch with a non-band aid fix (or can close for now)

jorisvandenbossche:
> this is a band aid
it shouldn’t be set in the first place like this

Can you be more specific than this? 
I don't think this fix is necessarily a band aid. 

Currently, the `index` object (which at the end of the `_generate` method is passed to `_simple_new`) is generated in some different ways:

- `cls._cached_range` -> returns DatetimeIndex
- `_generate_regular_range` -> returns DatetimeIndex
- passed through `conversion.tz_localize_to_utc` -> returns array
- `tools.to_datetime` -> returns DatetimeIndex

So possible fixes I see:

1) make sure that `index` is a DatetimeIndex in the end in all cases and update the final `_simple_new` call (this is what @reidy-p did)
2) make sure that each case results in a datetime64 array (this seems more work to do that conversion in each place)
3) just before the `_simple_new` call, check if `index` is a DatetimeIndex or not, and convert there to ndarray if needed
4) change `_simple_new` to convert DatetimeIndex to ndarray if passed one.

From those options, the first seems reasonable to me. I think 3) is also fine (although that is less explicit).

reidy-p:
@jorisvandenbossche thanks for that summary!

@jreback do you agree with the above comment or is this still a band-aid?

pandas/core/indexes/datetimes.py
```diff
@@ -588,7 +588,9 @@ def _generate(cls, start, end, periods, name, freq,
             index = index[1:]
         if not right_closed and len(index) and index[-1] == end:
             index = index[:-1]
-        index = cls._simple_new(index, name=name, freq=freq, tz=tz)
+
+        index = cls._simple_new(index.values, name=name, freq=freq, tz=tz)
```
jreback: i don’t like having 2 different construction paths generally: eg we always need to be an ndarray or already converted to a DTI by the time _simple_new gets called

what i would do is run all of the index tests and see what the current state is
then probably settle on an ndarray input to _simple_new and put an assertion to validate this

jreback (changes requested):

pandas/core/indexes/datetimes.py
jorisvandenbossche: Which 2 different construction paths do you mean? 
With the update above, `index` is always an index, and it's always the values that are passed to `_simple_new`. So it is ndarray input to `_simple_new`.

jreback: my point is that we need an assertion to validate this. it may be that everything is fixed, but we should actually test this.

pandas/core/indexes/datetimes.py
```diff
@@ -609,6 +611,8 @@ def _simple_new(cls, values, name=None, freq=None, tz=None,
                            dtype=dtype, **kwargs)
             values = np.array(values, copy=False)
 
+        assert isinstance(values, np.ndarray)
```
reidy-p: It was suggested above that we should have an assertion to check whether the input to ``_simple_new`` is actually always an ndarray with the new changes. It turns out that it's still not guaranteed to be an ndarray. In particular, ``_shallow_copy`` sometimes calls ``_simple_new`` with a non-ndarray input. Some of these cases are handled by the code directly above this new assert statement but one case that is not handled is the ``DatetimeIndex`` (i.e., it is not converted to an ndarray). This is why I have put code to convert a DTI in ``_shallow_copy`` to an ndarray, although I realise this may not be the correct way to handle this.

pandas/core/indexes/base.py
```diff
@@ -506,6 +506,9 @@ def _shallow_copy(self, values=None, **kwargs):
         attributes.update(kwargs)
         if not len(values) and 'dtype' not in kwargs:
             attributes['dtype'] = self.dtype
+        from pandas import DatetimeIndex
```
reidy-p: As discussed below, this code converts a DTI to ndarray before calling ``_simple_new``. All the other cases either seem to be an ndarray already or are converted to ndarray in the ``_simple_new`` function. I expect that there is probably a better way of handling this.

jreback: you can just do

with a comment
``values = getattr(values, 'values', values)``

pandas/core/indexes/datetimes.py
jreback: can you add .a message on the assert as well
can also ``assert is_integer_dtype(values)``

pandas/core/indexes/datetimes.py
```diff
@@ -607,6 +611,9 @@ def _simple_new(cls, values, name=None, freq=None, tz=None,
                            dtype=dtype, **kwargs)
             values = np.array(values, copy=False)
 
+        # values should be a numpy array
```
jreback: use the format, not commment is needed
``assert ...., "values are not an np.ndarray"``
assert the integer dtype as well

pandas/core/indexes/datetimes.py
```diff
@@ -552,10 +552,12 @@ def _generate(cls, start, end, periods, name, freq,
...
                 if tz is not None and getattr(index, 'tz', None) is None:
-                    index = conversion.tz_localize_to_utc(_ensure_int64(index),
-                                                          tz,
-                                                          ambiguous=ambiguous)
-                    index = index.view(_NS_DTYPE)
+                    arr = conversion.tz_localize_to_utc(_ensure_int64(index),
+                                                        tz,
+                                                        ambiguous=ambiguous)
+
+                    arr = arr.view(_NS_DTYPE)
```
jreback: I think can remove this arr.view(...)

jreback (changes requested):

pandas/core/indexes/datetimes.py
reidy-p: What's the intention of an ``assert is_integer_dtype(values)``? ``values`` is often an ndarray of datetime64[ns] at this stage which means this assert fails very often.

jreback: oh right, sorry the assert should be assert is_datetime64_dtype, it *always* should be this

jreback: its just better that we have certain guarantees in theses low level constructors

pandas/core/indexes/datetimes.py
```diff
@@ -2087,6 +2094,8 @@ def _generate_regular_range(start, end, periods, freq):
                              "if a 'period' is given.")
 
         data = np.arange(b, e, stride, dtype=np.int64)
+
+        # _simple_new is getting an array of int64 here
         data = DatetimeIndex._simple_new(data, None, tz=tz)
```
reidy-p: There is a new assert statement in ``_simple_new`` to check whether the input is an array of datetime64[ns]. But in this case data is an array of int64 so the assert statement fails. Is there a convenient way to rewrite this part to make data an array of datetime64[ns] before calling ``_simple_new`` so the assert works?

jreback: oh yes, ``.view(_NS_DTYPE)``

jreback:
@reidy-p if you can rebase. datetimes have been  changing a bit as getting ready for DatetimeArray cc @jbrockmendel

reidy-p:
Yeah sorry I just saw the new changes. I'll rebase.

pandas/core/indexes/datetimes.py
```diff
@@ -608,12 +610,14 @@ def _simple_new(cls, values, name=None, freq=None, tz=None,
                            dtype=dtype, **kwargs)
             values = np.array(values, copy=False)
 
-        if is_object_dtype(values):
-            return cls(values, name=name, freq=freq, tz=tz,
```
jreback: was this just never hit?

reidy-p: Yes I think it's never hit unless I'm missing something

jreback (approved):

jreback:
@reidy-p lgtm. ping on green.

jreback:
thanks @reidy-p

pandas/core/indexes/datetimes.py
```diff
@@ -608,12 +608,14 @@ def _simple_new(cls, values, name=None, freq=None, tz=None,
...
-                       dtype=dtype, **kwargs).values
-        elif not is_datetime64_dtype(values):
+        if not is_datetime64_dtype(values):
             values = _ensure_int64(values).view(_NS_DTYPE)
 
+        values = getattr(values, 'values', values)
```
jorisvandenbossche: Why is this one still needed? I thought the idea was now to ensure ndarrays are passed to `_simple_new` and not DatetimeIndexes?

reidy-p: Well-spotted! I inserted this when I was trying to investigate why some tests were failing and I meant to move it before pushing the commit but forgot. I think we can move this line to just before the call to ``_simple_new`` in this file:
https://github.com/pandas-dev/pandas/blob/1dd05ccf56b163dc51786e4f5ce76b279cc0dec3/pandas/core/arrays/datetimelike.py#L43-L53

Does this make sense? I did the same thing here:
https://github.com/pandas-dev/pandas/blob/1dd05ccf56b163dc51786e4f5ce76b279cc0dec3/pandas/core/indexes/base.py#L501-L513

PR #20913, opened 2018-05-01
DOC: update the Series.str.ismethods docstring

MrKriss:
- [ ] closes #xxxx
- [x] tests ~added~ / passed (using scripts/validate_docstrings.py)
- [ ] whatsnew entry

A result of the Pandas Sprint at PyData London 2018.

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -2401,17 +2401,16 @@ class StringMethods(NoNewAttributesMixin):
                               _shared_docs['swapcase'])
 
     _shared_docs['ismethods'] = ("""
-    Check whether all characters in each string in the Series/Index
-    are %(type)s.
+    Check whether all characters in each string are %(type)s.
 
     This is equivalent to running the Python string method
     :meth:`str.%(method)s` for each element of the Series/Index. If a string
-    has zero characters, `False` is returned for that check.
+    has zero characters, ``False`` is returned for that check.
 
     Returns
     -------
-    Series
-        Series of boolean values with the same length as the original
+    Series or Index of bool
+        Series or Index of boolean values with the same length as the original
         Series/Index.
 
     See Also
@@ -2430,9 +2429,8 @@ class StringMethods(NoNewAttributesMixin):
     --------
     **Checks for Alphabetic and Numeric Characters**
 
-    >>> s1 = pd.Series(['AB', 'C12', '42', ''])
+    >>> s1 = pd.Series(['one', 'one1', '1', ''])
 
-    >>> # All are alphabetic characters
     >>> s1.str.isalpha()
     0     True
     1    False
@@ -2440,7 +2438,6 @@ class StringMethods(NoNewAttributesMixin):
     3    False
     dtype: bool
 
-    >>> # All are numeric characters
     >>> s1.str.isnumeric()
     0    False
     1    False
@@ -2448,7 +2445,6 @@ class StringMethods(NoNewAttributesMixin):
     3    False
     dtype: bool
 
-    >>> # All are either alphabetic characters or numeric characters
     >>> s1.str.isalnum()
     0     True
     1     True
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -2401,11 +2401,137 @@ def rindex(self, sub, start=0, end=None):
...
+    0     True
+    1     True
+    2     True
+    3    False
+    dtype: bool
+
+    **Checks for Whitespace**
+
+    >>> # All characters represent whitespace
+    >>> s4 = pd.Series([' ','\\t\\r\\n ', ''])
```
datapythonista: I'd find a bit clearer to use `r'\t\r\n'` than `'\\t\\r\\n'`

pandas/core/strings.py
```diff
@@ -2401,11 +2401,137 @@ def rindex(self, sub, start=0, end=None):
...
+    Series.str.isupper : Check whether all characters are uppercase.
+    Series.str.istitle : Check whether all characters are titlecase.
+
+    Examples
+    --------
+    **Checks for Alphabetic and Numeric Characters**
+
+    >>> s1 = pd.Series(['AB', 'C12', '42', ''])
+
+    >>> # All are alphabetic characters
```
datapythonista: I find these comments before each function unnecessary.

pandas/core/strings.py
```diff
@@ -2401,11 +2401,137 @@ def rindex(self, sub, start=0, end=None):
...
+    Series.str.isspace : Check whether all characters are whitespace.
+    Series.str.islower : Check whether all characters are lowercase.
+    Series.str.isupper : Check whether all characters are uppercase.
+    Series.str.istitle : Check whether all characters are titlecase.
+
+    Examples
+    --------
+    **Checks for Alphabetic and Numeric Characters**
+
+    >>> s1 = pd.Series(['AB', 'C12', '42', ''])
```
datapythonista: We haven't done it anywhere yet, but I think it could make sense to add an index with the same values. It'd make very easy to follow the examples.

As a personal opinion, I find a bit distracting using too arbitrary examples. I'd prefer a real-world example, but as seems difficult for this case, I'd prefer something like `['one', 'one1', '1', '']`, which makes it obvious what is being shown, and don't let the users guessing why `42` and not `43`.

datapythonista (commented):
Really good job, added some comments with ideas, and about the standards followed.

pandas/core/strings.py
MrKriss: Sure, I can strip them out where it's more self-explanatory, though I'm thinking to keep the ones in the section on *More Detailed Checks for Numeric Characters* as I found this the most confusing part. i.e. why there are so many checks for numeric values and what the difference between them is. Or I could put these as text instead of comments.

datapythonista: Sounds good to me, the ones that you think the example is not clear by itself I think it's better to have the explanations as text and not as code comments.

pandas/core/strings.py
MrKriss: Just checking I understand, do you mean having the series values and index be the same so the values are displayed side by side like below?  

```
>>> s1 = pd.Series(data=['one', 'one1', '1', ''], 
                   index=['one', 'one1', '1', ''])

>>> s1.str.isalpha()
one      True
one1    False
1       False
        False
dtype: bool
```

datapythonista: It's just an idea, but yes, that's what I meant. We haven't done it in any docstring yet, afaik. But I think it makes very easy to see which value is true for each method.

pandas/core/strings.py
MrKriss: So using `r'\t\r\n'` seems to interfere with the validation script causing an error, likely because the newline and whitespace characters are still rendered as the docstring is parsed. I can correct for this if I make the whole docstring a raw string, but that seems a bit drastic, and not sure if it would have any other consequences. Would sticking with '\\t\\r\\n' be preferable over this?

pandas/core/strings.py
MrKriss: It does make comparisons easier, though I find having a blank in the last index position spoils it a bit. Could explicitly label it with 'empty string' as below?
```
>>> s1.str.isalpha()
one              True
one1            False
1               False
empty string    False
dtype: bool
```

PR #20347, opened 2018-03-14
API: str.cat will align on Series

h-vetinari:
Fixes issue #18657, fixed existing tests, added new test; all pass.

After I pushed everything and thought about it some more, I realised that one may argue about the default alignment-behavior, and whether it should be changed to `join=outer`. The behavior as implemented is compatible with the current requirement that everything be of the same length. To me, it is more intuitive that the concatenated `other` is added to the current series without enlarging it, but I can also see the argument why that restriction is unnecessary.

PS. This is my first PR, tried to follow all the rules. Sorry if I overlooked something.

Edit: Also fixes #20842

PR #20923, opened 2018-05-02
Follow-up #20347: incorporate review about _get_series_list

### LINKED ISSUE #20922: Followup to #20347 (str_cat)
cc @h-vetinari 

Some requests for simplification. Most of the complexity seems to be in the code inferring what kind of regime we're in (one or more alignable objects vs. one or more unalignable objects vs. mix). Being strict here (after deprecating) or requiring additional user input may be worthwhile from a maintenance perspective.

- [ ] https://github.com/pandas-dev/pandas/pull/20347/files/3f77b80b4ca7c8b04894b00bc08e980b62b97b97#diff-4627140a96de4a6390883e22e983f1b2
- [ ] https://github.com/pandas-dev/pandas/pull/20347/files/3f77b80b4ca7c8b04894b00bc08e980b62b97b97#diff-4627140a96de4a6390883e22e983f1b2
- [ ] https://github.com/pandas-dev/pandas/pull/20347#discussion_r185459889
- [ ] https://github.com/pandas-dev/pandas/pull/20347#discussion_r185460287
- [ ] https://github.com/pandas-dev/pandas/pull/20347#discussion_r185460641

h-vetinari:
Closes #20922

recap of #20347 :
Enabling alignment on top of the (v.0.22-)legal list of lists was relatively complex. The allowed argument types implemented in #20347 are as follows:

```
Type of (element of) "others"        | alone | within list-like
---------------------------------------------------------------------
pd.Series                            |  yes  | yes
pd.Index                             |  yes  | yes
np.ndarray (1-dim)                   |  yes  | yes
DataFrame                            |  yes  | no
np.ndarray (2-dim)                   |  yes  | no
iterators, dict-views, etc.          |  yes  | **
anything else w/ is_list_like = True |  yes  | **
```

** to be permitted, list-likes (say `nxt`) within a list-like `others` must pass (essentially)
```all(not is_list_like(x) for x in nxt)```
and not be a DF.

Open review points from #20347 from @jreback (and my responses) reproduced in review comments.

PS. Managed to have a brainfart and there's a bug in the RC docs - very sorry. Fixed in this PR.

### DIFF
--- a/doc/source/text.rst
+++ b/doc/source/text.rst
@@ -311,7 +311,7 @@ All one-dimensional list-likes can be arbitrarily combined in a list-like contai
 
     s
     u
-    s.str.cat([u, pd.Index(u.values), ['A', 'B', 'C', 'D'], map(int, u.index)], na_rep='-')
+    s.str.cat([u, pd.Index(u.values), ['A', 'B', 'C', 'D'], map(str, u.index)], na_rep='-')
 
 All elements must match in length to the calling ``Series`` (or ``Index``), except those having an index if ``join`` is not None:
 
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1943,21 +1943,21 @@ class StringMethods(NoNewAttributesMixin):
 
         Parameters
         ----------
-        input : Series, DataFrame, np.ndarray, list-like or list-like of
+        others : Series, DataFrame, np.ndarray, list-like or list-like of
             objects that are either Series, np.ndarray (1-dim) or list-like
         ignore_index : boolean, default False
-            Determines whether to forcefully align with index of the caller
+            Determines whether to forcefully align others with index of caller
 
         Returns
         -------
-        tuple : (input transformed into list of Series,
-                 Boolean whether FutureWarning should be raised)
+        tuple : (others transformed into list of Series,
+                 boolean whether FutureWarning should be raised)
         """
 
         # once str.cat defaults to alignment, this function can be simplified;
         # will not need `ignore_index` and the second boolean output anymore
 
-        from pandas import Index, Series, DataFrame, isnull
+        from pandas import Index, Series, DataFrame
 
         # self._orig is either Series or Index
         idx = self._orig if isinstance(self._orig, Index) else self._orig.index
@@ -1967,23 +1967,23 @@ class StringMethods(NoNewAttributesMixin):
                    'only objects of type Series/Index/list-like/np.ndarray)')
 
         if isinstance(others, Series):
-            fu_wrn = not others.index.equals(idx)
+            warn = not others.index.equals(idx)
             los = [Series(others.values, index=idx)
-                   if ignore_index and fu_wrn else others]
-            return (los, fu_wrn)
+                   if ignore_index and warn else others]
+            return (los, warn)
         elif isinstance(others, Index):
-            fu_wrn = not others.equals(idx)
+            warn = not others.equals(idx)
             los = [Series(others.values,
                           index=(idx if ignore_index else others))]
-            return (los, fu_wrn)
+            return (los, warn)
         elif isinstance(others, DataFrame):
-            fu_wrn = not others.index.equals(idx)
-            if ignore_index and fu_wrn:
+            warn = not others.index.equals(idx)
+            if ignore_index and warn:
                 # without copy, this could change "others"
                 # that was passed to str.cat
                 others = others.copy()
                 others.index = idx
-            return ([others[x] for x in others], fu_wrn)
+            return ([others[x] for x in others], warn)
         elif isinstance(others, np.ndarray) and others.ndim == 2:
             others = DataFrame(others, index=idx)
             return ([others[x] for x in others], False)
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: pandas/core/strings.py [hunk 1]
```diff
@@ -1967,65 +1967,58 @@ def _get_series_list(self, others, ignore_index=False):
...
-            if ignore_index and fu_wrn:
+            warn = not others.index.equals(idx)
+            if ignore_index and warn:
                 # without copy, this could change "others"
                 # that was passed to str.cat
                 others = others.copy()
                 others.index = idx
-            return ([others[x] for x in others], fu_wrn)
+            return ([others[x] for x in others], warn)
         elif isinstance(others, np.ndarray) and others.ndim == 2:
```
h-vetinari: @jreback :
> this is wrong
i don’t think we can align a ndarray at all like this
let’s can ndarray a that are > 1 dim

My response: 
The DF-constructor works as expected for a 2-dim ndarray, but I haven't checked if this is tested behaviour. (essentially, `df == DataFrame(df.values, columns=df.columns, index=df.index)`)

I would suggest not to can 2-dim ndarrays, because they are necessary to avoid alignment on the deprecation path for `join`:
> [...] To disable alignment (the behavior before v.0.23) and  silence this warning, use `.values` on any Series/Index/DataFrame in `others`. [...]

pandas/core/strings.py
```diff
@@ -1967,65 +1967,58 @@ def _get_series_list(self, others, ignore_index=False):
...
             others = DataFrame(others, index=idx)
             return ([others[x] for x in others], False)
         elif is_list_like(others):
             others = list(others)  # ensure iterators do not get read twice etc
             if all(is_list_like(x) for x in others):
                 los = []
-                fu_wrn = False
```
h-vetinari: @jreback :
> can u name this parameter just warn

Done

pandas/core/strings.py
```diff
@@ -1967,65 +1967,58 @@ def _get_series_list(self, others, ignore_index=False):
...
                         raise TypeError(err_msg)
 
-                    nxt, fwn = self._get_series_list(nxt,
+                    nxt, wnx = self._get_series_list(nxt,
                                                      ignore_index=ignore_index)
                     los = los + nxt
-                    fu_wrn = fu_wrn or fwn
-                return (los, fu_wrn)
-            # test if there is a mix of list-like and non-list-like (e.g. str)
-            elif (any(is_list_like(x) for x in others)
```
h-vetinari: @jreback 
> you can make this simpler by just checking for all is not list like (eg strings)
anything else will fail thru to the TypeError

Done

pandas/core/strings.py
```diff
@@ -1967,65 +1967,58 @@ def _get_series_list(self, others, ignore_index=False):
...
             return ([others[x] for x in others], False)
         elif is_list_like(others):
             others = list(others)  # ensure iterators do not get read twice etc
             if all(is_list_like(x) for x in others):
                 los = []
-                fu_wrn = False
+                warn = False
                 while others:
-                    nxt = others.pop(0)  # list-like as per check above
-                    # safety for iterators and other non-persistent list-likes
```
h-vetinari: @jreback :
> this whole section needs some work it’s way too hard to read and follow

pandas/core/strings.py
```diff
@@ -1967,65 +1967,58 @@ def _get_series_list(self, others, ignore_index=False):
...
                     no_deep = ((isinstance(nxt, np.ndarray) and nxt.ndim == 1)
                                or isinstance(nxt, (Series, Index)))
-                    # Nested list-likes are forbidden - elements of nxt must be
-                    # strings/NaN/None. Need to robustify NaN-check against
-                    # x in nxt being list-like (otherwise ambiguous boolean)
+                    # Nested list-likes are forbidden:
+                    # -> elements of nxt must not be list-like
                     is_legal = ((no_deep and nxt.dtype == object)
-                                or all((isinstance(x, compat.string_types)
-                                        or (not is_list_like(x) and isnull(x))
```
h-vetinari: @jreback :
>isnull already checks for None
only 1d objects are valid here (or all scalars)
do this check up front

I simplified the check, but I can't put it any further upfront (unless I remove the fast track). Need to convert iterators first, and the checking for fast-track for Series/Index etc. also needs to be done before

h-vetinari (commented):
Responding to review of @jreback

h-vetinari:
@TomAugspurger from #20922 
> Some requests for simplification. Most of the complexity seems to be in the code inferring what kind of regime we're in (one or more alignable objects vs. one or more unalignable objects vs. mix). Being strict here (after deprecating) or requiring additional user input may be worthwhile from a maintenance perspective.

What do you mean by being strict. Could you maybe edit the table from above to reflect what you think should be allowed?

h-vetinari:
@TomAugspurger 
Since I saw you're doing an RC2 due to #20924, would you consider merging this PR (or a separate one) for fixing the embarassing (for me) doc-error?

TomAugspurger:
Nope, just doing build stuff right now.

pandas/core/strings.py
jreback: can you add a test for this, basically see what happens when I have a non-default index with the Series (e.g. 2,3, 4) or something and it gets aligned with the 0,1,2 of the ndarray-converted-to-DataFrame. It will 'work' but is wrong.

jreback (changes requested):

jreback:
other than my comment above the cleanup looks good.

pandas/core/strings.py
h-vetinari: I added a general comment before all the cases for `others` because it seems this point wasn't clear:
```
# Generally speaking, all objects without an index inherit the index
# `idx` of the calling Series/Index - i.e. must have matching length.
# Objects with an index (i.d. Series/Index/DataFrame) keep their own
# index, *unless* ignore_index is set to True.
```

So, the case you mention does not happen, because an ndarray is always automatically align with the calling Series (of course, the lengths must match for this to work).

There are several tests with different indexes, both with `join is None` (so successful but raising a warning), as well as with the different `join`-types.

h-vetinari (commented):
Thanks for the feedback. I think the point you mention is a misunderstanding, but I pushed another commit to make it clearer what's happening.

doc/source/text.rst
```diff
@@ -311,7 +311,7 @@ All one-dimensional list-likes can be arbitrarily combined in a list-like contai
```
h-vetinari: Another point of @TomAugspurger in #20347, responding to the line
```All one-dimensional list-likes can be arbitrarily combined in a list-like container (including iterators, ``dict``-views, etc.):```
(which is just above this line pointer):

@TomAugspurger 
> question: what happens with a dict? Do we use the keys, or does it transform to a series and get aligned?

Response 1:
> I was talking about `d.keys()` or `d.values()`. For passing a `dict` directly, currently the keys would get read (as that's what `x in d` would return).

Response 2:
> I could of course add another safety that maps `d` to `d.values()`.
However, I tend to think that this could be left to the python-skills of the user as well -- a dictionary is not "list-like" from the POV of normal python usage (whereas its keys and values are, see `d = dict(zip(keys, values))`).

**Review by h-vetinari — COMMENTED** [review]
Before we close this, maybe this question from the last PR merits some further discussion. @TomAugspurger @jreback @jorisvandenbossche

File: doc/source/text.rst (hunk 6 above)
Comment (jreback) [reply doc/source/text.rst]: a dict is treated consistently here how we handle list-likes else where, we effectively call ``list`` on it

File: pandas/core/strings.py (hunk 1 above)
Comment (jreback) [reply pandas/core/strings.py]: i c, ok then

**Review by jreback — APPROVED** [review]

**Comment by h-vetinari** [comment]
> a dict is treated consistently here how we handle list-likes else where, we effectively call list on it

In that case, I think this PR is good to go.

**Comment by h-vetinari** [comment]
@jreback @TomAugspurger @jorisvandenbossche 
re: treatment of dicts:
FWIW, `concat` does a very similar kind of operation to `str.cat`, and does treat dicts differently (than calling list on them):
```
d = {'a' : pd.Series([0, 1]), 'b': pd.Series(['c', 'd'])}
pd.concat(d, axis=1)
#    a  b
# 0  0  c
# 1  1  d
pd.concat(d.values(), axis=1)
#    0  1
# 0  0  c
# 1  1  d
pd.concat(list(d), axis=1)
# TypeError
```

OTOH, I'm honestly fine to leave this PR as it is. Special treatment for dicts should be a separate issue, IMHO.

jreback:
thanks @h-vetinari good followup!

as far as dicts go, certainly welcome clarification if needed (PR / issue all ok), or can just wait and see if that usecase is seen in real life

h-vetinari:
@jreback 
Glad this PR is done, TBH. Thanks for all the reviews and feedback!

PR #20924, opened 2018-05-02
BLD: Use relative path in setup.py for pxd includes

TomAugspurger:
This caused failures on the conda-forge build

xref https://github.com/conda-forge/pandas-feedstock/pull/41


TomAugspurger:
I'll have to tag an rc2 for this. For whatever reason, the error only showed up on the windows worker, so I didn't catch it when running locally.

PR #20947, opened 2018-05-04
Allow drop bins when using the cut function

FANGOD:
- [ ] if cut(x, bins=[0, 1, 2, 3, 4, 5, 6, 6, 8, 8, 10], labels=False, retbins=True, right=False) will raise ValueError: You can drop duplicate edges by setting the 'duplicates' kwarg, so add 'duplicates' parameters to the cut function.


jreback (changes requested):
ok would need:

-  a doc-strings update
-  tests, see how testing is done for ``qcut``

pandas/core/reshape/tile.py
```diff
@@ -65,6 +65,8 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
         The precision at which to store and display the bins labels.
     include_lowest : bool, default False
         Whether the first interval should be left-inclusive or not.
+    duplicates : {default 'raise', 'drop'}, optional
+        If bin edges are not unique, raise ValueError or drop non-uniques.
```
jreback: add a versionadded tag

add a Raises part of the doc-string

jreback (changes requested):
the doc-string is fine, but we need an actual test (checking the duplicates kw in test_tile.py), follow the pattern of how qcut is tested

also need a whatsnew entry

FANGOD:
I don't know what to do.

pandas/core/reshape/tile.py
```diff
@@ -65,6 +65,12 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
         The precision at which to store and display the bins labels.
     include_lowest : bool, default False
         Whether the first interval should be left-inclusive or not.
+    duplicates : {default 'raise', 'drop'}, optional
+        If bin edges are not unique, raise ValueError("Bin edges must be
```
jreback: you dont need to show the error message here.

jreback: instead add a Returns section in the doc-string, showing that a ValueError can be raised

pandas/core/reshape/tile.py
```diff
@@ -144,6 +150,32 @@ def cut(x, bins, right=True, labels=None, retbins=False, precision=3,
...
+    >>> pd.cut(s, [0, 2, 4, 6, 8, 10], labels=False, retbins=True, right=False)
+    ... # doctest: +ELLIPSIS
+    (a    0.0
+     b    1.0
+     c    2.0
+     d    3.0
+     e    4.0
+     dtype: float64, array([0, 2, 4, 6, 8]))
+
+    ``duplicates=drop`` drop non-uniques
```
jreback: make this a sentence

pandas/tests/test_tile.py
```diff
@@ -0,0 +1,25 @@
+import numpy as np
+
+from pandas import DataFrame, Series
+import pandas.util.testing as tm
+
+from pandas import cut
+
+
+class TestCut(object):
+
```
jreback: don't add a new file, put in pandas/tests/reshaping/test_tile.py

jreback (changes requested):

pandas/tests/test_tile.py
FANGOD: Sorry, I didn't find it before.

jreback:
ok linted and added a test. ping on green.

jreback (approved):

jreback:
thanks @FANGOD

FANGOD:
Finally. Good experience.

PR #20956, opened 2018-05-05
ENH: Return DatetimeIndex or TimedeltaIndex bins for q/cut when input is datelike

### LINKED ISSUE #19891: ENH: cut return bins should be Timestamp/Timedelta objects for datelike data
```
In [1]: s = pd.Series(pd.date_range('20180101', periods=3))

In [2]: s
Out[2]:
0   2018-01-01
1   2018-01-02
2   2018-01-03
dtype: datetime64[ns]

In [3]: res, bins = pd.cut(s, 2, retbins=True)

In [4]: pd.cut(s, bins)  # Bins do not "roundtrip"

ValueError: bins must be of datetime64 dtype

In [5]: pd.cut(s, [pd.Timestamp(b) for b in bins]) == res
Out[5]:
0    True
1    True
2    True
dtype: bool
```

For convenience and argument of "roundtripping",  I think `cut(..., retbins=True)` with datelike data (datetime or timedelta data) should return bins that is an array of `Timestamp`/`Timedelta` objects instead of an array of integers.

mroeschke:
- [x] closes #19891
- [x] tests added / passed
- [x] whatsnew entry


doc/source/whatsnew/v0.23.0.txt
```diff
@@ -524,6 +524,7 @@ Other Enhancements
 - Added new writer for exporting Stata dta files in version 117, ``StataWriter117``.  This format supports exporting strings with lengths up to 2,000,000 characters (:issue:`16450`)
 - :func:`to_hdf` and :func:`read_hdf` now accept an ``errors`` keyword argument to control encoding error handling (:issue:`20835`)
 - :func:`date_range` now returns a linearly spaced ``DatetimeIndex`` if ``start``, ``stop``, and ``periods`` are specified, but ``freq`` is not. (:issue:`20808`)
+- :func:`cut` and :func:`qcut` now returns a ``DatetimeIndex`` or ``TimedeltaIndex`` bins when the input is datetime or timedelta dtype respectively and ``retbins=True`` (:issue:`19891`)
```
jreback: can you add ``:class:`` for DTI / TDI

pandas/core/reshape/tile.py
```diff
@@ -332,6 +332,8 @@ def _bins_to_cuts(x, bins, right=True, labels=None,
             result = result.astype(np.float64)
             np.putmask(result, na_mask, np.nan)
 
+    bins = _convert_bin_to_datelike_type(bins, dtype)
```
jreback: instead of this just call ``Index()`` which will do all of this inference

jreback (changes requested):

mroeschke:
Simplified this a bit with the Index constructor; however, I wanted to avoid turing all bins into an Index type unless this is an API change we want to make. 

Also, I created #20964 since the Index constructor does not correctly localize values with a datetimetz dtype specified.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -550,6 +550,7 @@ Other Enhancements
 - :func:`to_hdf` and :func:`read_hdf` now accept an ``errors`` keyword argument to control encoding error handling (:issue:`20835`)
 - :func:`cut` has gained the ``duplicates='raise'|'drop'`` option to control whether to raise on duplicated edges (:issue:`20947`)
 - :func:`date_range`, :func:`timedelta_range`, and :func:`interval_range` now return a linearly spaced index if ``start``, ``stop``, and ``periods`` are specified, but ``freq`` is not. (:issue:`20808`, :issue:`20983`, :issue:`20976`)
+- :func:`cut` and :func:`qcut` now returns a :class:`DatetimeIndex` or :class:`TimedeltaIndex` bins when the input is datetime or timedelta dtype respectively and ``retbins=True`` (:issue:`19891`)
```
jreback: move to 0.24

pandas/core/reshape/tile.py
```diff
@@ -428,6 +430,29 @@ def _convert_bin_to_numeric_type(bins, dtype):
...
+    ----------
+    bins : list-like of bins
+    dtype : dtype of data
+
+    Returns
+    -------
+    bins : Array-like of bins, DatetimeIndex or TimedeltaIndex if dtype is
+           datelike
+    """
+    # Can be simplified once GH 20964 is fixed.
```
jreback: would like to fix #20964 first

jreback (changes requested):

jreback (changes requested):
@mroeschke can you rebase & move note to 0.24.0. ping and will have another look.

mroeschke:
Updated and ready for a final review @jreback

pandas/core/reshape/tile.py
```diff
@@ -364,6 +365,8 @@ def _bins_to_cuts(x, bins, right=True, labels=None,
             result = result.astype(np.float64)
             np.putmask(result, na_mask, np.nan)
 
+    bins = _convert_bin_to_datelike_type(bins, dtype)
```
jreback: is there a way to make this logic flow better, e.g. ``_convert_bins_to_numeric_type`` and ``_convert_bin_to_datelike_type`` are disjointed (e.g. they are separate), maybe combine? I just find that we are doing conversions in 2 places here

jreback (changes requested):

pandas/core/reshape/tile.py
mroeschke: Well they both serve different purposes; 
`_convert_bins_to_numeric_type ` is used at the start to prep the bins (datelike data -> numeric data) before the main array is cut.

 `_convert_bin_to_datelike_type` is used at the end to "stylize" the bins (numeric  data -> datelike data) after the array is cut. 

Plus, I like the explicit naming of these two processes. Just my 2c.

jreback: ok that's fine then.

jreback (approved):

PR #20961, opened 2018-05-05
PERF: removed coercion to int64 in Categorical.from_codes

### LINKED ISSUE #18501: Categorical.from_codes shouldn't coerce to int64
Categorical.from_codes coerces its input to an array of np.int64 unconditionally even though the Categorical constructor immediately coerces the input to some other dtype using coerce_indexer_dtype. This coercion might cause a memory usage spike when codes is large. ISTM that we can just avoid the conversion in from_codes entirely and let coerce_indexer_dtype take care of any error case.

Version: master

nlee737:
- [x] closes #18501
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -578,7 +578,7 @@ class Categorical(ExtensionArray, PandasObject):
             unordered.
         """
         try:
-            codes = np.asarray(codes, np.int64)
+            codes = np.asarray(codes)
         except (ValueError, TypeError):
             raise ValueError(
                 "codes need to be convertible to an arrays of integers")


WillAyd:
Since this is performance related - is there an ASV for this? If not can you add and post result?

fivemok:
Different solution needed; removing coercion breaks causes type errors with reduce operations.

fivemok:
@TomAugspurger apologies for the lack of details. The `from_codes` function can take arrays of non-numeric types which is coerced to `np.int64`. This breaks NumPy reduction operations like `np.max`. I think only removing coercion for arrays of numeric dtype will work. I'll reopen this PR after running the tests.

@WillAyd should the full ASV be posted or just for `pd.Categoricals`?

WillAyd:
@fivemok just for `pd.Categoricals` and the relevant method(s)

fivemok:
@WillAyd @TomAugspurger I created a new PR [here](https://github.com/pandas-dev/pandas/pull/21000) for this issue.

PR #20979, opened 2018-05-08
Parametrization of indexes/test_base #4

WillAyd:
progress towards #20812

This should be it for parametrizing tests. Plan is to look at and clean up fixtures after this


jreback (approved):
lgtm

PR #20980, opened 2018-05-08
Preserve Alignment Between Index and Values for Non-Monotonic Stack

### LINKED ISSUE #20945: Data is mismatched with labels after stack with MultiIndex columns
```python
df = pd.DataFrame([
    ['DIM', 'A', 1, 2, 3, 4],
    ['DIM', 'B', 11, 22, 33, 44],
])
df.columns = ["dim1", "dim2", 'c', 'b', 'a', 'd']
df = df.set_index(["dim1", "dim2"])
df.columns.name = 'metrics'

print(df)
metrics     c   b   a   d
dim1 dim2                
DIM  A      1   2   3   4
     B     11  22  33  44

print(df.unstack('dim2').stack('metrics'))

dim2          A   B
dim1 metrics       
DIM  a        1  11
     b        2  22
     c        3  33
     d        4  44
```
#### Problem description

After stacking data by one level of MultiIndex columns, the output is unexpected. 'metrics' labels are sorted alphabetically, but data is show as is - without being matched with labels.

#### Expected Output

Either option shown below would be fine and expected:

```python
dim2          A   B
dim1 metrics       
DIM  a        3  33
     b        2  22
     c        1  11
     d        4  44

dim2          A   B
dim1 metrics       
DIM  c        1  11
     b        2  22
     a        3  33
     d        4  44
```

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.2.final.0
python-bits: 64
OS: Darwin
OS-release: 17.5.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.21.0
pytest: 3.2.3
pip: 9.0.1
setuptools: 38.2.3
Cython: None
numpy: 1.13.3
scipy: 1.0.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: None
patsy: 0.5.0
dateutil: 2.7.2
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 2.1.1
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: 1.0b10
sqlalchemy: None
pymysql: 0.7.11.None
jinja2: 2.9.6
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

WillAyd:
- [X] closes #20945
- [X] tests added / passed
- [ ] whatsnew entry

Not overly familiar with this code so submitting for review as there's probably a better way of going about it. The root cause of the referenced issue IIUC is that the index labels of the caller are non-monotonic. `stack` essentially takes values and labels from the level that is getting pushed down into the rows with an implicit assumption that both are monotonic, hence the index/values get misaligned.

This breaks at least one other test so not ready to merge, but looking for feedback on:

 - If there's a better way to align the values with the labels in this function AND/OR
 - If we make any guarantees about the order of the labels for the level(s) being moved in this function

### DIFF
--- a/pandas/core/reshape/reshape.py
+++ b/pandas/core/reshape/reshape.py
@@ -653,7 +653,13 @@ def _stack_multi_columns(frame, level_num=-1, dropna=True):
     # time to ravel the values
     new_data = {}
     level_vals = this.columns.levels[-1]
-    level_labels = sorted(set(this.columns.labels[-1]))
+    level_labels = list()
+    for label in this.columns.labels[-1]:
+        # GH 20945 if labels are not monotonic we were mangling
+        # alignment when moving to index; ensure we preserve order
+        if label not in level_labels:
+            level_labels.append(label)
+
     level_vals_used = level_vals[level_labels]
     levsize = len(level_labels)
     drop_cols = []
--- a/pandas/tests/frame/test_reshape.py
+++ b/pandas/tests/frame/test_reshape.py
@@ -414,20 +414,23 @@ class TestDataFrameReshape(TestData):
         assert_frame_equal(df3.stack(level=['animal', 0]),
                            animal_hair_stacked, check_names=False)
 
-    def test_stack_retains_index_order(self):
+    def test_stack_retains_index_order_non_monotonic(self):
         # GH 20945
-        col_index = pd.MultiIndex.from_product([
-            ['c', 'b', 'a', 'd'],
-            ['A', 'B']
+        df = pd.DataFrame([
+            ['DIM', 'A', 1, 2, 3, 4],
+            ['DIM', 'B', 11, 22, 33, 44],
         ])
+        df.columns = ["dim1", "dim2", 'c', 'b', 'a', 'd']
+        df.columns.name = 'foo'
+        df = df.set_index(["dim1", "dim2"])
 
-        df = pd.DataFrame([[1, 11, 2, 22, 3, 33, 4, 44]], columns=col_index)
-
-        expected_mi = pd.MultiIndex.from_product([[0], ['c', 'b', 'a', 'd']])
+        expected_mi = pd.MultiIndex.from_product([['DIM'], ['c', 'b', 'a', 'd']])
+        expected_mi.names = ['dim1', 'foo']
         expected = pd.DataFrame([[1, 11], [2, 22], [3, 33], [4, 44]],
                                 index=expected_mi, columns=['A', 'B'])
+        expected.columns.name = 'dim2'
 
-        result = df.stack(level=0)
+        result = df.unstack('dim2').stack(level=0)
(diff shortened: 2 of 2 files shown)


pandas/core/reshape/reshape.py
```diff
@@ -653,7 +653,13 @@ def _convert_level_number(level_num, columns):
     # time to ravel the values
     new_data = {}
     level_vals = this.columns.levels[-1]
```
jreback: we have a sort_monotonic function in a MI to do this

jreback (changes requested):

WillAyd:
Good to know, though I must have explained the issue incorrectly. Within the call to `_stack_multi_columns` on master here's how the column index of the failing example looks:

```python
MultiIndex(levels=[['A', 'B'], ['a', 'b', 'c', 'd']],
           labels=[[0, 0, 0, 0, 1, 1, 1, 1], [2, 1, 0, 3, 2, 1, 0, 3]],
           names=['dim2', 'foo'])
```

I built another DataFrame manually which was equivalent (at least according to `tm.assert_frame_equal`) and had a column index that look as follows:

```python
MultiIndex(levels=[['A', 'B'], ['c', 'b', 'a', 'd']],
           labels=[[0, 0, 0, 0, 1, 1, 1, 1], [0, 1, 2, 3, 0, 1, 2, 3]],
           names=['dim2', 'foo'])
```

The former yielded incorrect results at the end of the stack operation but the latter was fine, even though they were coming from two frames that look exactly the same. 

I believe the problem is that when iterating over the groups, `_stack_multi_columns` gets a slice of the frames' values. With the latter, getting a slice of the first 4 values at a time keeps it aligned with the first four labels since they sequential. That does not apply to the former example hence the issue.

Is the fact that the values of the DataFrame do not align with the labels of the column index by design?

jreback:
eg.
```
In [2]: mi = pd.MultiIndex(levels=[['A', 'B'], ['c', 'b', 'a', 'd']],
   ...:            labels=[[0, 0, 0, 0, 1, 1, 1, 1], [0, 1, 2, 3, 0, 1, 2, 3]],
   ...:            names=['dim2', 'foo'])
   ...:            

In [3]: mi._sort_levels_monotonic()
Out[3]: 
MultiIndex(levels=[['A', 'B'], ['a', 'b', 'c', 'd']],
           labels=[[0, 0, 0, 0, 1, 1, 1, 1], [2, 1, 0, 3, 2, 1, 0, 3]],
           names=['dim2', 'foo'])
```

WillAyd:
solved via #21043

Issue #11704, opened 2015-11-25
ENH: implement __iter__ for window objects


jreback:
xref #11603 

might be slightly non-trivial as the current impl does NOT use an explict iterator, rather does a sliding window in cython and marginal calcs.

kawochen:
I don't know if this is too ambitious, but for custom rolling aggregation, if you can let users supply a `StateVariables` object which is initialized using the first window, and supports two methods, `evolve()` and `compute()`, then you may be able to retain the efficiency when the evolution of the state variables is governed by a simple set of equations. `__iter__` becomes a special case of this where the `StateVariables` contains the whole window, and `evolve` moves the window forward/replaces the old window with the new window, and `compute` just returns the data structure inside.

selik:
Currently, iterating over a `pandas.core.window.Rolling` object will produce bizarre behavior. For DataFrames as it looks for columns named `0`, `1`, `2`, etc. until it hits a KeyError. For Series it raises AttributeError as it looks for a "columns" attribute.

Until the `__iter__` is implemented correctly, perhaps there should be a temporary measure of simply making it non-iterable by raising NotImplementedError.

jreback:
@selik if you want to submit a PR to raise NotImplementedError would be great!

antoinecomp:
Okay, how to cope with this problem then ?

I have an object, `EWM [com=9.5,min_periods=0,adjust=True,ignore_na=False,axis=0]` that I want to add to a dataframe:` predictions_df_list['ewma']` which raised this issue

selik:
@antoinecomp Could you share a bit more of your code and the error message?

antoinecomp:
@selik Sorry for answering so late. This is the code I used:

`predictions_df['ewma'] = pd.DataFrame.ewm(predictions_df["prices"], span=60)#, freq="D")`

And this is the error message: 

```
NotImplementedError                       Traceback (most recent call last)
<ipython-input-153-2a1b97ea7cf7> in <module>
----> 1 predictions_df['ewma'] = pd.DataFrame.ewm(predictions_df["prices"], span=60)#, freq="D")

C:\ProgramData\Anaconda3\lib\site-packages\pandas\core\frame.py in __setitem__(self, key, value)
   3117         else:
   3118             # set column
-> 3119             self._set_item(key, value)
   3120 
   3121     def _setitem_slice(self, key, value):

[...]

PR #20996, opened 2018-05-09
ENH: Raise useful error when iterating a Window

selik:
Until Issue #11704 is completed, raise a NotImplementedError to provide
a more clear error message when attempting to iterate over a Rolling
or Expanding window.

### DIFF
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -966,7 +966,7 @@ Other API Changes
 - Constructing a Series from a list of length 1 no longer broadcasts this list when a longer index is specified (:issue:`19714`, :issue:`20391`).
 - :func:`DataFrame.to_dict` with ``orient='index'`` no longer casts int columns to float for a DataFrame with only int and float columns (:issue:`18580`)
 - A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousins, will now *always* be passed a ``Series``, rather than a ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
-- Window types, such as Rolling and Expanding, raise `NotImplementedError` upon iteration. This will ideally be replaced by mimicking the (key, group) iteration of GroupBy (:issue:`11704`).
+- Rolling and Expanding types raise ``NotImplementedError`` upon iteration (:issue:`11704`).
 
 .. _whatsnew_0230.deprecations:
 
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -966,6 +966,7 @@ Other API Changes
 - Constructing a Series from a list of length 1 no longer broadcasts this list when a longer index is specified (:issue:`19714`, :issue:`20391`).
 - :func:`DataFrame.to_dict` with ``orient='index'`` no longer casts int columns to float for a DataFrame with only int and float columns (:issue:`18580`)
 - A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousins, will now *always* be passed a ``Series``, rather than a ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
+- Window types, such as Rolling and Expanding, raise `NotImplementedError` upon iteration. This will ideally be replaced by mimicking the (key, group) iteration of GroupBy (:issue:`11704`).
 
 .. _whatsnew_0230.deprecations:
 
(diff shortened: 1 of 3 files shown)


pandas/core/window.py
```diff
@@ -181,6 +181,10 @@ def __unicode__(self):
         return "{klass} [{attrs}]".format(klass=self._window_type,
                                           attrs=','.join(attrs))
 
+    def __iter__(self):
+        url = 'https://github.com/pandas-dev/pandas/issues/11704'
+        raise NotImplementedError('See issue #11704 %s' % (url,))
```
WillAyd: Can you use `.format` here instead? Alternately could get away with just referencing the issue number

WillAyd (changes requested):
Thanks for the submission - can you add a test to go with this?

TomAugspurger:
We'll want a test and a note in the whatsnew.

pandas/core/window.py
selik: @WillAyd Can you point me to the file where the test should be?

selik:
@TomAugspurger @WillAyd Would the test be to confirm that a `NotImplementedError` is raised or to fail when it's raised?

selik:
I added tests for both rolling and expanding in the new functional pytest style. I wasn't sure from the contribution guide whether to continue with the class-organized style of the existing tests or to use the new style.

One issue with the functional pytest style is that parameterizing on fixtures is awkward. https://github.com/pytest-dev/pytest/issues/349

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -966,6 +966,7 @@ Other API Changes
 - Constructing a Series from a list of length 1 no longer broadcasts this list when a longer index is specified (:issue:`19714`, :issue:`20391`).
 - :func:`DataFrame.to_dict` with ``orient='index'`` no longer casts int columns to float for a DataFrame with only int and float columns (:issue:`18580`)
 - A user-defined-function that is passed to :func:`Series.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, :func:`DataFrame.rolling().aggregate() <pandas.core.window.Rolling.aggregate>`, or its expanding cousins, will now *always* be passed a ``Series``, rather than a ``np.array``; ``.apply()`` only has the ``raw`` keyword, see :ref:`here <whatsnew_0230.enhancements.window_raw>`. This is consistent with the signatures of ``.aggregate()`` across pandas (:issue:`20584`)
+- Window types, such as Rolling and Expanding, raise `NotImplementedError` upon iteration. This will ideally be replaced by mimicking the (key, group) iteration of GroupBy (:issue:`11704`).
```
WillAyd: Use double backticks around literals like NotImplementedError. Could also simply by just saying "Rolling and Expanding types will now raise a \`\`NotImplementedError\`\` upon iteration (:issue:\`11704\`)" 

You may be right on how that eventually gets implemented but I wouldn't mention it here because who knows what change it could be subject to

pandas/tests/test_window.py
```diff
@@ -46,6 +46,27 @@ def win_types_special(request):
     return request.param
 
 
+# Issue 11704: Iteration over a Window
+
+@pytest.fixture
```
WillAyd: The fixtures here are a little overkill. Just parametrize a "klass" of `pd.Series` and `pd.DataFrame` (both can be constructed from the values provided).

pandas/tests/test_window.py
```diff
@@ -46,6 +46,27 @@ def win_types_special(request):
...
+def series():
+    return pd.Series([1, 2, 3, 4])
+
+@pytest.fixture
+def frame():
+    return pd.DataFrame({'a': [1, 2, 3, 4], 'b': [10, 20, 30, 40]})
+
+@pytest.mark.parametrize('which', [series(), frame()])
+def test_rolling_iterator(which):
```
WillAyd: Move this to `TestRolling` and call it `test_iterator_raises`

pandas/tests/test_window.py
```diff
@@ -46,6 +46,27 @@ def win_types_special(request):
...
+    with pytest.raises(NotImplementedError):
+        iter(which.rolling(2))
+
+@pytest.mark.parametrize('which', [series(), frame()])
+def test_expanding_iterator(which):
```
WillAyd: Move this to `TestExpanding` and call it `test_iterator_raises`

WillAyd (changes requested):

pandas/tests/test_window.py
```diff
@@ -533,6 +512,14 @@ def test_multi_index_names(self):
         tm.assert_index_equal(result.columns, df.columns)
         assert result.index.names == [None, '1', '2']
 
+    @pytest.mark.parametrize('cls', [pd.Series, pd.DataFrame])
```
WillAyd: Can you use "klass" instead of "cls"

pandas/tests/test_window.py
```diff
@@ -533,6 +512,14 @@ def test_multi_index_names(self):
...
+    def test_iter_raises(cls):
```
WillAyd: Missing "self" as first argument

WillAyd (changes requested):
Minor edits (apply to both test cases) otherwise lgtm

pandas/tests/test_window.py
selik: I strongly prefer `cls` in this case. Having grown up in Georgia, certain misspellings have a distasteful cultural undertone for me. Also, `cls` is an equally popular standard in other Python projects, if not more popular.

WillAyd: "klass" is most used in pandas so we should stick with that as the standard for now (thought admittedly not 100%). If you feel so inclined you can definitely bring up the conversation to standard as an issue or on the mailing list and get feedback. 

Would be easier to change in one sweep that having each contributor implement their own standard

```bash
grep "parametrize.*klass" -r pandas/tests/ | wc -l
      38
grep "parametrize.*box" -r pandas/tests/ | wc -l
      12
grep "parametrize.*cls" -r pandas/tests/ | wc -l
      10
```

selik: Ok, I'll make that change separately.

WillAyd (approved):
lgtm - thanks!

TomAugspurger:
Thanks!

PR #20999, opened 2018-05-10
API: Added axis to take

### LINKED ISSUE #20932: API: expose `axis` keyword in pandas.core.algorithms.take
I haven't quite replaced cyberpandas' take with pandas' yet. I think one reason is that I need to be able to pass axis through to `pandas.core.algorithms.take`

We'll have to be careful. The default for `take_nd` is 0, while for `ndarray.take` it's None. I think we want the default to be 0.

I still have to figure out one other thing, but then hopefully things will work.

TomAugspurger:
- [x] closes #20932 
- [x] tests added / passed

One thing that's worrisome here: for N-D arrays, the `fill_value` argument is set for each element. In the case of IPArray, this is fine, since my storage NA-value is (0, 0), so I would pass `fill_value=0`. But in general, when people are using an N-D array to back a 1-D column, that won't necessarily work. But at that point, maybe we recommend people implement take on their own.

### DIFF
--- a/pandas/core/algorithms.py
+++ b/pandas/core/algorithms.py
@@ -1448,7 +1448,7 @@ def _get_take_nd_function(ndim, arr_dtype, out_dtype, axis=0, mask_info=None):
     return func
 
 
-def take(arr, indices, allow_fill=False, fill_value=None):
+def take(arr, indices, axis=0, allow_fill=False, fill_value=None):
     """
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):
Looks good to me

jreback:
lgtm. thanks @TomAugspurger

Issue #14468, opened 2016-10-21
adding static type checking with mypy


jreback:
http://blog.zulip.org/2016/10/13/static-types-in-python-oh-mypy/

might be interesting if someone is looking for a project :>

MaximilianR:
https://github.com/pandas-dev/pandas/issues/12412

jorisvandenbossche:
See also https://github.com/pandas-dev/pandas2/issues/18 (a advantage for targetting this for 2.0 is we can directly use the python 3 syntax)

brmc:
continuing the conversation from pandas-dev/pandas2#18

I agree with @shoyer. type hints should be implemented as early as possible. I've taken a couple days off work, so I can get started on annotations for py2 tomorrow. Do you have any preference on how large pull requests should be? if it were up to me, given the relative safety of this sort of modification, I would probably do a PR per sub-module. Let me know if you'd like me to do it differently. After that i can move on to stubs

jreback:
@brmc sounds great!

I think main things to add are support for ``Series``, ``DataFrame``, ``Index`` methods (``Timestamp, Timedelta, Period`` would be nice to).

you are waning to add ``.pyi`` files where the module live? (.e.g. ``pandas/core/``)

will need to add a build with ``mypy`` as well for validation.

jreback:
A concern of mine is this:

essentially you are taking a file, say ``pandas/core/series.py`` and stripping out the impl, leaving the signatures. Then we validate that this works with mypy. I really really want to be sure this stays in sync with the actual signature / impl files. So we need a tests that compares these directly (in py3 this is actually not that hard, you can iterate over the definitions and simply compare the signatures).

Have others done this?

TomAugspurger:
We'll want to push these upstream to https://github.com/python/typeshed/ once we have reasonable coverage.

TomAugspurger:
Oh, and I hadn't heard of this, but https://github.com/python/mypy/blob/master/mypy/stubgen.py may be a good starting place? I guess it operates one module at a time so

```python
mkdir out
stubgen pandas.core.series
stubgen pandas.core.frame
stubgen pandas.core.index
```

brmc:
@jreback, to be honest, i've never used stubs as a core part of a project I was working on. I've only used them to supplement underlying dependencies or to workaround bugs in my IDE, so i've never thought too deeply about what consequences they might have

jreback:
[...]

PR #21008, opened 2018-05-11
MyPy cleanup and absolute imports in pandas.core.dtypes.common

WillAyd:
Starting the conversation towards #14468

There were a couple misplaced annotations below docstrings that were causing mypy to choke when running:

```bash
mypy pandas --ignore-missing-imports
```

A starred import in pandas.core.dtypes.common was responsible for a lot of complaints from mypy. These were being intentionally suppressed by flake8 but I figure it makes sense to clean up and convert into absolute imports


pandas/core/dtypes/common.py
```diff
@@ -5,17 +5,19 @@
...
-                     PeriodDtype, PeriodDtypeType,
-                     IntervalDtype, IntervalDtypeType,
-                     ExtensionDtype, PandasExtensionDtype)
-from .generic import (ABCCategorical, ABCPeriodIndex,
-                      ABCDatetimeIndex, ABCSeries,
-                      ABCSparseArray, ABCSparseSeries, ABCCategoricalIndex,
-                      ABCIndexClass, ABCDateOffset)
-from .inference import is_string_like, is_list_like
-from .inference import *  # noqa
+from pandas.core.dtypes.dtypes import (
```
jorisvandenbossche: There are quite some other places where we use relative imports as well (but of course not consistently ..), but so why change those and not others?

WillAyd: It wasn’t the relative import that mypy was complaining about as much as the starred import. I just figured I’d convert the relative to absolute while I was touching the module

PR #21009, opened 2018-05-11
ENH: Implement linspace behavior for timedelta_range and interval_range

### LINKED ISSUE #20976: ENH: linspace behavior for timedelta_range and interval_range
Behavior similar to `np.linspace` has been implemented for the regular `date_range` function in #20846, e.g. `pd.date_range(start='2018-01-01', end='2018-01-10', periods=50)`. 

This behavior is not supported by `timedelta_range` or `interval_range`, so it seems like it should be added for consistency with `date_range`.

Note that this behavior doesn't really make sense for `period_range` since a frequency is always required.   Likewise it doesn't really make sense for `bdate_range`, as the only distinguishing feature from `date_range` is the support for custom business day logic, and the default freq being `'B'`, both of which imply a frequency.  Doubly so for `cdate_range`, as it's been deprecated in favor of `bdate_range`.

jschendel:
- [X] closes #20976
- [X] tests added / passed
- [X] whatsnew entry

Summary:
- added linspace behavior support to `timedelta_range`
- added linspace behavior support to `interval_range`
    - refactored `test_interval_range.py` to use `parametrize`; split up some tests
- added a check to `bdate_range` to ensure that `freq` is specified
    - doesn't really make sense to support linspace behavior since `bdate_range` implies a frequency
    - same underlying code as `date_range`; could get linspace behavior by overriding `freq` to `None`
- updated and cleaned docstrings for all `*_range` functions except `period_range`

### DIFF
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -358,11 +358,6 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
                 msg = 'periods must be a number, got {periods}'
                 raise TypeError(msg.format(periods=periods))
 
-        if data is None and freq is None \
-                and com._any_none(periods, start, end):
-            raise ValueError("Must provide freq argument if no data is "
-                             "supplied")
-
         # if dtype has an embedded tz, capture it
         if dtype is not None:
             try:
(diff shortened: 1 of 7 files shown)


pandas/core/indexes/datetimes.py
```diff
@@ -358,11 +358,6 @@ def __new__(cls, data=None,
                 msg = 'periods must be a number, got {periods}'
                 raise TypeError(msg.format(periods=periods))
 
-        if data is None and freq is None \
-                and com._any_none(periods, start, end):
-            raise ValueError("Must provide freq argument if no data is "
-                             "supplied")
```
jschendel: small refactor here; moved this code a few lines down so all the `if data is None:` code is in one place.

jreback:
nice patch @jschendel !

Ideally we could add a small doc-update in timeseries: http://pandas-docs.github.io/pandas-docs-travis/timeseries.html#generating-ranges-of-timestamps

and in the timedelta / interval sections if appropriate.

PR #20989, opened 2018-05-09
COMPAT: 32-bit indexing compat

jreback:
xref #19539

PR #21010, opened 2018-05-11
CLN: Revert "Return Python integers" (not needed after #20989)

toobaz:
This reverts commit 8e3d4d08501dc83d4c2302686c8342e7448c78e5.

- [x] tests passed

I had introduced the explicit casting to solve broken tests on Windows, but @jreback's #20989 is the correct way to do it, so these are now redundant.

### DIFF
--- a/pandas/_libs/index.pyx
+++ b/pandas/_libs/index.pyx
@@ -195,7 +195,7 @@ cdef class IndexEngine:
         if count > 1:
             return indexer
         if count == 1:
-            return int(found[0])
+            return found[0]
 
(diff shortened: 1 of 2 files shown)


jreback:
looks ok, but oddly failing on 2.7 / windows

toobaz:
> looks ok, but oddly failing on 2.7 / windows

Uhm, OK: I thought it was irrelevant, but given that it's not, and I have no idea why... never mind. (I'm failing to understand how the dtype of ``np.where``... changes with the platform.

Issue #20998, opened 2018-05-10
Merging on int / float columns with NaNs gives spurious warning


jorisvandenbossche:
In 0.23 we added a warning when trying to merge on int / float columns where the float values are not actually integers (https://github.com/pandas-dev/pandas/pull/18352). No objections to that, but I think we should special case for NaNs, as this is a typical case that turns an integer column into floats, but does give the warning:

```
In [8]: df1 = pd.DataFrame({'col1': [0, 1], 'key': [1, 2]})
   ...: df2 = pd.DataFrame({'col2': ['a', 'b'], 'key': [2, np.nan]})

In [9]: pd.merge(df1, df2, on='key')
/home/joris/scipy/pandas/pandas/core/reshape/merge.py:969: UserWarning: You are merging on int and float columns where the float values are not equal to their int representation
  'representation', UserWarning)
Out[9]: 
   col1  key col2
0     1    2    a
```

So I think the above should not generate the warning. 
A fix is probably to just mask for missing values in the check (PR coming).

PR #21011, opened 2018-05-11
Don't raise warning on merging int and float with nan

jorisvandenbossche:
Closes https://github.com/pandas-dev/pandas/issues/20998


jreback:
thanks @jorisvandenbossche

Issue #20395, opened 2018-03-17
PERF: df.loc is 100x slower for CategoricalIndex than for normal Index


topper-123:
EDIT: After #21369 was merged the result of ``%timeit df2.loc['b']`` has improved to  **3.8 ms**.
EDIT: After #21618 was merged the result of ``%timeit df2.loc['b']`` has improved to  **3.3 ms**.
EDIT: After #21659 was merged the result of ``%timeit df2.loc['b']`` has improved to  **1.6 ms**.

#### Code Sample

```python
>>> n = 100_000
>>> df1 = pd.DataFrame(dict(A=range(n*3)), index=list('a'*n + 'b'*n + 'c'*n))
>>> df1.index.is_monotonic_increasing
True
>>> df2 = df1.copy()
>>> df2.index = pd.CategoricalIndex(df2.index)
>>> df2.index.is_monotonic_increasing
True
>>> %timeit df1.loc['b']
125 µs ± 2.95 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
>>> %timeit df2.loc['b']
13.8 ms ± 193 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
#### Problem description

Selecting on a ``CategoricalIndex`` is 100x slower than selecting on a normal ``Index``.

I've tested this on master ( a few days old)  and on v0.22, with same result for both versions. The speed is even worse than the speed for a full columns scan:

```python
>>> df3 = df2.reset_index()
>>> %timeit df3[df3['index'] == 'b']
6.58 ms ± 25.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

A guess is that the binary search is bypassed and a full index scan is being done + some extra stuff  so it's even slower than a normal full columns scan.

#### Expected Output

The output is as expected, but the speed is very slow for ``CategoricalIndex``.

#### Output of ``pd.show_versions()``

<details>

INSTALLED VERSIONS
------------------
commit: a7a7f8c1101aed1a9d37abbbcd80f77da414f0a8
python: 3.6.3.final.0
python-bits: 32
OS: Windows
OS-release: 10
machine: AMD64
processor: Intel64 Family 6 Model 78 Stepping 3, GenuineIntel
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.22.0.dev0+870.ga7a7f8c
pytest: 3.3.1
pip: 9.0.1
setuptools: 38.2.5
[...]

Issue #9748, opened 2015-03-29
ENH: support CategoricalIndex Followups


jreback:
followup to #9741 

These are all commented in `pandas/core/test_frame/test_categorical_indexing_index`
- [ ] support slice semantics
- [x] fix index name issue on selection for non-existant label that IS in the categories
- [x] fix index name issue on assignment for non-existan label that IS in the categories
- [ ] duplicated returns  Boolean array? / test
- [ ] in repr, use `max_categories/max_seq_len` to display a truncated view

These last 2 were 'solved' by putting the following in `core/internals/reindex_indexer`
but this caused some issues as the `MultiIndex._shallow_copy` is really a view semantic (so that needs to be fixed as well)

```
# make sure that we propogate the correct meta data for this axis
# we need to do this because we are not doing a take / reindex
# which would do this automatically
new_axis = self.axes[axis]._shallow_copy(new_axis)
```

```
In [1]: In [4]: df = DataFrame({'A' : np.arange(6,dtype='int64'),
   ...:    ...:                 'B' : Series(list('aabbca')).astype('category',categories=list('cabe')) }).set_index('B')

In [2]: df
Out[2]: 
   A
B   
a  0
a  1
b  2
b  3
c  4
a  5

In [3]: df.loc['e'] = 20

In [4]: df
Out[4]: 
    A
a   0
a   1
b   2
b   3
c   4
a   5
e  20
```

TomAugspurger:
This is tangentially related. I think that the `.categories` object should return a `CategoricalIndex` instead of a regular `Index` like is does now. My need right is to be able to reindex a categorical with a `.categories`. We decided to have `.reindex(Index)` have an `Index` and `.reindex(CategoricalIndex)` have a `CategoricalIndex`. Should I open a separate issue? That should be a quick fix.

jreback:
@TomAugspurger that sounds right. I thought we discussed that in another issue (but can't find right now).

TomAugspurger:
Ok, I'll do that now since it will simplify my categorical `value_counts` code a lot.

mroeschke:
I think a lot of these have already been addressed so closing

PR #21019, opened 2018-05-13
BUG: CategoricalIndex.searchsorted doesn't return a scalar if input was scalar

fjetter:
`CategoricalIndex.searchsorted` returns the wrong shape for scalar input. Numpy arrays and all other index types return a scalar if the input is a scalar, but the `CategoricalIndex` does not

For example
```
>>> import numpy as np
>>> np.array([1, 2, 3]).searchsorted(1)
0
>>> np.array([1, 2, 3]).searchsorted([1])
array([0])
>>> import pandas as pd
>>> pd.Index([1, 2, 3]).searchsorted(1)
0
>>> pd.Index([1, 2, 3]).searchsorted([1])
array([0])
```

This issue also affects slicing on sorted/ordered categoricals, which is why I've written another test for the slicing.

- [x] tests added / passed
- [x] whatsnew entry
- [ ] example in categoricals.rst

### DIFF
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -1289,9 +1289,6 @@ Indexing
 - Bug in performing in-place operations on a ``DataFrame`` with a duplicate ``Index`` (:issue:`17105`)
 - Bug in :meth:`IntervalIndex.get_loc` and :meth:`IntervalIndex.get_indexer` when used with an :class:`IntervalIndex` containing a single interval (:issue:`17284`, :issue:`20921`)
 - Bug in ``.loc`` with a ``uint64`` indexer (:issue:`20722`)
-- Bug in :func:`CategoricalIndex.searchsorted` where the method did not return a scalar when the input values was scalar (:issue:`21019`)
-- Bug in :class:`CategoricalIndex` where slicing beyond the range of the data raised a KeyError (:issue:`21019`)
-
 
 MultiIndex
 ^^^^^^^^^^
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -1289,6 +1289,9 @@ Indexing
 - Bug in performing in-place operations on a ``DataFrame`` with a duplicate ``Index`` (:issue:`17105`)
 - Bug in :meth:`IntervalIndex.get_loc` and :meth:`IntervalIndex.get_indexer` when used with an :class:`IntervalIndex` containing a single interval (:issue:`17284`, :issue:`20921`)
 - Bug in ``.loc`` with a ``uint64`` indexer (:issue:`20722`)
+- Bug in ``CategoricalIndex.searchsorted`` where the method didn't return a scalar when the input values was scalar (:issue:`21019`)
+- Bug in ``CategoricalIndex`` where slicing beyond the range of the data raised a KeyError (:issue:`21019`)
+
 
 MultiIndex
 ^^^^^^^^^^
--- a/doc/source/whatsnew/v0.23.0.txt
+++ b/doc/source/whatsnew/v0.23.0.txt
@@ -1289,8 +1289,8 @@ Indexing
 - Bug in performing in-place operations on a ``DataFrame`` with a duplicate ``Index`` (:issue:`17105`)
 - Bug in :meth:`IntervalIndex.get_loc` and :meth:`IntervalIndex.get_indexer` when used with an :class:`IntervalIndex` containing a single interval (:issue:`17284`, :issue:`20921`)
 - Bug in ``.loc`` with a ``uint64`` indexer (:issue:`20722`)
-- Bug in ``CategoricalIndex.searchsorted`` where the method didn't return a scalar when the input values was scalar (:issue:`21019`)
-- Bug in ``CategoricalIndex`` where slicing beyond the range of the data raised a KeyError (:issue:`21019`)
+- Bug in :func:`CategoricalIndex.searchsorted` where the method did not return a scalar when the input values was scalar (:issue:`21019`)
+- Bug in :class:`CategoricalIndex` where slicing beyond the range of the data raised a KeyError (:issue:`21019`)
 
 
 MultiIndex
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -89,8 +89,7 @@ Indexing
 - Bug in :class:`IntervalIndex` constructors where creating an ``IntervalIndex`` from categorical data was not fully supported (:issue:`21243`, issue:`21253`)
 - Bug in :meth:`MultiIndex.sort_index` which was not guaranteed to sort correctly with ``level=1``; this was also causing data misalignment in particular :meth:`DataFrame.stack` operations (:issue:`20994`, :issue:`20945`, :issue:`21052`)
 - Bug in :func:`CategoricalIndex.searchsorted` where the method did not return a scalar when the input values was scalar (:issue:`21019`)
-- Bug in :class:`CategoricalIndex` where slicing beyond the range of the data raised a KeyError (:issue:`21019`)
--
+- Bug in :class:`CategoricalIndex` where slicing beyond the range of the data raised a ``KeyError`` (:issue:`21019`)
 
 I/O
 ^^^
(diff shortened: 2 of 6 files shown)


pandas/core/indexes/category.py
```diff
@@ -432,13 +432,14 @@ def get_loc(self, key, method=None):
...
 
-        >>> non_monotonic_index = p.dCategoricalIndex(list('abcb'))
+        >>> non_monotonic_index = pd.CategoricalIndex(list('abcb'))
         >>> non_monotonic_index.get_loc('b')
         array([False,  True, False,  True], dtype=bool)
         """
-        codes = self.categories.get_loc(key)
-        if (codes == -1):
-            raise KeyError(key)
+        try:
```
jreback: this is redundant, KeyError is already raised by get_loc

pandas/core/arrays/categorical.py
```diff
@@ -1341,6 +1341,8 @@ def searchsorted(self, value, side='left', sorter=None):
 
         if -1 in values_as_codes:
             raise ValueError("Value(s) to be inserted must be in categories.")
+        if is_scalar(value):
```
jreback: would rather do this in ``pandas/core/base.py/searchsorted``

use ``is_scalar`` rather than a numpy function

pandas/tests/categorical/test_analytics.py
```diff
@@ -86,9 +86,9 @@ def test_searchsorted(self):
         # Searching for single item argument, side='left' (default)
         res_cat = c1.searchsorted('apple')
         res_ser = s1.searchsorted('apple')
-        exp = np.array([2], dtype=np.intp)
-        tm.assert_numpy_array_equal(res_cat, exp)
-        tm.assert_numpy_array_equal(res_ser, exp)
+        exp = np.int64(2)
```
jreback: hmm, odd that this doesn't fail, this should be a platform indexer (``intp``)

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1259,6 +1259,9 @@ Indexing
 - Bug in performing in-place operations on a ``DataFrame`` with a duplicate ``Index`` (:issue:`17105`)
 - Bug in :meth:`IntervalIndex.get_loc` and :meth:`IntervalIndex.get_indexer` when used with an :class:`IntervalIndex` containing a single interval (:issue:`17284`, :issue:`20921`)
 - Bug in ``.loc`` with a ``uint64`` indexer (:issue:`20722`)
+- Bug in ``CategoricalIndex.searchsorted`` where the method didn't return a scalar when the input values was scalar (:issue:`21019`)
```
jreback: use the ``:func:`` syntax

didn't -> did not

jreback (changes requested):

jreback:
does this have an associated issue? (pls do a search if not)

topper-123:
Could this be a reason the slowness seen in #20395?

fjetter:
@jreback The only issue I could find which seems to be related is https://github.com/pandas-dev/pandas/issues/9748 where an open TODO is the slicing on categoricals. From what I can see slicing is still not working for integers, though.

@topper-123 I don't think these are related. You should only hit the `searchsorted` code path during range slicing, e.g. `df[1:2]`

pandas/core/arrays/categorical.py
fjetter: * the issue is rather with the helper function `_get_codes_for_values` which always returns an array. I didn't want to change it there since the way it is written right now only works for array like objects. In `base.py` we're already calling `searchsorted` directly on the numpy array, i.e. it obeys the in/output shape
* I'm using `is_scalar` here, is this wrong? Are you referring to the `np.asscalar`? I couldn't find a suitable pandas function for that (other than ~ `values[0]`)

topper-123:
@fjetter, actually that operation is/should be slicing and use searchsorted, as the index is a monotonic index, and ``'b'`` is not unique.

I'll look into that again with this angle. I won't high-jack this thread anymore.

jreback (changes requested):

fjetter:
I refactored the tests and hope the intention is a bit clearer now. Slicing of the categorical should behave similar to a ordinary index (at least if it is ordered)

fjetter:
Tests fail because of an import error of `geopandas`. I have no idea what might cause this, though.

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1289,6 +1289,9 @@ Indexing
 - Bug in performing in-place operations on a ``DataFrame`` with a duplicate ``Index`` (:issue:`17105`)
 - Bug in :meth:`IntervalIndex.get_loc` and :meth:`IntervalIndex.get_indexer` when used with an :class:`IntervalIndex` containing a single interval (:issue:`17284`, :issue:`20921`)
 - Bug in ``.loc`` with a ``uint64`` indexer (:issue:`20722`)
+- Bug in :func:`CategoricalIndex.searchsorted` where the method did not return a scalar when the input values was scalar (:issue:`21019`)
```
jreback: move to 0.23.1

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1289,6 +1289,9 @@ Indexing
...
+- Bug in :class:`CategoricalIndex` where slicing beyond the range of the data raised a KeyError (:issue:`21019`)
+
```
jreback: use double backticks on KeyError

jreback: i see you added to 0.23.1 below, ok, make these changes there and revert this one

pandas/tests/indexing/test_categorical.py
```diff
@@ -627,15 +627,79 @@ def test_reindexing(self):
                       lambda: self.df2.reindex(['a'], limit=2))
 
     def test_loc_slice(self):
-        # slicing
-        # not implemented ATM
-        # GH9748
+        # Raises KeyError since the left slice 'a' is not unique
```
jreback: can you add this issue reference here (gh-....)

pandas/tests/indexing/test_categorical.py
```diff
@@ -627,15 +627,79 @@ def test_reindexing(self):
...
+        pytest.raises(KeyError, lambda: self.df.loc["a":"b"])
+        result = self.df.loc["b":"c"]
```
jreback: this tests is the same as the 3rd case? if so do we need both?  (or if not can you move them together and comment)

pandas/tests/indexing/test_categorical.py
```diff
@@ -627,15 +627,79 @@ def test_reindexing(self):
...
+            ),
+        )
+
+        # Although the edge is not within the slice, this should fall back
+        # to searchsorted slicing since the category is known and the index
+        # is ordered. Since we're selecting a value larger/lower than the
+        # right/left edge we should get the original slice again.
+        result = ordered_df.loc["a": "d"]
+        assert_frame_equal(result, ordered_df)
+
```
jreback: can you also test the left edge as well

jreback (changes requested):
lgtm. just some comment on tests. I think it might be worthwhile to update the categorical.rst with a small example of this as well.

jreback:
pls rebase on master (ci failures are fixed)

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -88,7 +88,8 @@ Indexing
 - Bug in :meth:`MultiIndex.set_names` where error raised for a ``MultiIndex`` with ``nlevels == 1`` (:issue:`21149`)
 - Bug in :class:`IntervalIndex` constructors where creating an ``IntervalIndex`` from categorical data was not fully supported (:issue:`21243`, issue:`21253`)
 - Bug in :meth:`MultiIndex.sort_index` which was not guaranteed to sort correctly with ``level=1``; this was also causing data misalignment in particular :meth:`DataFrame.stack` operations (:issue:`20994`, :issue:`20945`, :issue:`21052`)
--
+- Bug in :func:`CategoricalIndex.searchsorted` where the method did not return a scalar when the input values was scalar (:issue:`21019`)
```
jreback: can you move to 0.23.2

jreback (changes requested):
can you rebase

topper-123:
Hey @fjetter , want to follow this to the door?

pandas/core/arrays/categorical.py
topper-123: As I found out in #21699, numpy.searchsorted doesn't like python ints, but needs numpy ints to archieve its speed.

```python
>>> n = 1_000_000
>>> c = pd.Categorical(list('a' * n + 'b' * n + 'c' * n), ordered=True)
>>> %timeit c.codes.searchsorted(1)  # python int
7 ms ± 24.7 µs per loop
>>> c.codes.dtype
int8
>>> %timeit c.codes.searchsorted(np.int8(1))
2.46 µs ± 82.4 ns per loop
```

So the scalar version should be ``values_as_codes = values_as_codes[0]`` to avoid speed loss.

jreback:
can you merge master and can see where this is

PR #21025, opened 2018-05-14
PERF: improved performance of CategoricalIndex.is_monotonic*

topper-123:
- [x] whatsnew entry

```python
>>> n = 1000000
>>> %t ci.is_monotonic_increasing
22 ms # v0.22 and master
227 ns  # this commit
```

There seem to be a few more like this, where ``CategoricalIndex`` should use ``self._engine`` but doesn't. 

@TomAugspurger?

### DIFF
--- a/asv_bench/benchmarks/categoricals.py
+++ b/asv_bench/benchmarks/categoricals.py
@@ -173,3 +173,23 @@ class Isin(object):
 
     def time_isin_categorical(self, dtype):
         self.series.isin(self.sample)
+
+
+class IsMonotonic(object):
+
+    def setup(self):
+        N = 1000
+        self.c = pd.CategoricalIndex(list('a' * N + 'b' * N + 'c' * N))
+        self.s = pd.Series(self.c)
+
+    def time_categorical_index_is_monotonic_increasing(self):
+        self.c.is_monotonic_increasing
+
+    def time_categorical_index_is_monotonic_decreasing(self):
+        self.c.is_monotonic_decreasing
+
+    def time_categorical_series_is_monotonic_increasing(self):
+        self.s.is_monotonic_increasing
+
+    def time_categorical_series_is_monotonic_decreasing(self):
+        self.s.is_monotonic_decreasing
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -29,6 +29,7 @@ Deprecations
 Performance Improvements
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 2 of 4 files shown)


jreback:
this hit the same code path; so check this

topper-123:
Not sure I follow, but these two versions do not follow the same code path, as the old version required creating a new ``Int64Index`` which is expensive.

``CategoricalIndex.is_monotonic`` is already tested in ``indexes/test_category.py::TestCategoricalIndex::test_is_monotonic``.

jreback:
can you add additional tests using strings (and not just integers) in that same test. otherwise lgtm.

jreback:
do we have sufficient asv's for this?

topper-123:
There were no asv's for this. However, if you run my code snippet above, there is a huge spike in RAM usage, when run in the old version. I've even gotten a few ``MemoryErrors``.

So my ASV is done using only ``N = 1000`` to limit memory usage. The result is here 60 microseconds (old version) vs 260 ns (new version).

Also, Series.is_monotonic* wasn't added untill 0.19. should that be put in a try/except clause, to avoid failing on older versions of pandas?

asv_bench/benchmarks/categoricals.py
```diff
@@ -169,3 +169,19 @@ def setup(self, dtype):
...
+
+
+class IsMonotonic(object):
+
+    def setup(self):
+        N = 1000
+        self.c = pd.CategoricalIndex(list('a'*N + 'b'*N + 'c'*N))
+        self.s = pd.Series(self.c)
+
+    def time_categorical_index_is_monotonic(self):
```
jreback: these shouldn't be in the same asv, you can do this with params I think

jreback (commented):
minor comment on the asv. its ok if it fails under 0.19, that's pretty far back now

doc/source/whatsnew/v0.23.0.txt
```diff
@@ -1079,6 +1079,7 @@ Performance Improvements
 - Improved performance of :func:`pandas.core.groupby.GroupBy.pct_change` (:issue:`19165`)
 - Improved performance of :func:`Series.isin` in the case of categorical dtypes (:issue:`20003`)
 - Improved performance of ``getattr(Series, attr)`` when the Series has certain index types. This manifiested in slow printing of large Series with a ``DatetimeIndex`` (:issue:`19764`)
+- Improved performance of :meth:`CategoricalIndex.is_monotonic_increasing`, :meth:`CategoricalIndex.is_monotonic_decreasing` and :meth:`CategoricalIndex.is_monotonic` (:issue:`21025`)
```
jreback: will need to be in 0.23.1 (not yet in repo, soon)

jreback (changes requested):

doc/source/whatsnew/v0.23.0.txt
topper-123: Moved to 0.23.1.

jreback (approved):

PR #18848, opened 2017-12-19
Make DatetimeIndex iterator pickleable by dill

rmihael:
Currently, dill (https://github.com/uqfoundation/dill) cannot pickle iterators over DatetimeIndex because they are generators. This simple change removes that limitation.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] whatsnew entry

PR #21027, opened 2018-05-14
BUG: Iteration over DatetimeIndex stops at chunksize (GH21012)

### LINKED ISSUE #21012: Iteration over DatetimeIndex stops at 10000
#### Code Sample

```python
import pandas as pd

num_index = pd.Index(range(20000))
hour_index = pd.date_range('2000-01-01', periods=20000, freq='H')
minute_index = pd.date_range('2000-01-01', periods=20000, freq='min')

a = 0
for i in num_index: # OK
    a += 1

b = 0
for i in hour_index: # unexpectedly
    b += 1

c = 0
for i in minute_index: # unexpectedly
    c += 1

a, b, c # -> (20000, 10000, 10000)
```
#### Problem description

Iteration over DatetimeIndex stops unexpectedly when it's evaluated 10000 times.

**NOTE** :  this happened in 0.23.0rc2, not 0.22.0

#### Expected Output
 (20000, 20000, 20000)

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 3.10.0-693.21.1.el7.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: ja_JP.UTF-8
LOCALE: ja_JP.UTF-8

pandas: 0.23.0rc2
pytest: None
pip: 10.0.1
setuptools: 39.1.0
Cython: None
numpy: 1.14.3
scipy: None
pyarrow: None
xarray: None
IPython: None
sphinx: None
patsy: None
dateutil: 2.7.2
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: None
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

kittoku:
- [x] closes #21012
- [x] tests added / passed
- [ ] whatsnew entry

avoided using `yield` and changing DatetimeIndex itself into a iterator. @cbertinato 's advice was helpful.

### DIFF
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -1365,7 +1365,8 @@ class DatetimeIndex(DatelikeOps, TimelikeOps, DatetimeIndexOpsMixin,
             converted = libts.ints_to_pydatetime(data[start_i:end_i],
                                                  tz=self.tz, freq=self.freq,
                                                  box="timestamp")
-            return iter(converted)
+            for v in converted:
+                yield v
 
     def _wrap_union_result(self, other, result):
         name = self.name if self.name == other.name else None
--- a/pandas/tests/indexes/datetimes/test_datetime.py
+++ b/pandas/tests/indexes/datetimes/test_datetime.py
@@ -153,6 +153,17 @@ class TestDatetimeIndex(object):
             assert result._repr_base == expected._repr_base
             assert result == expected
 
+    @pytest.mark.parametrize('periods', [0, 9999, 10000, 10001])
+    def test_iteration_over_chunksize(self, periods):
+        # GH21012
+
+        index = date_range('2000-01-01 00:00:00', periods=periods, freq='min')
+        num = 0
+        for stamp in index:
+            assert index[num] == stamp
+            num += 1
+        assert num == len(index)
+
     def test_misc_coverage(self):
         rng = date_range('1/1/2000', periods=5)
         result = rng.groupby(rng.day)


cbertinato:
Couple of comments:
- Think about changing the name of the iterator class to something a bit less general than `Iterator`.
- Reviewers will eventually ask, but you will need a whatsnew entry.

pandas/core/indexes/datetimes.py
```diff
@@ -172,6 +172,50 @@ def _new_DatetimeIndex(cls, d):
     return result
 
 
+class Iterator:
+    def __init__(self, index):
+        self.data = index.asi8
```
jreback: we already have a ``BaseIterator`` class in ``pandas.io.common``, it would need to be moved to ``pandas.core.common`` to use it here

jreback: but why are you creating a class for this anyhow?

jreback (changes requested):

pandas/tests/indexes/datetimes/test_datetime.py
```diff
@@ -153,6 +153,17 @@ def test_iteration_preserves_tz(self):
...
             assert result == expected
 
+    @pytest.mark.parametrize('periods', [0, 9999, 10000, 10001])
+    def test_iteration_over_chunksize(self, periods):
+        # GH21012
+
+        index = date_range('2000-01-01 00:00:00', periods=periods, freq='min')
+        num = 0
+        for i, d in enumerate(index):
+            assert index[num] == d
```
TomAugspurger: Use `i` instead of `num`?

pandas/core/indexes/datetimes.py
TomAugspurger: Yes, and I'd prefer to keep this in `DatetimeIndex.__iter__` unless it's going to be used elsewhere.

TomAugspurger (commented):
This will also need a release note (for 0.23.0)

pandas/tests/indexes/datetimes/test_datetime.py
kittoku: I wanted to check boundary values including `periods == 0`.
Without `num`, it would be a bit complicated, isn't it ?

pandas/core/indexes/datetimes.py
kittoku: If `DatetimeIndex` class has `__next__` method and `__iter__` returns itself,  some other codes see `DatetimeIndex` as a iterator rather than a sequence. This cause some failures (#21012). So I wanted to hide `__next__` somehow. 

I found `BaseIterator`. But with it, it would become complicated because we have to make a subclass or override methods of a instance of `BaseIterator` (is this correct?).

kittoku:
I'm not sure where to write a whatsnew (this is not a bug in 0.22.0).
Please tell me.

pandas/tests/indexes/datetimes/test_datetime.py
TomAugspurger: That's fine then, though you can remove the `enumerate` since `i` isn't used anywhere I think.

pandas/core/indexes/datetimes.py
TomAugspurger: I think that most of the state tracking can be done inside `DatetimeIndex.__iter__`. Unless I'm misunderstanding, there's no need or desire to add `DatetimeIndex.__next__`.

cbertinato:
Good point on the whatsnew.  I suppose it's not necessary in this case.

pandas/core/indexes/datetimes.py
kittoku: hmm... sorry, but I can't come up with the way using only `__iter__` and avoiding `yield` (#18848).

jorisvandenbossche:
Yes, whatsnew is not necessary as it was not a bug before.

IMO, this PR adds a lot of complexity, but probably necessary if we want to keep both:

- the optimization to iterate over chunks (https://github.com/pandas-dev/pandas/pull/7720)
- the removal of yield to please dill (https://github.com/pandas-dev/pandas/pull/18848)

I don't know how important either of those are (eg couldn't this be fixed in dill itself? Other pickle libraries have no problem with it?), so difficult to assess for me. 
If one of the above restrictions could be removed, we wouldn't need this added complexity of returning a custom iterator object

cbertinato:
An alternative is the original fix proposed by @kittoku which entails making `DateTimeIndex` into an iterator and adding a condition in `from_tuples` to avoid the code breaking the test.  It is a kind of hack, but it does preclude the need to use a custom iterator.

kittoku:
So there are 4 options from what I see now.

1 : `DatetimeIndex.__iter__` returns `iter(self)`
**good** : the simplest , dill can pickle **bad** : low performance

2 : `DatetimeIndex.__iter__` returns generator using `yield` (like below).
**good** : simpler  **bad** : dill can't pickle
```
    # from GH7720
    def __iter__(self):
        """
        Return an iterator over the boxed values
        Returns
        -------
        Timestamps : ndarray
        """

        # convert in chunks of 10k for efficiency
        data = self.asi8
        l = len(self)
        chunksize = 10000
        chunks = int(l / chunksize) + 1
        for i in range(chunks):
            start_i = i*chunksize
            end_i = min((i+1)*chunksize,l)
            converted = tslib.ints_to_pydatetime(data[start_i:end_i], tz=self.tz, offset=self.offset, box=True)
            for v in converted:
            yield v
```

3 : `DatetimeIndex.__iter__` returns a separate iterator (like this PR).
**good** : not need to edit other codes , dill can pickle **bad** : complex, need to make a class

4 : `DatetimeIndex.__iter__` returns itself as a iterator and edit other codes (like below).
**good** : dill can pickle **bad** : complex
suggestions in #21012 and
```
#pandas\core\indexes\multi.py
def from_tuples(cls, tuples, sortorder=None, names=None):
    ...
    if not is_list_like(tuples):
        raise TypeError('Input must be a list / sequence of tuple-likes.')
    elif is_iterator(tuples): # add not isinstance(tuples, Index) here
        tuples = list(tuples)
```
```
#pandas/io/formats/printing.py
# exchange `if` and `elif` to catch DatetimeIndex as sequence before assumed as iterator
def pprint_thing(thing, _nest_lvl=0, escape_chars=None, default_escapes=False,
    ...
    if (compat.PY3 and hasattr(thing, '__next__')) or hasattr(thing, 'next'):
        return compat.text_type(thing)
    elif (isinstance(thing, dict) and
          _nest_lvl < get_option("display.pprint_nest_depth")):
        result = _pprint_dict(thing, _nest_lvl, quote_strings=True,
                              max_seq_items=max_seq_items)
    elif (is_sequence(thing) and
          _nest_lvl < get_option("display.pprint_nest_depth")):
        result = _pprint_seq(thing, _nest_lvl, escape_chars=escape_chars,
                             quote_strings=quote_strings,
                             max_seq_items=max_seq_items)
```

I'm not in need to pickle by dill, so option 2 is best to me. I thought option 3 was the second. But now option 4 looks not bad because an idea of DatetimeIndex as iterator seems not so strange.

jreback:
sure this is a bug, but not a blocker for 0.23.0

kittoku:
It is odd of me to say that, but could we discard my PR and cancel a commit of #18848 (take option 2)?

I can't find a reason to pickle `DatetimeIndex.__iter__()` while we can pickle `DatetimeIndex`.
I want to hear others' opinions.

kittoku:
May I leave the test?

jreback:
yes test is ok

jreback (approved):

jreback:
thanks @kittoku 

yeah the original (not-released) patched wasn't doc-ced / tested, and unclear how this actually solves a problem, while breaking this.

PR #21033, opened 2018-05-14
TST: add arithmetic operators fixture

jreback:


pandas/tests/series/test_operators.py
```diff
@@ -827,16 +827,52 @@ def test_sub_datetime64_not_ns(self, box, assert_func):
...
+        # 8260, 10763
+        # datetime64 with tz
+        tz = 'US/Eastern'
+        dt1 = Series(date_range('2000-01-01 09:00:00', periods=5,
+                                tz=tz), name='foo')
+        dt2 = dt1.copy()
+        dt2.iloc[2] = np.nan
+        td1 = Series(timedelta_range('1 days 1 min', periods=5, freq='H'))
+        td2 = td1.copy()
+        td2.iloc[1] = np.nan
```
WillAyd: Is this missing a check for `dt2` and `td2`?

pandas/conftest.py
```diff
@@ -77,6 +78,24 @@ def observed(request):
...
+                             '__truediv__', '__rtruediv__',
+                             '__pow__', '__rpow__']
+if not PY3:
+    _all_arithmetic_operators.extend(['__div__', '__rdiv__'])
+
+
+@pytest.fixture(params=_all_arithmetic_operators)
+def all_arithmetic_operators(request):
+    """
+    Fixture for common arithmetic operations
```
TomAugspurger: Maybe say that these are the dunder names? Otherwise unclear whether it's the like `__div__` or `div` or `operator.div`.

pandas/tests/series/test_operators.py
jreback: fixed

pandas/conftest.py
jreback: fixed

Issue #14772, opened 2016-11-30
ENH/DOC: Timestamp/DTI to epoch time


jreback:
Converting a Timestamp/DTI to epoch time involves a little recipe. We could make this a bit simpler in a couple of ways:

a) document  the below
b) provide a ``.to_epoch(unit=)`` method on Timestamp/DTI/.dt
c) allow an enhanced astype, IOW, ``.astype('int64[s]')
d) allow ``.to_numeric(...., unit=)``
e) allow Timestamp/DTI division with a Timedelta, e.g. ``s / Timedelta('1s')`` 

```
In [19]: s = Series(pd.date_range('20160101',periods=3))

In [20]: s
Out[20]: 
0   2016-01-01
1   2016-01-02
2   2016-01-03
dtype: datetime64[ns]

In [21]: ((s-Timestamp(0)) / Timedelta('1s')).astype('i8')
Out[21]: 
0    1451606400
1    1451692800
2    1451779200
dtype: int64

# this is the inverse
In [25]: pd.to_datetime(((s-Timestamp(0)) / Timedelta('1s')).astype('i8'), unit='s')
Out[25]: 
0   2016-01-01
1   2016-01-02
2   2016-01-03
dtype: datetime64[ns]
```

cc @sinhrks @jorisvandenbossche @chris-b1

jreback:
xref #11022, https://github.com/pandas-dev/pandas/issues/6741

jreback:
This also works, but exposing internal impl, verbose and not user friendly
```
In [36]: Series(s.values.astype('datetime64[s]').astype('i8'), index=s.index)
Out[36]: 
0    1451606400
1    1451692800
2    1451779200
dtype: int64
```

jorisvandenbossche:
If we would add user facing functionality, I think I would like `to_epoch()` most (certainly not something like `int64[s]` IMO)

jreback:
I changed this to make this an enhancement for a simple ``.to_epoch()`` method on Timestamp/DTI.

jbrockmendel:
since timestamp now has timestamp method, should,we use the same name for DTI?

jreback:
yes this would be reasonable (though to be honest the ``.timestamp()`` name is not very informative.....

jorisvandenbossche:
[...]

PR #21036, opened 2018-05-14
Warn on ndarray[int] // timedelta

### LINKED ISSUE #19761: BUG: regression in Timedelta.__rfloordiv__ with integer data
From a failing example in the docs (http://pandas-docs.github.io/pandas-docs-travis/timeseries.html#from-timestamps-to-epoch):

```
In [46]: stamps = pd.date_range('2012-10-08 18:15:05', periods=4, freq='D')

In [47]: stamps.view('int64') // pd.Timedelta(1, unit='s')
<ipython-input-47-9893b9bae6cf> in <module>()
----> 1 stamps.view('int64') // pd.Timedelta(1, unit='s')

/home/joris/scipy/pandas/pandas/_libs/tslibs/timedeltas.pyx in pandas._libs.tslibs.timedeltas.Timedelta.__rfloordiv__()

TypeError: Invalid dtype int64 for __floordiv__
```

This was working correctly in 0.22, and is mentioned in the docs explicitly as a way to get unix epoch data.

The previously working example (http://pandas.pydata.org/pandas-docs/stable/timeseries.html#from-timestamps-to-epoch) looks like:

```
In [42]: stamps.view('int64') // pd.Timedelta(1, unit='s')
Out[42]: array([1349720105, 1349806505, 1349892905, 1349979305])
```

cc @jbrockmendel

TomAugspurger:
Closes #19761.

```python
In [2]: pd.DatetimeIndex(['1931', '1970', '2017']).view('i8') // pd.Timedelta(1, unit='s')
pandas-dev/bin/ipython:1: FutureWarning: Floor division between integer array and Timedelta is
deprecated. Use 'array // timedelta.value' instead.

Out[2]: array([-1230768000,           0,  1483228800])
```

Long-term, we'll recommend using `to_epoch` for the case where people are doing this to do conversion to unix epoch. But https://github.com/pandas-dev/pandas/issues/14772 has a few design issues that will take some time to discuss. I think we should just recommend `.value` for now.

### DIFF
--- a/doc/source/timeseries.rst
+++ b/doc/source/timeseries.rst
@@ -257,7 +257,7 @@ Pass ``errors='coerce'`` to convert unparseable data to ``NaT`` (not a time):
 Epoch Timestamps
 ~~~~~~~~~~~~~~~~
 
-pandas supports converting integer or float epoch times to ``Timestamp`` and 
+pandas supports converting integer or float epoch times to ``Timestamp`` and
 ``DatetimeIndex``. The default unit is nanoseconds, since that is how ``Timestamp``
 objects are stored internally. However, epochs are often stored in another ``unit``
 which can be specified. These are computed from the starting point specified by the
@@ -304,11 +304,12 @@ To invert the operation from above, namely, to convert from a ``Timestamp`` to a
    stamps = pd.date_range('2012-10-08 18:15:05', periods=4, freq='D')
    stamps
 
-We convert the ``DatetimeIndex`` to an ``int64`` array, then divide by the conversion unit.
+We subtract the epoch (midnight at January 1, 1970 UTC) and then floor divide by the
+"unit" (1 second).
 
 .. ipython:: python
 
-   stamps.view('int64') // pd.Timedelta(1, unit='s').value
+   (stamps - pd.Timestamp("1970-01-01")) // pd.Timedelta('1s')
 
 .. _timeseries.origin:
 
--- a/doc/source/timeseries.rst
+++ b/doc/source/timeseries.rst
@@ -308,7 +308,7 @@ We convert the ``DatetimeIndex`` to an ``int64`` array, then divide by the conve
 
 .. ipython:: python
 
-   stamps.view('int64') // pd.Timedelta(1, unit='s')
+   stamps.view('int64') // pd.Timedelta(1, unit='s').value
 
 .. _timeseries.origin:
 
--- a/pandas/_libs/tslibs/timedeltas.pyx
+++ b/pandas/_libs/tslibs/timedeltas.pyx
@@ -1221,6 +1221,11 @@ class Timedelta(_Timedelta):
 
     def __rmod__(self, other):
         # Naive implementation, room for optimization
+        if hasattr(other, 'dtype') and other.dtype.kind == 'i':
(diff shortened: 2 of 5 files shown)


pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -1188,6 +1190,15 @@ class Timedelta(_Timedelta):
             if other.dtype.kind == 'm':
                 # also timedelta-like
                 return _broadcast_floordiv_td64(self.value, other, _rfloordiv)
+            elif other.dtype.kind == 'i':
+                # Backwards compatibility
+                # GH-19761
+                msg = textwrap.dedent("""\
```
jorisvandenbossche: you don't need the textwrap dedent if you put the string over multiple lines with implicit line continuation and string concatentation? 

```
msg = (" ... "
       " ... ")
```

(I would personally find that a bit cleaner (also don't need to `\`), but no big deal)

jorisvandenbossche (approved):

doc/source/timeseries.rst
```diff
@@ -308,7 +308,7 @@ We convert the ``DatetimeIndex`` to an ``int64`` array, then divide by the conve
 
 .. ipython:: python
 
-   stamps.view('int64') // pd.Timedelta(1, unit='s')
+   stamps.view('int64') // pd.Timedelta(1, unit='s').value
```
jreback: I would prob change this example to

```
In [11]: i
Out[11]: DatetimeIndex(['1931-01-01', '1970-01-01', '2017-01-01'], dtype='datetime64[ns]', freq=None)

In [12]: (i-pd.Timestamp(0)) // pd.Timedelta(1, unit='s')
Out[12]: Int64Index([-1230768000, 0, 1483228800], dtype='int64')

```

jreback (approved):

doc/source/timeseries.rst
TomAugspurger: That makes sense for now.

jreback:
thanks!

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -1188,6 +1190,15 @@ class Timedelta(_Timedelta):
...
+                Floor division between integer array and Timedelta is
+                deprecated. Use 'array // timedelta.value' instead.
```
jorisvandenbossche: Since you updated the example in the docs, maybe we should reflect this as well in the deprecation message?

TomAugspurger:
The maybe tricky thing there is that not all ndarray[int] // timedelta
operations are necessarily for epoch conversions. They would be the
majority though...

jorisvandenbossche:
I don't know who else would be doing that :-) But it's true that it's not necessarily the case. I can do a PR to add that explanation in addition to the current one if you like.

TomAugspurger:
That'd be great. Tracking down some dask issues to verify we don't have any
other regressions right now, so we've got some time.

germannp:
Should users also get warned on timedelta // int, which happens with timedelta % int too? Maybe the warning should pop up everytime a number is implicitly converted to a timestamp, as the current behavior does not seem very consistent with addition for example.

PR #20472, opened 2018-03-23
DOC: update the pandas.Series.shift docstring

ZackStone:
Checklist for the pandas documentation sprint (ignore this if you are doing an unrelated PR):

- [x] PR title is "DOC: update the <your-function-or-method> docstring"
- [x] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [x] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [x] It has been proofread on language by another sprint participant

Please include the output of the validation script below between the "```" ticks:

```
################################################################################
####################### Docstring (pandas.Series.shift)  #######################
################################################################################

Shift index by desired number of periods with an optional time freq.

When freq is not passed, shift the index without realign the data.
If freq is passed (in this case, the index must be date or datetime),
the index will be increased using the periods and the freq.

Parameters
----------
periods : int
    Number of periods to move, can be positive or negative.
freq : DateOffset, timedelta, or time rule string, optional
    Increment to use from the tseries module or time rule (e.g. 'EOM').
    See Notes.
axis : int or str
    Shift direction. {0, 'index'}.

Notes
-----
If freq is specified then the index values are shifted but the data
is not realigned. That is, use freq if you would like to extend the
index when shifting and preserve the original data.

Returns
-------
shifted : Series

Examples
--------
>>> df = pd.DataFrame({'Col1': [10, 20, 30, 20, 15, 30, 45],
...                    'Col2': [13, 23, 33, 23, 18, 33, 48],
...                    'Col3': [17, 27, 37, 27, 22, 37, 52]})
>>> print(df)
   Col1  Col2  Col3
0    10    13    17
1    20    23    27
2    30    33    37
3    20    23    27
4    15    18    22
5    30    33    37
6    45    48    52

>>> df.shift(periods=3)
   Col1  Col2  Col3
0   NaN   NaN   NaN
1   NaN   NaN   NaN
2   NaN   NaN   NaN
3  10.0  13.0  17.0
4  20.0  23.0  27.0
5  30.0  33.0  37.0
6  20.0  23.0  27.0

>>> df.shift(periods=1, axis=1)
   Col1  Col2  Col3
0   NaN  10.0  13.0
1   NaN  20.0  23.0
2   NaN  30.0  33.0
3   NaN  20.0  23.0
4   NaN  15.0  18.0
5   NaN  30.0  33.0
[...]

PR #21039, opened 2018-05-14
Docstring shift

### LINKED ISSUE #20492: issue when shifting with Timedelta in a groupby
Hello the awesome Pandas team!

Consider the example below

```
data = pd.DataFrame({'mydate' : [pd.to_datetime('2016-06-06'),
                                 pd.to_datetime('2016-06-08'),
                                 pd.to_datetime('2016-06-09'),
                                 pd.to_datetime('2016-06-10'),
                                 pd.to_datetime('2016-06-12'),
                                 pd.to_datetime('2016-06-13')],
                     'myvalue' : [1, 2, 3, 4, 5, 6],
                     'group' : ['A', 'A', 'A', 'B', 'B', 'B']})

data.set_index('mydate', inplace = True)
Out[58]: 
           group  myvalue
mydate                   
2016-06-06     A        1
2016-06-08     A        2
2016-06-09     A        3
2016-06-10     B        4
2016-06-12     B        5
2016-06-13     B        6

```
Now I need to compute the difference between the current value of `myvalue` and its lagged value, where by lagged I actually mean **lagged by 1 day** (if possible).

So this returns a result, but its not what I need

```
data['delta_one'] = data.groupby('group').myvalue.transform(lambda x: x - x.shift(1))

data
Out[56]: 
           group  myvalue  delta_one
mydate                              
2016-06-06     A        1        nan
2016-06-08     A        2     1.0000
2016-06-09     A        3     1.0000
2016-06-10     B        4        nan
2016-06-12     B        5     1.0000
2016-06-13     B        6     1.0000
```

This is what I need, but it does not work

```
data['delta_two'] = data.groupby('group').myvalue.transform(lambda x: x - x.shift(1, pd.Timedelta('1 days')))

  File "C:\Users\john\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\core\internals.py", line 120, in __init__
    len(self.mgr_locs)))

ValueError: Wrong number of items passed 4, placement implies 3
```

Any ideas? Is this a bug? I think I am using the correct pandonic syntax here.

Thanks!

PyJay:
- [x] closes #20492
- [x] tests passed
- [ ] whatsnew entry


WillAyd:
I think this is missing the main purpose of that thread, namely that shift will not do any implicit inference of ordering or continuity for you. 

While the solution you provided "answers" that question, I think it would be more appropriate to document shift with / without reindexing prior to the fact to contrast the difference

jythonic:
@WillAyd  Thanks for the feedback.

I am not sure how I could document the case without re-indexing without resulting in an error. I decided to print the intermediary result in an attempt to make this clear. Do you suggest I show the column with the shift applied instead of having it hide behind the lambda operation?

WillAyd:
I don't think whatever you show should have any lambda (UDFs should typically be a "last resort").

I would suggest doing a shift with the base frame, pointing out that values from 06-06 simply get moved down to 06-08 without any regard for the fact that there is a two day gap in between those. Then contrast that by reindexing then shifting to show how to move from 06-06 to 06-07

jythonic:
Okay - will change this to remove usage of lambda.

Doing a shift with the base frame without any re-indexing _data.shift(1, pd.Timedelta('1 days')))_ moves values from 06 to 07 but I believe the re-indexing and left join is needed because of the mismatch of indexing between the original and resulting dataframe as a result of the shift.

image

pandas/core/generic.py
```diff
@@ -7807,6 +7807,66 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
         is not realigned. That is, use freq if you would like to extend the
         index when shifting and preserve the original data.
 
+        Examples
+        --------
+        Compute the difference between a column in a dataframe and
+        its shifted version
+
+        >>> data = pd.DataFrame({'mydate': [pd.to_datetime('2016-06-06'),
```
TomAugspurger: maybe use `pd.Timestamp('2016-06-06')` or `pd.to_datetime(['2016-06-06', ...])` to convey that `to_datetime` should be used for arrays.

pandas/core/generic.py
```diff
@@ -7807,6 +7807,66 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
         is not realigned. That is, use freq if you would like to extend the
         index when shifting and preserve the original data.
 
+        Examples
+        --------
+        Compute the difference between a column in a dataframe and
+        its shifted version
```
TomAugspurger: End with a `.`

pandas/core/generic.py
TomAugspurger: Move `mydate` to `index=` and remove the `set_index` below.

pandas/core/generic.py
```diff
@@ -7807,6 +7807,66 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+        ...                                 pd.to_datetime('2016-06-10'),
+        ...                                 pd.to_datetime('2016-06-12'),
+        ...                                 pd.to_datetime('2016-06-13')],
+        ...                     'myvalue': [1, 2, 3, 4, 5, 6],
+        ...                     'group': ['A', 'A', 'A', 'B', 'B', 'B']})
+
+        >>> data.set_index('mydate', inplace=True)
+        >>> data
+                        myvalue group
+            mydate
```
TomAugspurger: Is this indented too far? I thought the leftmost output should be directly below the leftmost `>`.

pandas/core/generic.py
```diff
@@ -7807,6 +7807,66 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+                   2016-06-11    NaN
+                   2016-06-12    NaN
+                   2016-06-13    1.0
+                   2016-06-14    NaN
+            Name: myvalue, dtype: float64
+
+        Merge result as a column named *delta* to the original data
+
+        >>> result.name = 'delta'
+        >>> data.reset_index().merge(
```
TomAugspurger: I think the `reset_index` and `set_index` can be avoided in 0.23+ (can join on mix of columns and index names).

pandas/core/generic.py
PyJay: It is useful to have a named index as the label (mydate) is used later in the process (for the left join). Would you prefer if I use `index= ` and then set the label using  `data.index.name =  ` ?

pandas/core/generic.py
PyJay: noted

pandas/core/generic.py
PyJay: noted

pandas/core/generic.py
PyJay: I have had a play and looks like the` reset_index` and `set_index` are necessary. reset index changes the delta (shifted values) from a series to a DataFrame and it exposes mydate as column on the original data which is needed for the left join. And the set index just sets up mydate as an index like in the original dataset.

pandas/core/generic.py
TomAugspurger: Agreed. How about 

```
index=pd.DatetimeIndex(['2016-06-08', ... '2016-06-13'], name='mydate')
```.

File: pandas/core/generic.py (hunk 4 above)
Comment (TomAugspurger) [reply pandas/core/generic.py]: https://github.com/pandas-dev/pandas/issues/21220 for that.

I think `pd.merge(data, result.to_frame(), on=['mydate', 'group'])` may work


```
In [45]: pd.merge(data, result.to_frame(), on=['group', 'mydate'])
Out[45]:
            myvalue_x group  myvalue_y
mydate
2016-06-06          1     A        NaN
2016-06-08          2     A        NaN
2016-06-09          3     A        1.0
2016-06-10          4     B        NaN
2016-06-12          5     B        NaN
2016-06-13          6     B        1.0
```

**Comment by PyJay** [comment]
@TomAugspurger I have updated the docs as per the feedback.

**Comment by WillAyd** [comment]
I still think this is missing the point. Can you show:

 - The "as is shift", calling out that the values are simply moved down (i.e. 06-06 value goes to 06-08) 
 - A reindexed version where the value from 06-06 subsequently moves to 06-07

Please remove the use of the anonymous function

**Comment by PyJay** [comment]
@WillAyd I have updated to address to your first two points. I am still unsure about removing the lambda since the only alternative I can think of is having a one line function customized to the use case to do the same. Maybe I'm missing a  better way here?

File: pandas/core/generic.py [hunk 5]
```diff
@@ -7810,6 +7810,86 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
         is not realigned. That is, use freq if you would like to extend the
         index when shifting and preserve the original data.
 
+        Examples
+        --------
+        Compute the difference between a column in a dataframe
```
WillAyd: Whenever you reference it be sure to capitalize appropriately the word `DataFrame`

WillAyd: This first line isn't necessary - can simply delete

pandas/core/generic.py
```diff
@@ -7810,6 +7810,86 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+        with grouped data, and its shifted version.
+
+        >>> data = pd.DataFrame({'myvalue': [1, 2, 3, 4, 5, 6],
```
WillAyd: Use the variable name `df` instead of `data` for consistency

pandas/core/generic.py
```diff
@@ -7810,6 +7810,86 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+        2016-06-12        5     B
+        2016-06-13        6     B
+
+        For the groups compute the difference between current `myvalue` and
+        `myvalue` shifted forward by 1 day.
+
+        If the dataframe is shifted without passing a freq argument than the
+        values simply move down
+
+        >>> data[data.group=='A'].myvalue.shift(1)
```
WillAyd: Instead of filtering to group `A` you'd be better served to work with the entire frame

pandas/core/generic.py
```diff
@@ -7810,6 +7810,86 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+        2016-06-08        2     A
+        2016-06-09        3     A
+        2016-06-10        4     B
+        2016-06-12        5     B
+        2016-06-13        6     B
+
+        For the groups compute the difference between current `myvalue` and
+        `myvalue` shifted forward by 1 day.
+
+        If the dataframe is shifted without passing a freq argument than the
```
WillAyd: Simple typo - "then" instead of "than" here

pandas/core/generic.py
```diff
@@ -7810,6 +7810,86 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+        values simply move down
+
+        >>> data[data.group=='A'].myvalue.shift(1)
+        mydate
+        2016-06-06    NaN
```
WillAyd: Others might have a differing opinion but I find these examples rather confusing as you really need to think through how the data is indexed.

I mentioned it before but I think the cleanest approach would be the `reindex` / `fill` set of operations I posted in the original PR - is there any reason why we can't use that here instead of the UDF? I think it more clearly explains the situation and it will certainly scale better on larger datasets, hence why I'd rather we suggest that type of usage in the documentation.

WillAyd (changes requested):

pandas/core/generic.py
PyJay: Which first line? This one?  "Compute the difference between a column in a dataframe
        with grouped data, and its shifted version."

pandas/core/generic.py
PyJay: Agreed. Your approach seems more explicit. You have suggested
>>> dt_rng = pd.date_range(data.index.min(), data.index.max())
>>> data = data.reindex(rng)
>>> data['group'] = data['group'].ffill()
>>> data.groupby('group')['myvalue'].transform(lambda x: x-x.shift())

I think I can avoid the UDF by doing

>>> df['myvalue'] - df.groupby('group')['myvalue'].shift(1)

I get the same answer, just wanted to confirm that it's the correct thing to do here?
Thanks.

WillAyd: That’s correct

pandas/core/generic.py
TomAugspurger: Agreed. The group stuff seems to be a distraction from `shift` at this point.

pandas/core/generic.py
TomAugspurger: I believe @WillAyd was referring to the "Compute the difference..." line.

PyJay:
@WillAyd @TomAugspurger Changes made as per feedback.

pandas/core/generic.py
```diff
@@ -8029,6 +8029,91 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+        2016-06-07    NaN
+        2016-06-08    NaN
+        2016-06-09    1.0
+        2016-06-10    NaN
+        2016-06-11    NaN
+        2016-06-12    NaN
+        2016-06-13    1.0
+        Freq: D, Name: myvalue, dtype: float64
+
+        Concatenate result as a column named `delta` to the original data
```
WillAyd: I appreciate the thoroughness you are aiming for here, but let's get rid of this section to the example. `concat` will have a separate docstring with examples to teach that piece to users should they want to know

pandas/core/generic.py
```diff
@@ -8029,6 +8029,91 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
         is not realigned. That is, use freq if you would like to extend the
         index when shifting and preserve the original data.
 
+        Examples
+        --------
+
+        >>> df = pd.DataFrame({'group': ['A', 'A', 'A', 'B', 'B', 'B'],
```
WillAyd: This probably isn't enforceable and maybe it's just a personal preference but I think using `key` and `val` would be preferable to `group` and `myvalue`. I've seen the former used in quite a few more places so for consistency would be better to use those names.

@datapythonista maybe something to think about

pandas/core/generic.py
```diff
@@ -8029,6 +8029,91 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+                    group	myvalue
+        mydate
+        2016-06-06	A	1
+        2016-06-08	A	2
+        2016-06-09	A	3
+        2016-06-10	B	4
+        2016-06-12	B	5
+        2016-06-13	B	6
+
+        For the groups compute the difference between current `myvalue` and
```
WillAyd: Maybe I'm missing the point but I don't think you need this sentence

pandas/core/generic.py
```diff
@@ -8029,6 +8029,91 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+        `myvalue` shifted forward by 1 day.
+
+        If `myvalue` is shifted then the values will simply move down.
```
WillAyd: Good start but I think the point to stress here is that "move down" makes no consideration of the dates being in order. So maybe to emphasize better can append something like "...move down one row, regardless of the chronology of the dates"

pandas/core/generic.py
```diff
@@ -8029,6 +8029,91 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+        mydate
+        2016-06-06    NaN
+        2016-06-08    1.0
+        2016-06-09    2.0
+        2016-06-10    3.0
+        2016-06-12    4.0
+        2016-06-13    5.0
+        Name: myvalue, dtype: float64
+
+        We only want to shift `myvalue` forward by one day before computing
```
WillAyd: Just for clearer wording I'd say "If instead you wanted to shift values forward by a day you can do this by reindexing first and filling `key`"

pandas/core/generic.py
```diff
@@ -8029,6 +8029,91 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+        2016-06-09     A      3.0
+        2016-06-10     B      4.0
+        2016-06-11     B      NaN
+        2016-06-12     B      5.0
+        2016-06-13     B      6.0
+
+        After considering the grouping we can calculate the difference
+        as follows
+
+        >>> result = df['myvalue'] - df.groupby('group')['myvalue'].shift(1)
```
WillAyd: Not sure what you are aiming for with the subtraction but I think it's clearer if you just do `df.groupby('key').shift()` here and show the result

WillAyd (changes requested):
Almost there!

PyJay:
Ahh thanks for the feedback @WillAyd I realise this has dragged on a bit! Will hopefully be able to finalise this weekend. @datapythonista were you thinking of taking this work over?

pandas/core/generic.py
```diff
@@ -8029,6 +8029,91 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
...
+        2016-06-08	 A	   2.0	     NaN
+        2016-06-09	 A	   3.0	     1.0
+        2016-06-10	 B	   4.0	     NaN
+        2016-06-11	 B	   NaN	     NaN
+        2016-06-12	 B	   5.0	     NaN
+        2016-06-13	 B	   6.0	     1.0
+
         Returns
         -------
         shifted : %(klass)s
```
datapythonista: Can you move `Return` before the examples?

pandas/core/generic.py
```diff
@@ -8012,12 +8012,12 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
                           errors=errors)
 
     _shared_docs['shift'] = ("""
-        Shift index by desired number of periods with an optional time freq
+        Shift index by desired number of periods with an optional time freq.
 
         Parameters
         ----------
         periods : int
```
datapythonista: Can you add the default `1`

pandas/core/generic.py
datapythonista: In general I think it makes things clearer if the example is with real world data. In this case, if we just show basic examples of `shift` (and we avoid using `ffill`, `reindexing`, `groupby` and `concat`), I don't think we need the column `group`.

I think the examples are quite good for the cookbook, or the time series documentation, but for the docstring, I think we should show the basic stuff. In this case what I'd do:
- Create a simple dataframe (one column and 4 rows for example)
- Show `.shift()` with default arguments
- Show how `period` can change the shifted periods
- Show how `freq` can be used

And I don't know exactly what `axis` does in this method, it could be shown too.

For all the rest of the stuff, feel free to contribute it to the other documents.

datapythonista:
Duplicate of #20472

PR #21046, opened 2018-05-15
DOC: Fix find_undoc_args script to use getfullargspec()

shanamatthews:
Use [getfullargspec](https://docs.python.org/3/library/inspect.html#inspect.getfullargspec) instead of [getargspec](https://docs.python.org/3/library/inspect.html#inspect.getargspec) to fix the find_undoc_args.py script


WillAyd:
Thanks for the PR. Just a few questions:

 - Is this fixing a current issue?
 - Is this backwards compatible (we still support Py27)?
 - Can you add tests?

shanamatthews:
Oops, didn't realize these scripts were supposed to support Py27 as well ☺️

PR #21066, opened 2018-05-15
Prevent Unlimited Agg Recursion with Duplicate Col Names

### LINKED ISSUE #21063: segfault in DataFrame.apply with duplicate column names and multiple aggfuncs
```python
In [1]: import pandas as pd

In [2]: df = pd.DataFrame([['a', 'b'], ['c', 'd']], columns=['a', 'a'])

In [3]: df.agg(['min'])
Fatal Python error: Cannot recover from stack overflow.
```

This was present at least in 0.22.0. Haven't looked further.

WillAyd:
- [X] closes #21063
- [X] tests added / passed
- [x] whatsnew entry

The `_gotitem` implementation for `DataFrame` seems a little strange so there may be a more comprehensive approach, but this should prevent the issue for the time being

Didn't add whatsnew yet since none existed for 0.23.1. Happy to add if we are OK with this fix

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -5731,7 +5731,12 @@ class DataFrame(NDFrame):
     # ----------------------------------------------------------------------
     # Function application
 
-    def _gotitem(self, key, ndim, subset=None):
+    def _gotitem(self,
+                 key,           # type: Union[str, List[str]]
+                 ndim,          # type: int
+                 subset=None    # type: Union[Series, DataFrame, None]
+                 ):
+        # type: (...) -> Union[Series, DataFrame]
         """
         sub-classes to define
         return a sliced object
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -5746,6 +5746,8 @@ class DataFrame(NDFrame):
         """
         if subset is None:
             subset = self
+        elif subset.ndim == 1:
+            subset = self._constructor(subset)
 
         # TODO: _shallow_copy(subset)?
         return subset[key]
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -5748,7 +5748,7 @@ class DataFrame(NDFrame):
             subset = self
 
         # TODO: _shallow_copy(subset)?
-        return self[key]
+        return subset[key]
 
     _agg_doc = dedent("""
     The aggregation operations are always performed over an axis, either the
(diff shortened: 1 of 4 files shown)


pandas/core/frame.py
```diff
@@ -5731,7 +5731,12 @@ def diff(self, periods=1, axis=0):
     # ----------------------------------------------------------------------
     # Function application
 
-    def _gotitem(self, key, ndim, subset=None):
+    def _gotitem(self,
```
WillAyd: I realize we don't have an overall strategy for annotations just yet but I had to think through this as I was debugging anyway, so figured I'd put here explicitly for when we turn this on

jreback: ok!

pandas/core/frame.py
```diff
@@ -5746,9 +5751,11 @@ def _gotitem(self, key, ndim, subset=None):
         """
         if subset is None:
             subset = self
+        elif subset.ndim == 1:
```
jreback: is this line actually hit in tests? return self._constructor here doesn't make sense

jreback (changes requested):

pandas/core/frame.py
WillAyd: Yes this is actually hit by the test case and code that was in place. A Series is a valid value for the subset parameter so I’m forcing it to a DataFrame or else the subsequent slice would fail. All for a better way if you think there’s one

TomAugspurger:
FYI, the whasnew is available now.

pandas/core/frame.py
jreback: what operation actually hits this code?

WillAyd: This line (which was changed to prevent the unlimited calls):

https://github.com/WillAyd/pandas/blob/4a24f734047d387ce242c36dba16eb69388a3ca1/pandas/core/base.py#L595

This would pass a Series before (unless column names were duplicated), though it never raised an error because of the code in `_got_item`. It would accept the Series and even use it in a condition, but would always just return a subset of itself...

Definitely convoluted - I think it was inadvertent before the way `_got_item` was implemented by DataFrame

jreback: no the issue is why you are wrapping it with ``self._constructor`` which is a ``DataFrame`` here, then you are selecting it out again, just do

``return subset`` i think is enough?

WillAyd: Hmm you're probably right with that - I suppose could just return immediately if `ndim == 1`. Will try locally and push if it works

jreback (approved):

TomAugspurger:
Looks good. I'll keep an eye on the appveyor test in case it's relevant.

PR #21082, opened 2018-05-16
BUG: to_records() fails for MultiIndex DF (#21064)

### LINKED ISSUE #21064: to_records() fails for empty dataframes after groupby
Calling `to_records()` on an empty dataframe after a group-by results in  a `ValueError`:

```python
In [1]: import pandas as pd
In [2]: mydf = pd.DataFrame({'A':[10,-10], 'B':[400,400], 'C':["bla","bla"]})
In [3]: new_pd2 = pd.DataFrame(mydf.groupby(['B','C'])['A'].agg(lambda x: x.sum()))
In [4]: new_pd3 = new_pd2[new_pd2['A'] != 0 ]
In [5]: new_pd3.to_records()
<ipython-input-5-ce958ec44125> in <module>()
----> 1 new_pd3.to_records()

/.../lib/python2.7/site-packages/pandas-0.22.0+ahl1-py2.7-linux-x86_64.egg/pandas/core/frame.pyc in to_records(self, index, convert_datetime64)
   1233         return np.rec.fromarrays(
   1234             arrays,
-> 1235             dtype={'names': names, 'formats': formats}
   1236         )
   1237 

/.../.eggs/numpy-1.14.2-py2.7-linux-x86_64.egg/numpy/core/records.pyc in fromarrays(arrayList, dtype, shape, formats, names, titles, aligned, byteorder)
    605 
    606     if dtype is not None:
--> 607         descr = sb.dtype(dtype)
    608         _names = descr.names
    609     else:

ValueError: 'names', 'formats', 'offsets', and 'titles' dict entries must have the same length

```
#### Problem description

Calling `to_records()` on an empty dataframe after a group-by results in  a `ValueError`. I expect this to not throw and return the records.

```
pandas==0.22.0
numpy==1.14.2
```

_The same code above without column 'C' (just using 2 columns) works fine._

The problem is here:
https://github.com/pandas-dev/pandas/blob/master/pandas/core/frame.py#L1419

```
return np.rec.fromarrays(
            arrays,
            dtype={'names': names, 'formats': formats}
        )
```

where `names` and `formats` have different lengths and make `numpy` upset.

This seems to work in pandas==0.16.2 and returns the following

```
In [5]: new_pd3.to_records()
Out[5]: 
rec.array([], 
      dtype=[('B', '<i8')])
```

fersarr:
- [x] closes #21064
- [x] tests added / passed
- [x] whatsnew entry

This fixes the bug that prevents using `to_records` on an empty dataframe that has a MultiIndex

`self.index.values` returns an empty array (no tuples) for an empty MultiIndex DF and that would be used for `ix_values`. Instead, `ix_values should have an array of empty arrays, each one with the correct dtype according to the MultiIndex level/column.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -1392,7 +1392,13 @@ class DataFrame(NDFrame):
             else:
                 if isinstance(self.index, MultiIndex):
                     # array of tuples to numpy cols. copy copy copy
-                    ix_vals = lmap(np.array, zip(*self.index.values))
+                    tuples = self.index.values
+                    if len(tuples):
+                        ix_vals = lmap(np.array, zip(*tuples))
+                    else:
+                        # empty MultiIndex DF
+                        ix_vals = [np.array([], dtype=lev.dtype)
+                                   for lev in self.index.levels]
                 else:
                     ix_vals = [self.index.values]
 
--- a/pandas/tests/frame/test_convert_to.py
+++ b/pandas/tests/frame/test_convert_to.py
@@ -328,3 +328,35 @@ class TestDataFrameConvertTo(TestData):
         result = DataFrame.from_dict(result, orient='index')[cols]
         expected = DataFrame.from_dict(expected, orient='index')[cols]
         tm.assert_frame_equal(result, expected)
+
+    def test_to_records_with_multiindex(self):
+        size = 4
+        idx1 = [u'a', u'a', u'b', u'b']
+        idx2 = [u'x', u'y', u'x', u'y']
+        tup = zip(idx1, idx2)
+        index = MultiIndex.from_tuples(tup)
+        random_data = np.random.randn(size, size)
+        df = DataFrame(random_data, index=index)
+
+        result = df.to_records(index=True)
+
+        col_arrays = [idx1, idx2] + [col for col in random_data.T]
+        expected = np.rec.fromarrays(
+            col_arrays, dtype=np.dtype([('level_0', '<U1'), ('level_1', '<U1'),
+                                        ('0', '<f8'), ('1', '<f8'),
+                                        ('2', '<f8'), ('3', '<f8')]))
+
+        tm.assert_numpy_array_equal(result, expected)
+
+    def test_to_records_with_empty_multiindex(self):
+        # GH 21064
+        multi = MultiIndex([['a'], ['b']], labels=[[], []])
+        df = DataFrame(columns=['A'], index=multi)
+
+        expected = np.rec.fromarrays([[], [], []],
+                                     dtype=np.dtype([('level_0', 'O'),
+                                                     ('level_1', 'O'),
+                                                     ('A', 'O')]))
+
+        result = df.to_records(index=True)
+        tm.assert_numpy_array_equal(result, expected)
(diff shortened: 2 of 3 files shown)


pandas/tests/frame/test_convert_to.py
```diff
@@ -328,3 +328,26 @@ def test_to_dict_index_dtypes(self, into, expected):
         result = DataFrame.from_dict(result, orient='index')[cols]
         expected = DataFrame.from_dict(expected, orient='index')[cols]
         tm.assert_frame_equal(result, expected)
+
+    def test_to_records_with_multiindex(self):
```
fersarr: I guess I could remove this test `test_to_records_with_multiindex` since `test_to_records_index_name` is already there on line 130 and deals with multi index?

chris-b1:
Seems reasonable to me, @toobaz, want to have a look?

pandas/tests/frame/test_convert_to.py
```diff
@@ -328,3 +328,26 @@ def test_to_dict_index_dtypes(self, into, expected):
...
+        size = 4
+        tup = zip(*[['a', 'a', 'b', 'b'], ['x', 'y', 'x', 'y']])
+        index = MultiIndex.from_tuples(tup)
+        df = DataFrame(np.random.randn(size, size), index=index)
+
+        records = df.to_records(index=True)
+        assert len(records) == size
```
jreback: construct a recarray and compare using ``assert_numpy_array_equal``

pandas/tests/frame/test_convert_to.py
```diff
@@ -328,3 +328,26 @@ def test_to_dict_index_dtypes(self, into, expected):
...
+        # last record includes index
+        assert records[size - 1][0] == 'b'
+        assert records[size - 1][1] == 'y'
+
+    def test_to_records_with_empty_multiindex(self):
+        # GH 21064
+        multi = MultiIndex([['a'], ['b']], labels=[[], []])
+        df = DataFrame(columns=['A'], index=multi)
+
+        records = df.to_records(index=True)
```
jreback: same

pandas/core/frame.py
```diff
@@ -1392,7 +1392,13 @@ def to_records(self, index=True, convert_datetime64=None):
             else:
                 if isinstance(self.index, MultiIndex):
                     # array of tuples to numpy cols. copy copy copy
-                    ix_vals = lmap(np.array, zip(*self.index.values))
+                    tuples = self.index.values
```
jreback: so rather than doing this here, call ``.tolist()`` on this and fix the underlying code, like this:

```
In [30]: pd.MultiIndex.from_product([list('abc'), range(2)]).tolist()
Out[30]: [('a', 0), ('a', 1), ('b', 0), ('b', 1), ('c', 0), ('c', 1)]

In [31]: pd.MultiIndex.from_product([list('abc'), range(2)])[0:0].tolist()
Out[31]: []
```
so [31] should be a nested list arrays of the correct length/dtype

this might not work, but let's try

jreback (changes requested):

pandas/core/frame.py
jreback: yes you can override ``tolist()`` (its defined in pandas/core/base) in MultiIndex

jreback (changes requested):

jreback:
can you rebase

jreback:

PR #21093, opened 2018-05-16
Support for OO Optimization

### LINKED ISSUE #21071: pandas is no longer importable with -OO optimization

In your shell:
```
$ python -OO -c 'import pandas'
  File "<string>", line 1, in <module>
  File "/Users/shoyer/miniconda3/envs/xarray-py36/lib/python3.6/site-packages/pandas/__init__.py", line 42, in <module>
    from pandas.core.api import *
  File "/Users/shoyer/miniconda3/envs/xarray-py36/lib/python3.6/site-packages/pandas/core/api.py", line 10, in <module>
    from pandas.core.groupby.groupby import Grouper
  File "/Users/shoyer/miniconda3/envs/xarray-py36/lib/python3.6/site-packages/pandas/core/groupby/__init__.py", line 2, in <module>
  File "/Users/shoyer/miniconda3/envs/xarray-py36/lib/python3.6/site-packages/pandas/core/groupby/groupby.py", line 46, in <module>
    from pandas.core.index import (Index, MultiIndex,
  File "/Users/shoyer/miniconda3/envs/xarray-py36/lib/python3.6/site-packages/pandas/core/index.py", line 2, in <module>
    from pandas.core.indexes.api import *
  File "/Users/shoyer/miniconda3/envs/xarray-py36/lib/python3.6/site-packages/pandas/core/indexes/api.py", line 4, in <module>
    from pandas.core.indexes.base import (Index,
  File "/Users/shoyer/miniconda3/envs/xarray-py36/lib/python3.6/site-packages/pandas/core/indexes/base.py", line 7, in <module>
    from pandas._libs import (lib, index as libindex, tslib as libts,
  File "pandas/_libs/index.pyx", line 28, in init pandas._libs.index
  File "pandas/_libs/tslibs/period.pyx", line 59, in init pandas._libs.tslibs.period
  File "/Users/shoyer/miniconda3/envs/xarray-py36/lib/python3.6/site-packages/pandas/tseries/offsets.py", line 1092, in <module>
    class CustomBusinessMonthEnd(_CustomBusinessMonth):
  File "/Users/shoyer/miniconda3/envs/xarray-py36/lib/python3.6/site-packages/pandas/tseries/offsets.py", line 1093, in CustomBusinessMonthEnd
    __doc__ = _CustomBusinessMonth.__doc__.replace('[BEGIN/END]', 'end')
AttributeError: 'NoneType' object has no attribute 'replace'
```

#### Problem description

`-OO` optimization strips out docstrings, which may give a minor performance boost (I honestly don't know). Nonetheless, users requested that xarray import properly properly with the `-OO` flag (https://github.com/pydata/xarray/issues/1706), so we added a regression test that caught this in the latest pandas release (https://github.com/pydata/xarray/pull/1708).

#### Expected Output

Pandas should be imported without any errors.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Darwin
OS-release: 17.4.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.22.0
pytest: 3.5.0
pip: 9.0.1
setuptools: 39.0.1
Cython: None
numpy: 1.14.2
scipy: None
pyarrow: None
xarray: None
IPython: 6.3.0
sphinx: None
patsy: None
dateutil: 2.7.2
pytz: 2018.3
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: None
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

WillAyd:
- [X] closes #21071 
- [X] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -50,6 +50,7 @@ Groupby/Resample/Rolling
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
 - Bug in :func:`DataFrame.agg` where applying multiple aggregation functions to a :class:`DataFrame` with duplicated column names would cause a stack overflow (:issue:`21063`)
+- Bug in :func:`pandas.core.groupby.GroupBy.ffill` and :func:`pandas.core.groupby.GroupBy.bfill` where the fill within a grouping would not always be applied as intended due to the implementations' use of a non-stable sort (:issue:`21207`)
 
 Strings
 ^^^^^^^
@@ -64,6 +65,7 @@ Categorical
 ^^^^^^^^^^^
 
 - Bug in :func:`pandas.util.testing.assert_index_equal` which raised ``AssertionError`` incorrectly, when comparing two :class:`CategoricalIndex` objects with param ``check_categorical=False`` (:issue:`19776`)
+- Bug in :meth:`Categorical.fillna` incorrectly raising a ``TypeError`` when `value` the individual categories are iterable and `value` is an iterable (:issue:`21097`, :issue:`19788`)
 
 Conversion
 ^^^^^^^^^^
@@ -76,11 +78,13 @@ Indexing
 
 - Bug in :meth:`Series.reset_index` where appropriate error was not raised with an invalid level name (:issue:`20925`)
 - Bug in :func:`interval_range` when ``start``/``periods`` or ``end``/``periods`` are specified with float ``start`` or ``end`` (:issue:`21161`)
+- Bug in :meth:`MultiIndex.set_names` where error raised for a ``MultiIndex`` with ``nlevels == 1`` (:issue:`21149`)
 -
 
 I/O
 ^^^
 
+- Bug in IO methods specifying ``compression='zip'`` which produced uncompressed zip archives (:issue:`17778`, :issue:`21144`)
 - Bug in :meth:`DataFrame.to_stata` which prevented exporting DataFrames to buffers and most file-like objects (:issue:`21041`)
 -
 
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -96,11 +96,6 @@ Reshaping
 - Bug in :func:`concat` where error was raised in concatenating :class:`Series` with numpy scalar and tuple names (:issue:`21015`)
 -
 
-Categorical
-^^^^^^^^^^^
-
--
-
 Other
 ^^^^^
 
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -81,7 +81,7 @@ Indexing
 I/O
 ^^^
 
--
+- Bug in :meth:`DataFrame.to_stata` which prevented exporting DataFrames to buffers and most file-like objects (:issue:`21041`)
 -
 
 Plotting
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -94,3 +94,8 @@ Categorical
 ^^^^^^^^^^^
 
 -
+
+Other
+^^^^^
+
+- Bug preventing pandas from being importable with -OO optimization (:issue:`21071`)
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -46,8 +46,6 @@ Documentation Changes
 Bug Fixes
 ~~~~~~~~~
 
-- tab completion on :class:`Index` in IPython no longer outputs deprecation warnings (:issue:`21125`)
-
 Groupby/Resample/Rolling
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -106,4 +104,5 @@ Categorical
 Other
 ^^^^^
 
+- Tab completion on :class:`Index` in IPython no longer outputs deprecation warnings (:issue:`21125`)
 - Bug preventing pandas from being importable with -OO optimization (:issue:`21071`)
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -15,6 +15,8 @@ and bug fixes. We recommend that all users upgrade to this version.
 New features
 ~~~~~~~~~~~~
 
+- :meth:`Index.droplevel` is now implemented also for flat indexes, for compatibility with MultiIndex (:issue:`21115`)
+
 
 .. _whatsnew_0231.deprecations:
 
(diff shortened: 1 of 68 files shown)


pandas/tests/test_downstream.py
```diff
@@ -53,6 +55,15 @@ def test_xarray(df):
     assert df.to_xarray() is not None
 
 
+def test_xarray_oo_optimizable():
```
jorisvandenbossche: you can remove the "xarray" here, as it is generally useful to ensure this, not only for xarray

jorisvandenbossche (commented):
small comment, looks good to me

pandas/tseries/offsets.py
```diff
@@ -1090,12 +1090,15 @@ def apply(self, other):
 
 
 class CustomBusinessMonthEnd(_CustomBusinessMonth):
-    __doc__ = _CustomBusinessMonth.__doc__.replace('[BEGIN/END]', 'end')
+    if _CustomBusinessMonth.__doc__:
+        __doc__ = _CustomBusinessMonth.__doc__.replace('[BEGIN/END]', 'end')
     _prefix = 'CBM'
 
 
 class CustomBusinessMonthBegin(_CustomBusinessMonth):
```
jreback: I think better to do this with ``@subsitution`` and template this. We dont' use this pattern anywhere and might be confusing.

jreback (changes requested):

jreback (changes requested):
also pls add a whatsnew 0.23.1 ok

pandas/tseries/offsets.py
WillAyd: Makes sense - I'll take a look and repush

pandas/tests/test_downstream.py
```diff
@@ -53,6 +55,11 @@ def test_xarray(df):
     assert df.to_xarray() is not None
 
 
+def test_oo_optimizable():
+    # GH 21071
+    subprocess.check_call(["python", "-OO", "-c", "import pandas"])
```
mattpap: Does pandas import all modules at `import pandas`? If not, maybe [this](https://github.com/bokeh/bokeh/blob/73f60d791310f4a03e9a8c06eb743de496cc2ada/tests/test_python_execution_with_OO.py) will be helpful.

pandas/util/_decorators.py
```diff
@@ -19,46 +19,42 @@ def deprecate(name, alternative, version, alt_name=None,
 
     Parameters
     ----------
-    name : str
-        Name of function to deprecate
-    alternative : str
-        Name of function to use instead
+    name : func
```
WillAyd: Unless I am overlooking something I didn't see where this decorator was being used, but it seems like there is a mismatch between what was documented and the types it actually accepts (at least looking at `alternative`).

To make this work with substitution I took the liberty of assuming these arguments should functions and not strings whenever this decorator gets used

WillAyd:
The failures here are the result of `__doc__` not being mutable for classes in Python27. I did some research on alternatives and there don't appear to be any that are that simple.

I suppose this explains why the Substitution / Appender system wasn't used in the first place for the `CustomBusinessMonth` classes and I don't see it used on any other classes within the code base (just methods). 

@jreback would you be OK with keeping the `CustomBusinessMonthStart` and `CustomBusinessMonthBegin` classes as is?

jreback (commented):
lgtm

pandas/tests/test_downstream.py
jorisvandenbossche: @WillAyd do we want to go this way to also check some submodules?

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -94,3 +94,8 @@ Categorical
 ^^^^^^^^^^^
 
 -
+
+Other
```
jreback: can you move the tab completion entry here as well (its just below Bug Fixes)

pandas/tests/test_downstream.py
WillAyd: I can take a look - if it's not a lot of effort I suppose it doesn't hurt

WillAyd: I've been messing around with this for a bit but AFAICT it keeps getting choked up when trying to import modules containing optional dependencies (ex: `pandas.io.s3`). I can't think of a simple way to configure the test so that it can differentiate intentional ImportErrors as a result of missing dependencies.

For now I'm going to revert to just working with the top level import - can open a separate issue if we wanted this to work recursively

jorisvandenbossche: Yep, if it turns out difficult, on problem in starting with the basic test (which already tests the large majority of our API)

pandas/util/_decorators.py
jorisvandenbossche: > Unless I am overlooking something I didn't see where this decorator was being used, 

It's not used a lot, but at least it is used for argmin/argmax and for scatter_matrix (it is not used there as a decorator, so I searched for "deprecate("

pandas/util/_decorators.py
```diff
@@ -19,46 +19,42 @@ def deprecate(name, alternative, version, alt_name=None,
...
+    @Substitution(version=version, msg=msg)
+    @Appender(name.__doc__)
     @wraps(alternative)
     def wrapper(*args, **kwargs):
+        """
+        .. deprecated:: %(version)s
+
+           %(msg)s
+
+        """
```
jorisvandenbossche: I am not sure why, but it seems this does not give the correct docstring. 
Can you check that `pd.Series.argmin?` still gives the deprecation warning at the top? (I checked out this branch and didn't see it, while on master it is there)

jorisvandenbossche (changes requested):

pandas/tseries/offsets.py
```diff
@@ -1090,12 +1090,17 @@ def apply(self, other):
 
 
 class CustomBusinessMonthEnd(_CustomBusinessMonth):
-    __doc__ = _CustomBusinessMonth.__doc__.replace('[BEGIN/END]', 'end')
+    # TODO: Replace condition with Subsitution after dropping Py27 support
```
jreback: ideally we could standarize these todos so can find later, maybe TODO(py27), I think we are doing similar to EA t

jreback (approved):
lgtm.

WillAyd:
AppVeyor failure is the random ResourceWarning that's been popping up which I was hoping would be fixed through #21105 but apparently was not. I don't believe its related to this change

jreback:
ok this looks fine. @jorisvandenbossche do you have any comments.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

Issue #19984, opened 2018-03-04
BLD: wheel building is now failing


jreback:
start 3 days ago: https://travis-ci.org/MacPython/pandas-wheels/builds/348053404

I think something is showing a ``UserWarning`` that didn't before.

like between master and: 52559f535788b3c59c6c78d2cf987b03303b16f6

jreback:
cc @jbrockmendel @WillAyd @Liam3851 @mroeschke @topper-123 

(you had non-trivial changes since then)

jreback:
builds for wheels are on numpy 1.9.3

WillAyd:
Kind of strange that the error getting thrown is a `ResourceWarning`. Checking the [Python docs](https://docs.python.org/3.6/library/warnings.html#default-warning-filters) this is typically ignored unless Python is built in debug mode. I don't see anything in the pandas codebase that would explicitly set that save when `PANDAS_TESTING_MODE` contains "deprecate" but I didn't see that in the applicable config.

jbrockmendel:
Is it happening consistently?  Could be a file handle leak?  If so, maybe try just instantiating/reading the StringIO obj without the rest of the test?

jreback:
no this is new

there is a warning that is not being caught

jreback:
this is ok now

PR #21105, opened 2018-05-17
Replaced open with Context Mgrs in Parser Tests

### LINKED ISSUE #21102: CI: Failure on Appveyor Master
https://ci.appveyor.com/project/pandas-dev/pandas/build/job/0y6fvce0al2yax06?fullLog=true#L1989

cc @WillAyd 

Maybe #21066 was related? Not sure how though. Looking now.

WillAyd:
Maybe closes #21102 and #19984, though if not should still work as a general cleanup

### DIFF
--- a/pandas/tests/io/parser/common.py
+++ b/pandas/tests/io/parser/common.py
@@ -54,20 +54,22 @@ bar2,12,13,14,15
         # and C engine will raise UnicodeDecodeError instead of
         # c engine raising ParserError and swallowing exception
         # that caused read to fail.
-        handle = open(self.csv_shiftjs, "rb")
         codec = codecs.lookup("utf-8")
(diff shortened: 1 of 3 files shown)


alysivji:
Just did a search and there are a surprising number of non-context manager `open`s in the codebase. Is there a cleanup effort around this? Fix as you see them?

WillAyd:
@alysivji are those using IO buffers or file handles? If the former I don't think it's a necessity. The latter should definitely be cleaned up

jreback:
thanks @WillAyd 

if you can see if #21102 and #19984 are fixed would be great.

WillAyd:
Is that validated by physically restarting the failed jobs associated with those?

jreback:
yep - nothing to do now

PR #21107, opened 2018-05-17
PERF: Improve performance of CategoricalIndex.is_unique

topper-123:
``CategoricalIndex.is_unique`` creates an extraneous boolean array. By changing ``CategoricalIndex.is_unique`` to use ``CategoricalIndex._engine.is_unique`` instead, this array creation is avoided. We simultaneously get to set ``is_monotonic*`` for free, and therefore will save time, if that property is called later.

Demonstration
==========

Setup:

```python
>>> n = 1_000_000
```

Currently, ``ci.is_unique`` is about the same (disregarding``@readonly_cache``) as:

```python
>>> from pandas._libs.hashtable import duplicated_int64
>>> not duplicated_int64(ci.codes.astype('int64')).any()
False
>>> %timeit duplicated_int64(ci.codes.astype('int64')).any()
46.7 ms ± 4.18 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Notice that the ``duplicated_int64()`` creates an boolean array, which is not needed and slows the operation down.

If we instead use ``ci._engine.is_unique`` to check for uniqueness, the check is roughly similar to:
```python
>>> from pandas._libs.algos import is_monotonic_int64
>>> is_monotonic_int64(ci.codes.astype('int64'), False)
(True, False, False)  # (is_monotonic_inc, is_monotonic_dec, is_unique)
>>> %timeit is_monotonic_int64(ci.codes.astype('int64'), False)
23.3 ms ± 364 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

This is faster than the other version, as the intermediate boolean array is not created in this version. Also, is it (IMO) more idiomatic, as ``index._engine`` is in general supposed to be used for this kind of index content checks.

### DIFF
--- a/pandas/tests/indexes/test_category.py
+++ b/pandas/tests/indexes/test_category.py
@@ -581,6 +581,15 @@ class TestCategoricalIndex(Base):
         assert c.is_monotonic_increasing
         assert not c.is_monotonic_decreasing
 
+    @pytest.mark.parametrize('values, expected', [
+        ([1, 2, 3], True),
+        ([1, 3, 1], False),
+        (list('abc'), True),
+        (list('aba'), False)])
+    def test_is_unique(self, values, expected):
+        ci = CategoricalIndex(values)
+        assert ci.is_unique is expected
+
     def test_duplicates(self):
 
         idx = CategoricalIndex([0, 0, 0], name='foo')
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -30,6 +30,7 @@ Performance Improvements
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
 - Improved performance of :meth:`CategoricalIndex.is_monotonic_increasing`, :meth:`CategoricalIndex.is_monotonic_decreasing` and :meth:`CategoricalIndex.is_monotonic` (:issue:`21025`)
+- Improved performance of :meth:`CategoricalIndex.is_unique` (:issue:`21107`)
 -
 -
 
--- a/pandas/core/indexes/category.py
+++ b/pandas/core/indexes/category.py
@@ -378,7 +378,7 @@ class CategoricalIndex(Index, accessor.PandasDelegate):
     # introspection
     @cache_readonly
     def is_unique(self):
-        return not self.duplicated().any()
+        return self._engine.is_unique
 
     @property
     def is_monotonic_increasing(self):


pandas/tests/indexes/test_category.py
```diff
@@ -581,6 +581,16 @@ def test_is_monotonic(self, data, non_lexsorted_data):
         assert c.is_monotonic_increasing
         assert not c.is_monotonic_decreasing
 
+    @pytest.mark.parametrize('unique, not_unique',
```
jschendel: Might be slightly cleaner to do the parametrize like:
```python
@pytest.mark.parametrize('values, expected', [([1, 2, 3], True), ...])
```
Then just have one `assert` in the test:
```python
assert ci.is_unique is expected
```

topper-123: Yeah, that is cleaner. I've changed it.

pandas/tests/indexes/test_category.py
```diff
@@ -581,6 +581,15 @@ def test_is_monotonic(self, data, non_lexsorted_data):
...
         assert not c.is_monotonic_decreasing
 
+    @pytest.mark.parametrize('values, expected', [
+        ([1, 2, 3], True),
+        ([1, 3, 1], False),
+        (list('abc'), True),
+        (list('aba'), False)])
+    def test_is_unique(self, values, expected):
+        ci = CategoricalIndex(values)
+        assert ci.is_unique == expected
```
jschendel: Can you use `is` instead of `==`?  A little bit of paranoia on my end, but it'd guard against the unlikely event that `is_unique` mistakenly starts returning equivalent non-booleans (e.g. 0/1).

topper-123: Ok, I've changed that.

jreback (changes requested):
is there an asv for this? if not can you add (ideally for all index types)

topper-123:
``is_unique`` is a cached property, so I don't think an ASV makes sense in this case. (I assume ``setup`` is called once for each time method, but not once for every call to the time method, haven't checked)

jreback:
how does this diminish the value of an asv?

sure if the subsequent calls are cached the mean time is lower but this should still show an asv improvement yes?

topper-123:
My impression has been that these neasurement are done in a “best of 3” style (like %timeit). For such tests the uncached version always loses.

Haven’t looked into how asv does it, so I could be mistaken here.

topper-123:
It seems to me that ASV does "best of X" style of testing. from the [docs](https://asv.readthedocs.io/en/latest/writing_benchmarks.html#timing):

> The timing itself is based on the Python standard library’s timeit module, with some extensions for automatic heuristics shamelessly stolen from IPython’s %timeit magic function

I could add the timing tests, but I assume these improvement is not seen well in repeated calls to ``is_unique``. You want them added still?

jreback:
we do have asv's for others. sure if can tag so these are run once is better (not sure if that is possible)


```
(pandas) bash-3.2$ grep -r is_unique asv_bench/benchmarks/
asv_bench/benchmarks//algorithms.py:        # cache is_unique
asv_bench/benchmarks//algorithms.py:        self.idx_int_dup.is_unique
```

topper-123:
> we do have asv's for others. sure if can tag so these are run once is better (not sure if that is possible)
>
> (pandas) bash-3.2$ grep -r is_unique asv_bench/benchmarks/
> asv_bench/benchmarks//algorithms.py:        # cache is_unique
> asv_bench/benchmarks//algorithms.py:        self.idx_int_dup.is_unique

Those two are in a setup method, so timing are not done on them:

```python
class DuplicatedUniqueIndex(object):

    goal_time = 0.2

    def setup(self):
        N = 10**5
        self.idx_int_dup = pd.Int64Index(np.arange(N * 5))
        # cache is_unique
        self.idx_int_dup.is_unique

     def time_duplicated_unique_int(self):
```

So currently there is no tests for ``is_unique`` in asv, and it probably doesn't make sense  IMO, as ``is_unique`` is a cached property and timing it will just return the cached value, not the algorithm...

jreback:
hmm ok. can you open a new issue to create asv's in some form or another for this (and is_unique in general). w/o asv's someone will ineveitable change something which breaks this.

maybe should just temporary turn cached_properties off with a context manager when running them.

jreback (approved):

topper-123:
> maybe should just temporary turn cached_properties off with a context manager when running them.

Is there currently a context manager or option for this in Pandas?

jreback:
no i think would need to create one for purposes of measuring cached properties

PR #21109, opened 2018-05-17
DOC: Add sphinx spelling extension

### LINKED ISSUE #21079: DOC: Enable sphinx.contrib.spelling
http://sphinxcontrib-spelling.readthedocs.io/en/latest/run.html

Saw this in Carol Willing's talk on Sphinx: https://www.youtube.com/watch?v=0ROZRNZkPS8

We'll want to

- add the extension
- build a wordlist of exceptions
- fix the currently failing items
- fail the doc build when there are exceptions

FabioRosado:
This is my first time contributing to a project this big so I hope everything is okay
I am still working on fixing bugs and polishing a few things but here is what I have done so far:

- fixed typos
- updated a wordlist txt file with words to be ignored by the spellchecker
-  added the spellcheck method to the DocBuilder class (should this be named spelling?)

I've re-used the _sphix_build method to run the spelling command, to do this I had to update the _sphix_build method to include 'spelling' as a kind.  I'd like to know if this is the best approach of If I should just replicate the code.

I have added a few configuring options in order to get better results when using the spellcheck, the first one was to use the wordlist text file, the second was to ignore known PyPI packages names and finally to show suggestions of a misspelt word. 

At the moment the spellcheck will run if we type the command: `python make.py spellcheck`  and the spellcheck method is called, I haven't yet figured out how to fail the build when there are exceptions like the Issue 21079 suggested.

----
- [x] closes #21079 
- [x] tests added / passed
- [x] whatsnew entry


doc/source/10min.rst
```diff
@@ -645,7 +645,7 @@ the quarter end:
    ts.index = (prng.asfreq('M', 'e') + 1).asfreq('H', 's') + 9
    ts.head()
 
-Categoricals
+Categorical
 ------------
```
TomAugspurger: Let's keep this, to not break links.

doc/source/advanced.rst
```diff
@@ -387,7 +387,7 @@ Furthermore you can *set* the values using the following methods.
    df2.loc(axis=0)[:, :, ['C1', 'C3']] = -10
    df2
 
-You can use a right-hand-side of an alignable object as well.
+You can use a right-hand-side of an align object as well.
```
TomAugspurger: I think alignable is OK.

doc/source/advanced.rst
```diff
@@ -593,7 +593,7 @@ Take Methods
 
 Similar to NumPy ndarrays, pandas Index, Series, and DataFrame also provides
 the ``take`` method that retrieves elements along a given axis at the given
-indices. The given indices must be either a list or an ndarray of integer
+indexes. The given indexes must be either a list or an ndarray of integer
```
TomAugspurger: Prefer `indices`, since that's the argument name to `take`.

doc/source/advanced.rst
```diff
@@ -711,7 +711,7 @@ order is ``cab``).
 
    df2.sort_index()
 
-Groupby operations on the index will preserve the index nature as well.
+Group by operations on the index will preserve the index nature as well.
```
TomAugspurger: Can probably add groupby to the wordlist.

doc/source/cookbook.rst
```diff
@@ -723,7 +723,7 @@ Rolling Apply to multiple columns where function returns a Scalar (Volume Weight
    s = pd.concat([ (pd.Series(vwap(df.iloc[i:i+window]), index=[df.index[i+window]])) for i in range(len(df)-window) ]);
    s.round(2)
 
-Timeseries
```
TomAugspurger: Add timeseries to the wordlist?

TomAugspurger (commented):
Thanks! Just gave a quick skim.

FabioRosado:
I apologise for taking so long to reply but this took me longer than I expected. I've added all the words to the wordlist (this file is now huge) and corrected some stuff, I will need your help in regards to some changes that I have done:

- changed `indicies` to `indices` since indices is already in the plural form (I only changed comments and docstrings, there are some variables using this indicies but I didn't want to touch them unless you agree)
- changed `performant` to `efficient` this term was used plenty of times and I'm not sure if changing it is desirable (if not I'll revert the changes)

The spelling extension is funny as it divides words by special characters (anything like `-` or `'`) so I had to add some words that don't exist in the dictionary but make sense when attached to others. (hope this makes sense?)

I've also given it some thought and I think we could fail the build by raising an exception if there are any words in this output.txt file, the idea is to:
- open the file
- readlines
- pass if empty
Not sure if this is a smart way to approach the problem though. 

Finally, the time it takes to spellcheck. If every single file needs to be generated then it will take about 2-3minutes to generate and read the files and about 1 minute to spellcheck and output misspelt words.

If the files are already generated then it runs in 1 minute. This is with the `spelling_show_suggestions` option set to False because this option seems to increase the time it takes to spellcheck to about 3x worse.

WillAyd (changes requested):
I haven't gotten all the way through yet but I see some consistent themes that could use updating and have labeled as such.

To answer your other question, stick with just DOC updates for the time being. If you wanted to go back and change variable names that should be a separate change if at all

FabioRosado:
Thanks for the feedback @WillAyd some of these changes where made yesterday and after working today on it I got a better understanding of a few terms. I also forgot to double check the previous commit, my apologies for it.

I will revert the changes from efficient back to performant. I'll try to do all the changes before submitting another commit so I'll wait a bit 😃👍

FabioRosado:
I got some time to work on this PR and here is the updated version:
- Reverted changes from `efficient` back to `performant`
- Revert `lex-sorted` back to `lexsorted`
- Fixed lint issues
- Added new section on documentation to inform how to run the spell check and how to add words to the file

As for the MultiIndex I changed two terms but as I searched for multi-index it seems that this term appears a fair amount of time so I'm not sure if I should change ALL of them from multi-index to MultiIndex or just leave it as it is.

On the documentation about the spell check should I explain better about the output.txt file and how to find the word that is being marked as spell check?

Finally, should I raise an exception if any errors were found after running the spellcheck command? I think the best way to do this would be opening the output.txt file, check if it's empty and if not raise an exception. Not sure if this i the best way to approach this problem though 😃

TomAugspurger:
> Finally, should I raise an exception if any errors were found after running the spellcheck command?

Yes, ideally. Do you know how long the spellchecker takes to run over our docs? We're close to our time limit on Travis for the doc build.

FabioRosado:
Okay I’ll add that bit of code when i get home later on. The spell checker takes 1 minute to run but if the docs weren’t generated yet they are generated first so it takes up to 6 minutes to run

FabioRosado:
The code will not raise an exception if there are misspelt words.  I was unsure which exception to raise so I went with SyntaxError with a message that perhaps will be enough to let people know why the exception was raised 😃

doc/source/categorical.rst
```diff
@@ -370,7 +370,7 @@ Renaming categories is done by assigning new values to the
 
 .. note::
 
-    Be aware that assigning new categories is an inplace operation, while most other operations
+    Be aware that assigning new categories is an in place operation, while most other operations
```
WillAyd: Still think inplace as a term is fine, especially since it aligns with the keyword used for the concept throughout pandas

doc/source/comparison_with_sql.rst
```diff
@@ -228,9 +228,9 @@ Grouping by more than one column is done by passing a list of columns to the
 JOIN
 ----
 JOINs can be performed with :meth:`~pandas.DataFrame.join` or :meth:`~pandas.merge`. By default,
-:meth:`~pandas.DataFrame.join` will join the DataFrames on their indices. Each method has
+:meth:`~pandas.DataFrame.join` will join the DataFrames on their indexes. Each method has
```
WillAyd: Sorry if I missed it before but any reason we changed these? `indices` seemed preferable to me

doc/source/contributing.rst
```diff
@@ -436,6 +436,25 @@ the documentation are also built by Travis-CI. These docs are then hosted `here
...
+contains any spelling errors. Sphinx provides an easy way to spell check documentation
+and docstrings.
+
+Running the spell check is easy. Just navigate to your local ``pandas/doc/`` directory and run::
+
+    python make.py spellcheck
+
+The spellcheck will take a few minutes to run (between 1 to 6 minutes). Sphinx will alert you
+with warnings and misspelt words - these misspelt words will be added to a file called
+``output.txt`` and you can find it on your local directory ``pandas/doc/build/spelling/``.
```
WillAyd: Not saying this is a bad thing but any reason you chose to output to a text file instead of to STDOUT? All of the other checks I can think of off the top of my head would write to the latter

doc/source/contributing.rst
```diff
@@ -436,6 +436,25 @@ the documentation are also built by Travis-CI. These docs are then hosted `here
...
+Spell checking documentation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When contributing to documentation to **pandas** it's good to check if your work
+contains any spelling errors. Sphinx provides an easy way to spell check documentation
+and docstrings.
+
+Running the spell check is easy. Just navigate to your local ``pandas/doc/`` directory and run::
+
+    python make.py spellcheck
```
WillAyd: Could we alternately add a rule to the Makefile so that `make spellcheck` works?

doc/source/spelling_wordlist.txt
```diff
@@ -0,0 +1,2557 @@
...
+Sakuma
+Hirschfeld
+Adrián
+Cañones
+Castellano
+Arcos
+Hoese
+Stansby
+Kamau
+Niederhut
```
WillAyd: Not saying we need to do this here but wonder if its cleaner to have a separate contributors file containing their names that could be pulled here in as well as into the docs

doc/source/whatsnew/v0.15.2.txt
```diff
@@ -230,11 +230,11 @@ Bug Fixes
 - Bug where index name was still used when plotting a series with ``use_index=False`` (:issue:`8558`).
 - Bugs when trying to stack multiple columns, when some (or all) of the level names are numbers (:issue:`8584`).
 - Bug in ``MultiIndex`` where ``__contains__`` returns wrong result if index is not lexically sorted or unique (:issue:`7724`)
-- BUG CSV: fix problem with trailing whitespace in skipped rows, (:issue:`8679`), (:issue:`8661`), (:issue:`8983`)
+- BUG CSV: fix problem with trailing white space in skipped rows, (:issue:`8679`), (:issue:`8661`), (:issue:`8983`)
 - Regression in ``Timestamp`` does not parse 'Z' zone designator for UTC (:issue:`8771`)
 - Bug in `StataWriter` the produces writes strings with 244 characters irrespective of actual size (:issue:`8969`)
 - Fixed ValueError raised by cummin/cummax when datetime64 Series contains NaT. (:issue:`8965`)
-- Bug in Datareader returns object dtype if there are missing values (:issue:`8980`)
+- Bug in Data reader returns object dtype if there are missing values (:issue:`8980`)
```
WillAyd: This was a class that could just be placed in backticks like `StatWriter` above. Also capitalize R so `DataReader`

doc/source/whatsnew/v0.18.0.txt
```diff
@@ -414,7 +414,7 @@ New Behavior:
    df.loc[ix, 'b'] = df.loc[ix, 'b']
    df.dtypes
 
-When a DataFrame's integer slice is partially updated with a new slice of floats that could potentially be downcasted to integer without losing precision, the dtype of the slice will be set to float instead of integer.
+When a DataFrame's integer slice is partially updated with a new slice of floats that could potentially be down casted to integer without losing precision, the dtype of the slice will be set to float instead of integer.
```
WillAyd: down-casted?

WillAyd (changes requested):

doc/source/comparison_with_sql.rst
FabioRosado: I saw both terms in the documentation and since the spelling marked indices as wrong I changed a lot of them that's the reason why. I can revert these changes 😃 👍

doc/source/contributing.rst
FabioRosado: Unfortunately, I didn't choose such approach. The spelling library is coded in this way, it will always output a text file with all the misspelt words. To be honest I would much rather work with STDOUT and avoid calling open on a file to check if it was empty

doc/source/contributing.rst
FabioRosado: I didn't think about this my apologies I'll add the spellcheck on the makefile 👍

doc/source/release.rst
```diff
@@ -2582,7 +2582,7 @@ Improvements to existing features
   docstrings for Panel flex methods. (:issue:`5336`)
 - ``NDFrame.drop()``, ``NDFrame.dropna()``, and ``.drop_duplicates()`` all
   accept ``inplace`` as a keyword argument; however, this only means that the
-  wrapper is updated inplace, a copy is still made internally.
+  wrapper is updated in place, a copy is still made internally.
```
WillAyd: Same here

doc/source/release.rst
```diff
@@ -2625,7 +2625,7 @@ API Changes
   :issue:`5744`, :issue:`5756`)
 - Default export for ``to_clipboard`` is now csv with a sep of `\t` for
   compat (:issue:`3368`)
-- ``at`` now will enlarge the object inplace (and return the same)
+- ``at`` now will enlarge the object in place (and return the same)
```
WillAyd: Here too

WillAyd (changes requested):
minor comments but otherwise lgtm. If you want to create a make rule for this could be a logical follow up *(separate PR)

FabioRosado:
Ops I thought I had reverted all the terms but I missed those two sorry for that. 
They are now changed and I added the makefile rule hope all is okay now, thanks for the help 😄 👍

Makefile
```diff
@@ -23,3 +23,4 @@ doc:
 	cd doc; \
 	python make.py clean; \
 	python make.py html
+	python make.py spellcheck
```
WillAyd: Hmm well I was thinking with would be a separate rule but the way you've done it doc is fine and may even be preferable. Just calling out here in case another reviewer has a differing opinion

WillAyd (approved):
lgtm thanks for tackling these updates!

doc/source/names_wordlist.txt
```diff
@@ -0,0 +1,1652 @@
...
+Shahapure
+Pölsterl
+Rubbert
+Sinhrks
+Siu
+Kwan
+Seabold
+Carrucciu
+Hoyer
+Pascoe
```
jreback: these look like the names of our contributors? can we simply generate this?

something like: `git log --format='%an' | sort|uniq` (I guess would have to split on whitespace as well)

jreback (changes requested):
is this added to the build?

doc/source/names_wordlist.txt
FabioRosado: The spelling extension checks every single file for text/docstrings so the names of the contributors needed to be whitelisted in order to be ignored by the spellchecker.

How would you advise on generating this? The reason why I'm asking is that the spelling extension uses `path.join` on anything that you pass as a wordlist (one or two files) so we only get an exception when trying to use let's say a list of names.

I suppose we could run the command and write it to file on every run? Perhaps using something like a tempfile to get all those names and use that file on every run?  

Also, I have encountered a few issues when decoding and splitting every name - the spelling extension makes some weird word divisions at times, when I try to decode and split the list  Yan Facai name will be split like this: `['颜发才（Yan', 'Facai）']` since the wordlist expects one line per word, then this name will always be marked as wrong due to how the split divided the name.  I suppose using regex could help with this matter but not sure.

TomAugspurger:
We already have
https://github.com/pandas-dev/pandas/blob/master/scripts/announce.py for
getting a list of contributors.

We could refactor that to update the wordlist of contributors when it's rn.

FabioRosado:
I see. But the names need to be added to a wordlist like the file named `spelling_wordlist.txt`. Since the extension captures all text within the project every term will have to be whitelisted (unless its a grammatical error) which means that the name of every contributor needs to be included in a file somewhere. 

It makes sense that perhaps the contributor names should be in a separate file like WillAyd suggested. Also, the extension needs very specific ways to whitelist terms - one term per line. As I pointed out, the problem with scrappy contributors name is that the names will need to be split into both newlines + spaces so every name is a whitelisted word.

Due to the fact that the extension divides words by special characters, it could potentially cause trouble with names such as the one from  Yan Facai (example: 颜发才（Yan Facai） ) so the name needs to be added with both the special characters and normal characters - you can see this on the wordlist.

So I'm unsure what to do now in regards to jreback suggestions. If anyone else could give a hand in order for me to solve this issue and the PR merged I'll be happy to fix/modify anything 😃 👍

TomAugspurger:
> It makes sense that perhaps the contributor names should be in a separate file

Agreed. I would

1. Modify `scripts/announce.py`  to take an optional flag that disables the `+` at the end of the name of new contributors
2. Run `scripts/announce.py v0.4.0..HEAD` to get a list of all the contributors
3. Update `scripts/announce.py` to update the list of contributors (will have to do some de-duplication)
4. Update the docbuild to run scripts/announce.py` on each build.

I've been meaning to change the docs around so that 4 is done anyone. I'll hopefully get to it by the end of the week.

WillAyd:
@TomAugspurger that would be nice but I'm of the opinion that should be a separate PR, if only because that will introduce some code changes that exceed what we would normally allow in a DOC change. OK to merge this and have that as a follow up?

TomAugspurger:
I'm fine with doing it as a followup, as long as the doc build is currently
passing. I image this will be hard to keep up to date with master.

jreback:
i am ok with only merging the actual spelling changes

the machinery needs to be done in a maintenance aware manner (no hard coding)

so if this were split up then ok

TomAugspurger:
What hard coding are you referring to, the contributors list?

It'll have to always be "hard coded" in the sense that it's a static text
file. IIUC, in the meantime if we have a new contributor name, the doc
build will fail, and they'll need to manually add their name to the
contributors list?

jreback:
yes my point is that it needs to be auto generated 
otherwise this will become a maintenance burden

if u want to make this a manual action that is ok too

but it should not be part of he build (is it? my question was never answered)

TomAugspurger:
It's not currently part of the doc build on Travis, so for now it's the ability to run spellcheck + fixing a bunch of typos.

Opened https://github.com/pandas-dev/pandas/issues/21354 for the followup.

TomAugspurger:
Thanks @FabioRosado! If you're interested in the followup changes to the build processes feel free to claim it on #21354.

doc/source/whatsnew/v0.14.0.txt
```diff
@@ -781,7 +781,7 @@ Enhancements
   noon, January 1, 4713 BC.  Because nanoseconds are used to define the time
   in pandas the actual range of dates that you can use is 1678 AD to 2262 AD. (:issue:`4041`)
 - ``DataFrame.to_stata`` will now check data for compatibility with Stata data types
-  and will upcast when needed.  When it is not possible to losslessly upcast, a warning
+  and will upcast when needed.  When it is not possible to lossless upcast, a warning
```
h-vetinari: This change is wrong - needs to be an adverb.

doc/source/whatsnew/v0.10.0.txt
```diff
@@ -370,7 +370,7 @@ Updated PyTables Support
       df1.get_dtype_counts()
 
 - performance improvements on table writing
-- support for arbitrarily indexed dimensions
+- support for arbitrary indexed dimensions
```
h-vetinari: This should stay an adverb.

doc/source/io.rst
```diff
@@ -2356,7 +2356,7 @@ Read a URL and match a table that contains specific text:
 
 Specify a header row (by default ``<th>`` or ``<td>`` elements located within a
 ``<thead>`` are used to form the column index, if multiple rows are contained within
-``<thead>`` then a multiindex is created); if specified, the header row is taken
+``<thead>`` then a multi-index is created); if specified, the header row is taken
```
h-vetinari: All above changes were to `MultiIndex` - here `multi-index`?

doc/source/whatsnew/v0.16.1.txt
```diff
@@ -133,7 +133,7 @@ groupby operations on the index will preserve the index nature as well
 reindexing operations, will return a resulting index based on the type of the passed
 indexer, meaning that passing a list will return a plain-old-``Index``; indexing with
 a ``Categorical`` will return a ``CategoricalIndex``, indexed according to the categories
-of the PASSED ``Categorical`` dtype. This allows one to arbitrarly index these even with
+of the PASSED ``Categorical`` dtype. This allows one to arbitrary index these even with
```
h-vetinari: adverb

doc/source/whatsnew/v0.20.0.txt
```diff
@@ -462,7 +462,7 @@ Selecting via a scalar value that is contained *in* the intervals.
 Other Enhancements
 ^^^^^^^^^^^^^^^^^^
 
-- ``DataFrame.rolling()`` now accepts the parameter ``closed='right'|'left'|'both'|'neither'`` to choose the rolling window-endpoint closedness. See the :ref:`documentation <stats.rolling_window.endpoints>` (:issue:`13965`)
+- ``DataFrame.rolling()`` now accepts the parameter ``closed='right'|'left'|'both'|'neither'`` to choose the rolling window-endpoint closed. See the :ref:`documentation <stats.rolling_window.endpoints>` (:issue:`13965`)
```
h-vetinari: "Closedness" might not be the prettiest word, but "to choose the [...] closed" is definitely wrong

h-vetinari (commented):
Had a quick look through the diff, and some adverbs seem to have fallen through the cracks. Also "closedness". 

Furthermore, there's an inconsistency with "MultiIndex" vs "multi-index" (and potentially vs. "multi-value").

doc/source/release.rst
```diff
@@ -4115,7 +4115,7 @@ Bug Fixes
   datetime64 when calling DataFrame.apply. (:issue:`2374`)
 - Raise exception when calling to_panel on non uniquely-indexed frame (:issue:`2441`)
 - Improved detection of console encoding on IPython zmq frontends (:issue:`2458`)
-- Preserve time zone when .append-ing two time series (:issue:`2260`)
+- Preserve time zone when .appending two time series (:issue:`2260`)
```
h-vetinari: The hyphen was surely intentional...

TomAugspurger:
@h-vetinari this is merged so your comments are probably going to be lost.
Feel free to make a new PR with the fixes.

FabioRosado:
Even though this PR was already merged I'm happy to fix these issues. 
In regards to MultiIndex vs Multi-index it's a bit confusing because both terms seem to be in use (for examples in the `io.rst` the term Multi-index is used on a title but on advanced MultiIndex is used) which one you think is right @TomAugspurger ?

Also apologies for replying here :)

WillAyd:
MultiIndex is what we want

PR #21110, opened 2018-05-17
PERF: Speedup in printing DataFrames of long strings

david-liu-brattle-1:
Currently, when printing DataFrames, each element of a table to be printed is first formatted in its entirety by `GenericArrayFormatter._format_strings` and then truncated by `_make_fixed_width`.  When processing DataFrames that contain very long strings, there is a noticeable computation delay when calling `DataFrame.__repr__`.

The bottleneck is in `_format_strings`, where each string runs several `str.replace` to format it for printing.  We can eliminate wasted time by slicing any any strings longer than the maximum column width **before** pretty formatting the string, leaving enough buffer space such that formatting the string won't bring it below the maximum column width.

### Performance Demonstration
Setup:
```python
>>> data = pd.DataFrame([str(list(range(5000000)))]*50)
>>> len(data.values[0][0]) #length of each string
3888890
>>> _ = data.__repr__()
```

Final steps goes from 0.97386 seconds to 0.0089 seconds to execute.

### DIFF
--- a/pandas/tests/reshape/test_pivot.py
+++ b/pandas/tests/reshape/test_pivot.py
@@ -161,6 +161,24 @@ class TestPivotTable(object):
 
         tm.assert_frame_equal(result, expected)
 
+        # gh-21378
+        df = pd.DataFrame(
+            {'A': pd.Categorical(['left', 'low', 'high', 'low', 'high'],
+                                 categories=['low', 'high', 'left'],
+                                 ordered=True),
+             'B': range(5)})
+
+        result = df.pivot_table(index='A', values='B', dropna=dropna)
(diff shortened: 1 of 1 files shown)


chris-b1 (changes requested):
Could you also add an asv that covers this case? Could add a new `strings.py` to the io subfolder.

https://github.com/pandas-dev/pandas/tree/master/asv_bench/benchmarks/io

See docs here
https://pandas.pydata.org/pandas-docs/stable/contributing.html#running-the-performance-test-suite

jreback (changes requested):
yep pls add an asv & a whatsnew note.

does this break any current tests?

PR #21111, opened 2018-05-17
BUG: to_clipboard fails to format output for Excel

### LINKED ISSUE #21104: pandas.DataFrame.to_clipboard with excel option does not parse into columns with 0.23
#### Problem description

With Version 0.23 of pandas, the function .to_clipboard(), with the option excel=True no longer produces a copy of the dataframe which is Excel friendly when pasting the data into an Excel worksheet.  All the data comes into a single column in the Excel workbook.  Previously, the data would parse across all the columns.

The code that demonstrates this situation is:

```
Import pandas as pd
df = pd.DataFrame(np.random.randn(5,10))

df.to_clipboard(excel=True)
```

The output in excel ends up looking like this:

[https://www.screencast.com/t/DPTOukwE](url)

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Darwin
OS-release: 17.5.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.23.0
pytest: 3.5.1
pip: 10.0.1
setuptools: 39.1.0
Cython: 0.28.2
numpy: 1.13.3
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: 1.7.4
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.2.1
tables: 3.4.3
numexpr: 2.6.5
feather: None
matplotlib: 2.2.2
openpyxl: 2.5.3
xlrd: 1.1.0
xlwt: 1.2.0
xlsxwriter: 1.0.4
lxml: 4.2.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.7
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

david-liu-brattle-1:
`DataFrame.to_clipboard` has been broken for pasting to excel.  Tables are copied with spaces as delimiters instead of tabs (#21104).  

This issue originated in https://github.com/pandas-dev/pandas/commit/e1d5a2738235fec22f3cfad4814e09e3e3786f8c#diff-3f25860d9237143c1952a1f93c3aae18R102
which I've partially reverted.  

By setting the delimiter to `r'\t'`, a 2 character string, obj.to_csv raised an error, but is was caught and passed silently.  I reverted the separator to `'\t'`. 

Similar issue in `read_clipboard` also fixed.

- [x] closes #21104


WillAyd:
Can you add a test to make sure this produces the correct output?

WillAyd (changes requested):
Can you add a test to make sure this produces the correct output? Also could use a whatsnew note

pandas/tests/io/test_clipboard.py
```diff
@@ -123,6 +124,15 @@ def test_read_clipboard_infer_excel(self):
         exp = pd.read_clipboard(**clip_kwargs)
 
         tm.assert_frame_equal(res, exp)
+        
+    def test_excel_clipboard_format(self):
+        for dt in self.data_types:
+            for sep in ['\t', None]:
+                data = self.data[dt]
+                data.to_clipboard(excel=True, sep=sep)
```
chris-b1: Can you also test with `excel=False`?  In <=0.22 it would also be tab delimited

david-liu-brattle-1: You mean to test `data.to_clipboard(excel=False, sep=sep)`?  I don't think that was (or should be) tab delimited.  Do you mean `excel=None`?

chris-b1: Sorry, right, I meant to the default `df.to_clipboard()`  (sep=None, excel=None)

```
# pandas 0.22
In [72]: df = pd.DataFrame({'a': [1, 3], 'b': [4, 5]})

In [73]: df.to_clipboard()

# <paste>
In [74]: res = """	a	b
    ...: 0	1	4
    ...: 1	3	5"""

In [75]: res.count('\t')
Out[75]: 6

Comment (david-liu-brattle-1) [reply pandas/tests/io/test_clipboard.py]: Good point.  Should be added

File: pandas/tests/io/test_clipboard.py [hunk 2]
```diff
@@ -69,18 +72,36 @@ def teardown_class(cls):
     def check_round_trip_frame(self, data_type, excel=None, sep=None,
                                encoding=None):
         data = self.data[data_type]
-        data.to_clipboard(excel=excel, sep=sep, encoding=encoding)
-        if sep is not None:
-            result = read_clipboard(sep=sep, index_col=0, encoding=encoding)
+        if excel in [None, True] and sep is not None and len(sep) > 1:
```
WillAyd: Admittedly not overly familiar with this code but having a hard time making sense of it - is this possible to parametrize? We typically avoid a lot of conditionals and loops in test scenarios

pandas/tests/io/test_clipboard.py
```diff
@@ -124,6 +145,34 @@ def test_read_clipboard_infer_excel(self):
 
         tm.assert_frame_equal(res, exp)
 
+    def test_excel_clipboard_tabs(self):
+        for dt in self.data_types:
```
WillAyd: Same thing here - can we replace the loops with parametrization?

pandas/tests/io/test_clipboard.py
```diff
@@ -60,6 +60,9 @@ def setup_class(cls):
         # unicode round trip test for GH 13747, GH 12529
         cls.data['utf8'] = pd.DataFrame({'a': ['µasd', 'Ωœ∑´'],
                                          'b': ['øπ∆˚¬', 'œ∑´®']})
+        # Test for quotes and common delimiters in text
+        cls.data['delim_symbols'] = pd.DataFrame({'a': ['"a,\t"b|c', 'd\tef´'],
```
WillAyd: What's the point of this?

pandas/tests/io/test_clipboard.py
```diff
@@ -124,6 +145,34 @@ def test_read_clipboard_infer_excel(self):
...
+                        kwargs['excel'] = excel
+                    if sep != 'default':
+                        kwargs['sep'] = sep
+                    data = self.data[dt]
+                    data.to_clipboard(**kwargs)
+                    if sep in ['\t', None, 'default'] and excel is not False:
+                        # Expect tab delimited
+                        result = read_clipboard(sep='\t', index_col=0)
+                        tm.assert_frame_equal(data, result, check_dtype=False)
+                        assert clipboard_get().count('\t') > 0
```
WillAyd: Instead of counting the number of tabs can you construct the expected output and match exactly against that?

pandas/tests/io/test_clipboard.py
```diff
@@ -69,18 +72,36 @@ def teardown_class(cls):
     def check_round_trip_frame(self, data_type, excel=None, sep=None,
                                encoding=None):
         data = self.data[data_type]
-        data.to_clipboard(excel=excel, sep=sep, encoding=encoding)
-        if sep is not None:
-            result = read_clipboard(sep=sep, index_col=0, encoding=encoding)
+        if excel in [None, True] and sep is not None and len(sep) > 1:
+            with tm.assert_produces_warning():
```
WillAyd: Typically for raising exceptions and warnings we split this off into separate tests

pandas/io/clipboards.py
```diff
@@ -108,8 +112,9 @@ def to_clipboard(obj, excel=True, sep=None, **kwargs):  # pragma: no cover
                 text = text.decode('utf-8')
             clipboard_set(text)
             return
-        except:
-            pass
+        except TypeError:
+            warnings.warn('to_clipboard in excel mode requires a single \
+            character separator. Set "excel=false" or change the separator')
```
WillAyd: Capitalize `False`

WillAyd (changes requested):

pandas/tests/io/test_clipboard.py
david-liu-brattle-1: Excel can be very picky about how it handles copying/pasting text that includes delimiters or quote characters.  This dataframe caused some of the original clipboard tests to fail until I changed the to_clipboard and read_clipboard functions.

pandas/tests/io/test_clipboard.py
david-liu-brattle-1: Do you mean like this one? https://github.com/pandas-dev/pandas/blob/e033c0616158d3ba974456b4f84810492936b1fe/pandas/tests/io/test_common.py#L183-L188

WillAyd: Yes that's one example. There's plenty of them in the tests so feel free to poke around

jreback (changes requested):
yeah pls re-factor these tests to make it clear what is going on. ideally you can refactor in a commit, then do your changes in the following (or even a separate PR) to refactor first.

jreback:
@david-liu-brattle-1 can you rebase / update

pandas/io/clipboards.py
```diff
@@ -108,8 +112,9 @@ def to_clipboard(obj, excel=True, sep=None, **kwargs):  # pragma: no cover
                 text = text.decode('utf-8')
             clipboard_set(text)
             return
-        except:
-            pass
+        except TypeError:
+            warnings.warn('to_clipboard in excel mode requires a single \
+            character separator. Set "excel=False" or change the separator')
```
jorisvandenbossche: Can you use implicit line continuation instead of `\` ?

Like
```
            warnings.warn('to_clipboard in excel mode requires a single '
                          'character separator. Set "excel=False" or change '
                          'the separator')
```

jorisvandenbossche (commented):
You removed all the tests? (not fully clear if that was asked by the other reviewers) 

@WillAyd @chris-b1 can you take a new look?

jorisvandenbossche:
And the failing tests are actually clipboard tests that are failing, so should be addressed.

david-liu-brattle-1:
@jorisvandenbossche Separate pr for the clipboard tests in #21163

jorisvandenbossche:
Ah, thanks, didn't see that! 
Only, the existing tests in this PR should still pass / be fixed for this to be merged.

pandas/io/clipboards.py
```diff
@@ -55,11 +56,17 @@ def read_clipboard(sep=r'\s+', **kwargs):  # pragma: no cover
 
     counts = {x.lstrip().count('\t') for x in lines}
     if len(lines) > 1 and len(counts) == 1 and counts.pop() != 0:
-        sep = r'\t'
+        sep = '\t'
 
     if sep is None and kwargs.get('delim_whitespace') is None:
```
jreback: can you add some comments on these cases (including the existing ones).

pandas/io/clipboards.py
```diff
@@ -55,11 +56,17 @@ def read_clipboard(sep=r'\s+', **kwargs):  # pragma: no cover
...
         sep = r'\s+'
 
+    if sep == r'\s+' and kwargs.get('engine') is None:
+        kwargs['engine'] = 'python'
+
+    if kwargs.get('engine') == 'python' and compat.PY2:
```
jreback: why is the engine important here, shouldn't we alway encode in PY2?

pandas/io/clipboards.py
```diff
@@ -108,8 +115,10 @@ def to_clipboard(obj, excel=True, sep=None, **kwargs):  # pragma: no cover
                 text = text.decode('utf-8')
             clipboard_set(text)
             return
-        except:
-            pass
+        except TypeError:
+            warnings.warn('to_clipboard in excel mode requires a single '
+                          'character separator. Set "excel=False" or change '
```
jreback: is this tested?

test (2).py
```diff
@@ -0,0 +1,752 @@
...
+# util
+# extension for pseudo-safely moving bytes into mutable buffers
+_move_ext = Extension('pandas.util._move',
+                      depends=[],
+                      sources=['pandas/util/move.c'])
+extensions.append(_move_ext)
+
+# The build cache system does string matching below this point.
+# if you change something, be careful.
+
```
jreback: looks like you copied in a random file, pls remove.

test (2).py
```diff
@@ -0,0 +1,752 @@
...
+setup(name=DISTNAME,
+      maintainer=AUTHOR,
+      version=versioneer.get_version(),
+      packages=find_packages(include=['pandas', 'pandas.*']),
```
jreback: pls add tests

jreback (changes requested):

jorisvandenbossche (commented):
@david-liu-brattle-1 doesn't this need a test for the actual case that broke? (so the `excel=True` case)

david-liu-brattle-1:
@jorisvandenbossche 
#21163 should have tests for all of the issues fixed in this PR.  I think that one will be merged first? In which case adding more tests here would be redundant right?

jreback:
@david-liu-brattle-1 

- can you rebase on master
- add a whatsnew 0.23.2, IIRC this is a bug fix (IO section)
- remove the setup.py that was added.

ping on green.

jorisvandenbossche:
> #21163 should have tests for all of the issues fixed in this PR. I think that one will be merged first? In which case adding more tests here would be redundant right?

Ah, OK, the other PR is merged now, so then in this PR the `xfail`s should be removed?

david-liu-brattle-1:
@jreback 
I think we're all set except for the setup.py file.  I'm not sure what exactly is happening with that, from my end it doesn't look like there are any changes to the existing file and I don't think I created that file.

chris-b1:
Looks like file permissions were modified on `setup.py` (the `100755 → 100644` in the diff)

Should be able to just to reset to the current setup.py to back out the change `git checkout origin/master setup.py`

chris-b1:
See also - https://stackoverflow.com/questions/1257592/how-do-i-remove-files-saying-old-mode-100755-new-mode-100644-from-unstaged-cha - might have some funky git settings

jreback (approved):
rather minor point, other lgtm.

jreback:
@WillAyd merge when satisfied

pandas/io/clipboards.py
```diff
@@ -55,11 +56,27 @@ def read_clipboard(sep=r'\s+', **kwargs):  # pragma: no cover
...
+    # Edge case where sep is specified to be None, return to default
     if sep is None and kwargs.get('delim_whitespace') is None:
         sep = r'\s+'
 
+    # Regex separator currently only works with python engine.
+    # Default to python if separator is multi-character (regex)
+    if len(sep) > 1 and kwargs.get('engine') is None:
+        kwargs['engine'] = 'python'
+    elif len(sep) > 1 and kwargs.get('engine') == 'c':
+        warnings.warn('from_clipboard with regex separator does not work'
```
WillAyd: What was the reasoning for going with a warning here instead of an error? Curious what actually happens if this comes up (question maybe applicable to other errors as well)

jorisvandenbossche: I think that's a good question, we should check what `read_csv` does (warning or erroring) 
But anyhow, if we change it to an error, that would be for 0.24.0 IMO, so another PR.

pandas/io/clipboards.py
```diff
@@ -121,14 +121,13 @@ def to_clipboard(obj, excel=True, sep=None, **kwargs):  # pragma: no cover
...
             text = buf.getvalue()
-            if compat.PY2:
+            if PY2:
                 text = text.decode('utf-8')
             clipboard_set(text)
             return
         except TypeError:
             warnings.warn('to_clipboard in excel mode requires a single '
-                          'character separator. Set "excel=False" or change '
-                          'the separator')
```
jorisvandenbossche: @david-liu-brattle-1 I removed the last sentence of this warning, as setting `excel=False` does not help as then you get the warning that the separator is ignored. It's simply that `to_clipboard` does not support multiple character separator at all.

jorisvandenbossche (approved):

PR #19670, opened 2018-02-13
ENH-19629: Adding numpy nansun/nanmean, etc etc to _cython_table

AaronCritchley:
- [x] closes #19629 
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

As per the issue, here's proof of solution:
```python
Python 3.6.4 |Anaconda, Inc.| (default, Jan 16 2018, 12:04:33) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas as pd
>>> import numpy as np
>>> pd.Series([1,2,3,4]).agg(np.sum)
10
>>> pd.Series([1,2,3,4]).agg(np.nansum)
10
```

Where should I add tests for these changes? I wasn't sure on the best fit and how to most effectively test. Also, is a whatsnew entry needed here? If yes, any guidance on what it should be?

Issue #8383, opened 2014-09-24
nanmin no longer works with series


rstoneback:
Using NumPy's nanmin on a pandas Series worked as expected, now however the Series is returned with each element set to the minimum, rather than just getting the minimum. Not sure if this is a Numpy of Pandas issue.

```
np.nanmin(pds.Series([1,2,3,4]))
Out[14]: 
0    1
1    1
2    1
3    1
dtype: int64

np.min(pds.Series([1,2,3,4]))
Out[15]: 1
```

topper-123:
This issue has been solved:

```python
>>> np.nanmin(pds.Series([1,2,3,4]))
1
```

I'm on numpy 1.13.1 and pandas 0.22.

jreback:
yup. though its possibly only for certainly versions of numpy. @topper-123 can you add some tests and see?

topper-123:
This is solved only for numpy >= 1.13. Where would tests for this be placed?

jreback:
test_nanops

PR #19753, opened 2018-02-18
TST: add test for numpy ops, esp. nanmin/max bug for np<1.13

topper-123:
- [x] closes #8383
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

The bug in #8383 concerns np.nanmin and np.nanmax only, don't know if I'm testing too much.

PR #21123, opened 2018-05-18
ENH: add np.nan funcs to _cython_table

### LINKED ISSUE #19629: series.agg(np.nansum) etc. returns a  series, scalar expected
```python
>>> pd.Series([1,2,3,4]).agg(np.sum)
10
>>> pd.Series([1,2,3,4]).agg(np.nansum)
0    1
1    2
2    3
3    4
dtype: int64
```
I would expect the return values of ``.agg(np.nansum)`` and ``agg(np.sum)`` to be the same. 

the same issue is with np.nanmean, np.nanmax etc.

#### Problem description

The underlying issue is that ``SelectionMixIn._cython_table`` has records for ``np.sum``, ``np.mean`` etc, but not ``np.nansum``, ``np.nanmean`` etc.

From a quick look at it, it seems that the issue is easily fixable, by adding the ``np.nan*`` function to cython_table.

<details>

INSTALLED VERSIONS
------------------
commit: 7590b8723516a9d595c287067fa639fbb9e19ec1
python: 3.6.3.final.0
python-bits: 32
OS: Windows
OS-release: 10
machine: AMD64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.22.0.dev0+710.g7590b87.dirty
pytest: 3.3.1
pip: 9.0.1
setuptools: 38.2.5
Cython: 0.26.1
numpy: 1.13.3
scipy: 1.0.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: 1.6.3
patsy: 0.4.1
dateutil: 2.6.1
pytz: 2017.3
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 2.1.0
openpyxl: 2.4.9
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: None
lxml: None
bs4: None
html5lib: 1.0b10
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

### LINKED ISSUE #21134: BUG: df.agg(sum, axis=1) gives wrong result when Nan value is in frame

```python
>>> df.agg(sum, axis=0)
0    3.0
1    6.0
dtype: float64
>>> df.agg(sum, axis=1)
0    NaN
1    7.0
dtype: float64
```

```python
>>> n = 1_000
```

#### Problem description

#### Expected Output

The expected output is:

```python
>>> df.agg(sum, axis=1)
0    2.0
1    7.0
dtype: float64
```

#### Solution proposal

topper-123:
closes #19629 
closes #21134
- [x] tests added / passed
- [x] whatsnew entry

This started as a copy of #19670 by @AaronCritchley, but has solved two bugs that the tests surfaced along the way.

Bug 1:
there is currently a bug in ``df.aggregate``, where the method incorrectly defers to ``df.apply`` in a corner case. This only shows up in the result when using numpy < 1.13 and passing np.nan* functions to ``df.aggregate``. This is the reason for the change in ``base.py`` line 571. (see #8383 for further details on the bug in numpy<1.13 and how it affects pandas.)

Bug 2:
Passing builtins to ``df.aggregate`` is ok when ``axis=0``, but gives wrong result,when ``axis=1`` (#21134). The reason for this difference is that ``df.aggregate`` defers to ``df._aggregate`` when ``axis=0,`` but defers to ``df.apply``, when ``axis=1``, and these give different result when passed funcions  and the series/frame contains Nan values. This can be solved by transposing df and defering the transposed frame to its ``_aggragate`` method when ``axis=1``.

The added tests have been heavily parametrized (this helped unearth the bugs above). Thet have been placed in ``series/test_apply.py`` and ``frame/test_apply``, as a lot of other tests for ser/df.aggregate were already there.

### DIFF
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -94,3 +94,8 @@ Categorical
 ^^^^^^^^^^^
 
 -
+
+Numeric
+^^^^^^^
+
+- :meth:`~DataFrame.agg` now correctly handles numpy NaN-aware methods like :meth:`numpy.nansum` (:issue:`19629`)
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -99,3 +99,4 @@ Numeric
 ^^^^^^^
 
 - :meth:`~DataFrame.agg` now correctly handles numpy NaN-aware methods like :meth:`numpy.nansum` (:issue:`19629`)
+- :meth:`~DataFrame.agg` now correctly handles built-in methods like ``sum`` when axis=1 (:issue:`21134`)
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -149,3 +149,20 @@ def tz_aware_fixture(request):
     Fixture for trying explicit timezones: {0}
     """
     return request.param
+
+
+@pytest.fixture(
+    # params: Python 3.5 randomizes dict access and xdist doesn't like that
+    # in fixtures. In order to get predetermined values we need to sort
+    # the list deterministically
+    # GH 21123
+    params=list(sorted(pd.core.base.SelectionMixin._cython_table.items(),
+                       key=lambda x: x[0].__name__)),
+    ids=lambda x: "({}-{!r})".format(x[0].__name__, x[1]),
+)
+def cython_table_items(request):
+    """
+    Fixture for returning the items in
+    pandas.core.base.SelectionMixin._cython_table
+    """
+    return request.param
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -158,7 +158,7 @@ def tz_aware_fixture(request):
     # GH 21123
     params=list(sorted(pd.core.base.SelectionMixin._cython_table.items(),
                        key=lambda x: x[0].__name__)),
-    ids=lambda x: "({}-{!r})".format(x[0].__name__, x[1]),
+    ids=lambda x: "({}-{!r})_fixture".format(x[0].__name__, x[1]),
 )
 def cython_table_items(request):
     """
--- a/pandas/core/base.py
+++ b/pandas/core/base.py
@@ -22,7 +22,8 @@ from pandas.errors import AbstractMethodError
 from pandas.core import common as com, algorithms
 import pandas.core.nanops as nanops
 import pandas._libs.lib as lib
-from pandas.compat.numpy import function as nv
+from pandas.compat.numpy import (function as nv, _np_version_under1p10,
+                                 _np_version_under1p12)
 from pandas.compat import PYPY
 from pandas.util._decorators import (Appender, cache_readonly,
                                      deprecate_kwarg, Substitution)
@@ -191,16 +192,30 @@ class SelectionMixin(object):
         np.all: 'all',
         np.any: 'any',
         np.sum: 'sum',
+        np.nansum: 'sum',
         np.mean: 'mean',
+        np.nanmean: 'mean',
         np.prod: 'prod',
         np.std: 'std',
+        np.nanstd: 'std',
         np.var: 'var',
+        np.nanvar: 'var',
         np.median: 'median',
+        np.nanmedian: 'median',
         np.max: 'max',
+        np.nanmax: 'max',
         np.min: 'min',
+        np.nanmin: 'min',
         np.cumprod: 'cumprod',
-        np.cumsum: 'cumsum'
+        np.cumsum: 'cumsum',
     }
+
+    if not _np_version_under1p10:
+        _cython_table[np.nanprod] = 'prod'
+
+    if not _np_version_under1p12:
+        _cython_table[np.nancumprod] = 'cumprod'
+        _cython_table[np.nancumsum] = 'cumsum'
 
     @property
     def _selection_name(self):
@@ -553,8 +568,8 @@ class SelectionMixin(object):
             result = None
 
         f = self._is_cython_func(arg)
-        if f and not args and not kwargs:
-            return getattr(self, f)(), None
+        if f:
+            return getattr(self, f)(*args, **kwargs), None
 
         # caller can react
         return result, True
--- a/pandas/core/base.py
+++ b/pandas/core/base.py
@@ -568,7 +568,7 @@ class SelectionMixin(object):
             result = None
 
         f = self._is_cython_func(arg)
-        if f:
+        if f is not None:
             return getattr(self, f)(*args, **kwargs), None
 
         # caller can react
--- a/pandas/core/base.py
+++ b/pandas/core/base.py
@@ -331,13 +331,14 @@ class SelectionMixin(object):
 
         raise ValueError("{arg} is an unknown string function".format(arg=arg))
 
-    def _aggregate(self, arg, *args, **kwargs):
+    def _aggregate(self, arg, axis=0, *args, **kwargs):
         """
         provide an implementation for the aggregators
 
         Parameters
         ----------
         arg : string, dict, function
+        axis : int
         *args : args to pass on to the function
         **kwargs : kwargs to pass on to the function
 
@@ -350,17 +351,18 @@ class SelectionMixin(object):
         how can be a string describe the required post-processing, or
         None if not required
         """
+        obj = self if axis == 0 else self.T
         is_aggregator = lambda x: isinstance(x, (list, tuple, dict))
         is_nested_renamer = False
 
         _axis = kwargs.pop('_axis', None)
         if _axis is None:
-            _axis = getattr(self, 'axis', 0)
+            _axis = getattr(obj, 'axis', 0)
         _level = kwargs.pop('_level', None)
 
         if isinstance(arg, compat.string_types):
-            return self._try_aggregate_string_function(arg, *args,
-                                                       **kwargs), None
+            return obj._try_aggregate_string_function(arg, *args,
+                                                      **kwargs), None
 
         if isinstance(arg, dict):
 
@@ -368,7 +370,7 @@ class SelectionMixin(object):
             if _axis != 0:  # pragma: no cover
                 raise ValueError('Can only pass dict with axis=0')
 
-            obj = self._selected_obj
+            selected_obj = obj._selected_obj
 
             def nested_renaming_depr(level=4):
                 # deprecation of nested renaming
@@ -403,16 +405,16 @@ class SelectionMixin(object):
                     if isinstance(v, dict):
                         is_nested_renamer = True
 
-                        if k not in obj.columns:
+                        if k not in selected_obj.columns:
                             msg = ('cannot perform renaming for {key} with a '
                                    'nested dictionary').format(key=k)
                             raise SpecificationError(msg)
                         nested_renaming_depr(4 + (_level or 0))
 
-                    elif isinstance(obj, ABCSeries):
+                    elif isinstance(selected_obj, ABCSeries):
                         nested_renaming_depr()
-                    elif isinstance(obj, ABCDataFrame) and \
-                            k not in obj.columns:
+                    elif isinstance(selected_obj, ABCDataFrame) and \
+                            k not in selected_obj.columns:
                         raise KeyError(
                             "Column '{col}' does not exist!".format(col=k))
 
@@ -422,8 +424,8 @@ class SelectionMixin(object):
                 # deprecation of renaming keys
                 # GH 15931
                 keys = list(compat.iterkeys(arg))
-                if (isinstance(obj, ABCDataFrame) and
-                        len(obj.columns.intersection(keys)) != len(keys)):
+                if (isinstance(selected_obj, ABCDataFrame) and len(
+                        selected_obj.columns.intersection(keys)) != len(keys)):
                     nested_renaming_depr()
 
             from pandas.core.reshape.concat import concat
@@ -432,7 +434,7 @@ class SelectionMixin(object):
                 """
                 aggregate a 1-dim with how
                 """
-                colg = self._gotitem(name, ndim=1, subset=subset)
+                colg = obj._gotitem(name, ndim=1, subset=subset)
                 if colg.ndim != 1:
                     raise SpecificationError("nested dictionary is ambiguous "
                                              "in aggregation")
@@ -442,8 +444,8 @@ class SelectionMixin(object):
                 """
                 aggregate a 2-dim with how
                 """
-                colg = self._gotitem(self._selection, ndim=2,
-                                     subset=obj)
+                colg = obj._gotitem(obj._selection, ndim=2,
+                                    subset=selected_obj)
                 return colg.aggregate(how, _level=None)
 
             def _agg(arg, func):
@@ -473,20 +475,22 @@ class SelectionMixin(object):
 
                 else:
 
-                    if self._selection is not None:
+                    if obj._selection is not None:
                         keys = None
 
             # some selection on the object
-            elif self._selection is not None:
+            elif obj._selection is not None:
 
-                sl = set(self._selection_list)
+                sl = set(obj._selection_list)
 
                 # we are a Series like object,
                 # but may have multiple aggregations
                 if len(sl) == 1:
 
-                    result = _agg(arg, lambda fname,
-                                  agg_how: _agg_1dim(self._selection, agg_how))
+                    result = _agg(
+                        arg,
+                        lambda fname, agg_how: _agg_1dim(
+                            obj._selection, agg_how))
 
                 # we are selecting the same set as we are aggregating
                 elif not len(sl - set(keys)):
@@ -531,7 +535,7 @@ class SelectionMixin(object):
                 return concat([result[k] for k in keys],
                               keys=keys, axis=1), True
 
-            elif isinstance(self, ABCSeries) and is_any_series():
+            elif isinstance(obj, ABCSeries) and is_any_series():
 
                 # we have a dict of Series
                 # return a MI Series
@@ -556,20 +560,20 @@ class SelectionMixin(object):
 
                 # we have a dict of scalars
                 result = Series(result,
-                                name=getattr(self, 'name', None))
+                                name=getattr(obj, 'name', None))
 
             return result, True
         elif is_list_like(arg) and arg not in compat.string_types:
             # we require a list, but not an 'str'
-            return self._aggregate_multiple_funcs(arg,
-                                                  _level=_level,
-                                                  _axis=_axis), None
+            return obj._aggregate_multiple_funcs(arg,
+                                                 _level=_level,
+                                                 _axis=_axis), None
         else:
             result = None
 
-        f = self._is_cython_func(arg)
+        f = obj._is_cython_func(arg)
         if f is not None:
-            return getattr(self, f)(*args, **kwargs), None
+            return getattr(obj, f)(*args, **kwargs), None
 
         # caller can react
         return result, True
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -5818,13 +5818,11 @@ class DataFrame(NDFrame):
     def aggregate(self, func, axis=0, *args, **kwargs):
         axis = self._get_axis_number(axis)
 
-        # TODO: flipped axis
         result = None
-        if axis == 0:
-            try:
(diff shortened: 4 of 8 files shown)


pandas/tests/test_nanops.py
```diff
@@ -994,6 +994,49 @@ def prng(self):
         return np.random.RandomState(1234)
 
 
+@pytest.fixture(params=[
+    pd.Series([1, 2, 3, 4, 5, 6]),
+    pd.DataFrame([[1, 2, 3], [4, 5, 6]])
+])
+def nan_test_object(request):
```
WillAyd: Should we add NA data to these?

pandas/core/base.py
```diff
@@ -553,8 +568,8 @@ def is_any_frame():
             result = None
 
         f = self._is_cython_func(arg)
-        if f and not args and not kwargs:
-            return getattr(self, f)(), None
+        if f:
+            return getattr(self, f)(*args, **kwargs), None
```
WillAyd: Is this equivalent? Wondering if there's any case where providing args / kwargs before would have routed the function to a different place

pandas/tests/test_nanops.py
```diff
@@ -994,6 +994,49 @@ def prng(self):
...
+    (np.sum, np.nansum),
+    (np.mean, np.nanmean),
+    (np.std, np.nanstd),
+    (np.var, np.nanvar),
+    (np.median, np.nanmedian),
+    (np.max, np.nanmax),
+    (np.min, np.nanmin),
+])
+def test_np_nan_functions(standard, nan_method, nan_test_object):
+    tm.assert_almost_equal(nan_test_object.agg(standard),
```
WillAyd: Might be a silly question but can we not use the frame / series equals methods here? Don't think precision is that much of a factor with the fixtures

WillAyd (changes requested):

pandas/core/base.py
topper-123: It's not, but this is better :-). This fixes the bug in that was causing #19629 to fail.

The issue is subtle, but it has to do with a bug in numpy < 1.13 np.nan* functions. Numpy < 1.13 handles e.g. ``np.nanmin(pd_obj)`` incorrectly while it handles ``np.min(pd_obj)`` correctly. Numpy >= 1.13 handles both correctly. See #19753 for another issue regarding the same numpy problem.

Anyway, if ``f`` is a string, you *should* call ``getattr(self, f)(*args, **kwargs)``, so it was perhaps more luck than design that the previous version did work:-)

pandas/tests/test_nanops.py
topper-123: I think these are ok as ``nan_test_object`` can be either a series or a frame. I think the name should be changed though, to e.g. ``series_or_frame``.

pandas/tests/test_nanops.py
topper-123: Added

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -89,3 +89,9 @@ Categorical
 ^^^^^^^^^^^
 
 -
+
+Numeric
+^^^^^^^
+
+- :meth:`~DataFrame.agg` now correctly handles numpy NaN-aware methods like :meth:`numpy.nansum` (:issue:`19629`)
+- :meth:`~DataFrame.agg` now correctly handles built-in methods like ``sum`` when axis=1 (:issue:`19629`)
```
WillAyd: Don't think you meant to add this in this PR?

pandas/core/frame.py
```diff
@@ -5818,13 +5818,12 @@ def _gotitem(self,
...
         axis = self._get_axis_number(axis)
 
-        # TODO: flipped axis
         result = None
-        if axis == 0:
-            try:
-                result, how = self._aggregate(func, axis=0, *args, **kwargs)
-            except TypeError:
-                pass
+        df = self if not axis else self.T
```
WillAyd: Same thing?

pandas/tests/frame/test_apply.py
```diff
@@ -1056,3 +1056,37 @@ def test_non_callable_aggregates(self):
...
 
         assert result == expected
+
+    @pytest.mark.parametrize("test_input", [
+        # is either frame or tuple (frame, args, kwargs),
+        pd.DataFrame([[1, 2], [3, 4]]),
+        pd.DataFrame([[np.nan, 2], [3, 4]]),
+        pd.DataFrame(),
+    ])
+    @pytest.mark.parametrize(
```
WillAyd: Hmm wonder if this would be better as a shared fixture - thoughts?

pandas/tests/frame/test_apply.py
```diff
@@ -1056,3 +1056,37 @@ def test_non_callable_aggregates(self):
...
+        # pd.core.base.SelectionMixin._cython_table give the same result
+        # as the related strings (values) when used in df.agg. Examples:
+        # - ``df.agg(np.nansum)`` should give the same result as
+        #   ``df.agg('sum')``
+        # - ``df.agg(sum)`` should give the same result as ``df.agg('sum')``
+        # etc.
+        # GH21123
+        np_func, str_func = cython_table_items
+
+        if isinstance(test_input, pd.DataFrame):
```
WillAyd: Don't need this type check since the parameters are all DataFrames

WillAyd (changes requested):

doc/source/whatsnew/v0.23.1.txt
topper-123: Well, without those the tests in test_apply don't pass. At the same time, the tests in test_apply are sufficient for testing for this bug, so these two issues are very related...

pandas/core/frame.py
topper-123: Yeah, a bit. ``df.agg`` and ``df.apply`` with ``axis=1`` give in many cases identical output, but not in all, as ``df.apply`` doesn't make the lookup in ``_cython_table``.

AFAIK, transposition is cheap in numpy/pandas, so this is an ok approach?

pandas/tests/frame/test_apply.py
topper-123: You are right. I was preparing for eventual tests where test_input would be tuple([frame, args, kwargs]), but of course looks silly, when I could find suchs tests that make sense. 

I remove that unless someone comes up with a tests that requires args and/or kwargs

topper-123: Also, the assert should now be ``tm.assert_frame_equal``. I'll change that.

topper-123:
The travis failures say "Different tests were collected between gw1 and gw0. The difference is:..."

I don't *think* this has anything to do with my PR, anyone knows?

pandas/core/frame.py
jreback: I would rather you not do this, this is a quite hacky way of handling this, there is a small bug on the lower level i think.

pandas/tests/frame/test_apply.py
jreback: yes pls make this a fixture (in conftest.py)

pandas/tests/series/test_apply.py
```diff
@@ -587,3 +587,34 @@ def test_map_missing_mixed(self, vals, mapping, exp):
...
         tm.assert_series_equal(result, pd.Series(exp))
+
+    @pytest.mark.parametrize("test_input", [
+        # is either series or tuple  (series, args, kwargs),
+        pd.Series([1, 2, 3, 4]),
+        pd.Series([np.nan, 2, 3, 4]),
+        pd.Series(),
+    ])
+    @pytest.mark.parametrize(
+        "cython_table_items",
```
jreback: same

jreback (changes requested):

pandas/core/frame.py
topper-123: I don't think so, as ``df._aggregate`` currently doesn't take an ``axis`` parameter. I could add an ``axis`` parameter to ``df._aggregate`` and do the transposition there instead?

jreback: you can try. this should be handled on a much lower level.

topper-123:
All the failures are of the type 

```
Different tests were collected between gw0 and gw1. The difference is:
--- gw0
+++ gw1
```

Everything passes locally. Don't think this has anything to do with my PR. I can try push again, but this has been every time so far, so don't think that will help...

topper-123:
The failure above has been solved, it was a mistake concerning python3.5 dicts in xdist in ``conftest.py``: python3.5 randomizes dict retrieval, and of course  xdist doesn`t like randomizing work distribution. I`ve made the params parameter in ``cython_table`` deterministic, so it doesnt happen again.

pandas/core/frame.py
topper-123: I've added the ``axis`` parameter to ``df._aggregate``, so this is handled there.

pandas/core/groupby/groupby.py
```diff
@@ -4086,7 +4086,10 @@ def _post_process_cython_aggregate(self, obj):
     def aggregate(self, arg, *args, **kwargs):
 
         _level = kwargs.pop('_level', None)
-        result, how = self._aggregate(arg, _level=_level, *args, **kwargs)
+        _agg_kwargs = kwargs.copy()
```
jreback: you can just list axis as a kwarg

pandas/conftest.py
```diff
@@ -149,3 +149,20 @@ def tz_aware_fixture(request):
...
+@pytest.fixture(
+    # params: Python 3.5 randomizes dict access and xdist doesn't like that
+    # in fixtures. In order to get predetermined values we need to sort
+    # the list deterministically
+    # GH 21123
+    params=list(sorted(pd.core.base.SelectionMixin._cython_table.items(),
+                       key=lambda x: x[0].__name__)),
+    ids=lambda x: "({}-{!r})".format(x[0].__name__, x[1]),
+)
+def cython_table_items(request):
```
jreback: add _fixture to the end of the name

pandas/tests/frame/test_apply.py
```diff
@@ -1056,3 +1056,26 @@ def test_non_callable_aggregates(self):
...
+        # pd.core.base.SelectionMixin._cython_table give the same result
+        # as the related strings (values) when used in df.agg. Examples:
+        # - ``df.agg(np.nansum)`` should give the same result as
+        #   ``df.agg('sum')``
+        # - ``df.agg(sum)`` should give the same result as ``df.agg('sum')``
+        # etc.
+        # GH21123
+        np_func, str_func = cython_table_items
+
+        tm.assert_almost_equal(df.agg(np_func),
```
jreback: use assert_frame_equal it provides stronger guarantees

pandas/tests/series/test_apply.py
```diff
@@ -587,3 +587,23 @@ def test_map_missing_mixed(self, vals, mapping, exp):
...
+        # pd.core.base.SelectionMixin._cython_table give the same result
+        # as the related strings (values), when used in ser.agg. Examples:
+        # - ``ser.agg(np.nansum)`` should give the same result as
+        #   ``ser.agg('sum')``
+        # - ``ser.agg(sum)`` should give the same result as ``ser.agg('sum')``
+        # etc.
+        # GH21123
+        np_func, str_func = cython_table_items
+
+        tm.assert_almost_equal(series.agg(np_func),
```
jreback: use assert_series_equal

pandas/tests/frame/test_apply.py
```diff
@@ -1056,3 +1056,26 @@ def test_non_callable_aggregates(self):
...
+
+    @pytest.mark.parametrize("df", [
+        pd.DataFrame([[1, 2], [3, 4]]),
+        pd.DataFrame([[np.nan, 2], [3, 4]]),
+        pd.DataFrame(),
+    ])
+    def test_agg_function_input(self, df, cython_table_items):
+        # test whether the functions (keys) in
+        # pd.core.base.SelectionMixin._cython_table give the same result
+        # as the related strings (values) when used in df.agg. Examples:
```
jreback: add an example which actually tests (say for sum, nansum) axis=1, IOW contruct the resultant frame

this test doesn't actually tests that axis=1 works, just that it matches with a string (which doesn't have tests itself)

jreback (changes requested):

pandas/tests/frame/test_apply.py
topper-123: Most of these aggregate to series (so ``assert_series_equal``)), but ``cumprod`` and ``cumsum`` is in ``_cython_table``, so in that case a DataFrame is returned.

I could add a conditional, so the more correct assert is used each time.

pandas/conftest.py
topper-123: ok

pandas/core/groupby/groupby.py
topper-123: That breaks a test. The issue if ``axis`` in is in ``_kwargs``, you may get (from the breaking test):

```python
>>> _level, args, _agg_kwargs = None, (80,), {'axis': 0}
>>> self._aggregate(arg, _level=_level, *args, **_agg_kwargs)
TypeError: _aggregate() got multiple values for argument 'axis'
```

topper-123:
I`ve rewritten the tests to now have expected results also.

pandas/core/base.py
```diff
@@ -316,13 +331,14 @@ def _try_aggregate_string_function(self, arg, *args, **kwargs):
 
         raise ValueError("{arg} is an unknown string function".format(arg=arg))
 
-    def _aggregate(self, arg, *args, **kwargs):
+    def _aggregate(self, arg, axis=0, *args, **kwargs):
```
WillAyd: I still think this should be a separate PR. I know you mentioned there was some interweaving of test dependency, but I feel like we are injecting this keyword in here without any regard to existing test coverage for axis=1.

pandas/tests/frame/test_apply.py
```diff
@@ -1056,3 +1056,66 @@ def test_non_callable_aggregates(self):
...
+        }],
+    ])
+    @pytest.mark.parametrize("axis", [0, 1], ids=lambda x: "axis {}".format(x))
+    def test_agg_function_input(self, cython_table_items, inputs, axis):
+        # GH21123
+        np_func, str_func = cython_table_items
+        df = inputs[0]
+        expected = inputs[1][str_func]
+
+        if isinstance(expected, type) and issubclass(expected, Exception):
```
WillAyd: Anything that raises should be done in a separate test, i.e. `test_agg_function_input_raises`

pandas/tests/frame/test_apply.py
```diff
@@ -1056,3 +1056,66 @@ def test_non_callable_aggregates(self):
...
+            with pytest.raises(expected):
+                # e.g. DataFrame(['a b'.split()]).cumprod() will raise
+                df.agg(np_func, axis=axis)
+                df.agg(str_func, axis=axis)
+        elif str_func in ('cumprod', 'cumsum'):
+            tm.assert_frame_equal(df.agg(np_func, axis=axis), expected)
```
WillAyd: For readability / consistency with other tests create a variable called result and assign to it before the call to `assert_frame_equal`

pandas/tests/series/test_apply.py
```diff
@@ -331,6 +331,67 @@ def test_non_callable_aggregates(self):
...
+
+        if isinstance(expected, type) and issubclass(expected, Exception):
+            with pytest.raises(expected):
+                series.agg(np_func)
+                series.agg(str_func)
+        elif str_func in ('cumprod', 'cumsum'):
+            tm.assert_series_equal(series.agg(np_func), expected)
+            tm.assert_series_equal(series.agg(str_func), expected)
+        else:
+            tm.assert_almost_equal(series.agg(np_func), expected)
```
WillAyd: Should be `assert_series_equal` no?

WillAyd (changes requested):

pandas/tests/series/test_apply.py
topper-123: ``series.agg(np_func)`` resturns a scalar. I could use ``assert np.isclose(...)``?

pandas/tests/frame/test_apply.py
topper-123: I agree in principle, but this test iterates over all items in ``_cython_table``, of which some will fail on some inputs.

So I'd  have to construct the tests quite a bit differently and probably the fixture in conftest.py couldn't be used (because it returns all combinations and I now have to select the relevant ones for each test method). So something like:

```
(builtins.sum, 'sum': 0),
(np.sum, 'sum', 0),
(np.nansum: 'sum', 0),
etc...
```

which will be very inelegant and repetitive IMO. Is it not possible to bend this rule on this one (or give hint on how to do it elegantly)?...

pandas/core/base.py
topper-123: I can look into making separate PR. That will have to be pulled in before this one, so the tests of this PR won't break.

pandas/tests/frame/test_apply.py
topper-123: I've uploaded a changed version.

pandas/tests/frame/test_apply.py
WillAyd: Hmm OK understood. May make sense as an exception then - don't have anything off the top of my head to improve but will think more about it

pandas/core/groupby/groupby.py
WillAyd: Not sure I understand this - what test is breaking? Perhaps the test is configured incorrectly?

pandas/core/frame.py
```diff
@@ -5818,13 +5818,11 @@ def _gotitem(self,
...
-        if axis == 0:
-            try:
-                result, how = self._aggregate(func, axis=0, *args, **kwargs)
-            except TypeError:
-                pass
+        try:
+            result, how = self._aggregate(func, axis=axis, *args, **kwargs)
+        except TypeError:
+            pass
         if result is None:
```
WillAyd: Related to the axis change, do we still hit this condition?

WillAyd (changes requested):

pandas/core/groupby/groupby.py
topper-123: Both ``80`` and ``0`` may be the value for parameter ``axis``. The signature is ``def _aggregate(self, arg, axis=0, *args, **kwargs)``, so the second unnamed argument (``80``) will be considered to be ``axis``, but this clashes with the parameter in kwargs (``{'axis': 0}``), causing the exception.

To avoid this we'd prefer the signature to be ``def _aggregate(self, arg, *args, axis=0, **kwargs)``, but this syntax is only supported in Python3...

pandas/core/frame.py
topper-123: The axis change is related to #21134. So if I move that to a separate PR, this will move too.

pandas/core/base.py
WillAyd: If timing is a concern you can also xfail the axis=1 tests. Rebase thereafter would be minor

pandas/core/groupby/groupby.py
WillAyd: Where is that test located?

topper-123: ``pandas\tests\groupby\test_groupby.py::test_pass_args_kwargs``

pandas/core/base.py
topper-123: This has now been created as #21222

jreback:
as @WillAyd indicates, can you split this up into a cython table PR and on top of that the agg fixes? 

first should be straightforward and we can get it quickly. pls put tests, changes and whatsnew for that one (in this PR is fine), and issue another PR for other changes.

pandas/conftest.py
```diff
@@ -149,3 +149,20 @@ def tz_aware_fixture(request):
     Fixture for trying explicit timezones: {0}
     """
     return request.param
+
+
+@pytest.fixture(
+    # params: Python 3.5 randomizes dict access and xdist doesn't like that
+    # in fixtures. In order to get predetermined values we need to sort
```
jreback: so would like to have this fixture in your other PR

topper-123:
Closing in favor of #22109.

PR #21122, opened 2018-05-18
Resolution docstring

chalmerlowe:
- [ ] closes #xxxx
- [ ] tests added / passed
- [X] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] whatsnew entry

```
################################################################################
################### Docstring (pandas.Timedelta.resolution)  ###################
################################################################################

Return a string representing the lowest (i.e. smallest) time resolution.

Each timedelta has a defined resolution that represents the lowest OR
most granular level of precision. Each level of resolution is
represented by a short string as defined below:

============ ============
Resolution   Return value
============ ============
Days         ``'D'``
Hours        ``'H'``
Minutes      ``'T'``
Seconds      ``'S'``
Milliseconds ``'L'``
Microseconds ``'U'``
Nanoseconds  ``'N'``
============ ============

Returns
-------
str
    Time resolution.

Examples
--------
**Using string input**

>>> td = pd.Timedelta('1 days 2 min 3 us 42 ns')
>>> td.resolution
'N'

>>> td = pd.Timedelta('1 days 2 min 3 us')
>>> td.resolution
'U'

>>> td = pd.Timedelta('2 min 3 s')
>>> td.resolution
'S'

**Using integer input**

>>> td = pd.Timedelta(36, unit='us')
>>> td.resolution
'U'

################################################################################
################################## Validation ##################################
################################################################################

Errors found:
	See Also section not found
```

PR #21135, opened 2018-05-19
DOC: Improve docstring of Timedelta.delta

chalmerlowe:
- [ ] closes #xxxx
- [ ] tests added / passed
- [X] passes `git diff upstream/master -u -- "*.pyx" | flake8 --diff`
- [ ] whatsnew entry

```
###################### Docstring (pandas.Timedelta.delta) ######################

Return the timedelta in nanoseconds (ns), for internal compatibility.

Returns
-------
int
    Timedelta in nanoseconds.

Examples
--------
**Using string input**

>>> td = pd.Timedelta('1 days 42 ns')
>>> td.delta
86400000000042

>>> td = pd.Timedelta('3 s')
>>> td.delta
3000000000

>>> td = pd.Timedelta('3 ms 5 us')
>>> td.delta
3005000

**Using integer input**

>>> td = pd.Timedelta(42, unit='ns')
>>> td.delta
42

Errors found:
	No extended summary found
	See Also section not found
```

### DIFF
--- a/pandas/_libs/tslibs/timedeltas.pyx
+++ b/pandas/_libs/tslibs/timedeltas.pyx
@@ -803,7 +803,6 @@ cdef class _Timedelta(timedelta):
         Timedelta.components : Return all attributes with assigned values (i.e.
             days, hours, minutes, seconds, milliseconds, microseconds, nanoseconds).
 
-
         Examples
         --------
         **Using string input**
@@ -812,12 +811,11 @@ cdef class _Timedelta(timedelta):
         >>> td.nanoseconds
         42
 
-        **Using integer inputs**
+        **Using integer input**
 
         >>> td = pd.Timedelta(42, unit='ns')
         >>> td.nanoseconds
         42 
-
         """
         self._ensure_components()
         return self._ns
--- a/pandas/_libs/tslibs/timedeltas.pyx
+++ b/pandas/_libs/tslibs/timedeltas.pyx
@@ -800,8 +800,9 @@ cdef class _Timedelta(timedelta):
 
         See Also
         --------
-        Timedelta.components : Return all attributes with assigned values (i.e.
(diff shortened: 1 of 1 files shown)


pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -770,7 +799,48 @@ cdef class _Timedelta(timedelta):
 
     @property
     def resolution(self):
-        """ return a string representing the lowest resolution that we have """
+        """
```
WillAyd: Isn't this already a part of #21122?

WillAyd (changes requested):

pandas/_libs/tslibs/timedeltas.pyx
chalmerlowe: @WillAyd 

Yes...
I am not sure why that is in there.
I created separate branches for each of the changes I am proposing to:

* delta
* resolution
* nanoseconds

etcetera, but the commit stream above seems to be sucking all of them into this PR.
super confused.

WillAyd: Hmm this should just reflect what you have locally - any chance you merged your branches there accidentally?

jreback: when creating a new branch, do it off of upstream/master (or if you keep a local master then off of this). each branch should be independent.

you will need to revert the resolution part of this change (as its in the other PR). or can just close that one and do it here is ok too)

jreback (changes requested):

chalmerlowe:
i did some looking at how to `git` rid of (pun intended) a bunch of the changes that are cluttering up the works.
think I am gonna just close this, try to submit a new pull request starting with a clean slate.
See PR #21138

PR #21139, opened 2018-05-19
CLN: Refactor open into context manager in io tests

alysivji:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

Building on #21105 by @WillAyd, I cleaned up some more `open` statements.

### DIFF
--- a/pandas/tests/io/test_packers.py
+++ b/pandas/tests/io/test_packers.py
@@ -128,9 +128,8 @@ class TestAPI(TestPackers):
         with ensure_clean(self.path) as p:
 
             s = df.to_msgpack()
-            fh = open(p, 'wb')
-            fh.write(s)
-            fh.close()
+            with open(p, 'wb') as fh:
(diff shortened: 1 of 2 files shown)


pandas/tests/io/test_packers.py
```diff
@@ -128,9 +128,8 @@ def test_string_io(self):
         with ensure_clean(self.path) as p:
 
             s = df.to_msgpack()
-            fh = open(p, 'wb')
-            fh.write(s)
-            fh.close()
+            with open(p, 'wb') as fh:
```
WillAyd: Do we have a name used more often than not for the file handle? I notice in this change we have two different ones. The last file I touched they were all `as f:` - if that's used more often than not then let's update these to stay consistent

gfyoung (approved):

jreback:
thanks!

PR #21144, opened 2018-05-20
BUG: set keyword argument so zipfile actually compresses

minggli:
- [x] closes https://github.com/pandas-dev/pandas/issues/17778
- [x] tests added / passed
- [x] whatsnew entry

zipfile.ZipFile has default compression mode `zipfile.ZIP_STORED`. It creates an uncompressed archive member. Whilst it doesn't cause issue, it is a strange default to have given users would want to compress files.

In order for zip compression to actually reduce file size, keyword argument `compression=zipfile.ZIP_DEFLATED` is added.

### DIFF
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -15,6 +15,8 @@ and bug fixes. We recommend that all users upgrade to this version.
 New features
 ~~~~~~~~~~~~
 
+- :meth:`Index.droplevel` is now implemented also for flat indexes, for compatibility with MultiIndex (:issue:`21115`)
+
 
 .. _whatsnew_0231.deprecations:
 
@@ -44,6 +46,8 @@ Documentation Changes
 Bug Fixes
 ~~~~~~~~~
 
+- tab completion on :class:`Index` in IPython no longer outputs deprecation warnings (:issue:`21125`)
+
 Groupby/Resample/Rolling
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -53,7 +57,10 @@ Strings
 ^^^^^^^
 
 - Bug in :meth:`Series.str.replace()` where the method throws `TypeError` on Python 3.5.2 (:issue: `21078`)
--
+
+Timedelta
+^^^^^^^^^
+- Bug in :class:`Timedelta`: where passing a float with a unit would prematurely round the float precision (:issue: `14156`)
 
 Categorical
 ^^^^^^^^^^^
@@ -70,12 +77,14 @@ Indexing
 ^^^^^^^^
 
 - Bug in :meth:`Series.reset_index` where appropriate error was not raised with an invalid level name (:issue:`20925`)
+- Bug in :func:`interval_range` when ``start``/``periods`` or ``end``/``periods`` are specified with float ``start`` or ``end`` (:issue:`21161`)
 -
 
 I/O
 ^^^
 
 - Bug in :class:`pandas.io.common.BytesZipFile` where zip compression produces uncompressed zip archive (:issue:`17778`)
+- Bug in :meth:`DataFrame.to_stata` which prevented exporting DataFrames to buffers and most file-like objects (:issue:`21041`)
 -
 
 Plotting
@@ -87,10 +96,5 @@ Plotting
 Reshaping
 ^^^^^^^^^
 
--
--
-
-Categorical
-^^^^^^^^^^^
-
+- Bug in :func:`concat` where error was raised in concatenating :class:`Series` with numpy scalar and tuple names (:issue:`21015`)
 -
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -75,7 +75,7 @@ Indexing
 I/O
 ^^^
 
--
+- Bug in :class:`pandas.io.common.BytesZipFile` where zip compression produces uncompressed zip archive (:issue:`17778`)
 -
 
 Plotting
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -83,7 +83,7 @@ Indexing
 I/O
 ^^^
 
-- Bug in :class:`pandas.io.common.BytesZipFile` where zip compression produces uncompressed zip archive (:issue:`17778`)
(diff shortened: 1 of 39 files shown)


pandas/io/common.py
```diff
@@ -437,10 +437,10 @@ class BytesZipFile(ZipFile, BytesIO):
     bytes strings into a member of the archive.
     """
     # GH 17778
-    def __init__(self, file, mode='r', **kwargs):
+    def __init__(self, file, mode, compression=zipfile.ZIP_DEFLATED, **kwargs):
         if mode in ['wb', 'rb']:
             mode = mode.replace('b', '')
-        super(BytesZipFile, self).__init__(file, mode, **kwargs)
+        super(BytesZipFile, self).__init__(file, mode, compression, **kwargs)
```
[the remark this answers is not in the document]
jreback: no this is a bug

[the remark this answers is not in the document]
gfyoung: Fair enough, though tests and `whatsnew` are still needed.

jreback (changes requested):
pls add a whatsnew & a test (not sure what that should look like)

pandas/io/common.py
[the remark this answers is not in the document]
minggli: thanks. added whatsnew and tests.

pandas/tests/frame/test_to_csv.py
```diff
@@ -943,6 +943,22 @@ def test_to_csv_compression(self, compression):
             with tm.decompress_file(filename, compression) as fh:
                 assert_frame_equal(df, read_csv(fh, index_col=0))
 
+    def test_to_csv_compression_size(self, compression):
```
WillAyd: Since these are all the same tests I think it makes more sense to put in `test_common` and parametrize for the different writers, rather than splitting out across the various modules

WillAyd (changes requested):

pandas/tests/frame/test_to_csv.py
minggli: done.

minggli:
@WillAyd @jreback @gfyoung changes implemented, comments welcome. 👍

pandas/tests/test_common.py
```diff
@@ -222,3 +224,22 @@ def test_standardize_mapping():
...
+def test_compression_size(method, compression):
+
+    df = pd.concat(100 * [DataFrame([[0.123456, 0.234567, 0.567567],
+                                     [12.32112, 123123.2, 321321.2]],
+                                    columns=['X', 'Y', 'Z'])],
+                   ignore_index=True)
+    s = df.iloc[:, 0]
+
+    with tm.ensure_clean() as filename:
+        for obj in [df, s]:
```
WillAyd: Can even parametrize the Series and Frame instead of a loop

pandas/tests/test_common.py
```diff
@@ -222,3 +224,22 @@ def test_standardize_mapping():
...
+            getattr(obj, method)(filename, compression=compression)
+            file_size = os.path.getsize(filename)
+            getattr(obj, method)(filename, compression=None)
+            uncompressed_file_size = os.path.getsize(filename)
+            if compression:
```
WillAyd: Shouldn't need this conditional

WillAyd (changes requested):

pandas/tests/test_common.py
minggli: ok

pandas/tests/test_common.py
minggli: done.

pandas/tests/test_common.py
```diff
@@ -222,3 +224,22 @@ def test_standardize_mapping():
 
     dd = collections.defaultdict(list)
     assert isinstance(com.standardize_mapping(dd), partial)
+
+
+@pytest.mark.parametrize('frame', [
+    pd.concat(100 * [DataFrame([[0.123456, 0.234567, 0.567567],
```
WillAyd: Instead of using concat just multiply your list of lists by 100 within the constructor - will be more performant and idiomatic

pandas/tests/test_common.py
```diff
@@ -222,3 +224,22 @@ def test_standardize_mapping():
 
     dd = collections.defaultdict(list)
     assert isinstance(com.standardize_mapping(dd), partial)
+
+
+@pytest.mark.parametrize('frame', [
```
WillAyd: Since this is either a frame or a series don't use `frame` as the variable name, as that's slightly confusing in case of the former being passed. `obj` should be fine

WillAyd (changes requested):

pandas/tests/test_common.py
minggli: good point.

WillAyd (approved):
lgtm - will see what others say but thanks for the PR!

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -83,6 +83,7 @@ Indexing
 I/O
 ^^^
 
+- Bug in :class:`pandas.io.common.BytesZipFile` where zip compression produces uncompressed zip archive (:issue:`17778`)
```
jreback: can you reference 21144 (as well is fine) as that other issue was closed for 0.23.0

pandas/tests/test_common.py
```diff
@@ -222,3 +223,21 @@ def test_standardize_mapping():
...
+                     [12.32112, 123123.2, 321321.2]],
+              columns=['X', 'Y', 'Z']),
+    Series(100 * [0.123456, 0.234567, 0.567567], name='X')])
+@pytest.mark.parametrize('method', ['to_pickle', 'to_json', 'to_csv'])
+def test_compression_size(obj, method, compression):
+    if not compression:
+        pytest.skip("only test compression case.")
+
+    with tm.ensure_clean() as filename:
+        getattr(obj, method)(filename, compression=compression)
```
jreback: does this fail under 0.23.0? (and works here clearly)

doc/source/whatsnew/v0.23.1.txt
minggli: Hi @jreback @WillAyd , tested 21144 on older version of pandas 0.20 on windows and same issue occurred. I think it's an existing issue unrelated to this PR. This PR only address zip compression and doesn't touch gzip at all.

jreback (approved):

jreback:
thanks @minggli !

minggli:
happy to help!

PR #21160, opened 2018-05-22
ENH: Integer NA Extension Array 

### LINKED ISSUE #20700: API: integer Extension Array
xref #8640

Could easily imagine an ``ExtensionArray`` which uses as an implementation a numpy array of the appropriate dtype and a bitmask in order to fully support Integer NA across the board. I don't think this would be too hard. As a bonus, would be zero-copy compat with pyarrow impl.

- [ ] Integer
- [ ] UnsignedInteger
- [ ] Boolean

making these the actual default (e.g. when integers are inferred with or w/o nulls) might be non-trivial, but let's implement first. These would give rise to a hierarchy of dtypes, e.g. ``IntegerDtype``, ``Int8Dtype``

### LINKED ISSUE #20747: ExtensionArray construction with given dtype (sort of shallow_copy?)
Many of the extension arrays tests are skipped for `Categorical` because the reconstruction of the expected result does not preserve the categoricals (so kind of the "metadata" of the dtype).

For example:

https://github.com/pandas-dev/pandas/blob/78fee04e95e3c53c83c938285580c39e7761ddc8/pandas/tests/extension/category/test_categorical.py#L70-L72

because in the actual test, the expected result is constructed from a list:

https://github.com/pandas-dev/pandas/blob/78fee04e95e3c53c83c938285580c39e7761ddc8/pandas/tests/extension/base/reshaping.py#L47-L50

(still with `type(data)(..)`, but replacing that with `data._constructor_from_sequence(..)` in https://github.com/pandas-dev/pandas/pull/20746).

This is kind of a recurrent pattern, so that might indicate we should find a solution for this?

So do we want a canonical way in the extension array interface to construct an ExtensionArray that has a certain dtype?

Possible solution is to add a `dtype` keyword to `_constructor_from_sequence`

jreback:
closes #20700 
closes #20747

```
In [1]: df = pd.DataFrame({
'A': pd.Series([1, 2, np.nan], dtype='Int64'), 
'B': pd.Series([1, np.nan, 3], dtype='UInt8'), 
'C': [1, 2, 3]})

In [2]: df
Out[2]: 
     A    B  C
0    1    1  1
1    2  NaN  2
2  NaN    3  3

In [3]: df.dtypes
Out[3]: 
A    Int64
B    UInt8
C    int64
dtype: object

In [4]: df.A + df.B
Out[4]: 
0      2
1    NaN
2    NaN
dtype: Int64

In [5]: df.A + df.C
Out[5]: 
0      2
1      4
2    NaN
dtype: Int64

In [6]: (df.A + df.C) * 3
Out[6]: 
0      6
1     12
2    NaN
dtype: Int64
In [7]: (df.A + df.C) * 3 == 1
Out[7]: 
0    False
1    False
2    False
dtype: bool

In [8]: (df.A + df.C) * 3 == 12
Out[8]: 
0    False
1     True
2    False
dtype: bool
```


jreback:
cc @TomAugspurger @jorisvandenbossche @cpcloud @jbrockmendel @Dr-Irv

pandas/core/arrays/base.py
```diff
@@ -567,6 +581,34 @@ def copy(self, deep=False):
         """
         raise AbstractMethodError(self)
 
+    def append(self, other):
+        """
+        Append a collection of Arrays together
+
+        Parameters
+        ----------
+        other : ExtenionArray or list/tuple of ExtenionArrays
```
jbrockmendel: typo Extenion-->Extension

pandas/tests/extension/json/array.py
```diff
@@ -170,6 +170,9 @@ def _values_for_argsort(self):
         return np.array(frozen, dtype=object)[1:]
 
 
+JSONDtype.array_type = JSONArray
```
jbrockmendel: Rather than pin this attribute ex-post, could it be a property defined directly in JSONDtype?

jbrockmendel:
There's a lot here, so preliminary thoughts based on a quick pass:

- Getting integer NA implemented will be a big win!
- If there are bite-sized pieces that can be broken off, that will make life easier
- For a bitmask is there a good heuristic for when to use sparse vs dense?  "Too early to worry about that" is a reasonable answer.

- For ops, this skips several of the previously-discussed steps in transitioning to One True Implementation.
    - IIRC there are inconsistencies in Series vs Index division by zero behavior for numeric types.  I'd prefer to get these aligned before transitioning the dispatch logic.
    - Should IntegerArray also back `Int64Index` and `UInt64Index`?

Dr-Irv:
@jreback Here is my take on the operators implementation, based on what I did in #20889.  Looking at your implementation in `pandas/core/arrays/integer.py`, you've defined the operators on the arrays themselves.  I did set things up in #20889 so that if someone did that with `ExtensionArray`, there is a dispatch to those methods.  The important thing in  `pandas/core/ops.py`  is that you can't assume anything about the underlying implementation of `ExtensionArray`, and it's possible that your changes to `ops.py` might be assuming that `ExtensionArray` is using a `numpy` implementation under the hood. I think (but I'm not 100% sure) that if you take what I did in `pandas/core/ops.py` with what you've done here, things would work correctly.

TomAugspurger (commented):
Only partway through. Will try to take a closer look later today.

pandas/core/dtypes/base.py
```diff
@@ -156,6 +156,12 @@ def name(self):
         """
         raise AbstractMethodError(self)
 
+    @property
+    def array_type(self):
```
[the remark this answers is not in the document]
jreback: this is not a problem, you simply have multiple arrays and multiple dtypes (as I did here).

jreback:
I split out #21185 with the changes to EA. To answer some of @jbrockmendel questions

> Getting integer NA implemented will be a big win!
yes

> If there are bite-sized pieces that can be broken off, that will make life easier
see above

> For a bitmask is there a good heuristic for when to use sparse vs dense? "Too early to worry about that" is a reasonable answer.

I think this would overcomplicate things.

> For ops, this skips several of the previously-discussed steps in transitioning to One True Implementation.

how so, the implementation is barely changed here

> IIRC there are inconsistencies in Series vs Index division by zero behavior for numeric types. I'd prefer to get these aligned before transitioning the dispatch logic.

yes this is tricky, we are matching on the current impl. if we were to change this we should adjust.

> Should IntegerArray also back Int64Index and UInt64Index?

sure, but we don't really support the concept of an ExtensionIndex yet.

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,489 @@
...
+                values.dtype))
+
+    elif not (is_integer_dtype(values) or is_float_dtype(values)):
+        raise TypeError("{} cannot be converted to an IntegerDtype".format(
+            values.dtype))
+
+    if mask is None:
+        mask = isna(values)
+    else:
+        assert len(mask) == len(values)
```
jbrockmendel: should this raise a ValueError instead of AssertionError?

pandas/core/ops.py
```diff
@@ -1003,8 +1003,18 @@ def _arith_method_SERIES(cls, op, special):
...
+        # handle extension array ops
+        # TODO(extension)
+        # the ops *between* non-same-type extension arrays are not
+        # very well defined
+        if (is_extension_array_dtype(x) or is_extension_array_dtype(y)):
+            if (op_name.startswith('__r') and not
+                    is_extension_array_dtype(y) and not
+                    is_scalar(y)):
+                y = x.__class__._from_sequence(y)
+            return op(x, y)
```
jbrockmendel: overlap in comment in #21191 since IntegerArray has an `__init__` that I'm pretty sure can take `left` here (actually below in `wrapper`), I'm hopeful that in this case we can save a lot of trouble by using `dispatch_to_index_op` directly (passing the EA subclass rather than an Index subclass)

pandas/core/arrays/integer.py
jreback: no, this is an internal construction error, need to satisfy the input guarantees

Dr-Irv (commented):
In pandas/core/arrays/base.py, I think you are missing `__rmod__` in the list of arithmetic operators

jreback:
thanks @Dr-Irv added to #21191

jreback:
ok finally rebased on top of current ops impl. @jbrockmendel a couple of comments I have added to make vectorized ops work correctly.

pandas/core/ops.py
```diff
@@ -135,7 +135,20 @@ def rfloordiv(left, right):
 
 
 def rmod(left, right):
-    return right % left
+    # cannot use % as an operand becaue this
```
jreback: cc @jbrockmendel

pandas/core/ops.py
```diff
@@ -1002,16 +1015,37 @@ def dispatch_to_extension_op(op, left, right):
 
     # The op calls will raise TypeError if the op is not defined
     # on the ExtensionArray
+    # TODO(jreback)
+    # we need to listify to avoid ndarray, or non-same-type extension array
+    # dispatching
+
     if is_extension_array_dtype(left):
-        res_values = op(left.values, right)
+
```
jreback: cc @jbrockmendel 

not real happy here because ndarrays take over the ops, but not sure what else to do

pandas/tests/extension/base/methods.py
```diff
@@ -126,11 +127,26 @@ def test_combine_add(self, data_repeated):
         orig_data1, orig_data2 = data_repeated(2)
         s1 = pd.Series(orig_data1)
         s2 = pd.Series(orig_data2)
-        result = s1.combine(s2, lambda x1, x2: x1 + x2)
+
+        # fundamentally this is not a great operation
```
jreback: FYI @Dr-Irv

pandas/core/ops.py
jbrockmendel: Good catch; that's an annoying corner case.  Would it be simpler to just raise if `right` is stringy?

Is there a test for this?

Typo becaue --> because

pandas/core/ops.py
```diff
@@ -1002,16 +1015,37 @@ def dispatch_to_extension_op(op, left, right):
...
+        new_left = list(left.values)
+        new_right = right
+
+    res_values = op(new_left, new_right)
     res_name = get_op_result_name(left, right)
-    return left._constructor(res_values, index=left.index,
-                             name=res_name)
+
+    # divmod returns a tuple
+    def compute_result(res_values):
```
jbrockmendel: I think `_construct_result` and `_construct_divmod_result` exist for this purpose.

pandas/core/ops.py
jbrockmendel: Yah definitely not my favorite part of this diff.  Making sure I grok: this is to make it dispatched to reversed ops correctly?

Instead of trying to "trick" it into calling the reversed ops via `NotImplemented`, what if we defined `reversed_ops = {operator.add: radd, operator.sub: rsub, ...}` and then made the calls directly?  We'd have to be careful about `RecursionError`s, but it _might_ be cleaner.

pandas/core/ops.py
```diff
@@ -1049,6 +1082,20 @@ def na_op(x, y):
...
+        try coercion to object type if the native types are not compatible
+
+        Parameters
+        ----------
+        lvalues : array-like
+        rvalues : array-like
+
+        Raises
+        ------
+        invalid operation raises TypeError
```
jbrockmendel: nit: `TypeError: explanation` I think

pandas/core/ops.py
```diff
@@ -1049,6 +1082,20 @@ def na_op(x, y):
         return result
 
     def safe_na_op(lvalues, rvalues):
```
jbrockmendel: Ideally I'd like to get rid of this altogether to de-nest the closures, catch a more specific exception, and do this directly in `na_op`

pandas/core/ops.py
```diff
@@ -1059,14 +1106,21 @@ def safe_na_op(lvalues, rvalues):
...
         if isinstance(right, ABCDataFrame):
             return NotImplemented
 
         left, right = _align_method_SERIES(left, right)
         res_name = get_op_result_name(left, right)
 
-        if is_datetime64_dtype(left) or is_datetime64tz_dtype(left):
+        if is_categorical_dtype(left):
+            raise TypeError("{typ} cannot perform the operation "
+                            "{op}".format(typ=type(left).__name__, op=str_rep))
```
jbrockmendel: Not necessarily part of this PR, but this could just dispatch to the `Categorical` and it would raise `TypeError` correctly.  The exception message would be slightly changed.

pandas/core/ops.py
```diff
@@ -1078,15 +1132,6 @@ def wrapper(left, right):
                                     index=left.index, name=res_name,
                                     dtype=result.dtype)
 
-        elif is_categorical_dtype(left):
-            raise TypeError("{typ} cannot perform the operation "
-                            "{op}".format(typ=type(left).__name__, op=str_rep))
-
-        elif (is_extension_array_dtype(left) or
-              (is_extension_array_dtype(right) and
-               not is_categorical_dtype(right))):
```
jbrockmendel: Is it obvious why the `is extension_array_dtype(right) and not is_categorical_dtype(right)` is dropping the second part of that check?

pandas/core/ops.py
```diff
@@ -1158,6 +1203,9 @@ def _comp_method_SERIES(cls, op, special):
     masker = _gen_eval_kwargs(op_name).get('masker', False)
 
     def na_op(x, y):
+        # TODO:
+        # should have guarantess on what x, y can be type-wise
+        # Extension Dtypes are not called here
```
jbrockmendel: Yah, ideally we'll do that below in `wrapper` and de-nest this closure too

pandas/core/ops.py
jreback: i changed to catch up higher (where we already dispatch on categorical types)

pandas/core/ops.py
jreback: I think that dispatch_to_extension_op should subsume any is_categorial dispatching directly (but for later)

pandas/core/ops.py
jreback: sure, separate PR :>

pandas/core/ops.py
jreback: done

pandas/core/ops.py
jreback: done

pandas/core/ops.py
jreback: yeah this prob could use some refactoring, but left for after everything works

pandas/core/ops.py
jreback: no existing tests AFAICT, just some ones I added to test non-senscial ops.

pandas/tests/extension/base/methods.py
Dr-Irv: @jreback Not sure what you mean.  I added the test to see that `Series.combine()` works for EA.

Dr-Irv (changes requested):
In `pandas/core/arrays/integer.py` is the function `make_data()`.  This follows the pattern in the EA *test* classes decimal, json, etc., but it seems odd to see it there. So I think the `make_data()` function should really be in the test file `pandas/tests/extension/integer/test_integer.py` (and maybe make similar changes in tests/extension/decimal and tests/extension/json as well).

pandas/core/ops.py
Dr-Irv: @jreback If you're making the change in the wrapper in `_arith_method_SERIES()` to change when you test for categorical and EA types, then you may need similar changes in the wrapper in `_comp_method_SERIES()`.

pandas/tests/extension/base/methods.py
jreback: yes i may do that

jreback:
@Dr-Irv moved ``make_data()``

Dr-Irv:
@jreback looked at the `make_data()` move and it looks good

jreback:
going to add reduce & groupby operations in next PR. so pls have a look @TomAugspurger @jorisvandenbossche @jbrockmendel @WillAyd

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,563 @@
...
+        res = u("%s(%s%s)") % (klass, data, prepr)
+
+        return res
+
+    @property
+    def nbytes(self):
+        return self.data.nbytes + self.mask.nbytes
+
+    def isna(self):
+        return self.mask
```
jbrockmendel: Should `pd.isna` dispatch to this?  (or does it already?)

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,563 @@
...
+        return np.nan
+
+    @classmethod
+    def _concat_same_type(cls, to_concat):
+        data = np.concatenate([x.data for x in to_concat])
+        mask = np.concatenate([x.mask for x in to_concat])
+        return cls(data, mask=mask, dtype=to_concat[0].dtype)
+
+    def astype(self, dtype, copy=True):
+        """Cast to a NumPy array with 'dtype'.
```
jbrockmendel: Should docstring reflect cases in which another EA is returned?  (currently just IntegerDtype)

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,563 @@
...
+            maybe = maybe_downcast_to_dtype(nonans, self.dtype.numpy_dtype)
+            if not is_integer_dtype(maybe):
+                result[mask] = np.nan
+                return result
+
+        return type(self)(result, mask=mask, dtype=self.dtype, copy=False)
+
+    @classmethod
+    def _create_arithmetic_method(cls, op):
+        def integer_arithmetic_method(self, other):
```
jbrockmendel: In `core.ops` many things first attempt to dispatch to `expressions.evaluate` for `numexpr` boost.  Is that worth implementing here?  (this also comes up in the ongoing effort to make sure Index and Series arith ops behave identically)

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,563 @@
...
+
+            op_name = op.__name__
+            mask = None
+            if isinstance(other, (ABCSeries, ABCIndexClass)):
+                other = getattr(other, 'values', other)
+
+            if isinstance(other, IntegerArray):
+                other, mask = other.data, other.mask
+            elif getattr(other, 'ndim', 0) > 1:
+                raise TypeError("can only perform ops with 1-d structures")
```
jbrockmendel: return `NotImplemented` for `DataFrame`?

pandas/core/arrays/integer.py
jreback: yes it already does, see ``pandas.core.dtypes.missing._isndarray_like``

pandas/core/arrays/integer.py
jreback: done

pandas/core/arrays/integer.py
jreback: for EA I think directly to here is best, the the EA author can choose what to do. If we want to expose some helpers in core/ops would be great though

pandas/core/arrays/integer.py
jreback: fixed

pandas/core/arrays/base.py
```diff
@@ -88,14 +88,16 @@ class ExtensionArray(object):
...
+    def _from_sequence(cls, scalars, dtype=None, copy=False):
         """Construct a new ExtensionArray from a sequence of scalars.
 
         Parameters
         ----------
         scalars : Sequence
             Each element will be an instance of the scalar type for this
             array, ``cls.dtype.type``.
+        dtype : Dtype, optional
+            consruct for this particular dtype
```
jorisvandenbossche: This needs to be a dtype compatible with the ExtensionArray?

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,564 @@
...
+    mask : boolean 1D array, optional
+    copy : boolean, default False
+        if True, copy the input
+
+    Returns
+    -------
+    tuple of (values, mask)
+    """
+
+    if isinstance(values, IntegerArray):
```
jorisvandenbossche: do we need to check the dtype argument in this case as well?

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,564 @@
...
+    We represent an IntegerArray with 2 numpy arrays
+    - data: contains a numpy integer array of the appropriate dtype
+    - mask: a boolean array holding a mask on the data, False is missing
+    """
+
+    @cache_readonly
+    def dtype(self):
+        return _dtypes[str(self.data.dtype)]
+
+    def __init__(self, values, mask=None, dtype=None, copy=False):
```
jorisvandenbossche: can you add a docstring?

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,564 @@
...
+        self.data, self.mask = coerce_to_array(
+            values, dtype=dtype, mask=mask, copy=copy)
+
+    @classmethod
+    def _from_sequence(cls, scalars, mask=None, dtype=None, copy=False):
+        return cls(scalars, mask=mask, dtype=dtype, copy=copy)
```
jorisvandenbossche: remove mask argument

pandas/core/indexes/base.py
```diff
@@ -266,18 +267,32 @@ def __new__(cls, data=None, dtype=None, copy=False, name=None,
...
+            data = np.asarray(data)
+            if not (dtype is None or is_object_dtype(dtype)):
+
+                # coerce to the provided dtype
+                data = dtype.construct_array_type()(
+                    data, dtype=dtype, copy=False)
+
+            # coerce to the object dtype
+            data = data.astype(object)
+            return Index(data, dtype=object, copy=True, name=name,
```
TomAugspurger: Can this be `copy=copy`, the one from the index `__init__`?

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,564 @@
...
+        ndarray
+            The transformed values should maintain the ordering between values
+            within the array.
+
+        See Also
+        --------
+        ExtensionArray.argsort
+        """
+        data = self.data.copy()
+        data[self.mask] = data.min() - 1
```
jorisvandenbossche: can you add a comment here why? 
also, so the above means the nan values are sorted first and not last?

pandas/core/ops.py
```diff
@@ -135,6 +135,13 @@ def rfloordiv(left, right):
 
 
 def rmod(left, right):
+    # check if right is a string as % is the string
+    # formatting operation; this is a TypeError
+    # otherwise perform the op
+    if isinstance(right, compat.string_types):
+        raise TypeError("{typ} cannot perform the operation mod".format(
+            typ=type(left).__name__))
```
jorisvandenbossche: Why is this needed?

jorisvandenbossche (commented):
some quick comments

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,573 @@
...
+        # if we are filling
+        # we only fill where the indexer is null
+        # not existing missing values
+        # TODO(jreback) what if we have a non-na float as a fill value?
+        if allow_fill and notna(fill_value):
+            fill_mask = np.asarray(indexer) == -1
+            result[fill_mask] = fill_value
+            mask = mask ^ fill_mask
+
+        return type(self)(result, mask=mask, dtype=self.dtype)
```
TomAugspurger: Can you do `copy=False` here? IIUC, `result` is never a view on self.data?

jorisvandenbossche:
```
In [54]: s = pd.Series([1, 2, 3], dtype='Int64')

In [55]: s + 1
Out[55]: 
0    2
1    3
2    4
dtype: Int64

In [56]: s + 1.
Out[56]: 
0    2
1    3
2    4
dtype: Int64

In [57]: s + 1.1
Out[57]: 
0    2.1
1    3.1
2    4.1
dtype: float64
```

second case should also be float

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,573 @@
...
+        ndarray
+            The transformed values should maintain the ordering between values
+            within the array.
+
+        See Also
+        --------
+        ExtensionArray.argsort
+        """
+        data = self.data.copy()
+        data[self.mask] = data.min() - 1
```
TomAugspurger: potentially wraparound issues here?

pandas/core/arrays/base.py
jreback: yes

jorisvandenbossche: Can you clarify that in the docstring? 
Because basically what it comes down to is to be able to pass parameters of the dtype (in this case the actual dtype (int8, int64, ..), or in case of the Categorical the categories/ordered attributes, to have a dtype faithfully round-trip

jreback: done

pandas/core/arrays/integer.py
jreback: done

pandas/core/arrays/integer.py
jreback: done

pandas/core/arrays/integer.py
jreback: not sure  I think nans are basically ignored and handled at a higher level.

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,521 @@
...
+
+def coerce_to_array(values, dtype, mask=None, copy=False):
+    """
+    Coerce the input values array to numpy arrays with a mask
+
+    Parameters
+    ----------
+    values : 1D list-like
+    dtype : integer dtype
+    mask : boolean 1D array, optional
```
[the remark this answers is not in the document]
jreback: done

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,521 @@
...
+        assert len(mask) == len(values)
+
+    if not values.ndim == 1:
+        raise TypeError("values must be a 1D list-like")
+    if not mask.ndim == 1:
+        raise TypeError("mask must be a 1D list-like")
+
+    # avoid float->int numpy conversion issues
+    if is_object_dtype(values):
+        mask |= isna(values)
```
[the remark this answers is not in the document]
jreback: was old code, removed

pandas/core/indexes/base.py
jreback: yeah we don't have great tests around this

pandas/core/arrays/integer.py
jreback: done

pandas/core/arrays/integer.py
jreback: yeah haven't really looked at this that much

pandas/core/arrays/integer.py
jorisvandenbossche: Yes, it indeed works because `sort_values` is handling this. 
Anyhow, opened an issue here to fix this in general: https://github.com/pandas-dev/pandas/issues/21801

pandas/core/ops.py
jorisvandenbossche: OK, I agree that the above is not the best error message (as I assume the catching above is only for a better error message?)

Can you add to the message "... with a string" at the end?

jreback:
fixed up if you want to have a relook @jorisvandenbossche @TomAugspurger @jbrockmendel 

ps. numpy scalars should be banned :<

jorisvandenbossche:
Still on the branch I fetched this afternoon (so might already be fixed), but astype was not yet working:

```
In [123]: s = pd.Series([1, 2, 3], dtype='Int64')

In [124]: s.astype('Int8')
Out[124]: 
0    1
1    2
2    3
dtype: int8    <--- note lower case here
```

```
In [125]: s = pd.Series([1, 2, 3, None], dtype='Int64')

In [127]: s.astype('Int8')
...
ValueError: cannot convert float NaN to integer
```

jreback:
https://github.com/pandas-dev/pandas/pull/21160/commits/1ff394a04b805358feb9a2167f24dc09deaf6c76 fixes ``.astype()`` generally for ExtensionArrays when the string is specified for a dtyep.
cc @jorisvandenbossche

jreback:
@jorisvandenbossche @TomAugspurger should be good

jreback:
finally passing again

jreback:
if any final comments. @pandas-dev/pandas-core

pandas/core/internals.py
```diff
@@ -632,9 +633,13 @@ def _astype(self, dtype, copy=False, errors='raise', values=None,
 
             return self.make_block(Categorical(self.values, dtype=dtype))
 
+        # convert dtypes if needed
+        try:
+            dtype = pandas_dtype(dtype)
+        except TypeError:
+            pass
```
toobaz: I'm missing the rationale for not raising here

jorisvandenbossche:
> if any final comments. @pandas-dev/pandas-core

Please give some time if possible, we're at Scipy so might take a few days. Or is that very annoying for a next PR?

pandas/core/ops.py
Dr-Irv: @jreback see comment above - the dispatching to `dispatch_to_extension_op` is not consistent between `_arith_method_SERIES()` and `_comp_method_SERIES()`.

Dr-Irv (commented):
@jreback So is this a desired behavior?
```
>>> pd.core.arrays.integer.IntegerArray([1.2, 2.3, 3.7])
IntegerArray([1, 2, 3], dtype='Int64')
```
In other words, if you specify an `IntegerArray`, and pass floats, shouldn't there be an error?  Because if you do:
```
pd.Series([1.2, 2.3, 3.7], dtype='int64')
```
a `ValueError: Trying to coerce float values to integers` is raised.

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,570 @@
...
+        return self._coerce_to_ndarray()
+
+    def __iter__(self):
+        """Iterate over elements of the array.
+
+        """
+        # This needs to be implemented so that pandas recognizes extension
+        # arrays as list-like. The default implementation makes successive
+        # calls to ``__getitem__``, which may be slower than necessary.
+        for i in range(len(self)):
```
jschendel: should import `range` from `pandas.compat` to make this an iterator in py2

pandas/core/arrays/base.py
```diff
@@ -88,14 +88,17 @@ class ExtensionArray(object):
...
-    def _from_sequence(cls, scalars, copy=False):
+    def _from_sequence(cls, scalars, dtype=None, copy=False):
         """Construct a new ExtensionArray from a sequence of scalars.
 
         Parameters
         ----------
         scalars : Sequence
             Each element will be an instance of the scalar type for this
             array, ``cls.dtype.type``.
+        dtype : Dtype, optional
```
gfyoung: Nit: `dtype` instead of `Dtype`?

pandas/core/internals.py
jreback: hmm, good point. I think I can take this out, previously this was ``np.dtype`` which would raise on lots of things (like pandas dtypes). ``pandas_dtype`` only raises on completely invalid things.

jreback:
@Dr-Irv so that is a good point

numpy silently converts floats to ints. We *could* warn or raise.
```
In [2]: np.array([1.2, 2.3, 3.7]).astype('int64')
Out[2]: array([1, 2, 3])
```

We normally convert things to new dtypes that can be directly converted w.o loss of precision. So I am inclined to raise here (note that numpy *does* raise on the above if a NaN appears, but we don't have that issue :)

pandas/core/arrays/base.py
jreback: done

pandas/core/arrays/integer.py
jreback: done

wesm:
@jreback could you summarize the memory format for this extension dtype in the PR description? 

For example, Arrow implements nullable integer as follows:

* 1 contiguous memory buffer containing null bits. A set bit (1) indicates non-null where a cleared bit indicates null

* 1 contiguous memory buffer containing the integer values

If you wanted, you could clone this using the bitarray (https://pypi.org/project/bitarray/) package if you are not already

jreback:
latest push raises on casting non-equivalent floats -> ints (numpy allows this, but raising is better I think).

jreback:
addressing memory layout (@wesm); these are currently hidden behind private attributes.

- uses a int8 mask where set is null and clear is non-null (could prob flip this)
- uses a contiguous ndarray of the same dtype to hold the data

so we have a bit of overhead compared to bitmasks (we use a byte per). I will look at ``bitarray``
should be simple to use, though it would add a dep (possibly could do this an an optional depenedecy - if easy may just do that).

but would do this in a followup.

wesm:
Ideally the user would not be exposed to these details at least unless they are constructing from large-ish datasets. Using bits would help with zero copy from Arrow-land

jorisvandenbossche:
> latest push raises on casting non-equivalent floats -> ints (numpy allows this, but raising is better I think).


\+ 1

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,592 @@
...
+
+            result[mask] = True if op_name == 'ne' else False
+            return result
+
+        name = '__{name}__'.format(name=op.__name__)
+        return set_function_name(cmp_method, name, cls)
+
+    def _maybe_mask_result(self, result, mask, other, op_name):
+        """
+        Parameters
```
jorisvandenbossche: Can you give this an explanation summary?

pandas/core/dtypes/common.py
```diff
@@ -1976,21 +1986,38 @@ def pandas_dtype(dtype):
...
+
+    Raises
+    ------
+    TypeError if not a dtype
+
     """
 
+    # short-circuit
+    if isinstance(dtype, np.ndarray):
+        return dtype.dtype
```
jorisvandenbossche: Is this needed? That seems like code smell that you are not sure it is a dtype or an array.

(it should be documented and tested otherwise, this is a public function)

pandas/core/ops.py
jorisvandenbossche: Can you do this one?

pandas/core/ops.py
jorisvandenbossche: Is this difference intended:

```
In [55]: s = pd.Series([1, 2, np.nan], dtype='Int64')

In [56]: a = s.values

In [57]: np.array([1, 2, 3]) + a
Out[57]: array([2, 4, nan], dtype=object)

In [58]: a + np.array([1, 2, 3]) 
Out[58]: IntegerArray([2, 4, nan], dtype='Int64')
```

(I thought the code here tried to avoid this difference, although I am not fully sure if that is needed)

pandas/tests/extension/decimal/array.py
```diff
@@ -54,8 +53,8 @@ def __init__(self, values, copy=False):
         # self._values = self.values = self.data
 
     @classmethod
-    def _from_sequence(cls, scalars, copy=False):
-        return cls(scalars)
+    def _from_sequence(cls, scalars, dtype=None, copy=False):
+        return cls(scalars, copy=copy)
```
jorisvandenbossche: minor nit: copy does not need to be passed here

pandas/tests/extension/decimal/test_decimal.py
```diff
@@ -1,13 +1,19 @@
...
 
 from pandas.tests.extension import base
 
-from .array import DecimalDtype, DecimalArray, make_data
+from .array import DecimalDtype, DecimalArray
+
+
+@pytest.fixture
+def make_data():
+    return [decimal.Decimal(random.random()) for _ in range(100)]
```
jorisvandenbossche: What is the reason you moved this? There is already the `data` fixture

pandas/tests/extension/integer/test_integer.py
```diff
@@ -0,0 +1,693 @@
...
+                original = original.fillna(0).astype('int')
+
+            original = original.astype('float')
+            original[mask] = np.nan
+            self.assert_series_equal(original, expected.astype('float'))
+
+        # assert our expected result
+        self.assert_series_equal(result, expected)
+
+    def test_arith_integer_array(self, data, all_arithmetic_operators):
```
jorisvandenbossche: I would move all integer specific tests to `tests/arrays/integer/`, and only keep here those that are actually subclassed from the base extension tests.

jorisvandenbossche (commented):
Additional non-inline comments:

- Can you add a section to the documentation as well? (not sure where)

- I think we should consider not using `np.nan` for the missing value indicator (as the value what you get back on scalar access, on conversion to numpy array

- How does a user check for the dtype?

    ```
    In [17]: df.dtypes == 'int64'
    Out[17]: 
    a    False
    b     True
    dtype: bool

    In [18]: df.dtypes == 'Int64'
    Out[18]: 
    a    False
    b    False
    dtype: bool
    ```

- For DataFrame the arithmetic ops do not work yet (with example from the whatsnew):

    ```
    In [45]: df = pd.DataFrame({'A': s, 'B': [1, 1, 3], 'C': list('aab')})

    In [46]: df
    Out[46]: 
         A  B  C
    0    1  1  a
    1    2  1  a
    2  NaN  3  b

    In [47]: df[['A', 'B']] + 1
    ...
    TypeError: Could not compare [1] with block values

    In [48]: df[['A']] + 1
    ...
    TypeError: Could not compare [1] with block values
    ```

pandas/tests/extension/decimal/array.py
Dr-Irv: @jorisvandenbossche On your example above where you ask "Is this difference intended:", there is no way to avoid this.  If you do `np.array([1, 2, 3]) + a`, then the numpy `__add__()` method is called. If you did `pd.Series([1,2,3]) + a` in your example, it works, because then the `Series` method sees that the right operand is an EA, and dispatches to the reverse operator of the EA.

pandas/core/ops.py
jorisvandenbossche: @Dr-Irv said (putting it in the right place):

@jorisvandenbossche On your example above where you ask "Is this difference intended:", there is no way to avoid this. If you do np.array([1, 2, 3]) + a, then the numpy __add__() method is called. If you did pd.Series([1,2,3]) + a in your example, it works, because then the Series method sees that the right operand is an EA, and dispatches to the reverse operator of the EA.

jreback:
@jorisvandenbossche 

> I think we should consider not using np.nan for the missing value indicator (as the value what you get back on scalar access, on conversion to numpy array

what is the reasoning here? users are accustomed to using ``np.nan`` and ``pd.NaT`` exclusively. (and not really ``None``), and an already constructed float array that is using integers + nan is a natural conversion.

pandas/tests/extension/decimal/test_decimal.py
jreback: these were previously in the ``array.py`` module itself, which is not a great idea.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -26,6 +27,61 @@ See the :ref:`ExtensionArray Operator Support
 <extending.extension.operator>` documentation section for details on both
 ways of adding operator support.
 
+.. _whatsnew_0240.enhancements.intna:
+
+Integer NA Support
+^^^^^^^^^^^^^^^^^^
+
+Pandas has gained the ability to hold integer dtypes with missing values. This long requested feature is enabled thru the use of ``ExtensionTypes`` . Here is an example of the usage.
```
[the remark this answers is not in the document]
jreback: changed

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -26,6 +27,61 @@ See the :ref:`ExtensionArray Operator Support
 <extending.extension.operator>` documentation section for details on both
 ways of adding operator support.
 
+.. _whatsnew_0240.enhancements.intna:
+
+Integer NA Support
```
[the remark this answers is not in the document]
jreback: done

pandas/core/arrays/base.py
```diff
@@ -88,14 +88,17 @@ class ExtensionArray(object):
...
+    def _from_sequence(cls, scalars, dtype=None, copy=False):
         """Construct a new ExtensionArray from a sequence of scalars.
 
         Parameters
         ----------
         scalars : Sequence
             Each element will be an instance of the scalar type for this
             array, ``cls.dtype.type``.
+        dtype : dtype, optional
+            construct for this particular dtype. This should be a Dtype
```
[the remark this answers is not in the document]
jreback: done

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,592 @@
+import sys
+import warnings
+import copy
+import numpy as np
+
+from pandas.compat import u, range
+from pandas.core.dtypes.generic import ABCSeries, ABCIndexClass
+from pandas.util._decorators import cache_readonly
+from pandas.compat import set_function_name
+from pandas.api.types import (is_integer, is_scalar, is_float,
```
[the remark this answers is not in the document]
jreback: yep

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,592 @@
...
+    """
+    Parameters
+    ----------
+    values : 1D list-like
+    dtype : dtype, optional
+        dtype to coerce
+
+    Returns
+    -------
+    infer and return an integer array
```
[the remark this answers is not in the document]
jreback: done

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,592 @@
...
+        if is_integer(item):
+            if self._mask[item]:
+                return self.dtype.na_value
+            return self._data[item]
+        return type(self)(self._data[item],
+                          mask=self._mask[item],
+                          dtype=self.dtype)
+
+    def _coerce_to_ndarray(self):
+        """ coerce to an ndarary, preserving my scalar types """
```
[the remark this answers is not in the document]
jreback: hmm, they are python types interesting.

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,592 @@
...
+        mask : 1D list-like, optional
+        dtype : subclass of _IntegerDtype, optional
+        copy : bool, default False
+
+        Returns
+        -------
+        IntegerArray
+        """
+        self._data, self._mask = coerce_to_array(
+            values, dtype=dtype, mask=mask, copy=copy)
```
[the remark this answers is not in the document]
jreback: this is not possible, as quite a lot of checking / coercion needs to occur on the input. e.g. make it an ndarray if not, coerce dtype. this is the point of ``coerce_to_array``, to make it DRY (as I also have to do this with setitem).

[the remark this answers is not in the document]
jorisvandenbossche: It *is* possible, just a different API design choice. 
I personally think it makes it cleaner code-wise, to clearly separate those case where coercion and checking is needed (there we would still use `coerce_to_array` or another method calling that under the hood) and those internal places where it is not needed (as those also exist: eg getitem). 
It is very similar to the `fastpath` keyword in many constructors, but I think we agree on is not a good design to mix both in the same constructor.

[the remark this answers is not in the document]
shoyer: I agree with @jorisvandenbossche, we certainly could make a simpler constructor which doesn't call `coerce_to_array()` internally. This is a perfect use case for an alternative constructor, e.g., `IntegerArray.from_values()`.

In general I think this is a better/cleaner pattern to use for internal APIs in pandas, because it involves less introspection (which tends to be error/bug prone).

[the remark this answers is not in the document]
jreback: did you see my response above?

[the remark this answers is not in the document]
shoyer: `pd.Series(array, dtype='Int64')` could call `IntegerArray.from_values(array, dtype='Int64')` internally. It doesn't need to call the `IntegerArray()` constructor directly.

jreback:
latest push fixes dtype comparisons with strings & xfails DataFrame tests (they will be implemented in a followup, #21789) 

I haven't moved any tests yet. have to think about how to do this for all extension array types. We for sure want to test the interface, but don't want to split arbitrarily.

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,599 @@
...
+
+    @classmethod
+    def _from_sequence(cls, scalars, dtype=None, copy=False):
+        return cls(scalars, dtype=dtype, copy=copy)
+
+    @classmethod
+    def _from_factorized(cls, values, original):
+        return cls(values, dtype=original.dtype)
+
+    def __getitem__(self, item):
```
WillAyd: Without trying to overcomplicated things have we considered moving some of these items to a `MaskedEAMixin`? I'm thinking of taking a stab at the Boolean EA next and can see this being generalizable along with a few other methods (`__iter__`, `__setitem__`, perhaps `take`, etc...)

jreback: certainly could, though I think might be better just to just directly subclass ``IntegerArray`` and the dtype, but that's for another PR :>

jreback:
anyhow any final comments @pandas-dev/pandas-core 

certainly will be follows in any event.

jreback:
@jorisvandenbossche

chris-b1:
Made a couple doc follow-up notes in #22003

jorisvandenbossche:
>> I think we should consider not using np.nan for the missing value indicator (as the value what you get back on scalar access, on conversion to numpy array

> what is the reasoning here? users are accustomed to using np.nan and pd.NaT exclusively. (and not really None), and an already constructed float array that is using integers + nan is a natural conversion.

I am not necessarily saying that we should use None, but the fact is that we currently somewhat 'misuse' `np.nan` as missing value indicator (for lack of good alternative, for sure). But when we start adding the capability to more data types than float, I think we should consider using a separate value for this. 
For example in arrow there is a distinction between NaN and NA (there was recently some discussion on that). I think in the end we should also do that. I am not sure when we should do that, but I would like to at least see some discussion about it.

pandas/core/arrays/integer.py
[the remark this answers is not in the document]
jorisvandenbossche: Can you respond here further?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -31,6 +32,62 @@ See the :ref:`ExtensionArray Operator Support
...
+   # indexing
+   s.iloc[1:3]
+
+   # operate with other dtypes
+   s + s.iloc[1:3].astype('Int8')
+
+   # coerce when needed
+   s + 0.01
+
+These dtypes can operate as part of of ``DataFrame``.
```
jorisvandenbossche: "of of " -> "of a"

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,599 @@
...
+        values = safe_cast(values, dtype, copy=False)
+
+    return values, mask
+
+
+class IntegerArray(ExtensionArray, ExtensionOpsMixin):
+    """
+    We represent an IntegerArray with 2 numpy arrays
+    - data: contains a numpy integer array of the appropriate dtype
+    - mask: a boolean array holding a mask on the data, False is missing
```
jorisvandenbossche: > False is missing

I don't think this is true currently?

But given the discussion earlier, I think it would be good to actually implement what you stated there to follow the example of arrow?

jorisvandenbossche:
> I disagree that None is any way superior here, and is just plain confusing.

Again, I didn't say we should use None for this. I also think we should not use that, it is already used in many other contexts.

jreback:
ok, let's certainly discuss. In reality is a pretty cosmetic detail (mostly for actual printing).

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,599 @@
...
+        # if we are astyping to an existing IntegerDtype we can fastpath
+        if isinstance(dtype, _IntegerDtype):
+            result = self._data.astype(dtype.numpy_dtype,
+                                       casting='same_kind', copy=False)
+            return type(self)(result, mask=self._mask,
+                              dtype=dtype, copy=False)
+
+        # coerce
+        data = self._coerce_to_ndarray()
+        return data.astype(dtype=dtype, copy=False)
```
jorisvandenbossche: Should we treat converting to float separately here? (that could be easily made more performant, as a probably common use case for astype)

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,599 @@
...
+            if isinstance(other, IntegerArray):
+                other, mask = other._data, other._mask
+            elif is_list_like(other):
+                other = np.asarray(other)
+                if other.ndim > 0 and len(self) != len(other):
+                    raise ValueError('Lengths must match to compare')
+
+            # numpy will show a DeprecationWarning on invalid elementwise
+            # comparisons, this will raise in the future
+            with warnings.catch_warnings(record=True):
```
jorisvandenbossche: Do we want to catch those warnings if it will raise in the future?

pandas/core/arrays/integer.py
jorisvandenbossche: This one

pandas/core/arrays/integer.py
```diff
@@ -0,0 +1,599 @@
...
+            if isinstance(other, (ABCSeries, ABCIndexClass)):
+                other = getattr(other, 'values', other)
+
+            if isinstance(other, IntegerArray):
+                other, mask = other._data, other._mask
+            elif getattr(other, 'ndim', 0) > 1:
+                raise NotImplementedError(
+                    "can only perform ops with 1-d structures")
+            elif is_list_like(other):
+                other = np.asarray(other)
```
jorisvandenbossche: Should we try to convert to IntegerArray here if possible? 
eg `s + s.tolist()` gives floats (in case `s` is a series with int-na dtype)

jreback:
bombs away

pandas/core/ops.py
jorisvandenbossche: This one

pandas/core/ops.py
```diff
@@ -1048,16 +1055,39 @@ def dispatch_to_extension_op(op, left, right):
...
+    # dispatching
+
     if is_extension_array_dtype(left):
-        res_values = op(left.values, right)
+
+        new_left = left.values
+        if isinstance(right, np.ndarray):
+
+            # handle numpy scalars, this is a PITA
+            # TODO(jreback)
```
jorisvandenbossche: Can you clarify this to do comment?

jorisvandenbossche:
Ahum, I was actually reviewing this now

pandas/tests/extension/integer/test_integer.py
jorisvandenbossche: We need to discuss how to handle this (it will pop-up in all the other internal extension dtypes as well), so now is a good time. 
I would personally move all specific tests that you add here that are not in the parent base tests to `tests/arrays/integer`, and only subclass here to check that the base tests make sense

jorisvandenbossche (commented):
It would make it easier to review if you do not rebase your commits, once you do that github's features to automatically let me see what has changed since the last review does not work anymore.

To come back to my previous comment: can you add a section to the actual documentation? (not only whatsnew)

pandas/core/arrays/__init__.py
```diff
@@ -1,7 +1,10 @@
 from .base import (ExtensionArray,    # noqa
+                   ExtensionOpsMixin,
                    ExtensionScalarOpsMixin)
 from .categorical import Categorical  # noqa
 from .datetimes import DatetimeArrayMixin  # noqa
 from .interval import IntervalArray  # noqa
 from .period import PeriodArrayMixin  # noqa
 from .timedeltas import TimedeltaArrayMixin  # noqa
+from .integer import (  # noqa
+    IntegerArray, to_integer_array)
```
jorisvandenbossche: What was the goal of exposing `to_integer_array` here? Is it used somewhere else?

jreback:
well I'll adress these in a followup. and for the record I didn't actually rebase this.

jreback:
> Can you respond here further?

your comment about the __init__. Well I already address this and it is simply not possible to have the __init__ do nothing. I am -1 on adding from_* methods. This is how it is implemented. If you want to change it you are welcome to submit a PR.

shoyer:
> This is how it is implemented. If you want to change it you are welcome to submit a PR.

Sorry, but I don't think this should be an acceptable response. I think we should either roll back this PR or disable all public interfaces for this functionality until the (collective) development team is happy with the new features. As of now, I would consider this a blocker for new releases.

As a general policy, I think we should require explicit approval from at least one other core developer before merging pull requests. I am all for avoiding large PRs and doing collaborative development in master, but we should not merge PRs unless they pass code review  (certainly not for large changes like this one).

jreback:
@shoyer that is a disengenuos response. If you want to do code review on the 100's of open PR's be my guest. This PR has been in a ready state for weeks.

jreback:
> As a general policy, I think we should require explicit approval from at least one other core developer before merging pull requests. I am all for avoiding large PRs and doing collaborative development in master, but we should not merge PRs unless they pass code review (certainly not for large changes like this one)

If folks respond in a timely manner, sure. If you haven't noticed I approve every other PR in a pretty timely manner. After round and rounds of comments. 

I believe I have earned, and deserve the same consideration.

jorisvandenbossche:
> If you want to do code review on the 100's of open PR's be my guest.

This PR is not just "one of the 100's of open PRs", but a major change that needs more attention than smaller PRs. I agree with Stephan that it would be good for such PRs to have the explicit +1 of other core devs.
I did my best to review again as fast as possible, but have been very busy the last days. Time of other core devs is scarce, that is a reality we have to live with.

But OK, although I would have preferred to iterate further in this PR, let's move forward in follow-up PRs!
It would be helpful for that if you could respond to some of my comments above (of which some were already there in my previous review).

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -31,6 +32,62 @@ See the :ref:`ExtensionArray Operator Support
...
+These dtypes can be merged & reshaped & casted.
+
+.. ipython:: python
+
+   pd.concat([df[['A']], df[['B', 'C']]], axis=1).dtypes
+   df['A'].astype(float)
+
+.. warning::
+
+   The Integer NA support currently uses the captilized dtype version, e.g. ``Int8`` as compared to the traditional ``int8``. This may be changed at a future date.
```
toobaz: captilized -> capitalized?

jreback:
> This PR is not just "one of the 100's of open PRs"

This misses the point. I spend an inordinate amount of time reviewing practically everything. I don't expect everyone to do this, rather for my PR's, which I DO expect review time from other folks.

At the same time, endless comments, while certainly debatable, we don't force on anyone. Many many times, we have done follow up PR's just to cut down on the endless review cycle.

In any event I will address your comments.

PR #21163, opened 2018-05-22
Cleanup clipboard tests

david-liu-brattle-1:
As requested in #21111, I've refactored the original clipboard tests as well as adding my own.  The set of tests have been improved to include testing of:

- failure of tab delimited tables (#21104)
- errors relating to copying and reading space-delimited tables
- copy and pasting tables where the text contains quotes or delimiters

As a result, these tests will fail on the current master branch, but should succeed on the commit in #21111

@WillAyd

### DIFF
--- a/pandas/tests/io/test_clipboard.py
+++ b/pandas/tests/io/test_clipboard.py
@@ -32,7 +32,11 @@ def build_kwargs(sep, excel):
     return kwargs
 
 
-def gen_df(data_type):
+@pytest.fixture(params=['delims', 'utf8', 'string', 'long', 'nonascii',
+                        'colwidth', 'mixed', 'float', 'int'])
+def df(request):
+    data_type = request.param
+
     if data_type == 'delims':
         return pd.DataFrame({'a': ['"a,\t"b|c', 'd\tef´'],
                              'b': ['hi\'j', 'k\'\'lm']})
@@ -68,19 +72,14 @@ def gen_df(data_type):
         return mkdf(5, 3, data_gen_f=lambda *args: randint(2),
                     c_idx_type='s', r_idx_type='i',
                     c_idx_names=[None], r_idx_names=[None])
-
-
-@pytest.fixture(params=['delims', 'utf8', 'string', 'long', 'nonascii',
-                        'colwidth', 'mixed', 'float', 'int'])
-def df(request):
-    return gen_df(request.param)
+    else:
+        raise ValueError
 
 
 @pytest.mark.single
 @pytest.mark.skipif(not _DEPS_INSTALLED,
                     reason="clipboard primitives not installed")
 class TestClipboard(object):
-
     def check_round_trip_frame(self, data, excel=None, sep=None,
                                encoding=None):
         data.to_clipboard(excel=excel, sep=sep, encoding=encoding)
@@ -113,16 +112,15 @@ class TestClipboard(object):
     # Two character separator is not supported in to_clipboard
     # Test that multi-character separators are not silently passed
     @pytest.mark.xfail(reason="Not yet implemented.  Fixed in #21111")
-    def test_excel_sep_warning(self):
+    def test_excel_sep_warning(self, df):
         with tm.assert_produces_warning():
-            gen_df('string').to_clipboard(excel=True, sep=r'\t')
+            df.to_clipboard(excel=True, sep=r'\t')
 
     # Separator is ignored when excel=False and should produce a warning
-    # Fails, Fixed in #21111
-    @pytest.mark.xfail
-    def test_copy_delim_warning(self):
+    @pytest.mark.xfail(reason="Not yet implemented.  Fixed in #21111")
+    def test_copy_delim_warning(self, df):
         with tm.assert_produces_warning():
-            gen_df('string').to_clipboard(excel=False, sep='\t')
+            df.to_clipboard(excel=False, sep='\t')
 
     # Tests that the default behavior of to_clipboard is tab
     # delimited and excel="True"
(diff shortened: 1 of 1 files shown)


jreback (changes requested):
so what you can do is xfail the cases which don't work now but will work in the new PR.

david-liu-brattle-1:
@jreback 
I made some changes to the testing setup.  Is this what you had in mind?

pandas/tests/io/test_clipboard.py
```diff
@@ -22,73 +22,118 @@
...
+        result = read_clipboard()
+        assert df.to_string() == result.to_string()
+        assert df.shape == result.shape
+
+    @pytest.mark.xfail
+    def test_excel_sep_warning(self):
+        with tm.assert_produces_warning():
+            gen_df('string').to_clipboard(excel=True, sep=r'\t')
+
+    @pytest.mark.xfail
```
jreback: for tests that xfail, pls provide a reason

pandas/tests/io/test_clipboard.py
```diff
@@ -22,73 +22,118 @@
...
-            self.check_round_trip_frame(dt, excel=False)
-
-    def test_round_trip_frame(self):
-        for dt in self.data_types:
-            self.check_round_trip_frame(dt)
+    @pytest.mark.xfail
+    def test_round_trip_frame(self, df):
+        self.check_round_trip_frame(df)
+
+    def test_round_trip_frame_sep(self, df):
```
jreback: you can parameterize on the sep

jreback (changes requested):
do all of these tests actually need xfailing?

pandas/tests/io/test_clipboard.py
david-liu-brattle-1: Should I add comments in the code with the reason?

jreback: yes if they are xfailing, reference the issue number and a brief description. (when the bug is fixed will remove the xfails, but that's in a subsequent PR).

pandas/tests/io/test_clipboard.py
```diff
@@ -22,73 +23,139 @@
...
+    elif data_type == 'int':
+        return mkdf(5, 3, data_gen_f=lambda *args: randint(2),
+                    c_idx_type='s', r_idx_type='i',
+                    c_idx_names=[None], r_idx_names=[None])
+
+
+@pytest.fixture(params=['delims', 'utf8', 'string', 'long', 'nonascii',
+                        'colwidth', 'mixed', 'float', 'int'])
+def df(request):
+    return gen_df(request.param)
```
jreback: I would just move gen_df inside the fixture itself, I think makes it more clear.

pandas/tests/io/test_clipboard.py
```diff
@@ -22,73 +23,139 @@
...
+    @pytest.mark.xfail
+    def test_copy_delim_warning(self):
+        with tm.assert_produces_warning():
+            gen_df('string').to_clipboard(excel=False, sep='\t')
+
+    # Tests that the default behavior of to_clipboard is tab
+    # delimited and excel="True"
+    # Fails because to_clipboard defaults to space delim.
+    # Issue in #21104, Fixed in #21111
+    @pytest.mark.xfail
```
jreback: every xfail needs a reason

jreback (changes requested):

pandas/tests/io/test_clipboard.py
david-liu-brattle-1: gen_df is called directly from some of the tests.  The fixture is called using a a pytest object and gen_df needs to be called with a string argument.  Is there a better workaround than separating the two functions?

david-liu-brattle-1: I can change it to something like this, but I think separating the two out is still preferable.
```
@pytest.fixture(params=['delims', 'utf8', 'string', 'long', 'nonascii',
                        'colwidth', 'mixed', 'float', 'int'])
def df(request):
    if type(request) is str:
        data_type = request
    else:
        data_type = request.param

    if data_type == 'delims':
        return pd.DataFrame({'a': ['"a,\t"b|c', 'd\tef´'],
                             'b': ['hi\'j', 'k\'\'lm']})
    elif data_type == 'utf8':
        return pd.DataFrame({'a': ['µasd', 'Ωœ∑´'],
                             'b': ['øπ∆˚¬', 'œ∑´®']})
```

pandas/tests/io/test_clipboard.py
david-liu-brattle-1: Can you take a look at the latest version? Is that what you meant for reason?

pandas/tests/io/test_clipboard.py
jreback: oh didn't see that, though then this begs the question of why you are not simply passing the constructed df in?

jreback (changes requested):
minor comment. all of the currently xfailed ones are fixed by #21111 ?

pandas/tests/io/test_clipboard.py
david-liu-brattle-1: That's a good point, I'll update

pandas/tests/io/test_clipboard.py
```diff
@@ -22,73 +23,136 @@
...
+        result = read_clipboard()
+        assert df.to_string() == result.to_string()
+        assert df.shape == result.shape
+
+    # Two character separator is not supported in to_clipboard
+    # Test that multi-character separators are not silently passed
+    @pytest.mark.xfail(reason="Not yet implemented.  Fixed in #21111")
+    def test_excel_sep_warning(self):
+        with tm.assert_produces_warning():
+            gen_df('string').to_clipboard(excel=True, sep=r'\t')
```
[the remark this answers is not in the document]
david-liu-brattle-1: Thanks, that makes a lot of sense and I'll keep that in mind for future cases.  I went a different way and just ran that test function on the entire `df` fixture, which is a little cleaner.

david-liu-brattle-1:
@jreback 
Yes, all xfails pass in #21111

jreback (approved):

jreback:
thanks @david-liu-brattle-1 !

PR #21171, opened 2018-05-22
BUG: group with multiple named results

### LINKED ISSUE #19029: Warn on duplicate names in MI?
Opening a new issue so this isn't lost.

In https://github.com/pandas-dev/pandas/pull/18882 banned duplicate names in a MultiIndex. I think this is a good change since allowing duplicates hit a lot of edge cases when you went to actually do something. I want to make sure we understand all the cases that actually produce duplicate names in the MI though, specifically groupby.apply.

```python
In [1]: import dask.dataframe as dd

In [2]: import pandas as pd

In [3]:     pdf = pd.DataFrame({'a': [1, 2, 3, 4, 5, 6, 7, 8, 9],
   ...:                         'b': [4, 5, 6, 3, 2, 1, 0, 0, 0]},
   ...:                        index=[0, 1, 3, 5, 6, 8, 9, 9, 9]).set_index("a")
   ...:
   ...:

In [4]: pdf.groupby(pdf.index).apply(lambda x: x.b)
```

Another, more realistic example: groupwise drop_duplicates:

```python
In [18]: df = pd.DataFrame({"B": [0, 0, 0, 1, 1, 1, 2, 2, 2]}, index=pd.Index([0, 1, 1, 2, 2, 2, 0, 0, 1], name='a'))

In [19]: df
Out[19]:
   B
a
0  0
1  0
1  0
2  1
2  1
2  1
0  2
0  2
1  2

In [20]: df.groupby('a').apply(pd.DataFrame.drop_duplicates)
Out[20]:
     B
a a
0 0  0
  0  2
1 1  0
  1  2
2 2  1
```

Is it possible to throw a warning on this for now, in case duplicate names are more common than we thought?

guenteru:
- [x] closes #19029 
- [x] tests added / passed
- [ ] whatsnew entry

This bugfix gets rid of duplicated names that can be the result of groupby operations (#19029).
I opted to implement one of the ideas proposed by [toobaz](https://github.com/pandas-dev/pandas/issues/19029#issuecomment-354684340): duplicated names get suffixed by their corresponding position, i.e. ['name','name'] gets transformed into ['name0', 'name1'].

* A few testcases have been added.
* One particular testcase had to be changed ([test_crosstab_dup_index_names](https://github.com/pandas-dev/pandas/blob/master/pandas/tests/reshape/test_pivot.py))
  - This is because with the new bugfix crosstab does not yield a ValueError anymore.

### DIFF
--- a/pandas/core/groupby/groupby.py
+++ b/pandas/core/groupby/groupby.py
@@ -2300,14 +2300,14 @@ class BaseGrouper(object):
     def names(self):
         # GH 19029
         # add suffix to level name in case they contain duplicates (GH 19029):
-        orig_names =  [ping.name for ping in self.groupings]
-        # if no names were assigned return the original names 
+        orig_names = [ping.name for ping in self.groupings]
+        # if no names were assigned return the original names
         if all(x is None for x in orig_names):
             return orig_names
         # in case duplicates are contained rename all of them
         if len(set(orig_names)) < len(orig_names):
-            orig_names = [''.join([str(x),str(i)])
-                                for i,x in enumerate(orig_names)]
+            orig_names = [''.join([str(x), str(i)])
+                          for i, x in enumerate(orig_names)]
 
         return orig_names
 
--- a/pandas/core/groupby/groupby.py
+++ b/pandas/core/groupby/groupby.py
@@ -2298,7 +2298,18 @@ class BaseGrouper(object):
 
     @property
     def names(self):
-        return [ping.name for ping in self.groupings]
+        # GH 19029
+        # add suffix to level name in case they contain duplicates (GH 19029):
+        orig_names =  [ping.name for ping in self.groupings]
+        # if no names were assigned return the original names 
+        if all(x is None for x in orig_names):
+            return orig_names
+        # in case duplicates are contained rename all of them
+        if len(set(orig_names)) < len(orig_names):
+            orig_names = [''.join([str(x),str(i)])
+                                for i,x in enumerate(orig_names)]
+
+        return orig_names
 
     def size(self):
         """
--- a/pandas/core/groupby/groupby.py
+++ b/pandas/core/groupby/groupby.py
@@ -2298,18 +2298,28 @@ class BaseGrouper(object):
 
     @property
     def names(self):
-        # GH 19029
         # add suffix to level name in case they contain duplicates (GH 19029):
         orig_names = [ping.name for ping in self.groupings]
         # if no names were assigned return the original names
         if all(x is None for x in orig_names):
             return orig_names
-        # in case duplicates are contained rename all of them
-        if len(set(orig_names)) < len(orig_names):
-            orig_names = [''.join([str(x), str(i)])
-                          for i, x in enumerate(orig_names)]
 
-        return orig_names
+        suffixes = collections.defaultdict(int)
+        dups = {n: count for n, count in
+                collections.Counter(orig_names).items() if count > 1}
+        new_names = []
+        for name in orig_names:
+            if name not in dups:
+                new_names.append(name)
+            else:
+                if name is not None:
+                    new_name = '{0}_{1}'.format(name, suffixes[name])
+                else:
+                    new_name = '{0}'.format(suffixes[name])
+                suffixes[name] += 1
+                new_names.append(new_name)
+
+        return new_names
 
     def size(self):
         """
--- a/pandas/tests/groupby/test_categorical.py
+++ b/pandas/tests/groupby/test_categorical.py
@@ -558,14 +558,6 @@ def test_as_index():
     result = df.groupby(['cat', s], as_index=False, observed=True).sum()
     tm.assert_frame_equal(result, expected)
 
-    # GH 19029: conflicitng names should not raise a value error anymore
-    raised = False
-    try:
-        df.groupby(['cat', s.rename('cat')], observed=True).sum()
-    except ValueError:
-        raised = True
-    assert raised is False
-
     # is original index dropped?
     group_columns = ['cat', 'A']
     expected = DataFrame(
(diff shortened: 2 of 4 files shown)


pandas/core/groupby/groupby.py
```diff
@@ -2298,7 +2298,18 @@ def levels(self):
...
-        return [ping.name for ping in self.groupings]
+        # GH 19029
+        # add suffix to level name in case they contain duplicates (GH 19029):
+        orig_names = [ping.name for ping in self.groupings]
+        # if no names were assigned return the original names
+        if all(x is None for x in orig_names):
+            return orig_names
+        # in case duplicates are contained rename all of them
+        if len(set(orig_names)) < len(orig_names):
+            orig_names = [''.join([str(x), str(i)])
```
WillAyd: Use `.format` instead of `.join`. I'd also suggest adding an underscore like we do in merges

pandas/core/groupby/groupby.py
```diff
@@ -2298,7 +2298,18 @@ def levels(self):
 
     @property
     def names(self):
-        return [ping.name for ping in self.groupings]
+        # GH 19029
+        # add suffix to level name in case they contain duplicates (GH 19029):
```
WillAyd: Can remove ref to GH here, since it's on the line above

pandas/tests/groupby/test_categorical.py
```diff
@@ -558,9 +558,13 @@ def test_as_index():
     result = df.groupby(['cat', s], as_index=False, observed=True).sum()
     tm.assert_frame_equal(result, expected)
 
-    # GH18872: conflicting names in desired index
-    with pytest.raises(ValueError):
+    # GH 19029: conflicitng names should not raise a value error anymore
+    raised = False
```
WillAyd: Rather than doing this you should say what the `expected` object is and compare to that the result of the groupby

pandas/tests/groupby/test_groupby.py
```diff
@@ -1674,3 +1674,42 @@ def test_tuple_correct_keyerror():
                                                           [3, 4]]))
     with tm.assert_raises_regex(KeyError, "(7, 8)"):
         df.groupby((7, 8)).mean()
+
+
+def test_dup_index_names():
+    # dup. index names in groupby operations should be renamed (GH 19029):
+    df = pd.DataFrame({'date': list(pd.date_range('5.1.2018', '5.3.2018')),
```
WillAyd: Don't think you need to wrap the date_range in a list (?)

pandas/tests/groupby/test_groupby.py
```diff
@@ -1674,3 +1674,42 @@ def test_tuple_correct_keyerror():
...
+                       'vals': list(range(3))})
+
+    mi = pd.MultiIndex.from_product([[5], [1, 2, 3]], names=['date0', 'date1'])
+    expected = pd.Series(data=list(range(3)), index=mi, name='vals')
+
+    failed = False
+    try:
+        result = df.groupby([df.date.dt.month, df.date.dt.day])['vals'].sum()
```
WillAyd: Same comment as before - don't need this try...catch block. For what it's worth though let's not duplicate tests, so either have this check here in its own method and remove from the other test, or just update the other test and leave this. See which one makes more sense

pandas/tests/groupby/test_groupby.py
```diff
@@ -1674,3 +1674,42 @@ def test_tuple_correct_keyerror():
...
+    except ValueError:
+        failed = True
+
+    assert failed is False
+
+    tm.assert_series_equal(result, expected)
+
+
+def test_empty_index_names():
```
WillAyd: What happens if say 2 out of 3 levels use `None` as a name - would these tests still hold?

WillAyd (changes requested):

pandas/core/groupby/groupby.py
```diff
@@ -2298,7 +2298,28 @@ def levels(self):
 
     @property
     def names(self):
-        return [ping.name for ping in self.groupings]
+        # add suffix to level name in case they contain duplicates (GH 19029):
```
jreback: this is super complicated, what exactly are you trying to do here?

pandas/tests/groupby/test_groupby.py
```diff
@@ -1674,3 +1674,44 @@ def test_tuple_correct_keyerror():
                                                           [3, 4]]))
     with tm.assert_raises_regex(KeyError, "(7, 8)"):
         df.groupby((7, 8)).mean()
+
+
+def test_dup_index_names():
+    # dup. index names in groupby operations should be renamed (GH 19029):
```
jreback: can you parameterize this

pandas/tests/reshape/test_pivot.py
```diff
@@ -1705,9 +1705,21 @@ def test_crosstab_with_numpy_size(self):
         tm.assert_frame_equal(result, expected)
 
     def test_crosstab_dup_index_names(self):
-        # GH 13279, GH 18872
+        # duplicated index name should get renamed (GH 19029)
         s = pd.Series(range(3), name='foo')
-        pytest.raises(ValueError, pd.crosstab, s, s)
+        failed = False
+        try:
```
jreback: if you are asserting that this works, simply collect the result and compare vs expected

jreback (changes requested):

pandas/core/groupby/groupby.py
guenteru: Hi. The goal here is to add a suffix to duplicate entries:
```python
# takes care of multiplicity:
['x', 'x', 'y', 'y']  => ['x_0', 'x_1', 'y_0', 'y_1']
```
Before the current version the code just enumerated all the entries regardless of their multiplicity:
```python
['x', 'x', 'y', 'y']  => ['x_0', 'x_1', 'y_2', 'y_3']
```
For some reason I thought that the current version would be better suited.
I can switch back to the old (and probably less confusing) version if you want.

jreback:

PR #21176, opened 2018-05-22
BUG: read_csv with specified kwargs

### LINKED ISSUE #21141: read_csv errors when low_memory=True, index_col is not None, and nrows=0
#### Code Sample

```python
import pandas as pd
from io import StringIO

data = """A,B,C,D,E
2000-01-03 00:00:00,0.980268513777,3.68573087906,-0.364216805298,-1.15973806169,foo
2000-01-04 00:00:00,1.04791624281,-0.0412318367011,-0.16181208307,0.212549316967,bar
2000-01-05 00:00:00,0.498580885705,0.731167677815,-0.537677223318,1.34627041952,baz
2000-01-06 00:00:00,1.12020151869,1.56762092543,0.00364077397681,0.67525259227,qux
2000-01-07 00:00:00,-0.487094399463,0.571454623474,-1.6116394093,0.103468562917,foo2"""

df = pd.read_csv(StringIO(data), low_memory=True, index_col=0, nrows=0)
```
#### Problem description

The above code results in `TypeError: 'NoneType' object is not iterable`. `read_csv` behaves correctly if `low_memory=False`, `index_col=None` or `nrows>0`.

Traceback:
```
  File "/home/peter/workspace/ray_env/lib/python3.6/site-packages/pandas/io/parsers.py", line 1848, in read
    data = self._reader.read(nrows)
  File "pandas/_libs/parsers.pyx", line 876, in pandas._libs.parsers.TextReader.read
  File "pandas/_libs/parsers.pyx", line 916, in pandas._libs.parsers.TextReader._read_low_memory
StopIteration

  File "pd_bug.py", line 12, in <module>
    df = pd.read_csv(StringIO(data), low_memory=True, index_col=0, nrows=0)
  File "/home/peter/workspace/ray_env/lib/python3.6/site-packages/pandas/io/parsers.py", line 678, in parser_f
    return _read(filepath_or_buffer, kwds)
  File "/home/peter/workspace/ray_env/lib/python3.6/site-packages/pandas/io/parsers.py", line 446, in _read
    data = parser.read(nrows)
  File "/home/peter/workspace/ray_env/lib/python3.6/site-packages/pandas/io/parsers.py", line 1036, in read
    ret = self._engine.read(nrows)
  File "/home/peter/workspace/ray_env/lib/python3.6/site-packages/pandas/io/parsers.py", line 1855, in read
    dtype=self.kwds.get('dtype'))
  File "/home/peter/workspace/ray_env/lib/python3.6/site-packages/pandas/io/parsers.py", line 3215, in _get_empty_meta
    data = [Series([], dtype=dtype[name]) for name in index_names]
TypeError: 'NoneType' object is not iterable
```

#### Expected Output
No error.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.13.0-36-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.23.0
pytest: 3.3.1
pip: 9.0.1
setuptools: 38.4.0
Cython: 0.27.3
numpy: 1.14.3
scipy: 1.0.0
pyarrow: 0.8.0
xarray: None
IPython: 6.2.1
sphinx: None
patsy: None
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: 0.4.0
matplotlib: 2.1.1
openpyxl: 2.5.3
xlrd: 1.1.0
xlwt: None
xlsxwriter: None
lxml: 4.2.0
bs4: 4.6.0
html5lib: 0.999999999
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

r00ta:
- [+] closes #21141 
- [+] tests added / passed
- [+] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] whatsnew entry

Solves the issue 21141.

### DIFF
--- a/doc/source/whatsnew/v0.23.2.txt
+++ b/doc/source/whatsnew/v0.23.2.txt
@@ -64,6 +64,7 @@ Bug Fixes
 
 **I/O**
 
+- Bug in :func:`read_csv` that caused it to incorrectly raise an error when ``nrows=0``, ``low_memory=True``, and ``index_col`` was not ``None`` (:issue:`21141`)
 -
 -
 
--- a/pandas/io/parsers.py
+++ b/pandas/io/parsers.py
@@ -3209,12 +3209,22 @@ def _get_empty_meta(columns, index_col, index_names, dtype=None):
             col = columns[k] if is_integer(k) else k
             dtype[col] = v
 
-    if index_col is None or index_col is False:
+    # Even though we have no data, the "index" of the empty DataFrame
+    # could for example still be an empty MultiIndex. Thus, we need to
+    # check whether we have any index columns specified, via either:
+    #
+    # 1) index_col (column indices)
+    # 2) index_names (column names)
+    #
+    # Both must be non-null to ensure a successful construction. Otherwise,
+    # we have to create a generic emtpy Index.
+    if (index_col is None or index_col is False) or index_names is None:
         index = Index([])
     else:
         data = [Series([], dtype=dtype[name]) for name in index_names]
         index = _ensure_index_from_sequences(data, names=index_names)
         index_col.sort()
+
         for i, n in enumerate(index_col):
             columns.pop(n - i)
 
--- a/pandas/tests/io/parser/common.py
+++ b/pandas/tests/io/parser/common.py
@@ -238,6 +238,21 @@ c,4,5
         out = self.read_csv(StringIO(data))
         tm.assert_frame_equal(out, expected)
 
+    def test_read_csv_low_memory_no_rows_with_index(self):
+        if self.engine == "c" and not self.low_memory:
+            pytest.skip("This is a low-memory specific test")
+
+        # see gh-21141
+        data = """A,B,C
+1,1,1,2
+2,2,3,4
+3,3,4,5
+"""
+        out = self.read_csv(StringIO(data), low_memory=True,
+                            index_col=0, nrows=0)
+        expected = DataFrame(columns=["A", "B", "C"])
+        tm.assert_frame_equal(out, expected)
+
     def test_read_csv_dataframe(self):
         df = self.read_csv(self.csv1, index_col=0, parse_dates=True)
         df2 = self.read_table(self.csv1, sep=',', index_col=0,


pandas/tests/io/parser/common.py
```diff
@@ -238,6 +238,17 @@ def test_csv_mixed_type(self):
         out = self.read_csv(StringIO(data))
         tm.assert_frame_equal(out, expected)
 
+    def test_csv_index_col_and_nrows(self):
+        data = """A,B,C
```
jschendel: Can you add a comment with the GitHub issue number here?

pandas/tests/io/parser/common.py
```diff
@@ -238,6 +238,17 @@ def test_csv_mixed_type(self):
...
+1,1,1,2
+2,2,3,4
+3,3,4,5
+"""
+        out = self.read_csv(StringIO(data), low_memory=True, index_col=0,
+                            nrows=0)
+        tm.assert_index_equal(out.columns, pd.Index(['A', 'B', 'C']))
```
mroeschke: Could you instead specify an expected `DataFrame` and use `tm.assert_frame_equal` to the output?

```
result = self.read_csv(...)
expected = pd.DataFrame(...)

```

pandas/io/parsers.py
```diff
@@ -3208,8 +3208,7 @@ def _get_empty_meta(columns, index_col, index_names, dtype=None):
         for k, v in compat.iteritems(_dtype):
             col = columns[k] if is_integer(k) else k
             dtype[col] = v
-
-    if index_col is None or index_col is False:
+    if index_col is None or index_col is False or index_names is None:
```
WillAyd: Is it possible to simplify this to just be `if not (index_col or index_names)`?

r00ta: Nope. ìndex_col`can be a list ( like in the issue #21141 ) and `bool([0] or None)` is `True`

jreback: can you add a comment here on what is going on

jreback (changes requested):

pandas/io/parsers.py
cgopalan: If I can help, the reason the error happens is because `index_names` is `None` and it is being iterated over in https://github.com/r00ta/pandas/blob/master/pandas/io/parsers.py#L3215
To prevent this, there's an additional check for it being not None. 
@r00ta you could add a comment here saying `# also create empty index if index_names is None`
@jreback is that clear enough?

pandas/tests/io/parser/common.py
cgopalan: Also, I think it needs a more descriptive test name - how about `test_read_csv_low_memory_no_index_cols_rows`?

pandas/io/parsers.py
jreback: I mean in the code itself, e.g. why we are checking these things

jreback (changes requested):
some comments from @mroeschke and need a whatsnew entry

pandas/io/parsers.py
cgopalan: @WillAyd if it was possible to simplify it, it would be `if not (index_col and index_names)` but I suspect it would be not be easy due to the complexity of the code. They can be `None` or other values like `0` and the code was checking for specific things.

cgopalan: @jreback the only thing that changed here is the addition of `index_names` in the if statement so as to divert the code flow into the creation of the empty index. Hard to determine why the check is being made here for the other values. Maybe @r00ta can explain further.

r00ta: @cgopalan exactly. According to the issue #21141 if `nrow=0, `index_names`is `None` it raises an Exception. To avoid this i check that `index_names` is not `None`. If it's better to handle this corner case in another way let me know!

cgopalan: @r00ta I tried other ways but they seemed to go into other code flows with complicated things further. I am not a member of the pandas team, just trying to contribute to this PR by reviewing it. AFAIK, the code is fine with me. Just need to address the comments on commenting the code and on the test.

WillAyd: My hesitation is that this is already somewhat convoluted, and when that's the case typically adding things just because "they work" tends to yield some other issue down the road.

I'm not an expert on this module but can you check the following:
- There is a function called `_clean_index_names` that I assume gets called before all of this. If there's no index column it returns `None`, but what if that returned an empty list instead? 
- There is a function called `_is_index_col` in the module which appears to do the first 2/3 items of this conditional - should we be using that here?

AFAICT with the former the "cleaning" is supposed to help when the `index_cols` are `None` in instances like below, so maybe that function just shouldn't be returning `None`?

https://github.com/r00ta/pandas/blob/1d680be82e757d79fbf3e3d6d87de3586ea5e06a/pandas/io/parsers.py#L1767

https://github.com/r00ta/pandas/blob/1d680be82e757d79fbf3e3d6d87de3586ea5e06a/pandas/io/parsers.py#L2112

cgopalan: @WillAyd if i change the code such that `index_names` is `[]` instead of `None` it branches to the `else` condition and gives the following error at [this line](https://github.com/r00ta/pandas/blob/1d680be82e757d79fbf3e3d6d87de3586ea5e06a/pandas/io/parsers.py#L3215)
`ValueError: Must pass non-zero number of levels/labels`

It seems like the right branch would be to go the `if index_col is None or index_col is False:` but since the `index_col` passed is an int, this is being bypassed.

The real culprit seems to be the fact that `nrows` is zero. 

Why does pandas allow `nrows=0` as a value for `read_csv`? Shouldnt it send back an error saying nrows should be greater than zero? Is there a value to reading 0 rows into a dataframe?

If we do allow `nrows` to be zero, then we can alter `index_col` to be `False` since any other value for it seems meaningless. I can look further into that, but just was curious about `nrows` allowing zero.

WillAyd: I can't think of a good use case for `nrows` to be 0 so alternately could just raise if that is the case instead of all of the changes here (unless someone else does have a use case for that)

r00ta: @WillAyd The parser already checks that the parameter `nrows`is not a negative integer ( https://github.com/pandas-dev/pandas/blob/master/pandas/io/parsers.py#L382 ). It raises `ValueError: 'nrows' must be an integer >=0`. For this reason i thought that by contract the case `nrows=0` is allowable

cgopalan: @WillAyd @r00ta The solution where we validate nrows to be >= 1 looks good to me. This will take care of this weird corner case.

The changes for that can be made [here](https://github.com/pandas-dev/pandas/blob/master/pandas/tests/io/test_excel.py#L998) to pass in a min_val argument of 1.
Also the tests need to be changed [here](https://github.com/pandas-dev/pandas/blob/master/pandas/tests/io/parser/common.py#L367) and [here](https://github.com/pandas-dev/pandas/blob/master/pandas/tests/io/test_excel.py#L998)

WillAyd: @cgopalan do you want to submit that as a separate PR? I think that would be preferable to this.

Just glancing at your code / tests make sure you cover the case that `nrows` equals 0

gfyoung:
Superseded by https://github.com/pandas-dev/pandas/pull/21431.  I think given the discussions above, I don't think there is much objection to that...

(if there is, please comment, and we can reopen no problem)

gfyoung:
Per https://github.com/pandas-dev/pandas/pull/21431#issuecomment-396761132, we're back!

gfyoung:
Marking this for `0.23.2` because if this still isn't merged by then, I'll close it out.

cgopalan:
@gfyoung the main difference between the `low_memory=True` and `low_memory=False` case seems to be that in the True case, the read method [here](https://github.com/r00ta/pandas/blob/1d680be82e757d79fbf3e3d6d87de3586ea5e06a/pandas/io/parsers.py#L1848) raises an exception, so it goes into `_get_empty_meta`. Would it be good to check for the combination of args before that read method is called? Or would it be somewhere much earlier?

gfyoung:
The patch should be in `_get_empty_meta`.

cgopalan:
`_get_empty_meta` doesn't to have any knowledge of `n_rows`. How exactly do you propose to patch it up there?

cgopalan:
@gfyoung @r00ta one possible solution is to add `nrows` parameter to `_get_empty_meta` and then check `nrows=0` instead of checking `index_names` in the changed if clause. Would that be an acceptable solution?

gfyoung:
@cgopalan : The handling I think should be independent of `nrows` as it is in the current PR.

cgopalan:
@gfyoung so the current changes in the PR are fine?

gfyoung:
@jreback : All is green.

gfyoung:
@jreback : Friendly ping.

pandas/tests/io/parser/common.py
```diff
@@ -238,6 +238,21 @@ def test_csv_mixed_type(self):
         out = self.read_csv(StringIO(data))
         tm.assert_frame_equal(out, expected)
 
+    def test_read_csv_low_memory_no_rows_with_index(self):
```
jreback: I think there is a class that you can put the will specifically only have low_memory set, just search for self.low_memory

doc/source/whatsnew/v0.23.2.txt
```diff
@@ -59,6 +59,7 @@ Bug Fixes
 
 **I/O**
 
+- Bug in :func:`read_csv` when ``nrows=0``, ``low_memory=True``, and ``index_col`` was not ``None`` (:issue:`21141`)
```
jreback: can you add a bit about that has changed from a user POV

jreback (changes requested):
minor comments.

pandas/tests/io/parser/common.py
gfyoung: Indeed, there is, but I deliberately allowed this test to run for both the C and Python engines.  As the Python engine doesn't support `low_memory`, I need to do this.

doc/source/whatsnew/v0.23.2.txt
gfyoung: Sure thing.  Done.

jreback (approved):

PR #21185, opened 2018-05-24
ENH: add in extension dtype registry

jreback:
precursor to #21160

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -10,6 +10,22 @@ New features
 
 - ``ExcelWriter`` now accepts ``mode`` as a keyword argument, enabling append to existing workbooks when using the ``openpyxl`` engine (:issue:`3441`)
 
+.. _whatsnew_0240.enhancements.extension_array_operators
+
+``ExtensionArray`` operator support
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A ``Series`` based on an ``ExtensionArray`` now supports arithmetic and comparison
+operators. (:issue:`19577`). There are two approaches for providing operator support for an ``ExtensionArray``:
+
+1. Define each of the operators on your ``ExtensionArray`` subclass.
+2. Use an operator implementation from pandas that depends on operators that are already defined
+   on the underlying elements (scalars) of the ``ExtensionArray``.
+
+See the :ref:`ExtensionArray Operator Support
+<extending.extension.operator>` documentation section for details on both
+ways of adding operator support.
+
 .. _whatsnew_0240.enhancements.other:
 
 Other Enhancements
@@ -118,6 +134,7 @@ Datetimelike API Changes
 
 - For :class:`DatetimeIndex` and :class:`TimedeltaIndex` with non-``None`` ``freq`` attribute, addition or subtraction of integer-dtyped array or ``Index`` will return an object of the same class (:issue:`19959`)
 - :class:`DateOffset` objects are now immutable. Attempting to alter one of these will now raise ``AttributeError`` (:issue:`21341`)
+- :class:`PeriodIndex` subtraction of another ``PeriodIndex`` will now return an object-dtype :class:`Index` of :class:`DateOffset` objects instead of raising a ``TypeError`` (:issue:`20049`)
 
 .. _whatsnew_0240.api.extension:
 
@@ -127,6 +144,7 @@ ExtensionType Changes
 - ``ExtensionDtype`` has gained the ability to instantiate from string dtypes, e.g. ``decimal`` would instantiate a registered ``DecimalDtype``; furthermore
   the ``ExtensionDtype`` has gained the method ``construct_array_type`` (:issue:`21185`)
 - The ``ExtensionArray`` constructor, ``_from_sequence`` now take the keyword arg ``copy=False`` (:issue:`21185`)
+- Bug in :meth:`Series.get` for ``Series`` using ``ExtensionArray`` and integer index (:issue:`21257`)
 - :meth:`Series.combine()` works correctly with :class:`~pandas.api.extensions.ExtensionArray` inside of :class:`Series` (:issue:`20825`)
 - :meth:`Series.combine()` with scalar argument now works for any function type (:issue:`21248`)
 -
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -45,14 +45,13 @@ Other Enhancements
 
 .. _whatsnew_0240.api_breaking:
 
-
 Backwards incompatible API changes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. _whatsnew_0240.api.datetimelike.normalize:
 
 Tick DateOffset Normalize Restrictions
---------------------------------------
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 Creating a ``Tick`` object (:class:`Day`, :class:`Hour`, :class:`Minute`,
 :class:`Second`, :class:`Milli`, :class:`Micro`, :class:`Nano`) with
@@ -129,13 +128,6 @@ Previous Behavior:
     In [3]: pi - pi[0]
     Out[3]: Int64Index([0, 1, 2], dtype='int64')
 
-Datetimelike API Changes
-^^^^^^^^^^^^^^^^^^^^^^^^
-
-- For :class:`DatetimeIndex` and :class:`TimedeltaIndex` with non-``None`` ``freq`` attribute, addition or subtraction of integer-dtyped array or ``Index`` will return an object of the same class (:issue:`19959`)
-- :class:`DateOffset` objects are now immutable. Attempting to alter one of these will now raise ``AttributeError`` (:issue:`21341`)
-- :class:`PeriodIndex` subtraction of another ``PeriodIndex`` will now return an object-dtype :class:`Index` of :class:`DateOffset` objects instead of raising a ``TypeError`` (:issue:`20049`)
-
 .. _whatsnew_0240.api.extension:
 
 ExtensionType Changes
@@ -154,10 +146,10 @@ ExtensionType Changes
 Other API Changes
 ^^^^^^^^^^^^^^^^^
 
-.. _whatsnew_0240.api.other.incompatibilities:
+.. _whatsnew_0240.api.incompatibilities:
 
 Series and Index Data-Dtype Incompatibilities
----------------------------------------------
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 ``Series`` and ``Index`` constructors now raise when the
 data is incompatible with a passed ``dtype=`` (:issue:`15832`)
@@ -179,6 +171,18 @@ Current Behavior:
     Out [4]:
     ...
     OverflowError: Trying to coerce negative values to unsigned integers
+
+Datetimelike API Changes
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+- For :class:`DatetimeIndex` and :class:`TimedeltaIndex` with non-``None`` ``freq`` attribute, addition or subtraction of integer-dtyped array or ``Index`` will return an object of the same class (:issue:`19959`)
+- :class:`DateOffset` objects are now immutable. Attempting to alter one of these will now raise ``AttributeError`` (:issue:`21341`)
+- :class:`PeriodIndex` subtraction of another ``PeriodIndex`` will now return an object-dtype :class:`Index` of :class:`DateOffset` objects instead of raising a ``TypeError`` (:issue:`20049`)
+
+.. _whatsnew_0240.api.other:
+
+Other API Changes
+^^^^^^^^^^^^^^^^^
 
 - :class:`DatetimeIndex` now accepts :class:`Int64Index` arguments as epoch timestamps (:issue:`20997`)
 - Invalid construction of ``IntervalDtype`` will now always raise a ``TypeError`` rather than a ``ValueError`` if the subdtype is invalid (:issue:`21185`)
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -125,7 +125,7 @@ ExtensionType Changes
 ^^^^^^^^^^^^^^^^^^^^^
 
 - ``ExtensionArray`` has gained the abstract methods ``.dropna()`` and ``.append()``, and attribute ``array_type`` (:issue:`21185`)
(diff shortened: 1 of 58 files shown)


doc/source/whatsnew/v0.24.0.txt
```diff
@@ -14,14 +14,22 @@ Other Enhancements
...
 Backwards incompatible API changes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. _whatsnew_0240.api.extension:
+
+ExtensionType Changes
+^^^^^^^^^^^^^^^^^^^^^
+
+- ``ExtensionArray`` has gained the abstract methods ``.dropna()`` and ``.append()``, and attribute ``array_type`` (:issue:`21185`)
+- ``ExtensionDtype`` has gained the ability to instantiate from string dtypes, e.g. ``decimal`` would instaniate a registered ``DecimalDtype`` (:issue:`21185`)
```
jbrockmendel: typo instantiate

Dr-Irv (commented):
It seems to me that there are changes in this PR that go beyond just adding in the extension dtype registry.  Wasn't your goal to just make this PR about the registry?

jreback:
split out ops to a separate PR #21191

pandas/core/dtypes/dtypes.py
```diff
@@ -2,12 +2,70 @@
...
+    These are tried in order for inference.
+    """
+    dtypes = OrderedDict()
+
+    @classmethod
+    def register(self, dtype, constructor=None):
+        """
+        Parameters
+        ----------
+        dtype : PandasExtension Dtype
```
jorisvandenbossche: Can you document constructor parameter? 
and "PandasExtension Dtype" -> "ExtensionDtype"

pandas/core/dtypes/dtypes.py
```diff
@@ -2,12 +2,70 @@
...
+        """
+        if not issubclass(dtype, (PandasExtensionDtype, ExtensionDtype)):
```
jorisvandenbossche: Shouldn't ExtensionDtype be enough? (our internals one subclass both I think?)

pandas/core/dtypes/dtypes.py
```diff
@@ -564,6 +622,17 @@ def construct_from_string(cls, string):
...
+    @classmethod
+    def construct_from_string_strict(cls, string):
+        """
+        Strict construction from a string, raise a TypeError if not
+        possible
+        """
+        if string.startswith('period[') or string.startswith('Period['):
+            # do not parse string like U as period[U]
+            return PeriodDtype.construct_from_string(string)
+        raise TypeError("could not construct PeriodDtype")
```
jorisvandenbossche: Should we consider rather changing `construct_from_string` itself? As those don't really adhere to the ExtensionDtype requirements I think?

If we keep both, I would make this one private.

pandas/tests/extension/base/ops.py
```diff
@@ -0,0 +1,6 @@
+from .base import BaseExtensionTests
+
+
+class BaseOpsTests(BaseExtensionTests):
+    """Various Series and DataFrame ops methos."""
+    pass
```
jorisvandenbossche: This is a left-over from splitting it from the previous PR?

jorisvandenbossche (commented):
Didn't look at all the tests yet.

In addition:

- We need to discuss what is the exact contract for this registry and `array_type`: where is it used, and how is it used? 
  I think now it is used in constructors or astype if you specify a (string) dtype? But then it calls the class constructor of the extension array? Shouldn't this rather be the `_from_sequence` method ?
- Can you add some explanation of the registry to the extension array documentation?
- I think this also gives an enhancement for the existing types (interval, datetimetz, period, ..) that you can now specify string dtypes? (since they are added to the registry) In that case, can you add tests for this and a whatsnew notice?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -14,14 +14,22 @@ Other Enhancements
...
 
 Backwards incompatible API changes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+.. _whatsnew_0240.api.extension:
+
+ExtensionType Changes
+^^^^^^^^^^^^^^^^^^^^^
+
+- ``ExtensionArray`` has gained the abstract methods ``.dropna()`` and ``.append()``, and attribute ``array_type`` (:issue:`21185`)
```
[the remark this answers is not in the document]
jreback: no, these are necessary for followons. The api is not fleshed out enough.

pandas/core/arrays/base.py
```diff
@@ -82,14 +85,16 @@ class ExtensionArray(object):
...
+    def _from_sequence(cls, scalars, copy=False):
         """Construct a new ExtensionArray from a sequence of scalars.
 
         Parameters
         ----------
         scalars : Sequence
             Each element will be an instance of the scalar type for this
             array, ``cls.dtype.type``.
+        copy : boolean, default True
+            if True, copy the underlying data
```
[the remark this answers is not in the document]
jreback: same guarantee with copy we have already, if its True, then copy if possible.

pandas/core/dtypes/cast.py
```diff
@@ -647,6 +647,11 @@ def conv(r, dtype):
 def astype_nansafe(arr, dtype, copy=True):
     """ return a view if copy is False, but
         need to be very careful as the result shape could change! """
+
+    # dispatch on extension dtype if needed
+    if is_extension_array_dtype(dtype):
+        return dtype.array_type._from_sequence(arr, copy=copy)
```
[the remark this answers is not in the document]
jreback: this validation is a contract of the Array itself

pandas/core/dtypes/dtypes.py
jreback: not yet, are extension dtypes are not all ExtensiDtypes yet (this is why we have the whole PandasExtensionDtype bizness)

pandas/core/dtypes/dtypes.py
jreback: done

jreback:
all fixed up @jorisvandenbossche  (except for docs on Registry in Extension section)

jreback:
see if can make this dispatch to create_array_type as a function

jreback:
updated docs

@jorisvandenbossche @TomAugspurger if any final comments.

pandas/tests/extension/base/constructors.py
```diff
@@ -45,3 +46,14 @@ def test_series_given_mismatched_index_raises(self, data):
         msg = 'Length of passed values is 3, index implies 5'
         with tm.assert_raises_regex(ValueError, msg):
             pd.Series(data[:3], index=[0, 1, 2, 3, 4])
+
+    def test_from_dtype(self, data):
+        # construct from our dtype & string dtype
+        dtype = data.dtype
+
+        expected = pd.Series(data)
+        result = pd.Series(np.array(data), dtype=dtype)
```
jorisvandenbossche: Do we have a restriction on what the result of `__array__` should be? Maybe not necessarily the scalars? 

So maybe better to do `list(data)` or `data.astype(object)`

pandas/tests/extension/category/test_categorical.py
```diff
@@ -52,7 +52,9 @@ def data_for_grouping():
 
 
 class TestDtype(base.BaseDtypeTests):
-    pass
+
+    def test_array_type_with_arg(self, data, dtype):
+        assert dtype.construct_array_type('foo') is Categorical
```
jorisvandenbossche: Why should this not raise?

pandas/core/dtypes/base.py
```diff
@@ -156,6 +161,22 @@ def name(self):
         """
         raise AbstractMethodError(self)
 
+    @classmethod
+    def construct_array_type(cls, array=None):
+        """Return the array type associated with this dtype
+
+        Parameters
+        ----------
+        array : array-like, optional
```
jorisvandenbossche: Do we actually have a use case for this at the moment ?

pandas/tests/extension/decimal/array.py
```diff
@@ -27,7 +41,7 @@ def construct_from_string(cls, string):
 class DecimalArray(ExtensionArray):
     dtype = DecimalDtype()
 
-    def __init__(self, values):
+    def __init__(self, values, copy=False):
```
jorisvandenbossche: this change should not be needed?

pandas/tests/extension/decimal/test_decimal.py
```diff
@@ -92,15 +92,21 @@ def assert_frame_equal(self, left, right, *args, **kwargs):
...
     pass
 
 
 class TestConstructors(BaseDecimal, base.BaseConstructorsTests):
-    pass
+
+    @pytest.mark.xfail(reason="not implemented constructor from dtype")
+    def test_from_dtype(self, data):
+        # construct from our dtype & string dtype
+        pass
```
jorisvandenbossche: Can you actually implement this? (only change needed would be to register DecimalDtype I think?) 
As that way we actually have a test for external dtypes registering?

pandas/tests/extension/decimal/test_decimal.py
```diff
@@ -147,6 +153,10 @@ class TestGroupby(BaseDecimal, base.BaseGroupbyTests):
     pass
 
 
+# TODO(extension)
+@pytest.mark.xfail(reason=(
+    "raising AssertionError as this is not implemented, "
+    "though easy enough to do"))
```
jorisvandenbossche: I think you can change the AssertionError to a ValueError in the code, and then we can still test this.

pandas/tests/extension/json/test_json.py
```diff
@@ -130,7 +132,11 @@ def test_custom_asserts(self):
 
 
 class TestConstructors(BaseJSON, base.BaseConstructorsTests):
-    pass
+
+    @pytest.mark.xfail(reason="not implemented constructor from dtype")
+    def test_from_dtype(self, data):
+        # construct from our dtype & string dtype
+        pass
```
jorisvandenbossche: Can we test here that an appropriate error is raised (instead of skipping the test) ?

pandas/tests/dtypes/test_dtypes.py
```diff
@@ -767,3 +773,24 @@ def test_update_dtype_errors(self, bad_dtype):
...
+     PeriodDtype, IntervalDtype])
+def test_registry(dtype):
+    assert dtype in registry.dtypes
+
+
+@pytest.mark.parametrize(
+    'dtype, expected',
+    [('int64', None),
+     ('interval', IntervalDtype()),
+     ('interval[int64]', IntervalDtype()),
```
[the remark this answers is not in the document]
jorisvandenbossche: (although that is maybe covered with the period or datetime64 one below. Is `'interval[datetime64[ns]]'` already tested for IntervalDtype?)

pandas/core/dtypes/base.py
```diff
@@ -109,6 +109,11 @@ class ExtensionDtype(_DtypeOpsMixin):
     * name
     * construct_from_string
 
+    Optionally one can override construct_array_type for construction
+    with the name of this dtype via the Registry
```
jorisvandenbossche: I would give some additional explanation of what the Registry is, because now this is not explained here?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -30,12 +30,22 @@ Datetimelike API Changes
 
 - For :class:`DatetimeIndex` and :class:`TimedeltaIndex` with non-``None`` ``freq`` attribute, addition or subtraction of integer-dtyped array or ``Index`` will return an object of the same class (:issue:`19959`)
 
+.. _whatsnew_0240.api.extension:
+
+ExtensionType Changes
+^^^^^^^^^^^^^^^^^^^^^
+
+- ``ExtensionArray`` has gained the abstract methods ``.dropna()`` and ``.append()`` (:issue:`21185`)
```
jorisvandenbossche: As I said before, I am convinced we should necessarily add those methods (certainly append), so would personally leave this for another PR to discuss

jorisvandenbossche (commented):
Added some more comments. 

From my previous review:

> We need to discuss what is the exact contract for this registry and array_type: where is it used, and how is it used?

Can you add some explanation about this to the docs/docstrings? (eg that it relies on `_from_sequence`)

pandas/core/dtypes/dtypes.py
```diff
@@ -8,6 +8,64 @@
 from .base import ExtensionDtype, _DtypeOpsMixin
 
 
+class Registry(object):
```
TomAugspurger: Without looking at the uses yet, could we simplify this a by just allowing string lookup? Ideally, `registry` would be a simple class holding a dict mapping `dtype.name -> ExtensionDtype`.

pandas/core/indexes/interval.py
```diff
@@ -800,7 +800,7 @@ def astype(self, dtype, copy=True):
     @cache_readonly
     def dtype(self):
         """Return the dtype object of the underlying data"""
-        return IntervalDtype.construct_from_string(str(self.left.dtype))
+        return IntervalDtype(str(self.left.dtype))
```
TomAugspurger: `left.dtype.name`? I'm not sure when these differ, but using `.name` seems safer.

pandas/core/dtypes/dtypes.py
jorisvandenbossche: I think the current code also supports finding the dtype for eg `'interval[int64]'` and not just `interval` (so parametrized versions of the strings)

TomAugspurger: Oh, and I suppose we want that to support `.astype('interval[int64]')`. That's fair...

jreback:
ok all fixed up. @jorisvandenbossche @TomAugspurger if you want to look.

jreback:
@Dr-Irv you don;t *need* to register the EA types, this is *only* for the purpose of translating a string dtype name -> EA dtype (e.g. 'categorical' -> Categorical), basically for convenience.

so sure you *could* have a collision, but then user would have to deal with that (this wouldn't show up on pandas because we won't have colliding names).

jreback:
@Dr-Irv I had this an OrderdedDict but its actually not necessary, though maybe makes it more intuitve.

TomAugspurger (commented):
Aside from adding EA.append, I'm +1 on this.

Given that NumPy doesn't implement it, I don't think we should either.

jreback:
I had removed append FYI (it should not longer be in this PR)

TomAugspurger:
I see it at https://github.com/pandas-dev/pandas/pull/21185/files#diff-99633b1041c2ea6283aa3d5bcf79e8cdR584

Maybe need to push?

jreback:
pushed, yeah had it on the wrong branch.

jreback:
any final comments. going to rebase.

jreback:
@jorisvandenbossche @TomAugspurger

jreback:
merging.

PR #21191, opened 2018-05-24
ENH: extension ops

jreback:
builds on #21185 for extension array ops
pre-cursor to #21160

### DIFF
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -101,7 +101,7 @@ def all_arithmetic_operators(request):
 def all_compare_operators(request):
     """
     Fixture for dunder names for common compare operations
-   """
+    """
     return request.param
 
 
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -92,7 +92,16 @@ if not PY3:
 def all_arithmetic_operators(request):
     """
     Fixture for dunder names for common arithmetic operations
+   """
+    return request.param
+
+
+@pytest.fixture(params=['__eq__', '__ne__', '__le__',
+                        '__lt__', '__ge__', '__gt__'])
+def all_compare_operators(request):
     """
+    Fixture for dunder names for common compare operations
+   """
     return request.param
 
 
--- a/pandas/core/arrays/base.py
+++ b/pandas/core/arrays/base.py
@@ -681,6 +681,7 @@ class ExtensionArray(object):
         cls.__rpow__ = cls._make_arithmetic_op(ops.rpow)
         cls.__pow__ = cls._make_arithmetic_op(operator.pow)
         cls.__mod__ = cls._make_arithmetic_op(operator.mod)
+        cls.__rmod__ = cls._make_arithmetic_op(ops.rmod)
         cls.__floordiv__ = cls._make_arithmetic_op(operator.floordiv)
         cls.__rfloordiv__ = cls._make_arithmetic_op(ops.rfloordiv)
         cls.__truediv__ = cls._make_arithmetic_op(operator.truediv)
--- a/pandas/core/arrays/base.py
+++ b/pandas/core/arrays/base.py
@@ -690,6 +690,7 @@ class ExtensionArray(object):
             cls.__rdiv__ = cls._make_arithmetic_op(ops.rdiv)
 
         cls.__divmod__ = cls._make_arithmetic_op(divmod)
+        cls.__rdivmod__ = cls._make_arithmetic_op(ops.rdivmod)
 
     @classmethod
     def make_comparison_op(cls, op):
--- a/pandas/core/arrays/base.py
+++ b/pandas/core/arrays/base.py
@@ -9,6 +9,9 @@ import numpy as np
 
 from pandas.errors import AbstractMethodError
 from pandas.compat.numpy import function as nv
+from pandas.compat import set_function_name, PY3
+from pandas.core import ops
+import operator
 
 _not_implemented_message = "{} does not implement {}."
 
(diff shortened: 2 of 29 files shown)


pandas/conftest.py
```diff
@@ -92,7 +92,16 @@ def observed(request):
...
+   """
+    return request.param
+
+
+@pytest.fixture(params=['__eq__', '__ne__', '__le__',
+                        '__lt__', '__ge__', '__gt__'])
+def all_compare_operators(request):
     """
+    Fixture for dunder names for common compare operations
+   """
```
jbrockmendel: indentation

pandas/core/arrays/base.py
```diff
@@ -610,3 +656,54 @@ def _ndarray_values(self):
...
+        """ add in numeric methods """
+        cls.__add__ = cls._make_arithmetic_op(operator.add)
+        cls.__radd__ = cls._make_arithmetic_op(ops.radd)
+        cls.__sub__ = cls._make_arithmetic_op(operator.sub)
+        cls.__rsub__ = cls._make_arithmetic_op(ops.rsub)
+        cls.__mul__ = cls._make_arithmetic_op(operator.mul)
+        cls.__rmul__ = cls._make_arithmetic_op(ops.rmul)
+        cls.__rpow__ = cls._make_arithmetic_op(ops.rpow)
+        cls.__pow__ = cls._make_arithmetic_op(operator.pow)
+        cls.__mod__ = cls._make_arithmetic_op(operator.mod)
```
jbrockmendel: `__divmod__`

pandas/core/arrays/base.py
```diff
@@ -610,3 +656,54 @@ def _ndarray_values(self):
...
+        cls.__floordiv__ = cls._make_arithmetic_op(operator.floordiv)
+        cls.__rfloordiv__ = cls._make_arithmetic_op(ops.rfloordiv)
+        cls.__truediv__ = cls._make_arithmetic_op(operator.truediv)
+        cls.__rtruediv__ = cls._make_arithmetic_op(ops.rtruediv)
+        if not PY3:
+            cls.__div__ = cls._make_arithmetic_op(operator.div)
+            cls.__rdiv__ = cls._make_arithmetic_op(ops.rdiv)
+
+        cls.__divmod__ = cls._make_arithmetic_op(divmod)
```
jbrockmendel: rdivmod?

pandas/core/ops.py
```diff
@@ -1003,8 +1003,18 @@ def _arith_method_SERIES(cls, op, special):
...
+        # handle extension array ops
+        # TODO(extension)
+        # the ops *between* non-same-type extension arrays are not
+        # very well defined
+        if (is_extension_array_dtype(x) or is_extension_array_dtype(y)):
+            if (op_name.startswith('__r') and not
+                    is_extension_array_dtype(y) and not
+                    is_scalar(y)):
+                y = x.__class__._from_sequence(y)
+            return op(x, y)
```
jbrockmendel: Is it really necessary to put this in the closure?  It seems like this case could be handled directly on line 1061.

jbrockmendel:
Looks reasonable, with some comments on organization in `core.ops`.  I haven't looked at the tests closely yet, since I think a lot of them will disappear from the diff after #21185.

pandas/core/ops.py
jreback: the way the r ops are defined is not compatible with some types, pretty much if you have a 

``rop(et, ndarray)`` will fail for some types, e.g. say ndarray is a datetime64[ns] then et is an integer like the precedence is wrong (IOW need to defer to the op here)

pandas/core/ops.py
```diff
@@ -1136,6 +1150,14 @@ def na_op(x, y):
             # The `not is_scalar(y)` check excludes the string "category"
             return op(y, x)
 
+        # handle extension array ops
+        # TODO(extension)
+        # the ops *between* non-same-type extension arrays are not
+        # very well defined
+        elif (is_extension_array_dtype(x) or
+                is_extension_array_dtype(y)):
+            return op(x, y)
```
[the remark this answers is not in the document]
jreback: if you notice categorical is *already* here, this is just another case, not sure how we could dispatch on this.

pandas/core/ops.py
jbrockmendel: I think I understand what you're saying, but that it isn't what I'm getting at.  Let me try to rephrase:

In `wrapper` below, there is currently (in the PR)

```
        elif (is_extension_array_dtype(left) or
                is_extension_array_dtype(right)):
            pass
```

I'm suggesting that instead of `pass` there, the code above (currently 1006-1015) could go there.  The main reason why this might _not_ be appropriate is if the try/except in `safe_na_op` is relevant for this case.

pandas/core/ops.py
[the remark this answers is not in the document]
jbrockmendel: Yah, categorical has a few characteristics that make it a difficult case in this function.  I think you're right that (for now at least) _part_ of this is in the right place.  The thing to do here is to:

- take the `is_extension_array_dtype(x)` part of this condition and put it down on L1231 (right after the `is_categorical_dtype(self)` block) and mimic the categorical_dtype case there
- keep the `is_extension_array_dtype(y)` case as is here (should the op be reversed?)
- check if this needs a `and not is_scalar(y)` like the categorical above for some corner case
- consider merging the categorical_dtype blocks into this more general case

pandas/core/ops.py
```diff
@@ -1035,14 +1059,21 @@ def safe_na_op(lvalues, rvalues):
...
     def wrapper(left, right):
-
         if isinstance(right, ABCDataFrame):
             return NotImplemented
 
         left, right = _align_method_SERIES(left, right)
         res_name = get_op_result_name(left, right)
 
-        if is_datetime64_dtype(left) or is_datetime64tz_dtype(left):
+        if is_categorical_dtype(left):
```
jbrockmendel: This will be made cleaner if/when we implement these ops in Categorical itself (they'll just raise the TypeError below)

jreback:
@jbrockmendel note I haven't actually changed this much (IOW for your comments) yet.

jbrockmendel:
@jreback am I right in thinking that after #21885 goes through the diff for this PR will be much smaller (and therefore easier to review)?

jreback:
yes

jreback:
going to incude directly in #21160

Issue #8953, opened 2014-12-01
Use multi-row inserts for massive speedups on to_sql over high latency connections


maxgrenderjones:
I have been trying to insert ~30k rows into a mysql database using pandas-0.15.1, oursql-0.9.3.1 and sqlalchemy-0.9.4. Because the machine is as across the atlantic from me, calling `data.to_sql` was taking >1 hr to insert the data. On inspecting with wireshark, the issue is that it is sending an insert for every row, then waiting for the ACK before sending the next, and, long story short, the ping times are killing me.

However, following the instructions from [SQLAlchemy](http://docs.sqlalchemy.org/en/latest/core/dml.html#sqlalchemy.sql.expression.Insert.values), I changed 

``` python
def _execute_insert(self, conn, keys, data_iter):
    data = [dict((k, v) for k, v in zip(keys, row)) for row in data_iter]
    conn.execute(self.insert_statement(), data)
```

to 

``` python
def _execute_insert(self, conn, keys, data_iter):
    data = [dict((k, v) for k, v in zip(keys, row)) for row in data_iter]
    conn.execute(self.insert_statement().values(data))
```

and the entire operation completes in less than a minute. (To save you a click, the difference is between multiple calls to `insert into foo (columns) values (rowX)` and one massive `insert into foo (columns) VALUES (row1), (row2), row3)`). Given how often people are likely to use pandas to insert large volumes of data, this feels like a huge win that would be great to be included more widely.

Some challenges:
- Not every database supports multirow inserts (SQLite and SQLServer didn't in the past, though they do now). I don't know how to check for this via SQLAlchemy
- The MySQL server I was using didn't allow me to insert the data all in one go, I had to set the chunksize (5k worked fine, but I guess the full 30k was too much). If we made this the default insert, most people would have to add a chunk size (which might be hard to calculate, as it might be determined by the maximum packet size of the server).

The easiest way to do this, would be to add a `multirow=` boolean parameter (default `False`) to the `to_sql` function, and then leave the user responsible for setting the chunksize, but perhaps there's a better way?

Thoughts?

nhockham:
We've figured out how to monkey patch - might be useful to someone else. Have this code before importing pandas.

```
from pandas.io.sql import SQLTable

def _execute_insert(self, conn, keys, data_iter):
    print "Using monkey-patched _execute_insert"
    data = [dict((k, v) for k, v in zip(keys, row)) for row in data_iter]
[...]

PR #21199, opened 2018-05-25
ENH: 'to_sql()' add param 'method' to control insert statement (#21103)

### LINKED ISSUE #21103: "too many SQL variables" Error with pandas 0.23 - enable multivalues insert #19664 issue

```python
#let's import packages to use
import numpy as np
import sqlite3 as sl
import pandas as pd
import datetime as dt
from sqlalchemy import create_engine
from sqlite3 import OperationalError



# #use pandas to import data
df = pd.DataFrame(np.arange(0,20000,1))
#create the engine to connect pandas with sqlite3
engine = create_engine('sqlite://')
#create connection
conn = engine.connect()
#convert df to sql table
df.to_sql('test',engine, if_exists='fail')
#print results
result = conn.execute("select * from test")
for row in result:
    print(row['index'])
conn.close()

#################
#my actual case#####
#################

#let's import packages to use
import numpy as np
import sqlite3 as sl
import pandas as pd
import datetime as dt
from sqlalchemy import create_engine
from sqlite3 import OperationalError

# #use pandas to import data
df = pd.read_excel(pth + 'CGM.xlsx', sheet_name=0)
#create the engine to connect pandas with sqlite3
engine = create_engine('sqlite:///H:\\Tests\\CO_db.db')
#create a connection to the database to run SQL Queries
conn = engine.connect()

#deletes empty space
conn.execute('VACUUM')

#create the CGM table unless it already exists
try:
    
    df.to_sql('CGM',engine, if_exists='fail', chunksize=100)
    
except ValueError as e:
    print(e)
    if str(e) == "Table 'CGM' already exists.":
        pass
    else:
        raise


```
#### Problem description

In pandas 0.22 I could write a dataframe to sql of reasonable size without error. Now I receive this error "OperationalError: (sqlite3.OperationalError) too many SQL variables". I am converting a dataframe with ~20k+ rows to sql. After looking around I suspect the problem lies in the limit set by sqlite3: SQLITE_MAX_VARIABLE_NUMBER which is set to 999 by default. This can apparently be changed by recompiling sqlite and adjusting this variable accordingly. I also noticed that adjusting the chunksize in DataFrame.to_sql has no effect perhaps confirming this is the root cause. I can confirm that for a df of length (rows) 499 this works. I can also confirm that this test version works with a row lenght of 20k and a chunksize of 499 inputted with df_to_sql works. My actual case is connecting to a local sqllite database and this is failing regardless of the chunksize adjusted to be 1 or anything else or without. #19664 

#### Expected  #Output
 
runfile('H:/Tests/Pandas_0.23_test.py', wdir='H:/Tests')
0
1
2
3
4
5
6
7
8
9
10
11
12
13
14
15
16
17
18
19

<details>

pd.show_versions()

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Windows
OS-release: 7
machine: AMD64
byteorder: little
LC_ALL: None
LANG: en
LOCALE: None.None

pandas: 0.23.0
pytest: 3.5.1
pip: 10.0.1
setuptools: 39.1.0
Cython: 0.28.2
numpy: 1.14.3
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: 1.7.4
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.2.1
tables: 3.4.3
numexpr: 2.6.5
feather: None
matplotlib: 2.2.2
openpyxl: 2.5.3
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.4
lxml: 4.2.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.7
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

### LINKED ISSUE #21146: to_sql() performance regression (#19664) when DF contains many columns
After #19664 `to_sql()` performance is highly dependent on the number of columns.

#### Problem description

For a single column 0.23 is faster than 0.22. Adding more columns has little influence on 0.22 but quickly degrades the performance on 0.23. I.e. for 20 columns 0.23 takes 1.5 times longer than 0.22

Performance degradation happens regardless of parameter `chunksize` adjustment.

```python
import time

import numpy as np
import pandas as pd
from sqlalchemy import create_engine


start = time.time()

N_COLS = 20
df = pd.DataFrame({n: np.arange(0,20_000,1) for n in range(N_COLS)})

#create the engine to connect pandas with sqlite3
engine = create_engine('postgresql://user:@localhost/db')
#create connection
conn = engine.connect()

# convert df to sql table
df.to_sql('test',engine, if_exists='replace',chunksize=1_000)
result = conn.execute("select * from test")
conn.close()
print('WRITE: {}'.format(time.time() - start))
```

Tested using linux, postgres 9.6, python  3.6, SQLAlchemy 1.2.2

schettino72:
Also revert default insert method to NOT use multi-value.

This is WIP. I would like to gather feedback on API change before going on.

- support callables as parameter?
- support to_sql() when used on SQLite3 (without SQLAlchemy)

Sample file for performance benchmarking
```
import time

import numpy as np
import pandas as pd
from sqlalchemy import create_engine


N_COLS = 20
N_ROWS= 200000
# one of to_sql insert methods: None/'default', 'multi', 'copy'
METHOD = 'copy'

engine = create_engine('postgresql://postgres:@localhost/pandas_perf')
conn = engine.connect()

start = time.time()
df = pd.DataFrame({n: np.arange(0, N_ROWS, 1) for n in range(N_COLS)})

# convert df to sql table
df.to_sql('test', conn, index=False, if_exists='replace',
          chunksize=1000, method=METHOD)

print('WRITE: {}'.format(time.time() - start))
```

- [x] closes #21103 & closes #21146 & #8953
- [x] tests added / passed
- [ ] whatsnew entry
- [ ] update docstrings/docs

### DIFF
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -29,6 +29,8 @@ Fixed Regressions
 - Bug in :meth:`~DataFrame.to_csv` causes encoding error when compression and encoding are specified (:issue:`21241`, :issue:`21118`)
 - Bug preventing pandas from being importable with -OO optimization (:issue:`21071`)
 - Bug in :meth:`Categorical.fillna` incorrectly raising a ``TypeError`` when `value` the individual categories are iterable and `value` is an iterable (:issue:`21097`, :issue:`19788`)
+- Regression in :func:`pivot_table` where an ordered ``Categorical`` with missing
+  values for the pivot's ``index`` would give a mis-aligned result (:issue:`21133`)
 
 
 .. _whatsnew_0231.performance:
--- a/pandas/core/reshape/pivot.py
+++ b/pandas/core/reshape/pivot.py
@@ -1,8 +1,10 @@
 # pylint: disable=E1103
 
 
-from pandas.core.dtypes.common import is_list_like, is_scalar
+from pandas.core.dtypes.common import (
+    is_list_like, is_scalar, is_integer_dtype)
 from pandas.core.dtypes.generic import ABCDataFrame, ABCSeries
+from pandas.core.dtypes.cast import maybe_downcast_to_dtype
 
 from pandas.core.reshape.concat import concat
 from pandas.core.series import Series
@@ -79,8 +81,22 @@ def pivot_table(data, values=None, index=None, columns=None, aggfunc='mean',
                 pass
         values = list(values)
 
-    grouped = data.groupby(keys, observed=dropna)
+    # group by the cartesian product of the grouper
+    # if we have a categorical
+    grouped = data.groupby(keys, observed=False)
     agged = grouped.agg(aggfunc)
+    if dropna and isinstance(agged, ABCDataFrame) and len(agged.columns):
+        agged = agged.dropna(how='all')
+
+        # gh-21133
+        # we want to down cast if
+        # the original values are ints
+        # as we grouped with a NaN value
+        # and then dropped, coercing to floats
+        for v in [v for v in values if v in data and v in agged]:
+            if (is_integer_dtype(data[v]) and
+                    not is_integer_dtype(agged[v])):
+                agged[v] = maybe_downcast_to_dtype(agged[v], data[v].dtype)
 
     table = agged
     if table.index.nlevels > 1:
--- a/pandas/tests/reshape/test_pivot.py
+++ b/pandas/tests/reshape/test_pivot.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 
 from datetime import datetime, date, timedelta
 
@@ -14,6 +15,11 @@ from pandas.core.reshape.pivot import pivot_table, crosstab
 from pandas.compat import range, product
 import pandas.util.testing as tm
 from pandas.api.types import CategoricalDtype as CDT
+
+
+@pytest.fixture(params=[True, False])
+def dropna(request):
+    return request.param
 
 
 class TestPivotTable(object):
@@ -109,7 +115,6 @@ class TestPivotTable(object):
             index=exp_index)
         tm.assert_frame_equal(result, expected)
 
-    @pytest.mark.parametrize('dropna', [True, False])
     def test_pivot_table_dropna_categoricals(self, dropna):
         # GH 15193
         categories = ['a', 'b', 'c', 'd']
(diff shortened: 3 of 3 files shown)


schettino72:
For 200,000 rows, 20 columns, chunksize 1000. Running python 3.6, postgres: 

Method | time(seconds)
-----------|-------------------
default  | 36.6 
multi | 56.4 
copy | 2.2

jorisvandenbossche:
Does ``COPY`` work for all sqlalchemy flavors?

In the past, we always have limited the pandas implementation to sqlachemy core constructs that are backend agnostic, within its limitations. 
But given the big difference, we maybe should reconsider that.

schettino72:
> Does COPY work for all sqlalchemy flavors?

No. I do not have enough knowledge of various DB systems. I just know it works with Postgresql and does NOT work with SQLite.

I think the main point is to allow the insertion method to be modified without requiring any monkey-patching. 

I can drop the COPY handling from the patch if you guys wish... I included it to make sure the API was good enough to really use other methods. As the default and multi-value implementation are very similar.

IMO it does not hurt to include some code for a specific backend as long as the user needs to explicitly choose to do so, but I understand you guys dont want the burden to support code for every different backend.

jorisvandenbossche:
So as I said on the issue (https://github.com/pandas-dev/pandas/issues/21103#issuecomment-393333398), I think a good start might be to have this option and allow users to pass a custom function, but not to actually implement such a 'copy' method ourselves (but it is a good test case). 
(if we include the copy method, we should have much more extensive testing for different platforms and datatypes etc)

jorisvandenbossche:
Be sure to keep it as a test case in the tests, and maybe also a good example for the documentation

jreback:
for 0.23.1 ?

jorisvandenbossche:
At least something, if not this, then a revert of the original PR.

jreback:
fair enough. there are 2 or 3 issues open in 0.23.1. pls adjust to what makes sense.

schettino72:
Sorry, I rebased this to master and force pushed... It seems github decided to automatically close this. Re-opened as #21401

Issue #21097, opened 2018-05-17
Categorical `fillna` with custom objects raises TypeError


jbrockmendel:
This behavior appears new in 0.23.0.

Setup copied from #21002:

```
from pandas.core.base import StringMixin

class County(StringMixin):
    name = u'San Sebastián'
    state = u'PR'
    def __unicode__(self):
        return self.name + u', ' + self.state

cat = pd.Categorical([County() for n in range(61)])
>>> cat.fillna(cat[0])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/pandas/util/_decorators.py", line 177, in wrapper
    return func(*args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/pandas/core/arrays/categorical.py", line 1769, in fillna
    '"{0}"'.format(type(value).__name__))
TypeError: "value" parameter must be a scalar, dict or Series, but you passed a "County"
```

I don't see any reason why passing one of the categories to `fillna` should be disallowed, so it looks like the issue is with being too strict about what qualifies as a scalar.

jbrockmendel:
@TomAugspurger I think this might be up your alley.

TomAugspurger:
Comes from https://github.com/pandas-dev/pandas/pull/19684

We check `is_scalar(value)`, and for `County` that's false.

`is_scalar` doesn't quite seem like the right check. We want to see if it's in the categories... Here's a simpler repro:

```python
In [25]: c = pd.Categorical([(0, 1), (1, 2)])

In [26]: c.fillna((0, 1))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-26-69f5393b61c6> in <module>()
----> 1 c.fillna((0, 1))

~/sandbox/pandas/pandas/util/_decorators.py in wrapper(*args, **kwargs)
    175                 else:
    176                     kwargs[new_arg_name] = new_arg_value
--> 177             return func(*args, **kwargs)
    178         return wrapper
    179     return _deprecate_kwarg

~/sandbox/pandas/pandas/core/arrays/categorical.py in fillna(self, value, method, limit)
   1767                 raise TypeError('"value" parameter must be a scalar, dict '
   1768                                 'or Series, but you passed a '
-> 1769                                 '"{0}"'.format(type(value).__name__))
   1770
   1771         return self._constructor(values, categories=self.categories,

[...]

Issue #19788, opened 2018-02-20
Categorical.fillna doesn't accept tuples


TomAugspurger:
This should maybe work.

```python
In [1]: import pandas as pd

In [2]: pd.Categorical([(1, 2), None]).fillna((1, 2))
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-73c5b4ab00c2> in <module>()
----> 1 pd.Categorical([(1, 2), None]).fillna((1, 2))

~/sandbox/pandas-ip/pandas/pandas/util/_decorators.py in wrapper(*args, **kwargs)
    136                 else:
    137                     kwargs[new_arg_name] = new_arg_value
--> 138             return func(*args, **kwargs)
    139         return wrapper
    140     return _deprecate_kwarg

~/sandbox/pandas-ip/pandas/pandas/core/arrays/categorical.py in fillna(self, value, method, limit)
   1664                 raise TypeError('"value" parameter must be a scalar, dict '
   1665                                 'or Series, but you passed a '
-> 1666                                 '"{0}"'.format(type(value).__name__))
   1667
   1668         return self._constructor(values, categories=self.categories,

TypeError: "value" parameter must be a scalar, dict or Series, but you passed a "tuple"
```

xref https://github.com/pandas-dev/pandas/pull/19684

This could be a can of worms if we want to accept `Series(categorical).fillna(tuple)`, since it's not allowed there. Would accept incremental improvements here of course.

PR #21215, opened 2018-05-26
BUG: Categorical.fillna with iterables

TomAugspurger:
Closes https://github.com/pandas-dev/pandas/issues/19788
Closes https://github.com/pandas-dev/pandas/issues/21097

Note that `Series.fillna` still doesn't allow iterables. I'm not sure that we should allow that, as it's potentially confusing what that would do.


jorisvandenbossche (approved):

jreback:
thanks @TomAugspurger

PR #21216, opened 2018-05-26
API/BUG: DatetimeIndex correctly localizes integer data 

### LINKED ISSUE #20964: BUG: Index with integer data and datetime64[ns, tz] dtype does not localize correctly
```
In [1]: pd.__version__
Out[1]: '0.23.0rc2+16.gccf4b96.dirty'

In [2]: val = [pd.Timestamp('2018-01-01', tz='US/Pacific').value]

In [3]: pd.Index(val, dtype='datetime64[ns, US/Pacific]')
Out[3]: DatetimeIndex(['2018-01-01 08:00:00-08:00'], dtype='datetime64[ns, US/Pacific]', freq=None)
```

The localization appears to localize directly to the timezone instead of localizing first to UTC and therefore does not roundtrip correctly from the timestamp value.

#20956 can be simplified once this is fixed.

Expected:

```
Out[3]: DatetimeIndex(['2018-01-01 00:00:00-08:00'], dtype='datetime64[ns, US/Pacific]', freq=None)
```

### LINKED ISSUE #20997: API: Int64Index cannot be converted into DatetimeIndex?
Is there a historical reason why an Int64Index (defaulting to represent epoch timestamps) cannot be converted to a DatetimeIndex?

```
In [7]: pd.__version__
Out[7]: '0.23.0rc2+27.geff1faf.dirty'

In [8]: val = [pd.Timestamp('2018-01-01').value]

In [9]: pd.DatetimeIndex(pd.Index(val))
<ipython-input-9-77aedfdf1582> in <module>()
----> 1 pd.DatetimeIndex(pd.Index(val))

/mnt/c/Users/Matt Roeschke/Projects/pandas-mroeschke/pandas/core/indexes/datetimes.py in __new__(cls, data, freq, start, end, periods, tz, normalize, closed, ambiguous, dayfirst, yearfirst, dtype, copy, name, verify_integrity)
    439             # must be integer dtype otherwise
    440             if isinstance(data, Int64Index):
--> 441                 raise TypeError('cannot convert Int64Index->DatetimeIndex')
    442             if data.dtype != _INT64_DTYPE:
    443                 data = data.astype(np.int64)

TypeError: cannot convert Int64Index->DatetimeIndex

# Expected behavior of In[9]
In [10]: pd.DatetimeIndex(pd.Index(val).values)
Out[10]: DatetimeIndex(['2018-01-01'], dtype='datetime64[ns]', freq=None)
```

mroeschke:
closes #20997 
closes #20964
- [x] tests added / passed
- [x] whatsnew entry

I did a little refactoring of `Datetime.__new__` that converts passed data but made it easier to fix this issue. I had to modify some methods (join, normalize, delete, insert) and tests that weren't assuming integer data weren't epoch timestamps.


pandas/tests/indexes/datetimes/test_construction.py
```diff
@@ -33,16 +33,15 @@ def test_construction_with_alt(self):
         tm.assert_index_equal(i, i2)
         assert i.tz.zone == 'US/Eastern'
 
-        i2 = DatetimeIndex(i.tz_localize(None).asi8, tz=i.dtype.tz)
+        i2 = DatetimeIndex(i.asi8, tz=i.dtype.tz)
         tm.assert_index_equal(i, i2)
         assert i.tz.zone == 'US/Eastern'
 
-        i2 = DatetimeIndex(i.tz_localize(None).asi8, dtype=i.dtype)
```
jreback: i actually like this construction, can you add a test that does both? (e.g. both uses and does not localize), maybe split it out and parameterize

pandas/tests/indexes/datetimes/test_construction.py
```diff
@@ -469,6 +468,14 @@ def test_constructor_with_non_normalized_pytz(self, tz):
         result = DatetimeIndex(['2010'], tz=non_norm_tz)
         assert pytz.timezone(tz) is result.tz
 
+    @pytest.mark.parametrize('klass', [Index, DatetimeIndex])
+    @pytest.mark.parametrize('box', [np.array, list])
+    def test_constructor_with_int_tz(self, klass, box):
+        ts = Timestamp('2018-01-01', tz='US/Pacific')
```
jreback: can you add the issue number

pandas/tests/indexes/test_base.py
```diff
@@ -402,26 +402,30 @@ def test_constructor_dtypes_to_timedelta(self, cast_index, vals):
...
-        # pass values without timezone, as DatetimeIndex localizes it
-        pd.date_range('2011-01-01', periods=5).values,
-        pd.date_range('2011-01-01', periods=5).asi8])
+    @pytest.mark.parametrize("attr, utc", [
+        ['values', False],
+        ['asi8', True]])
     @pytest.mark.parametrize("klass", [pd.Index, pd.DatetimeIndex])
-    def test_constructor_dtypes_datetime(self, tz_naive_fixture, values,
+    def test_constructor_dtypes_datetime(self, tz_naive_fixture, attr, utc,
                                          klass):
```
jreback: can you add a bit of commentary on what is tested here

jreback (changes requested):
ok looks pretty good. 

- can you add a whatsnew note (0.24.0)
- can you test with .astype as well
- can you search thru some issue for timezone / DTI. IIRC I think this might solve some (just a guess)

mroeschke:
Thanks @jreback. Moved the whatsnew to v0.24, added some additional logic to get `astype` working. It doesn't appear to fix any of the other reported timezone issues.

pandas/core/indexes/datetimes.py
```diff
@@ -806,8 +792,9 @@ def _mpl_repr(self):
 
     @cache_readonly
     def _is_dates_only(self):
+        """Return a boolean if we are only dates (and don't have a timezone)"""
         from pandas.io.formats.format import _is_dates_only
-        return _is_dates_only(self.values)
+        return _is_dates_only(self.values) and self.tz is None
```
jreback: why did this need changing?

pandas/core/indexes/datetimes.py
```diff
@@ -394,57 +394,43 @@ def __new__(cls, data=None,
...
+                data = conversion.tz_localize_to_utc(data.view('i8'), tz,
+                                                     ambiguous=ambiguous)
+            subarr = data.view(_NS_DTYPE)
         else:
             # must be integer dtype otherwise
-            if isinstance(data, Int64Index):
-                raise TypeError('cannot convert Int64Index->DatetimeIndex')
+            # assume this data are epoch timestamps
             if data.dtype != _INT64_DTYPE:
                 data = data.astype(np.int64)
```
jreback: could
``data = data.astype(np.int64, copy=False)``

pandas/tests/indexes/datetimes/test_astype.py
```diff
@@ -225,6 +225,13 @@ def _check_rng(rng):
         _check_rng(rng_eastern)
         _check_rng(rng_utc)
 
+    def test_integer_index_astype_datetimetz_dtype(self):
```
jreback: can you use tz_naive_fixture here

pandas/tests/indexes/datetimes/test_construction.py
```diff
@@ -26,25 +27,28 @@ def test_construction_caching(self):
                                                  freq='ns')})
         assert df.dttz.dtype.tz.zone == 'US/Eastern'
 
-    def test_construction_with_alt(self):
-
+    @pytest.mark.parametrize('kwargs', [
```
jreback: can you parameterize on tz_aware_fixture (doesn't include None), then parameterize as you are doing

pandas/tests/indexes/datetimes/test_construction.py
```diff
@@ -26,25 +27,28 @@ def test_construction_caching(self):
...
-        i2 = DatetimeIndex(
-            i.tz_localize(None).asi8, dtype=i.dtype, tz=i.dtype.tz)
-        tm.assert_index_equal(i, i2)
-        assert i.tz.zone == 'US/Eastern'
+        kwargs = {key: attrgetter(val)(i) for key, val in kwargs.items()}
+        result = DatetimeIndex(i, **kwargs)
+        tm.assert_index_equal(i, result)
+        assert result.tz.zone == 'US/Eastern'
+
+    @pytest.mark.parametrize('kwargs', [
```
jreback: same

pandas/tests/indexes/datetimes/test_construction.py
```diff
@@ -26,25 +27,28 @@ def test_construction_caching(self):
...
+        {'tz': 'dtype.tz'},
+        {'dtype': 'dtype'},
+        {'dtype': 'dtype', 'tz': 'dtype.tz'}])
```
jreback: can you explain how this test is different than the one above?

pandas/tests/indexes/datetimes/test_construction.py
```diff
@@ -469,6 +473,15 @@ def test_constructor_with_non_normalized_pytz(self, tz):
         result = DatetimeIndex(['2010'], tz=non_norm_tz)
         assert pytz.timezone(tz) is result.tz
 
+    @pytest.mark.parametrize('klass', [Index, DatetimeIndex])
+    @pytest.mark.parametrize('box', [np.array, list])
+    def test_constructor_with_int_tz(self, klass, box):
```
jreback: parameterize on tz_naive_fixture

pandas/core/indexes/datetimes.py
jreback: I don't think we have any tests with a non-int64 dtype, can you add something

jreback (changes requested):

jreback (approved):

jreback:
looks good. @jschendel @WillAyd if you can have a look if any comments.

pandas/tests/indexes/datetimes/test_construction.py
mroeschke: I am preserving the construction of `tz_localize(None)` that was here before but split it out in its own test as discussed here https://github.com/pandas-dev/pandas/pull/21216#discussion_r191286077

pandas/core/indexes/datetimes.py
mroeschke: After this refactor, I was getting errors where the repr didn't look correct but the i8 data was correct in the case when a tz was passed in the `DatetimeIndex` constructor.

`_is_dates_only` tries to:
> return a boolean if we are only dates (and don't have a timezone)

but checks for timezones by re-passing `self.values` into `DatetimeIndex`. I am not sure the historical reason why this was the case, but checking `self.tz is None` fixed the repr issue (and seems to achieve what that function was trying to do in the first place)

jreback:
thanks @mroeschke

pandas/core/indexes/base.py
```diff
@@ -1175,6 +1175,10 @@ def astype(self, dtype, copy=True):
             return CategoricalIndex(self.values, name=self.name, dtype=dtype,
                                     copy=copy)
         try:
+            if is_datetime64tz_dtype(dtype):
+                from pandas.core.indexes.datetimes import DatetimeIndex
```
jreback: side issue (another PR). The imports inside functions should be as simple as possible, e.g.
from pandas import DatetimeIndex (several occurrences of this)

PR #21222, opened 2018-05-27
BUG: df.agg(sum, axis=1) uses different method than when axis=0

### LINKED ISSUE #21134: BUG: df.agg(sum, axis=1) gives wrong result when Nan value is in frame

topper-123:
- [x] closes #21134
- [x] xref #21123
- [x] tests added / passed
- [x] whatsnew entry

This is a splitoff from #21123, to only fix #21134. #19629 will be fixed in a separate PR afterwards.

Passing builtins to ``df.agg`` is ok when ``axis=0``, but can give wrong result, when ``axis=1``  when NaNs are supplied.

Explanation
-------------
Passing the functions in ``SelectionMixin._cython_table`` to ``df.agg`` should defer to use the relevant cython functions. This currently works as expected when ``axis=0``, but not always when ``axis=1``.

The reason for this difference is that ``df.aggregate`` currently defers to ``df._aggregate`` when ``axis=0``, but defers to ``df.apply``, when ``axis=1``, and these give different result when passed funcions and the series/frame contains Nan values. I've solved this by transposing df in ``_aggragate`` when ``axis=1``.

The tests have been heavily parametrized, helping ensure that the various ways to call ``df.agg`` now give correct result.

### DIFF
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -94,3 +94,8 @@ Categorical
 ^^^^^^^^^^^
 
 -
+
+Numeric
+^^^^^^^
+
+- :meth:`~DataFrame.agg` now correctly handles built-in methods like ``sum`` when axis=1 (:issue:`21134`)
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -149,3 +149,20 @@ def tz_aware_fixture(request):
     Fixture for trying explicit timezones: {0}
     """
     return request.param
+
+
+@pytest.fixture(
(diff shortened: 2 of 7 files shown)


topper-123:
I’ve thought of a couple issues that should be tested. I mark this a WIP untill this is done.

PR #21224, opened 2018-05-27
BUG: df.agg, df.transform and df.apply use different methods when axis=1 than when axis=0

### LINKED ISSUE #16679: agg method with list of functions does not work with axis=1
#### Code Sample

```python
import numpy
import pandas

numpy.random.seed(1234)

frame = pandas.DataFrame(
    data=numpy.random.random((5, 3))
)
print(
    frame
)
# Works
print(
    frame.agg(['mean', 'std'])
)
# Expected output
print(
    frame.transpose().agg(['mean', 'std']).transpose()
)
# Does not work as expected
print(
    frame.agg(['mean', 'std'], axis=1)
)

```
#### Problem description

Calling the `agg` method on a frame with `axis=1` raises a `TypeError`:

```python

  File "/home/lafras/tmp/buggy.py", line 13, in <module>
    frame.agg(['mean', 'std'], axis=1)
  File "/home/lafras/.local/lib/python3.5/site-packages/pandas/core/frame.py", line 4154, in aggregate
    return self.apply(func, axis=axis, args=args, **kwargs)
  File "/home/lafras/.local/lib/python3.5/site-packages/pandas/core/frame.py", line 4262, in apply
    ignore_failures=ignore_failures)
  File "/home/lafras/.local/lib/python3.5/site-packages/pandas/core/frame.py", line 4358, in _apply_standard
    results[i] = func(v)
TypeError: ("'list' object is not callable", 'occurred at index 0')

```

#### Expected Output
`frame.agg(['mean', 'std'], axis=1)` should produce this:

```
       mean       std
0  0.417119  0.216033
1  0.612642  0.294504
2  0.678825  0.357107
3  0.578248  0.267557
4  0.588805  0.189838
```

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.2.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.0-79-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_ZA.UTF-8
LOCALE: en_ZA.UTF-8

pandas: 0.20.2
pytest: 3.0.4
pip: 9.0.1
setuptools: 36.0.1
Cython: 0.25.2
numpy: 1.13.0
scipy: 0.19.0
xarray: 0.8.2
IPython: 6.1.0
sphinx: None
patsy: 0.4.1
dateutil: 2.6.0
pytz: 2017.2
blosc: None
bottleneck: None
tables: 3.3.0
numexpr: 2.6.1
feather: None
matplotlib: 2.0.0
openpyxl: None
xlrd: 1.0.0
xlwt: None
xlsxwriter: None
lxml: None
bs4: 4.4.1
html5lib: 0.9999999
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.9.5
s3fs: 0.0.7
pandas_gbq: None
pandas_datareader: 0.2.1
</details>

### LINKED ISSUE #21134: BUG: df.agg(sum, axis=1) gives wrong result when Nan value is in frame

topper-123:
- [x] closes #16679
- [x] tests added / passed
- [x] whatsnew entry

This is a splitoff from #21123, to only fix #16679. #19629 will be fixed in a separate PR afterwards.

Passing functions to ``df.agg``, ``df.transform`` and ``df.apply`` may use different methods when ``axis=1``, than when,``axis=0``, and give different results when NaNs are supplied.

Explanation
-------------
Passing the functions in ``SelectionMixin._cython_table`` to ``df.agg`` should defer to use the relevant cython functions. This currently works as expected when ``axis=0``, but not when ``axis=1``.

The reason for this difference is that ``df.aggregate`` currently defers to ``df._aggregate`` when ``axis=0``, but defers to ``df.apply``, when ``axis=1``, and these may give different result when passed functions and the series/frame contains Nan values. I've solved this by transposing df in ``DataFrame._aggragate`` when ``axis=1``, and passing the possibly transposed on to the super method.

Also, ``df.apply`` delegates back to ``df.agg``, when given lists or dicts as inputs, but only works when axis=0. This PR fixes this, so axis=1 works the as axis=0.

The tests have been heavily parametrized, helping ensure that various ways to call the methods now give correct results for both axes.

@WillAyd @jreback (reviewers of #21123)

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -475,11 +475,9 @@ Numeric
 - Bug in :class:`Series` ``__rmatmul__`` doesn't support matrix vector multiplication (:issue:`21530`)
 - Bug in :func:`factorize` fails with read-only array (:issue:`12813`)
 - Fixed bug in :func:`unique` handled signed zeros inconsistently: for some inputs 0.0 and -0.0 were treated as equal and for some inputs as different. Now they are treated as equal for all inputs (:issue:`21866`)
-- Bug in :meth:`DataFrame.agg`, :meth:`DataFrame.transform` and :meth:`DataFrame.apply` when ``axis=1``.
-  Using ``apply`` with a list of functions and axis=1 (e.g. ``df.apply(['abs'], axis=1)``)
-  previously gave a TypeError. This fixes that issue.
-  As ``agg`` and ``transform`` in some cases delegate to ``apply``, this also
-  fixed this issue for them (:issue:`16679`).
+- Bug in :meth:`DataFrame.agg`, :meth:`DataFrame.transform` and :meth:`DataFrame.apply` where,
+  when supplied with a list of functions and ``axis=1`` (e.g. ``df.apply(['sum', 'mean'], axis=1)``),
+  a ``TypeError`` was wrongly raised. For all three methods such calculation are now done correctly. (:issue:`16679`).
 -
 
 Strings
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -475,7 +475,11 @@ Numeric
 - Bug in :class:`Series` ``__rmatmul__`` doesn't support matrix vector multiplication (:issue:`21530`)
 - Bug in :func:`factorize` fails with read-only array (:issue:`12813`)
 - Fixed bug in :func:`unique` handled signed zeros inconsistently: for some inputs 0.0 and -0.0 were treated as equal and for some inputs as different. Now they are treated as equal for all inputs (:issue:`21866`)
--
+- Bug in :meth:`DataFrame.agg`, :meth:`DataFrame.transform` and :meth:`DataFrame.apply` when ``axis=1``.
+  Using ``apply`` with a list of functions and axis=1 (e.g. ``df.apply(['abs'], axis=1)``)
+  previously gave a TypeError. This fixes that issue.
+  As ``agg`` and ``transform`` in some cases delegate to ``apply``, this also
+  fixed this issue for them (:issue:`16679`).
 -
 
 Strings
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -6,7 +6,7 @@ import pytest
 import pandas
 import numpy as np
 import pandas as pd
-from pandas.compat import PY3
+from pandas.compat import PY3, PY36
 import pandas.util._test_decorators as td
 
 
@@ -64,7 +64,7 @@ def spmatrix(request):
                 ids=lambda x: "axis {!r}".format(x))
 def axis(request):
     """
-     Fixture for returning the axis numbers of a dataframe.
+     Fixture for returning the axis numbers of a DataFrame.
      """
     return request.param
 
@@ -72,7 +72,7 @@ def axis(request):
 @pytest.fixture(params=[0, 'index'], ids=lambda x: "axis {!r}".format(x))
 def axis_series(request):
     """
-     Fixture for returning the axis numbers of a series.
+     Fixture for returning the axis numbers of a Series.
      """
     return request.param
 
@@ -117,6 +117,18 @@ def all_arithmetic_operators(request):
     """
     Fixture for dunder names for common arithmetic operations
     """
+    return request.param
+
+
+_cython_table = list(pd.core.base.SelectionMixin._cython_table.items())
+if not PY36:
+    # dicts have random order in Python<3.6, which xdist doesn't like
+    _cython_table = sorted(((key, value) for key, value in _cython_table),
+                           key=lambda x: x[0].__class__.__name__)
+
+
+@pytest.fixture(params=_cython_table)
+def cython_table_items(request):
     return request.param
 
 
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -60,8 +60,8 @@ def spmatrix(request):
     return getattr(sparse, request.param + '_matrix')
 
 
-@pytest.fixture(params=[0, 1],
-                ids=lambda x: "axis {}".format(x))
+@pytest.fixture(params=[0, 1, 'index', 'columns'],
+                ids=lambda x: "axis {!r}".format(x))
 def axis(request):
     """
      Fixture for returning the axis numbers of a dataframe.
@@ -69,7 +69,7 @@ def axis(request):
     return request.param
 
 
-@pytest.fixture(params=[0], ids=lambda x: "axis {}".format(x))
+@pytest.fixture(params=[0, 'index'], ids=lambda x: "axis {!r}".format(x))
 def axis_series(request):
     """
      Fixture for returning the axis numbers of a series.
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -60,6 +60,23 @@ def spmatrix(request):
     return getattr(sparse, request.param + '_matrix')
 
 
+@pytest.fixture(params=[0, 1],
+                ids=lambda x: "axis {}".format(x))
+def axis(request):
+    """
+     Fixture for returning the axis numbers of a dataframe.
+     """
+    return request.param
+
+
+@pytest.fixture(params=[0], ids=lambda x: "axis {}".format(x))
+def axis_series(request):
+    """
+     Fixture for returning the axis numbers of a series.
+     """
+    return request.param
+
+
 @pytest.fixture
 def ip():
     """
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -6,7 +6,7 @@ import pytest
 import pandas
 import numpy as np
 import pandas as pd
-from pandas.compat import PY3, PY36
+from pandas.compat import PY3
 import pandas.util._test_decorators as td
 
 
@@ -69,6 +69,9 @@ def axis(request):
     return request.param
 
 
+axis_frame = axis
+
+
 @pytest.fixture(params=[0, 'index'], ids=lambda x: "axis {!r}".format(x))
 def axis_series(request):
     """
@@ -120,16 +123,39 @@ def all_arithmetic_operators(request):
     return request.param
 
 
-_cython_table = list(pd.core.base.SelectionMixin._cython_table.items())
-if not PY36:
-    # dicts have random order in Python<3.6, which xdist doesn't like
-    _cython_table = sorted(((key, value) for key, value in _cython_table),
-                           key=lambda x: x[0].__class__.__name__)
+# use sorted as dicts in py<3.6 have random order, which xdist doesn't like
+_cython_table = sorted(((key, value) for key, value in
+                        pd.core.base.SelectionMixin._cython_table.items()),
+                       key=lambda x: x[0].__class__.__name__)
 
 
 @pytest.fixture(params=_cython_table)
 def cython_table_items(request):
     return request.param
+
+
+def _get_cython_table_params(ndframe, func_names_and_expected):
+    """combine frame, functions from SelectionMixin._cython_table
+    keys and expected result.
+
+    Parameters
+    ----------
+    ndframe : DataFrame or Series
+    func_names_and_expected : Sequence of two items
+        The first item is a name of a NDFrame method ('sum', 'prod') etc.
+        The second item is the expected return value
+
+    Returns
+    -------
+    results : list
+        List of three items (DataFrame, function, expected result)
+    """
+    results = []
+    for func_name, expected in func_names_and_expected:
+        results.append((ndframe, func_name, expected))
+        results += [(ndframe, func, expected) for func, name in _cython_table
+                    if name == func_name]
+    return results
 
 
 @pytest.fixture(params=['__eq__', '__ne__', '__le__',
--- a/pandas/core/apply.py
+++ b/pandas/core/apply.py
@@ -5,6 +5,8 @@ from pandas._libs import reduction
 from pandas.core.dtypes.generic import ABCSeries
 from pandas.core.dtypes.common import (
     is_extension_type,
+    is_dict_like,
+    is_list_like,
     is_sequence)
 from pandas.util._decorators import cache_readonly
 
@@ -106,7 +108,7 @@ class FrameApply(object):
         """ compute the results """
 
         # dispatch to agg
-        if isinstance(self.f, (list, dict)):
+        if is_list_like(self.f) or is_dict_like(self.f):
             return self.obj.aggregate(self.f, axis=self.axis,
                                       *self.args, **self.kwds)
 
--- a/pandas/core/apply.py
+++ b/pandas/core/apply.py
@@ -105,6 +105,11 @@ class FrameApply(object):
     def get_result(self):
         """ compute the results """
 
+        # dispatch to agg
+        if isinstance(self.f, (list, dict)):
+            return self.obj.aggregate(self.f, axis=self.axis,
+                                      *self.args, **self.kwds)
+
         # all empty
         if len(self.columns) == 0 and len(self.index) == 0:
             return self.apply_empty_result()
@@ -307,15 +312,6 @@ class FrameApply(object):
 
 class FrameRowApply(FrameApply):
     axis = 0
-
-    def get_result(self):
-
-        # dispatch to agg
-        if isinstance(self.f, (list, dict)):
-            return self.obj.aggregate(self.f, axis=self.axis,
-                                      *self.args, **self.kwds)
-
-        return super(FrameRowApply, self).get_result()
 
     def apply_broadcast(self):
         return super(FrameRowApply, self).apply_broadcast(self.obj)
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -6080,7 +6080,7 @@ class DataFrame(NDFrame):
         return result
 
     def _aggregate(self, arg, axis=0, *args, **kwargs):
-        if axis == 1:
+        if axis in {1, 'columns'}:
             result, how = (super(DataFrame, self.T)
                            ._aggregate(arg, *args, **kwargs))
             result = result.T if result is not None else result
@@ -6090,7 +6090,7 @@ class DataFrame(NDFrame):
     agg = aggregate
 
     def transform(self, func, axis=0, *args, **kwargs):
-        if axis == 1:
+        if axis in {1, 'columns'}:
             return super(DataFrame, self.T).transform(func, *args, **kwargs).T
         return super(DataFrame, self).transform(func, *args, **kwargs)
 
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -6080,7 +6080,8 @@ class DataFrame(NDFrame):
         return result
 
     def _aggregate(self, arg, axis=0, *args, **kwargs):
-        if axis in {1, 'columns'}:
+        axis = self._get_axis_number(axis)
+        if axis == 1:
             result, how = (super(DataFrame, self.T)
                            ._aggregate(arg, *args, **kwargs))
             result = result.T if result is not None else result
@@ -6090,7 +6091,8 @@ class DataFrame(NDFrame):
     agg = aggregate
 
     def transform(self, func, axis=0, *args, **kwargs):
-        if axis in {1, 'columns'}:
+        axis = self._get_axis_number(axis)
+        if axis == 1:
             return super(DataFrame, self.T).transform(func, *args, **kwargs).T
         return super(DataFrame, self).transform(func, *args, **kwargs)
 
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -6070,16 +6070,19 @@ class DataFrame(NDFrame):
     def aggregate(self, func, axis=0, *args, **kwargs):
         axis = self._get_axis_number(axis)
 
-        # TODO: flipped axis
         result = None
-        if axis == 0:
-            try:
-                result, how = self._aggregate(func, axis=0, *args, **kwargs)
-            except TypeError:
-                pass
+        try:
+            result, how = self._aggregate(func, axis=axis, *args, **kwargs)
+        except TypeError:
+            pass
         if result is None:
             return self.apply(func, axis=axis, args=args, **kwargs)
         return result
+
+    @Appender(NDFrame._aggregate.__doc__, indents=2)
+    def _aggregate(self, arg, axis=0, *args, **kwargs):
+        obj = self.T if axis == 1 else self
+        return super(DataFrame, obj)._aggregate(arg, *args, **kwargs)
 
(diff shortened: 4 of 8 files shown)


pandas/core/frame.py
```diff
@@ -5818,17 +5818,20 @@ def _gotitem(self,
...
-                pass
+        try:
+            result, how = self._aggregate(func, axis=axis, *args, **kwargs)
+        except TypeError:
+            pass
         if result is None:
             return self.apply(func, axis=axis, args=args, **kwargs)
         return result
 
+    @Appender(NDFrame._aggregate.__doc__, indents=2)
```
jreback: I don't think you need this doc string here

pandas/tests/frame/test_apply.py
```diff
@@ -1056,3 +1056,73 @@ def test_non_callable_aggregates(self):
...
+    @pytest.mark.parametrize("axis", [0, 1], ids=lambda x: "axis {}".format(x))
+    def test_agg_cython_table(self, cython_table_items,
+                              frame, expected_dict, axis):
+        # GH21224
+        # test if using items in pandas.core.base.SelectionMixin._cython_table
+        # in agg gives correct results
+        np_func, str_func = cython_table_items
+        expected = expected_dict[str_func]
+
+        if isinstance(expected, type) and issubclass(expected, Exception):
```
jreback: I really do not like mixing exceptions with good cases in a single test. can you split.

pandas/tests/frame/test_apply.py
```diff
@@ -1056,3 +1056,73 @@ def test_non_callable_aggregates(self):
...
+            with pytest.raises(expected):
+                # e.g. DataFrame(['a b'.split()]).cumprod() will raise
+                frame.agg(np_func, axis=axis)
+            with pytest.raises(expected):
+                frame.agg(str_func, axis=axis)
+            return
+
+        result = frame.agg(np_func, axis=axis)
+        result_str_func = frame.agg(str_func, axis=axis)
```
jreback: rather than doing an if like this, can you create another fixture (or 2), that splits the functions into 2 groups: aggregators and transformers

pandas/tests/series/test_apply.py
```diff
@@ -331,6 +331,78 @@ def test_non_callable_aggregates(self):
...
+                              series, expected_dict):
+        # GH21224
+        # test if using items in pandas.core.base.SelectionMixin._cython_table
+        # in agg gives correct results
+        np_func, str_func = cython_table_items
+        expected = expected_dict[str_func]
+
+        if isinstance(expected, type) and issubclass(expected, Exception):
+            with pytest.raises(expected):
+                # e.g. Series('a b'.split()).cumprod() will raise
```
jreback: same comments as above

jreback (changes requested):

pandas/tests/frame/test_apply.py
topper-123: I agree in principle, but think will make the code less readable, as I explained in https://github.com/pandas-dev/pandas/pull/21123#discussion_r191036159.

I've made an version where the different types of expected values are split into separate test functions, but IMO it's not an improvement in legibility...Can this not be kept as an exception?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -119,7 +119,7 @@ Offsets
 Numeric
 ^^^^^^^
 
--
+- :meth:`~DataFrame.agg` now handles built-in methods like ``sum`` in the same manner when axis=1 as when axis=0 (:issue:`21224`)
```
jreback: just say.....``sum`` with ``axis=1``. (no need for the rest of the sentence)

pandas/conftest.py
```diff
@@ -170,3 +170,11 @@ def string_dtype(request):
     * 'U'
     """
     return request.param
+
+
+@pytest.fixture(params=[0, 1], ids=lambda x: "axis {}".format(x))
```
jreback: actually could also add ``'index', 'columns'`` here as well, *may* need to call this ``axis_frame``

jreback: may as well add ``axis_series``

pandas/core/indexing.py
```diff
@@ -1795,7 +1795,7 @@ def error():
                     error()
                 raise
             except:
-                error()
```
jreback: why did this need changin?

pandas/tests/frame/test_apply.py
```diff
@@ -21,6 +22,38 @@
...
+        The first item is a name of a NDFrame method ('sum', 'prod') etc.
+        The second item is the expected return value
+
+    Returns
+    -------
+    results : list
+        List of three items (DataFrame, function, expected result)
+    """
+    table = pd.core.base.SelectionMixin._cython_table
+    if compat.PY36:
```
jreback: can you make this first part a fixture in conftest

jreback: then not sure you really need this function, can you not compute this directly give the name of the function and table?

jreback (changes requested):

pandas/core/indexing.py
topper-123: The old way hides the cause of the exception.

For example if the exception is in ``ax.contains`` you won't see where the actual error was in the traceback, as Python thinks the exeption is **in** ``error``, while in reality is is somet´where else, hence the need to reraise.

Thought on closer inspection, just removing the two lines ``except: error()`` would be even clearer than reraising.

pandas/tests/frame/test_apply.py
topper-123: > can you make this first part a fixture in conftest

Not possible; this function is not a test function, so does not take fixtures

topper-123: > then not sure you really need this function, can you not compute this directly give the name of the function and table?

I'd want the the functions to take the lefthand side functions in ``_cython_table`` as input, so each one is tested individually. 

An alternative would be to pass the strings into the test functions (e.g. "sum"), and then get the relevant functions inside the test functions. That would mean that each test would run several subtests, which is a different kind of issue/problem.

So I think you have to choose between different kinds of ugliness here. I'd personally prefer my original solution. I'f someone has an idea for improvement, I'm wiling to give it another run, though

pandas/conftest.py
topper-123: I'm thinking that this fixture could replace all instances where there's a ``parametrize``. There are quite a few in the tests. Many of those don't support "index" and "columns" and it would give some failures (29 failures when I ran it with "index" and "columns").

I could add a ``axis_all`` fixture where params would be [0, 1, "index", "columns"]? Then people could differentiate between which one they need.

jreback:
can you rebase

topper-123:
Rebased and green.

topper-123:
Hold a bit pulling this in. I think I've found an issue that needs fixing.

topper-123:
I discovered a bug in my ``df.aggregate`` implentation. Solving that unearthed several other bugs, as:
* df.transform relies on df.aggregate
* Fixing the bug for df.transform required tests for df.transform(func, axis=1), which unearthed a bug in df.apply(func, axis=1), which also needed fixing

Very much fun!

The most basic bug is that ``df.apply([func}, axis=1)`` doesn't work in master. This fixes that, and subsequent bugs in df.transform and df.agg.

```python
>>> from pandas.tests.frame.test_apply import TestDataFrameAggregate
>>> x = TestDataFrameAggregate()
>>> x.frame.apply([np.sqrt], axis=0)  # ok
                   A         B         C         D
                sqrt      sqrt      sqrt      sqrt
J5bdGdv0g8       NaN  0.790097       NaN  1.261288
bLhjMokuwH  0.704260       NaN       NaN  0.582152
nt9hGglqqX  1.192715  0.858367  0.809943  0.658500
qzqonGSWI2       NaN       NaN  0.213483  0.329113
HKzrPBstjP       NaN       NaN  1.009700  1.478508
h9KtufwkNe       NaN       NaN       NaN       NaN
zqPXLsFNhc  1.035181       NaN  1.034206  0.932663
LVWWuXcDeN  0.391793       NaN  0.946918  0.403030
pNPW6YbGlX  0.573936  0.709931       NaN       NaN
1tDta2tJkA       NaN  0.968199       NaN       NaN
OtOh41yTwD  1.170893       NaN       NaN       NaN
NBcn0qeBUA       NaN  0.390018       NaN  0.979592
iV03qoDpDz       NaN       NaN       NaN       NaN
OsRUIhjUji  0.614736       NaN  0.722526       NaN
uUzaTh9V8S  0.895671  0.687541       NaN       NaN
EwkkR0nUCN  1.248047       NaN       NaN       NaN
RIDiWUqfhY       NaN       NaN  0.221734  0.681125
EIA9BP3RDu       NaN       NaN       NaN       NaN
ZS94HtjpTW  0.930240       NaN  0.423507       NaN
Bq4BMF3C3c  1.192806       NaN  1.530339  0.319647
YQGtbDKliH       NaN       NaN       NaN  0.943571
idohmGIF1n  0.230102  1.281673       NaN  1.273232
htMOE93MLi  0.572996       NaN  0.703781  0.650777
BMgE6NPlk6  0.648459  0.087043       NaN  0.958977
aZnHOmLKIc  0.834899       NaN  0.630469  1.079947
ArHDDm2ZTJ  1.583401       NaN       NaN       NaN
IQCqXHHzK1  0.180034       NaN  1.084064       NaN
U9s5JKOGnm  0.829343       NaN  1.392410       NaN
ss8c5sdDKb  0.765316       NaN  0.701255       NaN
QcBJ1vuAja  0.311198       NaN       NaN       NaN
>>> x.frame.apply([np.sqrt], axis=1) 
TypeError: ("'list' object is not callable", 'occurred at index J5bdGdv0g8')  # master
# result from this PR is below
                        A         B         C         D
yohEBbNfVO sqrt       NaN       NaN       NaN  0.932832
nAmb8pifpF sqrt  0.672436  1.337469  0.754772  1.198730
v134xZDU2L sqrt       NaN       NaN  1.673506  0.783993
LosL3nZo29 sqrt       NaN  0.682876       NaN       NaN
TNPVQRqTJb sqrt       NaN  0.780793  0.230658  0.482869
7DG2GzTEuF sqrt       NaN       NaN       NaN       NaN
rSngToOl4I sqrt       NaN       NaN       NaN       NaN
8KDnTlyiro sqrt  1.169174  0.455202       NaN  1.483356
bJBgtadC2a sqrt       NaN       NaN       NaN       NaN
ayAAwzg9iv sqrt       NaN       NaN  1.133000  0.662161
uyY52dHxcj sqrt       NaN  0.624445       NaN       NaN
xkGYEt8DY1 sqrt       NaN       NaN  0.765723       NaN
MMrlaZJwaN sqrt  0.458500       NaN       NaN       NaN
3CTF3dT63C sqrt  1.232379       NaN       NaN  0.816069
zTepySfz0l sqrt       NaN  0.867734       NaN  0.874975
CmW5sWb3pn sqrt       NaN       NaN       NaN  1.057664
LfPNb7mKPq sqrt  0.596743       NaN  1.158681       NaN
zcPe3o0s35 sqrt       NaN  1.265513  1.364592  0.545380
CFW0u4jjJV sqrt       NaN       NaN  0.762759       NaN
XMfNwT2R6e sqrt       NaN  0.857021       NaN       NaN
lw2ZlN5fp5 sqrt  1.471999       NaN  0.718156  0.848762
9fj0fYmPRR sqrt       NaN       NaN  0.706411  1.029278
ecqwt01Eqz sqrt  1.126164       NaN       NaN  0.393378
zMMKCgOswt sqrt  0.467303       NaN  0.535867  1.188402
DKCTE0k3SE sqrt       NaN       NaN  0.931693  0.741451
h0EtVtS73i sqrt  0.567195  0.299606  0.645672       NaN
8a7amQnx7P sqrt       NaN       NaN  0.617702  0.823773
tP6HCPlwlN sqrt       NaN  0.989050  0.792688  0.572421
5GNZ27tezt sqrt  0.575549  0.741495  1.175171       NaN
n3bIDyezrL sqrt  0.785265       NaN  1.599316  0.583166
```

Similar problem where in master when calling ``df.transform([func], axis=1)`` and ``df.agg([func], axis=1)``.

pandas/core/apply.py
```diff
@@ -308,15 +313,6 @@ def wrap_results(self):
 class FrameRowApply(FrameApply):
     axis = 0
 
-    def get_result(self):
-
-        # dispatch to agg
-        if isinstance(self.f, (list, dict)):
-            return self.obj.aggregate(self.f, axis=self.axis,
-                                      *self.args, **self.kwds)
```
topper-123: ``if isinstance(self.f, (list, dict))`` should also be called when axis=1, so moved up to ``FrameApply.get_results``.

pandas/core/generic.py
```diff
@@ -9090,16 +9090,14 @@ def ewm(self, com=None, span=None, halflife=None, alpha=None,
...
+    @Appender(_shared_docs['transform'] % _shared_doc_kwargs)
+    def transform(self, func, *args, **kwargs):
+        result = self.agg(func, *args, **kwargs)
+        if is_scalar(result) or len(result) != len(self):
+            raise ValueError("transforms cannot produce "
+                             "aggregated results")
 
-            return result
-
-        cls.transform = transform
```
topper-123: when fransform was added by calling ``_add_series_or_dataframe_operations`` that method shadowed the method on ``DataFrame``. As transform is doesnt need to be added in any special way, I just move to be a normal instance method.

pandas/tests/frame/test_apply.py
```diff
@@ -813,81 +846,97 @@ def test_consistency_for_boxed(self, box):
...
 
-    def test_agg_transform(self):
+    def test_agg_transform(self, axis):
+        other_axis = abs(axis -1)
 
         with np.errstate(all='ignore'):
 
-            f_sqrt = np.sqrt(self.frame)
             f_abs = np.abs(self.frame)
+            f_sqrt = np.sqrt(self.frame)
```
topper-123: having "absolute" come before "sqrt" maintains alphabetical ordering, and makes creating multindexes easier below.

topper-123:
@jreback ?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -405,7 +405,11 @@ Numeric
 
 - Bug in :class:`Series` ``__rmatmul__`` doesn't support matrix vector multiplication (:issue:`21530`)
 - Bug in :func:`factorize` fails with read-only array (:issue:`12813`)
--
```
jreback: use double-back ticks around TypeError. you don't need the 2nd sentence, just list the issue.

pandas/conftest.py
```diff
@@ -60,6 +60,23 @@ def spmatrix(request):
     return getattr(sparse, request.param + '_matrix')
 
 
+@pytest.fixture(params=[0, 1],
```
jreback: add 'index', 'columns'

pandas/conftest.py
```diff
@@ -60,6 +60,23 @@ def spmatrix(request):
...
+                ids=lambda x: "axis {}".format(x))
+def axis(request):
+    """
+     Fixture for returning the axis numbers of a dataframe.
+     """
+    return request.param
+
+
+@pytest.fixture(params=[0], ids=lambda x: "axis {}".format(x))
```
jreback: add 'index'

pandas/core/apply.py
```diff
@@ -105,6 +105,11 @@ def agg_axis(self):
     def get_result(self):
         """ compute the results """
 
+        # dispatch to agg
+        if isinstance(self.f, (list, dict)):
```
jreback: can you use is_list_like and is_dict_like

pandas/tests/frame/test_apply.py
jreback: this can easily BE a fixture, IOW you put this function in conftest.py and call it to create the params FOR a fixture.

pandas/tests/series/test_apply.py
```diff
@@ -17,6 +18,38 @@
...
+    func_names_and_expected : Sequence of two items
+        The first item is a name of a NDFrame method ('sum', 'prod') etc.
+        The second item is the expected return value
+
+    Returns
+    -------
+    results : list
+        List of three items (Series, function, expected result)
+    """
+    table = pd.core.base.SelectionMixin._cython_table
```
jreback: see my above comment

jreback (changes requested):

pandas/tests/frame/test_apply.py
topper-123: I've almost dried my brain looking into how to do this, I just can't see it... The function ``_get_cython_table_params`` can't take fixtures, as it's not a test function. Do you mean simply import it from conttest.py? (``from pandas.conftest import cython_table`` or similar) I assume that's not what you mean, tough.

Could you maybe spell it out how you see the implementation for me? I would be very grateful for that.

topper-123:
Comments addressed, except the cython_table one, where I ask for clarification.

pandas/tests/frame/test_apply.py
topper-123: I've added a fixture for cython table items. Can't use fixtures in ``_get_cython_table_params``, so do an import from conttest.py.

The build in CirceCi is a resourceerror, so unrelated to this PR.

pandas/conftest.py
```diff
@@ -60,6 +60,23 @@ def spmatrix(request):
     return getattr(sparse, request.param + '_matrix')
 
 
+@pytest.fixture(params=[0, 1, 'index', 'columns'],
+                ids=lambda x: "axis {!r}".format(x))
+def axis(request):
```
jreback: maybe also add axis_frame just for consistency? (its exactly equal to axis)

pandas/conftest.py
```diff
@@ -103,6 +120,18 @@ def all_arithmetic_operators(request):
     return request.param
 
 
+_cython_table = list(pd.core.base.SelectionMixin._cython_table.items())
```
jreback: why not always sort them?

pandas/core/frame.py
```diff
@@ -6068,19 +6068,32 @@ def _gotitem(self,
...
+        if axis == 1:
+            result, how = (super(DataFrame, self.T)
+                           ._aggregate(arg, *args, **kwargs))
+            result = result.T if result is not None else result
+            return result, how
+        return super(DataFrame, self)._aggregate(arg, *args, **kwargs)
+
     agg = aggregate
 
+    def transform(self, func, axis=0, *args, **kwargs):
```
jreback: does this doc-string get updated somewhere?

pandas/core/frame.py
```diff
@@ -6068,19 +6068,32 @@ def _gotitem(self,
...
+        try:
+            result, how = self._aggregate(func, axis=axis, *args, **kwargs)
+        except TypeError:
+            pass
         if result is None:
             return self.apply(func, axis=axis, args=args, **kwargs)
         return result
 
+    def _aggregate(self, arg, axis=0, *args, **kwargs):
+        axis = self._get_axis_number(axis)
```
jreback: can you add a comment here on what you are doing

pandas/core/generic.py
```diff
@@ -9156,16 +9156,14 @@ def ewm(self, com=None, span=None, halflife=None, alpha=None,
 
         cls.ewm = ewm
 
-        @Appender(_shared_docs['transform'] % _shared_doc_kwargs)
-        def transform(self, func, *args, **kwargs):
-            result = self.agg(func, *args, **kwargs)
-            if is_scalar(result) or len(result) != len(self):
-                raise ValueError("transforms cannot produce "
-                                 "aggregated results")
+    @Appender(_shared_docs['transform'] % _shared_doc_kwargs)
```
jreback: I think you need to add this appender in frame.py as well

pandas/tests/series/test_apply.py
```diff
@@ -17,6 +18,33 @@
 from .common import TestData
 
 
+def _get_cython_table_params(series, func_names_and_expected):
```
jreback: I guess I would like this to be a centralized function, can you move to pandas.util.testing (as a private function like you have now), and consolidate with the one from frame).

jreback (changes requested):

pandas/conftest.py
topper-123: for Python >= 3.6 dicts are already sorted and I think maintaining dict order makes the most sense.

The sorting is only because xdist fails for Python 3.5, so predefined order is necessary for Python < 3.6 to satisfy xdist.

jreback: i understand, but this is for testing where we want determinism. Having alphabeting ordering is good.

topper-123: It is deterministic for python 3.6 also, as per the ordered dicts. My thinking is that the if not if ``if`` part can more easily be removed when support for python 3.5 is dropped and ``PY36`` is removed from compat...

However, this is obviously not a huge issue for me, I'll can change it.

pandas/core/generic.py
topper-123: Yes, I'll do that

pandas/core/frame.py
topper-123: This probably isn't necessary here, as _aggregate is an internal function and ``axis = self._get_axis_number(axis)`` is already called in ``aggregate´´. I'll just delete that line an dexplain the rest.

pandas/core/frame.py
topper-123: I'll add the Appender

pandas/tests/series/test_apply.py
topper-123: ok, done.

pandas/conftest.py
topper-123: sure, ok

topper-123:
comments addressed.

pandas/util/testing.py
```diff
@@ -2826,3 +2826,28 @@ def skipna_wrapper(x):
             return alternative(nona)
 
     return skipna_wrapper
+
+
+def _get_cython_table_params(ndframe, func_names_and_expected):
```
jreback: hmm didn't realize you are actually importing from conftest, so on 2nd though, move this to ``pandas.conftest`` and import from there (you have to explicity import non-fixtures FYI)

jreback (changes requested):
small change, otherwise lgtm. ping on green.

topper-123:
green

jreback (approved):

PR #21235, opened 2018-05-28
pct change bug issue 21200

### LINKED ISSUE #21200: BUG: groupby.pct_change() does not work properly in Pandas 0.23.0. Grouping is ignored.
#### Code Sample

```python
>>>import pandas as pd
>>>import numpy as np

>>>df = pd.DataFrame(data=np.random.rand(8, 1), columns={'a'})
>>>df['grp']=1
>>>df.loc[::2, 'grp']=2
>>>df['%_groupby']=df.groupby('grp')['a'].pct_change()
>>>df['%_shift']=df.groupby('grp')['a'].shift(0)/df.groupby('grp')['a'].shift(1)-1
>>>print(df)

```
#### Problem description

When there are different groups in a dataframe, by using `groupby` it is expected that the `pct_change` function be applied on each group. However, combining `groupby` with `pct_change` does not produce the correct result.

Output:
```
     a  grp  %_groupby   %_shift
0  1.0    2        NaN       NaN
1  1.1    1   0.100000       NaN
2  1.2    2   0.090909  0.200000
3  1.3    1   0.083333  0.181818
4  1.4    2   0.076923  0.166667
5  1.5    1   0.071429  0.153846
6  1.6    2   0.066667  0.142857
7  1.7    1   0.062500  0.133333
```
#### Expected Output
```
     a  grp  %_groupby   %_shift
0  1.0    2        NaN       NaN
1  1.1    1        NaN       NaN
2  1.2    2   0.200000  0.200000
3  1.3    1   0.181818  0.181818
4  1.4    2   0.166667  0.166667
5  1.5    1   0.153846  0.153846
6  1.6    2   0.142857  0.142857
7  1.7    1   0.133333  0.133333
```

<details>

**INSTALLED VERSIONS**

------------------
commit: None
python: 3.6.3.final.0
python-bits: 64
OS: Darwin
OS-release: 17.5.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.23.0
pytest: 3.2.1
pip: 10.0.1
setuptools: 36.5.0.post20170921
Cython: 0.26.1
numpy: 1.14.3
scipy: 0.19.1
pyarrow: None
xarray: None
IPython: 6.1.0
sphinx: 1.6.3
patsy: 0.4.1
dateutil: 2.6.1
pytz: 2018.3
blosc: None
bottleneck: 1.2.1
tables: 3.4.2
numexpr: 2.6.2
feather: None
matplotlib: 2.1.0
openpyxl: 2.4.8
xlrd: 1.1.0
xlwt: 1.2.0
xlsxwriter: 1.0.2
lxml: 4.1.1
bs4: 4.6.0
html5lib: 0.9999999
sqlalchemy: 1.1.13
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

### LINKED ISSUE #21621: Groupby Pct_change Bug Report
#### Groupby Pct_change Bug Report
```python
from io import StringIO
import pandas as pd

txt = StringIO("""Group;Date;Value
  A     ;01-02-2016     ;16 
  A     ;01-03-2016     ;15 
  A     ;01-04-2016     ;14 
  A     ;01-05-2016     ;17 
  A     ;01-06-2016     ;19 
  A     ;01-07-2016     ;20 
  B     ;01-02-2016     ;16 
  B     ;01-03-2016     ;13 
  B     ;01-04-2016     ;13 
  C     ;01-02-2016     ;16 
  C     ;01-03-2016     ;16 """)

df = pd.read_table(txt,sep=';',parse_dates=[1])
d1 = df.set_index(['Date', 'Group']).Value
d2 = d1.groupby(level='Group').pct_change()
print(d2)
```
#### Problem description
I tried to calculate value changes by group in pandas , according this stackoverflow answer, https://stackoverflow.com/questions/41453325/python-pandas-groupby-calculate-change 

the expected result should be 
#### Expected Output
```python
Date        Group
2016-01-02  A             NaN
2016-01-03  A       -0.062500
2016-01-04  A       -0.066667
2016-01-05  A        0.214286
2016-01-06  A        0.117647
2016-01-07  A        0.052632
2016-01-02  B             NaN
2016-01-03  B       -0.187500
2016-01-04  B        0.000000
2016-01-02  C             NaN
2016-01-03  C        0.000000
Name: Value, dtype: float64
```
First pecent change value of each group should be NaN, while in my computer with pandas version 0.23.1, the result shows as below.
 ```python
Date        Group   
2016-01-02    A              NaN
2016-01-03    A        -0.062500
2016-01-04    A        -0.066667
2016-01-05    A         0.214286
2016-01-06    A         0.117647
2016-01-07    A         0.052632
2016-01-02    B        -0.200000
2016-01-03    B        -0.187500
2016-01-04    B         0.000000
2016-01-02    C         0.230769
2016-01-03    C         0.000000
Name: Value, dtype: float64
```

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
processor: AMD64 Family 23 Model 1 Stepping 1, AuthenticAMD
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.23.1
pytest: 3.5.1
pip: 10.0.1
setuptools: 39.1.0
Cython: 0.28.2
numpy: 1.14.3
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: 1.7.4
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.2.1
tables: 3.4.3
numexpr: 2.6.5
feather: None
matplotlib: 2.2.2
openpyxl: 2.5.3
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.4
lxml: 4.2.1
bs4: 4.6.0
html5lib: 0.9999999
sqlalchemy: 1.2.7
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

### PR DESCRIPTION
**Pull request by simonariddell**
closes #21200
closes #21621
- [x] tests added / passed
- [x] whatsnew entry

This addresses [#21200](https://github.com/pandas-dev/pandas/issues/21200?_pjax=%23js-repo-pjax-container) . When there are different groups in a dataframe, by using groupby it is expected that the pct_change function be applied on each group. However, combining groupby with pct_change does not produce the correct result.

**Explanation:**
Currently the groupby method in the pandas series and pandas dataframe pct change method can implement a vectorized solution, rather than calling apply, if certain conditions are met. For the pandas series method, the vectorized solution is the only option.
This is certainly inappropriate in cases where the groupby object is non-monotonic in its group order. To solve this I've added a check for monotonicity in both the series and dataframe implementation, as well as adding the opportunity to call apply for the series method.
In addition, I have augmented the UT to accept a parameter that can shuffle the dataframe, in order to ensure that the correct calculation occurs.

**Concern**
One concern I have is that depending on whether the apply or vectorized solution is used within the pct change method (e.g. depending on whether the groupby object is monotonic or not), the result returned to the user may have a different index structure. While this was the case prior to the PR, It's not clear to me if (1) this is an acceptable design within the pandas infrastructure, and (2) whether or not this is within the scope of a single PR that was originally opened to address a very specific bug. 

As this is my first pandas PR, I would certainly appreciate feedback, and will incorporate any constructive feedback into future issues.

### CODE REVIEW

**Comment by WillAyd** [comment]
Just eyeballing this I do not like that the implementation is going to vary depending on the monotonicity of the group as that has relatively large performance implications that the standard user will not understand nor should they care to

Is there any way to preserve the vectorized solution and come to the same result?

**Comment by SimonAlecks** [comment]
@WillAyd It's hard for me to think of a scenario where the vectorized solution would work if the data is not in order with respect to groups, although perhaps the monotonicity requirement is overkill? I'm new to Pandas source code, so I'll investigate the existing helper functions that may preserve the intended logic but lessen the requirements. 

I'm also a little concerned though that the vectorized solution may be getting the first element of each group wrong, if it's calculating its percentage change with respect to the last element of the previous group. I'll check this when I get home.

**Comment by SimonAlecks** [comment]
After changing the UT to have a value in the 0th mock array element, the UT begins to fail. The non-apply based vectorized solution doesn't seem to work, as it calculates a pct diff between the first element of a group, and the last element of the preceding group.

It seems to me the simplest solution to solve this issue is to remove the potential to use a vectorized solution. In addition to solving the original issue, it also solves the other issue I discovered above, as well as the additional bonus of letting the user expect the same index structure returned whether or not the vectorized style was called or not (granted, this could easily be fixed).

I suppose we could try preserving the vectorized solution, then writing a little algo to find the first element of each group and setting it to nan. I have no prior knowledge as to whether this would improve performance vs. calling apply. My intuition says at this point you'd be better off just using apply in the first place, but I could very well be wrong.

I also left my updated UT in, as it would catch these cases in case they occurred again, or if a vectorized solution was reimplemented. 

PS: Does anyone have a link they can share with me on how to troubleshoot/debug the continuous-integration/travis-ci/pr failure?

**Review by jreback — CHANGES_REQUESTED** [review]
you are killing performance here, this is why it was changed originally. The real fix is to do this in a cythonized version, similar to the way shift/fillna are done.

File: pandas/tests/groupby/test_transform.py [hunk 1]
```diff
@@ -743,14 +753,18 @@ def get_result(grp_obj):
...
     if test_series:
-        exp = pd.Series(exp_vals * 2)
-        exp.name = 'vals'
+        exp = exp.loc[:, 'A']
         grp = grp['vals']
         result = get_result(grp)
+        df.insert(0, 'A', result)
+        result = df.sort_values(by='key')
+        result = result.loc[:, 'A']
+        result = result.reset_index(drop=True)
```
jreback: don't use inplace in testing

pandas/tests/groupby/test_transform.py
```diff
@@ -722,19 +722,29 @@ def interweave(list_obj):
...
-    exp_vals = Series(vals).pct_change(periods=periods,
-                                       fill_method=fill_method,
-                                       limit=limit).tolist()
-
-    df = DataFrame({'key': ['a'] * len(vals) + ['b'] * len(vals),
-                    'vals': vals * 2})
+def test_pct_change(test_series, shuffle, periods, fill_method, limit):
+    vals = [3, np.nan, 1, 2, 4, 10, np.nan, np.nan]
+    keys = ['a', 'b']
+    key_v = [k for j in list(map(lambda x: [x] * len(vals), keys)) for k in j]
```
jreback: if you really want to do this, use ``np.repeat``

pandas/tests/groupby/test_transform.py
```diff
@@ -722,19 +722,29 @@ def interweave(list_obj):
...
+    df = DataFrame({'key': key_v, 'vals': vals * 2})
+    if shuffle:
+        order = np.random.RandomState(seed=42).permutation(len(df))
+        df = df.reindex(order).reset_index(drop=True)
+
+    manual_apply = []
+    for k in keys:
```
jreback: why aren't you just using this? ``df.loc[df.key == k, 'vals']``, which is a Series

pandas/tests/groupby/test_transform.py
```diff
@@ -722,19 +722,29 @@ def interweave(list_obj):
...
+        subgroup = Series(df.loc[df.key == k, 'vals'].values)
+        manual_apply.append(subgroup.pct_change(periods=periods,
+                                                fill_method=fill_method,
+                                                limit=limit))
+    exp_vals = pd.concat(manual_apply).reset_index(drop=True)
+    exp = pd.DataFrame(exp_vals, columns=['A'])
```
jreback: concat already returns a frame, use ``ignore_index=True`` in the concat

jreback (changes requested):

WillAyd:
Just looking at this can't you just changed the following line:

https://github.com/pandas-dev/pandas/blob/c85ab083919b59ce84c220d5baf7d34ff4a0bcf2/pandas/core/groupby/groupby.py#L2081

To be `self.shift` instead of `filled.shift`? I think that emulates exactly what the user is doing in the example cited so may just be a typo in the first place. That would preserve the performance and be much preferable if it works

SimonAlecks:
In my most recent commit I tried to preserve vectorization, while also not making costly calls to is_monotonic, and preventing group contamination within the vectorization (e.g. first element of group B being divided by an element of group A).

I did this by only allowing the code to attempt vectorization if it can confirm that the series is monotonic, *and* if the truth value of whether or not it is monotonic already exists in the cache. If these conditions are met, it will do a vectorized pct change operation, and then a vectorized operation to set any values that shouldn't exist to np.nan (e.g. for shift=1 the first element of a group would be divided by the last element of the prior group, but this is an invalid value).

I expect this is trying to be too fancy given the problem and my knowledge of pandas source, so I'm prepared to nix it if this is the case, and just keep it simple and/or try to implement it in cython. Please just let me know.  Else I will extend it to the dataframe method as well. 

Hi William,
I tried what you suggested -- but I don't _think_ it makes sense. I think the current logic is correct in so far as shifted should simply be the filled series that is shifted once. Another way to say it, is changing the code to your suggestion wouldn't address the issue that non-monotonic group ordering is handled incorrectly, and that even if it were monotonic, the first element of a group may have a pct change with respect to the last element of the previous group.

pandas/tests/groupby/test_transform.py
```diff
@@ -722,35 +722,56 @@ def interweave(list_obj):
...
-def test_pct_change(test_series, periods, fill_method, limit):
-    vals = [np.nan, np.nan, 1, 2, 4, 10, np.nan, np.nan]
-    exp_vals = Series(vals).pct_change(periods=periods,
-                                       fill_method=fill_method,
-                                       limit=limit).tolist()
-
-    df = DataFrame({'key': ['a'] * len(vals) + ['b'] * len(vals),
-                    'vals': vals * 2})
+    (-1, 'bfill', None), (-1, 'bfill', 1),
+    (-1, None, None), (-1, None, 1),
```
WillAyd: `None` is not a valid `fill_method` on master so this is changing the behavior

pandas/core/groupby/groupby.py
```diff
@@ -2070,17 +2070,10 @@ def shift(self, periods=1, freq=None, axis=0):
...
-                                                     fill_method=fill_method,
-                                                     limit=limit, freq=freq,
-                                                     axis=axis))
-
-        filled = getattr(self, fill_method)(limit=limit).drop(
-            self.grouper.names, axis=1)
-        shifted = filled.shift(periods=periods, freq=freq)
-
-        return (filled / shifted) - 1
+        return self.apply(lambda x: x.pct_change(periods=periods,
```
WillAyd: Isn't this essentially eliminating a vectorized solution across the board? How are the ASVs comparing?

WillAyd (changes requested):
This isn't the correct approach and is way too convoluted. My previous suggestion worked fine for your tests, save the tests that introduce allowing None as an argument. 

If however you can find a scenario where what I suggested doesn't work then this needs to be implemented in Cython.

pandas/core/groupby/groupby.py
SimonAlecks: pct_change is blacklisted in benchmarks/groupby.py.

SimonAlecks:
For the reasons I mentioned above a vectorized solution will return incorrect values for pct_change under a set of conditions. I attempted to solve these conditions in the CR above, which William pointed out was too convoluted (I agree). 

The additional change you suggested (self.shift instead of filled.shift) does not work either for the reasons I mentioned above. The new test I wrote would fail under that change.

My suggestion is to update the test, remove the vectorized solution, and create a new issue to implement this in cython, which I'm interested in doing.

WillAyd:
Not sure what you mean by the benchmarks being blacklisted but for illustration purposes here is how your latest compares compares to master:

```bash
     [415012f4]       [eaede342]
+         571±1μs          975±4ms  1708.33  groupby.GroupByMethods.time_dtype_as_group('float', 'pct_change', 'transformation')
+         572±2μs        974±0.6ms  1702.23  groupby.GroupByMethods.time_dtype_as_group('float', 'pct_change', 'direct')
+         592±3μs          635±8ms  1072.93  groupby.GroupByMethods.time_dtype_as_group('int', 'pct_change', 'transformation')
+         606±8μs          628±3ms  1036.87  groupby.GroupByMethods.time_dtype_as_group('int', 'pct_change', 'direct')
+        559±10μs         514±10ms   920.59  groupby.GroupByMethods.time_dtype_as_field('float', 'pct_change', 'direct')
+       529±0.9μs          481±5ms   908.79  groupby.GroupByMethods.time_dtype_as_field('float', 'pct_change', 'transformation')
+         603±8μs         460±20ms   762.64  groupby.GroupByMethods.time_dtype_as_field('int', 'pct_change', 'direct')
+         593±1μs          445±3ms   750.02  groupby.GroupByMethods.time_dtype_as_field('int', 'pct_change', 'transformation')

```

As you can see (and I think have already recognized) this isn't a great solution and kills performance.

> The additional change you suggested (self.shift instead of filled.shift) does not work either for the reasons I mentioned above. The new test I wrote would fail under that change.

I'm not entirely clear on what your reasons are. It's not a requirement that grouped elements be monotonic in any of the Cythonized code. 

If you are getting an error on the test can you be explicit about *what* exactly that error is? IIRC correctly with the last set of tests what I suggested worked fine. I have a hunch that there may be confusion about the test as it was before (i.e. explicitly passing `None` would fail, but that was also expected). I could be wrong but in any case clarifying the failures you are seeing would be very helpful.

Final note - thanks for helping on this. I hope you are taking this feedback constructively and not in any negative fashion (I believe you are but just want to be explicit). Understanding how / when to interface with the Cythonized code is one of the more complicated aspects of this package

SimonAlecks:
I'm very grateful for the feedback, and I definitely appreciate your time. I apologize for the delay since the last commit. I've been working on understanding the code more clearly.

In this situation I've attempted to solve the problem by incorporating your prior feedback. My previous posts were incorrect, and you correctly pointed out to me how the shift operator works.

In this latest CR I've attempted to fix the problem in line with the spirit of the original implementation. The first step is I added an explicit catch for a fill_method, which did not exist before.

I initially attempted to implement your 1-line suggestion, but I noticed that I couldn't simply call self.shift(), as it first needed to have the proper fill method. However, the fill method returns the obj attribute, not the groupby object. 

In order to remedy this, without creating a function that changed the state of the code, I used the following strategy:
https://github.com/SimonAlecks/pandas/blob/64097ccd308db909dc99cbb88eeb6206beedb820/pandas/core/groupby/groupby.py#L2079
I first recreate a new groupby object, but I solve the fill problem as part of this group. I remember I then had to reset the cache in order for some of the meta-data to align properly. I convinced myself this was necessary at the time, but I didn't write down why, so I'm going to go revisit that now to remind myself that I did this for a good reason. More generally, I could see it being the case that this solution is a hack or not idiomatic within pandas. If so, could you please let me know the correct way to achieve this same result?

Once this is done, the result is relatively self-explanatory. and is pretty similar to the original code.

Lastly, I ran an ASV benchmark locally, and the performance seemed similar, but I need to spend some more time teaching myself how the benchmarking system and code works, as it's possible I made an error.

pandas/core/groupby/groupby.py
```diff
@@ -2075,12 +2075,19 @@ def pct_change(self, periods=1, fill_method='pad', limit=None, freq=None,
                                                      fill_method=fill_method,
                                                      limit=limit, freq=freq,
                                                      axis=axis))
+        if fill_method:
+            new = DataFrameGroupBy(self._obj_with_exclusions,
+                                   grouper=self.grouper)
+            new.obj = getattr(new, fill_method)(limit=limit)
+            new._reset_cache()
```
WillAyd: What's the point of these calls?

pandas/tests/groupby/test_transform.py
```diff
@@ -722,19 +722,30 @@ def interweave(list_obj):
...
+    vals = [3, np.nan, 1, 2, 4, 10, np.nan, 4]
+    keys = ['a', 'b']
+    key_v = np.repeat(keys, len(vals))
+    df = DataFrame({'key': key_v, 'vals': vals * 2})
+    if shuffle:
+        order = np.random.RandomState(seed=42).permutation(len(df))
+        df = df.reindex(order).reset_index(drop=True)
+
+    manual_apply = []
+    for k in keys:
```
WillAyd: This loop / concat mechanism is rather non-idiomatic - whats the reasoning for this instead of just doing a comparison to the anonymous function as was done in the original issue?

pandas/tests/groupby/test_transform.py
```diff
@@ -743,14 +754,21 @@ def get_result(grp_obj):
                                   limit=limit)
 
     if test_series:
-        exp = pd.Series(exp_vals * 2)
-        exp.name = 'vals'
+        exp = exp.loc[:, 'A']
         grp = grp['vals']
         result = get_result(grp)
```
WillAyd: Not that your change introduced it but can we get rid of the `get_result` function? Not sure why this was set up this way and adds unnecessary fluff

pandas/tests/groupby/test_transform.py
```diff
@@ -743,14 +754,21 @@ def get_result(grp_obj):
...
+        df.insert(0, 'A', result)
```
WillAyd: What is the point of all of the insert / sort calls? Would rather just construct the expected value and simply compare that to the result of a pct_change call

WillAyd (changes requested):
Getting closer

pandas/tests/groupby/test_transform.py
SimonAlecks: I chose to do this as it means the test will ultimately assert that the result of a groupby apply is the same as looping over the keys and doing a normal apply, as the two should be identical. Or,  more specifically, the groupby apply should ought to replicate the results of a normal apply looped over the groups.
I'm experimenting now with what an anonymous comparison would look like, but my naive thought is that a comparison to a normal apply is more to-the-point on what we want to test?

pandas/tests/groupby/test_transform.py
SimonAlecks: Removed

pandas/tests/groupby/test_transform.py
SimonAlecks: Agree. Fixed.

SimonAlecks:
Incorporated most requested changes, with a few clarifying questions asked on a few of them. Thanks for your patience!

mroeschke:
@SimonAlecks do you believe this change would also solve #2162?1 If so, could you add a test cases for that issue?

WillAyd (changes requested):

pandas/core/groupby/groupby.py
SimonAlecks: I can simplify those three lines, but I don't believe I can simply use self.ffill(). Let me try to explain:
The first line can be replaced with new = copy(self). However, the reason I call the fill method and reassign it to the .obj attribute, is that later on line 2087 I want to call the groupby implementation of the shift method. Whereas type(self.ffill()) returns a DataFrame. I used this 'trick' of doing the fill and reassigning it back to the .obj attribute in order to still have access to the groupby helper methods. If there is a more idiomatic way to do this, please let me know.

pandas/tests/groupby/test_transform.py
```diff
@@ -722,35 +722,36 @@ def interweave(list_obj):
...
+        df = df.reindex(order).reset_index(drop=True)
+
+    df = getattr(df.groupby('key'), fill_method)(limit=limit)
     grp = df.groupby('key')
-
-    def get_result(grp_obj):
-        return grp_obj.pct_change(periods=periods,
-                                  fill_method=fill_method,
-                                  limit=limit)
+    exp = grp['vals'].shift(0) / grp['vals'].shift(periods) - 1
```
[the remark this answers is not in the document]
SimonAlecks: I'll change exp to expected, and wait for a response on the other comment before updating. I'm still active on this, apologies for the delay in responding to your helpful comments.

mroeschke:
Since you mentioned that this PR fixes #21621, can you go ahead and add a test for that issue?

SimonAlecks:
@mroeschke I think the test I already wrote (well, augmented), should already cover it. It's essentially the exact problem I have attempted to solve here. If you disagree can you enumerate on the point? I don't mind adding a new test if I'm missing something, but it's not clear to me it's necessary.

pandas/tests/groupby/test_transform.py
```diff
@@ -722,35 +722,36 @@ def interweave(list_obj):
...
-    vals = [np.nan, np.nan, 1, 2, 4, 10, np.nan, np.nan]
-    exp_vals = Series(vals).pct_change(periods=periods,
-                                       fill_method=fill_method,
-                                       limit=limit).tolist()
-
-    df = DataFrame({'key': ['a'] * len(vals) + ['b'] * len(vals),
-                    'vals': vals * 2})
+    (-1, 'bfill', None), (-1, 'bfill', 1),
+])
+def test_pct_change(test_series, shuffle, periods, fill_method, limit):
```
mroeschke: Could you add a comment with the issue number for this test (like `GH #21200, #21621`), looks like the other issue is covered here as well.

pandas/core/groupby/groupby.py
WillAyd: Probably just easier if I share code myself at this point :-). Building off of what you have here's my stab at implementation:

https://github.com/WillAyd/pandas/blob/ebe4fffd935f2f16b6e5c635cfa4883e6bacc5ea/pandas/core/groupby/groupby.py#L2070

AFAICT it passes all of the tests you have and therefore provides the same behavior. However, it has the advantage of being more terse and not requiring the construction of a secondary `GroupBy` object.

Not saying you need to exactly mimic this but I'm sharing to highlight how I think this can be done more succinctly again without the need of constructing entirely new objects. If you know of where this would fail then we probably want to add a test case to handle whatever that difference is. Otherwise if they are the same I'd prefer if you adopt something a little more in line with what I've suggested.

No worries on timing - this is open source so any and all contributions are welcome at any time.

WillAyd (changes requested):

pandas/core/groupby/groupby.py
```diff
@@ -2075,12 +2075,19 @@ def pct_change(self, periods=1, fill_method='pad', limit=None, freq=None,
...
+        else:
+            new = self
 
-        filled = getattr(self, fill_method)(limit=limit).drop(
-            self.grouper.names, axis=1)
-        shifted = filled.shift(periods=periods, freq=freq)
+        obj = new.obj.drop(self.grouper.names, axis=1)
+        shifted = new.shift(periods=periods, freq=freq).\
```
WillAyd: Side note - I thought this would fail LINTing but regardless we tend to prefer implicit line continuation, so would be better to do `.drop(` on the first line and continue along with the subsequent code on the next line, so long as that all fits

pandas/core/groupby/groupby.py
SimonAlecks: Thanks for sharing some code, that was very helpful. I updated my test to highlight a case where your code would fail, but what I wrote wouldn't. I agree making a second object seems wrong. However, It's not clear to me how to avoid this.
I'll continue to think about this problem.

WillAyd:
Hmm OK thanks. So I see the difference now between the two sets of code we have. Here's a very small example to illustrate:

```python
df = pd.DataFrame([['a', 1], ['a', np.nan], ['a', 2]], columns=['key', 'val'])
```

Your code yields the following:
```python
In [4]: df.groupby('key').pct_change()
Out[4]: 
   val
0  NaN
1  0.0
2  1.0
```

whereas my branch would do as follows:
```python
In [4]: df.groupby('key').pct_change()
Out[4]: 
   val
0  NaN
1  0.0
2  NaN
```

There difference is whether or not the fill method is getting applied to both the numerator and denominator or just the denominator.

Your's is preferable behavior though the code should still be simplified. Can you see if there's a way to still emulate your behavior borrowing some of my code? It may just need a small tweak, but doing that would be preferable to the copy and private method calls you are currently making

WillAyd (changes requested):

pandas/tests/groupby/test_transform.py
```diff
@@ -722,36 +722,43 @@ def interweave(list_obj):
...
+    # GH  21200, 21621
+    vals = [3, np.nan, np.nan, np.nan, 1, 2, 4, 10, np.nan, 4]
+    keys = ['a', 'b']
+    key_v = np.repeat(keys, len(vals))
+    df = DataFrame({'key': key_v, 'vals': vals * 2})
+    if shuffle:
+        order = np.random.RandomState(seed=42).permutation(len(df))
```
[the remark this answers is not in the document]
SimonAlecks: Eh, you're right. Nixed it.

pandas/tests/groupby/test_transform.py
```diff
@@ -722,36 +722,43 @@ def interweave(list_obj):
...
+        df = df.reindex(order).reset_index(drop=True)
+
+    if fill_method:
+        df_g = getattr(df.groupby('key'), fill_method)(limit=limit)
+        grp = df_g.groupby('key')
+    else:
+        grp = df.groupby('key')
+
+    expected = grp['vals'].obj / grp['vals'].shift(periods) - 1
+    expected = expected.to_frame('vals')
```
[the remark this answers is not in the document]
SimonAlecks: Good point. Fixed.

SimonAlecks:
> Hmm OK thanks. So I see the difference now between the two sets of code we have. Here's a very small example to illustrate:

I cleaned up the code a fair amount. Let me know if anything is still unpalatable or not in line with what you were thinking.

pandas/core/groupby/groupby.py
```diff
@@ -2076,9 +2076,8 @@ def pct_change(self, periods=1, fill_method='pad', limit=None, freq=None,
                                                      limit=limit, freq=freq,
                                                      axis=axis))
         if fill_method:
-            new = copy.copy(self)
+            new = self
```
SimonAlecks: Unfortunately doing this will change the state of self, which seems unwise. However, I also need to get this data into the obj attribute of a groupby object in order to call shift on it later. 
Is there a helper function I'm missing that will solve this issue for me?

pandas/core/groupby/groupby.py
```diff
@@ -3948,14 +3947,11 @@ def _apply_to_column_groupbys(self, func):
     def pct_change(self, periods=1, fill_method='pad', limit=None, freq=None):
         """Calcuate pct_change of each value to previous entry in group"""
         if fill_method:
-            new = SeriesGroupBy(self.obj, grouper=self.grouper)
-            new.obj = getattr(new, fill_method)(limit=limit)
-            new._reset_cache()
-        else:
-            new = self
+            self.obj = getattr(self, fill_method)(limit=limit)
```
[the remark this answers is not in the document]
WillAyd: Generally the state management could be improved (you'll note that in other places in the module). If that's the case can you use the context manager of `_group_selection_context` instead?

pandas/core/groupby/groupby.py
WillAyd: Don't think the branch here is necessary. If state management is the issue and the context manager outlined below works should just be able to assign to a local variable from the context manager and leave `self` untouched

WillAyd (changes requested):

WillAyd:
Thanks for the updates @SimonAlecks . Note that we've reorganized some of the modules here so you'll want to merge in master on your next PR and resolve any merge conflicts

pandas/core/groupby/generic.py
```diff
@@ -1220,11 +1220,14 @@ def _apply_to_column_groupbys(self, func):
...
-        filled = getattr(self, fill_method)(limit=limit)
-        shifted = filled.shift(periods=periods, freq=freq)
-
-        return (filled / shifted) - 1
+        """Calcuate pct_change of each value to previous entry in group"""
+        with _group_selection_context(self) as new:
+            if fill_method:
+                new = copy.copy(new)
+                new.obj = getattr(new, fill_method)(limit=limit)
+                new._reset_cache('_selected_obj')
```
SimonAlecks: Despite being in the context manager, this line is still necessary. The context manager exit calls _reset_group_selection on line 471 of groupby/groupby.py. Here since if self._group_selection is not None does not evaluate to True, it does not make a call to reset cache.

pandas/core/groupby/generic.py
```diff
@@ -1220,11 +1220,14 @@ def _apply_to_column_groupbys(self, func):
...
     def pct_change(self, periods=1, fill_method='pad', limit=None, freq=None):
-        """Calculate percent change of each value to previous entry in group"""
-        filled = getattr(self, fill_method)(limit=limit)
-        shifted = filled.shift(periods=periods, freq=freq)
-
-        return (filled / shifted) - 1
+        """Calcuate pct_change of each value to previous entry in group"""
+        with _group_selection_context(self) as new:
+            if fill_method:
+                new = copy.copy(new)
```
SimonAlecks: I continue to need this for state management. On line 835 of groupby/generic.py I see something similar done i.e.
 obj = copy.copy(obj)
 obj._reset_cache()

I know you previously mentioned state management isn't always great, should I take that to mean that it's okay if a call to pct_change changes the state of .obj in groupby? I'll defer to you on this.

pandas/core/groupby/groupby.py
```diff
@@ -1887,12 +1888,10 @@ def pct_change(self, periods=1, fill_method='pad', limit=None, freq=None,
                                                      fill_method=fill_method,
                                                      limit=limit, freq=freq,
                                                      axis=axis))
-
         with _group_selection_context(self) as new:
```
SimonAlecks: Unlike in the series groupby, the context manager here successfully resets the cache, although it provides no solution for preserving the state.

pandas/tests/groupby/test_transform.py
```diff
@@ -766,35 +766,38 @@ def test_pad_stable_sorting(fill_method):
 
 @pytest.mark.parametrize("test_series", [True, False])
 @pytest.mark.parametrize("periods,fill_method,limit", [
+    (1, None, None), (1, None, 1),
```
WillAyd: `None` is not a valid `fill_method` value in the existing code base. If you wanted to add that feature should be a separate change, but for this one let's remove this test case and the corresponding code

WillAyd (changes requested):
@SimonAlecks sorry for the delay in getting back to you. I know we've had some back and forth on this so wanted to take time and review thoroughly in detail.

To get us on the same page, I think the conversation before about the state of the GroupBy object was a red herring. The problem was that the call to fill `self` would return a Series object, and the subsequent call to shift would not respect the original grouping that is a part of `self`. 

Your approach was to try copying the object, which obviously did the trick but isn't quite the idiomatic approach and comes with some other drawbacks. As another option, here is some code that I dropped into the `SeriesGroupBy` implementation which still passes the tests you've created:

```python
filled = getattr(self, fill_method)(limit=limit)        
fill_grp = filled.groupby(self.grouper.labels)        
shifted = fill_grp.shift(periods=periods, freq=freq)

return (filled / shifted) - 1
```

The main thing here is the second line; to reiterate the fill call returns a `Series` and in the existing implementation the subsequent call to `shift` does not respect the groupings. By reusing the labels attached to the `Grouper` object of `self` you can ensure that the same groupings are respected and keep all of the data aligned.

Does that make sense to you? If so can you try to use something similar to that for the `SeriesGroupBy` and `DataFrameGroupBy` implementations instead of the copy calls and context management in there now?

WillAyd:
@SimonAlecks can you merge master?

SimonAlecks:
Hi William,

I did let it go stale -- but I would like to finish this if that's okay.

Thanks,
Simon

WillAyd:
Thanks! Ping if you need more help or clarification

SimonAlecks:
Thanks for your thoughtful response. Using self.grouper.labels seems to be exactly what I was looking for -- thanks for pointing that out! In adding these changes to series and dataframe groupby, I also noticed that the vectorized operation would break for certain usages of frequency. Specifically, the code would *break* on the [following line](
https://github.com/pandas-dev/pandas/pull/21235/commits/5d3fc2dd6752a04ddeb56989d397ea9ee85c8228#diff-bfee1ba9e7cb79839776fac1a57ed940R1231), when an additional index was added. In addition, the shifted/filled series would have mismatched datetime indices, which was causing other issues. As a result I added a conditional to simply call the normal pct change in that case, as well as an additional test check.

WillAyd:
Looking a lot better. Can you provide a concrete example of where `freq` is breaking things? Really want to avoid using an anonymous func as it kills performance.

FYI you also have what I assume to be a stylistic error causing the Travis failure. Make sure to run `ci/code_checks.sh` locally to see what may be causing it

WillAyd:
Thanks for the thorough example and understood. I opened #23918 for that as a separate issue.

Instead of falling back to the lambda could we inspect `shifted` when `periods` is non-zero and a `freq` is specified and simply drop the first index level? If we add a comment referencing the other issue I think it's clearer that that code is a workaround and should be adjusted accordingly whenever the other item gets fixed

SimonAlecks:
No problem, thanks for your response.

My first attempt at implementing your suggestion had some unexpected outcomes. 

After this line I added this:
https://github.com/pandas-dev/pandas/blob/5d3fc2dd6752a04ddeb56989d397ea9ee85c8228/pandas/core/groupby/generic.py#L1231

> ```python
> >>> shifted = fill_grp.shift(periods=periods, freq=freq)
> >>> if isinstance(shifted.index, MultiIndex):
> >>>         shifted.reset_index(level=0, drop=True, inplace=True)

From here, using the test data, we can notice:
> ```python
> >>> print(filled)

> Out[28]: 
2010-01-01     3.0
2010-01-02     3.0
2010-01-03     NaN
2010-01-04     NaN
2010-01-05     1.0
2010-01-06     2.0
2010-01-07     4.0
2010-01-08    10.0
2010-01-09    10.0
2010-01-10     4.0
2010-01-01     3.0
2010-01-02     3.0
2010-01-03     NaN
2010-01-04     NaN
2010-01-05     1.0
2010-01-06     2.0
2010-01-07     4.0
2010-01-08    10.0
2010-01-09    10.0
2010-01-10     4.0
Name: vals, dtype: float64

> ```python
> >>> print(shifted)

> Out[29]: 
2010-01-02     3.0
2010-01-03     3.0
2010-01-04     NaN
2010-01-05     NaN
2010-01-06     1.0
2010-01-07     2.0
2010-01-08     4.0
2010-01-09    10.0
2010-01-10    10.0
2010-01-11     4.0
2010-01-02     3.0
2010-01-03     3.0
2010-01-04     NaN
2010-01-05     NaN
2010-01-06     1.0
2010-01-07     2.0
2010-01-08     4.0
2010-01-09    10.0
2010-01-10    10.0
2010-01-11     4.0
Name: vals, dtype: float64

> ```python
> >>> print((filled / shifted) - 1)

> (filled / shifted) - 1
Out[30]: 
2010-01-01    NaN
2010-01-01    NaN
2010-01-02    0.0
2010-01-02    0.0
2010-01-02    0.0
2010-01-02    0.0
2010-01-03    NaN
2010-01-03    NaN
2010-01-03    NaN
2010-01-03    NaN
2010-01-04    NaN
2010-01-04    NaN
2010-01-04    NaN
2010-01-04    NaN
2010-01-05    NaN
2010-01-05    NaN
2010-01-05    NaN
2010-01-05    NaN
2010-01-06    1.0
2010-01-06    1.0
2010-01-06    1.0
2010-01-06    1.0
2010-01-07    1.0
2010-01-07    1.0
2010-01-07    1.0
2010-01-07    1.0
2010-01-08    1.5
2010-01-08    1.5
2010-01-08    1.5
2010-01-08    1.5
2010-01-09    0.0
2010-01-09    0.0
2010-01-09    0.0
2010-01-09    0.0
2010-01-10   -0.6
2010-01-10   -0.6
2010-01-10   -0.6
2010-01-10   -0.6
2010-01-11    NaN
2010-01-11    NaN
Name: vals, dtype: float64


I suspect I could solve this with a little investigation of the root cause for the index doubling, which seems to be due to having duplicates of dates in each series, and potentially trying to modify the index structure. In fact, I noticed this same issue would occur in the canonical pct_change method if there are duplicate date values in the series. 

Do you think it's worth trying to solve this issue?

WillAyd:
No - the more I think about it the referenced issue is going to cause misalignment of values so division probably won’t work. As you have it is probably the best we can do here. 

Can you add a TODO comment referencing the issue and fix up the Travis failure (it’s coming from ci/code_checks.sh)

pandas/core/groupby/generic.py
```diff
@@ -1221,10 +1221,18 @@ def _apply_to_column_groupbys(self, func):
...
+        if freq:
+            return self.apply(lambda x: x.pct_change(periods=periods,
+                                                     fill_method=fill_method,
+                                                     limit=limit, freq=freq))
         filled = getattr(self, fill_method)(limit=limit)
-        shifted = filled.shift(periods=periods, freq=freq)
+        fill_grp = filled.groupby(self.grouper.labels)
+        shifted = fill_grp.shift(periods=periods, freq=freq)
 
+        # TODO: Remove this conditional when #23918 is fixed
```
WillAyd: I don't think you need this and the condition on `freq` right? OK to just have the former but move this comment up

WillAyd (changes requested):

pandas/core/groupby/generic.py
SimonAlecks: Right, good point. 
Although fixing #23918 won't necessarily solve the problem here, right? As even if it does not return a multi-index, it still won't handle division as you pointed up to. 

For now I'll just stick to the relegation to the anonymous function, and remove the check for the multi-index.

SimonAlecks: (which is what I think you were suggesting, but correct me if I'm wrong)

WillAyd: You've got it

pandas/tests/groupby/test_transform.py
```diff
@@ -765,36 +765,48 @@ def test_pad_stable_sorting(fill_method):
...
+    (-1, 'bfill', None), (-1, 'bfill', 1),
+])
+def test_pct_change(test_series, test_freq, periods, fill_method, limit):
+    # GH  21200, 21621
+    vals = [3, np.nan, np.nan, np.nan, 1, 2, 4, 10, np.nan, 4]
+    keys = ['a', 'b']
+    key_v = np.repeat(keys, len(vals))
+    df = DataFrame({'key': key_v, 'vals': vals * 2})
+
+    if test_freq:
```
WillAyd: It would be cleaner if you just parametrize `test_freq` to be `None` and 'D'. You can `xfail` `D` for now

pandas/tests/groupby/test_transform.py
```diff
@@ -765,36 +765,48 @@ def test_pad_stable_sorting(fill_method):
...
+
+    expected = grp['vals'].obj / grp['vals'].shift(periods) - 1
 
     if test_series:
-        exp = pd.Series(exp_vals * 2)
-        exp.name = 'vals'
-        grp = grp['vals']
-        result = get_result(grp)
-        tm.assert_series_equal(result, exp)
+        result = df.groupby('key')['vals'].pct_change(periods=periods,
```
WillAyd: Style nit but this is more readable if you break `periods=periods` on a newline and move up the remaining parameters to that same line (within the 79 char limit)

pandas/tests/groupby/test_transform.py
```diff
@@ -765,36 +765,48 @@ def test_pad_stable_sorting(fill_method):
...
+
+    if test_freq:
+        freq = 'D'
+        dt_idx = pd.DatetimeIndex(start='2010-01-01', freq=freq,
+                                  periods=len(vals))
+        df.index = np.concatenate([dt_idx.values] * 2)
+    else:
+        freq = None
+
+    if fill_method:
```
WillAyd: Since we always have a `fill_method` do we still need this condition?

WillAyd (changes requested):

pandas/tests/groupby/test_transform.py
SimonAlecks: I think you're right, let me run the tests.

pandas/tests/groupby/test_transform.py
```diff
@@ -765,36 +765,37 @@ def test_pad_stable_sorting(fill_method):
...
+
+    expected = grp['vals'].obj / grp['vals'].shift(periods) - 1
 
     if test_series:
-        exp = pd.Series(exp_vals * 2)
-        exp.name = 'vals'
-        grp = grp['vals']
-        result = get_result(grp)
-        tm.assert_series_equal(result, exp)
+        result = df.groupby('key')['vals'].pct_change(
```
WillAyd: Do you need the selection of ['vals'] here? Think it should get the same result without it no? If so you could make this expression the same as the one in the else block, but then move it out of the condition altogether

pandas/tests/groupby/test_transform.py
```diff
@@ -765,36 +765,37 @@ def test_pad_stable_sorting(fill_method):
...
-
-    def get_result(grp_obj):
-        return grp_obj.pct_change(periods=periods,
-                                  fill_method=fill_method,
-                                  limit=limit)
+    (-1, 'bfill', None), (-1, 'bfill', 1),
+])
+def test_pct_change(test_series, freq, periods, fill_method, limit):
+    # GH  21200, 21621
+    if freq == 'D':
```
WillAyd: Sorry I was not thinking correctly about this before; no need to xfail

WillAyd (changes requested):

pandas/tests/groupby/test_transform.py
SimonAlecks: By selecting on ['vals'] here, it means the pct_change is called on the series, and since there is a separate method for pct_change for series vs. dataframe, it seems necessary in order to test each. By moving it out of the conditional we would end up only testing one of those two groupby methods.

pandas/tests/groupby/test_transform.py
SimonAlecks: In that case I'll need to add back in some of the previously deleted code in order to ensure we have a datetime index.
i.e.

> ```python
>        dt_idx = pd.DatetimeIndex(start='2010-01-01', freq='D', periods=len(vals))
>        idx = np.concatenate([dt_idx.values]*2)

SimonAlecks: (although I think this makes more sense)

pandas/tests/groupby/test_transform.py
WillAyd: Ignore my comment this is fine as is

pandas/tests/groupby/test_transform.py
WillAyd: Let's just hold off on this until the fix for freq

WillAyd (approved):
OK I think this is in a good enough state with appropriate follow ups documented. @jreback over to you

jreback (changes requested):
looks ok. this needs a whatnew entry. I think its important enough to warrant a sub-section, with a small example of what it was doing before and the correct result. you an put it in right below the grouping section in Bug Fixes I think is ok.

SimonAlecks:
Added the whatsnew. I checked for previous examples in the docs to follow style guidelines, but please let me know if it's too verbose or I made a stylistic mistake, and I'll quickly fix it. Thanks.

jreback (changes requested):

doc/source/whatsnew/v0.23.5.txt
```diff
@@ -42,7 +42,33 @@ Bug Fixes
 **Groupby/Resample/Rolling**
 
 - Bug in :meth:`DataFrame.resample` when resampling ``NaT`` in ``TimeDeltaIndex`` (:issue:`13223`).
--
+- Bug where calling :func:`SeriesGroupBy.pct_change` or :func:`DataFrameGroupBy.pct_change` would ignore groups when calculating the percent change (:issue:`21235`)
+
+.. ipython:: python
+
+   df = pd.DataFrame({'a': [1.0, 1.1, 2.2], 'grp': ['a', 'a', 'b']})
+
```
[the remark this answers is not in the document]
SimonAlecks: Added

doc/source/whatsnew/v0.23.5.txt
```diff
@@ -42,7 +42,33 @@ Bug Fixes
...
+        a
+   0  NaN
+   1  0.1
+   2  1.0
+
+**New behavior**:
+
+Grouping respected
+
+.. code-block:: ipython
```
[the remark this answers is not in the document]
SimonAlecks: Fixed. Didn't properly read contributor documentation on how to do documentation -- sorry for the churn.

WillAyd (changes requested):
We actually aren't doing a v0.23.5 release anymore. If you can merge master and move the whatsnew to v0.24 would be great

WillAyd (approved):

jreback (changes requested):

doc/source/whatsnew/v0.24.0.rst
```diff
@@ -379,6 +379,33 @@ Backwards incompatible API changes
 - ``max_rows`` and ``max_cols`` parameters removed from :class:`HTMLFormatter` since truncation is handled by :class:`DataFrameFormatter` (:issue:`23818`)
 - :meth:`read_csv` will now raise a ``ValueError`` if a column with missing values is declared as having dtype ``bool`` (:issue:`20591`)
 
+Percentage change called on groupby now returns correct values
```
jreback: call this 'Percentage change on groupby changes'

doc/source/whatsnew/v0.24.0.rst
```diff
@@ -379,6 +379,33 @@ Backwards incompatible API changes
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Fixed a bug where calling :func:`SeriesGroupBy.pct_change` or :func:`DataFrameGroupBy.pct_change` would ignore groups when calculating the percent change (:issue:`21235`).
```
jreback: Can you add a tiny bit more here. A user reading this will want to know what is changing, IOW the groupby percentage worked across groups, now it correctly works per group.

pandas/tests/groupby/test_transform.py
```diff
@@ -765,36 +765,37 @@ def test_pad_stable_sorting(fill_method):
...
+        pytest.xfail("'freq' test not necessary until #23918 completed and"
+                     "freq is used in the vectorized approach")
```
jreback: you could put the xxfail inthe parametrize, something like this format

```
@pytest.mark.parametrize("engine,ext", [
    pytest.param('openpyxl', '.xlsx', marks=pytest.mark.skipif(
        not td.safe_import('openpyxl'), reason='No openpyxl')),
```

but of course its an xfail and not a skip and this is not an excel test :<

jreback (changes requested):
pls merge master as well. small changes. ping on green.

doc/source/whatsnew/v0.24.0.rst
jreback: can you also list both issues here

pandas/tests/groupby/test_transform.py
WillAyd: Would this actually fail or be non-performant compared to the other argument combinations? I think the latter and if so probably OK to just remove xfail / skip altogether; we have the TODO in the actual code in regards to the perf

SimonAlecks: It would be non-performant, but it wouldn't fail (well it technically would fail, but because the index in the test is not a date-time). Personally, I think it makes sense to make the index date-time, and not xfail this. Let me know if you agree, and I'll do that.

WillAyd: Ah OK. No I think that would make the test here more complicated than it needs to be; if you could follow the approach as outlined by @jreback would be preferable

SimonAlecks: Sure!

jreback (approved):

jreback:
thanks @SimonAlecks

SimonAlecks:
@jreback and @WillAyd thanks so much for your patience on my first PR! 
Any suggestions for the next bug I tackle? I'd like to learn the index structure. (I'm perfectly capable of finding a bug to work on on my own, just curious if anything in particular that you have a preference for).

WillAyd:
Since it came out of this PR #23918 would be a logical follow up

pandas/tests/groupby/test_transform.py
jbrockmendel: I'm trying to understand this xfail and it seems like the behavior is correct bc, as @simonariddell mentioned, it _should_ raise without a DatetimeIndex.  what is the xfailed case supposed to be testing?

PR #21238, opened 2018-05-29
BUG: Support to create DataFrame from list subclasses

mitar:
- [x] closes #21226
- [x] tests added / passed
- [x] whatsnew entry

Alternative fix could be that instead of doing `isinstance` check in frame's `_to_arrays` method:

```
if isinstance(data[0], (list, tuple)):
```

We would do:

```
if type(data[0]) in (list, tuple):
```

This would then assure that `to_object_array` is called really just with exactly lists. But currently there is a mismatch, `if` checks with subtypes, while `to_object_array` does not support them. I think it is better to support them so this merge request adds support for subtypes/subclasses of list.

### DIFF
--- a/pandas/_libs/lib.pyx
+++ b/pandas/_libs/lib.pyx
@@ -2208,7 +2208,7 @@ def map_infer(ndarray arr, object f, bint convert=1):
     return result
 
 
-def to_object_array(rows: list, min_width: int=0):
+def to_object_array(rows: object, int min_width=0):
     """
     Convert a list of lists into an object array.
 
@@ -2229,20 +2229,22 @@ def to_object_array(rows: list, min_width: int=0):
     cdef:
         Py_ssize_t i, j, n, k, tmp
         ndarray[object, ndim=2] result
+        list input_rows
         list row
 
-    n = len(rows)
+    input_rows = <list>rows
+    n = len(input_rows)
 
     k = min_width
     for i in range(n):
-        tmp = len(rows[i])
+        tmp = len(input_rows[i])
         if tmp > k:
             k = tmp
 
     result = np.empty((n, k), dtype=object)
 
     for i in range(n):
-        row = rows[i]
+        row = <list>input_rows[i]
 
         for j in range(len(row)):
             result[i, j] = row[j]
--- a/asv_bench/benchmarks/frame_ctor.py
+++ b/asv_bench/benchmarks/frame_ctor.py
@@ -91,4 +91,17 @@ class FromNDArray(object):
         self.df = DataFrame(self.data)
 
 
+class FromLists(object):
+
+    goal_time = 0.2
+
+    def setup(self):
+        N = 1000
+        M = 100
+        self.data = [[j for j in range(M)] for i in range(N)]
+
+    def time_frame_from_lists(self):
+        self.df = DataFrame(self.data)
+
+
 from .pandas_vb_common import setup  # noqa: F401
--- a/doc/source/whatsnew/v0.24.0.rst
+++ b/doc/source/whatsnew/v0.24.0.rst
@@ -1606,6 +1606,7 @@ Other
 - Bug in :meth:`DataFrame.combine_first` in which column types were unexpectedly converted to float (:issue:`20699`)
 - Bug where C variables were declared with external linkage causing import errors if certain other C libraries were imported before Pandas. (:issue:`24113`)
 - Constructing a DataFrame with an index argument that wasn't already an instance of :class:`~pandas.core.Index` was broken in `4efb39f <https://github.com/pandas-dev/pandas/commit/4efb39f01f5880122fa38d91e12d217ef70fad9e>`_ (:issue:`22227`).
+- Bug in :func:`to_object_array` prevented list subclasses to be used to create :class:`DataFrame` (:issue:`21226`)
 
 .. _whatsnew_0.24.0.contributors:
 
--- a/pandas/tests/frame/test_constructors.py
+++ b/pandas/tests/frame/test_constructors.py
@@ -2165,6 +2165,15 @@ class TestDataFrameConstructors(TestData):
         result = DataFrame({'A': range(5)}, dtype=dtype)
         tm.assert_frame_equal(result, expected)
 
+    def test_frame_from_list_subclass(self):
+        # GH21226
+        class List(list):
+            pass
+
+        expected = DataFrame([[1, 2, 3], [4, 5, 6]])
+        result = DataFrame(List([List([1, 2, 3]), List([4, 5, 6])]))
+        tm.assert_frame_equal(result, expected)
+
 
 class TestDataFrameConstructorWithDatetimeTZ(TestData):
 


WillAyd (changes requested):

pandas/_libs/src/inference.pyx
```diff
@@ -1508,8 +1508,10 @@ def to_object_array(list rows, int min_width=0):
     cdef:
         Py_ssize_t i, j, n, k, tmp
         ndarray[object, ndim=2] result
+        list input_rows
         list row
 
+    input_rows = <list>rows
```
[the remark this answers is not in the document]
mitar: It is just to type check. :-) So this will throw an exception if it is not really a list or its subclass. I can also use it later on.

jorisvandenbossche:
Can you give a timing of the benchmark that you added (I would just do it with `%timeit` using the same code, that's a bit easier / more reliable) before/after? 
I don't expect a difference, but just to be sure.

jorisvandenbossche (approved):

pandas/_libs/src/inference.pyx
```diff
@@ -1487,7 +1487,7 @@ def map_infer(ndarray arr, object f, bint convert=1):
     return result
 
 
-def to_object_array(list rows, int min_width=0):
+def to_object_array(rows, int min_width=0):
```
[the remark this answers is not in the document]
mitar: No performance impact.

jorisvandenbossche:
What I meant was doing (before and after):

```
In [7]: N = 1000
   ...: M = 100
   ...: data = [[j for j in range(M)] for i in range(N)]

In [8]: %timeit DataFrame(data)
31.1 ms ± 357 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

as if you want to time a very specific thing, `%timeit` is both easier and more accurate than ASV in my experience (but asv is still important to keep track of this in long term / benchmark changes that can have wide impact).
But what you posted is fine as well! Performance is not affected.

mitar:
Oh, I didn't know about `%timeit`.

pandas/_libs/src/inference.pyx
```diff
@@ -1508,20 +1508,22 @@ def to_object_array(list rows, int min_width=0):
     cdef:
         Py_ssize_t i, j, n, k, tmp
         ndarray[object, ndim=2] result
+        list input_rows
```
WillAyd: Is there not a way to make the type declaration part of the call signature? Just wondering if that wouldn't be cleaner. Also curious if doing it in this fashion doesn't require more memory to construct a DataFrame, which maybe is an issue if dealing with very large lists

mitar: I am not an expert in Cython, but this approach is what I found using Google on their mailing list. :-) So not having it in the signature and cast it later. (Casting still throws an exception if impossible.)

I do not see why would this require any more memory?

WillAyd: May not - just not sure how the casting rules are applied. Looking at the module there appear to be other instances where the type is in the signature (see `to_object_array_tuples`). Seems like that would be preferable if its an option

mitar: What do you mean? The whole issue is that if you put the `list` in the signature, then Cython requires exactly the list signature and not a subclass. So removing the signature and doing casting fixes this. In `to_object_array_tuples` the argument type is `list`, which again means that only strict list type can be provided.

mitar: I can put `object` type there. It seems equivalent. This is what `infer_dtype` has.

jreback:
why is a List subclass actually useful?

mitar:
I mean, there are different places where you can get data which is a list subclass. So if you want to convert such data to DataFrame, then why not? For example, we are using a list subclass to have additional methods on the list and additional metadata.

In any case, Pandas is currently inconsistent and something has to be changed. Currently is is checking with `isnstance(data, list)` if you can call this method, and then calls it, which fails. So or that check is changed, or this method is changed.

I do not understsand why we would not change this method? There is no performance penalty, and it allows easier interoperability with other data. In general Pandas tries to be well integrated with Python ecosystem, so if something is an instance of list, in Python, this means that you can always use it as list. But Pandas currently does not support this.

So, my question is, why not? What is problematic in this pull request?

jorisvandenbossche:
I am fine with this patch. It is a minor change and easy to support it. 
In any case, the current error message is not really use friendly (due to the inconsistency @mitar notes between the isinstance check and the actual cython function)

jreback:
The problem is that, we have many many places where we actually check for lists for various reasons, e.g. its is not a scalar, not a tuple and also we care if things are array-like (IOW have a dtype) or not. So  sure this patch looks ok, but this is actually introducing way more inconsistencies that trying to solve one.

mitar:
Are you checking elsewhere with `type(...) is` or with `isinstance`?

jreback:
if you want to merge master and update, can see where this is

jreback:

rok:
> closing as stale. if you'd like to continue, pls ping.

Hey @jreback I'd like to help close this. Could you reopen the PR please so I commit changes to it?

rok:
Hey all, I think the discussion here was already completed and we can merge?

pandas/_libs/lib.pyx
```diff
@@ -2208,7 +2208,7 @@ def map_infer(ndarray arr, object f, bint convert=1):
     return result
 
 
-def to_object_array(rows: list, min_width: int=0):
+def to_object_array(rows, int min_width=0):
```
jreback: can you leave the type annotation

jreback: type rows as object

jreback (changes requested):
can you show the result of running that benchmarks (before and after). also can you run the DataFrame constructor benchmarks and report.

jreback:
@WillAyd if you have any commets

pandas/_libs/lib.pyx
rok: Will post the benchmark results when they complete.

rok: I've rerun twice overnight and reported significant changes don't overlap between runs. I assume this is just the benchmark being noisy. (Commit hashes changed because I rebased again)

```
     [d43ac979]       [8661bbb6]
     <master>         <from-list>
+        50.7±2ms         93.0±3ms     1.84  frame_methods.Iteration.time_iteritems_indexing
+     6.32±0.02μs         7.65±1μs     1.21  timestamp.TimestampOps.time_normalize(tzutc())
+     11.5±0.03μs         13.3±1μs     1.16  offset.OffestDatetimeArithmetic.time_apply(<BusinessMonthEnd>)
+     2.51±0.01μs       2.88±0.2μs     1.15  timeseries.DatetimeIndex.time_get('dst')
+     2.50±0.01μs       2.85±0.2μs     1.14  timeseries.DatetimeIndex.time_get('tz_naive')
+      11.8±0.2μs         13.4±1μs     1.13  timestamp.TimestampConstruction.time_parse_iso8601_tz
+     2.66±0.01μs       3.00±0.2μs     1.13  timeseries.DatetimeIndex.time_get('repeated')
+     20.8±0.07μs         23.1±2μs     1.11  timestamp.TimestampProperties.time_weekday_name(<DstTzInfo 'Europe/Amsterdam' LMT+0:20:00 STD>, None)
-      8.40±0.3ms       7.48±0.2ms     0.89  indexing.NumericSeriesIndexing.time_getitem_scalar(<class 'pandas.core.indexes.numeric.UInt64Index'>, 'nonunique_monotonic_inc')
-      9.62±0.3ms       8.52±0.3ms     0.89  indexing.NumericSeriesIndexing.time_ix_slice(<class 'pandas.core.indexes.numeric.UInt64Index'>, 'nonunique_monotonic_inc')
-      8.85±0.4ms       7.69±0.1ms     0.87  indexing.NumericSeriesIndexing.time_loc_scalar(<class 'pandas.core.indexes.numeric.UInt64Index'>, 'nonunique_monotonic_inc')
-      8.68±0.8μs      7.48±0.03μs     0.86  indexing.CategoricalIndexIndexing.time_get_loc_scalar('monotonic_incr')
-      1.13±0.2μs          896±6ns     0.79  period.PeriodProperties.time_property('min', 'day')
-      86.8±0.6ms          100±4μs     0.00  multiindex_object.GetLoc.time_large_get_loc

     [d43ac979]       [8661bbb6]
     <master>         <from-list>
+        82.4±1μs          104±1μs     1.26  indexing.NumericSeriesIndexing.time_getitem_scalar(<class 'pandas.core.indexes.numeric.Float64Index'>, 'nonunique_monotonic_inc')
+     7.09±0.06μs       8.21±0.8μs     1.16  indexing.NonNumericSeriesIndexing.time_getitem_scalar('datetime', 'unique_monotonic_inc')
+      18.5±0.3μs         20.6±1μs     1.11  indexing.NonNumericSeriesIndexing.time_get_value('datetime', 'unique_monotonic_inc')
-         803±6ms          714±4ms     0.89  gil.ParallelGroupbyMethods.time_parallel(8, 'mean')

```

jreback (approved):

mitar:
Thanks @rok.

PR #21246, opened 2018-05-29
BUG: Fix handling of encoding for the StataReader #21244

### LINKED ISSUE #21244: BUG: `read_stata` always uses 'utf8'

```python
import pandas
for chunk in data:
```
```python
import pandas
for chunk in data:
```

#### Problem description

```python
        return s.decode('utf-8')
```
to:
```python
        return s.decode('latin-1')
```
```python
```

#### Expected Output

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 2.7.12.final.0
python-bits: 64
OS: Linux
OS-release: 4.10.0-37-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: ro_RO.UTF-8
LANG: ro_RO.UTF-8
LOCALE: None.None

pytest: 3.2.3
pip: 9.0.3
setuptools: 36.6.0
Cython: 0.28.2
numpy: 1.13.3
scipy: 1.0.0
pyarrow: None
xarray: None
IPython: 5.1.0
sphinx: 1.6.3
patsy: None
dateutil: 2.7.3
pytz: 2017.3
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: 2.4.9
xlrd: 1.0.0
xlwt: 1.3.0
xlsxwriter: None
lxml: 3.8.0
bs4: None
html5lib: 0.999999999
sqlalchemy: 1.1.13
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

adrian-castravete:
- [x] closes #21244
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -146,7 +146,8 @@ MultiIndex
 I/O
 ^^^
 
--
+- :func:`pandas.read_stata` now honours the ``encoding`` parameter, and supports the 'utf-8'
+  encoding.  #21244
 -
 -
 
@@ -184,4 +185,3 @@ Other
 -
 -
 -
-
--- a/pandas/io/stata.py
+++ b/pandas/io/stata.py
@@ -37,7 +37,8 @@ from pandas.util._decorators import Appender
 from pandas.util._decorators import deprecate_kwarg
 
 VALID_ENCODINGS = ('ascii', 'us-ascii', 'latin-1', 'latin_1', 'iso-8859-1',
-                   'iso8859-1', '8859', 'cp819', 'latin', 'latin1', 'L1')
+                   'iso8859-1', '8859', 'cp819', 'latin', 'latin1', 'L1',
+                   'utf-8', 'utf8')
 
 _version_error = ("Version of given Stata file is not 104, 105, 108, "
                   "111 (Stata 7SE), 113 (Stata 8/9), 114 (Stata 10/11), "
@@ -1335,7 +1336,7 @@ class StataReader(StataParser, BaseIterator):
 
     def _decode(self, s):
         s = s.partition(b"\0")[0]
-        return s.decode('utf-8')
+        return s.decode(self._encoding or self._default_encoding)
 
     def _null_terminate(self, s):
         if compat.PY3 or self._encoding is not None:
--- a/pandas/tests/io/test_stata.py
+++ b/pandas/tests/io/test_stata.py
@@ -99,9 +99,9 @@ class TestStata(object):
 
         self.stata_dates = os.path.join(self.dirpath, 'stata13_dates.dta')
 
-    def read_dta(self, file):
+    def read_dta(self, file, encoding='latin-1'):
         # Legacy default reader configuration
-        return read_stata(file, convert_dates=True)
+        return read_stata(file, convert_dates=True, encoding=encoding)
 
     def read_csv(self, file):
         return read_csv(file, parse_dates=True)
@@ -268,7 +268,7 @@ class TestStata(object):
         tm.assert_frame_equal(parsed_117, expected, check_dtype=False)
 
     def test_read_dta18(self):
-        parsed_118 = self.read_dta(self.dta22_118)
+        parsed_118 = self.read_dta(self.dta22_118, encoding='utf-8')
         parsed_118["Bytes"] = parsed_118["Bytes"].astype('O')
         expected = DataFrame.from_records(
             [['Cat', 'Bogota', u'Bogotá', 1, 1.0, u'option b Ünicode', 1.0],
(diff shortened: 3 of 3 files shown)


doc/source/whatsnew/v0.24.0.txt
```diff
@@ -146,7 +146,8 @@ MultiIndex
 I/O
 ^^^
 
--
+- :func:`pandas.read_stata` now honours the ``encoding`` parameter, and supports the 'utf-8'
+  encoding.
```
jreback: add the issue number

jreback (changes requested):
is it possible include / and/or generate a utf-8 encoded file for testing?

jreback:
@bashtage can you have a look

bashtage:
You need to add a small Stata produce dta file that can replicate the issue and that this or fixes. Then please add a test that reads this dta file. It needs to be produced by Stata and not pandas.

bashtage:
118 does support utf8. But this needs to be tested using a real Stata file with both fixed width utf8 strings and StrL utf8 characters, and numbers.

doc/source/whatsnew/v0.24.0.txt
adrian-castravete: Adding now.

adrian-castravete:
I see. I thought that https://github.com/pandas-dev/pandas/blob/master/pandas/tests/io/data/stata14_118.dta contains Unicode, since the test passes with unicode characters. And the `TestStata.test_read_dta18` test from https://github.com/pandas-dev/pandas/blob/master/pandas/tests/io/test_stata.py has a check for that.
Though indeed I do not know if the generated file was made by Stata or not.

Unfortunately I don't own a copy of the software and it's neither in my _field_ to work with stata files as a Scientist/Statician. :)
My use case is for a converter for a simple view of the file.

bashtage:
It possible the standard strings might be latin-1 not unicode. Are you sure
that the reader doesn't already read utf8 without passing an encoding? The
dta 118 spec says all strings are unicode and so this suggests that the
encoding is always utf8 irrespective of the one passed in. If so, then the
reader shouldn't be changed and a note should be added that encoding is
ignored for 118 files.

More generally I think reader should remove encoding and always use latin-1
for dta <118 and utf8 for 118.

adrian-castravete:
When I tried `encoding` the string with unicode and then decoding it with `latin-1` the string was different, It showed two weird characters instead of the expected `Ü`.

As for the `encoding` argument: I agree if the specs say that 118 is always Unicode then it would make sense to add the necessary ifs where applicable. I've seen parts of the code where the encoding gets set to `latin-1`. 

If the encoding is to be determined via the version number, then this argument should also be removed and everything else like _documentation_ or _tests_ should be changed to reflect this.

pandas/io/stata.py
```diff
@@ -37,7 +37,8 @@
 from pandas.util._decorators import deprecate_kwarg
 
 VALID_ENCODINGS = ('ascii', 'us-ascii', 'latin-1', 'latin_1', 'iso-8859-1',
-                   'iso8859-1', '8859', 'cp819', 'latin', 'latin1', 'L1')
+                   'iso8859-1', '8859', 'cp819', 'latin', 'latin1', 'L1',
```
bashtage: These are not in general valid.  The set of valid depends on the reader.  For < 118 it is  

```
('ascii', 'us-ascii', 'latin-1', 'latin_1', 'iso-8859-1',
'iso8859-1', '8859', 'cp819', 'latin', 'latin1', 'L1')
```

for 118 it is `'utf-8', 'utf8'`.

pandas/io/stata.py
```diff
@@ -1335,7 +1336,7 @@ def _calcsize(self, fmt):
 
     def _decode(self, s):
         s = s.partition(b"\0")[0]
-        return s.decode('utf-8')
+        return s.decode(self._encoding or self._default_encoding)
```
bashtage: This should not be changed.

pandas/tests/io/test_stata.py
```diff
@@ -99,9 +99,9 @@ def setup_method(self, method):
 
         self.stata_dates = os.path.join(self.dirpath, 'stata13_dates.dta')
 
-    def read_dta(self, file):
+    def read_dta(self, file, encoding='latin-1'):
```
bashtage: Should use `None` as encoding so it can be overridden by the reader depending on the dta version.

pandas/tests/io/test_stata.py
```diff
@@ -99,9 +99,9 @@ def setup_method(self, method):
...
         # Legacy default reader configuration
-        return read_stata(file, convert_dates=True)
+        return read_stata(file, convert_dates=True, encoding=encoding)
```
bashtage: Same.

bashtage (commented):
Overall I think the encoding should be changd form settable to not settable.  This will require a deprecation in StataReader.  Users should not be encouraged to set encoding in `read_stata` since it isn't really settable (either latin-1 or utf-8)

bashtage:
I took a look at the dta spec and it is stricter than pandas enforces. dta < 118 claim to use ASCII only although Stata internally displays and works with latin-1.  dta 118 is utf-8 only.

bashtage:
@adrian-castravete Master just got updated with a fix for this. If you have a chance could you try it out with your dta file?  It should always use the correct encoding (automatically) now. If it fails, we might need to look into your dta file.

bashtage:
I think this has been resolved in master.

jreback:
deprecated encoding by #21400 (and bug is already fixed in master)

PR #21251, opened 2018-05-29
Append Mode for ExcelWriter with openpyxl

### LINKED ISSUE #3441: Allow ExcelWriter() to add sheets to existing workbook
The ability of ExcelWriter to save different dataframes to different worksheets is great for sharing those dfs with the python-deficient.  But this quickly leads to a need to _add_ worksheets to an existing workbook, not just creating one from scratch; something like:

```
df0=pd.DataFrame(np.arange(3))
df0.to_excel('foo.xlsx','Data 0')

df1=pd.DataFrame(np.arange(2))
df1.to_excel('foo.xlsx','Data 1')
```

The following little diff to io/parsers.py implements this behavior for *.xlsx files:

```
diff --git a/pandas/io/parsers.py b/pandas/io/parsers.py
index 89f892d..7f010ee 100644
--- a/pandas/io/parsers.py
+++ b/pandas/io/parsers.py
@@ -2099,12 +2099,19 @@ class ExcelWriter(object):
             self.fm_date = xlwt.easyxf(num_format_str='YYYY-MM-DD')
         else:
             from openpyxl.workbook import Workbook
-            self.book = Workbook()  # optimized_write=True)
-            # open pyxl 1.6.1 adds a dummy sheet remove it
-            if self.book.worksheets:
-                self.book.remove_sheet(self.book.worksheets[0])
+            from openpyxl.reader.excel import load_workbook
+
+            try:
+               self.book=load_workbook(filename = path)
+               self.sheets={wks.title:wks for wks in self.book.worksheets}
+            except InvalidFileException:
+                self.book = Workbook()  # optimized_write=True)
+                # open pyxl 1.6.1 adds a dummy sheet remove it
+                if self.book.worksheets:
+                    self.book.remove_sheet(self.book.worksheets[0])
+                self.sheets = {}
+
         self.path = path
-        self.sheets = {}
         self.cur_sheet = None
```

Doing this for *.xls files is a little harder.

WillAyd:
- [X] closes #3441
- [X] tests added / passed
- [X] whatsnew entry

### DIFF
--- a/pandas/io/excel.py
+++ b/pandas/io/excel.py
@@ -807,6 +807,8 @@ class ExcelWriter(object):
     mode : {'w' or 'a'}, default 'w'
         File mode to use (write or append).
 
+    .. versionadded:: 0.24.0
+
     Notes
     -----
     For compatibility with CSV writers, ExcelWriter serializes lists
@@ -1012,13 +1014,12 @@ class _OpenpyxlWriter(ExcelWriter):
             # Create workbook object with default optimized_write=True.
             self.book = Workbook()
 
-            # Openpyxl 1.6.1 adds a dummy sheet. We remove it.
             if self.book.worksheets:
                 try:
                     self.book.remove(self.book.worksheets[0])
                 except AttributeError:
 
-                    # compat
+                    # compat - for openpyxl <= 2.4
                     self.book.remove_sheet(self.book.worksheets[0])
 
     def save(self):
--- a/pandas/io/excel.py
+++ b/pandas/io/excel.py
@@ -804,6 +804,8 @@ class ExcelWriter(object):
     datetime_format : string, default None
         Format string for datetime objects written into Excel files
         (e.g. 'YYYY-MM-DD HH:MM:SS')
+    mode : {'w' or 'a'}, default 'w'
+        File mode to use (write or append).
 
     Notes
     -----
@@ -897,7 +899,8 @@ class ExcelWriter(object):
         pass
 
     def __init__(self, path, engine=None,
-                 date_format=None, datetime_format=None, **engine_kwargs):
+                 date_format=None, datetime_format=None, mode='w',
+                 **engine_kwargs):
         # validate that this engine can handle the extension
         if isinstance(path, string_types):
             ext = os.path.splitext(path)[-1]
@@ -918,6 +921,8 @@ class ExcelWriter(object):
             self.datetime_format = 'YYYY-MM-DD HH:MM:SS'
         else:
             self.datetime_format = datetime_format
+
+        self.mode = mode
 
     def __fspath__(self):
         return _stringify_path(self.path)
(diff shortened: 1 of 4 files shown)


pandas/io/excel.py
```diff
@@ -804,6 +804,8 @@ class ExcelWriter(object):
     datetime_format : string, default None
         Format string for datetime objects written into Excel files
         (e.g. 'YYYY-MM-DD HH:MM:SS')
+    mode : {'w' or 'a'}, default 'w'
+        File mode to use (write or append).
```
jreback: add a versionadded

pandas/io/excel.py
```diff
@@ -993,23 +998,28 @@ class _OpenpyxlWriter(ExcelWriter):
...
+        else:
+            # Create workbook object with default optimized_write=True.
+            self.book = Workbook()
 
-        # Openpyxl 1.6.1 adds a dummy sheet. We remove it.
-        if self.book.worksheets:
-            try:
-                self.book.remove(self.book.worksheets[0])
-            except AttributeError:
+            # Openpyxl 1.6.1 adds a dummy sheet. We remove it.
```
jreback: you can remove this comment (and maybe don't need this anymore), 1.6.1 is lower than supported version

jreback (changes requested):
hmm maybe do for 0.24.0

pandas/io/excel.py
WillAyd: Do you think it's worth adding this class to the API? Would be a little strange as its an ABCMeta but on the flip side may still be beneficial for those looking further at it

pandas/io/excel.py
WillAyd: Unless I'm overlooking it I don't see an explicit min version. Happy to add one - what would that be?

pandas/io/excel.py
gfyoung: > Do you think it's worth adding this class to the API?

Can't you just do `from pandas.io.excel import ExcelWriter` ?

WillAyd: It's not the import mechanism I was referring to as much as the documentation - this class is not part of the doc build

gfyoung: @WillAyd : Ah, thanks for clarifying.  I don't see why not then, but let's see what @jreback has to say about that.

jreback (approved):

jreback:
thanks @WillAyd

WillAyd:
@enterdisplaynamehere this feature is being released as part of 0.24 which we just issued a release candidate for over the past few days. You can try on the RC or here on master and if neither works open an issue per the contributing guide, but this wouldn't be expected to work on versions older than than

yehia123:
Hi,

How come when looking at ExcelWriter documentation on
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.ExcelWriter.html 
``>>> with ExcelWriter('path_to_file.xlsx', mode='a') as writer:"
...     df.to_excel(writer, sheet_name='Sheet3')"

But when I try using the same syntax I receive this error 
"ValueError: Append mode is not supported with xlsxwriter!"

I tried using engine='openpyxl' but it is very slow and sometimes causes the file to become corrupted and not being able to open. Any feedback is appreciated.


Thanks.

WillAyd:
@yehia123 this was only implemented for openpyxl and I'm not sure it's even possible to append with xlsxwriter. If you research and find otherwise certainly can open as an enhancement request on the issue tracker

yehia123:
@WillAyd Thank you for the quick response. I ended up making my file csv instead of xlsx and used to_csv(); the append mode is much faster than 'openpyxl'.

PR #21263, opened 2018-05-30
Color text based on background color when using `_background_gradient()`

### LINKED ISSUE #21258: Automatically adjust HTML table text color when using `style.background_gradient()`
#### Code Sample

```python
import pandas as pd
import numpy as np

np.random.seed(24)
df = pd.DataFrame(np.random.randn(8, 4), columns=list('ABCD'))
df.style.background_gradient()
```
image

#### Problem description

Since all text in the displayed HTML table is black, it is difficult to see values when the background is dark. This is currently possible to adjust manually with the `high` and `low` parameters, but this limits the range of the colormap.

A possible improvement would be to color text white automatically if the background color is too dark (low luminance). This approach is currently [implemented for annotated heatmaps](https://github.com/mwaskom/seaborn/blob/scatterplot/seaborn/matrix.py#L241-L242) in the visualization library `seaborn`, and similar criteria could be used with `pandas` styles ([luminance implementation](https://github.com/mwaskom/seaborn/blob/scatterplot/seaborn/utils.py#L547-L566) and [w3c's official definition](https://www.w3.org/WAI/GL/wiki/Relative_luminance)).

Automatic text coloring could be exposed through a `text_color` parameter to `style.background_gradient()`, which could be either `auto`, or a color value. Another alternative is to let the user adjust the luminance threshold directly, with a sound default value. A third approach would be to create a new method for controlling the text color in general, and call it with specific arguments inside `style.background_gradient()`.

If you think adding this functionality would be useful, I am happy to attempt a PR.

#### Expected Output

luminance threshold 0.408 (`seaborn` default)
screenshot-20180530-112153

luminance threshold 0.2
screenshot-20180530-112028

<details>

INSTALLED VERSIONS
------------------
commit: c85ab083919b59ce84c220d5baf7d34ff4a0bcf2
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.16.8-1-ARCH
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.21.0.dev+2049.gc85ab0839.dirty
pytest: 3.5.1
pip: 9.0.3
setuptools: 39.2.0
Cython: 0.28.2
numpy: 1.14.3
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: 1.7.4
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.2.1
tables: 3.4.3
numexpr: 2.6.5
feather: None
matplotlib: 2.2.2
openpyxl: 2.5.3
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.4
lxml: 4.2.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.7
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

joelostblom:
The purpose of this PR is to automatically color text dark or light based on the background color of the HTML table:

================`old behavior` ===========================`new behavior`========
image image

As described in #21258, I use the luminance-based approach from `seaborn`'s annotated heatmaps. Tagging @WillAyd who commented on that issue. A few comments on this PR

1. Initially, I was not sure if defining the `relative_luminance()` method within `_background_gradient()` was the right way to go, but I opted for this since I saw the same approach elsewhere in the file. Let me know if you prefer a different approach.

2. I am not sure how intuitive it is that a parameter named `text_color` takes a numeric argument and not a color, but I think it is a good name for discoverability. Naming it `luminance_threshold` or similar might be confusing for users looking for a way to change the text color.

3. I opted to make the light text not completely white. The focus should be the background color so the text should not pop out too much. Thoughts?

================`#ffffff` ============================`#f1f1f1` (current choice)===
image image

image image

4. I think 0.2 is a good default threshold value for `text_color` based on my own qualitative assessment, feel free to disagree. The seaborn default of 0.4 makes too much text white in my opinion. Since colors and contrast can be quite subjective, I thought I would include some comparisons.

============`text_color=0.4` ======================`text_color=0.2` (current choice)===
image image
image image
image image

This is my first PR, apologies if I have misunderstood something. 

- [x] closes #21258 
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -181,7 +181,7 @@ Reshaping
 Other
 ^^^^^
 
-- :meth: `~pandas.io.formats.style.Styler.background_gradient` now takes a ``text_color_threshold`` parameter to automatically lighten the text color based on the luminance of the background color. This improves readability with dark background colors without the need to limit the background colormap range. (:issue:`21258`, :issue:`21269`)
+- :meth: `~pandas.io.formats.style.Styler.background_gradient` now takes a ``text_color_threshold`` parameter to automatically lighten the text color based on the luminance of the background color. This improves readability with dark background colors without the need to limit the background colormap range. (:issue:`21258`)
 -
 -
 
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -181,7 +181,7 @@ Reshaping
 Other
 ^^^^^
 
--
+- :meth: `~pandas.io.formats.style.Styler.background_gradient` now takes a ``text_color_threshold`` parameter to automatically lighten the text color based on the luminance of the background color. This improves readability with dark background colors without the need to limit the background colormap range. (:issue:`21258`, :issue:`21269`)
 -
 -
 
--- a/pandas/io/formats/style.py
+++ b/pandas/io/formats/style.py
@@ -863,7 +863,7 @@ class Styler(object):
         return self
 
     def background_gradient(self, cmap='PuBu', low=0, high=0, axis=0,
-                            subset=None):
+                            subset=None, text_color=0.2):
         """
         Color the background in a gradient according to
         the data in each column (optionally row).
@@ -879,6 +879,10 @@ class Styler(object):
             1 or 'columns' for columnwise, 0 or 'index' for rowwise
         subset: IndexSlice
             a valid slice for ``data`` to limit the style application to
+        text_color: float or int
+            luminance threshold for determining text color. Facilitates text
+            visibility across varying background colors. From 0 to 1.
+            0 = all text is dark colored, 1 = all text is light colored.
 
         Returns
         -------
@@ -886,19 +890,19 @@ class Styler(object):
 
         Notes
         -----
-        Tune ``low`` and ``high`` to keep the text legible by
-        not using the entire range of the color map. These extend
-        the range of the data by ``low * (x.max() - x.min())``
-        and ``high * (x.max() - x.min())`` before normalizing.
+        Set ``text_color`` or tune ``low`` and ``high`` to keep the text
+        legible by not using the entire range of the color map. The range of
+        the data is extended by ``low * (x.max() - x.min())`` and ``high *
+        (x.max() - x.min())`` before normalizing.
         """
         subset = _maybe_numeric_slice(self.data, subset)
         subset = _non_reducing_slice(subset)
         self.apply(self._background_gradient, cmap=cmap, subset=subset,
-                   axis=axis, low=low, high=high)
+                   axis=axis, low=low, high=high, text_color=text_color)
         return self
 
     @staticmethod
-    def _background_gradient(s, cmap='PuBu', low=0, high=0):
+    def _background_gradient(s, cmap='PuBu', low=0, high=0, text_color=0.2):
         """Color background in a range according to the data."""
         with _mpl(Styler.background_gradient) as (plt, colors):
             rng = s.max() - s.min()
@@ -909,8 +913,33 @@ class Styler(object):
             # https://github.com/matplotlib/matplotlib/issues/5427
             normed = norm(s.values)
             c = [colors.rgb2hex(x) for x in plt.cm.get_cmap(cmap)(normed)]
-            return ['background-color: {color}'.format(color=color)
-                    for color in c]
+            if (not isinstance(text_color, (float, int)) or
+                    not 0 <= text_color <= 1):
+                msg = "`text_color` must be a value from 0 to 1."
+                raise ValueError(msg)
+
+            def relative_luminance(color):
+                """Calculate the relative luminance of a color according to W3C
+                standards, https://www.w3.org/WAI/GL/wiki/Relative_luminance
+                Parameters
+                ----------
+                color : matplotlib color. Hex code, rgb-tuple, or
+                HTML color name.
+                Returns
+                -------
+                luminance : float between 0 and 1
+                """
+                rgb = colors.colorConverter.to_rgba_array(color)[:, :3]
+                rgb = np.where(rgb <= .03928, rgb / 12.92,
+                               ((rgb + .055) / 1.055) ** 2.4)
+                lum = rgb.dot([.2126, .7152, .0722])
+                return lum.item()
+
+            text_colors = ['#f1f1f1' if relative_luminance(x) < text_color
+                           else '#000000' for x in c]
+
+            return ['background-color: {color};color: {tc}'.format(
+                    color=color, tc=tc) for color, tc in zip(c, text_colors)]
 
     def set_properties(self, subset=None, **kwargs):
         """
--- a/pandas/io/formats/style.py
+++ b/pandas/io/formats/style.py
@@ -863,7 +863,7 @@ class Styler(object):
         return self
 
     def background_gradient(self, cmap='PuBu', low=0, high=0, axis=0,
-                            subset=None, text_color=0.2):
+                            subset=None, text_color=0.408):
         """
         Color the background in a gradient according to
         the data in each column (optionally row).
@@ -902,7 +902,7 @@ class Styler(object):
         return self
 
     @staticmethod
-    def _background_gradient(s, cmap='PuBu', low=0, high=0, text_color=0.2):
+    def _background_gradient(s, cmap='PuBu', low=0, high=0, text_color=0.408):
         """Color background in a range according to the data."""
         with _mpl(Styler.background_gradient) as (plt, colors):
             rng = s.max() - s.min()
--- a/pandas/io/formats/style.py
+++ b/pandas/io/formats/style.py
@@ -863,7 +863,7 @@ class Styler(object):
         return self
 
     def background_gradient(self, cmap='PuBu', low=0, high=0, axis=0,
-                            subset=None, text_color=0.408):
+                            subset=None, text_color_threshold=0.408):
         """
         Color the background in a gradient according to
         the data in each column (optionally row).
@@ -879,7 +879,7 @@ class Styler(object):
             1 or 'columns' for columnwise, 0 or 'index' for rowwise
         subset: IndexSlice
             a valid slice for ``data`` to limit the style application to
-        text_color: float or int
+        text_color_threshold: float or int
             luminance threshold for determining text color. Facilitates text
             visibility across varying background colors. From 0 to 1.
             0 = all text is dark colored, 1 = all text is light colored.
@@ -892,19 +892,21 @@ class Styler(object):
 
         Notes
         -----
-        Set ``text_color`` or tune ``low`` and ``high`` to keep the text
-        legible by not using the entire range of the color map. The range of
-        the data is extended by ``low * (x.max() - x.min())`` and ``high *
+        Set ``text_color_threshold`` or tune ``low`` and ``high`` to keep the
+        text legible by not using the entire range of the color map. The range
+        of the data is extended by ``low * (x.max() - x.min())`` and ``high *
         (x.max() - x.min())`` before normalizing.
         """
         subset = _maybe_numeric_slice(self.data, subset)
         subset = _non_reducing_slice(subset)
         self.apply(self._background_gradient, cmap=cmap, subset=subset,
-                   axis=axis, low=low, high=high, text_color=text_color)
+                   axis=axis, low=low, high=high,
+                   text_color_threshold=text_color_threshold)
         return self
 
     @staticmethod
-    def _background_gradient(s, cmap='PuBu', low=0, high=0, text_color=0.408):
+    def _background_gradient(s, cmap='PuBu', low=0, high=0,
+                             text_color_threshold=0.408):
         """Color background in a range according to the data."""
         with _mpl(Styler.background_gradient) as (plt, colors):
             rng = s.max() - s.min()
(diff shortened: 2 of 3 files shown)


WillAyd (changes requested):
Thanks for the PR. Here are some initial comments

pandas/io/formats/style.py
```diff
@@ -879,26 +879,30 @@ def background_gradient(self, cmap='PuBu', low=0, high=0, axis=0,
             1 or 'columns' for columnwise, 0 or 'index' for rowwise
         subset: IndexSlice
             a valid slice for ``data`` to limit the style application to
+        text_color: float or int
+            luminance threshold for determining text color. Facilitates text
+            visibility across varying background colors. From 0 to 1.
+            0 = all text is dark colored, 1 = all text is light colored.
```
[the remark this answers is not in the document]
jreback: is this the common name for this field in mpl?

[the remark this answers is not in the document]
soxofaan: IMHO `text_color=0.2` looks very counterintuitive to me. It almost looks like the exact opposite of what this feature is about (not having a constant text color).
Shouldn't the name  at least contain "threshold", e.g. `text_color_threshold`?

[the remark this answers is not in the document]
joelostblom: I agree that `text_color_threshold` is a suitable, is it ok with multiple underscores in the parameter name? I don't think there is a common name for this in mpl (for coloring text in general, they tend to use just `color`, but I believe it would be easy to mistake that for the backgruond color here due to the method's name)

pandas/io/formats/style.py
```diff
@@ -909,8 +913,33 @@ def _background_gradient(s, cmap='PuBu', low=0, high=0):
             # https://github.com/matplotlib/matplotlib/issues/5427
             normed = norm(s.values)
             c = [colors.rgb2hex(x) for x in plt.cm.get_cmap(cmap)(normed)]
-            return ['background-color: {color}'.format(color=color)
-                    for color in c]
+            if (not isinstance(text_color, (float, int)) or
+                    not 0 <= text_color <= 1):
+                msg = "`text_color` must be a value from 0 to 1."
+                raise ValueError(msg)
```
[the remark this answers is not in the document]
joelostblom: I am having troubles getting the test correct for this. When I try the function manually, it raises `ValueError` when called with any one of the parameters in the test, but pytest keeps failing saying that it doesn't raise a `ValueError`. Would you have time to check my latest commits and advise?

pandas/tests/io/formats/test_style.py
```diff
@@ -1031,7 +1031,9 @@ def test_background_gradient(self):
 
         result = df.style.background_gradient(
             subset=pd.IndexSlice[1, 'A'])._compute().ctx
-        assert result[(1, 0)] == ['background-color: #fff7fb']
+
+        assert result[(1, 0)] == ['background-color: #fff7fb',
```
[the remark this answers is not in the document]
joelostblom: Added a suggestion

pandas/tests/io/formats/test_style.py
```diff
@@ -1028,10 +1028,25 @@ def test_background_gradient(self):
             assert all("#" in x[0] for x in result.values())
             assert result[(0, 0)] == result[(0, 1)]
             assert result[(1, 0)] == result[(1, 1)]
+            for res in result:
```
WillAyd: Make this a separate test called `test_text_color_threshold` to distinguish it from the gradient testing. 

Also I didn't really understand the point of the conditional - can we not be more assertive about the *exact* values we'd expect?

pandas/tests/io/formats/test_style.py
```diff
@@ -1028,10 +1028,25 @@ def test_background_gradient(self):
...
 
         result = df.style.background_gradient(
             subset=pd.IndexSlice[1, 'A'])._compute().ctx
-        assert result[(1, 0)] == ['background-color: #fff7fb']
+
+        assert result[(1, 0)] == ['background-color: #fff7fb',
+                                  'color: #000000']
+
+    @td.skip_if_no_mpl
+    def test_text_color_threshold(self):
```
WillAyd: Rename this to `test_text_color_threshold_raises`

pandas/tests/io/formats/test_style.py
```diff
@@ -1028,10 +1028,25 @@ def test_background_gradient(self):
...
+        df = pd.DataFrame([[1, 2], [2, 4]], columns=['A', 'B'])
+        for text_color_threshold in [1.1, '1', -1, [2, 2]]:
```
WillAyd: You can parametrize these (check `test_to_html.py` for examples if you don't know what I mean)

pandas/tests/io/formats/test_style.py
```diff
@@ -1028,10 +1028,25 @@ def test_background_gradient(self):
...
+            with pytest.raises(ValueError):
```
WillAyd: `tm.assert_raises_regex` would be preferable as you can use it to assert on the message as well. Can still be used as a context manager (can also see usage in `test_to_html.py`)

pandas/io/formats/style.py
```diff
@@ -863,7 +863,7 @@ def highlight_null(self, null_color='red'):
         return self
 
     def background_gradient(self, cmap='PuBu', low=0, high=0, axis=0,
-                            subset=None):
+                            subset=None, text_color_threshold=0.408):
         """
```
WillAyd: Can you also add a Raises section for the docstring?

WillAyd (changes requested):

pandas/tests/io/formats/test_style.py
joelostblom: Thanks for the pointers, I updated accordingly.

This test still complains that a `ValueError` is not raised for any of the parameters. When I call the the function outside the test, it does raise `ValueError: ('`text_color_threshold` must be a value from 0 to 1.', 'occurred at index A')`. The test works fine if I raise `ValueError(msg)` manually.

pandas/tests/io/formats/test_style.py
joelostblom: My thinking is that we want to test if the text color is light or dark *conditional on* the background color. If for some unexpected reason the background color changes, this test should not fail.

I added a section to assert if the background color takes on one of its 4 expected values to prevent that `test_text_color_threshold` doesn't test anything (I could move this section to `test_background_gradient` if you think that is more suitable).

pandas/tests/io/formats/test_style.py
WillAyd: This didn't raise for me locally, though it did when running the `_compute` method

```python
In [12]: df.style.background_gradient(text_color_threshold='1')
Out[12]: <pandas.io.formats.style.Styler at 0x110163630>

In [13]: df.style.background_gradient(text_color_threshold='1')._compute()
ValueError: ('`text_color_threshold` must be a value from 0 to 1.', 'occurred at index A')
```

pandas/tests/io/formats/test_style.py
```diff
@@ -1031,7 +1031,40 @@ def test_background_gradient(self):
...
             subset=pd.IndexSlice[1, 'A'])._compute().ctx
-        assert result[(1, 0)] == ['background-color: #fff7fb']
+
+        assert result[(1, 0)] == ['background-color: #fff7fb',
+                                  'color: #000000']
+
+    @td.skip_if_no_mpl
+    def test_text_color_threshold(self):
+        df = pd.DataFrame([[1, 2], [2, 4]], columns=['A', 'B'])
+        for c_map in [None, 'YlOrRd']:
```
WillAyd: Parametrize the `cmap`

pandas/tests/io/formats/test_style.py
joelostblom: Thanks, I missed that `_compute()` is called automatically in Notebooks via `_repr_html_`. I believe all comments are addressed now.

pandas/tests/io/formats/test_style.py
```diff
@@ -1031,7 +1031,40 @@ def test_background_gradient(self):
...
+            result = df.style.background_gradient(cmap=c_map)._compute().ctx
+            for res in result:
```
WillAyd: Instead of the loop just be explicit about the dict that you expect and compare it to the result

WillAyd (changes requested):

pandas/tests/io/formats/test_style.py
```diff
@@ -1036,26 +1036,24 @@ def test_background_gradient(self):
...
-                    assert text_color == '#000000'
-                elif bg_color in ['#800026', '#440154']:
-                    assert text_color == '#f1f1f1'
-            for res in result:
-                if result[res][0].split(' ')[1] in ['#fde725', '#ffffcc']:
-                    assert result[(res)][1].split(' ')[1] == '#000000'
-                elif result[res][0].split(' ')[1] in ['#800026', '#440154']:
-                    assert result[(res)][1].split(' ')[1] == '#f1f1f1'
+        result = df.style.background_gradient(cmap=c_map)._compute().ctx
+        test_colors = {None: {(0, 0): ('#440154', '#f1f1f1'),
```
WillAyd: Change the variable here to `expected` (standard in pandas tests)

WillAyd: Instead of using a nested dict you should also send this in via parametrization, so you would have `"c_map,expected"` and then send in a tuple of the `c_map` and it's expected result

pandas/tests/io/formats/test_style.py
```diff
@@ -1036,26 +1036,24 @@ def test_background_gradient(self):
...
+                              (1, 0): ('#fde725', '#000000')},
+                       'YlOrRd': {(0, 0): ('#ffffcc', '#000000'),
+                                  (1, 0): ('#800026', '#f1f1f1')}}
+        # Light text on dark background
+        assert result[0, 0][0].split(' ')[1] == test_colors[c_map][0, 0][0], (
```
WillAyd: Why are you splitting this? Just make your `expected` variable account for the exact string required. Unless I'm missing something you should just have one assertion here for `result == expected`

WillAyd (changes requested):

WillAyd (changes requested):

pandas/tests/io/formats/test_style.py
joelostblom: I was doing this to be able to raise different assertion messages for the background and foreground color. But since pytest shows the expected and current value anyways, this should be easy to trace done without separate assertion messages.

pandas/tests/io/formats/test_style.py
```diff
@@ -1031,7 +1031,33 @@ def test_background_gradient(self):
 
         result = df.style.background_gradient(
             subset=pd.IndexSlice[1, 'A'])._compute().ctx
-        assert result[(1, 0)] == ['background-color: #fff7fb']
+
+        assert result[(1, 0)] == ['background-color: #fff7fb',
+                                  'color: #000000']
+
+    @td.skip_if_no_mpl
```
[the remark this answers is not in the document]
joelostblom: Yes, they both call `background_gradient()` [which depends on mpl](https://github.com/pandas-dev/pandas/blob/master/pandas/io/formats/style.py#L903).

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -181,7 +181,7 @@ Reshaping
 Other
 ^^^^^
 
--
+- :meth: `~pandas.io.formats.style.Styler.background_gradient` now takes a ``text_color_threshold`` parameter to automatically lighten the text color based on the luminance of the background color. This improves readability with dark background colors without the need to limit the background colormap range. (:issue:`21258`, :issue:`21269`)
```
WillAyd: Can just reference the first issue here, since that is what you are closing anyway (second was a duplicate)

WillAyd:
Thanks for the changes. Assuming tests pass I'll approve on my end

joelostblom:
@WillAyd Thank you for all the help, much appreciated!

WillAyd (approved):

gfyoung:
@TomAugspurger : Could you take a look?

pandas/io/formats/style.py
```diff
@@ -879,26 +879,39 @@ def background_gradient(self, cmap='PuBu', low=0, high=0, axis=0,
             1 or 'columns' for columnwise, 0 or 'index' for rowwise
         subset: IndexSlice
             a valid slice for ``data`` to limit the style application to
+        text_color_threshold: float or int
```
TomAugspurger: Add `, default 0.408`.

And may a note as to why that's the default?

pandas/io/formats/style.py
```diff
@@ -879,26 +879,39 @@ def background_gradient(self, cmap='PuBu', low=0, high=0, axis=0,
...
-        and ``high * (x.max() - x.min())`` before normalizing.
+        Set ``text_color_threshold`` or tune ``low`` and ``high`` to keep the
+        text legible by not using the entire range of the color map. The range
+        of the data is extended by ``low * (x.max() - x.min())`` and ``high *
+        (x.max() - x.min())`` before normalizing.
+
+        Raises
+        ------
+        ValueError
+            If ``text_color_threshold`` is not a value from 0 to 1.
```
TomAugspurger: Single backtick for parameter names.

TomAugspurger (approved):

pandas/io/formats/style.py
joelostblom: Thanks, I did double tick because I saw it in other places, for example for `high` and `low` in the notes section. Do you want me to update them to single tick as well? Does the same go for `data` in the subset section and the expressions in the note section?

pandas/io/formats/style.py
joelostblom: We chose `0.408` to stay consistent with the Seaborn implementation. Should I ask the Seaborn author for the underlying reason, just reference seaborn, or leave the value without explanation?

TomAugspurger: No worries, was just curious.

pandas/io/formats/style.py
TomAugspurger: Yes, I noticed that after posting. Probably fine to be consistent w/ the rest of the docstring here.

TomAugspurger:
Thanks @joelostblom!

PR #21264, opened 2018-05-30
DOC: Fix renaming categories section

JarnoRFB:
The categories were not correctly saved so that the renaming using the
dict-like object had no effect.


WillAyd (approved):
In the future would be better to open an issue first, but lgtm

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

@WillAyd I think in this case for such a minor doc fix it is fine to directly open a PR (we mainly want to avoid people doing bug fixes which might not be fixes or new features we might not want to include)

PR #21273, opened 2018-05-31
Fix an attribute error in __init__.py

Rory101:
This fixes a problem, where a bool is treated like a string.

### DIFF
--- a/pandas/__init__.py
+++ b/pandas/__init__.py
@@ -67,7 +67,7 @@ json = _DeprecatedModule(deprmod='pandas.json',
 parser = _DeprecatedModule(deprmod='pandas.parser',
                            removals=['na_values'],
                            moved={'CParserError': 'pandas.errors.ParserError'})
-lib = _DeprecatedModule(deprmod='pandas.lib', deprmodto=False,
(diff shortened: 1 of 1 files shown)


WillAyd:
What problem is this fixing? Can you post the related issue?

jorisvandenbossche:
@Rory101 Can you provide some context as asked before?

Rory101:
I am running Python 2.7.14 on Windows 10, 64bit. 
In `pandas/util/_depr_module.py`  Line 59. The following function is called. `self._import_deprmod(self.deprmodto)`. In that function, `deprmodule = importlib.import_module(mod)`, this line of code is executed and function requires the `name` argument to be a string, not a boolean. 

I see now that my fix will not work either, because you rely on the boolean value to determine if the module still has to be deprecated or if it has already been deprecated.

Sorry for taking so long to get back to you.

jorisvandenbossche:
But when do you get this error? (and what error? can you post the full error traceback) 
From just importing pandas? Can you post a code snippet that for you gives this error?

jreback:
closing. this needs more info and probably an issue to track before a fix is put.

PR #21274, opened 2018-05-31
PERF: improve performance of NDFrame.describe

DataOmbudsman:
A one-line change that enables to calculate the percentiles in `describe` more efficiently. The point is that calculating percentiles in one pass is faster than separately.

`describe` (with default `percentiles` argument) becomes 25-30% faster than before for numerical Series and DataFrames.

### Setup 
```
import timeit

setup = '''
import numpy as np
import pandas as pd
np.random.seed(123)
s = pd.Series(np.random.randint(0, 100, 1000000))
'''
```

### Benchmark
```
min(timeit.Timer('s.describe()', setup=setup).repeat(100, 1))
```

### Results
On master:
```
0.06349272100487724
```

With this change:
```
0.04745814300258644
```

Results are similar for DataFrames.

- [ ] closes #xxxx
- [x] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/asv_bench/benchmarks/frame_methods.py
+++ b/asv_bench/benchmarks/frame_methods.py
@@ -519,7 +519,6 @@ class Describe(object):
     goal_time = 0.2
 
     def setup(self):
-        np.random.seed(123)
         self.df = DataFrame({
             'a': np.random.randint(0, 100, int(1e6)),
             'b': np.random.randint(0, 100, int(1e6)),
--- a/asv_bench/benchmarks/frame_methods.py
+++ b/asv_bench/benchmarks/frame_methods.py
@@ -513,6 +513,7 @@ class NSort(object):
     def time_nsmallest(self, keep):
         self.df.nsmallest(100, 'A', keep=keep)
 
+
 class Describe(object):
 
     goal_time = 0.2
--- a/asv_bench/benchmarks/frame_methods.py
+++ b/asv_bench/benchmarks/frame_methods.py
@@ -512,3 +512,21 @@ class NSort(object):
 
     def time_nsmallest(self, keep):
         self.df.nsmallest(100, 'A', keep=keep)
+
+class Describe(object):
+
+    goal_time = 0.2
+
+    def setup(self):
+        np.random.seed(123)
+        self.df = DataFrame({
+            'a': np.random.randint(0, 100, int(1e6)),
+            'b': np.random.randint(0, 100, int(1e6)),
+            'c': np.random.randint(0, 100, int(1e6))
+        })
+
+    def time_series_describe(self):
+        self.df['a'].describe()
+
+    def time_dataframe_describe(self):
+        self.df.describe()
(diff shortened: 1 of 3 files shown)


WillAyd:
Typically for performance-related changes we look for an ASV to measure and track over time. Can you add one to `asv_bench/benchmarks/frame_methods.py` and post the results of the benchmark here?

DataOmbudsman:
Sure. Thanks for the suggestion. Here are my ASV benchmarks. These also show the improvement.

### Setup

```
class Describe(object):

    goal_time = 0.2

    def setup(self):
        np.random.seed(123)
        self.df = DataFrame({
            'a': np.random.randint(0, 100, int(1e6)),
            'b': np.random.randint(0, 100, int(1e6)),
            'c': np.random.randint(0, 100, int(1e6)),
        })

    def time_series_describe(self):
        self.df['a'].describe()

    def time_dataframe_describe(self):
        self.df.describe()
```
### Results
|before|after|ratio|    |
|---|---|---|---|
|689±10ms|495±6ms|0.72|frame_methods.Describe.time_dataframe_describe|
|234±9ms|166±6ms|0.71|frame_methods.Describe.time_series_describe|

WillAyd:
OK thanks. Can you update your PR to include the benchmark and a whatsnew note for 0.24?

jorisvandenbossche (approved):

WillAyd (approved):

asv_bench/benchmarks/frame_methods.py
```diff
@@ -512,3 +512,22 @@ def time_nlargest(self, keep):
...
     def time_nsmallest(self, keep):
         self.df.nsmallest(100, 'A', keep=keep)
+
+
+class Describe(object):
+
+    goal_time = 0.2
+
+    def setup(self):
+        np.random.seed(123)
```
mroeschke: You can remove the random seed; this is handled when `setup` is imported at the top (`from .pandas_vb_common import setup`)

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -63,8 +63,7 @@ Removal of prior version deprecations/changes
 Performance Improvements
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
--
--
+- Improved performance of :func:`Series.describe` in case of numeric dtpyes
```
jreback: can you add the issue number (this pr number as we don't have an issue)

jreback (changes requested):

doc/source/whatsnew/v0.24.0.txt
DataOmbudsman: OK but I'm unsure about what format is expected. Do you think a link to an external URL (such as [here](https://github.com/pandas-dev/pandas/blob/67e6e6fcd19d1d89cb60abc3a78372bc85fd8e29/doc/source/whatsnew/v0.23.0.txt#L534)) would be appropriate? E.g., `` `pull request #21274 <https://github.com/pandas-dev/pandas/pull/21274/>`_ ``. Or something else?

jreback: same format as all the others, just use ``:issue:`number` ``

DataOmbudsman: I see now that the URL of the issue is translated to the URL of the PR. That's great.

jorisvandenbossche:
@DataOmbudsman Thanks!

PR #21279, opened 2018-05-31
BUG: Fix encoding for Stata format 118 files

### LINKED ISSUE #21244: BUG: `read_stata` always uses 'utf8'

bashtage:
Ensure that Stata 118 files always use utf-8 encoding

- [x] closes #21244 
- [x] tests added / passed
- [x] whatsnew entry


doc/source/whatsnew/v0.23.1.txt
```diff
@@ -23,7 +23,7 @@ New features
 Deprecations
 ~~~~~~~~~~~~
 
--
+- :func:`read_stata` and :class:`StataReader` have deprecated the ``encoding`` parameter. Stata files only support a single encoding and so this input has no effect. (:issue:`21244`)
```
jorisvandenbossche: I know you say it didn't have any effect before, but still, we should not introduce deprecation warnings in a bug fix release. So would either move this to 0.24.0 or split the PR

bashtage:
I removed the deprecation.

jreback:
@bashtage so the fix looks good. can your restore the encoding paramater then can merge into 0.23.1 (and then come back with another PR to deprecate it)? thxs

bashtage:
I removed the deprecation warning but encoding is now a no-op, which is the correct behavior.  

I'd rather not restore the (potentially) broken behavior where users are allowed to set an invalid encoding (e.g. latin-1 for more recent file formats), and this can wait to 24 if you think this is too much change for a bug fix release.

jorisvandenbossche:
> I'd rather not restore the (potentially) broken behavior where users are allowed to set an invalid encoding (e.g. latin-1 for more recent file formats), and this can wait to 24 if you think this is too much change for a bug fix release.

Yes, we should not remove the keyword for 0.23.1, even when it was a no-op, because that breaks people code who inadvertently used it. So I would add it back here, but then a next PR to deprecate it if you want.

bashtage:
The keyword is still available in read_Stata and Stata Reader and so code
won't break.

jorisvandenbossche:
Ah, sorry, missed that you only removed the internal passing through of the keyword. That's fine then!

jorisvandenbossche (approved):

jorisvandenbossche:
@bashtage Thanks!

Issue #21267, opened 2018-05-31
BUG: DatetimeIndex.time incorrect in case of timezones 


ssikdar1:
An offshoot of https://github.com/pandas-dev/pandas/issues/21230

http://pandas.pydata.org/pandas-docs/stable/timeseries.html?highlight=weekofmonth#time-date-components

From the docs:
There are several time/date properties that one can access from Timestamp or a collection of timestamps like a DatetimeIndex.
...
time | Returns datetime.time (does not contain timezone information)

But take this code here:

```
import pandas as pd

df1 = pd.DataFrame(data=[24, 25],
                   index=pd.DatetimeIndex(['2013-01-24 15:01:00+01:00',
                                           '2013-01-25 15:01:00+01:00'],
                                          dtype='datetime64[ns, CET]',
                                          name='Date', freq=None))

print(type(df1.index))
print("df1.index.time")
print(df1.index.time)
                 
```

returns:
```
<class 'pandas.core.indexes.datetimes.DatetimeIndex'>
df1.index.time
[datetime.time(14, 1, tzinfo=<DstTzInfo 'CET' CET+1:00:00 STD>)
 datetime.time(14, 1, tzinfo=<DstTzInfo 'CET' CET+1:00:00 STD>)]


```

- This is showing timezone information.

This may not be a bad thing, but if tz information is expected the docs should probably be changed?

[paste the output of ``pd.show_versions()`` here below this line]

>>> pd.show_versions()

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Linux
OS-release: 4.13.0-43-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.23.0
pytest: None
pip: 9.0.1
setuptools: 28.8.0
Cython: None
numpy: 1.14.3
scipy: None
pyarrow: None
xarray: None
IPython: None
sphinx: None
patsy: None
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: None
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

[...]

Issue #15938, opened 2017-04-07
BUG: invalid construction from repr of dt-aware index


jreback:
thought we had an issue about this....

The data is converted from string and is tz-aware, then the dtype convert is *also* applied.
```
In [1]: dt = pd.date_range('20160101', periods=1, tz='US/Eastern')

In [2]: dt
Out[2]: DatetimeIndex(['2016-01-01 00:00:00-05:00'], dtype='datetime64[ns, US/Eastern]', freq='D')

In [3]: r = DatetimeIndex(['2016-01-01 00:00:00-05:00'], dtype='datetime64[ns, US/Eastern]', freq='D')

In [4]: r
Out[4]: DatetimeIndex(['2016-01-01 05:00:00-05:00'], dtype='datetime64[ns, US/Eastern]', freq='D')

In [5]: dt.equals(r)
Out[5]: False
```

mroeschke:
After #21822 this now raises

```
In [3]: r = DatetimeIndex(['2016-01-01 00:00:00-05:00'], dtype='datetime64[ns, US/Eastern]', freq='D')
TypeError: data is already tz-aware pytz.FixedOffset(-300), unable to set specified tz: US/Eastern
```

Which I argue is correct since the user is attempting to localize the datetime with two methods

* dtype
* utcoffset

PR #21281, opened 2018-06-01
BUG: Using DatetimeIndex.date with timezone returns incorrect date #2…

### LINKED ISSUE #21230: Datetime.index.date returns incorrect date post upgrade to version 0.23

```python
df1 = pd.DataFrame(data=[24, 25],
                   index=pd.DatetimeIndex(['2013-01-24 15:01:00+01:00',
                                           '2013-01-25 15:01:00+01:00'],
                                          dtype='datetime64[ns, CET]',
                                          name='Date', freq=None))
print(df1.index.date)
```
the above code prints:
[datetime.date(2013, 1, 23) datetime.date(2013, 1, 24)]

#### Problem description
Datetime.index.date does return incorrect dates. The behavious worked until version 0.22, and seems to be incorrect post upgrade to version 0.23

#### Expected Output
`[datetime.date(2013, 1, 24) datetime.date(2013, 1, 25)]`

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None
pandas: 0.23.0
pytest: 3.5.1
pip: 10.0.1
setuptools: 39.1.0
Cython: None
numpy: 1.12.1
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: 1.7.4
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 2.2.2
openpyxl: None
xlrd: 1.1.0
xlwt: None
xlsxwriter: None
lxml: None
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

jamestran201:
- [ ] closes #21230, closes https://github.com/pandas-dev/pandas/issues/21267
- [x] tests added / passed
- [x] whatsnew entry

Added 2 tests for DatetimeIndex.date in tests/indexes/datetimes/test_datetime.py. Please let me know if I should move them somewhere else.


pandas/core/indexes/datetimes.py
```diff
@@ -2040,7 +2040,7 @@ def date(self):
         Returns numpy array of python datetime.date objects (namely, the date
         part of Timestamps without timezone information).
         """
-        return libts.ints_to_pydatetime(self.normalize().asi8, box="date")
+        return libts.ints_to_pydatetime(self.asi8, box="date")
```
mroeschke: I _think_ this needs to be `self._local_timestamps()` instead of `self.asi8` since those are essentially epoch (UTC) timestamps and we want to return local values.

pandas/tests/indexes/datetimes/test_datetime.py
```diff
@@ -27,6 +27,21 @@ def test_roundtrip_pickle_with_tz(self):
         unpickled = tm.round_trip_pickle(index)
         tm.assert_index_equal(index, unpickled)
 
+    def test_date_accessor_with_tz(self):
```
mroeschke: Instead of 2 separate tests, you can have one test `pytest.mark.parameterize` over the timezone aware and naive data. (e.g. `dtype='datetime64[ns, CET]''` and dtype=`None`)

pandas/core/indexes/datetimes.py
tmnhat2001: Thanks for the suggestion. That might be it.

pandas/core/indexes/datetimes.py
```diff
@@ -2040,7 +2040,11 @@ def date(self):
         Returns numpy array of python datetime.date objects (namely, the date
         part of Timestamps without timezone information).
         """
-        return libts.ints_to_pydatetime(self.normalize().asi8, box="date")
+        if (self.tz is None):
+            return libts.ints_to_pydatetime(self.normalize().asi8, box="date")
```
mroeschke: Do we need the `.normalize()` calls here? I think `ints_to_pydatetime` will strip the time information away when constructing the date

pandas/tests/indexes/datetimes/test_datetime.py
```diff
@@ -27,6 +27,18 @@ def test_roundtrip_pickle_with_tz(self):
...
 
+    @pytest.mark.parametrize("test_input", [
+        DatetimeIndex(['2013-01-24 15:01:00']),
+        DatetimeIndex(['2013-01-24 15:01:00'], dtype='datetime64[ns, CET]'),
+        DatetimeIndex(['2013-01-24 15:01:00'], dtype='datetime64[ns, EST]'),
+        DatetimeIndex(['2013-01-24 15:01:00'], dtype='datetime64[ns, UTC]')
+    ])
+    def test_date_accessor(self, test_input):
+        # GH 21230
+        from datetime import date
```
mroeschke: Looks like this is imported at the top already

pandas/tests/indexes/datetimes/test_datetime.py
```diff
@@ -27,6 +27,18 @@ def test_roundtrip_pickle_with_tz(self):
...
+
+        assert test_input.date == np.array(date(2013, 1, 24))
```
mroeschke: Use `tm.assert_numpy_array_equal` here.

pandas/tests/indexes/datetimes/test_datetime.py
```diff
@@ -27,6 +27,18 @@ def test_roundtrip_pickle_with_tz(self):
         unpickled = tm.round_trip_pickle(index)
         tm.assert_index_equal(index, unpickled)
 
+    @pytest.mark.parametrize("test_input", [
+        DatetimeIndex(['2013-01-24 15:01:00']),
```
mroeschke: Instead, you could leave the `DatetimeIndex` construction in the test, and parametrize over the dtypes and pass that into the constructor.

jorisvandenbossche (commented):
Thanks for the PR!

Can you do a similar change for `time` attribute (just above `date` in the code) ?

jorisvandenbossche:
> Can you do a similar change for time attribute (just above date in the code) ?

Then it would also close https://github.com/pandas-dev/pandas/issues/21267

tmnhat2001:
Sure. There's some discussion in #21267 about whether `time` should return a time with the timezone in it. Could you tell me which is the correct behavior?

jorisvandenbossche:
I think converting back to the 0.22 behaviour is the best for now (which I think is no timezone in the output?) We can still later discuss if we want to add the timezone or not.

mroeschke:
@jorisvandenbossche regarding my comment (https://github.com/pandas-dev/pandas/issues/21267#issuecomment-394060416), it looks like the `time` behavior looks correct but there was some confusion stemming from the construction of the DatetimeIndex (#15938).

pandas/core/indexes/datetimes.py
```diff
@@ -2040,7 +2040,11 @@ def date(self):
         Returns numpy array of python datetime.date objects (namely, the date
         part of Timestamps without timezone information).
         """
-        return libts.ints_to_pydatetime(self.normalize().asi8, box="date")
+        if (self.tz is None):
```
jreback: prob need
``if self.tz is not None and self.tz is not utc``
also write it more like
```
# add a comment here about tz conversions
if self.tz is not None and self.tz is not utc:
    stamps = self.asi8
else:
    stamps = self._local_timestamps()
return libts.init_to_pytdatetime(stamps, box="date")
```

pandas/tests/indexes/datetimes/test_datetime.py
```diff
@@ -27,6 +27,17 @@ def test_roundtrip_pickle_with_tz(self):
...
 
+    @pytest.mark.parametrize("dtype", [
+        None, 'datetime64[ns, CET]',
+        'datetime64[ns, EST]', 'datetime64[ns, UTC]'
+    ])
+    def test_date_accessor(self, dtype):
+        # GH 21230
+
+        index = DatetimeIndex(['2013-01-24 15:01:00'], dtype=dtype)
+        tm.assert_numpy_array_equal(index.date,
```
jreback: use 
```
result = 
expected = 
```

pandas/tests/indexes/datetimes/test_datetime.py
```diff
@@ -27,6 +27,17 @@ def test_roundtrip_pickle_with_tz(self):
         unpickled = tm.round_trip_pickle(index)
         tm.assert_index_equal(index, unpickled)
 
+    @pytest.mark.parametrize("dtype", [
```
jreback: this needs to go in ``test_timezones`` (same subdir)

pandas/tests/indexes/datetimes/test_datetime.py
jreback: can you add another date that is not ambiguous (e.g. its on the same date regardless of tz)
can you add a NaT as well

pandas/tests/indexes/datetimes/test_datetime.py
jreback: place it near other tests that look at .date

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -85,7 +85,7 @@ Indexing
 - Bug in :meth:`Series.reset_index` where appropriate error was not raised with an invalid level name (:issue:`20925`)
 - Bug in :func:`interval_range` when ``start``/``periods`` or ``end``/``periods`` are specified with float ``start`` or ``end`` (:issue:`21161`)
 - Bug in :meth:`MultiIndex.set_names` where error raised for a ``MultiIndex`` with ``nlevels == 1`` (:issue:`21149`)
--
+- Bug in :attr:`DatetimeIndex.date` where an incorrect date is returned when the input date has a timezone (:issue:`21230`)
```
jreback: this should be non-UTC timezone

jreback (changes requested):

tmnhat2001:
@jorisvandenbossche @mroeschke @jreback For #21267, it seems that whether the timezone should be returned still needs to be discussed. Is it ok if I change the behavior of `time` to match v0.22 for now (i.e: `time` does not return a Time with timezone information) since that was also an unintended change?

jreback:
wouldn’t change this as it’s correct
pls update the doc string

pandas/tests/indexes/datetimes/test_timezones.py
```diff
@@ -706,6 +706,19 @@ def test_join_utc_convert(self, join_type):
...
+    @pytest.mark.parametrize("dtype", [
+        None, 'datetime64[ns, CET]',
+        'datetime64[ns, EST]', 'datetime64[ns, UTC]'
+    ])
+    def test_date_accessor(self, dtype):
+        # Regression test for GH#21230
+        expected = np.array([date(2018, 6, 4), pd.NaT], ndmin=1)
+
+        index = DatetimeIndex(['2018-06-04 10:00:00', pd.NaT], dtype=dtype)
+        result = index.date
```
jreback: can u also test .time here

jreback (changes requested):

jorisvandenbossche:
> wouldn’t change this as it’s correct

No, adding the timezone was unintended, and is inconsistent with how Timestamp itself works, and inconsistent with standard library (they both return naive local time). 
So I would just change it back to how it was.

jorisvandenbossche:
@jreback sorry, missed the update of your comment. So yes, then we agree!

pandas/tests/indexes/datetimes/test_timezones.py
```diff
@@ -706,6 +706,32 @@ def test_join_utc_convert(self, join_type):
         assert isinstance(result, DatetimeIndex)
         assert result.tz.zone == 'UTC'
 
+    @pytest.mark.parametrize("dtype", [
+        None, 'datetime64[ns, CET]',
+        'datetime64[ns, EST]', 'datetime64[ns, UTC]'
+    ])
+    def test_date_accessor(self, dtype):
+        # Regression test for GH#21230
+        expected = np.array([date(2018, 6, 4), pd.NaT], ndmin=1)
```
jorisvandenbossche: is the `ndmin=1` needed here?

jorisvandenbossche (approved):
minor comment, looks good to me, thanks!

pandas/tests/indexes/datetimes/test_timezones.py
tmnhat2001: I needed it to avoid a shape of (1,) when creating a numpy array with only 1 element. I'll remove it now.

tmnhat2001:
Is there any way to trigger a re-build on CircleCI? That build failed with this error:
```
CondaHTTPError: HTTP 522 ORIGIN CONNECTION TIME-OUT for url <https://repo.anaconda.com/pkgs/main/linux-64/libpq-10.3-h1ad7b7a_0.tar.bz2>
Elapsed: 00:42.034156
CF-RAY: 426f9db31db895a4-IAD

An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.
```

jorisvandenbossche:
@tmnhat2001 Thanks a lot!

tmnhat2001:
Thanks everyone.

PR #13441, opened 2016-06-14
BUG: Rolling negative window issue fix #13383

priyankjain:
- [x] closes #13383 
- [x] tests added / passed
- [x] passes `git diff upstream/master | flake8 --diff`
- [x] whatsnew entry

Added functionality in validate function of Rolling to ensure that window size is non-negative.

PR #21291, opened 2018-06-02
BUG: invalid rolling window on empty input

### LINKED ISSUE #21286: rolling on 0 window causes core dump

pd.__version__ == 0.22.0

```python
import pandas as pd
series = pd.Series(range(1000))
series.rolling(0).min()
```
#### Problem description

Attempting to perform an operation on a series.rolling with window=0 

kills the kernel if running with jupyter. 

causes a core dump if running directly from python.  

 free(): invalid next size (fast): 0x00007fffbfcc5040 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x70bfb)[0x7fe14e0b0bfb]
/lib/x86_64-linux-gnu/libc.so.6(+0x76fc6)[0x7fe14e0b6fc6]
...

#### Expected Output
Should raise a ValueError instead of crashing.

uds5501:
Added a raise Value Error for the case when window==0, Kindly refer to issue #21286 for the same

- [x] closes #21286 
- [x] tests added / passed
- [x] A simple except cases added

### DIFF
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -52,7 +52,8 @@ Groupby/Resample/Rolling
 - Bug in :func:`DataFrame.agg` where applying multiple aggregation functions to a :class:`DataFrame` with duplicated column names would cause a stack overflow (:issue:`21063`)
 - Bug in :func:`pandas.core.groupby.GroupBy.ffill` and :func:`pandas.core.groupby.GroupBy.bfill` where the fill within a grouping would not always be applied as intended due to the implementations' use of a non-stable sort (:issue:`21207`)
 - Bug in :func:`pandas.core.groupby.GroupBy.rank` where results did not scale to 100% when specifying ``method='dense'`` and ``pct=True``
--Bug in :func:`pandas.DataFrame.rolling` and :func:`pandas.Series.rolling` which incorrectly accepted a 0 window size rather than raising (:issue:`21286`)
+- Bug in :func:`pandas.DataFrame.rolling` and :func:`pandas.Series.rolling` which incorrectly accepted a 0 window size rather than raising (:issue:`21286`)
+
 Strings
 ^^^^^^^
 
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -52,6 +52,7 @@ Groupby/Resample/Rolling
 - Bug in :func:`DataFrame.agg` where applying multiple aggregation functions to a :class:`DataFrame` with duplicated column names would cause a stack overflow (:issue:`21063`)
 - Bug in :func:`pandas.core.groupby.GroupBy.ffill` and :func:`pandas.core.groupby.GroupBy.bfill` where the fill within a grouping would not always be applied as intended due to the implementations' use of a non-stable sort (:issue:`21207`)
 - Bug in :func:`pandas.core.groupby.GroupBy.rank` where results did not scale to 100% when specifying ``method='dense'`` and ``pct=True``
+- Bug in :func:`pandas.core.windows.Windows.validate` where windows parameter was accpeting a numeric value of 0 (:issue:`21286`) 
 
 Strings
 ^^^^^^^
--- a/doc/source/whatsnew/v0.23.1.txt
+++ b/doc/source/whatsnew/v0.23.1.txt
@@ -52,8 +52,7 @@ Groupby/Resample/Rolling
 - Bug in :func:`DataFrame.agg` where applying multiple aggregation functions to a :class:`DataFrame` with duplicated column names would cause a stack overflow (:issue:`21063`)
 - Bug in :func:`pandas.core.groupby.GroupBy.ffill` and :func:`pandas.core.groupby.GroupBy.bfill` where the fill within a grouping would not always be applied as intended due to the implementations' use of a non-stable sort (:issue:`21207`)
 - Bug in :func:`pandas.core.groupby.GroupBy.rank` where results did not scale to 100% when specifying ``method='dense'`` and ``pct=True``
-- Bug in :func:`pandas.core.windows.Windows.validate` where windows parameter was accpeting a numeric value of 0 (:issue:`21286`) 
-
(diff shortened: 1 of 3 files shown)


uds5501:
Respected Seniors, this is literally my first PR. Kindly do guide me if I do anything wrong here

WillAyd:
Thanks for the PR! This appears to have broken a few tests. Be sure to check out the contributing to pandas guide and follow the steps locally to make sure you don't get any test errors before pushing to GitHub:

https://pandas.pydata.org/pandas-docs/stable/contributing.html

jorisvandenbossche:
Some of those tests were added explicitly to test this in https://github.com/pandas-dev/pandas/pull/13441 (it was a PR to disallow negative windows, but explicitly tested windows of zero not too raise *upon construction*). 
I am not fully sure why those were added, so I think you can actually change those tests.

jorisvandenbossche:
@uds5501 also general feedback: can you give the PR a more informative title? (you can still edit the title, see 'edit' button at the right of it)

uds5501:
@jorisvandenbossche I would do that soon and thanks for the link

WillAyd:
Can you add a test to ensure that this raises? Also will eventually want a whatsnew note (probably 0.24 at this point)

pandas/core/window.py
```diff
@@ -602,6 +604,8 @@ def validate(self):
         if isinstance(window, (list, tuple, np.ndarray)):
             pass
         elif is_integer(window):
+            if window == 0:
```
mroeschke: Instead, you can combine this check with the one below (i.e `window <= 0`) and then raise an appropriate error message

pandas/core/window.py
```diff
@@ -96,6 +96,8 @@ def is_freq_type(self):
         return self.win_type == 'freq'
 
     def validate(self):
+        if self.window == 0:
```
WillAyd: As @mroeschke pointed out can't you combine this with the check below? Don't think we need this

pandas/tests/test_window.py
```diff
@@ -402,16 +402,17 @@ def test_constructor(self, which):
             with pytest.raises(ValueError):
                 c(window=2, min_periods=1, center=w)
 
+    # these tests seems unnecessary here 21291
```
WillAyd: Don't need this comment

pandas/tests/test_window.py
```diff
@@ -402,16 +402,17 @@ def test_constructor(self, which):
...
     @pytest.mark.parametrize(
         'which', ['series', 'frame'])
     def test_constructor_with_win_type(self, which):
         # GH 13383
         o = getattr(self, which)
         c = o.rolling
-        c(0, win_type='boxcar')
         with pytest.raises(ValueError):
             c(-1, win_type='boxcar')
+
```
WillAyd: Errant extra line here will fail LINTing

WillAyd (changes requested):
Looks like there are some test failures that need to be addressed, but otherwise minor comments on current code

pandas/core/window.py
```diff
@@ -602,8 +604,8 @@ def validate(self):
         if isinstance(window, (list, tuple, np.ndarray)):
             pass
         elif is_integer(window):
-            if window < 0:
-                raise ValueError("window must be non-negative")
+            if window <= 0:
+                raise ValueError("window must be positive")
```
uds5501: @WillAyd  Combined them as pointed out by @mroeschke

WillAyd: My mistake in looking at the git diff I didn't realize they were two different methods. That said, are both required? Doesn't the call to super here raise already in the case where `window == 0`?

uds5501: @WillAyd To be really honest, I don't think that it's required either. Let me remove this

uds5501:
@WillAyd Please check the corrections now and see if it is good enough

pandas/core/window.py
```diff
@@ -602,8 +602,8 @@ def validate(self):
         if isinstance(window, (list, tuple, np.ndarray)):
             pass
         elif is_integer(window):
-            if window < 0:
-                raise ValueError("window must be non-negative")
+            if window <= 0:
+                raise ValueError("please enter a positive window")
```
jreback: maybe : window must be > 0

jreback (changes requested):
can you add a whatsnew note (will be for 0.23.2), so wait for that (soon)

pandas/core/window.py
uds5501: @jreback is this a suggestion for ```ValueError``` message?

jreback (changes requested):
actually if u can put a whatsnew note for 0.23.1
can get this in (bug fixes in the rolling section)

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -52,6 +52,7 @@ Groupby/Resample/Rolling
 - Bug in :func:`DataFrame.agg` where applying multiple aggregation functions to a :class:`DataFrame` with duplicated column names would cause a stack overflow (:issue:`21063`)
 - Bug in :func:`pandas.core.groupby.GroupBy.ffill` and :func:`pandas.core.groupby.GroupBy.bfill` where the fill within a grouping would not always be applied as intended due to the implementations' use of a non-stable sort (:issue:`21207`)
 - Bug in :func:`pandas.core.groupby.GroupBy.rank` where results did not scale to 100% when specifying ``method='dense'`` and ``pct=True``
+- Bug in :func:`pandas.core.windows.Windows.validate` where windows parameter was accpeting a numeric value of 0 (:issue:`21286`)
```
jreback: pls make this user facing, something like

``Bug in :func:`pandas.DataFrame.rolling` and :func:`pandas.Series.rolling` which incorrectly accepted a 0 window size rather than raising``

jreback (changes requested):
can you add a whatsnew note (will be for 0.23.2), so wait for that (soon)

jreback:
@WillAyd were your comments addressed?

jreback (approved):

jreback:
lgtm. @WillAyd

WillAyd (approved):
lgtm as well. @jreback let me know if / when you want me to merge

PR #18674, opened 2017-12-06
ERR: ValueError when merging on incompatible dtypes

reidy-p:
- [ ] closes #9780
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

PR #21310, opened 2018-06-04
REGR: allow merging on object boolean columns

### LINKED ISSUE #21119: pandas 0.23 fails on column<->index boolean merge

```python
import pandas as pd

df1 = pd.DataFrame({'x': pd.Series([True, False], dtype=bool)})
df2 = df1.set_index('x')

df1.merge(df2, left_on='x', right_index=True)
```

#### Problem description
On Pandas `0.23`, I get the following Exception:

```
venv/local/lib/python2.7/site-packages/pandas/core/frame.pyc in merge(self, right, how, on, left_on, right_on, left_index, right_index, sort, suffixes, copy, indicator, validate)
   6377                      right_on=right_on, left_index=left_index,
   6378                      right_index=right_index, sort=sort, suffixes=suffixes,
-> 6379                      copy=copy, indicator=indicator, validate=validate)
   6380
   6381     def round(self, decimals=0, *args, **kwargs):

venv/local/lib/python2.7/site-packages/pandas/core/reshape/merge.pyc in merge(left, right, how, on, left_on, right_on, left_index, right_index, sort, suffixes, copy, indicator, validate)
     58                          right_index=right_index, sort=sort, suffixes=suffixes,
     59                          copy=copy, indicator=indicator,
---> 60                          validate=validate)
     61     return op.get_result()
     62

venv/local/lib/python2.7/site-packages/pandas/core/reshape/merge.pyc in __init__(self, left, right, how, on, left_on, right_on, axis, left_index, right_index, sort, suffixes, copy, indicator, validate)
    552         # validate the merge keys dtypes. We may need to coerce
    553         # to avoid incompat dtypes
--> 554         self._maybe_coerce_merge_keys()
    555
    556         # If argument passed to validate,

venv/local/lib/python2.7/site-packages/pandas/core/reshape/merge.pyc in _maybe_coerce_merge_keys(self)
    976             # incompatible dtypes GH 9780, GH 15800
    977             elif is_numeric_dtype(lk) and not is_numeric_dtype(rk):
--> 978                 raise ValueError(msg)
    979             elif not is_numeric_dtype(lk) and is_numeric_dtype(rk):
    980                 raise ValueError(msg)

ValueError: You are trying to merge on bool and object columns. If you wish to proceed you should use pd.concat
```

#### Expected Output

```
       x
0   True
1  False
```

<details>

```
INSTALLED VERSIONS
------------------
commit: None
python: 2.7.13.final.0
python-bits: 64
OS: Linux
OS-release: 4.9.49-moby
machine: x86_64
processor:
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: None.None

pandas: 0.23.0
pytest: 3.4.0
pip: 9.0.3
setuptools: 39.0.1
Cython: 0.28.1
numpy: 1.14.0
scipy: 1.0.0
pyarrow: None
xarray: None
IPython: 5.5.0
sphinx: 1.6.5
patsy: 0.4.1
dateutil: 2.6.1
pytz: 2017.3
blosc: None
bottleneck: 1.2.1
tables: None
numexpr: 2.6.4
feather: None
matplotlib: 2.1.0
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.8.1
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
```
</details>

jorisvandenbossche:
Fixes https://github.com/pandas-dev/pandas/issues/21119

This is a quick/hacky fix for the specific case in that issue, i.e. merging boolean / object boolean columns. 

However, I have the feeling we should do a more general solution for cases where object dtyped columns might be actually mergeable. The problem is that inferring the type can be costly? (and eg when having actual strings object columns, this inferring was also unnecessary and thus will unneeded slow down `merge`?)


pandas/tests/reshape/merge/test_merge.py
```diff
@@ -1526,6 +1526,17 @@ def test_merge_on_ints_floats_warning(self):
             result = B.merge(A, left_on='Y', right_on='X')
             assert_frame_equal(result, expected[['Y', 'X']])
 
+    def test_merge_incompat_infer_object(self):
```
jreback: this is not a great test name, it should be related to bool.

pandas/tests/reshape/merge/test_merge.py
```diff
@@ -1526,6 +1526,17 @@ def test_merge_on_ints_floats_warning(self):
...
+        # GH21119
+        df1 = DataFrame({'key': Series([True, False], dtype=object)})
+        df2 = DataFrame({'key': [True, False]})
```
jreback: you need to have a NaN in the object dtype as well, this should test inferred and bool on both sides (with and w/o NaN).

pandas/tests/reshape/merge/test_merge.py
```diff
@@ -1526,6 +1526,17 @@ def test_merge_on_ints_floats_warning(self):
...
+
+        expected = DataFrame({'key': [True, False]}, dtype=object)
+        result = pd.merge(df1, df2, on='key')
```
jreback: need to test the error case (e.g. bool / inferred bool) on 1 side and numeric on the other (I *think* this will raise), prob have an existing test, just need to add to it

pandas/tests/reshape/merge/test_merge.py
```diff
@@ -1526,6 +1526,17 @@ def test_merge_on_ints_floats_warning(self):
...
+        assert_frame_equal(result, expected)
+        result = pd.merge(df2, df1, on='key')
```
jreback: blank lines between cases

pandas/tests/reshape/merge/test_merge.py
```diff
@@ -1526,6 +1526,17 @@ def test_merge_on_ints_floats_warning(self):
...
+        assert_frame_equal(result, expected)
```
jreback: ideally use the how fixture

jreback (changes requested):

jorisvandenbossche:
> this is not hacky it’s inline with what we normally do

What I mean is that it only solves a very specific case, while there might be other cases that pop-up later that might need similar fixes (although that the boolean case probably is the most important one). 

There are other reasons why you could have for some reason a column of `object` dtype, where you could perfectly merge it with another column of properly typed data, was it not for us prematurely erroring on it.

A more 'permissive' solution would be to only error for "clear" errors (eg the int on datetime, or tz-aware with non-tz aware), and leave anything with object columns as it was before (leave it to the merge algorithm to see whether there is something to merge).

jreback:
@jorisvandenbossche I don't know about that. We are pretty strict and only error on obvious things (e.g. we allow ints to merge regardless of kinds). so ok with this for now.

jorisvandenbossche:
> only error on obvious things

object dtyped columns are not *that* obvious, they can occur for many reasons. 

Eg in geopandas I had the problem that we created a columns of NaNs that was of object dtype in a specific case (due to a previous merge of an empty and non empty frame (and if you create an empty frame it gets object dtype if you don't specify anything specific), and then merging that again gave this error, while before it perfectly worked, as merging object NaNs on another numerical column is no problem.

Typically those issues are solvable (as my example case), but can be tricky to find, and leads to annoying errors in working code in downstream users before it gets fixed.

jorisvandenbossche:
For example, we could also have it as warnings instead of errors, to indicate something might be going wrong with your dtypes.

jreback:
we have been erroring in this for a while
warnings are a backwards step and imho generally ignored anyhow

jorisvandenbossche:
> we have been erroring in this for a while

I don't think that is correct. As far as I can see, we only started erroring in 0.23.0: https://github.com/pandas-dev/pandas/pull/18674

jreback:
that was in 0.22

crepererum:
> that was in 0.22

That's weird. The test case works in `0.22.0` but fails in `0.23.0`.

jorisvandenbossche:
@jreback on the time of merging it was 0.22, but that became 0.23 eventually :) Anyway, so it is only from the last release.

jorisvandenbossche:
To be more explicit about the the alternative I was thinking about, something like this (using pseudo-code):

```
if (is_object_dtype(left) and not is_object_dtype(right)) or (the other way around):
    warn("You are trying to merge columns of object and {other.dtype} dtype.
          This might indicate an error ...")
    coerce non-object to object (which is what was happening before)
    and continue merging
else:
    ...
    the existing more specific checks for dtypes that are *certainly* incompatible that raise errors (eg in case of merging int on datetime64)
```

that would keep the dubious cases (where one of both columns is object dtype) working but with a warning to indicate to users they need to check their dtypes. But also keeps the informative errors introduced in 0.23.0 for those cases where we are sure it is incompatible.

jreback:
I don't think warnings are a good idea. if you want to merge this, pls make the changes.

jorisvandenbossche:
> I don't think warnings are a good idea.

Can you explain why? 
A warning instead of error can give a similar indication of "there is something wrong with your dtypes", but without breaking working code.

jreback:
because we don't do this for anything else, it is simply out of place here. If you start warning for this, then you need to warn for practiclaly any non-matched but ok to merge errors. that would be so much noise it would be incomprehensible. Either a merge is good or its an error.

jorisvandenbossche:
> Either a merge is good or its an error.

The thing is that a merge *is* often good (meaning "working as expected", eg this broke a case in geopandas that was perfectly working before), and that it is very difficult for pandas to judge wether a merge is good or not.  
I proposed a warning to at least be somewhat informative that there is something fishy about the types.

jreback:
how is object type with not object type not an error? 

note i wouldn’t object to allowing extension types to be a separate clause here (which may solve the issue for geopandas)

jorisvandenbossche:
> how is object type with not object type not an error?

Because object can be by definition "anything". Whether we regard that as an error or not is something that we can discuss, but it is not that it "should" be an error (it can work, it worked before for certain cases). 

> note i wouldn’t object to allowing extension types to be a separate clause here (which may solve the issue for geopandas)

merging on two extension types will already work if they are equal dtypes. 
But do you mean with the above that you would not error on merging "object + extension type" ?

crepererum:
Could we agree on shipping a minimal solution that fixes the described bug and discuss a broader / more general solution in another issue? I could take over this PR if @jorisvandenbossche wants.

pandas/tests/reshape/merge/test_merge.py
jorisvandenbossche: This is also not used for the other tests on this, so leaving this for now

jorisvandenbossche:
Yes, this will certainly be merged for 0.23.1 so at least the boolean case will be fixed. Updated the PR now (was just waiting with it in case we could get agreement on a more thorough solution).

jreback (approved):

jreback:
lgtm @jorisvandenbossche if you want to create an issue to show a warning as suggested above then ok, otherwise merge on pass.

jorisvandenbossche:
(travis passed)

Issue #21140, opened 2018-05-20
Add Timedelta Support to JSON Reader with orient="table"


WillAyd:
Right now trying to parse JSON using orient='table' will fail if the object contains Timedeltas intentionally:

https://github.com/pandas-dev/pandas/blob/bc37ea2e05019a89adaa48159b220483598d1898/pandas/io/json/table_schema.py#L310

Since #19065 was merged I think this should now be possible if anyone wants to take a look

nprad:
I would like to work on this issue!

fjdiod:
Hello, should pd.Timedelta be used for parsing here? I mean, use it somewhere inside parse_table_schema?

WillAyd:
Yes IIRC correctly the limiting factor previously was that timedeltas are represented as ISO 8601 durations in the JSON table Schema, but when this was first built the Timedelta constructor would not parse that duration format as an argument. 

Since the PR in the original post that is no longer an issue so this should be possible to do now

fjdiod:
Now, if parse durations with Timedelta constructor in parse_table_schema, they get coerced into int64 after, in FrameParser from json.py(it seems that it tries coerce floats without deciaml digits). Is it expected behavior? Maybe their dtype should be changed from timedelta64 to timedelta64[ns]?

PR #21318, opened 2018-06-04
BUG: Fix empty Data frames to JSON round-trippable back to data frames

### LINKED ISSUE #21287: Empty data frames not round-trippable to JSON
#### Code Sample

```python
import pandas as pd
df = pd.DataFrame([], columns=['a', 'b', 'c'])
df.to_json('tmp.json', orient='table')
pd.read_json('tmp.json', orient='table')

>> KeyError: "['index' 'a' 'b' 'c'] not in index"

```
#### Problem description

Empty data frames saved as JSON fail to load back to data frames.
Quick fix: replace [this line](https://github.com/pandas-dev/pandas/blob/master/pandas/io/json/table_schema.py#L299) with

```python
df = DataFrame(table['data'], columns=col_order)[col_order]
```

#### Expected Output

```python
print(df)
Empty DataFrame
Columns: [a, b, c]
Index: []
```

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.3.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
processor: Intel64 Family 6 Model 158 Stepping 9, GenuineIntel
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.23.0
pytest: 3.2.3
pip: 10.0.1
setuptools: 36.6.0
Cython: 0.27.2
numpy: 1.13.3
scipy: 1.0.0
pyarrow: 0.7.1
xarray: 0.9.6
IPython: 6.2.1
sphinx: 1.6.5
patsy: 0.4.1
dateutil: 2.6.1
pytz: 2017.2
blosc: 1.5.1
bottleneck: 1.2.1
tables: 3.4.2
numexpr: 2.6.4
feather: 0.4.0
matplotlib: 2.1.0
openpyxl: None
xlrd: 1.1.0
xlwt: None
xlsxwriter: 1.0.2
lxml: None
bs4: 4.6.0
html5lib: 1.0b10
sqlalchemy: 1.2.3
pymysql: None
jinja2: 2.9.6
s3fs: None
fastparquet: 0.1.3
pandas_gbq: None
pandas_datareader: None

</details>

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Darwin
OS-release: 17.5.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.23.0
pytest: None
pip: 9.0.3
setuptools: 39.1.0
Cython: None
numpy: 1.14.3
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: None
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: 2.6.4
feather: None
matplotlib: 2.2.2
openpyxl: None
xlrd: 1.1.0
xlwt: None
xlsxwriter: 1.0.4
lxml: None
bs4: None
html5lib: 1.0.1
sqlalchemy: 1.2.7
pymysql: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

pyryjook:
[x] closes #21287
[x] tests added / passed
[x]passes git diff upstream/master -u -- "*.py" | flake8 --diff
[x]whatsnew entry

Fixes the bug occurring when empty DF, previously saved to JSON-file, is read from JSON back to DF.


WillAyd (changes requested):
Need test and whatsnew

pyryjook:
Sure, I'll add those both!

pandas/tests/io/json/test_json_table_schema.py
```diff
@@ -86,6 +87,16 @@ def test_multiindex(self):
         assert result == expected
 
 
+class TestParseSchema(object):
```
WillAyd: Shouldn't need a new class here - can you just move this to `TestTableOrientReader`?

pandas/tests/io/json/test_json_table_schema.py
```diff
@@ -86,6 +87,16 @@ def test_multiindex(self):
...
+
+    def test_empty_json_data(self):
```
WillAyd: Rename this to `test_empty_frame_roundtrip`

pandas/tests/io/json/test_json_table_schema.py
```diff
@@ -86,6 +87,16 @@ def test_multiindex(self):
...
+        # GH21287
+        df = pd.DataFrame([], columns=['a', 'b', 'c'])
+        json = df.to_json(None, orient='table')
+        df = parse_table_schema(json, True)
```
WillAyd: Let's use `pd.read_json` here instead

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -92,7 +92,7 @@ I/O
 
 - Bug in IO methods specifying ``compression='zip'`` which produced uncompressed zip archives (:issue:`17778`, :issue:`21144`)
 - Bug in :meth:`DataFrame.to_stata` which prevented exporting DataFrames to buffers and most file-like objects (:issue:`21041`)
--
+- Bug in IO JSON methods reading empty JSON schema back to DataFrame caused an error (:issue:`21287`)
```
WillAyd: Explicitly reference :func:\`read_json\` and make sure to qualify that this only applies when \``orient='table'\``

WillAyd: Can also update reference to :class:\`DataFrame\`

pandas/tests/io/json/test_json_table_schema.py
```diff
@@ -86,6 +87,16 @@ def test_multiindex(self):
...
+        assert df.empty
```
WillAyd: To make sure that we've preserved the frame metadata we should use the `tm.assert_frame_equal` function here. Typically with tests we will:

1. Create a variable called `expected`, which is very explicit about what we want (here that's just a copy of `df`)
2. Assign to a variable called `result`, which here would be the result of `pd.read_json`
3. Make sure result and expected match using `tm.assert_frame_equal`

WillAyd (changes requested):
Some comments but nothing major - just need to standardize and clean up the code here a bit

pyryjook:
Thanks for the comments! I'll look into this later today.

pandas/tests/io/json/test_json_table_schema.py
```diff
@@ -571,3 +561,11 @@ def test_multiindex(self, index_names):
...
         result = pd.read_json(out, orient="table")
         tm.assert_frame_equal(df, result)
+    
+    def test_empty_frame_roundtrip(self):
+        # GH 21287
+        df = pd.DataFrame([], columns=['a', 'b', 'c'])
+        expected = df.copy()
+        out = df.to_json(None, orient='table')
+        result = pd.read_json(out, orient='table')
+        tm.assert_frame_equal(expected, result)
```
pyryjook: This raises an assertion error:
```
E       AssertionError: DataFrame.index are different
E
E       DataFrame.index classes are not equivalent
E       [left]:  Index([], dtype='object')
E       [right]: Float64Index([], dtype='float64')
```

That's something I need to dig deeper, but any pointers on what might be the reason for this would be great.

ludaavics: Thanks for doing this PR! Beat me to it :)
A bit weak but what do we think of just
`pd.testing.assert_frame_equal(expected, actual, check_dtype=False)` ?

Otherwise I would guess we have to go down the road of including the dtypes in the JSON representation?

pyryjook: Good point! And actually it only works if I assert it like this:
`pd.testing.assert_frame_equal(expected, result, check_dtype=False, check_index_type=False)`
So both `check_dtype=False`and `check_index_type=False` have to be set to `False` in order to get the assertion right.

Thoughts on this?

WillAyd: Ignoring the dtype difference is not the solution. The point of this format is to persist that metadata.

What I would do is check that the proper type information for the index is being written out (you can use a io.StringIO instance instead of writing to `None`). If that appears correct then there would be an issue with the reader that is ignoring or casting the type of the index after the fact

pyryjook: Yeah, you're right, ignoring the `dtype` and the `index_type` will just hide the problem.

Did some initial testings and it seems that on the reading side empty `data` with `data.dtype == 'object'` gets coerced to Float64 without any clear reason.

I'll push a commit with fix proposal for comments.

pandas/io/json/json.py
```diff
@@ -686,7 +686,7 @@ def _try_convert_data(self, name, data, use_dtypes=True,
 
         result = False
 
-        if data.dtype == 'object':
+        if len(data) and data.dtype == 'object':
```
pyryjook: This is the fix that seem to solve the error.

Any thoughts on this?

WillAyd: The point of the JSON table schema is that we can be explicit about the types of the columns, so we shouldn't need to infer anything. Any way to avoid a call to this method altogether?

gfyoung: > The point of the JSON table schema is that we can be explicit about the types of the columns, so we shouldn't need to infer anything.

@WillAyd : I'm confused how this is relevant to the patch.

WillAyd: My point is that since the datatypes are explicitly defined in the schema that we shouldn't need any type of inference, which I get the impression this method is doing

gfyoung: > My point is that since the datatypes are explicitly defined in the schema that we shouldn't need any type of inference, which I get the impression this method is doing

Sure, but I'm still not seeing relevance.  The patch is pretty straightforward.

WillAyd: The relevance is this "fix" still wouldn't cover all use cases. It helps for the empty DataFrame, but what happens if the frame contains the value `1.` with a dtype of object? AFAICT would still cast to float, which again is not desirable and works against the explicitness of the JSON schema

pyryjook: Yeah, to me it sounds like the possible problematics related to the method itself calls for a new issue, don’t you think?

gfyoung: > Yeah, to me it sounds like the possible problematics related to the method itself calls for a new issue, don’t you think?

@pyryjook : Yes, I agree with this sentiment.

> The relevance is this "fix" still wouldn't cover all use cases. It helps for the empty DataFrame, but what happens if the frame contains the value 1. with a dtype of object?

@WillAyd : How is this concern relevant to the particular issue at hand, and wouldn't this have been a problem even before this patch.  We want PR's to be atomic, and I sense the ask here is a little too much for this PR.

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -92,7 +92,7 @@ I/O
 
 - Bug in IO methods specifying ``compression='zip'`` which produced uncompressed zip archives (:issue:`17778`, :issue:`21144`)
 - Bug in :meth:`DataFrame.to_stata` which prevented exporting DataFrames to buffers and most file-like objects (:issue:`21041`)
--
+- Bug in IO JSON :func:`read_json`reading empty JSON schema with ``orient='table'`` back to :class:DataFrame caused an error (:issue:`21287`)
```
gfyoung: @pyryjook : If you could address the merge conflicts, that would be great.

pyryjook: Sure, I’ll certainly do that when I’m on my laptop again!

pandas/io/json/json.py
WillAyd: Agreed that it is conceptually two different issues tied together. I think there's three ways to go about this:

1. Implement this without the "len() fix", knowing that empty DataFrames that can be coerced to float will, just like non-empty DataFrames will
2. Implement this with the "len() fix" and live with the fact that empty DataFrames can be round-tripped cleanly but non-empty DataFrames with "float-like" data may still be undesirably coerced to float OR
3. Find a way to prevent the coercion altogether along with this change

I'm not clear what the level of effort is for the third point above. If it is large then for sure it would make sense for one of the first two options. Otherwise I don't think it makes sense to introduce this inconsistency into the code base, as nuanced as it is

ludaavics: My personal opinion, I don't think option 2 makes much sense. As a user, I would rather have consistent, though undesirable behavior (and leave the test a bit weak, ignoring dtypes, or perhaps marking as expected failure?) than have empty and non empty data frame behave differently.
If at least non empty dataframe behaved as expected, and there was bad behavior on the corner case of empty ones, i could better see the case for living with inconsistent behavior.

Option 3 is obviously the best choice, though spontaneously seems a bit overkill to me? But, given #21140 I would be glad to take a look this week end and circle back with my best shot.

pyryjook: I swiftly went thru the past commits related to the method and that way tried to find out the reasoning behind the implementation. First of all it's quite old method (5 years) and it has seen only a few modifications during it's lifetime. On top of that, the functionality of it does seem to be quite complexly tied to multiple use cases. 

On that light, knowing its quite subtle look on the matter, it sure looks like quite fundamental task to re-think the purpose (or existence) of that method within this PR.

I completely get the point that the fix with the `len()` or without would only be a compromise when looking at the whole picture.

Clearly, my opinion of the complexity is affected by the fact that this is my first contribution to this library :)

WillAyd: I agree with @ludaavics that 2 is the least desirable. I would say if you don't see an apparent solution to number 3 above then create a separate issue about the coercion of object types to float with `read_json` and `table='orient'`. After that you can parametrize the test for `check_dtypes`, xfailing the strict check and placing a reference via a TODO comment to the issue around coercion.

gfyoung: @WillAyd : Yes, that sounds like a good plan.

pyryjook: Great, sounds like a plan! I'll make the new issue and the changes to the code accordingly. Thanks guys!

pandas/tests/io/json/test_json_table_schema.py
```diff
@@ -560,3 +560,13 @@ def test_multiindex(self, index_names):
...
         tm.assert_frame_equal(df, result)
+
+    def test_empty_frame_roundtrip(self):
+        # GH 21287
+        df = pd.DataFrame([], columns=['a', 'b', 'c'])
+        expected = df.copy()
+        out = df.to_json(orient='table')
+        result = pd.read_json(out, orient='table')
+        # TODO: After DF coercion issue (GH 21345) is resolved, tighten type checks
+        tm.assert_frame_equal(expected, result,
```
WillAyd: Minor nit but can you parametrize this with a "strict_check" parameter whose values can be `True` and `False`, with the former being marked as an xfail? You can see an example of this below:

https://github.com/pandas-dev/pandas/blob/0c65c57a279e755ab7093db925d1e580f9878dae/pandas/tests/io/json/test_json_table_schema.py#L498

The explicit xfail gives more visibility to the issue (I'm being overly cautious here)

pyryjook: Sure, I’ll make the change. Have to say that I appreciate your pedantics on these! 😊

WillAyd (approved):
Changes lgtm. I'm OK with merge if tests pass - thanks!

pyryjook:
Great, I just resolved the merge conflicts in the whatsnew file.

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -84,4 +85,4 @@ Reshaping
 
 Other
 
-- Tab completion on :class:`Index` in IPython no longer outputs deprecation warnings (:issue:`21125`)
```
gfyoung: Why is this showing up in the diff?

pyryjook: Yeah, just noticed the same. Should not be there, sry

gfyoung: No worries!  Merging branches can surprise you sometimes 😄

jreback: there was a removed newline, fixed

jreback:
thanks @pyryjook

doc/source/whatsnew/v0.23.1.txt
pyryjook: Great, thanks! I did not have change to fix it during the weekend, but great that it was resolved already.

PR #21330, opened 2018-06-05
DOC: fix mistake in Series.str.cat

h-vetinari:
Fix error in API-docstring that was introduced at the end of #20347 due to timepressure for the v.0.23-cutoff: removed functionality that categorical callers get categorical output, but forgot to adapt doc-string.

Unfortunately, this survived both #20347 and the follow-up, but since v.0.23.1 is coming soon, I didn't wanna let this opportunity pass.

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -2172,9 +2172,9 @@ class StringMethods(NoNewAttributesMixin):
 
         Returns
         -------
-        concat : str if `other is None`, Series/Index of objects if `others is
-            not None`. In the latter case, the result will remain categorical
-            if the calling Series/Index is categorical.
+        concat :
+            str if `others` is None, Series/Index of objects if `others` is not
+            None.
 
         See Also
         --------
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -2172,9 +2172,9 @@ class StringMethods(NoNewAttributesMixin):
 
         Returns
         -------
-        concat :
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -2172,9 +2172,9 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
 
         Returns
         -------
-        concat : str if `other is None`, Series/Index of objects if `others is
-            not None`. In the latter case, the result will remain categorical
-            if the calling Series/Index is categorical.
+        concat :
+            str if `others` is None, Series/Index of objects if `others` is not
+            None.
```
gfyoung: 1) I would still put all possible data-types right after the `concat: ` part.

2) OCD kicking in here: can we rephrase so that `None` isn't by itself?

h-vetinari:
@gfyoung 
It doesn't fit, unfortunately. Otherwise I would have left it there as well.

Current state:
fix_str_cat_doc_curr

Propsed:
fix_str_cat_doc_new

I'm guessing this is mainly due to the docstring parser + linelength-limit; it might work without a line break - I could do a `#noqa` line, I guess?

pandas/core/strings.py
h-vetinari: Fair enough. See new suggestion

gfyoung (approved):
Nice!  LGTM!

jorisvandenbossche (approved):

jorisvandenbossche:
@h-vetinari Thanks!

Issue #18824, opened 2017-12-18
DataFrame vs Series vs Index arithmetic Roundup


jbrockmendel:
Categorical:
- [x] #19513 Comparisons between Categorical, CategoricalIndex, and Series[category] are inconsistent
- [x] #8995 API: comparisons of categorical data and (scalar or list-like)

Series-specific:
- [x] #19252 Series has no `__rdivmod__` (closed by #23271)
- [x] #19322 Series division by zero does not get filled for some dtypes
- [x] #13843 BUG/API Series/Index integer floordiv/mod inconsistency

DataFrame
- [x] #15697 `pd.DataFrame([pd.NaT]).eq(pd.NaT)` returns `NaN` instead of bool.  `==` is OK. (closed by #22163)
- [x] #13128 `DataFrame.__eq__` is raising instead of returning bool (closed by #22163)
- [x] #12437 `DataFrame.sub` broadcasting problem (closed by (I think) #22163)
- [x] #8554 `DataFrame[datetime64] - datetime` is returning `DataFrame[datetime64]` instead of `DataFrame[timedelta64]` (closed by #22163)
- [x] #8932 `pd.Timestamp('2000-01-01') > pd.DataFrame({'x': range(5)})` returns 5 Trues instead of raising (closed by #22163)
- [ ] #9006 datetime64 comparisons raise incorrectly
- [x] #17559 `DataFrame[datetime64] != pd.Series([pd.NaT])` raises (closed by (I think) #22163)
- [x] #20088 ``DataFrame[timedelta64] / timedeltalike`` (closed by #22163)
- [x] #21610 dtype changed when DataFrame add DateOffset (closed by #22163)
- [x] #22005 DataFrame(Timestamp) add Timedelta fails (closed by #22163)
- [x] #18874 DataFrame[datetime64].sub non-nano datetime64 fails (closed by #22163)
- [x]   #22017 Timestamp comparison inconsistency (closed by #22163)

Bool/Bitwise Methods:
- [x] #5035 BUG: bool methods are wrong with DataFrame and Series (or at least wrong error)
- [ ] #16873 ERR: Remove bitwise operations on dtype=object?
- [x] #13896 API: Series/DataFrame logical op NaN propagation

Unsorted:
- [x] #19123 DatetimeIndex + ndarray[int] wrong, reverse op errors (TODO: track down what PR fixed it)
- [x] #15552 BUG: tz-aware datetime with column-wise comparisions failing
[...]

PR #18224, opened 2017-11-10
Tslibs offsets immutable

jbrockmendel:
This is the big hurdle towards making DateOffsets immutable.

- Make tslibs.offsets._BaseOffset a cython class, define `n` and `normalize` to be readonly
- Update tests that try to alter offsets inplace
- Poke at things until pickle tests stop failing <-- the part with the kludge

I'm not wild about the hack in pickle_compat, but haven't found any other way to avoid `TestCommon.test_pickle_v0_15_2` failing with:

```
cls = <class 'pandas.tseries.offsets.YearBegin'>, base = <type 'object'>, state = None

    def _reconstructor(cls, base, state):
        if base is object:
>           obj = object.__new__(cls)
E           TypeError: object.__new__(YearBegin) is not safe, use pandas._libs.tslibs.offsets._BaseOffset.__new__()
```

After this there will then be a) a bunch of other attributes to make readonly and b) the optimizations that are available once DateOffsets are immutable.  But first we need to either OK the pickle_compat hack or find another way around.

PR #21341, opened 2018-06-06
make DateOffset immutable

jbrockmendel:
Returning to the long-standing goal of making `DateOffset`s immutable (recall: `DateOffset.__eq__` calls `DateOffset._params` which is _very_ slow.  `_params` can't be cached ATM because `DateOffset` is mutable`)

Earlier attempts in this direction tried to make the base class a `cdef class`, but that has run into `pickle` problems that I haven't been able to sort out so far.  This PR goes the patch-`__setattr__` route instead.

Note: this PR does _not_ implement the caching that is the underlying goal.

I'm likely to make some other PRs in this area, will try to keep them orthogonal.


pandas/_libs/tslibs/offsets.pyx
```diff
@@ -304,6 +304,15 @@ class _BaseOffset(object):
...
     _attributes = frozenset(['n', 'normalize'])
 
+    def __init__(self, n=1, normalize=False):
+        n = self._validate_n(n)
+        object.__setattr__(self, "n", n)
+        object.__setattr__(self, "normalize", normalize)
+        object.__setattr__(self, "_cache", {})
+
+    def __setattr__(self, name, value):
+        raise AttributeError("DateOffset objects are immutable.")
```
shoyer: The standard approach is usually to use private attribute like `_n` and provide access via a properties. Is there a reason why that wouldn't make sense here?

jbrockmendel: That would absolutely make sense, but since micro-benchmarks are a concern I prefer to avoid the `property` overhead.  The best-case would be the implementation in #18824, but I can't figure out the pickle errors there.

shoyer: OK, seems reasonable to me.

jbrockmendel:
This is better than nothing since it will allow us to move forward with some serious perf improvements to PeriodIndex ops, but is still pretty ugly.  If there's a decent chance of getting someone more competent than me to look at the pickle issue holding up #18824 that would be the best-case.

pandas/_libs/tslibs/offsets.pyx
```diff
@@ -304,6 +304,15 @@ class _BaseOffset(object):
     _day_opt = None
     _attributes = frozenset(['n', 'normalize'])
 
+    def __init__(self, n=1, normalize=False):
+        n = self._validate_n(n)
+        object.__setattr__(self, "n", n)
```
jreback: I would just set these as regular attributes, I doubt this actually makes a signfiicant perf difference.

pandas/_libs/tslibs/offsets.pyx
```diff
@@ -304,6 +304,15 @@ class _BaseOffset(object):
...
+        object.__setattr__(self, "normalize", normalize)
+        object.__setattr__(self, "_cache", {})
```
jreback: what is _cache for?

jreback (changes requested):

pandas/_libs/tslibs/offsets.pyx
jbrockmendel: Not sure I understand.  These _are_ regular attributes, but the PR overrides `__setattr__` so we have to use `object.__setattr__`.

pandas/_libs/tslibs/offsets.pyx
jbrockmendel: If we don't create this explicitly in `__init__` then `@cache_readonly` lookups will try to create it, which will raise.

pandas/_libs/tslibs/offsets.pyx
jreback: oh sorry you are right. 

can you set these as cdef readonly? (and maybe type them) or not yet?

jbrockmendel: I really really wish I could, but until someone smarter than me looks at #18824 `cdef class` causes pickle test failures.

WillAyd: @jbrockmendel what pickle issue are you referring to in that change? That just looks like a master tracker but not sure the exact issue contained within

pandas/_libs/tslibs/offsets.pyx
jreback: can you put this on the class? (or is that really weird)?

pandas/_libs/tslibs/offsets.pyx
jbrockmendel: @WillAyd thanks for catching that, should be #18224.  Will fix above.

pandas/_libs/tslibs/offsets.pyx
jbrockmendel:
```
pd.DateOffset._cache = {}
off = pd.DateOffset(n=1)
hour = pd.offsets.BusinessHour(n=2)

>>> off._cache is hour._cache
True
```

jreback: right so we can't use the ``@cache_readonly`` decorator here? (or is that what you are doing by pre-emptively setting the cache)?

jbrockmendel: `cache_readonly.__get__` checks for a `_cache` attribute and creates one if it does not exist.  Creating it in `__init__` ensures that one exists, so a new one does not have to be created (since attempting to do so would raise)

jbrockmendel: i.e. this is necessary in order to retain the existing usages of `cache_readonly` on `DateOffset` subclasses.

jreback: ok makes sense

jreback:
@jbrockmendel ok this is ok. merge before your set_state/get_state change? does it matter?

jreback (approved):

jbrockmendel:
There will be a merge conflict I'll need to address.  Since I'm holding out hope WillAyd can help make #18224 viable (and this unnecessary), I'd rather the the getstate/setstate one get merged first.

jreback:
ok dokie

jreback (changes requested):
I *think* this needs a whatsnew as now technically DateOffsets are immutable yes? do we have some direct tests for this?

jbrockmendel:
> I think this needs a whatsnew as now technically DateOffsets are immutable yes? do we have some direct tests for this?

added

jreback (approved):

PR #21369, opened 2018-06-07
PERF: Add __contains__ to CategoricalIndex

topper-123:
- [x] progress towards #20395
- [x] xref #21022
- [ ] tests added / passed
- [x] benchmark added / passed
- [x] whatsnew entry

Currently, membership checks in ``CategoricalIndex`` is very slow as explained in #21022. This PR fixes the issue for ``CategoricalIndex``, while #21022 contains the fix for ``Categorical``. The difference between the two cases is the use of ``_engine`` for ``CategoricalIndex``, which makes this even faster than the ``Catagorical`` solution in #21022.

Tests exist already and can be found in ``tests/indexes/test_category.py::TestCategoricalIndex::test_contains``.

ASV:

```
     [0c65c57a]       [986779ab]
-      2.49±0.2ms       3.26±0.2μs     0.00  categoricals.Contains.time_contains

```

### DIFF
--- a/pandas/core/indexes/category.py
+++ b/pandas/core/indexes/category.py
@@ -324,18 +324,31 @@ class CategoricalIndex(Index, accessor.PandasDelegate):
     @Appender(_index_shared_docs['__contains__'] % _index_doc_kwargs)
     def __contains__(self, key):
         hash(key)
-        if isna(key):
+
+        if isna(key):  # if key is a NaN, check if any NaN is in self.
             return self.isna().any()
-        elif self.categories._defer_to_indexing:  # e.g. Interval values
+
+        # is key in self.categories? Then get its location.
+        # If not (i.e. KeyError), it logically can't be in self either
+        try:
             loc = self.categories.get_loc(key)
-            return np.isin(self.codes, loc).any()
-        elif key in self.categories:
-            return self.categories.get_loc(key) in self._engine
-        else:
+        except KeyError:
             return False
+
+        # loc is the location of key in self.categories, but also the value
+        # for key in self.codes and in self._engine. key may be in categories,
+        # but still not in self, check this. Example:
+        # 'b' in CategoricalIndex(['a'], categories=['a', 'b']) #  False
+        if is_scalar(loc):
+            return loc in self._engine
+        else:
+            # if self.categories is IntervalIndex, loc is an array
+            # check if any scalar of the array is in self._engine
+            return any(loc_ in self._engine for loc_ in loc)
 
     @Appender(_index_shared_docs['contains'] % _index_doc_kwargs)
     def contains(self, key):
+        hash(key)
         return key in self
 
     def __array__(self, dtype=None):
--- a/pandas/core/indexes/category.py
+++ b/pandas/core/indexes/category.py
@@ -324,20 +324,19 @@ class CategoricalIndex(Index, accessor.PandasDelegate):
     @Appender(_index_shared_docs['__contains__'] % _index_doc_kwargs)
     def __contains__(self, key):
         hash(key)
-
-        if self.categories._defer_to_indexing:
-            return key in self.categories
-
-        return key in self.values
+        if isna(key):
+            return self.isna().any()
+        elif self.categories._defer_to_indexing:  # e.g. Interval values
+            loc = self.categories.get_loc(key)
+            return np.isin(self.codes, loc).any()
+        elif key in self.categories:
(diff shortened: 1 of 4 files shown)


pandas/core/indexes/category.py
```diff
@@ -324,20 +324,19 @@ def _reverse_indexer(self):
     @Appender(_index_shared_docs['__contains__'] % _index_doc_kwargs)
     def __contains__(self, key):
         hash(key)
```
gfyoung: This might be a really silly question, but what does this line do?

pandas/core/indexes/category.py
```diff
@@ -324,20 +324,19 @@ def _reverse_indexer(self):
...
-
-        if self.categories._defer_to_indexing:
-            return key in self.categories
-
-        return key in self.values
```
gfyoung: Remind me: why do we need to check membership in `self.values` anymore?

pandas/core/indexes/category.py
topper-123: It just ensures that mutables are not passes into the function....

I think it should probably be removed, but that line is in various other places as well, so maybe a seperate PR, that goes through all similar cases? Or I can just remove it.

gfyoung: Well, you removed it in another part of the diff, hence why I'm asking.  Let's investigate for another time then, in which case I would put back the other one that you deleted.

pandas/core/indexes/category.py
topper-123: For indices, their indexing engine (i.e. ``._engine``) has a ``__contains__`` method which does the same thingbut  is faster.

gfyoung: Awesome, thanks for clarifying!

gfyoung (approved):
Nice!

cc @jreback

pandas/core/indexes/category.py
```diff
@@ -324,20 +324,21 @@ def _reverse_indexer(self):
...
+        if isna(key):
+            return self.isna().any()
+        try:
+            loc = self.categories.get_loc(key)
+        except KeyError:
+            return False
+        if is_scalar(loc):
+            return loc in self._engine
+        else:  # if self.categories is IntervalIndex, loc is an array
```
jreback: can you put a blank line between things, e.g.
```
if isna(...):
     ....

try:
   ....
except:
  ....

if is_scalar(...):
   ...

# no else needed here
return ...
```

jreback: also can you put comments before each case (not everythin needs a comment), but i find this hard to grok in its current form.

jreback (changes requested):

topper-123:
I've updated the PR.

jreback (approved):

jreback:
thanks @topper-123

jorisvandenbossche:
This is changing the implementation quite substantially, so let's move this to 0.24.0.txt?

jorisvandenbossche:
Any comments on my comment above about keeping this for 0.24.0 ?

topper-123:
Quility-wise this is ok to go into 23.2 IMO, the PRs are really not that complex, IMO, it's much faster and it doesn't change any APIs.

Also, my main motivation for writing this was speeding up slicing dataframes with a CategoricalIndex (see #20395), which previously was *very* slow (still is slow, but better than before, and now faster than fancy indexing, at least). I think a lot of people will appreciate this speedup.

jreback:
we tagged for 0.23.2 (and note is there)
it would be slightly tricky to change as there is another related change - both in 0.23.2 or 0.24
either way is fine

jorisvandenbossche:
The main reason that I am raising this is that `__contains__` checking has quite some implications (which is of course also the reason you speeded it up!), and I think it is rather easy to miss a small exotic corner case where the new implementation might differ. 
To be clear, @topper-123 , I am not questioning the quality of this PR! I just know from experience, also for us, that it is easy to miss an unintended API change (which we might not even decide to fix if it is debateble behaviour, but then that is still better to be left as 0.24.0). Since this is only about performance improvement (and not a regression), I would play on safe and give it more time in 0.24.0.


>  it would be slightly tricky to change as there is another related change - both in 0.23.2 or 0.24

Yep, it would be both in 0.23.2 or both in 0.24.0 (the other I actually tagged first as 0.24.0, but that was changed before merging)

topper-123:
Ok, your call, I won't object.

PR #21384, opened 2018-06-08
Accepts integer/float string with units and raises when unit is ambiguous

### LINKED ISSUE #12136: BUG/API: to_timedelta unit-argument ignored for string input
So, I don't know if this is by design, but it sure confused me. `to_timedelta()` operates differently on the number `1000` compared to the string `'1000'`:

``` python
In[3]: pd.to_timedelta(1000, unit='ms')
Out[3]: Timedelta('0 days 00:00:01')
In[4]: pd.to_timedelta(1000.0, unit='ms')
Out[4]: Timedelta('0 days 00:00:01')
```

and

``` python
In[6]: pd.to_timedelta('1000ms')
Out[6]: Timedelta('0 days 00:00:01')
```

while

``` python
In[5]: pd.to_timedelta('1000', unit='ms')
Out[5]: Timedelta('0 days 00:00:00.000001')
```

I've tried my best to come up with a reason for why this behaviour would be desirable, but I can't think of any.

Sup3rGeo:
I couldn't get tests to run on my machine so I will see if it passes on the CI servers.

- [x] closes #12136
- [ ] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/pandas/_libs/tslibs/timedeltas.pyx
+++ b/pandas/_libs/tslibs/timedeltas.pyx
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+# cython: profile=False
 import collections
 import textwrap
 import warnings
@@ -6,13 +7,13 @@ import warnings
 import sys
 cdef bint PY3 = (sys.version_info[0] >= 3)
 
-from cython import Py_ssize_t
+from cython cimport Py_ssize_t
 
-from cpython cimport Py_NE, Py_EQ, PyObject_RichCompare
+from cpython cimport PyUnicode_Check, Py_NE, Py_EQ, PyObject_RichCompare
 
 import numpy as np
 cimport numpy as cnp
-from numpy cimport int64_t
+from numpy cimport int64_t, ndarray
 cnp.import_array()
 
 from cpython.datetime cimport (datetime, timedelta,
@@ -32,7 +33,6 @@ from np_datetime cimport (cmp_scalar, reverse_ops, td64_to_tdstruct,
 
 from nattype import nat_strings, NaT
 from nattype cimport checknull_with_nat, NPY_NAT
-from offsets cimport to_offset
 
 # ----------------------------------------------------------------------
 # Constants
@@ -77,44 +77,6 @@ cdef dict timedelta_abbrevs = { 'D': 'd',
                                 'nanosecond': 'ns'}
 
 _no_input = object()
-
-
-# ----------------------------------------------------------------------
-# API
-
-def ints_to_pytimedelta(int64_t[:] arr, box=False):
-    """
-    convert an i8 repr to an ndarray of timedelta or Timedelta (if box ==
-    True)
-
-    Parameters
-    ----------
-    arr : ndarray[int64_t]
-    box : bool, default False
-
-    Returns
-    -------
-    result : ndarray[object]
-        array of Timedelta or timedeltas objects
-    """
-    cdef:
-        Py_ssize_t i, n = len(arr)
-        int64_t value
-        object[:] result = np.empty(n, dtype=object)
-
-    for i in range(n):
-
-        value = arr[i]
-        if value == NPY_NAT:
-            result[i] = NaT
-        else:
-            if box:
-                result[i] = Timedelta(value)
-            else:
-                result[i] = timedelta(microseconds=int(value) / 1000)
-
-    return result.base  # .base to access underlying np.ndarray
-
 
 # ----------------------------------------------------------------------
 
@@ -161,7 +123,7 @@ cpdef convert_to_timedelta64(object ts, object unit):
         if ts.astype('int64') == NPY_NAT:
             return np.timedelta64(NPY_NAT)
     elif is_timedelta64_object(ts):
-        ts = ts.astype("m8[{unit}]".format(unit=unit.lower()))
+        ts = ts.astype("m8[{0}]".format(unit.lower()))
     elif is_integer_object(ts):
         if ts == NPY_NAT:
             return np.timedelta64(NPY_NAT)
@@ -182,11 +144,7 @@ cpdef convert_to_timedelta64(object ts, object unit):
             ts = cast_from_unit(ts, unit)
             ts = np.timedelta64(ts)
     elif is_string_object(ts):
-        if len(ts) > 0 and ts[0] == 'P':
-            ts = parse_iso_format_string(ts)
-        else:
-            ts = parse_timedelta_string(ts)
-        ts = np.timedelta64(ts)
+        ts = np.timedelta64(parse_timedelta_string(ts))
     elif hasattr(ts, 'delta'):
         ts = np.timedelta64(delta_to_nanoseconds(ts), 'ns')
 
@@ -198,7 +156,7 @@ cpdef convert_to_timedelta64(object ts, object unit):
     return ts.astype('timedelta64[ns]')
 
 
-cpdef array_to_timedelta64(object[:] values, unit='ns', errors='raise'):
+cpdef array_to_timedelta64(ndarray[object] values, unit='ns', errors='raise'):
     """
     Convert an ndarray to an array of timedeltas. If errors == 'coerce',
     coerce non-convertible objects to NaT. Otherwise, raise.
@@ -206,7 +164,7 @@ cpdef array_to_timedelta64(object[:] values, unit='ns', errors='raise'):
 
     cdef:
         Py_ssize_t i, n
-        int64_t[:] iresult
+        ndarray[int64_t] iresult
 
     if errors not in ('ignore', 'raise', 'coerce'):
         raise ValueError("errors must be one of 'ignore', "
@@ -232,7 +190,7 @@ cpdef array_to_timedelta64(object[:] values, unit='ns', errors='raise'):
                 else:
                     raise
 
-    return iresult.base  # .base to access underlying np.ndarray
+    return iresult
 
 
 cpdef inline int64_t cast_from_unit(object ts, object unit) except? -1:
@@ -264,7 +222,7 @@ cpdef inline int64_t cast_from_unit(object ts, object unit) except? -1:
         m = 1L
         p = 0
     else:
-        raise ValueError("cannot cast unit {unit}".format(unit=unit))
+        raise ValueError("cannot cast unit {0}".format(unit))
 
     # just give me the unit back
     if ts is None:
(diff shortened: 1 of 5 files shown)


pandas/tests/scalar/timedelta/test_construction.py
```diff
@@ -210,3 +210,27 @@ def test_td_constructor_on_nanoseconds(constructed_td, conversion):
...
+class not_raises(object):
+    def __enter__(self):
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
+
+
+@pytest.mark.parametrize("redundant_unit, expectation", [
+    ("", not_raises()),
```
gfyoung: I believe `pytest.raises(None)` should suffice.

Sup3rGeo: It doesn't - it is still an open issue in pytest: https://github.com/pytest-dev/pytest/issues/1830

gfyoung: Odd that we can have an empty context manager with warnings assertions, but that's not implemented for exceptions.  Implementing the empty context manager wouldn't super difficult on our end (we have `pandas.util.testing.assert_raises_regex`).

@jreback @jorisvandenbossche : Thoughts?

pandas/tests/scalar/timedelta/test_construction.py
```diff
@@ -210,3 +206,27 @@ def test_td_constructor_on_nanoseconds(constructed_td, conversion):
...
+    ("us", pytest.raises(ValueError))])
+@pytest.mark.parametrize("unit", [
+    "d", "m", "s", "us"])
+@pytest.mark.parametrize("sign", [
+    +1, -1])
+@pytest.mark.parametrize("num", [
+    0.001, 1, 10])
+def test_string_with_unit(num, sign, unit, redundant_unit, expectation):
+    with expectation:
+        assert Timedelta(str(sign * num) + redundant_unit, unit=unit)\
```
gfyoung: nit: can't we somehow do multiline with an `assert` statement without the slash?

Sup3rGeo: well I could break the whole statement in two, moving the Timedelta first argument to a variable. This would reduce the line length. Should I?

gfyoung: That would also work.

gfyoung (approved):
The `not_raises` question notwithstanding (it's not a blocker unless we decide we want to do this), this looks good to me!

cc @jreback

pandas/tests/scalar/timedelta/test_construction.py
```diff
@@ -210,3 +206,28 @@ def test_td_constructor_on_nanoseconds(constructed_td, conversion):
...
+        pass
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        pass
+
+
+@pytest.mark.parametrize("redundant_unit, expectation", [
+    ("", not_raises()),
+    ("d", pytest.raises(ValueError)),
+    ("us", pytest.raises(ValueError))])
```
jorisvandenbossche: Is it needed to test that it is raising the value error with all the other combinations? (because this gives a combinatorial explosion of number of test .. ;-))

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -999,10 +1004,21 @@ class Timedelta(_Timedelta):
         if isinstance(value, Timedelta):
             value = value.value
         elif is_string_object(value):
-            if len(value) > 0 and value[0] == 'P':
+            # Check if it is just a number in a string
```
jorisvandenbossche: Do we want to support this? (do we now?)

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -245,7 +245,7 @@ cdef inline _decode_if_necessary(object ts):
     return ts
 
 
-cdef inline parse_timedelta_string(object ts):
+cdef inline parse_timedelta_string(object ts, specified_unit=None):
```
jreback: this should not be handled here at all

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -999,10 +1004,21 @@ class Timedelta(_Timedelta):
         if isinstance(value, Timedelta):
             value = value.value
         elif is_string_object(value):
-            if len(value) > 0 and value[0] == 'P':
+            # Check if it is just a number in a string
+            try:
```
jreback: just try to float here. if it succeeds it *must* have a unit specified, if not you must raise, e.g. ``Timedelta('1000')`` is not accepted

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -999,10 +1004,21 @@ class Timedelta(_Timedelta):
...
+                except (ValueError, TypeError):
+                    pass
+
+            if is_integer_object(value) or is_float_object(value):
+                value = convert_to_timedelta64(value, unit)
+            elif len(value) > 0 and value[0] == 'P':
                 value = parse_iso_format_string(value)
             else:
-                value = parse_timedelta_string(value)
+                value = parse_timedelta_string(value, unit)
```
jreback: no not pass here

pandas/tests/scalar/timedelta/test_construction.py
```diff
@@ -85,10 +85,6 @@ def test_construction():
     with pytest.raises(ValueError):
         Timedelta('10 days -1 h 1.5m 1s 3us')
 
-    # no units specified
-    with pytest.raises(ValueError):
-        Timedelta('3.1415')
```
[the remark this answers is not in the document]
jreback: we don't allow this, pls revert. a string w/o a unit must raise.

pandas/tests/scalar/timedelta/test_construction.py
```diff
@@ -210,3 +206,28 @@ def test_td_constructor_on_nanoseconds(constructed_td, conversion):
 def test_td_constructor_value_error():
     with pytest.raises(TypeError):
         Timedelta(nanoseconds='abc')
+
+
+class not_raises(object):
+    def __enter__(self):
+        pass
```
jreback: remove this its not idiomatic

pandas/tests/scalar/timedelta/test_construction.py
```diff
@@ -210,3 +206,28 @@ def test_td_constructor_on_nanoseconds(constructed_td, conversion):
...
+
+
+@pytest.mark.parametrize("redundant_unit, expectation", [
+    ("", not_raises()),
+    ("d", pytest.raises(ValueError)),
+    ("us", pytest.raises(ValueError))])
+@pytest.mark.parametrize("unit", [
+    "d", "m", "s", "us"])
+@pytest.mark.parametrize("sign", [
```
jreback: there are related tests for this pls move.

pandas/tests/scalar/timedelta/test_construction.py
```diff
@@ -210,3 +206,28 @@ def test_td_constructor_on_nanoseconds(constructed_td, conversion):
...
+    +1, -1])
+@pytest.mark.parametrize("num", [
+    0.001, 1, 10])
+def test_string_with_unit(num, sign, unit, redundant_unit, expectation):
```
jreback: this needs equivalent handling for the ``to_timedelta`` path (which is separate)

jreback (changes requested):

pandas/tests/scalar/timedelta/test_construction.py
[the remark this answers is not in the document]
jorisvandenbossche: Well, looking back at the issue https://github.com/pandas-dev/pandas/issues/12136, there was discussion about this and said we would allow this. So that is a good reason this is PR is doing that! :-) 

(but personally, I would simply not allow it)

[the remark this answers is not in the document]
jreback: allowing a string with a unit is fine
this does not have a unit

jreback:
can you rebase and fixup

jreback:
can you rebase and update according to comments

jreback:
closing as stale, but would take if rebased / updated

Sup3rGeo:
@jreback sorry for abandoning this, but I will start working back again to finish this PR, especially given that everyone has spent time to review it.

Should I start another one or can I reply to the comments on this one?

gfyoung:
@Sup3rGeo : You can keep pushing to this one.  I'll reopen.

pandas/tests/scalar/timedelta/test_construction.py
Sup3rGeo: I kinda needed it for parametrizing on the expectation. Should I then just add a if/else in the test to use or not pytest.raises?

pandas/tests/scalar/timedelta/test_construction.py
Sup3rGeo: I see, I just wanted to make sure kinda all cases were covered. I believe this does not give a lot of combinations, so I thought that for this case specifically it would be ok.

Sup3rGeo:
Hi!

> Yes, we used such a construct in one of the tests (in "pandas\tests\util\test_hashing.py:23: in TestHashing"), but you can correct this in the test.
> 
> That said, we should probably deprecate it first instead of simply changing.

Sorry but what do you mean by deprecating it instead of changing?

> Can you try:
> 
> ```
> git fetch upstream
> git merge upstream/master
> git push origin bugfix-string-timedelta
> ```
> to see if that fixes it?

Did it. Did it help? I think I screwed up with the rebase indeed. I could try starting fresh with a new branch if nothing helps.

jorisvandenbossche:
> Did it. Did it help? I think I screwed up with the rebase indeed. I could try starting fresh with a new branch if nothing helps.

Apparently not .. I also tried to rebase, and removed the commits that were not yours. That seemed to have fixed some of the problems, but not all (there are still changes in `timedeltas.pyx` in the diff that are unrelated). 

So maybe starting from a fresh branch and copying over your changes there might be the easiest to do.

> Sorry but what do you mean by deprecating it instead of changing?

So currently `pd.Timedelta("2000")` works (the string being interpreted as an integer), so we could keep that working for now, but give the user a warning that this will change in the future to raise an error. 
See also http://pandas-docs.github.io/pandas-docs-travis/contributing.html#backwards-compatibility, although that is more about deprecating a full method/function, while here you would be deprecating only a specific behaviour (but some of the explanation still holds)

Sup3rGeo:
Closing in favor of #23025

PR #21368, opened 2018-06-07
DOC: clean-up 0.23.1 whatsnew

jorisvandenbossche:
I added a "Fixed regressions" section and moved some over there. @TomAugspurger does that look ok to you? (then I'll maybe merge to the other PRs can add it there)

Also mad the bug fixes subsections into plain text to make it a bit less "heavy" (given the only few entries per section).

PR #21387, opened 2018-06-08
DOC: update whatsnew 0.23.1

jorisvandenbossche:
One of the last commits apparently added back the subsections


jorisvandenbossche:
ah, @jreback I see that you actually added that commit to that PR explicitly (I assumed is was due to some rebase or so). I removed the subsections on purpose in https://github.com/pandas-dev/pandas/pull/21368, because IMO for such a short list it is not really needed and rather adds noise to the toc (but can certainly be discussed)

jreback (commented):
i added these back because they were completely unmarked

bold is fine too

PR #21394, opened 2018-06-08
Bugfix timedelta notimplemented eq

### LINKED ISSUE #20829: Timedelta always returning False for equality tests for incompatible types

```python
import datetime
import pandas

class CustomClass:

    def __init__(self):
        pass
       
    def __eq__(self, other):
        raise Exception("Custom Class eq")

        
custom = CustomClass()
var1 = datetime.timedelta(seconds=1)
var2 = pandas.Timedelta("1s")    

# Following code raises CustomClass exception
var1 == custom

# Following code returns False, does not call CustomClass.__eq__
var2 == custom
```
#### Problem description

I am trying to implement a version of `pytest.approx` for Timedeltas, which basically returns a class with a custom `__eq__` implementation.

`Timedelta("5s") == AproxTimedelta(...)

However it does not work with pandas Timedelta because it always return False, so the `__eq__` of AproxTimedelta object is never called.

This is not the behavior for python datetime timedelta, as can be seen in the example script.

#### Expected Output

`Exception("Custom Class eq")` raised

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.3.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
processor: Intel64 Family 6 Model 158 Stepping 9, GenuineIntel
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None
pandas: 0.22.0
pytest: 3.5.0
pip: 10.0.0
setuptools: 38.5.1
Cython: 0.28.2
numpy: 1.14.1
scipy: 1.0.0
pyarrow: None
xarray: None
IPython: 6.3.1
sphinx: 1.7.1
patsy: None
dateutil: 2.6.1
pytz: 2017.2
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 2.1.1
openpyxl: 1.7.0
xlrd: 1.1.0
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

Sup3rGeo:
- [x] closes #20829 
- [x] tests added / passed
- [x] whatsnew entry


Sup3rGeo:
@jbrockmendel following up from the issue discussion, the tests look good it seems?

Except for python 2 that actually has different ordering behavior than python 3, this is why it it fails on python2:
https://docs.python.org/3.0/whatsnew/3.0.html#ordering-comparisons

What do you think? Could we just then add a special case checking for python 2 that raises TypeError?

jbrockmendel:
Seems reasonable.  What other tests are affected?  This may be a good opportunity to flesh them out.

Sup3rGeo:
I gave it a second thought and I think we should not have a special case for python 2 because if we raise manually from the `__richcmp__` function than we still won't allow the other object to try to implement the comparison.

So this last commit is only testing for `TypeError`s for python 3, letting python 2 comparisons (lt gt le ge) act as it is the default (not raising).

jbrockmendel:
Take a look at the Travis log; the test is raising before getting to the behavior you care about.

In nearly all cases we want the behavior to be identical in py2 and py3.  Is there a compelling reason to break that rule in this case?

Sup3rGeo:
Basically the comparisons are implemented differently in python 2 and python 3, so with the changes in this PR we are just following python itself. Which I believe is good for the following reasons:

- `pandas.Timedelta` won't try to change the default python behavior on comparisons, so no surprises.
- We don't break the ability of the other member of the comparison to implement it in a meaningful way (very important)

The way `pandas.Timedelta` is currently implemented in the released versions breaks item 2 because it does not follow item 1.

jbrockmendel:
> won't try to change the default python behavior on comparisons, so no surprises.

I'd argue that the default python2 behavior is itself surprising.  Having `pd.Timedelta` behave differently in py2 vs py3 seems very undesirable.  That said, it's not me you need to convince, it's @jreback.

Sup3rGeo:
I agree with you about python 2 having this surprising behavior, and that it would be desirable to have them work the same. Its just that given that our current implementation also brings undesired side-effects, if we have to choose between two evils, I would say it is better to just expose these intrinsic python2-3 differences to the user.

IMHO it breaks less "contracts" (and then surely this python2-3 differences aren't to blame pandas, but something they have to deal with when working with different backward-incompatible python versions, just like the different strings behavior and so on).

Well we might not 100% agree on this but thanks for the support so far @jbrockmendel. Let's see what @jreback thinks.

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -59,6 +59,7 @@ Data-type specific
 
 - Bug in :meth:`Series.str.replace()` where the method throws `TypeError` on Python 3.5.2 (:issue: `21078`)
 - Bug in :class:`Timedelta`: where passing a float with a unit would prematurely round the float precision (:issue: `14156`)
+- Bug in :class:`Timedelta`: Comparison with unknown types now return NotImplemented as typically expected (:issue: `20829`)
```
jreback: move to 0.24

pandas/tests/scalar/timedelta/test_timedelta.py
```diff
@@ -42,8 +43,10 @@ def test_ops_error_str(self):
             with pytest.raises(TypeError):
                 left + right
 
-            with pytest.raises(TypeError):
-                left > right
+            # GH 20829: python 2 comparison naturally does not raise TypeError
```
jreback: use compat.PY3

pandas/tests/scalar/timedelta/test_timedelta.py
```diff
@@ -103,6 +106,54 @@ def test_compare_timedelta_ndarray(self):
...
+        assert not (t == 1)
+        assert not (t == CustomClass())
+        assert not (t == CustomClass(cmp_result=False))
+
+        assert t < CustomClass(cmp_result=True)
+        assert not (t < CustomClass(cmp_result=False))
+
+        assert t == CustomClass(cmp_result=True)
+
+    @pytest.mark.parametrize("val", [
```
jreback: use a skip with the compat.PY3 here

pandas/tests/scalar/timedelta/test_timedelta.py
```diff
@@ -103,6 +106,54 @@ def test_compare_timedelta_ndarray(self):
         expected = np.array([False, False])
         tm.assert_numpy_array_equal(result, expected)
 
+    def test_custom_comparison_object(self):
```
jreback: test_compare_custom_object

pandas/tests/scalar/timedelta/test_timedelta.py
```diff
@@ -103,6 +106,54 @@ def test_compare_timedelta_ndarray(self):
...
+        assert not (t == CustomClass(cmp_result=False))
+
+        assert t < CustomClass(cmp_result=True)
+        assert not (t < CustomClass(cmp_result=False))
+
+        assert t == CustomClass(cmp_result=True)
+
+    @pytest.mark.parametrize("val", [
+        "string", 1])
+    def test_raise_comparisons_unknown_types(self, val):
```
jreback: test_compare_unknown_type

jreback:
IIRC there is a bug in the cpython impl for comparisons with timedelta in py2, maybe @cpcloud remembers where it actually is.

jreback (changes requested):

jreback:
so I think something broke as the circle tests are failing. needs to very carefully look at all failures as this is a far reaching change.

pandas/tests/scalar/timedelta/test_timedelta.py
Sup3rGeo: Done, but as your comment was in the code comment (which didn't change), it remained here after my changes.

Sup3rGeo:
I will further investigate the errors.

Sup3rGeo:
I might need some help to debug the failing test on CircleCI. Any idea how can I reproduce it locally on my machine?

jreback:
can you rebase / update

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -479,6 +479,7 @@ all-``NaT``.  This is for compatibility with ``TimedeltaIndex`` and
 Previous Behavior:
 
 .. code-block:: ipython
+- Comparing :class:`Timedelta` with unknown types now return NotImplemented instead of False (:issue:`20829`)
```
jreback: use double backticks around NotIMplemented and False.

this note is in an odd place

pandas/tests/scalar/timedelta/test_timedelta.py
```diff
@@ -103,6 +105,54 @@ def test_compare_timedelta_ndarray(self):
         expected = np.array([False, False])
         tm.assert_numpy_array_equal(result, expected)
 
+    def test_compare_custom_object(self):
+        # GH20829
```
jreback: add a comment to explain what / why this is testing

jreback (changes requested):

Sup3rGeo:
I tried installing ubuntu in a virtual machine and running the same scripts the the failing CircleCI build, and the failing test is still passing.

Sup3rGeo:
@jreback sorry what do you mean by "ping on green"?

Sup3rGeo:
@jreback, @jbrockmendel I think I am in a dead end because I cannot manage to reproduce this failure locally. I would need some help/pointers here.

jreback:
merge master and push again

Sup3rGeo:
Oh, seems like it's green now.

jreback (approved):

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -193,6 +193,7 @@ Other Enhancements
 - :func:`~DataFrame.to_csv`, :func:`~Series.to_csv`, :func:`~DataFrame.to_json`, and :func:`~Series.to_json` now support ``compression='infer'`` to infer compression based on filename extension (:issue:`15008`).
   The default compression for ``to_csv``, ``to_json``, and ``to_pickle`` methods has been updated to ``'infer'`` (:issue:`22004`).
 - :func:`to_timedelta` now supports iso-formated timedelta strings (:issue:`21877`)
+- Comparing :class:`Timedelta` with unknown types now return ``NotImplemented`` instead of ``False`` (:issue:`20829`)
```
jorisvandenbossche: NotImplemented -> raises a TypeError (that is what the user sees) 
Fixed in https://github.com/pandas-dev/pandas/pull/23355

PR #21397, opened 2018-06-08
DOC: Loading sphinxcontrib.spelling to sphinx only if it's available (#21396)

### LINKED ISSUE #21396: Missing dependency for sphinxcontrib-spelling
```
(pandas-dev) [mgarcia@xps doc]$ python make.py --num-jobs=8
Running Sphinx v1.7.5

Extension error:
Could not import extension sphinxcontrib.spelling (exception: No module named 'sphinxcontrib.spelling')
```
#### Problem description

In #21109 the package `sphinxcontrib-spelling` has been added to the sphinx configuration, but the package has not been included in the `conda` or `pip` dependencies. This is making the documentation script crash.

datapythonista:
- [X] closes #21396
- [ ] tests added / passed
- [ ] whatsnew entry

I think the optional dependencies go directly into the `.txt` files for pip and conda, and not into the yaml file. Please correct me if I'm wrong.

Seems like besides `sphinxcontrib-spelling`, a binary program `enchant` is required, and it's not available in conda for what I've seen. Not sure what's the right approach for it.

### DIFF
--- a/ci/environment-dev.yaml
+++ b/ci/environment-dev.yaml
@@ -13,3 +13,4 @@ dependencies:
   - pytz
   - setuptools>=24.2.0
   - sphinx
+  - sphinxcontrib-spelling
--- a/ci/requirements_dev.txt
+++ b/ci/requirements_dev.txt
@@ -9,3 +9,4 @@ python-dateutil>=2.5.0
 pytz
 setuptools>=24.2.0
 sphinx
+sphinxcontrib-spelling
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -16,9 +16,11 @@ import os
 import re
 import inspect
 import importlib
-from sphinx.ext.autosummary import _import_by_name
+import logging
 import warnings
+from sphinx.ext.autosummary import _import_by_name
 
+logger = logging.getLogger(__name__)
 
 try:
     raw_input          # Python 2
@@ -73,8 +75,15 @@ extensions = ['sphinx.ext.autodoc',
               'sphinx.ext.ifconfig',
               'sphinx.ext.linkcode',
               'nbsphinx',
-              'sphinxcontrib.spelling'
               ]
+
+try:
+    import sphinxcontrib.spelling
+except ImportError as err:
+    logger.warn(('sphinxcontrib.spelling failed to import with error "{}". '
+                '`spellcheck` command is not available.'.format(err)))
+else:
+    extensions.append('sphinxcontrib.spelling')
 
 exclude_patterns = ['**.ipynb_checkpoints']
 


gfyoung:
> I think the optional dependencies go directly into the .txt files for pip and conda, and not into the yaml file. Please correct me if I'm wrong.

Good call: I agree that it is an optional dependency.  You don't need it to install `pandas` for general use.

ci/requirements-optional-conda.txt
```diff
@@ -20,6 +20,7 @@ pytest-xdist
 s3fs
 scipy
 seaborn
+sphinxcontrib-spelling
```
jreback: these should go in requirements_dev.txt instead (only)

jreback (changes requested):

ci/requirements_dev.txt
```diff
@@ -9,3 +9,4 @@ python-dateutil>=2.5.0
 pytz
 setuptools>=24.2.0
 sphinx
+sphinxcontrib-spelling
```
jorisvandenbossche: this file indicates it is autogenerated

jorisvandenbossche:
As said on the issue, given the installation troubles, I would make this an optional one.

ci/environment-dev.yaml
```diff
@@ -13,3 +13,4 @@ dependencies:
   - pytz
   - setuptools>=24.2.0
   - sphinx
+  - sphinxcontrib-spelling
```
jorisvandenbossche: I think we should rather put it in the requirements-pip/conda-optional.txt

jorisvandenbossche (approved):
Apart from the comment where to put the dep, looks good to me!

ci/environment-dev.yaml
datapythonista: @jreback, you said it should be in `requirements_dev.txt`, may be you want to explain why? I also thought the optional files made more sense, but you two surely know better than me.

jreback: requirements_dev are for pure dev things
but i think things have gotten confused

jorisvandenbossche: OK, let's discuss that in another one then

kylebarron:
Just FYI, I just followed the [instructions to create a development environment with Conda](https://pandas.pydata.org/pandas-docs/stable/contributing.html#creating-a-python-environment). On the `python -m pip install -e .` step, I got
```
sphinxcontrib-spelling 4.1.0 requires PyEnchant>=1.6.5, which is not installed.
```
Is that desired behavior or should the `environment.yml` file also include `PyEnchant` as a dependency?

PR #21400, opened 2018-06-09
MAINT: Deprecate encoding from stata reader/writer

### LINKED ISSUE #21244: BUG: `read_stata` always uses 'utf8'

bashtage:
Deprecate the encoding parameter from all Stata reading and writing
methods and classes.  The encoding depends only on the file format and
cannot be changed by users.

- [X] closes #21244
- [X] tests added / passed
- [X] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -45,7 +45,7 @@ Other API Changes
 Deprecations
 ~~~~~~~~~~~~
 
--
+- :meth:`DataFrame.to_stata`, :meth:`read_stata`, :class:`StataReader` and :class:`StataWriter` have deprecated the ``encoding`` argument.  The encoding of a Stata dta file is determined by the file type and cannot be changed (:issue:`21244`).
 -
 -
 
--- a/pandas/tests/io/test_stata.py
+++ b/pandas/tests/io/test_stata.py
@@ -361,7 +361,8 @@ class TestStata(object):
 
         # GH 4626, proper encoding handling
         raw = read_stata(self.dta_encoding)
-        encoded = read_stata(self.dta_encoding, encoding="latin-1")
+        with tm.assert_produces_warning(FutureWarning):
+            encoded = read_stata(self.dta_encoding, encoding='latin-1')
         result = encoded.kreis1849[0]
 
         expected = raw.kreis1849[0]
@@ -369,8 +370,9 @@ class TestStata(object):
         assert isinstance(result, compat.string_types)
 
         with tm.ensure_clean() as path:
-            encoded.to_stata(path, encoding='latin-1',
-                             write_index=False, version=version)
+            with tm.assert_produces_warning(FutureWarning):
+                encoded.to_stata(path, write_index=False, version=version,
+                                 encoding='latin-1')
             reread_encoded = read_stata(path)
             tm.assert_frame_equal(encoded, reread_encoded)
 
@@ -1348,13 +1350,6 @@ class TestStata(object):
                 original.to_stata(path)
             assert 'ColumnTooBig' in cm.exception
             assert 'infinity' in cm.exception
-
-    def test_invalid_encoding(self):
-        # GH15723, validate encoding
-        original = self.read_csv(self.csv3)
-        with pytest.raises(ValueError):
-            with tm.ensure_clean() as path:
-                original.to_stata(path, encoding='utf-8')
 
     def test_path_pathlib(self):
         df = tm.makeDataFrame()
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -80,7 +80,8 @@ from pandas import compat
 from pandas.compat import PY36
 from pandas.compat.numpy import function as nv
 from pandas.util._decorators import (Appender, Substitution,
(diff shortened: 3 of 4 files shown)


gfyoung:
@bashtage : How come your PR is closing #21244, which is already closed?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -45,7 +45,7 @@ Other API Changes
 Deprecations
 ~~~~~~~~~~~~
 
--
+- :meth:`DataFrame.to_stata`, :meth:`read_stata`, :class:`StataReader` and :class:`StataWriter` have deprecated has deprecated ``encoding``.  The   encoding of a Stata dta file is determined by the file type and cannot be changed (:issue:`21244`).
```
gfyoung: * ``encoding`` --> the ``encoding`` argument
* "The  encoding" --> "The encoding" (extra space).

pandas/tests/io/test_stata.py
```diff
@@ -361,16 +361,15 @@ def test_encoding(self, version):
...
         result = encoded.kreis1849[0]
 
         expected = raw.kreis1849[0]
         assert result == expected
         assert isinstance(result, compat.string_types)
 
         with tm.ensure_clean() as path:
-            encoded.to_stata(path, encoding='latin-1',
-                             write_index=False, version=version)
+            encoded.to_stata(path, write_index=False, version=version)
```
gfyoung: Can we make sure that the `DeprecationWarning` is issued when we pass in the `encoding` argument?

bashtage:
I added a test and fixed the whatsnew

pandas/tests/io/test_stata.py
```diff
@@ -361,16 +361,20 @@ def test_encoding(self, version):
 
         # GH 4626, proper encoding handling
         raw = read_stata(self.dta_encoding)
-        encoded = read_stata(self.dta_encoding, encoding="latin-1")
+        with warnings.catch_warnings(record=True) as w:
+            encoded = read_stata(self.dta_encoding, encoding='latin-1')
+            assert len(w) == 1
```
gfyoung: Hmmm...I was thinking for that you use `tm.assert_produces_warning` or `pytest.warns`.

pandas/tests/io/test_stata.py
```diff
@@ -361,16 +361,22 @@ def test_encoding(self, version):
 
         # GH 4626, proper encoding handling
         raw = read_stata(self.dta_encoding)
-        encoded = read_stata(self.dta_encoding, encoding="latin-1")
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            encoded = read_stata(self.dta_encoding, encoding='latin-1')
+            assert len(w) == 1
```
gfyoung: (repost since GitHub hid it due to recent push): Hmmm...I was thinking that you use tm.assert_produces_warning or pytest.warns so that we explicitly ensure that the warning is produced.

jorisvandenbossche: Yes, I think this can be:

```
    encoded = read_stata(self.dta_encoding, encoding='latin-1')
```

(and the same for the other occurrence)

gfyoung (approved):
@bashtage : For some reason, I missed the notification that you had patched the warnings setup in the tests.  LGTM!

cc @jorisvandenbossche

jreback:
thanks @bashtage

hmgaudecker:
I cannot check right now, but I do not think that the sentence " The encoding depends only on the file format and cannot be changed by users." is quite true. If I remember correctly, pre-118 Stata files used, in an undocumented fashion, the native encoding, Latin-1 on Windows and MacRoman on OS X, I don't remember for Linux. Officially, special characters were not supported. 

I think there is a case to be made for the encoding keyword, but with old versions dying out I am not sure whether it is worth the effort.

Issue #21354, opened 2018-06-07
CI: Enable spellcheck in docbuild


TomAugspurger:
Followup to https://github.com/pandas-dev/pandas/pull/21109

We'll need to refactor `scripts/announce.py` to update the list of contributors on each build.

Then we can enable spellcheck in the Travis doc build.

FabioRosado:
I would like to take on this since this is related to my previous PR 😄

FabioRosado:
@daminisatya Apologies I had this tab open and your comment didnt show but you beat me to it so if you would like to take on this issue go ahead 😄 👍

WillAyd:
@daminisatya since you marked interest in this and the JSON issue how about you tackle the JSON one as a first contribution and let @FabioRosado continue with this one, since this is a logical follow up to a change he just made?

daminisatya:
Yes @WillAyd, I was about to comment the same :) @FabioRosado you can go ahead with the issue since you have already solved it you will have a better context than me! I took up another first-timer issue.

FabioRosado:
I am having a bit of issue with this script, when I run the command `./announce.py $GITHUB v1.11.0..v1.11.1 > announce.rst` from the script folder I get the error message `announce.py: error: unrecognized arguments: v1.11.0..v1.11.1` but the example given in the file says that this should work.

Is the script broken at the moment or is just some issue that I am having?

afeld:
Update: Appears this was added and then subsequently removed. In https://github.com/pandas-dev/pandas/pull/24287, the comment that:

> The spellcheck system was implemented, but with dependencies (i.e. `enchant`) that are no longer maintained, and there is no reasonable way to set them up.

doesn't seem to be true anymore. I'm seeing releases for both [enchant](https://github.com/AbiWord/enchant/releases) and [pyenchant](https://pypi.org/project/pyenchant/#history) in December 2020. In other words, I think we should consider adding spellcheck back in.

afeld:
Currently we run [`codespell`](https://github.com/codespell-project/codespell) on the source code, minus the test suite. 

https://github.com/pandas-dev/pandas/blob/00ddacfa191734adcd7a062e336e5625ea65605f/.pre-commit-config.yaml#L202-L205

We could expand it to the docs and see how it compares…?

MarcoGorelli:
@afeld yes, there's an open issue about that #38802

afeld:
Seems like the issues are duplicates, so closing this one for now.

PR #21402, opened 2018-06-09
CI: Enable Spellcheck in dockbuild

### LINKED ISSUE #21396: Missing dependency for sphinxcontrib-spelling

FabioRosado:
This is the initial PR to try and fix the issue #21354 the sphinx spelling extension needs every name to be divided and added to each line in order to get the name marked as spelt correctly. The file `names_wordlist.txt` is updated on every run of the function `get_authors()` located inside`scripts/announce.py`.

Since the issue #21396 is related to my previous PR I attempted to resolve this issue here as well, I followed the suggestion of @jorisvandenbossche and added a try/except to the `conf.py` if the dependency doesn't exist then nothing will happen.

@datapythonista since you submitted a PR(#21397) to try and fix the issue, perhaps you have an opinion about this bit of code. Also, if you would like to do the changes on your own PR I will just delete this bit and use your solution 😄 👍 

I would like some feedback on this first attempt, the generator `[names.extend(re.sub('\W+', ' ', x).split()) for x in cur.union(pre)]` seems a bit smelly but does the trick, should this be refactored to use a long form and make it more readable? I extended the list to names because split() was creating a list inside the main list and this makes every name to be split and added to the same list (this is probably very memory inefficient though)

I tried to run the script from my machine but I got an issue with it, probably I'm doing something wrong but when I run `./scripts/announce.py $GITHUB v0.15.1..v0.23.0` I get the error message:

```
usage: announce.py [-h] [--repo REPO] revision_range 
announce.py: error: unrecognized arguments: v0.15.1..v0.23.0
```

Not sure what I'm doing wrong, to be honest. If I run just the `get_authors()` function everything seems to be fine (including the v0.15.1..v0.23.0 format) and the names_wordlist is updated successfully.

I will look forward to read your opinions about this PR 👍 
----
- [ ] closes #21396 #21354
- [ ] tests added / passed
- [ ] whatsnew entry


scripts/announce.py
```diff
@@ -67,6 +67,17 @@ def get_authors(revision_range):
     cur.discard('Homu')
     pre.discard('Homu')
 
+    # Update doc/source/names_wordlist.txt with the names of every author
+    names = []
+    [names.extend(re.sub('\W+', ' ', x).split()) for x in cur.union(pre)]
```
WillAyd: Could use a generator expression instead of a list comprehension since you only iterate over `names` once

scripts/announce.py
```diff
@@ -67,6 +67,17 @@ def get_authors(revision_range):
...
+
+    path = os.path.sep.join(os.path.abspath(__file__).split(os.path.sep)[:-2])
```
WillAyd: Can use `os.path.dirname` to go up a couple directories or alternately `os.path.join` using `../..` (latter is more succinct though not completely sure of portability)

WillAyd (changes requested):

doc/source/conf.py
```diff
@@ -73,13 +73,18 @@
...
 
-spelling_word_list_filename = ['spelling_wordlist.txt', 'names_wordlist.txt']
-spelling_ignore_pypi_package_names = True
+# Tries to import sphixcontrib.spelling - if not available pass
+try:
+    import sphinxcontrib.spelling
+    extensions.append('sphinxcontrib.spelling')
+    spelling_word_list_filename = ['spelling_wordlist.txt', 'names_wordlist.txt']
+    spelling_ignore_pypi_package_names = True
+except ModuleNotFoundError:
```
jorisvandenbossche: Can you also catch ImportError? 
Because if you have installed `sphinxcontrib.spelling`, but not pyenchant, you get this (which will be a common case I think, since this is what you get if you simply pip/conda install it)

doc/source/conf.py
```diff
@@ -73,13 +73,18 @@
...
+    pass
```
jorisvandenbossche: Maybe we can print a message about no spell-check being done because it is not installed?

jorisvandenbossche (commented):
@FabioRosado do you plan to add more to this PR to actually enable it on CI? 
In that case, I would quickly submit another PR with just the try except change in conf.py if you want (so at least the doc build works again)

FabioRosado:
Heya yeah that is my plan but I have some busy days at work I plan to work on this tomorrow or Wednesday. 

I appreciate your help @jorisvandenbossche I need to check datapythonista’s PR and see if he changed anything and then I’ll check if I need to add something or not. 
Also, good point on the `ImportError` I didn’t thought about that. 

I was planning to add some sort of message saying why it failed but I wasn’t sure if I should add a print or some sort of logging (I just need to check what is used in pandas)

datapythonista:
@FabioRosado, I can work on the simple PR to make the documentation build work again later this afternoon. Then you can check for a better solution when you have time.

FabioRosado:
@datapythonista  that sounds awesome I didn't want to step in on your PR that's why I added that bit as a thing that could be done in case you did it differently haha my schedule just changed tomorrow so I will have to work on this on Wednesday :/

datapythonista:
No worries. Just updated the PR so the documentation can be built when the dependencies for the spellcheck are not present, as this is kind of urgent. Then, with no rush and whenever you have time, you can work on a good approach to manage the dependencies for the spellcheck.

FabioRosado:
I would like to apologise for the delay on this PR I have been a bit busy with work.
The changes submitted by datapythonista look great and I have just added the bit that makes spellcheck work (the wordlists and whatnot).

On `scripts/announce.py` I refactored the list comprehension and used a single for loop to remove any special characters and split the name into spaces and then write it to the file - I believe this should be okay since I would have to iterate over every single name as well, perhaps this will be a bit more efficient way to do it?

I got a question that I would like to have it clarified. The function `get_authors()` takes the `revision_range` param in the form of 'v0.22.01..v0.23.0' but I have noticed that  if I try to pass the latest versions located in `whatsnew`  I get an error `unknown revision or path not in the working tree.` 

The latest version I can pass to get the authors is `v0.23.0` so version v0.23.1, v0.23.2 and v0.24.0 won't be able to be used at the moment. I can't understand why though, is it because these versions haven't been tagged yet or something? git shortlog doesn't seem to find these versions in the logs.

Finally, I need to add this function to CI in order to update the names_wordlist, I'm can hardcode the earliest version but I'm not sure how to get the latest one, any ideas how to best do this? Thank you

scripts/announce.py
```diff
@@ -73,6 +73,32 @@ def get_authors(revision_range):
...
+
+    # Add author names to set from version 0.3.0 until latest version
+    for author in authors:
+        all_names = all_names.union(re.sub('[.+\-!?]', ' ', author).split())
+
+    with open(wordlist_path, 'w', encoding='utf-8') as wordlist:
+        for name in all_names:
+            wordlist.write('{}\n'.format(name))
+
+update_name_wordlist()
```
FabioRosado: I forgot to remove this from my code will fix on the next commit

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -18,7 +18,7 @@ Fixed Regressions
 **Comparing Series with datetime.date**
 
 We've reverted a 0.23.0 change to comparing a :class:`Series` holding datetimes and a ``datetime.date`` object (:issue:`21152`).
-In pandas 0.22 and earlier, comparing a Series holding datetimes and ``datetime.date`` objects would coerce the ``datetime.date`` to a datetime before comapring.
+In pandas 0.22 and earlier, comparing a Series holding datetimes and ``datetime.date`` objects would coerce the ``datetime.date`` to a datetime before comparing.
```
WillAyd: Don't think you want this whatsnew or the one for 24

FabioRosado: I was testing the spellcheck to see if all the names were marked as correct, on the process of the spellcheck it marked this and the other whatsnew with some grammatical errors, so I checked and it seems that they should be fixed that's why I included it here - sorry for the confusion since this change wasn't really related to the PR I could revert the changes and submit a new one with just those changes if you would like.

WillAyd: Ah OK - it's simple enough that I think that's fine. It just wasn't readily apparent to me that's why those notes were included but that makes sense.

FabioRosado: I understand I should have explained in my first comment. Could you help me with the CI integration, would it suffice to write a test and run the spellcheck and assert that no exception is raised?

ci/build_docs.sh
```diff
@@ -69,6 +69,10 @@ if [ "$DOC" ]; then
            pandas/core/reshape/reshape.py \
            pandas/core/reshape/tile.py
 
+    echo "Running spellcheck on documentation"
```
FabioRosado: I tried to follow the example on line 35 where `make.py` is being called. If I'm not mistaken we need to cd into the doc folder before running the make.py spellcheck since the script cd back to `$TRAVIS_BUILD_DIR`.

This is the first time I tried to write something like this so I'd like your opinion about it and if enabling the spellchheck within CI should be done in a different way.

Thank you

WillAyd: Why don't you just move this closer to line 35 so you don't need to keep changing directories?

FabioRosado:
I added the code down there because I thought I needed to let the whole script run and then do the spell check at the end. I have moved that bit closer to `./make.py` so this should run once the docs are build

doc/source/names_wordlist.txt
```diff
@@ -1,1652 +1,1987 @@
-Critchley
-Villanova
```
jreback: do we need this list at all? (as you are now getting it from the authors)?

jreback (changes requested):

FabioRosado:
Yes we do need the list, like I said on my previous PR, the Sphinx extension reads the words to be whitelisted from a txt file and currently there is no way to use stdout to whitelist words :smile:

doc/source/names_wordlist.txt
FabioRosado: Yeah the list is build when spellcheck is called. I have updated the spellcheck function located in make.py to call the function get_name_wordlist() which is located in scripts/announce.py

jreback: ok, so let's delete the list then.

jorisvandenbossche:
The doc build on Travis is failing with "ModuleNotFoundError: No module named 'git'" (check the last build to see the log output of the doc build)

doc/source/names_wordlist.txt
FabioRosado: Unfortunately, we still need the list, the only reason why we do need it is due to the fact that there are names(around 5-7) in the file `release.rst` that are not author names (or at least they are different so they don't get added when we run the function `get_names_wordlist`. 

If these names are not added somewhere then the build will always fail. If you are worried about having a file with all the names of every contributor I could just delete most of them and add only those in `release.rst` since the file will be updated with all the names.

Perhaps this file could be good to have just in case people need to add their own names (like they use a different one or something and it gets flagged as a spelling error like those in `release.rst`), so with this file people would know that they have to add those names in there and not in the `spelling_wordlist.txt`, what do you think?

FabioRosado:
@jorisvandenbossche The build that failed is that #21511 or the build from my previous commit? If it was the latter I was getting a problem from the lint script, but removing the `ModuleNotFoundError` from the try/except (when trying to import the spelling library) seem to fix the issue.

But now that you mention that, i'm importing the function from announce.py which uses `gitpython`
and `pygithub`, I guess this could potentially be a problem, I'll try to import these dependencies along with sphinx spelling and just log if they weren't found

doc/source/names_wordlist.txt
jorisvandenbossche: > I could just delete most of them 

Yes, I would only keep those that are not generated by the script (so the ones that are needed to get the script running)

jorisvandenbossche:
> The build that failed is that #21511 or the build from my previous commit?

No, it is from this PR. Each PR (and each commit that you add to a PR) is tested on travis. You can follow the link to that by clicking on the "All checks have passed" -> "Show all checks". 

Note that it is not actually failing (travis will show it as green), but you can see in the doc build log that it is actually failing. Eg for the last commit added here: https://travis-ci.org/pandas-dev/pandas/jobs/393561877

The reason this is failing is simply because the correct dependencies are not installed on travis (so the installation script needs to be updated, need to check where you can add this)

> I'll try to import these dependencies along with sphinx spelling and just log if they weren't found.

Yes, that's a good idea. 
Although that makes yet another optional dependency for a dev environment ...

jorisvandenbossche:
BTW, I think a very valuable contribution would be to add a feature to sphinxcontrib-spelling to skip a file ... (that would solve many problems here!)

jorisvandenbossche:
> BTW, I think a very valuable contribution would be to add a feature to sphinxcontrib-spelling to skip a file ... (that would solve many problems here!)

And it might be as simple as skipping based on the `docname` in https://github.com/sphinx-contrib/spelling/blob/fa8c4b6140d7be0dcaa2aefbf690b84090bb0884/sphinxcontrib/spelling/builder.py#L144

FabioRosado:
@jorisvandenbossche  Yeah you are right that would make things so much easier, I'll try to submit a PR to the spelling extension to add the possibility to avoid certain files. 

I guess this PR will have to either be placed on hold or closed since if this gets merged then the build will eventually fail when new names are added - unless we add another dependency to make this work on the build.

I will wait until I hear from you as to what is the best course of action with these issues 👍

doc/source/names_wordlist.txt
jorisvandenbossche: > Yes, I would only keep those that are not generated by the script (so the ones that are needed to get the script running)

Actually, another option is to have them all in there, and have the script update it on demand (so if the release manager adds the names to the notes, the scripts needs to be run). That would save the extra step of creating it on each doc build.

jorisvandenbossche:
> 'll try to submit a PR to the spelling extension to add the possibility to avoid certain files.

That would be really cool!

> I guess this PR will have to either be placed on hold or closed

I think we can go ahead with this for now, as long as the script is there to update the list of names to not have the build fail. We can later remove that again if the feature to ignore certain files is added.

FabioRosado:
I had to remove ModuleNotFoundError since linting kept marking this as non-existent as this except is only available on Python 3.6, I have updated the name list back to all the names as well.

I did leave the reference to the function that updates the wordlist on the spellcheck function, but if you run the announce.py script the list will also be updated (since its not called from `main`). Hope this is okay.

Will look forward to fixing some issue here once I figure out how to avoid files/folders in the spelling extension haha

jreback:
can you rebase / update

FabioRosado:
I thought this PR wasn't accepted as there was some different opinions about it

jreback:
rereading i think this is ok if we can just auto generate the list of authors from the commit history rather than hard coding everything

datapythonista:
@jreback @jorisvandenbossche are you happy merging this if @FabioRosado rebases?

My opinion is that the whole typo validation adds too much complexity to the project. And that being based in enchant, which is a deprecated project, it even makes less sense to have it. We can surely have a separate repo in the pandas org that checks in the pandas repo whether there is any typo with all this system we have, that would be very useful. But having it in pandas, when afaik none of the maintainers was able to set it up, doesn't seem to be worth.

In any case, can you let @FabioRosado if he should rebase so we can merge it, or just close the PR. Thanks!

PR #21405, opened 2018-06-10
parametrize tests, unify repeated tests

jbrockmendel:
There's more to do, but the diff is already pretty big as it is.

There are a handful of test methods that are defined in 5-6 test classes.  By moving those to the base class we a) get rid of duplicate code and b) run the tests for all (well, most) `DateOffset` subclasses.

### DIFF
--- a/pandas/tests/tseries/offsets/test_offsets.py
+++ b/pandas/tests/tseries/offsets/test_offsets.py
@@ -89,6 +89,7 @@ def test_to_m8():
 
 class Base(object):
     _offset = None
+    d = Timestamp(datetime(2008, 1, 2))
 
     timezones = [None, 'UTC', 'Asia/Tokyo', 'US/Eastern',
                  'dateutil/Asia/Tokyo', 'dateutil/US/Pacific']
@@ -149,7 +150,7 @@ class Base(object):
             pass
 
     def test_offsets_compare_equal(self):
-        # root cause of GH#456
+        # root cause of GH#456: __ne__ was not implemented
         if self._offset is None:
             return
         offset1 = self._offset()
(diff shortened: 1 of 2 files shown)


pandas/tests/tseries/offsets/test_offsets.py
```diff
@@ -148,6 +148,50 @@ def test_apply_out_of_range(self, tz):
             # so ignore
             pass
 
+    def test_offsets_compare_equal(self):
+        # root cause of GH#456
```
jreback: can you add a more informative comment

pandas/tests/tseries/offsets/test_offsets.py
```diff
@@ -148,6 +148,50 @@ def test_apply_out_of_range(self, tz):
...
+        if self._offset is None:
+            return
+        offset1 = self._offset()
+        offset2 = self._offset()
+        assert not offset1 != offset2
+        assert offset1 == offset2
+
+    def test_rsub(self):
+        if self._offset is None or not hasattr(self, 'd'):
+            # TODO: define `d` for classes where it is missing?
```
jreback: can you expand this coment (and similar fo radd)

pandas/tests/tseries/offsets/test_offsets.py
```diff
@@ -1424,33 +1429,18 @@ def testEQ(self):
         assert (CustomBusinessHour(holidays=['2014-06-27']) !=
                 CustomBusinessHour(holidays=['2014-06-28']))
 
+    def test_sub(self):
+        # override Base implementation
```
jreback: can you provide a better comment here. why is skipped?

jreback (changes requested):
looks good. some comment requests. ping on green.

jreback (approved):

jreback:
thanks @jbrockmendel

PR #21413, opened 2018-06-10
add dropna=False to crosstab example

testvinder:
```
>>> foo = pd.Categorical(['a', 'b'], categories=['a', 'b', 'c'])
>>> bar = pd.Categorical(['d', 'e'], categories=['d', 'e', 'f'])
>>> crosstab(foo, bar)  # 'c' and 'f' are not represented in the data,
    ...                     # but they still will be counted in the output
col_0  d  e  f
row_0
a      1  0  0
b      0  1  0
c      0  0  0
```

The above example code does not produce the output shown because dropna=True is default. Changing crosstab(foo, bar) to crosstab(foo, bar, dropna=False) fixes that and produces the shown output (which is also the expected and correct output).

- [ ] closes #xxxx
- [x] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/reshape/pivot.py
+++ b/pandas/core/reshape/pivot.py
@@ -446,7 +446,18 @@ def crosstab(index, columns, values=None, rownames=None, colnames=None,
     >>> foo = pd.Categorical(['a', 'b'], categories=['a', 'b', 'c'])
     >>> bar = pd.Categorical(['d', 'e'], categories=['d', 'e', 'f'])
     >>> crosstab(foo, bar)  # 'c' and 'f' are not represented in the data,
-    ...                     # but they still will be counted in the output
+                            # and will not be shown in the output because
+                            # dropna is True by default. Set 'dropna=False'
+                            # to preserve categories with no data
+    ... # doctest: +SKIP
+    col_0  d  e
+    row_0
+    a      1  0
+    b      0  1
+
+    >>> crosstab(foo, bar, dropna=False)  # 'c' and 'f' are not represented
+                            # in the data, but they still will be counted
+                            # and shown in the output
     ... # doctest: +SKIP
     col_0  d  e  f
     row_0


pandas/core/reshape/pivot.py
```diff
@@ -429,8 +429,9 @@ def crosstab(index, columns, values=None, rownames=None, colnames=None,
 
     >>> foo = pd.Categorical(['a', 'b'], categories=['a', 'b', 'c'])
     >>> bar = pd.Categorical(['d', 'e'], categories=['d', 'e', 'f'])
-    >>> crosstab(foo, bar)  # 'c' and 'f' are not represented in the data,
-    ...                     # but they still will be counted in the output
+    >>> crosstab(foo, bar, dropna=False)  # 'c' and 'f' are not represented
```
uds5501: @testvinder , just a suggestion, you could've actually just changed the output regarding the given code and mentioned in comments about the default value of `dropna`being set to `True`

uds5501 (commented):
Please refer to the comments. 

@gfyoung , are the suggestions viable?

gfyoung:
@testvinder : Thanks for this PR!  In general, we actual prefer that you open an issue first before proceeding with a PR just so that we can evaluate whether a PR is needed.  This issue that you bring up does ring a bell, but I'm not sure how we dealt with it before.

cc @jreback

pandas/core/reshape/pivot.py
gfyoung: I would actually just add an example instead of modifying this one.

gfyoung:
I completely misread this PR and thought this was a larger change than it actually was.  Sorry!

pandas/core/reshape/pivot.py
testvinder: @uds5501 I actually think it's better to set ´dropna` to `False`. The whole point of specifying the extra categories is to include them in the crosstab. It's a small change but the original is very confusing because the output cannot be reproduced.

testvinder:
@gfyoung yes, it's really just a tiny change that makes the documentation more clear.

pandas/core/reshape/pivot.py
gfyoung: @testvinder : Ah, yes, your point is well-taken.  Can you update the original example with the correct output and add your example with `dropna=False` ?

testvinder: @gfyoung Sure, will do.

gfyoung:
@testvinder : I just realized: your PR should be against the `master` branch, not `0.23.x`.  Could you fix that?

testvinder:
@gfyoung Done:-) Sorry for the mistake. This is all new to me.

gfyoung:
@testvinder : I re-applied your commits against `master` because the base-switching was a lot uglier than I was hoping it would be.

gfyoung (approved):
With the base fixed, LGTM!

cc @jreback

jreback:
thanks @testvinder

jorisvandenbossche:
@jreback just a small note, can you make sure the final commits gets a reasonable message? (normally you should see it in the github merge box), because now it is "Reapply all patches by @testvinder against master" :-)

jreback:
normally i remove all comments - guess missed that one

jorisvandenbossche:
Yeah, it's not about the comments (those go into the commit message body) but the title. But it's a bit strange github did it this wrongly, because normally the title is fine (probably because the commit was rebased and squashed)

PR #18882, opened 2017-12-20
API: disallow duplicate level names

toobaz:
- [x] closes #18872
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

PR #21423, opened 2018-06-11
API: re-allow duplicate index level names

jorisvandenbossche:
One possible solution for https://github.com/pandas-dev/pandas/issues/19029

WIP (need to clean up tests and possibly re-add some ones that have been removed in https://github.com/pandas-dev/pandas/pull/18882)


TomAugspurger:
+1 in general.

Will want a semi-detailed release not, potentially laying out future plans?

jorisvandenbossche:
> Will want a semi-detailed release note, potentially laying out future plans with respect to duplicate names?

Until we don't exactly know what those future plans are, I would rather keep it limited (only noting that duplicated level names are allowed again?)

jorisvandenbossche:
@jreback @toobaz ready for review

I went through the tests and added back some of the tests removed in https://github.com/pandas-dev/pandas/pull/18882

I also added back `MultiIndex._reference_duplicate_name` to use it in `stack`/`unstack` (so we can there raise an informative error message). However, I would personally prefer that `MultiIndex._get_level_number` directly raises this informative warning, then the extra checking in `stack`/`unstack` can be removed. However, that is a slight change in behaviour for eg `get_level_values` (KeyError -> ValueError), so I would rather leave that for 0.24.0.

jorisvandenbossche:
> However, I would personally prefer that MultiIndex._get_level_number directly raises this informative warning

That was also the original intent of:

https://github.com/pandas-dev/pandas/blob/a620e7255ce896c9185275d82172513268b1a719/pandas/core/indexes/multi.py#L771-L776

where already an informative ValueError is raised, but now the ValueError is catched, and then a generic KeyError is raised (this was changed 5 years ago: https://github.com/pandas-dev/pandas/commit/bdf270c77ef8c84a04fcf25f805b56b00e6725c9#diff-dc96e62847cdfe77b570cf9c6b0e8086R1536)

TomAugspurger:
> Until we don't exactly know what those future plans are, I would rather keep it limited (only noting that duplicated level names are allowed again?)

Agreed.

pandas/core/reshape/reshape.py
```diff
@@ -115,6 +115,12 @@ def __init__(self, values, index, level=-1, value_columns=None,
 
         self.index = index.remove_unused_levels()
 
+        if isinstance(self.index, MultiIndex):
```
jreback: I would rather not add API here. why don't you just do this check inside ``_get_level_number`` itself, and then also add to ``Index`` so you don't also have to add an instance check?

jreback (changes requested):

pandas/core/reshape/reshape.py
jorisvandenbossche: Did you read my comment in the main thread? 
This code above is exactly how it was before. As I said, I also think this is better in het_level_number, but propose to do a follow-up PR that can be merged for 0.24.0, as it gives an additional API change (change in error type)

jreback: then let's just wait till 0.24.0 to do this.

TomAugspurger: It seems like a good fix for the regression, why not merge it? It's no different that reverting a previous commit.

jorisvandenbossche: yes, it is how it was for a long time, and this is fixing a regression so it is the point to do it now and not wait for 0.24.0.

I can certainly do the PR for 0.24.0 to further clean this up, but that will involve an API change so I will not do that in this PR.

jreback (approved):

jreback:
fair enough, is there an issue to clean this up?

jorisvandenbossche:
@jreback I am doing a PR, but ran into some issues: https://github.com/pandas-dev/pandas/issues/21677

jorisvandenbossche:
Opened the PR to not forget his (https://github.com/pandas-dev/pandas/pull/21678), but as mentioned, we first need to discuss the other issue.

PR #21427, opened 2018-06-11
disallow normalize=True with Tick classes

### LINKED ISSUE #21434: Tick with normalize=True should be disallowed
See #21427 

The issue is that a (non-`Day`) `Tick` object with `normalize=True` breaks addition monotonicity/associativity:

```
now = pd.Timestamp.now()
tick = pd.offsets.Minute(n=3, normalize=True)

>>> now
Timestamp('2018-06-11 17:38:10.135400')
>>> now + tick
Timestamp('2018-06-11 00:00:00')

>>> now + 64*tick + 64*tick
Timestamp('2018-06-11 00:00:00')
>>> now + 128*tick
Timestamp('2018-06-12 00:00:00')
```

The `Day` tick doesn't have the same problem, but I'd prefer to disallow it rather than make it a double-special-case.  (See #20633 for more discussion about how `Day` should behave)

jbrockmendel:
The problem: allowing `Tick` objects with `normalize=True` causes addition to lose monotonicity/associativity

```
ts = pd.Timestamp.now()
tick = pd.offsets.Minute(n=4, normalize=True)
>>> ts
Timestamp('2018-06-11 10:50:14.419655')
>>> ts + tick
Timestamp('2018-06-11 00:00:00')
```

- [x] closes #21434
- [x] tests added/passed
- [x] passes flake8
- [x] whatsnew note


pandas/tseries/offsets.py
```diff
@@ -2219,6 +2219,9 @@ class Tick(SingleConstructorOffset):
     def __init__(self, n=1, normalize=False):
         # TODO: do Tick classes with normalize=True make sense?
```
gfyoung: It seems like this was a broader API design question.  Depending on how things go, creating an issue to "close" with this PR might be appropriate.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -112,7 +112,7 @@ Timezones
 Offsets
 ^^^^^^^
 
--
+- :class:`Tick` subclasses can no longer be created with the argument `normalize=True` (:issue:`????`)
```
gfyoung: In cases like these, reference the PR itself (unless an issue is created as pointed out <a href="https://github.com/pandas-dev/pandas/pull/21427#discussion_r194555863">here</a>).

gfyoung: The same goes for the test case (i.e. add a comment referencing the PR).

gfyoung:
@jbrockmendel : If this is the correct way to fix (I'm not 100% sure @jreback ), then your PR is good to go.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -112,7 +112,7 @@ Timezones
 Offsets
 ^^^^^^^
 
--
+- :class:`Tick` subclasses can no longer be created with the argument `normalize=True` (:issue:`21427`)
```
jreback: double backticks here. I would be a little more informative (say for example Minute, Second).

jreback (changes requested):
is there an associated issue?

gfyoung:
> is there an associated issue?

@jreback : No, there isn't AFAIK.  I suggested to @jbrockmendel that one be opened only if there was disagreement about the fix.

doc/source/whatsnew/v0.24.0.txt
jbrockmendel: Will do.

jbrockmendel:
Opened #21434, extends the example given in the OP and mentions that `Day` could be a special case (though I'd prefer it not)

jreback:
@jbrockmendel you didn’t need to make an issue, but ok

can u make a sub section with an example in whatsnew

jbrockmendel:
Just added a whatsnew section, not sure it adds much on top of the one-line note (also fleshed out)

pandas/tseries/offsets.py
gfyoung: @jbrockmendel : Could you remove this comment?  Looks like your fix is here to say. 😄

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -26,6 +26,33 @@ Backwards incompatible API changes
...
+--------------------------------------
+
+Creating a ``Tick`` object (:class:``Day``, :class:``Hour``, :class:``Minute``,
+:class:``Second``, :class:``Milli``, :class:``Micro``, :class:``Nano``) with
+`normalize=True` is no longer supported.  This prevents unexpected behavior
+where addition could fail to be monotone or associative.
+
+.. ipython:: python
+
+    ts = pd.Timestamp('2018-06-11 18:01:14')
```
jreback: show ts and tic here

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -26,6 +26,33 @@ Backwards incompatible API changes
...
+    tic = pd.offsets.Hour(n=2, normalize=True)
+
+Previous Behavior:
+
+.. code-block:: ipython
+
+    In [2]: ts = pd.Timestamp('2018-06-11 18:01:14')
```
jreback: you don't need to repeat ts & tic here, just the calculations

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -26,6 +26,33 @@ Backwards incompatible API changes
...
+
+    In [3]: tic = pd.offsets.Hour(n=2, normalize=True)
+
+    In [4]: ts + tic
+    Out [4]: Timestamp('2018-06-11 00:00:00')
+
+    In [5]: ts + tic + tic + tic == ts + (tic + tic + tic)
+    Out [5]: False
+
```
jreback: add the Current  behavior via an ipython block (obviously w/o normalize passed)

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -26,6 +26,33 @@ Backwards incompatible API changes
 
 .. _whatsnew_0240.api.datetimelike:
```
jreback: add a ref to this section (this is a new one, like: _whatsnew_0240.api.datetimelike.normalize )

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -112,7 +139,7 @@ Timezones
 Offsets
 ^^^^^^^
 
--
```
jreback: if you have a subsection, you don't need a note, the subsection IS the note.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -26,6 +26,33 @@ Backwards incompatible API changes
 
 .. _whatsnew_0240.api.datetimelike:
 
+Tick DateOffset Normalize Restrictions
+--------------------------------------
+
+Creating a ``Tick`` object (:class:``Day``, :class:``Hour``, :class:``Minute``,
+:class:``Second``, :class:``Milli``, :class:``Micro``, :class:``Nano``) with
+`normalize=True` is no longer supported.  This prevents unexpected behavior
+where addition could fail to be monotone or associative.
```
jreback: add the issue number here

jreback (changes requested):

doc/source/whatsnew/v0.24.0.txt
jbrockmendel: Will do.  Does showing the tic here work given that the line that defines it will raise in the new version?

doc/source/whatsnew/v0.24.0.txt
jbrockmendel: OK.  And just the same `ts + tic + tic + tic == ts + (tic + tic + tic)` demonstration?  With `normalize=False` that behavior is unchanged.

jreback (approved):

jreback:
thanks @jbrockmendel

PR #21431, opened 2018-06-11
 BUG: Nrows cannot be zero for read_csv. Fixes #21141

### LINKED ISSUE #21141: read_csv errors when low_memory=True, index_col is not None, and nrows=0

cgopalan:
- [x] closes #21141
- [x] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/tests/io/parser/common.py
+++ b/pandas/tests/io/parser/common.py
@@ -364,7 +364,7 @@ bar,foo"""
         df = self.read_csv(StringIO(self.data1), nrows=3.0)
         tm.assert_frame_equal(df, expected)
 
-        msg = r"'nrows' must be an integer >=0"
+        msg = r"'nrows' must be an integer >=1"
 
         with tm.assert_raises_regex(ValueError, msg):
             self.read_csv(StringIO(self.data1), nrows=1.2)
@@ -374,6 +374,9 @@ bar,foo"""
 
         with tm.assert_raises_regex(ValueError, msg):
             self.read_csv(StringIO(self.data1), nrows=-1)
+
+        with tm.assert_raises_regex(ValueError, msg):
+            self.read_csv(StringIO(self.data1), nrows=0)
 
     def test_read_chunksize(self):
         reader = self.read_csv(StringIO(self.data1), index_col=0, chunksize=2)
--- a/pandas/tests/io/test_excel.py
+++ b/pandas/tests/io/test_excel.py
@@ -1001,7 +1001,7 @@ class TestXlrdReader(ReadingTestsBase):
                           nrows='5')
 
     def test_read_excel_nrows_zero_parameter(self, ext):
-        # GH 16645
+        # GH 21141
         msg = "'nrows' must be an integer >=1"
         with tm.assert_raises_regex(ValueError, msg):
             pd.read_excel(os.path.join(self.dirpath, 'test1' + ext),
--- a/pandas/tests/io/test_excel.py
+++ b/pandas/tests/io/test_excel.py
@@ -995,10 +995,17 @@ class TestXlrdReader(ReadingTestsBase):
 
     def test_read_excel_nrows_non_integer_parameter(self, ext):
         # GH 16645
-        msg = "'nrows' must be an integer >=0"
+        msg = "'nrows' must be an integer >=1"
         with tm.assert_raises_regex(ValueError, msg):
             pd.read_excel(os.path.join(self.dirpath, 'test1' + ext),
                           nrows='5')
+
+    def test_read_excel_nrows_zero_parameter(self, ext):
+        # GH 16645
+        msg = "'nrows' must be an integer >=1"
+        with tm.assert_raises_regex(ValueError, msg):
+            pd.read_excel(os.path.join(self.dirpath, 'test1' + ext),
+                          nrows=0)
 
     def test_read_excel_squeeze(self, ext):
         # GH 12157
--- a/pandas/io/parsers.py
+++ b/pandas/io/parsers.py
@@ -1027,7 +1027,7 @@ class TextFileReader(BaseIterator):
         raise com.AbstractMethodError(self)
 
     def read(self, nrows=None):
-        nrows = _validate_integer('nrows', nrows)
+        nrows = _validate_integer('nrows', nrows, min_val=1)
 
         if nrows is not None:
             if self.options.get('skipfooter'):


gfyoung:
@jreback : Do we need to deprecate `nrow=0` before disallowing?  It's a (small) corner case, which is why I ask about it.

pandas/tests/io/test_excel.py
```diff
@@ -995,11 +995,18 @@ def test_read_excel_nrows_greater_than_nrows_in_file(self, ext):
 
     def test_read_excel_nrows_non_integer_parameter(self, ext):
```
WillAyd: Would be better to parametrize this test for cases like `[-1, 0, '1']` instead of having separate tests

WillAyd (changes requested):

pandas/tests/io/parser/common.py
```diff
@@ -375,6 +375,9 @@ def test_read_nrows(self):
         with tm.assert_raises_regex(ValueError, msg):
             self.read_csv(StringIO(self.data1), nrows=-1)
 
+        with tm.assert_raises_regex(ValueError, msg):
```
gfyoung: Can you reference the issue number in a comment above this line.

jschendel:
I'm not sure about disallowing `nrows=0`.  It's generally useful when you need to dynamically read a subset of columns, where you do a two pass procedure along the lines of:

1. Read in the columns only, e.g. `all_cols = pd.read_csv('file.csv', nrows=0).columns`
2. Dynamically filter the columns, e.g. `good_cols = [c for c in all_cols if c.startswith('foo')]`
3. Read in only the desired columns, e.g. `df = pd.read_csv('file.csv', usecols=good_cols)`

I doubt you lose much performance by using `nrows=1` instead, so it's not the end of the world if this does get removed.  However, there are some established StackOverflow answers with `nrows=0` similar to my procedure above, so I imagine there are some users that have `nrows=0` in their codebase.  If this is going to be removed, it definitely seems like it should be deprecated first.

gfyoung:
@jschendel : Good to know.  In the case of your example, couldn't you just read the first line of the file using the builtin `open` ?  Don't really need `pandas` for that. 🤷‍♂️

jschendel:
Yeah, I'm not saying that it's the _right_ solution, just that it's a solution people are likely using.  Certainly more concise than using `open` if you method chain to do the filtering.  I wouldn't complain if this was removed, but doing so without deprecating seems likely to break things for users.

gfyoung:
> I wouldn't complain if this was removed, but doing so without deprecating seems likely to break things for users.

jreback (changes requested):
nrows=0 is valid
the combinations of options is invalid

pr needs to limit scope

gfyoung:
> nrows=0 is valid
the combinations of options is invalid

@jreback : `nrows=0` has limited use and was a clean way to patch the corner case (though it seems like @jschendel has a case to keep it).  In any case, should we just disallow that combination of inputs then?

cgopalan:
@jreback @gfyoung i too agree that just the value of `low_memory` should not render the combination of values invalid. Because of this corner case, this is actually exhibiting weird behavior on an unrelated set of conditions. Better to suppress this corner case by removing/deprecating `nrows=0`.
Is there a way to deprecate just a value for an arg? I would assume we want to keep `nrows` as an argument.

jreback:
> @jreback @gfyoung i too agree that just the value of low_memory should not render the combination of values invalid. Because of this corner case, this is actually exhibiting weird behavior on an unrelated set of conditions. Better to suppress this corner case by removing/deprecating nrows=0.
Is there a way to deprecate just a value for an arg? I would assume we want to keep nrows as an argument.

no reason to remove or even deprecate ``nrows=0``, you simply need to addees for this particular combination of options

cgopalan:
@jreback so `read_csv(StringIO(data), low_memory=False, index_col=0, nrows=0)` is allowed but `read_csv(StringIO(data), low_memory=True, index_col=0, nrows=0)` ?

pandas/tests/io/test_excel.py
cgopalan: @WillAyd sure, I will make these changes as soon as we decide if we are keeping this PR :)

jreback:
yeah i thought that was fine

gfyoung:
@jreback : 🤦‍♂️ well, I guess we're heading back there then. 😄

cgopalan:
@gfyoung should i delete this branch?

gfyoung:
@cgopalan : Sure thing.

PR #21433, opened 2018-06-11
DOC: Add 0.23.2 whatsnew template

gfyoung:
Title is self-explanatory.

Copied (almost) directly from #21001.


jreback:
actually this is totally fine (and doesn't update the index yet anyhow).

PR #21442, opened 2018-06-12
0.23.1 backports 2

TomAugspurger:


TomAugspurger:
One of the backported tests used a fixture that was tagged for 0.24. Manually backported that fixture.

PR #21444, opened 2018-06-12
Resolves Issue 21344: provide a timedelta in a non-string format

### LINKED ISSUE #21344: pd.Timedelta.resolution is different from datetime.timedelta.resolution
`timedelta.resolution` is documented as:
> The smallest possible difference between non-equal timedelta objects, timedelta(microseconds=1).

while `pandas.Timedelta.resolution` is doc-string'd as:
>return a string representing the lowest resolution that we have
```
In [9]: pd.Timedelta(minutes=30).resolution
Out[9]: 'T'

In [10]: timedelta(minutes=30).resolution
Out[10]: datetime.timedelta(0, 0, 1)
```

I think this difference should be addressed by either:
1) Documenting the difference between both but also supporting `datetime.timedelta`'s notion of resolution in `Timedelta` (i.e. something that returns `Timedelta(nanoseconds=1)`)
2) Making `Timedelta.resolution` consistent with `timedelta.resolution` and creating a new property for "the lowest resolution" of a `Timedelta` instance

Thoughts? 

xref #21336

drewmassey:
- [x] closes #21344
- [x] tests added / passed
- [x] whatsnew entry

This extends the API - and there is a judgement call about how to treat nanosecond-resolution pd.Timedelta() objects in the context of datetime.timedelta() objects.


pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -813,6 +817,32 @@ cdef class _Timedelta(timedelta):
         else:
             return "D"
 
+    @property
+    def resolution_timedelta(self):
```
mroeschke: The general idea was to have this new property return `Timedelta(nanosecond=1)` akin to how `datetime.timedelta.resolution` returns `timedelta(microseconds=1)`

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -795,7 +795,11 @@ cdef class _Timedelta(timedelta):
 
     @property
     def resolution(self):
-        """ return a string representing the lowest resolution that we have """
+        """
+        Return a string representing the lowest resolution that we have.
+        Note that this is nonstandard behavior.
```
mroeschke: Instead, could you create a "See Also" section with the new method name, similar to this: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.isin.html

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -795,7 +795,11 @@ cdef class _Timedelta(timedelta):
 
     @property
     def resolution(self):
-        """ return a string representing the lowest resolution that we have """
+        """
+        Return a string representing the lowest resolution that we have.
```
mroeschke: Instead of `lowest resolution that we have`, I think I'd be clearer if it was `lowest resolution of the Timedelta instance`

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -813,6 +817,32 @@ cdef class _Timedelta(timedelta):
         else:
             return "D"
 
+    @property
+    def resolution_timedelta(self):
+        """
+        Return a timedelta object (rather than a string)
+        representing the lowest resolution we have.
```
mroeschke: Instead of `lowest resolution that we have`, I think I'd be clearer if it was `lowest resolution of the Timedelta class`

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -813,6 +817,32 @@ cdef class _Timedelta(timedelta):
...
+        to retrieve a string use the resolution property.
```
mroeschke: Instead, could you create a "See Also" section with the new method name, similar to this: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.isin.html

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -133,3 +133,4 @@ Bug Fixes
 
 - Tab completion on :class:`Index` in IPython no longer outputs deprecation warnings (:issue:`21125`)
 - Bug preventing pandas being used on Windows without C++ redistributable installed (:issue:`21106`)
+- Add `resolution_timedelta` to :class:`Timedelta` to get non-string representations of resolution (:issue: `21344`)
```
mroeschke: Could you move to `v0.23.2.txt`

jreback:
``Timedelta.resolution`` is used completely internally. rather have you change its name and leave ``.resolution`` as an inherited property. I didn't even know that ``timedelta`` has a property by this name.

This would then be an API change, but conform to ``timedelta``, which we in general want to do.

jreback:
can you rebase

jreback:

PR #21451, opened 2018-06-12
PERF: Use ccalendar.get_days_in_month over tslib.monthrange

jbrockmendel:
asv looks like a wash, probably to be expected since these calls make up such a small part of the methods they are used in.

tslibs.resolution uses the stdlib calendar.monthrange instead of tslib.monthrange, so the perf bump should be bigger there.


gfyoung:
@jbrockmendel : Any `asv`-related information + `whatsnew` in `0.24.0` would be good here.

jreback:
this is fine, no whatsnew needed if just internal refactoring generally (unless its 'big') or has an api impact.

PR #21452, opened 2018-06-12
PERF: typing and cdefs for tslibs.resolution

jbrockmendel:
Moving things lower-level will help improve performance (due in part to better Cython compilation).

### DIFF
--- a/pandas/_libs/tslibs/resolution.pyx
+++ b/pandas/_libs/tslibs/resolution.pyx
@@ -389,8 +389,8 @@ cdef class _FrequencyInferer(object):
     Not sure if I can avoid the state machine here
     """
     cdef public:
-        index
-        values
+        object index
+        object values
         bint warn
         bint is_monotonic
         dict _cache
--- a/pandas/_libs/tslibs/resolution.pyx
+++ b/pandas/_libs/tslibs/resolution.pyx
@@ -388,7 +388,12 @@ cdef class _FrequencyInferer(object):
     """
     Not sure if I can avoid the state machine here
     """
-    cdef public index, values, warn, is_monotonic, _cache
+    cdef public:
+        index
+        values
+        bint warn
+        bint is_monotonic
+        dict _cache
 
     def __init__(self, index, warn=True):
         self.index = index
@@ -485,10 +490,14 @@ cdef class _FrequencyInferer(object):
             bint calendar_start = True
             bint business_start = True
             bint cal
+            int32_t[:] years
+            int32_t[:] months
+            int32_t[:] days
 
-        years = self.fields['Y']
-        months = self.fields['M']
-        days = self.fields['D']
+        fields = self.fields
+        years = fields['Y']
+        months = fields['M']
+        days = fields['D']
         weekdays = self.index.dayofweek
 
         from calendar import monthrange
--- a/pandas/_libs/tslibs/resolution.pyx
+++ b/pandas/_libs/tslibs/resolution.pyx
@@ -5,7 +5,7 @@ from cython cimport Py_ssize_t
 
 import numpy as np
 cimport numpy as cnp
-from numpy cimport ndarray, int64_t
+from numpy cimport ndarray, int64_t, int32_t
 cnp.import_array()
 
 from util cimport is_string_object, get_nat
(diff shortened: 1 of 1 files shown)


pandas/_libs/tslibs/resolution.pyx
```diff
@@ -373,21 +373,27 @@ cdef unique_deltas(ndarray[int64_t] arr):
     return result
 
 
-def _is_multiple(us, mult):
+cdef inline bint _is_multiple(int64_t us, int64_t mult):
     return us % mult == 0
 
 
-def _maybe_add_count(base, count):
+cdef inline str _maybe_add_count(str base, count):
```
WillAyd: Can we add a int64_t type for `count` here?

pandas/_libs/tslibs/resolution.pyx
```diff
@@ -373,21 +373,27 @@ cdef unique_deltas(ndarray[int64_t] arr):
...
         return base
 
 
-class _FrequencyInferer(object):
+cdef class _FrequencyInferer(object):
     """
     Not sure if I can avoid the state machine here
     """
+    cdef public:
+        index
```
WillAyd: Do we need types here?

pandas/_libs/tslibs/resolution.pyx
```diff
@@ -475,16 +481,23 @@ class _FrequencyInferer(object):
...
-        months = self.fields['M']
-        days = self.fields['D']
+        cdef:
+            int32_t daysinmonth, y, m, d
+            bint calendar_end = True
+            bint business_end = True
+            bint calendar_start = True
+            bint business_start = True
+            bint cal
+            int32_t[:] years
```
WillAyd: These are memoryviews, right? Any reason you are choosing these instead of the `ndarray` declarations throughout the rest of the code base?

pandas/_libs/tslibs/resolution.pyx
```diff
@@ -373,21 +373,27 @@ cdef unique_deltas(ndarray[int64_t] arr):
...
 
-def _is_multiple(us, mult):
+cdef inline bint _is_multiple(int64_t us, int64_t mult):
     return us % mult == 0
 
 
-def _maybe_add_count(base, count):
+cdef inline str _maybe_add_count(str base, count):
     if count != 1:
         return '{count}{base}'.format(count=int(count), base=base)
```
WillAyd: If above comment is correct shouldn't need the int cast here

pandas/_libs/tslibs/resolution.pyx
jbrockmendel: Ill double-check, but I'm pretty sure past-me had a reason for leaving that one out.  Maybe there are cases when `None` is passed

pandas/_libs/tslibs/resolution.pyx
jbrockmendel: Not really viable since we can't specify `Index` as a type.  We just have to make the name public for wheels to turn.

pandas/_libs/tslibs/resolution.pyx
jbrockmendel: My understanding is that memoryviews are encouraged because they are lighter weight.

pandas/_libs/tslibs/resolution.pyx
jbrockmendel: Looks like you're right.  Just pushed a commit that added the suggested typing and removed the `int(`.  (and hopefully fixed the test errors)

pandas/_libs/tslibs/resolution.pyx
gfyoung: @jbrockmendel : That is indeed correct.  I agree with this design choice.

gfyoung:
@jbrockmendel : I feel like the performance help is relatively obvious given that you are moving things underneath the Python level.  Is there any `asv` that could illustrate this by any chance (not a blocker, but just curious)?  If so, that might be worthwhile to include in a `whatsnew`.

pandas/_libs/tslibs/resolution.pyx
jorisvandenbossche: In general, one of the main problems with memoryviews is with read-only arrays (and for that reason we do/should not use them solely in certain algos), but I don't think that's relevant in this case

pandas/_libs/tslibs/resolution.pyx
jreback: just specify them as object to make this explict

pandas/_libs/tslibs/resolution.pyx
jreback: you can make cdefs readonly when it makes sense FYI

jreback (changes requested):
minor comments. just run asv's to confirm no regressions. of course if improvements then shout it out!
ping on green after updating for comments.

jreback (approved):

jreback:
thanks!

PR #21453, opened 2018-06-12
Fix tests fragile to PATH

### LINKED ISSUE #21450: Python -00 test is fragile to PATH
https://github.com/pandas-dev/pandas/blob/ab668b0a56a9f2aee959bde787e9a0af4068d7a7/pandas/tests/test_downstream.py#L60

that assumes that first `python`  on the PATH is the one running the tests, which it may not be (e.g. if doing `./venv/bin/python`

We should replace `python` with something like `sys.executable`.

alimcmaster1:
Closes #21450

Get path of current python interpreter running opposed to one on users path.

### DIFF
--- a/pandas/tests/plotting/test_converter.py
+++ b/pandas/tests/plotting/test_converter.py
@@ -1,4 +1,5 @@
 import subprocess
+import sys
 import pytest
 from datetime import datetime, date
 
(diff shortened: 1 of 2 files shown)


WillAyd (approved):
lgtm. Only downside I can think of with this is when `sys.executable` returns an empty string, though I couldn't find definitive documentation on when that happens (embedded applications?) so maybe not an issue

cc @TomAugspurger

PR #15859, opened 2017-04-01
API/BUG: Handling Dtype Coercions in Series/Index (GH 15832)

ucals:
- [x] closes #15832
 - [x] tests added / passed
 - [x] passes ``git diff upstream/master --name-only -- '*.py' | flake8 --diff``
 - [x] whatsnew entry

Hey @jreback , need a quick help here. I was able to raise an overflow exception (case II), but I wasn't able to raise it on case III. Can you help?  Thanks

Issue #21311, opened 2018-06-04
BUG/API: Index constructor does not enforce specified dtype 


jorisvandenbossche:
#### Code Sample, a copy-pastable example if possible

Manually specifying a dtype does not garantuee the output is in that dtype. Eg with Series if incompatible data is passed, an error is raised, while for Index it just silently outputs another dtype:

```python
In [11]: pd.Series(['a', 'b', 'c'], dtype='int64')
...
ValueError: invalid literal for int() with base 10: 'a'

In [12]: pd.Index(['a', 'b', 'c'], dtype='int64')
Out[12]: Index(['a', 'b', 'c'], dtype='object')

```

jschendel:
xref https://github.com/pandas-dev/pandas/pull/21254#discussion_r191658787

Meant to create an issue similar to this for the `Categorical` --> `Interval` case:
```python
In [2]: cat = pd.Categorical([pd.Interval(0, 1), pd.Interval(1, 2), pd.Interval(0, 1)])

In [3]: pd.Index(cat, dtype='interval')
Out[3]: CategoricalIndex([(0, 1], (1, 2], (0, 1]], categories=[(0, 1], (1, 2]], ordered=False, dtype='category')
```

This happens because the `Index` code is structured so that categorical takes precedence over interval:
https://github.com/pandas-dev/pandas/blob/c85ab083919b59ce84c220d5baf7d34ff4a0bcf2/pandas/core/indexes/base.py#L262-L273

The code above could be restructured so that the `dtype` argument, if present, takes precedence over the type of `data`.

Not immediately sure what the fix for other scenarios entails, as `Categorical` and `Interval` are a bit of a special case.

PR #21456, opened 2018-06-13
API/BUG: Raise when int-dtype coercions fail

### LINKED ISSUE #15832: API/BUG: Handling Dtype Coercions in Series/Index
Off `master` (<a href="https://github.com/pandas-dev/pandas/commit/bd169dc0a91f50031f6c2240075ff84d6b296576">bd169d</a>):
~~~python
# Case I: Overflow on int64
>>> Index([np.iinfo(np.uint64).max-1],dtype='int64')
...
OverflowError: Python int too large to convert to C long

# Case II: Coercion to uint64
>>> Index([-1], dtype='uint64')
UInt64Index([18446744073709551615], dtype='uint64')

# Case III: Ignoring coercion to int
>>> Index([1, 2, 3.5], dtype=int)
Float64Index([1.0, 2.0, 3.5], dtype='float64')
~~~

So we got some coercions that fail but others that work.  Although all of these issues involve `Index`, the first two are also applicable to `Series` (in the last issue, it does coerce all elements to integer).

How should we handle failed coercions?  Is the second case even a failure?  Should iron out this.

xref #12758
xref #15187

gfyoung:
Related to the Index and Series constructors.

Closes #15832.

cc @ucals (since this is mostly based off what you did in #15859)

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -26,7 +26,7 @@ Other Enhancements
 Backwards incompatible API changes
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. _whatsnew_0240.api.datetimelike.normalize
+.. _whatsnew_0240.api.datetimelike.normalize:
 
 Tick DateOffset Normalize Restrictions
 --------------------------------------
@@ -72,6 +72,32 @@ Datetimelike API Changes
 
 Other API Changes
 ^^^^^^^^^^^^^^^^^
+
+.. _whatsnew_0240.api.other.incompatibilities:
+
+Series and Index Data-Dtype Incompatibilities
+---------------------------------------------
+
+``Series`` and ``Index`` constructors now raise when the
+data is incompatible with a passed ``dtype=`` (:issue:`15832`)
+
+Previous Behavior:
+
+.. code-block:: ipython
+
+    In [4]: pd.Series([-1], dtype="uint64")
+    Out [4]:
+    0    18446744073709551615
+    dtype: uint64
+
+Current Behavior:
+
+.. code-block:: ipython
+
+    In [4]: pd.Series([-1], dtype="uint64")
+    Out [4]:
+    ...
+    OverflowError: Trying to coerce negative values to unsigned integers
 
 - :class:`DatetimeIndex` now accepts :class:`Int64Index` arguments as epoch timestamps (:issue:`20997`)
 -
--- a/pandas/core/dtypes/cast.py
+++ b/pandas/core/dtypes/cast.py
@@ -20,6 +20,7 @@ from .common import (_ensure_object, is_bool, is_integer, is_float,
                      is_dtype_equal,
                      is_float_dtype, is_complex_dtype,
                      is_integer_dtype,
+                     is_unsigned_integer_dtype,
                      is_datetime_or_timedelta_dtype,
                      is_bool_dtype, is_scalar,
                      is_string_dtype, _string_dtypes,
@@ -1269,3 +1270,74 @@ def construct_1d_ndarray_preserving_na(values, dtype=None, copy=False):
         subarr = subarr2
 
     return subarr
+
+
+def maybe_cast_to_integer_array(arr, dtype, copy=False):
+    """
+    Takes any dtype and returns the casted version, raising for when data is
+    incompatible with integer/unsigned integer dtypes.
+
+    .. versionadded:: 0.24.0
+
+    Parameters
+    ----------
+    arr : array-like
+        The array to cast.
+    dtype : str, np.dtype
+        The integer dtype to cast the array to.
+    copy: boolean, default False
+        Whether to make a copy of the array before returning.
+
+    Returns
+    -------
+    int_arr : ndarray
+        An array of integer or unsigned integer dtype
+
+    Raises
+    ------
+    OverflowError : the dtype is incompatible with the data
+    ValueError : loss of precision has occurred during casting
+
+    Examples
+    --------
+    If you try to coerce negative values to unsigned integers, it raises:
+
+    >>> Series([-1], dtype="uint64")
+    Traceback (most recent call last):
+        ...
+    OverflowError: Trying to coerce negative values to unsigned integers
+
+    Also, if you try to coerce float values to integers, it raises:
+
+    >>> Series([1, 2, 3.5], dtype="int64")
+    Traceback (most recent call last):
+        ...
+    ValueError: Trying to coerce float values to integers
+    """
+
+    try:
+        if not hasattr(arr, "astype"):
+            casted = np.array(arr, dtype=dtype, copy=copy)
+        else:
+            casted = arr.astype(dtype, copy=copy)
+    except OverflowError:
+        raise OverflowError("The elements provided in the data cannot all be "
+                            "casted to the dtype {dtype}".format(dtype=dtype))
+
+    if np.array_equal(arr, casted):
+        return casted
+
+    # We do this casting to allow for proper
+    # data and dtype checking.
+    #
+    # We didn't do this earlier because NumPy
+    # doesn't handle `uint64` correctly.
+    arr = np.asarray(arr)
+
+    if is_unsigned_integer_dtype(dtype) and (arr < 0).any():
+        raise OverflowError("Trying to coerce negative values "
+                            "to unsigned integers")
+
+    if is_integer_dtype(dtype) and (is_float_dtype(arr) or
+                                    is_object_dtype(arr)):
+        raise ValueError("Trying to coerce float values to integers")
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -21,6 +21,7 @@ from pandas.core.dtypes.generic import (
     ABCPeriodIndex, ABCTimedeltaIndex,
     ABCDateOffset)
 from pandas.core.dtypes.missing import isna, array_equivalent
+from pandas.core.dtypes.cast import maybe_cast_to_integer_array
 from pandas.core.dtypes.common import (
     _ensure_int64,
     _ensure_object,
@@ -311,18 +312,15 @@ class Index(IndexOpsMixin, PandasObject):
                     if is_integer_dtype(dtype):
                         inferred = lib.infer_dtype(data)
                         if inferred == 'integer':
-                            try:
-                                data = np.array(data, copy=copy, dtype=dtype)
-                            except OverflowError:
-                                # gh-15823: a more user-friendly error message
-                                raise OverflowError(
-                                    "the elements provided in the data cannot "
-                                    "all be casted to the dtype {dtype}"
-                                    .format(dtype=dtype))
+                            data = maybe_cast_to_integer_array(data, dtype,
+                                                               copy=copy)
                         elif inferred in ['floating', 'mixed-integer-float']:
                             if isna(data).any():
                                 raise ValueError('cannot convert float '
                                                  'NaN to integer')
+
+                            if inferred == "mixed-integer-float":
+                                data = maybe_cast_to_integer_array(data, dtype)
 
                             # If we are actually all equal to integers,
                             # then coerce to integer.
@@ -352,7 +350,8 @@ class Index(IndexOpsMixin, PandasObject):
 
                 except (TypeError, ValueError) as e:
                     msg = str(e)
-                    if 'cannot convert float' in msg:
+                    if ("cannot convert float" in msg or
+                            "Trying to coerce float values to integer" in msg):
                         raise
 
             # maybe coerce to a sub-class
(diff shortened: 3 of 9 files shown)


pandas/core/dtypes/cast.py
```diff
@@ -1269,3 +1270,64 @@ def construct_1d_ndarray_preserving_na(values, dtype=None, copy=False):
...
+    ValueError: Trying to coerce float values to integers
+    """
+
+    try:
+        casted = arr.astype(dtype, copy=copy)
+    except OverflowError:
+        raise OverflowError("The elements provided in the data cannot all be "
+                            "casted to the dtype {dtype}".format(dtype=dtype))
+
+    if np.array_equal(arr, casted):
```
gfyoung: numpy-self-warning

Since this is in part due to `numpy`'s self-conflict regarding element-wise comparisons, should we care about this showing up (perhaps we swallow the warning)?

pandas/tests/indexes/test_numeric.py
```diff
@@ -451,6 +451,20 @@ def test_astype(self):
             i = Float64Index([0, 1.1, np.NAN])
             pytest.raises(ValueError, lambda: i.astype(dtype))
 
+    @pytest.mark.parametrize("int_dtype", ["uint8", "uint16", "uint32",
+                                           "uint64", "int32", "int64",
+                                           "int16", "int8"])
+    @pytest.mark.parametrize("float_dtype", ["float16", "float32", "float64"])
```
gfyoung: Perhaps these (i.e. `int_dtype`, `float_dtype`, `uint_dtype` below) should be `conftest.py` fixtures?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -36,6 +36,7 @@ Datetimelike API Changes
 Other API Changes
 ^^^^^^^^^^^^^^^^^
 
+- Series and Index constructors now raise when the data is incompatible with the specified dtype (:issue:`15832`)
```
jreback: double backticks around Series/Index

say with a passed ``dtype=``

pandas/core/dtypes/cast.py
```diff
@@ -1269,3 +1270,64 @@ def construct_1d_ndarray_preserving_na(values, dtype=None, copy=False):
...
     return subarr
+
+
+def maybe_cast_to_integer_array(arr, dtype, copy=False):
+    """
+    Takes any dtype and returns the casted version, raising for when data is
+    incompatible with integer/unsigned integer dtypes.
+
+    .. versionadded:: 0.24.0
+
```
jreback: this duplicates maybe_downcast_to_dtype which is used internally, rather have the doc-string of that updated / examples (and can add the copy=)

pandas/tests/indexes/test_numeric.py
jreback: yes we have multiple PRs trying to do this (and lots of existing code). can you create / find (I think there is one) an issue about this. PRs to fix even better :> (and may take multiple as using these fixtures can be non-trivial to refactor)

jreback (changes requested):

pandas/core/dtypes/cast.py
gfyoung: Hmm...that's going to take some time given that `maybe_downcast_to_dtype` is much more lenient than this function is.  That being said, I do see your point.  Let me see.

pandas/tests/indexes/test_numeric.py
gfyoung: I'll add them as fixtures to this PR and then can see afterwards what can be done with them.

doc/source/whatsnew/v0.24.0.txt
gfyoung: Yep, done.

pandas/tests/indexes/test_numeric.py
gfyoung: @jreback : BTW, I didn't realize that the comment in your "review" was referencing the one in this `diff` (I thought I was deleting a duplicate comment).  Oops.

pandas/conftest.py
```diff
@@ -170,3 +170,67 @@ def string_dtype(request):
     * 'U'
     """
     return request.param
+
+
+@pytest.fixture(params=["float16", "float32", "float64"])
+def float_dtype(request):
```
jschendel: I was also asked to add these fixtures in #21432.  I think it might be best to merge #21432 first to make backporting easier since it's tagged for 0.23.2?  Not sure if it really matters though.

Your code looks 95% identical to mine, so I'll just modify mine to match yours prior to pushing so there shouldn't be any significant issues regardless of which PR gets merged first.  The only potential sources of differences:
- I was asked to not include float16 since it's barely supported (https://github.com/pandas-dev/pandas/pull/21432#discussion_r194702410)
- Looks like the existing convention is to not include a blank line between the docstring and `return`

pandas/conftest.py
```diff
@@ -170,3 +170,67 @@ def string_dtype(request):
...
+    * uint8
+    * uint16
+    * uint32
+    * uint64
+    """
+
+    return request.param
+
+
+@pytest.fixture(params=SIGNED_INT_DTYPES)
```
jschendel: params should be `ALL_INT_DTYPES `

jschendel: Also, should we rename either the list or the fixture to make them consistent ("all" vs. "any")?

pandas/conftest.py
gfyoung: @jschendel : I'll address all of your questions:

* The merging doesn't particularly matter at this point.  I would hold on modifying your fixtures for the moment.  Let's ask @jreback in your PR.

* `float16`, ah that's true.  I'll remove that.

* I generally prefer a newline after the docstring.  Facilitates reading IMO.

pandas/conftest.py
gfyoung: * Oops, good catch 😄 

* So this was a decision in semantics.  The list is "all" of the integer dtypes, hence the name.  However, `all_int_dtype` as a parameter to the test makes less sense vs. `any_int_dtype.`

pandas/tests/series/test_constructors.py
```diff
@@ -542,12 +542,29 @@ def test_constructor_pass_nan_nat(self):
...
+        # see gh-15832
+        msg = 'Trying to coerce negative values to unsigned integers'
+        with tm.assert_raises_regex(OverflowError, msg):
+            Series([-1], dtype=uint_dtype)
+
+    def test_constructor_coerce_float_fail(self, any_int_dtype, float_dtype):
+        # see gh-15832
+        msg = "Trying to coerce float values to integers"
+        with tm.assert_raises_regex(ValueError, msg):
+            Series([1, 2, 3.5], dtype=any_int_dtype)
```
jreback: so this is going to cartesian product these, but these look like independent cases, can you split to 2 tests.

pandas/tests/indexes/test_numeric.py
```diff
@@ -451,6 +451,16 @@ def test_astype(self):
             i = Float64Index([0, 1.1, np.NAN])
             pytest.raises(ValueError, lambda: i.astype(dtype))
 
+    def test_type_coercion(self, any_int_dtype, float_dtype):
+
+        # see gh-15832
+        msg = "Trying to coerce float values to integers"
+        with tm.assert_raises_regex(ValueError, msg):
+            Index([1, 2, 3.5], dtype=any_int_dtype)
```
jreback: same here, can you split to 2 tests

pandas/tests/series/test_constructors.py
```diff
@@ -542,12 +542,29 @@ def test_constructor_pass_nan_nat(self):
         tm.assert_series_equal(Series(np.array([np.nan, pd.NaT])), exp)
 
     def test_constructor_cast(self):
-        pytest.raises(ValueError, Series, ['a', 'b', 'c'], dtype=float)
+        msg = "could not convert string to float"
+        with tm.assert_raises_regex(ValueError, msg):
+            Series(["a", "b", "c"], dtype=float)
+
```
jreback: do we have these tests for Index as well (I think we do)?

jreback (changes requested):

pandas/tests/series/test_constructors.py
gfyoung: 😮 : That's a bug!

~~~python
>>> Index([["a", "b", "c"], dtype=float)
Index([["a", "b", "c"], dtype=object)
~~~

jschendel: See #21311

gfyoung: @jreback : I'll add an `xfail` test for this.

pandas/tests/series/test_constructors.py
gfyoung: Done.

pandas/tests/indexes/test_numeric.py
gfyoung: Done.

jreback:
@gfyoung if you can rebase your fixures should cleanly merge

pandas/core/indexes/base.py
```diff
@@ -311,19 +312,16 @@ def __new__(cls, data=None, dtype=None, copy=False, name=None,
...
-                                    .format(dtype=dtype))
+                            data = maybe_cast_to_integer_array(data, dtype,
+                                                               copy=copy)
                         elif inferred in ['floating', 'mixed-integer-float']:
                             if isna(data).any():
                                 raise ValueError('cannot convert float '
                                                  'NaN to integer')
 
+                            if inferred == "mixed-integer-float":
+                                maybe_cast_to_integer_array(data, dtype)
```
jreback: do you need ``data = `` here?

jreback (changes requested):
question, also I think its worthwhile to make a sub-section in the whatsnew with a mini-example

pandas/core/indexes/base.py
gfyoung: Fixed.  Also added mini-section to `whatsnew`.

pandas/core/dtypes/cast.py
```diff
@@ -1269,3 +1270,74 @@ def construct_1d_ndarray_preserving_na(values, dtype=None, copy=False):
...
+    if np.array_equal(arr, casted):
+        return casted
+
+    # We do this casting to allow for proper
+    # data and dtype checking.
+    #
+    # We didn't do this earlier because NumPy
+    # doesn't handle `uint64` correctly.
+    arr = np.asarray(arr)
```
jreback: why do you need this? at this point its either an ndarray, Series, Index, which is all ok here

pandas/core/series.py
```diff
@@ -4068,6 +4069,9 @@ def _try_cast(arr, take_fast_path):
                 return arr
 
         try:
+            if is_float_dtype(dtype) or is_integer_dtype(dtype):
```
jreback: can you add a comment here

jreback (changes requested):
can you run some constructor asv's to see if any effect.

pandas/core/dtypes/cast.py
gfyoung: Not quite.  `list` can also propagate here.  This was prompted by #21432, which introduced an annoying but necessary corner case with `uint64`.

pandas/core/series.py
gfyoung: Yep, done.

pandas/core/dtypes/cast.py
jreback: ic, then add ``copy=False``

gfyoung:
@jreback : Tests are green again.  PTAL.

jreback (changes requested):
minor comment. perf check?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -71,6 +71,32 @@ Datetimelike API Changes
 Other API Changes
 ^^^^^^^^^^^^^^^^^
 
+.. _whatsnew_0240.api.other.incompatibilities
```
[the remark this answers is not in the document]
gfyoung: Oops.  Fixed.

gfyoung:
> perf check?

@jreback : Oh sorry, forgot to respond on that.  Perf looks good!

gfyoung:
@jreback : Comments addressed, perf looks good, and CI is green.  PTAL.

jreback (approved):

jreback:
thanks @gfyoung  nice patch!

pandas/core/dtypes/cast.py
```diff
@@ -1269,3 +1270,74 @@ def construct_1d_ndarray_preserving_na(values, dtype=None, copy=False):
...
+
+    if is_unsigned_integer_dtype(dtype) and (arr < 0).any():
+        raise OverflowError("Trying to coerce negative values "
+                            "to unsigned integers")
```
jbrockmendel, line 1339: @gfyoung should maybe_cast_to_integer_array also check for overflows like

```
arr = np.array([1, 200, 923442])
dtype = np.dtype(np.int8)
```

?

PR #21477, opened 2018-06-14
BUG: inconsistency between replace dict using integers and using strings (#20656)

### LINKED ISSUE #20656: DataFrame.replace with dict behaves inconsistently for integers and strings
From [SO](https://stackoverflow.com/q/49775130/4909087)

DataFrame.replace does not behave consistently when replacing integers vs replacing strings.

Reproducible example:
```
df = pd.DataFrame({'A': [0, 1]})
df.A.replace({0: 1, 1: 0})
```
This yields the result:

````
    df
    A
    1
    0
````

However, when using the same commands for string values

````
df = pd.DataFrame({'B': ['a', 'b']})
df.B.replace({'a': 'b', 'b': 'a'})
````
This gives

````
df
B
a
a
````

But it should give

````
B
b
a

````

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.0.final.0
python-bits: 64
OS: Darwin
OS-release: 17.4.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.22.0
pytest: None
pip: 9.0.3
setuptools: 38.5.1
Cython: 0.28.1
numpy: 1.14.0
scipy: 1.0.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: None
patsy: None
dateutil: 2.6.1
pytz: 2018.3
blosc: None
bottleneck: None
tables: None
numexpr: 2.6.4
feather: None
matplotlib: 2.1.2
openpyxl: None
xlrd: 0.9.4
xlwt: None
xlsxwriter: None
lxml: 3.8.0
bs4: 4.6.0
html5lib: 1.0b8
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

### PR DESCRIPTION
**Pull request by peterpanmj**
- [x] closes #20656
- [x] tests added / passed
- [x] whatsnew entry

### CODE REVIEW

**Comment by gfyoung** [comment]
@peterpanmj : Good start.  Need a `whatsnew` entry in `0.23.2`.

**Comment by peterpanmj** [comment]
@gfyoung : Which subsection under Bug fixes should I add my entry ?

File: pandas/core/internals.py [hunk 1]
```diff
@@ -3748,18 +3753,44 @@ def comp(s):
...
+                    # regular expression support needs to be improved.
+                    # If the replacement for the previous pattern
+                    # matches the next pattern, the value will be replaced
+                    # again with a different value from dest_list
+                    # i.e. when values is ['a', 'b']
+                    # src_list: [r'a*', r'b*'], dest_list: ['b', 'a']
+                    # result will be ['b', b'] after searching for pattern r'a'
+                    # and then changed to ['a', 'a'] for pattern r'b*'
+                    if regex:
+                        if b.dtype == np.object_:
```
jreback: can you use is_object_dtype here

pandas/core/internals.py
```diff
@@ -3748,18 +3753,44 @@ def comp(s):
...
+                            convert = i == src_len
+                            result = b.replace(s, d, inplace=inplace,
+                                               regex=regex,
+                                               mgr=mgr, convert=convert)
+                            new_rb = _extend_blocks(result, new_rb)
```
jreback: instead of this, can you add a new (private) method on the Block itself (and then override for object dtype). It will be much cleaner code and this part becomes really generic.

jreback (changes requested):

jreback (changes requested):

pandas/core/internals.py
```diff
@@ -3748,18 +3769,37 @@ def comp(s):
...
+                        else:
```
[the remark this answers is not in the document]
peterpanmj: I kept the logic in master untouched for regex mode. This means regex replace still behave incorrectly (just like how dict replace is behaving now)
 ``_maybe_compare`` can only do equality compare for now. That is the cause of it. I haven't figure out how to add regex support in ``_maybe_compare`` without breaking any existing test. Once _maybe_compare is fixed, this part can be removed. I think @Licht-T is working on this part #20656.

pandas/core/internals.py
```diff
@@ -1690,6 +1690,17 @@ def _nanpercentile(values, q, axis, **kw):
                               placement=np.arange(len(result)),
                               ndim=ndim)
 
+    def _replace_coerce(self, mask=None, dst=None, convert=False):
+        """replace value to dst where mask is true, value is coerce to target
```
jreback: can you add a full Parameters section here

pandas/core/internals.py
```diff
@@ -2538,9 +2549,17 @@ def re_replacer(s):
         block = self.make_block(new_values)
         if convert:
             block = block.convert(by_item=True, numeric=False)
-
         return block
 
+    def _replace_coerce(self, mask=None, dst=None, convert=False):
+        if mask.any():
```
jreback: same here

pandas/core/internals.py
jreback: can you change this

pandas/core/internals.py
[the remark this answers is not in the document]
jreback: well what I would do is add a ``regex=`` param to ``_replace_coerce`` and push the logic to the block

jreback (changes requested):

pandas/core/internals.py
peterpanmj: Do you mean add **kwargs and *args ?

peterpanmj: @jreback Or do you mean add a full docsstring with all parameters ?

jreback: yes full doc string

pandas/core/internals.py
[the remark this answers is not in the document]
peterpanmj: In addition to ``regex=`` param, it also needs ``src=`` param for the regex pattern to match. I think the ``_replace_coerce`` should only take a boolean array (``mask=``) indicating where to put the new value. Meanwhile, regex matching should be done at generating the mask before passing it to the ``_replace_coerce``.  Adding regex logic to it might makes it less clear for others and probably let someone else override ``_replace_coerce`` or reuse it for regex replace. What I want is keep ``_replace_coerce`` just like ``putmask`` and handle the regex comparison at ``_maybe_compare(values, s, operator.eq)``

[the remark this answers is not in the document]
jreback: I disagree, this is crazy code here and needs simplification. all of the real work should be done in the blocks themselves, e .g. object is different than the other blocks. 

Here should be a simple
```
rb = [b._replace(.....) for b in rb]
```

you can pass whatever args you want, but the top level logic is just too complicated here

peterpanmj:
Wondering am I working in the right direction ?

jreback (changes requested):

jreback:
@peterpanmj  yes going in good direction! need to rebase as we moved internals around a bit.

pandas/core/internals.py
```diff
@@ -5143,7 +5223,12 @@ def _vstack(to_stack, dtype):
         return np.vstack(to_stack)
 
 
-def _maybe_compare(a, b, op):
+def _maybe_compare(a, b, regex=False):
+    if not regex:
+        op = lambda x: operator.eq(x, b)
+    else:
```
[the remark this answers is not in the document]
peterpanmj: I want to add regex support in _maybe_compare. The comparing behavior is decided by param regex ( whether to use regex match or equality comparison) . The result will be a mask that  will be passed on to _replace_coerce. This can avoid the result of previous round of comparing overwritten in the succeeding ones, e.g in, {"a":"b", "b":"a"} .

pandas/core/internals.py
```diff
@@ -5155,9 +5240,8 @@ def _maybe_compare(a, b, op):
     # numpy deprecation warning if comparing numeric vs string-like
     elif is_numeric_v_string_like(a, b):
         result = False
-
     else:
```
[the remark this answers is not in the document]
peterpanmj: I removed a blank line. Should I keep it ?
And since the ``op`` function is modified based on b, as in the above lines. It wont need param b here.

jreback:
can you rebase

jreback (changes requested):
@jbrockmendel can you have a look

jbrockmendel:
Looks like a nice bit of cleanup in Manager.  For Block I wonder if it could share code with Index,
 (Or an object EA?) but that’s a question for another day.

pandas/core/internals/managers.py
```diff
@@ -571,12 +574,17 @@ def replace_list(self, src_list, dest_list, inplace=False, regex=False,
...
 
-        def comp(s):
+        def comp(s, reg=False):
             if isna(s):
                 return isna(values)
-            return _maybe_compare(values, getattr(s, 'asm8', s), operator.eq)
+            if hasattr(s, 'asm8'):
+                return _maybe_compare(maybe_convert_objects(values),
+                                      getattr(s, 'asm8'), reg)
+            if reg and is_re_compilable(s):
```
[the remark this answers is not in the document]
peterpanmj: My mistake. At first, I wanted to raise a ValueError when regex is True but replacer is not reg compilable. It might not be a good idea to do it here. I will delete the if condition.

pandas/core/internals/blocks.py
```diff
@@ -1689,6 +1689,44 @@ def _nanpercentile(values, q, axis, **kw):
...
 
+    def _replace_coerce(self, mask=None, src=None, dst=None, inplace=True,
+                        convert=False, regex=False, mgr=None):
+        """
+        Replace value corresponding to the given boolean array with another
+        value.
+
+        Parameters
+        ----------
+        mask : array_like of bool
```
jreback: can you add optional to all of these args

pandas/core/internals/blocks.py
```diff
@@ -1689,6 +1689,44 @@ def _nanpercentile(values, q, axis, **kw):
...
+            The mask of values to replace.
+        src : object
+            The value to replace. It is ignored if regex is False.
+        dst : object
+            The value to be replaced with.
+        convert : bool
+            If true, try to coerce any object types to better types.
```
jreback: add the inplace arg

pandas/core/internals/blocks.py
jreback: can you reorder these args to match as much as possible _replace_single

pandas/core/internals/blocks.py
```diff
@@ -2531,15 +2592,56 @@ def re_replacer(s):
...
         return block
 
+    def _replace_coerce(self, mask=None, src=None, dst=None, inplace=True,
+                        convert=False, regex=False, mgr=None):
+        """
+        Replace value corresponding to the given boolean array with another
+        value.
+
+        Parameters
+        ----------
```
jreback: same as above (you *can* make a shared doc-string if you want here)

pandas/core/internals/managers.py
```diff
@@ -571,12 +573,15 @@ def replace_list(self, src_list, dest_list, inplace=False, regex=False,
         # figure out our mask a-priori to avoid repeated replacements
         values = self.as_array()
 
-        def comp(s):
+        def comp(s, reg=False):
             if isna(s):
```
jreback: can you add a doc-string, rename reg -> regex

pandas/core/internals/managers.py
```diff
@@ -1890,7 +1891,28 @@ def _consolidate(blocks):
     return new_blocks
 
 
-def _maybe_compare(a, b, op):
+def _maybe_compare(a, b, regex=False):
```
jreback: can you make this slightly more verbose name, can you move to pandas/core/ops.py (cc @jbrockmendel good location)?

pandas/tests/series/test_replace.py
```diff
@@ -243,6 +243,13 @@ def test_replace_string_with_number(self):
         expected = pd.Series([1, 2, 3])
         tm.assert_series_equal(expected, result)
 
+    def test_repace_intertwined_key_value_dict(self):
+        # GH 20656
```
jreback: can you add a 1-liner explaining the test in a bit more detail

jreback (changes requested):
lgtm. some cosmetic things

pandas/tests/series/test_replace.py
jbrockmendel: typo repace --> replace (?)

pandas/core/internals/managers.py
jbrockmendel: Do we anticipate using it elsewhere?  If not I'd leave it here at least for now.

peterpanmj: like name it ``_compare_or_regex_match`` ?

jreback (approved):

jreback:
thanks @peterpanmj nice patch!

Issue #21471, opened 2018-06-14
to_csv regression in 0.23.1


francois-a:
Writing to gzip no longer works with 0.23.1:
```
with gzip.open('test.txt.gz', 'wt') as f:
    pd.DataFrame([0,1],index=['a','b'], columns=['c']).to_csv(f, sep='\t')
```
produces corrupted output. This works fine in 0.23.0.

Presumably this is related to #21241 and #21118.

WillAyd:
Please provide a reproducible example:

http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports

Liam3851:
@WillAyd Francois's example is reproducible for me on Windows 7 using master. The output file test.txt.gz is empty instead of containing data. 

If I let pandas do the compression it appears to work fine:
```
df = pd.DataFrame([0,1],index=['a','b'], columns=['c'])
df.to_csv('C:/temp/test.txt.gz', sep='\t', compression='gzip')
```

saidie:
Hi,
I also encountered a `to_csv` problem on 0.23.1 although my case is different to others:

```python
import sys
import pandas as pd
df = pd.DataFrame([0,1])
df.to_csv(sys.stdout)
```

This code writes the dataframe to **a file named `<stdout>`** while it is expected to be printed out to the stdout.

WillAyd:
@Liam3851 thanks - I misread the original post so I see the point now.

@saidie and @wildraid please do not add distinct issues to this. If you feel you have a different issue please open it separately

wildraid:
@WillAyd , I did a quick research.

It seems that all "file-like" objects which cannot be converted string file paths are affected. Gzip wrapper, stdout, FD's - all these problems have the same origin.

Example with FD:
```python
import pandas
import os

with os.fdopen(3, 'w') as f:
    print(f)
    pandas.DataFrame([0, 1]).to_csv(f)
```

Output:
```
<_io.TextIOWrapper name=3 mode='w' encoding='UTF-8'>
Traceback (most recent call last):
  File "gg.py", line 6, in <module>
    pandas.DataFrame([0, 1]).to_csv(f)
  File "/Users/wr/anaconda3/lib/python3.6/site-packages/pandas/core/frame.py", line 1745, in to_csv
    formatter.save()
  File "/Users/wr/anaconda3/lib/python3.6/site-packages/pandas/io/formats/csvs.py", line 166, in save
    f.write(buf)
[...]

PR #21249, opened 2018-05-29
EHN: to_csv compression accepts file-like object

minggli:
- [x] closes #21227 
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

Handle an unsupported case when a file-like object instead of path passed into to_csv with compression. According to documentation, compression keyword requires it to be a filename.

At the moment, when a handle is passed, it appears to be uncompressed.

Tentative enhancement.

Issue #21227, opened 2018-05-28
df.to_csv ignores compression when provided with a file handle


sbarlock:
#### Code at issue

```python
with open(outputfile, "w") as file_handle:
    df2.to_csv(file_handle, compression='gzip')
```
#### Problem description
The written file is not gzip compressed.

#### Expected Output
The output should be a gzip compressed csv file. Similar to what is obtained when using:
```python
df2.to_csv('/path/to/file.csv.gz',compression='gzip')
```

WillAyd:
Maybe related to #21144

minggli:
```
>>> import os
>>> import pandas as pd
>>> from pandas import *
>>>
>>> pd.__version__
u'0.22.0'
>>>
>>> df = DataFrame(100 * [[123, 234, 435]])
>>>
>>> with open('test_compressed', 'w') as fh:
...     df.to_csv(fh, compression='gzip')
...
>>> fh_size = os.path.getsize('test_compressed')
>>> df.to_csv('test_compressed', compression='gzip')
>>> f_size = os.path.getsize('test_compressed')
>>> assert fh_size == f_size
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AssertionError
```

looks like an existing error from previous version 0.22.

toninlg:
Hi,

Should the following code works with this merge?

```
import os
import sys
import pandas as pd
from pandas import *

print(sys.version)
print(pd.__version__)
df = DataFrame(100 * [[123, 234, 435]])
with open('test_compressed.gz', 'w') as fh:
    df.to_csv(fh)

fh_size = os.path.getsize('test_compressed.gz')
df.to_csv('test_compressed', compression='gzip')
f_size = os.path.getsize('test_compressed')

assert fh_size == f_size
```

3.6.7 (default, Dec  6 2019, 07:03:06) [MSC v.1900 64 bit (AMD64)]
0.25.1

3.7.7 (default, May  6 2020, 11:45:54) [MSC v.1916 64 bit (AMD64)]
1.0.5
I have an assertion error in both cases and if I add compression='gzip' to the first to_csv, I have RuntimeWarning: compression has no effect when passing file-like object as input.

Thank you

AvivAvital2:
@toninlg this worked for me
```python
import gzip
with io.StringIO() as buf:
    df.to_csv(buf)
    with open('test_compressed.gz', 'wb') as remote_file:
[...]

PR #21478, opened 2018-06-14
BUG/REG: file-handle object handled incorrectly in to_csv

minggli:
- [x] closes #https://github.com/pandas-dev/pandas/issues/21471
- [x] tests added / passed
- [x] whatsnew entry

This error related to PR #21249 and https://github.com/pandas-dev/pandas/issues/21227. This is never supported use case and to use file-handle in to_csv with compression, the file-object itself should be a compression archive such as:
```
with gzip.open('test.txt.gz', 'wt') as f:
    pd.DataFrame([0,1],index=['a','b'], columns=['c']).to_csv(f, sep='\t')
```

Regressed to 0.22 to_csv with support for zipfile. zipfile doesn't support writing csv strings to a zip archive using a file-handle. So buffer is used to catch the writing and dump into zip archive in one go. The other scenarios remain unchanged.

### DIFF
--- a/doc/source/whatsnew/v0.23.2.txt
+++ b/doc/source/whatsnew/v0.23.2.txt
@@ -56,7 +56,7 @@ Bug Fixes
 
 **I/O**
 
--
+- Bug in :meth:`to_csv` when handling file-like object incorrectly (:issue:`21471`)
 -
 
 **Plotting**
--- a/doc/source/whatsnew/v0.23.2.txt
+++ b/doc/source/whatsnew/v0.23.2.txt
@@ -16,7 +16,7 @@ and bug fixes. We recommend that all users upgrade to this version.
 Fixed Regressions
 ~~~~~~~~~~~~~~~~~
 
--
+- Fixed Regression in :meth:`to_csv` when handling file-like object incorrectly (:issue:`21471`)
 -
 
 .. _whatsnew_0232.performance:
@@ -56,7 +56,7 @@ Bug Fixes
 
 **I/O**
 
-- Bug in :meth:`to_csv` when handling file-like object incorrectly (:issue:`21471`)
+-
 -
 
 **Plotting**
--- a/doc/source/whatsnew/v0.23.2.txt
+++ b/doc/source/whatsnew/v0.23.2.txt
@@ -16,7 +16,7 @@ and bug fixes. We recommend that all users upgrade to this version.
 Fixed Regressions
 ~~~~~~~~~~~~~~~~~
 
-- Fixed Regression in :meth:`to_csv` when handling file-like object incorrectly (:issue:`21471`)
+- Fixed regression in :meth:`to_csv` when handling file-like object incorrectly (:issue:`21471`)
 -
 
 .. _whatsnew_0232.performance:
--- a/pandas/io/formats/csvs.py
+++ b/pandas/io/formats/csvs.py
@@ -174,10 +174,9 @@ class CSVFormatter(object):
             if is_zip:
                 # GH 17778 handles zip compression separately.
                 buf = f.getvalue()
-                try:
+                if hasattr(self.path_or_buf, 'write'):
                     self.path_or_buf.write(buf)
-                except AttributeError:
-                    # when path_or_buf is not file-like, create handle.
+                else:
                     f, handles = _get_handle(self.path_or_buf, self.mode,
                                              encoding=encoding,
                                              compression=self.compression)
--- a/pandas/io/formats/csvs.py
+++ b/pandas/io/formats/csvs.py
@@ -142,6 +142,11 @@ class CSVFormatter(object):
         elif hasattr(self.path_or_buf, 'write'):
             f = self.path_or_buf
             close = False
+            if self.compression:
+                import warnings
+                msg = ("compression has no effect when passing file-like "
+                       "object as input.")
+                warnings.warn(msg, RuntimeWarning, stacklevel=2)
         else:
             f, handles = _get_handle(self.path_or_buf, self.mode,
                                      encoding=encoding,
@@ -178,19 +183,6 @@ class CSVFormatter(object):
                 f.close()
                 for _fh in handles:
                     _fh.close()
-
-            # GH 17778 handles zip compression for byte strings separately.
-            # if path is not None:
-            #     buf = f.getvalue()
-            #     f, handles = _get_handle(path, self.mode,
-            #                              encoding=encoding,
-            #                              compression=self.compression)
-            #     f.write(buf)
-            #
-            #     if not py2zip:
-            #         f.close()
-            #         for _fh in handles:
-            #             _fh.close()
 
     def _save_header(self):
 
--- a/pandas/io/formats/csvs.py
+++ b/pandas/io/formats/csvs.py
@@ -130,17 +130,16 @@ class CSVFormatter(object):
         else:
             encoding = self.encoding
 
+        # GH 21227 internal compression is not used when file-like passed.
         if self.compression and hasattr(self.path_or_buf, 'write'):
             msg = ("compression has no effect when passing file-like "
                    "object as input.")
             warnings.warn(msg, RuntimeWarning, stacklevel=2)
 
-        if isinstance(self.path_or_buf, ZipFile) or (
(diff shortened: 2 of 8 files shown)


wildraid:
I am not a proper reviewer, but I am suspicious about the whole "ZipFile magic" part.

It looks really ugly to me, and this code may potentially cause spike of 2x memory usage due to StringIO accumulation and "getvalue()" call.

I do not fully understand what was the problem with 0.22 implementation. Yes, "read_csv" and "to_csv" were inconsistent, but it contained less "magic", and users could always wrap anything they need with few lines of code.

Also, compression algos usually have additional parameters. Users might want to change defaults later on, and they'll need manual wrapping for that anyway.

Just thoughts.

minggli:
agree that for extremely large zip output, memory usage will increase when writing csv to StringIO first and then call `getvalue()`. With 8G RAM MacOS, enough to write a dataframe [100m, 3]. the memory increase is only when zip compression, for other use cases, there should be no difference as 0.22. the custom class just provides a file-like that accepts string into zip archive when ZipFile class doesn't really offer that. 

The ability to write zip archive is added https://github.com/pandas-dev/pandas/issues/17778 where only read zip was supported, whereas we have round-trip ability for others. the motivation is consistency but also zip compression format is quite common and strange not to be supported.

the zip class is not only for to_csv but provides utility for to_json and to_pickle case too.

doc/source/whatsnew/v0.23.2.txt
```diff
@@ -56,7 +56,7 @@ Bug Fixes
 
 **I/O**
 
--
+- Bug in :meth:`to_csv` when handling file-like object incorrectly (:issue:`21471`)
```
jreback: can you move to the regression section

pandas/io/formats/csvs.py
```diff
@@ -128,19 +128,33 @@ def save(self):
...
-        # file-like output with compression as option. GH 21241, 21118
-        f = StringIO()
-        if not is_file_like(self.path_or_buf):
-            # path_or_buf is path
-            path_or_buf = self.path_or_buf
-        elif hasattr(self.path_or_buf, 'name'):
-            # path_or_buf is file handle
-            path_or_buf = self.path_or_buf.name
+        if self.compression and hasattr(self.path_or_buf, 'write'):
+            import warnings
```
jreback: you can move this import to the top

pandas/io/formats/csvs.py
```diff
@@ -128,19 +128,33 @@ def save(self):
...
-        # PR 21300 uses string buffer to receive csv writing and dump into
-        # file-like output with compression as option. GH 21241, 21118
-        f = StringIO()
-        if not is_file_like(self.path_or_buf):
-            # path_or_buf is path
-            path_or_buf = self.path_or_buf
-        elif hasattr(self.path_or_buf, 'name'):
-            # path_or_buf is file handle
-            path_or_buf = self.path_or_buf.name
+        if self.compression and hasattr(self.path_or_buf, 'write'):
```
jreback: can uou add a comment here (sure the warning says it all, but helpful nonetheless), also an issue reference

pandas/io/formats/csvs.py
```diff
@@ -128,19 +128,33 @@ def save(self):
...
+            import warnings
+            msg = ("compression has no effect when passing file-like "
+                   "object as input.")
+            warnings.warn(msg, RuntimeWarning, stacklevel=2)
+
+        if isinstance(self.path_or_buf, ZipFile) or (
```
jreback: you can just do

``is_zip = isinstance(......) or (.....)``

jreback: comment here on what is_zip means

pandas/io/formats/csvs.py
```diff
@@ -157,13 +171,18 @@ def save(self):
...
-                                         encoding=encoding,
-                                         compression=self.compression)
-                f.write(buf)
+            if is_zip:
+                # GH 17778 handles zip compression separately.
+                buf = f.getvalue()
+                try:
+                    self.path_or_buf.write(buf)
+                except AttributeError:
+                    f, handles = _get_handle(self.path_or_buf, self.mode,
```
jreback: can you add a comment on when the except happens?

pandas/tests/series/test_io.py
```diff
@@ -151,19 +152,19 @@ def test_to_csv_compression(self, s, encoding, compression):
...
-            with open(filename, 'w') as fh:
-                s.to_csv(fh, compression=compression, encoding=encoding,
-                         header=True)
-
+            # test the round trip using file handle - to_csv -> read_csv
+            f, _handles = _get_handle(filename, 'w', compression=compression,
+                                      encoding=encoding)
+            with f:
+                s.to_csv(f, encoding=encoding, header=True)
             result_fh = pd.read_csv(filename, compression=compression,
```
jreback: you can rename this to result (and below)

pandas/tests/test_common.py
```diff
@@ -246,19 +247,31 @@ def test_compression_size(obj, method, compression_only):
...
+        f, _handles = _get_handle(filename, 'w', compression=None)
+        with f:
+            getattr(obj, method)(f)
+            assert not f.closed
         uncompressed = os.path.getsize(filename)
         assert uncompressed > compressed
+
+
+def test_compression_warning(compression_only):
+    df = DataFrame(100 * [[0.123456, 0.234567, 0.567567],
```
jreback: can you add the issue number

jreback (changes requested):
looks good. mostly doc comments. ping on green.

pandas/tests/test_common.py
minggli: added.

pandas/io/formats/csvs.py
minggli: done.

pandas/tests/series/test_io.py
minggli: done.

doc/source/whatsnew/v0.23.2.txt
minggli: done.

pandas/io/formats/csvs.py
```diff
@@ -157,13 +171,19 @@ def save(self):
...
-            if path_or_buf:
-                f, handles = _get_handle(path_or_buf, self.mode,
-                                         encoding=encoding,
-                                         compression=self.compression)
-                f.write(buf)
+            if is_zip:
+                # GH 17778 handles zip compression separately.
+                buf = f.getvalue()
+                try:
+                    self.path_or_buf.write(buf)
```
gfyoung: They say that it's sometimes easier to ask for forgiveness than it is for permission.

That being said, your treatment of `self.path_or_buf.write` is inconsistent here compared to:

https://github.com/pandas-dev/pandas/pull/21478/files#diff-eaa887c826bfb361d98db1ddb668c7deR150

In addition, the logic seems somewhat similar.  Can we potentially deduplicate this logic?

doc/source/whatsnew/v0.23.2.txt
```diff
@@ -16,7 +16,7 @@ and bug fixes. We recommend that all users upgrade to this version.
 Fixed Regressions
 ~~~~~~~~~~~~~~~~~
 
--
+- Fixed Regression in :meth:`to_csv` when handling file-like object incorrectly (:issue:`21471`)
```
gfyoung: nit: "Regression" --> "regression"

pandas/io/formats/csvs.py
minggli: see your point. changed to if else to make it consistent.

doc/source/whatsnew/v0.23.2.txt
minggli: done.

minggli:
@jreback comments carried out.

gfyoung (approved):

jreback (approved):

jreback:
thanks @minggli nice followup patch!

jreback:
@TomAugspurger this might be tricky to backport

minggli:
@jreback @gfyoung happy to help. thanks for reviewing.

Issue #17667, opened 2017-09-25
DataFrame.any(axis={0, 1}) returns inconsistent values for non-zero timedelta


nmusolino:
#### Steps to reproduce

```python
In [1]: import pandas

In [2]: df = pandas.DataFrame({'a': pandas.Series([0, 0]),
   ...:                        't': pandas.Series([pandas.to_timedelta(0, 's'), pandas.to_timedelta(1, 'ms')])})

In [3]: df
Out[3]:
   a               t
0  0        00:00:00
1  0 00:00:00.001000

In [4]: df.dtypes
Out[4]:
a              int64
t    timedelta64[ns]
dtype: object

In [5]: df.any(axis=0)
Out[5]:
a    False
t     True
dtype: bool

In [6]: df.any(axis=1)
Out[6]:
0    False
1    False
dtype: bool
```

#### Problem description

pandas.DataFrame.any() returns "whether any element is True over the selected axis".

In the example above, the results of `df.any(axis=ax)` is clearly inconsistent for the two axis directions (cells [5] and [6]). 

If an element in column `"t"` leads to a result of `True` in `df.any(axis=0)`, then the same element should lead to a result of `True` in `df.any(axis=1)`.

#### Expected output

As a user, I expected `DataFrame.any()` to return True for rows or columns with a _non-zero_ timedelta value.  That is, I expected a `timedelta` column to have the same semantics as integers: zero values are treated as False, while other values are treated as True.

In other words, this is the output I expected:

```
In [6]: df.any(axis=1)
Out[6]:
0    False
1    True
dtype: bool
```

#### Output of ``pd.show_versions()``

<details>

```
In [7]: pandas.show_versions()

INSTALLED VERSIONS
------------------
commit: None
python: 3.4.5.final.0
python-bits: 64
OS: Windows
OS-release: 7
machine: AMD64
processor: Intel64 Family 6 Model 60 Stepping 3, GenuineIntel
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: None.None

pandas: 0.19.1
nose: 1.3.7
pip: 9.0.1
setuptools: 27.2.0
Cython: 0.24.1
numpy: 1.11.2
scipy: 0.18.1
statsmodels: 0.6.1
xarray: 0.8.2
IPython: 5.1.0
sphinx: 1.4.8
patsy: 0.4.1
dateutil: 2.6.0
pytz: 2016.7
blosc: 1.5.0
[...]

PR #21486, opened 2018-06-14
API/COMPAT: support axis=None for logical reduction (reduce over all axes)

### LINKED ISSUE #19976: CI: NumPy logical reductions (any, all) fail on DataFrame with NumPy master
https://travis-ci.org/pandas-dev/pandas/jobs/348215410#L2717

```
[gw1] linux -- Python 3.6.4 /home/travis/miniconda3/envs/pandas/bin/python
self = <pandas.tests.frame.test_analytics.TestDataFrameAnalytics object at 0x7f86fa5d54a8>
axis = 0
    @pytest.mark.parametrize("axis", [0, 1, None])
    def test_clip_against_frame(self, axis):
        df = DataFrame(np.random.randn(1000, 2))
        lb = DataFrame(np.random.randn(1000, 2))
        ub = lb + 1
    
>       clipped_df = df.clip(lb, ub, axis=axis)
pandas/tests/frame/test_analytics.py:2066: 
pandas/core/generic.py:5605: in clip
    if np.any(pd.isnull(lower)):
self = 0    False
1    False
dtype: bool
    def __nonzero__(self):
        raise ValueError("The truth value of a {0} is ambiguous. "
                         "Use a.empty, a.bool(), a.item(), a.any() or a.all()."
>                        .format(self.__class__.__name__))
E       ValueError: The truth value of a Series is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().
```

TomAugspurger:
- [x] closes #19976
- [x] tests added / passed
- [x] whatsnew entry

This is the minimal fix, just to get np.all / np.any working again.

Some followup items:

1. Expand to all aggregations, not just logical ones
2. Do logical reductions blockwiwse: https://github.com/pandas-dev/pandas/issues/17667. Currently, we do `DataFrame.values`, which isn't necessary for logical reductions.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -6865,7 +6865,7 @@ class DataFrame(NDFrame):
                 values = self.values
                 result = f(values)
 
-                if (filter_type == 'bool' and values.dtype.kind == 'O' and
+                if (filter_type == 'bool' and is_object_dtype(values) and
                         axis is None):
                     # work around https://github.com/numpy/numpy/issues/10489
                     # TODO: combine with hasattr(result, 'dtype') further down
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -6844,12 +6844,17 @@ class DataFrame(NDFrame):
 
     def _reduce(self, op, name, axis=0, skipna=True, numeric_only=None,
                 filter_type=None, **kwds):
-        axis = self._get_axis_number(axis)
+        if axis is None and filter_type == 'bool':
+            labels = None
+            constructor = None
+        else:
+            # TODO: Make other agg func handle axis=None properly
+            axis = self._get_axis_number(axis)
+            labels = self._get_agg_axis(axis)
+            constructor = self._constructor
 
         def f(x):
             return op(x, axis=axis, skipna=skipna, **kwds)
-
-        labels = self._get_agg_axis(axis)
 
         # exclude timedelta/datetime unless we are uniform types
         if axis == 1 and self._is_mixed_type and self._is_datelike_mixed_type:
@@ -6859,6 +6864,13 @@ class DataFrame(NDFrame):
             try:
                 values = self.values
                 result = f(values)
+
+                if (filter_type == 'bool' and values.dtype.kind == 'O' and
+                        axis is None):
+                    # work around https://github.com/numpy/numpy/issues/10489
+                    # TODO: combine with hasattr(result, 'dtype') further down
+                    # hard since we don't have `values` down there.
+                    result = np.bool_(result)
             except Exception as e:
 
                 # try by-column first
@@ -6925,7 +6937,9 @@ class DataFrame(NDFrame):
                 if axis == 0:
                     result = coerce_to_dtypes(result, self.dtypes)
 
-        return Series(result, index=labels)
+        if constructor is not None:
+            result = Series(result, index=labels)
+        return result
 
     def nunique(self, axis=0, dropna=True):
         """
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -9057,15 +9057,14 @@ _bool_doc = """
 
 Parameters
 ----------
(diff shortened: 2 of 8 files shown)


pandas/core/generic.py
```diff
@@ -9055,8 +9057,16 @@ def _doc_parms(cls):
 
 Parameters
 ----------
-axis : int, default 0
-    Select the axis which can be 0 for indices and 1 for columns.
+axis : {None, 0 or 'index', 1 or 'columns'}, default None
```
jorisvandenbossche: I don't think `default None` is correct? 
The default for *our* method is still 0? It's only when doing `np.all(..)` that we respect the `axis=None` ?

pandas/core/series.py
```diff
@@ -3238,6 +3238,8 @@ def _reduce(self, op, name, axis=0, skipna=True, numeric_only=None,
 
         """
         delegate = self._values
+        if axis is None:
+            axis = self._stat_axis_number
```
jorisvandenbossche: Why is this still needed?

pandas/core/generic.py
```diff
@@ -9651,22 +9671,17 @@ def _make_logical_function(cls, name, name1, name2, axis_descr, desc, f,
...
                   axis_descr=axis_descr, examples=examples, see_also=see_also)
     @Appender(_bool_doc)
-    def logical_func(self, axis=None, bool_only=None, skipna=None, level=None,
+    def logical_func(self, axis=0, bool_only=None, skipna=True, level=None,
                      **kwargs):
         nv.validate_logical_func(tuple(), kwargs, fname=name)
-        if skipna is None:
-            skipna = True
-        if axis is None:
-            axis = self._stat_axis_number
```
jorisvandenbossche: Shouldn't there be an override of this in case of Panel? (which is the only case where the stat_axis differs from 0)

pandas/core/frame.py
```diff
@@ -6859,6 +6864,13 @@ def f(x):
             try:
                 values = self.values
                 result = f(values)
+
+                if (filter_type == 'bool' and values.dtype.kind == 'O' and
```
jreback: use is_object_dtype

pandas/tests/frame/test_analytics.py
```diff
@@ -1159,11 +1159,34 @@ def test_any_all(self):
         self._check_bool_op('any', np.any, has_skipna=True, has_bool_only=True)
         self._check_bool_op('all', np.all, has_skipna=True, has_bool_only=True)
 
-        df = DataFrame(randn(10, 4)) > 0
```
jreback: I would make a new test function here

jreback (changes requested):

pandas/core/series.py
jorisvandenbossche: Can you have a look at this one?

jorisvandenbossche:
Tests are still failing: `np.any`/`np.all` don't seem to work on timedelta data (at least for certain versions of numpy, locally for me it is failing on 1.13)

pandas/core/series.py
TomAugspurger: Sorry, missed this earlier. We still call `self._get_axis_number(axis)` to validate that the user passed `axis` is correct. I'll modify things to be clearer.

jorisvandenbossche (approved):

jreback:
I think you need to change

```
    @pytest.mark.xfail(
        not _np_version_under1p15,
        reason="failing under numpy-dev gh-19976")
    @pytest.mark.parametrize("axis", [0, 1, None])
    def test_clip_against_frame(self, axis):
        df = DataFrame(np.random.randn(1000, 2))
```

in ``pandas/tests/frame/test_analytics.py`` which was the original failing test, to use your new decorator?

otherwise lgtm.

TomAugspurger:
Good catch. I've removed the xfail entirely, since I think it will pass on all versions now.

PR #20891, opened 2018-05-01
Read from multiple <tbody> within a <table>

adamhooper:
- [x] closes #20690 
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

Issue #17074, opened 2017-07-25
TestReadHtml.test_computer_sales_page - what's it doing?


jowens:
@chris-b1 or anyone else, help a brother out? Can you tell me what this test does? It's just expecting the parser to throw an error? The output from the test code (where it's failing) is at the bottom. It's a pretty weird HTML file.

<img width="269" alt="computer_sales_page_html" src="https://user-images.githubusercontent.com/976970/28598596-882515f2-7159-11e7-902f-1262a7c5d983.png">

Now, if I call it with my current in-progress code as `dfs = pd.read_html('computer_sales_page.html', header=[0, 1])`, I see:

```
Index([         (u'Unnamed: 0_level_0', u'Unnamed: 0_level_1'),
                (u'Unnamed: 1_level_0', u'Unnamed: 1_level_1'),
                     (u'Three months ended April?30', u'2013'),
              u'(u'Three months ended April\xa030', '2013').1',
       (u'Three months ended April?30', u'Unnamed: 4_level_1'),
                     (u'Three months ended April?30', u'2012'),
              u'(u'Three months ended April\xa030', '2012').1',
                (u'Unnamed: 7_level_0', u'Unnamed: 7_level_1'),
                       (u'Six months ended April?30', u'2013'),
                u'(u'Six months ended April\xa030', '2013').1',
        (u'Six months ended April?30', u'Unnamed: 10_level_1'),
                       (u'Six months ended April?30', u'2012'),
                u'(u'Six months ended April\xa030', '2012').1',
              (u'Unnamed: 13_level_0', u'Unnamed: 13_level_1')],
      dtype='object')
```

and if I call it without a header argument (`dfs = pd.read_html('computer_sales_page.html')`), I see:

```
Index([   (u'Unnamed: 0_level_0', u'Unnamed: 0_level_1', u'Unnamed: 0_level_2'),
          (u'Unnamed: 1_level_0', u'Unnamed: 1_level_1', u'Unnamed: 1_level_2'),
                      (u'Three months ended April?30', u'2013', u'In millions'),
                u'(u'Three months ended April\xa030', '2013', 'In millions').1',
        (u'Three months ended April?30', u'Unnamed: 4_level_1', u'In millions'),
                      (u'Three months ended April?30', u'2012', u'In millions'),
                u'(u'Three months ended April\xa030', '2012', 'In millions').1',
                 (u'Unnamed: 7_level_0', u'Unnamed: 7_level_1', u'In millions'),
[...]

Issue #17073, opened 2017-07-25
Q: correct behavior for read_html with rowspan/colspan for DataFrames?


jowens:
No code, just a question for proper behavior for rowspan/colspan with `read_html` of an HTML table into a DataFrame. (I'm not asking what currently happens with `read_html` now. I'm asking what _should_ happen.)

Below is a simple HTML table that uses both colspan (a,e,i,o,u) and rowspan (Fruit, schwa, honk, and the rightmost 0 in the table). It renders identically on each of {Chrome, Firefox, Safari}. With these renderers, both rowspans and colspans are basically rendered midway through the span, either vertically (rowspan) or horizontally (colspan).

<img width="446" alt="fruit_html" src="https://user-images.githubusercontent.com/976970/28595080-25038b14-7147-11e7-9e27-bb57ca9fb043.png">

Now, let's say we wanted to import this into pandas with `read_html`. It seems to me the behavior should be different for a pandas DataFrame than for a renderer:
- The header should have a MultiIndex, where the first column is `Fruit` and the second column is a combination of `a` and `Long`, etc. We don't "fill" a rowspan (the first column shouldn't be two `Fruit`s), but we do "fill" a colspan (`a` would appear in the 2nd and 3rd columns).
- The body should "fill" a rowspan or colspan with the provided values. So the rightmost column, instead of having one zero and two blanks on the three rows, should have a zero for each of the three rows. One would think a span in a DataFrame context within a body would mean "fill in the value for each cell in the span".

If this was the case, it would imply that we treat rowspans differently in header and body.

- If we "fill" a rowspan in a header, then we just repeat the header value in the MultiIndex output, which doesn't seem like what we want.
- If we don't "fill" a rowspan in the body, we leave some cells in the DataFrame blank, which also seems misguided.

I put the DataFrame that I _think_ we want below. It incorporates different behavior for rowspan for header and body. One thing I don't know, though: If I don't "fill" the rowspan name for rowspan > 1, what do I put instead? `None`? empty string? `False`? What does the input to `TextParser` look like when some column names are "taller" than others?

Thoughts? @chris-b1?   (relevant to https://github.com/pandas-dev/pandas/issues/17054)

```
             a          e          i          o          u               
    Fruit Long Short Long Short Long Short Long Short Long Short schwa honk
0   Apple    0     1    0     0    0     0    0     0    0     0     1    0
1  Banana    0     3    0     0    0     0    0     0    0     0     0    0
[...]

PR #17089, opened 2017-07-26
ENH: GH17054: read_html() handles rowspan/colspan and infers headers

jowens:
- [x] closes #17054
 - [x] tests added / passed
 - [x] passes ``git diff upstream/master -u -- "*.py" | flake8 --diff``
 - [x] whatsnew entry

PR #21487, opened 2018-06-14
read_html: Handle colspan and rowspan

### LINKED ISSUE #17054: ENH: read_html to handle rowspan, colspan

```python
import pandas as pd
pd.read_html('https://www.ssa.gov/policy/docs/statcomps/supplement/2015/5h.html')[0]
```

This has complex table headings:

annual_statistical_supplement__2015_-_beneficiary_families_with_oasdi_benefits_in_current-payment_status__5_h_

`read_html` output begins with:

```
                                                 Year                            Retired-worker families        Survivor families             Disabled-worker families                                Unnamed: 4_level_0                              Unnamed: 5_level_0 Unnamed: 6_level_0 Unnamed: 7_level_0 Unnamed: 8_level_0 Unnamed: 9_level_0 Unnamed: 10_level_0 Unnamed: 11_level_0  \
                                          Worker only                                  Worker and wife?a  Non-disabled widow only        Widowed mother or father and?                                       Worker only                            Worker, wife,?b and?  Worker and spouse Unnamed: 7_level_1 Unnamed: 8_level_1 Unnamed: 9_level_1 Unnamed: 10_level_1 Unnamed: 11_level_1
                                                  All                                                Men                    Women                              1?child                                        2?children                              3 or more children                All                Men              Women            1?child  2 or more children Unnamed: 11_level_2
0                                                 NaN                                 Number?(thousands)                      NaN                                  NaN                                               NaN                                             NaN                NaN                NaN                NaN                NaN                 NaN                 NaN
1                                                1945                                                416                      338                                   78                                               181                                              95              86.00              48.00              24.00              .?.?.               .?.?.               .?.?.
```

(row 0 of the output is probably something one would have to manually eliminate)
#### Problem description

For HTML headings with rowspan and colspan elements, `read_html` has undesirable behavior. Basically `read_html` packs all heading `<th>` elements in any particular row to the left, so any particular column no longer has any association with the `<th>` elements that are actually above it in the HTML table.

Ample discussion here about the analogous pandas+Excel test case: https://github.com/pandas-dev/pandas/issues/4679

Relevant web discussions:
- https://stackoverflow.com/questions/23703638/bug-in-pandas-read-html-method
- https://stackoverflow.com/questions/28763891/what-should-i-do-when-tr-has-rowspan
- https://johnricco.github.io/2017/04/04/python-html/   (using his example)

This may be an issue with the underlying parsers and cannot be solved well in pandas. This appears to be the behavior with both lxml and bs4/html5lib.

#### Expected Output

Each column should be associated with the `<th>` elements above it in the table. This might be a multi-row column name (as it is now) (a `MultiIndex`?) or a tuple (presumably if the argument `tupleize_cols` is set to `True`). Instead, currently, column _n_ is associated with the _n_ th `<th>` entry in the table row regardless of the settings of rowspan/colspan.

It may be this is possible to do properly in current pandas in which case I apologize for filing the issue (but I'd be happy to know how to do it).

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 2.7.13.final.0
python-bits: 64
OS: Darwin
OS-release: 16.7.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.US-ASCII
LOCALE: None.None

pandas: 0.20.3
pytest: None
pip: 9.0.1
setuptools: 36.2.0
Cython: 0.26
numpy: 1.13.1
scipy: 0.19.1
xarray: None
IPython: 5.3.0
sphinx: 1.6.3
patsy: None
dateutil: 2.6.0
pytz: 2017.2
blosc: None
bottleneck: 1.2.1
tables: 3.4.2
numexpr: 2.6.2
feather: None
matplotlib: 2.0.2
openpyxl: 2.4.7
xlrd: 1.0.0
xlwt: None
xlsxwriter: None
lxml: 3.7.3
bs4: 4.5.3
html5lib: 1.0b10
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
pandas_gbq: None
pandas_datareader: None
</details>

### LINKED ISSUE #21641: `header` argument in `read_html()` ignores empty trs, only within thead

With a table that has no `<thead>`, `read_html()`'s `header` argument looks at the correct row:

```
>>> pandas.read_html('''<table><tbody><tr><th></th><th></th></tr><tr><th>A</th><th>B</th></tr><tr><td>a</td><td>b</td></tr></tbody></table>''', header=0)
[  Unnamed: 0 Unnamed: 1
0          A          B
1          a          b]
```

... but when a row is in a `<thead>`, Pandas behaves differently:

```
>>> pandas.read_html('''<table><thead><tr><th></th><th></th></tr><tr><th>A</th><th>B</th></tr></thead><tbody><tr><td>a</td><td>b</td></tr></tbody></table>''', header=0)
[   A  B
0  a  b]
```

#### Problem description

Structurally, within HTML, `<thead>` and `<tbody>` serve the same purpose. It's unnatural to specify different `<tr>` indices depending on which element contains them. XPath and CSS selectors (e.g., `//table//tr`) are consistent, and `read_html()` isn't.

In code, the error is that Pandas deletes empty `<tr>`s within `<thead>` before passing them to `pandas.io.html._data_to_frame()`, which is where the `header` argument is parsed.

If the issue has not been resolved there, go ahead and file it in the issue tracker.

<details>

>>> pandas.show_versions()
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.13.0-37-generic
machine: x86_64
processor:
byteorder: little
LC_ALL: None
LANG: C.UTF-8
LOCALE: en_US.UTF-8
pandas: 0.23.0
pytest: None
pip: 10.0.1
setuptools: 39.2.0
Cython: None
numpy: 1.11.3
scipy: None
pyarrow: 0.8.0
xarray: None
IPython: None
sphinx: 1.7.5
patsy: None
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: 2.5.4
xlrd: 1.0.0
xlwt: None
xlsxwriter: None
lxml: 4.2.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: None
pymysql: None
psycopg2: 2.7.1 (dt dec pq3 ext lo64)
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

adamhooper:
This is essentially a rebased and squashed #17054 (mad props to @jowens
for doing all the hard thinking). My tweaks:

* test_computer_sales_page (see #17074) no longer tests for ParserError,
  because the ParserError was a bug caused by missing colspan support.
  Now, test that MultiIndex works as expected.
* I respectfully removed the fill_rowspan argument from #17073. Instead,
  the virtual cells created by rowspan/colspan are always copies of the
  real cells' text. This prevents _infer_columns() from naming virtual
  cells as "Unnamed: ..."
* I removed a small layer of abstraction to respect #20891 (multiple
  tbody support), which was implemented after @jowens' pull request.
  Now _HtmlFrameParser has _parse_thead_trs, _parse_tbody_trs and
  _parse_tfoot_trs, each returning a list of trs. That let me remove
  _parse_tr, Making All The Tests Pass.
* That caused a snowball effect. lxml does not fix malformed thead, as
  tested by spam.html. The previous hacky workaround was in
  _parse_raw_thead, but the new _parse_thead_trs signature returns nodes
  instead of text. The new hacky solution: return the thead itself,
  pretending it's a tr. This works in all the tests. A better solution
  is to use html5lib with lxml; but that might belong in a separate pull
  request.

- [x] closes #17054
- [x] closes #21641
- [x] tests added / passed
- [x] whatsnew entry


adamhooper:
Er, I mean, a rebased and squashed #17089.

jowens:
You're a hero, @adamhooper.

pandas/io/html.py
```diff
@@ -358,47 +336,115 @@ def _build_doc(self):
...
-            for tbody in tbodies:
-                raw_data.extend(self._parse_tr(tbody))
-        else:
-            raw_data.extend(self._parse_tr(table))
+        if not header_rows:
+            # The table has no <thead>. Treat first all-<th> rows as headers.
+            while body_rows and all(self._equals_tag(t, 'th') for t in
+                                    self._parse_td(body_rows[0])):
+                # this row should be a header row, move it from body to header
+                header_rows.append(body_rows.pop(0))
```
WillAyd: Hmm not so sure about this. What happens if there's no table header but the user explicitly provides a `header` argument?

pandas/io/html.py
```diff
@@ -358,47 +336,115 @@ def _build_doc(self):
...
+        footer_rows = self._parse_tfoot_tr(table_html)
 
-        if tbodies:
-            for tbody in tbodies:
-                raw_data.extend(self._parse_tr(tbody))
-        else:
-            raw_data.extend(self._parse_tr(table))
+        if not header_rows:
+            # The table has no <thead>. Treat first all-<th> rows as headers.
+            while body_rows and all(self._equals_tag(t, 'th') for t in
```
WillAyd: Could be overlooking something but why are you using `while` here - are you not just testing the truthiness of `body_rows`? If so, `if` would be preferable

pandas/io/html.py
```diff
@@ -358,47 +336,115 @@ def _build_doc(self):
...
+                                    self._parse_td(body_rows[0])):
+                # this row should be a header row, move it from body to header
+                header_rows.append(body_rows.pop(0))
+
+        if not footer_rows:
+            # The table has no <tfoot>. Treat last all-<th> rows as footers.
+            while body_rows and all(self._equals_tag(t, 'th') for t in
+                                    self._parse_td(body_rows[-1])):
+                # this row should be a footer row, move it from body to footer
+                footer_rows.insert(0, body_rows.pop())
```
WillAyd: Stylistically would prefer append, since that's also what you do above with `header_rows`

pandas/io/html.py
```diff
@@ -358,47 +336,115 @@ def _build_doc(self):
...
+            extracted_row = self._parse_td(row)
+            cols_text = [_remove_whitespace(
+                self._text_getter(col)) for col in extracted_row]
+            col_colspans = [int(col.get('colspan', 1))
+                            for col in extracted_row]
+            col_rowspans = [int(col.get('rowspan', 1))
+                            for col in extracted_row]
+            # expand cols using col_colspans
+            # maybe this can be done with a list comprehension, dunno
+            cols = list(zip(
```
WillAyd: Not sure how to read this on initial glance, though it seems like you are conscious of that from the comment. If you can't think of a better way to do via code any way you can speak to what you are trying to accomplish?

pandas/io/html.py
```diff
@@ -554,10 +600,9 @@ def _text_getter(self, obj):
         return obj.text_content()
 
     def _parse_td(self, row):
-        return row.xpath('.//td|.//th')
-
-    def _parse_tr(self, table):
-        return table.xpath('.//tr')
+        # Look for direct descendents only: the "row" element here may be a
```
WillAyd: Are you looking for children or descendants? Conflict between comment and code here?

pandas/io/html.py
```diff
@@ -637,41 +688,30 @@ def _build_doc(self):
...
 
-    def _parse_thead(self, table):
-        return table.xpath('.//thead')
+        for thead in table.xpath('.//thead'):
+            rows.extend(thead.xpath('./tr'))
 
-    def _parse_tfoot(self, table):
-        return table.xpath('.//tfoot')
+            # lxml does not clean up the clearly-erroneous
+            # <thead><th>foo</th><th>bar</th></thead>.
```
WillAyd: Probably my lack of understanding but what is erroneous about this? Missing `tr`?

pandas/io/html.py
```diff
@@ -637,41 +688,30 @@ def _build_doc(self):
...
 
-    def _parse_tbody(self, table):
-        return table.xpath('.//tbody')
+    def _parse_thead_tr(self, table):
+        rows = []
 
-    def _parse_thead(self, table):
-        return table.xpath('.//thead')
+        for thead in table.xpath('.//thead'):
+            rows.extend(thead.xpath('./tr'))
```
WillAyd: Does this work if the `thead` has multiple rows?

pandas/tests/io/test_html.py
```diff
@@ -359,7 +349,7 @@ def test_thousands_macau_stats(self):
                              attrs={'class': 'style1'})
         df = dfs[all_non_nan_table_index]
 
-        assert not any(s.isna().any() for _, s in df.iteritems())
+        assert not any(s.isnull().any() for _, s in df.iteritems())
```
WillAyd: What was this changed for?

pandas/io/html.py
```diff
@@ -210,32 +208,14 @@ def __init__(self, io, match, attrs, encoding, displayed_only):
...
-            A list of row elements.
-
-        text_getter : callable
-            A callable that gets the text from an individual node. This must be
-            defined by subclasses.
-
-        column_finder : callable
-            A callable that takes a row node as input and returns a list of the
-            column node in that row. This must be defined by subclasses.
+        """Parse and return all tables from the DOM.
```
WillAyd: Docstrings should always start on the line after the triple quotes

pandas/io/html.py
```diff
@@ -210,32 +208,14 @@ def __init__(self, io, match, attrs, encoding, displayed_only):
...
 
         Returns
         -------
-        data : list of list of strings
+        tables : list of parsed (header, body, footer) tuples from tables
```
WillAyd: Typically for Returns unless you return more than one object you don't need to provide name(s), just types on the first line and a description on the subsequent line (also add period at end of description)

pandas/io/html.py
```diff
@@ -257,7 +237,7 @@ def _parse_td(self, obj):
 
         Parameters
         ----------
-        obj : node-like
+        obj : an HTML row element
```
WillAyd: Is this supposed to be a type or a description? Not sure the exactly terminology but I feel like the type of `node-like` was preferable with this as the description (on subsequent line)

WillAyd (changes requested):

pandas/io/html.py
adamhooper: That's a great question, but I don't think this pull request affects that one way or another.

From my reading of `_data_to_frame(**kwargs)`: `_data_to_frame()` creates one big table, `head + body + foot`. If the user does not supply a `header` kwarg, then the function sets `header = lrange(len(head))`

That makes this more-reliable header inference backwards-incompatible: after this pull request, users who didn't supply `header` on ugly HTML tables that had multi-row headers will get a MultiIndex instead of an Index. Those users will be negatively-affected, and they may want to add an explicit `header` argument to mimic previous behavior.

Let's put that in perspective. Correct handling of rowspan/colspan is also backwards-incompatible. It could break existing code for a giant fraction of `read_html()` users who have coded workarounds for headers pandas misplaced.

I suspect most tables in the wild with multi-row headers also use rowspan/colspan. (Otherwise, why would they have multiple rows?)

Header inference doesn't change pandas' behavior for either of its sample HTML files (spam.html and computer_sales_page.html); the rowspan/colspan fix makes pandas change behavior with both. That's a random sample, but it matches my suspicion: rowspan/colspan is going to change pandas' behavior to an enormous degree, eclipsing the change in header inference.

pandas/io/html.py
adamhooper: `body_rows` changes length each iteration. I just made the conditional shorter to make that clearer.

pandas/io/html.py
adamhooper: Added a comment on why not to do that

pandas/io/html.py
adamhooper: I rewrote this logic as procedural.

pandas/io/html.py
adamhooper: added comment, thanks

pandas/io/html.py
adamhooper: That's the word I was looking for! d'oh, thanks :)

pandas/io/html.py
adamhooper: yes. (Previous code returned a list of <thead>, treating it as a list of lists. New code effectively flattens the return value.)

pandas/tests/io/test_html.py
adamhooper: Bad merge. Reverted.

pandas/io/html.py
adamhooper: Changed throughout the file.

pandas/io/html.py
adamhooper: Changed throughout the file

pandas/io/html.py
adamhooper: changed

pandas/io/html.py
```diff
@@ -258,161 +260,261 @@ def _parse_td(self, obj):
...
-        body = self._parse_raw_tbody(table)
-        footer = self._parse_raw_tfoot(table)
+    def _parse_thead_tbody_tfoot(self, table_html):
+        """
+        Given a table, return parsed header, body, and foot.
+
+        Header and body are lists-of-lists. Top level list is a list of
+        rows. Each row is a list of str text.
+
+        Logic: Use <thead>, <tbody>, <tfoot> elements to identify
```
jreback: can you put this in Notes

pandas/io/html.py
```diff
@@ -590,6 +697,12 @@ def _parse_tables(self, doc, match, kwargs):
                              .format(patt=pattern))
         return tables
 
+    def _equals_tag(self, obj, tag):
```
jreback: aren't these the same as in the super class?

jreback (changes requested):
looks pretty good. can you add a whatsnew note (0.24.0). if you can do it in 1 line great (e.g. what a user would care about the changes, what can it do now that it couldn't before). also a sub-section would be ok too.

pandas/io/html.py
WillAyd: I thought this too on initial glance but the attribute of `obj` used to determine equality differs across implementations. An alternate approach could be to make that attribute a property that the subclasses need to implement, though not sure that is that much cleaner

pandas/io/html.py
adamhooper: done

pandas/io/html.py
WillAyd: Ah OK that makes more sense. Is that a safe operation though? Can't say I've been personally bitten by this but SO seems to advise against modifying the container that you loop over:

https://stackoverflow.com/questions/1637807/modifying-list-while-iterating

pandas/io/html.py
adamhooper: My answer is either "no" or "I don't understand the question" :).

But I did notice and nix `_contains_tag()`, which did not belong. So thanks for focusing here :).

pandas/io/html.py
WillAyd: Ah OK that makes more sense. Is that a safe operation though? Can't say I've been personally bitten by this but SO seems to advise against modifying the container that you loop over:

https://stackoverflow.com/questions/1637807/modifying-list-while-iterating

adamhooper:
@jreback Thanks for the review! There was already an entry in whatsnew, but I rephrased it such that users will better be able to predict how the change will affect their data. Do you think that's enough, or should I add a subsection?

pandas/io/html.py
```diff
@@ -258,161 +260,263 @@ def _parse_td(self, obj):
...
-        else:
-            raw_data.extend(self._parse_tr(table))
+    def _expand_colspan_rowspan(self, rows):
+        """
+        Given a list of <tr>s, return a list of text rows that copy cell
+        text across rowspans/colspans.
+
+        Parameters
+        ----------
+        rows : list of <tr>s
```
WillAyd: Just say list after hyphen on the first line to denote type. I suppose you can add a comment on the subsequent line for a description

pandas/io/html.py
adamhooper: I'd argue that this loop doesn't iterate over the _list_. It just iterates over a condition.

pandas/io/html.py
adamhooper: The bizarre inheritance used in this module makes "node-like" values different across implementations. It isn't duck-typing: they're completely different beasts. That's what it was like before I started this pull request.

pandas/io/html.py
```diff
@@ -258,161 +260,263 @@ def _parse_td(self, obj):
...
+        for tr in rows:
+            texts = []  # the output for this row
+            next_remainder = []
+
+            index = 0
+            tds = self._parse_td(tr)
+            for td in tds:
+                # Append texts from previous rows with rowspan>1 that come
+                # before this <td>
+                while remainder and remainder[0][0] <= index:
```
WillAyd: Thanks for the comment before about iteration. Can you help point me to where `remainder` first gets values populated for it? I'm still somewhat concerned about the number of mutations done to that variable. 

I have the feeling some of this can be better and more safely expressed using enumerate but direction on where to begin with this would be helpful

WillAyd (changes requested):

pandas/io/html.py
adamhooper: `remainder` is initialized on line 447 and flipped with `next_remainder` on line 485.

It's valid to worry about an evil web page forcing lots of `remainder` mutations ... to a point. If I were generating a website to foil Pandas, here's what I'd do:

```html
<table>
  <tr>
    <td colspan="2147483647">This will crash your puny parser.</td>
  </tr>
</table>
```

I don't expect this to spin forever because of mutations: I expect it to crash with out-of-memory as it grows an enormous list.

Real-world tables will be fine. Usually colspan/rowspan are only in header cells, and there are rarely many items here.

Big-O analysis: for a table with `N` output cells (N = WxH), each `.pop(0)` generates a cell. That means there are at most `N` list-removals, meaning at most `N` list-insertions. Each insertion is `O(1)` but each removal is `O(W)`. (According to https://wiki.python.org/moin/TimeComplexity, list pop is O(k).) That removal is the slowest operation, so running time is `O(NW)`.

Shall I switch to a `collection.deque` to make it `O(N)`?

adamhooper: Incidentally: I, too, started by writing `enumerate()`. But that fell apart. The index of a cell in the output depends on two inputs: both `td`'s index in `row` and the number of items in `remainder` that pre-empt it.

An alternate strategy is be to make `remainder` a list of mostly-`None`, sometimes-tuple values and modify them in-place. That, too, would be `O(N)` ... but it wouldn't involve as many allocations.

pandas/io/html.py
```diff
@@ -258,161 +260,268 @@ def _parse_td(self, obj):
...
+
+        if not header_rows:
+            # The table has no <thead>. Move the top all-<th> rows from the
+            # <tbody> to the <thead>. (This is a common case because many
+            # tables in the wild have no <thead> or <tfoot>
+            while body_rows and row_is_all_th(body_rows[0]):
+                header_rows.append(body_rows.pop(0))
+
+        if not footer_rows:
+            # The table has no <tfoot>. Treat last all-<th> rows as footers.
```
WillAyd: Hmm what is the point in doing this? In the previous implementation it would just return an empty list for the footer right? Did that cause some kind of issue?

pandas/io/html.py
```diff
@@ -258,161 +260,268 @@ def _parse_td(self, obj):
...
+            while body_rows and row_is_all_th(body_rows[-1]):
```
WillAyd: What is this doing? I'm under the impression this would always be false on account of the `row_is_all_th` condition?

pandas/io/html.py
WillAyd: Have to think through that in more detail. I'm less concerned about the choice of container; I just feel like there's a more readable and concise way to approach this that would be easier for future contributors / maintainers to use

pandas/io/html.py
```diff
@@ -258,161 +260,268 @@ def _parse_td(self, obj):
...
+                    prev_i, prev_text, prev_rowspan = remainder.pop(0)
+                    texts.append(prev_text)
+                    if prev_rowspan > 1:
+                        next_remainder.append((prev_i, prev_text,
+                                               prev_rowspan - 1))
+                    index += 1
+
+                # Append the text from this <td>, colspan times
+                text = _remove_whitespace(self._text_getter(td))
+                rowspan = int(self._attr_getter(td, 'rowspan') or 1)
```
WillAyd: I think this can be simplified in one of two ways:

1. Simply use `td.get('rowspan', 1)` if td allows for that in both implementations
2. Move the implementation of `_attr_getter` into the base class

For point 2 above it looks like the subclasses both have the same implementation so it would be easier to just move to the base class

pandas/tests/io/test_html.py
```diff
@@ -448,6 +438,44 @@ def test_header_and_one_column(self):
...
+            </thead>
+            <tbody>
+                <tr>
+                    <td>Ukraine</td>
+                    <th>Odessa</th>
+                    <td>1944</td>
+                </tr>
+            </tbody>
+        </table>''')
+        data2 = StringIO('''<table>
```
WillAyd: Generally can add spacing between variables to make the tests more readable

pandas/tests/io/test_html.py
```diff
@@ -448,6 +438,44 @@ def test_header_and_one_column(self):
...
+            <tbody>
+                <tr>
+                    <td>Ukraine</td>
+                    <th>Odessa</th>
+                    <td>1944</td>
+                </tr>
+            </tbody>
+        </table>''')
+        res1 = self.read_html(data1)
+        res2 = self.read_html(data2, header=0)
```
WillAyd: What's the reason for `header=0` here?

pandas/tests/io/test_html.py
```diff
@@ -448,6 +438,44 @@ def test_header_and_one_column(self):
...
+        assert_framelist_equal(res1, res2)
```
WillAyd: Minor nit but there's a pseudo-standard to use `result` and `expected` when evaluating tests. Would be preferable to say `result1` and `result2` here (and in other tests)

pandas/tests/io/test_html.py
```diff
@@ -641,6 +669,224 @@ def test_different_number_of_rows(self):
...
+                   <tr>
+                     <th colspan="1">X</th>
+                     <th>Y</th>
+                     <th rowspan="1">Z</th>
+                     <th>W</th>
+                   </tr>
+                 </table>"""
+        expected = self.read_html(expected)[0]
+        res = self.read_html(out)[0]
+        tm.assert_frame_equal(expected, res)
```
WillAyd: Typically we use `result` as the first argument to this function and `expected` as the second. Would be nice to do that in these tests so failure output follows that pattern

WillAyd (changes requested):
Thanks for the iterations - getting closer

pandas/io/html.py
adamhooper: Great catch! It's cruft -- just nixed it. `_data_to_frame` simply does `body += foot` anyway.

pandas/io/html.py
adamhooper: Footer rows are often all-`<th>`: in a typical table, there may be an all-`<th>` "totals" row.

But this code cannot impact the user (because `_data_to_frame` simply does `body += foot` anyway), so I've nixed it.

pandas/io/html.py
adamhooper: Okay -- pulled up to the base class (though that seems strange to me)

pandas/tests/io/test_html.py
adamhooper: done in lots and lots of tests

pandas/tests/io/test_html.py
adamhooper: Moved the comment right there -- it's a regression test for #5048

pandas/tests/io/test_html.py
adamhooper: Okay, I've renamed variables (and fiddled with whitespace) in many, many tests.

pandas/tests/io/test_html.py
adamhooper: Okay, I've renamed variables (and fiddled with whitespace) in many, many tests.

adamhooper:
@jreback, @WillAyd I've implemented all the changes you requested. Could you please take a look?

@WillAyd I tested out an alternative algorithm, as you had concerns about the one in this pull request. The alternative is at https://github.com/adamhooper/pandas/commit/67dea69f604251d6de8185c8dd8c82f15eedfa2a. The diff is +53, -38. Maybe you find it more readable, though?

pandas/tests/io/test_html.py
```diff
@@ -441,11 +433,48 @@ def test_header_and_one_column(self):
...
         expected = DataFrame(data={'Header': 'first'}, index=[0])
-        result = self.read_html(data)[0]
+
+        tm.assert_frame_equal(result, expected)
+
+    def test_thead_without_tr(self):
+        """
+        Ensure parser adds <tr> within <thead> on malformed HTML.
+        """
+        expected = self.read_html('''<table>
```
WillAyd: Generally the `expected` value should be constructed directly and not via read_html - can you update any of the tests you've added to do that?

pandas/tests/io/test_html.py
```diff
@@ -765,11 +1054,11 @@ def test_keep_default_na(self):
                     </table>"""
 
         expected_df = DataFrame({'a': ['N/A', 'NA']})
-        html_df = read_html(html_data, keep_default_na=False)[0]
+        html_df = self.read_html(html_data, keep_default_na=False)[0]
```
WillAyd: What was the point of changing some of these?

pandas/tests/io/test_html.py
```diff
@@ -660,14 +940,15 @@ def test_parse_dates_combine(self):
...
     def test_computer_sales_page(self):
         data = os.path.join(DATA_PATH, 'computer_sales_page.html')
-        with tm.assert_raises_regex(ParserError,
-                                    r"Passed header=\[0,1\] are "
-                                    r"too many rows for this "
-                                    r"multi_index of columns"):
-            self.read_html(data, header=[0, 1])
+        # This table is unique because it has no <thead>, and its <th>-only
+        # rows are underneath an initial <td>-only row that has no content.
+        # After #kipping the empty row, header=[0,1] picks the two <th>-only
```
WillAyd: minor typo of skipping

pandas/tests/io/test_html.py
```diff
@@ -660,14 +940,15 @@ def test_parse_dates_combine(self):
 
     def test_computer_sales_page(self):
         data = os.path.join(DATA_PATH, 'computer_sales_page.html')
-        with tm.assert_raises_regex(ParserError,
-                                    r"Passed header=\[0,1\] are "
-                                    r"too many rows for this "
-                                    r"multi_index of columns"):
-            self.read_html(data, header=[0, 1])
+        # This table is unique because it has no <thead>, and its <th>-only
```
WillAyd: Hmm not sure about this. Certainly a crazy example but this changes the behavior of the arguments passed to `header` now, no?

pandas/io/html.py
```diff
@@ -637,41 +742,32 @@ def _build_doc(self):
...
-                for tr in trs:
-                    cols = [_remove_whitespace(x.text_content()) for x in
-                            self._parse_td(tr)]
+        for thead in table.xpath('.//thead'):
+            rows.extend(thead.xpath('./tr'))
 
-                    if any(col != '' for col in cols):
-                        res.append(cols)
-        return res
+            # HACK: lxml does not clean up the clearly-erroneous
```
WillAyd: The comment here mentions a situation we are missing `tr` elements but `test_header_inferred_from_th_elements` is constructed without a `thead` element instead - is that by design?

pandas/io/html.py
```diff
@@ -554,10 +657,9 @@ def _text_getter(self, obj):
         return obj.text_content()
 
     def _parse_td(self, row):
-        return row.xpath('.//td|.//th')
-
-    def _parse_tr(self, table):
-        return table.xpath('.//tr')
+        # Look for direct children only: the "row" element here may be a
```
WillAyd: Conceptually makes sense but is this comment necessary? We aren't enforcing that row be of a certain tag

WillAyd (changes requested):

pandas/tests/io/test_html.py
adamhooper: Done! Looks much nicer now.

pandas/tests/io/test_html.py
adamhooper: These tests are all run twice: once with `lxml` and once with `bs4`. But `read_html()`'s logic is, "try `lxml`, and if it fails use `bs4`." So it overcomes the entire purpose of the unit tests -- if the lxml backend _crashes_ while we're testing it, we won't know because `read_html()` will fall back to `bs4`.

From my reading of the code, `self.read_html()` is always correct, and `read_html()` is always a bug.

pandas/tests/io/test_html.py
adamhooper: fixed

pandas/tests/io/test_html.py
WillAyd: OK thanks for the clarification. I suppose these are fine to leave as is though if there are more better to do in a separate change to keep that cleanup orthogonal to the original issue

pandas/tests/io/test_html.py
adamhooper: This pull request doesn't change the behavior of the `header` argument. It changes the behavior one gets when the `header` argument is _omitted_. That change has no effect on this specific test.

When it comes to this specific unit test ("Test some old table in the wild"):

* Prior to this pull request, pandas erroneously crashed on this table. The unit test was bad -- it tested for a bug.
* This pull request includes a hack involving th-only rows. That hack would _seem_ to infer the correct `header=[0,1]` for this specific table, but it doesn't because this specific table is weird

This unit test now tests that pandas _works_, instead of testing that it is broken. The comment explains why the `header=[0,1]` is needed -- since the HTML makes it look like it shouldn't be.

pandas/io/html.py
adamhooper: That test is unrelated to this hack. This hack is tested in `test_thead_without_tr`.

pandas/tests/io/test_html.py
WillAyd: It looks like it broke previously for `header=[0, 1]` but worked for `header=[1, 2]`. Now it is working for `header=[0,1]` and that returns the same value that previously was returned with `header=[1,2]`, right?

Not saying that is necessarily a bad thing but looking for confirmation as in the off-chance something like this does occur in the wild that is a pretty subtle change for users that may have been expecting the former behavior

pandas/io/html.py
adamhooper: It is nonsensical that `_parse_td()` be called with a `thead` element, because it is nonsensical that a `thead` element contain `td` or `th` children. The `_parse_td` comment says parameter is a DOM `<tr>` node.

It's an ugly hack. I think that nonsense deserves a comment.

pandas/tests/io/test_html.py
adamhooper: No, that's not what changed. It broke before because it didn't support rowspan/colspan. The deletion of empty rows is existing behavior.

adamhooper: urgh, looking at the diff with master instead of my own. I see that `header=[1,2]` used to not crash. I'll investigate more to figure out exactly why that worked.

adamhooper: Okay, you're absolutely right -- the behavior changed.

Previously, pandas removed all-empty rows from the `<thead>` but _not_ from the `<tbody>`.

Now, pandas removes all-empty rows from anywhere in the table.

So `header` changed on this table because its header rows are inside the `<tbody>`.

Should I revert that change? It was unintentional. But the original behavior appears to be unintentional to me, too -- why would `<thead>` and `<tbody>` behave differently?

adamhooper: References:

* #13434 suggests read_html() should infer multiple table rows
* #15242 implements by deleting all-empty rows from `<thead>`

From these discussions, I gather that pandas shouldn't auto-delete data, but auto-deleting in the `<thead>` is kinda okay because otherwise we'd get a ParserError by default.

I'll auto-delete empty header rows. It seems to me `_data_to_frame()` is the correct place to do that. @WillAyd does that sound appropriate? Then I can revert the change to this unit test -- I was misunderstanding the error.

adamhooper: Fixed in commit ed120d8. I added new tests for behavior implied in #15242.

Unfortunately, `test_spam_header` tested for a bug -- it specified `header=1` instead of the intent, `header=2`. The test passed because of a bug in the previous implementation: indexing happened _after_ deleting empty header rows.

Let's track the bugs through history to see why `header=2` is correct:

1. Original implementation, commit b31c033f (2013) -- uses `header=0` to select `<td>` elements within `<thead>` that are not part of a `<tr>`. The bug that made this work: pandas erroneously treated `<thead>` as equivalent to `<tr>`, merging all `<th>` and `<td>` within the `<thead>` into one big row.
2. commit a8723a4b (2013): changed the test to `header=1` and changed the value it's looking for to be the value from the first `<tr>` in the `<tbody>`. Seems like a silly change. The author probably tested a body cell because lxml and bs4 began treating the (malformed) header differently as a result of some completely-unrelated change.
3. commit 0ab081345 (#15242, 2017): kept `header=1` unchanged, because two new behaviors cancelled out. First, `<thead>` handling was corrected to look for `<tr>`s inside the `<thead>` (meaning one would need to set `header=2`). Second, `_parse_raw_thead` was _incorrectly_ changed to delete empty rows (deleting one row from the header brings us back to `header=1`). That change made the unit test pass, but the unit test was testing the wrong behavior all along.
4. Today, I fixed the bug introduced in 2017, meaning `header=2` finds the cell that's been referenced since a8723a4b in 2013.

@WillAyd what do you think of this change? I think anybody would agree the previous behavior was a bug. And it seems silly to re-implement the bug in this pull request in the name of avoiding feature creep.

I imagine this applies to only a tiny number of tables in the wild.

WillAyd: Thanks for the clarification. I do agree that `header=2` appears to be more correct. Do you think there's an issue with other index-based keywords (like `skiprows`) in the current implementation? 

OK to address in a separate issue but it would be nice to have mention of this in a whatsnew in some fashion (again separate issue is fine)

WillAyd:
Need to clean up merge conflicts and maybe add whatsnew reference for proper indexing of tables with NA data. Otherwise lgtm

adamhooper:
Urgh. I did a rebase on upstream/master and pushed with `--force` and now Appveyor is complaining. Was that a bad idea on my part? How should I fix?

WillAyd:
I don't think the AppVeyor issue is a result of your rebase. That said, you'd be better off fetching master and merging it into your branch rather than rebase (maintains better comment history).

Looks like you have some file conflicts still anyway, so might was well resolve them locally and try merging from master into your branch this time before next push

adamhooper:
Indeed -- I rebased over adamhooper/master instead of pandas/master. I've learned my lesson: merge, don't rebase.

Fixed by ... er ... rebasing. Correctly, this time.

adamhooper:
^^ I've noticed a few other pull requests cause ResourceWarning failures on Travis. Does that error relate to this pull request?

WillAyd:
Unrelated - that intermittently seems to pop up on Travis

WillAyd (approved):

WillAyd:
@adamhooper can you fix merge conflict?

@jreback thoughts on your end?

adamhooper:
Merged :)

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -297,7 +298,7 @@ MultiIndex
 I/O
 ^^^
 
--
+- :func:`read_html()` no longer ignores all-whitespace ``<tr>`` within ``<thead>`` when considering the ``skiprows`` and ``header`` arguments. Previously, users had to decrease their ``header`` and ``skiprows`` values on such tables to work around the issue. (:issue:`21641`)
```
jreback: this is all ok, but would a subsection with a mini-example be instructive to the user about the revised functinaility? not saying 100% need to, but if a simple enough example (or even an image) would be helpful here

pandas/io/html.py
```diff
@@ -258,161 +261,257 @@ def _parse_td(self, obj):
...
 
-    def _build_table(self, table):
-        header = self._parse_raw_thead(table)
-        body = self._parse_raw_tbody(table)
-        footer = self._parse_raw_tfoot(table)
+    def _parse_thead_tbody_tfoot(self, table_html):
+        """
+        Given a table, return parsed header, body, and foot.
+
+        Notes
```
jreback: put Notes at the end (after Parameters / Returns)

pandas/io/html.py
```diff
@@ -258,161 +261,257 @@ def _parse_td(self, obj):
...
-            for tbody in tbodies:
-                raw_data.extend(self._parse_tr(tbody))
-        else:
-            raw_data.extend(self._parse_tr(table))
+    def _expand_colspan_rowspan(self, rows):
+        """
+        Given a list of <tr>s, return a list of text rows.
 
-        return self._parse_raw_data(raw_data)
+        Notes
```
jreback: same generally

pandas/io/html.py
jreback: ok happy to take follow-ups to make this simpler / easier for implementation.

jreback (changes requested):
just some doc comments, otherwise lgtm.

doc/source/whatsnew/v0.24.0.txt
adamhooper: I'm going to display my ignorance of the whatsnew format: ... where would this go?

jreback: oh, you make a new subsection in the top somewhere, e.g. model after this

``Series and Index Data-Dtype Incompatibilities`` 

just add a new one below and you can make an extended example

adamhooper:
@jreback Thanks -- added whatsnew entry.

Excited to see: is this last commit the final one?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -168,6 +168,120 @@ Current Behavior:
...
+            </tbody>
+          </table>
+        """)
+    Out [1]:
+    [   A  B  C
+    0  1  2  2]
+
+Calls that relied on the previous behavior will need to be changed.
+
+Also, :func:`read_html` previously ignored some ``<tr>`` elements when called
```
WillAyd: See what @jreback thinks but I don't consider this second half necessary as it addresses a minor bug. Otherwise lgtm

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -168,6 +168,120 @@ Current Behavior:
     ...
     OverflowError: Trying to coerce negative values to unsigned integers
 
+read_html Incompatibilities
```
jreback: how about read_html enhancements!

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -168,6 +168,120 @@ Current Behavior:
...
+            </tbody>
+          </table>
+        """)
+    Out [1]:
+    [   A  B   C
+    0  1  2 NaN]
+
+Current Behavior:
+
+.. code-block:: ipython
```
jreback: you can make this an ipython block (e.g. actually run it for current)

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -168,6 +168,120 @@ Current Behavior:
...
+
+:func:`read_html` previously ignored ``colspan`` and ``rowspan`` attributes.
+Now it understands them, treating them as a sequence of cells with the same
+value.
+
+Previous Behavior:
+
+.. code-block:: ipython
+
+    In [1]: pd.read_html("""
```
jreback: to avoid repeating the code, make an ipython block above where you define the html itself). then the preivious uses a code-block to show the output, and the new uses an ipython block

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -168,6 +168,120 @@ Current Behavior:
...
+                <td colspan="2">1</td><td>2</td>
+              </tr>
+            </tbody>
+          </table>
+        """)
+    Out [1]:
+    [   A  B  C
+    0  1  2  2]
+
+Calls that relied on the previous behavior will need to be changed.
```
jreback: this sentence is not needed

doc/source/whatsnew/v0.24.0.txt
jreback: yeah you can just list as a 1-liner in the io section is ok, no example for this needed (the latter half)

jreback (changes requested):

doc/source/whatsnew/v0.24.0.txt
adamhooper: Ah, I get it now. (I actually read the docs to preview what it looks like.) Thanks -- looks like a super improvement.

jreback (approved):

jreback:
thanks @adamhooper very nice!

adamhooper:
Woo-hoo! Thank you all for your guidance. What a thrill :).

clytemnestra:
Hello, when will this be included into a release? I see it's being included in v0.24, but when will that happen? I'm having this exact problem and would love to benefit from this fix.

PR #21493, opened 2018-06-15
PERF: improve speed of nans in CategoricalIndex

topper-123:
This is a minor follow-up to #21369.

```python
>>> n = 100_000
>>> ci = pd.CategoricalIndex(['a']*n + ['b']*n + ['c']*n + [np.nan])
>>> np.nan in ci
19.5 us  # master
114 ns  # this PR
```

Using ``self.hasnans`` to check for nans is faster than ``self.isna().any()`` because  it's cached.


jreback:
thanks @topper-123

PR #21503, opened 2018-06-15
De-duplicate code for indexing with list-likes of keys

toobaz:
- [x] tests passed

Just refactoring, removing duplicated code (I don't think the bug in ``Index._reindex_non_unique`` was actually appearing anywhere)

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -3627,7 +3627,7 @@ class Index(IndexOpsMixin, PandasObject):
             else:
 
                 # need to retake to have the same size as the indexer
-                indexer[~check] = 0
+                indexer[~check] = -1
 
                 # reset the new indexer to account for the new size
                 new_indexer = np.arange(len(self.take(indexer)))
--- a/pandas/core/indexing.py
+++ b/pandas/core/indexing.py
@@ -688,7 +688,8 @@ class _NDFrameIndexer(_NDFrameIndexerBase):
         if isinstance(indexer, tuple):
 
             # flatten np.ndarray indexers
-            ravel = lambda i: i.ravel() if isinstance(i, np.ndarray) else i
+            def ravel(i):
+                return i.ravel() if isinstance(i, np.ndarray) else i
             indexer = tuple(map(ravel, indexer))
 
             aligners = [not com.is_null_slice(idx) for idx in indexer]
--- a/pandas/core/indexing.py
+++ b/pandas/core/indexing.py
@@ -1132,33 +1132,31 @@ class _NDFrameIndexer(_NDFrameIndexerBase):
         """
         o = self.obj
         ax = o._get_axis(axis)
-        try:
-            # Have the index compute an indexer or return None
-            # if it cannot handle:
-            indexer, keyarr = ax._convert_listlike_indexer(key,
-                                                           kind=self.name)
-            # We only act on all found values:
-            if indexer is not None and (indexer != -1).all():
-                self._validate_read_indexer(key, indexer, axis,
-                                            raise_missing=raise_missing)
-                return ax[indexer], indexer
 
-            if ax.is_unique:
-                # If we are trying to get actual keys from empty Series, we
-                # patiently wait for a KeyError later on - otherwise, convert
-                if len(ax) or not len(key):
-                    key = self._convert_for_reindex(key, axis)
-                indexer = ax.get_indexer_for(key)
-                keyarr = ax.reindex(keyarr)[0]
-            else:
-                keyarr, indexer, new_indexer = ax._reindex_non_unique(keyarr)
-
-            self._validate_read_indexer(keyarr, indexer,
-                                        o._get_axis_number(axis),
+        # Have the index compute an indexer or return None
+        # if it cannot handle:
+        indexer, keyarr = ax._convert_listlike_indexer(key,
+                                                       kind=self.name)
+        # We only act on all found values:
+        if indexer is not None and (indexer != -1).all():
+            self._validate_read_indexer(key, indexer, axis,
                                         raise_missing=raise_missing)
-            return keyarr, indexer
-        except (KeyError, IndexingError) as detail:
-            raise self._exception(detail)
+            return ax[indexer], indexer
+
+        if ax.is_unique:
+            # If we are trying to get actual keys from empty Series, we
+            # patiently wait for a KeyError later on - otherwise, convert
+            if len(ax) or not len(key):
+                key = self._convert_for_reindex(key, axis)
+            indexer = ax.get_indexer_for(key)
+            keyarr = ax.reindex(keyarr)[0]
+        else:
+            keyarr, indexer, new_indexer = ax._reindex_non_unique(keyarr)
+
+        self._validate_read_indexer(keyarr, indexer,
+                                    o._get_axis_number(axis),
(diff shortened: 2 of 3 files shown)


toobaz:
(there is still room for improvement, by the way, but I will need to look in the ``Index`` methods)

toobaz:
ASV run (of ``indexing.py``):

```
     [9e982e18]       [c08a39fa]
+       146±0.7ms        167±0.7ms     1.14  frame_methods.Iteration.time_iteritems_indexing
-         103±2μs         93.2±2μs     0.91  indexing.DataFrameNumericIndexing.time_loc
-         213±2μs          193±1μs     0.91  indexing.DataFrameNumericIndexing.time_iloc_dups
-         250±2μs          224±1μs     0.90  indexing.DataFrameStringIndexing.time_boolean_rows
-         162±1μs          144±1μs     0.89  indexing.IntervalIndexing.time_getitem_list
-      53.1±0.2μs       46.3±0.3μs     0.87  indexing.IntervalIndexing.time_getitem_scalar
-     2.33±0.01ms      2.00±0.02ms     0.86  indexing.DataFrameNumericIndexing.time_loc_dups
-         494±8ns          424±5ns     0.86  indexing.MethodLookup.time_lookup_iloc
-         727±4μs         623±10μs     0.86  indexing.NumericSeriesIndexing.time_loc_list_like(<class 'pandas.core.indexes.numeric.Int64Index'>)
-        88.6±2μs       74.9±0.6μs     0.85  indexing.IntervalIndexing.time_loc_scalar
-     1.67±0.07ms      1.34±0.02ms     0.80  indexing.NumericSeriesIndexing.time_loc_array(<class 'pandas.core.indexes.numeric.Int64Index'>)
-     1.90±0.09ms      1.44±0.01ms     0.76  indexing.NumericSeriesIndexing.time_getitem_array(<class 'pandas.core.indexes.numeric.Int64Index'>)
-     1.77±0.07ms      1.30±0.01ms     0.73  indexing.NumericSeriesIndexing.time_ix_array(<class 'pandas.core.indexes.numeric.Int64Index'>)
-         799±8μs          519±4μs     0.65  indexing.NumericSeriesIndexing.time_ix_list_like(<class 'pandas.core.indexes.numeric.Int64Index'>)
-         319±5ms          201±4ms     0.63  indexing.NumericSeriesIndexing.time_ix_array(<class 'pandas.core.indexes.numeric.Float64Index'>)
-       307±0.9ms          185±2ms     0.60  indexing.NumericSeriesIndexing.time_ix_list_like(<class 'pandas.core.indexes.numeric.Float64Index'>)

```

toobaz:
@jreback @jorisvandenbossche ping

jreback (commented):
assumed this to make some other of your PRs simpler if this is a precursor?

toobaz:
> assumed this to make some other of your PRs simpler if this is a precursor?

Not sure I understand. This transforms three copies of the same code into one (and in doing so removes some superfluous calls, explaining the performance improvements). So yes, incidentally it will also make it easier to further rationalize the indexing code, and yes, I plan to work on it.

jreback:
@toobaz no that's what I was asking. thanks.

pandas/core/indexes/base.py
```diff
@@ -3627,7 +3627,7 @@ def _reindex_non_unique(self, target):
             else:
 
                 # need to retake to have the same size as the indexer
-                indexer[~check] = 0
+                indexer[~check] = -1
```
jreback: what effect did this have? e.g. was this a bug before or just not used?

pandas/core/indexing.py
```diff
@@ -927,28 +928,8 @@ def _multi_take(self, tup):
...
-                    key = self._convert_for_reindex(key, axis)
-                indexer = ax.get_indexer_for(key)
-                keyarr = ax.reindex(keyarr)[0]
-                self._validate_read_indexer(keyarr, indexer,
-                                            o._get_axis_number(axis))
-                d[axis] = (keyarr, indexer)
+            d = {axis: self._get_listlike_indexer(key, axis)
+                 for (key, axis) in zip(tup, o._AXIS_ORDERS)}
             return o._reindex_with_indexers(d, copy=True, allow_dups=True)
         except (KeyError, IndexingError) as detail:
```
jreback: This can *only* raise KeyError now, right?

jreback: and in fact, maybe can just use ``self._exception()`` in ``_get_listlike_indexer``? then can simply remove the try/except (here)? (and I think you do this below)

pandas/core/indexing.py
```diff
@@ -1124,63 +1105,83 @@ def _getitem_axis(self, key, axis=None):
...
-            return self.obj._take(inds, axis=axis)
-        else:
+        Returns
+        -------
+        keyarr: Index
+            New index (coinciding with 'key' if the axis is unique)
+        values : array-like
+            An indexer for the return object; -1 denotes keys not found
+        """
+        try:
```
jreback: can you move ops that won't fail outside the try

pandas/core/indexing.py
```diff
@@ -1124,63 +1105,83 @@ def _getitem_axis(self, key, axis=None):
...
+                                        o._get_axis_number(axis),
+                                        raise_missing=raise_missing)
+            return keyarr, indexer
+        except (KeyError, IndexingError) as detail:
+            raise self._exception(detail)
 
-                if new_indexer is not None:
-                    result = self.obj._take(indexer[indexer != -1], axis=axis)
+    def _getitem_iterable(self, key, axis=None):
+        if axis is None:
```
jreback: can you add a doc-string here

pandas/core/indexing.py
```diff
@@ -1124,63 +1105,83 @@ def _getitem_axis(self, key, axis=None):
...
-                        copy=True, allow_dups=True)
+        self._validate_key(key, axis)
 
-                else:
-                    self._validate_read_indexer(key, indexer, axis)
-                    result = self.obj._take(indexer, axis=axis)
+        labels = self.obj._get_axis(axis)
 
-                return result
+        if com.is_bool_indexer(key):
```
jreback: can you add a comment on these cases (e.g. what is being checked)

pandas/core/indexing.py
```diff
@@ -1214,6 +1220,10 @@ def _validate_read_indexer(self, key, indexer, axis):
                     u"None of [{key}] are in the [{axis}]".format(
                         key=key, axis=self.obj._get_axis_name(axis)))
 
+            if not(self.name == 'loc' and not raise_missing):
```
jreback: can you add a comment here

jreback (changes requested):
looks good on the cleanup, a few comments.

pandas/core/indexes/base.py
toobaz: To the best of my understanding, it was not used: not in the sense that this line was not hit, but because the locations for missing keys were then taken from ``new_indexer`` (see three lines below), where they were correctly marked with -1.

pandas/core/indexing.py
toobaz: In fact, I think the entire block was just superfluous: if we get here, it means we have collections of keys, and if a key is not found, it will raise later.

jreback (approved):

jreback:
very nice @toobaz 

more like this very welcome!

PR #17087, opened 2017-07-26
Use `python-requests` library, if installed, to support broad http(s) scenarios including Basic Authentication

NanSoftStu:
Use `python-requests` ([Python HTTP Requests for Humans](https://github.com/requests/requests)) , if installed, to enable additional scenarios as Basic Authentication, permit self-signed SSL certs, and other advanced scenarios.

A new optional parameter  `url_params` is added into `read_csv`, `read_html`, `read_excel` and `read_json` to allow passing in either a dict containing authentication, etc or even a requests session. It also allows passing in basic auth in url: http://<usr>:<pwd>@<fqdn>:<port>/  and fixes #Issue #16910 Also, better handling of specified `:<port>` in url.

If `python-requests` library is not installed, existing behavior prior to this check-in should continue as-is, which directly calls `urllib\urllib3`

 - [x] closes #16910, #16716
 - [x] tests added / passed
 - [x] passes ``git diff upstream/master -u -- "*.py" | flake8 --diff``
 - [x] whatsnew entry

PR #21504, opened 2018-06-15
wrap urlopen with requests

### LINKED ISSUE #16716: read_csv from HTTPs + basic-auth + custom port throws an error (urlopen error)

```python
import pandas as pd
df = pd.read_csv('https://user-name:pwd@my.fqdn.com:8080/get_content.csv')
# fails with urlopen error [Errno 11003] getaddrinfo failed Python2.7 as well as Python 3.x
```
#### Problem description

HTTPS basic auth is very common. This URL format works in Excel, other text editors, etc.  This url works in `requests ` library.  It seems like the scenario doesnt work because the underlying `urlopen` doesnt work
```python
# I think it fails because the underlying code here fails in pandas/io/common.py:
from urllib2 import urlopen as _urlopen
b = _urlopen('https://un-name:pwd@my.fqdn.com:8080/get_content.csv')
# fails with urlopen error [Errno 11003] getaddrinfo failed
```

see [stackoverflow issue urllib basic auth](https://stackoverflow.com/questions/635113/python-urllib2-basic-http-authentication-and-tr-im)
Only way to overcome this is using `requests` + `StringIO` ?

#### Expected Output
be able to get a CSV loaded dataframe

<details>
python: 2.7.13.final.0
python-bits: 64
pandas: 0.20.1
</details>

ocefpaf:
- [X] closes #16716
- [ ] tests added / passed
- [ ] whatsnew entry

Alternative to https://github.com/pandas-dev/pandas/pull/17087

I'll write the whatsnew entry, tests, and the session option for `html`, `json`, and `excel` if this path is OK with the devs.

Ping @skynss who is the original author of #17087 and @gfyoung who reviewed it.

Note that the main difference in this approach is that I made the option slightly simpler by allowing only the `session` instead of `http_params`.

You can see an example of this in action in:

http://nbviewer.jupyter.org/urls/gist.githubusercontent.com/ocefpaf/d6e9ab2c3569ff8fa181fc7885b6524d/raw/5d868fd4cbe2b81f84ccab7760b80251ef6e4651/pandas_test.ipynb

### DIFF
--- a/pandas/io/common.py
+++ b/pandas/io/common.py
@@ -104,7 +104,7 @@ def _urlopen(url, session=None):
             r = session.get(url)
         else:
             r = requests.get(url)
-        r.raise_for_status
+        r.raise_for_status()
         content = r.content
     except ImportError:
         r = urlopen(url)
--- a/pandas/io/common.py
+++ b/pandas/io/common.py
@@ -215,6 +215,14 @@ def get_filepath_or_buffer(filepath_or_buffer, encoding=None,
                          or buffer
     encoding : the encoding to use to decode py3 bytes, default is 'utf-8'
     mode : str, optional
+    compression : {'infer', 'gzip', 'bz2', 'zip', 'xz', None}, default 'infer'
+        For on-the-fly decompression of on-disk data. If 'infer' and
+        `filepath_or_buffer` is path-like, then detect compression from the
+        following extensions: '.gz', '.bz2', '.zip', or '.xz' (otherwise no
+        decompression). If using 'zip', the ZIP file must contain only one data
+        file to be read in. Set to None for no decompression.
+
+    .. versionadded:: 0.18.1 support for 'zip' and 'xz' compression.
 
     Returns
     -------
--- a/pandas/io/common.py
+++ b/pandas/io/common.py
@@ -30,14 +30,13 @@ _NA_VALUES = {'-1.#IND', '1.#QNAN', '1.#IND', '-1.#QNAN', '#N/A N/A', '#N/A',
 
 if compat.PY3:
     from urllib.request import urlopen, pathname2url
-    _urlopen = urlopen
     from urllib.parse import urlparse as parse_url
     from urllib.parse import (uses_relative, uses_netloc, uses_params,
                               urlencode, urljoin)
     from urllib.error import URLError
     from http.client import HTTPException  # noqa
 else:
-    from urllib2 import urlopen as _urlopen
+    from urllib2 import urlopen as urlopen2
     from urllib import urlencode, pathname2url  # noqa
     from urlparse import urlparse as parse_url
     from urlparse import uses_relative, uses_netloc, uses_params, urljoin
@@ -46,10 +45,10 @@ else:
     from contextlib import contextmanager, closing  # noqa
     from functools import wraps  # noqa
 
-    # @wraps(_urlopen)
+    # @wraps(urlopen2)
     @contextmanager
     def urlopen(*args, **kwargs):
-        with closing(_urlopen(*args, **kwargs)) as f:
+        with closing(urlopen2(*args, **kwargs)) as f:
             yield f
 
 
@@ -89,6 +88,34 @@ def _is_url(url):
         return parse_url(url).scheme in _VALID_URLS
     except Exception:
         return False
+
+
+def _urlopen(url, session=None):
+    compression = None
+    content_encoding = None
+    try:
+        import requests
+        if session:
+            if not isinstance(session, requests.sessions.Session):
+                raise ValueError(
+                    'Expected a requests.sessions.Session object, '
+                    'got {!r}'.format(session)
+                )
+            r = session.get(url)
+        else:
+            r = requests.get(url)
+        r.raise_for_status
+        content = r.content
+    except ImportError:
+        r = urlopen(url)
+        content = r.read()
+        content_encoding = r.headers.get('Content-Encoding', None)
+    r.close()
+    if content_encoding == 'gzip':
+        # Override compression based on Content-Encoding header.
+        compression = 'gzip'
+    reader = BytesIO(content)
+    return reader, compression
 
 
 def _expand_user(filepath_or_buffer):
@@ -177,7 +204,7 @@ def is_gcs_url(url):
 
 
 def get_filepath_or_buffer(filepath_or_buffer, encoding=None,
-                           compression=None, mode=None):
+                           compression=None, mode=None, session=None):
     """
     If the filepath_or_buffer is a url, translate and return the buffer.
     Otherwise passthrough.
@@ -199,13 +226,7 @@ def get_filepath_or_buffer(filepath_or_buffer, encoding=None,
     filepath_or_buffer = _stringify_path(filepath_or_buffer)
 
     if _is_url(filepath_or_buffer):
-        req = _urlopen(filepath_or_buffer)
-        content_encoding = req.headers.get('Content-Encoding', None)
-        if content_encoding == 'gzip':
-            # Override compression based on Content-Encoding header
-            compression = 'gzip'
-        reader = BytesIO(req.read())
-        req.close()
+        reader, compression = _urlopen(filepath_or_buffer, session=session)
         return reader, encoding, compression, True
 
     if is_s3_url(filepath_or_buffer):
--- a/pandas/io/common.py
+++ b/pandas/io/common.py
@@ -106,11 +106,11 @@ def _urlopen(url, session=None):
             r = requests.get(url)
         r.raise_for_status()
         content = r.content
+        r.close()
     except ImportError:
-        r = urlopen(url)
-        content = r.read()
-        content_encoding = r.headers.get('Content-Encoding', None)
-    r.close()
+        with urlopen(url) as r:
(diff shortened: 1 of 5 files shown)


pandas/io/common.py
```diff
@@ -47,10 +46,10 @@
     from contextlib import contextmanager, closing  # noqa
     from functools import wraps  # noqa
 
-    # @wraps(_urlopen)
+    # @wraps(urlopen2)
     @contextmanager
     def urlopen(*args, **kwargs):
-        with closing(_urlopen(*args, **kwargs)) as f:
+        with closing(urlopen2(*args, **kwargs)) as f:
```
ocefpaf: I'm not sure this workaround is used anywhere.
Maybe we can remove this?

pandas/io/common.py
```diff
@@ -92,6 +91,34 @@ def _is_url(url):
         return False
 
 
+def _urlopen(url, session=None):
+    compression = None
+    content_encoding = None
+    try:
+        import requests
```
ocefpaf: There is no need to check if `requests` is installed when using a `session` object b/c `requests` is needed to create the object, this is slitly simpler than the approach in #17087 and introduces less code to maintain.

pandas/io/common.py
```diff
@@ -192,13 +219,7 @@ def get_filepath_or_buffer(filepath_or_buffer, encoding=None,
...
 
     if _is_url(filepath_or_buffer):
-        req = _urlopen(filepath_or_buffer)
-        content_encoding = req.headers.get('Content-Encoding', None)
-        if content_encoding == 'gzip':
-            # Override compression based on Content-Encoding header
-            compression = 'gzip'
-        reader = BytesIO(req.read())
-        req.close()
+        reader, compression = _urlopen(filepath_or_buffer, session=session)
```
ocefpaf: `compression` is not mentioned in the docs and I wonder if `session` should.
If so we can use the template/Appender approach, just not sure how to do that in a clean way.

pandas/io/common.py
```diff
@@ -92,6 +91,34 @@ def _is_url(url):
...
+        if session:
+            if not isinstance(session, requests.sessions.Session):
+                raise ValueError(
+                    'Expected a requests.sessions.Session object, '
+                    'got {!r}'.format(session)
+                )
+            r = session.get(url)
+        else:
+            r = requests.get(url)
+        r.raise_for_status
```
gfyoung: What does this line do?

pandas/io/common.py
gfyoung: 1) `compression` should be mentioned in the docs.  Would be great to add that.
2) Given my first point, `session` should be as well.

And yes, template / Appender approach sounds like a good idea!

pandas/io/common.py
ocefpaf: NP. I'll leave it there. Just renamed the `urlopen` variants to try to make it a little bit clear of what is going on.

pandas/io/common.py
ocefpaf: I need some guidance on how to do this when crossing modules, `common.py` in `get_filepath_or_buffer` and `parsers.py` for all the parsers defined in `_parser_params`.

The options are:

| options               | `get_filepath_or_buffer`    | `_parser_params`  |
|-----------------------|-----------------------------|-------------------|
| `filepath_or_buffer`  |  present                    | NA                |
| `mode`                |  present                    | NA                |
| `encoding`            |  present                    | present           |
| `compression`         |  missing (added in this PR) | present           |
| `session`             |  missing                    | present           |

It is unclear to me how to fix this. Some questions I have:
- Should the `compression` and `encoding` docstring be standardized?
- If yes on the question above what is the best strategy for this? Should `encoding`, `compression`, and `session` be defined in `common.py` and imported in `parsers.py` for appending in `_parser_params`? Or the other way around?

My guess is that I should do get the common options from `parser.py` and add it to `common.py` like:

```python

_common_params = r"""
Parameters
----------
encoding : str, default None
    Encoding to use for UTF when reading/writing (ex. 'utf-8'). `List of Python
    standard encodings
    <https://docs.python.org/3/library/codecs.html#standard-encodings>`_
session : requests.Session
    object with the a requests session configuration for remote file.
    (requires the requests library)
compression : {'infer', 'gzip', 'bz2', 'zip', 'xz', None}, default 'infer'
    For on-the-fly decompression of on-disk data. If 'infer' and
    `filepath_or_buffer` is path-like, then detect compression from the
    following extensions: '.gz', '.bz2', '.zip', or '.xz' (otherwise no
    decompression). If using 'zip', the ZIP file must contain only one data
    file to be read in. Set to None for no decompression.

    .. versionadded:: 0.18.1 support for 'zip' and 'xz' compression.
"""
```

Then compose the `get_filepath_or_buffer` with the missing `filepath_or_buffer` and `mode`, and later `import` `_common_params` in `parsers.py` to compose `_parser_params`. Does that sound OK?

PS: if that is correct I'd prefer if a pandas doc expert did this instead of me, I'm kind of lost in the `docstrings` and `Appender`s :smile:

pandas/io/common.py
ocefpaf: It will raise an exception if there is any issue reading the URL (any non-200 OK status).

gfyoung: But don't you need to actually call the function?  i.e. `r.raise_for_status()` (with parentheses)

ocefpaf: Yep. I totally forgot that when copying-and-pasting :smile:

rsignell-usgs:
Tested this on a username/password protected CSV endpoint and it's working!
```python
import requests
session = requests.sessions.Session()
session.auth = ('Nope', 'NoWay!')
url = 'https://cgoms.coas.oregonstate.edu/erddap/tabledap/CE01ISSM-BUOY-001-GPS.csv'

import pandas as pd
df = pd.read_csv(url, session=session)
```

pandas/io/excel.py
```diff
@@ -304,7 +304,8 @@ def read_excel(io,
                **kwds):
 
     if not isinstance(io, ExcelFile):
-        io = ExcelFile(io, engine=engine)
+        session = kwds.get('session', None)
```
jreback: just list session as a kwarg in read_excel (and in ExcelFile), then just pass it in

pandas/io/json/json.py
```diff
@@ -406,6 +406,7 @@ def read_json(path_or_buf=None, orient=None, typ='frame', dtype=True,
     compression = _infer_compression(path_or_buf, compression)
     filepath_or_buffer, _, compression, should_close = get_filepath_or_buffer(
         path_or_buf, encoding=encoding, compression=compression,
+        session=session,
```
jreback: add a versionadded tag when you add session

pandas/io/parsers.py
```diff
@@ -297,6 +297,9 @@
     If a filepath is provided for `filepath_or_buffer`, map the file object
     directly onto memory and access the data directly from there. Using this
     option can improve performance because there is no longer any I/O overhead.
+session : requests.Session
+    object with the a requests session configuration for remote file.
+    (requires the requests library)
```
jreback: add a versionadded tag

jreback (changes requested):

WillAyd:
@ocefpaf can you address latest comments and merge in changes from master?

WillAyd:
Can you merge master?

WillAyd:
You are better off doing the following on your branch locally:

```bash
git fetch upstream
git merge upstream/master
```

That retains the history of comments on GitHub so makes it easier to track.

Looks like you have some test failures with the latest push - if you could take a look would be great

ocefpaf:
> You are better off doing the following on your branch locally:
> 
> ```shell
> git fetch upstream
> git merge upstream/master
> ```

That is what I did :-/

> Looks like you have some test failures with the latest push - if you could take a look would be great

I'll take a look ASAP.

WillAyd:
> That is what I did :-/

Hmm OK. Saw the force push here so that's why I thought you had done something else. Shouldn't need the `-f` flag when you do it this way

datapythonista:
@ocefpaf CI is red, can you fix any problem, and push again so there are no tests failures.

ocefpaf:
> @ocefpaf CI is red, can you fix any problem, and push again so there are no tests failures.

I believe that when I first submitted this PR the [`with-statement` wrapper](https://github.com/pandas-dev/pandas/pull/21504/files#r195869676) was not used or untested, that is why it passed. I tried to fix that but my Python knowledge and time is limited to do it before the end of the year. If anyone has an idea please let me know. (IMHO I would remove the wrapper b/c it will be removed anyway when pandas drop py27.)

datapythonista:
Not sure what the problem is. I'll close this for now, as we can't merge with the conflicts and the CI in red, but if you find a solution for what you said, feel free to reopen.

ocefpaf:
> Not sure what the problem is. I'll close this for now, as we can't merge with the conflicts and the CI in red, but if you find a solution for what you said, feel free to reopen.

Would you be interested in a Python 3 only version of this PR? That is working and I guess that all the Python 2.7 code will be removed soon anyway.

WillAyd:
@ocefpaf we drop Py2 support for 0.25 so sure I think Python 3 only is fine for that milestone onwards

PR #21265, opened 2018-05-30
TST: Add failing tests for minute rounding

Safrone:
For some minute frequencies, `.ceil()` will round up even if it should not.

- [ ] closes #21262
- [x] tests added
- [ ] tests passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] whatsnew entry

PR #21507, opened 2018-06-15
Fix Timestamp rounding

### LINKED ISSUE #21262: Timestamp ceil rounds up when it should not when using the '15min' frequency.
#### Code Sample

```python
>>> import pandas as pd
>>> pd.Timestamp('2018-01-12 18:15:00').ceil('15T')
Timestamp('2018-01-12 18:30:00')

>>> pd.Timestamp('2018-01-12 00:30:00').ceil('15T')
Timestamp('2018-01-12 00:45:00')

```
#### Problem description

When the frequency is 15min `.ceil()` rounds up when the timestamp is divisible by the frequency.

From some experimentation, this also occurs with some other minute frequencies, notably: `2`, `4`, `15`, `20` 

Minute frequencies that are factors of 60 should have consistent behavior in my opinion (not sure about the behavior of non-factors but those would be rare edge cases). 

#### Expected Output

```python
>>> pd.Timestamp('2018-01-12 18:15:00').ceil('15T')
Timestamp('2018-01-12 18:15:00')

>>> pd.Timestamp('2018-01-12 00:30:00').ceil('15T')
Timestamp('2018-01-12 00:30:00')
```

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.5.1.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.0-124-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.23.0
pytest: 2.8.5
pip: 9.0.1
setuptools: 39.2.0
Cython: 0.23.4
numpy: 1.14.3
scipy: 0.17.1
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: 1.3.1
patsy: 0.4.1
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.0.0
tables: 3.2.2
numexpr: 2.5.2
feather: None
matplotlib: 1.5.1
openpyxl: 2.3.2
xlrd: 0.9.4
xlwt: 1.0.0
xlsxwriter: 0.8.4
lxml: 3.6.0
bs4: 4.4.1
html5lib: None
sqlalchemy: 1.0.12
pymysql: None
psycopg2: 2.6.2 (dt dec pq3 ext lo64)
jinja2: 2.8
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

alimcmaster1:
- Closes #21262
- Tests added (thanks  @Safrone [PR](https://github.com/pandas-dev/pandas/pull/21265))

This change-set is to avoid rounding a timestamp when the timestamp is a multiple of the frequency string passed in.

"Values" param passed into round_ns can either be a np array or int. So relevant handling added for both.

FYI I havn't used Cython much before so keen to get peoples thoughts/feedback.

Thanks


mroeschke:
A few thoughts:

1) With your solution, I think it's better to make `round_ns` only take arrays (therefore wrap scalar inputs as array) to avoid checking the type of the input (and moreover `cdef`ing them) and iterating over values. We do something similar with `tz_localize_to_utc`.

2) Include tests for `DatetimeIndex.round`

3) Looks like there are some linting errors (your spacing seems to be 2 instead of 4 in some places).

alimcmaster1:
Thanks @mroeschke for the comments. I've done (2) and (3) added a few additional test cases to the DateTimeIndex ceil/floor and one that clearly shows how round should behave for this bug.

1) Yes this definitely makes sense. Just to clarify are you thinking in 
`_round` we check the type of `value` if its an int we wrap in an np array. Hence we can cdef `round_ns`?

Best,

Alistair

mroeschke:
Rather have `round_ns` (which we can at best `cpdef` since it's imported by a python file) _only_ accept numpy arrays. So since this method is used by timestamps, we would have to wrap it's `self.value` input as a numpy array. The relevant rounding code for `Timestamp`s would turn into:

```
def _round(self, freq, rounder):
    if self.tz is not None:
        value = self.tz_localize(None).value
    else:
        value = self.value
    value = np.array([value], dtype=np.int64)
    r = round_ns(value, rounder, freq)
```

pandas/tests/scalar/timestamp/test_unary_ops.py
```diff
@@ -118,6 +118,27 @@ def test_ceil_floor_edge(self, test_input, rounder, freq, expected):
...
+        ('2018-01-01 00:20:00', '20T', '2018-01-01 00:20:00'),
+        ('2018-01-01 03:00:00', '3H', '2018-01-01 03:00:00'),
+    ])
+    def test_round_minute_freq(self, test_input, freq, expected):
+        # Ensure timestamps that shouldnt round dont!
+        # GH#21262
+        dt = Timestamp(test_input)
+        expected = Timestamp(expected)
+
+        result_ceil = dt.ceil(freq)
```
mroeschke: Could you also parametrize over the rounding methods (`ceil`, `floor`, and `round`)? It would help reduce this duplication

alimcmaster1: Sure makes sense to me

jreback (changes requested):

Safrone:
So it shows up on the issue properly: #21262

alimcmaster1:
Cleaned up my implementation here
1) Params added in test cases
2) round_ns now only takes np.array ( hence cleaned up the logic in here )

Thoguhts @mroeschke ?

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -59,43 +59,54 @@ cdef inline object create_timestamp_from_ts(int64_t value,
...
 
-        r = (unit * rounder((values * (divisor / float(unit))) / divisor)
-             .astype('i8'))
+            r = (unit * rounder((value * (divisor / float(unit))) / divisor)
+                 .astype('i8'))
 
-    return r
+        return r
+
+    return np.fromiter((_round_non_int_multiple(item) for item in values), np.int64)
```
jreback: why are you iterating? this doesn’t make any sense to do so with. vectorizes function

jreback (changes requested):

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -59,43 +59,54 @@ cdef inline object create_timestamp_from_ts(int64_t value,
...
 def round_ns(values, rounder, freq):
+
     """
     Applies rounding function at given frequency
 
     Parameters
     ----------
-    values : int, :obj:`ndarray`
-    rounder : function
+    values : np.array
```
mroeschke: I'd leave this as :obj:`ndarray`

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -59,43 +59,54 @@ cdef inline object create_timestamp_from_ts(int64_t value,
...
+    rounder : function, eg. 'Ceil', 'Floor', 'round'
     freq : str, obj
 
     Returns
     -------
-    int or :obj:`ndarray`
+    np.array
```
mroeschke: Same

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -649,7 +660,10 @@ class Timestamp(_Timestamp):
         else:
             value = self.value
 
-        r = round_ns(value, rounder, freq)
+        value = np.array([value], dtype=np.int64)
+
+        # Will only ever contain 1 element for timestamp
+        r = round_ns(value, rounder, freq).item()
```
mroeschke: Nit. I think just indexing into this array (i.e. `round_ns(value, rounder, freq)[0]`) is just fine. Looks like `item` returns a copy of a Python scalar (and we may want to keep this a numpy scalar just in case)

mroeschke:
Thanks for the revision. As @jreback mentions, I dont think it's necessary to iterate, You should be able to perform the adjustments with vectorization. At a high level the logic should look like:

```
mask = value % unit == 0
if mask.all():
    return value
values[~mask] = _round_non_int_multiple(values[~mask])
return values
```

alimcmaster1:
That @mroeschke your logic above seems much neater let me refactor!

pandas/_libs/tslibs/timestamps.pyx
mroeschke: `self.value` from `Timestamp` will always be a scalar, so we implicitly know the result of this is be a one element array.

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -59,43 +59,58 @@ cdef inline object create_timestamp_from_ts(int64_t value,
...
+
     """
     Applies rounding function at given frequency
 
     Parameters
     ----------
-    values : int, :obj:`ndarray`
-    rounder : function
+    values : :obj:`ndarray`
+    rounder : function, eg. 'Ceil', 'Floor', 'round'
```
mroeschke: Nit: Could you lowercase ceil and floor

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -59,43 +59,58 @@ cdef inline object create_timestamp_from_ts(int64_t value,
...
-        # GH19206
-        # to deal with round-off when unit is large
-        if unit >= 1e9:
-            divisor = 10 ** int(np.log10(unit / 1e7))
-        else:
-            divisor = 10
 
-        r = (unit * rounder((values * (divisor / float(unit))) / divisor)
-             .astype('i8'))
+    values = np.copy(values)
```
mroeschke: Why do we need a copy here?

alimcmaster1: This is to handle the case where 'NaT' exists in the DateTimeIndex. Removing it will cause `test_ceil_floor_edge` in `test_scalar_compact.py` to fail.

I found that datetimelike.py `self.hasnans` in `_maybe_mask_results`will return False if we don't do the copy, by copying we ensure that we arn't referencing the base of the input array. Think that is the issue here, what do you think?

mroeschke: cc @jreback.

Thanks for the investigation. That sounds reasonable; but I am not too familiar with `nan`/`Nat` ops with respect to references.

jreback: you are mutating in place, so you DO need to copy.

that's fine. though pls use ``values.copy()``

jreback (changes requested):
i haven't fully reviewd yet

alimcmaster1:
Thanks @jreback, ive updated, let me know if there is anything else!

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -59,43 +59,58 @@ cdef inline object create_timestamp_from_ts(int64_t value,
...
-            divisor = 10 ** int(np.log10(unit / 1e7))
-        else:
-            divisor = 10
 
-        r = (unit * rounder((values * (divisor / float(unit))) / divisor)
-             .astype('i8'))
+    values = values.copy()
 
-    return r
+    # GH21262 If the Timestamp is multiple of the freq str
```
jreback: i think it simpler to put this condition inside ththe above function

then you don’t even need a function !

iow i think you just needed to add that case above  (in the original code)

jreback (changes requested):

pandas/_libs/tslibs/timestamps.pyx
alimcmaster1: Right yep I see what you are saying, just to confirm before I commit, something like [this](https://gist.github.com/alimcmaster1/824ea777641c5c54f58e4f279a334210) ?

jreback: yep!

alimcmaster1:
Done! @jreback

mroeschke:
Travis flagged some linting errors:

```
pandas/_libs/tslibs/timestamps.pyx:92:14: E128 continuation line under-indented for visual indent
pandas/_libs/tslibs/timestamps.pyx:92:80: E501 line too long (80 > 79 characters)
pandas/_libs/tslibs/timestamps.pyx:103:80: E501 line too long (87 > 79 characters)
pandas/_libs/tslibs/timestamps.pyx:104:14: E128 continuation line under-indented for visual indent
```

jreback (changes requested):
ok lgtm. can you add a whatsnew note in Fixed Regression for 0.23.2, ping on green.

jreback (approved):

alimcmaster1:
thanks @jreback and @mroeschke for helping review!

PR #21508, opened 2018-06-16
PERF: add method Categorical.__contains__

topper-123:
- [x] closes #21022 
- [x] xref #21369
- [x] tests added / passed
- [x] benchmark added / passed
- [x] whatsnew entry

Currently, membership checks in ``Categorical`` is very slow as explained by @fjetter in #21022. This PR fixes the issue. See also #21369 which fixed the similar issue for ``CategoricalIndex``.

Tests didn't exist beforehand and have been added.

ASV:

```
     [9e982e18]       [28461f0c]
-      4.26±0.7ms         134±20μs     0.03  categoricals.Contains.time_categorical_contains

```

### DIFF
--- a/doc/source/whatsnew/v0.23.2.txt
+++ b/doc/source/whatsnew/v0.23.2.txt
@@ -26,7 +26,7 @@ Performance Improvements
 
 - Improved performance of membership checks in :class:`CategoricalIndex`
   (i.e. ``x in ci``-style checks are much faster). :meth:`CategoricalIndex.contains`
-  is likewise much faster (:issue:`21369`)
+  is likewise much faster (:issue:`21369`, :issue:`21508`)
 - Improved performance of :meth:`MultiIndex.is_unique` (:issue:`21522`)
 -
 
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -1846,6 +1846,43 @@ class Categorical(ExtensionArray, PandasObject):
         """Returns an Iterator over the values of this Categorical."""
         return iter(self.get_values().tolist())
 
+    @staticmethod
+    def _contains(key, categories, container):
+        """Returns True if `key` is in `categories` and the
+        location of `key` in `categories` is in `container`.
+
+        This is a helper method used in :method:`Categorical.__contains__`
+        and in :class:`CategoricalIndex.__contains__`.
+        """
+
+        # is key in categories? Then get its location in categories.
+        # If not (i.e. KeyError), its location logically can't be in
+        # container either.
+        try:
+            loc = categories.get_loc(key)
+        except KeyError:
+            return False
+
+        # loc is the location of key in categories, but also the value
+        # for key in container. So, key may be in categories,
+        # but still not in container, this must be checked. Example:
+        # 'b' in Categorical(['a'], categories=['a', 'b']) #  False
+        if is_scalar(loc):
+            return loc in container
+        else:
+            # if categories is an IntervalIndex, loc is an array.
+            # Check if any scalar of the array is in the container
+            return any(loc_ in container for loc_ in loc)
+
+    def __contains__(self, key):
+        """Returns True if `key` is in this Categorical."""
+        hash(key)
+
+        if isna(key):  # if key is a NaN, check if any NaN is in self.
+            return self.isna().any()
+
+        return self._contains(key, self.categories, container=self._codes)
+
     def _tidy_repr(self, max_vals=10, footer=True):
         """ a short repr displaying only max_vals and an optional (but default
         footer)
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -157,6 +157,55 @@ def _maybe_to_categorical(array):
     return array
 
 
+def contains(cat, key, container):
+    """
+    Helper for membership check for ``key`` in ``cat``.
+
+    This is a helper method for :method:`__contains__`
+    and :class:`CategoricalIndex.__contains__`.
+
+    Returns True if ``key`` is in ``cat.categories`` and the
+    location of ``key`` in ``categories`` is in ``container``.
+
+    Parameters
+    ----------
+    cat : :class:`Categorical`or :class:`categoricalIndex`
+    key : a hashable object
+        The key to check membership for.
+    container : Container (e.g. list-like or mapping)
+        The container to check for membership in.
+
+    Returns
+    -------
+    is_in : bool
+        True if ``key`` is in ``self.categories`` and location of
+        ``key`` in ``categories`` is in ``container``, else False.
+
+    Notes
+    -----
+    This method does not check for Nan values. Do that separately
+    before calling this method.
+    """
+    # get location of key in categories.
+    # If a KeyError, the key isn't in categories, so logically
+    #  can't be in container either.
+    try:
+        loc = cat.categories.get_loc(key)
+    except KeyError:
+        return False
+
+    # loc is the location of key in categories, but also the *value*
+    # for key in container. So, `key` may be in categories,
+    # but still not in `container`. Example ('b' in categories,
+    # but not in values):
+    # 'b' in Categorical(['a'], categories=['a', 'b'])  # False
+    if is_scalar(loc):
+        return loc in container
+    else:
+        # if categories is an IntervalIndex, loc is an array.
+        return any(loc_ in container for loc_ in loc)
+
+
 _codes_doc = """The category codes of this categorical.
 
 Level codes are an array if integer which are the positions of the real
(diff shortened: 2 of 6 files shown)


pandas/core/arrays/categorical.py
```diff
@@ -1847,6 +1847,31 @@ def __iter__(self):
         """Returns an Iterator over the values of this Categorical."""
         return iter(self.get_values().tolist())
 
+    def __contains__(self, key):
+        """Returns True if `key` is in this Categorical."""
+        hash(key)
```
jreback: isn’t this very similar to what u did in CI ?
can we unify

jreback (changes requested):

pandas/core/arrays/categorical.py
topper-123: I've made a unified version.

pandas/tests/categorical/test_algos.py
```diff
@@ -71,6 +71,23 @@ def test_isin_empty(empty):
     tm.assert_numpy_array_equal(expected, result)
 
 
+def test_contains():
+
+    c = pd.Categorical(list('aabbca'), categories=list('cab'))
```
gfyoung: I know this is not a bug, but can we reference the original issue anyway?

pandas/tests/categorical/test_algos.py
```diff
@@ -71,6 +71,23 @@ def test_isin_empty(empty):
...
+    assert 'b' in c
+    assert 'z' not in c
+    assert np.nan not in c
+
+    # assert codes NOT in index
+    assert 0 not in c
+    assert 1 not in c
+
+    c = pd.Categorical(list('aabbca') + [np.nan], categories=list('cab'))
+
```
gfyoung: Nit: unnecessary newline IMO

pandas/core/arrays/categorical.py
```diff
@@ -1847,6 +1847,43 @@ def __iter__(self):
...
         return iter(self.get_values().tolist())
 
+    @staticmethod
+    def _contains(key, categories, container):
+        """Returns True if `key` is in `categories` and the
+        location of `key` in `categories` is in `container`.
+
+        This is a helper method used in :method:`Categorical.__contains__`
+        and in :class:`CategoricalIndex.__contains__`.
+        """
```
gfyoung: As this is not a one-liner doc-string:

* The docstring should start underneath the triple parentheses.
* The first line is a summary should only be one line
* List of parameters and their dtypes
* Return description / type

gfyoung:
@topper-123 : A bunch of nit-picking things, but overall this looks pretty good!

topper-123:
Comments addressed.

pandas/core/arrays/categorical.py
```diff
@@ -1847,6 +1847,66 @@ def __iter__(self):
...
+        This helper method is used in :method:`Categorical.__contains__`
+        and in :class:`CategoricalIndex.__contains__`.
+
+        Returns True if ``key`` is in ``categories`` and the
+        location of ``key`` in ``categories`` is in ``container``.
+
+        Notes
+        -----
+        This method does not check for Nan values. Do that separately
+        before calling this method.
```
gfyoung: Nice!  Couple of things:

* Notes section goes at the bottom
* The first line ("Helper for membership check...") should be underneath the triple quotes.

topper-123: Ok, both done.

But I've never noticed "the first line should be underneath the triple quotes"-rule and it's not widely applied in pandas. Are you sure that's a style rule?

gfyoung (dismissed):
LGTM!

cc @jreback

topper-123:
Ok, won't hold you up to that approval, fixed :-)

gfyoung (approved):
Appreciate it 😄 .  LGTM!

cc @jreback

doc/source/whatsnew/v0.23.2.txt
```diff
@@ -27,6 +27,8 @@ Performance Improvements
 - Improved performance of membership checks in :class:`CategoricalIndex`
   (i.e. ``x in ci``-style checks are much faster). :meth:`CategoricalIndex.contains`
   is likewise much faster (:issue:`21369`)
+- Improved performance of membership checks in :class:`Categorical`
```
jreback: can you make this refernce this PR number (or just coming the what's new into 1 line either way)

pandas/core/indexes/category.py
```diff
@@ -328,23 +327,8 @@ def __contains__(self, key):
         if isna(key):  # if key is a NaN, check if any NaN is in self.
             return self.hasnans
 
-        # is key in self.categories? Then get its location.
```
jreback: why can't you just do
```return key in self._data``?

File: pandas/core/arrays/categorical.py [hunk 8]
```diff
@@ -1847,6 +1847,67 @@ def __iter__(self):
         """Returns an Iterator over the values of this Categorical."""
         return iter(self.get_values().tolist())
 
+    @staticmethod
+    def _contains(key, categories, container):
```
jreback: this seems pretty convoluted, see my comment below

jreback (changes requested):

pandas/core/indexes/category.py
topper-123: Ok, actually I'v'e got an good idea for this, will reimplement.

pandas/core/arrays/categorical.py
jreback: so it looks very similiar. why cannot you just use ``key in self._data`` for the CI

jreback (changes requested):

topper-123:
@jreback , using ``._data`` would be slower, as it doesn't use the hashed value in ``._engine``:

```python
>>> n = 100_000
%timeit 'b' in ci._engine
2.6 µs ± 28.8 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
%timeit 'b' in ci._data
183 µs ± 9.4 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

So there neeeds to be different implementations.

pandas/core/arrays/categorical.py
```diff
@@ -158,6 +158,55 @@ def _maybe_to_categorical(array):
...
+
+    Returns
+    -------
+    is_in : bool
+        True if ``key`` is in ``self.categories`` and location of
+        ``key`` in ``categories`` is in ``container``, else False.
+
+    Notes
+    -----
+    This method does not check for Nan values. Do that separately
```
jreback: Nan -> NaN.

pandas/core/arrays/categorical.py
```diff
@@ -1847,6 +1896,15 @@ def __iter__(self):
         """Returns an Iterator over the values of this Categorical."""
         return iter(self.get_values().tolist())
 
+    def __contains__(self, key):
+        """Returns True if `key` is in this Categorical."""
+        hash(key)
```
jreback: I think your hash check needs to go into contains itself (null check is ok where it is).

can you add a test on something that isn't hashable as well

pandas/core/arrays/categorical.py
```diff
@@ -1847,6 +1896,15 @@ def __iter__(self):
...
+
+        if isna(key):  # if key is a NaN, check if any NaN is in self.
```
jreback: can you put comments on the line above

pandas/tests/categorical/test_algos.py
```diff
@@ -71,6 +71,22 @@ def test_isin_empty(empty):
     tm.assert_numpy_array_equal(expected, result)
 
 
+def test_contains():
```
jreback: move to test_indexing, should be other checks already there, de-duplicate with those.

jreback (changes requested):
this ok. some comments. ping when green.

pandas/tests/categorical/test_algos.py
topper-123: Lookef at test_indexing, didn't find anything very similar. I moved it to test_operators, which is more similar, IMO. Is that ok?

jorisvandenbossche (commented):
What is the reason you changed the private  ``._contains`` to public  ``contains``? (it's not needed for the implementation?) 
It's just that I am not sure it is worth adding a new public method.

topper-123:
@jorisvandenbossche, to my understanding these modules are themselves private and it’s not necessary for functions in private modules to be marked as private - that would be superfluous.

When it was a method on Categorical, it *was* needed to mark it private, as Categorical is a public class.

topper-123:
The Travis error is an HTTP error:

```
CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://repo.anaconda.com/pkgs/main/linux-64/qt-5.9.5-h7e424d6_0.tar.bz2>
Elapsed: -
An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.
real	3m14.046s
user	1m7.620s
sys	0m3.084s
The command "ci/install_travis.sh" failed and exited with 1 during .
Your build has been stopped.
```

so the Travis failure has nothing to do with this PR.

jorisvandenbossche:
> @jorisvandenbossche, to my understanding these modules are themselves private and it’s not necessary for functions in private modules to be marked as private - that would be superfluous.

@topper-123 whoops, sorry, only looked very briefly at the incremental diff and missed that it was a function and not a method.

jreback (approved):

jreback:
thanks @topper-123

PR #21511, opened 2018-06-16
DOC: Improve code example for Index.get_indexer

topper-123:
Make code example clearer for ``Index.get_indexer``

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -3148,8 +3148,12 @@ class Index(IndexOpsMixin, PandasObject):
 
         Examples
         --------
-        >>> indexer = index.get_indexer(new_index)
-        >>> new_values = cur_values.take(indexer)
+        >>> index = Index(['c', 'a', 'b'])
+        >>> index.get_indexer(['a', 'b', 'x'])
+        array([ 1,  2, -1], dtype=int32)
+
+        Notice that the return value is an array of locations in ``index``
+        and ``x`` is marked by -1, as it is not in ``index``.
 
         Returns
         -------
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -3146,21 +3146,22 @@ class Index(IndexOpsMixin, PandasObject):
 
             .. versionadded:: 0.21.0 (list-like tolerance)
 
-        Examples
-        --------
-        >>> index = Index(['c', 'a', 'b'])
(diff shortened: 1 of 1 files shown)


gfyoung (approved):
Nice!

cc @jreback

pandas/core/indexes/base.py
```diff
@@ -3148,8 +3148,12 @@ def droplevel(self, level=0):
 
         Examples
         --------
-        >>> indexer = index.get_indexer(new_index)
-        >>> new_values = cur_values.take(indexer)
+        >>> index = %(target_klass)s(['c', 'a', 'b'])
```
jschendel: There are docs being generated for both [`MultiIndex.get_indexer`](https://pandas-docs.github.io/pandas-docs-travis/generated/pandas.MultiIndex.get_indexer.html) and [`IntervalIndex.get_indexer`](https://pandas-docs.github.io/pandas-docs-travis/generated/pandas.IntervalIndex.get_indexer.html) that are using this template:

https://github.com/pandas-dev/pandas/blob/9e982e18678c47c83ce2bb5932f844e2d9ee21f3/pandas/core/indexes/multi.py#L1962-L1963

https://github.com/pandas-dev/pandas/blob/9e982e18678c47c83ce2bb5932f844e2d9ee21f3/pandas/core/indexes/interval.py#L1090-L1091

Other indexes, like `DatetimeIndex`, don't appear to currently have docs generated for this, but inherit the method and would produce a docstring viewable with a command like `DatetimeIndex.get_indexer?` in IPython.

In any of these cases, the `%(target_klass)s` (I think this should actually be `%(klass)s`?) would get filled in with one of the aforementioned index types, which would produce an invalid example.

I think it's reasonable to just hardcode this in as `Index(['c', 'a', 'b'])` and have the user infer the behavior for the specific type of index they're using?

jschendel (changes requested):

pandas/core/indexes/base.py
jorisvandenbossche: > I think it's reasonable to just hardcode this in as Index(['c', 'a', 'b']) and have the user infer the behavior for the specific type of index they're using?

+ 1

jschendel (approved):
Thanks!

jorisvandenbossche (approved):

jorisvandenbossche:
@topper-123 Thanks!

PR #21531, opened 2018-06-18
BUG: Allow IOErrors when attempting to retrieve default client encoding.

### LINKED ISSUE #21552: BUG: mod_wsgi restricting access to stdout throws unhandled IOError.
#### Problem description
We're using pandas on the backend of our internal web server for a few utilities. Any attempt to import the library causes an IOError to be thrown due to our setup of mod_wsgi restricting stdout access by default (WSGIRestrictStdout On).

The cause of this error is an attempt to get the client encoding on init when registering options. A call is made to detect_client_encoding() in console.py under pandas/io/formats/ which is throwing the IOError. The call to sys.stdout.encoding in this function is contained inside a try/except block, however, the except statement is limited to an AttributeError.

I propose we expand this except statement to allow for both AttributeErrors and IOErrors.

#### Details
<details>
INSTALLED VERSIONS

------------------

commit: e2d5958bb111cb9a87b307a963f57ea8cc775646
python: 2.7.15.final.0
python-bits: 64
OS: Linux
OS-release: 4.16.14-300.fc28.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: None.None

pandas: 0.24.0.dev0+118.ge2d5958bb
pytest: 3.6.1
pip: 9.0.3
setuptools: 39.2.0
Cython: 0.28.3
numpy: 1.14.5
scipy: None
pyarrow: None
xarray: None
IPython: None
sphinx: None
patsy: None
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: None
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

JayOfferdahl:
- [X] closes #21552 

When using mod_wsgi, access to sys.stdout is restricted by default. To handle this case, catch a base exception instead of a more specific AttributeError during this process to include the thrown IOError by mod_wsgi.


WillAyd:
- Is this referencing a particular issue?
- Is there any other option besides catching base Exception (tend to avoid this if we can)
- Can you add tests?

JayOfferdahl:
* Is this referencing a particular issue?
    - I haven't raised an issue for this as it just came up this morning.
* Is there any other option besides catching base Exception (tend to avoid this if we can)
    - We could instead widen the exception clause to just AttributeError and IOError.
* Can you add tests?
    - Will do.

WillAyd:
OK thanks - if you could open an issue and reference it then that would be preferable (typically PRs are only allowed without an issue if they are DOC updates).

You'll also want a whatsnew note to describe the bug and reference the issue. Next minor release is 0.23.2 at this point

pandas/io/formats/console.py
```diff
@@ -21,7 +21,7 @@ def detect_console_encoding():
     encoding = None
     try:
         encoding = sys.stdout.encoding or sys.stdin.encoding
-    except AttributeError:
+    except Exception:
```
gfyoung: Why can't we just `except` both on `AttributeError` and `IOError` ?  I generally try to avoid using the broad `Exception` unless we have good reason to.

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,34 @@
...
+    raise error
+
+
+def test_detect_console_encoding_stdout(monkeypatch):
+    monkeypatch.setattr('sys.stdin.encoding', '')
+    monkeypatch.setattr('sys.stdout.encoding', 'foo')
+    assert detect_console_encoding() == 'foo'
+
+
+def test_detect_console_encoding_stdin(monkeypatch):
```
jreback: can you add some comments on what is being tested here.

doc/source/whatsnew/v0.23.2.txt
```diff
@@ -65,7 +65,7 @@ Bug Fixes
 **I/O**
 
 - Bug in :func:`read_csv` that caused it to incorrectly raise an error when ``nrows=0``, ``low_memory=True``, and ``index_col`` was not ``None`` (:issue:`21141`)
--
+- Bug in :meth:`detect_client_encoding` where potential IOError goes unhandled when importing in a mod_wsgi process due to restricted access to stdout. (:issue:`21552`)
```
jreback: double backticks on IOError

jreback (changes requested):
looks ok, getting to pass travis might be a bit tricky.

JayOfferdahl:
Thanks for the feedback, made some changes to tests to cover more cases and provide some insight as to what they're doing. Is there anything special I need to do to get this working with travis?

jreback:
looks like it might be tricky to monkeypatch how you are doing it: https://travis-ci.org/pandas-dev/pandas/jobs/394566147

does this test run locally?

JayOfferdahl:
Interesting...it runs locally just fine...I'll probably need to patch stdout/stdin like the other tests and provide the property on that patched object.

JayOfferdahl:
I've updated the test to mock out stdout/stdin instead of trying to change their encoding attribute. I do the same in other tests so I'm assuming this will work fine with travis.

Apologies for the slow turn around on this one!

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,51 @@
...
+])
+def test_detect_console_encoding_from_stdout_stdin(monkeypatch, empty, filled):
+    # Ensures that when sys.stdout.encoding or sys.stdin.encoding is used when
+    # they have values filled.
+    class MockEncoding(object):
+        def __init__(self, encoding):
+            super(MockEncoding, self).__init__()
+            self.encoding = encoding
+
+    monkeypatch.setattr('sys.{}'.format(empty), MockEncoding(''))
```
WillAyd: Any chance you saw the context manager for monkeypatch? Seems to be preferred for patching standard library items (may require a version bump for pytest dependency).

JayOfferdahl:
@WillAyd I've taken a look at the context manager and have updated these tests. Looks like I missed the 0.23.2 release...my own fault, of course. Should I put my note in the 0.24 whatsnew document? Not sure how you all handle bugs in non-minor release versions.

Side note: I did have to upgrade pytest to 3.6.3, I'm almost certain that this will break travis since it won't have that version (maybe I'm wrong) -- where does this get specified?

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,54 @@
...
+def mock_raises_exception(error=Exception):
+    raise error
+
+
+@pytest.mark.parametrize('empty,filled', [
+    ['stdin', 'stdout'],
+    ['stdout', 'stdin']
+])
+def test_detect_console_encoding_from_stdout_stdin(monkeypatch, empty, filled):
+    # Ensures that when sys.stdout.encoding or sys.stdin.encoding is used when
```
WillAyd: Can add a reference to the issue for this and the other new tests, i.e. `# GH 21552`

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,54 @@
...
+        assert detect_console_encoding() == filled
+
+
+@pytest.mark.parametrize('stdEncoding', [
+    partial(mock_raises_exception, AttributeError),
+    partial(mock_raises_exception, IOError),
+    lambda: 'ascii'
+])
+def test_detect_console_encoding_fallback_to_locale(monkeypatch, stdEncoding):
+    monkeypatch.setattr('locale.getpreferredencoding', lambda: 'foo')
```
WillAyd: Can put this in the context manager, no?

WillAyd (changes requested):
@JayOfferdahl no worries on timing - we tentatively have a 0.23.3 bug fix release that you can move the whatsnew entry to. Looks like you also have some LINTing errors to address

@jreback do you have any opposition to increasing the min version of pytest to 3.6? It's a relatively new release but the context manager for monkey patching I think is a useful and clean feature

jreback:
no problem bumping pytest version (separate PR)

JayOfferdahl:
Made the linter changes...Have there been any consideration into putting that step in a pre-commit hook? Other than that, it appears that the 0.23.3 release already went out (July 7th) so I put the line in the 0.23.4 whatsnew file; let me know if this was an error.

doc/source/whatsnew/v0.23.2.txt
```diff
@@ -89,6 +89,8 @@ Bug Fixes
 
 - Bug in :func:`read_csv` that caused it to incorrectly raise an error when ``nrows=0``, ``low_memory=True``, and ``index_col`` was not ``None`` (:issue:`21141`)
 - Bug in :func:`json_normalize` when formatting the ``record_prefix`` with integer columns (:issue:`21536`)
+- Bug in :meth:`detect_client_encoding` where potential ``IOError`` goes unhandled when importing in a mod_wsgi process due to restricted access to stdout. (:issue:`21552`)
```
jreback: remove from here

doc/source/whatsnew/v0.23.4.txt
```diff
@@ -41,7 +41,7 @@ Bug Fixes
 
 **I/O**
 
--
+- Bug in :meth:`detect_client_encoding` where potential ``IOError`` goes unhandled when importing in a mod_wsgi process due to restricted access to stdout. (:issue:`21552`)
```
jreback: let’s put in 0.24 instead

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,57 @@
+from functools import partial
+import pytest
+
+from pandas.io.formats.console import detect_console_encoding
+
+
+def mock_raises_exception(error=Exception):
```
jreback: what is this for?

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,57 @@
...
+@pytest.mark.parametrize('stdEncoding', [
+    partial(mock_raises_exception, AttributeError),
+    partial(mock_raises_exception, IOError),
+    lambda: 'ascii'
+])
+def test_detect_console_encoding_fallback_to_locale(monkeypatch, stdEncoding):
+    # GH 21552
+    with monkeypatch.context() as context:
+        context.setattr('locale.getpreferredencoding', lambda: 'foo')
+        context.setattr('sys.stdout', stdEncoding)
```
jreback: just directly use

with pytest raises

jreback (changes requested):
the mock raises is
plain confusing here - pls simplify

JayOfferdahl:
Thanks for the feedback on the mock_raises_exception. Using pytest.raises is much simpler. I moved the whatsnew entry to 0.24 and took it out of both of the others.

JayOfferdahl:
Any update here?

jreback (commented):
so these tests fail w/o the fix?

JayOfferdahl:
Well this is embarrassing: No. When I updated the tests to use pytest.raises in the parameterized sections, I failed to include a case where IOError was raised. The pytest.raises ExecInfo object was then sent into the first try block and raised it's own AttributeError (which was caught) because it doesn't have an encoding attribute. 

The reason I originally had the 'mock_raises_exception' madness was because I want the input to raise an exception when accessed, so that it may be caught inside the function. As far as I can tell, pytest doesn't offer this functionality (using Mock, I would just have this as a side_effect on my mocked object).

Am I missing something with pytest.raises? I'm not trying to assert that an exception being raised was the end of the test, but rather that the function handled the exception and moved on.

WillAyd:
`pytest.raises` is typically used as a context manager so I don't think it serves the purpose for what you are trying to do. Does raising via a UDF not work?

JayOfferdahl:
I pushed an attempt at solving this problem. I ended up moving that MockEncoding class outside of the test class and extended it a little bit. Yes, I know this is a dirty solution -- let me know if you guys can think of any way to improve it.

WillAyd:
Hmm yea this is kind of a corner case...would be nice with `mock` but I don't think that's feasible to manage with our Py27 compat. Unless @jreback has any objection I think we can stick with what you have, though add a comment to the class that says `TODO(py27): replace with mock` so we can get rid of it when we drop support

JayOfferdahl:
Sounds good to me. I did get a pylint warning about the TODO -- is this going to fail in Travis?

WillAyd:
Looks like it. You can move the comment inline with the class name

JayOfferdahl:
Inline with the class name gave the same error...ended up putting this in the docstring. This should still be just as easy to discover, I assume.

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,71 @@
+import pytest
+
+from pandas.io.formats.console import detect_console_encoding
+
+
+class MockEncoding(object):
+    """
+    TODO(py27): replace with mock
```
WillAyd: I didn't see the LINT errors before but maybe you got it mixed up with the one I've commented on below? I still think inline with the `class` block opening would be preferable. Certainly a nit, but I haven't seen TODOs embedded in docstrings like this before

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,71 @@
...
+    side effect should be an exception that will be raised.
+    """
+    def __init__(self, encoding):
+        super(MockEncoding, self).__init__()
+        self.val = encoding
+
+    @property
+    def encoding(self):
+        return raiseOrReturn(self.val)
+
```
WillAyd: The current LINT error is because you are missing a blank line here

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,71 @@
...
+def raiseOrReturn(val):
```
WillAyd: Prefer PEP8 naming, so let's call this `raise_or_return`

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,71 @@
...
+    # Ensures that when sys.stdout.encoding or sys.stdin.encoding is used when
+    # they have values filled.
+    # GH 21552
+    with monkeypatch.context() as context:
+        context.setattr('sys.{}'.format(empty), MockEncoding(''))
+        context.setattr('sys.{}'.format(filled), MockEncoding(filled))
+        assert detect_console_encoding() == filled
+
+
+@pytest.mark.parametrize('stdEncoding', [
```
WillAyd: PEP8 so `std_encoding` or simply `encoding`

WillAyd (changes requested):

JayOfferdahl:
Changes made. I was using the regular pylint package and not flake8, which explains the difference. Apologies, and thanks for spending so much time on this, I really appreciate it.

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,71 @@
...
+    def __init__(self, encoding):
+        super(MockEncoding, self).__init__()
+        self.val = encoding
+
+    @property
+    def encoding(self):
+        return raise_or_return(self.val)
+
+
+def raise_or_return(val):
```
WillAyd: Can this be moved as an instance method of `MockEncoding`?

JayOfferdahl: I'm also using it down below to test when defaulting to system default -- it covers the same goal of throwing an exception like above here. I could make it a static method of MockEncoding, but I thought using that object for both cases wouldn't make sense since the encoding property is only being consumed from accesses by stdout or stdin. What do you think?

JayOfferdahl: @WillAyd did you see this?

WillAyd: Let's make this a class method just so all of the relevant constructs stay together

pandas/tests/io/formats/test_console.py
```diff
@@ -0,0 +1,71 @@
...
+    # GH 21552
+    with monkeypatch.context() as context:
+        context.setattr('locale.getpreferredencoding', lambda: 'foo')
+        context.setattr('sys.stdout', encoding)
+        assert detect_console_encoding() == 'foo'
+
+
+@pytest.mark.parametrize('std,locale', [
+    [MockEncoding('ascii'), lambda: 'ascii'],
+    [MockEncoding('ascii'), lambda: raise_or_return(Exception)],
```
WillAyd: Instead of mixing usage of `raise_or_return` here for each parameter just do it in the body of the function; should work for all parameters since in case of a string it returns the  string anyway

WillAyd (changes requested):
2 minor comments

WillAyd (approved):
lgtm so long as it goes green. @jreback

WillAyd:
@JayOfferdahl can you rebase?

JayOfferdahl:
Of course, is the general work flow to just squash all of my changes down into one commit and rebase on top of master or should I merge master into this?

WillAyd:
Merge master in

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -673,7 +673,7 @@ I/O
 
 - :func:`read_html()` no longer ignores all-whitespace ``<tr>`` within ``<thead>`` when considering the ``skiprows`` and ``header`` arguments. Previously, users had to decrease their ``header`` and ``skiprows`` values on such tables to work around the issue. (:issue:`21641`)
 - :func:`read_excel()` will correctly show the deprecation warning for previously deprecated ``sheetname`` (:issue:`17994`)
--
+- Bug in :meth:`detect_client_encoding` where potential ``IOError`` goes unhandled when importing in a mod_wsgi process due to restricted access to stdout. (:issue:`21552`)
```
WillAyd: Maybe move to 0.23.5? @jreback

gfyoung: Is this a regression?  If not, I would probably stay with `0.24.0` unless there is a compelling reason to backport-patch this bug.

WillAyd (approved):
lgtm @jreback

JayOfferdahl:
Any update here?

gfyoung (approved):
Sorry that this fell through the cracks!  This looks good to merge from me as well.

cc @jreback

jreback (changes requested):
do we have other console tests that can go in here?

jreback (approved):

jreback:
this is fine. can you rebase, and ping on green (@WillAyd  / @gfyoung ok to merge then)

WillAyd:
Thanks @JayOfferdahl !

JayOfferdahl:
Thank you, and thanks for the help throughout the process!

PR #21532, opened 2018-06-18
fixed unicode issue (21499) using requests library

### LINKED ISSUE #21499: Python Pandas read_html fails when reading tables from Wikipedia
I am trying to read the tables from a Wikipedia page using the following code:

```python
import pandas as pd
pd.read_html('https://en.wikipedia.org/wiki/2013–14_Premier_League')
```

Doing that generates the following error:

```python
UnicodeEncodeError: 'ascii' codec can't encode character '\u2013' in    position 14: ordinal not in range(128)
```

I have tried

```python
pd.read_html('https://en.wikipedia.org/wiki/2013–14_Premier_League', encoding='utf-8')
```

But still get the same error. The following works:

```python 
import requests
r = requests.get('https://en.wikipedia.org/wiki/2017–18_Premier_League')
c = r.content
dfs = pd.read_html(c)
```
What I want to know is how to get pd.read_html() to work directly on the url without requests. What is it that I don't understand about encoding or is this a problem with Pandas?

I am running an Anaconda distribution of Pandas 0.21.1 and Python 3.5.4. Thanks for any help.

hannah2048:
- [x ] closes #21499 
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/io/html.py
+++ b/pandas/io/html.py
@@ -127,8 +127,8 @@ def _read(obj):
     raw_text : str
     """
     if _is_url(obj):
-        with urlopen(obj) as url:
-            text = url.read()
+        import requests
+        text = requests.get(obj).content
     elif hasattr(obj, 'read'):
         text = obj.read()
     elif isinstance(obj, char_types):
--- a/pandas/io/html.py
+++ b/pandas/io/html.py
@@ -985,3 +985,4 @@ def read_html(io, match='.+', flavor=None, header=None, index_col=None,
                   decimal=decimal, converters=converters, na_values=na_values,
                   keep_default_na=keep_default_na,
                   displayed_only=displayed_only)
+  


hannah2048 (commented):
added missing newline

WillAyd:
Hmm well it looks like this is going to conflict with the work being done in #21504. @ocefpaf can you see if your PR accounts for this issue (adding a test case, accordingly)? If so would be easier just to bundle this in with your change

ocefpaf:
My guess is that https://github.com/pandas-dev/pandas/pull/21504 could supersede this one. I'll add tests soon. I have a limited time I can work on this. (And I'm waiting for the docstrings questions I asked there before continuing :smile:).

pandas/io/html.py
```diff
@@ -127,8 +127,8 @@ def _read(obj):
     raw_text : str
     """
     if _is_url(obj):
-        with urlopen(obj) as url:
```
ocefpaf: @WillAyd in my PR all `urlopen` calls are wrap with `requests` if it is available so I guess we cover this one in #21504

WillAyd:
Closing in favor of #21504

PR #21533, opened 2018-06-19
remove daytime attr, move getstate and setstate to base class

jbrockmendel:
If/when #18224 is revived to make `_BaseOffset` a `cdef class`, we'll need to move `__getstate__` and `__setstate__` into the base class (and then make some changes).  This moves the two methods pre-emptively so that we'll have a smaller diff for the next steps.

Also removes an attr `self.daytime` that is not used anywhere else.


jreback:
lgtm. ping on green.

jreback:
thanks!

PR #21543, opened 2018-06-19
Fixed HDFSTore.groups() performance.  

### LINKED ISSUE #21372: Improve HDFStore.groups performance
I ran into this issue (#17593) recently, when I was trying to figure out what was causing the order of magnitude difference between `h5ls` and `HDFStore.groups()` or `HDFStore.keys()` and was trying to figure out more information about the problem.  This was shrugged off as a Tables problem at the time and then closed.

I've come to the conclusion that this is actually Pandas problem.  Hopefully, I can convince you of that as well.

If I look at a large *flat* HDF5 file (one where I have only created pandas dataframes at the root node `/`), then:

```
%%timeit
groups_iter_nodes = set()
for x in store._handle.iter_nodes(store._handle.root):
    groups_iter_nodes |= {x._v_pathname} 
```
```
78.7 ms ± 1.54 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

where:

```
%%timeit
groups_walk_nodes = set()
for x in store._handle.walk_nodes(store._handle.root):
    groups_walk_nodes |= {x._v_pathname}
```
```
7.08 s ± 168 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

And the `iter_nodes` method has found every node that is relevant.  The bottom version is (an abbreviated) version of what the Pandas library does, while the top version is just looking at the first level of keys.

Obviously, just looking at the root level of the HDF file doesn't work for the vast majority of cases, but I think there is still *significant* room to improve.

The root cause is that there are a whole bunch of leaves of groups that don't need to be looked at because we already know the answer.  In my file, each of those groups has the following arrays under it: `['axis0', 'axis1', 'block0_items', 'block0_values']`.  These arrays are where the data lies for the pandas dataframe.  In the above example, `walk_nodes` must actually look at every one of these nodes, while the `iter_nodes` version doesn't.

If we check if there are any children *not* in this list:

```
%%timeit
groups_iter_nodes = set()
for x in store._handle.iter_nodes(store._handle.root):
    if len(set(x._v_children.keys()) - {'axis0', 'axis1', 'block0_items', 'block0_values'}) > 0:
        print("oops")
    groups_iter_nodes |= {x._v_pathname}
```
```
88.4 ms ± 427 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

This is still much faster than actually visiting the nodes in order to determine that they aren't pandas dataframes.

I'm not entirely sure how pandas deals with dataframes in hdf files.  But if pandas really doesn't ever store anything except as children of groups, then this is a better way to look for all the groups:

```
%%timeit
groups_walk_nodes = set()
for x in store._handle.walk_groups(store._handle.root):
    groups_walk_nodes |= {x._v_pathname}
```
```
149 ms ± 2.65 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Ultimately, I don't know enough about the history of pandas and HDFStores, or the details about how pandas stores dataframes in hdf files to fix this -- and I expect that there is a lot of historical backwards compatibility baggage that needs to be navigated -- but I think it is clear that this could be much smarter about how it finds groups.

If backwards compatability isn't an issue, I would be happy to submit a pull request for this.  I can only assume that `walk_groups` is the least that could be done to improve things.

spott:
No longer walks every node, but rather every group.

- [x] closes #21372
- [x] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.23.2.txt
+++ b/doc/source/whatsnew/v0.23.2.txt
@@ -27,7 +27,9 @@ Performance Improvements
 - Improved performance of membership checks in :class:`CategoricalIndex`
   (i.e. ``x in ci``-style checks are much faster). :meth:`CategoricalIndex.contains`
   is likewise much faster (:issue:`21369`)
--
+- Improved performance of :meth:`HDFStore.groups` (and dependent functions like
+  :meth:`~HDFStore.keys`.  (i.e. ``x in store`` checks are much faster)
+  (:issue:`21372`)
 
 Documentation Changes
 ~~~~~~~~~~~~~~~~~~~~~
--- a/pandas/io/pytables.py
+++ b/pandas/io/pytables.py
@@ -1098,7 +1098,7 @@ class HDFStore(StringMixin):
         _tables()
         self._check_if_open()
         return [
-            g for g in self._handle.walk_nodes()
+            g for g in self._handle.walk_groups()
             if (not isinstance(g, _table_mod.link.Link) and
                 (getattr(g._v_attrs, 'pandas_type', None) or
                  getattr(g, 'table', None) or


spott:
I'm not sure if this deserves it's own 'whatsnew' entry... it is pretty simple.

spott:
This is my first pull request for pandas.  "asv"?

I've updated the whatsnew.

jreback:
http://pandas.pydata.org/pandas-docs/stable/contributing.html#running-the-performance-test-suite

spott:
Thanks.

I ran the `io.hdf.*` benchmarks and got the following:
```
     [49188296]       [38b70fcf]
+     7.02±0.07ms         25.1±1ms     3.57  io.hdf.HDFStoreDataFrame.time_query_store_table
+      28.9±0.4ms           50.7ms     1.75  io.hdf.HDFStoreDataFrame.time_read_store_mixed
+      16.3±0.4ms           28.3ms     1.74  io.hdf.HDFStoreDataFrame.time_query_store_table_wide
+     25.5±0.01ms           38.0ms     1.49  io.hdf.HDFStoreDataFrame.time_read_store
+        68.1±2ms           98.1ms     1.44  io.hdf.HDFStoreDataFrame.time_read_store_table_mixed
+        41.8±1ms         57.5±7ms     1.37  io.hdf.HDFStoreDataFrame.time_write_store_mixed
-          22.5ms           19.5ms     0.86  io.hdf.HDFStoreDataFrame.time_read_store_table_wide
-          29.8ms         25.1±1ms     0.84  io.hdf.HDFStoreDataFrame.time_write_store
-          10.4μs       8.63±0.9μs     0.83  io.hdf.HDFStoreDataFrame.time_store_repr
-      6.04±0.1ms       4.61±0.2ms     0.76  io.hdf.HDFStoreDataFrame.time_store_info
```

Which is obviously an issue.

Because this doesn't make sense (some of those tests don't even appear to be touching the part of the code that I changed...), I ran it again:

```
     [49188296]       [38b70fcf]
+      28.2±0.3ms         41.0±1ms     1.45  io.hdf.HDFStoreDataFrame.time_write_store_mixed
+         200±3ms            284ms     1.42  io.hdf.HDFStoreDataFrame.time_write_store_table_dc
+        44.5±2ms         60.7±2ms     1.36  io.hdf.HDFStorePanel.time_read_store_table_panel
+     5.84±0.03μs       7.64±0.3μs     1.31  io.hdf.HDFStoreDataFrame.time_store_repr
+        49.3±3ms           58.0ms     1.18  io.hdf.HDFStoreDataFrame.time_write_store_table
+      70.6±0.4ms         80.8±2ms     1.15  io.hdf.HDFStoreDataFrame.time_write_store_table_mixed
-      28.4±0.2ms       24.4±0.2ms     0.86  io.hdf.HDFStoreDataFrame.time_read_store
-         159±2ms         136±10ms     0.85  io.hdf.HDF.time_read_hdf('table')
-         152±4ms          128±5ms     0.84  io.hdf.HDF.time_write_hdf('table')
-      16.6±0.9ms       13.1±0.2ms     0.79  io.hdf.HDFStoreDataFrame.time_query_store_table_wide
-      5.10±0.4ms      3.04±0.08ms     0.60  io.hdf.HDFStoreDataFrame.time_store_info
```

And I get what looks to be a completely different set of tests that are different.

A third time gives me:

```
     [49188296]       [38b70fcf]
+      27.6±0.1ms       36.1±0.7ms     1.31  io.hdf.HDFStoreDataFrame.time_read_store_mixed
+      19.1±0.3ms       22.5±0.7ms     1.18  io.hdf.HDFStoreDataFrame.time_read_store_table_wide
-      71.7±0.7ms       56.6±0.1ms     0.79  io.hdf.HDFStoreDataFrame.time_write_store_table_mixed
-     26.0±0.07ms       19.0±0.2ms     0.73  io.hdf.HDFStoreDataFrame.time_write_store
-        60.6±3ms         41.4±1ms     0.68  io.hdf.HDFStorePanel.time_read_store_table_panel
-        42.7±2ms       26.9±0.1ms     0.63  io.hdf.HDFStoreDataFrame.time_write_store_mixed
```

Which is at least a little better.  I think this test is fairly sensitive to what else is going on on my computer.  In order to get a more accurate result, I think I'll need to clean boot and run the tests without anything else running.

jreback:
yeah there don't appear to be any benchmarks for this (except for ``.info()``) which is related. can you see if it makes sense to add any benchmarks here (i don't remember exactly what .info hits)

doc/source/whatsnew/v0.23.2.txt
```diff
@@ -27,7 +27,9 @@ Performance Improvements
 - Improved performance of membership checks in :class:`CategoricalIndex`
   (i.e. ``x in ci``-style checks are much faster). :meth:`CategoricalIndex.contains`
   is likewise much faster (:issue:`21369`)
--
+- Improved performance of :meth:`HDFStore.groups` (and dependent functions like
```
jreback: you don't need to say dependent functions, you can just list .keys

jreback (changes requested):

doc/source/whatsnew/v0.23.2.txt
spott: There are a number of places that this hits (.keys, .info, .__iterate__, __len__, items, and those that call those functions.).  Should I ignore them, list the ones that I know of, or just drop the "dependent functions" part and keep it simple?

spott:
`.info()` hits `.keys()`, so there is some benchmarking supported here.

This particular performance issue wasn't really a problem till I hit a couple hundred (thousand?) keys in a hdf file, so I don't know if we want to create a dataset to test this.

doc/source/whatsnew/v0.23.2.txt
jreback: on 2nd thought this is ok

jreback:
I rebased, so ping on green.

jreback (approved):

jreback:
thanks @spott

Issue #13877, opened 2016-08-02
ERR/API: non-string types


michaelaye:
xref #5602, #9343, #13806 (comments)
#### Code Sample, a copy-pastable example if possible

```
In [2]: s = pd.Series(['a', 123, 'b'])

In [3]: s.str.startswith('a')
Out[3]:
0     True
1      NaN
2    False
dtype: object
```
#### Expected Output

raise an exception
Reasoning: If I provide NaN, it's okay to return NaN, but silently convert is hard to find. The user cannot easily drop hidden integers in an 'O' series.
## INSTALLED VERSIONS

commit: None
python: 3.5.2.final.0
python-bits: 64
OS: Darwin
OS-release: 14.5.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8

pandas: 0.18.1
nose: None
pip: 8.1.2
setuptools: 23.0.0
Cython: 0.24.1
numpy: 1.11.1
scipy: 0.18.0
statsmodels: None
xarray: 0.7.2
IPython: 5.0.0
sphinx: 1.4.1
patsy: None
dateutil: 2.5.3
pytz: 2016.6.1
blosc: None
bottleneck: None
tables: 3.2.2
numexpr: 2.6.1
matplotlib: 1.5.1
openpyxl: None
xlrd: 1.0.0
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: 0.999999999
httplib2: None
apiclient: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.8
boto: None
pandas_datareader: None

michaelaye:
Potential connection: #9343

shoyer:
@michaelaye I fixed your example -- `Series.startswith` does not exist (you need to use the `str` prefix)

michaelaye:
ah, sorry. thanks!

sinhrks:
I think all related string methods should have following kwds to cover most of usecases:

- ``errors``: whether to raise or coerce non-str input
- ``na``: specify value to fill ``NaN``  (only exists in few methods)

```
pd.Series(['a', 2, 3]).str.startswith('a')
# 0    True
# 1     NaN
# 2     NaN
# dtype: object

pd.Series(['a', 2, 3]).str.startswith('a', na=False)
# 0     True
# 1    False
# 2    False
# dtype: bool
```

related to #5672 and #13806 (comments)

[...]

PR #21547, opened 2018-06-19
ENH: set accessor for Series (WIP)

### LINKED ISSUE #4480: Add set operations to Series objects
from [this SO question](http://stackoverflow.com/questions/18079563/finding-the-intersection-between-two-series-in-pandas)

h-vetinari:
Closes #4480, and a fair bit more that was discussed in comments. In particular, it makes the `numpy` methods NaN-safe, index aware, and adds a bit of wrapping convenience.

This is still a WIP, so no tests yet. Also more work to do on docstrings, whatsnew, and some overview documentation similar to `pandas/doc/source/text.rst`. The implementation borrows heavily from `pandas/core/strings.py`.

Before I write any tests etc., I'd like to have a discussion on the API - which functions to add (I only added union/intersection/difference/symmetric difference), how to name them, and what arguments they should have. I've followed the suggestions in #13877 (not yet realised for the `.str`-accessor) to have an `errors`- and a `fill_value`-parameter, as well as the `join`-parameter from #20347.

Here's a sample API documentation:

Here's some more usage examples. The basic idea is to follow the numpy set methods, which work with `| & ^ - ` between arrays and arrays (as well as arrays and singletons), but are not NaN-safe. Basiscs:
```
s = pd.Series([{1, 2}, {2, 4}, {3, 1}])
s
# 0    {1, 2}
# 1    {2, 4}
# 2    {1, 3}
# dtype: object

s.set.union()  # apply (sequentially) to elements of calling Series
# {1, 2, 3, 4}

s.set.union({2})  # broadcast like "s.values | {2}"
# 0       {1, 2}
# 1       {2, 4}
# 2    {1, 2, 3}
# dtype: object
```

With another Series:
```
t = pd.Series([{2, 3}, {1, 2}, np.nan])
t
# 0    {2, 3}
# 1    {1, 2}
# 2       NaN
# dtype: object

s.values | t.values  # remember...
# TypeError

s.set.union(t)
# 0    {1, 2, 3}
# 1    {1, 2, 4}
# 2          NaN
# dtype: object

s.set.union(t, fill_value={5})
# 0    {1, 2, 3}
# 1    {1, 2, 4}
# 2    {1, 3, 5}
# dtype: object
```

With different indices (`fill_value` also works for NaNs created by alignment):
```
u = pd.Series(t.values, index=[1, 2, 3])
u
# 1    {2, 3}
# 2    {1, 2}
# 3       NaN
# dtype: object

s.set.union(u)
# 0          NaN
# 1    {2, 3, 4}
# 2    {1, 2, 3}
# dtype: object

s.set.union(u, join='outer')
# 0          NaN
# 1    {2, 3, 4}
# 2    {1, 2, 3}
# 3          NaN
# dtype: object

s.set.union(u, join='outer', fill_value={5})
# 0    {1, 2, 5}
# 1    {2, 3, 4}
# 2    {1, 2, 3}
# 3          {5}
# dtype: object
```

Finally, the behaviour of the `errors`-parameter. Since strings are not list-like, but *can* be coerced into a set, I made a distinction between `'coerce'` and `'wrap'`, which is the most permissive (but treats strings as singletons).
```
v = pd.Series([{1, 2}, [2, 3], 'abcd', 4, np.nan])
v
# 0    {1, 2}
# 1    [2, 3]
# 2      abcd
# 3         4
# 4       NaN
# dtype: object

v.set.union(set(), errors='raise')  # default
# ValueError

v.set.union(set(), errors='ignore')
# 0    {1, 2}
# 1    {2, 3}
# 2       NaN
# 3       NaN
# 4       NaN
# dtype: object

v.set.union(set(), errors='coerce')
# 0          {1, 2}
# 1          {2, 3}
# 2    {a, c, b, d}
# 3             NaN
# 4             NaN
# dtype: object

v.set.union(set(), errors='wrap')
# 0    {1, 2}
# 1    {2, 3}
# 2    {abcd}
# 3       {4}
# 4       NaN
# dtype: object
```

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -686,6 +686,36 @@ strings and apply several methods to it. These can be accessed like
        Series.dt
        Index.str
 
+Set handling
+~~~~~~~~~~~~~~~
+``Series.set`` can be used to access the values of the series as
+sets and apply several methods to it. These can be accessed like
+``Series.set.<function/property>``.
+
+.. autosummary::
+   :toctree: generated/
+   :template: autosummary/accessor_method.rst
+
+   Series.set.union
+   Series.set.intersect
+   Series.set.xor
+   Series.set.diff
+   Series.set.len
+
+..
+    The following is needed to ensure the generated pages are created with the
+    correct template (otherwise they would be created in the Series/Index class page)
+
+..
+    .. autosummary::
+       :toctree: generated/
+       :template: autosummary/accessor.rst
+
+       Series.str
+       Series.cat
+       Series.dt
+       Series.set
+
 .. _api.categorical:
 
 Categorical
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -77,6 +77,7 @@ from pandas.util._validators import validate_bool_kwarg
 from pandas._libs import index as libindex, tslib as libts, lib, iNaT
 from pandas.core.config import get_option
 from pandas.core.strings import StringMethods
+from pandas.core.sets import SetMethods
 
 import pandas.plotting._core as gfx
 
@@ -158,7 +159,7 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         Copy input data
     """
     _metadata = ['name']
-    _accessors = set(['dt', 'cat', 'str'])
+    _accessors = set(['dt', 'cat', 'str', 'set'])
     _deprecations = generic.NDFrame._deprecations | frozenset(
         ['asobject', 'sortlevel', 'reshape', 'get_value', 'set_value',
          'from_csv', 'valid'])
@@ -3992,6 +3993,7 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
     # Accessor Methods
     # ----------------------------------------------------------------------
     str = CachedAccessor("str", StringMethods)
+    set = CachedAccessor("set", SetMethods)
     dt = CachedAccessor("dt", CombinedDatetimelikeProperties)
     cat = CachedAccessor("cat", CategoricalAccessor)
     plot = CachedAccessor("plot", gfx.SeriesPlotMethods)
--- a/pandas/core/sets.py
+++ b/pandas/core/sets.py
@@ -1,0 +1,587 @@
+import numpy as np
+
+from pandas.core.dtypes.generic import ABCSeries
+from pandas.core.dtypes.missing import isna
+from pandas.core.dtypes.common import is_list_like
+
+import pandas.compat as compat
+from pandas.core.base import NoNewAttributesMixin
+from pandas.util._decorators import Appender
+import re
+from functools import reduce
+from operator import __or__, __xor__, __and__, __sub__
+import pandas._libs.lib as lib
+
+_shared_docs = dict()
+
+
+def _na_map(f, arr, na_result=np.nan, dtype=object):
+    # should really _check_ for NA
+    return _map(f, arr, na_mask=True, na_value=na_result, dtype=dtype)
+
+
+def _map(f, arr, na_mask=False, na_value=np.nan, dtype=object):
+    if not len(arr):
+        return np.ndarray(0, dtype=dtype)
+
+    if isinstance(arr, ABCSeries):
+        arr = arr.values
+    if not isinstance(arr, np.ndarray):
+        arr = np.asarray(arr, dtype=object)
+    if na_mask:
+        mask = isna(arr)
+        try:
+            convert = not all(mask)
+            result = lib.map_infer_mask(arr, f, mask.view(np.uint8), convert)
+        except (TypeError, AttributeError) as e:
+            # Reraise the exception if callable `f` got wrong number of args.
+            # The user may want to be warned by this, instead of getting NaN
+            if compat.PY2:
+                p_err = r'takes (no|(exactly|at (least|most)) ?\d+) arguments?'
+            else:
+                p_err = (r'((takes)|(missing)) (?(2)from \d+ to )?\d+ '
+                         r'(?(3)required )positional arguments?')
+
+            if len(e.args) >= 1 and re.search(p_err, e.args[0]):
+                raise e
+
+            def g(x):
+                try:
+                    return f(x)
+                except (TypeError, AttributeError):
+                    return na_value
+
+            return _map(g, arr, dtype=dtype)
+        if na_value is not np.nan:
+            np.putmask(result, mask, na_value)
+            if result.dtype == object:
+                result = lib.maybe_convert_objects(result)
+        return result
+    else:
+        return lib.map_infer(arr, f)
+
+
+def _noarg_wrapper(f, docstring=None, **kargs):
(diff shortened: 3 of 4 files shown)


jreback (changes requested):
tests - they should be the first thing you write

h-vetinari:
@jreback IMO there's something that needs to come before tests, and that's having a good idea about the API (which I can't decide by myself; hence the discussion here). Otherwise, a lot of time is sunk in writing obsolete tests.

Could you please comment on your thoughts for the API choices so far? The rendered docstring should give a good starting point.

h-vetinari:
@jreback I added over 200 nicely parametrized tests that cover everything I could think of. It's hopefully modular enough that renaming/changing the API won't be too much of a hassle.

I also added the same `'errors'/'fill_value'` parameters to `set.len`, which I could therefore refactor (and cut out a lot of code I had pasted from `pandas/core/strings.py`).

jreback:
at a glance it looks ok, but you have way too much complexity here. all of the ``errors`` options and the ``fill_value``. some simpler ``errors`` are ok, but still to the ``coerce, ignore, raise`` options. 

Furthermore this is fundamentally flawed in that you can use this with *anything*, in reality this should be be restricted to an ``inferred_dtype=='set'``, so that's the first thing to change.

jorisvandenbossche:
To be the devil's advocate here: given the new accessor registry api: isn't this a perfect case for having this in an external package?

h-vetinari:
@jreback I don't care much if it's only `'raise'|'ignore'|'coerce'`, just added `'wrap'` for convenience and `'skip'` for performance. Again, the `'errorrs'/'fill_value'` pair is used in many places, and propsed for `.str`-methods in #13877.

To me, the whole point of giving an `'errors'`-parameter is to *be able* to use it with anything, and whether its lists or strings or whatnot, this should be made available IMO. Especially since it's such a bitch to prepare real-world data to a point where it's just sets (not least because `fillna` doesn't take lists or sets).

Aside from the templatization to avoid repetition, I went out of my way to keep complexity at a minimum. Could you tell me what you find complex? There's one clean-up method for Series, one clean-up for `others`, and one method for executing the numpy method, all nicely segregated...

h-vetinari:
@jorisvandenbossche I maybe don't fully understand the implications/requirements of accessor registry, but since this is really just a wrapper to make the *existing* numpy methods nan-safe and index-aware, I think this should be part of the core, and not an external package?

jreback:
@h-vetinari in reality this is much better done as a real ExtensionArray type. The point of the accessors is to extend with dtype specific functionaility. But first you have to have actual set operations supported.

h-vetinari:
@jreback To which I responded as follows
> To me, the whole point of giving an 'errors'-parameter is to be able to use it with anything, [...]

I agree this is an important discussion - but there's normally lots of preprocessing necessary to get this to work with dirty real-world data, and every restriction makes it harder to actually transform real data. Something about consenting adults and all that. ;-)

jreback:
> agree this is an important discussion - but there's normally lots of preprocessing necessary to get this to work with dirty real-world data, and every restriction makes it harder to actually transform real data. Something about consenting adults and all that. ;-)

sure but that's what pandas does. set intelligent defaults.

The issue I have is that this is the very first thing that works on containers *inside* a pandas object. Everything else is a scalar. Sure its 'ok' now, you 'can' use whatever you want but there is no first class support. So need to think about best way / if to provide this support.

h-vetinari:
@jreback 
Re: 
> @h-vetinari in reality this is much better done as a real ExtensionArray type. The point of the accessors is to extend with dtype specific functionaility. But first you have to have actual set operations supported.

But then there are no methods to *transform* the data to a point where the accessor can be used. At least `.str` has `.astype(str)`. Of course, then the `'errors'/'fill_value'` kwargs could be position in a potential
`.astype(set, errors, fill_value)` call, which does have a certain appeal, I admit.

> Furthermore these are quite non-performant.

This is not a pressing issue IMO; the numpy methods are orders of magnitude faster than writing for-loops already.

h-vetinari:
@jreback 
> The issue I have is that this is the very first thing that works on containers inside a pandas object. Everything else is a scalar. Sure its 'ok' now, you 'can' use whatever you want but there is no first class support. So need to think about best way / if to provide this support.

Lots of things work on containers already, like `.str.len()`, `.str[slice]`, for lists etc.  Also, `.str.extract` and `str.split` with `expand=False` yield lists in Series, for example.

And people *are* using pandas in this way a lot - for clean data, one can just call `map(set)`, `map(len)` etc. and continue, but often the clean-up is the biggest headache. Granted, sometimes it's just an intermediate processing step on the way to a DF or a MultiIndex, but sometimes its genuinely a set/list that needs per-row processing. There's several SO questions on this, for example.

h-vetinari:
@jreback How do you propose to continue with this? As I said above, I like the idea of:
* a set dtype
* having all the conversion logic in `.astype(set, errors, fill_value)`
* having the `.set` accessor only for that dtype, and then the methods don't need the kwargs themselves

But the problem is, I have no idea what creating a new dtype would imply. It would be easy enough to write the code for `.astype(set, errors, fill_value)`, but I'm guessing that's not nearly enough.

jreback:
http://pandas-docs.github.io/pandas-docs-travis/extending.html#extension-types

this should be an extension array. the actual impl is pretty easy, it could just e a list of sets. The problem is adding an accessor to disambiguate an object dtype is a bit odd. I am not totally against it, but having a defined dtype is much much cleaner.

h-vetinari:
@jreback Not sure I follow completely - I'm afraid I'm getting a bit confused between implementing an EA (your link; apparently not so hard) and implementing a new dtype `set` (very hard?).

Further up you said:
> [...] this should be be restricted to an inferred_dtype=='set', so that's the first thing to change.

which sounds like implementing a new dtype, and now:
> [...] adding an accessor to disambiguate an object dtype is a bit odd.

which sounds (more) like the version I already implemented (or the same thing translated to EA)? Again, it's not clear to me what the requirements/recommendations are at this point.

As a side note, I think that sets are such a fundamental building block in python (as opposed to IP addresses in cyperpandas or gps coordinates in geopandas), that the methods for them shouldn't be externalised to some package - this is what the EA documentation says, if I understand correctly.

jreback:
@h-vetinari whether this is external or not is slightly orthogonal to this discussion, the point of EA is that it *can* be external (or maybe start that way).

EA is an extension array & a dtype, both of which are first class. That's how I'd like to see sets (and lists and dicts)

jreback:
closing as out of scope. This should be an EA type (could potentially be internal), but needs to be restructured to that.

Florents-Tselai:
I also needed something like this, but as @jorisvandenbossche said, initially makes sense to have it as an extension. `pip install pandas-sets`

You can have a look here https://github.com/Florents-Tselai/pandas-sets - related post: https://tselai.com/pandas-sets.html.

The implementation is very primitive (and based on `StringMethods`) , but it does the job. 

May come back later with a PR.

Issue #21526, opened 2018-06-18
ENH: allow ranges to be set on Styler .bar()


psychemedia:
When generating in table bar charts using the [Styler `.bar()`](https://github.com/pandas-dev/pandas/blob/master/pandas/io/formats/style.py):

- extreme outliers in a column mask differences in other values;
- comparisons across columns are not supported.

One solution would be to allow the setting of explicit bounds on the range of values used to limit the bar extrema using explicit values rather than `s.min()` and `s.max()`.

Cell values exceeding the range limits could be extending to the cell edge / maximum limit value and then coloured with an "overflow" colour.

psychemedia:
Pandas version 0.23.0

In the following example a dataframe is created with a column containing two outlier values (`x1`) and two columns where it would be useful to depict and compare the values over the same range even though in this example the ranges differ:

```
import pandas as pd
import numpy as np

df=pd.DataFrame({'x1':list(np.random.randint(-10,10,size=10))+[-500,1000, -1000],
               'y1':list(np.random.randint(-5,5,size=13)), 'y2':list(np.random.randint(-2,3,size=13)) })

df.style.bar( align='zero', color=[ '#5fba7d','#d65f5f'])
```

In column `x1`, the outlier values set the max/min range values used to display the styled bar and the variation in the non-outlier values is masked. It would be useful to be able to set max / min limits to so that variation in the non-outlier values can be depicted. For example, setting `min,max=(-10,10)` and then colouring squares with outlier values [-500,1000, -1000] differently to show they are outside the depicted range.

For columns `y1` and `y2`, it may be desired to display the bars over the same range, for example `min,max=(-5,5)` so that a direct visual comparison can be made across columns `y1` and `y2`.

soxofaan:
FYI: pull request #21548  now includes support for `vmin` and `vmax` arguments in `Styler.bar`

domoritz:
Is there a way to set the color range in the currently released pandas version?

soxofaan:
@domoritz  there is a `color` argument, see http://pandas.pydata.org/pandas-docs/stable/generated/pandas.io.formats.style.Styler.bar.html . Also: please specify which pandas version you are using ("currently released" doesn't say a lot).

attack68:
this issue would appear to be solved by the above PRs and acceptance by OP.

PR #21259, opened 2018-05-30
ENH: Add support for tablewise application of style.background_gradient with axis=None

soxofaan:
- [x] closes #15204
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

PR #21548, opened 2018-06-19
style.bar: add support for axis=None (tablewise application instead of rowwise or columnwise)

### LINKED ISSUE #21525: NotImplementedError: 'AnnAssign' nodes are not implemented but numpy-like syntax works
My dataframe ``df`` has a column called "Unnamed:_6" which is of dtype ``object``.

Running the following command works just fine:
```python
df[df['Unnamed:_6'] == "Foo"]
```

However, the equivalent line:
```python
df.query('Unnamed:_6 == "Foo"')
```
raises the following exception:
```python
NotImplementedError: 'AnnAssign' nodes are not implemented
```

soxofaan:
- eliminate code duplication related to style.bar with different align modes
- add support for axis=None
- fix minor bug with align 'zero' and width < 100
- make generated CSS gradients more compact

- [x] Closes #21525
- [x] tests added / passed
- [x] passes `git diff origin/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -728,6 +728,7 @@ Other
 - :meth: `~pandas.io.formats.style.Styler.background_gradient` now takes a ``text_color_threshold`` parameter to automatically lighten the text color based on the luminance of the background color. This improves readability with dark background colors without the need to limit the background colormap range. (:issue:`21258`)
 - Require at least 0.28.2 version of ``cython`` to support read-only memoryviews (:issue:`21688`)
 - :meth: `~pandas.io.formats.style.Styler.background_gradient` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None`` (:issue:`15204`)
+- :meth: `~pandas.io.formats.style.Styler.bar` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None``
 -
 -
 -
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -725,10 +725,10 @@ Build Changes
 Other
 ^^^^^
 
-- :meth: `~pandas.io.formats.style.Styler.background_gradient` now takes a ``text_color_threshold`` parameter to automatically lighten the text color based on the luminance of the background color. This improves readability with dark background colors without the need to limit the background colormap range. (:issue:`21258`)
+- :meth:`~pandas.io.formats.style.Styler.background_gradient` now takes a ``text_color_threshold`` parameter to automatically lighten the text color based on the luminance of the background color. This improves readability with dark background colors without the need to limit the background colormap range. (:issue:`21258`)
 - Require at least 0.28.2 version of ``cython`` to support read-only memoryviews (:issue:`21688`)
-- :meth: `~pandas.io.formats.style.Styler.background_gradient` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None`` (:issue:`15204`)
-- :meth: `~pandas.io.formats.style.Styler.bar` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None``
+- :meth:`~pandas.io.formats.style.Styler.background_gradient` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None`` (:issue:`15204`)
+- :meth:`~pandas.io.formats.style.Styler.bar` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None`` and setting clipping range with ``vmin`` and ``vmax`` (:issue:`21548` and :issue:`21526`)
 -
 -
 -
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -728,7 +728,7 @@ Other
 - :meth:`~pandas.io.formats.style.Styler.background_gradient` now takes a ``text_color_threshold`` parameter to automatically lighten the text color based on the luminance of the background color. This improves readability with dark background colors without the need to limit the background colormap range. (:issue:`21258`)
 - Require at least 0.28.2 version of ``cython`` to support read-only memoryviews (:issue:`21688`)
 - :meth:`~pandas.io.formats.style.Styler.background_gradient` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None`` (:issue:`15204`)
-- :meth:`~pandas.io.formats.style.Styler.bar` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None`` and setting clipping range with ``vmin`` and ``vmax`` (:issue:`21548` and :issue:`21526`)
+- :meth:`~pandas.io.formats.style.Styler.bar` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None`` and setting clipping range with ``vmin`` and ``vmax`` (:issue:`21548` and :issue:`21526`). ``NaN`` values are also handled properly.
 -
 -
 -
--- a/pandas/io/formats/style.py
+++ b/pandas/io/formats/style.py
@@ -1053,36 +1053,40 @@ class Styler(object):
 
         Parameters
         ----------
-        subset: IndexSlice, optional
-            a valid slice for ``data`` to limit the style application to
-        axis: int, default 0, meaning column-wise
-        color: str or 2-tuple/list
+        subset : IndexSlice, optional
+            A valid slice for `data` to limit the style application to.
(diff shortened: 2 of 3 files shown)


soxofaan:
example with `axis=0` (top) and `axis=None` (bottom):
screen shot 2018-06-20 at 01 11 39

soxofaan:
related  for `style.background_gradient` PR: #21259

jreback:
can you search the issues to see if we have something an issue for this? IIRC we do

soxofaan:
@jreback   did a quick search but couldn't find anything yet

note to self: #21526  should be straightforward to include in another iteration of this issue

soxofaan:
I added a commit to handle issue #21526

TomAugspurger:
@soxofaan can you update the original post to include `Closes #21525`

will try to review this sometime soon.

soxofaan:
Maybe also to be considered in this pull request:
I find the align mode names `mid` and `zero` very confusing:
- `mid` claims "midpoint is centered in the cell", but if all values are strictly positive or negative, this is not the case as zero is forced into the value range
- `zero` is about "zero is centered in the cell", which I would have called `center` (because there already is `left`), but this sounds a lot like `mid` of course.

Also: the `left` mode can lead to poor visualisation and is a bad choice as default IMHO. The bars do not start from zero at the left but from the minimum value. When this minimum is negative (or all values are negative), you get in the territory of poor/confusing visualisation.

In my refactor, I kept the original modes to avoid "bike shedding" issues, but I think I should at least raise the issue here. 
What do you think?

soxofaan:
(rebased)

pandas/io/formats/style.py
```diff
@@ -993,126 +993,63 @@ def set_properties(self, subset=None, **kwargs):
...
+            )
 
     def bar(self, subset=None, axis=0, color='#d65f5f', width=100,
-            align='left'):
+            align='left', vmin=None, vmax=None):
         """
-        Color the background ``color`` proportional to the values in each
-        column.
+        Draw bars in the cell backgrounds, with a width proportional to
+        the values
```
TomAugspurger: First sentence should be less than 80 characters and end with a `.`

pandas/io/formats/style.py
```diff
@@ -993,126 +993,63 @@ def set_properties(self, subset=None, **kwargs):
...
         Excludes non-numeric data by default.
 
         Parameters
         ----------
         subset: IndexSlice, default None
             a valid slice for ``data`` to limit the style application to
-        axis: int
+        axis: int or None
```
TomAugspurger: "or None" -> "optional"

pandas/io/formats/style.py
```diff
@@ -1127,40 +1064,35 @@ def bar(self, subset=None, axis=0, color='#d65f5f', width=100,
             - 'mid' : the center of the cell is at (max-min)/2, or
               if values are all negative (positive) the zero is aligned
               at the right (left) of the cell
+        vmin: float or None: minimum bar value, defining the left hand limit
```
TomAugspurger: "or None" -> "optional"

Description should start on the next line.

https://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html

pandas/tests/io/formats/test_style.py
```diff
@@ -350,10 +350,10 @@ def test_bar_align_left(self):
             (0, 0): ['width: 10em', ' height: 80%'],
             (1, 0): ['width: 10em', ' height: 80%',
                      'background: linear-gradient('
-                     '90deg,#d65f5f 50.0%, transparent 0%)'],
+                     '90deg,#d65f5f 50.0%, transparent 50.0%)'],
```
TomAugspurger: Why are these tests changing?

soxofaan: The original implementation generated gradient stops that were invalid: the positions (percentages) should not decrease. In the example above the first stop (`#d65f5f`) is at 50% and de second (`transparent`) was originally at 0%, but it's more correct to have the second one explicitly at 50% as well. (Browsers are expected to guess the correct intention of these incorrect cases, but you never know of course.)

In other test cases, the original css was also bit redudant, e.g.:
```
transparent 0%, transparent 10.0%, #5fba7d 10.0%
```
can be simplified to 
```
transparent 10.0%, #5fba7d 10.0%
````

or even worse, sometimes things like these were part of the gradient: `transparent 0%, transparent 0.0%, #5fba7d 0.0%,` which is completely redundant.

The new implementation improves on all these points, so original tests were adapted.

**Comment by soxofaan** [comment]
@TomAugspurger : I addressed the docstring style remarks and rebased

File: pandas/tests/io/formats/test_style.py (hunk 4 above)
Comment (soxofaan) [reply pandas/tests/io/formats/test_style.py]: And there was also a test case (`test_bar_align_zero_pos_and_neg`) with bar size bug in the original implementation when `width < 100`, which is also fixed in new implementation

**Comment by TomAugspurger** [comment]
Thanks.

There's a linting failure: https://travis-ci.org/pandas-dev/pandas/jobs/403324861#L2928.

If you could fix that, then +1 from me.

**Comment by soxofaan** [comment]
@TomAugspurger done and rebased

File: pandas/io/formats/style.py [hunk 5]
```diff
@@ -993,132 +993,67 @@ def set_properties(self, subset=None, **kwargs):
...
-                     'transparent {w:.1f}%)')
-
-        return [attrs_pos.format(c=color[1], w=(50 + x)) if x >= 0
-                else attrs_neg.format(c=color[0], w=(50 + x))
-                for x in normed]
+    def _bar(s, align, colors, width=100, vmin=None, vmax=None):
+        """Draw bar chart in dataframe cells"""
+
+        # Get input value range.
+        smin = s.values.min() if vmin is None else vmin
```
TomAugspurger: What are the types of `s`? Series or DataFrame?

We don't want `.values.min`, since if it has any missing values the output will be missing. Presumably, we'll want to skip missing values here...

Pandas doesn't yet support reducing over all axes with min / max, so you'll likely need to handle series / frame separately here

pandas/io/formats/style.py
```diff
@@ -1129,38 +1064,37 @@ def bar(self, subset=None, axis=0, color='#d65f5f', width=100,
               at the right (left) of the cell
 
               .. versionadded:: 0.20.0
+        vmin: float, optional
```
TomAugspurger: Can you add `..versionadded` for both of these.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -571,6 +571,7 @@ Other
 - :meth: `~pandas.io.formats.style.Styler.background_gradient` now takes a ``text_color_threshold`` parameter to automatically lighten the text color based on the luminance of the background color. This improves readability with dark background colors without the need to limit the background colormap range. (:issue:`21258`)
 - Require at least 0.28.2 version of ``cython`` to support read-only memoryviews (:issue:`21688`)
 - :meth: `~pandas.io.formats.style.Styler.background_gradient` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None`` (:issue:`15204`)
+- :meth: `~pandas.io.formats.style.Styler.bar` now also supports tablewise application (in addition to rowwise and columnwise) with ``axis=None`` and setting clipping range with ``vmin`` and ``vmax`` (:issue:`21548` and :issue:`21526`)
```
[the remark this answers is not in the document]
soxofaan: Ok, I'll fix that in the preceding two bullet points as well then

pandas/io/formats/style.py
soxofaan: Good point. Apparently the original implementation doesn't handle this properly either

soxofaan:
@TomAugspurger done and rebased

pandas/io/formats/style.py
```diff
@@ -993,132 +993,69 @@ def set_properties(self, subset=None, **kwargs):
...
-                     'transparent {w:.1f}%)')
-
-        return [attrs_pos.format(c=color[1], w=(50 + x)) if x >= 0
-                else attrs_neg.format(c=color[0], w=(50 + x))
-                for x in normed]
+    def _bar(s, align, colors, width=100, vmin=None, vmax=None):
+        """Draw bar chart in dataframe cells"""
+
+        # Get input value range.
+        smin = s.min().min() if vmin is None else vmin
```
TomAugspurger: This will fail for object dtypes, which *may* happen in this context. Even though it's slightly wordier, I'd prefer

```python
smin = s.min() if vmin is None else vmin
if isinstance(smin, ABCSeries):
    smin = smin.min()
```

to reduce it down to a scalar.

LGTM otherwise.

soxofaan:
@TomAugspurger done and rebased

TomAugspurger:
Thanks @soxofaan!

PR #21549, opened 2018-06-19
use ccalendar instead of np_datetime

jbrockmendel:
Do Not Merge

AFAICT the np_datetime.c versions of these functions are noticeably more performant than the cython versions.  I could use help verifying this observation since asv doesn't work well for me.

Best guess is that lookups in np_datetime.c's `days_per_month_table[2][12]` are more efficient than lookups in `ccalendar.days_per_month_array`, as there does not appear to be a way to instantiate the C data structure in cython (at least not as a module-global).

### DIFF
--- a/pandas/_libs/tslibs/offsets.pyx
+++ b/pandas/_libs/tslibs/offsets.pyx
@@ -304,6 +304,15 @@ class _BaseOffset(object):
     _day_opt = None
     _attributes = frozenset(['n', 'normalize'])
 
+    def __init__(self, n=1, normalize=False):
+        n = self._validate_n(n)
+        object.__setattr__(self, "n", n)
+        object.__setattr__(self, "normalize", normalize)
+        object.__setattr__(self, "_cache", {})
+
+    def __setattr__(self, name, value):
+        raise AttributeError("DateOffset objects are immutable.")
+
     @property
     def kwds(self):
         # for backwards-compatibility
@@ -395,13 +404,17 @@ class _BaseOffset(object):
             kwds = {key: odict[key] for key in odict if odict[key]}
             state.update(kwds)
 
-        self.__dict__ = state
+        if '_cache' not in state:
+            state['_cache'] = {}
+
+        self.__dict__.update(state)
+
         if 'weekmask' in state and 'holidays' in state:
             calendar, holidays = _get_calendar(weekmask=self.weekmask,
                                                holidays=self.holidays,
                                                calendar=None)
-            self.calendar = calendar
-            self.holidays = holidays
+            object.__setattr__(self, "calendar", calendar)
+            object.__setattr__(self, "holidays", holidays)
 
     def __getstate__(self):
         """Return a pickleable state"""
--- a/pandas/_libs/tslibs/offsets.pyx
+++ b/pandas/_libs/tslibs/offsets.pyx
@@ -18,12 +18,12 @@ cnp.import_array()
 from util cimport is_string_object, is_integer_object
 
 from ccalendar import MONTHS, DAYS
+from ccalendar cimport get_days_in_month, dayofweek
 from conversion cimport tz_convert_single, pydt_to_i8
 from frequencies cimport get_freq_code
 from nattype cimport NPY_NAT
 from np_datetime cimport (pandas_datetimestruct,
-                          dtstruct_to_dt64, dt64_to_dtstruct,
-                          is_leapyear, days_per_month_table, dayofweek)
+                          dtstruct_to_dt64, dt64_to_dtstruct)
 
 # ---------------------------------------------------------------------
 # Constants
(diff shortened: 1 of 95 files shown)


jreback:
what does the generated code look like?

pandas/_libs/tslibs/offsets.pyx
```diff
@@ -437,12 +437,6 @@ class BaseOffset(_BaseOffset):
 # ----------------------------------------------------------------------
 # RelativeDelta Arithmetic
 
-@cython.wraparound(False)
```
jreback: can you use ``days_per_month_array`` in ``ccalendar.pyx``?

jbrockmendel: > can you use `days_per_month_array` in ccalendar.pyx?

We could import it from the C file, but we can't define it directly in cython.  Since part of the point of ccalendar was to move to pure-cython, I'm not sure we want to re-introduce the C dependency.  Or if we do, we might as well keep the entire implementation in the C file.

Cython 0.28 supposedly allows for "verbatim C" in the form:

```
cdef extern from *:
     """
     foo;
     bar;
     """
```

which could be useful for small snipped.  But when I've tried this I get compile-time errors.

jbrockmendel:
> what does the generated code look like? (for that function)

After stripping out a bunch of comments:

```
static CYTHON_INLINE __pyx_t_5numpy_int32_t __pyx_f_6pandas_5_libs_6tslibs_9ccalendar_get_days_in_month(int __pyx_v_year, Py_ssize_t __pyx_v_month, CYTHON_UNUSED int __pyx_skip_dispatch) {
  __pyx_t_5numpy_int32_t __pyx_r;

  __pyx_r = (__pyx_v_6pandas_5_libs_6tslibs_9ccalendar_days_per_month_array[(((12 * __pyx_f_6pandas_5_libs_6tslibs_9ccalendar_is_leapyear(__pyx_v_year)) + __pyx_v_month) - 1)]);
  goto __pyx_L0;

  /* function exit code */
  __pyx_L0:;
  return __pyx_r;
}```

**Comment by chris-b1** [comment]
I think this will all get inlined so performance shouldn't differ anyways, but FYI you can equivalently define/import `days_per_month_array` if you wanted to. 

```python
# ccalender.pyx
cdef int* days_per_month_array

#tslib.pyx
from ccalender cimport days_per_month_array
```

jbrockmendel:
@chris-b1 I'm not sure we're on the same page as to the goal/problem.

- Seemingly-equivalent implementations in C appear to continue outperforming those in cython
- I think it has to do with the type of array that is used in the C file that can't be constructed as a global in the cython file. (see also comment above regarding cython's new "verbatim C" feature)
- I've been an advocate for reducing C dependencies and implementing things directly in cython; if there are unavoidable performance implications, I may need to rethink that position.
    - not having cython boilerplate leads (I think) to smaller .so files and (I expect) quicker compilation.  Not sure how much either of those matter.
- Profiling these functions is tricky, so all of this comes with a grain of salt

chris-b1:
Thanks for the color, I was attempting to say that I think this part is false, or at least seemed to work with what I showed above
 > the type of array that is used in the C file that can't be constructed as a global in the cython file. (see also comment above regarding cython's new "verbatim C" feature)

Interesting (surprising) that performance is different, I will look closer

chris-b1:
At least with `days_in_month` getting similar performance both ways
```python
# master
big_ts = pd.date_range('1999-01-01', periods=1_000_000, freq='H')
%timeit big_ts.daysinmonth
91.6 ms ± 11.7 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

# patch
big_ts = pd.date_range('1999-01-01', periods=1_000_000, freq='H')
%timeit big_ts.daysinmonth
94.7 ms ± 6.98 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

patch
```diff
diff --git a/pandas/_libs/tslibs/fields.pyx b/pandas/_libs/tslibs/fields.pyx
index ccf67e765..be53dfe8b 100644
--- a/pandas/_libs/tslibs/fields.pyx
+++ b/pandas/_libs/tslibs/fields.pyx
@@ -14,13 +14,19 @@ from numpy cimport ndarray, int64_t, int32_t, int8_t
 cnp.import_array()

 from ccalendar import get_locale_names, MONTHS_FULL, DAYS_FULL
-from ccalendar cimport (get_days_in_month, is_leapyear, dayofweek,
+from ccalendar cimport (is_leapyear, dayofweek,
                         get_week_of_year, get_day_of_year)
 from np_datetime cimport (pandas_datetimestruct, pandas_timedeltastruct,
+                          days_per_month_table,
                           dt64_to_dtstruct, td64_to_tdstruct)
 from nattype cimport NPY_NAT


+@cython.wraparound(False)
+@cython.boundscheck(False)
+cdef inline int get_days_in_month(int year, int month) nogil:
+    return days_per_month_table[is_leapyear(year)][month - 1]
+
 def get_time_micros(ndarray[int64_t] dtindex):
     """
     Return the number of microseconds in the time component of a
```

jreback:
@jbrockmendel status of this? (as the DO NOT MERGE at the top is clear!)

jbrockmendel:
If I understand @chris-b1 correctly then the perf discrepancy may not be an issue, in which case this is a clear (albeit small) improvement, so merge is OK.

jreback:
thanks @jbrockmendel

PR #21568, opened 2018-06-21
[DOC]: Updating merge.rst to resolve rendering issues

### LINKED ISSUE #21562: DOC: Wrong documentation rendering when a space is missing after backticks
It seems like the documentation is not rendering correctly when a space is not present after using backticks.

For example `making ``DataFrame`` render with monospace font` will work as expected.

But `using plural ``DataFrame``s does not have a space between `` and s` does not identify the closing backticks and the whole paragraph may render with monospace.

This image shows a real case in https://pandas.pydata.org/pandas-docs/stable/merging.html#ignoring-indexes-on-the-concatenation-axis:

pandas_doc_render_backticks

Couldn't find a way to obtain the expected result (monospace text followed by normal text with a space in between). Not sure if there is a way, but if it's not possible, I think it would be better to use something like the next options (instead of the wrong rendering):
- Avoid having the `s` for plurals
- Leave a space before the `s`
- Use a single quote (e.g. ` ``DataFrame``'s `) which seems to be rendering correctly as ``DataFrame``'s.

uds5501:
Fixing Documentation according to rendering requirements

- [x] xref #21562
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/doc/source/merging.rst
+++ b/doc/source/merging.rst
@@ -279,7 +279,7 @@ need to be:
 
 Ignoring indexes on the concatenation axis
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-For ``DataFrame``s which don't have a meaningful index, you may wish to append
+For ``DataFrame`` s which don't have a meaningful index, you may wish to append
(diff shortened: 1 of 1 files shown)


jreback:
thanks!

PR #21569, opened 2018-06-21
REF: multi_take is now able to tackle all list-like (non-bool) cases

toobaz:
- [x] tests passed

This is basically a consequence of #21503 - the code path for indexing when ``key`` is a collection is now the same for multi_take and for single take.

(Next step will be to merge the other code paths too, and approach with multi_take _all_ cases in which mutiple axes are being indexed, in whatever fashion)


jreback:
lgtm. merge on green.

Issue #19026, opened 2018-01-01
PERF/BENCHMARK: comprehensive cat-groupby benchmarks


jreback:
we have a number of groupby benchmarks with categoricals, but I think we need a comprehensive set to exercise combinations of:

xref [SO](https://stackoverflow.com/questions/48042952/pandas-dataframe-aggregate-on-column-whos-dtype-category-leads-to-slow-perf)

groupby on cat/object columns
cython function (e.g. ``first/max/....``)
``.agg`` variants of cython functions

```
In [4]: import pandas as pd
   ...: import numpy as np
   ...: animals = ['Dog', 'Cat']
   ...: days = ['Sunday', 'Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday','Saturday']
   ...: N = 1000000
   ...: df = pd.DataFrame({'animals': np.array(animals).take(np.random.randint(0, len(animals), size=N)),
   ...:                    'days': np.array(days).take(np.random.randint(0, len(days), size=N))})
   ...: df2 = df.copy()
   ...: df2['animals'] = df2['animals'].astype('category')
   ...: 
   ...: df3 = df2.copy()
   ...: df3['animals'] = df3['animals'].cat.codes
   ...: 
   ...: # group on object, aggregate cat
   ...: print('groupby on object')
   ...: %timeit df.groupby('days').agg({'animals': 'first'})
   ...: %timeit df2.groupby('days').agg({'animals': 'first'})
   ...: 
   ...: 
   ...: # group on cat, aggregate cat
   ...: print('groupby on cat / codes / agg')
   ...: %timeit df.groupby('animals').agg({'animals': 'first'})
   ...: %timeit df2.groupby('animals').agg({'animals': 'first'})
   ...: %timeit df3.groupby('animals').agg({'animals': 'first'})
   ...: 
   ...: print('groupby on cat / codes / cython')
   ...: %timeit df2.groupby('animals').first()
   ...: %timeit df3.groupby('animals').first()
   ...: 
[1] groupby on object
270 ms +- 5.22 ms per loop (mean +- std. dev. of 7 runs, 1 loop each)
118 ms +- 1.96 ms per loop (mean +- std. dev. of 7 runs, 10 loops each)
[2] groupby on cat / codes / agg
147 ms +- 2.53 ms per loop (mean +- std. dev. of 7 runs, 10 loops each)
[...]

PR #15885, opened 2017-04-03
BUG: groupby's first/last functions maintain Series rather than convert to numpy array (#15884)

chrisaycock:
- [x] closes #15884 
 - [x] tests added / passed
 - [x] passes ``git diff upstream/master --name-only -- '*.py' | flake8 --diff``
 - [x] whatsnew entry

PR #21573, opened 2018-06-21
BUG: first/last lose timezone in groupby with as_index=False

### LINKED ISSUE #15884: BUG: first() loses the timezone in groupby
need a systematic test of all groupby methods here.

This works as expected:

```
pd.DataFrame({'time': [pd.Timestamp('2012-01-01 13:00:00+00:00')],
              'A': [3]}).groupby('A', as_index=False).head(1)
```

However, this loses the timezone:

```
pd.DataFrame({'time': [pd.Timestamp('2012-01-01 13:00:00+00:00')],
              'A': [3]}).groupby('A', as_index=False).first()
```

reidy-p:
- [ ] closes #15884
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -225,7 +225,7 @@ Plotting
 Groupby/Resample/Rolling
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
--
+- Bug in :func:`pandas.core.groupby.first` and :func:`pandas.core.groupby.last` with ``as_index=False`` leading to the loss of timezone information (:issue:`15884`)
 -
 -
 
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -225,7 +225,7 @@ Plotting
 Groupby/Resample/Rolling
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-- Bug in :func:`pandas.core.groupby.first` and :func:`pandas.core.groupby.last` with ``as_index=False`` leading to the loss of timezone information (:issue:`15884`)
+- Bug in :func:`pandas.core.groupby.GroupBy.first` and :func:`pandas.core.groupby.GroupBy.last` with ``as_index=False`` leading to the loss of timezone information (:issue:`15884`)
 -
 -
 
--- a/pandas/core/groupby/groupby.py
+++ b/pandas/core/groupby/groupby.py
@@ -4740,10 +4740,7 @@ class DataFrameGroupBy(NDFrameGroupBy):
 
     def _wrap_agged_blocks(self, items, blocks):
         if not self.as_index:
-            if blocks[0].values.ndim > 1:
-                index = np.arange(blocks[0].values.shape[1])
-            else:
-                index = np.arange(blocks[0].values.shape[0])
+            index = np.arange(blocks[0].values.shape[-1])
             mgr = BlockManager(blocks, [items, index])
             result = DataFrame(mgr)
 
--- a/pandas/core/groupby/groupby.py
+++ b/pandas/core/groupby/groupby.py
@@ -4740,7 +4740,10 @@ class DataFrameGroupBy(NDFrameGroupBy):
 
     def _wrap_agged_blocks(self, items, blocks):
         if not self.as_index:
-            index = np.arange(blocks[0].values.shape[1])
+            if blocks[0].values.ndim > 1:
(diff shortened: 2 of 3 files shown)


pandas/core/groupby/groupby.py
```diff
@@ -4740,7 +4740,10 @@ def _wrap_transformed_output(self, output, names=None):
 
     def _wrap_agged_blocks(self, items, blocks):
         if not self.as_index:
-            index = np.arange(blocks[0].values.shape[1])
+            if blocks[0].values.ndim > 1:
```
reidy-p: In a case such as:
```
pd.DataFrame({'time': [pd.Timestamp('2012-01-01 13:00:00+00:00')],
              'A': [3]}).groupby('A', as_index=False).first()
```
the ``blocks[0].values`` is a ``DatetimeIndex`` and not an ``array`` so trying to call ``shape[1]`` on a ``DTI`` results in an out of index error and then the compat routine for ``first`` or ``last`` are called which leads to the timezone being lost.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -225,7 +225,7 @@ Plotting
 Groupby/Resample/Rolling
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
--
+- Bug in :func:`pandas.core.groupby.first` and :func:`pandas.core.groupby.last` with ``as_index=False`` leading to the loss of timezone information (:issue:`15884`)
```
jschendel: I think the `:func:` links should be `pandas.core.groupby.GroupBy.first` (and likewise for `last`)

pandas/core/groupby/groupby.py
mroeschke: > the compat routine for first or last are called which leads to the timezone being lost.

Is there a case where DatetimeTZ data can go through the compat routine (maybe with `as_index=True`?) It would be great for the data type to also be preserved there as well.

pandas/core/groupby/groupby.py
```diff
@@ -4740,7 +4740,7 @@ def _wrap_transformed_output(self, output, names=None):
 
     def _wrap_agged_blocks(self, items, blocks):
         if not self.as_index:
-            index = np.arange(blocks[0].values.shape[1])
+            index = np.arange(blocks[0].values.shape[-1])
```
jreback: @reidy-p pushed a simplification. but maybe need some additional tests that do this when a column is selected

e.g. ``df.groupby('id', as_index=False)['foo'].first()``

jreback (changes requested):

pandas/core/groupby/groupby.py
reidy-p: Yeah I have also been wondering about the same thing. The following case goes through the compat routine and consequently loses the timezone:
```
In [2]: df = pd.DataFrame({'group': [1, 1, 2],
                           'category_string': pd.Series(list('abc')).astype('category'),
                           'datetimetz': pd.date_range('20130101', periods=3, tz='US/Eastern'})
In [3]: df.groupby('group').first()
Out[3]: 
      category_string          datetimetz                                                
group                                                                                                                                                     
1                   a 2013-01-01 05:00:00                                                                      
2                   c 2013-01-03 05:00:00    
```

But if we exclude the categorical column it doesn't go through the compat routine and preserves the timezone information:
```
In[4]: df[['group', 'datetimetz']].groupby('group').first()
Out[4]:
                     datetimetz                                                       
group                                                                  
1     2013-01-01 00:00:00-05:00                                                            
2     2013-01-03 00:00:00-05:00
```

So if we have the categorical column do we want to legitimately go through the compat routine? And, if so, should we preserve the timezone in the compat routine? I think this might actually be quite straightforward (see #15885)

doc/source/whatsnew/v0.24.0.txt
reidy-p: Yep, you're right, thanks!

pandas/core/groupby/groupby.py
reidy-p: Nice simplification. I added some new tests.

mroeschke:
@reidy-p 
>So if we have the categorical column do we want to legitimately go through the compat routine? And, if so, should we preserve the timezone in the compat routine?

I am not too familiar of the conditions in which the data gets passed through the compat routine, but timezones should be preserved in the compat routine.  Yes, looks like #15885 should solve that issue (and offer a performance boost for Categoriacals as well xref #19026)

jreback (approved):

jreback:
thanks @reidy-p very nice!

PR #21579, opened 2018-06-21
DOC: update the Series.any / Dataframe.any docstring

strickvl:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -463,7 +463,7 @@ class NDFrame(PandasObject, SelectionMixin):
 
         See Also
         --------
-        ndarray.ndim
+        ndarray.ndim : Number of array dimensions.
 
         Examples
         --------
@@ -487,7 +487,7 @@ class NDFrame(PandasObject, SelectionMixin):
 
         See Also
         --------
-        ndarray.size
+        ndarray.size : Number of elements in the array.
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -9420,7 +9420,12 @@ def _doc_parms(cls):
 _any_see_also = """\
 See Also
 --------
-pandas.DataFrame.all : Return whether all elements are True.
+any : Standard library version of this method.
```
WillAyd: Can't recall any other instance where we've listed a standard library method like this before; I'd remove as this is about as implicit as it gets

strickvl: There are a few. Was at a Python Sprint in London and the organiser suggested it. But I'll remove. Thank you!

WillAyd: Hmm OK. Well if that organizer was @datapythonista then his word trumps mine :-)

strickvl: It was! Should I revert back to how it was?

WillAyd: I'd still keep it as is because its trivial either way so the least amount of effort at this point (i.e. doing nothing) I think is the best course of action. Fine if others disagree

datapythonista: Not really sure if we've got a discussion about it. But yes, there are some other methods like `str.split` where equivalent standard library were listed.

I'm happy having or not having them. I guess some user new to Python could be happy to discover that Python itself does it. And I don't think the `See Also` is never that long. But happy if we don't include them.

jreback:
@datapythonista ?

jreback:
thanks @strickvl

PR #21584, opened 2018-06-22
BUG: Let IntervalIndex constructor override inferred closed

### LINKED ISSUE #19370: Allow IntervalIndex constructor closed parameter to override inferred closed

The `closed` parameter of `IntervalIndex` is currently a bit inconsistent, and it's behavior is input dependent.  By my count, there are 3 different behaviors:

**1.  `closed` parameter overrides the inferred `closed` when the input data is empty or purely NA**
```python
In [2]: pd.IntervalIndex([], closed='both')
Out[2]:
IntervalIndex([]
              closed='both',
              dtype='interval[int64]')

In [3]: pd.IntervalIndex([np.nan, np.nan], closed='neither')
Out[3]:
IntervalIndex([nan, nan]
              closed='neither',
              dtype='interval[float64]')
```

**2. `closed` is ignored when the input data is an `IntervalIndex`:**
```python
In [4]: ii = pd.interval_range(0, 3, closed='both')

In [5]: ii
Out[5]:
IntervalIndex([[0, 1], [1, 2], [2, 3]]
              closed='both',
              dtype='interval[int64]')

In [6]: pd.IntervalIndex(ii, closed='neither')
Out[6]:
IntervalIndex([[0, 1], [1, 2], [2, 3]]
              closed='both',
              dtype='interval[int64]')
```

**3. `closed` raises a `ValueError` when a list-like of `Interval` objects with a conflicting value for `closed` is passed as input data:**
```python
In [7]: ivs = [pd.Interval(0, 1, closed='both'), pd.Interval(1, 2, closed='both')]

In [8]: pd.IntervalIndex(ivs, closed='neither')
ValueError: conflicting values for closed: constructor got 'neither', inferred from data 'both'
```
xref #18421 where I implemented this behavior; I'm no longer sure it's the best approach to take.

#### Problem description
This behavior is inconsistent, and could lead to confusion.  As it's currently implemented, the `closed` parameter is useless in the majority of cases, either doing nothing or raising an error.

#### Expected Output
I'd expect there to be consistency as to how the `closed` parameter is handled.  My preference would be to have it override the inferred `closed` if explicitly passed by the user.  Note that this is consistent with how the `dtype` parameter is used in constructors throughout the codebase.

xref #19371 (may render this issue moot)

jschendel:
- [X] closes #19370
- [X] tests added / passed
- [X] whatsnew entry

Makes `IntervalIndex` constructor behavior consistent: the `closed` parameter, if specified, takes priority over the inferred `closed`.

Comments:
- Modified the `intervals_to_interval_bounds` function to optionally raise if mixed values of `closed` are  encountered instead of automatically raising.
   - Allow creating an `IntervalIndex` from mixed closed lists, e.g.`[Interval(0, 1, closed='left'), Interval(2, 3, closed='right')]`, only if `closed` is specified during construction.
  -  The above will still raise if `closed` is not passed to the constructor.
   - This appears to only be called in the constructor currently.
- Added an Interval subsection to the Bug Fixes section of the 0.24.0 whatsnew, since I anticipate that there will be a non-negligible number of interval related fixes.

### DIFF
--- a/pandas/tests/indexes/interval/test_construction.py
+++ b/pandas/tests/indexes/interval/test_construction.py
@@ -317,13 +317,7 @@ class TestClassConstructors(Base):
         pass
 
     def test_constructor_errors(self, constructor):
-        # mismatched closed inferred from intervals vs constructor.
-        ivs = [Interval(0, 1, closed='both'), Interval(1, 2, closed='both')]
-        msg = 'conflicting values for closed'
-        with tm.assert_raises_regex(ValueError, msg):
-            constructor(ivs, closed='neither')
-
-        # mismatched closed within intervals
+        # mismatched closed within intervals with no constructor override
         ivs = [Interval(0, 1, closed='right'), Interval(2, 3, closed='left')]
         msg = 'intervals must all be closed on the same side'
         with tm.assert_raises_regex(ValueError, msg):
@@ -340,6 +334,24 @@ class TestClassConstructors(Base):
                "is not an interval")
         with tm.assert_raises_regex(TypeError, msg):
             constructor([0, 1])
+
+    @pytest.mark.parametrize('data, closed', [
+        ([], 'both'),
+        ([np.nan, np.nan], 'neither'),
+        ([Interval(0, 3, closed='neither'),
+          Interval(2, 5, closed='neither')], 'left'),
+        ([Interval(0, 3, closed='left'),
+          Interval(2, 5, closed='right')], 'neither'),
+        (IntervalIndex.from_breaks(range(5), closed='both'), 'right')])
+    def test_override_inferred_closed(self, constructor, data, closed):
+        # GH 19370
+        if isinstance(data, IntervalIndex):
+            tuples = data.to_tuples()
+        else:
+            tuples = [(iv.left, iv.right) if notna(iv) else iv for iv in data]
+        expected = IntervalIndex.from_tuples(tuples, closed=closed)
+        result = constructor(data, closed=closed)
+        tm.assert_index_equal(result, expected)
 
 
 class TestFromIntervals(TestClassConstructors):
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -194,6 +194,13 @@ Strings
 -
 -
 
+Interval
+^^^^^^^^
+
+- Bug in the :class:`IntervalIndex` constructor where the ``closed`` parameter did not always override the inferred ``closed`` (:issue:`19370`)
+-
+-
+
 Indexing
 ^^^^^^^^
 
--- a/pandas/_libs/interval.pyx
+++ b/pandas/_libs/interval.pyx
@@ -335,11 +335,17 @@ cdef class Interval(IntervalMixin):
 
 @cython.wraparound(False)
 @cython.boundscheck(False)
-cpdef intervals_to_interval_bounds(ndarray intervals):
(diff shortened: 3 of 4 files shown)


pandas/tests/indexes/interval/test_construction.py
```diff
@@ -317,13 +317,7 @@ def test_generic_errors(self, constructor):
         pass
 
     def test_constructor_errors(self, constructor):
-        # mismatched closed inferred from intervals vs constructor.
```
jreback: hmm, I think this is an invalid interval as its closed on both sides?

jreback (changes requested):
question

pandas/tests/indexes/interval/test_construction.py
jschendel: Yeah, my logic here is that if there's a conflict, then the user had to explicitly pass the conflicting value to the constructor since the default for `closed` is `None`.  Being done explicitly seems to indicate (at least a degree of) user intention, so seems reasonable to defer to the user passed value if there's a conflict, as opposed to forcing preprocessing onto the user.

Can certainly see an argument for raising in this case though, so can modify to make the behavior consistent in regards to raising if that's the preferred behavior.

jreback: so thinking of the scenario where you have a list of Intervals and you are then overriding to make it uniform, but this actually changes what this represents; the user doesn't even know this. maybe a warning would be in order? ISince the override is explicit the user 'knows what they are doing', but do they really?

jreback: cc @TomAugspurger

TomAugspurger: Hmm, not sure. I think it would be OK to override without warning or error, but not really sure.

jschendel: Not opposed to adding a warning; will do that within the next day.  Shouldn't be much work and will be easy enough to remove if we decide we don't want it later.

Another use case for overriding is to change `closed` for an existing `IntervalIndex`.  A little bit different than the list of intervals case, but I think we'd want the constructor to have consistent behavior for list of intervals input and `IntervalIndex` input.

On master the `closed` parameter is ignored for `IntervalIndex` input:
```python
In [2]: index = pd.interval_range(0, 3, closed='both')

In [3]: pd.IntervalIndex(index, closed='neither')
Out[3]:
IntervalIndex([[0, 1], [1, 2], [2, 3]]
              closed='both',
              dtype='interval[int64]')
```
And the best workaround (to my knowledge) is to deconstruct and pass to one of the `from_*` methods, which seems unnecessarily tedious, e.g.
```python
In [4]: pd.IntervalIndex.from_arrays(index.left, index.right, closed='neither')
Out[4]:
IntervalIndex([(0, 1), (1, 2), (2, 3)]
              closed='neither',
              dtype='interval[int64]')

In [5]: pd.IntervalIndex.from_tuples(index.to_tuples(), closed='neither')
Out[5]:
IntervalIndex([(0, 1), (1, 2), (2, 3)]
              closed='neither',
              dtype='interval[int64]')
```

jreback: maybe think about adding a ``set_closed`` method (similar to how we have accessors for names/labels/levels and so on on Indexes), though in the construtor is fine as well.

ok so merge this as is and do  followup? (either way ok)

jschendel: > ok so merge this as is and do followup?

Merging as is.  To be clear, for the followup are you referring to adding warnings, or just the `set_closed`?

Issue #21500, opened 2018-06-15
TST: Use numeric dtype fixtures for testing


jschendel:
Some new numeric dtype fixtures were implemented in #21432:
https://github.com/pandas-dev/pandas/blob/9e982e18678c47c83ce2bb5932f844e2d9ee21f3/pandas/conftest.py#L183-L243

It would be nice to use these fixtures where appropriate in the existing tests, e.g.
https://github.com/pandas-dev/pandas/blob/9e982e18678c47c83ce2bb5932f844e2d9ee21f3/pandas/tests/indexes/datetimes/test_construction.py#L527-L528

Additional comments:
- These changes can be split across multiple PR's so no need to do all changes at once
  - Not how many instances of this will need to be replaced, so this will need to be investigated
- This might involve expanding the dtype coverage for some tests
   - The example test shown above could also be tested over uint dtypes
- Can probably drop `float16` dtypes from tests since it's barely supported
  - See https://github.com/pandas-dev/pandas/pull/21432#discussion_r194702410
- Might be necessary to create additional fixtures
   - Perhaps a fixture for the combination of float + signed integer + unsigned integer dtypes

gfyoung:
They have propagated since this issue was opened.

PR #21588, opened 2018-06-22
TST: Use int fixtures in test_construction.py

gfyoung:
Title is self-explanatory.

Partially addresses #21500.


jreback:
thanks!

PR #21589, opened 2018-06-22
clarifying regex pipe behavior

mitchnegus:
[Current documentation for `pandas.Series.str.contains()`](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.str.contains.html) states that the `|` acts as a regex **and** operator ("Returning ‘house’ and ‘parrot’ within same string."), while I believe it is actually an **or** operator. I've updated the phrase and example to demonstrate this behavior in what I think is a clearer way, though I'm definitely open to suggestions to improve it further.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -335,11 +335,11 @@ def str_contains(arr, pat, case=True, flags=0, na=np.nan, regex=True):
     4    False
     dtype: bool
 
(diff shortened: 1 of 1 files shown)


jreback:
thanks, yeah this was a typo

PR #21590, opened 2018-06-22
API/REGR: (re-)allow neg/pos unary operation on object dtype

### LINKED ISSUE #21380: pandas 0.23 broke unary negative expression on Decimal data type

```python
import pandas as pd
from decimal import Decimal as D
series = pd.Series([D(1)])
print(series)
print(-(series))
```
#### Problem description

I'm dealing with decimal data where exact representation is required, thus I use Python's Decimal type with pandas. With the update from 0.22 to 0.23, the unary negative expression broke.

#### Expected Output (from 0.22)
```
>>> import pandas as pd
>>> from decimal import Decimal as D
>>> series = pd.Series([D(1)])
>>> print(series)
0    1
dtype: object
>>> print(-(series))
0    -1
dtype: object
```

#### Actual Output (from 0.23)
```
>>> import pandas as pd
>>> from decimal import Decimal as D
>>> series = pd.Series([D(1)])
>>> print(series)
0    1
dtype: object
>>> print(-(series))
  File "python3.6/site-packages/pandas/core/generic.py", line 1124, in __neg__
    .format(values.dtype))
TypeError: Unary negative expects numeric dtype, not object
```
#### Workaround (in 0.23)
Broadcasting against 0 has the expected effect:
```
>>> 0-series
0    -1
dtype: object
>>> (0-series).iloc[0]
Decimal('-1')
```

<details>
>>> pd.show_versions()

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.16.13-300.fc28.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.23.0
pytest: None
pip: 9.0.3
setuptools: 38.5.1
Cython: None
numpy: 1.14.4
scipy: 1.0.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: None
patsy: 0.4.1
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 2.1.0
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: 0.999999999
sqlalchemy: 1.1.15
pymysql: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

jorisvandenbossche:
closes #21380

This is an easy fix to simply be more forgiveable and just try the operation:

- For object dtype, this is IMO the better way: simply let the scalar values decide if it can do those operations or not. 
  The scalar values (eg strings) already raise an informative error message. Eg for a series with string you get `TypeError: bad operand type for unary -: 'str'` for unary negative operation.
- But with two caveats to discuss:
  - For datetime64 data, the error message is not as user friendly as our custom one: "TypeError: ufunc 'negative' did not contain a loop with signature matching types dtype('<M8[ns]') dtype('<M8[ns]')"
  - For the unary positive operation, numpy has apparently different rules: it does allow datetime64 data and returns itself (the same for object string data). Because of this, the current tests fail (it expects TypeError for `+s_string`)

<details>
<summary>Details with examples</summary>

```
In [14]: s_string = pd.Series(['a', 'b'])

In [15]: -s_string
<ipython-input-15-91308cf04ad2> in <module>()
----> 1 -s_string

~/scipy/pandas/pandas/core/generic.py in __neg__(self)
   1119             arr = operator.inv(values)
   1120         else:
-> 1121             arr = operator.neg(values)
   1122         return self.__array_wrap__(arr)
   1123 

TypeError: bad operand type for unary -: 'str'

In [16]: +s_string
Out[16]: 
0    a
1    b
dtype: object

In [17]: s_datetime = pd.Series(pd.date_range('2012', periods=3))

In [18]: -s_datetime
<ipython-input-18-b460860ba74c> in <module>()
----> 1 -s_datetime

~/scipy/pandas/pandas/core/generic.py in __neg__(self)
   1119             arr = operator.inv(values)
   1120         else:
-> 1121             arr = operator.neg(values)
   1122         return self.__array_wrap__(arr)
   1123 

TypeError: ufunc 'negative' did not contain a loop with signature matching types dtype('<M8[ns]') dtype('<M8[ns]')

In [19]: +s_datetime
Out[19]: 
0   2012-01-01
1   2012-01-02
2   2012-01-03
dtype: datetime64[ns]

In [20]: 

In [20]: from decimal import Decimal
    ...: s_decimal = pd.Series([Decimal(1)])

In [21]: -s_decimal
Out[21]: 
0    -1
dtype: object

In [22]: +s_decimal
Out[22]: 
0    1
dtype: object


```
</details>

### DIFF
--- a/doc/source/whatsnew/v0.23.2.txt
+++ b/doc/source/whatsnew/v0.23.2.txt
@@ -18,7 +18,7 @@ Fixed Regressions
 
 - Fixed regression in :meth:`to_csv` when handling file-like object incorrectly (:issue:`21471`)
 - Bug in both :meth:`DataFrame.first_valid_index` and :meth:`Series.first_valid_index` raised for a row index having duplicate values (:issue:`21441`)
--
+- Fixed regression in unary negative operations with object dtype (:issue:`21380`)
 
 .. _whatsnew_0232.performance:
 
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1118,11 +1118,8 @@ class NDFrame(PandasObject, SelectionMixin):
         values = com._values_from_object(self)
         if is_bool_dtype(values):
             arr = operator.inv(values)
-        elif is_numeric_dtype(values) or is_timedelta64_dtype(values):
-            arr = operator.neg(values)
-        elif (is_object_dtype(values) and
-              lib.infer_dtype(values) not in ['string', 'bytes', 'unicode']):
-            # explicity allow object dtypes that are not strings, gh-21380
+        elif (is_numeric_dtype(values) or is_timedelta64_dtype(values)
+                or is_object_dtype(values)):
             arr = operator.neg(values)
         else:
             raise TypeError("Unary negative expects numeric dtype, not {}"
@@ -1133,11 +1130,8 @@ class NDFrame(PandasObject, SelectionMixin):
         values = com._values_from_object(self)
         if (is_bool_dtype(values) or is_period_arraylike(values)):
             arr = values
-        elif is_numeric_dtype(values) or is_timedelta64_dtype(values):
-            arr = operator.pos(values)
-        elif (is_object_dtype(values) and
-              lib.infer_dtype(values) not in ['string', 'bytes', 'unicode']):
-            # explicity allow object dtypes that are not strings, gh-21380
+        elif (is_numeric_dtype(values) or is_timedelta64_dtype(values)
+                or is_object_dtype(values)):
             arr = operator.pos(values)
         else:
             raise TypeError("Unary plus expects numeric dtype, not {}"
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1117,22 +1117,16 @@ class NDFrame(PandasObject, SelectionMixin):
         values = com._values_from_object(self)
         if is_bool_dtype(values):
             arr = operator.inv(values)
-        elif (is_numeric_dtype(values) or is_timedelta64_dtype(values)):
-            arr = operator.neg(values)
         else:
-            raise TypeError("Unary negative expects numeric dtype, not {}"
-                            .format(values.dtype))
+            arr = operator.neg(values)
         return self.__array_wrap__(arr)
 
     def __pos__(self):
         values = com._values_from_object(self)
         if (is_bool_dtype(values) or is_period_arraylike(values)):
             arr = values
-        elif (is_numeric_dtype(values) or is_timedelta64_dtype(values)):
-            arr = operator.pos(values)
         else:
-            raise TypeError("Unary plus expects numeric dtype, not {}"
-                            .format(values.dtype))
+            arr = operator.pos(values)
         return self.__array_wrap__(arr)
 
     def __invert__(self):
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -10,7 +10,7 @@ import json
 import numpy as np
 import pandas as pd
 
-from pandas._libs import tslib, properties, lib
+from pandas._libs import tslib, properties
 from pandas.core.dtypes.common import (
     _ensure_int64,
     _ensure_object,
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
cc @deniederhut 

For the unary positive, if we care I can check for datetime64 and raise a TypeError for that (although we can also follow numpy's example). For the object dtype (eg strings), that's not really possible since it is exactly the idea to allow it for those types in an object column that allow it.

jreback:
code change looks good (though you simply may need to add ``is_object_type`` to the check as we are strict on it).

jbrockmendel:
+1.  Thoughts on applying the same pass-through approach to `Index.__sub__`?

jorisvandenbossche:
> though you simply may need to add is_object_type to the check as we are strict on it

Do you mean to only specifically call `operator.neg(values)` for object dtype (as we do for numeric), so it still fails for datetime64?

jreback:
> Do you mean to only specifically call operator.neg(values) for object dtype (as we do for numeric), so it still fails for datetime64?

I meant more just to extend the if from the original code, like

```
        elif (is_numeric_dtype(values) or is_timedelta64_dtype(values) or is_object_type(values)):
            arr = operator.neg(values)
       else:
            raise TypeError
```

jreback:
@jorisvandenbossche I pushed a fix to make this work

jorisvandenbossche:
yeah, sorry for the delay, getting to it today

I am fine with the `is_object_dtype` check (to explicitly not allow the datetime64), but personally I don't think we should infer anything if it is object. The error message is already fine in case of `__neg__`, and I don't really care about `+s_string` working (it is consistent with numpy). So I would rather do what you posted above in https://github.com/pandas-dev/pandas/pull/21590#issuecomment-400118383

jreback:
> I am fine with the is_object_dtype check (to explicitly not allow the datetime64), but personally I don't think we should infer anything if it is object. The error message is already fine in case of __neg__, and I don't really care about +s_string working (it is consistent with numpy). So I would rather do what you posted above in #21590 (comment)

I tried that and it failed your tests :> So maybe they are too strict. I didn't dive too deep here.

jreback:
in particular this (on master)

```
In [9]: df = pd.DataFrame({'A': ['a', 'b']})

In [10]: + df
Out[10]: 
   A
0  a
1  b

In [11]: - df
<ipython-input-11-42e20adaf419> in <module>()
----> 1 - df

~/pandas/pandas/core/generic.py in __neg__(self)
   1125               not is_string_like_dtype(np.array(values))):#lib.infer_dtype(values) not in ['string', 'bytes', 'unicode']):
   1126             # explicity allow object dtypes that are not strings, gh-21380
-> 1127             arr = operator.neg(values)
   1128         else:
   1129             raise TypeError("Unary negative expects numeric dtype, not {}"

TypeError: bad operand type for unary -: 'str'
```

jorisvandenbossche:
Yes, not inferring strings means that we need to update the test (but only for `__pos__`), because `+s_strings` would now work while it fails on master. 
I didn't do that initially because I first wanted to get an idea if you were OK with allowing something like `+s_strings`.

jorisvandenbossche:
> Thoughts on applying the same pass-through approach to Index.__sub__?

Yes, I think that is a good idea, but let's leave that for a separate PR / 0.24.0

jorisvandenbossche:
I think the policy of numpy regarding `+` is simply to return the original object regardless whether it would otherwise support the notion of negative/positive or addition.

> ok with matching numpy (for pos only).

Yes (for object dtype now, for datetime we still raise), and for `neg` our behaviour was already consistent with numpy before.

jreback:
thank you sir!

PR #21595, opened 2018-06-22
TST: Add interval closed fixture to top-level conftest

jschendel:
Noticed identical copies of this fixture in multiple places, so seemed reasonable to dedupe and move to the top-level `conftest.py`.  I think I removed all instances of such a fixture/parametrize being used, but could have missed one.  I imagine this fixture will be used in more places in the future when interval related tests are expanded.

### DIFF
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -137,6 +137,14 @@ def nselect_method(request):
     return request.param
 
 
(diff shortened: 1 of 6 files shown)


jreback:
thanks. its possible this *might* conflict with 'closed' as used in thinks like ``pd.date_range`` and ``resample`` which only have ``left/right`` but no big deal.

PR #21599, opened 2018-06-22
DOC: Change release and whatsnew

TomAugspurger:
Some cleanup & changes to facilitate release automation

* We will include the correct (latest on master or maintenance branch)
  whatsnew directly in the index.rst toctree
* Contributors are included in the whatsnew for each version (automatically)
* Removed release.rst
* Added new releases.rst which has toctrees for each release
screen shot 2018-06-22 at 3 10 23 pm

Incidental changes

* Updated style.ipynb. Writing the jinja template was confusing sphinx. We
  included it in the git source now.
* Fixing some inconsitent header levels (will do more)
* Refactored announce.py to support auto-generated contributors

TODO:

- [x] Finish up the rest of the whatsnews

cc @jorisvandenbossche, @datapythonista


doc/source/conf.py
```diff
@@ -120,7 +122,10 @@
 templates_path = ['../_templates']
 
 # The suffix of source filenames.
-source_suffix = '.rst'
+source_suffix = [
+    '.rst',
+    '.txt',
```
TomAugspurger: required so that the whatsnew files can be included in the toctree.

doc/source/conf.py
```diff
@@ -298,8 +303,39 @@
     for page in moved_api_pages
 }
 
+
+common_imports = """\
```
TomAugspurger: Previously we had this at the top of whatsnew.rst. Now it has to be in every `whatsnew/v0.x.x.txt`, as they are no longer literally included below this. Usage is just `{{ common_imports }}`

doc/source/conf.py
```diff
@@ -298,8 +303,39 @@
...
+
+    return announce.build_string('{}..{}'.format(start, end), None)
+
+
+contributors = {
+    'v0.24.0': announce.build_string('v0.23.1..HEAD'),
+    'v0.23.2': announce.build_string('v0.23.1..HEAD'),
+    'v0.23.1': announce.build_string('v0.23.0..v0.23.1'),
+    'v0.23.0': announce.build_string('v0.22.0..v0.23.0'),
+}
```
TomAugspurger: This is run every time we build the docs. It'll probably be run even for single page / docstring ones. I was concerned it would take a while, but it's surprisingly fast.

Once I finish the rest of the whatsnew, I'll time it again and look into ways of speeding it up if is a problem.

doc/source/conf.py
```diff
@@ -649,7 +685,23 @@ def process_class_docstrings(app, what, name, obj, options, lines):
 ]
 
 
+def rstjinja(app, docname, source):
+    """
+    Render our pages as a jinja template for fancy templating goodness.
+    """
+    # http://ericholscher.com/blog/2016/jul/25/integrating-jinja-rst-sphinx/
+    # Make sure we're outputting HTML
+    if app.builder.format != 'html':
```
TomAugspurger: Hmm I should look into what happens for PDF. We should be able to handle it just fine.

doc/source/index.rst.template
```diff
@@ -118,7 +118,7 @@ See the package overview for more detail about what's in the library.
     {{ single_doc }}
     {% endif -%}
     {% if not single_doc -%}
-    whatsnew
+    What's New <whatsnew/v0.24.0>
```
TomAugspurger: We'll update this as part of the post-release process.

datapythonista:
Looks really good. There is just one thing that I'd consider changing. I don't think it's a good practice to import functions from `scripts/`. Whn in the future someone make changes to the scripts, I don't think we will feel that we need to keep compatibility on any function defined there, as they can be imported by other parts of the code like the doc.

What I'd do instead is creating somewhere a directory "dev utils", where we can not only have the code of `build_string` but also the generation of contributors (and anything that could potentially be reused by the docs, the release, and may be the web). And if possible, I'd also automatically generate there the list of releases (0.23.0, 0.23.1...) so `releases.rst` and anything that requires the list of releases can be generated with the info from there (`releases.rst` would be a template as `index.rst` in that case).

doc/source/conf.py
```diff
@@ -298,8 +303,94 @@
...
+
+import re
+import git
+
+xpr = re.compile(r'^v(\d+)\.(\d+)\.(\d+)$')
+repo = git.Repo('..')
+tags = [x.name for x in repo.tags if xpr.match(str(x))]
+versions = sorted(zip(tags,
+                      [tuple(map(int, xpr.match(x).groups())) for x in tags]),
+                  key=lambda x: x[1])
```
datapythonista: In case you don't know it, `distutils` provides a class to parse, sort... versions. You'd need to get rid of the heading `v` in `v0.23.1`. But then you can do this stuff automatically. For example:

```
>>> import distutils.version
>>> tags = ['0.22.0', '0.23.0', '0.23.1', '0.21.1', '0.21.0']
>>> versions = map(distutils.version.StrictVersion, tags)
>>> versions = sorted(versions)
>>> versions[-1].version
(0, 23, 1)
```

TomAugspurger:
Sorry I didn't mean to push those changes. Not sure whether it's worth auto-discovering.

jorisvandenbossche:
Cool! Didn't have time to look into detail yet (will do that tomorrow), but some very quick feedback: 

- The docs are not yet building on travis: ``ModuleNotFoundError: No module named 'announce'``
- I get "/home/joris/scipy/pandas/doc/source/whatsnew/v0.20.3.txt: WARNING: document not readable. Ignored." for all whatsnew pages locally, but didn't further look into what is the cause (might be a problem on my side)
- Big +1 on having the separate whatsnew files to separate html pages (already wanted to propose that for some time). However, the only problem is that this will break quite some links?
- From a usability perspective of the "releases page", it would maybe be nice to include the highlights there, if this would be possible in some way.
- Generating all the `announce.build_string` takes a non-negligible amount of time (when building the full docs probably not significant, but for a single page it might get annoying). Do we want to keep the script to generate it, but then actually include the content in the actual files (and use the script to generate it on release time) instead of generating it dynamically? (or otherwise skip it in certain cases, eg when building a single page that is not the releases pages)

jorisvandenbossche:
> Looks really good. There is just one thing that I'd consider changing. I don't think it's a good practice to import functions from scripts/

\+ 1. We can maybe also put things in or alongside doc/make.py ?

doc/source/conf.py
jorisvandenbossche: Is there actually a reason they are not *.rst ?

TomAugspurger:
> alongside doc/make.py

Sure.

> this will break quite some links?

Which ones? Anything using the "canonical" format like `/version/x.y.z/` should be fine, since we aren't changing those.

> it would maybe be nice to include the highlights there, if this would be possible in some way.

Agreed. I think `highlights.rst` will be a stub or something that gets included in several places

> I would slightly prefer to generate it dynamically. My goal is to remove the "final doc update" commit that's currently part of the release process. In theory, it should be possible to make this a Jinja2 macro that is called inside the actual whatsnew, rather than ahead of time.

TomAugspurger:
I think it used to be to avoid warnings that they weren't included in a
toctree? Maybe we can just change them.

jorisvandenbossche:
> > this will break quite some links?

> Which ones? Anything using the "canonical" format like /version/x.y.z/ should be fine, since we aren't changing those.

Because it are now all separate pages, the url changes. For example, the first link in the highlights section of 0.23.0 points to http://pandas.pydata.org/pandas-docs/stable/whatsnew.html#whatsnew-0230-enhancements-round-trippable-json. However, with this change the link will end in .../whatsnew/v0.23.0.html#whatsnew-0230-enhancements-round-trippable-json

scripts/announce.py
```diff
@@ -98,27 +99,33 @@ def get_pull_requests(repo, revision_range):
...
     lst_release, cur_release = [r.strip() for r in revision_range.split('..')]
-
-    # document authors
     authors = get_authors(revision_range)
-    heading = u"Contributors"
-    print()
-    print(heading)
-    print(u"=" * len(heading))
-    print(author_msg % len(authors))
+    uline = '=' * len(heading)
```
jorisvandenbossche: this header does not always match the header levels in the whatsnew file

doc/source/conf.py
```diff
@@ -298,8 +303,94 @@
...
+    'v0.19.1': announce.build_string('v0.10.0..v0.19.1'),
+    'v0.19.2': announce.build_string('v0.10.0..v0.19.2'),
+    'v0.20.0': announce.build_string('v0.10.0..v0.20.0'),
+    'v0.20.1': announce.build_string('v0.10.0..v0.20.1'),
+    'v0.20.2': announce.build_string('v0.10.0..v0.20.2'),
+    'v0.20.3': announce.build_string('v0.10.0..v0.20.3'),
+    'v0.21.0': announce.build_string('v0.10.0..v0.21.0'),
+    'v0.21.1': announce.build_string('v0.10.0..v0.21.1'),
+    'v0.22.0': announce.build_string('v0.10.0..v0.22.0'),
+    'v0.22.0': announce.build_string('v0.22.0..v0.23.0'),
```
jorisvandenbossche: I suppose you know you still have some typos here (a lot of "v0.10.0" above, and this line is duplicated)

TomAugspurger:
Yeah, I'm removing all this. Didn't mean to push it.

I tried to write a new sphinx directive, so we can just do `..
contributors:: v0.23.1..v0.23.2`, but didn't get the rendering of the
inserted text to work yet.

TomAugspurger:
Ok, cleaned up my mess from yesterday. Replaced with a custom directive so we write

```
.. contributors:: v0.23.0..v0.23.1
```

rather than building them up ahead of time.

I haven't updated all the old files yet.

TomAugspurger:
Will think about the links as well.

Writing the directive took a long time (I didn't realize I had to build the list of elements myself. I was hoping I could just pass it a string and have sphinx parse it), so I'm not sure I'll have time to finish this for 0.23.2.

jorisvandenbossche:
> so I'm not sure I'll have time to finish this for 0.23.2.

no problem, we can still go the normal way for that

TomAugspurger:
> > > this will break quite some 

> > Which ones? Anything using the "canonical" format like /version/x.y.z/ should be fine, since we aren't changing those.

> Because it are now all separate pages, the url changes.

Hmm I'm not sure I see what will break.

For internal links, we're fine, since we use `:ref:`. The ones in the highlights will be fine.

For external links, we're hopefully using the canonical link with `/version/x.y.z/` wherever it matters (e.g. announce). Places that don't use the canonical link will break, but I think that's acceptable.

TomAugspurger:
And it looks like sharing the highlights between the release notes, announcement, and website will be difficult.

jorisvandenbossche:
> For external links, we're hopefully using the canonical link with /version/x.y.z/ wherever it matters (e.g. announce). Places that don't use the canonical link will break, but I think that's acceptable.

Did you see my comment above (https://github.com/pandas-dev/pandas/pull/21599#issuecomment-400223120)? 
It is not about the "base url" (/version/0.23.0/ vs /stable/ or so), but about the part that comes after it (now "whatsnew.html#internal-link-label", with this PR "whatsnew/v0.23.2.html#internal-link-label", even if we keep all "internal-link-label" the same, the actual url will change)

Anyhow, if we want to do a bit bigger docs reorganisation in the future as well, we will have to live with such breakages in any case ..

jreback:
can you rebase or close

jorisvandenbossche:
> I assume (hope) that people have the expectation that /stable is a snapshot, and the structure could change.

I am not sure that assumption is correct :-) (I think people often point to stable (or at least, I also do that), to make sure to always point to the latest version)

But that said, that doesn't mean we shouldn't do it. As I said above, if we want to do bigger reorganisation of the docs, we will have such issues anyway. 

I could look into autogenerating some redirects (but not right now).

Didn't look at the latest updates yet.

datapythonista (approved):
lgtm, just I think for v0.4 the contributors are missing.

Needs a rebase, but I'd merge this as soon as it's rebased. It's a good improvement, and I'd better have to fix something later, than having to keep fixing the conflicts (that will keep happening fast in the latest whatsnew file).

jreback:
@TomAugspurger can you rebase

TomAugspurger:
I was thinking a bit... for links like `http://pandas.pydata.org/pandas-docs/stable/whatsnew.html#whatsnew-0190-enhancements-other` I *think* we're completely out of luck for redirecting. nginx (our web server) doesn't see anchor tags (everything after the `#`) so we wouldn't be able to rewrite the url.

I'm going to rebase and I think we should merge. Every release we delay this is another releases's worth of URLs we'll break in the future if we ever end up doing this :)

TomAugspurger:
Since we renamed the whatsnew suffixes from .txt to .rst, they're getting linted now. I'm not planning to go through the old whatsnews to fix linting errors, but new ones will be linted (and 0.24.0 is clean right now).

jorisvandenbossche:
> I think we're completely out of luck for redirecting. nginx (our web server) doesn't see anchor tags (everything after the #) so we wouldn't be able to rewrite the url.

So you mean we can't "catch" the fact that a user is visiting eg `http://pandas.pydata.org/pandas-docs/stable/whatsnew.html#whatsnew-0190-enhancements-other` ?

Since the `whatsnew` page is still there, in principle we could add a bunch of generated tags there with a link for people to click on? 
Or with some javascript (https://stackoverflow.com/questions/1305211/javascript-to-redirect-from-anchor-to-a-separate-page) 

(but that does not need to hold up this PR)

TomAugspurger:
> So you mean we can't "catch" the fact that a user is visiting eg http://pandas.pydata.org/pandas-docs/stable/whatsnew.html#whatsnew-0190-enhancements-other ?

If we agree that "catch" means *before* the user actually gets there, then yes that's what I'm saying.

> Since the whatsnew page is still there, in principle we could add a bunch of generated tags there with a link for people to click on?

Sounds pretty ugly :)

A javascript-based solution seems like the best bet.

jorisvandenbossche:
> If we agree that "catch" means before the user actually gets there, then yes that's what I'm saying.

I am a complete web noob, so not sure what I meant :-)

TomAugspurger:
@datapythonista it'd be nice if warnings like https://travis-ci.org/pandas-dev/pandas/jobs/454525382#L2853

```
doc/source/whatsnew/v0.24.0.rst:- Bug in :func:`pandas.concat` when concatenating a multicolumn DataFrame with tz-aware data against a DataFrame with a different number of columns (:issue`22796`)
```

printed the source line number. Is that possible? Do we have an issue for that?

jorisvandenbossche:
"Check for backticks incorrectly rendering because of missing spaces" is actually also a strange description of the problem in this case?

TomAugspurger:
I think that one is a bug. It's really checking for things like :class:\`DataFrame\`s.

doc/source/whatsnew/v0.23.1.rst
```diff
@@ -1,7 +1,10 @@
 .. _whatsnew_0231:
 
-v0.23.1 (June 12, 2018)
------------------------
+********************
+What's New in 0.23.1
+********************
+
+{{ common_imports }}
```
h-vetinari: Missing the `**v0.23.1 (June 12, 2018)**` here

doc/source/whatsnew/v0.23.2.rst
```diff
@@ -1,7 +1,10 @@
 .. _whatsnew_0232:
 
-v0.23.2 (July 5, 2018)
-----------------------
+********************
+What's New in 0.23.2
+********************
+
+{{ common_imports }}
```
h-vetinari: Missing the `**v0.23.2 (July 5, 2018)**` here

doc/source/style.ipynb
```diff
@@ -1216,7 +1166,7 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python [default]",
+   "display_name": "Python 3",
    "language": "python",
    "name": "python3"
```
h-vetinari: When building the docs locally, I always have to rename the kernel to `pandas-dev` here. Not necessarily relevant for this PR, but it would be great if that could be modularized.

doc/source/whatsnew/v0.24.0.rst
```diff
@@ -8,6 +8,11 @@ v0.24.0 (Month XX, 2018)
    Starting January 1, 2019, pandas feature releases will support Python 3 only.
```
h-vetinari: Header format of v.0.24 not changed here? Then maybe the v.0.23-ones should stay in line with the style before/after v.0.23.

doc/source/whatsnew/v0.23.3.txt
```diff
@@ -1,7 +0,0 @@
-.. _whatsnew_0233:
-
-v0.23.3 (July 7, 2018)
-----------------------
```
h-vetinari: Not sure v.0.23.3 should be removed completely? Leaves a puzzling gap between v.0.23.2 and v.0.23.4 otherwise

TomAugspurger:
I think we have +1s here, so merging once the doc build passes.

doc/source/whatsnew/v0.23.0.rst
```diff
@@ -1,7 +1,13 @@
...
-v0.23.0 (May 15, 2018)
-----------------------
+
+********************
+What's New in 0.23.0
+********************
+
+{{ common_imports }}
+
+**v0.23.0 (May 15, 2018)**
```
h-vetinari: Changing the header formatting from `v.0.23` is very likely intended, but just pointing out that it is only used for `0.23.*` and not before or after. (also: not sure I'm the biggest fan with the redundancy of having the version twice)

TomAugspurger:
Updated the headers.

doc/source/whatsnew/v0.23.0.rst
h-vetinari: This header `**v0.23.0 (May 15, 2018)**` is still redundant

doc/source/whatsnew/v0.24.0.rst
h-vetinari: If the new style starting from `0.23` should be: 
```diff
- v0.23.0 (May 15, 2018)
- ----------------------
+ What's new in 0.23.0 (May 15, 2018)
+ -----------------------------------
```

then this should be continued with v.0.24 (or reverted for the 0.23.* ones).

doc/source/whatsnew/v0.20.2.rst
```diff
@@ -1,7 +1,7 @@
 .. _whatsnew_0202:
 
-v0.20.2 (June 4, 2017)
-----------------------
+What's new in v0.20.2 (June 4, 2017)
+------------------------------------
```
h-vetinari: Are you sure you wanted to do this in `v.0.20.2` (which further increases the inconsistency) rather than `v.0.24.0`?

h-vetinari (commented):
Looks like you inadvertently edited the wrong file

doc/source/whatsnew/v0.20.2.rst
h-vetinari: Alternatively, you could add "What's new in " to all versions (which would be even better than introducing it starting with 0.23.0)

TomAugspurger:
I think I needed to change the 0.23+ headers to get the levels consistent. Not super interested in changing the old ones.

doc/source/whatsnew/v0.23.0.rst
```diff
@@ -1,7 +1,9 @@
 .. _whatsnew_0230:
 
-v0.23.0 (May 15, 2018)
-----------------------
+What's new in 0.23.0 (May 15, 2018)
```
h-vetinari: Not a big deal, but "new" is capitalized for all the others. I'd go with lower-case for all.

doc/source/releases.rst
```diff
@@ -0,0 +1,203 @@
...
+*************
+Release Notes
+*************
+
+This is the list of changes to pandas between each release. For full details,
+see the commit logs at http://github.com/pandas-dev/pandas. For install and
+upgrade instructions, see :ref:`install`.
+
+0.24 release
+------------
```
h-vetinari: I think it would look better (also looking at your render from the OP) to have
```diff
+ Version 0.24
- 0.24 release
  ------------
```

etc.

h-vetinari (commented):
Some tiny style comments, but nothing blocking. Also, you seem to have closed the PR by accident.

> I think I needed to change the 0.23+ headers to get the levels consistent. Not super interested in changing the old ones.

Fair enough.

TomAugspurger:
All green. Merging before we have more conflicts :)

I think this will require all outstanding PRs adding a release note to merge master. But contributors shouldn't have to manually fix any conflicts; git will handle things automatically.

TomAugspurger:
Fixed (avoided) in https://github.com/pandas-dev/pandas/pull/23714.

I think you're right about the clone depth. Not a huge deal for the CI,
though I'm a bit concerned things
will be broken when we get to release time, and we'll need to scramble last
minute to fix it.

jorisvandenbossche:
@TomAugspurger Thanks!

It would still be nice to have it run at the CI as well, as I would imagine that at some point we would like to rely on some CI service to build the docs and upload that version, instead of having the release manager it to do locally?

TomAugspurger:
Yeah, I'm not sure what's best here...

For now, we could hard code a `git fetch origin
refs/tags/v0.23.4:refs/tags/v0.23.4` before doing the doc build.
But then that's one more thing to remember to update manually, which was
the entire point of this PR :)

PR #21602, opened 2018-06-23
TST: Use multiple instances of parametrize instead of product in tests

jschendel:
- All instances `product` being used inside a `@pytest.mark.parametrize` have been converted to separate instances of `@pytest.mark.parametrize`
- Extracted `product` from within tests where it was straightforward
- There are still multiple instances of `product` being used within tests for looping purporses
   - Most of the ones left do not appear to be trivial to extract
   - Some instances we might not want to remove/are clearer as-is
   - Could maybe open an issue for some of the more obvious ones?

### DIFF
--- a/pandas/tests/frame/test_rank.py
+++ b/pandas/tests/frame/test_rank.py
@@ -10,7 +10,6 @@ from numpy import nan
 from pandas.util.testing import assert_frame_equal
 from pandas.tests.frame.common import TestData
 from pandas import Series, DataFrame
-from pandas.compat import product
 
 
 class TestRank(TestData):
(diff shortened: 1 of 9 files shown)


pandas/tests/frame/test_rank.py
```diff
@@ -257,43 +264,38 @@ def _check2d(df, expected, method='average', axis=0):
...
-                     [1., 2. / 3, 1. / 3]]),
-                   ("average",
-                    [[2. / 3, 1., 1.],
-                     [2. / 3, 0.5, 2. / 3],
-                     [2. / 3, 0.5, 1. / 3]]),
-                   ("first",
-                    [[1. / 3, 1., 1.],
-                     [2. / 3, 1. / 3, 2. / 3],
-                     [3. / 3, 2. / 3, 1. / 3]])])
-def test_rank_pct_true(method, exp):
```
jschendel: This test was the only one in that was outside the class, which seemed a bit strange, so I moved it inside the class.

pandas/tests/groupby/test_whitelist.py
```diff
@@ -8,7 +8,6 @@
 import numpy as np
 from pandas import DataFrame, Series, compat, date_range, Index, MultiIndex
 from pandas.util import testing as tm
-from pandas.compat import lrange, product
 
 AGG_FUNCTIONS = ['sum', 'prod', 'min', 'max', 'median', 'mean', 'skew',
                  'mad', 'std', 'var', 'sem']
```
jreback: this should be a fixture (future PR)

jreback (approved):
lgtm merge on green

PR #21582, opened 2018-06-21
cache DateOffset attrs now that they are immutable

jbrockmendel:
TL;DR ~6x speedup in `set_index` for `PeriodIndex`-like column.

Alright! Now that DateOffset objects are immutable (#21341), we can can start caching stuff.  This was pretty much the original motivation that brought me here, so I'm pretty psyched to finally make this happen.

The motivating super-slow operation is `df.set_index`.  Profiling before/after with:
```
idx = pd.period_range('May 1973', freq='M', periods=10**5)
df = pd.DataFrame({"A": 1, "B": idx})
out = df.set_index("B", append=True)
```

Total Runtime Before: 32.708 seconds
Total Runtime After: 5.340 seconds

pstats output (truncated) before:
```
        1    0.000    0.000   31.903   31.903 pandas/core/frame.py:3807(set_index)
        1    0.000    0.000   31.897   31.897 pandas/core/indexes/base.py:4823(_ensure_index_from_sequences)
        1    0.000    0.000   31.896   31.896 pandas/core/indexes/multi.py:1246(from_arrays)
        1    0.000    0.000   31.896   31.896 pandas/core/arrays/categorical.py:2590(_factorize_from_iterables)
        2    0.001    0.000   31.896   15.948 pandas/core/arrays/categorical.py:2553(_factorize_from_iterable)
        2    0.000    0.000   31.895   15.948 pandas/core/arrays/categorical.py:318(__init__)
        2    0.000    0.000   31.512   15.756 pandas/util/_decorators.py:136(wrapper)
        2    0.002    0.001   31.512   15.756 pandas/core/algorithms.py:576(factorize)
  1600011    1.168    0.000   28.211    0.000 pandas/tseries/offsets.py:338(__ne__)
        4    1.820    0.455   28.042    7.010 {method 'argsort' of 'numpy.ndarray' objects}
  1600011    4.016    0.000   27.042    0.000 pandas/tseries/offsets.py:324(__eq__)
  3200022   16.987    0.000   21.856    0.000 pandas/tseries/offsets.py:291(_params)
        2    0.000    0.000    3.460    1.730 pandas/core/algorithms.py:449(_factorize_array)
        1    0.617    0.617    3.445    3.445 {method 'get_labels' of 'pandas._libs.hashtable.PyObjectHashTable' objects}
  3200023    3.200    0.000    3.200    0.000 {sorted}
3400729/3400727    1.235    0.000    1.235    0.000 {isinstance}
[...]

PR #21606, opened 2018-06-23
More speedups for Period comparisons

jbrockmendel:
Following #21582, the biggest avoidable overheads in Period comparisons are in a) `isinstance` calls that can be short-circuited and b) `__ne__` call overhead.  This PR moves `__eq__` and `__ne__` to the cython file, removing the `__ne__` overhead, and changes an `isinstance` check to a try/except.

Using the same profile code from #21582, this brings the `set_index` runtime from 6.603 seconds down to 2.165 seconds.


pandas/_libs/tslibs/offsets.pyx
```diff
@@ -313,6 +313,30 @@ class _BaseOffset(object):
     def __setattr__(self, name, value):
         raise AttributeError("DateOffset objects are immutable.")
 
+    def __eq__(self, other):
+        if is_string_object(other):
+            from pandas.tseries.frequencies import to_offset
```
jreback: maybe you want to define a function to return this import, we use this in a number of places?

pandas/_libs/tslibs/offsets.pyx
```diff
@@ -313,6 +313,30 @@ class _BaseOffset(object):
...
+
+    def __ne__(self, other):
+        return not self == other
+
+    def __hash__(self):
+        return hash(self._params)
+
+    @property
+    def _params(self):
+        from pandas.errors import AbstractMethodError
```
jreback: can you import this at the top?

jreback (changes requested):
add this issue to your existing perf issue (always do that on follow ons)

pandas/_libs/tslibs/offsets.pyx
jbrockmendel: Wouldn't this just be trading one one-liner for another?

jreback: no then then its a function which we can import at the top (and of course if we eventually are able to actually move to_offset to cython its a big win). but its 'cleaner' I think. (certainly can followup on this in new PR). I just don't really like the imports hidden in the functions (as we must do currently), and rather centrailize in a single place.

pandas/tseries/offsets.py
```diff
@@ -182,6 +182,8 @@ def __add__(date):
 
     Since 0 is a bit weird, we suggest avoiding its use.
     """
+    _params = cache_readonly(BaseOffset._params.fget)
+
```
jreback: can you remove this blank line

pandas/_libs/tslibs/offsets.pyx
```diff
@@ -313,6 +313,37 @@ class _BaseOffset(object):
...
+
+    def __ne__(self, other):
+        return not self == other
+
+    def __hash__(self):
+        return hash(self._params)
+
+    @property  # NB: non-cython subclasses override with cache_readonly
+    def _params(self):
+        all_paras = self.__dict__.copy()
```
jreback: can you give an expl of what this is doing (and you can move the comment inside the function too)

jreback (changes requested):
just a minor comment. ping on green.

jreback (approved):

Issue #18262, opened 2017-11-13
DEPR: let's deprecate


jreback:
xref #18202 

We have some cruft, let's deprecate it (I have noted some which already have an issue associated).

### From ndarray

- [x] `Series.compress()` (#21930)
- [x] `Series.imag` / `Series.real` (#27106) 
- [x] `Series.nonzero()` (#24048) 
- [x] `Series.put()` (#27106)
- [x] `Series.itemsize`
- [x] `Series.flags`
- [x] `Series.strides`

### timeseries specific

- [ ] `Series.first()`
    let head / tail take a timedelta
- [ ] `Series.last()`

### other

- [ ] `Series.swapaxes()`
  - No reason once panel is gone.

### non-controversial
- [x] ``MultiIndex.to_hierarchical`` (only used for ``Panel``) (#21613)
- [x] ``Series/DataFrame.compound()`` (#26405)
- [x] ``Series.ptp()`` (#21614)
- [x] ``Series.from_array`` (#18213)
- [x] ``Series.valid()`` (#18800)
- [x] ``Series/DataFrame.slice_shift()``  (#37601)
- [x] ``Series/DataFrame.tshift()`` (https://github.com/pandas-dev/pandas/pull/34545)
- [x] ``Series/DataFrame.get_values()`` (https://github.com/pandas-dev/pandas/issues/19617)
- [x] ``Index.dtype_str`` (#27106)
- [x] ``Index.summary()`` (#18217)
- [x] ``.get_ftype_counts`` (#18243) (#20404)
- [x] ``.get_dtype_counts`` #27145 
- [x] ``Index/Series.asobject`` (#18237) (#18572)
- [x] ``Index.to_native_types()`` (make private) (https://github.com/pandas-dev/pandas/pull/36418)
- [x] ``DataFrame/Series.as_matrix`` (#18458)
- [x] ``.clip_upper/.clip_lower`` (replace by ``.clip``) (#24203)

### Potentially
- [x] ``.ftypes`` (#18243) (#26744) 
- [ ] ``.xs()`` (#6249)
- [ ] ``.iat/.at``
- [ ] ``.take``
- [x] ``.lookup`` (#35224)
- [x] ``DataFrame.from_items``
- [ ] ~``Series/DataFrame.add_prefix/add_suffix``~ (https://github.com/pandas-dev/pandas/pull/18347)
  - Maybe add suffix / prefix to concat?
- [ ] ``NDFrame.filter``

[...]

PR #18258, opened 2017-11-13
DEPR: deprecate (Sparse)Series.from_array

jorisvandenbossche:
Closes #18213. 

It's a bit annoying that I have to keep it as internal method. On the other hand this might signal that `from_array` actually has some use case ..

PR #21613, opened 2018-06-24
DEPR: MultiIndex.to_hierarchical

KalyanGokhale:
xref #18262
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/pandas/core/indexes/multi.py
+++ b/pandas/core/indexes/multi.py
@@ -189,7 +189,6 @@ class MultiIndex(Index):
     from_product
     set_levels
     set_labels
-    to_hierarchical
     to_frame
     is_lexsorted
     sortlevel
--- a/pandas/core/indexes/multi.py
+++ b/pandas/core/indexes/multi.py
@@ -1182,6 +1182,8 @@ class MultiIndex(Index):
 
     def to_hierarchical(self, n_repeat, n_shuffle=1):
         """
+        DEPRECATED: to_hierarchical will be removed in a future version.
+
         Return a MultiIndex reshaped to conform to the
         shapes given by n_repeat and n_shuffle.
 
@@ -1216,6 +1218,9 @@ class MultiIndex(Index):
         # Assumes that each label is divisible by n_shuffle
         labels = [x.reshape(n_shuffle, -1).ravel(order='F') for x in labels]
         names = self.names
+        warnings.warn("Method .to_hierarchical is deprecated and will "
+                      "be removed in a future version",
+                      FutureWarning, stacklevel=2)
         return MultiIndex(levels=levels, labels=labels, names=names)
 
     @property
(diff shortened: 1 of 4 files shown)


pandas/tests/indexes/test_multi.py
```diff
@@ -1704,6 +1704,11 @@ def test_to_hierarchical(self):
         tm.assert_index_equal(result, expected)
         assert result.names == index.names
 
+        # GH21613
+        # .to_hierarchical will be deprecated
+        with tm.assert_produces_warning(FutureWarning):
+            result = index.to_hierarchical(2)
```
jreback: you don't need the result =

pandas/tests/indexes/test_multi.py
```diff
@@ -1704,6 +1704,11 @@ def test_to_hierarchical(self):
         tm.assert_index_equal(result, expected)
```
jreback: you need to catch all uses of to_hierarchical (or you get the warnings)

pandas/core/indexes/multi.py
```diff
@@ -1182,6 +1182,8 @@ def to_frame(self, index=True):
 
     def to_hierarchical(self, n_repeat, n_shuffle=1):
         """
```
jreback: to_hierarchical is listed elsewhere in this file in a doc-string

jreback (changes requested):

pandas/core/indexes/multi.py
KalyanGokhale: Thanks - it is mentioned as a part of enumeration of methods of `MultiIndex` - should it be removed?

pandas/tests/indexes/test_multi.py
KalyanGokhale: Thanks - not sure if I understand. 
The only use of `to_hierarchical` was in `pandas/core/panel.py`- this has been substituted. 
This existing test is specifically for `to_hierarchical` - which can be removed once this method is removed in a future version - should these tests be removed as part of this PR?

pandas/tests/indexes/test_multi.py
KalyanGokhale: Thanks - yes had missed it - will update together along with other comments

pandas/tests/indexes/test_multi.py
KalyanGokhale: @jreback I think after reviewing changes in #18258 I probably get what is expected - please ignore the comment above - will update

jreback (changes requested):

pandas/core/indexes/multi.py
KalyanGokhale: Thanks - updated for all of the latest review comments

jreback (approved):

jreback:
thanks!

PR #21618, opened 2018-06-24
CLN: make CategoricalIndex._create_categorical a classmethod

topper-123:
Currently, ``CategoricalIndex._create_categorical`` is a staticmethod, and is being called internally using *either* instances or classes as its first argument, e.g.:
* in ``_is_dtype_compat`` *an instance* is supplied as the first argument,
* in ``__new__`` *a class* is supplied as the first argument.

This is confusing and makes the code paths different depending on how the method is called. It makes it difficult to reason about the precise output of the method.

This PR cleans this up by making ``_create_categorical`` a classmethod. This simplifies stuff, and we can also remove the method's first parameter when calling it.

Calling ``_create_categorical`` unneccesarily is one reason for the slowness of #20395. After this cleanup PR I will do another that should get #20395 come down to 1.6 ms as well as give some other related performance improvements.

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -1130,7 +1130,8 @@ class Index(IndexOpsMixin, PandasObject):
         """
 
         from pandas import DataFrame
-        result = DataFrame(self._shallow_copy(), columns=[self.name or 0])
+        name = self.name or 0
+        result = DataFrame({name: self.values.copy()})
 
         if index:
             result.index = self
--- a/pandas/core/indexes/category.py
+++ b/pandas/core/indexes/category.py
@@ -85,11 +85,11 @@ class CategoricalIndex(Index, accessor.PandasDelegate):
             name = data.name
 
         if isinstance(data, ABCCategorical):
-            data = cls._create_categorical(cls, data, categories, ordered,
+            data = cls._create_categorical(data, categories, ordered,
                                            dtype)
         elif isinstance(data, CategoricalIndex):
             data = data._data
-            data = cls._create_categorical(cls, data, categories, ordered,
+            data = cls._create_categorical(data, categories, ordered,
                                            dtype)
         else:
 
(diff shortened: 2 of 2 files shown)


pandas/core/indexes/base.py
```diff
@@ -1130,7 +1130,8 @@ def to_frame(self, index=True):
         """
 
         from pandas import DataFrame
-        result = DataFrame(self._shallow_copy(), columns=[self.name or 0])
```
topper-123: _shallow_copy now does not guarantee that _data is copied, so a test fails. ``to_frame`` should get a copy, else smething like this could happen:

```python
>>> ci = pd.CategoricalIndex(['a', 'b', 'c'])
>>> df = ci.to_frame()
>>> df.iloc[0] = 'c'
>>> df
   0
c  c  # index label changed!
b  b
c  c
```

By using ``self.values.copy()`` rather than ``_shallow_copy``, this issue is avoided.

doc/source/whatsnew/v0.23.2.txt
```diff
@@ -25,8 +25,8 @@ Fixed Regressions
 Performance Improvements
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-- Improved performance of membership checks in :class:`CategoricalIndex`
-  (i.e. ``x in ci``-style checks are much faster). :meth:`CategoricalIndex.contains`
+- Improved performance of membership checks in :class:`CategoricalIndex` and :class:`Categorical`
+  (i.e. ``x in cat``-style checks are much faster). :meth:`CategoricalIndex.contains`
```
topper-123: This is just a cleanup of the whatsnew for #21508, not related to this PR.

jreback: put this in a separate PR then.

jreback (changes requested):
this is fine, pls do the 0.23.2 change separately though.

topper-123:
Comments addressed & green

jreback (approved):

PR #21623, opened 2018-06-25
DOC: Do no use 'type' as first word when specifying a return type (#21622)

### LINKED ISSUE #21622: Pycharm is confused by 'type' in docstrings
#### Problem description
In quite some places the return type of functions is specified like this:
```
Returns
-------
swapped : type of caller (new object)
```
Pycharm now thinks the function returns an object of type `type` and warns about legitimate usage.

See #17079 for a related issue.

The problem could be mitigated by a wording avoiding any actual type as the first word.
Proposal how the change could look like: #21623

aberres:
Pycharm is otherwise confused and expects objects of type
'type' to be returned.

- [x] closes #21622 
- [x] tests added / passed (no code changes)
- [x] whatsnew entry (not needed I guess)

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4675,7 +4675,7 @@ class DataFrame(NDFrame):
 
         Returns
         -------
-        swapped : type of caller (new object)
+        swapped : same type as caller (new object)
 
         .. versionchanged:: 0.18.1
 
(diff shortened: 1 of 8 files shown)


WillAyd (approved):

jreback:
thanks @aberres

PR #21612, opened 2018-06-23
TST: Clean old timezone issues PT2

mroeschke:
- [x] closes #11679
- [x] closes #11365
- [x] closes #12064
- [x] closes #12601
- [x] closes #9243
- [x] closes #11792
- [x] closes #13051
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

xref #21491, cleaning up older timezone issues and moved the whatsnew entries in the first cleanup PR from v0.23.2 to v0.24.0 as discussed https://github.com/pandas-dev/pandas/pull/21491#discussion_r197077733

PR #21631, opened 2018-06-26
Move tz cleanup whatsnew entries to v0.24

mroeschke:
Pre-req for #21612. https://github.com/pandas-dev/pandas/pull/21612#pullrequestreview-131551269

Moving tz cleanup whatsnew entries from 0.23.2 to 0.24.0 added in #21491

cc @jreback


jorisvandenbossche:
Thanks!

PR #21514, opened 2018-06-17
split up pandas/tests/indexes/test_multi.py #18644

elmq0022:
- [x] closes #18644
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

The file is split out, the number of tests remains 377 (2 skips on my machine; all other test cases pass), the code uses fixtures in all instances, new fixtures are in a local conftest.py file, and there are no dependencies on classes.  

I'm happy to receive and address any feedback.

PR #21645, opened 2018-06-26
ENH: add return_inverse to duplicated for DataFrame/Series/Index/MultiIndex

### LINKED ISSUE #21357: ENH: adding .unique() to DF (or return_inverse for duplicated)
When deduplicating, I need to get the mapping from the deduplicated records back to the originals.

Unfortunately:
* `drop_duplicates` and `duplicated` provide no information about this mapping
* `np.unique` has a `return_inverse`-parameter, but does not work with `object` data due to numpy/numpy#641

Let's say I have a large DF (`df`) with several object columns (say, `sig`) that are the deduplication signature (`subset`-parameter of `duplicated`). 

Then I have (here only for a subset of 100k records):
```
sig = ['list', 'of', 'column', 'names']
N = 100000
tic = timeit.default_timer()
### no reverse mapping
df[sig].sample(N).duplicated(keep=False)
toc = timeit.default_timer()
print(f'duplicated: {toc-tic:.2f} sec.')

tic = timeit.default_timer()
### no reverse mapping
pd.unique([x for x in df[sig].sample(N).itertuples(name='', index=False)])
toc = timeit.default_timer()
print(f'pd.unique: {toc-tic:.2f} sec.')

tic = timeit.default_timer()
### this creates the reverse mapping
dups = df.sample(N).groupby(sig).groups.values() # indices per group
redup = pd.concat([pd.Series(x[0], index = x) for x in dups], axis=0) # assume sorted index; otherwise "x.min()"
toc = timeit.default_timer()
print(f'with groupby: {toc-tic:.2f} sec.')
```
resulting in:
``` 
duplicated: 2.45 sec.
pd.unique: 1.64 sec.
with groupby: 16.89 sec.
```

The performance gets worse with size, so for a few millions, I need to wait ~20min to get the inverse, whereas a pure `duplicated` call only takes around 10 sec.

This caused me to hunt down the inner workings of `duplicated`, to see where the inverse gets lost. Turns out until almost the very end of `duplicated`, the indices are there: https://github.com/pandas-dev/pandas/blob/3147a86e1b20571766b488a8444c74cef29729ad/pandas/core/frame.py#L4384

I extracted and applied the guts of `duplicated` and once having `ids` for the whole set (~10sec), I have the following:
```
[calculate ids]
tic = timeit.default_timer()
### no reverse mapping
duplicated_int64(ids, keep=False)
toc = timeit.default_timer()
print(f'pd.duplicated_int64: {toc-tic:.2f} sec.')

tic = timeit.default_timer()
### no reverse mapping
np.unique(ids)
toc = timeit.default_timer()
print(f'np.unique, no inverse: {toc-tic:.2f} sec.')

tic = timeit.default_timer()
### this creates the reverse mapping
np.unique(ids, return_inverse=True)
toc = timeit.default_timer()
print(f'np.unique, with inverse: {toc-tic:.2f} sec.')
```
yielding
```
pd.duplicate_int64: 1.60 sec.
np.unique, no inverse: 0.31 sec.
np.unique, with inverse: 0.64 sec.
```

So, by exposing a `return_inverse` argument -- either in `.duplicated`, or in a separate `.unique` function for DF -- this would accelerate my problem from 20min to around 11sec.

I wouldn't know how to adapt the cython code for `duplicate_int64`, but since it's 5x slower than `np.unique` anyway, there's no point, IMHO. I *would* know how to call the numpy function at the appropriate time...

h-vetinari:
Closes #21357 

I followed the recommendation of @jreback there to change `duplicated`, and to make the output a tuple if `return_inverse=True`.

I didn't find a test for the behavior of `DataFrame.duplicated()` -- there's a test of `Index.duplicated()` and `Series.duplicated()` in `tests/test_base.py`, but nothing I could find in `tests/frame/`.

So I added a file to test to test current behavior, as well as the new kwarg. Notably, the backend between the different Series in the output tuple is different (`pandas._libs.hashtable.duplicated_int64` vs. `np.unique`), and so I also added several tests that these results are exactly compatible.


jschendel:
Can you add a whatsnew entry to 0.24.0?

pandas/core/frame.py
```diff
@@ -4354,15 +4354,24 @@ def duplicated(self, subset=None, keep='first'):
               first occurrence.
             - ``last`` : Mark duplicates as ``True`` except for the
               last occurrence.
-            - False : Mark all duplicates as ``True``.
+            - False : Mark all duplicates as ``True``. This option is not
+              compatible with ``return_inverse``.
+        return_inverse boolean, default False
+            Determines whether the mapping from unique elements to the original
+            index should be returned. If true, the output is a tuple.
```
jschendel: Can you add a `.. versionadded:: 0.24.0` to this?

pandas/tests/frame/test_duplicates.py
```diff
@@ -0,0 +1,106 @@
...
+import pytest
+
+import numpy as np
+from pandas import Series, DataFrame
+
+from pandas.util.testing import assert_series_equal, assert_frame_equal
+import pandas.util.testing as tm
+
+
+class TestDataFrameDuplicated(object):
```
jschendel: I think these tests should be in `tests/frame/test_analytics.py`.  It looks like there are a couple bug `DataFrame.duplicated` tests there, though nothing expansive, e.g.

https://github.com/pandas-dev/pandas/blob/4c4579d5629d7aed1719dbc8aaa423681b343772/pandas/tests/frame/test_analytics.py#L1634-L1649

The `Series.duplicated` tests are also in the associated `series/tests_analytics.py` file, so makes sense to mirror that setup.

pandas/tests/frame/test_duplicates.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def test_duplicated_keep(self):
+        df = DataFrame({'A': [0, 1, 1, 2, 0], 'B': ['a', 'b', 'b', 'c', 'a']})
+
+        # keep = 'first'
+        exp = Series([False, False, True, False, True])
+        assert_series_equal(df.duplicated(keep='first'), exp)
```
jschendel: My preference would be to use a slightly more explicit pattern throughout these tests along the lines of:
```python
expected = ...
result = ...
assert_series_equal(expected, result)
```

WillAyd (changes requested):
Thanks for the PR

pandas/tests/frame/test_duplicates.py
```diff
@@ -0,0 +1,106 @@
...
+        # keep = 'last'
+        exp = Series([True, True, False, False, False])
+        assert_series_equal(df.duplicated(keep='last'), exp)
+
+        # keep = False
+        exp = Series([True, True, True, False, True])
+        assert_series_equal(df.duplicated(keep=False), exp)
+
+    def test_duplicated_nan_none(self):
+        # np.nan and None are considered equal
```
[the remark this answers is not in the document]
h-vetinari: This is how it works currently - stuff like this becomes visible when writing tests. ;-)

Of course someone can open a separate issue on this.

pandas/core/frame.py
```diff
@@ -4354,15 +4354,24 @@ def duplicated(self, subset=None, keep='first'):
...
 
         Returns
         -------
-        duplicated : Series
+        duplicated : Series or tuple of Series if return_inverse is True
```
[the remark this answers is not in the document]
h-vetinari: It's absolutely essential that it's a Series (otherwise it'd have to be two arrays at least), because - other than in numpy - we also need to keep track of the original index.

So, the second Series here links two arrays - which original index gets mapped to which index in the deduplicated one. This is how the reconstruction can succeed:
``` 
df == unique.reindex(inv.values).set_index(inv.index)
```

h-vetinari:
@jschendel @WillAyd 
Thanks for the feedback, working to incorporate all the comments.

Another question that I realised while writing the whatsnew entry - should this kwarg be made available to Series/Index as well? In this case I'm guessing I'd have to adapt `core.base.IndexOpsMixin.duplicated`?

h-vetinari:
@jschendel Re:
> Also, looks like we have asvs for DataFrame.duplicated. Might be nice to add methods for return_inverse=True to these. Not sure if this is strictly necessary though.

Haven't ever worked with the asv's yet - will have a look later.

h-vetinari:
I added three runs in the ASV, here are the results against current master. I don't know why `frame_methods.Duplicated.time_frame_duplicated` failed on master, that shouldn't be the case - all the others are explained by the missing `return_inverse` kwarg on master.

```
 [a620e725]       [d3b5d0a2]
       failed            234ms      n/a  frame_methods.Duplicated.time_frame_duplicated
       78.1ms           78.1ms     1.00  frame_methods.Duplicated.time_frame_duplicated_mixed
       failed            125ms      n/a  frame_methods.Duplicated.time_frame_duplicated_mixed_inverse
       failed            125ms      n/a  frame_methods.Duplicated.time_frame_duplicated_mixed_inverse_last
        188ms            203ms     1.08  frame_methods.Duplicated.time_frame_duplicated_wide
       failed            219ms      n/a  frame_methods.Duplicated.time_frame_duplicated_wide_inverse
```

asv_bench/benchmarks/frame_methods.py
```diff
@@ -420,14 +420,33 @@ def setup(self):
...
         self.df.duplicated()
 
     def time_frame_duplicated_wide(self):
         self.df2.duplicated()
 
+    def time_frame_duplicated_wide_inverse(self):
+        # will take fastpath for no duplicates
+        self.df2.duplicated(return_inverse=True)
+
+    def time_frame_duplicated_mixed(self):
```
mroeschke: Instead of the 3 separate benchmarks below, you can parameterize over the `return_inverse` and `keep` keyword arguments. 

Here's an example: https://github.com/h-vetinari/pandas/blob/094139bffa07c101882bbfbf35a1621eadc00adc/asv_bench/benchmarks/frame_methods.py#L219-L231

h-vetinari: Thanks for the hint, coming in the commit for the next feedback round.

h-vetinari: Is it normal that the times are then not split up depending on the parameter?

I get:
```
[...]
[ 16.67%] ··· Running frame_methods.Duplicated.time_frame_duplicated         469ms;...
[ 33.33%] ··· Running frame_methods.Duplicated.time_frame_duplicated_mixed   125ms;...
[ 50.00%] ··· Running frame_methods.Duplicated.time_frame_duplicated_wide    203ms;...
[...]
```

mroeschke: I believe the semicolons are splitting the time depending on the parameter. The format usually looks better with `asv continuous`

h-vetinari: @mroeschke The output I posted _is_ an extract from `asv continuous`, but I hadn't looked into `asv compare yet`, where the results were hiding. :)

h-vetinari:
Test failure in circleci is unrelated. Failed in `TestPythonParser.test_dtype_with_converter` with 
```
E           AssertionError: Caused unexpected warning(s): ['ResourceWarning'].
```

asv_bench/benchmarks/frame_methods.py
h-vetinari: However, I get no outputs anymore for `master` (tried again three times...) -- could it be that a single parameter failing (like `return_inverse=True` will do on master) fails the whole test?

h-vetinari:
@mroeschke 
I got the results now, although I had to add a temporary if-condition (otherwise `master` would stumble over the unknown kwarg). Will remove that condition again, but I just wanted to have the same hash show up in the comparison as well as the commit I just pushed. Here are the results:

```
   before      after       ratio
 [a620e725]  [36b03a1c]
      203ms       203ms     1.00  frame_methods.Duplicated.time_frame_duplicated('first', False)
     failed       578ms      n/a  frame_methods.Duplicated.time_frame_duplicated('first', True)
+     203ms       250ms     1.23  frame_methods.Duplicated.time_frame_duplicated('last', False)
     failed       438ms      n/a  frame_methods.Duplicated.time_frame_duplicated('last', True)
     78.1ms      78.1ms     1.00  frame_methods.Duplicated.time_frame_duplicated_mixed('first', False)
     failed       125ms      n/a  frame_methods.Duplicated.time_frame_duplicated_mixed('first', True)
     78.1ms      78.1ms     1.00  frame_methods.Duplicated.time_frame_duplicated_mixed('last', False)
     failed       125ms      n/a  frame_methods.Duplicated.time_frame_duplicated_mixed('last', True)
      219ms       203ms     0.93  frame_methods.Duplicated.time_frame_duplicated_wide('first', False)
     failed       203ms      n/a  frame_methods.Duplicated.time_frame_duplicated_wide('first', True)
      203ms       203ms     1.00  frame_methods.Duplicated.time_frame_duplicated_wide('last', False)
     failed       203ms      n/a  frame_methods.Duplicated.time_frame_duplicated_wide('last', True)
```

To me, the benchmarks seem a bit unstable, otherwise I don't know where the performance differences come from.

doc/source/whatsnew/v0.23.1.txt
```diff
@@ -1,7 +1,7 @@
 .. _whatsnew_0231:
 
-v0.23.1
--------
+v0.23.1 (June 12, 2018)
```
jreback: don't modify things, separate PR if you would like to change this

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -8,12 +8,45 @@ v0.24.0
 New features
 ~~~~~~~~~~~~
 
-- ``ExcelWriter`` now accepts ``mode`` as a keyword argument, enabling append to existing workbooks when using the ``openpyxl`` engine (:issue:`3441`)
+.. _whatsnew_0240.enhancements.duplicated_inverse:
+
+``DataFrame.duplicated`` has gained the ``return_inverse`` kwarg
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Previously, there was no way to determine how duplicate rows in a ``DataFrame`` got mapped to the deduplicated, unique subset. This made it hard to push back
```
jreback: too much explanation. make this a sentence or (all of the text). you don't need to explain why

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -8,12 +8,45 @@ v0.24.0
...
+Therefore, the ``duplicated``-method has now gained a ``return_inverse`` keyword. Specifying ``return_inverse=True`` will change the output from a single Series
+to a tuple of two Series (in the following example, the index is not just a simple order, to illustrate that the inverse correctly takes it into account):
+
+.. ipython:: python
+
+    df = pd.DataFrame({'A': [0, 1, 1, 2, 0], 'B': ['a', 'b', 'b', 'c', 'a']},
+                      index=[1, 4, 9, 16, 25])
+    df
+    isdup, inv = df.duplicated(return_inverse=True)  # default: keep='first'
+    isdup
```
jreback: spell out names

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -8,12 +8,45 @@ v0.24.0
...
+    inv
+
+This allows to reconstruct the original DataFrame as follows:
+
+.. ipython:: python
+
+    unique = df.loc[~isdup]  # same as df.drop_duplicates()
```
jreback: this example is confusing because you are using different code here

jreback (changes requested):
still have yet to review the actual code

doc/source/whatsnew/v0.24.0.txt
h-vetinari: Not sure I understand. It refers to the same objects as in the code block above - it's just a continuation.

h-vetinari:
I finally finished adding all the tests, turns out it's quite complicated to do anything for `Index`, haha. ;-)

The core algorithm changes almost vanish compared to the amount of test changes. Some of the things I added tests for were not really tested so far (bare `duplicated` was only tested implicitly - and then only sometimes - for stuff like `drop_duplicates`). I also split up `pandas/tests/indexes/multi/test_unique_and_duplicates.test_duplicated`, which had a large amount of tests in one.

ASVs should be in line with what I posted above, because as long as `return_inverse=False`, I never calculate anything additional.

I needed to implement a different signature for `pandas.core.algorithms.duplicated`, which is inspired by `np.unique`, in the sense that it has both `return_index` and `return_inverse` (with the same meanings). This method is not public as far as I can tell? In any case, the docstring reflects the changes.
Finally, I added a `stabilize`-kwarg, which *stabilises* against the implicit sorting that numpy does. It's on by default, but turning it off is a performance improvement in case the index/inverse are immediately plugged together again (as is the case for `Series`/`DataFrame`).

Tried to explain all the code as well as possible, but happy to receive feedback (and/or answer questions).

pandas/tests/frame/test_duplicates.py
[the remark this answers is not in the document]
h-vetinari: Opened #21720

pandas/core/algorithms.py
```diff
@@ -786,16 +787,84 @@ def duplicated(values, keep='first'):
           occurrence.
         - ``last`` : Mark duplicates as ``True`` except for the last
           occurrence.
-        - False : Mark all duplicates as ``True``.
+        - False : Mark all duplicates as ``True``. This option is not
+          compatible with ``return_index`` or ``return_inverse``.
+    return_index : boolean, default False
```
jreback: pls pls let's keep the scope to the original. just *return_inverse*. If you want to do more things, make them PR's on top.

jreback (changes requested):

h-vetinari:
> pls pls let's keep the scope to the original. just return_inverse. If you want to do more things, make them PR's on top.

@jreback I did keep strictly with the scope of just adding `return_inverse` to `duplicated`. The issue is that `Series/DataFrame` need a *slightly* different treatment than `Index`:
* the inverse for `Series/DataFrame` needs to take into account two indexes (the original index, and the one of the deduplicated data). This needs to combine the outputs `return_index=True` and `return_inverse=True` from `np.unique`.
* the inverse for `Index` is just a simple `np.ndarray`, which would in principle just need `return_inverse=True` from `np.unique`, except we need to undo the implicit sorting that `np.unique` does.

There are two options from my POV:
1. implement two different `duplicated`-methods for `Series/DataFrame` vs. `Index`, which would however overlap if `return_inverse=False`.
1. (current state of PR) implement one central `duplicated`-method, but with enough flexibility to cover both cases, and have thin wrappers around that. Since -- AFAIU -- `pandas.core.algorithms.duplicated` is not public API, I decided to add that extra flexibility.

Does that make the situation clearer? I don't mind implementing option 1, but it would lead to code duplication and would IMHO be messier.

h-vetinari:
Ugh, resolving conflicts on windows is such a pain (conflict resolution commits not recognised by git). Had to do a massive rebase...

Also added some examples to the separate `duplicated`-docstrings.

h-vetinari:
@jreback 

I refactored `core.algorithms.duplicated` and removed the offending kwargs. This means that for the `Series/DataFrame` case, there is some unnecessary `argsort`-ing happening, but I'm guessing that performance penalty (up to 2x as long) is more welcome than having a different method-signature. For completeness, the ASV-comparison is below.

I also reworked the `Series/DataFrame/Index` docstrings - it might be instructive to have a look at the examples. :)

```
     before           after       ratio
   [93bd97f3]       [18b15eb1]
        234ms            250ms     1.07  frame_methods.Duplicated.time_frame_duplicated('first', False)
+       438ms            828ms     1.89  frame_methods.Duplicated.time_frame_duplicated('first', True)
-       219ms            188ms     0.86  frame_methods.Duplicated.time_frame_duplicated('last', False)
+       656ms            1.02s     1.55  frame_methods.Duplicated.time_frame_duplicated('last', True)
-       250ms            219ms     0.88  frame_methods.Duplicated.time_frame_duplicated(False, False)
       78.1ms           85.9ms     1.10  frame_methods.Duplicated.time_frame_duplicated_mixed('first', False)
+       125ms            188ms     1.50  frame_methods.Duplicated.time_frame_duplicated_mixed('first', True)
       85.9ms           78.1ms     0.91  frame_methods.Duplicated.time_frame_duplicated_mixed('last', False)
+       172ms            203ms     1.18  frame_methods.Duplicated.time_frame_duplicated_mixed('last', True)
       85.9ms           85.9ms     1.00  frame_methods.Duplicated.time_frame_duplicated_mixed(False, False)
        219ms            203ms     0.93  frame_methods.Duplicated.time_frame_duplicated_wide('first', False)
+       203ms            234ms     1.15  frame_methods.Duplicated.time_frame_duplicated_wide('first', True)
        203ms            203ms     1.00  frame_methods.Duplicated.time_frame_duplicated_wide('last', False)
        203ms            219ms     1.08  frame_methods.Duplicated.time_frame_duplicated_wide('last', True)
        203ms            219ms     1.08  frame_methods.Duplicated.time_frame_duplicated_wide(False, False)
   3.77±0.3ms         3.47±0ms     0.92  series_methods.Duplicated.time_series_duplicated('first', False)
      141±8ms         141±10ms     1.00  series_methods.Duplicated.time_series_duplicated('first', True)
   3.35±0.2ms       3.91±0.2ms    ~1.17  series_methods.Duplicated.time_series_duplicated('last', False)
      156±8ms          172±8ms     1.10  series_methods.Duplicated.time_series_duplicated('last', True)
   3.47±0.2ms       3.91±0.3ms    ~1.12  series_methods.Duplicated.time_series_duplicated(False, False)
```

pandas/core/frame.py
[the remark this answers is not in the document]
h-vetinari: I added examples to the docstrings that illustrate the usage (as well as the need to keep track of the index of the deduplicated object in case of Series/DataFrame).

h-vetinari:
@jschendel @WillAyd @mroeschke @jreback 
gentle ping since everything is green and I now incorporated (resp. responded to) all your feedback.

h-vetinari:
@jschendel @WillAyd @mroeschke @jreback
I know this ping is quite soon after the last - please excuse. I'll be gone for a while in about a week, and would prefer to get this done before, instead of having to do lots of rebasing afterwards.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -66,10 +64,60 @@ Current Behavior:
...
     result
 
+.. _whatsnew_0240.enhancements.duplicated_inverse:
+
+The ``duplicated``-method has gained the ``return_inverse`` kwarg
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``duplicated``-method for ``Series``, ``DataFrame`` and all flavours of ``Index`` has gained a ``return_inverse`` keyword,
+which is False by default. Specifying ``return_inverse=True`` will add an object to the output (which therefore becomes a tuple)
+that allows reconstructing the original object from the deduplicated, unique subset.
```
jschendel: Can you add the issue number at the end of this sentence using the usual whatsnew conventions?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -66,10 +64,60 @@ Current Behavior:
...
+    reconstruct.equals(df)
+
+The keyword works as expected for ``keep='first'|'last'``, but cannot be used together with ``keep=False`` (since discarding all duplicates makes it impossible
+to construct an inverse).
+
 .. _whatsnew_0240.enhancements.other:
 
 Other Enhancements
 ^^^^^^^^^^^^^^^^^^
+- ``ExcelWriter`` now accepts ``mode`` as a keyword argument, enabling append to existing workbooks when using the ``openpyxl`` engine (:issue:`3441`)
```
jschendel: I think this got added by accident?

pandas/core/algorithms.py
```diff
@@ -786,16 +786,67 @@ def duplicated(values, keep='first'):
...
+
     values, dtype, ndtype = _ensure_data(values)
     f = getattr(htable, "duplicated_{dtype}".format(dtype=ndtype))
-    return f(values, keep=keep)
+    isdup = f(values, keep=keep)
+    if not return_inverse:
+        return isdup
+    elif not isdup.any():
+        # no need to calculate inverse if no duplicates
+        inv = np.array(range(len(values)))
```
jschendel: use `np.arange` instead of `np.array(range(...))`

pandas/core/algorithms.py
```diff
@@ -786,16 +786,67 @@ def duplicated(values, keep='first'):
...
+        values = values[::-1]
+        _, o2u, u2o = np.unique(values, return_index=True,
+                                return_inverse=True)
+        # the values in "values" correspond(ed) to the index of "values",
+        # which is simply np.array(range(len(values))).
+        # By flipping "values" around, we need to do the same for the index,
+        # ___because o2u and u2o are relative to that order___.
+        # Finally, to fit with the original order again, we need to flip the
+        # result around one last time.
+        o2u, u2o = np.array(range(len(values)))[::-1][o2u], u2o[::-1]
```
jschendel: use `np.arange` instead of `np.array(range(...))`

pandas/core/base.py
```diff
@@ -1242,16 +1242,39 @@ def drop_duplicates(self, keep='first', inplace=False):
...
+
         if isinstance(self, ABCIndexClass):
             if self.is_unique:
-                return np.zeros(len(self), dtype=np.bool)
-            return duplicated(self, keep=keep)
-        else:
+                isdup = np.zeros(len(self), dtype=np.bool)
+                if not return_inverse:
+                    return isdup
+                return isdup, np.array(range(len(self)))
```
jschendel: use `np.arange` instead of `np.array(range(...))`

doc/source/whatsnew/v0.24.0.txt
h-vetinari: When I started working on this, there was no "big feature" under "New Features" (usually having a descriptive section each) except this one-liner. I moved it down to "Other Enhancements"

In the same vein, I removed a duplicated "Other API Changes" section (while keeping to the order in the previous whatsnews).

jschendel: Oops...missed that the entry was deleted from further up in the whatsnew.  Thought it was a remnant from a rebase.

h-vetinari: @jschendel Thx for the review, pushed a new commit incorporating the changes.

pandas/tests/indexes/multi/test_unique_and_duplicates.py
```diff
@@ -122,14 +121,26 @@ def test_duplicate_meta_data():
...
+    major_labels = np.array([0, 0, 1, 2, 3, 3])
+    minor_labels = np.array([0, 1, 0, 1, 0, 1])
+    index_names = ['first', 'second']
+    mi = pd.MultiIndex(levels=[major_axis, minor_axis],
+                       labels=[major_labels, minor_labels],
+                       names=index_names, verify_integrity=False)
+    assert not mi.has_duplicates
+    assert mi.append(mi).has_duplicates
+
+    index = pd.MultiIndex(levels=[[0, 1], [0, 1, 2]],
```
WillAyd: This check seems a little "imbalanced" compared to what's above. Can we simplify this test through parametrization instead?

pandas/tests/indexes/multi/test_unique_and_duplicates.py
```diff
@@ -7,7 +7,6 @@
 import pandas as pd
 import pandas.util.testing as tm
 import pytest
-from pandas import MultiIndex
```
WillAyd: Same comment as before - for a change that's already pretty large would be preferable to do these minor cleanups separately in the future

pandas/tests/indexes/multi/test_unique_and_duplicates.py
```diff
@@ -194,46 +207,155 @@ def f(a):
...
+def test_duplicated_inverse(keep):
+    # check that return_inverse kwarg does not affect outcome;
+    # index of inverse must be correctly transformed as well
+    major_axis = pd.Index(['foo', 'bar'])
+    minor_axis = pd.Index(['one', 'two'])
+
+    major_labels = np.array([0, 0, 1, 0, 1, 1])
+    minor_labels = np.array([0, 1, 0, 1, 0, 1])
+    index_names = ['first', 'second']
+    mi = pd.MultiIndex(levels=[major_axis, minor_axis],
```
WillAyd: Since this is shared across quite a few tests should make a module fixture

pandas/tests/series/test_analytics.py
```diff
@@ -1045,6 +1030,89 @@ def test_drop_duplicates_bool(self):
...
+
+        # test that inv works (and fits together with expected_isdup)
+        unique = s.loc[~expected_isdup]
+        reconstr = unique.reindex(result_inv)
+        # Series has no set_index (GH21684)
+        reconstr.index = result_inv.index
+        tm.assert_series_equal(reconstr, s)
+
+    @pytest.mark.parametrize('keep', ['first', 'last'])
+    def test_duplicated_inverse_fastpath(self, keep):
```
WillAyd: What is the point of all of the `fastpath` tests? Couldn't an array of all distinct values simply be sent via parametrization to the other tests?

pandas/tests/series/test_analytics.py
```diff
@@ -1045,6 +1030,89 @@ def test_drop_duplicates_bool(self):
...
+        result = s.duplicated(keep=keep)
+        tm.assert_series_equal(result, expected)
+
+    @pytest.mark.parametrize('keep, expected', [
+        ('first', Series([False, False, True, False, True])),
+        ('last', Series([True, True, False, False, False])),
+        (False, Series([True, True, True, False, True]))
+    ])
+    def test_duplicated_nan_none(self, keep, expected):
+        # np.nan and None are considered equal
```
WillAyd: not considered equal here, no?

WillAyd (changes requested):

pandas/tests/frame/test_analytics.py
```diff
@@ -1545,6 +1545,143 @@ def test_isin_empty_datetimelike(self):
     # ----------------------------------------------------------------------
     # Row deduplication
 
+    @pytest.mark.parametrize('subset', ['a', ['a'], ['a', 'B']])
+    def test_duplicated_with_misspelled_column_name(self, subset):
```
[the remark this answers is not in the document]
h-vetinari: The heading was there already, I just added all the `duplicated` tests in one spot (and moved two related ones to the same place).

Fair enough about keeping the diffs small though. Next time would be enough, or do you want me to revert these changes?

pandas/tests/frame/test_analytics.py
```diff
@@ -1545,6 +1545,143 @@ def test_isin_empty_datetimelike(self):
...
+                        'C': [np.nan, 3, 3, None, np.nan]})
+
+        if subset is None:
+            subset = list(df.columns)
+        elif isinstance(subset, string_types):
+            # need to have a DataFrame, not a Series
+            # -> select columns with singleton list, not string
+            subset = [subset]
+
+        expected = df[subset].duplicated(keep=keep).rename(name=None)
```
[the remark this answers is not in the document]
h-vetinari: Good catch, that was a remnant from when I didn't have the string-check (because `Series.duplicated` keeps the name of the original). Removed.

pandas/tests/indexes/common.py
```diff
@@ -351,6 +351,95 @@ def test_duplicates(self, indices):
         assert not idx.is_unique
         assert idx.has_duplicates
 
+    @pytest.mark.parametrize('keep', ['first', 'last', False])
+    def test_duplicated(self, indices, keep):
+        if type(indices) is not self._holder:
+            return
```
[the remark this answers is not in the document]
h-vetinari: To be honest, I couldn't grok right away what's all happening in the index-test, but copied the conditions you point out verbatim from `test_duplicates` directly above. Skipping would be cleaner, that's for sure, but a blank `return` is a common pattern in `test/indexes/common.py`. In light of your comment above, I'm not changing those right now, to not further blow up this PR/diff.

pandas/tests/indexes/multi/test_unique_and_duplicates.py
h-vetinari: This is mainly due to overlaps with #21514 - I extracted the fixture from the class-based test, because it was hard to read for me before. I didn't attack the rest of the file though, so now you see this fixture in several places, I'm guessing.

The clean-up was to unify/harmonize some of what had happened with #21514...

pandas/tests/indexes/multi/test_unique_and_duplicates.py
h-vetinari: I don't understand what unbalanced is supposed to mean here. It is exactly how this test worked pre #21514.

pandas/tests/series/test_analytics.py
h-vetinari: Yes, this was a relict from the first df-tests.

pandas/tests/series/test_analytics.py
h-vetinari: I just made a separate test to make sure that path was hit. I don't fully understand the parametrization request - each test should run on an array and its deduplicated version?

pandas/tests/indexes/multi/test_unique_and_duplicates.py
h-vetinari: I agree. This would have been something for #21514, but I'll pick it up.

h-vetinari (commented):
Thanks for the feedback, commit coming soonish.

pandas/tests/indexes/multi/test_unique_and_duplicates.py
h-vetinari: Done (git mistakenly does not show this as outdated).

jreback (changes requested):
i have a large number of comments
but won’t get to till next week

main issue is all of the orthogonal changes / cleanups in tests need to be in a separate P

this diff is way to big

h-vetinari:
@jreback @WillAyd Rebased on the merged PRs and the most current state of #21902.

h-vetinari:
@jreback @WillAyd Now that #21902 is in, all preparatory PRs should be out of the way, and this is ready to be discussed again.

h-vetinari:
@jreback @WillAyd This has been lying around for about a month or so. Could you guys have a look please? :)

WillAyd:
Traveling for a few days but can review again next week. Larger PRs just take more time to review - definitely still appreciate the contribution (and patience!)

h-vetinari:
@WillAyd No problem. :)
TBH, part of the reason for my patience is that I still want to get #22225 in beforehand, but that PR itself is waiting for #22486, which is blocked by #22236, haha...

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -159,6 +159,56 @@ This is the same behavior as ``Series.values`` for categorical data. See
 :ref:`whatsnew_0240.api_breaking.interval_values` for more.
 
 
+.. _whatsnew_0240.enhancements.duplicated_inverse:
+
+The ``duplicated``-method has gained the ``return_inverse`` kwarg
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The ``duplicated``-method for ``Series``, ``DataFrame`` and all flavors of ``Index`` has gained a ``return_inverse`` keyword,
```
WillAyd: Double backticks are for literals and code samples, so let's put `False` in double backticks and everything else that you have here rendered with just single backticks, save `return_inverse=True` which is OK

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -159,6 +159,56 @@ This is the same behavior as ``Series.values`` for categorical data. See
...
+which is False by default. Specifying ``return_inverse=True`` will add an object to the output (which therefore becomes a tuple)
+that allows reconstructing the original object from the deduplicated, unique subset (:issue:`21357`).
+
+For ``Index`` objects, the inverse is an ``np.ndarray``:
+
+.. ipython:: python
+
+    idx = pd.Index(['a', 'b', 'b', 'c', 'a'])
+    idx.has_duplicates
```
WillAyd: Can remove this expression

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -159,6 +159,56 @@ This is the same behavior as ``Series.values`` for categorical data. See
...
+    isduplicate, inverse = df.duplicated(keep='last', return_inverse=True)
+    isduplicate
+    inverse
+
+    unique = df.loc[~isduplicate]  # same as df.drop_duplicates(keep='last')
+    unique
+    reconstruct = unique.reindex(inverse.values).set_index(inverse.index)
+    reconstruct.equals(df)
+
+The keyword works as expected for ``keep='first'|'last'``, but cannot be used together with ``keep=False`` (since discarding all duplicates makes it impossible
```
WillAyd: Not sure what keyword in particular this is referencing, though concept is clear. Better to move into Notes section of docstring imo

pandas/tests/frame/test_duplicates.py
```diff
@@ -88,6 +88,75 @@ def test_duplicated_subset(subset, keep):
     tm.assert_series_equal(result, expected)
 
 
+@pytest.mark.parametrize('keep, expected_inv_values', [
+    ('first', [1, 4, 4, 16, 1]),
+    ('last', [25, 9, 9, 16, 25])
+])
+def test_duplicated_inverse(keep, expected_inv_values):
+    # check that return_inverse kwarg does not affect outcome;
```
WillAyd: Can just reference issue number instead of this comment

pandas/tests/frame/test_duplicates.py
```diff
@@ -88,6 +88,75 @@ def test_duplicated_subset(subset, keep):
...
+    # unsorted index important to check 'first'/'last' functionality
+    df = DataFrame(np.random.randint(0, 10, (10000, 3)),
+                   columns=list('ABC')).sample(5000)
+
+    expected_isdup = df.duplicated(keep=keep, subset=subset)
+    result_isdup, result_inv = df.duplicated(keep=keep, subset=subset,
+                                             return_inverse=True)
+    tm.assert_series_equal(result_isdup, expected_isdup)
+
+    if subset is None:
```
WillAyd: Can you just parametrize lists instead of this code here? Seems to unnecessarily complicate the test case

pandas/tests/frame/test_duplicates.py
```diff
@@ -88,6 +88,75 @@ def test_duplicated_subset(subset, keep):
...
+
+    rgx = 'The parameters return_inverse=True and keep=False cannot be.*'
+    with tm.assert_raises_regex(ValueError, rgx):
+        df.duplicated(keep=False, return_inverse=True)
+
+
+@pytest.mark.parametrize('keep', ['first', 'last'])
+@pytest.mark.parametrize('subset', [None, ['A', 'B'], 'A'])
+def test_duplicated_inverse_large(subset, keep):
+    # unsorted index important to check 'first'/'last' functionality
```
WillAyd: Not instantly apparent the link between this comment and the frame construction below. Does this need to be random data or can you explicitly construct to ensure proper testing?

pandas/tests/indexes/common.py
```diff
@@ -378,6 +378,82 @@ def test_duplicated(self, indices, keep):
         result = idx.duplicated(keep=keep)
         tm.assert_numpy_array_equal(result, expected)
 
+    @pytest.mark.parametrize('keep', ['first', 'last'])
+    def test_duplicated_inverse(self, indices, keep):
+        # check that return_inverse kwarg does not affect outcome
+        if type(indices) is not self._holder:
+            pytest.skip('Can only check if we have the correct type')
```
WillAyd: Hmm when does this happen?

pandas/tests/indexes/common.py
```diff
@@ -378,6 +378,82 @@ def test_duplicated(self, indices, keep):
...
+        idx = self._holder(idx.values[duplicated_selection])
+
+        expected_isdup = idx.duplicated(keep=keep)
+        if keep == 'first':
+            _, tmp_ind, tmp_inv = np.unique(idx, return_index=True,
+                                            return_inverse=True)
+        else:  # 'last'
+            # switch order before calling unique then restore correct ordering
+            # for tmp_ind, tmp_inv
+            _, tmp_ind, tmp_inv = np.unique(idx[::-1], return_index=True,
```
WillAyd: Is there no way to simplify this? Pretty difficult to tell what is going on with all of the reverse stepping

pandas/tests/indexes/multi/test_duplicates.py
```diff
@@ -264,3 +241,84 @@ def test_get_duplicates():
...
             tm.assert_numpy_array_equal(mi.duplicated(),
                                         np.zeros(len(mi), dtype='bool'))
+
+
+@pytest.mark.parametrize('keep, expected', [
+    ('first', np.array([False, False, False, True, True, False])),
+    ('last', np.array([False, True, True, False, False, False])),
+    (False, np.array([False, True, True, True, True, False]))
+])
+def test_duplicated(idx_dup, keep, expected):
```
WillAyd: Was there a change to this and subsequent test? Unclear why this is showing in diff

pandas/core/algorithms.py
```diff
@@ -785,16 +785,67 @@ def duplicated(values, keep='first'):
...
+                         "instances of a duplicate).")
+
     values, dtype, ndtype = _ensure_data(values)
     f = getattr(htable, "duplicated_{dtype}".format(dtype=ndtype))
-    return f(values, keep=keep)
+    isdup = f(values, keep=keep)
+    if not return_inverse:
+        return isdup
+    elif not isdup.any():
+        # no need to calculate inverse if no duplicates
```
WillAyd: Is this always going to hold true? For example, if we work with a Series that is not sequentially indexed starting at 0 but doesn't contain duplicates is this going to return the appropriate result?

WillAyd (changes requested):
Here's some more feedback. Generally anything we can do to simplify code is going to be beneficial from a review and long term maintenance perspective

h-vetinari (commented):
Incorporated the requested changes, where appropriate

pandas/core/series.py
```diff
@@ -1604,64 +1604,131 @@ def duplicated(self, keep='first'):
...
+        construct an inverse).
 
         Examples
         --------
         By default, for each set of duplicated values, the first occurrence is
         set on False and all others on True:
 
-        >>> animals = pd.Series(['lama', 'cow', 'lama', 'beetle', 'lama'])
+        >>> animals = pd.Series(['lama', 'cow', 'lama', 'beetle', 'lama'],
+                                index=[1, 4, 9, 16, 25])
```
jorisvandenbossche: to get this to pass the doctests, you need to add `...` here at the beginning of a 'continued' line. See here for some details: http://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html#tips-for-getting-your-examples-pass-the-doctests

h-vetinari: @jorisvandenbossche Thanks a lot!

h-vetinari:
@WillAyd I incorporated your feedback - any open questions?

@jreback care to take a look?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -159,6 +159,52 @@ This is the same behavior as ``Series.values`` for categorical data. See
 :ref:`whatsnew_0240.api_breaking.interval_values` for more.
 
 
+.. _whatsnew_0240.enhancements.duplicated_inverse:
+
+The `duplicated`-method has gained the `return_inverse` kwarg
```
jorisvandenbossche: double backticks were you now use single backticks (also on the lines below)

(in rst, double backticks give code-styled text)

pandas/core/frame.py
```diff
@@ -4353,14 +4353,157 @@ def duplicated(self, subset=None, keep='first'):
               first occurrence.
             - ``last`` : Mark duplicates as ``True`` except for the
               last occurrence.
-            - False : Mark all duplicates as ``True``.
+            - False : Mark all duplicates as ``True``. This option is not
+              compatible with ``return_inverse``.
+        return_inverse : boolean, default False
+            If True, also return the selection from the index from the
```
jorisvandenbossche: "selection from the index" sounds strange to me, and it does not really say what the values are (I understand what you mean, but I think we should think about a better wording). 
Also the following lines are rather complex to understand (eg the "boolean complement of the first output")

pandas/core/frame.py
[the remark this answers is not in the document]
jorisvandenbossche: The index is still available in the first output, so it is not *that* essential that the second output is a Series (there is still a design issue for sure)

jorisvandenbossche (commented):
Didn't look in detail to the implementation, but some general questions:

- What is the reason you decided to add it to `duplicated` and not `drop_duplicates` ? (because from quickly looking at the issue, and from the docstring examples, that is actually what you want?) 
  I also ask because I think it would be a more natural fit there, as it is not literally an "inverse" for the result of `duplicated` (as it is for `drop_duplicates` or as for `unique` in numpy)

- I think we should have more discussion (or at least motivation for the current design) about the actual return value (series vs array, index labels vs positional integers). Always returning an array of integers would also be an option (and is eg robust for duplicates in the index)

doc/source/whatsnew/v0.24.0.txt
jorisvandenbossche: I see I commented now again the reverse .. 
But @WillAyd, I think you gave the wrong recommendation, the "Double backticks are for literals and code samples" is correct, but we see method names as "code samples" (although for duplicated it is actually better to use the link-markup `` :meth:`~DataFrame.duplicated` ``). 
In general I don't think there should be much single backticks in normal .rst files (except for those sphinx mark-up cases like refs)

In docstrings we *do* use single backticks to refer to parameter names.

h-vetinari:
@jorisvandenbossche

> What is the reason you decided to add it to duplicated and not drop_duplicates ? (because from quickly looking at the issue, and from the docstring examples, that is actually what you want?)
I also ask because I think it would be a more natural fit there, as it is not literally an "inverse" for the result of duplicated (as it is for drop_duplicates or as for unique in numpy)

I wanted to add `.unique` method in #21357, but got told by @jreback that:
> -1 on adding any new methods, but would be ok with adding a return_inverse kwarg (clearly defaulting to False), to .duplicated() to easily recovering the indexer.

I still think it makes sense for `duplicated`, which is invoked by `drop_duplicates`, and so would need to support that anyway if one wanted to add it to `drop_duplicates`.

> I think we should have more discussion (or at least motivation for the current design) about the actual return value (series vs array, index labels vs positional integers). Always returning an array of integers would also be an option (and is eg robust for duplicates in the index)

numpy returns two arrays of indexes, one relative to the original set and one relative to the unique values. I think this information fundamentally belongs together and since pandas supports the index/values distinction, I strongly believe that they should live in the same object. I don't like reusing the index of `duplicated` itself, because then I need to keep around *two* separate objects to reconstruct. But happy to discuss alternative approaches.

h-vetinari:
@jorisvandenbossche 
I incorporated your feedback. Since no compelling alternatives have come up for the return type, I left that aspect as-is.

h-vetinari:
@jreback @WillAyd @jorisvandenbossche 

How can I/we move this PR forward?

jorisvandenbossche:
@h-vetinari sorry for the slow answer

I think my main worry is that we are adding a `return_inverse` keyword which actually does not return the inverse for that function (it does return the inverse for another function), and that it is in name similar to numpy's keyword, but in usage also different.

Numpy's `unique` is more comparable to `drop_duplicates`. Therefore I think such a keyword would be  a more natural fit there.


> numpy returns two arrays of indexes

Can you clarify this? You are talking about `np.unique`?  (but I think that only returns a single array of indices?)

Side note: it might make sense to add this to `pd.unique` / `Series.unique` as well? (not necessarily at the same time; or might actually be an easier starter)

h-vetinari:
@jorisvandenbossche 

> I think my main worry is that we are adding a return_inverse keyword which actually does not return the inverse for that function (it does return the inverse for another function), and that it is in name similar to numpy's keyword, but in usage also different.

Fair point. I would prefer a `df.unique` as well, but was following what I was told in #21357
> -1 on adding any new methods, but would be ok with adding a return_inverse kwarg (clearly defaulting to False), to .duplicated() to easily recovering the indexer.

Re: 
> >  numpy returns two arrays of indexes

> Can you clarify this? You are talking about np.unique? (but I think that only returns a single array of indices?)

https://docs.scipy.org/doc/numpy/reference/generated/numpy.unique.html has both `return_indices` and `return_inverse`, and both are necessary to reconstruct from the pandas side (the difference for us being that we also need to keep track of the index, whereas numpy just has `arange` as index).

> Side note: it might make sense to add this to pd.unique / Series.unique as well? (not necessarily at the same time; or might actually be an easier starter)

Agreed. Actually, for me the best solution would be: 
* add `df.unique` 
* have `df.unique`, `series.unique`, `index.unique` with `return_inverse`-kwarg
* dispatch from `pd.unique` to appropriate code branch

pandas/core/algorithms.py
```diff
@@ -785,16 +785,67 @@ def duplicated(values, keep='first'):
...
+        inv = np.arange(len(values))
+        return isdup, inv
+
+    if keep == 'first':
+        # o2u: original indices to indices of ARRAY of unique values
+        # u2o: reduplication from array of unique values to original array
```
jreback: I'd rather you not use ``np.unique`` at all, its not as performant as ``pd.unique``, doesn't handle all dtypes, and sorts. 

Further pls, pls use actual names here, and really avoid using abbreviations in any library code.

jreback (changes requested):
i haven't looked at this in any detail, there is a lot going on here.

pandas/core/algorithms.py
h-vetinari: @jreback 
> I'd rather you not use `np.unique` at all, its not as performant as `pd.unique`, doesn't handle all dtypes, and sorts.

Yes, it would be nicer to have this implemented in the cython hashtable functions, but that performance improvement is for a follow-up. `np.unique` is an easy solution (and we're only calling it on a series of ints, not objects, because the factorization for that hasn't changed!).

> Further pls, pls use actual names here, and really avoid using abbreviations in any library code.

There's a fully commented, thoroughly explained and very localized part where these appear. Not sure how this is unclear, but will adapt...

h-vetinari: Finally, the core changes are minimal (but I understand that it looks like a lot). TLDR: Implementation moves to `core.algorithms`, everything else wraps around that (+ doc improvements). Tests are expanded, and ASVs added.

jreback:
i am not against adding this to .unique, though as you have mentioned that is indeed just drop_duplicates

h-vetinari:
@jorisvandenbossche 
Thanks for your answer:

> I have wanted a DataFrame.unique for a long time, but the question is how it would be different compared to drop_duplicates?

I would suggest that `unique` just acts as a thin wrapper around `drop_duplicates`. The latter has been in the pandas API forever, but the former is even more established, and I think they could peacefully coexist.

> I think the numpy's return_indices is rather unrelated here? It is about the index into the original array to get the unique ones, while you want to preserve the index of the original array?

Assuming `np.unique` didn't sort (which I have to laboriously undo in this PR) and `keep='first'`, we would have:
```
isduplicated = some_series.duplicated()
_, values2unique, unique2values = np.unique(some_series.values,
                                            return_index=True, return_inverse=True)
inverse = Series(some_series.index[values2unique][unique2values], index=some_series.values)
```

and neither `values2unique` (the indices) nor `unique2values` (the inverse) can be omitted. Again, the difference to `numpy` here is that `some_series` can have an index different from `arange(len(some_series))`, which is why we need to first select the right indices of the index (to get the uniques) and the broadcast them back to the full shape to get the full inverse. If we didn't care about the index of `some_series`, then the indices would indeed be not necessary, but then we wouldn't be creating an inverse in the pandas sense (because two Series with different indexes are not equal).

> Adding a return_inverse to the existing unique function / methods would certainly be a welcome change I think.

I had already collected some thoughts for this during the day and opened a (largely orthogonal to this PR) issue for this: #22824

jorisvandenbossche:
> Assuming np.unique didn't sort (which I have to laboriously undo in this PR) 

That is because for the inverse you rely on `np.unique` instead of our own implementations? 
Another option might be to actually rely on our own algos for that? (I assume the hashtables already support that, as eg we need it for factorizing. Which led me think: `pd.unique` with a `return_inverse` argument is actually basically the same as `pd.factorize` ?)

I see how you use both `return_index` and `return_inverse` there, but that is to construct the inverse (so which would be hidden for the user), I thought you were speaking about the actual API for return_inverse towards the user. So therefore the confusion.

h-vetinari:
Alright, I'm closing this for now, as the lack of cython-inverse has now been solved by #23400, but for `.unique` rather than `.duplicated`. Therefore I'll follow that line of development (it's also closer to the intention of #21357).

PR #21649, opened 2018-06-27
DOC: Fix versionadded directive typos in IntervalIndex

jschendel:
There needs to be a space between the `..` and `versionadded::`.  Currently being rendered as plaintext, which doesn't look too great:

image

Couldn't find any other instances of this in the codebase for `versionadded` or any other commonly used directives.  Pretty minor so tagging this as 0.23.2.


jreback (approved):
sure - merge away (and add backport label)

maybe worth trying to catch this in lint.sh ?

jreback:
thanks!

Issue #21677, opened 2018-06-29
API: unclear what integer level name references: name or position?


jorisvandenbossche:
I assumed that in eg `get_level_values`, `stack`, `unstack`, ... (which all boil down to the behaviour of `_get_level_number`) using an integer to specify the level, would always mean positional because we have no way to disambiguate between position or name in those cases. 

But, this seems very inconsistent:

```
In [40]: mi = pd.MultiIndex.from_product([[0, 1], [2, 3], [4, 5]], names=[1, 1, 2])

In [41]: mi._get_level_number(0)     # <--- positional
Out[41]: 0

In [43]: mi._get_level_number(1)     # <--- positional (as name should raise an error given duplicates)
Out[43]: 1

In [44]: mi._get_level_number(2)     # <--- positional / name is the same
Out[44]: 2

In [45]: mi = pd.MultiIndex.from_product([[0, 1], [2, 3], [4, 5]], names=[2, 1, 0])

In [46]: mi._get_level_number(0)     # <--- name
Out[46]: 2

In [47]: mi._get_level_number(2)     # <--- name
Out[47]: 0

In [48]: mi = pd.MultiIndex.from_product([[0, 1], [2, 3], [4, 5]], names=[2, 1, 1])

In [49]: mi._get_level_number(2)     # <--- name
Out[49]: 0

In [50]: mi._get_level_number(1)     # <--- positional (as name should raise an error given duplicates)
Out[50]: 1

In [51]: mi._get_level_number(0)     # <--- positional
Out[51]: 0

In [52]: mi = pd.MultiIndex.from_product([[0, 1], [2, 3], [4, 5]], names=[1, 0, 1])

In [53]: mi._get_level_number(1)     # <--- positional (as name should raise an error given duplicates)
Out[53]: 1

In [54]: mi._get_level_number(0)     # <--- name
Out[54]: 1

In [55]: mi = pd.MultiIndex.from_product([[0, 1], [2, 3], [4, 5]], names=[0, 0, 1])

In [56]: mi._get_level_number(0)     # <--- positional (as name should raise an error given duplicates)
Out[56]: 0

In [57]: mi._get_level_number(1)     # <--- name
Out[57]: 2

In [58]: mi._get_level_number(2)     # <--- positional
```

Am I missing something? Was this discussed before?

cc @toobaz

jreback:
there are some issues that i believe this was discussed

[...]

PR #21678, opened 2018-06-29
API: referring to duplicate level name raises ValueError

jorisvandenbossche:
follow-up on: https://github.com/pandas-dev/pandas/pull/21423

But, this currently has a lot of failures, because of https://github.com/pandas-dev/pandas/issues/21677 that we first need to resolve.


jorisvandenbossche:
Changed the implementation a bit to keep the fallback logic for integer names, to decouple it from the discussion in #21677

jreback (changes requested):
wow much simpler 

minor comment - pls add whatsnew note

any downsides to doing this?

jorisvandenbossche:
> wow much simpler

I don't think the code is simpler, the diff is just smaller :-)

> any downsides to doing this?

I don't think so. 
It's an api change, but only a minor one I think. In principle we could also keep it as KeyError (I don't think there is necessarily a more correct one, although KeyError is typically used for non-existing key, while here the key is existing but duplicate), but then encountering this in eg `stack` would change from ValueError to KeyError (so a api change as well).

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -229,6 +229,9 @@ Other API Changes
 ^^^^^^^^^^^^^^^^^
 
 - :class:`DatetimeIndex` now accepts :class:`Int64Index` arguments as epoch timestamps (:issue:`20997`)
+- Accessing a level of a ``MultiIndex`` with a duplicate name (e.g. in
+  :meth:~MultiIndex.get_level_values) now raises a ValueError instead of
```
jreback: double backticks around the errors

jreback (changes requested):
minor comment

jreback (approved):

PR #21681, opened 2018-06-29
Allow merging on object / non-object column

### LINKED ISSUE #23733: Merging on int64 and object columns, pandas 0.23.4
#### Problem description
Hi, I was trying to merge 2 data frame based on a column named ID, which consists of integer.

In the case the ID column is of type numpy.int64 in one df, and of python native int in the other df. In the latest version of 0.23.4, I received this error 
```
ValueError: Your are tryign to merge on int64 and object columns, If you wish to proceed you should use pd.concat
```
Attached the screenshot of the problem

image

I have never encountered this situation in earlier panda version such as 0.22.0. The same code ran without an issue.

May I know what is the purpose of adding this constraint in the upgraded version?

jorisvandenbossche:
Follow-up on discussion in https://github.com/pandas-dev/pandas/pull/21310 (cc @jreback )

What this does: allow `merge` on an object and a non-object column (eg object and int, object and categorical, ..), by coercing the non-object column to object dtype.

Reasoning:
- object dtype can be anything, so IMO pandas should not try to be smart and think to know the merge is not valid
- all other, certainly invalid dtype combinations (eg int and datetime64) keep raising an error

The main disadvantage is that we no longer detect the case of merging object strings with eg integer column, but IMO we cannot avoid such cases as long as we don't have a proper string dtype but use object dtype for this (you can also mix string and ints in a column, we also don't raise for that for a similar reason).

Additional option would be to issue a warning in this case.

Closes https://github.com/pandas-dev/pandas/issues/23733


jreback:
-1 in this - if u what to merge an object column with a non object column then coerce to object 

this will propagate objects columns very easily - against the principle of strongly dtyping things 
what is the use case that is motivating this?

jorisvandenbossche:
There was actually a question on SO today reporting a problem in geopandas caused by this: https://stackoverflow.com/questions/51138156/how-can-i-use-sjoin-to-merge-two-geodataframes-without-an-error.

In a certain way, you could also see this PR as a regression fix. We probably should have made such a change with a deprecation warning first (if we want the original change).

jreback:
I disagree here. merging on non-alike dtypes is simply wrong. objects cannot merge with non-object, full stop. Sure we have a special case for boolean, but allowing object with integer is just a disaster. '

If you want to try a bit more inference I am ok with that. e.g. say you have an object column, then am ok with running ``infer_dtype`` on it. and allowing that to merge with a similar type (e.g. integer). That is a trivial coercion.

jorisvandenbossche:
> objects cannot merge with non-object, full stop.

They can perfectly well, depending on what is inside the object dtype column. It's not about "cannot", it is about "do we allow it or not". And for the record, until one month ago pandas has always allowed it.

> If you want to try a bit more inference I am ok with that.

I have thought about that as well, but inference can be costly, so I would rather not do it. 
Also, it will not work in all cases, eg with an empty frame.

Are you OK with raising an informative warning instead? Where we clearly indicate to the user what is happening and this typically points to a problem with their types? (but at least it allow people to ignore the warning, or to fix it only later) (I thought you hinted on being OK with that in the previous PR, but not sure)

jorisvandenbossche:
BTW, input from some other devs is certainly welcome, Jeff and I clearly disagree on this .. :-)

jreback:
> Are you OK with raising an informative warning instead? Where we clearly indicate to the user what is happening and this typically points to a problem with their types? (but at least it allow people to ignore the warning, or to fix it only later) (I thought you hinted on being OK with that in the previous PR, but not sure)

I guess that is ok. The original reason that this was changed is to avoid the all-NaN problem (e.g. you merge with something that has no matches because of the dtype mismatch). I don't want to go back to that as its worse from a user POV.

open to compromise but would rather be strict here. (and inference is not generally expensive compared to a borked merge :>)

WillAyd:
I'm +1 on being strict. At the very least if the end user is explicit about types they stand to benefit performance- and resource-wise instead of pandas falling back to `object`.

jorisvandenbossche:
> (and inference is not generally expensive compared to a borked merge :>)

yeah, that is true .. :) 
Thinking this a bit more through, then the main problem seems to how to handle the result. For example then need to add logic that eg 'integer' and 'mixed-integer-float' *might* be OK, which is repeating the logic (and also need to special case the 'empty'). 
Of course, I could also do the full existing check based on the inferred types hierarchy (to avoid such duplicated logic), assuming that for non-object dtypes this inferring of the type is cheap.

> The original reason that this was changed is to avoid the all-NaN problem  .. I don't want to go back to that as its worse from a user POV.

I just realized that I actually think a warning can be better from a user POV. Instead of the error, you would then *see* the NaNs, together with the informative warning explaining why the NaNs are there and how to solve it. I think this could actually be better to get what is happening than only the error. 

And this also alleviates the direct code breakages caused by the change in 0.23.0 to be more strict.

> At the very least if the end user is explicit about types they stand to benefit performance- and resource-wise instead of pandas falling back to object.

Yes, but note that there can be good reasons to want or to end up with object data, and not all use cases of pandas are performance sensitive (eg small data).

jreback:
@jorisvandenbossche can you rebase this

jreback (changes requested):

jreback:
@jorisvandenbossche can you merge master

h-vetinari:
I've had colleages come and complain to me about this recently - especially since this used to work.

I would advocate to be strict, but based on the inferred type. The case where it's an all-integer index hiding behind an object dtype would be easily solved through that, and the other cases would continue to emit the warning.

jreback:
@jorisvandenbossche can you merge master

TomAugspurger:
Fixed the merge conflicts. Haven't taken a look at the changes yet.

jreback (approved):

pandas/tests/reshape/merge/test_merge.py
```diff
@@ -1055,22 +1042,53 @@ def test_merge_incompat_infer_boolean_object(self):
         result = pd.merge(df2, df1, on='key')
         assert_frame_equal(result, expected)
 
+    @pytest.mark.parametrize('df1_vals, df2_vals, left_type, right_type', [
+
+        # infer to numeric
+        ([0, 1, 2], ["0", "1", "2"],
+         is_integer_dtype, is_object_dtype),
```
TomAugspurger: I don't understand this first one... Why is the dtype of `A` int?

```python
In [7]: a = pd.DataFrame({"A": [1, 2], "B": [3, 4]})

In [8]: b = pd.DataFrame({"A": ['1', '2'], "B": [3, 4]})

In [9]: pd.merge(a, b, on="A").A.dtype
Out[9]: dtype('int64')
```

I thought the expected behavior was to coerce both to object when the dtypes don't match.

pandas/core/reshape/merge.py
```diff
@@ -966,22 +973,24 @@ def _maybe_coerce_merge_keys(self):
...
-                    typ = lk.categories.dtype if lk_is_cat else object
-                    self.left = self.left.assign(
-                        **{name: self.left[name].astype(typ)})
-                if name in self.right.columns:
-                    typ = rk.categories.dtype if rk_is_cat else object
-                    self.right = self.right.assign(
-                        **{name: self.right[name].astype(typ)})
+            if name in self.left.columns:
+                typ = lk.categories.dtype if lk_is_cat else object
+                self.left = self.left.assign(
```
TomAugspurger: Do we want the implicit copy from assign here?

pandas/tests/reshape/merge/test_merge.py
jreback: we actually infer this because this is what numpy does. we could exclude non-numeric inference though, let me see

pandas/core/reshape/merge.py
jreback: yes, I think there is an implicit bug here as we really need to validate a categorical type vs an object dtype (e.g. imagine a categorical with M8[ns] categories vs merging on a M8[ns] dtype). but I think we have an issue about this already. Changing it was a rabbit hole, so can address in the future.

jreback:
@TomAugspurger

TomAugspurger (commented):
LGTM.

May leave this open for a day in case @jorisvandenbossche is around to look tomorrow, but I think this is good to go.

jreback:
yes anything with object just forces conversion to object 

we
could be more specific and disallow strings vs non strings

jreback:
added a commit to be more restrictive.

jreback:
@jorisvandenbossche if you have a chance

TomAugspurger (approved):
LGTM

jreback:
thanks @jorisvandenbossche !

jorisvandenbossche:
@jreback and thanks for finishing the PR!

One thing I note now, after your last comment to be a bit more strict on merging object and numeric columns, the whatsnew note is not fully correct any more.

I am also wondering a bit how costly this inferring of the type for object columns is.

jreback:
> One thing I note now, after your last comment to be a bit more strict on merging object and numeric columns, the whatsnew note is not fully correct any more.

yeah ok we are a bit more strict on strict.

inference is << cost of actually merging things, its a no-brainer and we do it all over the place.

kevalshah90:
I am running into this issue and found this thread. Was this change pushed to pandas or can we not merge object/non-object type. 

Here is my code: 

```
res_leads = pd.merge(df_temp,
                         df_market[['Tenant_Industry','Tenant','num_employees','size_calc']],
                         left_on = df_temp.index,
                         right_on = 'Tenant',
                         how='left')
```

Error traceback: 

```
  File "/Users/....../market.py", line 267, in render_table
    result = top_leads(company, market)
  File "/Users/...../CRE/Platform/Deploy/return_leads.py", line 77, in top_leads
    how='left')
  File "/anaconda3/lib/python3.7/site-packages/pandas/core/reshape/merge.py", line 61, in merge
    validate=validate)
  File "/anaconda3/lib/python3.7/site-packages/pandas/core/reshape/merge.py", line 555, in __init__
    self._maybe_coerce_merge_keys()
  File "/anaconda3/lib/python3.7/site-packages/pandas/core/reshape/merge.py", line 983, in _maybe_coerce_merge_keys
    raise ValueError(msg)
ValueError: You are trying to merge on int64 and object columns. If you wish to proceed you should use pd.concat
```

PR #21683, opened 2018-06-29
Release 0.23.2 backports part I

jorisvandenbossche:


pandas/tests/series/test_analytics.py
```diff
@@ -1866,6 +1866,189 @@ def s_main_dtypes():
     return df
 
 
+class TestMode(object):
```
jorisvandenbossche: the addition of this class is probably something that went wrong in fixing the conflict. Will see what the tests do, if this is was simply moved in another PR and the tests passes, would leave it like this here.

TomAugspurger: I thought I had an issue with a mode test in https://github.com/pandas-dev/pandas/pull/21442/files, but didn't see it in a quick scan. Could be misremembering.

PR #21686, opened 2018-06-29
API: remove ordered requirement in Categorical.searchsorted

### LINKED ISSUE #21667: PERF: make Categorical.searchsorted not require ordered=True
``searchsorted`` requires that the searched object is (monotonically) sorted to produce correct results. Orderedness is neither a necessary or sufficient condition to make searchsorted work correctly.

``Categorical.searchsorted`` has a hard check for if the Categorical is ordered:

https://github.com/pandas-dev/pandas/blob/e0f978d43c960e0af7944a9d7d410662e071c0c7/pandas/core/arrays/categorical.py#L1382-L1388

This is too strict, as unordered but sorted Categoricals could also benefit from using searchsorted.

I propose removing this check and (like for non-categoricals) let the user have the responsibility to ensure that the Categorical is sorted correctly. This would allow very quick lookups in all sorted Categoricals, whether they're ordered or not.

topper-123:
- [x] closes #21667
- [x] tests added / passed
- [x] whatsnew entry

``searchsorted`` is mainly a tool to find value locations in a sorted array. As such it doesn't make sense to restrict it to not work with unordered categoricals. For example, doing searchsorted on ordered, but unsorted arrays gives wrong result, so orderedness is a relatively irrelevant thing to check for in searchsorted.

The current solution to first convert to a ordered Categorical is inelegant and requires setting up different code paths depending on the situation:

```python
if categorical.ordered:
    val = c.searchsorted(key)
else:
    val = c.as_ordered().searchsorted(key)
```

I've also added to the doc string of searchsorted to make clear that sortedness is required for a correct result (this is not an issue that is limited to categoricals, but all Arrays)

### DIFF
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -1399,13 +1399,6 @@ class Categorical(ExtensionArray, PandasObject):
     @Substitution(klass="Categorical")
     @Appender(_shared_docs["searchsorted"])
     def searchsorted(self, value, side="left", sorter=None):
-        if not self.ordered:
-            raise ValueError(
-                "Categorical not ordered\nyou can use "
-                ".as_ordered() to change the Categorical to an "
-                "ordered one"
-            )
-
         from pandas.core.series import Series
 
         codes = _get_codes_for_values(Series(value).values, self.categories)
--- a/pandas/core/base.py
+++ b/pandas/core/base.py
@@ -1515,6 +1515,12 @@ class IndexOpsMixin:
         corresponding elements in `value` were inserted before the indices,
         the order of `self` would be preserved.
 
+        .. note::
+
+            The %(klass)s *must* be monotonically sorted, otherwise
+            wrong locations will likely be returned. Pandas does *not*
+            check this for you.
+
         Parameters
         ----------
         value : array_like
@@ -1540,6 +1546,7 @@ class IndexOpsMixin:
 
         See Also
         --------
+        sort_values
         numpy.searchsorted
 
         Notes
@@ -1578,6 +1585,13 @@ class IndexOpsMixin:
 
         >>> x.searchsorted(['bread'], side='right')
         array([3])
+
+        If the values are not monotonically sorted, wrong locations
+        may be returned:
+
+        >>> x = pd.Series([2, 1, 3])
+        >>> x.searchsorted(1)
+        0  # wrong result, correct would be 1
         """
 
     @Substitution(klass="Index")
--- a/pandas/tests/arrays/categorical/test_analytics.py
+++ b/pandas/tests/arrays/categorical/test_analytics.py
@@ -78,42 +78,36 @@ class TestCategoricalAnalytics:
         exp = Categorical(exp_mode, categories=categories, ordered=True)
         tm.assert_categorical_equal(res, exp)
 
-    def test_searchsorted(self):
+    def test_searchsorted(self, ordered_fixture):
         # https://github.com/pandas-dev/pandas/issues/8420
         # https://github.com/pandas-dev/pandas/issues/14522
 
-        c1 = Categorical(
+        cat = Categorical(
             ["cheese", "milk", "apple", "bread", "bread"],
             categories=["cheese", "milk", "apple", "bread"],
-            ordered=True,
+            ordered=ordered_fixture,
         )
-        s1 = Series(c1)
-        c2 = Categorical(
-            ["cheese", "milk", "apple", "bread", "bread"],
-            categories=["cheese", "milk", "apple", "bread"],
-            ordered=False,
-        )
-        s2 = Series(c2)
+        ser = Series(cat)
 
         # Searching for single item argument, side='left' (default)
-        res_cat = c1.searchsorted("apple")
+        res_cat = cat.searchsorted("apple")
         assert res_cat == 2
         assert is_scalar(res_cat)
 
-        res_ser = s1.searchsorted("apple")
+        res_ser = ser.searchsorted("apple")
         assert res_ser == 2
         assert is_scalar(res_ser)
 
         # Searching for single item array, side='left' (default)
-        res_cat = c1.searchsorted(["bread"])
-        res_ser = s1.searchsorted(["bread"])
+        res_cat = cat.searchsorted(["bread"])
+        res_ser = ser.searchsorted(["bread"])
         exp = np.array([3], dtype=np.intp)
         tm.assert_numpy_array_equal(res_cat, exp)
         tm.assert_numpy_array_equal(res_ser, exp)
 
         # Searching for several items array, side='right'
-        res_cat = c1.searchsorted(["apple", "bread"], side="right")
-        res_ser = s1.searchsorted(["apple", "bread"], side="right")
+        res_cat = cat.searchsorted(["apple", "bread"], side="right")
+        res_ser = ser.searchsorted(["apple", "bread"], side="right")
         exp = np.array([3, 5], dtype=np.intp)
         tm.assert_numpy_array_equal(res_cat, exp)
         tm.assert_numpy_array_equal(res_ser, exp)
@@ -121,22 +115,15 @@ class TestCategoricalAnalytics:
         # Searching for a single value that is not from the Categorical
         msg = r"Value\(s\) to be inserted must be in categories"
         with pytest.raises(KeyError, match=msg):
-            c1.searchsorted("cucumber")
+            cat.searchsorted("cucumber")
         with pytest.raises(KeyError, match=msg):
-            s1.searchsorted("cucumber")
+            ser.searchsorted("cucumber")
 
         # Searching for multiple values one of each is not from the Categorical
         with pytest.raises(KeyError, match=msg):
-            c1.searchsorted(["bread", "cucumber"])
+            cat.searchsorted(["bread", "cucumber"])
         with pytest.raises(KeyError, match=msg):
-            s1.searchsorted(["bread", "cucumber"])
-
-        # searchsorted call for unordered Categorical
-        msg = "Categorical not ordered"
-        with pytest.raises(ValueError, match=msg):
-            c2.searchsorted("apple")
-        with pytest.raises(ValueError, match=msg):
-            s2.searchsorted("apple")
+            ser.searchsorted(["bread", "cucumber"])
 
     def test_unique(self):
         # categories are reordered based on value when ordered=False
--- a/doc/source/whatsnew/v1.0.0.rst
+++ b/doc/source/whatsnew/v1.0.0.rst
@@ -173,7 +173,7 @@ Categorical
 
 - Added test to assert the :func:`fillna` raises the correct ValueError message when the value isn't a value from categories (:issue:`13628`)
 - Bug in :meth:`Categorical.astype` where ``NaN`` values were handled incorrectly when casting to int (:issue:`28406`)
--
+- :meth:`Categorical.searchsorted` and :meth:`CategoricalIndex.searchsorted` now work on unordered categoricals also (:issue:`21667`)
 -
 
 


jbrockmendel:
@topper-123 needs rebase

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -220,7 +220,8 @@ Bug Fixes
 Categorical
 ^^^^^^^^^^^
 
--
+- :meth:`Categorical.searchsorted` and :meth:`CategoricalIndex.searchsorted`
```
jbrockmendel: I think the standard practice is to keep this as one long line

pandas/core/base.py
```diff
@@ -1163,6 +1163,16 @@ def factorize(self, sort=False, na_sentinel=-1):
         corresponding elements in `value` were inserted before the indices,
         the order of `self` would be preserved.
 
+        .. note::
+
+            The %(klass)s *must* be monotonically sorted, else wrong
```
jbrockmendel: “Else wrong” is awkward wording

pandas/core/base.py
```diff
@@ -1199,7 +1210,7 @@ def factorize(self, sort=False, na_sentinel=-1):
         dtype: int64
 
         >>> x.searchsorted(4)
-        array([3])
+        array([3])  # Note: an array, not a scalar
```
jbrockmendel: Will this cause a doctest to fail?

pandas/tests/categorical/test_analytics.py
```diff
@@ -70,58 +70,46 @@ def test_mode(self, values, categories, exp_mode):
...
 
-        c1 = Categorical(['cheese', 'milk', 'apple', 'bread', 'bread'],
-                         categories=['cheese', 'milk', 'apple', 'bread'],
-                         ordered=True)
-        s1 = Series(c1)
-        c2 = Categorical(['cheese', 'milk', 'apple', 'bread', 'bread'],
-                         categories=['cheese', 'milk', 'apple', 'bread'],
-                         ordered=False)
-        s2 = Series(c2)
+        c = Categorical(['cheese', 'milk', 'apple', 'bread', 'bread'],
```
jbrockmendel: Mind calling these cat and ser or something like that.  1-letter names make grepping tough

pandas/core/base.py
topper-123: It does not fail, but is probably not needed. I'll just remove that comment.

jreback:
this needs evaluation
I don’t think we want to allow this
you can easily get incorrect results

topper-123:
I discuss some of my points in favor of this in #21667:

In ``searchsorted`` its already not possible to guard against incorrect result. Also letting this into the code base (.searchsorted on .codes) would just mimic existing ``sort_values`` behaviour (sort on .code):

```python
>>> c = pd.Categorical(['a', 'a', 'b', 'b', 'c'], categories=pd.Index(['b', 'a', 'c']))
>>> c.sort_values()
[b, b, a, a, c]
Categories (3, object): [b, a, c]
``` 

We can perhaps continue the discussion in #21667.

jorisvandenbossche:
I agree with @topper-123: it is indeed easy to get wrong results with searchsorted, but that is a general issue with that function, and not necessarily specific to `Categorical.searchsorted`. 
`searchsorted` requires the values to be sorted, but `ordered=True` is not necessary to sort a Categorical, so IMO should also not be needed for searchsorted

jreback:
``.sort_values()`` is a red-herring here, we are simply lexographically sorting which is expected. Its too easy to make mistakes with ``.searchsorted`` here. A doc warning is not great. I would leave this like it is.

topper-123:
```python
>>> c = pd.Categorical(['a', 'a', 'b', 'b', 'c'], categories=pd.Index(['b', 'a', 'c']))
>>> s1 = pd.Series(c)
>>> s1.is_monotonic_sorted
False
>>> s2 = 1.sort_values()
>>> s1.is_monotonic_sorted
True
>>> 
```

i.e. ``is_monotonic_sorted`` works on codes, like ``sort_values``. Seems logical to standardize ``searchsorted`` to work on codes also.

jreback:
> Yes, and this sorting will mean that the codes are sorted, so I don't understand why the above is a reason to not do this?

I think its pretty unituitive to sort by codes in an unordered categorical, we should actually disallow ``.sort_values()`` as well (or maybe allow it but sort by the lexographic categories themselves). 

e.g. we should raise just like we do for min/max.
```
In [1]: c = pd.Categorical(list('AABCD'))

In [2]: c.min()
TypeError: Categorical is not ordered for operation min
you can use .as_ordered() to change the Categorical to an ordered one


In [3]: c.sort_values()
Out[3]: 
[A, A, B, C, D]
Categories (4, object): [A, B, C, D]

```

jorisvandenbossche:
Being able to sort is very useful, we discussed this before, so I am -1 on removing that ability (but if you want further discuss this, I would open a new issue).

Note that, in the majority of the use cases for unordered categoricals, sorting by the codes or sorting lexicographically will be identical, exactly because we do sort the categories lexicographically on creation. 
So only when passing the categories manually and not setting `ordered=True`, this might be different.

topper-123:
I agree with @jorisvandenbossche.

IMO It is beneficial to be able find locations of groups of values unordered values, so ``lhs, rhs = c.searchssorted(val1, side='left'), c.searchsorted(val1, side='right')`` should be possible, also for unordered categoricals, because this the fastest way to look up locations (much faster than ``c[c==val1]``.

To do the above, unordered categoricals need to have ``sort_values``, ``is_monotonic*`` and ``searchsorted``...

mroeschke:
Looks like this stalled. In the meantime, this could use a rebase..

WillAyd:
Closing as stale. Ping if you'd like to continue

topper-123:
I would actually very much like to continue this, but was under the impression it wasn't going to get in. If you'll open it, I'll push a rebase.

jreback:
will have another look

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -0,0 +1,731 @@
+.. _whatsnew_0240:
+
```
jreback: this file is included somehow, pls revert

doc/source/whatsnew/v0.25.0.rst
```diff
@@ -127,7 +127,7 @@ Bug Fixes
 Categorical
 ^^^^^^^^^^^
 
--
+- :meth:`Categorical.searchsorted` and :meth:`CategoricalIndex.searchsorted` now work on unordered categoricals also (:issue:`21667`)
```
jreback: also -> as well

pandas/core/arrays/categorical.py
```diff
@@ -1374,19 +1374,15 @@ def memory_usage(self, deep=False):
...
+        value_codes = _get_codes_for_values(Series(value).values,
+                                            self.categories)
+        if -1 in value_codes:
             raise KeyError("Value(s) to be inserted must be in categories.")
+        value_codes = value_codes[0] if is_scalar(value) else value_codes
 
-        codes = codes[0] if is_scalar(value) else codes
-
-        return self.codes.searchsorted(codes, side=side, sorter=sorter)
+        return algorithms.searchsorted(self._codes, value_codes,
```
jreback: we don't have a guarantee that these codes are sorted right? so this seems likely to return an incorrect result if that is not the case

pandas/tests/arrays/categorical/test_analytics.py
```diff
@@ -71,55 +71,49 @@ def test_mode(self, values, categories, exp_mode):
...
+        res_cat = cat.searchsorted(['apple', 'bread'], side='right')
+        res_ser = ser.searchsorted(['apple', 'bread'], side='right')
         exp = np.array([3, 5], dtype=np.intp)
         tm.assert_numpy_array_equal(res_cat, exp)
         tm.assert_numpy_array_equal(res_ser, exp)
 
         # Searching for a single value that is not from the Categorical
-        pytest.raises(KeyError, lambda: c1.searchsorted('cucumber'))
-        pytest.raises(KeyError, lambda: s1.searchsorted('cucumber'))
+        with pytest.raises(KeyError):
```
jreback: add match= (if that is appropriate)

pandas/tests/arrays/categorical/test_analytics.py
```diff
@@ -71,55 +71,49 @@ def test_mode(self, values, categories, exp_mode):
...
 
-        c1 = Categorical(['cheese', 'milk', 'apple', 'bread', 'bread'],
-                         categories=['cheese', 'milk', 'apple', 'bread'],
-                         ordered=True)
-        s1 = Series(c1)
-        c2 = Categorical(['cheese', 'milk', 'apple', 'bread', 'bread'],
-                         categories=['cheese', 'milk', 'apple', 'bread'],
-                         ordered=False)
-        s2 = Series(c2)
+        cat = Categorical(['cheese', 'milk', 'apple', 'bread', 'bread'],
```
jreback: what about a test of the failing case, meaning an unsorted categorical

jreback (changes requested):
can you explain *why* you think its a good idea to enable this. This looks like it is just cause for trouble.

jreback:
if you can merge master and remove the 0.24.0 docs. pls ping.

topper-123:
I'll look into this over the weekend.

pandas/core/arrays/categorical.py
topper-123: I've added tests for this ``test_searchsorted_unordered``.

jreback:
@topper-123 can you merge master and we'll have another look see

topper-123:
I've rebased.

topper-123:
Yeah, I've rebased. The PR is quite old, so fixed some strange git issues by just merging the commits into one.

WillAyd:
Taking another look at this again. Read through history but I think I am also on the side of not doing this. Understood on the point about consistency with other aspects of the API but I personally find this confusing, and a doc note isn't a great guard against that.

So I think @jreback and I are -1 here, @topper-123 and @jorisvandenbossche are +1. Still accurate?

This is oldest PR in queue so would like to decide how to move forward so it isn't in limbo forever

jorisvandenbossche:
@WillAyd can you explain what exactly you find confusing? 
Note that the doc note is added to the generic `searchsorted` docstring, as this point is true for all dtypes (you always need to ensure yourself that the values are sorted when calling this method). There is nothing categorical specific about this gotcha with `searchsorted`. 

The only thing with Categoricals is that it can be a bit harder to know if your values are sorted or not. But, you *can* sort an unordered categorical with `sort_values`, you can check if they sorted with `is_monotonic`. 
And for `searchsorted` to work, the only requirement is sortedness, which an unordered categorical still has.

topper-123:
Agree with everything @jorisvandenbossche said.

I would like to add that ``searchsorted`` is used to get fast lookups, so for performance sensitive work, having ``searchsorted`` is necessary.

WillAyd:
I find  what is quoted in the docstring to be confusing:

```python
If the values are not monotonically sorted, wrong locations
   may be returned:

>>> x = pd.Series([2, 1, 3])
>>> x.searchsorted(1)
0  # wrong result, correct would be 1
```

I understand that this affects non-Categorical right now, but IMO this should be raising for non-monotonic instead of returning surprising results. I don't think expanding that coverage to non-sorted Categorical moves us in the right direction

jorisvandenbossche:
> I understand that this affects non-Categorical right now, but IMO this should be raising for non-monotonic instead of returning surprising results

Then maybe we could split those two issues? Let's open a separate issue to discuss raising for non-monotonic in general, and move forward with this PR making searchsorted available consistently across dtypes?

jorisvandenbossche:
Any more thoughts here?

jreback:
I guess this is fine, though IMHO this is scope creep for ordered categoricals. by-definition unordered ones are not ordered (well they *have* an order, but it is not exposed to the users per se). that being said, I guess for perf reasons this is desireable.

topper-123:
I've rebased.

WillAyd:
Thanks @topper-123 for the PR and @jorisvandenbossche for input

PR #21691, opened 2018-07-01
move and de-privatize _localize_pydatetime, move shift_day to liboffsets

jbrockmendel:
The only non-test place where tslib._localize_pydatetime is used is in tseries.offsets.  This moves that func to liboffsets, de-privatizes it, and moves tseris.offsets.shift_day up to liboffsets.

To avoid import proliferation this keeps localize_pydatetime in the tslib namespace so the tests can get at it.

Adds some typing to localize_pydatetime and fleshes out the docstring.

### DIFF
--- a/pandas/_libs/tslib.pyx
+++ b/pandas/_libs/tslib.pyx
@@ -54,6 +54,7 @@ from tslibs.nattype cimport checknull_with_nat, NPY_NAT
 from tslibs.timestamps cimport (create_timestamp_from_ts,
                                 _NS_UPPER_BOUND, _NS_LOWER_BOUND)
 from tslibs.timestamps import Timestamp
+from tslibs.offsets import localize_pydatetime  # noqa:F841
 
 cdef bint PY2 = str == bytes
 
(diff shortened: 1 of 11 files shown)


pandas/_libs/tslib.pyx
```diff
@@ -54,6 +54,7 @@ from tslibs.nattype cimport checknull_with_nat, NPY_NAT
 from tslibs.timestamps cimport (create_timestamp_from_ts,
                                 _NS_UPPER_BOUND, _NS_LOWER_BOUND)
 from tslibs.timestamps import Timestamp
+from tslibs.offsets import localize_pydatetime  # noqa:F841
```
jreback: would be ok with not doing this import and changing the tests

pandas/_libs/tslibs/offsets.pyx
```diff
@@ -494,6 +498,58 @@ class BaseOffset(_BaseOffset):
 # ----------------------------------------------------------------------
 # RelativeDelta Arithmetic
 
+cpdef inline datetime localize_pydatetime(datetime dt, object tz):
+    """
+    Take a datetime/Timestamp in UTC and localizes to timezone tz.
```
jreback: should this be in timezones.pyx?

jreback (changes requested):

pandas/_libs/tslib.pyx
jbrockmendel: After the current set of PRs I'm going to pitch an idea to 1) finish off `tslib` and 2) expose selected functions/classes in `tslibs.__init__`.  I'd rather avoid futzing with imports until then.

pandas/_libs/tslibs/offsets.pyx
jreback: ok, let's do that

jreback (changes requested):

pandas/_libs/tslib.pyx
jreback: ok that's fine

jreback (approved):

jreback:
thanks @jbrockmendel

PR #21692, opened 2018-07-01
move is_null_datetimelike to nattype for self-containment

jbrockmendel:
ATM `is_null_datetimelike` is used _only_ in `tslibs.period`.  The dependency structure is simplified nicely if we put that function is `tslibs.nattype` instead of in `_libs.missing`.

This also takes a couple of missing-themed functions out of inference.pyx

Finally, moves the libperiod block in setup.py to down with the rest of the tslibs blocks to restore alphabetical order.

### DIFF
--- a/pandas/_libs/missing.pxd
+++ b/pandas/_libs/missing.pxd
@@ -2,7 +2,6 @@
 # cython: profile=False
 
 from tslibs.nattype cimport is_null_datetimelike
-# pass-through so other modules can cimport is_null_datetimelike from  here
 
 cpdef bint checknull(object val)
 cpdef bint checknull_old(object val)
--- a/pandas/_libs/missing.pxd
+++ b/pandas/_libs/missing.pxd
@@ -1,6 +1,12 @@
 # -*- coding: utf-8 -*-
 # cython: profile=False
 
-cdef bint is_null_datetimelike(object val)
+from tslibs.nattype cimport is_null_datetimelike
+# pass-through so other modules can cimport is_null_datetimelike from  here
+
 cpdef bint checknull(object val)
 cpdef bint checknull_old(object val)
+
+cdef bint is_null_datetime64(v)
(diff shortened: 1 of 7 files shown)


pandas/_libs/missing.pxd
```diff
@@ -1,6 +1,12 @@
 # -*- coding: utf-8 -*-
 # cython: profile=False
 
-cdef bint is_null_datetimelike(object val)
+from tslibs.nattype cimport is_null_datetimelike
+# pass-through so other modules can cimport is_null_datetimelike from  here
```
jreback: why is this comment needed?  this is for api simplicity yes? if so i don’t think really necessary for a comment

jreback:
needs a rebase

pandas/_libs/tslibs/nattype.pyx
```diff
@@ -587,3 +588,28 @@ cdef inline bint checknull_with_nat(object val):
...
+
+    Returns
+    -------
+    null_datetimelike : bool
+    """
+    if util._checknull(val):
+        return True
+    elif val is NaT:
+        return True
+    elif util.is_timedelta64_object(val):
```
jorisvandenbossche: why import this from  `util`? Didn't you move it to `tslibs.missing` ?

jbrockmendel: Not sure I understand the question.  This PR is moving this function from `_libs.missing` to `tslibs.nattype` because doing so simplifies the dependency structure.

jorisvandenbossche: and by moving the function should not change the import?

jreback: these util functions of basically c versions of isinstance checks, so this is ok.

jreback:
thanks @jbrockmendel

PR #21693, opened 2018-07-01
use liboffsets to_offset function

jbrockmendel:


jreback:
thanks!

jbrockmendel:
Thanks for tracking this down.  A couple questions: is there anything weird about your Windows environment, e.g. running on top of cygwin or something?  Or more generally since I don't use Windows, any other reason why your system might be special?  Long-shot since you're already doing a lot here: can you reproduce these results on non-Windows?

If you re-run the same asv a few times, are the results pretty consistent?  Often asv results can have a lot of noise.

qwhelan:
- Nothing unusual from a Windows perspective: relatively clean Windows 10 install with benchmarks run under PowerShell. Only standard antivirus, with benchmarks excluded from AV checks.
- None that come to mind, actually did a fresh install a few months back due to upgrading the SSD and strictly used this machine for remote desktop until recently (so very little installed on it).
- Actually, it also looks like the trend is visible on my Ubuntu 18.04 machine, so seems like it might be a cross-platform issue. Haven't run a bisection yet, but could this evening.

And yes, sadly consistent.

setup.py
```diff
@@ -538,7 +538,8 @@ def pxd(name):
                      '_libs/tslibs/ccalendar',
                      '_libs/tslibs/timedeltas',
                      '_libs/tslibs/timezones',
-                     '_libs/tslibs/nattype'],
+                     '_libs/tslibs/nattype',
+                     '_libs/tslibs/offsets'],
```
jbrockmendel: @jreback I thought we had made the "pxdfiles" obsolete when moving to `cythonize`.  Am I remembering wrong?

jbrockmendel: Never mind, thought I was commenting on an open PR, not an old one.

PR #21690, opened 2018-07-01
Misc cleanups, test parametrization

jbrockmendel:
No coherent theme: remove an unused variable in inference.pyx, make some imports more direct, parametrize a big test, and move a tslib-specific test to test_tslibs

PR #21694, opened 2018-07-01
move _add__ and __sub__ to liboffsets, implement _Tick

jbrockmendel:
Last one for now.

Moving `__add__` and `__sub__` to liboffsets has been viable for a while now, just getting around to it.

Implements a dummy `_Tick` class that gets mixed in to `Tick`.  Then in `tslibs.period` we can replace `isinstance(obj, offsets.Tick)` checks with `isinstnace(obj, _Tick)`, obviating the need for the import of `tseries.offsets`.  This moves us closer to being able to import `Period` directly in `_libs.__init__`, which has been a long-time goal (see the comment there).  In conjunction with #21693 and #21690, I think that gets us the rest of the way there.

### DIFF
--- a/pandas/_libs/tslibs/offsets.pyx
+++ b/pandas/_libs/tslibs/offsets.pyx
@@ -378,8 +378,8 @@ class _BaseOffset(object):
         if isinstance(other, datetime):
             raise TypeError('Cannot subtract datetime from offset.')
         elif type(other) == type(self):
-            return self.__class__(self.n - other.n, normalize=self.normalize,
-                                  **self.kwds)
+            return type(self)(self.n - other.n, normalize=self.normalize,
+                              **self.kwds)
         else:  # pragma: no cover
             return NotImplemented
 
@@ -387,8 +387,8 @@ class _BaseOffset(object):
         return self.apply(other)
 
     def __mul__(self, other):
-        return self.__class__(n=other * self.n, normalize=self.normalize,
-                              **self.kwds)
+        return type(self)(n=other * self.n, normalize=self.normalize,
+                          **self.kwds)
 
     def __neg__(self):
         # Note: we are defering directly to __mul__ instead of __rmul__, as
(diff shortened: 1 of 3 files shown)


pandas/_libs/tslibs/offsets.pyx
```diff
@@ -491,6 +510,12 @@ class BaseOffset(_BaseOffset):
         return -self + other
 
 
+class _Tick(object):
+    # dummy class to mix into tseries.Tick so that in tslibs.period we can
```
jreback: can u use triple quotes here

jreback (changes requested):

pandas/_libs/tslibs/offsets.pyx
```diff
@@ -364,6 +364,25 @@ class _BaseOffset(object):
...
+        try:
+            return self.apply(other)
+        except ApplyTypeError:
+            return NotImplemented
+
+    def __sub__(self, other):
+        if isinstance(other, datetime):
+            raise TypeError('Cannot subtract datetime from offset.')
+        elif type(other) == type(self):
+            return self.__class__(self.n - other.n, normalize=self.normalize,
```
jreback: can you change to use ``type(self)`` here, I don't think we use the long form anywhere.

jreback (changes requested):
tiny comment ping on green.

pandas/_libs/tslibs/offsets.pyx
jbrockmendel: its used below in `__mul__`.  change it there too while I'm at it?

jreback: yeah might as well, its a bit more clean.

jreback (approved):
minor comments, ping on green.

jreback:
lgtm. ping on green.

jbrockmendel:
ping

jreback:
thank you sir!

PR #21700, opened 2018-07-01
implement tslibs.arraylike

jbrockmendel:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


pandas/_libs/tslib.pyx
```diff
@@ -37,6 +37,9 @@ from cython cimport Py_ssize_t
 import pytz
 UTC = pytz.utc
 
+from tslibs.arraylike import (  # noqa:F841
+    format_array_from_datetime, array_to_datetime, array_with_unit_to_datetime,
```
jreback: isn’t the point to not import back here, rather import directly to where it’s needed

jreback (changes requested):

pandas/_libs/tslib.pyx
jbrockmendel: Sure.  For the first pass I like to keep the diff small, but now I'll update.

jbrockmendel: Actually on second thought there are a _lot_ of usages of these I'd rather not update.  I think the thing to do is leave this where it is for now and put it in the `tslibs` namespace a few PRs from now when we finish transitioning `tslib`

pandas/_libs/tslibs/arraylike.pyx
```diff
@@ -0,0 +1,717 @@
+# -*- coding: utf-8 -*-
+
+cimport numpy as cnp
+from numpy cimport int64_t, ndarray, float64_t
+import numpy as np
```
jreback: so some of this should be put in conversion. or is there a reason you are not doing that? i would like to shy away from creating a new module unless it has a clear purpose. can you move what is logical first.

pandas/_libs/tslibs/arraylike.pyx
```diff
@@ -0,0 +1,717 @@
+# -*- coding: utf-8 -*-
+
```
jorisvandenbossche: if you end up creating this file, can you take the opportunity of the refactor to add a module docstring here to better document what / what type of functionality is included in this file? 

But I agree with Jeff that from a quick view some of those functions seems to be about parsing / conversion?

jorisvandenbossche (changes requested):

jbrockmendel:
Woops, closed the wrong one.

jbrockmendel:
Closing as non-actionable for now

PR #21702, opened 2018-07-02
CLN: to_datetime internals

mroeschke:
- [x] tests passed

The internals of `to_datetime` is getting a bit unwieldy, so split out the `_convert_listlike` logic and origin shifting logic to ` _convert_listlike_datetime` and `adjust_to_origin` methods respectively outside of `to_datetime`. The logic was not changed.


jreback:
lgtm. can you do a perf check to make sure nothing changed (as the caching logic was lightly touched here)

pandas/core/tools/datetimes.py
```diff
@@ -38,7 +39,7 @@ def _guess_datetime_format_for_array(arr, **kwargs):
         return _guess_datetime_format(arr[non_nan_elements[0]], **kwargs)
 
 
-def _maybe_cache(arg, format, cache, tz, convert_listlike):
```
jorisvandenbossche: The tz here was not used?

mroeschke: tz was passed into the `convert_listlike` function further down, but now I am embedding it into `convert_listlike` with `functools.partial` in `to_datetime`

pandas/core/tools/datetimes.py
```diff
@@ -140,6 +139,226 @@ def _return_parsed_timezone_results(result, timezones, box, tz):
...
+
+    except ValueError as e:
+        try:
+            values, tz = conversion.datetime_to_datetime64(arg)
+            return DatetimeIndex._simple_new(values, name=name, tz=tz)
+        except (ValueError, TypeError):
+            raise e
+
+
+def _adjust_to_origin(arg, origin, unit):
```
jbrockmendel: +1 for separating this out

pandas/core/tools/datetimes.py
```diff
@@ -140,6 +139,226 @@ def _return_parsed_timezone_results(result, timezones, box, tz):
...
+    ----------
+    arg : list, tuple, ndarray, Series, Index
+        date to be adjusted
+    origin : 'julian' or Timestamp
+        origin offset for the arg
+    unit : string
+        passed unit from to_datetime, must be 'D'
+    Returns
+    -------
+    ndarray of adjusted dates
```
jbrockmendel: Is it necessarily an `ndarray`?  Couldn't it be a `Timestamp`?

need newline before `Returns`

mroeschke: Good catch, yeah this can be a scalar value. Will fix that tonight.

mroeschke:
My asv setup is still a little broken, but here's a benchmark showing no performance hit (with cache)

```
In [1]: from pandas import *

In [3]: N = 100

In [4]: dup_string_with_tz = ['2000-02-11 15:00:00-0800'] * N

# this branch
In [5]: %timeit to_datetime(dup_string_with_tz, cache=True)
1.31 ms ± 44.5 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [6]: %timeit to_datetime(dup_string_with_tz, cache=False)
2.3 ms ± 41.2 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# Master
In [4]: %timeit to_datetime(dup_string_with_tz, cache=True)
1.29 ms ± 13.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [5]: %timeit to_datetime(dup_string_with_tz, cache=False)
2.31 ms ± 68.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

jorisvandenbossche (approved):

jreback:
thanks @mroeschke happily take refactorings to clean up things / make more readable

PR #21710, opened 2018-07-02
remove unused cimport, parametrize test

jbrockmendel:


pandas/_libs/writers.pyx
```diff
@@ -16,8 +16,6 @@ cimport numpy as cnp
 from numpy cimport ndarray, uint8_t
 cnp.import_array()
 
-cimport util
```
jreback: do you have a linter that catches this?

pandas/tests/sparse/test_libsparse.py
```diff
@@ -596,22 +596,10 @@ def _check_case(xloc, xlen, yloc, ylen, eloc, elen):
...
-
-# too cute? oh but how I abhor code duplication
-check_ops = ['add', 'sub', 'mul', 'truediv', 'floordiv']
-
-
-def make_optestf(op):
-    def f(self):
-        sparse_op = getattr(splib, 'sparse_%s_float64' % op)
-        python_op = getattr(operator, op)
+    @pytest.mark.parametrize('opname',
```
jreback: worth making a fixture in conftest?

pandas/_libs/writers.pyx
jbrockmendel: We've discussed this a couple times and haven't found a good solution for this.  It's on my list to bring up with the group  at the sprint.

pandas/tests/sparse/test_libsparse.py
jbrockmendel: I think this is a pretty specific set of strings that wouldn't get used often.  But I didn't look at this closely, just saw that the existing formatting was crying out to be updated.

pandas/_libs/writers.pyx
jreback: right, just asking if you found a soln!

pandas/tests/sparse/test_libsparse.py
jreback: ok, that's fine

gfyoung (approved):

PR #21717, opened 2018-07-03
TST: Parametrize tests in tools/test_numeric.py

gfyoung:
Title is self-explanatory.

### DIFF
--- a/pandas/tests/tools/test_numeric.py
+++ b/pandas/tests/tools/test_numeric.py
@@ -162,12 +162,17 @@ class TestToNumeric(object):
         tm.assert_series_equal(res, expected)
 
     def test_type_check(self):
-        # GH 11776
-        df = pd.DataFrame({'a': [1, -3.14, 7], 'b': ['4', '5', '6']})
-        with tm.assert_raises_regex(TypeError, "1-d array"):
+        # see gh-11776
+        df = pd.DataFrame({"a": [1, -3.14, 7], "b": ["4", "5", "6"]})
+        error_ctx = tm.assert_raises_regex(TypeError, "1-d array")
+
+        # Check default parameters.
+        with error_ctx:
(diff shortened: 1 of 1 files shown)


pandas/tests/tools/test_numeric.py
```diff
@@ -227,17 +232,17 @@ def test_str(self):
         res = pd.to_numeric(idx.values)
         tm.assert_numpy_array_equal(res, exp)
 
-    def test_datetimelike(self):
-        for tz in [None, 'US/Eastern', 'Asia/Tokyo']:
-            idx = pd.date_range('20130101', periods=3, tz=tz, name='xxx')
-            res = pd.to_numeric(idx)
-            tm.assert_index_equal(res, pd.Index(idx.asi8, name='xxx'))
+    @pytest.mark.parametrize("tz", [None, "US/Eastern", "Asia/Tokyo"])
```
jschendel: Should the common `tz_naive_fixture` be used here?  It has a few extra timezones, but I'm not sure if it tests anything extra/if there's a benefit to including the extra ones:

https://github.com/pandas-dev/pandas/blob/7cd2679982138e1e7f9f1a759827647e6e521d43/pandas/conftest.py#L221-L230

jreback: yes ideally we re-use our fixtures as much as possible

jreback (changes requested):

pandas/tests/tools/test_numeric.py
gfyoung: That makes sense to me.  Done.

gfyoung:
@jreback @jschendel : Comments addressed, and all is still green.  PTAL.

jreback (approved):

jreback:
thanks @gfyoung very nice!

PR #21721, opened 2018-07-04
implement _local_timestamps in DatetimeArrayMixin

jbrockmendel:
Broken out of #21715 because it is a blocker for a handful of orthogonal follow-ons.

### DIFF
--- a/pandas/core/arrays/datetimes.py
+++ b/pandas/core/arrays/datetimes.py
@@ -2,9 +2,10 @@
 import warnings
 
 import numpy as np
+from pytz import utc
 
 from pandas._libs.tslib import Timestamp, NaT, iNaT
-from pandas._libs.tslibs import timezones
+from pandas._libs.tslibs import conversion, timezones
 
 from pandas.util._decorators import cache_readonly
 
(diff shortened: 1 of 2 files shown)


pandas/core/indexes/datetimes.py
```diff
@@ -603,14 +603,8 @@ def _local_timestamps(self):
...
         else:
-            values = self.asi8
-            indexer = values.argsort()
-            result = conversion.tz_convert(values.take(indexer), utc, self.tz)
-
-            n = len(indexer)
-            reverse = np.empty(n, dtype=np.int_)
-            reverse.put(indexer, np.arange(n))
-            return result.take(reverse)
+            # fall back to non-optimized implementation
```
jreback: I would remove this comment

pandas/core/arrays/datetimes.py
```diff
@@ -108,3 +109,16 @@ def _sub_datelike_dti(self, other):
             mask = (self._isnan) | (other._isnan)
             new_values[mask] = iNaT
         return new_values.view('timedelta64[ns]')
+
+    # -----------------------------------------------------------------
+    # Timezone Conversion and Localization Methods
+
+    def _local_timestamps(self):
+        values = self.asi8
```
jreback: add a doc-string here (e.g. what are the return guarantees here)

jreback (changes requested):
looks good, just some doc comments

pandas/core/arrays/datetimes.py
```diff
@@ -108,3 +109,22 @@ def _sub_datelike_dti(self, other):
...
+    def _local_timestamps(self):
+        """
+        Convert to an i8 (unix-like nanosecond timestamp) representation
+        while keeping the local timezone and not using UTC.
+        This is used to calculate time-of-day information as if the timestamps
+        were timezone-naive.
+        """
+        values = self.asi8
+        indexer = values.argsort()
+        result = conversion.tz_convert(values.take(indexer), utc, self.tz)
```
jbrockmendel: Does this only make sense if `tz_convert` requires its input to be sorted?  I don't think it does.

jreback: yeah not sure why that indexer is there

jbrockmendel: In the next pass I'll see if this can be simplified.

jreback (approved):

PR #21723, opened 2018-07-04
Code deduplication in tz_convert and related

jbrockmendel:
`tz_convert_tzlocal_to_utc` and `tz_convert_utc_to_tzlocal` are identical except for a `+` changed to a `-`.  This implements `_tz_convert_tzlocal_utc` which has an argument that flips the sign.  IIUC because it is inlined, that check should be free.

An upcoming PR will do the same thing for the `get_dst_info` cases in `tz_convert` and `tz_convert_single`, at which point a bunch of duplicate code will be able to be removed.  To make it more obvious how that will work, this rearranges the to-be-de-duplicated code blocks to use (near) symmetric condition-ordering and indentation.


pandas/_libs/tslibs/conversion.pyx
```diff
@@ -1205,7 +1187,6 @@ def is_date_array_normalized(ndarray[int64_t] stamps, tz=None):
         for i in range(n):
             # Adjust datetime64 timestamp, recompute datetimestruct
             pos = trans.searchsorted(stamps[i]) - 1
-            inf = tz._transition_info[pos]
```
jbrockmendel: unrelated to the rest of the PR

PR #21724, opened 2018-07-04
CLN: Delay converting bins to datetimelike until necessary

mroeschke:
- [x] tests added / passed

xref #20956, `q/cut` now returns datetimelike bins when `retbins=True`. Instead of always trying to convert bins to datetimelike (even when `retbins=False`), moving the conversion to `_postprocess_for_cut` until after `retbins=True` is checked

cc @jreback


jreback:
thanks!

PR #21727, opened 2018-07-04
Implement _tz_convert_tzlocal_utc for de-duplication

jbrockmendel:
There is a bunch of code in conversion (and a bit in resolution, period, and tslib) that can be de-duplicated.  This is one part of ~5 steps in that process.

### DIFF
--- a/pandas/_libs/tslibs/conversion.pyx
+++ b/pandas/_libs/tslibs/conversion.pyx
@@ -537,7 +537,7 @@ cdef inline void localize_tso(_TSObject obj, tzinfo tz):
     elif obj.value == NPY_NAT:
         pass
     elif is_tzlocal(tz):
-        local_val = tz_convert_utc_to_tzlocal(obj.value, tz)
+        local_val = _tz_convert_tzlocal_utc(obj.value, tz, to_utc=False)
         dt64_to_dtstruct(local_val, &obj.dts)
     else:
         # Adjust datetime64 timestamp, recompute datetimestruct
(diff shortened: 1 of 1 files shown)


pandas/_libs/tslibs/conversion.pyx
```diff
@@ -1205,7 +1187,6 @@ def is_date_array_normalized(ndarray[int64_t] stamps, tz=None):
         for i in range(n):
             # Adjust datetime64 timestamp, recompute datetimestruct
             pos = trans.searchsorted(stamps[i]) - 1
-            inf = tz._transition_info[pos]
```
jbrockmendel: unrelated to the rest of the PR

pandas/_libs/tslibs/conversion.pyx
```diff
@@ -608,32 +608,33 @@ cpdef inline datetime localize_pydatetime(datetime dt, object tz):
 # ----------------------------------------------------------------------
 # Timezone Conversion
 
-cdef inline int64_t tz_convert_tzlocal_to_utc(int64_t val, tzinfo tz):
+cdef inline int64_t _tz_convert_tzlocal_utc(int64_t val, tzinfo tz,
+                                            bint to_utc=True):
```
jreback: looks good. can you add what this function does in doc-string. and mark as this is private.

jreback (changes requested):
minor comments. prefer as a non-private, though if its*only* for this module then private is ok.

pandas/_libs/tslibs/conversion.pyx
jbrockmendel: Mind if I put the docsting in the follow-up?  There’s a couple more coming up in the next few hours.

jbrockmendel:
Yes, the private function is just for this module.

jreback (approved):

pandas/_libs/tslibs/conversion.pyx
jreback: sure

PR #21730, opened 2018-07-04
implement _tz_convert_dst for de-duplication

jbrockmendel:
Analogous to #21727, this handles the `get_dst_info` case whereas that handled the `tzlocal` case.

Fleshes out docstring as discussed in #21727 

All the other uses of `get_dst_info` are handled _slightly_ differently.  Need to figure out whether those are intentional to see if more de-duplication (or perf) is available.

### DIFF
--- a/pandas/_libs/tslibs/conversion.pyx
+++ b/pandas/_libs/tslibs/conversion.pyx
@@ -643,17 +643,16 @@ cdef inline int64_t[:] _tz_convert_dst(ndarray[int64_t] values, tzinfo tz,
         # if all NaT, return all NaT
         return values
 
-    #posn = trans.searchsorted(tt, side='right')
+    posn = trans.searchsorted(tt, side='right')
 
-    #j = 0
+    j = 0
     for i in range(n):
         v = values[i]
         if v == NPY_NAT:
             result[i] = v
         else:
-            pos = trans.searchsorted(v) - 1
-            # pos = posn[j] - 1
-            #j += 1
+            pos = posn[j] - 1
+            j += 1
             if pos < 0:
                 raise ValueError('First time before start of DST info')
             result[i] = v - deltas[pos]
--- a/pandas/_libs/tslibs/conversion.pyx
+++ b/pandas/_libs/tslibs/conversion.pyx
@@ -608,6 +608,59 @@ cpdef inline datetime localize_pydatetime(datetime dt, object tz):
 # ----------------------------------------------------------------------
 # Timezone Conversion
 
+cdef inline int64_t[:] _tz_convert_dst(ndarray[int64_t] values, tzinfo tz,
+                                       bint to_utc=True):
+    """
+    tz_convert for non-UTC non-tzlocal cases where we have to check
+    DST transitions pointwise.
+
+    Parameters
+    ----------
+    values : ndarray[int64_t]
+    tz : tzinfo
+    to_utc : bool
+        True if converting _to_ UTC, False if converting _from_ utc
+
+    Returns
+    -------
+    result : ndarray[int64_t]
+    """
+    cdef:
+        Py_ssize_t n = len(values)
+        Py_ssize_t i, j, pos
+        ndarray[int64_t] result = np.empty(n, dtype=np.int64)
+        ndarray[int64_t] tt, trans, deltas
+        ndarray[Py_ssize_t] posn
+        int64_t v
+
+    trans, deltas, typ = get_dst_info(tz)
+    if not to_utc:
+        # We add `offset` below instead of subtracting it
+        deltas = -1 * deltas
+
+    tt = values[values != NPY_NAT]
+    if not len(tt):
+        # if all NaT, return all NaT
+        return values
+
+    #posn = trans.searchsorted(tt, side='right')
+
+    #j = 0
+    for i in range(n):
+        v = values[i]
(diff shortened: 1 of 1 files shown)


pandas/_libs/tslibs/conversion.pyx
```diff
@@ -672,6 +729,7 @@ cpdef int64_t tz_convert_single(int64_t val, object tz1, object tz2):
         Py_ssize_t pos
         int64_t v, offset, utc_date
         pandas_datetimestruct dts
+        ndarray[int64_t] arr  # TODO: Is there a lighter-weight way to do this?
```
jreback: ``int64_t[:] arr`` is a memory view, should work

pandas/_libs/tslibs/conversion.pyx
```diff
@@ -683,29 +741,23 @@ cpdef int64_t tz_convert_single(int64_t val, object tz1, object tz2):
...
-    offset = deltas[pos]
-    return utc_date + offset
+    else:
+        # Convert UTC to other timezone
+        arr = np.array([utc_date])
+        # Note: at least with cython 0.28.3, doing a looking `[0]` in the next
+        # line is sensitive to the declared return type of _tz_convert_dst;
+        # if it is declared as returning ndarray[int64_t], a compile-time error
+        # is raised.
+        return _tz_convert_dst(arr, tz2, to_utc=False)[0]
```
jreback: I think you could do an assignment here that then specifies the type

liek

```
arr = _tz_convert.....
cdef int64_t result = arr[0]
return result
```

pandas/_libs/tslibs/conversion.pyx
```diff
@@ -608,9 +608,66 @@ cpdef inline datetime localize_pydatetime(datetime dt, object tz):
...
+    if not to_utc:
+        # We add `offset` below instead of subtracting it
+        deltas = -1 * deltas
+
+    tt = values[values != NPY_NAT]
+    if not len(tt):
+        # if all NaT, return all NaT
+        return values
+
+    posn = trans.searchsorted(tt, side='right')
```
jbrockmendel: The other places where `get_dst_info` is used don't do this `tt` thing.

`_normalize_local`, `period.localize_dt64_arr_to_period`, `resolution._reso_local` all do:

```
trans, deltas, typ = get_dst_info(tz)
_pos = trans.searchsorted(stamps, side='right') - 1
 if _pos.dtype != np.int64:
        _pos = _pos.astype(np.int64)
pos = _pos
```

`is_date_array_normalized` and `tslib.ints_to_pydatetime` makes `searchsorted` calls pointwise

```
trans, deltas, typ = get_dst_info(tz)
for i in range(n):
```

I'm checking to see if anything breaks if we change it here to match the others.

Perf Considerations:
- The `tt` thing here presumably improves perf in cases with many `NaT`s
- Without the `values != NPY_NAT` check we could pass `values` as an `int64_t[:]` which I _think_ would be more performant.
- There is a `cnp.PY_ArraySearchSorted` that could be used to avoid reaching into python-space, but apparently there is a bug in cython's numpy.pxd that makes this non-trivial (https://stackoverflow.com/questions/28184211/calling-pyarray-searchsorted-from-cython-3-or-4-arguments?rq=1)

pandas/_libs/tslibs/conversion.pyx
jbrockmendel: why does that help in this context?

jbrockmendel: Ahh gotcha.  Fortunately the version here where _tz_convert_dst returns a memoryview is preferred anyway.  I could remove the comment, but it did take me a few rounds to figure out that was where the problem was coming from.

pandas/_libs/tslibs/conversion.pyx
jreback: hmm, ok to make that change here if you can

pandas/_libs/tslibs/conversion.pyx
jreback: > Without the values != NPY_NAT check we could pass values as an int64_t[:] which I think would be more performant.

why? NPY_NAT is i8

jreback: not sure why you can search an array of timestamps to searchsorted here?

jbrockmendel: > why? NPY_NAT is i8

AFAICT `int64_t[:]` doesn't support `__eq__`.  You can basically iterate through it and little else.

jbrockmendel: > not sure why you can search an array of timestamps to searchsorted here?

I don't understand the question.  `trans.searchsorted(x)` assumes (doesn't check) that `trans` is sorted, doesn't care about `x` (which can be array or scalar)

pandas/_libs/tslibs/conversion.pyx
jbrockmendel: > hmm, ok to make that change here if you can

Not an option yet.  To do this we need to get rid of the `values != NPY_NAT` check in `_tz_convert_dst`, which entails a non-trivial redesign of that function.  Need to a) make sure it doesn't break things and b) check perf.

ATM the options are either wrap in a 1-element array or keep duplicated code.

PR #21738, opened 2018-07-04
Implement tslibs.__init__, move easy bits of tslib

jbrockmendel:
Easy answers to where `tslib.ints_to_pytimedelta` and `tslib.normalize_date` belong, so this moves those.

Starts populating the `tslibs.__init__` namespace; leaves updating imports for later.

Cleans up a couple of unnecessary imports elsewhere.


pandas/_libs/tslibs/__init__.py
```diff
@@ -1,2 +1,8 @@
 # -*- coding: utf-8 -*-
-# cython: profile=False
+# flake8: noqa
+
+from .conversion import normalize_date, localize_pydatetime, tz_convert_single
+from .nattype import NaT, iNaT
```
jreback: next PR should have a test that exercises what is in this API, name it test_api.py and check that the namespace is locked down)

pandas/_libs/tslibs/conversion.pyx
```diff
@@ -1025,6 +1025,35 @@ cdef inline str _render_tstamp(int64_t val):
 # ----------------------------------------------------------------------
 # Normalization
 
+
+def normalize_date(object dt):
+    """
```
jreback: we ought to fix these names (prob rename date_normalize) in future PR

pandas/_libs/tslibs/conversion.pyx
```diff
@@ -1025,6 +1025,35 @@ cdef inline str _render_tstamp(int64_t val):
...
+    Normalize datetime.datetime value to midnight. Returns datetime.date as a
+    datetime.datetime at midnight
+
+    Parameters
+    ----------
+    dt : date, datetime, or Timestamp
+
+    Returns
+    -------
+    normalized : datetime.datetime or Timestamp
```
jreback: add a raises section

jreback (approved):
going to merge, but pls followup the comments in next PR

PR #21739, opened 2018-07-05
CLN/TST: Cleanup tests in test_ujson.py

gfyoung:
* Actually test the outputs
* Add more parameterization
* Take advantage of `pytest` features, such as exception contexts

This test module really needed an update 😄

### DIFF
--- a/pandas/tests/io/json/test_ujson.py
+++ b/pandas/tests/io/json/test_ujson.py
@@ -6,6 +6,7 @@ except ImportError:
     import simplejson as json
 import math
 import pytz
+import locale
 import pytest
 import time
 import datetime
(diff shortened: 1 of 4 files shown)


pandas/tests/io/json/test_ujson.py
```diff
@@ -40,116 +68,113 @@ def test_encodeDecimal(self):
...
         decoded = ujson.decode(encoded)
         assert sut == decoded
 
-    def test_encodeNonCLocale(self):
+    def test_encode_non_c_locale(self):
         import locale
-        savedlocale = locale.getlocale(locale.LC_NUMERIC)
+        saved_locale = locale.getlocale(locale.LC_NUMERIC)
+
         try:
```
jreback: i think we have a contest manager for locale setting

gfyoung: * We do, but it doesn't do the `Exception` checking that I think this test wants to do.
* I think I would want to tap into `_can_set_locale` instead.  Any reason why I shouldn't utilize into this private method?

jreback: not familiar with that
but happy to use existing code (so u know why it’s private?)

gfyoung:
@jreback : Everything is green.  PTAL.

jreback:
thanks @gfyoung nice cleanups! keep em coming!

PR #21740, opened 2018-07-05
CLN: Remove Legacy MultiIndex Index Compatibility

### LINKED ISSUE #21654: TST/CLN: remove MultiIndex compat
we have these 2 pickles

```
(pandas) bash-3.2$ ls pandas/tests/indexes/data/
mindex_073.pickle       multiindex_v1.pickle
```

should blow these away and the older MultiIndex formats. These are really really old.

note will conflict slightly with #21514 (IOW path will be different after)

elmq0022:
- [x] closes #21654 
- [x] tests added / passed
- [x] what's new - removing legacy MultiIndex internal formats


pandas/tests/indexes/multi/test_conversion.py
```diff
@@ -87,46 +85,6 @@ def test_to_hierarchical():
     assert result.names == index.names
 
 
-@pytest.mark.skipif(PY3, reason="testing legacy pickles not support on py3")
-def test_legacy_pickle(datapath):
-
-    path = datapath('indexes', 'multi', 'data', 'multiindex_v1.pickle')
-    obj = pd.read_pickle(path)
```
jreback: can u also remove the legacy code itself from MultiIndex

jreback (changes requested):
pls add a whatsnew note - removing legacy MuktiIndex internal formats

pandas/tests/indexes/multi/test_conversion.py
elmq0022: So, I came to the conclusion that I can just delete the `__setstate__` method as there is no associated `__getstate__` method.  I hope this approach is correct.

jreback (changes requested):

elmq0022:
I am not sure what happened with the circleci build.  I could use some help with the postmortem as I haven't dealt with circleci directly before.

jbrockmendel:
Looks like circleCI fail may be unrelated.  Pushing a new commit will have it try again.

elmq0022:
Thanks.  The queue seemed really full that day.  Will give it another go.

jreback (changes requested):

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -444,7 +444,7 @@ Missing
 MultiIndex
 ^^^^^^^^^^
 
--
+- Removing legacy MultiIndex internal formats
```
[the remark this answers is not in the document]
elmq0022: Okay, will add the issue number tonight and resubmit.  Looks like the first failure was something on AppVeyor's side.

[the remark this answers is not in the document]
jorisvandenbossche: Also, the above message is unclear to me what it is about (as an external reader, didn't look at the rest of the PR)

elmq0022:
@jreback I rebased after the latest merge should be good to go.  Thanks!

pandas/tests/indexes/multi/test_analytics.py
```diff
@@ -15,8 +15,8 @@ def test_shift(idx):
     pytest.raises(NotImplementedError, idx.shift, 1, 2)
 
 
-def test_bounds(idx):
-    idx._bounds
+# def test_bounds(idx):
```
jreback: remote this entirely

jreback (changes requested):
minor comment. ping on green.

elmq0022:
Nice catch.  Thanks!  Will get to this tonight.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -513,7 +513,7 @@ Missing
 MultiIndex
 ^^^^^^^^^^
 
--
+- Removed legacy internal formats (:issue:`21654`)
```
jorisvandenbossche: Putting it here again: the above message is unclear to me what it is about. 

Either this change is not relevant for the user (eg only internal clean-up), and not whatsnew notice is needed. Or it is, but then needs to be clarified (eg do we drop support for very old pickle files?)

jorisvandenbossche (changes requested):

doc/source/whatsnew/v0.24.0.txt
jorisvandenbossche: if that is the case, I would maybe just remove the what's new notice? (since the 0.13 threshold is already documented I assume?)

jreback: yeah I don't think this has any user visible effects, but still having a whatsnew note just documents what we did. I would expand the note slightly but leave it.

jorisvandenbossche: We don't always add a whatsnew note for internal code clean-up (if it has no user visible change), so I still don't think it is needed here. Anyhow, if we keep one, it should be comprehensible for users (the current one is not even comprehensible for me without looking at the issue)

elmq0022:
Okay thanks.  It's a fair point.  I will try to summarize the issue better
in the what's new text and resubmit.

elmq0022:
@jreback everything is green.  Thank you for all your help on refactoring this MultiIndex text suite.

jorisvandenbossche (approved):
Thanks for the update, that's clear now!

elmq0022:
No problem. I'm glad you brought up the issue.  Thanks!

PR #21782, opened 2018-07-07
REF: move more methods to EAMixins

jbrockmendel:
With the exception of the comparison ops, this is the last stuff that can be moved before I have to implement the constructors (and hence a buttload of tests).  Let's get this out of the way first.

(For the comparison ops there is some super jujitsu that needs to be un-wound, so will do that separately)

### DIFF
--- a/pandas/core/arrays/datetimelike.py
+++ b/pandas/core/arrays/datetimelike.py
@@ -2,7 +2,7 @@
 
 import numpy as np
 
-from pandas._libs import iNaT, NaT
+from pandas._libs import lib, iNaT, NaT
 from pandas._libs.tslibs.timedeltas import delta_to_nanoseconds
 from pandas._libs.tslibs.period import (
     DIFFERENT_FREQ_INDEX, IncompatibleFrequency)
@@ -33,6 +33,12 @@ class DatetimeLikeArrayMixin(object):
         """
         raise com.AbstractMethodError(self)
 
+    def _box_values(self, values):
+        """
+        apply box func to passed values
+        """
+        return lib.map_infer(values, self._box_func)
+
     def __iter__(self):
         return (self._box_func(v) for v in self.asi8)
 
@@ -45,6 +51,9 @@ class DatetimeLikeArrayMixin(object):
     def asi8(self):
         # do not cache or you'll create a memory leak
         return self.values.view('i8')
+
+    def __len__(self):
+        return len(self._data)
 
     # ------------------------------------------------------------------
     # Null Handling
--- a/pandas/core/arrays/datetimes.py
+++ b/pandas/core/arrays/datetimes.py
@@ -107,6 +107,32 @@ class DatetimeArrayMixin(DatetimeLikeArrayMixin):
         warnings.warn(msg, FutureWarning, stacklevel=2)
         self.freq = value
 
+    # ----------------------------------------------------------------
+    # Array-like Methods
+
+    def __iter__(self):
+        """
+        Return an iterator over the boxed values
+
+        Yields
+        -------
+        tstamp : Timestamp
+        """
+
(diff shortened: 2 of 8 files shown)


pandas/core/arrays/datetimes.py
```diff
@@ -107,6 +107,32 @@ def offset(self, value):
...
+    # ----------------------------------------------------------------
+    # Array-like Methods
+
+    def __iter__(self):
+        """
+        Return an iterator over the boxed values
+
+        Yields
+        -------
+        tstamp : Timestamp
```
jbrockmendel: changed from existing version on DatetimeIndex where it says 

```
        Returns
        -------
        Timestamps : ndarray
```

pandas/core/arrays/timedelta.py
```diff
@@ -48,9 +60,53 @@ def _sub_datelike(self, other):
...
+    def total_seconds(self):
+        """
+        Return total duration of each element expressed in seconds.
+
+        This method is available directly on TimedeltaArray, TimedeltaIndex
+        and on Series containing timedelta values under the ``.dt`` namespace.
+
+        Returns
+        -------
+        seconds : ndarray, Float64Index, or Series
```
jbrockmendel: updated docstring to include ndarray

pandas/core/indexes/period.py
```diff
@@ -702,16 +696,7 @@ def _add_delta(self, other):
...
-
-        asi8 = self.asi8
-        new_data = asi8 - other.ordinal
-        new_data = np.array([self.freq * x for x in new_data])
-
-        if self.hasnans:
-            new_data[self._isnan] = tslib.NaT
+        new_data = PeriodArrayMixin._sub_period(self, other)
 
         # TODO: Should name=self.name be passed here?
```
jbrockmendel: I _think_ the answer to this is a) yes, but b) if we remove this method entirely then when the inherited method is called from `__sub__` it will correctly get wrapped.

pandas/core/arrays/datetimelike.py
```diff
@@ -33,6 +33,12 @@ def _box_func(self):
         """
         raise com.AbstractMethodError(self)
 
+    def _box_values(self, values):
```
jreback: so for future, any methods that we implement in the sub classes but not here, should be an AbstractMethodError (not specific to this comment, just generally)

pandas/core/arrays/timedelta.py
```diff
@@ -48,9 +60,53 @@ def _sub_datelike(self, other):
         raise TypeError("cannot subtract a datelike from a {cls}"
                         .format(cls=type(self).__name__))
 
+    def _evaluate_with_timedelta_like(self, other, op):
+        if isinstance(other, ABCSeries):
+            # GH#19042
+            return NotImplemented
+
+        opstr = '__{opname}__'.format(opname=op.__name__).replace('__r', '__')
```
jreback: we ought to have a method to do exactly this in the base ExtensionOpsMixin (but for later cleanup)

pandas/core/arrays/timedelta.py
```diff
@@ -48,9 +60,53 @@ def _sub_datelike(self, other):
...
+        """
+        Return total duration of each element expressed in seconds.
+
+        This method is available directly on TimedeltaArray, TimedeltaIndex
+        and on Series containing timedelta values under the ``.dt`` namespace.
+
+        Returns
+        -------
+        seconds : ndarray, Float64Index, or Series
+            When the calling object is a TimedeltaArray, the return type
```
jreback: can you make this in to a list of these return values

jreback (approved):
minor comments, either  or in a followup is ok. merge away

Issue #21793, opened 2018-07-07
BUG: PeriodIndex comparisons break on listlike


jbrockmendel:
```
idx = pd.period_range('2016', periods=3, freq='M')

>>> idx == idx.values
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/pandas/core/indexes/period.py", line 107, in wrapper
    other = Period(other, freq=self.freq)
  File "pandas/_libs/tslibs/period.pyx", line 1777, in pandas._libs.tslibs.period.Period.__new__
TypeError: unhashable type: 'numpy.ndarray'
```

mroeschke:
This looks to work on master now. Could use a test

```
In [56]: pd.__version__
Out[56]: '1.1.0.dev0+1536.ga7fb88fd5'

In [57]: idx == idx.values
Out[57]: array([ True,  True,  True])
```

PR #21798, opened 2018-07-07
move PeriodIndex comparisons, implement PeriodArray constructor

jbrockmendel:
PeriodIndex comparisons turn out to be easier to move than the others (though partly because ATM they are broken: #21793), so this moves them over first.

Following that, most of the remaining PeriodArray arithmetic methods require `shift`, which in turns requires the constructors, so I went ahead and implemented those in PeriodArrayMixin.  Still holding off on TimedeltaArray and DatetimeArray constructors.

Found a few more things that could be harmlessly moved and some docstring fixups.

CC: @jorisvandenbossche @TomAugspurger @jreback

### DIFF
--- a/pandas/core/arrays/period.py
+++ b/pandas/core/arrays/period.py
@@ -5,15 +5,17 @@ import warnings
 import numpy as np
 
 from pandas._libs import lib
-from pandas._libs.tslib import NaT
+from pandas._libs.tslib import NaT, iNaT
 from pandas._libs.tslibs.period import (
     Period, IncompatibleFrequency, DIFFERENT_FREQ_INDEX,
     get_period_field_arr)
 from pandas._libs.tslibs.timedeltas import delta_to_nanoseconds
 from pandas._libs.tslibs.fields import isleapyear_arr
 
+from pandas import compat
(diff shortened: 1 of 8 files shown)


pandas/core/arrays/period.py
```diff
@@ -33,6 +35,47 @@ def f(self):
     return property(f)
 
 
+def _period_array_cmp(opname, cls):
+    """
```
jreback: in the future can you add some comments / doc-string here

pandas/core/arrays/period.py
```diff
@@ -161,3 +300,16 @@ def _maybe_convert_timedelta(self, other):
         msg = "Input has different freq from {cls}(freq={freqstr})"
         raise IncompatibleFrequency(msg.format(cls=type(self).__name__,
                                                freqstr=self.freqstr))
+
+    @classmethod
```
jreback: should this use ExtensionOpsMixin ?

jreback (changes requested):

jreback (approved):

jreback:
thanks @jbrockmendel

PR #21803, opened 2018-07-07
implement constructors for TimedeltaArray, DatetimeArray

jbrockmendel:
Implements AttributesMixin since this is a pattern we might want to formalize.

The `_simple_new` that this implements for TimedeltaArray and DatetimeArray are pretty close to the "full" versions in the Index classes.  The `__new__` methods though are very much stripped down for the now.

With these in place we get to move `tz_convert` and `tz_localize`.  Could do a bunch more if I wasn't too much of a wuss to do it all in one go.


pandas/core/arrays/datetimelike.py
```diff
@@ -56,9 +86,61 @@ def asi8(self):
...
+        if lib.is_scalar(key) and not is_int:
+            raise IndexError("only integers, slices (`:`), ellipsis (`...`), "
+                             "numpy.newaxis (`None`) and integer or boolean "
+                             "arrays are valid indices")
+
+        getitem = self._data.__getitem__
+        if is_int:
+            val = getitem(key)
+            return self._box_func(val)
+        else:
```
jreback: don't need the else here

pandas/core/arrays/datetimes.py
```diff
@@ -66,6 +74,50 @@ class DatetimeArrayMixin(DatetimeLikeArrayMixin):
...
+
+        result = object.__new__(cls)
+        result._data = values
+        result._freq = freq
+        tz = timezones.maybe_get_tz(tz)
+        result._tz = timezones.tz_standardize(tz)
+        return result
+
+    def __new__(cls, values, freq=None, tz=None):
+        if (freq is not None and not isinstance(freq, DateOffset) and
```
jreback: this condition could actually be handled to by ``to_offset`` (future PR) (except for the not-None part)

jreback (changes requested):
minor comments

jbrockmendel:
Just updated to de-conflict changes to tz_convert/tz_localize docstrings.

jreback (approved):

pandas/core/arrays/datetimelike.py
jreback: thanks!

jreback:
thanks @jbrockmendel

PR #21806, opened 2018-07-07
Consistent Naming Standard When Using Dict in GroupBy .agg

### LINKED ISSUE #21790: API: inconsistencies between grouped and grouped[['col']]  in groupby
One a DataFrameGroupBy object, you can already select some of the columns you want to apply your functions on, but apart from that, I would not expect it to influence the output structure:

```
In [56]: df = pd.DataFrame({'A' : ['foo', 'bar', 'foo', 'bar'],
    ...:                    'B' : np.random.randn(4),
    ...:                    'C' : np.random.randn(4)})
    ...:                    

In [57]: 

In [57]: gr = df.groupby('A')

In [58]: gr.agg({'B': {'r': np.sum}, 'C': {'r2': np.sum}})
/home/joris/scipy/pandas/pandas/core/groupby/groupby.py:4650: FutureWarning: using a dict with renaming is deprecated and will be removed in a future version
  return super(DataFrameGroupBy, self).aggregate(arg, *args, **kwargs)
Out[58]: 
            C         B
           r2         r
A                      
bar -0.371496  1.230346
foo  0.091779  2.487224

In [59]: gr[['C', 'B']].agg({'B': {'r': np.sum}, 'C': {'r2': np.sum}})
/home/joris/scipy/pandas/pandas/core/groupby/groupby.py:4650: FutureWarning: using a dict with renaming is deprecated and will be removed in a future version
  return super(DataFrameGroupBy, self).aggregate(arg, *args, **kwargs)
Out[59]: 
           r2         r
A                      
bar -0.371496  1.230346
foo  0.091779  2.487224
```

So in the above, doing `gr[['C', 'B']]` causes the final result not to have the MultiIndex, which IMO is very surprising.

The above example if for a deprecated case, so not sure how important it is (didn't directly see a similar case for a non-deprecated call), but when we clean-up the deprecations, we should certainly check this.

cc @jreback @WillAyd

WillAyd:
- [X] closes #21790
- [] tests added / passed
- [ ] whatsnew entry

This is a WIP as all of the tests in the groupby directory pass, but there are some outside of that which need touch ups. This is a relatively aggressive change so I figured I'd get buy in before going outside of that to modify tests.

This not only addresses the referenced issue but provides consistent expectations around columns returned when using a `dict` argument in a `DataFrameGroupBy`, namely that the returned object will have a `MultiIndex` where the top level is the column name and the subsequent level is the aggregation performed.

To play devil's advocate, this may be considered undesirable as it now returns a MultiIndex in *some* cases where that was flat before. However, I'd counter that this:

- Behavior is more consistent and easier to communicate
- Provides a return value whose column labeling is less ambiguous AND
- Simplifies the code base, providing us a clear path forward for whatever the renaming solution actually SHOULD be

Curious to hear other's feedback @jreback @TomAugspurger @jorisvandenbossche


pandas/tests/groupby/test_groupby.py
```diff
@@ -1693,3 +1701,25 @@ def test_groupby_agg_ohlc_non_first():
...
+
+
+@pytest.mark.parametrize("select_columns", [True, False])
+@pytest.mark.parametrize("agg_argument", [
+    {'B': 'sum', 'C': 'min'},  # Scalar result
+    {'B': 'sum', 'C': ['min']},  # Scalar and list
+    {'B': ['sum'], 'C': ['min']},  # Lists
+    {'B': {'sum': 'sum'}, 'C': {'min': 'min'}}  # deprecated call
+])
+def test_agg_dict_naming_consistency(select_columns, agg_argument):
```
WillAyd: FYI plan to move this to the aggregate sub-directory of tests

WillAyd:
Closing as this should be part of larger discussion

PR #21807, opened 2018-07-07
BENCH: asv csv reading benchmarks now rewind StringIO objects

tylerjereddy:
In short, the `asv` benchmarks used for timing `read_csv` are actually reading in empty file-like objects most of the time at the moment. This is because the `setup()` method of the benchmarks is only called between **repeats** and not between **iterations** of repeats--see [writing benchmarks docs](http://asv.readthedocs.io/en/latest/writing_benchmarks.html#timing).

You can confirm this by simply printing the `size` of the dataframes you get from `read_csv` in any StringIO-dependent benchmarks--the **first** iteration of a repeat should have the correct size, but the rest will all be 0. You could also just print the memory address (StringIO object)--they'll all be the same after the first iteration of a repeat. I did this with @stefanv and @mattip in the NumPy context.

This can be quite confusing--a bit different than the paradigm you might expect in i.e, a unit test context--for more gory details see [my related comment in NumPy loadtxt() asv benchmarks](https://github.com/numpy/numpy/pull/11422#issuecomment-403247951). As noted there, one could probably avoid this by using an actual file object instead / if preferred, but realistically this just boils down to [how timeit works](https://stackoverflow.com/questions/8220801/how-to-use-timeit-module#comment11535453_8220943). Confounding the results with the seek is unfortunate, so working with file objects may be preferred if dumping files in the bench env is acceptable.

### DIFF
--- a/asv_bench/benchmarks/io/csv.py
+++ b/asv_bench/benchmarks/io/csv.py
@@ -54,7 +54,14 @@ class ToCSVDatetime(BaseIO):
         self.data.to_csv(self.fname, date_format='%Y%m%d')
 
 
-class ReadCSVDInferDatetimeFormat(object):
+class StringIORewind(object):
+
+    def data(self, stringio_object):
+        stringio_object.seek(0)
+        return stringio_object
+
+
+class ReadCSVDInferDatetimeFormat(StringIORewind):
 
     goal_time = 0.2
     params = ([True, False], ['custom', 'iso8601', 'ymd'])
@@ -66,10 +73,12 @@ class ReadCSVDInferDatetimeFormat(object):
                    'iso8601': '%Y-%m-%d %H:%M:%S',
                    'ymd': '%Y%m%d'}
         dt_format = formats[format]
-        self.data = StringIO('\n'.join(rng.strftime(dt_format).tolist()))
+        self.StringIO_input = StringIO('\n'.join(
+                                       rng.strftime(dt_format).tolist()))
 
     def time_read_csv(self, infer_datetime_format, format):
-        read_csv(self.data, header=None, names=['foo'], parse_dates=['foo'],
+        read_csv(self.data(self.StringIO_input),
+                 header=None, names=['foo'], parse_dates=['foo'],
                  infer_datetime_format=infer_datetime_format)
 
 
@@ -95,7 +104,7 @@ class ReadCSVSkipRows(BaseIO):
         read_csv(self.fname, skiprows=skiprows)
 
 
-class ReadUint64Integers(object):
+class ReadUint64Integers(StringIORewind):
 
     goal_time = 0.2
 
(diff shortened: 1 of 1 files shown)


asv_bench/benchmarks/io/csv.py
```diff
@@ -69,6 +69,7 @@ def setup(self, infer_datetime_format, format):
         self.data = StringIO('\n'.join(rng.strftime(dt_format).tolist()))
 
     def time_read_csv(self, infer_datetime_format, format):
+        self.data.seek(0)
```
jreback: hmm can we instead make ``data`` a property which does this, otherwise this is likely to be forgotten the next time a benchmark is written. can you show a before /after on these benchmarks?

jreback (changes requested):

tylerjereddy:
Revised accordingly with before / after log files from `asv run -e -b "ReadCSV*" >& bench_results.txt` below. Note that this won't scrape in all of the affected benchmarks because not all CSV reading benchmarks follow a consistent naming scheme for regular expression selection, but should pull in many of them. Not sure if pandas CI does a quick (`--quick`) check of asv suite to make sure there are no bench errors on PRs (proposals to do this are in progress for NumPy / SciPy).

Before rewinding `StringIO` objects in csv reading benchmarks (commit hash: 7829ad32): [old_bench_results.txt](https://github.com/pandas-dev/pandas/files/2189026/old_bench_results.txt)

After rewinding `StringIO` objects in csv reading benchmarks: [bench_results.txt](https://github.com/pandas-dev/pandas/files/2188990/bench_results.txt)

I agree that using the property approach may alleviate future forgetting-to-rewind issues, though you'll likely notice that one of the classes where two `data1/2` objects are used instead of parametrization becomes slightly more complicated to think about.

Most `csv` reading benchmarks are now a bit slower (based on visual inspection of above logs), but this is obviously going to happen when you don't read in empty file objects by mistake for most of the iterations, and there's also the conflated time from adding in rewinds now too.

There's another alternative here if you prefer--setting `number = 1 ` in each of the class objects will force `setup()` to run between each iteration; I've preferred not to do that and instead allow the decisions about appropriate number of repeats & iterations to be offloaded to `asv` and `timeit` under the hood.

asv_bench/benchmarks/io/csv.py
```diff
@@ -54,7 +54,15 @@ def time_frame_date_formatting(self):
         self.data.to_csv(self.fname, date_format='%Y%m%d')
 
 
-class ReadCSVDInferDatetimeFormat(object):
+class StringIORewind(object):
+
+    @property
```
jreback: just make this a function I think and pass in the StringIO object

jreback (changes requested):

tylerjereddy:
Revised / squashed / rebased / force pushed. Hopefully better now.

jreback (approved):

jreback:
thanks @tylerjereddy

PR #21808, opened 2018-07-08
More doc warnings

TomAugspurger:


jreback:
thanks @TomAugspurger looks like still a couple of warnings, but reduced from before!

Issue #13962, opened 2016-08-11
TST: fail test suite on ResourceWarnings


jreback:
see comments [here](https://github.com/pydata/pandas/pull/13940#issuecomment-239126585)

should just require a setting in

`pandas.util.testing.TestCase.setUpClass` to set when the suite is running.

jreback:
cc @agraboso

agraboso:
I just tried simply adding `warnings.simplefilter('error', compat.ResourceWarning)` to `pandas.util.testing.TestCase.setUpClass` (and `warnings.simplefilter('ignore', compat.ResourceWarning)` to `pandas.util.testing.TestCase.tearDownClass` for good measure). The warnings appear in the output of the test runner, but the tests themselves do not fail.

For example, reverting [this change](https://github.com/pydata/pandas/pull/13940/files#diff-0d7b5a2c72b4dfc11d80afe159d45ff8R330) results in `pandas.io.tests.parser.common.ParserTests.test_memory_map` issuing the following:

```
Exception ignored in: <_io.FileIO name='/Users/agraboso/Code/pandas/pandas/io/tests/parser/data/test_mmap.csv' mode='rb' closefd=True>
ResourceWarning: unclosed file <_io.TextIOWrapper name='/Users/agraboso/Code/pandas/pandas/io/tests/parser/data/test_mmap.csv' mode='r' encoding='UTF-8'>
```

The same thing happens if the responsibility of closing some handle belongs to the test code itself instead of the tested code. For example, commenting out [this line](https://github.com/pydata/pandas/pull/13940/files#diff-fe58f4e9e077f2d5c7811982556792bfR1583) gives

```
Exception ignored in: <_io.FileIO name=7 mode='rb+' closefd=True>
ResourceWarning: unclosed file <_io.TextIOWrapper name=7 mode='w+' encoding='UTF-8'>
```

in `pandas.io.tests.parser.common.ParserTests.test_temporary_file`.

jreback:
hmm, if what if you do it in ``set_testing_mode``?

agraboso:
Same thing.

jreback:
hmm ok

well if can think of a way to make this error would be great!

h-vetinari:
It's July 2018, and I have unrelated `ResourceWarning`s appearing in one of the jobs on almost every second Travis run. Sometime they are caught and cause a direct error (see e.g. https://travis-ci.org/pandas-dev/pandas/jobs/403728466)
```
[...]

PR #21813, opened 2018-07-08
BUG: rolling with MSVC 2017 build

### LINKED ISSUE #21786: .rolling().std() only returns NaN in Python3.7
```python
import pandas as pd
d = {"col": [1, 23, 231, 231, 4, 353, 62, 3, 56, 43, 354, 43, 231, 21, 7]}
df = pd.DataFrame(data=d)
std = df["col"].std()
df["mean5"] = df["col"].rolling(5).mean()
df["std5"] = df["col"].rolling(5).std()

print(std)
print(df[["mean5", "std5"]])

# OUTPUT
130.20855066648528
    mean5  std5 
0     NaN   NaN
1     NaN   NaN
2     NaN   NaN
3     NaN   NaN
4    98.0   NaN
5   168.4   NaN
6   176.2   NaN
7   130.6   NaN
8    95.6   NaN
9   103.4   NaN
10  103.6   NaN
11   99.8   NaN
12  145.4   NaN
13  138.4   NaN
14  131.2   NaN
```

#### Problem description
`.std()` and `.rolling().mean()` work as intended, but `.rolling().std()` only returns NaN
I just upgraded from Python 3.6.5 where the same code did work perfectly.
I am now on Python 3.7, pandas 0.23.2

#### Expected Output
```python
130.20855066648528
    mean5        std5
0     NaN         NaN
1     NaN         NaN
2     NaN         NaN
3     NaN         NaN
4    98.0  108.855868
5   168.4  134.226078
6   176.2  126.458531
7   130.6  138.965607
8    95.6  131.085621
9   103.4  126.482568
10  103.6  126.877264
11   99.8  128.342355
12  145.4  126.337010
13  138.4  131.941805
14  131.2  137.803338
```

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.7.0.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
processor: Intel64 Family 6 Model 94 Stepping 3, GenuineIntel
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: None.None

pandas: 0.23.2
pytest: None
pip: 10.0.1
setuptools: 39.0.1
Cython: None
numpy: 1.14.5
scipy: None
pyarrow: None
xarray: None
IPython: None
sphinx: None
patsy: None
dateutil: 2.7.3
pytz: 2018.5
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: 4.6.0
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: None
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
None

</details>

chris-b1:
- [x] closes #21786
- [x] tests added / passed
- [x] whatsnew entry

For now just trying 3.7 on appveyor, not sure if all necessary conda packages are there


appveyor.yml
```diff
@@ -25,6 +25,12 @@ environment:
       CONDA_PY: "36"
       CONDA_NPY: "113"
 
+    - CONDA_ROOT: "C:\\Miniconda3_64"
+      PYTHON_VERSION: "3.7"
+      PYTHON_ARCH: "64"
+      CONDA_PY: "37"
+      CONDA_NPY: "113"
```
jreback: can you make this 114?

jreback: this also very likely won't work, very little is built for 3.7 :<

TomAugspurger:
The `c3i_test` channel may have some missing packages.

chris-b1:
Well, good news is we got a repro of this on our CI - it's not a python 3.7 issue, but instead a Visual Studio 2017 one.
https://ci.appveyor.com/project/pandas-dev/pandas/build/1.0.15961/job/r9axp415no6jubnr#L2059

Still not convinced this isn't a compiler bug, but having trouble to get a simple case to reproduce - one option would be to build our wheels VS 2015 - but annoying, b/c python forces the newest version, so can't have 2017 installed.  I'll keep trying some things.

chris-b1:
This is kind of horrible, but it does fix the problem (just a lint issue I pushed for) and I'm out of ideas - any thoughts or suggestions @jreback, @TomAugspurger ?

pandas/_libs/window.pyx
```diff
@@ -654,16 +654,21 @@ cdef inline void add_var(double val, double *nobs, double *mean_x,
                          double *ssqdm_x) nogil:
     """ add a value from the var calc """
     cdef double delta
-
```
jreback: doing
```
if val != val:
    return

# rest of code
```

might work (and be more clear here)

TomAugspurger:
@chris-b1 could you summarize the issue? You think MSVC is just optimizing away `if val == val`, which isn't true when `val` is NaN?

pandas/_libs/window.pyx
chris-b1: Great suggestion, sadly didn't work

TomAugspurger:
Ok, thanks...

@jjhelmus do we know anyone at Microsoft who works on their compilers that might be able to provide insight on https://github.com/pandas-dev/pandas/pull/21813#issuecomment-404501626 and https://github.com/pandas-dev/pandas/pull/21813#issuecomment-404571051?

jjhelmus:
@zooba might be able to connect you to the right team at Microsoft

zooba:
I can try, but I haven't been having as much luck recently.

Since this is Cython, can someone post the generated code for the comparison? There's a much better chance of getting something useful if they're looking at the code they actually compile, and not the code someone else is compiling.

chris-b1:
It's Cython, so of course the bigger file is a tangled mess, but the particular function is easy to read - lines of gist here.  
https://gist.github.com/chris-b1/cd974fb6b4165285a48b6d6c12129b40#file-window-cpp-L11028-L11096

Called in a few places, but for instance, here is a problem one.
https://gist.github.com/chris-b1/cd974fb6b4165285a48b6d6c12129b40#file-window-cpp-L11948

Cython isn't doing anything particular interesting - it's a `cdef` function so basically translates line for line into c++.

I can post it later, not on this machine, but I've also tried to write a simpler, plain c++ example that triggers the problem, but it always worked - really not sure why.

zooba:
I can reproduce it with this snippet (has a few tricks to try and defeat optimisations, but they are likely unnecessary):

```c
#include <math.h>
#include <stdio.h>

double get_value(int i) {
    return i > 1 ? NAN : 7;
}

int check_default(double v1, double v2) {
    return ((v1 == v2) != 0);
}

#pragma float_control(precise, off, push)
int check_fast(double v1, double v2) {
    return ((v1 == v2) != 0);
}
#pragma float_control(pop)


int main(int argc, char **argv) {
    double x1 = get_value(argc);
    double x2 = get_value(argc);
    fprintf(stdout, "default: %d\nfast:    %d\n",
        check_default(x1, x2), check_fast(x1, x2)
    );
}
```

The comparison doesn't behave properly when you disable precise floating point operations. You may be doing this in the build with `/fp:fast` (if you build this snippet with that option, both checks return '1'; if you build with the default `/fp:precise`, it works). In fact, the [docs for this flag](https://docs.microsoft.com/en-us/cpp/build/reference/fp-specify-floating-point-behavior) specifically call out NaN comparisons as something that only work in `precise` mode.

So I'd guess you need to update the compiler flags for this one, or possibly add an explicit `isnan()` call.

pandas/_libs/window.pyx
jbrockmendel: Sometimes I see compiler warnings about tautologically false comparisons.  Is there a chance the compiler threw out this check?

jbrockmendel:
Is the MSVC info something that can be inferred from platform.uname (at compile-time)?

chris-b1:
Thanks for taking a look at this @zooba - unfortunately that's not exactly the same problem I'm seeing here, the compiled code is acting as if `(v1 == v2)` always evaluates to False/0.  We are building with `/fp:precise` (via defaults, but I've also tried manually overriding)

chris-b1:
Interestingly using `isnan` from <math.h> does solve the problem here - still feels like MSVC is doing something wrong in the former case, but that's a reasonable enough workaround.

zooba:
> the compiled code is acting as if `(v1 == v2)` always evaluates to False/0

Do you have a disassembly showing this? This would be very strange to see as a compiler optimization (unless you mean `(v == v)` always evaluates to True/1?)

> b/c python forces the newest version, so can't have 2017 installed

Just saw this while rereading the thread. If you start with all the tools on PATH and set `DISTUTILS_USE_SDK=1` then it'll skip the autodetection and use whatever you provide. (And if you help encourage the ecosystem to get off distutils completely, then more of these problems go away :) )

jbrockmendel:
Could this SO answer be relevant?

https://stackoverflow.com/questions/8200311/checking-for-nan-in-cython/20031818#20031818

chris-b1:
@zooba - do you know how to pull symbols into the dissambly from a `.pyd`?  I was playing around with dumpbin / compile flags, but never got anything better than, e.g. below, where everything is still in offsets.

```
 dumpbin.exe /DISASM pandas\_libs\window.cp36-win_amd64.pyd /OUT:tmp.txt
```

image

chris-b1:
> do you know how to pull symbols into the dissambly from a .pyd

Answering my own question - I was missing `'/DEBUG'` on `extra_link_args`

chris-b1:
Alright @zooba - I  think I have this to a reasonable bit of a dis-assembly - pushing the bounds of my knowledge but looks like a code-generation bug.

Here's MSVC 2015 (printfs surround a call to `add_var`)  which works correctly.  In particular the highlighted lines are the NaN check.
https://gist.github.com/chris-b1/1049fdabafc56b2a25de7ad377016baa#file-msvc2015-asm-L6-L8

And here is MSVC 2017. The function call doesn't get inlined - half of the NaN check gets done (`jn`) - then the stack is being set up, then the other half the check (`jne`).  I believe in setting up the stack, the ZF flag originally set by `ucomisd` gets set back to 0 (from the `sub`) so the `jne` check will always be true, and the function unconditionally exits early.
https://gist.github.com/chris-b1/199f818c1fbb8f8c4108aa9e430df4f7#file-msvc2017-asm-L15-L19

zooba:
@chris-b1 Yep, you're absolutely right. Thanks for tracking that down, I'll pass it on!

zooba:
Here is the public side of the report: https://developercommunity.visualstudio.com/content/problem/294290/incorrect-codegen-for-double-equality.html Feel free to add pings there whenever you like

chris-b1:
Thanks @zooba!

and thanks @jbrockmendel - that's what I'll have to do to fix the 2.7 build, more modern MSVCs are basically standards compliant and have `isnan` where it should be.

jreback:
does using the fix proposed in #21940 help here? e.g. ``npy_isnan``?

pandas/_libs/window.pyx
```diff
@@ -650,20 +651,20 @@ cdef inline double calc_var(int64_t minp, int ddof, double nobs,
     return result
 
 
-cdef inline void add_var(double val, double *nobs, double *mean_x,
+cdef inline void add_var(const double val, double *nobs, double *mean_x,
```
jreback: is this const needed?

0.23.4 is still on an older cython

chris-b1: No, just me trying random stuff to fool the compiler, I'll back that out

chris-b1:
@jreback  - any idea on the circle CI failure? seems unrelated, but consistently failing on this branch

<details>

```bash
__________________ TestPythonParser.test_dtype_with_converter __________________

self = <pandas.tests.io.parser.test_parsers.TestPythonParser object at 0x7f2235d65b70>

    def test_dtype_with_converter(self):
        data = """a,b
    1.1,2.2
    1.2,2.3"""
        # dtype spec ignored if converted specified
        with tm.assert_produces_warning(ParserWarning):
            result = self.read_csv(StringIO(data), dtype={'a': 'i8'},
>                                  converters={'a': lambda x: str(x)})

pandas/tests/io/parser/dtypes.py:340: 
../miniconda3/envs/pandas/lib/python3.5/contextlib.py:66: in __exit__
    next(self.gen)

expected_warning = <class 'pandas.errors.ParserWarning'>
filter_level = 'always', clear = None, check_stacklevel = True

    @contextmanager
    def assert_produces_warning(expected_warning=Warning, filter_level="always",
                                clear=None, check_stacklevel=True):
        """
        Context manager for running code expected to either raise a specific
        warning, or not raise any warnings. Verifies that the code raises the
        expected warning, and that it does not raise any other unexpected
        warnings. It is basically a wrapper around ``warnings.catch_warnings``.
    
        Parameters
        ----------
        expected_warning : {Warning, False, None}, default Warning
            The type of Exception raised. ``exception.Warning`` is the base
            class for all warnings. To check that no warning is returned,
            specify ``False`` or ``None``.
        filter_level : str, default "always"
            Specifies whether warnings are ignored, displayed, or turned
            into errors.
            Valid values are:
    
            * "error" - turns matching warnings into exceptions
            * "ignore" - discard the warning
            * "always" - always emit a warning
            * "default" - print the warning the first time it is generated
              from each location
            * "module" - print the warning the first time it is generated
              from each module
            * "once" - print the warning the first time it is generated
    
        clear : str, default None
            If not ``None`` then remove any previously raised warnings from
            the ``__warningsregistry__`` to ensure that no warning messages are
            suppressed by this context manager. If ``None`` is specified,
            the ``__warningsregistry__`` keeps track of which warnings have been
             shown, and does not show them again.
        check_stacklevel : bool, default True
            If True, displays the line that called the function containing
            the warning to show were the function is called. Otherwise, the
            line that implements the function is displayed.
    
        Examples
        --------
        >>> import warnings
        >>> with assert_produces_warning():
        ...     warnings.warn(UserWarning())
        ...
        >>> with assert_produces_warning(False):
        ...     warnings.warn(RuntimeWarning())
        ...
            ...
        AssertionError: Caused unexpected warning(s): ['RuntimeWarning'].
        >>> with assert_produces_warning(UserWarning):
        ...     warnings.warn(RuntimeWarning())
            ...
        AssertionError: Did not see expected warning of class 'UserWarning'.
    
        ..warn:: This is *not* thread-safe.
        """
        with warnings.catch_warnings(record=True) as w:
    
            if clear is not None:
                # make sure that we are clearning these warnings
                # if they have happened before
                # to guarantee that we will catch them
                if not is_list_like(clear):
                    clear = [clear]
                for m in clear:
                    try:
                        m.__warningregistry__.clear()
                    except Exception:
                        pass
    
            saw_warning = False
            warnings.simplefilter(filter_level)
            yield w
            extra_warnings = []
    
            for actual_warning in w:
                if (expected_warning and issubclass(actual_warning.category,
                                                    expected_warning)):
                    saw_warning = True
    
                    if check_stacklevel and issubclass(actual_warning.category,
                                                       (FutureWarning,
                                                        DeprecationWarning)):
                        from inspect import getframeinfo, stack
                        caller = getframeinfo(stack()[2][0])
                        msg = ("Warning not set with correct stacklevel. "
                               "File where warning is raised: {actual} != "
                               "{caller}. Warning message: {message}"
                               ).format(actual=actual_warning.filename,
                                        caller=caller.filename,
                                        message=actual_warning.message)
                        assert actual_warning.filename == caller.filename, msg
                else:
                    extra_warnings.append((actual_warning.category.__name__,
                                           actual_warning.message,
                                           actual_warning.filename,
                                           actual_warning.lineno))
            if expected_warning:
                msg = "Did not see expected warning of class {name!r}.".format(
                    name=expected_warning.__name__)
                assert saw_warning, msg
            assert not extra_warnings, ("Caused unexpected warning(s): {extra!r}."
>                                       ).format(extra=extra_warnings)
E           AssertionError: Caused unexpected warning(s): [('ResourceWarning', ResourceWarning("unclosed file <_io.BufferedReader name='/tmp/tmpuehs2h5ztest_file.zip'>",), '/home/ubuntu/miniconda3/envs/pandas/lib/python3.5/importlib/_bootstrap.py', 967)].

pandas/util/testing.py:2535: AssertionError
```
</details>

h-vetinari:
I've had similar experiences recently (although it always happened to me on Travis) - I found the following issue and commented there #13962.

TomAugspurger:
FYI @chris-b1 circle CI has a CLI for running stuff locally. I think it's safe to ignore here though.

jreback:
needs a rebase. also Circle still seems to be failing.

jreback:
@chris-b1 this looks good. can you rebase once more. 

perf looks ok? (even if its slightly worse this is a bug :>)

Go ahead and merge on green.

chris-b1:
@jreback thanks, yeah perf will be fine, just slightly slower on Windows for this particular function

jreback:
cc @chris-b1 do you want to see if you can backprot to 23.x branch?

PR #21802, opened 2018-07-07
CLN: Remove unused imports in pyx files

mroeschke:
cc @jbrockmendel 

Removed unused imports flagged by PyCharm. Compiled fine locally, waiting for CI to pass.

PR #21816, opened 2018-07-08
update tslibs imports

jbrockmendel:
Another try at #21802.  Removing `Timedelta` from the `tslib` namespace broke a ton of imports, so this goes through and updates all of them.


mroeschke:
Linting error `pandas/core/resample.py:27:1: F401 'pandas._libs.tslib' imported but unused` otherwise LGTM

jreback:
thanks!

Issue #10043, opened 2015-05-01
iloc breaks on read-only dataframe


amueller:
This is picking up #9928 again. I don't know if the behavior is expected, but it is a bit odd to me. Maybe I'm doing something wrong, I'm not that familiar with the pandas internals.

We call ``df.iloc[indices]`` and that breaks with a read-only dataframe. I feel that it shouldn't though, as it is not writing.

Minimal reproducing example:
```python
import pandas as pd
import numpy as np
array = np.eye(10)
array.setflags(write=False)

X = pd.DataFrame(array)
X.iloc[[1, 2, 3]]
```
> ValueError buffer source array is read-only

Is there a way to slice the rows of the dataframe in another way that doesn't need a writeable array?

amueller:
Maybe to clarify: I would be happy if the thing sliced DF is a copy, I just don't want to copy the original data [I think].

jreback:
by definition what you are doing is ALWAYS going to be a copy. Frames are stored in columns.

```
In [10]: df = DataFrame(np.arange(20).reshape(5,4))

In [11]: df
Out[11]: 
    0   1   2   3
0   0   1   2   3
1   4   5   6   7
2   8   9  10  11
3  12  13  14  15
4  16  17  18  19

In [12]: result = df.iloc[[1,2,3]]

In [13]: result
Out[13]: 
    0   1   2   3
1   4   5   6   7
2   8   9  10  11
3  12  13  14  15

In [14]: result.values.base
Out[14]: 
array([[ 4,  5,  6,  7],
       [ 8,  9, 10, 11],
       [12, 13, 14, 15]])

In [15]: df._data
Out[15]: 
BlockManager
Items: Int64Index([0, 1, 2, 3], dtype='int64')
Axis 1: Int64Index([0, 1, 2, 3, 4], dtype='int64')
IntBlock: slice(0, 4, 1), 4 x 5, dtype: int64


In [17]: df._data.blocks[0].values
Out[17]: 
array([[ 0,  4,  8, 12, 16],
       [ 1,  5,  9, 13, 17],
       [ 2,  6, 10, 14, 18],
       [ 3,  7, 11, 15, 19]])
```

you can just access using [17] if you want to get the actual numpy array. Keep in mind, this is ONLY valid for a single-dtyped frame.

TomAugspurger:
Would it be OK for scikit-learn to rely on `._data` / blocks in library code?

[...]

PR #21817, opened 2018-07-08
Modernize the fix for #10043

xhochy:
- [ ] tests added / passed
- [x] whatsnew entry


xhochy:
This fails for non-writeable `object` NumPy arrays as they cannot be declared as `const` in Cython. This leaves us with either the case that we don't support them or that we reject this PR again.

jreback:
hmm, that's annoying. so we still the the work around but only for ``object``.

jreback:
you could leave the workaround in place and use the const keyword for non-object. i suppose that's an improvement.

jorisvandenbossche:
> This fails for non-writeable object NumPy arrays as they cannot be declared as const in Cython. 

Do you know if that is an inherent limitation, or something that can be improved in Cython? (in which case we should report this there)

xhochy:
Looking a bit deeper into this, I think this could be a limitation of Cython but I'm missing the necessary background knowledge, thus I opened a bug: https://github.com/cython/cython/issues/2485

jreback:
closing as pending a cython fix :<

PR #21820, opened 2018-07-08
CLN: reorg groupby to multiple modules

jreback:
makes the impl of groupby a bit easier to grok. 

cc @WillAyd @TomAugspurger @jorisvandenbossche


pandas/core/panel.py
```diff
@@ -911,7 +911,7 @@ def groupby(self, function, axis='major'):
         -------
         grouped : PanelGroupBy
         """
-        from pandas.core.groupby.groupby import PanelGroupBy
+        from pandas.core.groupby.panel import PanelGroupBy
```
jorisvandenbossche: I would make this importing just from `pandas.core.groupby`

jreback:
if any comments. This has no user facing changes.

pandas/core/resample.py
```diff
@@ -5,12 +5,13 @@
...
-
+from pandas.core.groupby.base import GroupByMixin
+from pandas.core.groupby.ops import BinGrouper, Grouper
 from pandas.core.groupby.groupby import (
-    BinGrouper, Grouper, _GroupBy, GroupBy, SeriesGroupBy, groupby,
-    PanelGroupBy, _pipe_template
+    _GroupBy, GroupBy, groupby, _pipe_template
 )
+from pandas.core.groupby.series import SeriesGroupBy
+from pandas.core.groupby.panel import PanelGroupBy
```
jorisvandenbossche: Same here, I would import only from `pandas.core.groupby`, and expose there everything that is needed elsewhere in pandas. That way we don't always have to change this if the groupby module is internally refactored.

jorisvandenbossche: And doing that makes it also clear what is used elsewhere in pandas, as this is what is in the `groupby/__init__.py`

pandas/core/groupby/ops.py
```diff
@@ -0,0 +1,1528 @@
+"""
+perform the groupby operations. These are not exposed
+to the top-level interace in groupby.py
```
jorisvandenbossche: Grouper is exposed and used in resample code? So that doesn't seem fully true?

pandas/core/groupby/ops.py
```diff
@@ -0,0 +1,1528 @@
+"""
+perform the groupby operations. These are not exposed
```
jorisvandenbossche: Can you expand a bit on "perform the groupby operations". That's not directly clear to me from reading that. (it is the "split" (finding the groups) part of "split-apply-combine ?)

jorisvandenbossche (commented):
There are still a couple of open groupby doc PRs (that we have been merging the last days). Merging this will make updating those harder (but merging those PRs first also give a possibly painful merge here)

I would personally leave frame and series stuff together with NDFrameGroupby, as those are quite related (that still makes that the file is splitted, so still an improvement)

WillAyd (approved):
Seems logical

pandas/core/groupby/frame.py
```diff
@@ -0,0 +1,368 @@
...
+from pandas.core.groupby.groupby import GroupBy
+from pandas.core.groupby import base
+from pandas.util._decorators import Appender
+from pandas.core.index import MultiIndex, CategoricalIndex
+from pandas.core.arrays.categorical import Categorical
+from pandas.core.frame import DataFrame
+from pandas.core.groupby.groupby import NDFrameGroupBy
+from pandas.core.groupby.series import SeriesGroupBy
+from pandas._libs.lib import count_level_2d
+from pandas.plotting._core import boxplot_frame_groupby
```
jbrockmendel: since stuff is being moved en-masse, this is a good time to clean up imports.  After the standard `__future__`-->stdlib-->3rd party blocks, I advocate blocks in order of dependencies.  So imports from `_libs`, then `util` and `compat`, then `core` (I like to put `core.dtypes` before other `core`), and last `plotting`.  YMMV.

jreback:
> There are still a couple of open groupby doc PRs (that we have been merging the last days). Merging this will make updating those harder (but merging those PRs first also give a possibly painful merge here)

better to do file splitting sooner rather than later

pandas/core/panel.py
jreback: done

pandas/core/resample.py
jreback: this not following the general pattern. It is trivial to update this if imports are changed and IMHO it makes the structure much more clear here.

pandas/core/groupby/frame.py
jreback: updated

jreback:
ok this is looking pretty good now. split logically. if any more pass. want to merge this soonish to avoid conflicts.

pandas/core/groupby/ops.py
```diff
@@ -0,0 +1,916 @@
+"""
+Provide classe to perform the groupby aggregate operations.
```
jbrockmendel: typo classe

jreback: done

jreback:
if any other comments

@pandas-dev/pandas-core pretty straightforward split.

jreback:
merging tomorrow if any comments.

PR #19937, opened 2018-02-28
Making to_datetime('today') and Timestamp('today') consistent

shangyian:
This PR changes `pd.to_datetime` to be consistent with `Timestamp('today')`.

- [x] closes #19935
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

PR #21822, opened 2018-07-08
BUG/API: to_datetime preserves UTC offsets when parsing datetime strings

### LINKED ISSUE #11736: Regression: DatetimeIndex ignores timezones
In 0.16.2, `DatetimeIndex` functions used to handle the timezone correctly:

```
In [1]: import pandas as pd
In [2]: df = pd.DataFrame({'x': ["2015-05-27 19:05:17.178000-02:00"]})
In [3]: pd.DatetimeIndex(df.x)
Out[3]: DatetimeIndex(['2015-05-27 19:05:17.178000-02:00'], dtype='datetime64[ns]', freq=None, tz='tzoffset(None, -7200)')
```

In 0.17.0, it's now ignoring timezones:

```
In [1]: import pandas as pd

In [2]: df = pd.DataFrame({'x': ["2015-05-27 19:05:17.178000-02:00"]})

In [3]: pd.DatetimeIndex(df.x)
Out[3]: DatetimeIndex(['2015-05-27 21:05:17.178000'], dtype='datetime64[ns]', freq=None)
```

This slipped thru since the tests are very simple and don't cover many cases:

https://github.com/pydata/pandas/blob/cda9da82a68aed42421f7f788fb0c1e949e7a60c/pandas/tests/test_format.py#L3342

I'll attempt a patch later but if someone can give pointers where the breaking change happened I appreciate.

### LINKED ISSUE #17697: Timestamp(foo) vs to_datetime(foo)
I assumed that these should be the same:

```
In [2]: pd.to_datetime("2015-11-18 15:30:00+05:30")
Out[2]: Timestamp('2015-11-18 10:00:00')

In [3]: pd.Timestamp("2015-11-18 15:30:00+05:30")
Out[3]: Timestamp('2015-11-18 15:30:00+0530', tz='pytz.FixedOffset(330)')
```

Is the mismatch intentional?

### LINKED ISSUE #21864: API/BUG: to_datetime(..., box=True) should always return an Index
Currently `box=True` is specified in the docstring to return a DatetimeIndex, but sometimes this is not possible due to how the dates are parsed and a different Index subclass is returned. However, this policy is not consistent:

```
In [8]: pd.__version__
Out[8]: '0.24.0.dev0+312.gd48f34141.dirty'

In [9]: idx = pd.Index([15e9], name='name')

In [10]: pd.to_datetime(idx, errors='ignore', box=True, unit='s')  # Index returned
Out[10]: Float64Index([15000000000.0], dtype='float64')

In [11]: pd.to_datetime(idx, errors='ignore', box=False, unit='s')
Out[11]: array([15000000000.0], dtype=object)

In [12]: malformed = np.array(['1/100/2000', np.nan], dtype=object)

In [13]: pd.to_datetime(malformed, errors='ignore', box=True)  # Index not returned
Out[13]: array(['1/100/2000', nan], dtype=object)

In [14]: pd.to_datetime(malformed, errors='ignore', box=False)
Out[14]: array(['1/100/2000', nan], dtype=object)
``` 

For consistency, I think `box=True` should always attempt to wrap the argument in an `Index`

Addressing this in #21822

mroeschke:
- [x] closes #17697
- [x] closes #11736
- [x] closes #21864
- [x] precursor to fixing #15938
- [x] tests added / passed
- [x] whatsnew entry

This PR makes `to_datetime(ts_string_with_offset)` and `DatetimeIndex([ts_string_with_offset])` to now match `Timestamp(ts_string_with_offset)`
```
This branch

In [2]: Timestamp("2015-11-18 15:30:00+05:30")
Out[2]: Timestamp('2015-11-18 15:30:00+0530', tz='pytz.FixedOffset(330)')

In [3]: to_datetime("2015-11-18 15:30:00+05:30")
Out[3]: Timestamp('2015-11-18 15:30:00+0530', tz='pytz.FixedOffset(330)')

In [4]: DatetimeIndex(["2015-11-18 15:30:00+05:30"])
Out[4]: DatetimeIndex(['2015-11-18 15:30:00+05:30'], dtype='datetime64[ns, pytz.FixedOffset(330)]', freq=None)
```


jbrockmendel:
Looking over the test failures, most of them are clearly the-test-is-wrong.

pandas/core/tools/datetimes.py
```diff
@@ -266,14 +266,17 @@ def _convert_listlike_datetimes(arg, box, format, name=None, tz=None,
...
                 arg,
                 errors=errors,
                 utc=tz == 'utc',
                 dayfirst=dayfirst,
                 yearfirst=yearfirst,
                 require_iso8601=require_iso8601
             )
+            if tz_parsed is not None:
+                return DatetimeIndex._simple_new(result, name=name,
+                                                 tz=tz_parsed)
```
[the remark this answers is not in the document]
mroeschke: That case will result in `tz_parsed = None` so this branch will not be hit.

pandas/_libs/tslib.pyx
```diff
@@ -651,7 +675,22 @@ cpdef array_to_datetime(ndarray[object] values, errors='raise',
...
+            #    parsed dates so it can behave nicely with DatetimeIndex
+            # 2) If the offsets are different, then force the parsing down the object path
+            #    where an array of datetimes (with individual datutil.tzoffsets) are returned
+
+            # Faster to compare integers than to compare objects
+            is_same_offsets = (out_tzoffset_values[0] == out_tzoffset_values).all()
+            if not is_same_offsets:
+                raise TypeError
+            else:
+                # Open question: should this return dateutil offset or pytz offset?
```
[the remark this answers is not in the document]
mroeschke: For this PR then, is it okay that parsing through `Timestamp` will produce a `pytz.FixedOffset` and `to_datetime` will producea `dateutil.tz.tzoffset`?

I should probably start a larger discussion whether we should be migrating from pytz to dateutil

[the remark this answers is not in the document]
jbrockmendel: My mistake, forgot that `Timestamp` defaulted to `pytz.FixedOffset`.  Sharing code with the `Timestamp` constructor is definitely a higher priority than defaulting to `dateutil.tz`.

pandas/_libs/tslib.pyx
```diff
@@ -449,26 +449,51 @@ cpdef array_to_datetime(ndarray[object] values, errors='raise',
                         dayfirst=False, yearfirst=False,
                         format=None, utc=None,
                         require_iso8601=False):
+    """
+    Converts a 1D array of date-like values to a numpy array of either:
+        1) datetime64[ns] data
+        2) datetime.datetime objects, if OutOfBoundsDatetime or TypeError
+           is encountered
+
+    Also returns a pytz.FixedOffset if an array of strings with the same
```
jbrockmendel: In principle other tzinfos could be returned, specifically if it falls back to dateutil

pandas/_libs/tslib.pyx
```diff
@@ -449,26 +449,51 @@ cpdef array_to_datetime(ndarray[object] values, errors='raise',
...
+    timezone offset if passed and utc=True is not passed
+
+    Handles datetime.date, datetime.datetime, np.datetime64 objects, numeric,
+    strings
+
+    Returns
+    -------
+    (ndarray, timezone offset)
```
jbrockmendel: underscore between timezone and offset?

pandas/_libs/tslib.pyx
```diff
@@ -649,7 +677,22 @@ cpdef array_to_datetime(ndarray[object] values, errors='raise',
...
                 raise TypeError
 
-        return result
+        if seen_datetime_offset and not utc_convert:
+            # GH 17697
+            # 1) If all the offsets are equal, return one pytz.FixedOffset for
+            #    the parsed dates to (maybe) pass to DatetimeIndex
+            # 2) If the offsets are different, then force the parsing down the
+            #    object path where an array of datetimes
+            #    (with individual datutil.tzoffsets) are returned
```
jbrockmendel: typo datutil

pandas/_libs/tslib.pyx
```diff
@@ -649,7 +677,22 @@ cpdef array_to_datetime(ndarray[object] values, errors='raise',
...
+
+            # Faster to compare integers than to compare objects
+            is_same_offsets = (out_tzoffset_vals[0] == out_tzoffset_vals).all()
+            if not is_same_offsets:
+                raise TypeError
```
jbrockmendel: This (pre-existing) pattern is pretty opaque to a first-time reader.  What if instead of raising `TypeError` the fallback block became its own function that gets called from here?

pandas/tests/indexes/datetimes/test_tools.py
```diff
@@ -484,15 +485,17 @@ def test_to_datetime_tz_psycopg2(self, cache):
...
+                                                         name=None))
         assert is_datetime64_ns_dtype(i)
 
         # tz coerceion
         result = pd.to_datetime(i, errors='coerce', cache=cache)
         tm.assert_index_equal(result, i)
 
         result = pd.to_datetime(i, errors='coerce', utc=True, cache=cache)
-        expected = pd.DatetimeIndex(['2000-01-01 13:00:00'],
+        expected = pd.DatetimeIndex(['2000-01-01 08:00:00'],
```
jbrockmendel: reason for this change?

pandas/tests/tslibs/test_array_to_datetime.py
```diff
@@ -52,46 +54,64 @@ def test_parsers_iso8601_invalid(self, date_str):
...
+        ['2013-01-01T08:00:00.000000000+0800', pytz.FixedOffset(480)],
+        ['2012-12-31T16:00:00.000000000-0800', pytz.FixedOffset(-480)],
+        ['12-31-2012 23:00:00-01:00', None]])
+    def test_parsing_timezone_offsets(self, dt_string, expected_tz):
         # All of these datetime strings with offsets are equivalent
         # to the same datetime after the timezone offset is added
+
+        # TODO: Appears that parsing non-ISO strings adjust the date to UTC
+        # but don't return the offset. Not sure if this is the intended
+        # behavior of non-iso strings in np_datetime_strings
```
jbrockmendel: np_datetime_strings doesn't handle non-ISO.  That case ends up going through dateutil (via `parse_datetime_string`).  I'm not totally sure what the TODO is for.  `'01-01-2013 00:00:00'` comes back tz-naive, right?

pandas/_libs/tslib.pyx
mroeschke: This is specifying that `array_to_datetime` function itself can return a `pytz.FixedOffset` or `None` from the C parser output. If it goes through the dateutil parser in the non-`ValueError` branch, I don't think there's a way to recover the timezone?

pandas/tests/indexes/datetimes/test_tools.py
mroeschke: The `+00:00` in `i` above required a `tz_convert` post construction while I think the original intention was to `tz_localize` it to the passed `psycopg.tz` in the constructor (leading to this change)

But to reflect the intension of the original test, I just removed the `+00:00` instead

pandas/_libs/tslib.pyx
jbrockmendel: When `parse_datetime_string` is called if there's a timezone it should return a tz-aware `datetime` object, so the tzinfo can be pulled off that can't it?

mroeschke: Oh right that's true, good catch. Sure I will try to add some tests and functionality to hit the dateutil parser before the object branch.

mroeschke: So I am using a `set` to store the parsed timezone offsets (which should be more performant that using an array in theory / I was having some trouble using an array due to duplicates), however `dateutil.tz.tzoffset`s cannot be hashed: https://github.com/dateutil/dateutil/issues/792

So instead, I am saving the `total_seconds()` of the dateutil tzoffset in the set instead and reconstructing the offsets as `pytz.FixedOffset`s

jreback (changes requested):
looks really good. some comments.

pandas/tests/tslibs/test_array_to_datetime.py
```diff
@@ -52,46 +54,68 @@ def test_parsers_iso8601_invalid(self, date_str):
...
+        dt_string = '01-01-2013T00:00:00.000000000+0000'
+        arr = np.array([dt_string], dtype=object)
+        result, result_tz = tslib.array_to_datetime(arr)
+        expected = np.array([np.datetime64('2013-01-01 00:00:00.000000000')])
+        tm.assert_numpy_array_equal(result, expected)
+        assert result_tz is pytz.FixedOffset(0)
+
+    def test_parsing_different_timezone_offsets(self):
+        # GH 17697
+        data = ["2015-11-18 15:30:00+05:30", "2015-11-18 15:30:00+06:30"]
```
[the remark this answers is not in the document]
mroeschke: Yup. Similar tests are here: https://github.com/mroeschke/pandas/blob/4ff7cb377d886ef8b515c4ca1ceccffde99c4201/pandas/tests/indexes/datetimes/test_tools.py#L595

pandas/tests/indexes/datetimes/test_arithmetic.py
```diff
@@ -423,13 +423,13 @@ def test_dti_shift_tzaware(self, tz_naive_fixture):
         tm.assert_index_equal(idx.shift(0, freq='H'), idx)
         tm.assert_index_equal(idx.shift(3, freq='H'), idx)
 
-        idx = pd.DatetimeIndex(['2011-01-01 10:00', '2011-01-01 11:00'
```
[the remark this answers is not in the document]
mroeschke: I am surprised this didn't raise earlier, but this test was failing locally as I was working on this PR

pandas/_libs/tslib.pyx
```diff
@@ -679,36 +764,67 @@ cpdef array_to_datetime(ndarray[object] values, errors='raise',
...
+        elif is_string_object(val):
+            if len(val) == 0 or val in nat_strings:
+                oresult[i] = 'NaT'
+                continue
+            try:
+                oresult[i] = parse_datetime_string(val, dayfirst=dayfirst,
+                                                   yearfirst=yearfirst)
+                pydatetime_to_dt64(oresult[i], &dts)
+                check_dts_bounds(&dts)
+            except Exception:
```
jbrockmendel: This could probably be more specific.  parse_datetime_string could raise ValueError or OverflowError, check_dts_bounds raises OutOfBoundsDatetime (which subclasses ValueError), and I think thats it.

jbrockmendel:
Does this make `Timestamp("now")` match `to_datetime("now")`?  Or "today"?

mroeschke:
This doesn't fix "now". ("today" looks like it was handled in #19937)

jreback:
can you rebase

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -224,6 +224,62 @@ For situations where you need an ``ndarray`` of ``Interval`` objects, use
...
+or :class:`DatetimeIndex` would automatically convert the datetime to UTC
+without timezone localization. This is inconsistent from parsing the same
+datetime string with :class:`Timestamp` which would preserve the UTC
+offset in the ``tz`` attribute. Now, :func:`to_datetime` preserves the UTC
+offset in the ``tz`` attribute when all the datetime strings have the same
+UTC offset (:issue:`17697`, :issue:`11736`)
+
+*Previous Behavior*:
+
+.. code-block:: ipython
```
jreback: I think this directiive you just be python (rather than ipython here) @TomAugspurger ?

pandas/tests/tslibs/test_array_to_datetime.py
```diff
@@ -123,11 +147,11 @@ def test_coerce_of_invalid_datetimes(self):
 
         # Without coercing, the presence of any invalid dates prevents
         # any values from being converted
-        result = tslib.array_to_datetime(arr, errors='ignore')
+        result = tslib.array_to_datetime(arr, errors='ignore')[0]
```
jreback: I like writing these like

``result, _ = ......`` I think its more clear

jreback (approved):
minor comments. @jbrockmendel any comments?

jreback:
@mroeschke can you also run some timeseries benchmarks generally to see if any effects (I would expected *some* small slowdown for parsing but but not significant). You *may* need to add some benchmarks to capture the new code paths.

jbrockmendel:
> any comments? 

Just that I’m pretty stoked to see this fixed.

mroeschke:
From a new asv I added, parsing strings with different offests will be slower, but I think its acceptable given the new (more correct) behavior

```
     [d30c4a06]       [807a2513]
+         1.17±0s       2.37±0.02s     2.02  timeseries.ToDatetimeNONISO8601.time_different_offset

```

doc/source/whatsnew/v0.24.0.txt
mroeschke: Looking at previous whatsnews, looks like either ipython or python works for this directive.

jreback:
ncie patch @mroeschke !

pandas/_libs/tslib.pyx
```diff
@@ -595,6 +633,30 @@ cpdef array_to_datetime(ndarray[object] values, errors='raise',
...
+                        seen_datetime_offset = 1
+                        if tz == dateutil_utc():
+                            # dateutil.tz.tzutc has no offset-like attribute
+                            # Just add the 0 offset explicitly
+                            out_tzoffset_vals.add(0)
+                        elif tz == tzlocal():
+                            # is comparison fails unlike other dateutil.tz
+                            # objects. Also, dateutil.tz.tzlocal has no
+                            # _offset attribute like tzoffset
+                            offset_seconds = tz._dst_offset.total_seconds()
```
pganssle: Note that this is using an internal method and is fragile.

jbrockmendel: Good point.  Suggestion for a better way to do this?

pganssle: I'm not sure I understand what this function is supposed to do, but from what I *do* understand it looks like a mistake.

mroeschke: My original intent with this section was to collect the dateutil timezone objects in a `set` and determine if more than 1 distinct timezone object was collected.

However since dateutil timezones cannot be hashed (thanks for starting progress on the associated issue btw @pganssle), I instead store the timezone's UTC offsets in seconds instead (which is what I essentially care about).

It wasn't too apparent to me if there was a public way to access a dateutil timezone's UTC offset, hence why I am use a private method here. Definitely open to a better method.

pganssle: I think if you can't find a public method you probably shouldn't dive into the private API. There are a few assumptions in here that won't survive future versions of `dateutil` for sure. The `tzlocal` thing also actually gives you the *wrong answer*, because `tzlocal` is not a fixed offset and you're taking the DST offset.

If you want the offset, you should do `dt.utcoffset()`, regardless of whether it's `dateutil` or not.

pganssle: If you are significantly worried about speed, you can probably use the fact that in `dateutil >= 2.7.0`, `dateutil.tz.tzutc()` returns a singleton, and `dateutil.tz.tzoffset(*args1)` is the same object as any other `dateutil.tz.tzoffset(*args2)` where `args1 == args2`. So you can probably store a mapping between `id(obj)` and the result of `obj.tzoffset(datetime(1970, 1, 1))` for that specific subclass (since it is also guaranteed to have no DST).

PR #19191, opened 2018-01-11
Hand written ISO parser for Timedelta construction

WillAyd:
- [X] closes #19103 
- [X] tests added / passed
- [ ] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] whatsnew entry

```
Running 16 total benchmarks (2 commits * 1 environments * 8 benchmarks)
[  0.00%] · For pandas commit hash 76475dd8:
[  0.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...............................................................
[  0.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[  6.25%] ··· Running timedelta.TimedeltaConstructor.time_from_components                                      15.4±0.4μs
[ 12.50%] ··· Running timedelta.TimedeltaConstructor.time_from_datetime_timedelta                              8.06±0.3μs
[ 18.75%] ··· Running timedelta.TimedeltaConstructor.time_from_int                                             6.24±0.2μs
[ 25.00%] ··· Running timedelta.TimedeltaConstructor.time_from_iso_format                                      12.1±0.4μs
[ 31.25%] ··· Running timedelta.TimedeltaConstructor.time_from_missing                                        2.07±0.03μs
[ 37.50%] ··· Running timedelta.TimedeltaConstructor.time_from_np_timedelta                                   5.06±0.09μs
[ 43.75%] ··· Running timedelta.TimedeltaConstructor.time_from_string                                          6.52±0.1μs
[ 50.00%] ··· Running timedelta.TimedeltaConstructor.time_from_unit                                            6.66±0.1μs
[ 50.00%] · For pandas commit hash 8acdf801:
[ 50.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[ 50.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[ 56.25%] ··· Running timedelta.TimedeltaConstructor.time_from_components                                      15.5±0.3μs
[...]

PR #21827, opened 2018-07-09
ENH: Add Timedelta Support to JSON Reader with orient=table (#21140)

### LINKED ISSUE #21140: Add Timedelta Support to JSON Reader with orient="table"

fjdiod:
- [ ] closes #21140
- [ ] tests added / passed
- [ ] whatsnew entry


pandas/io/json/table_schema.py
```diff
@@ -306,10 +306,9 @@ def parse_table_schema(json, precise_float):
         raise NotImplementedError('table="orient" can not yet read timezone '
                                   'data')
 
-    # No ISO constructor for Timedelta as of yet, so need to raise
-    if 'timedelta64' in dtypes.values():
-        raise NotImplementedError('table="orient" can not yet read '
-                                  'ISO-formatted Timedelta data')
+    for col, dtype in dtypes.items():
```
WillAyd: Is this block necessary? Assumed the subsequent `astype` call would take care of this

WillAyd (changes requested):
Nice change - one comment to address

pandas/io/json/table_schema.py
```diff
@@ -306,10 +306,9 @@ def parse_table_schema(json, precise_float):
...
+        if dtype == 'timedelta64[ns]':
+            df[col] = df[col].apply(Timedelta)
```
jreback: you would not convert this way, use ``pd.to_timedelta``

pandas/io/json/table_schema.py
```diff
@@ -306,10 +306,9 @@ def parse_table_schema(json, precise_float):
         raise NotImplementedError('table="orient" can not yet read timezone '
                                   'data')
 
-    # No ISO constructor for Timedelta as of yet, so need to raise
```
jreback: was this not hit in tests prior to this PR?

jreback (changes requested):

pandas/io/json/table_schema.py
WillAyd: Should have been covered by `test_read_json_table_orient_raises` through parametrization

pandas/io/json/table_schema.py
fjdiod: So, support for iso-format timedeltas should be added to `astype` function?

pandas/io/json/table_schema.py
fjdiod: Can `pd.to_timedelta` parse iso-formated timedeltas now? Should support for this format be added to this function?

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -183,7 +183,10 @@ cpdef convert_to_timedelta64(object ts, object unit):
             ts = cast_from_unit(ts, unit)
             ts = np.timedelta64(ts)
     elif is_string_object(ts):
-        ts = np.timedelta64(parse_timedelta_string(ts))
+        if len(ts) > 0 and ts[0] == 'P':
```
WillAyd: Hmm what problem are you trying to solve with this addition? ISO format support should have been added back in #19191 as a precursor to this PR

pandas/io/json/table_schema.py
```diff
@@ -308,7 +308,7 @@ def parse_table_schema(json, precise_float):
 
     for col, dtype in dtypes.items():
         if dtype == 'timedelta64[ns]':
-            df[col] = df[col].apply(Timedelta)
+            df[col] = to_timedelta(df[col])
 
     df = df.astype(dtypes)
```
WillAyd: Is there any conflict between this and the loop above? Wondering if we should be removing any timedelta calls from the `dtypes` dict since they get accounted for in the loop

WillAyd (changes requested):

pandas/_libs/tslibs/timedeltas.pyx
fjdiod: `pd.to_timedelta` was not converting iso-formated strings
In `pd.Timedelta` constructor `parse_iso_format_string` function is used:
```python
if len(value) > 0 and value[0] == 'P':
    value = parse_iso_format_string(value)
else:
    value = parse_timedelta_string(value)
```
while in `pd.to_timedelta` `array_to_timedelta64` is used. Which is not using `parse_iso_format_string`, actually there're only two occurrences of this function in `timedeltas.pyx`, its definition and inside `pd.Timedelta`.
Also, why have `parse_iso_format_string` and `parse_timedelta_string` simultaneously?

WillAyd: Ah OK I see. If that's the case then it would be better to do a separate PR as a pre-cursor to this one. That way we can communicate the enhancement and ensure appropriate testing.

Since you started on that, do you want to open an issue about `pd.to_timedelta` not accepting the ISO duration and send a separate PR to fix accordingly? Can come back to this one after the fact

fjdiod: OK, I'll open an issue

fjdiod: I've opened an issue  #21877

pandas/io/json/table_schema.py
WillAyd: Should be able to change this now

pandas/io/json/table_schema.py
```diff
@@ -6,7 +6,7 @@
 import warnings
 
 import pandas._libs.json as json
-from pandas import DataFrame
+from pandas import DataFrame, Timedelta
```
WillAyd: Import `to_timedelta` instead

WillAyd (changes requested):

pandas/io/json/table_schema.py
fjdiod: Oh, sorry, I've pushed this accidentally

fjdiod:
Hello, what is wrong with Travis test? Is it a problem on the testing side or on mine?

fjdiod:
@WillAyd is there anything else that I should change?

WillAyd (approved):
Only one minor fix up but otherwise lgtm

pandas/io/json/table_schema.py
jreback: hmm, are you sure you need to do this? I believe ``.astype(dtypes)`` should handle this, and if not we should fix that.

jreback (changes requested):

pandas/io/json/table_schema.py
fjdiod: For now 
```python
delta = pd.Timedelta(1e9).isoformat()
pd.DataFrame([delta]).astype('timedelta64[ns]')
```
results in error.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -121,6 +121,7 @@ Other Enhancements
 - :class:`IntervalIndex` has gained the :meth:`~IntervalIndex.set_closed` method to change the existing ``closed`` value (:issue:`21670`)
 - :func:`~DataFrame.to_csv` and :func:`~DataFrame.to_json` now support ``compression='infer'`` to infer compression based on filename (:issue:`15008`)
 - :func:`to_timedelta` now supports iso-formated timedelta strings (:issue:`21877`)
+- :func:`read_json` now parse timedelta with `orient='table'` (:issue:`21140`)
```
WillAyd: I might have missed this on previous comment but there's a small typo here. `parse` -> `parses`

fjdiod: @WillAyd I've pushed fix, but now there is a merge conflict. How should I deal with it on github?

WillAyd: Better to deal locally and re-push. Assuming you have things set up as mentioned in the pandas contributing guide do this on your local branch

```sh
git fetch upstream
git merge upstream/master
```

You'll probably get a merge conflict there, so fix that up and do:

```sh
git merge --continue
```

Everything should be resolved then so re-push after that

fjdiod:
Hello, @jreback 
How should I proceed with `astype`? Should support for iso-formatted timedeltas be added there?
Now `astype` could not parse timedeltas and datetimes from strings. So, wouldn't it be inconsistent to add support for iso-format?

jreback:
i believe we handle astype for date times by calling to_datetme we should do the same for timedelta (which then would make this automatically work)

WillAyd:
@fjdiod if you are interested it would be preferable to have a new issue and PR, similar to what you just did for the `to_timedelta` change. Sorry for all of the pre-cursors!

jreback:
can you merge master and update

jreback:
can you merge master and update

WillAyd:
Closing as stale though would be great to have this. Please ping if you'd like to pick this back up

PR #21843, opened 2018-07-10
[REF] Move constructor helpers to EA Mixin classes

jbrockmendel:
After this:

- Tests
- Comparisons
- Tests
- Deal with the fact that DatetimeIndex caches ranges, only feasible because of immutability
- Tests!
- formatting/`__repr__` methods
- Teeeeeests
- Double-check docstring accuracy

### DIFF
--- a/pandas/core/arrays/datetimelike.py
+++ b/pandas/core/arrays/datetimelike.py
@@ -403,7 +403,7 @@ class DatetimeLikeArrayMixin(AttributesMixin):
             td = Timedelta(self.freq)
             return op(self, td * other)
 
-        # We should only get here with DatetimeIndex; dispatch
+        # We should only get here with Datetime Array/Index; dispatch
         # to _addsub_offset_array
         assert not is_timedelta64_dtype(self)
         return op(self, np.array(other) * self.freq)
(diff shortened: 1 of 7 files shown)


jorisvandenbossche:
Are there methods left that can rather simply be moved?

jbrockmendel:
> Are there methods left that can rather simply be moved?

For the arithmetic/comparison methods (and their dependencies) that I've mostly been focused on, not really.  Most of what's left either a) dispatches to the base `Index` class and so is non-trivial, or b) requires more of `__new__` to be ported (`DatetimeIndexOpsMixin.shift` passes `start` and `end` kwargs to the constructor which are not currently handled).

There are a few more that can be moved with Index-wrapping left in the Index subclasses.  `DatetimeIndex.to_period`, `PeriodIndex.to_timestamp`, ...

There are a few that _would_ be simple but I'm not yet sure if they _should_ be moved: `inferred_type`, `is_all_dates`, `is_type_compatible`, `PeriodIndex.tz_convert`, `PeriodIndex.tz_localize`

jbrockmendel:
Based on tests it looks like this is _not_ the correct next step.  Closing.

PR #21844, opened 2018-07-10
[CLN] cleanup import reverse-dependencies

jbrockmendel:
Parts of the code are made more difficult to reason about by depending on `pandas.io.formats`, which has dependencies all over the code.

This PR starts in on this by replacing `import Foo` with `import ABCFoo` where possible.  Along the way it cleans up import order.

### DIFF
--- a/pandas/io/clipboards.py
+++ b/pandas/io/clipboards.py
@@ -1,7 +1,10 @@
 """ io on the clipboard """
-from pandas import compat, get_option, option_context, DataFrame
-from pandas.compat import StringIO, PY2, PY3
 import warnings
+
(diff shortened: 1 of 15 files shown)


WillAyd:
Think this will have any impact on the overall import time? @TomAugspurger do you know where that original issue was?

jreback:
thanks @jbrockmendel 

yeah this is almost cosmetic as python optimizes the imports, but it is nicer if can consolidate the imports at the top and avoid importing-chain deps.

if you can contrive a linter to actually conform the import orderings (e.g. _libs, util, compat, core), would be great, but might be tricky.

PR #21703, opened 2018-07-02
BUG : Adding *kwargs for name retention in Index args

uds5501:
- [x] closes #21697
- [ ] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] whatsnew entry

PR #21848, opened 2018-07-10
Propagate name in more cases when using `to_datetime`

### LINKED ISSUE #21697: BUG: to_datetime with Index arg does not retain name in all cases
```
#Example

In [17]: idx = pd.Index([15e9], name='name')

In [18]: pd.to_datetime(idx, errors='ignore', box=True, unit='s').name

In [19]: pd.__version__
Out[19]: '0.24.0.dev0+193.gdc45fba'
```
There are certain cases where the name attribute of the passed `Index` are not retained after `to_datetime`.

The `name` kwarg looks to be missing from these locations:

https://github.com/pandas-dev/pandas/blob/dc45fbafef172e357cb5decdeab22de67160f5b7/pandas/core/tools/datetimes.py#L102

https://github.com/pandas-dev/pandas/blob/dc45fbafef172e357cb5decdeab22de67160f5b7/pandas/core/tools/datetimes.py#L139

https://github.com/pandas-dev/pandas/blob/dc45fbafef172e357cb5decdeab22de67160f5b7/pandas/core/tools/datetimes.py#L344

eyurtsev:
Closes #21697

Haven't ran tests yet

### DIFF
--- a/pandas/core/tools/datetimes.py
+++ b/pandas/core/tools/datetimes.py
@@ -248,6 +248,8 @@ def _convert_listlike_datetimes(arg, box, format, name=None, tz=None,
                         if box:
                             from pandas import Index
                             return Index(tz_results, name=name)
+                        else:
+                            return tz_results
                 except tslibs.OutOfBoundsDatetime:
                     if errors == 'raise':
                         raise
--- a/pandas/core/tools/datetimes.py
+++ b/pandas/core/tools/datetimes.py
@@ -98,13 +98,13 @@ def _convert_and_box_cache(arg, cache_array, box, errors, name=None):
     result = Series(arg).map(cache_array)
     if box:
         if errors == 'ignore':
-            return Index(result)
+            return Index(result, name=name)
         else:
             return DatetimeIndex(result, name=name)
     return result.values
 
 
-def _return_parsed_timezone_results(result, timezones, box, tz):
+def _return_parsed_timezone_results(result, timezones, tz):
     """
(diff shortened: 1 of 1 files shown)


pandas/core/tools/datetimes.py
```diff
@@ -250,8 +243,11 @@ def _convert_listlike_datetimes(arg, box, format, name=None, tz=None,
                     result, timezones = array_strptime(
                         arg, format, exact=exact, errors=errors)
                     if '%Z' in format or '%z' in format:
-                        return _return_parsed_timezone_results(
-                            result, timezones, box, tz)
+                        tz_results = _return_parsed_timezone_results(
```
eyurtsev: missing return here actually

jreback:
https://github.com/pandas-dev/pandas/pull/21703 is also proposed for this issue, though its currently failing. So happy to have you add tests.

eyurtsev:
I was trying to stick to the convention in the file, but wanted is there a reason for local imports? Is that the convention in the codebase or is it OK to float the imports to the top?

```+                            from pandas import Index
+                            return Index(tz_results, name=name)```

jreback (changes requested):
needs tests. also you are changing some extraneous code, can you limit changes to what is needed to solve the issue.

jreback (changes requested):
needs tests and a whatsnew note

jreback:
closing as stale. would still take this PR if updated.

PR #21849, opened 2018-07-11
BUG: groupby with no non-empty groups, #21624

### LINKED ISSUE #21624: groupby w/ only NULL-groups crashes since 0.23

```python
import numpy as np
import pandas as pd
import pandas.testing as pdt


# works:
df1 = pd.DataFrame({
    'g': [None, 'a'],
    'x': 1,
})
actual1 = df1.groupby('g')['x'].transform('sum')
expected1 = pd.Series([np.nan, 1.], name='x')

pdt.assert_series_equal(
    actual1,
    expected1,
)


# crashes:
df2 = pd.DataFrame({
    'g': [None],
    'x': 1,
})
actual2 = df2.groupby('g')['x'].transform('sum')
expected2 = pd.Series([np.nan], name='x')  # crashes with "ValueError: Length of passed values is 1, index implies 0"

pdt.assert_series_equal(
    actual2,
    expected2,
)
```
#### Problem description
`groupby` ignores groups that contain NULL-elements in any of the group columns. In that case, results of `transform` are NULL (NaN for floats, NaT for time, None for objects). The question is what happens if there are only "NULL groups":

- before pandas 0.23: an object column with `None` objects is created
- pandas 0.23: crash (`ValueError: Length of passed values is 1, index implies 0`)

#### Expected Output
A NULL-column according to the input data type.

<details>

```
INSTALLED VERSIONS
------------------
commit: None
python: 2.7.13.final.0
python-bits: 64
OS: Linux
OS-release: 4.9.49-moby
machine: x86_64
processor:
byteorder: little
LC_ALL: en_US.UTF-8
LANG: en_US.UTF-8
LOCALE: None.None

pandas: 0.23.1
pytest: 3.4.0
pip: 10.0.1
setuptools: 39.1.0
Cython: 0.28.2
numpy: 1.14.3
scipy: 1.0.0
pyarrow: 0.9.0
xarray: None
IPython: 5.7.0
sphinx: 1.6.7
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.2.1
tables: None
numexpr: 2.6.5
feather: None
matplotlib: 2.2.2
openpyxl: 2.5.0
xlrd: None
xlwt: None
xlsxwriter: 0.8.4
lxml: 3.8.0
bs4: None
html5lib: 1.0.1
sqlalchemy: 1.2.7
pymysql: None
jinja2: 2.8.1
s3fs: 0.1
fastparquet: None
pandas_gbq: None
pandas_datareader: None
```
</details>

lopez86:
Should close #21624. Didn't see any issues running test_fast.sh on my laptop so I don't think it broke anything in the process. The problem here is that when there is only a null group, then `self.result_index` is empty, but `ids` is just a list of -1s, so a ValueError is raised when trying to create the series.


jreback:
tests!

lopez86:
Added in a test for this specific case. The other case where there are some but not all nans present is already covered extensively in the tests.

WillAyd (changes requested):

pandas/tests/groupby/test_transform.py
```diff
@@ -782,3 +782,14 @@ def test_any_all_np_func(func):
...
     tm.assert_series_equal(res, exp)
+
+
+def test_transform_with_all_nan():
+    # GH 21624
+    df = DataFrame({'groups': [np.nan, np.nan, np.nan],
+                    'values': [1, 2, 3]})
+
+    grouped = df.groupby('groups')
+    summed = grouped['values'].transform('sum')
```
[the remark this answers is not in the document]
lopez86: Sure

pandas/tests/groupby/test_transform.py
```diff
@@ -782,3 +782,14 @@ def test_any_all_np_func(func):
...
+
+
+def test_transform_with_all_nan():
+    # GH 21624
+    df = DataFrame({'groups': [np.nan, np.nan, np.nan],
+                    'values': [1, 2, 3]})
+
+    grouped = df.groupby('groups')
+    result = grouped['values'].transform('sum')
+    expected = Series([np.nan, np.nan, np.nan], name='values')
```
WillAyd: Hmm maybe I'm missing the point but shouldn't these be `6` and not `np.nan`?

WillAyd (changes requested):

pandas/tests/groupby/test_transform.py
lopez86: Currently, the behavior is to output `np.nan` for anything in a group with label `np.nan`. Making this output `6` would be changing the interpretation of `np.nan` from `not in any group` to `in a group with label np.nan`, which is going to be a pretty significant change in how things work. The treatment here of ignoring NA values is specified in the groupby docs. 

For example, see `test_groupby_transform_with_nan_group()` in `test_transform.py`

jreback:
pls rebase, just reorged the groupby code a bit.

lopez86:
Should be ready for review now

pandas/core/groupby/ops.py
```diff
@@ -235,7 +235,7 @@ def size(self):
         if ngroup:
             out = np.bincount(ids[ids != -1], minlength=ngroup)
         else:
-            out = ids
+            out = []
```
WillAyd: Hmm something just seems strange to me about this. So your test example works fine, but what happens if we have a DataFrame that has other groups besides the all-NA one? Can you add a test for that and make sure that still works?

WillAyd (changes requested):

pandas/core/groupby/ops.py
lopez86: Done. In that case, the else block should never be used

pandas/tests/groupby/test_transform.py
```diff
@@ -782,3 +782,23 @@ def test_any_all_np_func(func):
...
+    tm.assert_series_equal(result, expected)
+
+    # Make sure the standard case still works too
+    df = DataFrame({'groups': [np.nan, 'A', 'A', 'B', 'B'],
```
WillAyd: Since this follows the same pattern as the code above it would be preferable to parametrize it instead of re-coding

WillAyd (changes requested):
Just thinking more through this as other similar issues have popped up - are we now invalidating the docs below?

http://pandas.pydata.org/pandas-docs/stable/missing_data.html#na-values-in-groupby

If so we should do a separate PR to update docs as well, though somewhat surprised this didn't break any other tests

lopez86:
I think nothing is breaking because this shouldn't be changing anything other than preventing a ValueError in an edge case. I think there is some ambiguity in the docs where we want to ignore NA values in group labels but also have transform return a series that's the same shape as the input. It would definitely be good to have the behavior of transform in the presence of NA group labels explicitly spelled out somewhere.

pandas/tests/groupby/test_transform.py
lopez86: This is done in the latest update

WillAyd (approved):

jreback (changes requested):
can you check out the current 23.X branch and see if your new test pass). its possible this was fixed in master, but I don't think the change was backported.

WillAyd:
@jreback just tried locally and this wasn't working on master, so I think this change is still necessary

WillAyd (changes requested):
Can you add a whatsnew note for v0.23.5 and merge latest updates from master?

WillAyd:

PR #21854, opened 2018-07-11
[CLN] resolve circular Period dependency, prepare setup.py

jbrockmendel:
For a long time there has been a comment in `_libs.__init__` saying it would be nice to import `Period` directly but that is not possible due to circular imports.  This resolves that issue.

Also does some cleanup in setup.py, motivated by the goals of a) using `cythonize` and b) implementing test coverage for cython files.  I've gotten those working locally, but they involve big diffs, so this gets some of the easy stuff out of the way.


setup.py
```diff
@@ -103,27 +97,30 @@ def is_platform_mac():
 
 
 class build_ext(_build_ext):
-    def build_extensions(self):
+    @classmethod
+    def render_templates(cls, pxifiles):
```
jbrockmendel: I needed to be able to call this method separately to get `cythonize` working.

setup.py
```diff
@@ -53,24 +45,26 @@ def is_platform_mac():
 }
```
jbrockmendel: If there are no objections, I'd like to move this dictionary up above the `min_cython_version = '0.28.2` line

pandas/_libs/tslibs/resolution.pyx
```diff
@@ -574,6 +572,10 @@ cdef class _FrequencyInferer(object):
         if len(self.ydiffs) > 1:
             return None
 
+        # lazy import to prevent circularity
+        # TODO: Avoid non-cython dependency
+        from pandas.core.algorithms import unique
+
```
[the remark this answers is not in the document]
jreback: sure that would be ok

setup.py
jreback: sure

setup.py
```diff
@@ -53,24 +45,26 @@ def is_platform_mac():
...
 
 
+# The import of Extension must be after the import of Cython, otherwise
+# we do not get the appropriately patched class.
```
jreback: this seems odd?

setup.py
```diff
@@ -53,24 +45,26 @@ def is_platform_mac():
...
 from distutils.command.build import build  # noqa:E402
-from distutils.command.build_ext import build_ext as _build_ext  # noqa:E402
 
 try:
     if not _CYTHON_INSTALLED:
         raise ImportError('No supported version of Cython installed.')
     try:
-        from Cython.Distutils.old_build_ext import old_build_ext as _build_ext  # noqa:F811,E501
+        from Cython.Distutils.old_build_ext import old_build_ext as _build_ext
     except ImportError:
```
jreback: yes I think that is ok to remove

setup.py
```diff
@@ -471,8 +468,6 @@ def pxd(name):
 
 tseries_depends = np_datetime_headers + ['pandas/_libs/tslibs/np_datetime.pxd']
 
-# some linux distros require it
```
jreback: hmm do we need this?

jreback (changes requested):

setup.py
jbrockmendel: Found the appropriate reference in the cython [docs](https://cython.readthedocs.io/en/latest/src/reference/compilation.html):


> Note that when using setuptools, you should import it before Cython as setuptools may replace the Extension class in distutils. Otherwise, both might disagree about the class to use here.

setup.py
jbrockmendel: Apparently not.  It has apparently sat there unused for a while.

pandas/_libs/tslibs/resolution.pyx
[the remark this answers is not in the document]
jbrockmendel: great, will do this in follow up

setup.py
jreback: that is so weird, but ok

jreback (approved):

jreback:
thanks! happy to take these cleaning PRs!

Issue #21852, opened 2018-07-11
[CLN] Replace np_datetime code with numpy versions


jbrockmendel:
@jreback we discussed this briefly, hopefully you'll have better luck with it than me.

pandas/_libs/src/datetime/np_datetime.c and np_datetime_strings.c share a ton of code with numpy:

https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/datetime.c
https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/_datetime.h
https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/datetime_strings.c
https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/datetime_strings.h

My attempts so far to "get at" these numpy functions via `#include` have failed, someone with stronger C-fu than me will have to try.  The testing ground I've been trying on is src/ujson/python/objToJSON.c since all of the np_datetime functions it uses have drop-in counterparts in numpy:

PANDAS_DATETIMEUNIT --> NPY_DATETIMEUNIT
PANDAS_FR_ns --> NPY_FR_ns (ditto s, ms, us)
pandas_datetimestruct --> npy_datetimestruct

get_datetime_iso_8601_strlen --> identical

make_iso_8601_datetime has fewer args in the pandas version, largely since we removed unnecessary ones a few months ago.  Adding hard-coded args in should make them equivalent.

convert_pydatetime_to_datetimestruct also looks like dummy arguments need to be re-introduced to make them identical.

convert_datetime_to_datetimestruct has a different first-argument (again one we simplified) but is functionally identical.
convert_datetimestruct_to_datetime ditto.

parse_iso_8601_datetime has non-trivial differences IIRC.

jbrockmendel:
Is the conclusion from #21863 that this is either infeasible or undesirable?

jreback:
no this is possible ; just wanted to see if we could document the original reasons for using vendors approaches

jbrockmendel:
This is not actionable until https://github.com/numpy/numpy/issues/9675 is addressed there.  I'll keep an eye on that; closing.

PR #21863, opened 2018-07-11
Replaced PANDAS_DATETIMEUNIT with NPY_DATETIMEUNIT

WillAyd:
progress towards #21852

- [X] tests added / passed
- [ ] whatsnew entry

Working on my C-fu as well. There may be a better way to do this (ex: even adding NPY_DATETIMEUNIT to Cython/includes) but figured I'd offer this up for review

@jbrockmendel

### DIFF
--- a/pandas/_libs/tslibs/conversion.pxd
+++ b/pandas/_libs/tslibs/conversion.pxd
@@ -32,21 +32,3 @@ cdef maybe_datetimelike_to_i8(object val)
 cdef int64_t tz_convert_utc_to_tzlocal(int64_t utc_val, tzinfo tz)
 
 cpdef datetime localize_pydatetime(datetime dt, object tz)
-
-cdef extern from "numpy/ndarraytypes.h":
-    ctypedef enum NPY_DATETIMEUNIT:
-        NPY_FR_ERROR
-        NPY_FR_Y
-        NPY_FR_M
-        NPY_FR_W
-        NPY_FR_D
-        NPY_FR_h
-        NPY_FR_m
-        NPY_FR_s
-        NPY_FR_ms
-        NPY_FR_us
-        NPY_FR_ns
-        NPY_FR_ps
-        NPY_FR_fs
-        NPY_FR_as
-        NPY_FR_GENERIC
--- a/pandas/_libs/tslibs/conversion.pxd
+++ b/pandas/_libs/tslibs/conversion.pxd
@@ -7,6 +7,7 @@ from numpy cimport int64_t, int32_t
 
 from np_datetime cimport pandas_datetimestruct
 
+
 cdef class _TSObject:
     cdef:
         pandas_datetimestruct dts      # pandas_datetimestruct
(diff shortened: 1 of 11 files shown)


pandas/_libs/tslibs/conversion.pxd
```diff
@@ -33,3 +32,21 @@ cdef maybe_datetimelike_to_i8(object val)
 cdef int64_t tz_convert_utc_to_tzlocal(int64_t utc_val, tzinfo tz)
 
 cpdef datetime localize_pydatetime(datetime dt, object tz)
+
+cdef extern from "numpy/ndarraytypes.h":
```
WillAyd: Not sure if there is a better way to do this declaration. Right now I copy/pasted in the three files where it is need but is there a shared place to put that for shared access across the modules?

jbrockmendel: Since this is done in np_datetime.pxd, I'd just cimport from there rather than copying this.

jbrockmendel:
Aside from that note on repeated imports, this looks like a step in the right direction.  The other super-easy change-over is `pandas_datetimestruct`.

pandas/_libs/tslibs/conversion.pxd
jbrockmendel: Since it is in np_datetime.pxd, it can be `cimport`ed from there.

jbrockmendel:
I still can’t find a way to access functions in numpy’s _datetime.h.  Not even sure what is possible in C-land.  @chris-b1 are you the right person to ask about this?

WillAyd:
@jbrockmendel not sure what you are looking for out of that but it doesn't appear that header is included in numpy's `core/include` directory along with the rest of the headers. You might have to mess around with `setup.py` if you are trying to include header files outside of that location:

https://github.com/pandas-dev/pandas/blob/5d0daa0522730a6f999ceaf328f63f03dd62d0b4/setup.py#L128

Though that most likely violates a lot of best practices

WillAyd:
Though looking at it I don't believe `src/multiarray/_datetime.h` is included as part of your numpy installation in site-packages so may not be possible with how the code is currently structured (could be wrong)

jbrockmendel:
> not sure what you are looking for out of that but it doesn't appear that header is included in numpy's core/include directory along with the rest of the headers

The hope was that it is somehow passed through via an `#include` in one of the files that is part of core/include.  After all, numpy itself somehow manages to call these functions

chris-b1:
Yeah, even if it were possible I don't think we should be depending on private API of numpy - presumably why some bits (e.g. datetime string parsing) were vendored in the first place.

jreback:
cc @charris

do you recall if these existed in numpy < 1.6 ?

WillAyd:
AFAICT goes at least back to 1.4

jbrockmendel:
AFAICT the code was vendored b/c at the time pandas supported older numpy versions that did not yet have `datetime64`.  Also because pandas handles timezones differently in `np_datetime_strings`.  It'd still be nice to not re-implement the large portion that is identical.

It would also be super-convenient to be able to use the numpy versions in src/ujson/python/obj2JSON.c, as that would let us de-couple the remaing src/datetime code.

jreback:
thanks @WillAyd and @jbrockmendel for the comments. Agree that the vendored code, just wanted the rationale documented here :>

PR #21865, opened 2018-07-11
ENH: Change DatetimeIndex.time to also return timezone info

### LINKED ISSUE #21358: ENH: Add timetz attribute to DatetimeIndex and Timestamp
Currently `DatetimeIndex.time` returns a numpy array of local `datetime.time`s. In #18461, the resulting `datetime.time`s gained timezone information (but was an undocumented changed) and was subsequently reverted in #21281. Having `.time` return timezone information would still be a nice enhancement.

If we allow `DatetimeIndex.time` to return timezone information, [`timeseries.rst`](http://pandas.pydata.org/pandas-docs/stable/timeseries.html?highlight=weekofmonth#time-date-components) will need to be changed since it currently indicates that `.time` does not return timezone information.

jquinon:
- [ ] closes #21358
- [ ] tests added / passed
- [ ] whatsnew entry

Mainly just reverted changes from GH#21281 and updated docs. Also had to change one test added in GH#21281.

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -84,7 +84,7 @@ Other Enhancements
 - :meth:`Series.nlargest`, :meth:`Series.nsmallest`, :meth:`DataFrame.nlargest`, and :meth:`DataFrame.nsmallest` now accept the value ``"all"`` for the ``keep`` argument. This keeps all ties for the nth largest/smallest value (:issue:`16818`)
 - :class:`IntervalIndex` has gained the :meth:`~IntervalIndex.set_closed` method to change the existing ``closed`` value (:issue:`21670`)
 - :func:`~DataFrame.to_csv` and :func:`~DataFrame.to_json` now support ``compression='infer'`` to infer compression based on filename (:issue:`15008`)
--
+- Changed DatetimeIndex.time to also return timezone information. Mainly undid changes from :issue:`21821. (:issue:`21358)
 
 .. _whatsnew_0240.api_breaking:
 
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -84,7 +84,7 @@ Other Enhancements
 - :meth:`Series.nlargest`, :meth:`Series.nsmallest`, :meth:`DataFrame.nlargest`, and :meth:`DataFrame.nsmallest` now accept the value ``"all"`` for the ``keep`` argument. This keeps all ties for the nth largest/smallest value (:issue:`16818`)
 - :class:`IntervalIndex` has gained the :meth:`~IntervalIndex.set_closed` method to change the existing ``closed`` value (:issue:`21670`)
 - :func:`~DataFrame.to_csv` and :func:`~DataFrame.to_json` now support ``compression='infer'`` to infer compression based on filename (:issue:`15008`)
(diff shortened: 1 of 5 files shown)


doc/source/whatsnew/v0.24.0.txt
```diff
@@ -84,7 +84,7 @@ Other Enhancements
 - :meth:`Series.nlargest`, :meth:`Series.nsmallest`, :meth:`DataFrame.nlargest`, and :meth:`DataFrame.nsmallest` now accept the value ``"all"`` for the ``keep`` argument. This keeps all ties for the nth largest/smallest value (:issue:`16818`)
 - :class:`IntervalIndex` has gained the :meth:`~IntervalIndex.set_closed` method to change the existing ``closed`` value (:issue:`21670`)
 - :func:`~DataFrame.to_csv` and :func:`~DataFrame.to_json` now support ``compression='infer'`` to infer compression based on filename (:issue:`15008`)
--
+- Changed DatetimeIndex.time to also return timezone information. (:issue:`21358)
```
jschendel: Missing a closing backtick for the issue number. Can you also add an :attr: ref for DatetimeIndex.time?

jquinon: Does that look good?

pandas/core/arrays/datetimes.py
```diff
@@ -577,15 +577,7 @@ def time(self):
         """
         Returns numpy array of datetime.time. The time part of the Timestamps.
```
mroeschke: Could you expand this docstring to also mention that the `datetime.time` objects are in local time with associated timezone information.

jquinon: Does that look good?

mroeschke: Small nit: could you replace `, and contains` with `with`. Otherwise LGTM

jquinon: No problem

jreback:
does the standard library preserve for ``datetime.datetime.time()`` ?

mroeschke:
`.time()` does not but `.timetz()` does:

```
In [12]: pytz.UTC.localize(datetime(2000, 1, 1)).time()
Out[12]: datetime.time(0, 0)

In [13]: pytz.UTC.localize(datetime(2000, 1, 1)).timetz()
Out[13]: datetime.time(0, 0, tzinfo=<UTC>)
```

I guess if we want to follow the standard library naming it should be naive, but I think its useful knowing in which timezone the local time resides.

An alternative is to depreciate the `.time` property for a `time(tz=False)` method that mirrors the standard library, has an option to return tz information and is maybe less API breaking.

Thoughts @jreback?

jreback:
hmm its annoying that the std library does this with 2 methods, but I guess can't easily 'remove' a tz from a time (e.g. no ``.tz_localize(None)``.

mroeschke:
Yeah that's why I am more inclined to chose the method route to mimic the standard library but at least have a keyword argument (and not a second method) to include the tz in the output or not.

jquinon:
I can modify or cancel the PR as you guys see fit, just let me know when a decision is made!

jorisvandenbossche:
I would personally follow the standard library and leave it as is as a naive time. 
You can always get the timezone separately if you want.

jorisvandenbossche:
@jquinon thanks for working on it anyhow!

PR #21870, opened 2018-07-12
[CLN] De-privatize commonly-used functions

jbrockmendel:
Also updated numpy_helper to not use things from numpy's deprecated C API.  This won't get rid of the warnings since cython still causes them, but it's still nice.

Not sure how to lint for this (or if we really want to), will see if google knows.


jreback:
looks ok, something failing.

jbrockmendel:
Messed up whitespace in rebasing, should be fixed now.

jbrockmendel:
CI failure was a ResourceWarning.  Had to edit something to re-push, so de-privatized _ensure_categorical.

jbrockmendel:
ping.  I'd like to get this out of the way before inevitable merge conflicts arise.

jreback:
ahh i merged in a funny order, can you rebase

jbrockmendel:
Done

pandas/core/common.py
```diff
@@ -120,7 +120,7 @@ def is_bool_indexer(key):
     return False
 
 
-def _default_index(n):
+def default_index(n):
```
jreback: this entire module, is technically public, though we have a mix of public/private here anyway, so have to be careful about changing public functions here (obviously this is private -> public).

jreback: let' move this to ``pandas.index.base`` , ok to call it ``default_index``

jreback (changes requested):

pandas/core/common.py
jbrockmendel: done

jreback (approved):

PR #21872, opened 2018-07-12
[REF] Move comparison methods to EAMixins, share code

jbrockmendel:
Changes an old usage numpy's C API that is deprecated.  This won't get rid of the warnings because cython hasn't changed it, but still.

Also stops making copies of offsets since they are now immutable.

Handles a handful of changes requested in the last pass: de-privatizes _quarter_to_myear (plus bonus docstring), renames _generate --> _generate_range, comments in is_list_like

renames arrays.timedelta --> arrays.timedeltas to match core.indexes

Implements comparison methods in DatetimeArray and TimedeltaArray, cleans up some Index code that is no longer needed as a result.

Makes some progress on sharing code between TDI and DTI constructors (most of which we want to move up to the array classes)


pandas/_libs/tslibs/period.pyx
```diff
@@ -1877,6 +1897,8 @@ def _quarter_to_myear(year, quarter, freq):
             year -= 1
 
     return year, month
+    # FIXME: if quarter is None then `month` won't be defined here.
+    #   just disallow quarter == None?
```
gfyoung: cc @jreback

pandas/_libs/tslibs/period.pyx
```diff
@@ -1860,13 +1860,33 @@ cdef int64_t _ordinal_from_fields(year, month, quarter, day,
...
     if quarter is not None:
-        year, month = _quarter_to_myear(year, quarter, freq)
+        year, month = quarter_to_myear(year, quarter, freq)
 
     return period_ordinal(year, month, day, hour,
                           minute, second, 0, 0, base)
 
 
-def _quarter_to_myear(year, quarter, freq):
+def quarter_to_myear(year, quarter, freq):
```
jreback: ideally type these, where is quarter passed as None?

pandas/core/arrays/timedeltas.py
```diff
@@ -21,6 +21,19 @@
 from pandas.tseries.frequencies import to_offset
```
jreback: can you change to absolute imports, I think we are moving away from the relative ones fro readability.

pandas/core/arrays/timedeltas.py
```diff
@@ -219,6 +267,19 @@ def _evaluate_with_timedelta_like(self, other, op):
 
         return NotImplemented
 
+    # ----------------------------------------------------------------
+    # Comparison Methods
+
+    @classmethod
+    def _add_comparison_methods(cls):
```
jreback: so I would define the comparison methods on the class itself, then this should be trivial to change to use the EAMixins

jreback (changes requested):

pandas/core/arrays/datetimelike.py
```diff
@@ -435,3 +469,68 @@ def _addsub_offset_array(self, other, op):
...
             kwargs['freq'] = 'infer'
         return type(self)(res_values, **kwargs)
+
+    # --------------------------------------------------------------
+    # Comparison Methods
+
+    def _evaluate_compare(self, other, op):
+        """
+        We have been called because a comparison between
+        8 aware arrays. numpy >= 1.11 will
```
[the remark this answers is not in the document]
jbrockmendel: This is the existing docstring in indexes.datetimelike.  I'm open to suggestions.

pandas/_libs/tslibs/period.pyx
jbrockmendel: It is used once in period.pyx and once in arrays.period, always with quarter non-None.  I'll make that required.

pandas/core/arrays/timedeltas.py
jbrockmendel: I agree we should move in this direction, but prefer to do it separately.  ATM the PR still broadly resembles cut/paste.

jbrockmendel:
Ping.  I’d like to get a move on #21907 and the caching issues that will come up with this.

jreback (approved):

jreback:
thanks @jbrockmendel

PR #21878, opened 2018-07-12
Cleanup cimports, implement bits of numpy_helper in util.pxd

jbrockmendel:
Removes a few unnecessary uses of `cnp.import_array()`

Uses cython's conditional compilation to avoid runtime PY2/PY3 checks

Arranges cython imports in stdlib-->3rd party--> internal order.

Separates out the one part of `libresolution._FrequencyInferer` that we actually _do_ want to keep in cython

Couple of small lintings in groupby.pyx

### DIFF
--- a/pandas/_libs/index.pyx
+++ b/pandas/_libs/index.pyx
@@ -9,14 +9,12 @@ from cpython.slice cimport PySlice_Check
 import numpy as np
 cimport numpy as cnp
 from numpy cimport (ndarray, float64_t, int32_t,
-                    int64_t, uint8_t, uint64_t, intp_t)
+                    int64_t, uint8_t, uint64_t, intp_t,
+                    # Note: NPY_DATETIME, NPY_TIMEDELTA are only available
+                    # for cimport in cython>=0.27.3
+                    NPY_DATETIME, NPY_TIMEDELTA)
 cnp.import_array()
 
-cdef extern from "numpy/arrayobject.h":
-    # These can be cimported directly from numpy in cython>=0.27.3
-    cdef enum NPY_TYPES:
-        NPY_DATETIME
-        NPY_TIMEDELTA
 
 cimport util
 
--- a/pandas/_libs/tslib.pyx
+++ b/pandas/_libs/tslib.pyx
@@ -1,16 +1,9 @@
 # -*- coding: utf-8 -*-
 # cython: profile=False
-
-cimport numpy as cnp
-from numpy cimport int64_t, ndarray, float64_t
-import numpy as np
-cnp.import_array()
-
+cimport cython
+from cython cimport Py_ssize_t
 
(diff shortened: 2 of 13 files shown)


pandas/_libs/index.pyx
```diff
@@ -9,14 +9,12 @@ from cpython.slice cimport PySlice_Check
 import numpy as np
 cimport numpy as cnp
 from numpy cimport (ndarray, float64_t, int32_t,
-                    int64_t, uint8_t, uint64_t, intp_t)
+                    int64_t, uint8_t, uint64_t, intp_t,
+                    # Note: NPY_DATETIME, NPY_TIMEDELTA are only available
```
jreback: prob can remove this comment (future PR ok)

pandas/_libs/tslib.pyx
```diff
@@ -54,7 +52,8 @@ from tslibs.timestamps cimport (create_timestamp_from_ts,
                                 _NS_UPPER_BOUND, _NS_LOWER_BOUND)
 from tslibs.timestamps import Timestamp
 
-cdef bint PY2 = str == bytes
+
+DEF PY2 = str == bytes
```
jreback: we should define in util.pxd maybe (future ok)

pandas/_libs/tslibs/offsets.pyx
```diff
@@ -334,8 +334,6 @@ class _BaseOffset(object):
             # other is not a DateOffset object
             return False
 
-        return self._params == other._params
```
jreback: this didn't break anything?

pandas/_libs/tslibs/period.pyx
```diff
@@ -46,14 +46,14 @@ from conversion cimport tz_convert_utc_to_tzlocal
...
-from parsing import parse_time_string, NAT_SENTINEL
+from parsing import parse_time_string
 from resolution import Resolution
 from nattype import nat_strings, NaT, iNaT
 from nattype cimport _nat_scalar_rules, NPY_NAT, is_null_datetimelike
 from offsets cimport to_offset
 from offsets import _Tick
 
-cdef bint PY2 = str == bytes
+DEF PY2 = str == bytes
```
jreback: same

jreback (approved):

jreback:
thanks @jbrockmendel a couple of things ought to do in the future PRs

pandas/_libs/tslibs/offsets.pyx
jbrockmendel: Was unreachable; there was a compile-time warning about it.

pandas/_libs/tslib.pyx
jbrockmendel: I doubt that works for DEF values, but not much downside.

PR #21883, opened 2018-07-13
TST: Parametrize tests in tests/util/test_hashing.py

jschendel:
Noticed that these tests could be parametrized when investigating a comment in the `IntervalArray` PR, and originally started doing to work there.  The `IntervalArray` PR is big enough already, and this is unrelated, so creating a separate PR.


gfyoung (approved):
Nice!

cc @jreback

jreback:
thanks @jschendel

PR #21886, opened 2018-07-13
use npy_datetimestruct instead of pandas_datetimestruct

jbrockmendel:
Follow-on to #21863


jbrockmendel:
@WillAyd thoughts?  Should be easy follow-up to PANDAS_DATETIMEUNIT change

jreback:
thanks @jbrockmendel

jreback:
when i rebuilt locally after merging this I got an error about finding pandas_datetimeunit, so I think we are missing a dep somewhere (on a util.pxd?)

works after a ``make clean``

PR #21888, opened 2018-07-13
Multi line comment

Keerthivasan1990:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


gfyoung:
@Keerthivasan1990 : Thanks for the PR! In general, we use the # if the comments are not docstrings, so I actually would prefer to leave that comment as is.

Issue #21881, opened 2018-07-12
Series with dtype=object does unexpected type conversion


AllenDowney:
#### Code Sample, a copy-pastable example if possible

```python
# Example 1

timestamp = pd.Timestamp(1412526600000000000)
series = pd.Series([], dtype=object)
series['timestamp'] = timestamp
type(series.timestamp)

# Example 2

series = pd.Series([], dtype=object)
series['anything'] = 300.0
series['timestamp'] = timestamp
type(series.timestamp)

```
#### Problem description

In the first example, the timestamp is still a Timestamp.

In the second example, the timestamp gets converted to int.

#### Expected Output

I expected the timestamp to continue to be a Timestamp, especially because the dtype of the Series is object.  Why are the types of the values getting converted?

#### Output of ``pd.show_versions()``

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.0-128-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.utf8
LOCALE: en_US.UTF-8

pandas: 0.23.0
pytest: 3.5.1
pip: 10.0.1
setuptools: 39.1.0
Cython: 0.28.2
numpy: 1.14.3
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: 1.7.4
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.2.1
tables: 3.4.3
numexpr: 2.6.5
feather: None
matplotlib: 2.2.2
openpyxl: 2.5.3
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.4
lxml: 4.2.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.7
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

WillAyd:
This is probably the same issue as #13910

AllenDowney:
Thanks for the quick reply!  Yes, looks like it is the same issue.

Any suggestions on a workaround?

I thought with `dtype=object`, Series might stop trying to infer types.  Is
there any way to just turn off type inference?

makbigc:
[...]

PR #21893, opened 2018-07-13
BUG: Fixes unwanted casting in .isin (GH21804)

### LINKED ISSUE #21804: BUG: unwanted casting in .isin
xref #19508  for a partial PR

```
In [3]: values = [1, 0.5]
   ...: pd.Series([1, 0]).isin(values)
   ...: 
Out[3]: 
0    True
1    True
dtype: bool
```
should be [True, False]

KalyanGokhale:
- [x] closes #21804 
- [x] tests added / passed
- [x] whatsnew entry

Results from running asvs on algorithms are:
```
     [365eac4d]       [ee66578f]
+        1.91±0ms      2.28±0.02ms     1.20  algorithms.Hashing.time_series_timedeltas
-     2.34±0.02ms         1.88±0ms     0.80  algorithms.Hashing.time_series_int

```


KalyanGokhale:
The following tests are failing on my local machine after rebasing since yesterday - tried investigating it a little, but seems unrelated to this fix - will wait for Travis-CI to complete and then re-investigate if required.
`pandas/tests/indexes/datetimes/test_tools.py::TestToDatetimeUnit::test_unit_rounding`
`pandas/tests/scalar/timedelta/test_timedelta.py::TestTimedeltas::test_rounding_on_int_unit_construction`
`pandas/tests/scalar/timestamp/test_timestamp.py::TestTimestampProperties::{test_resolution, test_unit}`

Also, the following fails as part of the test suite, but passes when only this module is executed individually - not sure why??
`pandas/tests/scalar/timestamp/test_timestamp.py::TestClipboard::test_round_trip_frame`

KalyanGokhale:
Revised benchmarks after 84be606
```
     [365eac4d]       [84be606b]
+        1.82±0ms         2.23±0ms     1.23  algorithms.Hashing.time_series_int
-        2.24±0ms         1.83±0ms     0.82  algorithms.Hashing.time_series_float

```

pandas/tests/test_algos.py
```diff
@@ -509,42 +509,23 @@ def test_invalid(self):
...
-        result = algos.isin(Series(['a', 'b']), set(['a']))
-        expected = np.array([True, False])
-        tm.assert_numpy_array_equal(result, expected)
-
-        result = algos.isin(['a', 'b'], [1])
-        expected = np.array([False, False])
+    @pytest.mark.parametrize("comps,values,expected", [
+        ([1, 2], [1], [True, False]),
+        ([1, 0], [1, 0.5], [True, False]),
+        ([1.0, 0], [1, 0.5], [True, False]),
```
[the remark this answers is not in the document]
KalyanGokhale: Thanks @gfyoung

pandas/core/algorithms.py
```diff
@@ -413,35 +415,41 @@ def isin(comps, values):
...
+            values_types = values_types.pop()
+            if (is_int(comps_types) and is_int(values_types)):
+                values = values.astype('int64', copy=False)
+                comps = comps.astype('int64', copy=False)
+                f = lambda x, y: htable.ismember_int64(x, y)
+            elif (is_float(comps_types) and is_float(values_types)):
+                values = values.astype('float64', copy=False)
+                comps = comps.astype('float64', copy=False)
+                checknull = isna(values).any()
+                f = lambda x, y: htable.ismember_float64(x, y, checknull)
```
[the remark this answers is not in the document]
KalyanGokhale: In the earlier code only the dtype of comps was being checked for being either *int* or *float*, and then the values where force casted, which would have necessitated a `try-except`
Now, both comps and values are being explicitly checked to be either *int* or *float* before their conversion to `int64` or `float64`

pandas/core/algorithms.py
```diff
@@ -413,35 +415,41 @@ def isin(comps, values):
...
 
-    comps, dtype, _ = _ensure_data(comps)
-    values, _, _ = _ensure_data(values, dtype=dtype)
+    is_time_like = lambda x: (is_datetime_or_timedelta_dtype(x)
+                              or isinstance(x, Timestamp))
 
-    # faster for larger cases to use np.in1d
-    f = lambda x, y: htable.ismember_object(x, values)
+    is_int = lambda x: ((x == np.int64) or (x == int))
+
```
jreback: woa, what are you doing? this is way less understandable that before. too many more if/thens here. pls fit this into the existing structure.

jreback (changes requested):

pandas/core/algorithms.py
KalyanGokhale: Have edited to retain much of the existing structure

KalyanGokhale: Revised benchmarks after code restructure f8cc271 
```
     [365eac4d]       [f8cc271b]
-     2.31±0.01ms      1.91±0.03ms     0.82  algorithms.Hashing.time_series_int

```

KalyanGokhale: Revised asv benchmarks on algorithms after 5416711
```
     [365eac4d]       [54167114]
-     2.25±0.01ms      1.87±0.02ms     0.83  algorithms.Hashing.time_series_timedeltas

```

KalyanGokhale: @jreback thanks for the review - any other edits needed? 
Have tried to retain the existing structure and reduced if-blocks, though have removed the redundant try-except blocks. Have also tried to cluster the blocks logically.

doc/source/whatsnew/v0.23.4.txt
```diff
@@ -31,8 +31,7 @@ Bug Fixes
 
 **Conversion**
 
--
--
+- Unwanted casting of float to int in :func:`isin` (:issue:`21804`)
```
jreback: can you be more specific here, as a user I have no idea what this means.

pandas/core/algorithms.py
```diff
@@ -415,33 +417,47 @@ def isin(comps, values):
...
+        values, _, _ = _ensure_data(values, dtype=dtype)
+    else:
+        values, _, _ = _ensure_data(values)
 
     # faster for larger cases to use np.in1d
-    f = lambda x, y: htable.ismember_object(x, values)
+    f = lambda x, y: htable.ismember_object(x.astype(object), y.astype(object))
+
+    comps_types = set(type(v) for v in comps)
+    values_types = set(type(v) for v in values)
```
jreback: you are going to great lengths to circument the structure below. This needs to be an extension of the if/then. I don't want to see ``int_flg`` or anything like that. This is also not performant. The reason we check dtypes in this way is to avoid conversions and materialization.

jreback (changes requested):

pandas/core/algorithms.py
KalyanGokhale: Thanks - done
ASV results on algorithms after 82386c3 are similar to the most recent one 5416711 
```
     [537b65cb]       [82386c31]
-     2.36±0.02ms      1.92±0.02ms     0.82  algorithms.Hashing.time_series_timedeltas

```

KalyanGokhale:
Any other edits needed? Thanks

pandas/core/algorithms.py
```diff
@@ -415,33 +417,40 @@ def isin(comps, values):
     comps = com._values_from_object(comps)
 
     comps, dtype, _ = _ensure_data(comps)
-    values, _, _ = _ensure_data(values, dtype=dtype)
+
+    is_time_like = lambda x: (is_datetime_or_timedelta_dtype(x)
```
jreback: i already indicate that this is not the path forward

this is much slower than the existing

you need to keep it along the current structure

and post benchmarks

jreback (changes requested):

pandas/core/algorithms.py
KalyanGokhale: @jreback Thanks - have removed the flags `int_flg` / `float_flg` as suggested per your earlier review and posted the benchmarks after each commit. 
Re-pasting the benchmarks after the last commit
>ASV results on algorithms after 82386c3 are similar to the most recent one 5416711
```
     [537b65cb]       [82386c31]
-     2.36±0.02ms      1.92±0.02ms     0.82  algorithms.Hashing.time_series_timedeltas

```
Also, could not get around writing a custom function `is_time_like` - had tried it in 
9fca52c 

Please feel free to close this PR - since I think I am not sure I understand what is the expectation in terms of maintaining the existing structure

jreback: there are many benchmarks

the point is that this code is highly sensistive to changes and requires a lot of benchmark running to avoid regressions

KalyanGokhale:
Should I run the complete ASV suite? Earlier I only ran it for algorithms...
If the results are not promising, we can just close this PR. If they are we can re-think the approach - any suggestions are welcome as I think I am stuck in my thought pattern. Thanks

jreback:
run a battery of tests anything relating to value_counts, unique, factorize

KalyanGokhale:
@jreback you are correct - indeed many benchmarks have worsened (I actually didn't know where to start, so ran the full asv suite). Let me rebase and run the asv again and then will rethink the approach.

Current results below (pasting only the ones where it worsened):
```ruby
     [537b65cb]       [82386c31]
+        1.46±0ms      15.9±0.08ms    10.86  series_methods.IsIn.time_isin('int64')
+          9.20ms           78.3ms     8.51  categoricals.Isin.time_isin_categorical('int64')
+      9.56±0.4ms       79.7±0.4ms     8.33  categoricals.Isin.time_isin_categorical('object')
+     2.47±0.02ms       15.0±0.1ms     6.05  series_methods.IsIn.time_isin('object')
+     8.21±0.08ms       20.5±0.2ms     2.50  indexing.MultiIndexing.time_index_slice
+     2.89±0.01ms       4.49±0.1ms     1.56  binary_ops.Ops.time_frame_comparison(True, 1)
+        77.3±2ms          120±3ms     1.56  binary_ops.Ops.time_frame_comparison(False, 1)
+      6.74±0.4ms       9.97±0.4ms     1.48  groupby.Categories.time_groupby_nosort
+      78.4±0.7ms         115±10ms     1.47  binary_ops.Ops.time_frame_comparison(False, 'default')
+     2.93±0.02ms      4.29±0.01ms     1.46  binary_ops.Ops.time_frame_add(False, 'default')
+     2.94±0.05ms       4.06±0.2ms     1.38  binary_ops.Ops.time_frame_add(False, 1)
+     3.98±0.07ms       5.07±0.2ms     1.28  gil.ParallelRolling.time_rolling('mean')
+          86.0μs            108μs     1.25  frame_methods.XS.time_frame_xs(0)
+     2.32±0.05ms      2.87±0.01ms     1.24  rolling.VariableWindowMethods.time_rolling('DataFrame', '1h', 'float', 'count')
+      69.6±0.3μs       85.7±0.2μs     1.23  groupby.GroupByMethods.time_dtype_as_group('int', 'any', 'transformation')
+     4.40±0.04ms      5.32±0.08ms     1.21  rolling.Methods.time_rolling('Series', 1000, 'float', 'std')
+     2.49±0.01ms      2.98±0.04ms     1.20  ctors.SeriesConstructors.time_series_constructor(<function SeriesConstructors.<lambda> at 0x7f06c40db1e0>, True)
+          12.9ms           15.1ms     1.17  categoricals.CategoricalSlicing.time_getitem_bool_array('non_monotonic')
+     1.93±0.07ms       2.27±0.1ms     1.17  frame_methods.NSort.time_nsmallest('last')
+     2.07±0.09ms         2.35±0ms     1.13  algorithms.Hashing.time_series_dates
+      7.29±0.2μs           8.26μs     1.13  categoricals.CategoricalSlicing.time_getitem_list_like('non_monotonic')
+     1.94±0.04ms       2.19±0.1ms     1.13  frame_methods.NSort.time_nlargest('all')
+      57.2±0.5ms       64.5±0.4ms     1.13  groupby.Groups.time_series_groups('object_small')
+           313ms            347ms     1.11  gil.ParallelReadCSV.time_read_csv('float')
+     35.7±0.06μs         39.5±1μs     1.11  ctors.SeriesConstructors.time_series_constructor(<function SeriesConstructors.<lambda> at 0x7f06c40db2f0>, True)
```

KalyanGokhale:
@jreback @gfyoung 

Did some further tinkering today, and with the original code for `.isin` the results are as follows:
```python
>>> import pandas as pd
>>> import numpy as np
>>> import pandas.core.algorithms as algos
>>> comps=[1,0]
>>> values=[1,0.5]
>>> algos.isin(comps, values)
array([ True, False])
```
Above is as expected, whereas the result below is not...
```python
>>> pd.Series(comps).isin(values)
0    True
1    True
dtype: bool
```
Now thinking about it, its not an isolated casting issue in `.isin` (which actually seems to be working fine) - but rather a fundamental issue of casting with `Series` in general or something else upstream? 
e.g.
```python
>>> s = pd.Series([])
>>> s[0]=100
>>> s
0    100
dtype: int64
>>> s[0]=100.589
>>> s
0    100
dtype: int64
```
(was also thinking of the issue #21881 and related ones) - not sure of whether my thinking is on the correct lines...

The current issue can be addressed (probably even improving the perf) - but now seems rather like a band-aid, unless we address the casting issue(s) for Series....(have not yet checked if this was already discussed and the consensus is to not do it for some valid reason...)
Thoughts?

p.s:
This means that the current test cases I had written certainly need to be updated specifically to include `Series` - though had tested it specifically for `Series` on command line with the original test case :)

jreback:
can you rebase and update to comments

KalyanGokhale:
> can you rebase and update to comments

@jreback yes - will do over coming few days

jreback:

PR #21896, opened 2018-07-13
DEPR: Deprecate Series.to_csv signature

### LINKED ISSUE #19715: Conform Series.to_csv to DataFrame.to_csv
Thank you for this awesome package!

#### Problem description
When indexing/selecting/slicing a `DataFrame`, pandas may return a `Series`. Both classes support the `to_csv` method, but the methods have subtle differences. These subtle differences may break your code, or demand a call to `to_frame`. Perhaps it would be beneficial to conform the `Series.to_csv` method to the `DataFrame.to_csv` method?

At least, the following two examples seem relevant (and are the reason I came here):

- `DataFrame.to_csv` accepts the keyword parameter `path_or_buf`. This seems to be completely analogous to the differently named keyword parameter `path` of `Series.to_csv`.
- `DataFrame.to_csv` accepts the keyword parameter `line_terminator`. There does not seem to be an analogous keyword parameter of `Series.to_csv`.

* * *

Addendum: Other differences

The following two keyword parameters are implemented for `DataFrame.to_csv`, but not for `Series.to_csv`: `compression`, `chunksize`. I suppose these are mostly relevant if you have very large dataframes, but I do not see any reason why they should not have analogous methods for Series.

Some of the keyword parameters of `DataFrame.to_csv` are specific to multi-column data and may therefore not make immediate sense for `Series.to_csv`. On the other hand, a user may wish to use quoting in a single-column file in order to maintain conformity with other parts of a bigger project. In this relation, the following keyword parameters are implemented for `DataFrame.to_csv` but do not for `Series.to_csv`:  `quoting`, `quotechar`,`doublequote`, `escapechar`.

<details>

INSTALLED VERSIONS
------------------

commit: None
python: 3.6.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.13.0-32-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL:
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.22.0
pytest: None
pip: 9.0.1
setuptools: 38.5.1
Cython: None
numpy: 1.14.0
scipy: None
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: None
patsy: None
dateutil: 2.6.1
pytz: 2017.3
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: 2.5.0
xlrd: 1.1.0
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: None
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
None
</details>

toobaz:
- [x] closes #19715
- [x] tests added / passed
- [x] whatsnew entry

<strike>Just a proof of concept for discussion (misses docs, whatsnew, new tests). Based on #21868</strike>

@dahlbaek @gfyoung


dahlbaek:
This is the solution that feels the most "just get it done" to me, while not breaking any "correctly written" code (as far as I can tell...?).

pandas/core/generic.py
```diff
@@ -9161,6 +9161,103 @@ def first_valid_index(self):
...
+            the new, expanded format, where each MultiIndex column is a row
+            in the CSV (if False).
+        date_format : string, default None
+            Format string for datetime objects
+        decimal: string, default '.'
+            Character recognized as decimal separator. E.g. use ',' for
+            European data
+
+        """
+
```
dahlbaek: I think a few lines from the pull request of @gfyoung are missing here, something like
```python

    df = self if isinstance(self, DataFrame) else DataFrame(self)
```

If I am not mistaken, the purpose of those lines is to prepare for this method to be called by `Series`.

pandas/core/generic.py
```diff
@@ -9161,6 +9161,103 @@ def first_valid_index(self):
...
+        if tupleize_cols is not None:
+            warnings.warn("The 'tupleize_cols' parameter is deprecated and "
+                          "will be removed in a future version",
+                          FutureWarning, stacklevel=2)
+        else:
+            tupleize_cols = False
+
+        from pandas.io.formats.csvs import CSVFormatter
+        formatter = CSVFormatter(self, path_or_buf,
```
dahlbaek: If you agree with to comment above, `self` should be replaced by `df` here.

toobaz:
> while not breaking any "correctly written"

Do you have in mind "uncorrectly written" code it could break?

dahlbaek:
Just the example of passing `"y"` instead of `True` as a positional argument to `index`.

toobaz:
> Just the example of passing "y" instead of True as a positional argument to index.

Right, the previous version was wrong and incomplete. Should be fixed now, thanks.

``` python
In [2]: s = pd.Series(list('abcd'))

In [3]: s.to_csv('/tmp/test1.csv', 'y')
/home/nobackup/repo/pandas/pandas/core/series.py:3812: UserWarning: Argument 'header' has changed default value to True: please pass an explicit value to suppress this warning
  warnings.warn("Argument 'header' has changed default value to "

In [4]: cat /tmp/test1.csv
y0
0ya
1yb
2yc
3yd

In [5]: s.to_csv('/tmp/test1.csv', True)
/usr/bin/ipython3:1: FutureWarning: The signature of `Series.to_csv` was  aligned to that of `DataFrame.to_csv`. Note that the order of arguments changed, and the new one has 'sep' in first place, for which "True" is not a valid value. The old order will cease to be supported in a future version. Please refer to the documentation for `DataFrame.to_csv` when updating your function calls.
  #! /bin/sh
/home/nobackup/repo/pandas/pandas/core/series.py:3812: UserWarning: Argument 'header' has changed default value to True: please pass an explicit value to suppress this warning
  warnings.warn("Argument 'header' has changed default value to "

In [6]: cat /tmp/test1.csv
,0
0,a
1,b
2,c
3,d
```

pandas/core/generic.py
toobaz: (added)

jreback:
is this replacing #21868 .  - haven't been following 2 closely @toobaz @gfyoung

gfyoung:
@jreback : No.  Actually would like your or @jorisvandenbossche input, as the conversations have produced three possible fixes:

#19745 (@dahlbaek)
#21868 (@gfyoung)
#21896 (@toobaz)

jreback:
#21896 and #21868 (this one) look very similar at first glance, what is the substantive diff?

I prefer either of these over #19745 which I find much harder to read.

gfyoung:
@jreback : @toobaz does major argument introspection by changing the signature to `Series(*args, **kwargs)`.  I (#21868) kept things a little simpler by adding only `**kwargs` to the current signature and doing argument introspection on `**kwargs`.

toobaz:
> @toobaz does major argument introspection

Yes; one additional important change is that this PR tries to support both the old and the new positional signature (where "tries" means that it will never know for sure if some user had the idea to pass ``index="y"`` in place of ``index=True``).

gfyoung:
@jreback @jorisvandenbossche : Any further thoughts on #21896 and #21868?  I think @toobaz, @dahlbaek , and I have exhausted our own thoughts 😄 , so fresh eyes would be good.

gfyoung:
@jreback @jorisvandenbossche : Another friendly ping on this.

jreback:
I like this version. I think bending over to give a helpful error message is pretty useful to the user. This is not completely crazy and doesn't add a huge amount of code. I think a couple of TODO's were left in the code.

gfyoung:
@toobaz : Don't forget `whatsnew` entry as well.

toobaz:
> @toobaz : Don't forget whatsnew entry as well.

Tests are missing too, but I tought that maybe the simplest thing would be for you to take my code in your PR. Otherwise I can do the opposite (not sure when I'll find the time though).

gfyoung:
@toobaz : I can push to your branch if that works.

toobaz:
> @toobaz : I can push to your branch if that works.

Sure, feel free (I saw now only that you closed the other one)

jorisvandenbossche (commented):
Only looking at this PR (I didn't follow all the other discussions), this looks good to me

pandas/tests/series/test_io.py
```diff
@@ -37,18 +37,36 @@ def read_csv(self, path, **kwargs):
...
                 assert_series_equal(depr_ts, ts)
 
+    @pytest.mark.parametrize("arg", ["path", "header", "both"])
+    def test_to_csv_deprecation(self, arg):
+        # see gh-19715
+        with ensure_clean() as path:
+            if arg == "path":
+                kwargs = dict(path=path, header=False)
+            elif arg == "header":
+                kwargs = dict(path_or_buf=path)
```
jorisvandenbossche: Why would this case raise the warning? (as `path_or_buf` is the future keyword?)

pandas/tests/test_common.py
```diff
@@ -122,11 +131,12 @@ def test_standardize_mapping():
...
-        compressed = os.path.getsize(filename)
-        getattr(obj, method)(filename, compression=None)
-        uncompressed = os.path.getsize(filename)
-        assert uncompressed > compressed
+        with catch_to_csv_depr():
+            getattr(obj, method)(filename, compression=compression_only)
+            compressed = os.path.getsize(filename)
+            getattr(obj, method)(filename, compression=None)
+            uncompressed = os.path.getsize(filename)
+            assert uncompressed > compressed
```
jorisvandenbossche: why would those cases raise any warnings? 
(and same for the others below)

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -383,6 +383,7 @@ Deprecations
 - :meth:`MultiIndex.to_hierarchical` is deprecated and will be removed in a future version  (:issue:`21613`)
 - :meth:`Series.ptp` is deprecated. Use ``numpy.ptp`` instead (:issue:`21614`)
 - :meth:`Series.compress` is deprecated. Use ``Series[condition]`` instead (:issue:`18262`)
+- The signature in :meth:`Series.to_csv` has been deprecated. Please follow the signature in :meth:`DataFrame.to_csv` instead (:issue:`19715`)
```
jorisvandenbossche: I would give a bit more details here (or at least mention the most important ones: change in name of first argument + second positional argument is different, amongst other order changes)

jorisvandenbossche (commented):
Did you test all cases of the deprecation warnings? (maybe I missed some, but I don't see any tests for the positional args case)

pandas/core/series.py
```diff
@@ -3762,56 +3763,67 @@ def from_csv(cls, path, sep=',', parse_dates=True, header=None,
...
+                          "'header' will change its default value to "
+                          "True: please pass an explicit value to suppress "
+                          "this warning.", warning_klass,
+                          stacklevel=stack_level)
+            kwargs["header"] = False  # Backwards compatibility.
+        return self.to_frame().to_csv(**kwargs)
+
+    # This entire method is going to be removed, so the shared docstring
+    # mechanism is overkill
+    to_csv.__doc__ = generic.NDFrame.to_csv.__doc__
```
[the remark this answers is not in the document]
toobaz: > This is not safe for builds without docstrings I think?

Didn't know about those (link?) In general if ``func`` has no docstring, ``func.__doc__`` is just ``None``... but again, I might be missing something.

> Using that will simply be a one-line as well.

Wouldn't that mean changing ``pandas/core/generic.py`` to insert the docstring inside ``_shared_docs`` (and undo after deprecation)?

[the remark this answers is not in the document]
jorisvandenbossche: Yeah, forget about the comment about builds without docstring, I always forget the exact semantics: they don't remove the attribute, just set it to None. So it is things like `to_csv.__doc__ = generic.NDFrame.to_csv.__doc__ + "appending"` that then does not work.

Anyhow, you can do a 

```
@Appender(generic.NDFrame.to_csv.__doc__)
```

which will do the same but with our machinery (and then you don't need the comment, 2 lines saved :-))

pandas/core/series.py
```diff
@@ -3762,56 +3763,67 @@ def from_csv(cls, path, sep=',', parse_dates=True, header=None,
...
-        # result is only a string if no path provided, otherwise None
-        result = df.to_csv(path, index=index, sep=sep, na_rep=na_rep,
-                           float_format=float_format, header=header,
-                           index_label=index_label, mode=mode,
-                           encoding=encoding, compression=compression,
-                           date_format=date_format, decimal=decimal)
-        if path is None:
-            return result
+    def to_csv(self, *args, **kwargs):
+        warning_klass = FutureWarning
```
toobaz: @gfyoung I don't understand the need for this variable (and the following)

pandas/core/series.py
```diff
@@ -3762,56 +3763,64 @@ def from_csv(cls, path, sep=',', parse_dates=True, header=None,
...
-                           float_format=float_format, header=header,
-                           index_label=index_label, mode=mode,
-                           encoding=encoding, compression=compression,
-                           date_format=date_format, decimal=decimal)
-        if path is None:
-            return result
+    @Appender(generic.NDFrame.to_csv.__doc__)
+    def to_csv(self, *args, **kwargs):
+        warning_klass = FutureWarning
+        stack_level = 2
```
toobaz: I don't understand why we need those variables - they don't save code (actually the opposite) nor increase readability.

jorisvandenbossche: agreed, I think you can just change this

gfyoung: We use those variables in multiple places in this method to issue warnings about keyword arguments and the general deprecations.  If you want to change this to copy and paste it everywhere, go ahead.

pandas/tests/series/test_io.py
gfyoung: Because `header` will default to `True` in the future (instead of `False` as it was the case in `Series.to_csv`).  Thus, users who don't specify the `header` argument will be "surprised" when that change in the default results in different output.  Thus, we are warning when `header` is not passed in explicitly to let them know that the default is changing.

pandas/tests/test_common.py
gfyoung: Same answer as [above](https://github.com/pandas-dev/pandas/pull/21896#discussion_r205529792).

pandas/core/series.py
toobaz: > If you want to change this to copy and paste it everywhere, go ahead.

It's actually how it was, but OK, I will :-)

toobaz:
Rebased (on #22011 ) and added missing ``.. versionchanged:: 0.24.0`` for ``header`` argument. 

Lint complains with ``pandas/core/strings.py:296:80: E501 line too long (87 > 79 characters)`` but that's not a file I edited...

gfyoung (approved):
LGTM!

cc @jreback @jorisvandenbossche

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #21902, opened 2018-07-13
TST/CLN: correctly skip in indexes/common; add test for duplicated

h-vetinari:
Splitting up #21645 

* Added tests for `duplicated`
* Following https://github.com/pandas-dev/pandas/pull/21645#discussion_r202192191, turned several blank `return` statements (which falsely pass the test)  into `pytest.skip`.

### DIFF
--- a/pandas/tests/indexes/common.py
+++ b/pandas/tests/indexes/common.py
@@ -35,10 +35,6 @@ class Base(object):
         assert indices.equals(unpickled)
 
     def test_pickle_compat_construction(self):
-        # this is testing for pickle compat
-        if self._holder is None:
-            pytest.skip('Skip check for uncertain type')
-
         # need an object to create with
         pytest.raises(TypeError, self._holder)
 
--- a/pandas/tests/indexes/common.py
+++ b/pandas/tests/indexes/common.py
@@ -37,7 +37,7 @@ class Base(object):
     def test_pickle_compat_construction(self):
         # this is testing for pickle compat
         if self._holder is None:
-            return
+            pytest.skip('Skip check for uncertain type')
 
         # need an object to create with
         pytest.raises(TypeError, self._holder)
@@ -236,7 +236,7 @@ class Base(object):
 
         # don't tests a MultiIndex here (as its tested separated)
         if isinstance(indices, MultiIndex):
-            return
+            pytest.skip('Skip check for MultiIndex')
         original_name = indices.name
         new_ind = indices.set_names([new_name])
         assert new_ind.name == new_name
@@ -333,7 +333,8 @@ class Base(object):
         from copy import copy, deepcopy
 
         if isinstance(indices, MultiIndex):
-            return
+            pytest.skip('Skip check for MultiIndex')
+
         for func in (copy, deepcopy):
             idx_copy = func(indices)
             assert idx_copy is not indices
@@ -344,18 +345,46 @@ class Base(object):
 
     def test_duplicates(self, indices):
         if type(indices) is not self._holder:
-            return
+            pytest.skip('Can only check if we have the correct type')
         if not len(indices) or isinstance(indices, MultiIndex):
-            return
+            # MultiIndex tested separately in:
+            # tests/indexes/multi/test_unique_and_duplicates
+            pytest.skip('Skip check for empty Index and MultiIndex')
+
         idx = self._holder([indices[0]] * 5)
         assert not idx.is_unique
         assert idx.has_duplicates
+
+    @pytest.mark.parametrize('keep', ['first', 'last', False])
+    def test_duplicated(self, indices, keep):
+        if type(indices) is not self._holder:
+            pytest.skip('Can only check if we know the index type')
+        if not len(indices) or isinstance(indices, MultiIndex):
+            # MultiIndex tested separately in:
+            # tests/indexes/multi/test_unique_and_duplicates
+            pytest.skip('Skip check for empty Index and MultiIndex')
+
+        idx = self._holder(indices)
+        if idx.has_duplicates:
+            # We need to be able to control creation of duplicates here
+            # This is slightly circular, as drop_duplicates depends on
+            # duplicated, but in the end, it all works out because we
+            # cross-check with Series.duplicated
+            idx = idx.drop_duplicates()
+
+        n, k = len(idx), 10
+        duplicated_selection = np.random.choice(n, k * n)
+        expected = pd.Series(duplicated_selection).duplicated(keep=keep).values
+        idx = self._holder(idx.values[duplicated_selection])
+
+        result = idx.duplicated(keep=keep)
+        tm.assert_numpy_array_equal(result, expected)
 
     def test_unique(self, indices):
         # don't test a MultiIndex here (as its tested separated)
         # don't test a CategoricalIndex because categories change (GH 18291)
         if isinstance(indices, (MultiIndex, CategoricalIndex)):
-            return
+            pytest.skip('Skip check for MultiIndex/CategoricalIndex')
 
         # GH 17896
         expected = indices.drop_duplicates()
@@ -375,7 +404,7 @@ class Base(object):
     def test_get_unique_index(self, indices):
         # MultiIndex tested separately
         if not len(indices) or isinstance(indices, MultiIndex):
-            return
+            pytest.skip('Skip check for empty Index and MultiIndex')
 
         idx = indices[[0] * 5]
         idx_unique = indices[[0]]
@@ -394,7 +423,7 @@ class Base(object):
 
         # nans:
         if not indices._can_hold_na:
-            return
+            pytest.skip('Skip na-check if index cannot hold na')
 
         if needs_i8_conversion(indices):
             vals = indices.asi8[[0] * 5]
@@ -423,7 +452,7 @@ class Base(object):
 
     def test_mutability(self, indices):
         if not len(indices):
-            return
+            pytest.skip('Skip check for empty Index')
         pytest.raises(TypeError, indices.__setitem__, 0, indices[0])
 
     def test_view(self, indices):
@@ -761,7 +790,7 @@ class Base(object):
         # GH9947, GH10637
         index_a = self.create_index()
         if isinstance(index_a, PeriodIndex):
-            return
+            pytest.skip('Skip check for PeriodIndex')
 
         n = len(index_a)
         index_b = index_a[0:-1]
@@ -989,11 +1018,11 @@ class Base(object):
         # not implemented for tuple searches in MultiIndex
         # or Intervals searches in IntervalIndex
         if isinstance(indices, (MultiIndex, IntervalIndex)):
-            return
+            pytest.skip('Skip check for MultiIndex/IntervalIndex')
 
         # nothing to test if the index is empty
         if indices.empty:
-            return
+            pytest.skip('Skip check for empty Index')
         value = indices[0]
 
         # determine the expected results (handle dupes for 'right')
--- a/pandas/tests/indexes/common.py
+++ b/pandas/tests/indexes/common.py
@@ -366,10 +366,12 @@ class Base(object):
 
         idx = self._holder(indices)
         if idx.has_duplicates:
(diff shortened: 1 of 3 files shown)


pandas/tests/indexes/common.py
```diff
@@ -344,18 +345,46 @@ def test_copy_and_deepcopy(self, indices):
...
-            return
+            # MultiIndex tested separately in:
+            # tests/indexes/multi/test_unique_and_duplicates
+            pytest.skip('Skip check for empty Index and MultiIndex')
+
         idx = self._holder([indices[0]] * 5)
         assert not idx.is_unique
         assert idx.has_duplicates
 
+    @pytest.mark.parametrize('keep', ['first', 'last', False])
```
jreback: is this a duplicate these (pun intended) or are not testing indices duplicated currently?

jreback (changes requested):

pandas/tests/indexes/common.py
h-vetinari: @jreback `.duplicated` is hardly ever tested directly, only indirectly for stuff like `.drop_duplicates`. Regardless of the changes to `.duplicated` in #21645, I think duplicated should be tested separately.

jreback: so these are not relevant?
```
pandas/tests/indexes/common.py:    def test_duplicates(self, indices):
pandas/tests/indexes/test_category.py:    def test_duplicates(self):
pandas/tests/indexes/test_range.py:    def test_duplicates(self):
```

h-vetinari: @jreback No, `test_duplicates` (at least for `pandas/tests/indexes/common.py`, which is what this PR is about) tests `.is_unique` and `.has_duplicates`, but not the `.duplicated`-method itself.

jreback: can you point to the coverage that shows this is NOT tested?

h-vetinari: @jreback I never said that it is not tested, just that it is only tested *implicitly*. Any call to `.drop_duplicates` will invoke `duplicated`, and so obviously the coverage works out.

pandas/tests/indexes/common.py
```diff
@@ -342,20 +343,48 @@ def test_copy_and_deepcopy(self, indices):
...
+    def test_duplicated(self, indices, keep):
+        if type(indices) is not self._holder:
```
jreback: can you use isinstance here

pandas/tests/indexes/common.py
```diff
@@ -342,20 +343,48 @@ def test_copy_and_deepcopy(self, indices):
...
+            pytest.skip('Can only check if we know the index type')
+        if not len(indices) or isinstance(indices, MultiIndex):
+            # MultiIndex tested separately in:
+            # tests/indexes/multi/test_unique_and_duplicates
+            pytest.skip('Skip check for empty Index and MultiIndex')
+
+        idx = self._holder(indices)
+        if idx.has_duplicates:
+            # We need to be able to control creation of duplicates here
```
jreback: this comment is a bit obtuse, can you reword

jreback (changes requested):

pandas/tests/indexes/common.py
h-vetinari: @jreback `isinstance` of what? I copied the checks from other tests, because I didn't know all the different cases that flow into `common.py`.

jreback: isinstande of self._holder, is this what we do elsewhere?

h-vetinari: not in `indexes/common.py`, as far as I can see. There's several `isinstance` of course, but never for `self._holder`, which can apparently be `None` as well.

h-vetinari: @jreback If I change
```
if type(indices) is not self._holder:
```
to
```
if not isinstance(indices, self._holder):
```
I get 9 failures (instead of skips), all of which are from

`tests/indexes/test_numeric.TestInt64Index`, when run against `DatetimeIndex`, `PeriodIndex` or `TimedeltaIndex`, mainly because of `TypeError: Unsafe NumPy casting, you must explicitly cast` it seems.

h-vetinari: @jreback Seem to have found something: in my normal workbook, I get
```
pd.__version__
# '0.24.0.dev0+321.g0fe6ded52.dirty'
isinstance(pd.PeriodIndex, pd.Int64Index)
# False
```
but for some reason, within `tests/indexes/common.py`, this is the opposite:
```
    @pytest.mark.parametrize('keep', ['first', 'last', False])
    def test_duplicated(self, indices, keep):
        if not isinstance(indices, self._holder):
            pytest.skip('Can only check if we know the index type')
        if not len(indices) or isinstance(indices, (MultiIndex)):
            # MultiIndex tested separately in:
            # tests/indexes/multi/test_unique_and_duplicates
            pytest.skip('Skip check for empty Index and MultiIndex')
        if isinstance(indices, (PeriodIndex, DatetimeIndex)):
>           raise ValueError(f'{type(indices).__name__}, {self._holder.__name__}, {isinstance(indices, self._holder)}')
E           ValueError: PeriodIndex, Int64Index, True
```
Same happens for `DatetimeIndex`.

h-vetinari:
@jreback All green. Any more feedback / comments?

pandas/tests/indexes/common.py
```diff
@@ -37,7 +37,7 @@ def verify_pickle(self, indices):
     def test_pickle_compat_construction(self):
         # this is testing for pickle compat
         if self._holder is None:
-            return
```
jreback: is this actually hit?

pandas/tests/indexes/common.py
```diff
@@ -342,20 +343,50 @@ def test_copy_and_deepcopy(self, indices):
...
+            # exactly which indices are duplicate and how (for the result).
+            # This is not possible if "idx" has duplicates already, which we
+            # therefore remove. This is seemingly circular, as drop_duplicates
+            # invokes duplicated, but in the end, it all works out because we
+            # cross-check with Series.duplicated, which is tested separately.
+            idx = idx.drop_duplicates()
+
+        n, k = len(idx), 10
+        duplicated_selection = np.random.choice(n, k * n)
+        expected = pd.Series(duplicated_selection).duplicated(keep=keep).values
```
jreback: I hate to check as a numpy array, much prefer to check the type and use assert_index_equal or assert_series_equal. is this how the other tests are?

jreback (changes requested):

pandas/tests/indexes/common.py
h-vetinari: I don't know, I didn't write these tests, and the inner workings of `indexes/common.py` are not immediately apparent (which files call it, what do they fill `indices` with, etc.).

pandas/tests/indexes/common.py
h-vetinari: Um, `index.duplicated()` (without `return_inverse`) yields a numpy array - this is the documented signature (I guess because selecting on an `Index` really only needs an `ndarray`).

All the manual `duplicated`-tests actually create their own data, and _know_ what the correct outcome should be. Here, we're feeding tons of different things through that test, so we need to determine - as I'm doing with `duplicated_selection` what is actually duplicate; `idx.values[duplicated_selection]` is then a duplicate Index of the correct type, but we know where its duplicates are (from inspecting `duplicated_selection`), and therefore can validate.

pandas/tests/indexes/common.py
h-vetinari: Happy to leave the bare return in there, if that's preferred.

jreback: well, put a breakpoint there and see if you would.

h-vetinari: I'm not able to do work on any of this for 2 weeks now. I'll have a look after.

h-vetinari:
> is this actually hit?

@jreback The if-branch you mentioned in `test_pickle_compat_construction` was not hit, therefore I removed it. Please re-review.

h-vetinari:
The travis failure is unrelated. There seems to be a problem with the 3.5 job when collecting parametrized tests, in that the order is not stable, which yields a failure due to different tests

```
==================================== ERRORS ====================================
_____________________________ ERROR collecting gw1 _____________________________
Different tests were collected between gw0 and gw1. The difference is:

[...]

 pandas/tests/frame/test_apply.py::TestDataFrameAggregate::()::test_agg_cython_table[axis 0-df0-sum-expected0]
-pandas/tests/frame/test_apply.py::TestDataFrameAggregate::()::test_agg_cython_table[axis 0-df1-func1-expected1]
-pandas/tests/frame/test_apply.py::TestDataFrameAggregate::()::test_agg_cython_table[axis 0-df2-sum-expected2]
+pandas/tests/frame/test_apply.py::TestDataFrameAggregate::()::test_agg_cython_table[axis 0-df1-sum-expected1]
+pandas/tests/frame/test_apply.py::TestDataFrameAggregate::()::test_agg_cython_table[axis 0-df2-func2-expected2]
 pandas/tests/frame/test_apply.py::TestDataFrameAggregate::()::test_agg_cython_table[axis 0-df3-max-expected3]
 pandas/tests/frame/test_apply.py::TestDataFrameAggregate::()::test_agg_cython_table[axis 0-df4-amax-expected4]
 pandas/tests/frame/test_apply.py::TestDataFrameAggregate::()::test_agg_cython_table[axis 0-df5-func5-expected5]

[...]
```

pandas/tests/indexes/common.py
h-vetinari: Opened a follow-up: #22211

h-vetinari: Nevermind, I had gotten confused between instances and classes.

So `DatetimeIndex`, `PeriodIndex` and `TimedeltaIndex` are subclasses of `Int64Index` - but are unsafe to cast to Int64...? I guess this is intentional?

In any case, under these circumstances, I'm even more convinced that it's best to just stay with the `if type(indices) is not self._holder:` condition.

h-vetinari:
@jreback ping. Should be ready to go - travis failure is unrelated.

jreback:
@h-vetinari actually no, your PR is causing the fail. you have non-determinism in the test generation. usually this is because the ordering of the fixtures / parameters is based on a dictionary.

h-vetinari:
@jreback 
> actually no, your PR is causing the fail

I honestly can't see how this would be the case. I only change tests (no fixtures or anything) in
- `pandas/tests/indexes/common.py`
- `pandas/tests/indexes/test_category.py`
- `pandas/tests/indexes/test_range.py`

and the collection errors are in `pandas/tests/frame/test_apply.test_agg_cython_table`.

The parametrization of that test is
```
    @pytest.mark.parametrize("df, func, expected", chain(
        _get_cython_table_params(
            DataFrame(), [
                ('sum', Series()),
                ('max', Series()),
                ('min', Series()),
                ('all', Series(dtype=bool)),
                ('any', Series(dtype=bool)),
                ('mean', Series()),
                ('prod', Series()),
                ('std', Series()),
                ('var', Series()),
                ('median', Series()),
            ]),
        _get_cython_table_params(
            DataFrame([[np.nan, 1], [1, 2]]), [
                ('sum', Series([1., 3])),
                ('max', Series([1., 2])),
                ('min', Series([1., 1])),
                ('all', Series([True, True])),
                ('any', Series([True, True])),
                ('mean', Series([1, 1.5])),
                ('prod', Series([1., 2])),
                ('std', Series([np.nan, 0.707107])),
                ('var', Series([np.nan, 0.5])),
                ('median', Series([1, 1.5])),
            ]),
    ))
```

but I cannot hunt down further because the top imports `from pandas.conftest import _get_cython_table_params`, but `pandas.conftest` does not contain anything of that name.

h-vetinari:
rebased again, let's see if it helps

h-vetinari:
@jreback all green

h-vetinari:
Btw, that issue with the test order is related to #22156, #22157

jreback (approved):

jreback:
thanks @h-vetinari

PR #21903, opened 2018-07-13
[REF] implement internals as dir

jbrockmendel:
In the name of a) cleaning up internals and b) isolating BlockManager from everything else, this separates core.internals into `internals.managers`, `internals.blocks`, `internals.concat`.


pandas/core/internals/managers.py
```diff
@@ -0,0 +1,1892 @@
...
+
+import numpy as np
+
+from pandas._libs import lib, internals as libinternals
+
+from pandas.util._validators import validate_bool_kwarg
+
+from pandas.compat import range, map, zip
+
+import pandas.core.dtypes.generic as gt
```
jbrockmendel: I noticed this usage in `tests.dtypes.test_generic`, think it's a nice way to cut down massive import blocks/namespaces.

pandas/core/internals/blocks.py
```diff
@@ -5393,518 +3469,117 @@ def _putmask_preserve(nv, n):
...
+    if blocks is None:
+        blocks = []
+    if isinstance(result, list):
+        for r in result:
+            if isinstance(r, list):
+                blocks.extend(r)
             else:
-                concat_values = concat_values.copy()
+                blocks.append(r)
+    elif isinstance(result, BlockManager):
```
jbrockmendel: This is the only place where internals.blocks depends on either `internals.concat` or `internals.managers`.  DAG FTW

jorisvandenbossche:
I am not fully convinced this is needed, *before* we actually have concrete plans to actively refactor the code. For example, splitting for just splitting makes git blame not work anymore

jbrockmendel:
The git blame point is reasonable, but I'd like to make a case that this is a step in the right direction.

1) The long-term goal of ripping out the BlockManager will be made easier if core.internals is better isolated from the rest of the code.
   a) Decreasing the exposed surface of core.internals will help
2) The Block classes themselves need a lot of work (e.g. several datetime64 and timedelta64 interpolation issues), which will be _much_ easier to do if they can be reasoned about in isolation of the rest of the core.internals machinery.

That said, while I do think this would make working in this area much easier, my opinion isn't strong enough to make a big deal about it.

jreback:
I am ok with this. internals is just a giant file, things are much easier to grok for bug fixes / refactorings if its not so giant. still have to review.

jbrockmendel:
>  still have to review.

The diff GH shows for the blocks.py file is way more complicated than it needs to be since this is pretty much cut/paste.  LMK if doing this in smaller pieces would be easier (i.e. first move internals.py unchanged to internals.\_\_init\_\_.py before breaking out the independent modules)

jreback (changes requested):

jreback:
> The diff GH shows for the blocks.py file is way more complicated than it needs to be since this is pretty much cut/paste. LMK if doing this in smaller pieces would be easier (i.e. first move internals.py unchanged to internals.__init__.py before breaking out the independent modules)

yes let's do that

jreback (approved):
lgtm merge on green

jbrockmendel:
> yes let's do that

done.

jreback:
needs a rebase

jreback:
merge as soon as travis is green.

jorisvandenbossche:
@jbrockmendel to come back to your points why this is an improvement

> The long-term goal of ripping out the BlockManager will be made easier if core.internals is better isolated from the rest of the code.

But Blocks and the BlockManager would be ripped out together, so they are already isolated in that sense. And the code / implementation of both are strongly intertwined., slitting those both up does make it IMO not easier to work with.

> The Block classes themselves need a lot of work

In general, by using extension arrays for our internal ones, the code in the specific Blocks should decrease, as we should move towards using ExtensionBlock for all those.

So I am still not convinced of the benefit of this change in general (there might be smaller (eg generally useable) parts that can be refactored / moved out).

jreback:
@jorisvandenbossche this is just in general a good change. Making things simplier / easier to grok is a huge +1. Wether / IF block manager is every completely removed is completely not the point here. In order to even attempt that, you have to gradually move away from it. So making code more clear is the first step.

Trying to make sweeping changes is always a disaster, they take too long to review and are very disruptive. Incremental changes are much much much better.

jreback:
> In general, by using extension arrays for our internal ones, the code in the specific Blocks should decrease, as we should move towards using ExtensionBlock for all those.

maybe, though the ExtensionBlock themselves need quite a bit of work too. As things are moving closer together it is simply better to have a way to refactor code intelligently. 2000+ line modules is not the way forward.

jbrockmendel:
> And the code / implementation of both are strongly intertwined., slitting those both up does make it IMO not easier to work with.

If you recall the previous version of the PR, the implementations were not especially "intertwined", the dependencies ran almost entirely one-direction.

There is also a bunch of internals-aware code in core.reshape that I consider "intertwined", hopefully we can tighten the exposed surface there.  (I don't know that area of the code especially well, so not sure how hard that will be)

jorisvandenbossche:
> If you recall the previous version of the PR, the implementations were not especially "intertwined", the dependencies ran almost entirely one-direction.

That's what I meant with "intertwined" (BlockManager methods calling a lot of Block methods, and many Block methods only used in BlockManager)

PR #21904, opened 2018-07-13
BUG: issues with hash-function for Float64HashTable (GH21866)

### LINKED ISSUE #21866: Issues with hash-function for float64 version of klib's hash-map
#### Problem description

Hash-maps for float64 use the following [hash-function](https://github.com/pandas-dev/pandas/blob/master/pandas/_libs/src/klib/khash_python.h#L22)

```
#define kh_float64_hash_func(key) (khint32_t)((asint64(key))>>33^(asint64(key))^(asint64(key))<<11)
```

However, in order to guarantee consistent behavior, the following constrains must be met:

  1. `==` must be an equivalence relation
  2. `x==y`  => `hash(x)==hash(y)`

Following IEEE-754, floats aren't an equivalence relation, this is fixed [defining](https://github.com/pandas-dev/pandas/blob/master/pandas/_libs/src/klib/khash_python.h#L23) "equal" as
```
#define kh_float64_hash_equal(a, b) ((a) == (b) || ((b) != (b) && (a) != (a)))
```
Thus, apart from trivial equivalence classes, there are the following two:

 1. `{0.0, -0.0}`
 2. `{x | x is NAN} `

for which the second constrain doesn't hold:

  * `hash(0.0)=0 != 1073741824=2^30=hash(-0.0)`
  * `hash(float("nan")) = 1073479680 != 2147221504 =hash(-float("nan"))`

Due to the way klib [uses this hash value](https://github.com/pandas-dev/pandas/blob/master/pandas/_libs/src/klib/khash.h#L290), the values "0.0" and "-0.0" end up in different buckets only if there are more than 2^29 (ca. 6e8) elements in the hash-map already. The same holds for `float("nan")` and `-float("nan")`. However in the case of not-a-number there are much more elements in the equivalence class, so the inconsistent behavior can be triggered for much smaller sizes, for example for

```
import struct
NAN1=struct.unpack("d", struct.pack("L", 9221120237041090560))[0]
NAN2=struct.unpack("d", struct.pack("L", 9221120237041090561))[0]
```

### Proposed solution:

There are already some workarounds for this problem scattered throughout pandas, for example for `pd.unique(...)` [here](https://github.com/pandas-dev/pandas/blob/8cbee356da1161c56c64f6f89cb5548bcadc3e44/pandas/_libs/hashtable_class_helper.pxi.in#L284) or for `isin` [here](https://github.com/pandas-dev/pandas/blob/ec5956ed350d33ac2cee07bf9a24ea5315529443/pandas/_libs/hashtable_func_helper.pxi.in#L274).

However, this

 1. is not the right place to work around the bug, as it makes the code not really  clearer
 2. fixes only the not-a-number case, but not `{0.0, -0.0}`
 3. most probably everyone reusing the hash-map-implementation in the fututre will first get the NAN-case wrong and then fix it with a further workaround.
 
A better approach would be to fix the hash-function of float64-hash-map, a possibility would be:

```
   #include <math.h>
    //right for all but not -0.0 and NANs
    #define kh_float64_hash_func_0_NAN(key) (khint32_t)((f64_to_i64(key))>>33^(f64_to_i64(key))^(f64_to_i64(key))<<11)
    //right for all except NANs
    #define kh_float64_hash_func_NAN(key) ((key)==0.0 ? kh_float64_hash_func_0_NAN(0.0) : kh_float64_hash_func_0_NAN(key))
    //right for all, also- 0.0 and NANs
    #define kh_float64_hash_func(key) ((key) != (key) ? kh_float64_hash_func_NAN(NAN) : kh_float64_hash_func_NAN(key))
```

realead:
The following issues

   1)  hash(0.0) != hash(-0.0)
   2)  hash(x) != hash(y) for different x,y which are nans

are solved by setting:

   1) hash(-0.0):=hash(0.0)
   2) hash(x):=hash(np.nan) for every x which is nan

- [x] closes #21866
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -471,6 +471,7 @@ Numeric
 
 - Bug in :class:`Series` ``__rmatmul__`` doesn't support matrix vector multiplication (:issue:`21530`)
 - Bug in :func:`factorize` fails with read-only array (:issue:`12813`)
+- Fixed bug in :func:`unique` handled signed zeros inconsistently: for some inputs 0.0 and -0.0 were treated as equal and for some inputs as different. Now they are treated as equal for all inputs (:issue:`21866`)
 -
 -
 
--- a/pandas/_libs/src/klib/khash_python.h
+++ b/pandas/_libs/src/klib/khash_python.h
@@ -19,7 +19,20 @@ khint64_t PANDAS_INLINE asint64(double key) {
   memcpy(&val, &key, sizeof(double));
   return val;
 }
-#define kh_float64_hash_func(key) (khint32_t)((asint64(key))>>33^(asint64(key))^(asint64(key))<<11)
+
+// correct for all inputs but not -0.0 and NaNs
+#define kh_float64_hash_func_0_NAN(key) (khint32_t)((asint64(key))>>33^(asint64(key))^(asint64(key))<<11)
+
+// correct for all inputs but not NaNs
+#define kh_float64_hash_func_NAN(key) ((key) == 0.0 ?                       \
+                                        kh_float64_hash_func_0_NAN(0.0) : \
+                                        kh_float64_hash_func_0_NAN(key))
+
+// correct for all
+#define kh_float64_hash_func(key) ((key) != (key) ?                       \
+                                   kh_float64_hash_func_NAN(Py_NAN) :     \
+                                   kh_float64_hash_func_NAN(key))
+
 #define kh_float64_hash_equal(a, b) ((a) == (b) || ((b) != (b) && (a) != (a)))
 
 #define KHASH_MAP_INIT_FLOAT64(name, khval_t)								\
--- a/pandas/tests/test_algos.py
+++ b/pandas/tests/test_algos.py
@@ -7,6 +7,7 @@ from numpy.random import RandomState
 from numpy import nan
 from datetime import datetime
 from itertools import permutations
+import struct
 from pandas import (Series, Categorical, CategoricalIndex,
                     Timestamp, DatetimeIndex, Index, IntervalIndex)
 import pandas as pd
@@ -499,6 +500,23 @@ class TestUnique(object):
         expected = np.array(['foo', None], dtype=object)
 
         tm.assert_numpy_array_equal(result, expected, strict_nan=True)
+
+    def test_signed_zero(self):
+        a = np.array([-0.0, 0.0])
+        result = pd.unique(a)
+        expected = np.array([-0.0])  # 0.0 and -0.0 are equivalent
+        tm.assert_numpy_array_equal(result, expected)
+
+    def test_different_nans(self):
+        # create different nans from bit-patterns:
+        NAN1 = struct.unpack("d", struct.pack("=Q", 0x7ff8000000000000))[0]
+        NAN2 = struct.unpack("d", struct.pack("=Q", 0x7ff8000000000001))[0]
+        assert NAN1 != NAN1
+        assert NAN2 != NAN2
+        a = np.array([NAN1, NAN2])  # NAN1 and NAN2 are equivalent
+        result = pd.unique(a)
+        expected = np.array([np.nan])
+        tm.assert_numpy_array_equal(result, expected)
 
 
 class TestIsin(object):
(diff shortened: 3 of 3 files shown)


jbrockmendel:
Test failure appears unrelated.  Can you push again to re-run it?

jreback:
can you run an asv and report any anomalies

jreback (changes requested):
does this have any user level visible effects?
can you add a whatsnew note (0.24.0), depending on your answer above either API changes or other Enhancements section.

realead:
@jbrockmendel  No, it was a problem in the one of the added test cases: somewhat naively 16GB memory were reserved (but not committed!), yet obviously different operating systems with different resources reacts differently to such a request.

It is a little bit strange, that the testing just died and didn't recover from that...

I removed these silly tests and now it looks better (at least it is clear, what goes wrong).

pandas/_libs/src/klib/khash_python.h
```diff
@@ -19,7 +19,18 @@ khint64_t PANDAS_INLINE asint64(double key) {
   memcpy(&val, &key, sizeof(double));
   return val;
 }
-#define kh_float64_hash_func(key) (khint32_t)((asint64(key))>>33^(asint64(key))^(asint64(key))<<11)
+
+// correct for all inputs but not -0.0 and NaNs
+#define kh_float64_hash_func_0_NAN(key) (khint32_t)((asint64(key))>>33^(asint64(key))^(asint64(key))<<11)
```
jreback: can you add a blank between cases

realead: @jreback Sorry for the silly question: Do you expect me to add a new commit with the improvements to the branch and you will fixup it when merging or should I amend the current commit?

jreback: generally push new commits.

jreback: needs to add a whatsnew in any event

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -235,7 +235,7 @@ Other API Changes
   a ``KeyError`` (:issue:`21678`).
 - Invalid construction of ``IntervalDtype`` will now always raise a ``TypeError`` rather than a ``ValueError`` if the subdtype is invalid (:issue:`21185`)
 - Trying to reindex a ``DataFrame`` with a non unique ``MultiIndex`` now raises a ``ValueError`` instead of an ``Exception`` (:issue:`21770`)
--
+- :class:`Float64HashTable` handles zeros/signed zeros and all flavors of NaNs consistently: it is no longer possible to have both, zero and signed-zero, as keys at the same time in a table, also there can be at most one NaN-key in a table (:issue:`21866`)
```
jreback: this is not public, my question below was whether this has a public change?

jreback (changes requested):

doc/source/whatsnew/v0.24.0.txt
realead: It fixes the bug #21866, i.e. produces right results for some esoteric corner cases. This change of behavior can be observed by the end-user, but is this then a public change worth mentioning?

jreback: https://github.com/pandas-dev/pandas/issues/21866#issuecomment-404402552

is this the public case? you can list this, but just make it related to the effect on ``.unique()``

jreback: also add a test w.r.t. unique

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -84,6 +84,7 @@ Other Enhancements
 - :meth:`Series.nlargest`, :meth:`Series.nsmallest`, :meth:`DataFrame.nlargest`, and :meth:`DataFrame.nsmallest` now accept the value ``"all"`` for the ``keep`` argument. This keeps all ties for the nth largest/smallest value (:issue:`16818`)
 - :class:`IntervalIndex` has gained the :meth:`~IntervalIndex.set_closed` method to change the existing ``closed`` value (:issue:`21670`)
 - :func:`~DataFrame.to_csv` and :func:`~DataFrame.to_json` now support ``compression='infer'`` to infer compression based on filename (:issue:`15008`)
+- :class:`Float64HashTable` handles zeros/signed zeros and all flavors of NaNs consistently: it is no longer possible to have both, zero and signed-zero, as keys at the same time in a table, also there can be at most one NaN-key in a table (:issue:`21866`)
```
jreback: right, can you reword to just focus on ``.unique()``

pandas/tests/test_algos.py
```diff
@@ -500,6 +501,23 @@ def test_obj_none_preservation(self):
 
         tm.assert_numpy_array_equal(result, expected, strict_nan=True)
 
+    def test_signed_zero(self):
+        a = np.array([-0.0, 0.0])
```
jreback: can you add the issue number here as a comment (and on test below)

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -84,6 +84,7 @@ Other Enhancements
 - :meth:`Series.nlargest`, :meth:`Series.nsmallest`, :meth:`DataFrame.nlargest`, and :meth:`DataFrame.nsmallest` now accept the value ``"all"`` for the ``keep`` argument. This keeps all ties for the nth largest/smallest value (:issue:`16818`)
 - :class:`IntervalIndex` has gained the :meth:`~IntervalIndex.set_closed` method to change the existing ``closed`` value (:issue:`21670`)
 - :func:`~DataFrame.to_csv` and :func:`~DataFrame.to_json` now support ``compression='infer'`` to infer compression based on filename (:issue:`15008`)
+- :func:`unique` handles signed zeros consistently: it is no longer possible to have both, 0.0 and -0.0, in the same resulting array (:issue:`21866`)
```
jreback: move to bug fix / Numeric section.

jreback (changes requested):
doc comments. rebase & ping on green.

jreback (approved):

jreback:
@jbrockmendel any comments? if ok pls merge.

jbrockmendel:
@realead good job catching a subtle bug.  Thanks for taking point on this.

Issue #21908, opened 2018-07-14
[EA] ExtensionArray should support ndim==2


jbrockmendel:
There are a bunch of issues (indirectly) caused by the fact that we can't reshape a `DatetimeIndex[tz]` as 2D.

#13287, #17539, #19197, and each of the several issues about interpolation with `datetime64[tz]` dtypes.  At the end of `DataFrame._init_ndarray` is:

```
        if dtype is None and is_object_dtype(values):
            values = maybe_infer_to_datetimelike(values)

        return create_block_manager_from_blocks([values], [columns, index])
```

but in the relevant cases, `maybe_infer_to_datetimelike` returns a tz-aware `DatetimeIndex`, which `create_block_manager_from_blocks` raises on because it is expecting `values` to be 2D.

I cleanest way to fix this seems to be supporting 2D shapes for `DatetimeArray` (and ideally others for compat)

jbrockmendel:
> We deliberately restricted it to 1D, and I think that is still the good way to go.

I'll try to track down this discussion and see the original reasoning.  AFAICT the main thing we get out of the 1D restriction is not having to worry about `axis` arguments, but we're paying for it by sacrificing a lot of the generality people think of with "array"s.

jreback:
i agree that we should restrict to 1D and just move datetime to EA

jbrockmendel:
> i agree that we should restrict to 1D and just move datetime to EA

We _are_ moving datetime to EA, that is currently in progress.  What I'm saying is that there are bugs that are caused by the inability to reshape e.g. `dtarr.reshape((N, 1))` and I'm not clear what we get in exchange for that restriction (still looking for the old discussion).

jorisvandenbossche:
> What I'm saying is that there are bugs that are caused by the inability to reshape 

We will need to update the code to handle with 1D instead of 2D, for sure.

> AFAICT the main thing we get out of the 1D restriction is not having to worry about axis arguments, 

Yes, I think that is one of the main reasons. This fact is one of the things that make the current block manager really complex.

jbrockmendel:
Fair enough.  Closing, will chime in in the appropriate place.

Issue #19199, opened 2018-01-12
Interpolating of datetime64 values ignored (at best)


jbrockmendel:
tz naive case:
```
index = pd.Index([23, 26, 30])
dti = pd.DatetimeIndex(['2015-09-23', '2015-09-26', '2015-09-30'])
ser = pd.Series(dti, index=index).reindex(range(23, 31))
df = pd.DataFrame(ser)

>>> ser    # for reference
23   2015-09-23
24          NaT
25          NaT
26   2015-09-26
27          NaT
28          NaT
29          NaT
30   2015-09-30
dtype: datetime64[ns]

>>> ser.interpolate()
23   2015-09-23
24          NaT
25          NaT
26   2015-09-26
27          NaT
28          NaT
29          NaT
30   2015-09-30
dtype: datetime64[ns]

>>> df.interpolate()
            0
23 2015-09-23
24        NaT
25        NaT
26 2015-09-26
27        NaT
28        NaT
29        NaT
30 2015-09-30
```

tz-aware case
```
index = pd.Index([23, 26, 30])
dti = pd.DatetimeIndex(['2015-09-23', '2015-09-26', '2015-09-30'], tz='US/Central')
ser = pd.Series(dti, index=index).reindex(range(23, 31))
df = pd.DataFrame(ser)

>>> ser
23   2015-09-23 00:00:00-05:00
24                         NaT
25                         NaT
26   2015-09-26 00:00:00-05:00
27                         NaT
28                         NaT
29                         NaT
30   2015-09-30 00:00:00-05:00
dtype: datetime64[ns, US/Central]

>>> ser.interpolate()
23   2015-09-23 00:00:00-05:00
24                         NaT
25                         NaT
26   2015-09-26 00:00:00-05:00
27                         NaT
28                         NaT
29                         NaT
30   2015-09-30 00:00:00-05:00
dtype: datetime64[ns, US/Central]

>>> df.interpolate()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/pandas/core/generic.py", line 4738, in interpolate
    'object') == len(_maybe_transposed_self.T):
  File "/usr/local/lib/python2.7/site-packages/pandas/core/frame.py", line 1909, in transpose
    return super(DataFrame, self).transpose(1, 0, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/pandas/core/generic.py", line 599, in transpose
[...]

Issue #19198, opened 2018-01-12
BUG: DataFrame creation from a tz-aware Index


jbrockmendel:
```
dti = pd.date_range('2016-09-23', periods=3, tz='US/Central')
df = pd.DataFrame(dti)

>>> df.T
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pandas/core/frame.py", line 1941, in transpose
    return super(DataFrame, self).transpose(1, 0, **kwargs)
  File "pandas/core/generic.py", line 616, in transpose
    new_values = self.values.transpose(axes_numbers)
  File "pandas/core/base.py", line 697, in transpose
    nv.validate_transpose(args, kwargs)
  File "pandas/compat/numpy/function.py", line 54, in __call__
    self.defaults)
  File "pandas/util/_validators.py", line 218, in validate_args_and_kwargs
    validate_kwargs(fname, kwargs, compat_args)
  File "pandas/util/_validators.py", line 157, in validate_kwargs
    _check_for_default_values(fname, kwds, compat_args)
  File "pandas/util/_validators.py", line 69, in _check_for_default_values
    format(fname=fname, arg=key)))
ValueError: the 'axes' parameter is not supported in the pandas implementation of transpose()
```

Looks like this is b/c `df.values` is a `DatetimeIndex`

jreback:
xref to https://github.com/pandas-dev/pandas/issues/13287

jreback:
this is a duplicate and the same cause as #13287

Issue #11701, opened 2015-11-25
Interpolate NaT


scls19fr:
Hello,

`interpolate` doesn't work with `NaT`
see http://stackoverflow.com/questions/33921795/fill-timestamp-nat-with-a-linear-interpolation/33922824#33922824

Here is a trivial example to show the situation:

```
s = pd.Series(pd.date_range('2015-01-01' , '2015-01-30'), name='t')

s[3], s[4], s[5] = pd.NaT, pd.NaT, pd.NaT
s[13], s[14], s[15] = pd.NaT, pd.NaT, pd.NaT
print(s)

0    2015-01-01
1    2015-01-02
2    2015-01-03
3           NaT
4           NaT
5           NaT
6    2015-01-07
7    2015-01-08
8    2015-01-09
9    2015-01-10
10   2015-01-11
11   2015-01-12
12   2015-01-13
13          NaT
14          NaT
15          NaT
16   2015-01-17
17   2015-01-18
18   2015-01-19
19   2015-01-20
20   2015-01-21
21   2015-01-22
22   2015-01-23
23   2015-01-24
24   2015-01-25
25   2015-01-26
26   2015-01-27
27   2015-01-28
28   2015-01-29
29   2015-01-30
Name: t, dtype: datetime64[ns]

print(s.interpolate())
0    2015-01-01
1    2015-01-02
2    2015-01-03
3           NaT
4           NaT
5           NaT
6    2015-01-07
7    2015-01-08
8    2015-01-09
9    2015-01-10
10   2015-01-11
11   2015-01-12
12   2015-01-13
13          NaT
14          NaT
15          NaT
16   2015-01-17
17   2015-01-18
18   2015-01-19
19   2015-01-20
20   2015-01-21
21   2015-01-22
22   2015-01-23
23   2015-01-24
24   2015-01-25
25   2015-01-26
26   2015-01-27
27   2015-01-28
28   2015-01-29
29   2015-01-30
Name: t, dtype: datetime64[ns]

assert s.interpolate().isnull().sum() == 0
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-150-8a59e397a174> in <module>()
----> 1 assert s.interpolate().isnull().sum() == 0

AssertionError:
```

Kind regards

jreback:
this is not implemented ATM on datetimes. pull-requests are welcome.

scls19fr:
[...]

PR #21915, opened 2018-07-14
[BUG] Fix interpolation for datetimelike dtypes

### LINKED ISSUE #11312: ENH: interpolation of datetime values (NaT)
I didn't directly found an open issue about it, but is there a reason we do not implement interpolation of datetime _values_ (so not in the index)?

Example:

```
In [14]: s = pd.Series(pd.date_range('2012-01-01', periods=5))

In [16]: s[2] = np.nan

In [17]: s
Out[17]:
0   2012-01-01
1   2012-01-02
2          NaT
3   2012-01-04
4   2012-01-05
dtype: datetime64[ns]

In [18]: s.interpolate()
Out[18]:
0   2012-01-01
1   2012-01-02
2          NaT
3   2012-01-04
4   2012-01-05
dtype: datetime64[ns]
```

A crude manual work around:

```
In [20]: s2 = pd.Series(s.values.astype('int64'))

In [21]: s2[s2<0] = np.nan

In [22]: s2
Out[22]:
0    1.325376e+18
1    1.325462e+18
2             NaN
3    1.325635e+18
4    1.325722e+18
dtype: float64

In [23]: pd.to_datetime(s2.interpolate(), unit='ns')
Out[23]:
0   2012-01-01
1   2012-01-02
2   2012-01-03
3   2012-01-04
4   2012-01-05
dtype: datetime64[ns]
```

jbrockmendel:
- [x] closes #11312, #11701, #19199,
- [x] tests added / passed
- [x] whatsnew entry

Two bugs here, one fixed, one avoided.  First is in `Block._interpolate` where datetimelike values were not cast correctly.  Second is that `DataFrame.transpose` will raise in some conditions (see #19198, also motivated #21908).

This adds dtype-handling code in `Block._interpolate`.  An alternative would be to override `_interpolate` in subclasses.  Either way works for me.

NB: This is only implemented for `method='linear'`.  I made that explicit in the tests as a reminder to follow-up with others.

(#19199 is marked as  a duplicate but it includes a bug report for the TZ-aware case that is separate bug.)

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -683,7 +683,11 @@ class NDFrame(PandasObject, SelectionMixin):
 
         new_axes = self._construct_axes_dict_from(self, [self._get_axis(x)
                                                          for x in axes_names])
-        new_values = self.values.transpose(axes_numbers)
+        values = self.values
+        if isinstance(values, DatetimeIndex):
+            # kludge for tz-aware case.  See GH#19198
+            values = values.astype('O').values.reshape(self.shape)
+        new_values = values.transpose(axes_numbers)
         if kwargs.pop('copy', None) or (len(args) and args[-1]):
             new_values = new_values.copy()
 
@@ -6097,8 +6101,11 @@ class NDFrame(PandasObject, SelectionMixin):
             raise ValueError("Only `method=linear` interpolation is supported "
                              "on MultiIndexes.")
 
-        if _maybe_transposed_self._data.get_dtype_counts().get(
-                'object') == len(_maybe_transposed_self.T):
+        dtype_counts = _maybe_transposed_self._data.get_dtype_counts()
+        if ('object' in dtype_counts and
+                dtype_counts.get('object') == len(_maybe_transposed_self.T)):
+            # Try to short-circuit tranposing to avoid superfluous dimension
+            # errors GH#13287, GH#17539, GH#19197
             raise TypeError("Cannot interpolate with all NaNs.")
 
         # create/use the index
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -684,8 +684,6 @@ class NDFrame(PandasObject, SelectionMixin):
         new_axes = self._construct_axes_dict_from(self, [self._get_axis(x)
                                                          for x in axes_names])
         new_values = self.values.transpose(axes_numbers)
-
-        new_values = values.transpose(axes_numbers)
         if kwargs.pop('copy', None) or (len(args) and args[-1]):
             new_values = new_values.copy()
 
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -683,10 +683,8 @@ class NDFrame(PandasObject, SelectionMixin):
 
         new_axes = self._construct_axes_dict_from(self, [self._get_axis(x)
                                                          for x in axes_names])
-        values = self.values
-        if isinstance(values, DatetimeIndex):
-            # kludge for tz-aware case.  See GH#19198
-            values = values.astype('O').values.reshape(self.shape)
+        new_values = self.values.transpose(axes_numbers)
+
         new_values = values.transpose(axes_numbers)
         if kwargs.pop('copy', None) or (len(args) and args[-1]):
             new_values = new_values.copy()
--- a/pandas/core/internals.py
+++ b/pandas/core/internals.py
@@ -15,6 +15,7 @@ from pandas._libs import internals as libinternals
 
 from pandas.core.base import PandasObject
 
+import pandas.core.dtypes.common as ct
 from pandas.core.dtypes.dtypes import (
     ExtensionDtype, DatetimeTZDtype,
     PandasExtensionDtype,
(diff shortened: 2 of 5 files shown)


pandas/core/generic.py
```diff
@@ -684,6 +684,8 @@ def transpose(self, *args, **kwargs):
         new_axes = self._construct_axes_dict_from(self, [self._get_axis(x)
                                                          for x in axes_names])
         new_values = self.values.transpose(axes_numbers)
```
jreback: shouldn’t u remove the previous line?

pandas/core/internals.py
```diff
@@ -1199,20 +1199,35 @@ def _interpolate_with_fill(self, method='pad', axis=0, inplace=False,
...
                      fill_value=None, axis=0, limit=None,
                      limit_direction='forward', limit_area=None,
                      inplace=False, downcast=None, mgr=None, **kwargs):
         """ interpolate using scipy wrappers """
 
         inplace = validate_bool_kwarg(inplace, 'inplace')
-        data = self.values if inplace else self.values.copy()
 
         # only deal with floats
-        if not self.is_float:
```
jreback: i prefer to import these rather than use ct

pandas/core/internals.py
```diff
@@ -1199,20 +1199,35 @@ def _interpolate_with_fill(self, method='pad', axis=0, inplace=False,
...
+        if ct.needs_i8_conversion(self.dtype):
+            if ct.is_period_dtype(self.dtype):
+                raise NotImplementedError("PeriodDtype columns/Series don't "
```
jreback: let’s move code to subclasses as appropriate
this gets messy really fast otherwise

pandas/core/internals.py
```diff
@@ -1236,6 +1250,20 @@ def func(x):
 
         # interp each column independently
         interp_values = np.apply_along_axis(func, axis, data)
+        if ct.needs_i8_conversion(self.dtype):
```
jreback: same

pandas/tests/frame/test_missing.py
```diff
@@ -855,3 +858,71 @@ def test_interp_ignore_all_good(self):
...
         assert_frame_equal(result, df[['B', 'D']])
+
+    @pytest.mark.parametrize('use_idx', [True, False])
+    @pytest.mark.parametrize('tz', [None, 'US/Central'])
+    def test_interpolate_dt64_values(self, tz, use_idx):
+        # GH#21915
+        dti = pd.date_range('2016-01-01', periods=10, tz=tz)
+        index = dti if use_idx else None
+
+        # Copy to avoid corrupting dti, see GH#21907
```
jreback: can u make this more informative ; this is very crytptix

pandas/tests/frame/test_missing.py
```diff
@@ -855,3 +858,71 @@ def test_interp_ignore_all_good(self):
...
+        ser = pd.Series(dti, index=index).copy()
+        ser[::3] = pd.NaT
+
+        expected = pd.Series(dti, index=index)
+        expected.iloc[0] = pd.NaT
+        expected.iloc[-1] = expected.iloc[-2]
+
+        df = ser.to_frame()
```
jreback: why are you converting to frames? can u just start with frames?

pandas/tests/series/test_missing.py
```diff
@@ -1317,3 +1317,38 @@ def test_series_interpolate_intraday(self):
         result = ts.reindex(new_index).interpolate(method='time')
 
         tm.assert_numpy_array_equal(result.values, exp.values)
+
+    # TODO: De-duplicate with similar tests in test.frame.test_missing?
```
jreback: this is much more complicated we have generic tests that already do a lot of this

you can also move these tests there

jreback (changes requested):

pandas/core/generic.py
jbrockmendel: Yah, thought I had removed this no-longer-needed bit, left a line behind.  Will fix.

pandas/core/internals.py
jbrockmendel: Darn, I was hoping that would catch on (I don't like the giant namespaces).  Will change.

pandas/tests/frame/test_missing.py
```diff
@@ -360,7 +360,10 @@ def test_fillna_categorical_nan(self):
         cat = Categorical([np.nan, 2, np.nan])
         val = Categorical([np.nan, np.nan, np.nan])
         df = DataFrame({"cats": cat, "vals": val})
-        res = df.fillna(df.median())
+        with tm.assert_produces_warning(RuntimeWarning):
```
jbrockmendel: Unrelated, but nice to catch.

pandas/tests/series/test_missing.py
jbrockmendel: I'll take a look for a better place.

This is also the answer to the previous question about why not starting with DataFrame: bc the test originally tested Series and DataFrame in the same test.

jbrockmendel:
I'll hold off on pushing this until #21870 (and possibly #21903) go through to avoid rebasing hell

pandas/tests/series/test_missing.py
jbrockmendel: > this is much more complicated we have generic tests that already do a lot of this

Both for this PR and the upcoming slew of Datetime/Timedelta/PeriodArray tests, there are a bunch of for FooArray/FooIndex/Series[foo]/DataFrame[foo-column] that are highly duplicative and I think things are falling through the cracks.  Think we can find a place to put them so they can be parametrized?

jbrockmendel:
Closing temporarily to clear the queue.

nielsuit227:
Looks like this was lost somewhere in the process. Pity, would be great to see this somewhen :)

jbrockmendel:
@nielsuit227 you're welcome to pick this up and run with it

PR #18852, opened 2017-12-19
ENH: df.assign accepting dependent **kwargs (#14207)

datajanko:
Specifically, 'df.assign(b=1, c=lambda x:x['b'])'
does not throw an exception in python 3.6 and above.
Further details are discussed in Issues #14207 and #18797.

closes #14207 
closes #18797
- [X] tests added / passed
- [X] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [X] whatsnew entry

PR #21917, opened 2018-07-14
DOC: Updated the DataFrame.assign docstring

aeltanawy:
Updated the DataFrame.assign docstring example to use np.arange instead of np.random.randn to pass the validation test.


jschendel (commented):
Thanks, lgtm aside from one small comment

pandas/core/frame.py
```diff
@@ -3353,38 +3353,39 @@ def assign(self, **kwargs):
 
         Examples
         --------
-        >>> df = pd.DataFrame({'A': range(1, 11), 'B': np.random.randn(10)})
+        >>> df = pd.DataFrame({'A': range(1, 11),
+        ...                    'B':np.arange(-1.0, 2.0, 0.3)})
```
jschendel: needs a space between `'B':` and `np.arange`

aeltanawy: Thanks @jschendel for spotting this! I'll send a fix right away.

datapythonista:
Thanks for the fix, that's much better than using random.

What would you think about simplifying the example even further? I don't think we need 10 rows to show what assign does. And using a "simpler" function than log (e.g. `lambda x: x ** 2`) would probably help users understand the example faster.

aeltanawy:
Thanks @datapythonista and @jorisvandenbossche! I'll shorten the table(s) to hold only 2 rows.

jreback:
lgtm. @aeltanawy normally don't put 2 commits like this together (e.g. the statsmodels change is already in master)

pandas/core/frame.py
```diff
@@ -3353,38 +3353,23 @@ def assign(self, **kwargs):
 
         Examples
         --------
-        >>> df = pd.DataFrame({'A': range(1, 11), 'B': np.random.randn(10)})
+        >>> df = pd.DataFrame({'A': range(1, 3),
+        ...                    'B': np.arange(-1.0, 2.0, 1.5)})
```
datapythonista: Now that there are only few items I think using something like `[1, 2, 3]` would be clearer than `range(1, 3)`. And same for the other parameter.

pandas/core/frame.py
```diff
@@ -3353,38 +3353,23 @@ def assign(self, **kwargs):
...
-        3   4 -0.269708  1.386294
-        4   5 -0.274002  1.609438
-        5   6 -0.500792  1.791759
-        6   7  1.649697  1.945910
-        7   8 -1.495604  2.079442
-        8   9  0.549296  2.197225
-        9  10 -0.758542  2.302585
+            A    B      ln_A
+         0  1 -1.0  0.000000
+         1  2  0.5  0.693147
```
datapythonista: I agree with @jorisvandenbossche that not using `lambda` would be better. But I don't think that's possible, given that the column from the dataframe need to be specified. So, unless there is an easy way to avoid lambda, I'd prefer a function where the user can quickly compute the result mentally, and understand faster what happened in the operation. And if it's a real example, even better, I think that makes things even simpler for the reader.

Not sure if this is too complex, but something like that would be better in my opinion:

```
import pandas as pd

df = pd.DataFrame([('liquid', 100.),
                   ('liquid', 356.73),
                   ('gas', -252.87)],
                  index=['water', 'mercury', 'hydrogen'],
                  columns=('state', 'boiling_point_c'))

df.assign(state=lambda x: pd.Categorical(x.state),
          boiling_point_f=lambda x: x.boiling_point_c * 9 / 5 + 32)
```

jorisvandenbossche:
> I agree with @jorisvandenbossche that not using lambda would be better.

Ah, yes, missed that the `np.log` example was also using `lambda` .. Indeed not easy to avoid in this case.

aeltanawy:
> @aeltanawy normally don't put 2 commits like this together (e.g. the statsmodels change is already in master)

Ah! Sorry for that, I messed up my branch when fetching upstream then rebasing. Hopefully all sorted out now.

@datapythonista inspired by your example, how about using simply temperatures in Celsius and Fahrenheit indexed with cities?

datapythonista:
Sure, I just gave an example to illustrate my point, and not leave an abstract comment. But feel free to use the example that you find more useful/clear for users.

aeltanawy:
I apologize for adding already committed commits, again! Anyone willing to show me what I'm doing wrong? (please let me know if this is not the place to ask)

My workflow is as follows:
1. edit the script in my local branch
2. git add, the git commit
3. then git pull origin <local branch>
4. then git push origin <local branch>

jorisvandenbossche:
@aeltanawy no problem! 

Looking at your workflow, the one thing that is normally not needed is third step (`git pull origin doc`).

Now, I am not really sure how doing that ended up in having those extra commits, but let's try to solve it.

I would first try:

```
git pull upstream master
git push origin doc
```

and check if that fixes the commits here in the PR (the above updates the PR with changes that have happened in the meantime in the upstream (the pandas-dev/pandas repo) master branch). 

If that didn't fix it, another approach is doing a rebase:

```
git fetch upstream
git rebase upstream/master
git push origin doc --force
```

(note here you will need to force push because rebasing means rewriting the history of the branch, not simply adding commits to the branch as merging does).

If you have more questions, feel free to ask!

aeltanawy:
I see now what I was doing wrong. I should have amend my commit instead of creating new ones, and not to fetch upstream and merge origin master. 

@jschendel, shall I close this PR and start a new one with the new edits?

datapythonista:
@aeltanawy I fix the git problems for you. If you simply do a `git pull` in your branch, you should be able to continue working with the docstring.

jorisvandenbossche (approved):
Looks very nice now!

pandas/core/frame.py
```diff
@@ -3354,38 +3354,23 @@ def assign(self, **kwargs):
 
         Examples
         --------
-        >>> df = pd.DataFrame({'A': range(1, 11), 'B': np.random.randn(10)})
+        >>> df = pd.DataFrame({'temp_c': (17.0, 25.0)},
+                                index=['portland', 'berkeley'])
```
datapythonista: Do you mind displaying `df` content after creating it? I think it'll help users understand faster that a new column is created in the assign later on.

Also, do you think it could look better capitalizing the names of the cities?

pandas/core/frame.py
```diff
@@ -3354,38 +3354,23 @@ def assign(self, **kwargs):
...
-        5   6 -0.500792  1.791759
-        6   7  1.649697  1.945910
-        7   8 -1.495604  2.079442
-        8   9  0.549296  2.197225
-        9  10 -0.758542  2.302585
+        >>> newcol = df['temp_c'] * 9 / 5 + 32
+        >>> df.assign(temp_f=newcol)
+                  temp_c  temp_f
+        portland    17.0    62.6
+        berkeley    25.0    77.0
```
datapythonista: May be we could manually create a `Series` with Kelvin degrees and assign it. I personally find it a bit confusing that both examples do the same thing, but in a different way. Or we should say it if you think this is better.

I think a single example:
```
>>> df.assign(temp_f=lambda...
>>>                 temp_k=kelvin_series)
```

could illustrate both cases in a simpler way.

pandas/core/frame.py
```diff
@@ -3354,38 +3354,23 @@ def assign(self, **kwargs):
...
 
         Where the keyword arguments depend on each other
```
datapythonista: I think this last example is not useful anymore, or do you think it is?

datapythonista (commented):
Looks quite good, just couple of ideas to make the examples simpler.

pandas/core/frame.py
jschendel: This example is still necessary in some form, as it shows a non-obvious Python 3.6+ feature that was recently added (#18852).  It could definitely be improved though.

Reversing the setup and starting from a DataFrame with Fahrenheit data:
```python
In [2]: df = pd.DataFrame({'temp_f': [-40, 0, 32, 100, 212]})

In [3]: df
Out[3]:
   temp_f
0     -40
1       0
2      32
3     100
4     212
```

Then what this is section is trying to illustrate is that, in Python 3.6+, you can create multiple columns within the same `assign` where one of the columns depends on another one defined within the same `assign`.

For example, if you wanted to add both Celsius and Kelvin columns, you could use a formula to build Celsius from Fahrenheit, then use the Celsius column to create the Kelvin one (instead of using a more complex formula referencing Fahrenheit), all from within the same `assign`:
```python
In [4]: df.assign(temp_c=lambda df: 5 / 9 * (df['temp_f'] - 32), 
                  temp_k=lambda df: df['temp_c'] + 273.15)
Out[4]:
   temp_f      temp_c      temp_k
0     -40  -40.000000  233.150000
1       0  -17.777778  255.372222
2      32    0.000000  273.150000
3     100   37.777778  310.927778
4     212  100.000000  373.150000
```

There have also been some more generic questions about how to add multiple columns using `assign` (e.g. [StackOverflow](https://stackoverflow.com/q/42101382/3339965)).  It might be beneficial to first give an example similar to the one above, but where you create Kelvin from Fahrenheit, to simply show how to add multiple columns within the same `assign`.  Then the 3.6+ behavior could be discussed using something similar to my example above.

pandas/core/frame.py
jschendel: Maybe expanding the exposition could clear this up a bit?  E.g, "Where the value already exists and is inserted:" --> "Alternatively, the same behavior can be achieved by directly referencing an existing Series or list-like:". 

Would be fine with something along the lines of your suggestion too though.

pandas/core/frame.py
aeltanawy: I second your thoughts! Will send an update soon.

pandas/core/frame.py
aeltanawy: I'm leaning more towards keeping this example but to change its exposition to: "Alternatively, the same behavior can be achieved by directly referencing an existing Series or list-like".

pandas/core/frame.py
aeltanawy: Ah! I understand now why this example is needed. Let me add a different one that better portrays its usage.

jreback:
@datapythonista merge when satisfied

aeltanawy:
Here are the recent changes to the DataFrame.assign docstring:

1. Showed the output of the initial dataframe.
2. Re-wrote the third example to illustrate python3.6+ feature of being able to assign multiple columns that depends on each other. The example is derived from @datapythonista and @jschendel  comments.

pandas/core/frame.py
```diff
@@ -3250,48 +3250,34 @@ def assign(self, **kwargs):
 
         Examples
         --------
-        >>> df = pd.DataFrame({'A': range(1, 11), 'B': np.random.randn(10)})
+        >>> df = pd.DataFrame({'temp_c': (17.0, 25.0)},
```
datapythonista: Can you use a list instead of a tuple for the data? I think it's more conventional.

pandas/core/frame.py
```diff
@@ -3250,48 +3250,34 @@ def assign(self, **kwargs):
...
+                                index=['Portland', 'Berkeley'])
```
datapythonista: This is missing `...` at the same level of the `>>>`, and the indentation is not correct

pandas/core/frame.py
```diff
@@ -3250,48 +3250,34 @@ def assign(self, **kwargs):
...
-         2  3  3  6
+        >>> df.assign(temp_f=lambda x: x.temp_c * 9 / 5 + 32)
+                  temp_c  temp_f
+        Portland    17.0    62.6
+        Berkeley    25.0    77.0
+
+        Alternatively, the same behavior can be achieved by directly
+        referencing an existing Series or list-like:
+        >>> newcol = df['temp_c'] * 9 / 5 + 32
+        >>> df.assign(temp_f=newcol)
```
datapythonista: I'd use the expression directly in the assignment instead (i.e. `>>> df.assign(temp_f=df['temp_c'] * 9 / 5 + 32)`),  but it's your call

pandas/core/frame.py
```diff
@@ -3250,48 +3250,34 @@ def assign(self, **kwargs):
...
+                  temp_c  temp_f
+        Portland    17.0    62.6
+        Berkeley    25.0    77.0
+
+        In Python 3.6+, you can create multiple columns within the same assign
+        where one of the columns depends on another one defined within the same
+        assign:
+        >>> df.assign(temp_f=lambda x: x['temp_c'] * 9 / 5 + 32,
+                        temp_k=lambda x: (x['temp_f'] +  459.67) * 5 / 9)
```
datapythonista: same as before regarding `...` and indentation

datapythonista (changes requested):
Thanks for the changes, the example looks great, except for the formatting issues.

Can you run `./scripts/validate_docstrings.py pandas.DataFrame.assign` after you perform the requested changes? that should tell you if all the formatting problems I commented are fixed.

pandas/core/frame.py
aeltanawy: From my understanding, this example is to show that you can refer to an existing list or series and assign it to the df. The example above it is already using direct assignment. Do you think it is not necessary to show this?

datapythonista (commented):
Besides the previous comments, can you remove the `df` in the Returns section (just leave DataFrame in the first line).

And also, add the `**` to the `kwargs` parameter, and change the type, so the first line is `**kwargs : dict of {str: callable or Series}`

Thanks!

aeltanawy:
@datapythonista, I have adjusted DataFrame.assign docstring with your comments. I've also ran ./scripts/validate_docstrings.py pandas.DataFrame.assign till I got 'the Docstring for "pandas.DataFrame.assign" correct. :)'.

Let me know your thoughts on this commit. Thanks!

pandas/core/frame.py
```diff
@@ -3220,7 +3220,7 @@ def assign(self, **kwargs):
 
         Parameters
         ----------
-        kwargs : keyword, value pairs
+        **kwargs : dict of {str: callable or series}
```
datapythonista: can you use `Series` (with capital `S`) please?

pandas/core/frame.py
```diff
@@ -3250,48 +3250,34 @@ def assign(self, **kwargs):
...
-         0  1  1  2
-         1  2  2  4
-         2  3  3  6
+        >>> df.assign(temp_f=lambda x: x.temp_c * 9 / 5 + 32)
+                  temp_c  temp_f
+        Portland    17.0    62.6
+        Berkeley    25.0    77.0
+
+        Alternatively, the same behavior can be achieved by directly
+        referencing an existing Series or list-like:
```
datapythonista: IMO `sequence` would be more accurate than `list-like`. A tuple is a sequence, and could be used, but I wouldn't say it's a list-like (it doesn't implement the API of list).

datapythonista (commented):
lgtm, just two details that if you can change them, it'd be great.

aeltanawy:
Thanks @datapythonista! Adjusted `series` and `list-like` to `Series` and `sequence` respectively.

datapythonista (approved):
lgtm, thanks a lot for the work in this @aeltanawy

aeltanawy:
Thanks a lot @datapythonista! I've learned a lot!

Is there is any next steps I should do for this case, such as closing this pull request?

datapythonista:
can you merge master into your branch and update the PR please? Seems like some checks failed, and this way they'll rerun and we can merge the changes. Other than that, it's in our side.

In case it's not clear what I meant, in your branch do:
`git fetch upstream ; git merge upstream/master ; git push`

aeltanawy:
Do I need to rebase my changes against master?

datapythonista:
It's like the other way round I'd say. You get the latest changes from pandas `git fetch master`, you merge those changes into your branch `git merge upstream/master`, so your branch has your changes on top of the latest pandas, not the pandas of the time you started the changes, and then you update the PR with them `git push`. Besides having your changes on the latest development version of pandas, any update to the PR will make the continuous integration to run again. So the checks that failed (unrelated to your changes), will luckily pass this time.

TomAugspurger:
Git changes look OK.

@aeltanawy could you remove `-assign` from https://github.com/pandas-dev/pandas/blob/0480f4c183a95712cb8ceaf5682c5b8dd02e0f21/ci/doctests.sh#L24? Then the docstring will be tested on each PR.

aeltanawy:
Looks like all checks passed.

@TomAugspurger, I'll create another pull request for removing -assign so as not to confuse this one which is entirely a DOC case. How does this sound?

datapythonista:
@aeltanawy it's better if you do it in this one, as removing the `-assign` will make the CI validate that your examples pass the doctests already.

aeltanawy:
ah! 
removed -assign from pandas/ci/doctests.sh

datapythonista:
Thanks @aeltanawy, merging on green.

jschendel:
Thanks @aeltanawy!

jorisvandenbossche:
@aeltanawy Thanks a lot!

PR #21919, opened 2018-07-15
[BUG][BLD] revert DEF component of #21878

jbrockmendel:
#21878 introduced a subtle build problem that is not caught by the CI.  Running `python setup.py build_ext --inplace` followed by `python3 setup.py build_ext --inplace` causes compile-time errors in py3 (or if running these in the opposite order, errors in py2).  These are fixed by running `python setup.py clean` in between.

This reverts the relevant changes.

For anyone curious: cython supports syntax:

```
DEF foo = [...]

IF foo:
    [...]
ELSE:
    [...]
```

and these IF/ELSE conditions get evaluated at compile-time.  #21878 incorrectly assumed that "compile-time" in this context meant ".c -> .so" time, not ".pyx -> .c" time.  In this we used:

```
DEF PY2 = str == bytes

IF PY2:
    [...]
ELSE:
    [...]
```

so after running `setup.py build_ext --inplace` in py2, the ".c" file we end up with has already gotten rid of the PY3 branches.  When we run `python3 setup.py build_ext --inplace` cython uses the existing .c file, tries to compile it to .so, and breaks.

### DIFF
--- a/pandas/_libs/src/numpy_helper.h
+++ b/pandas/_libs/src/numpy_helper.h
@@ -30,6 +30,23 @@ PANDAS_INLINE PyObject* get_value_1d(PyArrayObject* ap, Py_ssize_t i) {
     return PyArray_Scalar(item, PyArray_DESCR(ap), (PyObject*)ap);
 }
 
+// returns ASCII or UTF8 (py3) view on python str
+// python object owns memory, should not be freed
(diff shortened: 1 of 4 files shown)


jbrockmendel:
@jreback @WillAyd quick take before I merge?

jreback:
lgtm

yeah ideally these would be compile time checks

PR #21922, opened 2018-07-15
Concatenation of series of differing types should lead to object

### LINKED ISSUE #21792: pd.concat of Series with int64 column and Series with int64-ExtensionArray yields int64
#### Code Sample

```python
import fletcher as fr
import pandas as pd

df_ext = pd.DataFrame({'a': fr.FletcherArray([1, 2])})
df_ext.info()
# <class 'pandas.core.frame.DataFrame'>
# RangeIndex: 2 entries, 0 to 1
# Data columns (total 1 columns):
# a    2 non-null fletcher[int64]
# dtypes: fletcher[int64](1)
# memory usage: 100.0 bytes

df_normal = pd.DataFrame({'a': [3, 4]})
df_normal.info()
# <class 'pandas.core.frame.DataFrame'>
# RangeIndex: 2 entries, 0 to 1
# Data columns (total 1 columns):
# a    2 non-null int64
# dtypes: int64(1)
# memory usage: 96.0 bytes

# Works
pd.concat([df_ext, df_normal]).info()
# <class 'pandas.core.frame.DataFrame'>
# Int64Index: 4 entries, 0 to 1
# Data columns (total 1 columns):
# a    4 non-null object
# dtypes: object(1)
# memory usage: 64.0+ bytes

# yield int64 instead of object
pd.concat([df_ext['a'], df_normal['a']]).dtype
# dtype('int64')
```

#### Problem description

This currently leads `BaseReshapingTests.test_concat_mixed_dtypes` to fail on ExtensionArrays that can be converted to any numeric data NumPy datatype.

xhochy:
- [x] closes #21792
- [ ] tests added / passed
- [x] whatsnew entry
 
I see no simple solution on how to make a test for this without introducing a whole new ExtensionArray class that is actually of scalar type `'i'`.

cc @jreback I'm wondering a bit why this has not yet been a problem for https://github.com/pandas-dev/pandas/pull/21160 

I have check that this fixes the problems I'm seeing in `fletcher` on `int` and `float` dtypes.

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -257,6 +257,7 @@ ExtensionType Changes
 - Bug in :meth:`Series.get` for ``Series`` using ``ExtensionArray`` and integer index (:issue:`21257`)
 - :meth:`Series.combine()` works correctly with :class:`~pandas.api.extensions.ExtensionArray` inside of :class:`Series` (:issue:`20825`)
 - :meth:`Series.combine()` with scalar argument now works for any function type (:issue:`21248`)
+- Bug in :func:`concat` that lead to inconsistent behaviour for ExtensionArrays where the scalar representation was int or float (:issue:`21792`)
 -
 
 .. _whatsnew_0240.api.incompatibilities:
--- a/pandas/core/dtypes/concat.py
+++ b/pandas/core/dtypes/concat.py
@@ -180,6 +180,8 @@ def _concat_compat(to_concat, axis=0):
     extensions = [is_extension_array_dtype(x) for x in to_concat]
     if any(extensions) and axis == 1:
         to_concat = [np.atleast_2d(x.astype('object')) for x in to_concat]
+    elif any(extensions):
+        to_concat = [x.astype('object') for x in to_concat]
 
     if not nonempty:
         # we have all empties, but may need to coerce the result dtype to


jreback:
this is not generally the correct solution, the EA should decide what the result should be. we need additional API for this I think in EA.

jreback:
best also to do an example with the test EA class

jorisvandenbossche:
> this is not generally the correct solution, the EA should decide what the result should be. we need additional API for this I think in EA.

As long as we do not have such additional api that handles negotation between different EA types on what the result should be, I think this PR is the good solution.

> best also to do an example with the test EA class

@xhochy mentioned there is no test EA class that can do this (eg with the DecimalArray test case you don't get this bug) 
How much code would it be to have a minimal EA test class to test this bug?

jreback:
> As long as we do not have such additional api that handles negotation between different EA types on what the result should be, I think this PR is the good solution.

and this PR should propose such a solution. otherwise everything is just a band aid.

jorisvandenbossche:
> without introducing a whole new ExtensionArray class that is actually of scalar type 'i'.

The question is maybe also in general if it is needed to have a 'i' kind in your fletcher dtypes (instead of always 'O'). I am not sure to what extent we currently use that information for EAs.

jorisvandenbossche:
> and this PR should propose such a solution. otherwise everything is just a band aid.

Why? We already do exactly the same for dataframes, there we simply always convert to object if they are not matching dtypes. 
I think it is good to already makes this uniform with Series, regardless of whether / when we introduce such a more general solution.

jreback:
> Why? We already do exactly the same for dataframes, there we simply always convert to object if they are not matching dtypes.

not sure what you are talking about, this is the obvious case of having an EA be able to do something intelligent on concat rather than simply coercing. It can always raise not implemented error and then this soln is fine.

In any event this needs tests.

jorisvandenbossche:
> not sure what you are talking about

Did you see the example in the issue? () Concatting a dataframe already works (if converting to object is assumed to be the correct behaviour, which it is at the moment IMO), concatting a series not. This PR is fixing that inconsistency.

jorisvandenbossche:
As I said before, I think this is a reasonable fix. @jreback can respond again (also to my last comment, this is basically doing for Series what already is done for DataFrame. The main difficulty is tests)

jreback:
I am still not clear what this is fixing, this would need a general extension test.

jreback:
closing. this needs tests, if you want to continue pls ping.

Issue #13715, opened 2016-07-20
CI: setup 32-build linux build on travis


jreback:
so if anyone is interested: https://github.com/matthew-brett/multibuild/tree/7abbd01f230be6067842f0510a98b06a6320ddf9

hosts the builder for the wheels, it does a 32-build on linux via a docker image. So we could just directly pull that same image and build (of master).

cc @matthew-brett 

xref #13584

gfyoung:
> Azure does possibly support 32bit builds directly? (to check)

@jorisvandenbossche : My guess is probably not, but couldn't hurt to check.

@datapythonista : Who was your contact at Microsoft regarding Azure builds?

gfyoung:
@datapythonista : Cool, thanks!

@vtbassmatt : Does Azure have support for 32-bit testing environments?

vtbassmatt:
We don't offer hosted 32-bit environments but we do ship a 32-bit Windows
and Linux agent. You can set up private agents with that.

gfyoung:
@vtbassmatt : That's good to know.  Thanks!

jbrockmendel:
we've got a 32 bit build on azure.  closeable?

PR #21923, opened 2018-07-15
[BUG] change types to Py_ssize_t to fix #21905

### LINKED ISSUE #21905: Are tests failing on Windows?
Running the tests on windows fail:

```
>>> pytest pandas\tests\tseries\offsets\test_offsets.py -x
――――――――――――――――――――― TestCommon.test_apply[BusinessDay0] ―――――――――――――――――――――

self = <pandas.tests.tseries.offsets.test_offsets.TestCommon object at 0x066D4530>
offset_types = <class 'pandas.tseries.offsets.BusinessDay'>

    def test_apply(self, offset_types):
        sdt = datetime(2011, 1, 1, 9, 0)
        ndt = np_datetime64_compat('2011-01-01 09:00Z')

        for dt in [sdt, ndt]:
            expected = self.expecteds[offset_types.__name__]
            self._check_offsetfunc_works(offset_types, 'apply', dt, expected)

            expected = Timestamp(expected.date())
            self._check_offsetfunc_works(offset_types, 'apply', dt, expected,
>                                        normalize=True)

pandas\tests\tseries\offsets\test_offsets.py:357:
pandas\tests\tseries\offsets\test_offsets.py:320: in _check_offsetfunc_works
    result = func(dt_tz)
pandas\tseries\offsets.py:85: in wrapper
    result = result.normalize()
pandas\_libs\tslibs\timestamps.pyx:1086: in pandas._libs.tslibs.timestamps.Timestamp.normalize
    normalized_value = normalize_i8_timestamps(
pandas\_libs\tslibs\conversion.pyx:1095: in pandas._libs.tslibs.conversion.normalize_i8_timestamps
    result = _normalize_local(stamps, tz)

>   pos = trans.searchsorted(stamps, side='right') - 1
E   ValueError: Buffer dtype mismatch, expected 'int64_t' but got 'int'

pandas\_libs\tslibs\conversion.pyx:1151: ValueError
```

Several tests fail. 

By going back in commit history until the tests pass, it looks like this is caused by #21826.

jbrockmendel:
May close #21905, will need to check with OP.

### DIFF
--- a/pandas/_libs/tslibs/conversion.pyx
+++ b/pandas/_libs/tslibs/conversion.pyx
@@ -527,7 +527,7 @@ cdef inline void localize_tso(_TSObject obj, tzinfo tz):
     cdef:
         ndarray[int64_t] trans, deltas
         int64_t delta, local_val
-        Py_ssize_t posn
+        Py_ssize_t pos
 
     assert obj.tzinfo is None
 
@@ -782,7 +782,6 @@ def tz_convert(ndarray[int64_t] vals, object tz1, object tz2):
     cdef:
         ndarray[int64_t] utc_dates, tt, result, trans, deltas
         Py_ssize_t i, j, pos, n = len(vals)
-        ndarray[Py_ssize_t] posn
         int64_t v, offset, delta
         npy_datetimestruct dts
 
@@ -1124,7 +1123,8 @@ cdef ndarray[int64_t] _normalize_local(ndarray[int64_t] stamps, object tz):
     cdef:
         Py_ssize_t n = len(stamps)
         ndarray[int64_t] result = np.empty(n, dtype=np.int64)
-        ndarray[int64_t] trans, deltas, pos
+        ndarray[int64_t] trans, deltas
+        Py_ssize_t[:] pos
         npy_datetimestruct dts
 
     if is_utc(tz):
--- a/pandas/_libs/tslibs/period.pyx
+++ b/pandas/_libs/tslibs/period.pyx
@@ -934,7 +934,8 @@ cdef ndarray[int64_t] localize_dt64arr_to_period(ndarray[int64_t] stamps,
     cdef:
         Py_ssize_t n = len(stamps)
         ndarray[int64_t] result = np.empty(n, dtype=np.int64)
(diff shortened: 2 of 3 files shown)


topper-123:
I've run the complete test suite on this. Everything passes beutifully.

gfyoung:
@jbrockmendel : Can we add a `whatsnew` entry for this?

jbrockmendel:
> Can we add a `whatsnew` entry for this?

I don't think so, the relevant bug was only introduced a few days ago.

topper-123:
Is it possible to add 32-bit tests to the testing systems, to guard against similar bugs in the future?

gfyoung:
@topper-123 : I don't see why not.  Feel free to propose some if you have ideas.

jbrockmendel:
@topper-123 might want to discuss in #13715.

@jreback this should be good to go.

topper-123:
uh, I 've never set such things up before, so it will proabbly take more time than I'd like. 

It's very seldom that things like this happen, so maybe it's so rare, that building it into the testing systems will be overkill?

gfyoung:
> uh, I 've never set such things up before, so it will proabbly take more time than I'd like.

It's very seldom that things like this happen, so maybe it's so rare, that building it into the testing systems will be overkill?

@topper-123 : Not overkill but not straightforward to do.  I tried once emulating what `numpy` did (the start of the fun can be found [here](https://github.com/numpy/numpy/blob/master/.travis.yml#L42)), but even that was surprisingly error-prone.

jreback:
32 bit testing happens on another CI on a daily basis

it’s a PITA because it’s not easily testable (eg you can run your PR against it) - this is a limited of travis currently

but haven’t put much effort because we should just drop 32 bit anyways

OTOH datetime changes often force these 32 bugs to the surface

jbrockmendel:
Unless there’s an objection, let’s move the CI discussion to #13715 and push this bug fix through.

jreback:
ok let's see.

topper-123:
I just built the newest master release with this PR and tested it.

Everything built correctly and all tests passed.

PR #21924, opened 2018-07-15
move rename functionality out of internals

### LINKED ISSUE #16045: CLN: move rename functionality out of internals
See comment https://github.com/pandas-dev/pandas/pull/13766#r71974566. Currently the `rename` logic lives in `internals`, cleaner would be to do this on the actual index

jbrockmendel:
- [ ] <s>closes #16045</s><b>update</b>Not anymore

@jorisvandenbossche can you confirm this is what you had in mind in that issue?


pandas/core/generic.py
```diff
@@ -1060,8 +1057,13 @@ def f(x):
             baxis = self._get_block_manager_axis(axis)
             if level is not None:
                 level = self.axes[axis]._get_level_number(level)
-            result._data = result._data.rename_axis(f, axis=baxis, copy=copy,
-                                                    level=level)
+
+            # TODO: we already did a copy above, can we avoid doing again?
```
jreback: I think this makes it more confusing here, reaching into internals to do this.

I would rather move _transform_index and items_overlap_with_suffix from internals to Index.base (or appropriate), these are just renaming / transform helpers for an index.

Then this would be much cleaner.

jreback (changes requested):

pandas/core/generic.py
jbrockmendel: > I think this makes it more confusing here, reaching into internals to do this.

At least add_suffix and add_prefix are clear wins.  I can de-`_data` this a little bit by using `NDFrame.set_axis`.  I'll take a look at transform_index and items_overlap_with_suffix, see if there is a more reasonable place for them.

> I would rather [...]

I agree, am working along similar lines.

jbrockmendel:
@jorisvandenbossche gentle ping; the issue that motivated this PR was yours

jbrockmendel:
@jorisvandenbossche this has been scaled back significantly, but still addresses part of the Issue you opened.  Thoughts?

pandas/core/internals/managers.py
```diff
@@ -169,27 +169,17 @@ def set_axis(self, axis, new_labels):
     def rename_axis(self, mapper, axis, copy=True, level=None):
         """
         Rename one of axes.
-
-        Parameters
+         Parameters
```
jorisvandenbossche: can you undo this change?

pandas/core/internals/managers.py
```diff
@@ -169,27 +169,17 @@ def set_axis(self, axis, new_labels):
...
         ----------
         mapper : unary callable
         axis : int
         copy : boolean, default True
         level : int, default None
-
-        """
+         """
```
jorisvandenbossche: this one as well

jorisvandenbossche (commented):
No objection to this

pandas/core/internals/managers.py
jbrockmendel: Copy/paste snafu, thanks for pointing this out.  Will fix shortly.

jreback (changes requested):
i will have a look soon

jreback (approved):

jreback:
thanks!

PR #21926, opened 2018-07-16
Docstrings, de-duplicate EAMixin/DatetimeLikeIndex __new__ code

jbrockmendel:
There's a lot of duplication in the constructors and constructor-helpers.  This starts to whittle that down, writes some docstrings long the way.

Also use `ExtensionOpsMixin` to define comparison operators on the EAMixin classes.

We determined that the DatetimeIndex._local_timestamps method had an unecessary monotonicy check, so took that out.


pandas/core/arrays/base.py
```diff
@@ -634,6 +634,11 @@ class ExtensionOpsMixin(object):
     """
     A base class for linking the operators to their dunder names
     """
+
+    @classmethod
```
jreback: ideally should add doc-strings & to the top-level docs, as this is API

pandas/core/arrays/datetimelike.py
```diff
@@ -551,3 +557,97 @@ def validate_periods(periods):
...
+
+    Raises
+    ------
+    ValueError : on tzinfo mismatch
+    """
+    if dtype is not None:
+        try:
+            dtype = DatetimeTZDtype.construct_from_string(dtype)
+            dtz = getattr(dtype, 'tz', None)
+            if dtz is not None:
```
jreback: can you do this ?

jreback (changes requested):
looks fine. if you can just update the docs in arrays/base.py when you add api there

pandas/core/arrays/base.py
jbrockmendel: OK.  I think I'll revert this part of the PR as out-of-scope and do the docs bit separately.

pandas/core/arrays/datetimelike.py
```diff
@@ -23,10 +24,13 @@
     is_timedelta64_dtype,
     is_object_dtype)
 from pandas.core.dtypes.generic import ABCSeries, ABCDataFrame, ABCIndexClass
+from pandas.core.dtypes.dtypes import DatetimeTZDtype
 
 import pandas.core.common as com
 from pandas.core.algorithms import checked_add_with_arr
 
+from .base import ExtensionOpsMixin
```
jreback: can you do absolute imports

pandas/core/arrays/datetimelike.py
```diff
@@ -464,7 +468,10 @@ def _addsub_offset_array(self, other, op):
                       "{cls} not vectorized"
                       .format(cls=type(self).__name__), PerformanceWarning)
 
-        res_values = op(self.astype('O').values, np.array(other))
```
jreback: is it worth a branch here in is_extension_dtype?

jreback (changes requested):
minor comments

pandas/core/arrays/datetimelike.py
jbrockmendel: Tried it, causes import-time errors

pandas/core/arrays/datetimelike.py
jbrockmendel: I think `values_from_object` is a pretty clean solution to this.

pandas/core/arrays/datetimelike.py
jreback: i see, we import in __init__ all of the extension arrays which import this. hmm.

pandas/core/arrays/datetimelike.py
jreback: can you update this

jreback (changes requested):

pandas/core/arrays/datetimelike.py
jbrockmendel: Already done

jreback: hmm, does tz_compare not take None as an arg? hmm make should add the check there? as I don't think we check for None's elsewhere (not really sure about this)

jreback (changes requested):

pandas/core/arrays/datetimelike.py
jbrockmendel: None-tz comparisons are a pretty special case that get handled separately.  I’m OK with tzcompare not taking None.

jbrockmendel:
@jreback since we're on a roll I'd like to push this through, get started on the next steps over the weekend.  Thoughts?

jreback (approved):

jreback:
sure. my comments were minor.

PR #21934, opened 2018-07-16
BUG: Properly handle lists for .mask

### LINKED ISSUE #21891: Incorrect assumption in Series.mask()

```python
pd.Series([1, 2]).where([True, False])
# 0    1.0
# 1    NaN
# dtype: float64
pd.Series([1, 2]).mask([True, False])
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
#   File "/usr/local/lib/python3.7/site-packages/pandas/core/generic.py", line 7792, in mask
#     return self.where(~cond, other=other, inplace=inplace, axis=axis,
# TypeError: bad operand type for unary ~: 'list'
```
#### Problem description

With the same `cond`, `Series.where(cond)` succeeded but `Series.mask(cond)` failed.
It's because  `Series.mask()` assumes `cond` has `__invert__()` while it's not always ture.

https://github.com/pandas-dev/pandas/blob/365eac4d27353693a2b758c87fc4bceabb74817f/pandas/core/generic.py#L7944-L7946
#### Expected Output
`Series.mask(cond)` succeeds.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.7.0.final.0
python-bits: 64
OS: Darwin
OS-release: 17.6.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: zh_CN.UTF-8
LOCALE: zh_CN.UTF-8

pandas: 0.23.3
pytest: None
pip: 10.0.1
setuptools: 39.2.0
Cython: None
numpy: 1.14.5
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: None
sphinx: None
patsy: None
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: 4.6.0
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: None
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

gfyoung:
Title is self-explanatory.

Closes #21891.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -7941,6 +7941,10 @@ class NDFrame(PandasObject, SelectionMixin):
         inplace = validate_bool_kwarg(inplace, 'inplace')
         cond = com._apply_if_callable(cond, self)
 
+        # see gh-21891
+        if not hasattr(cond, "__invert__"):
+            cond = np.array(cond)
+
         return self.where(~cond, other=other, inplace=inplace, axis=axis,
                           level=level, try_cast=try_cast,
                           errors=errors)
--- a/pandas/tests/frame/test_indexing.py
+++ b/pandas/tests/frame/test_indexing.py
@@ -2966,6 +2966,13 @@ class TestDataFrameIndexing(TestData):
         assert_frame_equal(rs, df.mask(df <= 0, other))
         assert_frame_equal(rs, df.mask(~cond, other))
 
+        # see gh-21891
+        df = DataFrame([1, 2])
+        res = df.mask([[True], [False]])
+
+        exp = DataFrame([np.nan, 2])
+        tm.assert_frame_equal(res, exp)
+
     def test_mask_inplace(self):
         # GH8801
         df = DataFrame(np.random.randn(5, 3))
--- a/pandas/tests/series/indexing/test_boolean.py
+++ b/pandas/tests/series/indexing/test_boolean.py
@@ -617,6 +617,13 @@ def test_mask():
     expected = Series([1, 2, np.nan, np.nan])
     assert_series_equal(result, expected)
 
+    # see gh-21891
+    s = Series([1, 2])
+    res = s.mask([True, False])
+
+    exp = Series([np.nan, 2])
+    tm.assert_series_equal(res, exp)
+
 
 def test_mask_inplace():
     s = Series(np.random.randn(5))
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -160,7 +160,7 @@ For situations where you need an ``ndarray`` of ``Interval`` objects, use
 :meth:`numpy.asarray` or ``idx.astype(object)``.
 
 .. ipython:: python
-  
+
    np.asarray(idx)
    idx.values.astype(object)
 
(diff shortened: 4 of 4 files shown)


pandas/core/generic.py
```diff
@@ -7941,6 +7941,10 @@ def mask(self, cond, other=np.nan, inplace=False, axis=None, level=None,
         inplace = validate_bool_kwarg(inplace, 'inplace')
         cond = com._apply_if_callable(cond, self)
 
+        # see gh-21891
```
jreback: I would do this in ``.where`` itself

pandas/tests/frame/test_indexing.py
```diff
@@ -3010,6 +3010,14 @@ def test_mask_callable(self):
         tm.assert_frame_equal(result,
                               (df + 2).mask((df + 2) > 8, (df + 2) + 10))
 
+    def test_mask_list(self):
```
jreback: can you pas a bunch of list-likes in a fixture (the old box argument)

pandas/tests/series/indexing/test_boolean.py
```diff
@@ -582,6 +582,15 @@ def test_where_dt_tz_values(tz_naive_fixture):
     assert_series_equal(exp, result)
 
 
+def test_where_list():
```
jreback: same

jreback: can you test where as well with array-likes for the cond

jreback (changes requested):

pandas/core/generic.py
gfyoung: Do what exactly in `.where` ?

jreback: oh i c we are using ``~`` here, ok then.

pandas/tests/series/indexing/test_boolean.py
gfyoung: > can you test where as well with array-likes for the cond

That's tested pretty heavily for `DataFrame` and `Series` already in fact.

pandas/tests/frame/test_indexing.py
gfyoung: Old box argument?  Sorry, I'm not sure what you mean here (or I forgot what it meant if you have mentioned it to me before).

pandas/tests/series/indexing/test_boolean.py
jreback: ok can u integrate this test in a related one then?

gfyoung: I thought about that, but I was more comfortable leaving it as a separate test semantically speaking.  After all, it is more atomic that way IMO.

jreback: right but it is not clear not where things are tested. we are apparently testing list-likes (ex list) somewhere else, so this is confusing to a new reader and is just future technical debt.

gfyoung: Fair enough.  Integrated.

pandas/core/generic.py
holymonson: `cond ` could be a `callable `, this patch seems wouldn't work.

holymonson (changes requested):

pandas/core/generic.py
gfyoung: Yes, it would.  Because `com._apply_if_callable` (right above) returns an `ndarray`.

jreback (approved):

jreback:
thanks @gfyoung

PR #21941, opened 2018-07-17
Add option for alternative correlation types to `DataFrame.corrwith()`

### LINKED ISSUE #21925: Allow different methods of correlation when using corrwith
Hi,

The `corr` method for DataFrames is exceedingly useful, especially since one can specify methods other than the Pearson correlation.  I think it would be nice if this same argument were available when using `corrwith`.  Perhaps there is another workaround, but right now if one wants to calculate all the Spearman (say) correlations with a specific variable, it seems necessary to calculate the full Spearman correlation matrix and subset to the column you care about.

Thanks.

samwaterbury:
- [x] closes #21925 

This adds a `method` argument to `DataFrame.corrwith()` which calls the corresponding correlation function which already exists within pandas. The default option "pearson" was originally implemented directly in this method, I changed it to match the other options for consistency (and a slight speed improvement).

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -6891,7 +6891,7 @@ class DataFrame(NDFrame):
 
         return self._constructor(baseCov, index=idx, columns=cols)
 
-    def corrwith(self, other, axis=0, drop=False):
+    def corrwith(self, other, axis=0, method='pearson', drop=False):
         """
         Compute pairwise correlation between rows or columns of two DataFrame
         objects.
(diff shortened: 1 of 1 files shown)


pandas/core/frame.py
```diff
@@ -6891,7 +6891,7 @@ def cov(self, min_periods=None):
 
         return self._constructor(baseCov, index=idx, columns=cols)
 
-    def corrwith(self, other, axis=0, drop=False):
+    def corrwith(self, other, axis=0, method='pearson', drop=False):
```
gfyoung: Hmm...it would be safest to add it to the end of the signature, as this modification as it stands is not backwards compatible.  Alternatively, you could issue a `DeprecationWarning` if a boolean is passed to `method` and re-assign it to `drop`.

mroeschke:
> The default option "pearson" was originally implemented directly in this method, I changed it to match the other options for consistency (and a slight speed improvement)

Could you demonstrate this speed improvement with the ASV? I am a little dubious that the iterative method is more performant than the existing implementation.

PR #21943, opened 2018-07-17
apply_frame_axis0 applies function to first group twice

### LINKED ISSUE #21609: df.groupby(.).plot.scatter() creates a spurious initial plot

```python
df = pd.DataFrame([[1, 2], [3, 4], [5, 6], [7, 8]], columns=['x', 'y'])
df['cat'] = [1, 1, 1, 1]
df.groupby('cat').plot.scatter(x='x', y='y')
```

#### Problem description

The above creates 2 plots rather than one (notice there is only one category).

Tested in Jupyter 5.4.0 with ``%matplotlib inline``.

#### Expected Output

A single plot.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.5.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.9.0-6-amd64
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: it_IT.UTF-8
LOCALE: it_IT.UTF-8

pandas: 0.24.0.dev0+141.gf1ffc5fae
pytest: 3.5.0
pip: 9.0.1
setuptools: 39.2.0
Cython: 0.25.2
numpy: 1.14.3
scipy: 0.19.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: 1.5.6
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.2.0dev
tables: 3.3.0
numexpr: 2.6.1
feather: 0.3.1
matplotlib: 2.2.2.post1153+gff6786446
openpyxl: 2.3.0
xlrd: 1.0.0
xlwt: 1.3.0
xlsxwriter: 0.9.6
lxml: 4.1.1
bs4: 4.5.3
html5lib: 0.999999999
sqlalchemy: 1.0.15
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: 0.2.1

</details>

javadnoorb:
- [x] closes #21609
- [x] tests added / passed
- [x] whatsnew entry


javadnoorb:
Sorry for the git history mess

jbrockmendel:
Don’t worry about the git history.  We’ll definitely need a test though.

javadnoorb:
It passes pytest locally, but has failed travis/appveyor. As of the test it's not clear to me how to do it. The extra plot does not produce a handle (3 plots lead to 2 axes objects), so there's no straightforward way to count it. Basically the function  gets called, but no trace of it is in the output of `apply_frame_axis0`/`fast_apply`. Does using some sort of static flags and incrementing them in cython seem like a good idea? Or possibly a private flag in `fast_apply`? Any suggestions?

jbrockmendel:
> It passes pytest locally, but has failed travis/appveyor

Did you run the test suite in both py2 and py3?  Linux or OSX or Windows?

Sometimes there are Travis failures unrelated to the PR, which are often solved by just re-pushing to make Travis run again.

Plots in general are tough to test.  If you can get a test working using pytest-mpl that could be _really_ helpful.

javadnoorb:
It was in py3 OSX. I'm repushing to see if that works. 

I'll try to come up with a test for this in the next couple of days.

javadnoorb:
I wrote a test that checks the number of runs for an arbitrary function (np.mean). It may be possible to write something similar for plots with class inheritance from pd.core.frame.DataFrame or instead use the suggested pytest-mpl. But given that this test already captures the issue in a more general sense, do we need an explicit test for scatter plots?

I'm still not sure why so many tests keep failing, but there seem to exist lines such as the following which failed by pytest:
https://github.com/pandas-dev/pandas/blob/de664e82086dc147ac63a09848f0b2f13befde5f/pandas/tests/groupby/test_groupby.py#L1398

I suspect there may be more cases like that. It seems to me that these need to be analyzed one by one and removed if the PR is expected go through. Should I go ahead and remove the ones that seem to be conflicting? I'm worried this may have too many unwanted consequences.

pandas/_libs/reduction.pyx
```diff
@@ -518,6 +519,7 @@ def apply_frame_axis0(object frame, object f, object names,
         except:
             raise InvalidApply('Let this error raise above us')
 
+    piece = result
```
toobaz: The ``try...except`` above is based on the assumption that ``result = f(chunk)`` can fail. I have no idea of when/why it does, but if it does, ``result`` is not defined and this line fails.

pandas/tests/groupby/test_groupby.py
```diff
@@ -23,6 +23,38 @@
...
 import pandas as pd
 
+from pandas._libs import reduction
+
+
+def demean(arr):
+    return arr - arr.mean()
+
+
+def test_apply_frame_axis0_runs_each_group_once(skip_first_index=True):
```
toobaz: What is the role of a named parameter in a test?

toobaz: (please add a reference to this bug)

pandas/_libs/reduction.pyx
```diff
@@ -494,7 +494,8 @@ class InvalidApply(Exception):
 
 
 def apply_frame_axis0(object frame, object f, object names,
-                      ndarray[int64_t] starts, ndarray[int64_t] ends):
+                      ndarray[int64_t] starts, ndarray[int64_t] ends,
+                      skip_first_index = True):
```
toobaz: Would you please add a docstring, or at least clarify the role of ``skip_first_index``? (I can't get it)

pandas/_libs/reduction.pyx
```diff
@@ -529,13 +531,14 @@ def apply_frame_axis0(object frame, object f, object names,
             item_cache.clear()  # ugh
 
             object.__setattr__(slider.dummy, 'name', names[i])
-            piece = f(slider.dummy)
+            if i >= int(skip_first_index):
```
toobaz: I would add a comment (here or above) about what you are doing, and a reference to the issue.

pandas/tests/groupby/test_groupby.py
```diff
@@ -23,6 +23,38 @@
...
+    class CountCalls:
+        def __init__(self):
+            self.counter = 0
+
+        def f(self, x):
+            self.counter += 1
+            return demean(x)
```
toobaz: What about

```
from itertools import count
f = count()
...
assert f() = 3
```
?

toobaz (changes requested):
Some comments. I think I liked the previous version better... maybe because I don't understand the use for ``skip_first_index=False``

pandas/_libs/reduction.pyx
javadnoorb: sorry. `skip_first_index` was to do debugging quicker with cython. `skip_first_index = True` is with the changes. `skip_first_index = False` is without. I'll remove it and make a commit.

pandas/_libs/reduction.pyx
javadnoorb: I assumed if exception is raised, program terminates and wouldn't reach this line. To be safe I'll move it inside the try clause, or maybe should rename result to piece

pandas/tests/groupby/test_groupby.py
javadnoorb: Good idea. I'm not familiar with using count in this way without `__next__` method. Could you explain a bit more?

I tried the following, and it works:

```python
    from itertools import count
    cc = count()
    f = lambda x: cc.__next__()  # need a function with at least one input, otherwise there will be an error
    ...
    assert f(None) == 2
```

pandas/_libs/reduction.pyx
toobaz: OK, no problem

pandas/_libs/reduction.pyx
toobaz: Sorry, you are absolutely right. What I should have written is: this will fail if ``n=0``. Then I have no idea if it can happen, but the code that follows doesn't fail.

pandas/tests/groupby/test_groupby.py
toobaz: Sorry, my mistake again.

I think your solution only works with python 3.*, while
```
f = lambda x : next(cc)
```
should also work in 2.7. And yes, it will be ``==2``, not ``==3``. Maybe ``assert next(cc) == 2`` is slightly cleaner though.

TomAugspurger: In the past, pytest's assertion rewriting had trouble with statements that have side-effects like this. I'd recommend

```python
result = next(cc)
assert result == 2
```

pandas/_libs/reduction.pyx
jreback: not sure why you need to add a parameter here? fast_apply (which calls this) is the first group operation, so by definition this is never called twice. The following code in pandas/core/groupby/ops.py *can* call the groups again, but that is only if the fast path has failed.

jreback (changes requested):

javadnoorb:
@jreback, sorry, that parameter was just for debugging. I'll remove in the next commit. Some tests related in tests/groupby fail. I'll make a commit as soon as I find out the reason. 

Regarding your comment, I suspect you're referring to this line:
https://github.com/pandas-dev/pandas/blob/1f6ddc4592728c1a09fc0a81ca2a00c685b60a8e/pandas/core/groupby/ops.py#L181

But the two calls to f(group #0) don't happen in fast_apply. They happen inside apply_frame_axis0:
First here:
https://github.com/pandas-dev/pandas/blob/1f6ddc4592728c1a09fc0a81ca2a00c685b60a8e/pandas/_libs/reduction.pyx#L515
and then here:
https://github.com/pandas-dev/pandas/blob/1f6ddc4592728c1a09fc0a81ca2a00c685b60a8e/pandas/_libs/reduction.pyx#L532

javadnoorb:
I think PR name was confusing. I renamed it.

pandas/_libs/reduction.pyx
```diff
@@ -529,16 +529,20 @@ def apply_frame_axis0(object frame, object f, object names,
             item_cache.clear()  # ugh
 
             object.__setattr__(slider.dummy, 'name', names[i])
-            piece = f(slider.dummy)
+
+            # issue #21609. Skipping the first group to avoid running
+            # the functions twice. For plots this leads to an extra
+            # plot of the first group.
+            if (i > 0) | (hasattr(piece, 'index')):
```
javadnoorb: Catching mutation status of the first group turned out to be a headache. The way I have written this right now it only avoids double-runs of first group if function `f` does not output a dataframes/etc (which have index attribute), which is the case for plots.

pandas/tests/groupby/test_groupby.py
```diff
@@ -1385,7 +1408,7 @@ def foo(x):
 
 
 def test_group_name_available_in_inference_pass():
-    # gh-15062
+    # gh-15062, issue #21609
```
javadnoorb: If all double-runs (and not just the ones with output `index` attribute) were fixed, this test wouldn't have passed. So I included the issue number.

pandas/_libs/reduction.pyx
```diff
@@ -529,16 +529,20 @@ def apply_frame_axis0(object frame, object f, object names,
...
+                piece = f(slider.dummy)
 
             # I'm paying the price for index-sharing, ugh
-            try:
+
+            if hasattr(piece, 'index'):
                 if piece.index is slider.dummy.index:
```
javadnoorb: Changed `try/except` to more explicit `if`. No difference otherwise.

javadnoorb:
The pytest-mpl test works locally (on a separate script file). How/where should I add it? Where does the baseline file go?

jbrockmendel:
I may have jumped the gun by suggesting pytest-mpl.  If you have another way to test the relevant behavior, go for it.

javadnoorb:
Ok, I'm gonna comment out the pytest-mpl for now to make sure the other tests pass, and wait for feedback on it. Catching the extra graph is not easy without manipulating plot functions (adding counters) so I actually don't think the test adds much beyond the one I've already written in tests/groupby.

pandas/tests/plotting/test_groupby.py
```diff
@@ -11,6 +11,40 @@
...
+#     f = mlplt.plot
+#     names = pd.Int64Index([1, 2], dtype='int64', name='cat')
+#     starts = np.array([0, 3])
+#     ends = np.array([3, 5])
+#
+#     results, mutated = reduction.apply_frame_axis0(
+#         sdata, f, names, starts, ends)
+#
+#     return results[0]
+
```
javadnoorb: Commenting this out until we figure if it's needed and/or where to put it.

javadnoorb:
For the record, after the changes, running the following in Jupyter lab:

```python
%matplotlib inline
import pandas as pd
df = pd.DataFrame([[1, 2], [3, 4], [5, 6], [7, 8]], columns=['x', 'y'])
df['cat'] = [1, 1, 1, 1]
df.groupby('cat').plot.scatter(x='x', y='y')
```

I get:

image

without the changes the output is:

image

javadnoorb:
One Appveyor job keeps getting canceled  without a log. I see other tasks there with similar status. Does anyone know if this is a server issue or from my code?

TomAugspurger:
I don't think we need pytest-mpl.

I'm a bit unsure about the approach here. I'd prefer to see a proper API for groupby that avoids the need to infer the output type, which the groupby plots can use.

javadnoorb:
@TomAugspurger Do you mean to skip `apply_frame_axis0 ` (or `fast_apply`) and instead take the slow path and loop through the groups when plots are passed as function?

TomAugspurger:
I'm not sure exactly what the API would be, but the goal is to give callers the ability to disable the "apply called twice" behavior. But this is a larger issue.

Since these groupby methods are internal, we have the ability to hardcode which functions aren't called twice. We do this already for things like `.hist`, `.boxplot`.

The following diff shows an example.

```diff
diff --git a/pandas/core/groupby/base.py b/pandas/core/groupby/base.py
index b2c5a8cff..65e6bdf35 100644
--- a/pandas/core/groupby/base.py
+++ b/pandas/core/groupby/base.py
@@ -59,6 +59,7 @@ class GroupByMixin(object):
 # special case to prevent duplicate plots when catching exceptions when
 # forwarding methods from NDFrames
 plotting_methods = frozenset(['plot', 'boxplot', 'hist'])
+extra_plotting_methods = frozenset(['scatter'])
 
 common_apply_whitelist = frozenset([
     'last', 'first',
diff --git a/pandas/core/groupby/groupby.py b/pandas/core/groupby/groupby.py
index 4b0143b3e..de59673dc 100644
--- a/pandas/core/groupby/groupby.py
+++ b/pandas/core/groupby/groupby.py
@@ -311,6 +311,7 @@ class GroupByPlot(PandasObject):
         def attr(*args, **kwargs):
             def f(self):
                 return getattr(self.plot, name)(*args, **kwargs)
+            f.__name__ = name
             return self._groupby.apply(f)
         return attr
 
diff --git a/pandas/core/groupby/ops.py b/pandas/core/groupby/ops.py
index f2c55a56b..1ba80cbf9 100644
--- a/pandas/core/groupby/ops.py
+++ b/pandas/core/groupby/ops.py
@@ -177,6 +177,7 @@ class BaseGrouper(object):
         # oh boy
         f_name = com._get_callable_name(f)
         if (f_name not in base.plotting_methods and
+                f_name not in base.extra_plotting_methods and
                 hasattr(splitter, 'fast_apply') and axis == 0):
             try:
                 values, mutated = splitter.fast_apply(f, group_keys)
```

Not claiming that's the perfect solution, but I think it's a bit cleaner than changing `apply_frame_axis0`

javadnoorb:
Thanks, didn't realize this was already there

javadnoorb:
@TomAugspurger Since you already have this written up do you want to push your version?

TomAugspurger:
FYI, you can apply diffs with `git apply`

```python
git apply foo.patch
```

You'll want to include other plot methods than scatter, revert the changes to `apply_frame_axis_0`, update the tests, etc.

javadnoorb:
Ok, thanks.

javadnoorb:
@TomAugspurger your patch (plus added plot names) fixes all plot functions, when I try:
```python
%matplotlib inline
import pandas as pd
df = pd.DataFrame([[1, 2], [3, 4], [5, 6], [7, 8]], columns=['x', 'y'])
df['cat'] = [1, 1, 1, 1]

extra_plotting_methods = ['line', 'bar', 'barh', 'box', 'kde', 'density', 'area', 'pie', 'scatter', 'hexbin']
p = df.groupby('cat').plot

for plotting_method in extra_plotting_methods:
    ax = p.__getattr__(plotting_method)(x='x', y='y');
    ax.iloc[0].set_title(plotting_method)
```

Still not sure how to implement a test though.

pandas/tests/plotting/test_groupby.py
```diff
@@ -11,6 +11,47 @@
...
+        def __init__(self, groupby):
+            super(extGroupByPlot, self).__init__(groupby)
+
+        def __getattr__(self, name):
+            def attr(*args, **kwargs):
+                counter_obj = count()
+
+                def f(self):
+                    getattr(self.plot, name)(*args, **kwargs)
+                    # I have no idea why this is needed twice
```
javadnoorb: RESOLVED: Function was returning next(counter_obj) when I wrote it.

TomAugspurger:
@javadnoorb looks like there are unrelated commits on this PR now, do you need help with git stuff? https://pandas.pydata.org/pandas-docs/stable/contributing.html#working-with-the-code might be helpful as well.

pandas/tests/plotting/test_groupby.py
```diff
@@ -11,6 +11,45 @@
...
+@pytest.mark.parametrize(
+    'plotting_method',
+    ['line', 'bar', 'barh', 'box',
+     'area', 'pie', 'scatter', 'hexbin',
+     pytest.param('density', marks=td.skip_if_no_scipy),
+     pytest.param('kde', marks=td.skip_if_no_scipy)
+     ])
+@td.skip_if_no_mpl
+def test_no_double_plot_for_first_group(plotting_method):
+    class extGroupByPlot(pd.core.groupby.groupby.GroupByPlot):
```
TomAugspurger: I think this test can be simplified.

How about

```python
df = pd.DataFrame({'cat': [1, 1, 2, 2], 'x': [1, 3, 5, 7], 'y': [2, 4, 6, 8]})
df.groupby('cat').plot.scatter(x='x', y='y')
assert plt.get_fignums() == [1, 2]
```

or something like that.

TomAugspurger: You'd need to replace `.plot.scatter` with your `getattr(gr, plotting_method)(x='x', y='y')`, but the basic idea is to test the output, the number of produced figures, is the number of groups.

javadnoorb:
@TomAugspurger yeah. I messed up while syncing my fork. There were too many conflicts after the merge and I manually removed mine against the master, and some newlines were left here and there. Was I supposed to do this manually?

pandas/tests/plotting/test_groupby.py
javadnoorb: That looks much better. I didn't know about `get_fignums`.

javadnoorb: Do you by any chance know how to reset this? It works for the first plot method, but then keep accumulating as `pytest.mark.parametrize` is looped through.

javadnoorb: Run `plt.close()` twice?

TomAugspurger: Each test should close all its state. `plt.close('all')` should do it.

javadnoorb:
@jreback @toobaz @TomAugspurger. Just leaving a note. This should be ready for review. -thanks

TomAugspurger (commented):
It looks like there are two uses of `pandas.core.groupby.base.plotting_methods`.

1. core/groupby/ops.py, which you've updated
2. core/rgoupby/groupby.py in `_make_wrapper`, which hasn't been updated.

Can you try just adding these new methods to `plotting_methods` and seeing if anything breaks? Ideally we would just have the one frozenset. I initially added `extra_plotting_methods` just to test things out.

javadnoorb:
@TomAugspurger, I don't think `plotting_methods` and `extra_plotting_methods` should be merged. It leads to some errors like:

`AttributeError: type object 'Series' has no attribute 'hexbin'`

`plotting_methods` affects:

https://github.com/pandas-dev/pandas/blob/26b3e7de56fdebaf1842b01d747a1a30a126c54a/pandas/core/groupby/base.py#L80

and 

https://github.com/pandas-dev/pandas/blob/26b3e7de56fdebaf1842b01d747a1a30a126c54a/pandas/core/groupby/base.py#L86

which are then used as methods for dataframes and series.

I think only functions used in this doc can be methods of `groupby.plot`:

http://pandas.pydata.org/pandas-docs/version/0.16.2/generated/pandas.core.groupby.DataFrameGroupBy.plot.html

`hist` is the only exception because it has dual functionality (both for series/dataframes and groups). But I think it should be sufficient to have it in `plotting_methods`.

Following the other point, adding `extra_plotting_methods` to core/goupby/groupby.py:`_make_wrapper` doesn't seem to do anything when I try it. These `curried_with_axis` functions are pretty old and have no issue number. I can't track down why they were written, but seems like they just explicitly add an axis to the plot if none is provided.

TomAugspurger:
Thanks for the clarification on merging those two sets. Is the make_wrapper code dead? Does removing it break any tests? I’ve never touched them.

javadnoorb:
No, it's not dead code. Just that it doesn't run when `groupby.plot.apply` is called. Commenting it out causes some tests to fail:
```
___________________________________________________________________________________________________ TestDataFrameGroupByPlots.test_series_groupby_plotting_nominally_works ___________________________________________________________________________________________________

self = <pandas.tests.plotting.test_groupby.TestDataFrameGroupByPlots object at 0x1a1fb72278>

    def test_series_groupby_plotting_nominally_works(self):
        n = 10
        weight = Series(np.random.normal(166, 20, size=n))
        height = Series(np.random.normal(60, 10, size=n))
        with tm.RNGContext(42):
            gender = np.random.choice(['male', 'female'], size=n)
    
        weight.groupby(gender).plot()
        tm.close()
>       height.groupby(gender).hist()

pandas/tests/plotting/test_groupby.py:49: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pandas.core.groupby.generic.SeriesGroupBy object at 0x1a1fc10d30>, attr = 'hist'

    def __getattr__(self, attr):
        if attr in self._internal_names_set:
            return object.__getattribute__(self, attr)
        if attr in self.obj:
            return self[attr]
        # if hasattr(self.obj, attr):
        #     return self._make_wrapper(attr)
    
        raise AttributeError("%r object has no attribute %r" %
>                            (type(self).__name__, attr))
E       AttributeError: 'SeriesGroupBy' object has no attribute 'hist'

pandas/core/groupby/groupby.py:537: AttributeError
```

jreback:
closing as stale. if you'd like to continue working, pls ping.

Issue #18894, opened 2017-12-21
DEPS: warning about dropping python 2.7 at end of 2018


jreback:
xref #18241

We should add a big note on the top of each whatsnew that we are planning on dropping 2.7 support as of the end of 2018.

topper-123:
+1

I think this note should put in ``install.rst`` as well (under "Python version support").

PR #21944, opened 2018-07-17
DOC add Python 2.7 warning to recent whatsnew; include 23.3

h-vetinari:
#18894 laid out:
> We should add a big note on the top of each whatsnew that we are planning on dropping 2.7 support as of the end of 2018.

So far, this warning only appeared in v0.23.0, and can be easily missed these days. This PR adds them to the recent whatsnews, and also includes v0.23.3 in `whatsnew.rst`


jreback:
thanks

PR #21947, opened 2018-07-17
BUG: fix df.where(cond) when cond is empty

pajachiet:
- when cond is empty, cond.dtypes are objects, which raised
`ValueError: Boolean array expected for the condition, not object
`
- [ ] closes #xxxxx 
- [ ] tests added / passed
- [ ] whatsnew entry


gfyoung:
@pajachiet : Thanks for the PR!  Can you add a test and a `whatsnew` entry in `0.24.0.txt` as well?

jreback:
how is this not an error?

if you don’t have a conf then where doesn’t make sure

gfyoung:
> how is this not an error?

@jreback : That's what I was wondering too myself.  Perhaps a better error message is in order?

pajachiet:
Hello

I wanted to write a test, but I did not find where were the unit tests for the 'where' function.

This is indeed a corner case, when both df and cond are empty. Example: 
```
import pandas as pd
data = [1]
df = pd.DataFrame(columns=['a'], data=data)
cond = df.applymap(lambda x: x > 0)
df.where(cond)  # Ok

data = []
df = pd.DataFrame(columns=['a'], data=data)
cond = df.applymap(lambda x: x > 0)
df.where(cond)  # Error
```

This kind of corner case do happen : 
- when you do unit testing 
- when your dataframe comes from real 'empty' data.

I had the same kind of corner case with df.duplicated in previous version of pandas.

jreback:
yeah certainly could have a better error message (e.g. you can detect the .empty), but this should just be an error.

jreback:
the ``df`` empty is ok, but the cond cannot be empty (unless the df itself is empty).

pajachiet:
Yes, `cond` can be empty if and only if `df` is empty. 

But it is a bug to raise an error when they are both empty, no ?

As the bug is on `cond`'s dtype checking, I thought the easiest way to solve it was to avoid dtypes checking on empty dataframe (as it does not makes sense anyway).

One could also explicitly check for `df.empty`, but this would be weird as it already works when `cond` is empty with correct dtypes : 

```
data = []
df = pd.DataFrame(columns=['a'], data=data)
cond = df > 0 
df.where(cond)  # Ok
cond.dtypes  # bool
```

jreback (changes requested):
needs a test

add in pandas/tests/frame/test_indexing (you may need one for series, though I don't know if this path is hit, in pandas/tests/series/tests_indexing)

pajachiet:
Hello,
I added a simple test as requested. 
I changed the commit two times because of pep8 mistake, then trying to fix the build, but it is still failing for a reason I do not understand (it does not seem related to the change).
 Could you help fixing it ?

pandas/tests/frame/test_indexing.py
```diff
@@ -2939,6 +2939,15 @@ def test_where_callable(self):
...
                               (df + 2).where((df + 2) > 8, (df + 2) + 10))
 
+    def test_where_empty_df_and_empty_cond_having_non_bool_dtypes(self):
+        # GH 21947
+        data = []
+        df = pd.DataFrame(columns=['a'], data=data)
+        cond = df.applymap(lambda x: x > 0)
+        # Should not raise ValueError:
+        # Boolean array expected for the condition, not object
+        df.where(cond)
```
jreback: can you check the result and assert e.g.

```
result = df.where(cond)
expected = ....
tm.assert_frame_equal(...)
```

pandas/tests/frame/test_indexing.py
```diff
@@ -2939,6 +2939,15 @@ def test_where_callable(self):
         tm.assert_frame_equal(result,
                               (df + 2).where((df + 2) > 8, (df + 2) + 10))
 
+    def test_where_empty_df_and_empty_cond_having_non_bool_dtypes(self):
+        # GH 21947
+        data = []
+        df = pd.DataFrame(columns=['a'], data=data)
```
jreback: just construct like
``df = pd.DataFrame(columns=['a'])``

pandas/tests/frame/test_indexing.py
```diff
@@ -2939,6 +2939,15 @@ def test_where_callable(self):
         tm.assert_frame_equal(result,
                               (df + 2).where((df + 2) > 8, (df + 2) + 10))
 
+    def test_where_empty_df_and_empty_cond_having_non_bool_dtypes(self):
```
jreback: can you move this tests near other tests of where which test empty

jreback (changes requested):

jreback (changes requested):
pls add a whatsnew entry in v0.24.0 / bug fixes / reshaping

jreback (changes requested):
use this PR number as the issue number (as I don't think there is an issue about this)

pandas/tests/frame/test_indexing.py
pajachiet: Modification done. 
Even though I initially wrote the test this way on purpose, as I only want to test if an error is raised.

pandas/tests/frame/test_indexing.py
pajachiet: done

pandas/tests/frame/test_indexing.py
pajachiet: Moved close to test_where_none

jreback:

pajachiet:
I _believe_ the bug in the CI was not caused by changes in this PR.

This PR fixed a simple but real bug. I managed to fix it in another way for my client, so I only did this PR if it can improve pandas, especially in those edges cases that can be really annoying. 

I added a test, a whatsnew entry and answered your comments. 
Could you please manage to merge it ? 

Otherwise keep it close, it's too much an investment for such a minor improvment.

gfyoung:
@pajachiet : Do you want to finish this up?  I think it requires a rebase, that's all.

jreback:
> Otherwise keep it close, it's too much an investment for such a minor improvment.

well if it’s a bug then it will probably come up again

it’s always helpful to improve things

gfyoung:
@jreback : Is there a reason why we can't attempt to merge this for `0.24.0` ?  It's green.  Did you have any other comments to add?

gfyoung:
Just saw your email regarding `0.24.0`.  That answers my question [above](https://github.com/pandas-dev/pandas/pull/21947#issuecomment-436126125).

jreback (approved):
minor comment.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -1305,6 +1305,7 @@ Reshaping
 - Bug in :func:`pandas.concat` when joining resampled DataFrames with timezone aware index (:issue:`13783`)
 - Bug in :meth:`Series.combine_first` with ``datetime64[ns, tz]`` dtype which would return tz-naive result (:issue:`21469`)
 - Bug in :meth:`Series.where` and :meth:`DataFrame.where` with ``datetime64[ns, tz]`` dtype (:issue:`21546`)
+- Bug in :meth:`Dataframe.where` with empty ``df`` and empty ``cond`` having non-bool dtype (:issue:`21947`)
```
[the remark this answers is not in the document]
gfyoung: Done.

jreback:
thanks!

pajachiet:
Many thanks ! :)

Issue #7662, opened 2014-07-03
read_csv/to_csv sep/delimiter inconsistency


dsm054:
related #7615

Bit of a UI problem here (although it's behaving as the docstring says it does, so it doesn't quite qualify as a bug):

```
>>> df = pd.DataFrame({"A": [1,2,3], "B": [4,5,6]})
>>> df.to_csv("tmp.csv", sep=";")
>>> !cat tmp.csv
;A;B
0;1;4
1;2;5
2;3;6
>>> df.to_csv("tmp.csv", delimiter=";")
>>> !cat tmp.csv
,A,B
0,1,4
1,2,5
2,3,6
```

`read_csv` accepts both `sep` and `delimiter` but `to_csv` silently ignores `delimiter`.  Someone was recently tripped up by this on SO.  I'm fine with either teaching `to_csv` to behave the same way `read_csv` does or, alternatively, raising if `delimiter` is found as a keyword.  

That is, I'm less bothered by the inconsistency than the silent unexpected behaviour.

indera:
👍   Spent about an hour trying to figure out why passing "delimiter" does not affect the "to_csv" call ...

indera:
Will try to fix this if nobody started.

Relevant code:

to_csv => https://github.com/pandas-dev/pandas/blob/ee374ee3bc49b82754e8269f69cf8f79d488bd5e/pandas/core/frame.py#L1303

read_csv => https://github.com/pandas-dev/pandas/blob/master/pandas/io/parsers.py#L565

gfyoung:
@indera : Just what are you trying to fix exactly?  The `**kw` option has been removed in `to_csv`, so are you trying to add `delimiter` to `to_csv`?

jreback:
@dsm054 @gfyoung is right, this *used* to be a problem, but no-longer.

indera:
Yes, this is fixed in pandas==0.19.0 

I get an error as expected:

TypeError: to_csv() got an unexpected keyword argument 'delimiter'

indera:
@gfyoung - we should probably now remove support for "delimiter" in read_csv for consistency ;)

jorisvandenbossche:
@indera the difference is that for `read_csv` the `delimeter` keyword actually works, so not sure if deprecating/removing is worth it. For `to_csv` it never worked, so that is another situation.

gfyoung:
[...]

PR #21954, opened 2018-07-17
DEPR: pd.read_table

### LINKED ISSUE #21948: DEPR: read_table
Went through #18262  looking for an easy task, and deprecating `read_table` looked like one. Currently, there's a function factory `_make_parser_function` which produces both `read_csv` and `read_table` [here](https://github.com/pandas-dev/pandas/blob/537b65cb0fd2aa318e089c5e38f09e81d1a3fe35/pandas/io/parsers.py#L685-L689):

```python
read_csv = _make_parser_function('read_csv', sep=',')
read_csv = Appender(_read_csv_doc)(read_csv)


read_table = _make_parser_function('read_table', sep='\t')
read_table = Appender(_read_table_doc)(read_table)
```

Seems to me the simplest solution is wrapping `read_table` like this:

```python
def depr_read_table(func):
    """Deprecates read_table."""
    from functools import wraps
    @wraps(read_table)
    def depr_func(*args, **kwargs):
        if len(args) > 1 or "sep" in kwargs or "delimiter" in kwargs:
            warnings.warn("read_table is deprecated, use read_csv instead",
                          FutureWarning, stacklevel=2)
        else:
            warnings.warn("read_table is deprecated, use read_csv with "
                          "sep='\\t' instead",
                          FutureWarning, stacklevel=2)
        return func(*args, **kwargs)
    return depr_func
read_table = depr_read_table(read_table)
```

I would have to find out how to change the docs to declare the function deprecated, but I suppose that is easy...?

Result:

```pycon
In [1]: import pandas as pd

In [2]: df = pd.DataFrame({"a": [1], "b": [2]})

In [3]: df.to_csv("test.csv", sep="\t")

In [4]: pd.read_table("test.csv", index_col=0)
/home/dahlbaek/virtualenvs/pandas-dev/bin/ipython:1: FutureWarning: read_table is deprecated, use read_csv with sep='\t' instead
  #!/home/dahlbaek/virtualenvs/pandas-dev/bin/python
Out[4]: 
   a  b
0  1  2

In [5]: pd.read_table("test.csv", "\t", index_col=0)
/home/dahlbaek/virtualenvs/pandas-dev/bin/ipython:1: FutureWarning: read_table is deprecated, use read_csv instead
  #!/home/dahlbaek/virtualenvs/pandas-dev/bin/python
Out[5]: 
   a  b
0  1  2

In [6]: pd.read_table("test.csv", sep="\t", index_col=0)
/home/dahlbaek/virtualenvs/pandas-dev/bin/ipython:1: FutureWarning: read_table is deprecated, use read_csv instead
  #!/home/dahlbaek/virtualenvs/pandas-dev/bin/python
Out[6]: 
   a  b
0  1  2

In [7]: pd.read_table("test.csv", delimiter="\t", index_col=0)
/home/dahlbaek/virtualenvs/pandas-dev/bin/ipython:1: FutureWarning: read_table is deprecated, use read_csv instead
  #!/home/dahlbaek/virtualenvs/pandas-dev/bin/python
Out[7]: 
   a  b
0  1  2

In [8]: df
Out[8]: 
   a  b
0  1  2
```

Any thoughts?

dahlbaek:
`pd.read_table` is deprecated and replaced by `pd.read_csv`.

- [x] closes #21948
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -478,6 +478,7 @@ Deprecations
 - :meth:`Series.ptp` is deprecated. Use ``numpy.ptp`` instead (:issue:`21614`)
 - :meth:`Series.compress` is deprecated. Use ``Series[condition]`` instead (:issue:`18262`)
 - :meth:`Categorical.from_codes` has deprecated providing float values for the ``codes`` argument. (:issue:`21767`)
+- :func:`pandas.read_table` is deprecated. Use :func:`pandas.read_csv` instead (:issue:`21948`)
 
 .. _whatsnew_0240.prior_deprecations:
 
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -478,7 +478,7 @@ Deprecations
 - :meth:`Series.ptp` is deprecated. Use ``numpy.ptp`` instead (:issue:`21614`)
 - :meth:`Series.compress` is deprecated. Use ``Series[condition]`` instead (:issue:`18262`)
 - :meth:`Categorical.from_codes` has deprecated providing float values for the ``codes`` argument. (:issue:`21767`)
-- :func:`pandas.read_table` is deprecated. Use :func:`pandas.read_csv` instead (:issue:`21948`)
+- :func:`pandas.read_table` is deprecated. Instead, use :func:`pandas.read_csv` passing ``sep='\t'`` if necessary (:issue:`21948`)
 
 .. _whatsnew_0240.prior_deprecations:
 
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -1594,11 +1594,11 @@ class DataFrame(NDFrame):
                       "for from_csv when changing your function calls",
                       FutureWarning, stacklevel=2)
 
-        from pandas.io.parsers import read_table
-        return read_table(path, header=header, sep=sep,
-                          parse_dates=parse_dates, index_col=index_col,
-                          encoding=encoding, tupleize_cols=tupleize_cols,
-                          infer_datetime_format=infer_datetime_format)
+        from pandas.io.parsers import read_csv
+        return read_csv(path, header=header, sep=sep,
+                        parse_dates=parse_dates, index_col=index_col,
+                        encoding=encoding, tupleize_cols=tupleize_cols,
+                        infer_datetime_format=infer_datetime_format)
 
     def to_sparse(self, fill_value=None, kind='block'):
         """
--- a/pandas/io/clipboards.py
+++ b/pandas/io/clipboards.py
@@ -9,7 +9,7 @@ from pandas import compat, get_option, option_context
 
 def read_clipboard(sep=r'\s+', **kwargs):  # pragma: no cover
     r"""
-    Read text from clipboard and pass to read_table. See read_table for the
+    Read text from clipboard and pass to read_csv. See read_csv for the
     full argument list
 
     Parameters
@@ -31,7 +31,7 @@ def read_clipboard(sep=r'\s+', **kwargs):  # pragma: no cover
             'reading from clipboard only supports utf-8 encoding')
 
     from pandas.io.clipboard import clipboard_get
-    from pandas.io.parsers import read_table
+    from pandas.io.parsers import read_csv
     text = clipboard_get()
 
     # try to decode (if needed on PY3)
@@ -51,7 +51,7 @@ def read_clipboard(sep=r'\s+', **kwargs):  # pragma: no cover
     # that this came from excel and set 'sep' accordingly
     lines = text[:10000].split('\n')[:-1][:10]
 
-    # Need to remove leading white space, since read_table
+    # Need to remove leading white space, since read_csv
     # accepts:
     #    a  b
     # 0  1  2
@@ -80,7 +80,7 @@ def read_clipboard(sep=r'\s+', **kwargs):  # pragma: no cover
     if kwargs.get('engine') == 'python' and PY2:
         text = text.encode('utf-8')
 
-    return read_table(StringIO(text), sep=sep, **kwargs)
+    return read_csv(StringIO(text), sep=sep, **kwargs)
 
 
 def to_clipboard(obj, excel=True, sep=None, **kwargs):  # pragma: no cover
--- a/pandas/io/parsers.py
+++ b/pandas/io/parsers.py
@@ -331,6 +331,10 @@ Read CSV (comma-separated) file into DataFrame
 """ % (_parser_params % (_sep_doc.format(default="','"), _engine_doc))
 
 _read_table_doc = """
+
+.. deprecated:: 0.24.0
+   Use :func:`pandas.read_csv` instead, passing `sep='\t'` if necessary.
+
 Read general delimited file into DataFrame
 
 %s
@@ -540,9 +544,13 @@ _deprecated_args = {
 }
 
 
-def _make_parser_function(name, sep=','):
+def _make_parser_function(name, default_sep=','):
 
-    default_sep = sep
+    # prepare read_table deprecation
+    if name == "read_table":
+        sep = False
+    else:
+        sep = default_sep
 
     def parser_f(filepath_or_buffer,
                  sep=sep,
@@ -611,11 +619,24 @@ def _make_parser_function(name, sep=','):
                  memory_map=False,
                  float_precision=None):
 
+        # deprecate read_table GH21948
+        if name == "read_table":
+            if sep is False and delimiter is None:
+                warnings.warn("read_table is deprecated, use read_csv "
+                              "instead, passing sep='\\t'.",
+                              FutureWarning, stacklevel=2)
+            else:
+                warnings.warn("read_table is deprecated, use read_csv "
+                              "instead.",
+                              FutureWarning, stacklevel=2)
+            if sep is False:
+                sep = default_sep
+
         # Alias sep -> delimiter.
         if delimiter is None:
             delimiter = sep
 
-        if delim_whitespace and delimiter is not default_sep:
+        if delim_whitespace and delimiter != default_sep:
             raise ValueError("Specified a delimiter with both sep and"
                              " delim_whitespace=True; you can only"
                              " specify one.")
@@ -687,10 +708,10 @@ def _make_parser_function(name, sep=','):
     return parser_f
 
 
-read_csv = _make_parser_function('read_csv', sep=',')
+read_csv = _make_parser_function('read_csv', default_sep=',')
 read_csv = Appender(_read_csv_doc)(read_csv)
 
-read_table = _make_parser_function('read_table', sep='\t')
+read_table = _make_parser_function('read_table', default_sep='\t')
 read_table = Appender(_read_table_doc)(read_table)
 
 
--- a/pandas/io/parsers.py
+++ b/pandas/io/parsers.py
@@ -333,7 +333,7 @@ Read CSV (comma-separated) file into DataFrame
 _read_table_doc = """
 
 .. deprecated:: 0.24.0
-   Use :func:`pandas.read_csv` instead, passing `sep='\t'` if necessary.
+   Use :func:`pandas.read_csv` instead, passing ``sep='\t'`` if necessary.
 
 Read general delimited file into DataFrame
 
--- a/pandas/tests/io/conftest.py
+++ b/pandas/tests/io/conftest.py
@@ -1,5 +1,5 @@
 import pytest
-from pandas.io.parsers import read_table
+from pandas.io.parsers import read_csv
 
 
 @pytest.fixture
@@ -17,7 +17,7 @@ def jsonl_file(datapath):
 @pytest.fixture
 def salaries_table(datapath):
     """DataFrame with the salaries dataset"""
-    return read_table(datapath('io', 'parser', 'data', 'salaries.csv'))
+    return read_csv(datapath('io', 'parser', 'data', 'salaries.csv'), sep='\t')
 
 
 @pytest.fixture
--- a/pandas/tests/io/parser/test_unsupported.py
+++ b/pandas/tests/io/parser/test_unsupported.py
@@ -14,7 +14,7 @@ import pandas.util.testing as tm
 
 from pandas.compat import StringIO
 from pandas.errors import ParserError
-from pandas.io.parsers import read_csv, read_table
+from pandas.io.parsers import read_csv
 
 import pytest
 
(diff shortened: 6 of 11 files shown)


WillAyd:
Initial comments:

- Needs test(s)
- Since we are replacing `read_table` with `read_csv` and a sep of `\t`you should reroute the call to `read_table` to reflect this
 - Needs whatsnew note

dahlbaek:
Added a whatsnew note.

I'm not sure what to do about tests. `pd.read_csv` has been around for as long as I've been using pandas, so I would suspect that tests are already plentiful.

I changed the warning message so that it will reflect whether or not the user is supplying a value to `sep`. The result is now similar to what I proposed in #21948 but without the decorator, as suggested by @gfyoung. Result:

```pycon
In [1]: import pandas as pd

In [2]: df = pd.DataFrame({"a": [1], "b": [2]})
   ...: df.to_csv("test.csv", sep="\t")

In [3]: pd.read_table("test.csv", index_col=0)
/home/dahlbaek/virtualenvs/pandas-dev/bin/ipython:1: FutureWarning: read_table is deprecated, use read_csv with sep='\t' instead.
  #!/home/dahlbaek/virtualenvs/pandas-dev/bin/python
Out[3]: 
   a  b
0  1  2

In [4]: pd.read_table("test.csv", "\t", index_col=0)
/home/dahlbaek/virtualenvs/pandas-dev/bin/ipython:1: FutureWarning: read_table is deprecated, use read_csv instead.
  #!/home/dahlbaek/virtualenvs/pandas-dev/bin/python
Out[4]: 
   a  b
0  1  2

In [5]: pd.read_table("test.csv", sep="\t", index_col=0)
/home/dahlbaek/virtualenvs/pandas-dev/bin/ipython:1: FutureWarning: read_table is deprecated, use read_csv instead.
  #!/home/dahlbaek/virtualenvs/pandas-dev/bin/python
Out[5]: 
   a  b
0  1  2

In [6]: pd.read_table("test.csv", delimiter="\t", index_col=0)
/home/dahlbaek/virtualenvs/pandas-dev/bin/ipython:1: FutureWarning: read_table is deprecated, use read_csv instead.
  #!/home/dahlbaek/virtualenvs/pandas-dev/bin/python
Out[6]: 
   a  b
0  1  2

In [7]: pd.read_csv("test.csv", "\t", index_col=0)
Out[7]: 
   a  b
0  1  2

In [8]: pd.read_csv("test.csv", sep="\t", index_col=0)
Out[8]: 
   a  b
0  1  2

In [9]: pd.read_csv("test.csv", delimiter="\t", index_col=0)
Out[9]: 
   a  b
0  1  2
```

gfyoung:
We definitely have tests for `read_table` in the codebase.  Modify them to assert that a warning is produced.

gfyoung:
@dahlbaek : Pass in `check_stacklevel=False` in your tests.

dahlbaek:
> @dahlbaek : Pass in check_stacklevel=False in your tests.

Perfect, that's what I figured.

dahlbaek:
@WillAyd @gfyoung Does this seem reasonable?

pandas/tests/io/conftest.py
```diff
@@ -17,7 +18,10 @@ def jsonl_file(datapath):
 @pytest.fixture
 def salaries_table(datapath):
     """DataFrame with the salaries dataset"""
-    return read_table(datapath('io', 'parser', 'data', 'salaries.csv'))
+    with tm.assert_produces_warning(FutureWarning,
+                                    check_stacklevel=False):
+        df = read_table(datapath('io', 'parser', 'data', 'salaries.csv'))
+    return df
```
dahlbaek: Maybe this should just be

```python
@pytest.fixture
def salaries_table(datapath):
    """DataFrame with the salaries dataset"""
    return read_csv(datapath('io', 'parser', 'data', 'salaries.csv'), sep="\t")
```

instead?

dahlbaek:
I wonder if it would be better to get rid of the function factory `_make_parser_function` altogether and define `read_csv` and `read_table` separately. It would cause some code duplication, but the code would be easier to read. Also, in the future I would like to propose aligning parameter names of `read_csv` and `to_csv`, and perhaps propose changes to the current `sep`/`delimiter` situation (#7662), but there is no reason those changes should affect `read_table`.

WillAyd:
Not sure I completely follow your logic there but it doesn't really seem worth the effort to do, especially since you are deprecating `read_table` here

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -319,6 +319,7 @@ Deprecations
 - :meth:`DataFrame.to_stata`, :meth:`read_stata`, :class:`StataReader` and :class:`StataWriter` have deprecated the ``encoding`` argument.  The encoding of a Stata dta file is determined by the file type and cannot be changed (:issue:`21244`).
 - :meth:`MultiIndex.to_hierarchical` is deprecated and will be removed in a future version  (:issue:`21613`)
 - :meth:`Series.ptp` is deprecated. Use ``numpy.ptp`` instead (:issue:`21614`)
+- :func:`pandas.read_table` is deprecated. Use ``pandas.read_csv`` instead (:issue:`21948`)
```
jreback: need a ``:func:`` on read_csv

pandas/tests/io/parser/test_unsupported.py
```diff
@@ -43,23 +45,43 @@ def test_c_engine(self):
...
+            with tm.assert_produces_warning(FutureWarning):
+                read_table(StringIO(data), engine='c', quotechar=chr(128))
         with tm.assert_raises_regex(ValueError, msg):
-            read_table(StringIO(data), engine='c', skipfooter=1)
+            with tm.assert_produces_warning(FutureWarning):
+                read_table(StringIO(data), engine='c', skipfooter=1)
 
         # specify C-unsupported options without python-unsupported options
         with tm.assert_produces_warning(parsers.ParserWarning):
+            warnings.simplefilter("ignore", category=FutureWarning)
```
jreback: why are you adding these as read_table? is it to support the read_table deprecation itself? if so, then make a new test function and localize the testing there. if not, then why are you doing it?

pandas/tests/test_multilevel.py
```diff
@@ -519,7 +519,8 @@ def test_xs_level_multiple(self):
 a   q   20      4     0.4473  1.4152  0.2834  1.00661  0.1744
 x   q   30      3    -0.6662 -0.5243 -0.3580  0.89145  2.5838"""
 
-        df = read_table(StringIO(text), sep=r'\s+', engine='python')
+        with tm.assert_produces_warning(FutureWarning):
```
jreback: anywhere non-essentiall, just change to use read_csv

jreback (changes requested):

jreback:
there is also a check for using pd.read_table in the top-level api that I am not sure is affected but pls check (pandas/tests/api/test_api.py)

pandas/tests/io/parser/test_unsupported.py
dahlbaek: I wanted to assert that both a `parsers.ParserWarning` and a `FutureWarning` is being emitted, and couldn't figure out any other way than first ignoring the `parsers.ParserWarning`, then ignoring the `FutureWarning`. Should I change the `read_table` calls to `read_csv` calls instead? Or simply ignore the `FutureWarning`?

jreback: I think you can use a list of the warnings to assert_produces_warning

dahlbaek: I see, you're right. Maybe it is best to change to `read_csv` anyway, since the tests do not seem to be `read_table` specific?

dahlbaek:
> there is also a check for using pd.read_table in the top-level api that I am not sure is affected but pls check (pandas/tests/api/test_api.py)

I think that test checks the pandas namespace, which I suppose should not change until `pandas.read_table` is actually removed completely. So no change required at this point in time, unless I am mistaken.

doc/source/whatsnew/v0.24.0.txt
dahlbaek: Fixed

pandas/tests/test_multilevel.py
dahlbaek: Did my best to find non-essential places and do this.

pandas/io/parsers.py
```diff
@@ -606,11 +614,24 @@ def parser_f(filepath_or_buffer,
                  memory_map=False,
                  float_precision=None):
 
+        # deprecate read_table
+        if name == "read_table":
```
jreback: can you add the issue number here

pandas/tests/io/test_common.py
```diff
@@ -169,6 +175,23 @@ def test_read_fspath_all(self, reader, module, path, datapath):
         mypath = CustomFSPath(path)
         result = reader(mypath)
         expected = reader(path)
+
+        if path.endswith('.pickle'):
```
jreback: why was this change needed?

pandas/tests/io/test_common.py
```diff
@@ -169,6 +175,23 @@ def test_read_fspath_all(self, reader, module, path, datapath):
...
+            # categorical
+            tm.assert_categorical_equal(result, expected)
+        else:
+            tm.assert_frame_equal(result, expected)
+
+    def test_read_fspath_all_read_table(self, datapath):
+        pytest.importorskip('os')
```
jreback: why would os not be available?

jreback (changes requested):
some comments. can you run the test suite and make sure you are catching all deprecation warnings for read_table.

jreback:
@gfyoung if you have any comments

pandas/core/frame.py
```diff
@@ -1592,11 +1592,11 @@ def from_csv(cls, path, header=0, sep=',', index_col=0, parse_dates=True,
...
-        from pandas.io.parsers import read_table
-        return read_table(path, header=header, sep=sep,
-                          parse_dates=parse_dates, index_col=index_col,
-                          encoding=encoding, tupleize_cols=tupleize_cols,
-                          infer_datetime_format=infer_datetime_format)
+        from pandas.io.parsers import read_csv
+        return read_csv(path, header=header, sep=sep,
+                        parse_dates=parse_dates, index_col=index_col,
+                        encoding=encoding, tupleize_cols=tupleize_cols,
+                        infer_datetime_format=infer_datetime_format)
```
gfyoung: As it stands, this is a breaking API change.

@jreback : `0.21.0` deprecations are fair game to remove in `0.24.0`, no?

pandas/io/clipboards.py
```diff
@@ -9,7 +9,7 @@
 
 def read_clipboard(sep=r'\s+', **kwargs):  # pragma: no cover
     r"""
-    Read text from clipboard and pass to read_table. See read_table for the
+    Read text from clipboard and pass to read_csv. See read_csv for the
```
gfyoung: Hmm...this is arguably breaking (the `sep` default is different for both functions), but since we explicitly have a default arg for `sep`, this *probably* should be okay (for at least the vast majority of users).

pandas/io/parsers.py
```diff
@@ -606,11 +614,24 @@ def parser_f(filepath_or_buffer,
...
+            if sep is False and delimiter is None:
+                warnings.warn("read_table is deprecated, use read_csv "
+                              "instead, passing sep='\\t'.",
+                              FutureWarning, stacklevel=2)
+            else:
+                warnings.warn("read_table is deprecated, use read_csv "
+                              "instead.",
+                              FutureWarning, stacklevel=2)
+            if sep is False:
+                sep = default_sep
```
gfyoung: Why do we need this `if` condition here?

pandas/core/frame.py
dahlbaek: As far as I know, the only difference between `read_csv` and `read_table` is the default value of `sep`. Here, `sep` is explicitly set, so I don't see how this changes anything?

pandas/tests/io/test_common.py
dahlbaek: There is no change. These six lines repeat at the bottom of `test_read_fspath_all_read_table` below, `diff` just does not realise that those six lines are the ones that "ought" to be highlighted instead of these six lines.

pandas/tests/io/test_common.py
dahlbaek: I would not know… frankly I just went with what was already written in `test_read_fspath_all`, plugging in `tm.assert_produces_warning` where appropriate.

pandas/io/parsers.py
dahlbaek: If the user is setting `sep` or `delimiter`, I want to suggest using `read_csv`, but if neither is set, I also want to suggest passing `sep='\t'`. This was the simplest solution I could come up with which handles the case correclt where the user is passing `sep=';'` to `read_table`.

pandas/io/clipboards.py
dahlbaek: Same, maybe I am missing something, but I think the only difference between `read_table` and `read_csv` is anyway being overridden here, thus making the change non-breaking.

dahlbaek:
> some comments. can you run the test suite and make sure you are catching all deprecation warnings for read_table.

I'm not seeing any `FutureWarning`s from `read_table`, but I am getting the following failed test:

```
(pandas-dev) $ pytest test_errors.py
=============================== test session starts ===============================
platform linux -- Python 3.6.6, pytest-3.6.3, py-1.5.4, pluggy-0.6.0
rootdir: /home/dahlbaek/git/pandas, inifile: setup.cfg
plugins: xdist-1.22.2, forked-0.2, cov-2.5.1
collected 12 items                                                                

test_errors.py ..........F.                                                 [100%]

==================================== FAILURES =====================================
________________________________ test_error_rename ________________________________

    def test_error_rename():
        # see gh-12665
        from pandas.errors import ParserError
        from pandas.io.common import CParserError
    
        try:
            raise CParserError()
        except ParserError:
            pass
    
        try:
            raise ParserError()
        except CParserError:
            pass
    
        with catch_warnings(record=True):
            try:
>               raise ParserError()
E               pandas.errors.ParserError

test_errors.py:52: ParserError
======================= 1 failed, 11 passed in 0.07 seconds =======================
(pandas-dev) $ 
```

Not sure exactly what is happening here…

pandas/core/frame.py
gfyoung: You're right: for the vast majority, it's not going to be breaking.  It's only in a corner case situation (e.g. you pass in `None`) that you will see a change.  This is fine then.

pandas/io/clipboards.py
gfyoung: Same as above.  It should probably be fine.  Don't worry about it.

pandas/io/parsers.py
gfyoung: Hmm...I see.  Okay, fair enough.

pandas/tests/io/test_common.py
gfyoung: Let's remove that line then.

gfyoung:
@dahlbaek : That's an odd test failure.  Seems to be fine on Travis though.  Also, the test logs don't seem to have any deprecation warnings related to `read_table` AFAICT.

jreback:
@dahlbaek we don't usually run *from* the test dir itself, rather from the base dir e.t.
``pytest path/to/test``. running from the test dir might not find the fixtures.

jreback (approved):

jreback:
@jorisvandenbossche @TomAugspurger any objections?

jreback:
@dahlbaek see my comments above. The tests are not designed to be run from subdirs, nor are they on our CI. pytest can't find things.

jreback:
@dahlbaek hmm that looks like a real error

dahlbaek:
@jreback I'll try to look into it when I have time and make a separate issue.

dahlbaek:
@jreback I re-downloaded the repo and installed a fresh python3.7 environment. I don't know what was causing the problem, but now it is gone.

jreback:
@TomAugspurger if you have any issues here

gfyoung:
@dahlbaek : This looks good to me, though if you could fix the merge conflicts, that would be good.

dahlbaek:
@gfyoung Done.

TomAugspurger:
Pushed a doc formatting fix.

Thanks @dahlbaek!

PR #21962, opened 2018-07-18
Trim unncessary code in datetime/np_datetime.c

jbrockmendel:
`pydatetime_to_datetimestruct` does a ton of checking that boils down to "is this a valid datetime object?"  Since the function only gets called after a type-check, we can assume it is a date/datetime and be a lot less verbose about it.

This also rips out an unnecessary layer of functions `convert_datetime_to_datetimestruct`, `convert_timedelta_to_timedeltastruct`.

cc @WillAyd you mentioned wanting to work on your C-foo.  There's a comment about figuring out how to import the cpython datetime C-API.  Any thoughts?

### DIFF
--- a/pandas/_libs/tslibs/np_datetime.pyx
+++ b/pandas/_libs/tslibs/np_datetime.pyx
@@ -147,6 +147,9 @@ cdef inline void td64_to_tdstruct(int64_t td64,
 
 cdef inline int64_t pydatetime_to_dt64(datetime val,
                                        npy_datetimestruct *dts):
+    """
+    Note we are assuming that the datetime object is timezone-naive.
+    """
     dts.year = PyDateTime_GET_YEAR(val)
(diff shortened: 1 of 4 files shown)


pandas/_libs/tslibs/np_datetime.pyx
```diff
@@ -147,6 +147,9 @@ cdef inline void td64_to_tdstruct(int64_t td64,
 
 cdef inline int64_t pydatetime_to_dt64(datetime val,
                                        npy_datetimestruct *dts):
+    """
```
jreback: add an explicit check?

jbrockmendel: This is called by some very perf-sensitive code, all uses of which already handle this carefully.  I think its sufficiently internal to be OK without.

jreback:
thanks @jbrockmendel

PR #21974, opened 2018-07-19
CLN:Unused Variables

mroeschke:

Used PyCharm inspections to check:

- Unused local variables
- Unreachable code
- `set([..])` to `{...}`

### DIFF
--- a/pandas/_libs/internals.pyx
+++ b/pandas/_libs/internals.pyx
@@ -390,7 +390,7 @@ def get_blkno_indexers(int64_t[:] blknos, bint group=True):
     start = 0
     cur_blkno = blknos[start]
 
-    if group == False:
+    if not group:
         for i in range(1, n):
             if blknos[i] != cur_blkno:
                 yield cur_blkno, slice(start, i)
--- a/pandas/compat/pickle_compat.py
+++ b/pandas/compat/pickle_compat.py
@@ -15,9 +15,6 @@ def load_reduce(self):
     args = stack.pop()
     func = stack[-1]
 
-    if len(args) and type(args[0]) is type:
-        n = args[0].__name__  # noqa
-
     try:
         stack[-1] = func(*args)
         return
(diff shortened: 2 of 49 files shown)


jreback (changes requested):
there are some very aggressive changes here

pandas/_libs/internals.pyx
```diff
@@ -390,7 +390,7 @@ def get_blkno_indexers(int64_t[:] blknos, bint group=True):
     start = 0
     cur_blkno = blknos[start]
 
-    if group == False:
+    if not group:
```
jbrockmendel: Group is false

pandas/compat/pickle_compat.py
```diff
@@ -15,9 +15,6 @@ def load_reduce(self):
     args = stack.pop()
     func = stack[-1]
 
-    if len(args) and type(args[0]) is type:
-        n = args[0].__name__  # noqa
```
jbrockmendel: Do we know what this is for? “Noqa” suggests author knewflake would complain

pandas/core/generic.py
```diff
@@ -1022,9 +1022,6 @@ def rename(self, *args, **kwargs):
         copy = kwargs.pop('copy', True)
         inplace = kwargs.pop('inplace', False)
         level = kwargs.pop('level', None)
-        axis = kwargs.pop('axis', None)
```
jbrockmendel: I think part of the point of this is to empty out the kwargs dict, no?

jbrockmendel:
I thought flake8 checked for unused variables, unreachable code.  Are we using an old version or something?

The set comprehensions are a good fix.  Might be worth adding that flake8 plugin while you’re at it.

mroeschke:
I'm not sure if PyCharm is using flake8 under the hood or they have their own introspection method, but it was catching some things that flake8 wasn't.

Nonetheless, I'll scale this back and break it into smaller PR's

PR #21981, opened 2018-07-19
ENH: Implement subtraction for object-dtype Index

### LINKED ISSUE #19369: Index.__add__ OK but __sub__ not?
Curious about why `Index.__add__` tries does the natural thing but `Index.__sub__` raises `TypeError`.

```
    def __add__(self, other):
        return Index(np.array(self) + other)

    def __radd__(self, other):
        return Index(other + np.array(self))

    __iadd__ = __add__

    def __sub__(self, other):
        raise TypeError("cannot perform __sub__ with this index type: "
                        "{typ}".format(typ=type(self)))
```

jbrockmendel:
- [x] closes #19369
- [x] tests added / passed
- [x] whatsnew entry

@jreback @jorisvandenbossche discussed briefly at the sprint.  Merits more thorough testing, but I'd like to get the go-ahead to separate out arithmetic tests that are common to EA/Index/Series/Frame[1col] that are highly duplicative first.

### DIFF
--- a/pandas/tests/indexes/test_base.py
+++ b/pandas/tests/indexes/test_base.py
@@ -903,7 +903,7 @@ class TestIndex(Base):
             'foo' - index
 
         with pytest.raises(TypeError):
-            np.array([2, pd.Timestamp.now()]) - index
+            np.array([True, pd.Timestamp.now()]) - index
 
     def test_map_identity_mapping(self):
         # GH 12766
--- a/pandas/tests/indexes/test_base.py
+++ b/pandas/tests/indexes/test_base.py
@@ -871,6 +871,18 @@ class TestIndex(Base):
         pytest.raises(TypeError, lambda: index - index.tolist())
         pytest.raises(TypeError, lambda: index.tolist() - index)
 
+    def test_sub_object(self):
+        # GH#19369
+        from decimal import Decimal
+        index = pd.Index([Decimal(1), Decimal(2)])
+        expected = pd.Index([Decimal(0), Decimal(1)])
+
+        result = index - Decimal(1)
+        tm.assert_index_equal(result, expected)
+
+        result = index - pd.Index([Decimal(1), Decimal(1)])
+        tm.assert_index_equal(result, expected)
+
     def test_map_identity_mapping(self):
         # GH 12766
         # TODO: replace with fixture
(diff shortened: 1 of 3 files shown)


pandas/tests/indexes/test_base.py
```diff
@@ -871,6 +871,18 @@ def test_sub(self):
...
+        from decimal import Decimal
+        index = pd.Index([Decimal(1), Decimal(2)])
+        expected = pd.Index([Decimal(0), Decimal(1)])
+
+        result = index - Decimal(1)
+        tm.assert_index_equal(result, expected)
+
+        result = index - pd.Index([Decimal(1), Decimal(1)])
+        tm.assert_index_equal(result, expected)
+
```
gfyoung: Can we also add tests where element-wise subtraction should fail.

jbrockmendel: There are already some in the test just above this, but more couldn’t hurt.  Will do.

gfyoung: Ah, yes, thanks for pointing that out.  Two things would be good there:

* We should rename that test above to be `test_sub_fail`
* Add a couple of tests where element-wise subtraction should fail

jorisvandenbossche (approved):
Yes, I am +1 on this.

Regarding testing, I think you should only add `rsub` case to the test?

jorisvandenbossche:
>  but I'd like to get the go-ahead to separate out arithmetic tests that are common to EA/Index/Series/Frame[1col] that are highly duplicative first.

You will need to give some more details, but let's do such a change in a separate PR anyway, so I would simply fix up and merge this one first.

jreback (changes requested):

pandas/tests/indexes/test_base.py
jbrockmendel: You got it.

jbrockmendel:
> Regarding testing, I think you should only add rsub case to the test?

> I don't think you are exercising rsub?

rsub is tested implicitly in the `result = index - pd.Index([Decimal(1), Decimal(1)])` line.  I'll make that explicit.

jorisvandenbossche (approved):

jorisvandenbossche:
@jbrockmendel thanks!

PR #21982, opened 2018-07-19
REF: No need to delegate to index check of whether an int is an int

toobaz:
- [x] tests passed
- [x] whatsnew entry

Just fixing the following nonsensical error:

``` python
In [2]: pd.Series(range(10)).iloc['a']
<ipython-input-2-4a32011255fe> in <module>()
----> 1 pd.Series(range(10)).iloc['a']
[...]
TypeError: cannot do positional indexing on <class 'pandas.core.indexes.range.RangeIndex'> with these indexers [a] of <class 'str'>
```

(you just cannot do positional indexing with a ``str``, regardless of the index)

... and adding a docstring while I was at it.

### DIFF
--- a/pandas/core/indexing.py
+++ b/pandas/core/indexing.py
@@ -2124,7 +2124,25 @@ class _iLocIndexer(_LocationIndexer):
         return values
 
     def _validate_integer(self, key, axis):
-        # return a boolean if we have a valid integer indexer
+        """
+        Check that 'key' is a valid position in the desired axis.
+
+        Parameters
+        ----------
+        key : int
+            Requested position
+        axis : int
+            Desired axis
+
+        Returns
+        -------
+        None
+
+        Raises
+        ------
+        IndexError
+            If 'key' is not a valid position in axis 'axis'
+        """
 
         ax = self.obj._get_axis(axis)
         l = len(ax)
@@ -2215,8 +2233,6 @@ class _iLocIndexer(_LocationIndexer):
 
         # a single integer
         else:
-            key = self._convert_scalar_indexer(key, axis)
-
             if not is_integer(key):
                 raise TypeError("Cannot index by location index with a "
                                 "non-integer key")
--- a/pandas/tests/indexing/test_floats.py
+++ b/pandas/tests/indexing/test_floats.py
@@ -50,7 +50,7 @@ class TestFloatIndexers(object):
             def f():
                 s.iloc[3.0]
             tm.assert_raises_regex(TypeError,
-                                   'cannot do positional indexing',
+                                   'Cannot index by location index',
                                    f)
 
             def f():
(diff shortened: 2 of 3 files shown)


pandas/core/indexing.py
```diff
@@ -2215,8 +2233,6 @@ def _getitem_axis(self, key, axis=None):
 
         # a single integer
         else:
-            key = self._convert_scalar_indexer(key, axis)
```
jorisvandenbossche: Can you check if there is any remaining occasion where `_convert_scalar_indexer` is used for `iloc` (in principle it shouldn't?). Because if not, `_convert_scalar_indexer` has some redundant code

doc/source/whatsnew/v0.23.4.txt
```diff
@@ -32,6 +32,7 @@ Bug Fixes
 
 - Bug where calling :func:`DataFrameGroupBy.agg` with a list of functions including ``ohlc`` as the non-initial element would raise a ``ValueError`` (:issue:`21716`)
 - Bug in ``roll_quantile`` caused a memory leak when calling ``.rolling(...).quantile(q)`` with ``q`` in (0,1) (:issue:`21965`)
+- Passing a non-int scalar to ``.iloc`` now raises a more appropriate error message (:issue:`21982`)
```
jorisvandenbossche: Not sure if it is worth to add a whatsnew for this.

jorisvandenbossche (commented):
Good change! (I have also always been annoyed if I saw that error :-))

pandas/core/indexing.py
toobaz: Yes, there is, and it was related to a commit I pushed but then removed, because unfortunately it broke some tests and I don't have time to check them in the next days. So I went for the obvious change only, will add the rest in some other PR (which I think will be quite large, as I will also touch code in indexes).

doc/source/whatsnew/v0.23.4.txt
toobaz: OK, removing

pandas/core/indexing.py
jorisvandenbossche: OK. Can you point to where it is currently used? (to have this on record) 
Is it worth to add a todo comment there?

toobaz: I'll rebase my other PR on this once merged, push it and open an issue referring to it

jreback (changes requested):
can you add the test in top of the PR?

pandas/tests/indexing/test_floats.py
```diff
@@ -50,7 +50,7 @@ def test_scalar_error(self):
             def f():
                 s.iloc[3.0]
             tm.assert_raises_regex(TypeError,
-                                   'cannot do positional indexing',
+                                   'Cannot index by location index',
```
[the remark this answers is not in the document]
toobaz: We already have both messages in the code (I did not write a new message).

The only difference is that now this case raises the correct one.

jreback (approved):

PR #21983, opened 2018-07-19
ENH GH20927 Added convert_dates parameter (default=True) to read_sas()

paul-lilley:
Added parameter which is passed down to the underlying
SAS7BDATReader.  This allows SAS datasets with dates too large
for pandas.datetime to be read in with the dates/datetimes as
SAS native float64 values, rather than trying (and failing) to convert
to pandas.datetime.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


jbrockmendel:
Could these be converted to python datetime?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -458,6 +458,7 @@ I/O
 
 - :func:`read_html()` no longer ignores all-whitespace ``<tr>`` within ``<thead>`` when considering the ``skiprows`` and ``header`` arguments. Previously, users had to decrease their ``header`` and ``skiprows`` values on such tables to work around the issue. (:issue:`21641`)
 - :func:`read_excel()` will correctly show the deprecation warning for previously deprecated ``sheetname`` (:issue:`17994`)
+- :func:`read_sas()` now has new parameter convert_dates(default=True) which is passed through to SAS7BDATReader. Setting convert_dates=False allows SAS datasets with dates larger than pd.Timestamp.max (2262-04-11 23:47:16.854775807) to be read in with all date/datetime columns as their native float64.  The SAS dates can then be converted to datetime.date/datetime objects post-import, or the values capped to pd.Timestamp.max and converted with pd.to_datetime()  (:issue:`20927`)
```
gfyoung: Can we make this shorter by any chance?  Our entries are generally two lines.

pandas/tests/io/sas/test_sas7bdat.py
```diff
@@ -188,3 +188,26 @@ def test_zero_variables(datapath):
...
     with pytest.raises(EmptyDataError):
         pd.read_sas(fname)
+
+
+def test_max_sas_date_exception(datapath):
+    # GH 20927
+    fname = datapath("io", "sas", "data", "max_sas_date.sas7bdat")
+    with pytest.raises(pd._libs.tslibs.np_datetime.OutOfBoundsDatetime):
+        pd.read_sas(fname)
+        assert True
```
gfyoung: Why do we need this line exactly?

pandas/tests/io/sas/test_sas7bdat.py
```diff
@@ -188,3 +188,26 @@ def test_zero_variables(datapath):
...
+    df = pd.read_sas(fname, encoding='iso-8859-1', convert_dates=False)
+    # SAS likes to left pad strings with spaces - lstrip before comparing
+    str_cols = df.select_dtypes(['object']).columns
+    df[str_cols] = df[str_cols].apply(lambda x: x.str.lstrip(' '))
+    fname = datapath("io", "sas", "data", "max_sas_date.csv")
+    df0 = pd.read_csv(
+        fname, dtype={'dt_as_float': np.float64,
+                      'dt_as_dt': np.float64, 'date_as_date': np.float64,
+                      'date_as_num': np.float64})
+    tm.assert_frame_equal(df, df0)
```
gfyoung: * I would just construct the expected CSV from scratch
* Instead of `df0`, can we use a more descriptive name like `expected` ?

gfyoung:
> Could these be converted to python datetime?

@jbrockmendel : I suspect you'll have to dig into the `SAS7BDATReader` internals to find out more.  I think the change as it stands makes sense.

doc/source/whatsnew/v0.24.0.txt
paul-lilley: I'll reduce to 2 lines.

pandas/tests/io/sas/test_sas7bdat.py
paul-lilley: Fair point - I'll construct the expected dataframe from scratch and give a more descriptive name (the comparison df vs df0 was to continue to style of other existing tests but I agree that it's not the best naming).

paul-lilley:
re: Could these be converted to python datetime?

@jbrockmendel, @gfyoung:
Thanks for comment and review.

I did initially look to catch the exception and convert to datetime.datetime instead of Pandas datetime, however it breaks some existing tests so I thought it simpler to expose the convert_dates parameter from SAS7BDATReader and allow the user to explicitly deal with the very large dates.
I'm happy to look further into the conversion to python datetime if you think this is a better solution (and have code I can submit if it is of interest)

gfyoung:
@paul-lilley : In the interests of modularity, I would not make any changes to `SAS7BDATReader` for this PR.  You are more than welcome to investigate and submit a PR for the `SAS7BDATReader` if you would like following this one.

jreback (changes requested):
you don't need a new parameter here. you simply need to catch the OutOfBoundsDatetime and return an object array, rather than a datetime converted object.

paul-lilley:
@jreback 
Hi, thanks for the review.  
Just to clarify, do you mean to keep the current conversion, returning pd.datetime if all ok, catch OutOfBoundsDatetime if not and return an object array of datetime.datetime objects? 
This would return different types depending on the range of dates in the SAS dataset, as SAS7BDATReader is an iterator this could give some chunks converted to pd.datetime and others to datetime.datetime within a given SAS dataset, but I guess this will unlikely break any existing code and retains the more efficient pd.datetime (ie. a numpy datetime) for most SAS datasets.

jreback:
> https://github.com/pandas-dev/pandas/pull/21983#issuecomment-407642117

@paul-lilley yes that's what I would do. This is how pandas works elsewhere and would be least suprising.

paul-lilley:
Thanks - I'll make the changes and update the PR title, probably after Europython this week.

jreback:
can you merge master and update to comments

jreback:

paul-lilley:
I'll pick this up in a new PR, resolving the issue as discussed with jreback above on 25 Jul 2018

PR #21984, opened 2018-07-19
[CLN] Collect __repr__ and rendering methods together

jbrockmendel:
Motivation here is in looking at options for rendering EA subclasses.

There's a decent argument to be made that `to_html` and/or `to_latex` should be included in "rendering methods", in large part because they both call `fmt.Formatter`.  Holding off on that pending feedback.


jbrockmendel:
Closing to clear the queue.

PR #21986, opened 2018-07-20
CLN: Remove unused variables

mroeschke:
Breaking up #21974.

Removes non-noqa, seemingly non-controversial, unused local variables according to PyCharm. These are mostly redefined elsewhere or not used.

I added some TODO comments about other unused local variables that seem misused.


pandas/core/generic.py
```diff
@@ -1024,6 +1024,7 @@ def rename(self, *args, **kwargs):
         level = kwargs.pop('level', None)
         axis = kwargs.pop('axis', None)
         if axis is not None:
+            # TODO: axis is unused, is this just validating?
             axis = self._get_axis_number(axis)
```
gfyoung: I suspect the answer is yes, in which case, not assigning the variable is fine.

pandas/io/json/json.py
```diff
@@ -547,6 +547,7 @@ def _get_object_parser(self, json):
 
         if typ == 'series' or obj is None:
             if not isinstance(dtype, bool):
+                # TODO: dtype is unused. Should this be an update on kwargs?
                 dtype = dict(data=dtype)
```
gfyoung: If we can find an example that breaks because of this unused variable, that would be good for another PR.

gfyoung: FYI, comment applies to many of the other variables in which you questioned whether it actually should be used, but previous authors didn't get around to implementing functionality with it.

pandas/core/nanops.py
```diff
@@ -479,6 +479,9 @@ def nanvar(values, axis=None, skipna=True, ddof=1):
 
 @disallow('M8', 'm8')
 def nansem(values, axis=None, skipna=True, ddof=1):
+    # This checks if non-numeric-like data is passed with numeric_only=False
+    # and raises a TypeError otherwise
+    var = nanvar(values, axis, skipna, ddof=ddof)
```
gfyoung: If `var` isn't used, don't assign the result of `nanvar` to a variable.  Just call it without.

jreback (changes requested):
rather than comments I would just like to see if changing breaks things. it might be that we are not testing that case or its unused code

pandas/core/indexes/category.py
```diff
@@ -133,7 +133,7 @@ def _create_from_codes(self, codes, categories=None, ordered=None,
         if name is None:
             name = self.name
         cat = Categorical.from_codes(codes, categories=categories,
-                                     ordered=self.ordered)
+                                     ordered=ordered)
```
jbrockmendel: We’re there cases before where the wrong thing was passed? I.e. None != ordered != self.ordered?

pandas/core/series.py
```diff
@@ -2479,7 +2478,8 @@ def sort_values(self, axis=0, ascending=True, inplace=False,
         dtype: object
         """
         inplace = validate_bool_kwarg(inplace, 'inplace')
-        axis = self._get_axis_number(axis)
+        # Valida the axis parameter
```
jbrockmendel: Validate typo

jbrockmendel:
Mostly looks good.  There are a few places where missing arguments are added that may merit tests.

jreback:
can you rebase

pandas/core/indexes/category.py
jreback: is this tested anywhere? I agree this should have broken things

jreback (changes requested):

pandas/core/indexes/category.py
mroeschke: It appears that all instances of this private method `_create_from_codes`  in the codebase never passed an arg to `ordered` (i.e.  `ordered` always defaulted to `None` which always got reassigned to `self.ordered` here)

mroeschke:
```
(pandas-dev) matthewroeschke:pandas-mroeschke matthewroeschke$ grep -R --include="*.py" _create_from_codes .
./pandas/core/indexes/category.py:    def _create_from_codes(self, codes, categories=None, ordered=None,
./pandas/core/indexes/category.py:                new_target = self._create_from_codes(codes)
./pandas/core/indexes/category.py:        return self._create_from_codes(taken)
./pandas/core/indexes/category.py:        return self._create_from_codes(np.delete(self.codes, loc))
./pandas/core/indexes/category.py:        return self._create_from_codes(codes)
./pandas/core/indexes/category.py:        result = self._create_from_codes(codes, name=name)
./pandas/core/indexes/category.py:        # if name is None, _create_from_codes sets self.name
```

pandas/core/internals/blocks.py
```diff
@@ -1248,7 +1248,7 @@ def take_nd(self, indexer, axis, new_mgr_locs=None, fill_tuple=None):
         if fill_tuple is None:
             fill_value = self.fill_value
             new_values = algos.take_nd(values, indexer, axis=axis,
-                                       allow_fill=False)
+                                       allow_fill=False, fill_value=fill_value)
```
mroeschke: This modification is deep within the internals (block's definition of `take_nd`), and I am not sure how to really add a test for this.

pandas/tests/io/json/test_pandas.py
```diff
@@ -642,6 +642,13 @@ def test_series_from_json_precise_float(self):
         result = read_json(s.to_json(), typ='series', precise_float=True)
         assert_series_equal(result, s, check_index_type=False)
 
+    def test_series_with_dtype(self):
```
mroeschke: This test should hit the modification made here: https://github.com/mroeschke/pandas/blob/0d7f07783ad9a42bb3fc7f0e3dda0c01b877fe57/pandas/io/json/json.py#L550

mroeschke:
I think everything should generally be covered with a test besides the `take_nd` modification explained above.

pandas/core/internals/blocks.py
jreback: this actually isn't used since allow_fill=False

pandas/core/nanops.py
```diff
@@ -635,7 +637,6 @@ def nankurt(values, axis=None, skipna=True):
         adj = 3 * (count - 1) ** 2 / ((count - 2) * (count - 3))
         numer = count * (count + 1) * (count - 1) * m4
         denom = (count - 2) * (count - 3) * m2**2
-        result = numer / denom - adj
```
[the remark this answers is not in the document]
jreback: is this not used?

jreback (approved):
minor questions, rebase and ping on green.

mroeschke:
@jreback all green

jreback:
thanks @mroeschke

PR #21987, opened 2018-07-20
allow using Iterable in Series and DataFrame constructor

### LINKED ISSUE #2193: Cannot produce DataFrame from dict-yielding iterator.
Hi there,

I have a codebase that (so far) consumes data series in the form of dict-yielding iterators, similar in form to how DataFrame's constructor accepts a list of such. Unfortunately it seems from current implementation of that constructor, there is no straightforward way to consume from an iterator straight into a DataFrame.

For example, I have a row factory for sqlite3 written in C that produces dicts. If pandas would accept an iterator, it should be possible to consume a series directly from speedy storage code into speedy data structure code. :)

Is there an inherent limitation preventing this? Otherwise with a few pointers I would be willing to try contribute the patch myself.

Thanks,

David

holymonson:
- [X] closes #2193
- [x] tests added / passed
- [ ] whatsnew entry
---

@TomAugspurger Hope you can review this.

BTW, perhpas you may want to change `Iterable` in `is_list_like()` to `Sequence`.

```python
gen = (i for i in range(10))
pandas.core.dtypes.inference.is_list_like(gen)
# True
len(gen)
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
# TypeError: object of type 'generator' has no len()
gen[0]
# Traceback (most recent call last):
#   File "<stdin>", line 1, in <module>
# TypeError: 'generator' object is not subscriptable
```

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -267,7 +267,7 @@ class DataFrame(NDFrame):
 
     Parameters
     ----------
-    data : numpy ndarray (structured or homogeneous), dict, or DataFrame
+    data : ndarray (structured or homogeneous), Iterable, dict, or DataFrame
         Dict can contain Series, arrays, constants, or list-like objects
 
         .. versionchanged :: 0.23.0
@@ -391,6 +391,8 @@ class DataFrame(NDFrame):
             else:
                 mgr = self._init_ndarray(data, index, columns, dtype=dtype,
                                          copy=copy)
+
+        # For data is list-like, or Iterable (will consume into list)
         elif (isinstance(data, collections.Iterable)
               and not isinstance(data, string_and_binary_types)):
             if not isinstance(data, collections.Sequence):
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -16,7 +16,6 @@ import functools
 import collections
 import itertools
 import sys
-import types
 import warnings
 from textwrap import dedent
 
@@ -75,7 +74,8 @@ from pandas.core.series import Series
 from pandas.core.arrays import Categorical, ExtensionArray
 import pandas.core.algorithms as algorithms
 from pandas.compat import (range, map, zip, lrange, lmap, lzip, StringIO, u,
-                           OrderedDict, raise_with_traceback)
+                           OrderedDict, raise_with_traceback,
+                           string_and_binary_types)
 from pandas import compat
 from pandas.compat import PY36
 from pandas.compat.numpy import function as nv
@@ -391,8 +391,9 @@ class DataFrame(NDFrame):
             else:
                 mgr = self._init_ndarray(data, index, columns, dtype=dtype,
                                          copy=copy)
-        elif isinstance(data, (list, types.GeneratorType)):
-            if isinstance(data, types.GeneratorType):
+        elif (isinstance(data, collections.Iterable)
+              and not isinstance(data, string_and_binary_types)):
+            if not isinstance(data, collections.Sequence):
                 data = list(data)
             if len(data) > 0:
                 if is_list_like(data[0]) and getattr(data[0], 'ndim', 1) == 1:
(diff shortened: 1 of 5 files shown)


holymonson:
This test failed. But a single-column `DataFrame` seems reasonable to me.
https://github.com/pandas-dev/pandas/blob/27ebb3e1e40513ad5f8919a5bbc7298e2e070a39/pandas/tests/frame/test_constructors.py#L868-L872
```python
pd.DataFrame((1, 2, 3))
#   0
#0  1
#1  2
#2  3
```

pandas/tests/series/test_constructors.py
```diff
@@ -156,10 +156,20 @@ def test_constructor_series(self):
...
-    def test_constructor_iterator(self):
+    def test_constructor_iterable(self):
+        class Iter():
+            def __iter__(self):
+                for i in range(10):
+                    yield i
 
         expected = Series(list(range(10)), dtype='int64')
-        result = Series(range(10), dtype='int64')
+        result = Series(Iter(), dtype='int64')
```
mroeschke: cant this simply be iter(range(10))?

holymonson: `range` is list-like, subclass of `Sequence`, which doesn't need  consuming by `data = list(data)`. I'd like to add back `range(10)` test and keep both.

holymonson:
Allow `pd.DataFrame((1, 2, 3))` because `pd.DataFrame([1, 2, 3])` is valid. They should be the same.

pandas/tests/frame/test_constructors.py
```diff
@@ -953,6 +947,16 @@ def __len__(self, n):
...
+    def test_constructor_iterable(self):
+        class Iter():
+            def __iter__(self):
+                for i in range(10):
+                    yield [1, 2, 3]
+
+        expected = DataFrame([[1, 2, 3]] * 10)
+        result = DataFrame(Iter())
+        tm.assert_frame_equal(result, expected)
+
```
gfyoung: Reference the issue number in a comment below the functional definition.  Applies to all newly-added tests.

TomAugspurger:
> Allow pd.DataFrame((1, 2, 3)) because pd.DataFrame([1, 2, 3]) is valid. They should be the same.

Context: That goes back a long way to DataMatrix: https://github.com/pandas-dev/pandas/commit/c34ac749382c2046548eb2470eb847d66ecc6d23#diff-225c6e44f997b9b698ab52a48b223018R39

+1 for allowing it.

TomAugspurger (approved):
Merging later today, unless someone else beats me to it.

jreback:
will have a look

don’t merge

pandas/core/frame.py
```diff
@@ -390,8 +390,9 @@ def __init__(self, data=None, index=None, columns=None, dtype=None,
             else:
                 mgr = self._init_ndarray(data, index, columns, dtype=dtype,
                                          copy=copy)
-        elif isinstance(data, (list, types.GeneratorType)):
-            if isinstance(data, types.GeneratorType):
+        elif (isinstance(data, collections.Iterable)
```
jreback: can you add a comment here (1 line above) which defines what this clause is checking

pandas/core/series.py
```diff
@@ -238,12 +238,12 @@ def __init__(self, data=None, index=None, dtype=None, name=None,
 
             elif is_extension_array_dtype(data):
                 pass
-            elif (isinstance(data, types.GeneratorType) or
-                  (compat.PY3 and isinstance(data, map))):
-                data = list(data)
             elif isinstance(data, (set, frozenset)):
                 raise TypeError("{0!r} type is unordered"
                                 "".format(data.__class__.__name__))
+            elif (isinstance(data, collections.Iterable)
```
jreback: same

jreback (changes requested):
pls add a whatsnew note, other enhancements in 0.24

jreback (approved):

jreback:
thanks @holymonson !

PR #21992, opened 2018-07-20
Move FrequencyInferer out of libresolution

jbrockmendel:
Discussed briefly, FrequencyInferer doesn't benefit much from cython, isn't needed elsewhere in tslibs, and with this move loses the dependency on khash.


pandas/_libs/tslibs/resolution.pyx
```diff
@@ -399,245 +358,14 @@ cdef object month_position_check(fields, weekdays):
         return None
 
 
-cdef inline bint _is_multiple(int64_t us, int64_t mult):
+@cython.returns(bint)
```
jreback: why don't you just type it and maybe cpdef?

pandas/tseries/frequencies.py
```diff
@@ -268,3 +282,224 @@ def infer_freq(index, warn=True):
...
+        return len(self.deltas_asi8) == 1
+
+    def get_freq(self):  # noqa:F811
+        if not self.is_monotonic or not self.index.is_unique:
+            return None
+
+        delta = self.deltas[0]
+        if libresolution.is_multiple(delta, _ONE_DAY):
+            return self._infer_daily_rule()
+        else:
```
jreback: remove the else & dedent

jreback (changes requested):

pandas/_libs/tslibs/resolution.pyx
jbrockmendel: It isn't called from cython, so no use for cpdef (and cython won't allow `def bint ...`).  For that matter might as well just move it along with FrequencyInferer.

jreback (approved):
just some doc-strings

pandas/tseries/frequencies.py
```diff
@@ -6,25 +6,32 @@
...
-from pandas._libs.tslibs import Timedelta
+from pandas._libs.tslibs import Timedelta, Timestamp
 
 import pandas._libs.tslibs.frequencies as libfreqs
 from pandas._libs.tslibs.frequencies import (  # noqa, semi-public API
     get_freq, get_base_alias, get_to_timestamp_base, get_freq_code,
     FreqGroup,
     is_subperiod, is_superperiod)
+from pandas._libs.tslibs.ccalendar import MONTH_ALIASES, int_to_weekday
+import pandas._libs.tslibs.resolution as libresolution
```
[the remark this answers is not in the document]
jbrockmendel: > maybe do one or the other here?

Other modules import from here.  Agreed that we need to clean that up, would prefer to do that in one of the periodic pure-cleanup PRs.

pandas/_libs/algos.pyx
```diff
@@ -71,6 +76,42 @@ class NegInfinity(object):
     __ge__ = lambda self, other: isinstance(other, NegInfinity)
 
 
+cpdef ndarray[int64_t, ndim=1] unique_deltas(ndarray[int64_t] arr):
```
jreback: as an aside I think we usually call this diff elsewhere, so maybe can share code, and might thing about renaming (future PR to think about)

pandas/_libs/tslibs/resolution.pyx
```diff
@@ -331,31 +314,7 @@ class Resolution(object):
...
-    kh_destroy_int64(table)
-
-    result = np.array(uniques, dtype=np.int64)
-    result.sort()
-    return result
-
-
-cdef object month_position_check(fields, weekdays):
+def month_position_check(fields, weekdays):
     cdef:
```
jreback: doc-string in future, does this need cdef?

pandas/tseries/frequencies.py
```diff
@@ -269,3 +283,246 @@ def infer_freq(index, warn=True):
 
     inferer = _FrequencyInferer(index, warn=warn)
     return inferer.get_freq()
+
+
+class _FrequencyInferer(object):
+    """
+    Not sure if I can avoid the state machine here
```
jreback: in future, pls fix the doc-string

pandas/tseries/frequencies.py
```diff
@@ -269,3 +283,246 @@ def infer_freq(index, warn=True):
 
     inferer = _FrequencyInferer(index, warn=warn)
     return inferer.get_freq()
+
+
+class _FrequencyInferer(object):
```
jreback: would not object to de-privatizing this

jreback (approved):
lgtm. some comments, pls follow up.

PR #21997, opened 2018-07-20
Sparse get dummies perf

TomAugspurger:
Previously, we did a scalar `elem == -1` for every element in the ndarray.

This replaces that check with a vectorized `array == -1`.

Running the ASV now. In the meantime, here's a simple timeit on the same problem

```python
# HEAD
In [3]: %timeit pd.get_dummies(s, sparse=True)
561 ms ± 4.96 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

# Master
In [3]: %timeit pd.get_dummies(s, sparse=True)
        2.18 s ± 273 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

### DIFF
--- a/asv_bench/benchmarks/reshape.py
+++ b/asv_bench/benchmarks/reshape.py
@@ -136,7 +136,6 @@ class PivotTable(object):
         self.df.pivot_table(index='key1', columns=['key2', 'key3'])
 
 
-
 class GetDummies(object):
     goal_time = 0.2
 
--- a/asv_bench/benchmarks/reshape.py
+++ b/asv_bench/benchmarks/reshape.py
@@ -1,7 +1,9 @@
+import string
 from itertools import product
 
 import numpy as np
 from pandas import DataFrame, MultiIndex, date_range, melt, wide_to_long
+import pandas as pd
 
 from .pandas_vb_common import setup  # noqa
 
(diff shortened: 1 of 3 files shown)


TomAugspurger:
Here's the ASV (only a 3x speedup).

```python
[100.00%] ··· Running reshape.GetDummies.time_get_dummies_1d_sparse                                                                                    1.79s       before           after         ratio
     [272bbdc7]       [bc658b03]
-           1.79s       1.05±0.02s     0.59  reshape.GetDummies.time_get_dummies_1d_sparse

```

asv_bench/benchmarks/reshape.py
```diff
@@ -132,3 +134,19 @@ def setup(self):
...
+class GetDummies(object):
+    goal_time = 0.2
+
+    def setup(self):
+        categories = list(string.ascii_letters[:12])
+        s = pd.Series(np.random.choice(categories, size=1_000_000),
+                      dtype=pd.api.types.CategoricalDtype(categories))
+        self.s = s
+
+    def time_get_dummies_1d(self):
```
mroeschke: Small nit: you can param over `sparce=False/True`

TomAugspurger: I have a slight preference for leaving them separate, since they're such distinct code paths and it's a tad easier to run just sparse with this layout. Happy to change if you feel strongly about this.

mroeschke: Sounds good, no strong preference to use `params` then.

jreback:
lgtm.

jreback:
thanks!

PR #22002, opened 2018-07-20
CLN: Remove Unneeded BlockManager methods

jbrockmendel:
BlockManager.get_scalar is never used outside of tests, 
BlockManager._is_indexed_like is only ever called from `merge`, which is only called once in Panel, before which the indexed_like check is already done.

Also BlockPlacements.iadd was more complicated than it needed to be, so tore out bits of that.


jorisvandenbossche (approved):

jreback:
thanks, always +1 on removing code!

PR #22006, opened 2018-07-21
Fix for issue #21150, using a simple lock to prevent an issue with multiple threads accessing an Index

### LINKED ISSUE #21150: BUG: race condition in Index.is_unique

Input:

```python
from concurrent.futures import ThreadPoolExecutor
import pandas as pd

x = pd.date_range('2001', '2020')
with ThreadPoolExecutor(2) as p:
    assert all(p.map(lambda x: x.is_unique, [x]*2))
```

Output:

```python
  File "bug.py", line 7, in <module>
    assert all(p.map(lambda x: x.is_unique, [x]*2))
AssertionError
```

#### Problem description

When calling `Index.is_unique` from multiple threads simultaneously, the wrong answer is returned.

#### Expected Output

Shouldn't raise.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.14.14-200.fc26.x86_64
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: C
LANG: C
LOCALE: None.None

pandas: 0.23.0
pytest: 3.5.1
pip: 10.0.1
setuptools: 39.1.0
Cython: 0.28.2
numpy: 1.13.3
scipy: 0.19.1
pyarrow: None
xarray: None
IPython: 4.2.1
sphinx: 1.7.4
patsy: 0.5.0
dateutil: 2.7.2
pytz: 2018.4
blosc: None
bottleneck: 1.2.1
tables: 3.4.3
numexpr: 2.6.2
feather: None
matplotlib: 2.2.2
openpyxl: None
xlrd: 1.1.0
xlwt: None
xlsxwriter: None
lxml: None
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.7
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: 0.1.5
pandas_gbq: None
pandas_datareader: None
</details>

kinow:
Hi,

First time working with Cython, and after uni did very little C/C++, so feel free to be a bit more serious about reviewing this pull request.

I am trying to practice more Python, and learn more about the code base of projects like Panda, and found the issue #21150, which happens a lot in Java (my main language), so decided to see how I'd go about fixing it.

Used the same approach found in `strptime.pyx`, to add a lock to prevent two threads of accessing the `mapping` Hashtable in the `_ensure_mapping_populated` method.

The problem occurred when two threads accessed the method around the same time, then one would create the `mapping` Hashtable instance, preventing the other from entering the `if` statement which also sets the `unique` flag to `true/1`.

I didn't find any other part that could cause an obvious issue due to this lock, or another thread-safety issue. However, I can imagine that moving the logic maybe to the constructor so that we immediately know whether the index is unique or not, could fix it too, and remove the need of the lock, at the expense of changing the current design (I believe the index is not tightly coupled with the values passed... later, after certain calls, the mapping hashtable gets populated, which gives better performance I believe).

Or, if core devs prefer to ask users to use a lock in their code base instead of implementing in Pandas, I'd be keen to close this PR and open a new one for the documentation with some note about it.

Feel free to suggest any changes or even leave comments to educate me :-) The documentation for setting up the development environment and for contributing was super easy to follow, I hope I didn't forget anything.

Thank you!
Bruno

- [ ] closes #21150
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/ci/travis-27.yaml
+++ b/ci/travis-27.yaml
@@ -8,6 +8,7 @@ dependencies:
   - cython=0.28.2
   - fastparquet
   - feather-format
+  - futures
   - gcsfs
   - html5lib
   - ipython
--- a/pandas/_libs/index.pyx
+++ b/pandas/_libs/index.pyx
@@ -23,6 +23,18 @@ from pandas._libs import algos, hashtable as _hash
 from pandas._libs.tslibs import Timestamp, Timedelta, period as periodlib
 from pandas._libs.missing import checknull
 
+# Python 2 vs Python 3
+try:
+    from thread import allocate_lock as _thread_allocate_lock
+except ImportError:
+    try:
+        from _thread import allocate_lock as _thread_allocate_lock
+    except ImportError:
+        try:
+            from dummy_thread import allocate_lock as _thread_allocate_lock
+        except ImportError:
+            from _dummy_thread import allocate_lock as _thread_allocate_lock
+
 cdef int64_t iNaT = util.get_nat()
 
 
@@ -53,6 +65,9 @@ def get_value_box(arr: ndarray, loc: object) -> object:
 # Don't populate hash tables in monotonic indexes larger than this
 _SIZE_CUTOFF = 1000000
 
+# Used in _ensure_mapping_populated to ensure is_unique behaves correctly
+# in multi-threaded code, see gh-21150
+_mapping_populated_lock = _thread_allocate_lock()
 
 cdef class IndexEngine:
 
@@ -236,17 +251,17 @@ cdef class IndexEngine:
 
     cdef inline _ensure_mapping_populated(self):
         # this populates the mapping
-        # if its not already populated
+        # if it is not already populated
         # also satisfies the need_unique_check
 
-        if not self.is_mapping_populated:
+        with _mapping_populated_lock:
+            if not self.is_mapping_populated:
+                values = self._get_index_values()
+                self.mapping = self._make_hash_table(len(values))
+                self._call_map_locations(values)
 
-            values = self._get_index_values()
-            self.mapping = self._make_hash_table(len(values))
-            self._call_map_locations(values)
-
-            if len(self.mapping) == len(values):
-                self.unique = 1
+                if len(self.mapping) == len(values):
+                    self.unique = 1
 
         self.need_unique_check = 0
 
--- a/ci/azure-windows-27.yaml
+++ b/ci/azure-windows-27.yaml
@@ -6,6 +6,7 @@ dependencies:
   - beautifulsoup4
   - bottleneck
   - dateutil
+  - futures
   - gcsfs
   - html5lib
   - jinja2=2.8
--- a/ci/circle-27-compat.yaml
+++ b/ci/circle-27-compat.yaml
@@ -5,6 +5,7 @@ channels:
 dependencies:
   - bottleneck=1.0.0
   - cython=0.28.2
+  - futures
   - jinja2=2.8
   - numexpr=2.4.4 # we test that we correctly don't use an unsupported numexpr
   - numpy=1.9.3
--- a/ci/travis-27-locale.yaml
+++ b/ci/travis-27-locale.yaml
@@ -5,6 +5,7 @@ channels:
 dependencies:
   - bottleneck=1.0.0
   - cython=0.28.2
+  - futures
   - lxml
   - matplotlib=1.4.3
   - numpy=1.9.3
--- a/pandas/tests/indexes/test_base.py
+++ b/pandas/tests/indexes/test_base.py
@@ -33,6 +33,8 @@ from pandas.core.indexes.datetimes import _to_m8
 import pandas as pd
 from pandas._libs.tslib import Timestamp
 
+from concurrent.futures import ThreadPoolExecutor
+
 
 class TestIndex(Base):
     _holder = Index
@@ -2507,6 +2509,21 @@ class TestIndexUtils(object):
     def test_ensure_index_from_sequences(self, data, names, expected):
         result = ensure_index_from_sequences(data, names)
         tm.assert_index_equal(result, expected)
+
+
+class TestThreadSafety(object):
+
+    @pytest.mark.slow
+    @pytest.mark.parametrize('execution_number', range(7))
+    def test_isunique(self, execution_number):
+        """This test is executed seven times, each time it uses a pool of
+        two threads to run a test that is very likely to fail without the
+        fix for gh-21150. It is not a deterministic test, as there is
+        still a chance it will pass even though the bug exists. But
+        with the fix, it must always work with not issues."""
+        x = pd.date_range('2001', '2020')
+        with ThreadPoolExecutor(2) as p:
+            assert all(p.map(lambda x: x.is_unique, [x] * 2))
 
 
 @pytest.mark.parametrize('opname', ['eq', 'ne', 'le', 'lt', 'ge', 'gt',


jbrockmendel:
Test file couldn’t import in py27.  Is there a backport of ThreadpoolExecutor?

For the most part pandas doesn’t promise threadsafety.  Is there a reason non-safety in this spot is particularly problematic?

kinow:
@jbrockmendel thanks for reviewing it so quickly. 

>Test file couldn’t import in py27. Is there a backport of ThreadpoolExecutor?

Oh, it's not included with py27. I believe it's included in py3, but for py27 we'd have to have something like `pip install futures`.

>For the most part pandas doesn’t promise threadsafety. Is there a reason non-safety in this spot is particularly problematic?

Only if users are using indexes in a multi threaded environment. Though, I believe we could suggest as workaround to ask users to put the lock before they call the index.

I'd be happy to close the PR and comment in the #21150 and go with that approach. At least I learned a few things while reading the issue & the code (-:

pandas/tests/indexes/test_threadsafety.py
```diff
@@ -0,0 +1,24 @@
...
+    @pytest.mark.parametrize('execution_number', range(7))
+    def test_isunique(self, execution_number):
+        """This test is executed seven times, each time it uses a pool of
+        two threads to run a test that is very likely to fail without the
+        fix for gh-21150. It is not a deterministic test, as there is
+        still a chance it will pass even though the bug exists. But
+        with the fix, it must always work with not issues."""
+        x = pd.date_range('2001', '2020')
+        with ThreadPoolExecutor(2) as p:
+            assert all(p.map(lambda x: x.is_unique, [x] * 2))
```
gfyoung: Hmm...I think we can incorporate this test into one of the existing files in this directory.

pandas/_libs/index.pyx
```diff
@@ -26,6 +26,18 @@ from pandas._libs import algos, hashtable as _hash
...
+    from thread import allocate_lock as _thread_allocate_lock
+except:
+    try:
+        from _thread import allocate_lock as _thread_allocate_lock
+    except:
+        try:
+            from dummy_thread import allocate_lock as _thread_allocate_lock
+        except:
+            from _dummy_thread import allocate_lock as _thread_allocate_lock
+
```
gfyoung: Is it possible to move this into `pandas/compat/__init__.py` and import from there?

gfyoung:
> For the most part pandas doesn’t promise threadsafety. Is there a reason non-safety in this spot is particularly problematic?

@jbrockmendel : We don't promise thread-safety, but IMO we should try to improve it if possible. 🙂

gfyoung:
@kinow : Going to also need a `whatsnew` entry for `0.24.0`.

pandas/_libs/index.pyx
jbrockmendel: why not just import it from tslibs.strptime where it is already?

gfyoung: In that case, we should move that import code to `pandas/compat` so it can be utilized by both modules (in more semantically logical arrangement IMO).

pandas/tests/indexes/test_threadsafety.py
kinow: Which one you think would accommodate this one? Happy to update the pull request with this change :-)

gfyoung: `test_base.py` should suffice.

kinow: Done!

kinow:
@gfyoung thanks for the thorough review. Here's a summary of my changes, but before, just a heads-up that I added a few more commits, but am happy to squash them all before merging.

* moved thread/allocate_lock imports to `pandas/compat/__init__.py`. I had one issue later, because I was renaming the builtin module `_thread` (Py3 in my env) to `thread`. Then trying to `from thread import allocate_lock`. I believe it's because the way Python treats builtins in a special way? I think the current import in the compat module should work, with the cons that in case we need more classes/methods from the `thread` module, we'd have to keep adding them to the import list
* took liberty of fixing `strptime` as well, which was mentioned earlier in this ticket, but happy to revert in case we don't want to mix the changes here, and instead submit another PR later for that?
* added `futures` to a few files under `ci/` folder, which I believe are used by the CI tools to test Pandas. With this, the Travis-CI error from before, related to not finding `ThreadpoolExecutor` should be gone (hopefully, it's still running as I write this)
* moved the unit test from its own class, to under test_base.py, still keeping it in its isolated class. So that maybe we can add more in the future if necessary? Otherwise, happy to move it to either `TestIndexUtils` or `TestIndex` (two other classes there).

Thanks
Bruno

jreback (changes requested):
i will have a look
but we make no guarantee on threading at all
so not inclined to add this

kinow:
@jreback 

>i will have a look
>but we make no guarantee on threading at all
>so not inclined to add this

(replying here)

Not a problem, no hard feelings :-) as I said, it was fun and I learned while working on this, so even if this PR is closed and not merged, that's still all right (and I'd only have to thanks everybody who reviewed/replied here too).

jbrockmendel:
> We don't promise thread-safety, but IMO we should try to improve it if possible.

Sure.  My point was that if we are going to start addressing thread-safety, we should triage.  I really have no idea if this should be at the top of the list, hence asking the OP.

Also curious how nicely locks play with cython code.

jbrockmendel:
> took liberty of fixing strptime as well, which was mentioned earlier in this ticket, but happy to revert in case we don't want to mix the changes here

Definitely revert.  A lot of time and effort has gone into getting the dependency structure of those modules right.  Adding a dependency on `pandas.compat` is not part of the gameplan.  (cc @gfyoung)

jreback:
so is it possible to use he cython / c-python API calls for the locks in strptime.pyx?

i agree getting this with a good foundation would be a nice start

gfyoung:
> Adding a dependency on pandas.compat is not part of the gameplan. (cc @gfyoung)

@jbrockmendel : So where do you want to put the import code then?  It should be moved if we're going to have it in two places in the Cython code.

gfyoung:
> but we make no guarantee on threading at all
so not inclined to add this

@jreback : Is there a reason to not make something thread safe if possible?

jreback:
I am not against threadsafeing at all. But say pandas is thread-safe is a huge thing, and so saying 'part' is threadsafe leads one to believe 'all'. so have to be really careful about this. Further the performance might (or might not be an issue).

jbrockmendel:
> Where do you want to put the import code then?

The import in strptime needs to stay where it is.  If another module needs it, get it from there.  Or in pandas.compat import it from strptime, then in index.pyx import it from pandas.compat

gfyoung:
> The import in strptime needs to stay where it is. If another module needs it, get it from there.

* Why does it need to stay there?
* I don't follow why we should make other code depond `strptime`, such as `pandas.compat`.  Semantically, that's confusing me a lot.  I should also point out that we do use `compat` in order Cython files like `parsers.pyx`.

jbrockmendel:
> Why does it need to stay there?

In order to import `Timestamp`, `Timedelta`, ... into `pandas._libs.__init__`, we need `_libs.tslibs` to have no import-time dependencies on non-`_libs` pandas.

jbrockmendel:
> Proposal: `_libs.compat` ?

There may be merit to this idea, but it would be for the _actual_ compat code that is currently in _libs/src/headers/ and src/compat_helper.h

Please leave strptime alone.  A lot of effort has gone into making tslibs logically independent for build/testing purposes.

kinow:
Updated pull request, now wait for Travis CI to confirm I didn't forget anything (test_base.py passing locally). Rebased and squashed commits too.

I'd be ok with leaving this PR as it is, and instead working on a broader ticket to define a plan to support multi-thread in Pandas - assuming there's a consensus to slowly introduce it, as I don't think it can happen in one-go.

Otherwise, I'm also keen to help documenting what can be used in a multi-threaded environment, and what cannot. We could simply add a few docstring comments, stating whether a class is thread-safe or not, or giving more info about how to use it in a multi-threaded environment. Normally that's a convention that I am used to in the [Apache Software Foundation with Java](https://github.com/apache/commons-collections/blob/ea305b08516979ff3e5688cd5268e8f7a7a1234e/src/main/java/org/apache/commons/collections4/ArrayStack.java#L25).

That way tickets like this could be closed in favour of good documentation. Happy to help working on either of these two, or even a third option. Because as [Captain Kirk said](https://www.youtube.com/watch?v=2S-0vIBIp4Y&feature=youtu.be&t=3m49s), if really necessary there's always a third alternative :nerd_face:

jreback:
@kinow I think we would welcome better multi thread support with a few caveats

- must have more benchmarks / in particular we don’t want to slow down single threaded too much
- create a central place to handle the code as much as possible (eg the lock cohde)
- most of this would be cython code - so need to support the idiomatic expressions
- ideally identify 2 or more issues so that we can have a generalized plan rather than one offs

the last point is the key one

jreback:
can you merge master and update?

kinow:
Sure thing @jreback. Done :tada:

ci/travis-27.yaml
```diff
@@ -8,6 +8,7 @@ dependencies:
   - cython=0.28.2
   - fastparquet
   - feather-format
+  - futures
```
jreback: is this just for testing?

pandas/_libs/index.pyx
TomAugspurger: Can you address this? We also recently fixed all our bare excepts.

ci/travis-27.yaml
kinow: Yes, in the test file `pandas/tests/indexes/test_base.py` we copied the code from the issue reported in GitHub that imports a `ThreadPoolExecutor`, available in Python 2.7 only with `futures` package. 

```python
from concurrent.futures import ThreadPoolExecutor
```

TomAugspurger: This will I think cause failures for people who just install pandas and run `pandas.test`.

I'd do that in a pytest `importorskip`.

pandas/_libs/index.pyx
TomAugspurger: That sounds fine, I didn't read through all the comments :)

jreback:

Issue #8554, opened 2014-10-14
Inconsistent Timestamp arithmetic


willpan:
Subtracting a scalar datetime from a series returns timedeltas, but doesn't do the same for a dataframe.

```
import pandas as pd 
series = pd.Series(pd.date_range('2000-1-1', periods=100), index=range(0, 100))
df = pd.DataFrame(series)
date = pd.Timestamp('1995-1-1')

series - date
#0    1826 days                                                                                 
#1    1827 days                                                                                 
#2    1828 days
# ....   

df - date
#             0
#0  1975-01-01                                                                                  
#1  1975-01-02                                                                                  
#2  1975-01-03
# ....
```

It seems like both should return timedeltas.

jbrockmendel:
Looked into this briefly, threw up my hands somewhere between `core.internals` and `core.ops`.  Is the spaghetti in `ops` carefully optimized or the result of years of frankensteining?

jreback:
> Looked into this briefly, threw up my hands somewhere between core.internals and core.ops. Is the spaghetti in ops carefully optimized or the result of years of frankensteining?

the code in ``ops`` is mostly to handle multiple dispatch (before it became a thing)

Issue #12437, opened 2016-02-24
BUG: datetimelike subtract incorrect when broadcasting


toobaz:
```
In [2]: base, step = 10**15, 10**14

In [3]: data = range(base, base+3*step, step)

In [4]: df = pd.DataFrame({i : data for i in range(3)}).astype('datetime64[ns]')

In [5]: s = pd.Series(data).astype('datetime64[ns]')

In [6]: df
Out[6]: 
                    0                   1                   2
0 1970-01-12 13:46:40 1970-01-12 13:46:40 1970-01-12 13:46:40
1 1970-01-13 17:33:20 1970-01-13 17:33:20 1970-01-13 17:33:20
2 1970-01-14 21:20:00 1970-01-14 21:20:00 1970-01-14 21:20:00

In [7]: (df - s)[0]
Out[7]: 
0   1970-01-01 00:00:00
1   1970-01-02 03:46:40
2   1970-01-03 07:33:20
Name: 0, dtype: datetime64[ns]

In [8]: (df[0] - s[0])
Out[8]: 
0   0 days 00:00:00
1   1 days 03:46:40
2   2 days 07:33:20
Name: 0, dtype: timedelta64[ns]
```

I would expect the two to give the same result (`timedelta64[ns]`).

```
In [9]: pd.show_versions()

INSTALLED VERSIONS
------------------
commit: None
python: 2.7.10.final.0
python-bits: 64
OS: Linux
OS-release: 4.3.0-1-amd64
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: it_IT.utf8

pandas: 0.18.0rc1+35.g4048193
nose: 1.3.6
pip: 1.5.6
setuptools: 18.4
Cython: 0.23.2
numpy: 1.10.4
scipy: 0.16.0
statsmodels: 0.8.0.dev0+755fa81
xarray: None
IPython: 2.4.1
sphinx: 1.3.1
patsy: 0.3.0-dev
dateutil: 2.2
pytz: 2012c
blosc: None
bottleneck: None
tables: 3.2.2
numexpr: 2.4.3
matplotlib: 1.5.0rc2
openpyxl: None
xlrd: 0.9.4
xlwt: 0.7.5
xlsxwriter: 0.7.3
lxml: None
bs4: 4.4.0
html5lib: 0.999
httplib2: 0.9.1
apiclient: None
sqlalchemy: 1.0.11
pymysql: None
[...]

PR #22008, opened 2018-07-21
core: try coerce result back to DatetimeBlock

### LINKED ISSUE #21610: dtype changed when DataFrame add DateOffset

```python
pd.DataFrame([pd.to_datetime('2018')]).dtypes
# 0    datetime64[ns]
# dtype: object
(pd.DataFrame([pd.to_datetime('2018')]) + pd.DateOffset(years=1)).dtypes
# 0    object
# dtype: object
```
#### Problem description
`Timestamp` + `DateOffset` = `Timestamp`, but `DataFrame(datetime64[ns])` + `DateOffset` = `DataFrame(object)`.

#### Expected Output
`DataFrame(datetime64[ns])` + `DateOffset` = `DataFrame(datetime64[ns])`. Don't change dtype.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Darwin
OS-release: 17.6.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: zh_CN.UTF-8
LOCALE: zh_CN.UTF-8

pandas: 0.23.0
pytest: None
pip: 10.0.1
setuptools: 39.2.0
Cython: None
numpy: 1.14.5
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: None
sphinx: 1.6.2
patsy: None
dateutil: 2.7.2
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 2.2.2
openpyxl: None
xlrd: 1.0.0
xlwt: None
xlsxwriter: None
lxml: 3.8.0
bs4: 4.5.3
html5lib: 1.0.1
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

holymonson:
- [x] closes #21610 #8554 #12437
- [x] tests added / passed
- [x] whatsnew entry
---
Hang up until `Block.eval` reformed.

### DIFF
--- a/pandas/core/internals/__init__.py
+++ b/pandas/core/internals/__init__.py
@@ -63,7 +63,9 @@ from pandas.core.dtypes.generic import (
     ABCSeries,
     ABCDatetimeIndex,
     ABCExtensionArray,
-    ABCIndexClass)
+    ABCIndexClass,
+    ABCDateOffset,
+)
 import pandas.core.common as com
 import pandas.core.algorithms as algos
 
@@ -2785,11 +2787,16 @@ class DatetimeBlock(DatetimeLikeBlockMixin, Block):
 
         self.values[locs] = values
 
-    def eval(self, try_cast=False, *args, **kwargs):
-        blocks = super().eval(try_cast=try_cast, *args, **kwargs)
+    def eval(self, func, other, try_cast=False, **kwargs):
+        block = super().eval(func, other, try_cast=try_cast, **kwargs)[0]
         if try_cast:
-            blocks = [self._try_coerce_result(block) for block in blocks]
-        return blocks
+            if isinstance(other,
+                          (tslibs.Timestamp, np.datetime64, datetime, date)):
+                block = TimeDeltaBlock(block.values, block.mgr_locs,
+                                       ndim=block.ndim)
+            elif isinstance(other, ABCDateOffset):
+                block = self._try_coerce_result(block)
+        return [block]
 
 
 class DatetimeTZBlock(NonConsolidatableMixIn, DatetimeBlock):
--- a/pandas/core/internals/__init__.py
+++ b/pandas/core/internals/__init__.py
@@ -2737,7 +2737,7 @@ class DatetimeBlock(DatetimeLikeBlockMixin, Block):
 
     def _try_coerce_result(self, result):
         """ reverse of try_coerce_args """
-        if isinstance(result, np.ndarray):
+        if isinstance(result, (np.ndarray, Block)):
             if result.dtype.kind in ['i', 'f', 'O']:
                 try:
                     result = result.astype('M8[ns]')
@@ -2784,6 +2784,12 @@ class DatetimeBlock(DatetimeLikeBlockMixin, Block):
             values = conversion.ensure_datetime64ns(values)
 
         self.values[locs] = values
+
+    def eval(self, try_cast=False, *args, **kwargs):
+        blocks = super().eval(try_cast=try_cast, *args, **kwargs)
+        if try_cast:
+            blocks = [self._try_coerce_result(block) for block in blocks]
+        return blocks
 
 
 class DatetimeTZBlock(NonConsolidatableMixIn, DatetimeBlock):
@@ -2920,6 +2926,8 @@ class DatetimeTZBlock(NonConsolidatableMixIn, DatetimeBlock):
         if isinstance(result, np.ndarray):
             if result.dtype.kind in ['i', 'f', 'O']:
                 result = result.astype('M8[ns]')
+        elif isinstance(result, Block):
+            result = self.make_block_same_class(result.values.flat)
         elif isinstance(result, (np.integer, np.float, np.datetime64)):
             result = tslibs.Timestamp(result, tz=self.values.tz)
         if isinstance(result, np.ndarray):
--- a/pandas/core/internals/__init__.py
+++ b/pandas/core/internals/__init__.py
@@ -2788,10 +2788,10 @@ class DatetimeBlock(DatetimeLikeBlockMixin, Block):
         self.values[locs] = values
 
     def eval(self, func, other, try_cast=False, **kwargs):
-        block = super().eval(func, other, try_cast=try_cast, **kwargs)[0]
+        block = super(DatetimeBlock, self).eval(func, other, try_cast=try_cast,
+                                                **kwargs)[0]
         if try_cast:
-            if isinstance(other,
-                          (tslibs.Timestamp, np.datetime64, datetime, date)):
+            if isinstance(other, (np.datetime64, date)):
                 block = TimeDeltaBlock(block.values, block.mgr_locs,
                                        ndim=block.ndim)
             elif isinstance(other, ABCDateOffset):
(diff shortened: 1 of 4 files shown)


WillAyd:
@jbrockmendel . Should this work with tz-aware date times as well?

pandas/core/internals.py
```diff
@@ -1362,10 +1368,15 @@ def eval(self, func, other, errors='raise', try_cast=False, mgr=None):
             values, values_mask, other, other_mask = self._try_coerce_args(
                 transf(values), other)
         except TypeError:
+            # `Timestamp` operate with `DateOffset` must cast to `object`,
```
jreback: i would rather see you override eval in the appropriate blocks

jreback (changes requested):

holymonson:
v2.0, override `DatetimeBlock.eval`.

Below test fails, but not a regression. Will look deeper later.
```python
    def test_timestamp_df_sub_timestamp(self):
        expected = pd.DataFrame([pd.Timedelta('365d')])
        result = pd.DataFrame([pd.Timestamp('2018')]) - pd.Timestamp('2017')
        tm.assert_frame_equal(expected, result)
```

pandas/core/internals/__init__.py
```diff
@@ -2785,6 +2787,17 @@ def set(self, locs, values, check=False):
 
         self.values[locs] = values
 
+    def eval(self, func, other, try_cast=False, **kwargs):
+        block = super().eval(func, other, try_cast=try_cast, **kwargs)[0]
+        if try_cast:
+            if isinstance(other,
+                          (tslibs.Timestamp, np.datetime64, datetime, date)):
```
holymonson: Is there an ABCTimestamp to check? At least 3 times use similar `isinstance()`.

jbrockmendel: > Is there an ABCTimestamp to check?

No, but Timestamp subclasses datetime, so could actually be removed from this check.

pandas/core/internals/__init__.py
```diff
@@ -2785,6 +2787,17 @@ def set(self, locs, values, check=False):
 
         self.values[locs] = values
 
+    def eval(self, func, other, try_cast=False, **kwargs):
+        block = super(DatetimeBlock, self).eval(func, other, try_cast=try_cast,
```
jbrockmendel: What happens here if instead of calling super eval you delegate to self.holder?

holymonson: In `DatetimeBlock - datetime` case and `DatetimeBlock + DateOffset` case, it require the result block changing (`DatetimeBlock` -> `TimeDeltaBlock `, `ObjectBlock` -> `DatetimeBlock` or `DatetimeTZBlock`). 
Delegating to `self.holder` couldn't change block type. Besides, `self._holder` is unaware of `other`.

In fact, what should be blamed is the last line in `Block.eval` carelessly make a wrong block. A proper fix should be 
1. spliting a method to make block regarding to `other`.
2. override `make_block` with `other` as argument. But this will lead to `self.make_block` returning another type of block.
https://github.com/pandas-dev/pandas/blob/0828c25913b5862d70322471a6ee4fa11d8f890d/pandas/core/internals/__init__.py#L1456

jbrockmendel: > In fact, what should be blamed is the last line in Block.eval carelessly make a wrong block. A proper fix should be

There has been an effort recently to fix mismatches between the behaviors of arithmetic ops in Index vs Series vs DataFrame.  For the timedelta and datetime cases, the canonical behavior is defined in TimedeltaIndex/DatetimeIndex.

The _best_ case here to to avoid making this fix in `core.internals` but instead do it in core.DataFrame (or core.ops).  The next base case is to have the FooBlock dispatch to the appropriate Index subclass to ensure the behavior is canonical.

holymonson: That will be great, but it may need a big rewrite of `eval` and `_try_coerce_args`, cause now `values` is all coerced to `ndarray[i8]` for `DatetimeLike`. (May be benefited from numpy bi-ops, but confused to convert back.)

Is there an issue to xfer? This PR would depand on it.

holymonson:
Hang up until `Block.eval` reformed. This will also close #8554 #12437.

pandas/tests/internals/test_internals.py
```diff
@@ -1238,7 +1238,6 @@ class TestCanHoldElement(object):
         (2**63, 'complex128'),
         (True, 'bool'),
         (np.timedelta64(20, 'ns'), '<m8[ns]'),
-        (np.datetime64(20, 'ns'), '<M8[ns]'),
```
jbrockmendel: Why removed?  (Besides no is exempt sub working)?

pandas/tests/internals/test_internals.py
```diff
@@ -1249,15 +1248,16 @@ class TestCanHoldElement(object):
...
+        skip = {
+            (operator.add, 'bool'),
+            (operator.sub, 'bool'),
+            (operator.mul, 'bool'),
+            (operator.truediv, 'bool'),
+            (operator.mod, 'i8'),
+            (operator.mod, 'complex128'),
+            (operator.mod, '<m8[ns]'),
+            (operator.pow, 'bool'),
+        }
```
jbrockmendel: Pls avoid formatting changes where convenient, makes it harder for reviewers to identify the actual changes.

pandas/tests/internals/test_internals.py
holymonson: This PR lead to `datetime64 OP datetime64 = timedelta64`, regardless of `OP`. Although meaningless, it will result `datetime64 + datetime64 = timedelta64`. 

The assumption `datetime64 OP datetime64 = datetime64` is meaningless for all `OP`. If allow this line, it will require extra `OP` check in `DatetimeBlock.eval`, which I don't think worth.

pandas/core/internals/__init__.py
```diff
@@ -2737,7 +2739,7 @@ def _try_coerce_args(self, values, other):
 
     def _try_coerce_result(self, result):
         """ reverse of try_coerce_args """
-        if isinstance(result, np.ndarray):
+        if isinstance(result, (np.ndarray, Block)):
```
jreback: this is not a good change, please remove. the point is that these are scalars/arrays NEVER blocks.

pandas/core/internals/__init__.py
```diff
@@ -2920,6 +2933,8 @@ def _try_coerce_result(self, result):
         if isinstance(result, np.ndarray):
             if result.dtype.kind in ['i', 'f', 'O']:
                 result = result.astype('M8[ns]')
+        elif isinstance(result, Block):
```
jreback: same, NEVER do this

jreback (changes requested):
I have a lot of concerns over the changes in the impl here.

jbrockmendel:
Two of the bugs this addresses (#21610, #8554) were closed by #22163.  The remaining issue #12437 is still open.  @holymonson can this be simplified if only aimed at that issue?

holymonson:
> Two of the bugs this addresses (#21610, #8554) were closed by #22163. The remaining issue #12437 is still open. @holymonson can this be simplified if only aimed at that issue?

This `Block`-level fix is not good if dispatching to `Series` could already fix some of them. I would rather close this (Or you might want to take the testcases).

Issue #22004, opened 2018-07-20
Defaulting to_csv to infer compression


dhimmel:
This issue follows up on https://github.com/pandas-dev/pandas/pull/17900 by thanks @Dobatymo and @gfyoung with review from @jreback. https://github.com/pandas-dev/pandas/pull/17900 added an `'infer'` option to compression in `_get_handle`. The main user-facing benefit here is that `df.to_csv` will be able to infer compression just like `pandas.read_csv`. However, unlike `read_csv` the default value for compression is `None` rather than `'infer'`

Unfortunately, much of the convenience of `compression='infer'` is lost if you have to explicitly specify it. In summary, I think there is a major convenience to the following command to work and automatically perform gzip compression:

```python
df.to_csv('path.csv.gz')
```

## Compatibility assessment

Defaulting to infer would only affect users who are currently using paths with compression extensions but not actually compressing. That's pretty bad practice IMO. Hence, I'm in favor of breaking backwards compatibility and changing the default for compression to infer. It looks like this would go into the major release 0.24?

WillAyd:
I agree conceptually. Probably need to handle cases where this would potentially conflict with the `compression` argument. PRs welcome

dhimmel:
I am happy to open a PR. I think the solution will be as simple as changing the compression default to infer in:

https://github.com/pandas-dev/pandas/blob/322dbf4f812f47d4eaee5d8afa10e17acf6dbc67/pandas/core/frame.py#L1714-L1716

Looks like `to_pickle` already defaults to infer:

https://github.com/pandas-dev/pandas/blob/322dbf4f812f47d4eaee5d8afa10e17acf6dbc67/pandas/io/pickle.py#L11

`to_json` should also probably be switched to default to infer:

https://github.com/pandas-dev/pandas/blob/322dbf4f812f47d4eaee5d8afa10e17acf6dbc67/pandas/io/json/json.py#L29-L32

I don't think the other [`to_*` methods](https://github.com/pandas-dev/pandas/issues/15008) have a compression argument but I should double check.

PR #20341, opened 2018-03-14
Preliminary format refactor

shangyian:
The branch for the other PR was messed up, so I created a new one.

- [x] based on discussion here: https://github.com/pandas-dev/pandas/pull/20032#pullrequestreview-101917292
- [ ] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [ ] whatsnew entry

This PR splits the various formatters into different files so that it's easier to refactor a specific formatter.

PR #22011, opened 2018-07-21
Default to_* methods to compression='infer'

dhimmel:
- [x] closes https://github.com/pandas-dev/pandas/issues/22004
- [x] tests added / passed
- [x] whatsnew entry

This PR does the following:

- Update default compression for `to_csv`, `to_json`, and `to_pickle` methods to infer.
- Adds `test_compression_defaults_to_infer` to test that compression='infer' is default for the relevant to_* methods.
- Fixes a bug in CSVFormatter where setting `compression='infer'` with a file object would produce a RuntimeWarning.
- Adds documentation to `test_compression_warning` which can fail due to a pytest bug.
- Cleans up how the encoding argument in CSVFormatter is processed.
- Moves compression tests from `pandas/tests/test_common.py` to `pandas/tests/io/test_common.py`


pandas/io/json/json.py
```diff
@@ -28,7 +28,7 @@
 # interface to/from
 def to_json(path_or_buf, obj, orient=None, date_format='epoch',
             double_precision=10, force_ascii=True, date_unit='ms',
-            default_handler=None, lines=False, compression=None,
+            default_handler=None, lines=False, compression='infer',
```
dhimmel: Not sure where to update the to_json docs... didn't see a docstring in this function.

WillAyd (changes requested):
Needs tests

pandas/io/json/json.py
WillAyd: https://github.com/pandas-dev/pandas/blob/322dbf4f812f47d4eaee5d8afa10e17acf6dbc67/pandas/core/generic.py#L1905

dhimmel:
The following test is failing:

https://github.com/pandas-dev/pandas/blob/1033e8b1195d4071253889ada60523832285354c/pandas/tests/test_common.py#L254-L264

Let me look into https://github.com/pandas-dev/pandas/issues/21227 as to what this test is for.

dhimmel:
> Needs tests

@WillAyd I believe the different values for `compression` are already being tested, such that we don't need to test that infer works. Are you saying that I should be testing that changing the default actually changes the default? It seems like that could create an excessive amount of tests were we to test every default argument, but am happy to proceed as you see fit.

minggli:
@dhimmel saw your comment. thanks for contributing! this `test_compression_warning` expects a RunTimeWarning when a file handle is passed to `to_csv` method with compression kwarg is supplied.

This is because it's not supported as stated in:
https://github.com/pandas-dev/pandas/blob/1033e8b1195d4071253889ada60523832285354c/pandas/io/formats/csvs.py#L134

I don't see why it would fail your work on `to_json`.

WillAyd:
> Are you saying that I should be testing that changing the default actually changes the default

Yep. Don't need to overthink it with all the various parameter combinations but at least need a test to ensure this now defaults to infer (since that is what you are changing)

dhimmel:
> Are you working off the latest master branch?

This PR currently branch off at 322dbf4f812f47d4eaee5d8afa10e17acf6dbc67, which is now two commits behind. I can rebase, but don't think that's the issue.

> I don't see why it would fail your work on to_json.

I was thinking the change of the to_csv compression default may have caused this issue, but it doesn't make sense to me. The test explicitly specifies compression in `df.to_csv(f, compression=compression_only)`, so I don't see how my PR would effect `test_compression_warning`.

minggli:
@dhimmel , it appears that moving

There appear a related issue on pytest with Python 2.7.  https://github.com/pytest-dev/pytest/issues/2917

The warning may have been raised in earlier test cases so in `test_compression_warning` fails in Python 2.7, but it doesn't explain why the test passes on master with Python 2.7. It's strange.

minggli:
@dhimmel,

I think the change of default in to_csv did change things. there is no infer_compression procedure in to_csv. I think that's what caused the failing of test.

minggli:
#17900 added infer in _get_handle but to_csv uses compression argument before calling _get_handle; therefore raises RuntimeWarning at https://github.com/dhimmel/pandas/blob/648bf4d1810a2c2b9cbff1d4b941ab7cb7bc0b35/pandas/io/formats/csvs.py#L133 because compression='infer' instead of None as it was before. It on itself shouldn't be a problem but https://github.com/pytest-dev/pytest/issues/2917 exists so that test_compression_warning fails because the warning has been raised in earlier test before this case. simply changing the order of the test is not thread-safe so I think the best way is:

inside to_csv:
compression = _infer_compression(path_or_buf, compression)
or inside csvs
self.compression = _infer_compression(path_or_buf, compression)

In regards to zip in docstring, it's what master is showing right and supported already.
```
import os
import pandas as pd

a = pd.DataFrame(10000 * [[123, 234, 435]], columns=['A', 'B', 'C'])
a.to_csv('test_compressed', index=False, compression='zip')

b = pd.read_csv('test_compressed', compression='zip')

assert a.equals(b)
os.remove('test_compressed')
```

dhimmel:
So I'm a bit concerned with what has happened since I last worked on compression & file IO. My impression was that we should start to use a unified API for inferring compression and opening files: see https://github.com/pandas-dev/pandas/issues/15008. `_get_handle` now delegates to `_infer_compression` . It seems however that `CSVFormatter.save` has now gotten further away from using the unified API. In https://github.com/pandas-dev/pandas/pull/22011/commits/be724fa7edcb80056cf0bdbc94351ca39d8f3d23 I replace `CSVFormatter.save` with the simplified workflow. If writing to zip is a feature that we want to support, then shouldn't this go in _get_handle?

jreback:
> _get_handle now delegates to _infer_compression . 

why is this a problem? this is clear separation of concerns. I don't think the goal has drifted from when you last worked on this. A number of bugs / consolidations have happened in the interim. Happy to take further cleanup.

dhimmel:
> why is this a problem?

Having `_get_handle` call `_infer_compression` is good. It simplifies the number of functions that individual `to_*` methods have to call. However, the issue with `CSVFormatter.save` is that performs a custom workflow for handling zip files, which has made the code quite complicated with several lines calling `_get_handle`. The issue came up because the failing test is part of this custom zip workflow CSVFormatter now contains. However, I apologize for getting a bit sidetracked. I should keep this PR focused on changing the default for compression to infer.

dhimmel:
# Why is `test_compression_warning` failing?

There is a [testing failure](https://travis-ci.org/pandas-dev/pandas/jobs/407342195#L3855) that has completely stumped me. It is only occurring in some Python 2 builds. The failing test is:

https://github.com/pandas-dev/pandas/blob/1033e8b1195d4071253889ada60523832285354c/pandas/tests/test_common.py#L254-L264

The error is:

```
E               AssertionError: Did not see expected warning of class 'RuntimeWarning'.
```

It doesn't make sense to me how the changes in this PR would effect whether the `RuntimeWarning`. This PR only changes the default compression value. The test specifies `compression` so the default should not matter.

In https://github.com/pandas-dev/pandas/pull/22011/commits/cebc0d98bd4ec7f68d3521f2814c0f41de818312 I added some debugging statements to help diagnose the testing failure

In CSVFormatter.save:

> warning-debug

In `test_compression_warning`:

> test-debug-changes

Here's the output:

```
------------------------------ Captured log call -------------------------------
test_common.py             261 WARNING  debug_1: gzip
test_common.py             266 WARNING  debug_2: gzip
csvs.py                    138 WARNING  debug_3: gzip
csvs.py                    139 WARNING  debug_4: <gzip open file '/tmp/tmpQYAr5L', mode 'wb' at 0x7f7c02e56db0 0x7f7c04698b10>
csvs.py                    141 WARNING  debug_5: True
csvs.py                    143 WARNING  debug_6: in loop, should RuntimeWarn
```

So the logging shows in the failing test that the code enters the if statement that triggers the RuntimeWarning. So I don't understand how the warning could not be raised or where this issue could be coming from.

@WillAyd other than this issue how do things look? I added a test for `to_csv` defaulting to inference.

pandas/tests/io/formats/test_to_csv.py
```diff
@@ -351,3 +352,15 @@ def test_to_csv_compression(self, compression_only,
             result = pd.read_csv(path, index_col=0,
                                  compression=read_compression)
             tm.assert_frame_equal(result, df)
+
+    def test_compression_defaults_to_infer(tmpdir):
+        """
+        Test that to_csv defaults to inferring compression from paths.
+        https://github.com/pandas-dev/pandas/pull/22011
+        """
+        df = DataFrame({"A": [1]})
```
WillAyd: While I suppose this "works" it is definitely focused on `gzip` and doesn't make any assertions about how the behavior works across a combination of `infer`, `None` and an explicit compression.

We have a fixture called `compression` in `conftest.py` that covers the various compression options - is there a way to parametrize that as part of this test and make more comprehensive assertions about the potential combinations of behavior?

WillAyd (changes requested):
Generally looks OK, though I think the test can be improved. Will have to take a look again after some of the logging things get cleaned up

WillAyd:
@minggli any thoughts on the issue @dhimmel is describing above?

dhimmel:
Thanks @minggli for tracking down the [pytest issue](https://github.com/pytest-dev/pytest/issues/2917) titled "asserting repeated warning does not work in Python 2.7". I think it's very likely this is issue, however, I'm still not sure where the first RuntimeWarning is occurring that is causing pytest to malfunction when detecting the second. An easy solution would be to skip `test_compression_warning` on Python 2. However, I'd like to know where in the tests changing the default for compression to infer is causing a RuntimeWarning, in case we should just fix that test.

pandas/tests/io/formats/test_to_csv.py
dhimmel: The possible values that compression can take for `to_csv` are already tested. The purpose of this test is simply to test that `compression` is defaulting to `'infer'`. I kept the test simple so that it ONLY tests that the default is infer and won't break or malfunction should other aspects of the API change. The test will break if the default for compression is put back to None (and hopefully not for other reasons).

I think it may make sense to make a similar test for `to_json`, but don't see how testing `compression='infer'` for all possible compression extensions is within scope of this PR as this PR just changes the default and is not creating the infer option. Is the worry that there is currently inadequate testing?

minggli:
@dhimmel ,

to_csv should not raise RuntimeWarning when user didn't explicitly specify a compression and pass a file-like object, so the tests work fine. In other words, now that there is 'infer' as default, the RuntimeWarning raises all the time, which is undesirable, and should only raise when user passes explicitly and a file-handle is given.

So, I feel fixing the test is wrong way about it (the test affected by default 'infer' is test_compression_size_fh) and we should instead pass inferred compression keyword into CSV formatter or during construction of it.

dhimmel:
@shangyian I saw you're the author of https://github.com/pandas-dev/pandas/pull/20341. Was there a specific reason why the encoding processing is performed in `CSVFormatter.save` rather than `CSVFormatter.__init__`? Does the movement of encoding logic as of https://github.com/pandas-dev/pandas/pull/22011/commits/eb7f9b5491defe035231e87f7902f1ea8786e86d raise any concerns?

pandas/tests/io/formats/test_to_csv.py
dhimmel: > it should be that much more work to leverage the existing compression fixture for this test to increase coverage

In https://github.com/pandas-dev/pandas/pull/22011/commits/abd19e3b02c77e1160d460ab50b34d4642d568c2, I modified an existing parametrized test to look for compression by default for paths where inference should occur. This actually caught an issue (we hadn't switched default for `Series.to_csv` -- fixed in https://github.com/pandas-dev/pandas/pull/22011/commits/2f670fedbe1694e64096c2a775d8e83799f04036).

Should I delete the gzip test? Note the parametrized test doesn't test that the _right_ compression is occurring, just that a compression is occurring.

dhimmel:
@WillAyd Travis builds have passed, if you want to take another look. No outstanding items from my end.

pandas/tests/io/formats/test_to_csv.py
WillAyd: Instead of going about it in this fashion can you not do a round trip with `to_csv` and `read_csv` using infer with a file extension for the former and the parametrized value as an argument for `compression` in the latter? So in pseudo-code:

```python
with tm.ensure_clean('compressed.csv.{}'.format(compression_only)) as path:
    df.to_csv(path)
    result = pd.read_csv(path, compression=compression_only)

tm.assert_frame_equal(result, df)
```

pandas/tests/test_common.py
```diff
@@ -217,13 +220,23 @@ def test_standardize_mapping():
...
-        getattr(obj, method)(filename, compression=compression_only)
-        compressed = os.path.getsize(filename)
-        getattr(obj, method)(filename, compression=None)
-        uncompressed = os.path.getsize(filename)
-        assert uncompressed > compressed
+    """
+    Tests that compression is occurring by comparing to the bytes on disk of
+    the uncompressed file.
+    """
+    extension = _compression_to_extension[compression_only]
```
WillAyd: Would prefer to keep this test unchanged and ensure that the test above is responsible for `infer` behavior

pandas/tests/test_common.py
```diff
@@ -251,8 +263,18 @@ def test_compression_size_fh(obj, method, compression_only):
         assert uncompressed > compressed
 
 
-# GH 21227
 def test_compression_warning(compression_only):
+    """
```
WillAyd: Minor nit but I'd rather see this as comments instead of a docstring, just so it doesn't clutter any test failure output

WillAyd (changes requested):
Couple things to give a better separation of tests. Can you also update the `to_json` docs?

pandas/tests/test_common.py
dhimmel: Didn't realize docstrings get printed when the test fails... that could be nice in many cases.

Happy to change docstring to comment here. But were this test to fail, wouldn't it be helpful for this message to be outputted?

pandas/tests/io/formats/test_to_csv.py
dhimmel: I'm not a big fan of the roundtrip approach here, since it never actually tests that the file is compressed on disk. Given that the to_* and read_* methods rely on much of the same compression infrastructure, I think it's possible to modify the code such that all compression gets disabled and the roudtrip works perfectly. Now hopefully there are enough other tests to catch such a situation.

pandas/tests/test_common.py
WillAyd: I'm not aware of anywhere else where we do this so don't want to introduce any special cases for now. Could always be changed in the future in perhaps a separate, more comprehensive PR if there's a need for it and other test docstrings

pandas/tests/test_common.py
dhimmel: I think the paremetrization of this test is ideal. Currently this tests compression X to_method X dataframe_or_series. Luckily this exhaustive parametrization caught that I hadn't switched the default for Series.to_csv. Also note that the functional changes to this test are minimal (just adding getting the size of an inferred file). The rest of the changes help readability.

Were we to leave `test_compression_size` unmodified, I'd want to repeat the parametrization. Is there a dataframe_or_series fixture?

dhimmel (commented):
@WillAyd thanks for the feedback. Just a few questions/comments before I do the next round of changes. See my line comment responses, but read the following paragraph first.

I'm thinking that one fully parametrized test for compression inference by default is enough. This could either be the current modifications to `test_compression_size` or it could be a rewritten `test_compression_defaults_to_infer`. The rewrite would take the parametrization from `test_compression_size`, but would switch to either the round-trip or compression-size method (or even both) of testing compression.

What do you think?

WillAyd:
> I'm thinking that one fully parametrized test for compression inference by default is enough. This could either be the current modifications to test_compression_size or it could be a rewritten test_compression_defaults_to_infer

Your point is understood but we typically like to have small, very directed unit tests to keep a logical separation of features tested. As such, there should (or already is in fact) one test that ensures that files actually get compressed as part of the process. A separate test entity for "ensuring that the default compression is set to 'infer'" is what you are introducing with this change. 

The structure I was hinting at would keep those two "features" distinct in tests and give a better logical separation. Again the smaller and more directed tests can be, the better. There are plenty of older tests in our suite that try to do a lot of things at once - if you don't believe me take a look at them :-)

lmk if that's not clear

pandas/tests/test_common.py
dhimmel: Changed test docstrings to comments in https://github.com/pandas-dev/pandas/pull/22011/commits/6be808d87f8470b733534969397613784eb35a1c. For the record, I think it's a bit awkward to use comments for test function descriptions, especially since it's less clear whether the comment documents the whole function or just a single part.

dhimmel:
> A separate test entity for "ensuring that the default compression is set to 'infer'" is what you are introducing with this change.

Okay, the only test that this now changes (besides comments) is a new test `test_compression_defaults_to_infer`. It uses a round-trip approach to testing compression. This test is parametrized as follows:

```
@pytest.mark.parametrize('input', [
    DataFrame([[1.0, 0, -4.4],
               [3.4, 5, 2.4]], columns=['X', 'Y', 'Z']),
    Series([0, 1, 2, 4], name='X'),
])
@pytest.mark.parametrize('methods', [
    ('to_csv', pandas.read_csv),
    ('to_json', pandas.read_json),
    ('to_pickle', pandas.read_pickle),
])
```

@WillAyd do you agree that this is the minimal test to exhaustively test that supported `to_*` methods default to infer? I think it's important we test Series AND DataFrame as well as to_json, to_csv, AND to_pickle. Hence, I put the test in `test_common.py` rather than `test_to_csv.py`.

dhimmel:
Had to split test into two in https://github.com/pandas-dev/pandas/pull/22011/commits/fadb943851e395cfd6e7dbe05ba86d15ea2ee6f2: one for series and one for dataframes. Basically, the DataFrame & Series roundtrips require different args. Still a WIP, but the filesize test is considerably simpler given the intricacies of roundtripping.

dhimmel:
@WillAyd round-tripping a series to a CSV with to_csv and read_csv is not straightforward. There is the deprecated pandas.Series.from_csv, but it doesn't support compression. Take a look at the current PR... do you still think round-tripping is preferable to file size comparing, given the complexity of round-tripping? Is there an easier way to roundtrip the CSV?

shangyian:
@dhimmel - there wasn't a specific reason, so yeah, go ahead and move it if it makes more sense!

pandas/tests/test_common.py
```diff
@@ -251,8 +256,53 @@ def test_compression_size_fh(obj, method, compression_only):
...
+    ('to_pickle', {}, pandas.read_pickle),
+])
+def test_dataframe_compression_defaults_to_infer(
+        write_method, write_kwargs, read_method, compression_only):
+    # Test that DataFrame.to_* methods default to inferring compression from
+    # paths. https://github.com/pandas-dev/pandas/pull/22011
+    input = DataFrame([[1.0, 0, -4.4], [3.4, 5, 2.4]], columns=['X', 'Y', 'Z'])
+    extension = _compression_to_extension[compression_only]
+    with tm.ensure_clean('compressed' + extension) as path:
+        # assumes that compression='infer' is the default
```
WillAyd: If you can reduce the amount of comments that would be ideal. Citing this as a specific example it doesn't add anything that's not already known from the test name. 

Generally we just go `# GHXXXXX` as a comment at the beginning of the test to reference the original issue

WillAyd (changes requested):
Haven't been able to review the latest changes in detail but you might also want to take a look at #19715 to make sure there aren't any conflicts. 

As far as `read_csv` is concerned you can use the `squeeze` keyword to return a Series. I like your idea to keep these consolidated in `test_common` so see if that doesn't help

dhimmel:
@WillAyd ready for re-review. Tests passing as of https://github.com/pandas-dev/pandas/pull/22011/commits/14c3945410288a8edd3ee989d5582b1e5df0e992 and the following commit just modifies comments.

> you might also want to take a look at #19715 to make sure there aren't any conflicts.

There may be some minor git conflicts since it looks like @toobaz's commits move where the `Series.to_csv` code is located. However, all we touch in that code is the compression default and docstring, so it shouldn't be too bad for @toobaz to rebase on top of these changes (I didn't see a WIP PR with the changes).

toobaz:
> There may be some minor git conflicts since it looks like @toobaz's commits move where the Series.to_csv code is located: see toobaz@191d2ab. However, all we touch in that code is the compression default and docstring, so it shouldn't be too bad for @toobaz to rebase on top of these changes

Sure, no problem




> (I didn't see a WIP PR with his/her changes)

#21896

pandas/tests/test_common.py
```diff
@@ -154,8 +158,50 @@ def test_compression_size_fh(obj, method, compression_only):
         assert uncompressed > compressed
 
 
-# GH 21227
```
jreback: can you move all of the compression tests to pandas/tests/io/test_common.py (including existing ones), may want to create a new class to organzie them

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -177,7 +177,8 @@ Other Enhancements
 - :func:`read_html` copies cell data across ``colspan`` and ``rowspan``, and it treats all-``th`` table rows as headers if ``header`` kwarg is not given and there is no ``thead`` (:issue:`17054`)
 - :meth:`Series.nlargest`, :meth:`Series.nsmallest`, :meth:`DataFrame.nlargest`, and :meth:`DataFrame.nsmallest` now accept the value ``"all"`` for the ``keep`` argument. This keeps all ties for the nth largest/smallest value (:issue:`16818`)
 - :class:`IntervalIndex` has gained the :meth:`~IntervalIndex.set_closed` method to change the existing ``closed`` value (:issue:`21670`)
-- :func:`~DataFrame.to_csv` and :func:`~DataFrame.to_json` now support ``compression='infer'`` to infer compression based on filename (:issue:`15008`)
+- :func:`~DataFrame.to_csv`, :func:`~Series.to_csv`, :func:`~DataFrame.to_json`, and :func:`~Series.to_json` now support ``compression='infer'`` to infer compression based on filename (:issue:`15008`).
+  The default compression for ``to_csv``, ``to_json``, and ``to_pickle`` methods has been updated to ``'infer'`` (:issue:`22004`).
```
jreback: can you add a tiny bit more detail that this means to infer based on the file extension.

any updates needed in io.rst?

pandas/core/frame.py
```diff
@@ -1715,7 +1715,7 @@ def to_panel(self):
 
     def to_csv(self, path_or_buf=None, sep=",", na_rep='', float_format=None,
                columns=None, header=True, index=True, index_label=None,
-               mode='w', encoding=None, compression=None, quoting=None,
+               mode='w', encoding=None, compression='infer', quoting=None,
```
jreback: can you add versionchanged in each of the modified doc-strings

jreback (changes requested):

pandas/core/frame.py
dhimmel: Done in https://github.com/pandas-dev/pandas/pull/22011/commits/1ba8f3a281a71db33d8baa6a9a4c9fd7e40c625a

doc/source/whatsnew/v0.24.0.txt
dhimmel: > can you add a tiny bit more detail that this means to infer based on the file extension.

Done in https://github.com/pandas-dev/pandas/pull/22011/commits/1ba8f3a281a71db33d8baa6a9a4c9fd7e40c625a

> any updates needed in io.rst?

I looked through https://pandas.pydata.org/pandas-docs/version/0.23.3/io.html and only saw the need for one small update (a `versionchanged` note) added in https://github.com/pandas-dev/pandas/pull/22011/commits/24e051e4d147d341de86e1919c74960a1b50253f.

pandas/tests/test_common.py
dhimmel: Done in https://github.com/pandas-dev/pandas/pull/22011/commits/387d1d29f2833e236e0d8c3e3167c94614676973

Didn't create a new class as I didn't want to make the line-length a bigger issue than it already is.

dhimmel:
Tests passing as of https://github.com/pandas-dev/pandas/pull/22011/commits/e3a0f56ccbf9cb513b613de8d1d9c51793b066ff. @WillAyd and @jreback ready for re-review.

jreback (changes requested):
lgtm. ping on green

pandas/tests/io/test_common.py
```diff
@@ -2,18 +2,18 @@
...
 """
 import mmap
-import pytest
 import os
-from os.path import isabs
+import pytest
 
 import pandas as pd
-import pandas.util.testing as tm
+import pandas.io.common as cmn
```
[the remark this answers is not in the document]
dhimmel: Done in https://github.com/pandas-dev/pandas/pull/22011/commits/af8c137f2d9fd643b4581e2c470d4388843bae13

pandas/core/frame.py
```diff
@@ -1750,10 +1750,13 @@ def to_csv(self, path_or_buf=None, sep=",", na_rep='', float_format=None,
         encoding : string, optional
             A string representing the encoding to use in the output file,
             defaults to 'ascii' on Python 2 and 'utf-8' on Python 3.
-        compression : {'infer', 'gzip', 'bz2', 'zip', 'xz', None}, default None
+        compression : {'infer', 'gzip', 'bz2', 'zip', 'xz', None},
+                      default 'infer'
             If 'infer' and `path_or_buf` is path-like, then detect compression
             from the following extensions: '.gz', '.bz2', '.zip' or '.xz'
             (otherwise no compression).
+            .. versionchanged:: 0.24.0
```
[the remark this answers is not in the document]
dhimmel: Done in https://github.com/pandas-dev/pandas/pull/22011/commits/f8829a60a24a17451134014da27bd5f1f93a3fea, but would be good to hear from @TomAugspurger and @datapythonista, since we have complex situations such as:

```rst
DOCLINE
DOCLINE
.. versionchanged:: 0.23.0
   here is what was added
.. versionchanged:: 0.24.0 here is what changed

DOCLINE
```

For example, is the above OKAY or do we need additional blanks?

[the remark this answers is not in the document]
datapythonista: @dhimmel I think you need the additional blank lines (before, and not sure if after).

The reason is not that much an standard in this case, but about sphinx understanding the directive. What we expect in the documentation, is that it's rendered like in the `validate` method of the `merge` docstring: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.merge.html

But if you don't leave the right blank lines, sphinx doesn't detect it's a directive, and the text is rendered as it is. See this case: https://pandas.pydata.org/pandas-docs/version/0.23.1/generated/pandas.IntervalIndex.from_tuples.html

So, the best is if you can build the documentation, and check that it's rendered all right. This can be done by `./doc/make.py html` (or `./doc/make.py html --single pandas.DataFrame.read_csv`)

Let me know if you have any issue.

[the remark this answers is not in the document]
dhimmel: Updated in https://github.com/pandas-dev/pandas/pull/22011/commits/eadf68e4e23c4d6715347d3096e4b50e99baaa0e and built the docs locally to confirm they're rendering properly.

Turns out the blank line before is required. After is not required. In between multiple statements is not required.

jreback (approved):

jreback:
lgtm 

@WillAyd merge when satisfied

pandas/tests/io/test_compression.py
```diff
@@ -0,0 +1,101 @@
...
+        getattr(input, write_method)(path, **write_kwargs)
+        output = read_method(path, compression=compression_only, **read_kwargs)
+    tm.assert_series_equal(output, input, check_names=False)
+
+
+def test_compression_warning(compression_only):
+    # Assert that passing a file object to to_csv while explicitly specifying a
+    # compression protocol triggers a RuntimeWarning, as per GH 21227.
+    # Note that pytest has an issue that causes assert_produces_warning to fail
+    # in Python 2 if the warning has occurred in previous tests
```
WillAyd: May be overlooking it but where in the links is there mention about Python 2 behavior? We get some random Resource Warnings in our tests which is maybe related so could be good info, but wasn't immediately apparent to me where that was

WillAyd (changes requested):
Small nits / question around comments

pandas/tests/io/test_compression.py
```diff
@@ -0,0 +1,101 @@
...
+
+@pytest.mark.parametrize('write_method, write_kwargs, read_method', [
+    ('to_csv', {'index': False}, pd.read_csv),
+    ('to_json', {}, pd.read_json),
+    ('to_pickle', {}, pd.read_pickle),
+])
+def test_dataframe_compression_defaults_to_infer(
+        write_method, write_kwargs, read_method, compression_only):
+    # Test that DataFrame.to_* methods default to inferring compression from
+    # paths. GH 22004
```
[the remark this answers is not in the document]
dhimmel: Done in https://github.com/pandas-dev/pandas/pull/22011/commits/cf5b62e31bacbc4b27744f45a60a10c6a7902396

pandas/tests/io/test_compression.py
dhimmel: The pytest issue at https://git.io/fNEBm / https://github.com/pytest-dev/pytest/issues/2917.

WillAyd (approved):
lgtm I'll merge after AppVeyor goes green. Thanks @dhimmel

dhimmel:
Not sure what the [Travis failure is about](https://travis-ci.org/pandas-dev/pandas/jobs/410842506#L1936):

```
_____________________________ ERROR collecting gw0 _____________________________
Different tests were collected between gw1 and gw0. The difference is:
--- gw1
+++ gw0
```

gfyoung:
@dhimmel : Looks like all of our required builds are passing!

However, I would like to double check our non-required builds first before merging this (cc @WillAyd ).

dhimmel:
Just wanted to add that I [posted on Steem](https://steemit.com/utopian-io/@dhimmel/automatic-compression-when-saving-files-in-pandas-how-changing-a-default-parameter-ballooned-into-a-monster-pr) about this pull request, as part of the https://utopian.io initiative. Today, the utopian account upvoted my post, thereby rewarding it, as an incentive for open source contributions! So thanks Utopian and happy to help any other Pandas contributors get set up with a Steem account  and use Utopian, just email me.

PR #22014, opened 2018-07-21
[REF] separate blocks.py out of internals.__init__

jbrockmendel:
Follow-up to #21903


pandas/core/internals/blocks.py
```diff
@@ -0,0 +1,3417 @@
...
+        # NonConsolidatable blocks can have a single item only, so we return
+        # one block per item
+        unstacker = unstacker_func(self.values.T)
+        new_items = unstacker.get_new_columns()
+        new_placement = new_columns.get_indexer(new_items)
+        new_values, mask = unstacker.get_new_values()
+
+        mask = mask.any(0)
+        new_values = new_values.T[mask]
+        new_placement = new_p
```
jreback: this prob doesn't belong here (move in future)

jreback: this doesn't belong here), actually a lot of things done't belong here. but I suppose that's for followups.

jreback (approved):

jreback:
thanks, ok for now. This needs further splitting (and ultimately everything should be moved out of ``__init__`` to separate modules with minimal dependencies on each other. But I know thats in the plan.

pandas/tests/extension/test_external_block.py
```diff
@@ -5,7 +5,8 @@
 
 import pandas as pd
 from pandas.core.internals import (
-    BlockManager, SingleBlockManager, NonConsolidatableMixIn, Block)
+    BlockManager, SingleBlockManager)
+from pandas.core.internals.blocks import Block, NonConsolidatableMixIn
```
jorisvandenbossche: For next PRs doing further splitting, can we please keep the existing imports? 

What I mean is: import all the things that are used in pandas outside of internals itself in `internals/__init_.py`, then no changes to imports are needed, and it is also clear which functions/classes are used in the rest of pandas.

jbrockmendel: > For next PRs doing further splitting, can we please keep the existing imports?

Yes.  I'm also marking which non-internals modules are importing what, in the hopes we can trim down the exposed surface.

PR #22015, opened 2018-07-21
ENH: Number formatting support for excel styles

### LINKED ISSUE #22027: No way to control display format when exporting to excel

```python
# format call does nothing
df.style.format("{:.2%}").to_excel('styled.xlsx')

```
#### Problem description

The current excel styler doesn't implement anything for the styler format method. To get number format there are two possible options. First is to connect this method to something. Second is to allow some way to set the number_format attribute on excel cells. The second option is a bit cleaner IMHO since data conversion like truncating decimal places can result in loss of information and number_format is also sort of how excel declares data types eg. displaying a number as a time.

marknsikora:
- [x] closes #22027 
- [x] tests added / passed
- [x] whatsnew entry

Add number formatting support to the excel styles using a fake css entry. 

```css
number-format: 0%;
```

Added new tests, updated docs, and manually verified outputs using both excel output engines that support styling.

### DIFF
--- a/doc/source/style.ipynb
+++ b/doc/source/style.ipynb
@@ -985,7 +985,10 @@
     "- `vertical-align`\n",
     "- `white-space: nowrap`\n",
     "\n",
-    "Only CSS2 named colors and hex colors of the form `#rgb` or `#rrggbb` are currently supported."
+    "Only CSS2 named colors and hex colors of the form `#rgb` or `#rrggbb` are currently supported.\n",
+    "\n",
+    "The following pseudo CSS properties are also available to set excel specific style properties:\n",
+    "- `number-format`\n"
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


WillAyd:
Haven't reviewed code in detail yet but does this reference an existing issue? If not can you open and add reference in whatsnew?

marknsikora:
Done

WillAyd (changes requested):
cc @gfyoung for input.

May just be my lack of familiarity with styling but what is the preferred way of using this from an end user perspective? `.apply` with a UDF? Might be worth adding an example to the style notebook to illustrate

gfyoung:
@marknsikora : This looks good at the moment.  Could you add `whatsnew` entry for `0.24.0` ?

TomAugspurger:
Thanks @marknsikora .

PR #22016, opened 2018-07-21
REF: move range-generation functions to EA mixin classes

jbrockmendel:
With the DatetimeArray range-generating functions moved, we are finally able to move `shift`, and in turn `__add__`, `__sub__`, etc.  

Two non-trivial changes made during the move process, see inline comments.

Upcoming commits will add docstrings and port tests.  ATM many tests are a PITA bc the Array contructors don't know how to handle lists of strings.  Trying to find a way to implement that in `__new__` without having it become a total mess.


pandas/core/arrays/datetimes.py
```diff
@@ -194,6 +201,117 @@ def __new__(cls, values, freq=None, tz=None):
...
+                    if start is not None:
+                        start = start.tz_localize(tz).asm8
+                    if end is not None:
+                        end = end.tz_localize(tz).asm8
+            else:
+                # Create a linearly spaced date_range in local time
+                start = start.tz_localize(tz)
+                end = end.tz_localize(tz)
+                arr = np.linspace(start.value, end.value, periods)
+                index = cls._simple_new(arr.astype('M8[ns]'), freq=None, tz=tz)
```
jbrockmendel: One of two things changed from the existing version.  ATM this is:

```
                index = tools.to_datetime(np.linspace(start.value,
                                                      end.value, periods),
                                          utc=True)
                index = index.tz_convert(tz)
```

Changed to avoid the dependency on `tools`.

pandas/core/arrays/datetimes.py
```diff
@@ -1085,3 +1203,91 @@ def to_julian_date(self):
...
+            tz = end.tz
+            end = end.to_pydatetime()
+
+        xdr = generate_range(start=start, end=end,
+                             periods=periods, offset=freq)
+
+        dates = list(xdr)
+        # utc = len(dates) > 0 and dates[0].tzinfo is not None
+        values = np.array([x.value for x in dates])
+        data = cls._simple_new(values, freq=freq, tz=tz)
```
jbrockmendel: This is the second change from the existing version.  ATM this reads:

```
        # utc = len(dates) > 0 and dates[0].tzinfo is not None
        data = tools.to_datetime(dates)
```

mroeschke: Maybe you can go ahead and blow away the commented line of code above and just directly use `xdr`:

`values = np.array([x.value for x in xdr])`

jbrockmendel: Good idea, will do.  I'd like to track down the source of the comment and make sure it isn't something important.

pandas/core/arrays/datetimelike.py
```diff
@@ -296,6 +312,34 @@ def resolution(self):
...
+        Parameters
+        ----------
+        index : DatetimeIndex or TimedeltaIndex
+            The index on which to determine if the given frequency is valid
+        freq : DateOffset
+            The frequency to validate
+        """
+        if is_period_dtype(cls):
+            # Frequency validation is not meaningful for Period Array/Index
+            return
```
jreback: be explicit about the return

pandas/core/arrays/datetimelike.py
```diff
@@ -477,6 +521,188 @@ def _addsub_offset_array(self, other, op):
...
+                # DatetimeIndex, ndarray[datetime64]
+                return self._add_datelike(other)
+            elif is_integer_dtype(other):
+                result = self._addsub_int_array(other, operator.add)
+            elif is_float_dtype(other) or is_period_dtype(other):
+                # Explicitly catch invalid dtypes
+                raise TypeError("cannot add {dtype}-dtype to {cls}"
+                                .format(dtype=other.dtype,
+                                        cls=type(self).__name__))
+            elif is_categorical_dtype(other):
```
jreback: better is is_extension_array_dtype

pandas/core/arrays/datetimelike.py
```diff
@@ -477,6 +521,188 @@ def _addsub_offset_array(self, other, op):
...
+            elif isinstance(other, ABCIndexClass):
+                raise TypeError("cannot subtract {cls} and {typ}"
+                                .format(cls=type(self).__name__,
+                                        typ=type(other).__name__))
+            elif is_float_dtype(other):
+                # Explicitly catch invalid dtypes
+                raise TypeError("cannot subtract {dtype}-dtype from {cls}"
+                                .format(dtype=other.dtype,
+                                        cls=type(self).__name__))
+            elif is_categorical_dtype(other):
```
jreback: same

pandas/core/arrays/datetimes.py
jreback: hmm this list conversion is problematic

pandas/core/arrays/datetimes.py
```diff
@@ -1085,3 +1203,91 @@ def to_julian_date(self):
...
+
+    return data
+
+
+def _infer_tz_from_endpoints(start, end, tz):
+    try:
```
jreback: for sure add doc-strings in future

jreback (changes requested):

jbrockmendel:
> hmm this list conversion is problematic

Why?

jreback:
> hmm this list conversion is problematic

you are turning an array into a list, only to turn it back to an array

jbrockmendel:
> you are turning an array into a list, only to turn it back to an array

It's the same in the existing implementation; we can open an issue if there's a real perf impact.

jreback:
> it's the same in the existing implementation; we can open an issue if there's a real perf impact.

ok if that's true, then this is prob an overblow concern. pls give a quick test to see if this is actually an issue.

rebase in an y event.

pandas/core/indexes/datetimelike.py
```diff
@@ -754,44 +593,6 @@ def isin(self, values):
 
         return algorithms.isin(self.asi8, values.asi8)
 
-    def shift(self, n, freq=None):
```
jorisvandenbossche: Should we keep a stub calling the super method here as well? (as you did with the arithmetic methods or with `_generate_range`) 
I know it is not strictly needed here, but I have the feeling that will make it more clear that then in the future we need to update this from calling super class method to calling the composited values method (although I suppose we have tests for it that will also be a reminder to do it)

jorisvandenbossche (approved):

jreback (approved):

jreback:
thanks @jbrockmendel, pls follow up with @jorisvandenbossche comments in next PR

pandas/core/indexes/datetimelike.py
jbrockmendel: For `_generate_range` and a few others it is needed because the Index method needs to set `result.name`.

There are a handful of other things things in the same situation as shift (e.g. several properties that are overriden by cache_readonlys) that will need to be edited if/when the change from subclassing to composition is made.  I'll make a note not to forget `shift` when the time comes.

jbrockmendel:
@TomAugspurger I think with this PeriodArrayMixin has everything needed for comparisons+arithmetic, so PeriodBlock should be viable.

jorisvandenbossche:
@jbrockmendel you are planning to do a PR for that?

jbrockmendel:
> you are planning to do a PR for that?

IIRC Tom wanted to be informed when the arith/compare parts were ready so he could forge ahead with PeriodBlock.

My current focus is on fixing DataFrame/Series ops inconsistencies by having DataFrame dispatch to Series (#22068, #22019) and getting thorough tests for these new EA subclasses (#22033)

TomAugspurger:
Thanks.

I'm hoping to knock out SparseArray this week or, so if anyone else wants
to do Period then by all means.
Otherwise I'll start on it after sparse.

jorisvandenbossche:
@jbrockmendel since you are now very familiar with code, I would say, certainly go ahead (if you have time of course)

Issue #22017, opened 2018-07-22
Timestamp comparison inconsistency


jbrockmendel:
Trying to make align the behavior of DataFrame/Series ops, a sticking point turns out to be Timestamp comparisons:

```
ts = pd.Timestamp.now()
df = pd.DataFrame({'A': [1.0, 2.0],
                   'B': [1, 2],
                   'C': ['foo', 'bar']})

>>> df < ts
      A     B     C
0  True  True  True
1  True  True  True
>>> ts < df
      A     B     C
0  True  True  True
1  True  True  True
```

So first off, this behavior is just nuts.  But let's look at Series behavior:

```
df['A'] < ts
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/pandas/core/ops.py", line 1283, in wrapper
    res = na_op(values, other)
  File "/usr/local/lib/python3.7/site-packages/pandas/core/ops.py", line 1167, in na_op
    result = method(y)
  File "pandas/_libs/tslibs/timestamps.pyx", line 170, in pandas._libs.tslibs.timestamps._Timestamp.__richcmp__
TypeError: Cannot compare type 'Timestamp' with type 'float'

>>> df['B'] == ts
0    False
1    False
Name: B, dtype: bool

>>> df['B'] > ts
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/pandas/core/ops.py", line 1283, in wrapper
    res = na_op(values, other)
  File "/usr/local/lib/python3.7/site-packages/pandas/core/ops.py", line 1167, in na_op
    result = method(y)
  File "pandas/_libs/tslibs/timestamps.pyx", line 170, in pandas._libs.tslibs.timestamps._Timestamp.__richcmp__
TypeError: Cannot compare type 'Timestamp' with type 'int'

>>> df['C']
0    foo
1    bar
Name: C, dtype: object
>>> df['C'] <= ts
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/pandas/core/ops.py", line 1283, in wrapper
    res = na_op(values, other)
  File "/usr/local/lib/python3.7/site-packages/pandas/core/ops.py", line 1143, in na_op
    result = _comp_method_OBJECT_ARRAY(op, x, y)
[...]

PR #22019, opened 2018-07-22
[CLN] Dispatch (some) Frame ops to Series, avoiding _data.eval

jbrockmendel:
ATM there are exactly two places where `BlockManager.eval` is called: `DataFrame._combine_match_columns` and `DataFrame._combine_const`.  This replaces the usage in `_combine_match_columns` with a dispatch-to-Series implementation.  Some output dtypes get changed (see edits in `test_axis_select_reindex`, `test_pivot`), and some errors get changed from `ValueError` to `TypeError` (see `test_operators`).

The other usage of `_data.eval` will be removed separately; that turns out to be a lot more trouble because a bunch of `DataFrame` behavior is currently incorrect (see #22017).

This PR also:
- Simplifies some of the special-casing in SparseDataFrame; trying to move towards not having separate implementations for these methods
- Dispatches `_combine_match_index` to avoid calling `self.values` when doing so would require coercing to object-dtype.


pandas/core/internals/__init__.py
```diff
@@ -86,6 +86,55 @@
...
+    """
+    if dtype is None:
+        dtype = left.dtype
+
+    if (is_integer_dtype(left) or is_bool_dtype(left) or
+            is_datetime64_dtype(left) or is_datetime64tz_dtype(left)):
+        pass
+    elif is_object_dtype(left) and lib.infer_dtype(left) == 'boolean':
+        # disguised is_bool_dtype
+        pass
```
jbrockmendel: This is nearly cut/paste from the existing `Block._try_cast_result` implementation.  The difference is this uses e.g. `is_integer_dtype(left)` instead of `left.is_integer`.  The `is_foo` attributes map directly onto the `is_foo_dtype` checks with the exception of `is_bool`, which requires this extra object-dtype check (this is the same way `Index.is_boolean` works).

pandas/core/frame.py
```diff
@@ -4921,21 +4922,64 @@ def _arith_op(left, right):
...
 
-        new_data = left._data.eval(func=func, other=right,
-                                   axes=[left.columns, self.index],
-                                   try_cast=try_cast)
-        return self._constructor(new_data)
+        new_data = [func(left.iloc[:, n], right.iloc[n])
+                    for n in range(len(left.columns))]
+
+        if try_cast:
+            new_data = [try_cast_result(left.iloc[:, n], new_data[n])
```
jreback: pls don't repeat code for the unique / non-unique case

pandas/core/internals/__init__.py
```diff
@@ -86,6 +86,55 @@
 from pandas.compat import range, map, zip, u
 
 
+def try_cast_result(left, result, dtype=None):
```
jreback: IF you need this function, it should be in panda/core/dtypes/cast.py. I say IF because this looks like lots of other functions there. please see if you can simplify. we want to remove things from internals not add them.

jreback (changes requested):

pandas/core/internals/__init__.py
jbrockmendel: > it should be in panda/core/dtypes/cast.py

Sure, though I'm hoping we can get rid of it before long.

> I say IF because this looks like lots of other functions there

This is based on the existing `Block._try_cast_result`, which is changed to just call this function.

> we want to remove things from internals not add them.

Very much so.  The point of this PR (along with another branch in progress) is to remove `Block.eval` and `BlockManager.eval`. (and in doing so ensure that Series/DataFrame ops all agree)

pandas/core/frame.py
jbrockmendel: This can be trimmed down a bit, but is based on the existing `_combine_frame` code.  I'm anticipating that after `combine_const` is "fixed" most of this will move over to core.ops and we'll put some effort into being less verbose.

jbrockmendel:
Changes in returns dtypes (see tests) merit discussion since they represent an API change.  Effectively some ops that currently return e.g. all-float will now return mixed float/int dtypes.

jbrockmendel:
@jreback just reverted a large chunk of non-central changes, made the remaining change in core.frame much less verbose.

This (along with the in-progress branch that gets rid of the other use of `BlockManager.eval`+`Block.eval`) will be a huge win.  We clear out a bunch of internals _and_ make Series/DataFrame ops consistent.  A _bunch_ of issues will be closed as a consequence of the latter.

This changes some return types, so will need an API Changes note if it moves forward.

jreback (changes requested):

pandas/core/frame.py
```diff
@@ -4922,21 +4922,30 @@ def _arith_op(left, right):
...
+        assert left.columns.equals(right.index)
 
-        new_data = left._data.eval(func=func, other=right,
-                                   axes=[left.columns, self.index],
-                                   try_cast=try_cast)
-        return self._constructor(new_data)
+        # Note: we use iloc instead of loc for compat with
+        # case with non-unique columns; same reason why we pin columns
+        # to result below instead of passing it to the constructor.
+        new_data = {n: func(left.iloc[:, n], right.iloc[n])
```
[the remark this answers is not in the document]
jbrockmendel: > why is moving operations code here a good idea, rather than internals? we want to ove code out of frame, not in

We want to move code out of _internals_.  There are a ton of Issues about `DataFrame` operations that are caused by `Block.eval` not handling things correctly that will be fixed by using the (much more careful and well-tested) `Series` implementations (which BTW are the ones that correctly defer to EA and `FooIndex` implementations).

`DataFrame` ops are all over the place right now: `op(DataFrame, DataFrame)` and `cmp(DataFrame, DataFrame)` operate column-wise (dispatching to `Series` implementation).   Other ops will either a) operate on `self.values` (doing expensive casting for mixed dtypes), or b) dispatch to `_data.eval`, which messes up on many dtypes and weird alignment behavior.

pandas/tests/reshape/test_pivot.py
```diff
@@ -1565,8 +1565,9 @@ def test_crosstab_normalize(self):
                               full_normal)
         tm.assert_frame_equal(pd.crosstab(df.a, df.b, normalize='index'),
                               row_normal)
-        tm.assert_frame_equal(pd.crosstab(df.a, df.b, normalize='columns'),
-                              col_normal)
+        tm.assert_frame_equal(
+            pd.crosstab(df.a, df.b, normalize='columns').astype('f8'),
```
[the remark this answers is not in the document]
jbrockmendel: The change is happening because ATM the casting is done at the Block-level, while with the PR casting is done at the column-level.  `pd.crosstab(df.a, df.b, normalize='columns')` ATM returns

```
b    3    4
a          
1  0.5  0.0
2  0.5  1.0
```

and under the PR gives:

```
b    3    4
a          
1  0.5  0
2  0.5  1
```

i.e. the `4` column is `int64`.  Note that the original inputs `df.a` and `df.b` are both int64 to begin with.  I don't know `crosstab` well enough to comment on the ideal output, but it's hard to call this less-correct.

jbrockmendel:
Note also this can be updated to be even less verbose after #22068.

@jorisvandenbossche thoughts on the dtype effects?

pandas/tests/reshape/test_pivot.py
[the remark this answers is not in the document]
TomAugspurger: I think the correct way to answer this is ask "was the upcast from int to float the result of coercing to a single dtype for multiple columns?" If so, then we should preserve integer dtype where we can, i.e.


For an all-integer `crosstab`, we preserve integer dtype:

```python
In [27]: a = pd.Series([1, 2])

In [28]: pd.crosstab(a, a)
Out[28]:
col_0  1  2
row_0
1      1  0
2      0  1
``` 

so this change seems good (and maybe worth explicitly testing rather than astyping the output, to ensure that we don't regress)?

pandas/core/frame.py
```diff
@@ -4922,21 +4922,30 @@ def _arith_op(left, right):
                                  copy=False)
 
     def _combine_match_index(self, other, func, level=None):
+        assert isinstance(other, Series)
         left, right = self.align(other, join='outer', axis=0, level=level,
                                  copy=False)
+        assert left.index.equals(right.index)
```
TomAugspurger: Are there cases where this might no be true? If not, then let's remove it, as it could be somewhat expensive (though maybe not, since the output of align shares indices.)

jbrockmendel: This always holds.  I added it largely so I wouldn't have to keep going back to core.ops to double-check what was being passed.

jbrockmendel:
@jreback thoughts on @TomAugspurger’s comments?  My main goals here are to a) simplify DataFrame ops and b) get rid of Block/BlockManager methods.

jreback (changes requested):
what is perf differences ?

i have a couple of unresolved comments above

jreback:
in the crosstab example what is the input look like?

pandas/tests/frame/test_axis_select_reindex.py
```diff
@@ -738,7 +738,7 @@ def test_align_int_fill_bug(self):
 
         result = df1 - df1.mean()
         expected = df2 - df2.mean()
-        assert_frame_equal(result, expected)
+        assert_frame_equal(result.astype('f8'), expected)
```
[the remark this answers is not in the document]
jbrockmendel: Same as [below](https://github.com/pandas-dev/pandas/pull/22019#discussion_r205946906)

jbrockmendel:
> in the crosstab example what is the input look like?

On L1569 in tests/reshape/test_pivot.py we call `pd.crosstab(df.a, df.b, normalize='columns')`.  In that, `df` is given by:

```
>>> df = pd.DataFrame({'a': [1, 2, 2, 2, 2], 'b': [3, 3, 4, 4, 4], 'c': [1, 1, np.nan, 1, 1]})
>>> df
   a  b    c
0  1  3  1.0
1  2  3  1.0
2  2  4  NaN
3  2  4  1.0
4  2  4  1.0
```

> i have a couple of unresolved comments above

Can you be more specific?  AFAICT all your comments have been responded to; not sure which ones you consider unresolved.

jbrockmendel:
> what is perf differences ?

Some %timeit trials:

```
df = pd.DataFrame({'a': np.arange(1000), 'b': np.arange(1000)*2})
%timeit out = pd.crosstab(df.a, df.b, normalize='columns')
--> 481 ms on PR, 289ms on master
--> presumably overhead from DataFrame construction, possibly consolidation

%timeit out = df + df.iloc[0]
--> 974 µs on PR, 398 on master
--> makes sense since the master version is taking advantage of contiguous blocks
```

So this does entail a perf hit.  I expect that to be ameliorated by #22284, but the fact remains that this is AFAIK the main upside of the BlockManager in the first place.

I maintain this performance hit is less important than the correctness-improvement this PR brings.  For example (not an extant test, but I'll add it and others, ideally after #22350):

```
now = pd.Timestamp.now()
df = pd.DataFrame([[now, now], [now, now]])
ser = pd.Series([now, now])
>>> df - ser
```

Under this PR we correctly get:
```
       0      1
0 0 days 0 days
1 0 days 0 days
```

In master we get dtypes wrong:
```
           0          1
0 1970-01-01 1970-01-01
1 1970-01-01 1970-01-01
```

pandas/core/ops.py
```diff
@@ -1632,6 +1632,7 @@ def dispatch_to_series(left, right, func, str_rep=None):
     right : scalar or DataFrame
     func : arithmetic or comparison operator
     str_rep : str or None, default None
+    axis : {None, "index", "columns"}
```
jreback: this is not fully correct, should be ``(0, "index"), (1, "columns")``

pandas/core/ops.py
```diff
@@ -1797,7 +1805,10 @@ def f(self, other, axis=default_axis, level=None, fill_value=None):
         if isinstance(other, ABCDataFrame):  # Another DataFrame
             return self._combine_frame(other, na_op, fill_value, level)
         elif isinstance(other, ABCSeries):
-            return _combine_series_frame(self, other, na_op,
+            # For these values of `axis`, we end up dispatching to Series op,
+            # so do not want the masked op.
+            pass_op = op if axis in [0, "columns", None] else na_op
```
jreback: this is pretty janky, can you make simpler?

pandas/tests/frame/test_axis_select_reindex.py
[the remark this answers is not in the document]
jreback: I think this PR needs a note in the what's new describing this change (as its effectievly an API change)

jreback (changes requested):

jbrockmendel:
PendingDeprecationWarning in xlrd

jreback:
that must be a new warning from xlrd?

can u pin the last version and create an issue to fix the warning?

jbrockmendel:
Rebase notwithstanding, thoughts on this?  If this goes through then there is only one use of BlockManger.eval left, and we can focus on fixing the last inconsistencies between Series vs DataFrame ops.

pandas/core/ops.py
```diff
@@ -1655,6 +1656,13 @@ def column_op(a, b):
             return {i: func(a.iloc[:, i], b.iloc[:, i])
                     for i in range(len(a.columns))}
 
+    elif isinstance(right, ABCSeries) and axis == "columns":
+        assert right.index.equals(left.columns)
```
jreback: can you add a doc-comment here (and below) to disambiguate the elifs

jreback (changes requested):
this looked generally ok. I think needs a whatsnew section as a change (good change), e.g. show the crossstab example.

jbrockmendel:
Travis error is unrelated on test_s3_roundtrip

jbrockmendel:
Same test_s3_roundtrip failure again.  Is there any chance that it isn't a coincidence?

jbrockmendel:
CircleCI fail unrelated, hypothesis turned up an overflow

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -532,6 +532,35 @@ Current Behavior:
     ...
     OverflowError: Trying to coerce negative values to unsigned integers
 
+.. _whatsnew_0240.api.crosstab_dtypes
+
+Crosstab Preserves Dtypes
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+`pd.crosstab` will preserve now dtypes in some cases that previously would
```
jreback: can you use ``:func:`crosstab` `` here

jreback (approved):
tiny doc comment. lgtm. push that change and ok to merge.

jreback:
acutally rebase and wait for green to merge, just to be sure

jbrockmendel:
Great.  Once this is in, #22880 or equivalent will be the end of BlockManager.eval

jbrockmendel:
@jreback travis fail is in s3; should I call that "green enough" or keep pushing until green-green?

PR #22028, opened 2018-07-23
Separate out internals.concat, internals.managers

jbrockmendel:
Follow-up to #22014.

Moved `get_blockno_placements` to libinternals since a) its a natural fit anyway and b) it is needed in both concat and managers and I didn't want to runtime-import it.

The only non-cut/paste edit is a change of `isinstance(self.block, ExtensionBlock)` to `self.block.is_extension`

`__init__` namespace is big in part because `io.packers` uses `getattr(internals, name)` and is just really tightly intertwined with internals.


TomAugspurger (approved):

jreback:
yep, thanks @jbrockmendel

PR #22031, opened 2018-07-23
REF/API: Stricter extension checking.

### LINKED ISSUE #22021: BLD: pandas-wheels are failing
https://travis-ci.org/MacPython/pandas-wheels/builds/406975380

looks like #21879 somehow broke this. The error messages I think are red herrings for somehow picking up a wrong version of numpy when things are compiled.

TomAugspurger:
Removes is_extension_array_dtype's handling of both arrays and dtypes.
Now it handles just arrays, and we provide `is_extension_dtype` for checking
whether a dtype is an extension dtype. It's the caller's responsibility to know
whether they have an array or dtype.

Closes #22021


jreback (changes requested):
why are you changing this? we do this for all other is_*

TomAugspurger:
> why are you changing this?

To fix #22021

> we do this for all other is_*

Yeah, and that's why `is_extension_array_dtype` originally followed the same pattern. But I think it's not a good idea to pass values to `pandas_dtype`. All the call sites knew whether they had values or a dtype.

I can make similar changes for the other methods, but we have backwards compatibility issues there.

jreback (changes requested):
I can't believe this didn't break anything. I am -1 on this though because this changes the definition of is_*_dtype. All of the routines work on an array correctly. Adding 2 different methods is very confusing as we on't do this for anything else and introduces a special case.

TomAugspurger:
> Wouldn't it be more logical to call this is_extension_array instead of is_extension_array_dtype? (I suppose the reason is backwards compatibility for at least one of the two use cases?)

Yes. These weren't in the public namespace earlier, so we don't have to worry about deprecating.

> Adding 2 different methods is very confusing

How so? Let's step back to the original issue: `pandas_dtype` shouldn't pass *data* to `np.dtype`. Say I have an object array `arr = [('f1', 'f2'), ('f2', 'f4')]`, as part of an operation that gets sliced to `arr[0]`, and we end up calling `pandas_dtype(subarr)`.

```python
In [28]: subarr = [('f1', 'f2')]

In [29]: pd.api.types.pandas_dtype(subarr)
Out[29]: dtype([('f1', '<f2')])
```

Is that what you would expect?

> we on't do this for anything else and introduces a special case.

As I said,

> I can make similar changes for the other methods, but we have backwards compatibility issues there.

jorisvandenbossche:
> I can't believe this didn't break anything

Why? As Tom said, "all the call sites knew whether they had values or a dtype.", and there are not yet that many usages in the code base currently for this. 
I personally like the fact we are explicit in the code about whether we are checking a dtype or an array-like.

> Adding 2 different methods is very confusing as we on't do this for anything else and introduces a special case.

Note that we actually have two methods for some other types as well: we have `is_categorical` and `is_categorical_dtype`, `is_datetimetz` and `is_datetime64tz_dtype` (where the first act only on the array and not dtype), and `is_sparse` which is also for array-like. 
On the other hand, `is_period` and `is_interval` then follow the other scheme where they are for the scalar objects, and the `_dtype` version for dtypes and array-likes. 
Anyway, mainly noting that it is already not fully consistent. But I agree it would be nice to follow the same scheme for all check functions.

jreback:
I think this is a really bad idea to change. We have had this convention for a LONG time esp on the basic types. this is throwing out everything.

TomAugspurger:
Throwing out?

TomAugspurger:
Are you -1 on this @jreback? If so you'll need an alternative fix to #22021, as I'm not sure how to avoid that without passing data to np.dtype.

jreback:
as I said I am -1 on this. This changes too much.

TomAugspurger:
That's unfortunate.

TomAugspurger:
Sorry, I closed this in frustration with you @jreback. But I think it's a good change for the project, so I'll continue to state my case for it.

Comments like

> I am not sure what the problem is

and

> This changes the meaning of the is_* for a single case, which is contrary to all of the existing code.

are frustrating, since they've already been addressed. I explain the original problem in the issue https://github.com/pandas-dev/pandas/issues/22021#issuecomment-407157789) and here (https://github.com/pandas-dev/pandas/pull/22031#issuecomment-407414822).

Joris explains in https://github.com/pandas-dev/pandas/pull/22031#issuecomment-407415472, this follows the precedent laid out by categorical and datetimetz.

jorisvandenbossche:
> > Wouldn't it be more logical to call this is_extension_array instead of is_extension_array_dtype? (I suppose the reason is backwards compatibility for at least one of the two use cases?)

> Yes. These weren't in the public namespace earlier, so we don't have to worry about deprecating.

But since they are not yet in the public namespace, we can choose whichever best name we can think of? Or maybe I didn't understand your answer :-) 

Anyway, I think we need to think a bit more on the naming scheme. 
`is_extension_dtype` (for the dtype checker) sounds good. But `is_extension_array_dtype` for array-like checking sounds a bit strange to me (although in the end you check the `dtype` of the array ...). `is_extension` might be OK here, but for the other dtypes will conflict with the scalar checkers. And `is_extension_array` is maybe too specific in "array" as it checks for any array-like that has a dtype (series, index, extensionarray, ..)? Maybe "arraylike" (but doesn't read that nice)?

pandas/core/dtypes/common.py
```diff
@@ -1688,38 +1688,69 @@ def is_extension_type(arr):
...
+    return isinstance(dtype, ExtensionDtype)
 
-    return isinstance(arr_or_dtype, (ExtensionDtype, ExtensionArray))
+
+def is_extension_dtype(dtype):
+    """Check if a dtype object is a pandas extension dtype.
+
+    Parameters
+    ----------
+    arr : object
```
jorisvandenbossche: array -> dtype

pandas/core/dtypes/common.py
```diff
@@ -1688,38 +1688,69 @@ def is_extension_type(arr):
...
+        dtype = getattr(arr, 'dtype')
+    except AttributeError:
+        return False
+
+    return isinstance(dtype, ExtensionDtype)
 
-    return isinstance(arr_or_dtype, (ExtensionDtype, ExtensionArray))
+
+def is_extension_dtype(dtype):
+    """Check if a dtype object is a pandas extension dtype.
```
jorisvandenbossche: One question is whether we want to accept strings here as well? 
Or keep it strictly for actual dtype objects?

TomAugspurger:
> But since they are not yet in the public namespace, we can choose whichever best name we can think of? Or maybe I didn't understand your answer :-)

Sorry for the confusion. I think you understood my answer, but I didn't push the change :) I think `is_extension_array` for checking whether an array-like is an extension array is good.

pandas/core/dtypes/common.py
TomAugspurger: That's a good point, I suspect we should accept strings that have been registered.

I'll see what I can do.

jreback:
this is exactly what pandas_dtype does

pandas/core/dtypes/common.py
jreback: meant for the comment here
this is what pandas_dtype does

TomAugspurger:
Yes, but `pandas_dtype` also potentially passes data to `np.dtype` in https://github.com/pandas-dev/pandas/blob/2c7c79779f1ef353d02446504d3b5f5c5cc68759/pandas/core/dtypes/common.py#L2013, which we need to avoid.

jorisvandenbossche:
> but pandas_dtype also potentially passes data to np.dtype in

Since you now splitted it in two functions, in the new `is_extension_dtype`, you can only use `pandas_dtype` when the input is a string, and then you should never run into those problems?

TomAugspurger:
Refactored.

1. No longer splitting `is_extension_dtype` and `is_extension_array_dtype`.
2. `is_extension_array_dtype` avoids calling `pandas_dtype`. There's no reason to go through all the stuff that `pandas_dtype` does, including `np.dtype(object)`, just to test if it's an extension array or dtype
3. DatetimeTZDtype and PeriodDtype were previously registered as extension dtypes. But this mixed dtypes that satisfy the interface and our own internal ones. For now, we'll need two registries, but this should be reduced to one by the time 0.24.0 comes out.

pandas/core/dtypes/dtypes.py
```diff
@@ -22,9 +22,9 @@ class Registry(object):
     --------
     registry.register(MyExtensionDtype)
     """
-    dtypes = []
+    def __init__(self):
+        self.dtypes = []
 
-    @classmethod
```
TomAugspurger: This wasn't actually a classmethod.

jreback:
@TomAugspurger looks a lot nicer!

TomAugspurger:
I don't think so. pandas will still have the issue of

```python
In [28]: subarr = [('f1', 'f2')]

In [29]: pd.api.types.pandas_dtype(subarr)
Out[29]: dtype([('f1', '<f2')])
```

this is just a workaround that avoids the problem.

jreback:
can you rebase

jreback (approved):

PR #22032, opened 2018-07-23
fix Error Indexing DafaFrame with a 0-d array

### LINKED ISSUE #21946: Error Indexing DafaFrame with a 0-d array
```python
>>> df = pd.DataFrame([[1, 2], [3, 4]], columns=['a', 'b'])
>>> ar = np.array(0)
>>> df.iloc[ar]
  ...
  ...
  File "/home/float2/Documents/Python/Pandas/pandas/pandas/core/indexes/base.py", line 529, in _shallow_copy_with_infer
    if not len(values) and 'dtype' not in kwargs:
TypeError: object of type 'numpy.int64' has no len()

```
A 0-d array passes [this](https://github.com/pandas-dev/pandas/blob/1a23779f09abc6ebf908d66ee88b973b767e2e3c/pandas/core/indexing.py#L2090) check cause it is an Iterable. But later, when **len()** is called on, TypeError is raised.
It would be nice to implement [PEP 357](https://www.python.org/dev/peps/pep-0357/) to prevent this error.
The matter has been raised at PyTorch project. [#9237](https://github.com/pytorch/pytorch/pull/9237)

illegalnumbers:
closes #21946

- [x ] closes #xxxx
- [x ] tests added / passed
- [x ] whatsnew entry

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -4178,6 +4178,8 @@ class Index(IndexOpsMixin, PandasObject):
             pass
         elif is_integer(key):
             pass
+        elif np.array(key).ndim == 0:
+            self._invalid_indexer(form, key)
         elif kind in ['iloc', 'getitem']:
             self._invalid_indexer(form, key)
         return key
--- a/pandas/core/indexing.py
+++ b/pandas/core/indexing.py
@@ -2692,8 +2692,9 @@ def is_nested_tuple(tup, labels):
 
 def is_list_like_indexer(key):
     # allow a list_like, but exclude NamedTuples which can be indexers
-    return is_list_like(key) and not (isinstance(key, tuple) and
-                                      type(key) is not tuple)
+    return (is_list_like(key)
+            and not (isinstance(key, tuple) and type(key) is not tuple)
+            and not np.array(key).ndim == 0)
 
 
 def is_label_like(key):
--- a/pandas/tests/indexing/test_indexing.py
+++ b/pandas/tests/indexing/test_indexing.py
@@ -890,6 +890,13 @@ class TestMisc(Base):
         del df
         assert wr() is None
 
+    def test_zero_index_iloc_raises(self):
+        df = pd.DataFrame([[1, 2], [3, 4]], columns=['a', 'b'])
+        ar = np.array(0)
+        msg = 'Cannot index by location index with a non-integer key'
+        with assert_raises_regex(TypeError, msg):
+            df.iloc[ar]
+
 
 class TestSeriesNoneCoercion(object):
     EXPECTED_RESULTS = [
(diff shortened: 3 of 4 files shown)


pandas/tests/indexing/test_indexing.py
```diff
@@ -890,6 +890,13 @@ def test_no_reference_cycle(self):
         del df
         assert wr() is None
 
+    def test_zero_index_iloc_raises(self):
+        df = pd.DataFrame([[1, 2], [3, 4]], columns=['a', 'b'])
+        ar = np.array(0)
+        msg = 'Cannot index by location index with a non-integer key'
+        tm.assert_raises_regex(TypeError, msg,
```
jbrockmendel: Can use assertraisesregex as a context here

pandas/core/indexing.py
```diff
@@ -2692,8 +2692,9 @@ def is_nested_tuple(tup, labels):
 
 def is_list_like_indexer(key):
     # allow a list_like, but exclude NamedTuples which can be indexers
-    return is_list_like(key) and not (isinstance(key, tuple) and
-                                      type(key) is not tuple)
+    return (is_list_like(key)
+            and not (isinstance(key, tuple) and type(key) is not tuple)
+            and not np.array(key).ndim == 0)
```
gfyoung: You're really just checking if you have a scalar.  Is it possible to use (and potentially modify) `is_scalar` from `pandas.core.dtypes.inference`?

pandas/tests/indexing/test_indexing.py
```diff
@@ -890,6 +890,13 @@ def test_no_reference_cycle(self):
         del df
         assert wr() is None
 
+    def test_zero_index_iloc_raises(self):
+        df = pd.DataFrame([[1, 2], [3, 4]], columns=['a', 'b'])
```
gfyoung: Reference issue as a comment below the function `def`.

pandas/core/indexing.py
illegalnumbers: Possibly! I'll look into that more tomorrow.

jreback (changes requested):

pandas/core/indexes/base.py
```diff
@@ -4178,6 +4178,8 @@ def _validate_indexer(self, form, key, kind):
             pass
         elif is_integer(key):
             pass
+        elif np.array(key).ndim == 0:
```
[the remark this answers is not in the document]
illegalnumbers: Yea I wasn't really happy with it as is anyways and since it seems to break some tests I'll probably change it in the AM.

pandas/tests/indexing/test_indexing.py
illegalnumbers: Can do.

illegalnumbers:
Hmm..it seems after reverting my changes and taking a look at my test again that the test still passes. I think I have either a bad test or this was fixed on master before I looked at it? I'm gonna try to build from master and repro in the AM.

illegalnumbers:
Yea confirmed. I checked on master today and we give out the expected error in this instance. ```TypeError: Cannot index by location index with a non-integer key``` so I'm gonna close this and comment on the other one as such. Pretty sad :'( but I'll look for something else to contribute with.

PR #22033, opened 2018-07-23
[POC] implement test_arithmetic.py

jbrockmendel:
There are a ton of scattered arithmetic tests for Index/Series/DataFrame that _should_ be testing the same things, but in fact are haphazard.  Fixing this given the current structure would entail an enormous about of code duplication.

This PR if a proof of concept for gathering all those tests in one test_arithmetic.py file, parametrizing them, and ensuring that the relevant behavior is identical across arraylike classes.

As Datetime/Timedelta/Period EA come online, the case for de-duplication will be even stronger.

In this form it is really easy to track (via xfails) what behavior needs fixing.


gfyoung:
@jbrockmendel : So even though this is a POC, you still want to merge this, right?

jbrockmendel:
If there is consensus about this goal, yes.

jbrockmendel:
@jreback related to discussion about test refactoring.

jorisvandenbossche:
The idea would be to add the arrays as when they exist?

general idea looks good to me

pandas/tests/test_arithmetic.py
```diff
@@ -0,0 +1,163 @@
...
+import pytest
+import numpy as np
+
+import pandas as pd
+import pandas.util.testing as tm
+
+from pandas import Timedelta
+
+
+def assert_equal(left, right, **kwargs):
```
jreback: move to utils.testing

pandas/tests/test_arithmetic.py
```diff
@@ -0,0 +1,163 @@
...
+        tm.assert_index_equal(left, right, **kwargs)
+    elif isinstance(left, pd.Series):
+        tm.assert_series_equal(left, right, **kwargs)
+    elif isinstance(left, pd.DataFrame):
+        tm.assert_frame_equal(left, right, **kwargs)
+    else:
+        raise NotImplementedError(type(left))
+
+
+def box_expected(expected, box_cls):
```
jreback: maybe put in conftest.py, these are test helper functions that we don't necessarily want to publicly expose

jreback (changes requested):

pandas/tests/test_arithmetic.py
jreback: alternatively, we should put this into _test_decorators (maybe rename though)

jreback (approved):

jreback:
all for this!

PR #22034, opened 2018-07-24
PERF/REF: improve performance of Series.searchsorted, PandasArray.searchsorted, collect functionality

### LINKED ISSUE #14565: PERF: better use of searchsorted for indexing performance
Follow-up on https://github.com/pandas-dev/pandas/pull/14551

When searchsorted is used (eg as `labels.searchsorted(loc)`), it is important that `loc` is first cast to the same dtype as `labels` for performance reasons.

PR https://github.com/pandas-dev/pandas/pull/14551 fixed one such example, but further actions:

- check the other usages of `searchsorted` for this pattern
- add benchmarks to track the improvement

topper-123:
- [x] closes #14565
- [x] ASV's added / passed
- [x] whatsnew entry

Numpy's searchsorted doesn't like being given values to search for that aren't of the same type as the array being searched. By ensuring that the input value has the same dtype as the underlying array, the search is sped up significantly.

```python
>>> n = 1_000_000
>>> s = pd.Series(([1] * n + [2] * n + [3] * n), dtype='int8')
>>> %timeit s.searchsorted(1)  # python int
15.2 ms  # master
9.75 µs  # this PR
```

## ASV results

```
     [b9754556]       [fd6c117b]
-        93.9±9μs         10.9±0μs     0.12  series_methods.SearchSorted.time_searchsorted('int64')
-        95.4±0μs         10.9±2μs     0.11  series_methods.SearchSorted.time_searchsorted('float64')
-        140±10μs       10.7±0.4μs     0.08  series_methods.SearchSorted.time_searchsorted('str')
-      1.34±0.1ms       10.8±0.3μs     0.01  series_methods.SearchSorted.time_searchsorted('uint16')
-     1.52±0.03ms         10.7±0μs     0.01  series_methods.SearchSorted.time_searchsorted('int16')
-      1.79±0.1ms         12.5±2μs     0.01  series_methods.SearchSorted.time_searchsorted('int8')
-      1.46±0.1ms       10.2±0.8μs     0.01  series_methods.SearchSorted.time_searchsorted('int32')
-        1.46±0ms         9.54±1μs     0.01  series_methods.SearchSorted.time_searchsorted('uint8')
-      1.46±0.1ms         9.38±2μs     0.01  series_methods.SearchSorted.time_searchsorted('float32')
-      1.71±0.2ms       10.4±0.9μs     0.01  series_methods.SearchSorted.time_searchsorted('uint32')
-      1.82±0.1ms         10.7±0μs     0.01  series_methods.SearchSorted.time_searchsorted('uint64')
-     2.14±0.04ms         10.7±0μs     0.00  series_methods.SearchSorted.time_searchsorted('float16')

```

The improventents are largest when the dtype  isn't int64 or float64, but for those cases the improvement is also significant (10x).

### DIFF
--- a/asv_bench/benchmarks/series_methods.py
+++ b/asv_bench/benchmarks/series_methods.py
@@ -124,6 +124,25 @@ class Dropna(object):
         self.s.dropna()
 
 
+class SearchSorted(object):
+
+    goal_time = 0.2
+    params = ['int8', 'int16', 'int32', 'int64',
+              'uint8', 'uint16', 'uint32', 'uint64',
+              'float16', 'float32', 'float64',
+              'str']
+    param_names = ['dtype']
+
+    def setup(self, dtype):
+        N = 10**5
+        data = np.array([1] * N + [2] * N + [3] * N).astype(dtype)
+        self.s = Series(data)
+
+    def time_searchsorted(self, dtype):
+        key = '2' if dtype == 'str' else 2
+        self.s.searchsorted(key)
+
+
 class Map(object):
 
     params = ['dict', 'Series']
--- a/doc/source/whatsnew/v0.25.0.rst
+++ b/doc/source/whatsnew/v0.25.0.rst
@@ -64,7 +64,8 @@ Performance Improvements
 
 - Significant speedup in `SparseArray` initialization that benefits most operations, fixing performance regression introduced in v0.20.0 (:issue:`24985`)
 - `DataFrame.to_stata()` is now faster when outputting data with any string or non-native endian columns (:issue:`25045`)
--
+- Improved performance of :meth:`Series.searchsorted`. The speedup is especially large when the dtype is int8/int16/int32 and the searched key is within
+  the integer bounds for the dtype(:issue:`22034`)
 
 
 .. _whatsnew_0250.bug_fixes:
--- a/doc/source/whatsnew/v0.25.0.rst
+++ b/doc/source/whatsnew/v0.25.0.rst
@@ -65,7 +65,7 @@ Performance Improvements
 - Significant speedup in `SparseArray` initialization that benefits most operations, fixing performance regression introduced in v0.20.0 (:issue:`24985`)
 - `DataFrame.to_stata()` is now faster when outputting data with any string or non-native endian columns (:issue:`25045`)
 - Improved performance of :meth:`Series.searchsorted`. The speedup is especially large when the dtype is
-  int8/int16/int32 and the searched key is within the integer bounds for the dtype(:issue:`22034`)
+  int8/int16/int32 and the searched key is within the integer bounds for the dtype (:issue:`22034`)
 
 
 .. _whatsnew_0250.bug_fixes:
--- a/doc/source/whatsnew/v0.25.0.rst
+++ b/doc/source/whatsnew/v0.25.0.rst
@@ -64,8 +64,8 @@ Performance Improvements
 
 - Significant speedup in `SparseArray` initialization that benefits most operations, fixing performance regression introduced in v0.20.0 (:issue:`24985`)
 - `DataFrame.to_stata()` is now faster when outputting data with any string or non-native endian columns (:issue:`25045`)
-- Improved performance of :meth:`Series.searchsorted`. The speedup is especially large when the dtype is int8/int16/int32 and the searched key is within
-  the integer bounds for the dtype(:issue:`22034`)
+- Improved performance of :meth:`Series.searchsorted`. The speedup is especially large when the dtype is
+  int8/int16/int32 and the searched key is within the integer bounds for the dtype(:issue:`22034`)
 
 
 .. _whatsnew_0250.bug_fixes:
--- a/pandas/core/algorithms.py
+++ b/pandas/core/algorithms.py
@@ -1734,24 +1734,25 @@ def searchsorted(arr, value, side="left", sorter=None):
 
     .. versionadded:: 0.25.0
 
-    Find the indices into a sorted array `self` (a) such that, if the
+    Find the indices into a sorted array `arr` (a) such that, if the
     corresponding elements in `value` were inserted before the indices,
-    the order of `self` would be preserved.
+    the order of `arr` would be preserved.
 
-    Assuming that `self` is sorted:
+    Assuming that `arr` is sorted:
 
     ======  ================================
     `side`  returned index `i` satisfies
     ======  ================================
-    left    ``self[i-1] < value <= self[i]``
-    right   ``self[i-1] <= value < self[i]``
+    left    ``arr[i-1] < value <= self[i]``
+    right   ``arr[i-1] <= value < self[i]``
     ======  ================================
 
     Parameters
     ----------
-    arr: numpy.array or ExtensionArray
-        array to search in. Cannot be Index, Series or PandasArray, as that
-        would cause a RecursionError.
+    arr: array-like
+        Input array. If `sorter` is None, then it must be sorted in
+        ascending order, otherwise `sorter` must be an array of indices
+        that sort it.
     value : array_like
         Values to insert into `arr`.
     side : {'left', 'right'}, optional
--- a/pandas/core/algorithms.py
+++ b/pandas/core/algorithms.py
@@ -19,7 +19,7 @@ from pandas.core.dtypes.common import (
     ensure_float64, ensure_int64, ensure_object, ensure_platform_int,
     ensure_uint64, is_array_like, is_bool_dtype, is_categorical_dtype,
     is_complex_dtype, is_datetime64_any_dtype, is_datetime64tz_dtype,
-    is_datetimelike, is_extension_array_dtype, is_float_dtype,
+    is_datetimelike, is_extension_array_dtype, is_float_dtype, is_integer,
     is_integer_dtype, is_interval_dtype, is_list_like, is_numeric_dtype,
     is_object_dtype, is_period_dtype, is_scalar, is_signed_integer_dtype,
     is_sparse, is_timedelta64_dtype, is_unsigned_integer_dtype,
@@ -1722,6 +1722,88 @@ def take_2d_multi(arr, indexer, out=None, fill_value=np.nan, mask_info=None,
 
     func(arr, indexer, out=out, fill_value=fill_value)
     return out
+
+
+# ---- #
+# searchsorted #
+# ---- #
+
+def searchsorted(arr, value, side="left", sorter=None):
+    """
+    Find indices where elements should be inserted to maintain order.
+
+    .. versionadded:: 0.25.0
+
+    Find the indices into a sorted array `self` (a) such that, if the
+    corresponding elements in `value` were inserted before the indices,
+    the order of `self` would be preserved.
+
+    Assuming that `self` is sorted:
+
+    ======  ================================
+    `side`  returned index `i` satisfies
+    ======  ================================
+    left    ``self[i-1] < value <= self[i]``
+    right   ``self[i-1] <= value < self[i]``
+    ======  ================================
+
+    Parameters
+    ----------
+    arr: numpy.array or ExtensionArray
+        array to search in. Cannot be Index, Series or PandasArray, as that
+        would cause a RecursionError.
+    value : array_like
+        Values to insert into `arr`.
+    side : {'left', 'right'}, optional
+        If 'left', the index of the first suitable location found is given.
+        If 'right', return the last such index.  If there is no suitable
+        index, return either 0 or N (where N is the length of `self`).
+    sorter : 1-D array_like, optional
+        Optional array of integer indices that sort array a into ascending
+        order. They are typically the result of argsort.
+
+    Returns
+    -------
+    array of ints
+        Array of insertion points with the same shape as `value`.
+
+    See Also
+    --------
+    numpy.searchsorted : Similar method from NumPy.
+    """
+    if sorter is not None:
+        sorter = ensure_platform_int(sorter)
+
+    if is_integer_dtype(arr) and (
+            is_integer(value) or is_integer_dtype(value)):
+        from .arrays.array_ import array
+        # if `arr` and `value` have different dtypes, `arr` would be
+        # recast by numpy, causing a slow search.
+        # Before searching below, we therefore try to give `value` the
+        # same dtype as `arr`, while guarding against integer overflows.
+        iinfo = np.iinfo(arr.dtype.type)
+        value_arr = np.array([value]) if is_scalar(value) else np.array(value)
+        if (value_arr >= iinfo.min).all() and (value_arr <= iinfo.max).all():
+            # value within bounds, so no overflow, so can convert value dtype
+            # to dtype of arr
+            dtype = arr.dtype
+        else:
+            dtype = value_arr.dtype
+
+        if is_scalar(value):
+            value = dtype.type(value)
+        else:
+            value = array(value, dtype=dtype)
+    elif not (is_object_dtype(arr) or is_numeric_dtype(arr) or
+              is_categorical_dtype(arr)):
+        from pandas.core.series import Series
+        # E.g. if `arr` is an array with dtype='datetime64[ns]'
+        # and `value` is a pd.Timestamp, we may need to convert value
+        value_ser = Series(value)._values
+        value = value_ser[0] if is_scalar(value) else value_ser
+
+    result = arr.searchsorted(value, side=side, sorter=sorter)
+    return result
 
 
 # ---- #
--- a/pandas/core/algorithms.py
+++ b/pandas/core/algorithms.py
@@ -1724,9 +1724,9 @@ def take_2d_multi(arr, indexer, out=None, fill_value=np.nan, mask_info=None,
     return out
 
 
-# ---- #
+# ------------ #
 # searchsorted #
-# ---- #
+# ------------ #
 
 def searchsorted(arr, value, side="left", sorter=None):
     """
@@ -1774,7 +1774,7 @@ def searchsorted(arr, value, side="left", sorter=None):
     if sorter is not None:
         sorter = ensure_platform_int(sorter)
 
-    if is_integer_dtype(arr) and (
+    if isinstance(arr, np.ndarray) and is_integer_dtype(arr) and (
             is_integer(value) or is_integer_dtype(value)):
         from .arrays.array_ import array
         # if `arr` and `value` have different dtypes, `arr` would be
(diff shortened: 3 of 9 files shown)


pandas/core/series.py
```diff
@@ -2077,8 +2077,11 @@ def __rmatmul__(self, other):
     def searchsorted(self, value, side='left', sorter=None):
         if sorter is not None:
             sorter = ensure_platform_int(sorter)
-        return self._values.searchsorted(Series(value)._values,
-                                         side=side, sorter=sorter)
+        if not is_extension_type(self._values):
+            value = np.asarray(value, dtype=self._values.dtype)
+            value = value[..., np.newaxis] if value.ndim == 0 else value
```
topper-123: Without this shim ``searchsorted`` returns a scalar, so this line is only to ensure that - like in master - an 1-dim array is returned.

I like better the numpy convention of returning a scalar from searchsorted when possible, but that can be for another PR.

topper-123:
The appveyor failure is a resource error, so unrelated to this PR.

pandas/core/series.py
jreback: can you add a comment here on why you are doing this (and the expl above as well)

jreback (changes requested):
ping on green.

topper-123:
green.

pandas/core/series.py
```diff
@@ -2077,8 +2077,14 @@ def __rmatmul__(self, other):
     def searchsorted(self, value, side='left', sorter=None):
         if sorter is not None:
             sorter = ensure_platform_int(sorter)
-        return self._values.searchsorted(Series(value)._values,
-                                         side=side, sorter=sorter)
+        if not is_extension_type(self._values):
```
jreback: hmm, this is orphaning pandas.core.base.searchsorted e.g. Index.searchsorted calls this. can you fix this up? (and add an asv for same). I think ok to make this if clause a helper function, maybe in algos.py and just call it in both places.

jreback (changes requested):

pandas/core/series.py
topper-123: Great idea  consolidating the two methods, I'll do that.

Now, there won't be a speed difference for Int64Index from this. But Uint64Index will benefit from a common implementation, as it today also experiences the casting problem:

```python
>>> i = pd.Index([1] * n + [2] * n + [3] * 3, dtype='uint64')
>>> %timeit i.searchsorted(2)
1.11 ms ± 4.98 µs per loop
```

So yeah, a common implentation is a great idea. I'll get to it.

topper-123:
This is as far I can make this for now. Still a few issues:
* does it give problems to downcast float64 to float32/16 or do we not lose precision
* Tests should be made for when value is out of bounds for integer dtype
* where do I put ASVs for Index.searchsorted?

I'll get back to this in about week, probably. Comments obviously welcome until then.

pandas/core/common.py
```diff
@@ -430,3 +431,46 @@ def _pipe(obj, func, *args, **kwargs):
...
+    dtype = arr.dtype
+    if sorter is not None:
+        sorter = ensure_platform_int(sorter)
+
+    # check integer bounds
+    iinfo = np.iinfo(dtype)
+    value_arr = np.array([value]) if is_scalar(value) else np.array(value)
+    if (value_arr < iinfo.min).any() or (value_arr > iinfo.max).any():
+        msg = "Value {} out of bound for dtype {}".format(value, dtype)
+        raise ValueError(msg)
```
topper-123: Notice I raise here when value is out of bound. I think this is reasonable behaviour and avoids some overflow surprises.

pandas/core/series.py
```diff
@@ -2080,10 +2080,24 @@ def __rmatmul__(self, other):
...
+        else:
+            # e.g. self is datetimelike and value is a pd.Timestamp
+            if sorter is not None:
+                sorter = ensure_platform_int(sorter)
+            value = Series(value)._values
+            result = self._values.searchsorted(value, side=side, sorter=sorter)
+
+        if is_scalar(result):
+            # ensure that a 1-dim array is returned
+            result = np.array([result])
```
topper-123: This bit is just to maintain backward compatability, to make <series.searchsorted always return a 1-dim array. I'd like to remove this in the future.

topper-123:
In the latest commit, Im automatically casting ints and uints to the correct dtype, bur also do some overflow checks, so we're not downcasting out-of-bound values. I think this is a good solution and is very fast.

But floats downcasting is more confusing, i.e. if the input in float64 and the array is float32 or float16, should we downcast?

Some options for floats:
1. Throw an error if input is not dtype compatible with array.
2. Silently force casting of input to the dtype of the array (fastest, but risks of wrong results?)
3. Keep current behaviour, i.e. just delegate decision to numpy (i.e. is slow, because we force upcasting of array for float32 and float16)
4. Keep current behaviour, but emit a warning if input and array are not dtype compatible (slow, but allows the user himself to fix the speed problem. Can be annoying with warnings)
5. Keep current behaviour, but add an boolean option to pd.options to emit a warning if input and array are not dtype compatible + add suitable explanation about the issue to the searchsorted doc string (Like option 4, but  avoids annoying with warnings)

I'm favoring option 4 or 5, but edge toward 5.

The reason is that in many use cases performance is really not the user's prime concern, and they just want things to work with minimal effort. Other users (beginners) may not understand types that well. So emitting dtype warnings would annoy/confuse such users. So, IMO option 5 would be best.

jreback:
don’t do anything for floats
ints are worth it because we commonly have non int64 and you do search on them

for floats you very rarely search and all of the above are too complex

topper-123:
I've made new version where I've made a common implementation.

I've skipped making changes for floats. That means searching in float16/32 Series with a float64 value is still slow, and we don't achieve the speed up for float16 and float32 that is shown in the table above.

Ints:
I've dropped making it an error, when the value is outside of bounds of the array dtype. Rather I delegate to numpy is that case. I think this is most consistent as we maintain an unchanged API, which is good. 
The "cost" is that searching for ints outside of bounds is slow, but that's ok IMO.

As the API is unchanged, I've not made any more tests. Series|Index/searchsorted is already tested in various places.

topper-123:
Is this ok?

Implementation on common for series and index, speed is much faster and it API-compatible with previous implementation...

pandas/core/common.py
```diff
@@ -430,3 +433,83 @@ def _pipe(obj, func, *args, **kwargs):
         return func(*args, **kwargs)
     else:
         return func(obj, *args, **kwargs)
+
+
+def ensure_integer_dtype(arr, value):
```
jreback: woa, what are you adding all of this code here for? its not the right place (don't add anything to core.common). can you make this much simpler. this is a very narrow case.

jreback (changes requested):

jreback:
@topper-123 if you would update

jreback:
needs a rebase

jreback:
@topper-123 can you comment on the added code

pandas/core/common.py
```diff
@@ -472,3 +471,84 @@ def f(x):
...
+        sorter = ensure_platform_int(sorter)
+
+    if is_integer_dtype(arr) and (
+            is_integer(value) or is_integer_dtype(value)):
+        return searchsorted_integer(arr, value, side=side, sorter=sorter)
+    if not (is_object_dtype(arr) or is_numeric_dtype(arr) or
+            is_categorical_dtype(arr)):
+        # E.g. if `arr` is an array with dtype='datetime64[ns]'
+        # and `value` is a pd.Timestamp, we may need to convert value
+        from pandas.core.series import Series
```
jreback: why are you doing this? this should accept an array, why are you then importing a Series???

jreback (changes requested):

jreback:
@topper-123 happy to have this if it can be done in a reasonbale way. pls update or close.

topper-123:
I'll look into in the coming week, maybe the weekend.

jreback:
@topper-123 if you have time to rebase this and address comment

topper-123:
I've now collected the esarchsorted functionality for Index, Series and PandasArray, making a common searchsorted implementation.

The Speed improvements are now also for PandasArray:

```python
>>> n = 100_000
>>> arr = pd.array([1]*n + [2]*n +[3]*n, dtype=np.int8)
>>> %timeit arr.searchsorted(2)
5.57 ms ± 332 µs per loop  # master
21.4 µs ± 3.55 µs per loop  # this PR
>>> ser = pd.Series(arr)
1.52 ms ± 9.33 µs per loop  # master
21 µs ± 2.81 µs  # this PR
```

I still need to make the ASV's for PandasArray, will make it later today.

pandas/core/common.py
```diff
@@ -470,3 +473,81 @@ def f(x):
...
+            dtype = value_arr.dtype
+
+        if is_scalar(value):
+            value = dtype.type(value)
+        else:
+            value = array(value, dtype=dtype)
+    elif not (is_object_dtype(arr) or is_numeric_dtype(arr) or
+              is_categorical_dtype(arr)):
+        from pandas.core.series import Series
+        # E.g. if `arr` is an array with dtype='datetime64[ns]'
```
jreback: put in algos.py

jreback (changes requested):

pandas/core/common.py
topper-123: do you mean algorithms.py?

jreback: yep

topper-123:
I've moved the searchsorted function to algorithms.py.

Where do PandasArray ASV's be placed? Can't find a logical location and having a pandas_array.py seems perhaps excessive for something that in most cases is a thin wrapper around numpy.array.

pandas/core/algorithms.py
```diff
@@ -1724,6 +1724,88 @@ def func(arr, indexer, out, fill_value=np.nan):
     return out
 
 
+# ---- #
```
jreback: can yo u make the 1st and 3rd lines match here

pandas/core/common.py
```diff
@@ -22,6 +21,8 @@
 from pandas.core.dtypes.inference import _iterable_not_string
 from pandas.core.dtypes.missing import isna, isnull, notnull  # noqa
 
+from pandas import compat
```
jreback: why was this changed?

pandas/core/algorithms.py
```diff
@@ -1724,6 +1724,88 @@ def func(arr, indexer, out, fill_value=np.nan):
...
+        sorter = ensure_platform_int(sorter)
+
+    if is_integer_dtype(arr) and (
+            is_integer(value) or is_integer_dtype(value)):
+        from .arrays.array_ import array
+        # if `arr` and `value` have different dtypes, `arr` would be
+        # recast by numpy, causing a slow search.
+        # Before searching below, we therefore try to give `value` the
+        # same dtype as `arr`, while guarding against integer overflows.
+        iinfo = np.iinfo(arr.dtype.type)
```
jreback: this can interactive with Nullable int types, is there testing on this?

jreback (changes requested):

pandas/core/algorithms.py
topper-123: It really should only let integer ndarrays reach this clause, else things get needlessly complex. I added a guard for this.

pandas/tests/arrays/test_array.py
```diff
@@ -254,3 +255,45 @@ def test_array_not_registered(registry_without_decimal):
     result = pd.array(data, dtype=DecimalDtype)
     expected = DecimalArray._from_sequence(data)
     tm.assert_equal(result, expected)
+
+
+class TestArrayAnalytics(object):
```
jreback: there are already tests in pandas/tests/extension/base/methods.py are these supplmental?

jreback (changes requested):

jreback:
pls merge master

pandas/tests/arrays/test_array.py
topper-123: Extension arrays and PandasArrays do not have the same interface, so the tests in pandas/tests/extension/base/methods.py will not work for PandasArrays. E.g. ``.value_counts``, ``.count`` etc. are not found in PandasArrays (or numpy arrays).

BTW, I copied these tests from pandas/tests/series/test_analytics.py.

topper-123:
The azure bugs*seem* unrelated to this PR, as I don't touch I/O or the html repr:

```
··· frame_methods.Repr.time_html_repr_trunc_mi                  failed
2019-02-03T09:40:31.5401600Z [ 11.57%] ···· Traceback (most recent call last):
2019-02-03T09:40:31.5401919Z                  File "/home/vsts/miniconda3/envs/pandas-dev/lib/python3.7/site-packages/asv/benchmark.py", line 1039, in main_run_server
2019-02-03T09:40:31.5402018Z                    main_run(run_args)
2019-02-03T09:40:31.5402334Z                  File "/home/vsts/miniconda3/envs/pandas-dev/lib/python3.7/site-packages/asv/benchmark.py", line 913, in main_run
2019-02-03T09:40:31.5402415Z                    result = benchmark.do_run()
2019-02-03T09:40:31.5402718Z                  File "/home/vsts/miniconda3/envs/pandas-dev/lib/python3.7/site-packages/asv/benchmark.py", line 412, in do_run
2019-02-03T09:40:31.5402779Z                    return self.run(*self._current_params)
2019-02-03T09:40:31.5403097Z                  File "/home/vsts/miniconda3/envs/pandas-dev/lib/python3.7/site-packages/asv/benchmark.py", line 506, in run
2019-02-03T09:40:31.5403275Z                    min_run_count=self.min_run_count)
2019-02-03T09:40:31.5403615Z                  File "/home/vsts/miniconda3/envs/pandas-dev/lib/python3.7/site-packages/asv/benchmark.py", line 569, in benchmark_timing
2019-02-03T09:40:31.5403707Z                    timing = timer.timeit(number)
2019-02-03T09:40:31.5403992Z                  File "/home/vsts/miniconda3/envs/pandas-dev/lib/python3.7/timeit.py", line 176, in timeit
2019-02-03T09:40:31.5404050Z                    timing = self.inner(it, self.timer)
2019-02-03T09:40:31.5404485Z                  File "<timeit-src>", line 6, in inner
2019-02-03T09:40:31.5404530Z                  File "/home/vsts/work/1/s/asv_bench/benchmarks/frame_methods.py", line 226, in time_html_repr_trunc_mi
2019-02-03T09:40:31.5404570Z                    self.df3._repr_html_()
2019-02-03T09:40:31.5404627Z                  File "/home/vsts/work/1/s/pandas/core/frame.py", line 651, in _repr_html_
2019-02-03T09:40:31.5404671Z                    import IPython
2019-02-03T09:40:31.5404898Z                  File "/home/vsts/miniconda3/envs/pandas-dev/lib/python3.7/site-packages/IPython/__init__.py", line 55, in <module>
2019-02-03T09:40:31.5404967Z                    from .terminal.embed import embed
2019-02-03T09:40:31.5405198Z                  File "/home/vsts/miniconda3/envs/pandas-dev/lib/python3.7/site-packages/IPython/terminal/embed.py", line 16, in <module>
2019-02-03T09:40:31.5405245Z                    from IPython.terminal.interactiveshell import TerminalInteractiveShell
2019-02-03T09:40:31.5405502Z                  File "/home/vsts/miniconda3/envs/pandas-dev/lib/python3.7/site-packages/IPython/terminal/interactiveshell.py", line 81, in <module>
2019-02-03T09:40:31.5405705Z                    if not _stream or not hasattr(_stream, 'isatty') or not _stream.isatty():
2019-02-03T09:40:31.5405746Z                ValueError: I/O operation on closed file
2019-02-03T09:40:31.5405794Z 
```

I don't think this  is related to this PR.

asv_bench/benchmarks/series_methods.py
```diff
@@ -124,6 +124,25 @@ def time_dropna(self, dtype):
...
 
+class SearchSorted(object):
+
+    goal_time = 0.2
+    params = ['int8', 'int16', 'int32', 'int64',
+              'uint8', 'uint16', 'uint32', 'uint64',
+              'float16', 'float32', 'float64',
+              'str']
+    param_names = ['dtype']
+
```
jreback: for a followup can add EA types here (Int8 and so on)

doc/source/whatsnew/v0.25.0.rst
```diff
@@ -64,7 +64,8 @@ Performance Improvements
 
 - Significant speedup in `SparseArray` initialization that benefits most operations, fixing performance regression introduced in v0.20.0 (:issue:`24985`)
 - `DataFrame.to_stata()` is now faster when outputting data with any string or non-native endian columns (:issue:`25045`)
--
+- Improved performance of :meth:`Series.searchsorted`. The speedup is especially large when the dtype is
+  int8/int16/int32 and the searched key is within the integer bounds for the dtype(:issue:`22034`)
```
jreback: need a space before the parens

pandas/core/algorithms.py
```diff
@@ -1724,6 +1724,88 @@ def func(arr, indexer, out, fill_value=np.nan):
...
+    ======  ================================
+    left    ``self[i-1] < value <= self[i]``
+    right   ``self[i-1] <= value < self[i]``
+    ======  ================================
+
+    Parameters
+    ----------
+    arr: numpy.array or ExtensionArray
+        array to search in. Cannot be Index, Series or PandasArray, as that
+        would cause a RecursionError.
```
jreback: not sure what this is referring. why is this not an array-like here?

pandas/tests/arrays/test_array.py
```diff
@@ -254,3 +255,45 @@ def test_array_not_registered(registry_without_decimal):
...
+        expected = np.array([2], dtype=np.intp)
+        tm.assert_numpy_array_equal(result, expected)
+
+    def test_searchsorted_numeric_dtypes_vector(self, any_real_dtype):
+        arr = pd.array([1, 3, 90], dtype=any_real_dtype)
+        result = arr.searchsorted([2, 30])
+        expected = np.array([1, 2], dtype=np.intp)
+        tm.assert_numpy_array_equal(result, expected)
+
+    def test_search_sorted_datetime64_scalar(self):
```
jreback: can you test for timedelta & datetime w/tz as well

jreback (changes requested):

pandas/core/algorithms.py
topper-123: Yes this text is wrong now, it can be indeed be array-like.

topper-123:
Comments addressed.

jreback (approved):

jreback:
thanks @topper-123

PR #22043, opened 2018-07-24
[WIP] Imprecise indexer

WeatherGod:
This is a work-in-progress to add a `tolerance` attribute to the `Index` class, and to plumb its use throughout the Index machinery. My immediate goal at this point is to not break anything. There is still a lot more work to do before this is ready for prime time, but hopefully I can get some inputs on best practices for mucking about in such deep internals of pandas.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/dtypes/missing.py
+++ b/pandas/core/dtypes/missing.py
@@ -369,7 +369,7 @@ def _isna_compat(arr, fill_value=np.nan):
     return True
 
 
-def array_equivalent(left, right, strict_nan=False):
+def array_equivalent(left, right, strict_nan=False, tolerance=None):
     """
     True if two arrays, left and right, have equal non-NaN elements, and NaNs
     in corresponding locations.  False otherwise. It is assumed that left and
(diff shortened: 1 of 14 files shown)


WillAyd:
Is this in reference to any particular issue or discussion?

jreback:
you can already do this with reindeer

what is the usecase?

WeatherGod:
Right, to summarize a bit, quite often with float64 indexes, you have two indexes which logically have similar keys, but because they were computed slightly differently, or came from different sources, they aren't binary identical.

I first tried implementing this just within Float64Index, but quickly ran into issues where I needed support implemented within the base class. Of course, once that happened, well, you need to implement a lot of this up into the other classes as well.

The basic premise of the design is that explicit will still always override implicit (which would be the tolerance attribute), which is why tolerance was added as an argument to many of the set operations. Also, any resulting indexes from these operations will have the tolerance that was used be set for its own tolerance attribute.

WeatherGod:
My employeer has changed priorities for me, so I have been unable to pursue this work any further, and I don't foresee any free time to spend on this. I hope someone else can take this work further, even if it is just going through and adding documentation.

The other major effort needed in this PR is to update the cython helpers for tolerance support, and unit tests.

Issue #13043, opened 2016-04-30
BUG: Series with datetime-like object ops raises TypeError


sinhrks:
#### Code Sample, a copy-pastable example if possible

```
pd.Series([pd.Timestamp('2015-01-01', tz='US/Eastern'),
           pd.Timestamp('2015-01-01', tz='Asia/Tokyo')]) + pd.Timedelta('1 days')
# TypeError: unsupported operand type(s) for +: 'numpy.ndarray' and 'Timedelta'

pd.Series([pd.Period('2015-01-01', freq='D')]) + pd.Timedelta('1 days')
# TypeError: unsupported operand type(s) for +: 'numpy.ndarray' and 'Timedelta'
```

#### Expected Output

ops must be applied to each elem

#### output of ``pd.show_versions()``

current master.

**NOTE:** This works:

```
pd.Series([pd.Timestamp('2015-01-01', tz='US/Eastern'),
           pd.Timestamp('2015-01-01', tz='Asia/Tokyo')]) + pd.tseries.offsets.Day()
# 0    2015-01-02 00:00:00-05:00
# 1    2015-01-02 00:00:00+09:00
# dtype: object

pd.Series([pd.Period('2015-01-01', freq='D')]) + pd.tseries.offsets.Day()
# 0   2015-01-02
# dtype: object
```

jreback:
@sinhrks thank you sir!

PR #22054, opened 2018-07-25
BUG: fix raise of TypeError when subtracting timedelta array

### LINKED ISSUE #21980: np.ndarray[object] - Timedelta raises
```
arr = np.array([pd.Timestamp.now(), pd.Timedelta('2D')])

>>> arr - pd.Timedelta('1D')
TypeError: unsupported operand type(s) for -: 'numpy.ndarray' and 'Timedelta'
```

It should attempt to operate element-wise.

illegalnumbers:
closes #21980

- [x] closes #xxxx
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -642,6 +642,7 @@ Timedelta
 - Bug in :class:`Series` with numeric dtype when adding or subtracting an an array or ``Series`` with ``timedelta64`` dtype (:issue:`22390`)
 - Bug in :class:`Index` with numeric dtype when multiplying or dividing an array with dtype ``timedelta64`` (:issue:`22390`)
 - Bug in :class:`TimedeltaIndex` incorrectly allowing indexing with ``Timestamp`` object (:issue:`20464`)
+- Fixed bug where subtracting :class:`Timedelta` from an object-dtyped array would raise ``TypeError`` (:issue:`21980`)
 -
 -
 
--- a/pandas/_libs/tslibs/timedeltas.pyx
+++ b/pandas/_libs/tslibs/timedeltas.pyx
@@ -541,10 +541,12 @@ def _binary_op_method_timedeltalike(op, name):
 
         elif hasattr(other, 'dtype'):
             # nd-array like
-            if other.dtype.kind not in ['m', 'M']:
-                # raise rathering than letting numpy return wrong answer
+            if other.dtype.kind in ['m', 'M']:
+                return op(self.to_timedelta64(), other)
+            elif other.dtype.kind == 'O':
+                return np.array([op(self, x) for x in other])
+            else:
                 return NotImplemented
-            return op(self.to_timedelta64(), other)
 
         elif not _validate_ops_compat(other):
             return NotImplemented
--- a/pandas/tests/scalar/timedelta/test_arithmetic.py
+++ b/pandas/tests/scalar/timedelta/test_arithmetic.py
@@ -200,6 +200,57 @@ class TestTimedeltaAdditionSubtraction(object):
         with pytest.raises(TypeError):
             2.0 - td
 
+    def test_td_sub_timedeltalike_object_dtype_array(self):
+        # GH 21980
+        arr = np.array([Timestamp('20130101 9:01'),
+                        Timestamp('20121230 9:02')])
+        exp = np.array([Timestamp('20121231 9:01'),
+                        Timestamp('20121229 9:02')])
+        res = arr - pd.Timedelta('1D')
+        tm.assert_numpy_array_equal(res, exp)
+
+    def test_td_sub_mixed_most_timedeltalike_object_dtype_array(self):
+        # GH 21980
+        now = pd.Timestamp.now()
+        arr = np.array([now,
+                        pd.Timedelta('1D'),
+                        np.timedelta64(2, 'h')])
+        exp = np.array([now - pd.Timedelta('1D'),
+                        pd.Timedelta('0D'),
+                        np.timedelta64(2, 'h') - pd.Timedelta('1D')])
+        res = arr - pd.Timedelta('1D')
+        tm.assert_numpy_array_equal(res, exp)
+
+    def test_td_rsub_mixed_most_timedeltalike_object_dtype_array(self):
+        # GH 21980
+        now = pd.Timestamp.now()
+        arr = np.array([now,
+                        pd.Timedelta('1D'),
+                        np.timedelta64(2, 'h')])
+        with pytest.raises(TypeError):
+            pd.Timedelta('1D') - arr
+
+    @pytest.mark.parametrize('op', [operator.add, ops.radd])
+    def test_td_add_timedeltalike_object_dtype_array(self, op):
+        # GH 21980
+        arr = np.array([Timestamp('20130101 9:01'),
+                        Timestamp('20121230 9:02')])
+        exp = np.array([Timestamp('20130102 9:01'),
+                        Timestamp('20121231 9:02')])
+        res = op(arr, pd.Timedelta('1D'))
+        tm.assert_numpy_array_equal(res, exp)
+
+    @pytest.mark.parametrize('op', [operator.add, ops.radd])
+    def test_td_add_mixed_timedeltalike_object_dtype_array(self, op):
+        # GH 21980
+        now = pd.Timestamp.now()
+        arr = np.array([now,
+                        pd.Timedelta('1D')])
+        exp = np.array([now + pd.Timedelta('1D'),
+                        pd.Timedelta('2D')])
+        res = op(arr, pd.Timedelta('1D'))
+        tm.assert_numpy_array_equal(res, exp)
+
 
 class TestTimedeltaMultiplicationDivision(object):
     """
@@ -616,3 +667,17 @@ class TestTimedeltaMultiplicationDivision(object):
 
         with pytest.raises(TypeError):
             divmod(np.array([22, 24]), td)
+
+    @pytest.mark.parametrize('op', [
+        operator.mul,
+        ops.rmul,
+        operator.truediv,
+        ops.rdiv,
+        ops.rsub])
+    @pytest.mark.parametrize('arr', [
+        np.array([Timestamp('20130101 9:01'), Timestamp('20121230 9:02')]),
+        np.array([pd.Timestamp.now(), pd.Timedelta('1D')])
+    ])
+    def test_td_op_timedelta_timedeltalike_array(self, op, arr):
+        with pytest.raises(TypeError):
+            op(arr, pd.Timedelta('1D'))


pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -542,10 +542,14 @@ def _binary_op_method_timedeltalike(op, name):
 
         elif hasattr(other, 'dtype'):
             # nd-array like
-            if other.dtype.kind not in ['m', 'M']:
+            if other.dtype.kind not in ['m', 'M', 'O']:
                 # raise rathering than letting numpy return wrong answer
                 return NotImplemented
-            return op(self.to_timedelta64(), other)
+            try:
+                converted_other = other.astype('datetime64[ns]')
```
jbrockmendel: Converting is wrong here, since it could be mixed, eg [Timestamp, Timedelta].

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -929,7 +933,7 @@ cdef class _Timedelta(timedelta):
     def nanoseconds(self):
         """
         Return the number of nanoseconds (n), where 0 <= n < 1 microsecond.
-       
+
```
jbrockmendel: Nice.  Want to add a line to lint.sh that checks for trailing white space in cython files?

illegalnumbers: Sure!

pandas/_libs/tslibs/timedeltas.pyx
illegalnumbers: The mixed test I added passes but I'm guessing this case is why I'm getting problems in other tests? Is there another approach I should take?

illegalnumbers:
Actually I'm a little confused now @jbrockmendel, checking on one of the failing tests that I have it is the one for ```test_ops_series_object``` (GH #13043),  it seems like based on the functionality that I'm changing the expectation there should change? Should the dtype now actually be a ```<M8[ns]``` when you do an addition rather than a ```O``` since we do the conversion properly now?

pandas/_libs/tslibs/timedeltas.pyx
illegalnumbers: Ah there was a bug in my test. Still looking around but definitely curious on another approach. I was thinking I could iterate over the entire array and do piecemeal conversions but that also seems wrong.

jbrockmendel:
> rather than a O since we do the conversion properly now?

The thing is that the line `converted_other = other.astype('datetime64[ns]')` doesn't belong at all.  Among other things the case to test is `other = np.array([pd.Timestamp.now(), pd.Timedelta('1D')])`.  That has an object-dtype, `__radd__` and `__rsub__` should both be valid, and the `astype` call would fail.

illegalnumbers:
Yea adding in the test for `other = np.array([pd.Timestamp.now(), pd.Timedelta('1D')])` exposes the failure you mentioned. I can remove that but moving forward I'm not sure how to implement that since the `O` type object fails to apply the lambda since Numpy just errors on the application of the lambda.

jbrockmendel:
> not sure how to implement that since the O type object fails to apply the lambda since Numpy just errors on the application of the lambda.

I'm not sure what lambda you're referring to, but I imagine this will end up looking something like:

```
if other.dtype.kind in ['m', 'M']:
     [do what it does now]
elif other.dtype.kind == 'O':
    return np.array([op(self, x) for x in other])
raise TypeError(...)
```

illegalnumbers:
Oh dang nice! I was just working through something similar in a REPL. I'll push up a revision in the next few minutes.

illegalnumbers:
Ok so I'm not sure why my build wouldn't output anything for the tests that did fail on TravisCI - @jbrockmendel is there a way I can retrigger travis? Is that known to be flakey? Or should I look into something else?

jbrockmendel:
Travis error looks unrelated.  When this happens I usually find some typo somewhere to fix and make a dummy commit to force it to re-run.

ci/lint.sh
```diff
@@ -49,6 +49,11 @@ if [ "$LINT" ]; then
     if [ $? -ne "0" ]; then
         RET=1
     fi
+
+    if [[ -n $(find **/*.pyx -type f -exec egrep -l " +$" {} \;) ]]
```
jbrockmendel: Will this print the list of offending files? Can you also include .pxd, .pxi, and .pxi.in?  Thanks for stepping up on this, linting these files is really tough.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -441,7 +441,7 @@ Datetimelike
 Timedelta
 ^^^^^^^^^
 
--
+- Fixed bug where array of timestamp and deltas raised a TypeError: unsupported operand type(s) for -: 'numpy.ndarray' and 'Timedelta' (:issue:`21980`)
```
jbrockmendel: Take a look at how quotation marks and backticks are used elsewhere In this file (iOS keyboard doesn’t make it easy to give an example directly)

pandas/tests/series/test_timeseries.py
```diff
@@ -1012,3 +1012,22 @@ def test_get_level_values_box(self):
         index = MultiIndex(levels=levels, labels=labels)
 
         assert isinstance(index.get_level_values(0)[0], Timestamp)
+
+    def test_diff_sub_timedelta(self):
```
jbrockmendel: These tests go in tests/scalar/timedelta/test_arithmetic.py

pandas/tests/series/test_timeseries.py
```diff
@@ -1012,3 +1012,22 @@ def test_get_level_values_box(self):
...
+        # GH 21980
+        arr = np.array([Timestamp('20130101 9:01'),
+                        Timestamp('20121230 9:02')])
+        exp = np.array([Timestamp('20121231 9:01'),
+                        Timestamp('20121229 9:02')])
+        res = arr - pd.Timedelta('1D')
+        tm.assert_numpy_array_equal(res, exp)
+
+    def test_diff_sub_timedelta_mixed(self):
```
jbrockmendel: This PR probably also fixes addition right?  If so, pls include a test.  Possibly also for reversed ops?

ci/lint.sh
illegalnumbers: Yea no worries! It won't display them as output I think just from this test but I can include the same find / grep dance under it or save it as a variable and echo it.

doc/source/whatsnew/v0.24.0.txt
illegalnumbers: I'll take a look!

pandas/tests/series/test_timeseries.py
illegalnumbers: Will move!

pandas/tests/series/test_timeseries.py
illegalnumbers: It should and I was thinking about reversed this morning actually so I'll include both today.

ci/lint.sh
jbrockmendel: > but I can include the same find / grep dance under it or save it as a variable and echo it.

Great.  If something breaks, we definitely want to know where to look for it.

illegalnumbers:
@jbrockmendel hmm upon further investigation this AM it seems that when doing the reverse subtract (ie 
```python
        arr = np.array([Timestamp('20130101 9:01'),
                        Timestamp('20121230 9:02')])
        exp = np.array([Timestamp('20121231 9:01'),
                        Timestamp('20121229 9:02')])
        res = pd.Timedelta('1D') - arr
        tm.assert_numpy_array_equal(res, exp)
```

I get `TypeError: descriptor '__sub__' requires a 'datetime.datetime' object but received a 'Timedelta'` in `pandas/_libs/tslibs/timestamps.pyx:330`. Still investigating but it seems like I might have to do a bigger change than I thought.

jbrockmendel:
> I get TypeError: descriptor '__sub__' requires a 'datetime.datetime' object but received a 'Timedelta' in pandas/_libs/tslibs/timestamps.pyx:330

The operation being run is `Timedelta - Timestamp` right?  That _should_ raise a `TypeError`, you're OK.

illegalnumbers:
Hmm actually, it appears that rsub doesn't work originally with `Timedelta` and `Timestamps`...maybe this is an existing issue?

```
>>> pd.Timedelta('1d') - Timestamp('20130101 9:01')
  File "pandas/_libs/tslibs/timestamps.pyx", line 332, in pandas._libs.tslibs.timestamps._Timestamp.__sub__
TypeError: descriptor '__sub__' requires a 'datetime.datetime' object but received a 'Timedelta'
```

illegalnumbers:
Ok so in that case I'll just check that these raise appropriately then.

illegalnumbers:
@jbrockmendel lemme know if there's anything else I need to do to get this guy merged! :)

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -441,7 +441,7 @@ Datetimelike
 Timedelta
 ^^^^^^^^^
 
--
+- Fixed bug where array of timestamp and deltas raised a TypeError: unsupported operand type(s) for -: ``numpy.ndarray`` and ``Timedelta`` (:issue:`21980`)
```
jbrockmendel: timestamp --> :class:`Timestamp`
double back ticks around ``TypeError``
don't need the full error message

Can you clarify what "array of timestamp and deltas" means?  e.g.

```
Bug where subtracting :class:`Timedelta` from an object-dtyped array would raise ``TypeError``
```

ci/lint.sh
```diff
@@ -71,6 +93,14 @@ if [ "$LINT" ]; then
             RET=1
         fi
     done
+
+    trailing_space_pxd=$(find ../**/*.pxd -type f -exec egrep -l " +$" {} \;)
+    if [[ -n $trailing_space_pxd ]]
+    then
+        RET=1
+        echo $trailing_space_pxd
+    fi
```
jbrockmendel: This looks great, thanks.  Any other ideas you have for linting cython files will be a big hit with the maintainers (separate PR(s))

pandas/tests/scalar/timedelta/test_arithmetic.py
```diff
@@ -616,3 +690,35 @@ def test_rdivmod_invalid(self):
 
         with pytest.raises(TypeError):
             divmod(np.array([22, 24]), td)
+
+    def test_td_div_timedelta_timedeltalike_array(self):
```
jbrockmendel: Would a valid case here be object-dtyped array containing all Timedelta objects (or some mix of timedelta, np.timedelta64)?

pandas/tests/scalar/timedelta/test_arithmetic.py
```diff
@@ -616,3 +690,35 @@ def test_rdivmod_invalid(self):
...
+
+    def test_td_mult_timedelta_mixed_timedeltalike_array(self):
+        # GH 21980
+        arr = np.array([pd.Timestamp.now(),
+                        pd.Timedelta('1D')])
+
+        with pytest.raises(TypeError):
+            pd.Timedelta('1D') * arr
+
+    def test_td_rmult_timedelta_mixed_timedeltalike_array(self):
```
jbrockmendel: Some of these you can probably de-duplicate using `pytest.mark.parametrize`.  Not a deal-breaker.

If there's a graceful way to work "object_dtype" into the test names that'd be ideal.

pandas/tests/scalar/timedelta/test_arithmetic.py
```diff
@@ -200,6 +200,80 @@ def test_td_rsub_numeric_raises(self):
...
+    def test_td_radd_timedelta_timedeltalike_array(self):
+        # GH 21980
+        arr = np.array([Timestamp('20130101 9:01'),
+                        Timestamp('20121230 9:02')])
+        exp = np.array([Timestamp('20130102 9:01'),
+                        Timestamp('20121231 9:02')])
+        res = pd.Timedelta('1D') + arr
+        tm.assert_numpy_array_equal(res, exp)
+
+    def test_td_radd_timedelta_mixed_timedeltalike_array(self):
```
jbrockmendel: The "timedelta" in "radd_timedelta" is redundant, since that is what the "td" in "td_radd" stands for.

Same comments as below re naming and parametrizing.

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -542,10 +542,15 @@ def _binary_op_method_timedeltalike(op, name):
...
-            if other.dtype.kind not in ['m', 'M']:
-                # raise rathering than letting numpy return wrong answer
+            try:
+                if other.dtype.kind in ['m', 'M']:
+                    return op(self.to_timedelta64(), other)
+                elif other.dtype.kind == 'O':
+                    return np.array([op(self, x) for x in other])
+                else:
+                    return NotImplemented
+            except:
```
jbrockmendel: What does the try/except catch?

ci/lint.sh
illegalnumbers: Sounds great! I am not sure if I'll have much time in the next little bit after this gets merged but I'll do my best.

pandas/_libs/tslibs/timedeltas.pyx
jbrockmendel: That explains the `return NotImplemented` on 551, but not the try/except on 552.  I'm pretty sure the try/except is not needed.

jreback (changes requested):

pandas/_libs/tslibs/timedeltas.pyx
illegalnumbers: well if calling the operation on 549 throws an exception wouldn't you want it to not error and instead return a proper message? that was why i caught any issue there. object types could be more than just timestamps, etc.

ci/lint.sh
```diff
@@ -60,6 +67,21 @@ if [ "$LINT" ]; then
             RET=1
         fi
     done
+
+    trailing_space_pxi_in=$(find ../**/*.pxi.in -type f -exec egrep -l " +$" {} \;)
+    if [[ -n $trailing_space_pxi_in ]]
+    then
```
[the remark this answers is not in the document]
illegalnumbers: sounds good yea it felt weird to include it here

pandas/_libs/tslibs/timedeltas.pyx
jbrockmendel: > well if calling the operation on 549 throws an exception wouldn't you want it to not error 

I'd want the traceback to show where the actual error came from.  Catching and re-raising obscures that.

illegalnumbers:
Status update: I'll try and finish wrapping this up sometime next weekend!

jbrockmendel:
@illegalnumbers any word?  Last I checked this was really close to ready.

illegalnumbers:
Sorry I suck! I haven't had any 20% time to commit to this and school kicked up so my priorities have been scattered. I will fix this up tonight and address the feedback and make a push.

pandas/_libs/tslibs/timedeltas.pyx
illegalnumbers: Done.

ci/lint.sh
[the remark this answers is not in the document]
illegalnumbers: Done.

pandas/tests/scalar/timedelta/test_arithmetic.py
illegalnumbers: Done

pandas/tests/scalar/timedelta/test_arithmetic.py
illegalnumbers: It might be a lot more effor to get this de-duped than I initially thought, would it be ok to submit as-is? I added in object_dtype to the method titles.

pandas/tests/scalar/timedelta/test_arithmetic.py
illegalnumbers: Done.

illegalnumbers:
Ok I think this should be good to go barring any other additional feedback! Thanks for being patient.

jreback (changes requested):

pandas/tests/scalar/timedelta/test_arithmetic.py
jreback: pls parameterize in this PR.

jreback (changes requested):

illegalnumbers:
I haven't used `pytest.mark.parametrize` before so you're going to have to wait until I have time to learn how to use that and work on this again. It might be a while.

illegalnumbers:
Additionally, an example of what you're looking for specifically would be wonderful!

pandas/tests/scalar/timedelta/test_arithmetic.py
jbrockmendel: @illegalnumbers the idea here is to notice that all four of these tests are special cases of a single test function.  In particular, you can write `arr * pd.Timedelta('1D')` as `operator.mul(arr, pd.Timedelta('1D')` and similarly for the others with `operator.truediv`, `ops.rdiv`, `ops.rmul`.  Then these can be parametrized as:

```
@pytest.mark.parametrize('op', [operator.mul, ops.rmul, operator.truediv, ops.rdiv])
def test_td_mul_object_array(self, op):
    arr = np.array([pd.Timestamp.now(), pd.Timedelta('1D')])

        op(arr, pd.Timedelta('1D'))
```

jreback:
see http://pandas.pydata.org/pandas-docs/stable/contributing.html#using-pytest

illegalnumbers:
Gonna see if I can finish this up today!

pandas/tests/scalar/timedelta/test_arithmetic.py
```diff
@@ -200,6 +200,92 @@ def test_td_rsub_numeric_raises(self):
...
 
+    def test_td_sub_timedeltalike_object_dtype_array(self):
+        # GH 21980
+        arr = np.array([Timestamp('20130101 9:01'),
+                        Timestamp('20121230 9:02')])
+        exp = np.array([Timestamp('20121231 9:01'),
+                        Timestamp('20121229 9:02')])
+        res = arr - pd.Timedelta('1D')
+        tm.assert_numpy_array_equal(res, exp)
+
```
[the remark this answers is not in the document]
illegalnumbers: Done, I parameterized the ones I could. The ones I could not (I kept getting errors relating to list concatenation, etc) I left as-is.

illegalnumbers:
Took a little while extra but it should be ready for review again. Barring any lint failures of course.

illegalnumbers:
I'm not sure if these are related to my ps? Kind of hard to read the build output.

jbrockmendel:
> EDIT: Seems like most are in the excel writer?

Yes.  I think this is fixed now.  If you rebase/push it _should_ go through alright.  I'll go over this one more time today, after which jreback will be called in for the final OK.

illegalnumbers:
Done! Like I said on the parameterize, I did the best I could without getting stuck. There were a few tests that seemed really similar but I kept getting exceptions when I tried to parameterize them so it seemed like more work than it was worth considering my experience.

pandas/tests/scalar/timedelta/test_arithmetic.py
```diff
@@ -616,3 +680,14 @@ def test_rdivmod_invalid(self):
...
+    @pytest.mark.parametrize('op', [operator.mul, ops.rmul, operator.truediv, ops.rdiv])
+    @pytest.mark.parametrize('arr', [
+                                      [Timestamp('20130101 9:01'),
+                                       Timestamp('20121230 9:02')],
+                                      [pd.Timestamp.now(),
+                                       pd.Timedelta('1D')]
+                                    ])
+    def test_td_op_timedelta_timedeltalike_array(self, op, arr):
+        with pytest.raises(TypeError):
+            arr * pd.Timedelta('1D')
```
jbrockmendel: You're going to need one more newline at the end of this file.

pandas/tests/scalar/timedelta/test_arithmetic.py
```diff
@@ -616,3 +680,14 @@ def test_rdivmod_invalid(self):
 
         with pytest.raises(TypeError):
             divmod(np.array([22, 24]), td)
+
+    @pytest.mark.parametrize('op', [operator.mul, ops.rmul, operator.truediv, ops.rdiv])
+    @pytest.mark.parametrize('arr', [
+                                      [Timestamp('20130101 9:01'),
```
jbrockmendel: shouldn't these be np.arrays, not lists?

pandas/tests/scalar/timedelta/test_arithmetic.py
```diff
@@ -200,6 +200,70 @@ def test_td_rsub_numeric_raises(self):
...
+        # GH 21980
+        arr = np.array([Timestamp('20130101 9:01'),
+                        Timestamp('20121230 9:02')])
+        exp = np.array([Timestamp('20130102 9:01'),
+                        Timestamp('20121231 9:02')])
+        res = op(arr, pd.Timedelta('1D'))
+        tm.assert_numpy_array_equal(res, exp)
+
+    @pytest.mark.parametrize('op', [operator.add, ops.radd])
+    def test_td_add_mixed_timedeltalike_object_dtype_array(self,op):
```
jbrockmendel: space after `,`.  Run `flake8 pandas/tests/scalar/timedelta/test_arithmetic.py` to get a full readout of places that need small edits

pandas/tests/scalar/timedelta/test_arithmetic.py
```diff
@@ -200,6 +200,70 @@ def test_td_rsub_numeric_raises(self):
...
+        res = arr - pd.Timedelta('1D')
+        tm.assert_numpy_array_equal(res, exp)
+
+    @pytest.mark.parametrize('arr', [
+                                      [Timestamp('20130101 9:01'),
+                                       Timestamp('20121230 9:02')],
+                                      [pd.Timestamp.now(),
+                                       pd.Timedelta('1D')]
+                                    ])
+    def test_td_rsub_timedeltalike_object_dtype_array(self,arr):
```
jbrockmendel: Could this be included in `test_td_op_timedelta_timedeltalike_array` by adding `ops.rsub` to the list of `op` args?

pandas/tests/scalar/timedelta/test_arithmetic.py
```diff
@@ -200,6 +200,70 @@ def test_td_rsub_numeric_raises(self):
...
+        # GH 21980
+        now = pd.Timestamp.now()
+        arr = np.array([now,
+                        pd.Timedelta('1D')])
+        exp = np.array([now - pd.Timedelta('1D'),
+                        pd.Timedelta('0D')])
+        res = arr - pd.Timedelta('1D')
+        tm.assert_numpy_array_equal(res, exp)
+
+    def test_td_sub_mixed_most_timedeltalike_object_dtype_array(self):
```
jbrockmendel: It seems like this makes `test_td_sub_mixed_timedeltalike_object_dtype_array` above unnecessary.  Am I missing something?

pandas/tests/scalar/timedelta/test_arithmetic.py
illegalnumbers: good catch, done

pandas/tests/scalar/timedelta/test_arithmetic.py
illegalnumbers: good catch, done

pandas/tests/scalar/timedelta/test_arithmetic.py
illegalnumbers: good catch, done

pandas/tests/scalar/timedelta/test_arithmetic.py
illegalnumbers: i was having problems with it last night but let me try again

illegalnumbers: yea i added that in

pandas/tests/scalar/timedelta/test_arithmetic.py
illegalnumbers: no you're right, on further inspection i removed it

illegalnumbers:
@jbrockmendel I think this should be good again!

pandas/_libs/tslibs/timedeltas.pyx
```diff
@@ -541,10 +541,12 @@ def _binary_op_method_timedeltalike(op, name):
 
         elif hasattr(other, 'dtype'):
             # nd-array like
-            if other.dtype.kind not in ['m', 'M']:
-                # raise rathering than letting numpy return wrong answer
+            if other.dtype.kind in ['m', 'M']:
+                return op(self.to_timedelta64(), other)
+            elif other.dtype.kind == 'O':
+                return np.array([op(self, x) for x in other])
```
jorisvandenbossche: @jbrockmendel  I am a bit confused here why simply returning NotImplemented is not sufficient ? 
(I tested it, and that doesn't seem to work. Although with a `datetime.timedelta` it does, and that one *does* return NotImplemented ..)

jbrockmendel: The `pd.Timedelta` version fails because both `arr.__add__(td)` and `td.__radd__(arr)` return `NotImplemented`.  `arr.__add__(td.to_pytimedelta())` returns OK, so presumably it is something on the numpy implementation.

jbrockmendel:
The tests could be parametrized a bit further, but at some point that just becomes equivalent to re-writing the method this PR is implementing.

@jreback LGTM.

illegalnumbers:
I'm super happy to have contributed! This was fun. Sorry it took so long.

pandas/tests/scalar/timedelta/test_arithmetic.py
```diff
@@ -200,6 +200,48 @@ def test_td_rsub_numeric_raises(self):
...
+        # GH 21980
+        now = pd.Timestamp.now()
+        arr = np.array([now,
+                        pd.Timedelta('1D'),
+                        np.timedelta64(2, 'h')])
+        exp = np.array([now - pd.Timedelta('1D'),
+                        pd.Timedelta('0D'),
+                        np.timedelta64(2, 'h') - pd.Timedelta('1D')])
+        res = arr - pd.Timedelta('1D')
```
jreback: can you also tests the rsub (flip the args)

jreback (changes requested):

pandas/tests/scalar/timedelta/test_arithmetic.py
illegalnumbers: i think Timedelta - Timestamp throws an exception iirc

illegalnumbers: which if that is true i think makes sense? unless td - timestamp should be some value

jbrockmendel: I'm guessing jreback's suggestion is to test exactly that, that it raises a TypeError

jreback: yes - you also should test ops and their reverse

pandas/_libs/tslibs/timedeltas.pyx
jorisvandenbossche: Was playing a bit with it, and I think Timedelta behaves differently as datetime.timedelta because of the `__array_priority__` we add to the `_Timedelta` class (by commenting it out, the simple example works), which is needed to get other behaviors working I suppose

jreback: @jorisvandenbossche ok otherwise this PR looks ok, are you suggesting that we remove that?

pandas/tests/scalar/timedelta/test_arithmetic.py
illegalnumbers: done

pandas/_libs/tslibs/timedeltas.pyx
illegalnumbers: I'm also confused.

jreback (approved):

PR #22058, opened 2018-07-25
DOC: Reword doc for filepath_or_buffer in read_csv

### LINKED ISSUE #22057: read_csv Documentation Rendering Poorly
Looks like the `filepath_or_buffer` argument is rendering poorly in the `read_csv` documentation:

image

Should be a pretty easy doc fix:

https://github.com/pandas-dev/pandas/blob/edb71fda022c6a155717e7a25679040ee0476639/pandas/io/parsers.py#L63

gfyoung:
Shortens dtype description following colon and moves it to a more expanded parameter description.

Closes #22057.

cc @WillAyd

### DIFF
--- a/pandas/io/parsers.py
+++ b/pandas/io/parsers.py
@@ -60,11 +60,16 @@ Additional help can be found in the `online docs for IO Tools
 
 Parameters
 ----------
-filepath_or_buffer : str, pathlib.Path, py._path.local.LocalPath or any \
-object with a read() method (such as a file handle or StringIO)
-    The string could be a URL. Valid URL schemes include http, ftp, s3, and
-    file. For file URLs, a host is expected. For instance, a local file could
-    be file://localhost/path/to/table.csv
+filepath_or_buffer : str, path object, or file-like object
+    Any valid string path is acceptable. The string could be a URL. Valid
+    URL schemes include http, ftp, s3, and file. For file URLs, a host is
+    expected. A local file could be: file://localhost/path/to/table.csv.
+
+    If you want to pass in a path object, pandas accepts either
+    ``pathlib.Path`` or ``py._path.local.LocalPath``.
+
+    By file-like object, we refer to objects with a ``read()`` method, such as
+    a file handler (e.g. via builtin ``open`` function) or ``StringIO``.
 %s
 delim_whitespace : boolean, default False
     Specifies whether or not whitespace (e.g. ``' '`` or ``'\t'``) will be


pandas/io/parsers.py
```diff
@@ -60,11 +60,18 @@
...
-    be file://localhost/path/to/table.csv
+filepath_or_buffer : str, path object, or file object
+    Any valid string path is acceptable. The string could be a URL. Valid
+    URL schemes include http, ftp, s3, and file. For file URLs, a host is
+    expected. A local file could be: file://localhost/path/to/table.csv.
+
+    If you want to pass in a path object, pandas accepts the following:
+
+    * pathlib.Path
+    * py._path.local.LocalPath
```
WillAyd: Is this supposed to be private? Not familiar with this but if that's the case don't feel like its worth mention here

gfyoung: The alias is `py.path.local`, so it is in fact public.  However, the alias resolves to that "private class", so no need to provide indirection for something like this.

pandas/io/parsers.py
```diff
@@ -60,11 +60,18 @@
 
 Parameters
 ----------
-filepath_or_buffer : str, pathlib.Path, py._path.local.LocalPath or any \
-object with a read() method (such as a file handle or StringIO)
-    The string could be a URL. Valid URL schemes include http, ftp, s3, and
-    file. For file URLs, a host is expected. For instance, a local file could
-    be file://localhost/path/to/table.csv
+filepath_or_buffer : str, path object, or file object
```
jorisvandenbossche: maybe "file-like" instead of "file" ? as eg StringIO is not really a actual file

pandas/io/parsers.py
jorisvandenbossche: Then I would use the public alias? or do they use it like that as well in their documentation?

Since it are only two, I would maybe put it in a normal paragraph instead of the bullet points (now it seems to get more weight than the other paragraphs)

jorisvandenbossche (commented):
A big improvement anyhow!

pandas/io/parsers.py
jorisvandenbossche: OK, if they use it like that, that's fine

gfyoung:
@jreback @jorisvandenbossche : Made the requested changes, and all is green.  PTAL.

jorisvandenbossche (approved):

PR #22063, opened 2018-07-26
PROPOSAL: use datetime64 and timedelta64 to eval

holymonson:
This PR aims to use datetime64 and timedelta64 in evaluation directly, instead of casting to integer and then casting back. 

Benefits:
1. `values` is saved in `ndarray[datetime64]`, casting to `int` to eval is non-intuitive. It seems back to the time before numpy 1.7 which implements `datetime64`.
2. `numpy.ufunc` can handle all the `dtype` change, e.g.  `datetime64 - datetime64 = timedelta64`. Additonally, nonsense operations are prohibited, e.g. `datetime64 + datetime64`.
3. numpy implements `datetime64('NaT')`, seems `values_mask` and `other_mask` are no longer needed for `nan`s.
4. `Series` can share the same call path with `DataFrame` in `Block`. Now `Series` is using `dispatch_to_index_op` to wrap as `Index`, but `Index` is 1d-array while `DataFrame` is 2d-array, so `DataFrame` can't use the same wrapper. `ndarray([Index])` is `dtype='O'` so can't be benefited from `numpy.ufunc`.

This commit has passed `tests/{internals,frame,dtypes}` , but not `tests/series`, because `dispatch_to_index_op` breaks it.

### DIFF
--- a/pandas/compat/numpy/__init__.py
+++ b/pandas/compat/numpy/__init__.py
@@ -68,6 +68,22 @@ def np_array_datetime64_compat(arr, *args, **kwargs):
     return np.array(arr, *args, **kwargs)
 
 
+# np.percentile should support datetime64 in upsteam. numpy/numpy#11620
+# TODO: change the version if the bug fixed
+if _nlv >= '99.99.99':
+    np_percentile_compat = np.percentile
+else:
+    def np_percentile_compat(values, q, **kw):
+        '''
+        provide compat for np.percentile supporting datetime64
+        '''
+        if values.dtype.kind == 'M':
+            result = np.percentile(values.view('i8'), q, **kw)
+            return result.astype(values.dtype)
+        else:
+            return np.percentile(values, q, **kw)
+
+
 __all__ = ['np',
            '_np_version_under1p10',
            '_np_version_under1p11',
--- a/pandas/core/common.py
+++ b/pandas/core/common.py
@@ -431,11 +431,3 @@ def _pipe(obj, func, *args, **kwargs):
         return func(*args, **kwargs)
     else:
         return func(obj, *args, **kwargs)
-
-
-# TODO: np.percentile not support datetime64, should be fixed in upsteam.
-def percentile(values, q, **kw):
-    if values.dtype.kind == 'M':
-        return np.percentile(values.view('i8'), q, **kw).astype(values.dtype)
-    else:
-        return np.percentile(values, q, **kw)
--- a/pandas/core/common.py
+++ b/pandas/core/common.py
@@ -431,3 +431,11 @@ def _pipe(obj, func, *args, **kwargs):
         return func(*args, **kwargs)
     else:
         return func(obj, *args, **kwargs)
+
+
+# TODO: np.percentile not support datetime64, should be fixed in upsteam.
+def percentile(values, q, **kw):
+    if values.dtype.kind == 'M':
+        return np.percentile(values.view('i8'), q, **kw).astype(values.dtype)
+    else:
+        return np.percentile(values, q, **kw)
--- a/pandas/core/internals/blocks.py
+++ b/pandas/core/internals/blocks.py
@@ -68,7 +68,6 @@ from pandas.core.arrays import Categorical
 from pandas.core.sparse.array import SparseArray
 
 from pandas.core.indexes.datetimes import DatetimeIndex
-from pandas.core.indexes.timedeltas import TimedeltaIndex
 from pandas.core.indexing import check_setitem_lengths
 
 from pandas.io.formats.printing import pprint_thing
@@ -738,7 +737,7 @@ class Block(PandasObject):
                 type(other).__name__,
                 type(self).__name__.lower().replace('Block', '')))
 
-        return values, False, other, False
+        return values, other
 
     def _try_coerce_result(self, result):
         """ reverse of try_coerce_args """
@@ -789,8 +788,7 @@ class Block(PandasObject):
         # try to replace, if we raise an error, convert to ObjectBlock and
         # retry
         try:
-            values, _, to_replace, _ = self._try_coerce_args(self.values,
-                                                             to_replace)
+            values, to_replace = self._try_coerce_args(self.values, to_replace)
             mask = missing.mask_missing(values, to_replace)
             if filter is not None:
                 filtered_out = ~self.mgr_locs.isin(filter)
@@ -841,7 +839,7 @@ class Block(PandasObject):
         # coerce if block dtype can store value
         values = self.values
         try:
-            values, _, value, _ = self._try_coerce_args(values, value)
+            values, value = self._try_coerce_args(values, value)
             # can keep its own dtype
             if hasattr(value, 'dtype') and is_dtype_equal(values.dtype,
                                                           value.dtype):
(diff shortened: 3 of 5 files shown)


jbrockmendel:
> `Series` can share the same call path with `DataFrame` in `Block`

We actually want to go in the other direction.  We want to make `DataFrame` dispatch to the `Series` (or `DatetimeIndex`/`TimedeltaIndex`) implementations.  Is there a viable way to do that here?  Maybe dispatch to `self._holder`?

pandas/core/common.py
```diff
@@ -431,3 +431,11 @@ def _pipe(obj, func, *args, **kwargs):
         return func(*args, **kwargs)
     else:
         return func(obj, *args, **kwargs)
+
+
+# TODO: np.percentile not support datetime64, should be fixed in upsteam.
+def percentile(values, q, **kw):
```
jbrockmendel: Probably belongs in `core.nanops`, definitely not `core.common`.

pandas/core/common.py
```diff
@@ -431,3 +431,11 @@ def _pipe(obj, func, *args, **kwargs):
...
+    if values.dtype.kind == 'M':
+        return np.percentile(values.view('i8'), q, **kw).astype(values.dtype)
```
jbrockmendel: Will this handle `iNaT` correctly?

pandas/core/internals/blocks.py
```diff
@@ -2134,7 +2134,7 @@ def _na_value(self):
 
     @property
     def fill_value(self):
-        return tslibs.iNaT
+        return tslibs.NaT.asm8
```
jbrockmendel: `iNaT` is the standard usage for this.

jreback (changes requested):
what problem are you trying to solve here?

pandas/tests/internals/test_internals.py
```diff
@@ -1250,14 +1250,20 @@ class TestCanHoldElement(object):
...
                 (operator.truediv, 'bool'),
+                (operator.truediv, '<M8[ns]'),
                 (operator.mod, 'i8'),
                 (operator.mod, 'complex128'),
                 (operator.mod, '<M8[ns]'),
                 (operator.mod, '<m8[ns]'),
-                (operator.pow, 'bool')}
+                (operator.pow, 'bool'),
+                (operator.pow, '<m8[ns]'),
+                (operator.pow, '<M8[ns]')}
```
jbrockmendel: These are good fixes.  Can you take it a step further and, instead of skipping, `pytest.raises(TypeError)` for the appropriate cases?

pandas/core/internals/blocks.py
holymonson: `iNaT ` is an `int`, but here wants a `numpy.datetime64('NaT')`.

pandas/core/common.py
holymonson: Can handle `numpy.datetime64('NaT')`, but not `iNaT`.

pandas/core/internals/blocks.py
```diff
@@ -2145,6 +2128,32 @@ def get_values(self, dtype=None):
...
+            pass
+        elif (hasattr(other, 'dtype')
+              and (is_timedelta64_dtype(other) or is_datetime64_dtype(other))):
+            other = getattr(other, 'values', other)
+        else:
+            # coercion issues
+            # let higher levels handle
+            raise TypeError
+
+        return values, other
```
holymonson: Seems `values` can be removed too. The other 2 `_try_coerce_args` overrides don't touch `values`, `_try_coerce_args` might rename to `_try_coerce_other`.

jbrockmendel:
> can only handle 1d-array

Can you expand on why this is a compelling reason?  Maybe there is some characteristic np.ufunc that I'm not considering?

> so `DataFrame` can't use the same wrapper

The idea is that `DataFrame` dispatches column-wise, like it already does in `DataFrame._combine_frame` and `DataFrame._compare_frame`

> Could someone show me the benefits of the Index-dispatch implement?

See #18824.  As you've noticed, there are many operations (particularly arithmetic and comparisons) where `DataFrame` behavior is incorrect, particularly for datetime64 and datetime64tz dtypes.  Until a few months ago, the same was true for Series.  We fixed the Series problems by having it dispatch to the Index subclasses, since the DatetimeIndex and TimedeltaIndex implementations are much more careful.  Now we're down to two implementations instead of three; we still should get it down to one.

pandas/compat/numpy/__init__.py
```diff
@@ -68,6 +68,22 @@ def np_array_datetime64_compat(arr, *args, **kwargs):
     return np.array(arr, *args, **kwargs)
 
 
+# np.percentile should support datetime64 in upsteam. numpy/numpy#11620
+# TODO: change the version if the bug fixed
+if _nlv >= '99.99.99':
```
jbrockmendel: Is this just a placeholder for "distant future"?

pandas/compat/numpy/__init__.py
```diff
@@ -68,6 +68,22 @@ def np_array_datetime64_compat(arr, *args, **kwargs):
...
+    np_percentile_compat = np.percentile
+else:
+    def np_percentile_compat(values, q, **kw):
+        '''
+        provide compat for np.percentile supporting datetime64
+        '''
+        if values.dtype.kind == 'M':
+            result = np.percentile(values.view('i8'), q, **kw)
```
jbrockmendel: Will this handle iNaT correctly?

pandas/compat/numpy/__init__.py
```diff
@@ -68,6 +68,22 @@ def np_array_datetime64_compat(arr, *args, **kwargs):
     return np.array(arr, *args, **kwargs)
 
 
+# np.percentile should support datetime64 in upsteam. numpy/numpy#11620
+# TODO: change the version if the bug fixed
+if _nlv >= '99.99.99':
+    np_percentile_compat = np.percentile
+else:
+    def np_percentile_compat(values, q, **kw):
```
jbrockmendel: the iNat concern aside, this looks like a nice function.  But I don't see why it should be in compat.  Might make more sense in nanops?

pandas/compat/numpy/__init__.py
```diff
@@ -68,6 +68,22 @@ def np_array_datetime64_compat(arr, *args, **kwargs):
...
+        '''
+        provide compat for np.percentile supporting datetime64
+        '''
+        if values.dtype.kind == 'M':
```
jbrockmendel: Does timedelta64 get handled correctly by numpy?

pandas/core/internals/blocks.py
```diff
@@ -789,8 +789,7 @@ def replace(self, to_replace, value, inplace=False, filter=None,
         # try to replace, if we raise an error, convert to ObjectBlock and
         # retry
         try:
-            values, _, to_replace, _ = self._try_coerce_args(self.values,
-                                                             to_replace)
+            values, to_replace = self._try_coerce_args(self.values, to_replace)
```
jbrockmendel: It's hard to tell, but a lot of this looks like it might be unrelated to the np_percentile_compat function.  Is that the case?  If so, that's great (simplifying try_coerce_args etc would be a big win), but should be done separately.

jbrockmendel:
@holymonson can you rebase?  Arithmetic ops have seen a lot of cleanup in the last couple months.

WillAyd:

PR #21930, opened 2018-07-16
DEPR: Series.compress

TomAugspurger:
xref #18262

PR #22064, opened 2018-07-26
TST: suppress deprecation warnings for compress

jorisvandenbossche:
xref https://github.com/pandas-dev/pandas/pull/21930#issuecomment-408068861

In case we find it OK `np.compress` with Series is also deprecated, this PR makes sure the warnings are catched in the tests.


TomAugspurger:
Copying from the other PR, how about we update the deprecation message to 

```
Use `series[condition] or np.asarray(series).compress(condition)`
```

We may also need something that `np.compress(series)` is also deprecated.

```python
Series.compress and np.compress(series, condition) are deprecated.

Use series[condition] or np.asarray(series).compress(condition) instead.
```

jreback:
can you rebase

jorisvandenbossche:
@TomAugspurger edited the deprecation message

TomAugspurger (approved):

PR #22070, opened 2018-07-26
Split fastpath IntegerArray constructor and general purpose constructor

jorisvandenbossche:
Related to the discussion we were having on the PR: https://github.com/pandas-dev/pandas/pull/21160/#discussion_r201209246 (@jreback @shoyer).

Opening this PR for discussion, to *show* what I meant there, to hopefully make it more concrete discussion.

Briefly the changes / rationale:

- Separate the "fastpath" and general purpose constructor (in master everything is passed through `coerce_to_array`:
  - `IntegerArray` is the fastpath if you know you have a proper `values` and `mask`
  - `integer_array` as general purpose constructor function, that eg parses lists, detects NaNs etc (exact API to discuss of course)
- IMO this clear separation makes the code cleaner (in all internal usage we know what we want). In the past we often added a `fastpath` keyword to the constructor, but IMO we should not repeat that pattern in new code (it is overloading the main constructor by combining things).
- This also has the benefit of improved performance. Eg on an array of 1e5 values, the current constructor has an overhead of ca 80 µs, not much of course but that means that a simple slice is 82 instead of 2 µs (of course, this improvement does not need the exact proposal in this PR, you could certainly do a PR adding a fastpath keyword that achieves the same boost).

Note that this is not yet a fully implemented PR:

- Docstrings still need to be updated
- Tests for new behaviour needs to be added
- `coerce_to_array` can be simplified a little bit, as currently the  `mask` argument is nowhere used (unless we foresee use cases for that?)


jreback:
looks interesting - will look tomorrow

jbrockmendel:
+1 on having a fastpath-like constructor, +1 on it not being accessed via a `fastpath` kwarg.  -0.5 on introducing a new naming convention for constructors.

We don't need to require it of downstream EA authors, but for pandas-internal EAs we should have some kind of consistent patterns for EA constructors.  Since `_simple_new`+`_shallow_copy` are already well-established (and used all over the Datetime/Timedelta/Period EA mixins), why not stick with that?

pandas/core/arrays/integer.py
```diff
@@ -76,7 +76,7 @@ def construct_from_string(cls, string):
                         "'{}'".format(cls, string))
 
 
-def to_integer_array(values, dtype=None):
+def integer_array(values, dtype=None, copy=False):
     """
     Infer and return an integer array of the values.
```
jreback: can you update doc-string

pandas/core/arrays/integer.py
```diff
@@ -219,25 +220,33 @@ def __init__(self, values, mask=None, dtype=None, copy=False):
         -------
         IntegerArray
         """
-        self._data, self._mask = coerce_to_array(
-            values, dtype=dtype, mask=mask, copy=copy)
+        if not (isinstance(values, np.ndarray)
+                and np.issubdtype(values.dtype, np.integer)):
```
jreback: is is_integer_dtype

pandas/core/arrays/integer.py
```diff
@@ -219,25 +220,33 @@ def __init__(self, values, mask=None, dtype=None, copy=False):
...
+            raise TypeError("values should be integer numpy array")
+        if not (isinstance(mask, np.ndarray) and mask.dtype == np.bool_):
```
jreback: use is_bool_dtype

pandas/core/arrays/integer.py
```diff
@@ -94,7 +94,8 @@ def to_integer_array(values, dtype=None):
     ------
     TypeError if incompatible types
     """
```
jreback: is there still a ref to to_integer_array? (I see it in the diff, but I also see that you changed it above)

jreback (changes requested):
looks pretty good to me, onboard with this change. may need some doc updates in whatsnew (and add this PR number to the original whatsnew).

pandas/core/arrays/integer.py
```diff
@@ -219,25 +220,33 @@ def __init__(self, values, mask=None, dtype=None, copy=False):
...
 
     def __getitem__(self, item):
         if is_integer(item):
             if self._mask[item]:
                 return self.dtype.na_value
             return self._data[item]
-        return type(self)(self._data[item],
-                          mask=self._mask[item],
-                          dtype=self.dtype)
+        return type(self)(self._data[item], self._mask[item])
```
jorisvandenbossche: @jreback additional question: what do you find of writing `IntegerArray(...)` instead of `type(self)(...)` ? 

Python perfectly allows that (and is the same here, as we don't subclass this one further), and I personally find that easier to read.

jreback: I like using ``type()`` as I expect this to be subclasses for ``BooleanArray``, though it could be that this not needed for that, but want to keep open that possiblitiy.

jorisvandenbossche:
So the failures is because I now changed the code so that a list of numpy scalars is always coerced to int64 (before it took the type of the scalars). 
Such a change is consistent with eg Series, but then we need to more thoroughly pass through the `dtype` to `_from_sequence` in the tests to get this passing.

jorisvandenbossche:
So the failures is because I now changed the code so that a list of numpy scalars is always coerced to int64 (before it took the type of the scalars). 
Such a change is consistent with eg Series, but then we need to more thoroughly pass through the `dtype` to `_from_sequence` in the tests to get this passing.

jorisvandenbossche:
@jreback can you have a look at this again?

jreback:
yes

jorisvandenbossche:
ping

jreback (approved):

jreback:
thanks @jorisvandenbossche nice improvement

jorisvandenbossche:
Thanks!

PR #22071, opened 2018-07-27
Remove unused, avoid uses of deprecated api

jbrockmendel:
`core.dtypes.common.is_period` is confusingly named and never used.  This PR gets rid of it.

The deprecated numpy API that causes zillions of warnings includes `ndarray.data`, so this replaces the old usage with the encouraged usage.

A bunch of `tslibs.util` isn't actually used in tslibs.util, so the relevant functions are moved up to `_libs.util`.

Docstrings for most of what's left in tslibs.util.

Simplify util._checknull by removing an unnecessary try/except.


pandas/_libs/src/numpy_helper.h
```diff
@@ -40,16 +38,10 @@ PANDAS_INLINE const char* get_c_string(PyObject* obj) {
...
-#if PY_VERSION_HEX >= 0x03000000
-    return PyUnicode_FromString(data);
-#else
-    return PyString_FromString(data);
-#endif
-}
-
 void set_array_not_contiguous(PyArrayObject* ao) {
-    ao->flags &= ~(NPY_ARRAY_C_CONTIGUOUS | NPY_ARRAY_F_CONTIGUOUS);
+    // Numpy>=1.8-compliant equivalent to:
```
jreback: we are numpy >= 1.9! can this be simplified?

pandas/_libs/tslibs/util.pxd
```diff
@@ -1,10 +1,18 @@
-from numpy cimport ndarray
```
jreback: wait, we have a tslibs/util.pxd and a util.pxd?

jreback (changes requested):

pandas/_libs/src/numpy_helper.h
jbrockmendel: No.  The existing implementation uses the deprecated numpy API.

pandas/_libs/tslibs/util.pxd
jbrockmendel: When util.pxd was moved from src to tslibs, we added _libs/util.pxd as an alias so as to keep existing cimports

jbrockmendel:
> we are numpy >= 1.9! can this be simplified?

An update/background on the ever-present numpy deprecation warnings.  Based on conversation on the cython [tracker](https://github.com/cython/cython/issues/2498) it looks like

a) numpy more or less forgot why they did that deprecation, have no plans to follow up on it
b) this _could_ be fixed in cython; I'll keep an eye on it but no one should hold their breath
c) _if_ this is fixed in cython, pandas has a handful of places that uses the old API that will need to be updated (many/most of which this PR does)

AFAICT we get the deprecation warning if we do one or more of:
i) cimport numpy as cnp
ii) from numpy cimport ...
iii) `cdef extern from ...` any file that has an `#include <numpy/...>`
iii) b) _unless_ we add to that file the lines:

```
#ifndef
#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
#endif
```

iii) c) But any C file we add this to becomes off-limits for cython files that have numpy cimports (or else we get compile-time errors)

----
The upshot being: there are a bunch of files for which we can get rid of these warnings.

Many of the files only  have numpy cimports because they need to a) call `import_array` in order to use `util.is_integer_object`, b) need `int64_t`, or c) need `ndarray`.

These can all be worked around if we get numpy cimports out of util (which this PR does)

pandas/_libs/tslibs/util.pxd
```diff
@@ -1,10 +1,18 @@
...
-cimport numpy as cnp
-cnp.import_array()
 
-cimport cpython
 from cpython cimport PyTypeObject
 
+cdef extern from *:
```
jreback: we don't need this in util.pxd? (the higher level one)?

jreback: does this change perf?

pandas/core/dtypes/common.py
```diff
@@ -304,34 +304,6 @@ def is_offsetlike(arr_or_obj):
     return False
 
 
-def is_period(arr):
```
jreback: is this not generally exposed to the user / higher level? I don't remember why we added this (and it has the wrong name)

pandas/core/dtypes/common.py
```diff
@@ -304,34 +304,6 @@ def is_offsetlike(arr_or_obj):
...
-    """
-    Check whether an array-like is a periodical index.
-
-    Parameters
-    ----------
-    arr : array-like
-        The array-like to check.
```
jreback: unfortunately I think we need to deprecate this (and remove in next version)

jreback: isn't this change orthogonal to this PR?

jreback (changes requested):

pandas/core/dtypes/common.py
jbrockmendel: > isn't this change orthogonal to this PR?

Sure, can be separated out.  This started out as a grab-bag PR.

> unfortunately I think we need to deprecate this (and remove in next version)

Darn, OK.

pandas/_libs/tslibs/util.pxd
jbrockmendel: > does this change perf?

No.

> we don't need this in util.pxd? (the higher level one)?

a) everything in this file is cimported into the other one, so it is available there.   b) No, this is only used by tslibs.period.

jbrockmendel:
> the warnings in the log seem more than before on interval.pyx. but could be just me.

A couple PRs back that fixed a bunch of warnings, changes to interval.pyx were reverted because they caused failures on windows.  So this file has been cleaned up much less than the others, might seem like more in comparison.

jreback (approved):

PR #22072, opened 2018-07-27
fix for TypeError: unorderable types" in when using set_index with multiple column names

### LINKED ISSUE #15457: "TypeError: unorderable types" in Python3 when column for MultiIndex contains tuple and int

```python
In [2]: df = pd.DataFrame([[2, 1], [4, (1,2)]]).set_index([0, 1])
/home/pietro/nobackup/repo/pandas/pandas/core/categorical.py in __init__(self, values, categories, ordered, name, fastpath)
    288             try:
--> 289                 codes, categories = factorize(values, sort=True)
    290             except TypeError:

/home/pietro/nobackup/repo/pandas/pandas/core/algorithms.py in factorize(values, sort, order, na_sentinel, size_hint)
    360         uniques, labels = safe_sort(uniques, labels, na_sentinel=na_sentinel,
--> 361                                     assume_unique=True)
    362 

/home/pietro/nobackup/repo/pandas/pandas/core/algorithms.py in safe_sort(values, labels, na_sentinel, assume_unique)
    258         # unorderable in py3 if mixed str/int
--> 259         ordered = sort_mixed(values)
    260     else:

/home/pietro/nobackup/repo/pandas/pandas/core/algorithms.py in sort_mixed(values)
    251                            dtype=bool)
--> 252         nums = np.sort(values[~str_pos])
    253         strs = np.sort(values[str_pos])

/usr/lib/python3/dist-packages/numpy/core/fromnumeric.py in sort(a, axis, kind, order)
    821         a = asanyarray(a).copy(order="K")
--> 822     a.sort(axis=axis, kind=kind, order=order)
    823     return a

TypeError: unorderable types: tuple() > int()

<ipython-input-2-b560307721b0> in <module>()
----> 1 df = pd.DataFrame([[2, 1], [4, (1,2)]]).set_index([0, 1])

/home/pietro/nobackup/repo/pandas/pandas/core/frame.py in set_index(self, keys, drop, append, inplace, verify_integrity)
   2907             arrays.append(level)
   2908 
-> 2909         index = MultiIndex.from_arrays(arrays, names=names)
   2910 
   2911         if verify_integrity and not index.is_unique:

/home/pietro/nobackup/repo/pandas/pandas/indexes/multi.py in from_arrays(cls, arrays, sortorder, names)
   1085         from pandas.core.categorical import _factorize_from_iterables
   1086 
-> 1087         labels, levels = _factorize_from_iterables(arrays)
   1088         if names is None:
   1089             names = [getattr(arr, "name", None) for arr in arrays]

/home/pietro/nobackup/repo/pandas/pandas/core/categorical.py in _factorize_from_iterables(iterables)
   2082         # For consistency, it should return a list of 2 lists.
   2083         return [[], []]
-> 2084     return map(list, lzip(*[_factorize_from_iterable(it) for it in iterables]))

/home/pietro/nobackup/repo/pandas/pandas/core/categorical.py in <listcomp>(.0)
   2082         # For consistency, it should return a list of 2 lists.
   2083         return [[], []]
-> 2084     return map(list, lzip(*[_factorize_from_iterable(it) for it in iterables]))

/home/pietro/nobackup/repo/pandas/pandas/core/categorical.py in _factorize_from_iterable(values)
   2054         codes = values.codes
   2055     else:
-> 2056         cat = Categorical(values, ordered=True)
   2057         categories = cat.categories
   2058         codes = cat.codes

/home/pietro/nobackup/repo/pandas/pandas/core/categorical.py in __init__(self, values, categories, ordered, name, fastpath)
    293                     # raise, as we don't have a sortable data structure and so
    294                     # the user should give us one by specifying categories
--> 295                     raise TypeError("'values' is not ordered, please "
    296                                     "explicitly specify the categories order "
    297                                     "by passing in a categories argument.")

TypeError: 'values' is not ordered, please explicitly specify the categories order by passing in a categories argument.
```
#### Problem description

I would understand the ``unorderable types`` error - after all, I'm asking to have different types in the index, and pandas maybe wants to compare them, and it can't (this error obviously only affects Python3). But then,

``` python
In [3]: pd.DataFrame([[2, 'a'], [4, (1,2)]]).set_index([0, 1])
Out[3]: 
Empty DataFrame
Columns: []
Index: [(2, a), (4, (1, 2))]
```
works fine, (even though ``'a' > (1,2)`` raises ``TypeError``), and even allows me to sort the resulting index! Moreover, [I understand this should be supported](https://github.com/pandas-dev/pandas/issues/11565#issuecomment-157691996).

#### Expected Output
``` python
Empty DataFrame
Columns: []
Index: [(4, 1), (2, (1, 2))]
```

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.2.final.0
python-bits: 64
OS: Linux
OS-release: 4.7.0-1-amd64
machine: x86_64
processor: 
byteorder: little
LC_ALL: None
LANG: it_IT.utf8
LOCALE: it_IT.UTF-8

pandas: 0.19.0+473.gf65a641
pytest: 3.0.6
pip: 8.1.2
setuptools: 28.0.0
Cython: 0.23.4
numpy: 1.12.0
scipy: 0.18.1
xarray: None
IPython: 5.1.0.dev
sphinx: 1.4.8
patsy: 0.3.0-dev
dateutil: 2.5.3
pytz: 2015.7
blosc: None
bottleneck: 1.2.0
tables: 3.2.2
numexpr: 2.6.0
feather: None
matplotlib: 2.0.0rc2
openpyxl: 2.3.0
xlrd: 1.0.0
xlwt: 1.1.2
xlsxwriter: 0.9.3
lxml: 3.6.4
bs4: 4.5.1
html5lib: 0.999
httplib2: 0.9.1
apiclient: 1.5.2
sqlalchemy: 1.0.15
pymysql: None
psycopg2: None
jinja2: 2.8
s3fs: None
pandas_datareader: 0.2.1

</details>

### LINKED ISSUE #16718: BUG: to_latex: Problems with MultiIndex / Mixed Index

```python

import pandas as pd
import numpy as np


df = pd.DataFrame()


for val0 in [3,2,1]:
  for val1 in range(val0):
    r0,r1 = np.random.uniform(size = 2)
    df = df.append({"val0" : val0,\
                          "val1":val1, 
                          "r0":r0,\
                          "r1":r1}, ignore_index=True)

print(df.to_latex())
print(df.to_latex(columns = ["r0","r1"],header = ["R0","R1"]))
  
df.set_index(['val0','val1'], inplace=True)

print(df.to_latex())
print(df.to_latex(columns = ["r0","r1"],header = ["R0","R1"]))

```

produces:

```latex

\begin{tabular}{lrrrr}
	\toprule
	{} &        r0 &        r1 &  val0 &  val1 \\
	\midrule
	0 &  0.631918 &  0.445245 &   3.0 &   0.0 \\
	1 &  0.418763 &  0.011685 &   3.0 &   1.0 \\
	2 &  0.742899 &  0.384768 &   3.0 &   2.0 \\
	3 &  0.275884 &  0.849828 &   2.0 &   0.0 \\
	4 &  0.371853 &  0.782338 &   2.0 &   1.0 \\
	5 &  0.763312 &  0.358247 &   1.0 &   0.0 \\
	\bottomrule
\end{tabular}

\begin{tabular}{lrr}
	\toprule
	{} &        R0 &        R1 \\
	\midrule
	0 &  0.631918 &  0.445245 \\
	1 &  0.418763 &  0.011685 \\
	2 &  0.742899 &  0.384768 \\
	3 &  0.275884 &  0.849828 \\
	4 &  0.371853 &  0.782338 \\
	5 &  0.763312 &  0.358247 \\
	\bottomrule
\end{tabular}

\begin{tabular}{llrr}
	\toprule
	&     &        r0 &        r1 \\
	val0 & val1 &           &           \\
	\midrule
	3.0 & 0.0 &  0.631918 &  0.445245 \\
	& 1.0 &  0.418763 &  0.011685 \\
	& 2.0 &  0.742899 &  0.384768 \\
	2.0 & 0.0 &  0.275884 &  0.849828 \\
	& 1.0 &  0.371853 &  0.782338 \\
	1.0 & 0.0 &  0.763312 &  0.358247 \\
	\bottomrule
\end{tabular}

\begin{tabular}{llrr}
	\toprule
	&     &        R0 &        R1 \\
	val0 & val1 &  0.631918 &  0.445245 \\
	\midrule
	3.0 & 0.0 &  0.418763 &  0.011685 \\
	& 1.0 &  0.742899 &  0.384768 \\
	& 2.0 &  0.275884 &  0.849828 \\
	2.0 & 0.0 &  0.371853 &  0.782338 \\
	& 1.0 &  0.763312 &  0.358247 \\
	\bottomrule
\end{tabular}


```
#### Problem description

Hi, 

just found out about the to_latex-functionality. Thumbs up for that! Unfortunately I seem to have stumbled upon a bug - or I'm not getting multiindices right.

I'd like to use to_latex with a multiindex-array, such that the hierarchical parameters don't repeat in the rows. The example code produces 4 tabular sections of latex code. The first two, without the multi-index seem okay. The 3rd, with "multiindex = True" starts to introduce an unnecessary shift in the header names. The multi-index headers are a level below the others. When using the "header" option with this (4th block), the data gets shifted and the last row of the multiindex is missing completely. 

It actually seems to be a problem with index mixing. If the index is created over all columns (df.set_index(['val0','val1',"r0","r1"], inplace=True)), then the table is complete again - there's however on more blank row above all header names.

So, this might be a bug... but I'd also accept that incomplete multi-indices are just not the way to go...

to_latex-bug

#### Expected Output

- all header names on the same level
- multiindex and non-multiindex results not shifted
- all parameters of the multiindex visible

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.5.3.final.0
python-bits: 32
OS: Windows
OS-release: 7
machine: AMD64
processor: AMD64 Family 16 Model 4 Stepping 3, AuthenticAMD
byteorder: little
LC_ALL: None
LANG: en
LOCALE: None.None

pandas: 0.20.2
pytest: 3.0.7
pip: 9.0.1
setuptools: 27.2.0
Cython: 0.25.2
numpy: 1.12.1
scipy: 0.19.0
xarray: None
IPython: 5.3.0
sphinx: 1.5.6
patsy: 0.4.1
dateutil: 2.6.0
pytz: 2017.2
blosc: None
bottleneck: 1.2.1
tables: 3.2.2
numexpr: 2.6.2
feather: None
matplotlib: 2.0.2
openpyxl: 2.4.7
xlrd: 1.0.0
xlwt: 1.2.0
xlsxwriter: 0.9.6
lxml: 3.7.3
bs4: 4.6.0
html5lib: 0.999
sqlalchemy: 1.1.9
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
pandas_gbq: None
pandas_datareader: None

</details>

dickreuter:
- closes #15457

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
-- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types. If (for instance) all labels are tuples of ints, there is no problem (:issue:`15457`)
+- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types, e.g. when some labels are tuples (:issue:`15457`)
 
 I/O
 ^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
-- Fix issue when creating :class:`MultiIndex` from mixed type arrays including tuples. In some cases df.set_index(['A','B']) raised a ``TypeError unorderable types`` when trying to set multiple columns as indices. (:issue:'15457')
+- Fix `TypeError` in Python 3 when creating `MultiIndex` in which some columns are of object type, e.g. when labels are tuples  (:issue:'15457')
 
 I/O
 ^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
-- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types, e.g. when labels are tuples (:issue:`15457`)
+- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types, e.g. when some labels are tuples (:issue:`15457`)
 
 I/O
 ^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
-- Fix `TypeError` in Python 3 when creating `MultiIndex` in which some columns are of object type, e.g. when labels are tuples  (:issue:'15457')
+- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some columns are of object type, e.g. when labels are tuples  (:issue:`15457`)
 
 I/O
 ^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
-- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some columns are of object type, e.g. when labels are tuples  (:issue:`15457`)
+- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types, e.g. when labels are tuples (:issue:`15457`)
 
 I/O
 ^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
-- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types, e.g. when some labels are tuples (:issue:`15457`)
+- Fix ``TypeError`` in Python 3 when creating :class:`MultiIndex` in which some levels have mixed types, e.g. when some labels are tuples (:issue:`15457`)
 
 I/O
 ^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
--
+- Fix issue when creating :class:`MultiIndex` from mixed type arrays including tuples. In some cases df.set_index(['A','B']) raised a ``TypeError unorderable types`` when trying to set multiple columns as indices. (:issue:'15457')
 
 I/O
 ^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
-- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types, e.g. when some labels are tuples (:issue:`15457`)
+- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types. If (for instance) all labels are tuples of ints, there is no problem (:issue:`15457`)
 
 I/O
 ^^^
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -2538,7 +2538,10 @@ def _factorize_from_iterable(values):
                                       ordered=values.ordered)
         codes = values.codes
     else:
-        cat = Categorical(values, ordered=True)
+        # The value of ordered is irrelevant since we don't use cat as such,
+        # but only the resulting categories, the order of which is independent
+        # from ordered. Set ordered to False as default due to issue #15457
+        cat = Categorical(values, ordered=False)
         categories = cat.categories
         codes = cat.codes
     return codes, categories
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -2540,7 +2540,7 @@ def _factorize_from_iterable(values):
     else:
         # The value of ordered is irrelevant since we don't use cat as such,
         # but only the resulting categories, the order of which is independent
-        # from ordered. Set ordered to False as default due to issue #15457
+        # from ordered. Set ordered to False as default. See GH #15457"
         cat = Categorical(values, ordered=False)
         categories = cat.categories
         codes = cat.codes
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -2540,7 +2540,7 @@ def _factorize_from_iterable(values):
     else:
         # The value of ordered is irrelevant since we don't use cat as such,
         # but only the resulting categories, the order of which is independent
-        # from ordered. Set ordered to False as default. See GH #15457"
+        # from ordered. Set ordered to False as default. See GH #15457
         cat = Categorical(values, ordered=False)
         categories = cat.categories
         codes = cat.codes
--- a/pandas/tests/frame/test_alter_axes.py
+++ b/pandas/tests/frame/test_alter_axes.py
@@ -261,7 +261,7 @@ class TestDataFrameAlterAxes(TestData):
         comp = comp.tz_localize(None)
         tm.assert_numpy_array_equal(result.values, comp.values)
 
-        # list of datetimes with a tzg
+        # list of datetimes with a tz
         df['D'] = i.to_pydatetime()
         result = df['D']
         assert_series_equal(result, expected, check_names=False)
--- a/pandas/tests/frame/test_alter_axes.py
+++ b/pandas/tests/frame/test_alter_axes.py
@@ -394,8 +394,9 @@ class TestDataFrameAlterAxes(TestData):
 
     def test_set_index_tuple(self):
         # test for fix of issue #15457, where the following raised a TypeError
-        df = DataFrame([[2, 1, 2], [4, (1, 2), 3]]).set_index([0, 1])
-        assert df.to_dict() == {2: {(2, 1): 2, (4, (1, 2)): 3}}
+        expected = pd.DataFrame([[2, 1], [4, (1, 2)]]).set_index([0, 1]).index
+        result = pd.MultiIndex.from_tuples([(2, 1), (4, (1, 2))], names=(0, 1))
+        tm.assert_index_equal(expected, result)
 
     # Renaming
 
--- a/pandas/tests/frame/test_alter_axes.py
+++ b/pandas/tests/frame/test_alter_axes.py
@@ -392,6 +392,11 @@ class TestDataFrameAlterAxes(TestData):
 
         assert new_index.freq == index.freq
 
+    def test_set_index_tuple(self):
+        # test for fix of issue #15457, where the following raised a TypeError
+        df = DataFrame([[2, 1, 2], [4, (1, 2), 3]]).set_index([0, 1])
+        assert df.to_dict() == {2: {(2, 1): 2, (4, (1, 2)): 3}}
+
     # Renaming
 
     def test_rename(self):
--- a/pandas/tests/frame/test_alter_axes.py
+++ b/pandas/tests/frame/test_alter_axes.py
@@ -261,7 +261,7 @@ class TestDataFrameAlterAxes(TestData):
         comp = comp.tz_localize(None)
         tm.assert_numpy_array_equal(result.values, comp.values)
 
-        # list of datetimes with a tz
+        # list of datetimes with a tzg
         df['D'] = i.to_pydatetime()
         result = df['D']
         assert_series_equal(result, expected, check_names=False)
@@ -391,12 +391,6 @@ class TestDataFrameAlterAxes(TestData):
         result = df.set_index(new_index)  # noqa
 
         assert new_index.freq == index.freq
-
-    def test_set_index_tuple(self):
-        # test for fix of issue #15457, where the following raised a TypeError
-        expected = pd.DataFrame([[2, 1], [4, (1, 2)]]).set_index([0, 1]).index
-        result = pd.MultiIndex.from_tuples([(2, 1), (4, (1, 2))], names=(0, 1))
-        tm.assert_index_equal(expected, result)
 
     # Renaming
 
--- a/pandas/tests/indexes/multi/test_constructor.py
+++ b/pandas/tests/indexes/multi/test_constructor.py
@@ -3,12 +3,11 @@
 import re
 
 import numpy as np
-import pytest
-from pandas._libs.tslib import Timestamp
-
 import pandas as pd
 import pandas.util.testing as tm
+import pytest
 from pandas import Index, MultiIndex, date_range
+from pandas._libs.tslib import Timestamp
 from pandas.compat import lrange, range
 from pandas.core.dtypes.cast import construct_1d_object_array_from_listlike
 
@@ -24,7 +23,7 @@ def test_constructor_single_level():
 
 def test_constructor_no_levels():
     tm.assert_raises_regex(ValueError, "non-zero number "
-                                       "of levels/labels",
+                           "of levels/labels",
                            MultiIndex, levels=[], labels=[])
     both_re = re.compile('Must pass both levels and labels')
     with tm.assert_raises_regex(TypeError, both_re):
@@ -57,7 +56,7 @@ def test_constructor_mismatched_label_levels(idx):
     labels = [np.array([1]), np.array([2]), np.array([3])]
     levels = ["a"]
     tm.assert_raises_regex(ValueError, "Length of levels and labels "
-                                       "must be the same", MultiIndex,
+                           "must be the same", MultiIndex,
                            levels=levels, labels=labels)
     length_error = re.compile('>= length of level')
     label_error = re.compile(r'Unequal label lengths: \[4, 2\]')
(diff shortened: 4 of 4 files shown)


gfyoung:
Going to need tests for this as well.

toobaz:
> closes #16718

Thanks for your attempt! But is this really the bug you're after?! (feel free to edit the description)

toobaz (changes requested):
The fix is good, some comments.

jreback (changes requested):
always write tests first - how else can you tell if you fixed something?

dickreuter:
added test and better doc

dickreuter:
Woudl like to check it locally but building pandas seems impossible. Is here no additional info what version of visual studio dev tools is needed exactly? https://stackoverflow.com/questions/51566895/unable-to-build-pandas-from-source-cl-exe-failed-with-exit-status-2

TomAugspurger:
http://pandas-docs.github.io/pandas-docs-travis/contributing.html#installing-a-c-compiler And 
https://devguide.python.org/setup/#build-dependencies may help

dickreuter:
I don't see how the error on the ci is related to my change (especially also since the tests passed yesterday)

toobaz:
> I don't see how the error on the ci is related to my change (especially also since the tests passed yesterday)

At least one is explained by
```
pandas/tests/arrays/categorical/test_indexing.py:131:41: W292 no newline at end of file
```
(see also the automatic [comment](https://github.com/pandas-dev/pandas/pull/22072#issuecomment-408561159)  by @pep8speaks )

dickreuter:
Still getting an error which I can't see how it's related to this: BufferError: could not get scalar buffer information

doc/source/whatsnew/v0.23.4.txt
```diff
@@ -51,7 +51,7 @@ Bug Fixes
 
 **Categorical**
 
--
+- Fix issue when creating :class:`MultiIndex` from mixed type arrays including tuples (:issue:'15457')
```
jreback: can you be even more explicit here, a user won't under stand this.

pandas/core/arrays/categorical.py
```diff
@@ -2520,7 +2520,10 @@ def _factorize_from_iterable(values):
                                       ordered=values.ordered)
         codes = values.codes
     else:
-        cat = Categorical(values, ordered=True)
+        # The value of ordered is irrelevant since we don't use cat as such,
```
jreback: can you add the issue reference

pandas/tests/arrays/categorical/test_indexing.py
```diff
@@ -121,3 +121,11 @@ def test_get_indexer_non_unique(self, idx_values, key_values, key_class):
 
             tm.assert_numpy_array_equal(expected, result)
             tm.assert_numpy_array_equal(exp_miss, res_miss)
+
+
+class TestSetIndex(object):
```
jreback: put in pandas/tests/frame/test_alter_axes.py.

put in a full comparison on the result table and use assert_fram_equal

jreback (changes requested):

pandas/tests/arrays/categorical/test_indexing.py
dickreuter: `df = DataFrame([[2, 1, 2], [4, (1, 2), 3]]).set_index([0, 1])` gives me
Out[90]: 
             2
0 1        
2 1        2
4 (1, 2)  3

How I construct this directly from the DataFrame constructor? The first fow seems empty?
df.to_dict() gives `{2: {(2, 1): 2, (4, (1, 2)): 3}}`

I tried:
DataFrame(['',2,3],
          columns=['2'],
          index=[['0','2','4'], ['1','1','(1,2)']])

but it's not equal (even though it looks equal), but 
DataFrame shape mismatch
[left]:  (2, 1)
[right]: (3, 1)

doc/source/whatsnew/v0.23.4.txt
```diff
@@ -51,7 +51,7 @@ Bug Fixes
 
 **Categorical**
 
--
+- Fix issue when creating :class:`MultiIndex` from mixed type arrays including tuples (:issue:'15457'). In some cases df.set_index(['A','B']) raised a TypeError unorderable types when trying to set multiple columns as indices.
```
jreback: move to 0.24

jreback (changes requested):

pandas/tests/frame/test_alter_axes.py
```diff
@@ -1186,3 +1186,11 @@ def test_set_axis_prior_to_deprecation_signature(self):
             with tm.assert_produces_warning(FutureWarning):
                 result = df.set_axis(axis, list('abc'), inplace=False)
             tm.assert_frame_equal(result, expected[axis])
+
```
jreback: don't add a new class, rather put this test next to the other set_index tests

jreback (changes requested):

pandas/tests/arrays/categorical/test_indexing.py
toobaz: > but it's not equal (even though it looks equal), but

Yes, they look equal, but you can see from how you built them they are not. ``[0, 1]`` are the index names, not index values.

```
pd.DataFrame([2,3], columns=['2'],
             index=pd.MultiIndex.from_arrays([['2','4'], ['1','(1,2)']], names=[0, 1]))
```
is what you're looking for.

toobaz: > I also tried

The original ``df`` has no strings at all, just integers. Why are you putting strings in the other (e.g. ``'2'``, ``'(1,2)'``?!)

dickreuter: Removing the two strings I get `ValueError: setting an array element with a sequence.`

toobaz: > Removing the two strings I get ValueError: setting an array element with a sequence.

OK, ``pd.MultiIndex.from_tuples([(2, 1), (4, (1,2))], names=[0, 1])``

pandas/tests/frame/test_alter_axes.py
```diff
@@ -392,6 +392,11 @@ def test_dti_set_index_reindex(self):
 
         assert new_index.freq == index.freq
 
+    def test_set_index_tuple(self):
+        # test for fix of issue #15457, where the following raised a TypeError
+        df = DataFrame([[2, 1, 2], [4, (1, 2), 3]]).set_index([0, 1])
+        assert df.to_dict() == {2: {(2, 1): 2, (4, (1, 2)): 3}}
```
jreback: use assert_frame_equal, IOW you must construct the expected frame. use ``result=`` and ``expected=`` here

jreback (changes requested):

pandas/tests/frame/test_alter_axes.py
dickreuter: Comparing from_arrays with from_tuples? Not sure I understand. As you said, this already works.

jreback (changes requested):

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -430,7 +430,7 @@ Bug Fixes
 Categorical
 ^^^^^^^^^^^
 
--
+- Fix issue when creating :class:`MultiIndex` from mixed type arrays including tuples (:issue:'15457'). In some cases df.set_index(['A','B']) raised a TypeError unorderable types when trying to set multiple columns as indices.
```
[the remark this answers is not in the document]
dickreuter: You want thesame issue number mentioned twice? I'm not aware of any second issue number.

dickreuter:
getting `ValueError: setting an array element with a sequence` when I test that. Not sure my fix is related to this bug?

dickreuter:
@toobaz I reverted to the test via dict as otherwise line 908 of cast.py causes an error when trying to create an array from  [1, (1, 2)], raising {ValueError}setting an array element with a sequence.

pandas/tests/frame/test_alter_axes.py
dickreuter: @toobaz I reverted to the test via dict as otherwise line 908 of cast.py causes an error when trying to create an array from [1, (1, 2)], raising {ValueError}setting an array element with a sequence. This may be an unrelated issue but not sure the test can be constructed in that way at the moment.

dickreuter:
@toobaz @jreback What's the next step? Can this be merged? tx

gfyoung:
https://travis-ci.org/pandas-dev/pandas/jobs/413344917#L3557

@dickreuter : You have linter failures.  I would fix those first before pinging.

pandas/tests/indexes/multi/test_constructor.py
```diff
@@ -463,3 +463,4 @@ def test_tuples_with_name_string():
         pd.Index(li, name='abc')
     with pytest.raises(ValueError):
         pd.Index(li, name='a')
+
```
gfyoung: Can you undo the change in this file?

(it's extraneous to your PR)

pandas/tests/frame/test_alter_axes.py
jreback: @dickreuter this has to be an assert on a frame, otherwise this test is not testing what we need here

jreback (changes requested):

pandas/tests/frame/test_alter_axes.py
toobaz: > {ValueError}setting an array element with a sequence

@dickreuter You are right, sorry. This should work instead:
```
expected = pd.DataFrame([[2, 1], [4, (1, 2)]]).set_index([0, 1]).index
result = pd.MultiIndex.from_tuples([(2, 1), (4, (1,2))], names=(0, 1))
```

@jreback see my comment below, the bug is unrelated to ``DataFrame``

pandas/core/arrays/categorical.py
```diff
@@ -2538,7 +2538,10 @@ def _factorize_from_iterable(values):
...
+        # but only the resulting categories, the order of which is independent
+        # from ordered. Set ordered to False as default due to issue #15457
```
toobaz: I wouldn't say it's "due to", as it would be worth doing even if there was no actual error. I think "See GH #15457" is enough.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -623,7 +623,7 @@ MultiIndex
 ^^^^^^^^^^
 
 - Removed compatibility for MultiIndex pickles prior to version 0.8.0; compatibility with MultiIndex pickles from version 0.13 forward is maintained (:issue:`21654`)
--
+- Fix issue when creating :class:`MultiIndex` from mixed type arrays including tuples (:issue:'15457'). In some cases df.set_index(['A','B']) raised a ``TypeError unorderable types`` when trying to set multiple columns as indices.
```
toobaz: We don't create the MultiIndex "from mixed type arrays", and I think the two sentences can be merged.

The following should work: "Fix ``TypeError`` in Python 3 when creating ``MultiIndex`` in which some labels are tuples"

Use backticks (not single quotes) around the issue number.

pandas/tests/frame/test_alter_axes.py
```diff
@@ -392,6 +392,12 @@ def test_dti_set_index_reindex(self):
 
         assert new_index.freq == index.freq
 
+    def test_set_index_tuple(self):
+        # test for fix of issue #15457, where the following raised a TypeError
```
toobaz: Please follow the style in other tests, e.g.
```
# GH 15457 : explanation
```
or
```
# GH 15457
# explanation
```

toobaz (changes requested):

pandas/core/arrays/categorical.py
```diff
@@ -2538,7 +2538,10 @@ def _factorize_from_iterable(values):
                                       ordered=values.ordered)
         codes = values.codes
     else:
-        cat = Categorical(values, ordered=True)
+        # The value of ordered is irrelevant since we don't use cat as such,
+        # but only the resulting categories, the order of which is independent
+        # from ordered. Set ordered to False as default. See GH #15457"
```
toobaz: Spurious double quote

doc/source/whatsnew/v0.24.0.txt
jreback: pls add the issue number

pandas/tests/frame/test_alter_axes.py
```diff
@@ -392,6 +392,12 @@ def test_dti_set_index_reindex(self):
...
+        expected = pd.DataFrame([[2, 1], [4, (1, 2)]]).set_index([0, 1]).index
+        result = pd.MultiIndex.from_tuples([(2, 1), (4, (1, 2))], names=(0, 1))
+        tm.assert_index_equal(expected, result)
```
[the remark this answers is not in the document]
jreback: well now that the index is good. let's just compare the DataFrame itself (after all that's the OP). not averse to @toobaz comment, but we need to have a test right here

jreback (changes requested):

doc/source/whatsnew/v0.24.0.txt
dickreuter: The problem seems to occur as well when the column types I want to index are of type object. Not sure why this is related. In that case they are unorderable (I think), but it should still be possible to set them as index I assume? I'll try to create a test for that as well.

toobaz:
> let's just compare the DataFrame itself (after all that's the OP)

@jreback My suggestion was based on the fact that the OP's error comes from an internal public API call. By testing this call directly (the index creation), we are not looking at the bug from another point of view: it is a test to exactly the same problematic code path, just better classified..

doc/source/whatsnew/v0.24.0.txt
toobaz: > The problem seems to occur as well when the column types I want to index are of type object.

@dickreuter Not sure I follow you: when "_some labels are tuples_", then the column types are _necessarily_ of type ``object``. That's the only dtype that can hold tuples.

> pls add the issue number

@jreback it's already there

jreback (approved):

jreback:
rebased

@toobaz merge when satisfied

doc/source/whatsnew/v0.24.0.txt
dickreuter: @toobaz Yes so the description should make reference to the object type rather than the tuple, because the former is the actual cause for the error.

toobaz: > the description should make reference to the object type rather than the tuple, because the former is the actual cause for the error.

Not sure what makes you think this, but
 
```
problem = (0, 1)
pd.DataFrame([[2, 1], [4, problem]]).set_index([0, 1])
```
works just fine if you replace the first line with ``problem = 'a'``.

(The ultimate reason why tuples are a problem is #22080.)

pandas/tests/indexes/multi/test_constructor.py
toobaz: @gfyoung (unless I'm wrong) this comment ended up showing here by mistake?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
--
+- Fix `TypeError` in Python 3 when creating `MultiIndex` in which some columns are of object type, e.g. when labels are tuples  (:issue:'15457')
```
toobaz: Apart from my comment on ``object type`` ( https://github.com/pandas-dev/pandas/pull/22072/review_comment/create#discussion_r209218334 ), you have lost the ":class:" before "`MultiIndex`".

doc/source/whatsnew/v0.24.0.txt
dickreuter: But if you have objects (I had ctype items) in the rows of column 'a', then it will fail. That's what I mean. It's not just the column name that matters.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
--
+- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some columns are of object type, e.g. when labels are tuples  (:issue:`15457`)
```
toobaz: @dickreuter OK, I think I see your point. What I'm saying is "not all ``object``" dtype give problems, what you are saying is "other ``object`` dtypes than tuples give problems". I think a more precise description then is "in which some levels have mixed types", as I guess happens in your case (no error comes if _all_ labels are ctype items, right?).

Then i think we're ready.

dickreuter: I believe in the example I first hit the error there where 3 columns. All of them were dtype and all the rows were homogenous of the same type as well.

dickreuter: @toobaz When you say mixed type, is the column name also relevant for that for each for? If yes, then I assume it’s a mixed type because the column names were strings in my case.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
--
+- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types, e.g. when labels are tuples (:issue:`15457`)
```
toobaz: Almost there... "when **_some_** labels are tuples". If (for instance) all labels are tuples of ints, there is no problem.

doc/source/whatsnew/v0.24.0.txt
toobaz: Your new version is almost ready (see my last comment), but for the records:

> I believe in the example I first hit the error there where 3 columns. All of them were dtype and all the rows were homogenous of the same type as well.

The error only comes out when two elements (in a same column) cannot be ordered, which in Python 3 happens when they are of non-comparable types. If you got an error with columns which homogeneous types, I doubt it was the same error.
Notice that I'm referring to (Python) _types_, not _dtypes_. If you put mixed _types_ in a single column, then its dtype will be just ``object``, but it is still a mixed type column.

> When you say mixed type, is the column name also relevant for that for each for?

No, column names are stored separately and should be unrelated to this bug and to any issue with mixed type columns/levels.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
--
+- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types. If (for instance) all labels are tuples of ints, there is no problem (:issue:`15457`)
```
toobaz: @dickreuter the previous one was almost OK, this one is not... in the whatsnew, we don't need examples in which "there is no problem" :-)

Please revert to the previous one, just adding "some", that is
```
Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types, e.g. when some labels are tuples (:issue:`15457`)
```

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -626,7 +626,7 @@ MultiIndex
 
 - Removed compatibility for :class:`MultiIndex` pickles prior to version 0.8.0; compatibility with :class:`MultiIndex` pickles from version 0.13 forward is maintained (:issue:`21654`)
 - :meth:`MultiIndex.get_loc_level` (and as a consequence, ``.loc`` on a :class:``MultiIndex``ed object) will now raise a ``KeyError``, rather than returning an empty ``slice``, if asked a label which is present in the ``levels`` but is unused (:issue:`22221`)
--
+- Fix `TypeError` in Python 3 when creating :class: `MultiIndex` in which some levels have mixed types, e.g. when some labels are tuples (:issue:`15457`)
```
toobaz: Double backticks around ``TypeError``, no space after ``:class:``

dickreuter:
@toobaz Why does travis fail with "Unable to download 3.5 archive. The archive may not exist. Please consider a different version"?

toobaz (approved):

toobaz:
Thanks @dickreuter !

gfyoung:
@dickreuter : FYI, those Travis failures that you saw [earlier](https://github.com/pandas-dev/pandas/pull/22072#issuecomment-412297820) are spurious.  I restarted the builds to get this PR merged 🙂

Issue #20588, opened 2018-04-02
STYLE: Add additional flake8 plugins


jschendel:
xref https://github.com/pandas-dev/pandas/pull/20563#issuecomment-377648415

Might be nice to add some additional `flake8` plugins to check various things:
  - Enforce using new-style classes?
     - The [hacking](https://github.com/openstack-dev/hacking) package looks to do this (cc @mroeschke)
     - Added a related check to `lint.sh` in #20563, which can be removed if this is added
  - Enforce PEP8 naming conventions for classes?
    - Doesn't look like this is currently enforced

Any other ideas are welcome.

TomAugspurger:
Import order :)

Linters for the docstrings and code sections of the prose docs (flake8 or otherwise) would be great.

jbrockmendel:
Couple of options for import order.
https://pypi.org/project/flake8-imports/
https://pypi.org/project/flake8-isort/
https://github.com/PyCQA/flake8-import-order  <-- tentatively looks like the best bet, need to figure out the config options

Docstrings (@datapythonista I'm going to back you up whatever your opinion is on this one)
https://pypi.org/project/flake8-docstrings/  <-- wraps pydocstyle
https://pypi.org/project/flake8-rst-docstrings/  <-- wraps docutils

Misc
https://pypi.org/project/flake8-self/  <-- check for private variable access
https://pypi.org/project/flake8-pep3101/  <-- old-style string formatting (514 hits ATM)
https://pypi.org/project/flake8-builtins/    <-- check for overriding built-in names
https://pypi.org/project/flake8-comprehensions/  <-- list/set/dict comprehensions (or do we already lint for that well enough?)

mroeschke:
+1 for flake8-comprehensions. Would close #18464 and replace some ad hoc regexes to check for unnecessary comprehensions.

jbrockmendel:
We've pushed pretty far down this road, and have been implementing custom linting for a while.  Is there anything else to do here, or closeable?

MarcoGorelli:
IMO this can be closed, any extra plugins can be discussed individually

Issue #11954, opened 2016-01-04
Style: removing unnecessary usages of `\` from the codebase


wesm:
I've found that using parentheses (which allow for automatic continuation) instead of `\` results in a lot more readable code. The lone exception is using `\` with long strings and avoiding a line break at the start of the string.

Worse:

```
return GOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOD and LET and THE and HATE and FLOW\
    and THROUGH and YOU
```

Better:

```
return (GOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOD and LET and THE and HATE and FLOW
        and THROUGH and YOU)
```

OK:

```
return """\
GOOOOOOOOOOOOOOOOOOOOOOOOOOOOOOD and LET and THE and HATE and FLOW
and THROUGH and YOU"""
```

jorisvandenbossche:
xref https://github.com/pydata/pandas/pull/11951#issuecomment-168562469

rockg:
Part of the underlying issue here is some editors nowadays (e.g., pycharm) have automatic behavior when it comes to continuation and uses '\' continuation by default.

wesm:
Can that be disabled?

rockg:
There's an automatic wrap option that will wrap lines automatically with '\' but if the statement was started with '(' then it will wrap without '\'.  If one does not wrap automatically then the syntax is up to the user.  If a portion of the statement is moved to the next line then '\' is used if not already wrapped in '()'.  There is no way to add '()' to these statements by default but this should probably be an editor option.  PyDev has no such behavior and relies on the user putting in the appropriate continuation and I'm unsure of other editors.

rockg:
So, I think if people get in the habit of using '()' at the outset the editor continuation won't be an issue.

jbrockmendel:
@jorisvandenbossche does `black` take care of this for us?

jorisvandenbossche:
For cases in python code, yes. So I think we can close this. 
There are some cases in docstrings where we still need it (but that's not the cases discussed above I think)

PR #22075, opened 2018-07-27
CLN/STYLE: Lint comprehensions

### LINKED ISSUE #18464: STYLE: Lint for comprehensions that use the list/set/dict functions
xref #18383

Ideally comprehensions should follow the general pattern:

`list(x for x in iterator)` --> `[x for x in iterator]`
`set(x for x in iterator)` --> `{x for x in iterator}`
`dict((x,x) for x in iterator)` --> `{x: x for x in iterator}`

Adding a lint check for the prior pattern in `ci/lint.sh` would be great.

mroeschke:
- [x] closes #18464
- [x] tests added / passed

Using flake8-comprehensions (xref https://github.com/pandas-dev/pandas/issues/20588#issuecomment-405057018) to lint for unnecessary comprehensions. I just added this to the 2.7-travis build since it appears to be our linting build.


jbrockmendel:
Syntax error in one of the edits.  +1 on the concept.

ci/travis-27.yaml
```diff
@@ -9,6 +9,7 @@ dependencies:
   - fastparquet
   - feather-format
   - flake8=3.4.1
+  - flake8-comprehensions
```
jreback: need to add this to the dev deps as well

jbrockmendel:
Tons of flake complaints, may need to implement gradually.

jbrockmendel:
Now it’s mostly upset about lambdas.  OTOH it’s reassuring to see mine isn’t the only PR that pytables is messing up.

mroeschke:
Is there a predefined set of lint codes we ignore?

I am a little puzzled why these lint checks now started to fire; I figure it has something to do with the scope of checks when `--ignore` is specified.

jbrockmendel:
> Is there a predefined set of lint codes we ignore?

in setup.cfg

jbrockmendel:
Possibly for the while-were-at-it pile: 

#11954 checks for unnecessary backslashes (could this be done with just a grep?)
Enable checks for bare “except:” statements in a subset of files and gradually expand?

jreback:
we have a fixed version of flake8 (an open issue about this too) - which doesn’t check bare excepts

jreback:
pls rebase

mroeschke:
The travis failure is from the reoccuring `ResourceWarning` that appears in `test_dialect_conflict`

jreback:
thanks @mroeschke 

if you think need to open additional issues for the skipped pep codes, pls do. alternatively maybe just add some docs on them on why we need to keep them there (e.g. so next person isn't wondering why those particular codes).

ci/lint.sh
```diff
@@ -10,42 +10,42 @@ if [ "$LINT" ]; then
...
-    flake8 pandas --filename=*.py --exclude pandas/_libs/src
+    flake8 pandas --filename=*.py --exclude pandas/_libs/src --ignore=C405,C406,C408,C409,C410,E402,E731,E741,W503
     if [ $? -ne "0" ]; then
         RET=1
     fi
     echo "Linting *.py DONE"
 
     echo "Linting setup.py"
-    flake8 setup.py
+    flake8 setup.py --ignore=C405,C406,C408,C409,C410,E402,E731,E741,W503
```
jreback: e.g. here if you'd document why these codes (general comments are enough), though if specific ones *can* be removed pls do that. also *maybe* want to add these to ``setup.cfg`` ?

ci/environment-dev.yaml
```diff
@@ -6,6 +6,7 @@ dependencies:
   - Cython>=0.28.2
   - NumPy
   - flake8
+  - flake8-comprehensions
```
jreback: I think need to run and commit scripts/convert_deps.py to have requirements_dev.txt updated. can you PR that?

PR #22083, opened 2018-07-27
BUG: maximum recursion error when replacing empty lists

### LINKED ISSUE #19266: Series/DataFrame.replace crashes without exception in some cases (concerning lists)
I have a Series of lists that I need to work with, and discovered some crashes without exceptions. Here's an overview of the behaviour (which is inconsistent to me).

```
import pandas as pd
import numpy as np
ser = pd.Series([['a', 'b'], np.nan, [1]])

ser.replace({np.nan : []}) # crashes w/o exception
ser.replace({np.nan : 'dummy'}) # works
# 0    [a, b]
# 1     dummy
# 2       [1]
# dtype: object

ser.replace({np.nan : ['dummy']}) # why does this unwrap?
# 0    [a, b]
# 1     dummy
# 2       [1]
# dtype: object

ser.replace({np.nan : ['dummy', 'alt']}) # crashes w/o exception

ser.fillna([]) # raises
ser.fillna({1 : []}) # this works!
# 0    [a, b]
# 1        []
# 2       [1]
# dtype: object

ser.fillna({1 : ['dummy', 'alt']}) # works as well!
# 0          [a, b]
# 1    [dummy, alt]
# 2             [1]
# dtype: object

# Dataframe has exact same behaviour as Series above
df = pd.DataFrame({'col' : ser})
df.replace({np.nan : []}) # crashes w/o exception
...
```

I agree that _interpreting_  a list as the argument to `.replace` makes no sense. But I don't understand why it's not possible to `fillna` a list (cf. other people asking this question https://stackoverflow.com/q/33199193/2965879).

There's no reason in my opinion why `.replace({np.nan : ['dummy', 'alt']})` or `.replace({np.nan : []})`couldn't work in principle - it's very clear what the intent is. Furthermore, it already works like that in `fillna` (of course with different interpretation of the dict-key). But even if forbidding lists is a design decision, the call shouldn't just crash, but raise an exception, at least.

In my case, I have to design an API that does replacements (as pre-/post-processing around actual work) with `pandas` in the background, and I'd like to be able to just pass through legal `dict`s to `Series.replace` / `DataFrame.replace` - e.g. `{r'\s*' : np.nan}` or `{np.nan : []}`. Otherwise I have to inspect every passed replacement parameter (with all the overhead that comes with allowing both `{search : replace}` and `{column : {search : replace}}`), extract special cases, and build complicated wrappers like in the answers of the above SO question.

Versions are the most recent on conda, details below.
<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.22.0
pytest: 3.2.1
pip: 9.0.1
setuptools: 38.4.0
Cython: 0.26.1
numpy: 1.13.3
scipy: 1.0.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: 1.6.3
patsy: 0.4.1
dateutil: 2.6.1
pytz: 2017.3
blosc: None
bottleneck: 1.2.1
tables: 3.4.2
numexpr: 2.6.4
feather: None
matplotlib: 2.1.0
openpyxl: 2.4.8
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.2
lxml: 4.1.0
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.1.13
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

### LINKED ISSUE #21977: Pandas Series: maximum recursion error when replacing empty lists
I have a pandas Series containing lists. I wanted to replace empty lists with NaN. My first approach was using .replace, but that unexpectedly gave me a maximum recursion error.

I am using pandas 0.21.1 on Python 3.4, but other users had the same problem on 0.23

```
import numpy as np
import pandas as 
ts = pd.Series([[1], [2, 3], [], [4]])
ts.replace([], np.nan)

RuntimeError: maximum recursion depth exceeded in comparison
```

I achieved the desired output using:

`ts[ts.apply(len) == 0] = np.nan`

but I don't understand the maximum recursion error

minggli:
- [x] closes #21977 
- [x] closes #19266 
- [x] tests added / passed
- [x] whatsnew entry

Internal mechanism in replace method is intertwined and can cause RecursionError (not very helpful). Discussed in #21977, there doesn't appear to be first class support for ExtensionArray yet and this issue is right now treated as error handling, having investigated treating it as new use case but not easily co-exist without large change.

The RecursionError is caused by `Block.replace` and `ObjectBlock.replace` calling each other under certain circumstances such as in the linked issues.

New approach is now to prevent re-casting of blocks to object blocks if it is already an ObjectBlock and instead raise the error caught as it is, because if it's an ObjectBlock already then ValueError and TypeError raised won't go away when re-casted to object and re-try, resulting in infinite loop and RecursionError.


pandas/core/internals/blocks.py
```diff
@@ -1000,8 +999,8 @@ def putmask(self, mask, new, align=True, inplace=False, axis=0,
                 if not (mask.shape[-1] == len(new) or
                         mask[mask].shape[-1] == len(new) or
                         len(new) == 1):
-                    raise ValueError("cannot assign mismatch "
```
jreback: does this need to change? (or is this what is actually causing the issue); this is not exposed to the user, correct?

jreback (changes requested):
can you add the example as a test

pandas/core/internals/blocks.py
minggli: Yes this ValueError caused Block.replace and ObjectBlock.replace to form infinite loop and ultimately raise RecursionError. It's because Block.replace catches exception include ValueError and delegate replace to ObjectBlock if fails. But ObjectBlock.replace also uses super(ObjectBlock, self), hence the loop. The exception handling includes TypeError and ValueError and they are both dependent on (as appear to me) by other casting cases.

Open to suggestion, not sure a better way to handle this but at the moment this change is needed to address the issue in question.

This can be exposed to user when:
```
import pandas as pd

s = pd.Series(range(10))
mask = s > 5

s[mask] = [5, 4, 3, 2, 1]
```

minggli:
added test cases too from two issues linked. 🚀

pandas/core/internals/blocks.py
```diff
@@ -993,15 +992,16 @@ def putmask(self, mask, new, align=True, inplace=False, axis=0,
...
-
                 if not (mask.shape[-1] == len(new) or
                         mask[mask].shape[-1] == len(new) or
                         len(new) == 1):
-                    raise ValueError("cannot assign mismatch "
-                                     "length to masked array")
+                    # GH 19266 and GH 21977
+                    # ValueError triggers try except block in Block.replace
+                    # causing RecursionError
+                    raise Exception("cannot assign mismatch length "
```
jreback: can you make this a bit more informative from a user POV?

jreback (changes requested):

pandas/core/internals/blocks.py
minggli: sure. this change here is exposed to user so when one tries to assign an array of n-dimension to mask of m-dimension, when m != n, this error throws. 

```python
import pandas as pd

s = pd.Series(range(10))
mask = s > 5

>>> s[mask].shape
(4,)

s[mask] = [5, 4, 3, 2, 1]

  File "/Users/mingli/.virtualenvs/cust-value/lib/python3.5/site-packages/pandas/core/internals.py", line 1006, in putmask
    raise ValueError("cannot assign mismatch "
ValueError: cannot assign mismatch length to masked array
```
This is also used internally by `replace` and ValueError it raises is causing infinite loop as per linked issues. Whilst changing that to not ValueError will solve it, it also makes it less clear and potentially breaks users' codes if they expect to catch ValueError running above code snippet.

Do you think it's ok to use Exception here as it would solve the issues linked but users might have to change their code?

pandas/core/internals/blocks.py
jreback: do you know why we are catching ValueError at a higher level, maybe can get away with only catching TypeError and let ValueError buble up?

minggli: Good point and it appears below test (only this test) fails when we only catch TypeError:

=================================== FAILURES ===================================
_________________ TestDataFrameReplace.test_replace_datetimetz _________________
[gw0] darwin -- Python 3.5.5 /Users/mingli/.virtualenvs/pd-dev/bin/python3.5

self = <pandas.tests.frame.test_replace.TestDataFrameReplace object at 0x1112dc400>

    def test_replace_datetimetz(self):
    
        # GH 11326
        # behaving poorly when presented with a datetime64[ns, tz]
        df = DataFrame({'A': date_range('20130101', periods=3,
                                        tz='US/Eastern'),
                        'B': [0, np.nan, 2]})
        result = df.replace(np.nan, 1)
        expected = DataFrame({'A': date_range('20130101', periods=3,
                                              tz='US/Eastern'),
                              'B': Series([0, 1, 2], dtype='float64')})
    
        result = df.fillna(1)
    
        result = df.replace(0, np.nan)
        expected = DataFrame({'A': date_range('20130101', periods=3,
                                              tz='US/Eastern'),
                              'B': [np.nan, np.nan, 2]})
    
        result = df.replace(Timestamp('20130102', tz='US/Eastern'),
                            Timestamp('20130104', tz='US/Eastern'))
        expected = DataFrame({'A': [Timestamp('20130101', tz='US/Eastern'),
                                    Timestamp('20130104', tz='US/Eastern'),
                                    Timestamp('20130103', tz='US/Eastern')],
                              'B': [0, np.nan, 2]})
    
        result = df.copy()
        result.iloc[1, 0] = np.nan
        result = result.replace(
            {'A': pd.NaT}, Timestamp('20130104', tz='US/Eastern'))
    
        # coerce to object
        result = df.copy()
        result.iloc[1, 0] = np.nan
        result = result.replace(
>           {'A': pd.NaT}, Timestamp('20130104', tz='US/Pacific'))

pandas/tests/frame/test_replace.py:1060: 
pandas/core/frame.py:3895: in replace
    method=method)
pandas/core/generic.py:6016: in replace
    convert=convert)
pandas/core/internals/managers.py:560: in replace
    return self.apply('replace', **kwargs)
pandas/core/internals/managers.py:421: in apply
    applied = getattr(b, f)(**kwargs)
pandas/core/internals/blocks.py:798: in replace
    blocks = self.putmask(mask, value, inplace=inplace)
pandas/core/internals/blocks.py:1800: in putmask
    new_values, _, new, _ = self._try_coerce_args(new_values, new)

self = DatetimeTZBlock: slice(0, 1, 1), 1 x 3, dtype: datetime64[ns, US/Eastern]
values = array([[ 1357016400000000000, -9223372036854775808,  1357189200000000000]])
other = Timestamp('2013-01-04 00:00:00-0800', tz='US/Pacific')

    def _try_coerce_args(self, values, other):
        """
            localize and return i8 for the values
    
            Parameters
            ----------
            values : ndarray-like
            other : ndarray-like or scalar
    
            Returns
            -------
            base-type values, values mask, base-type other, other mask
            """
        values_mask = _block_shape(isna(values), ndim=self.ndim)
        # asi8 is a view, needs copy
        values = _block_shape(values.asi8, ndim=self.ndim)
        other_mask = False
    
        if isinstance(other, ABCSeries):
            other = self._holder(other)
            other_mask = isna(other)
    
        if isinstance(other, bool):
            raise TypeError
        elif (is_null_datelike_scalar(other) or
              (lib.is_scalar(other) and isna(other))):
            other = tslibs.iNaT
            other_mask = True
        elif isinstance(other, self._holder):
            if other.tz != self.values.tz:
                raise ValueError("incompatible or non tz-aware value")
            other_mask = _block_shape(isna(other), ndim=self.ndim)
            other = _block_shape(other.asi8, ndim=self.ndim)
        elif isinstance(other, (np.datetime64, datetime, date)):
            other = tslibs.Timestamp(other)
            tz = getattr(other, 'tz', None)
    
            # test we can have an equal time zone
            if tz is None or str(tz) != str(self.values.tz):
>               raise ValueError("incompatible or non tz-aware value")
E               ValueError: incompatible or non tz-aware value

pandas/core/internals/blocks.py:2898: ValueError

minggli: @jreback made some changes now we don't touch above error but instead make sure we don't enter infinite loop.

pandas/core/internals/blocks.py
minggli: this is reverted.

minggli:
@jreback @WillAyd I think PR is done, comments welcome.

WillAyd (changes requested):

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -667,7 +667,8 @@ Reshaping
 - Bug in :meth:`DataFrame.replace` raises RecursionError when converting OutOfBounds ``datetime64[ns, tz]`` (:issue:`20380`)
 - :func:`pandas.core.groupby.GroupBy.rank` now raises a ``ValueError`` when an invalid value is passed for argument ``na_option`` (:issue:`22124`)
 - Bug in :func:`get_dummies` with Unicode attributes in Python 2 (:issue:`22084`)
--
+- Bug in :meth:`DataFrame.replace` raises RecursionError when replace empty lists (:issue:`22083`)
```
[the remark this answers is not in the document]
minggli: thanks! changed.

pandas/tests/frame/test_replace.py
```diff
@@ -603,6 +603,20 @@ def test_replace_list(self):
 
         assert_frame_equal(res, expec)
 
+    def test_replace_with_empty_list(self):
+        # GH 21977
+        s = pd.Series([['a', 'b'], np.nan, [1]])
```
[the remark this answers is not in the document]
minggli: good point!

WillAyd (approved):
lgtm

pandas/core/internals/blocks.py
```diff
@@ -802,14 +801,19 @@ def replace(self, to_replace, value, inplace=False, filter=None,
...
-            block = self.astype(object)
-            return block.replace(to_replace=original_to_replace,
-                                 value=value,
-                                 inplace=inplace,
-                                 filter=filter,
-                                 regex=regex,
-                                 convert=convert)
+            # GH 22083, TypeError or ValueError occurred within error handling
+            # causes infinite loop. Cast and retry only if not objectblock.
+            if self.dtype == 'object':
```
jreback: can you use is_object_dtype

pandas/core/internals/blocks.py
```diff
@@ -802,14 +801,19 @@ def replace(self, to_replace, value, inplace=False, filter=None,
...
+                raise
+            else:
```
jreback: don't need this else here

jreback (changes requested):
minor comments, ping on green.

jreback (approved):

jreback:
thanks @minggli

PR #22089, opened 2018-07-28
CLN: Remove vbench_to_asv.py

mroeschke:
Since pandas has been using ASV for performance benchmarking, and artifacts of vbench are gone, there doesn't appear a need for this script anymore.


jreback:
thanks @mroeschke

Issue #8425, opened 2014-09-30
BUG: DataFrame from_dict constructor ignores Ordered dict when orient='index'


aimboden:
Hello, 
I have been experimenting with OrderedDicts lately, and found a bug with the DataFrame from_dict constructor. Here is a sample code.

``` python
import collections
import pandas as pd

firstrow={}
firstrow['foo'] = 'bar'
firstrow['baz'] = 'buzz'

row1 = pd.Series(firstrow)

secondrow={}
secondrow['foo'] = 'bar2'
secondrow['baz'] = 'buzz2'

row2 = pd.Series(secondrow)

roworder = collections.OrderedDict()

roworder['zShould be first'] = row1
roworder['Should be second'] = row2

# Ordering is respected when sorting on columns
df = pd.DataFrame.from_dict(data=roworder, orient='columns')

# But not when sorting on rows
incorrectdf = pd.DataFrame.from_dict(data=roworder, orient='index')
correctdf = df.transpose()
```
## INSTALLED VERSIONS

commit: None
python: 3.3.5.final.0
python-bits: 64
OS: Windows
OS-release: 7
machine: AMD64
processor: Intel64 Family 6 Model 26 Stepping 5, GenuineIntel
byteorder: little
LC_ALL: None
LANG: fr_CH

pandas: 0.14.1
nose: 1.3.4
Cython: 0.20.1
numpy: 1.9.0
scipy: 0.13.3
statsmodels: 0.5.0
IPython: 2.2.0
sphinx: 1.2.3
patsy: 0.3.0
scikits.timeseries: None
dateutil: 2.2
pytz: 2013.9
bottleneck: None
tables: 3.1.1
numexpr: 2.4
matplotlib: 1.4.0
openpyxl: None
xlrd: 0.9.3
xlwt: None
xlsxwriter: 0.5.7
lxml: None
bs4: None
html5lib: None
httplib2: None
apiclient: None
rpy2: None
sqlalchemy: 0.9.4
pymysql: None
psycopg2: None

hamedhsn:
@jreback 
I have done based on what you said  and in the last part how can I pass the flag to fast_unique_multiple because it calls fast_unique_multiple_list(*args, **kwargs) and when I look at the lib.pyx it always sort the list at the end(uniques.sort())
any idea?

alichaudry:
[...]

PR #22094, opened 2018-07-28
MAINT: refactor from_items() using from_dict()

### LINKED ISSUE #21850: Un-deprecate DataFrame.from_items()
Under #18262 a `FutureWarning` was added suggesting that existing code like this:

    pd.DataFrame.from_items(x)

Should be changed to this:

    import collections
    pd.DataFrame.from_dict(collections.OrderedDict(x))

The fact that `from_items()` appeared only 6 times (now 8 times) in a Stack Overflow search was used as partial justification for removing it.  But if you search on GitHub, [`pd.DataFrame.from_items()`](https://github.com/search?q=pd.DataFrame.from_items&type=Code) appears more than 15,000 times in Python--almost half as many as `from_records()`!

We should celebrate the fact that this function doesn't cause enough confusion to appear often on Stack Overflow.  But it does occur (a lot!) in real code, and deprecating it is a mistake.

If constructing a temporary `OrderedDict` around items is the best way to construct a DataFrame, Pandas should implement that as a short function called `DataFrame.from_items()`, rather than asking thousands of people to busy themselves to accommodate this unnecessary API change.

I recommend removing the FutureWarning, and retaining this widely-used, longstanding function.

For reference, the `FutureWarning` starts in 0.23 and looks like this:

> FutureWarning: from_items is deprecated. Please use DataFrame.from_dict(dict(items), ...) instead. DataFrame.from_dict(OrderedDict(items)) may be used to preserve the key order.

jzwinck:
This removes the deprecation warnings introduced in #18262,
by reimplementing `DataFrame.from_items()` in the recommended
way using `DataFrame.from_dict()` and `collections.OrderedDict`.

This eliminates the maintenance burden of separate code for
`from_items()`, while allowing existing uses to keep working.

A small cleanup can be done once #8425 is fixed.

- [x] closes #21850
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -391,6 +391,7 @@ Removal of prior version deprecations/changes
 
 - The ``LongPanel`` and ``WidePanel`` classes have been removed (:issue:`10892`)
 - Several private functions were removed from the (non-public) module ``pandas.core.common`` (:issue:`22001`)
+- :meth: `~pandas.DataFrame.from_items` has been refactored and the deprecation warning from v0.23.0 has been removed (i.e. it is no longer deprecated)
 -
 -
 
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -54,7 +54,6 @@ from pandas.core.dtypes.common import (
     ensure_int64,
     ensure_platform_int,
     is_list_like,
-    is_nested_list_like,
     is_iterator,
     is_sequence,
     is_named_tuple)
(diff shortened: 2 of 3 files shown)


jzwinck:
CI is reporting an error in `TestHDFStore.test_calendar_roundtrip_issue` but it is not caused by this PR (other Pandas PRs report the same error).

jreback:
did u not see my comment on the issue?

pandas already has way too many apis

just because X number of people use something is only 1 factor in a deprecation

jzwinck:
@jreback I did see your comment, which read:

> there is a burden to maintain apis and code why should this be any different from other ‘highly used code’

This PR makes `from_items()` a simple wrapper around `from_dict()`, which surely imposes less burden than repeatedly answering questions from people whose code using `from_items()` now causes warnings (and soon, failures).  You might be under the impression that everyone can just replace `from_items(x, ...)` with `from_dict(OrderedDict(x), ...)`, but as this PR shows it is not quite so simple.  In fact there are three different "modes", one of which actually produces the wrong result if used without my wrapper (due to #8425).

I realize that you strongly prefer to eliminate code.  This PR reduces code relative to 0.22 and even more so when compared with master.  It adds a small amount of simple code relative to your preferred end state, but the benefit is that it allows thousands of existing applications and notebooks to continue working.

@jorisvandenbossche @gfyoung @chris-b1 @reidy-p @wesm I do not agree with the closure of this PR without a substantive discussion.  I do not know whether Pandas is unilaterally ruled by one person, but I would appreciate your consideration of this PR to avoid breaking a large number of programs.  I believe Pandas has a long future ahead, but this API brinksmanship erodes confidence.

jorisvandenbossche:
Let's keep the discussion on the issue https://github.com/pandas-dev/pandas/issues/21850 (but @jzwinck your PR is certainly useful for illustration purposes for the discussion as well), I'll reopen the issue.

PR #22110, opened 2018-07-29
BUG: _cython_table bug fix

topper-123:
#21224 introduced a bug, where the ``conftest._cython_table`` isn't sorted by the function's name, but by the functions class name (i.e. "function"). This can make the sorting indeterministic and make the build fail on Python 2.7 and 3.5 at times.

This PR makes the sort order follow the functions names again and therefore be deterministic.


jreback:
lgtm ping on green

topper-123:
All green.

PR #22114, opened 2018-07-29
BUG: TimedeltaIndex.intersection

### LINKED ISSUE #17391: BUG: TimedeltaIndex.intersection fails for decreasing TimedeltaIndex
Setup
```python
In [1]: import pandas as pd
   ...: idx1 = pd.to_timedelta(range(3), unit='s')
   ...: idx2 = pd.to_timedelta(range(2, -1, -1), unit='s')
   ...:

In [2]: idx1
Out[2]: TimedeltaIndex(['00:00:00', '00:00:01', '00:00:02'], dtype='timedelta64[ns]', freq=None)

In [3]: idx2
Out[3]: TimedeltaIndex(['00:00:02', '00:00:01', '00:00:00'], dtype='timedelta64[ns]', freq=None)
```

Intersection returns empty between `idx1` and `idx2`:
```python
In [4]: idx1.intersection(idx2)
Out[4]: TimedeltaIndex([], dtype='timedelta64[ns]', freq=None)

In [5]: idx2.intersection(idx1)
Out[5]: TimedeltaIndex([], dtype='timedelta64[ns]', freq=None)
```

Even an intersection of `idx2` with itself returns empty:
```python
In [6]: idx2.intersection(idx2)
Out[6]: TimedeltaIndex([], dtype='timedelta64[ns]', freq=None)
```

#### Problem description
The intersections are returning empty even though all elements are shared.

#### Expected Output
I'd expect all of the intersections to return a `TimedeltaIndex` containing all three elements.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.2.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.20.3
pytest: 3.1.2
pip: 9.0.1
setuptools: 27.2.0
Cython: 0.26
numpy: 1.13.1
scipy: 0.19.1
xarray: None
IPython: 6.1.0
sphinx: 1.5.6
patsy: 0.4.1
dateutil: 2.6.0
pytz: 2017.2
blosc: None
bottleneck: 1.2.1
tables: 3.2.2
numexpr: 2.6.2
feather: None
matplotlib: 2.0.2
openpyxl: 2.4.7
xlrd: 1.0.0
xlwt: 1.2.0
xlsxwriter: 0.9.6
lxml: 3.7.3
bs4: 4.6.0
html5lib: 0.999
sqlalchemy: 1.1.9
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
pandas_gbq: None
pandas_datareader: None

</details>

kirkhansen:
- [ ] closes #17391
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -1585,6 +1585,12 @@ class Index(IndexOpsMixin, PandasObject):
         return self.is_monotonic_increasing
 
     @property
+    def _is_strictly_monotonic(self):
+        """ Checks if the index is sorted """
+        return (self._is_strictly_monotonic_increasing or
+                self._is_strictly_monotonic_decreasing)
+
+    @property
     def is_monotonic_increasing(self):
         """
         Return if the index is monotonic increasing (only equal or
--- a/pandas/core/indexes/datetimelike.py
+++ b/pandas/core/indexes/datetimelike.py
@@ -576,38 +576,49 @@ class DatetimeIndexOpsMixin(DatetimeLikeArrayMixin):
         result.name = self.name
         return result
 
-    def _intersect_ascending(self, other):
-        # to make our life easier, "sort" the two ranges
-        if self[0] <= other[0]:
+    def _fast_intersection(self, other):
+        """
+        Speedy intersection that works only if certain assumptions are met.
+        See intersection for details.
+        Parameters
+        ----------
+        other
+
+        Returns
+        -------
+
+        """
+        # Coerce into the same order
+        ascending = self.is_monotonic_increasing
+        if ascending != other.is_monotonic_increasing:
+            other = other.sort_values(ascending=ascending)
+        # Lots of 'if ascending' calls here to setup mirrored function calls
+        first_comparison = '__le__' if ascending else '__ge__'
+        second_comparison = '__lt__' if ascending else '__gt__'
+        if getattr(self[0], first_comparison)(other[0]):
             left, right = self, other
         else:
             left, right = other, self
 
-        end = min(left[-1], right[-1])
-        start = right[0]
-
-        if end < start:
-            return []
-        return left.values[slice(*left.slice_locs(start, end))]
-
-    def _intersect_descending(self, other):
-        # this is essentially a flip of _intersect_ascending
-        if self[0] >= other[0]:
-            left, right = self, other
+        if ascending:
+            start = right[0]
+            end = min(left[-1], right[-1])
         else:
-            left, right = other, self
-
-        start = min(left[0], right[0])
-        end = right[-1]
-
-        if end > start:
-            return Index()
-        return left.values[slice(*left.slice_locs(start, end))]
+            start = min(left[0], right[0])
+            end = right[-1]
+        if getattr(end, second_comparison, start):
+            return left.values[slice(*left.slice_locs(start, end))]
+        return []
 
     def intersection(self, other):
         """
         Specialized intersection for DateTimeIndexOpsMixin objects.
         May be much faster than Index.intersection.
+
+        Fast intersection will occur if
+        1. Both are in a sorted order
+        2. Both indexes have a `freq` , and it's the same `freq`
+        3. Both are monotonic
 
         Parameters
         ----------
@@ -618,6 +629,9 @@ class DatetimeIndexOpsMixin(DatetimeLikeArrayMixin):
         Index
             A shallow copied intersection between the two things passed in
         """
+        # Run a few checks, and perform a regular intersection
+        # if the conditions aren't just right for fast intersection
+        # Perform a regular Index.intersection
         self._assert_can_do_setop(other)
 
         if self.equals(other):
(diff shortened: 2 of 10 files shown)


pandas/core/indexes/base.py
```diff
@@ -1412,6 +1412,12 @@ def is_monotonic(self):
         """ alias for is_monotonic_increasing (deprecated) """
         return self.is_monotonic_increasing
 
+    @property
```
jreback: are we actually using this? (it seems you in-lined it below), I actually prefer that to making a new method

pandas/core/indexes/datetimelike.py
```diff
@@ -679,6 +682,94 @@ def astype(self, dtype, copy=True):
...
+        if self[0] >= other[0]:
+            left, right = self, other
+        else:
+            left, right = other, self
+
+        start = min(left[0], right[0])
+        end = right[-1]
+
+        if end > start:
+            return Index()
```
jreback: why is this not []?

pandas/core/indexes/datetimelike.py
```diff
@@ -679,6 +682,94 @@ def astype(self, dtype, copy=True):
...
+        if self[0] <= other[0]:
+            left, right = self, other
+        else:
+            left, right = other, self
+
+        end = min(left[-1], right[-1])
+        start = right[0]
+
+        if end < start:
+            return []
```
jreback: worth to consolidate these into 1 routine? as you can call with the ``ascending`` arg

pandas/core/indexes/datetimelike.py
```diff
@@ -679,6 +682,94 @@ def astype(self, dtype, copy=True):
...
+            result = Index.intersection(self, other)
+            result = self._shallow_copy(result._values, name=result.name,
+                                        tz=getattr(self, 'tz', None),
+                                        freq=None
+                                        )
+            if result.freq is None:
+                result.offset = frequencies.to_offset(result.inferred_freq)
+            return result
+
+        # handle intersecting things like this
```
jreback: can you add a tiny bit more of expl

pandas/tseries/offsets.py
```diff
@@ -112,6 +112,41 @@ def wrapper(self, other):
     return wrapper
 
 
+def apply_index_wraps(func):
```
jreback: what is going on here?

jreback (changes requested):
pls add a whatsnew note in bug fixes / timedeltas

pandas/tseries/offsets.py
kirkhansen: Been awhile. I don't remember writing this code. I'll guess an artifact of my rebase.

pandas/core/indexes/base.py
kirkhansen: The names are similar, but I'm calling this function twice in the intersection function to help me deal with the boolean logic overloading. I can remove this, and put this inline if you really want it that way.

kirkhansen:
Having come back at this, I'm struggling with how to deal with all the edge cases that need to be accounted for before trying to apply this quick method of intersecting, and this is making me question whether or not to even include it. 
The quick method cannot allow any 'gaps' between values, unless both indexes have the same gaps, and having to check for that before intersecting seems a lot harder than just using `Index.intersection`, unless that method is worse than something we can write to check if if values are contiguous based on the frequency passed in or something.

I'm having a hard time describing it clearly, but hopefully this is making sense. If not, I can commit my broken code so you all can see what I'm talking about.

jorisvandenbossche:
A possibility to at least solve the bug would be to fall back to the `Index.intersection` implementation for the cases where monotonicity does not match?

Basically similar like `DatetimeIndex.intersection` also falls back to it when eg freq does not match:

https://github.com/pandas-dev/pandas/blob/59cfd8cf826c79f8633c3a47aab53cafa3927bac/pandas/core/indexes/datetimes.py#L1006-L1015

Will not be the most performant option, but at least could fix the bug

kirkhansen:
@jorisvandenbossche I thought I had been doing that, but it seems my check was backwards for that case :/. Correcting that appears to solve the issues. Thanks!

jreback:
if you can merge master

mroeschke:
You'll need to run `isort` on this file:

```
2018-12-12T21:36:13.4930067Z ERROR: /home/vsts/work/1/s/pandas/core/indexes/datetimelike.py Imports are incorrectly sorted.
```

pandas/tseries/offsets.py
```diff
@@ -103,6 +103,24 @@ def wrapper(self, other):
...
+    second: Index
+
+    Returns
+    -------
+    bool
+    """
+    first = getattr(first, 'freq', None)
+    second = getattr(second, 'freq', None)
+    are_offsets_equal = True
+    if first is None or second is None or first != second:
```
mroeschke: Just return this statement; no need for the intermediate variable

pandas/tests/io/test_pytables.py
```diff
@@ -4116,6 +4116,7 @@ def test_append_to_multiple_dropna(self):
         df2 = tm.makeTimeDataFrame().rename(columns=lambda x: "%s_2" % x)
         df1.iloc[1, df1.columns.get_indexer(['A', 'B'])] = np.nan
         df = concat([df1, df2], axis=1)
+        print(df)
```
mroeschke: This `print` should be removed.

pandas/core/indexes/datetimelike.py
```diff
@@ -572,6 +575,92 @@ def _time_shift(self, periods, freq=None):
         result.name = self.name
         return result
 
+    def _fast_intersection(self, other):
+        """
+        Speedy intersection that works only if certain assumptions are met.
+        See intersection for details.
+        Parameters
```
mroeschke: Could you give a bit more color to `Parameters` and `Returns

WillAyd:
Closing as stale ping if you'd like to continue

PR #22074, opened 2018-07-27
[Bug] Fix various DatetimeIndex comparison bugs

jbrockmendel:
Will have to see which issues this closes.  This is also a precursor to upcoming PR that dispatches DataFrame ops to Series (which in turn dispatches to Index).

Fixes #7830 (the OP's example is already fixed, this fixes examples in the comments to that issue)
Fixes DTI half of #19804

- [x] closes #7830
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

PR #22068, opened 2018-07-26
De-duplicate dispatch code, remove unreachable branches

jbrockmendel:
There are a couple of `DataFrame` methods that operate column-wise, dispatching to `Series` implementations.  This de-duplicates that code by implementing `ops.dispatch_to_series`.  Importantly, this function is going to be used a few more times as we move towards getting rid of `BlockManager.eval` and `Block.eval`, so putting it in place now makes for a more focused diff later.

Also removes a couple of no-longer-reachable cases in comparison ops.

- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

PR #22118, opened 2018-07-29
Centralize m8[ns] Arithmetic Tests

jbrockmendel:
There are a bunch still left in the TDI tests, but the diff is already pretty big as it is.

A _ton_ of `DataFrame` tests are xfailed here.  Most of these will be fixed by the follow-up to (#22068 + #22074).

xfails are made strict so that we get alerted when they are fixed.  Several existing xfails are removed since they have been XPASSing for a while.

@jreback I know you're going to want to make fixtures for a bunch of things.  I agree, but think that will be 19.6x easier after fixing some of the xfails.  Also this is step 2 of a many-step process and in general I'd prefer to do cleanup after getting more stuff in place.


jreback (changes requested):
this looks ok, and yes was going to see if we can use fixtures, but can do that later. however, I think test_arithmetic is going to get big fast, maybe make a directory structure on this, e.g.

pandas/tests/arithmetic/test_timedelta...... etc.?

jbrockmendel:
> maybe make a directory structure on this

I've been thinking about this too.  So far this file is all timedelta64[ns], so that's easy at least.  I'll split it in an upcoming pass.

jreback (approved):

PR #22127, opened 2018-07-30
Add Japanese version of Pandas_Cheat_Sheet

KangYoosam:
## Background
We are sponsoring `PyCon` on September in Japan. And we are planning to influence Pandas in Japan. To do so, we are planning to distribute `Pandas Cheet Sheet` with plastic file at `Pycon`.


jreback:
thanks @KangYoosam

PR #22132, opened 2018-07-30
Add timetz attribute to DatetimeIndex

### LINKED ISSUE #21358: ENH: Add timetz attribute to DatetimeIndex and Timestamp

jquinon:
- [x] closes #21358 
- [x] tests added / passed
- [x] whatsnew entry

I only added timetz to DatetimeIndex because there was already a timetz method in Timestamp that did essentially the same thing. If it is still worth adding to Timestamp though, I can certainly do that.


doc/source/whatsnew/v0.24.0.txt
```diff
@@ -180,6 +180,7 @@ Other Enhancements
 - :func:`~DataFrame.to_csv` and :func:`~DataFrame.to_json` now support ``compression='infer'`` to infer compression based on filename (:issue:`15008`)
 - :func:`to_timedelta` now supports iso-formated timedelta strings (:issue:`21877`)
 - :class:`Series` and :class:`DataFrame` now support :class:`Iterable` in constructor (:issue:`2193`)
+- :class:`DatetimeIndex` gained :attr:`~DatetimeIndex.timetz` attribute. Returns time with timezone information (:issue:`21358`)
```
mroeschke: 1) I don't think you need the `~` before `~DatetimeIndex.timetz`
2) Can you clarify that this returns _local_ time with timezone information

pandas/tests/indexes/datetimes/test_timezones.py
```diff
@@ -731,6 +731,18 @@ def test_time_accessor(self, dtype):
 
         tm.assert_numpy_array_equal(result, expected)
 
+    @pytest.mark.parametrize('tz', [
+        dateutil.tz.gettz('US/Eastern'), dateutil.tz.gettz('US/Pacific'),
+        dateutil.tz.gettz('UTC')])
+    def test_timetz_accessor(self, tz):
```
mroeschke: Instead could you use the `tz_naive_fixture`?

https://github.com/pandas-dev/pandas/blob/26b3e7de56fdebaf1842b01d747a1a30a126c54a/pandas/conftest.py#L290

mroeschke:
Looks good overall!

1) Could you add the `timetz` attribute to this list: http://pandas.pydata.org/pandas-docs/stable/timeseries.html?#time-date-components

2) Could you check if we have some tests for `Timestamp.timetz` and add one if we don't have any?

jquinon:
I didn't add a timetz attribute to Timestamp yet because it inherits a timetz method directly from standard library datetime. So should we add an attribute on top of that?

mroeschke:
I don't think you have to add a `timetz` property to `Timestamp`, but please go ahead and add some `timetz` tests for `Timestamp` if we don't have any so we can make sure the behavior is consistent with `datetime`.

jquinon:
Ok, gotcha

doc/source/api.rst
```diff
@@ -1739,6 +1739,7 @@ Time/Date Components
    DatetimeIndex.nanosecond
    DatetimeIndex.date
    DatetimeIndex.time
```
jreback: can add in the api section for Timestamp though

jreback (changes requested):
can we test for timetz on Timestamp (only in PY3 I think)?

doc/source/api.rst
jquinon: The timetz method was already there for Timestamp, and I added the property for DatetimeIndex

jreback:
I find an attribute for this is very confusing, as compared to the function call similar to  ``.time()``. would a function call be better here? (of course this would be a break with ``datetime.datetime``)

mroeschke:
For the PR, I think we should leave it as a property for now since we break this property vs. method for other `DatetimeIndex` attributes (`date`, `time`, `weekday`), and then in another PR consider converting these to method calls (maybe even before v0.24).

(Although I don't even know why these are method calls for `datetime` in the first place...)

jquinon:
@mroeschke how does it look?

mroeschke (approved):

mroeschke:
Any other comments @jreback?

jreback (approved):

jreback:
thanks @jquinon nice patch!

PR #22139, opened 2018-07-31
[TST] make xfails strict

jbrockmendel:
This leaves untouched xfails with reasons like "flaky on 3.7" and a few where the test itself is just a `pass`.

In the process a handful of previously xfailed cases are no longer xfailed.

### DIFF
--- a/pandas/tests/extension/json/test_json.py
+++ b/pandas/tests/extension/json/test_json.py
@@ -142,6 +142,7 @@ class TestInterface(BaseJSON, base.BaseInterfaceTests):
 
 class TestConstructors(BaseJSON, base.BaseConstructorsTests):
 
+    # TODO: Should this be pytest.mark.skip?
     @pytest.mark.xfail(reason="not implemented constructor from dtype")
     def test_from_dtype(self, data):
         # construct from our dtype & string dtype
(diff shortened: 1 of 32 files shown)


pandas/tests/extension/json/test_json.py
```diff
@@ -142,6 +142,7 @@ def test_custom_asserts(self):
 
 class TestConstructors(BaseJSON, base.BaseConstructorsTests):
 
+    # TODO: Should this be pytest.mark.skip?
```
jreback: yeah u can change to skip (and similar ones)

pandas/tests/indexes/multi/test_missing.py
```diff
@@ -83,7 +83,7 @@ def test_nulls(idx):
         idx.isna()
 
 
-@pytest.mark.xfail
+@pytest.mark.xfail(strict=True)
```
jreback: can u add a reason

pandas/tests/plotting/test_misc.py
```diff
@@ -212,7 +212,6 @@ def test_parallel_coordinates(self, iris):
         with tm.assert_produces_warning(FutureWarning):
             parallel_coordinates(df, 'Name', colors=colors)
 
-    @pytest.mark.xfail(reason="unreliable test")
     def test_parallel_coordinates_with_sorted_labels(self):
```
jreback: hah!

jreback (changes requested):

pandas/tests/indexes/multi/test_missing.py
jbrockmendel: Sure, will try to sort that out.  Also waiting to see if anything fails on the CI that passed locally.

jreback (approved):

jreback:
thanks, ideally also update the  contributing.rst docs where we use xfail as an example to include the strict param.

PR #22141, opened 2018-07-31
DOC: DataFrame.merge behaviour for suffix=(False, False)

### LINKED ISSUE #22045: [documentation] merge(suffixes=(False, False)) should cause an error if a suffix would be required to complete the merge but it is undocumented
#### Problem description

This was going to be a request for an enhancement, but after investigation it just became a reminder that pandas has an undocumented feature. 

About once every few months I get into a tricky debug situation where a column has ceased to exist because it was suffixed with _x or _y, and while I understand use cases where suffixing is good, it is in some cases bad and unintended, and can be tricky because the _x and _y columns may not matter until much later.  I wanted to be able to specify in merge that it should fail if it is going to suffix. 

Turns out you can, but it is undocumented.  Tracing code I eventually arrived at https://github.com/pandas-dev/pandas/blob/v0.23.3/pandas/core/internals.py#L5242

A quick check seems to indicate that merge(..., suffixes=(False,False)) causes the line to execute and doesn't bite if it shouldn't.  Which is nice, but I would encourage you to document it at:
https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.merge.html

~~~python
import pandas as pd
left = pd.DataFrame({'key': ['K0', 'K1', 'K2', 'K3'],
                    'A': ['A0', 'A1', 'A2', 'A3'],
                     'B': ['B0', 'B1', 'B2', 'B3']})


right = pd.DataFrame({'key': ['K0', 'K1', 'K2', 'K3'],
            'C': ['C0', 'C1', 'C2', 'C3'],
                  'B': ['D0', 'D1', 'D2', 'D3']})

right_2 = pd.DataFrame({'key': ['K0', 'K1', 'K2', 'K3'],
            'C': ['C0', 'C1', 'C2', 'C3'],
                  'D': ['D0', 'D1', 'D2', 'D3']})


result = pd.merge(left, right, on='key')
result

#should error and does
pd.merge(left, right, on='key', suffixes=(False, False))


#should not error and doesnt
pd.merge(left, right_2, on='key', suffixes=(False, False))
~~~

elmq0022:
- [x] closes #22045
- N/A tests added / passed
- N/A whatsnew entry

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -138,12 +138,11 @@ _numeric_only_doc = """numeric_only : boolean, default None
 """
 
 _merge_doc = """
-Merge DataFrame or named Series objects by performing a database-style join
-operation by columns or indexes.
+Merge DataFrame or named Series objects with a database-style join.
 
-If joining columns on columns, the DataFrame indexes *will be
-ignored*. Otherwise if joining indexes on indexes or indexes on a column or
-columns, the index will be passed on.
+The join is done on columns or indexes. If joining columns on
+columns, the DataFrame indexes *will be ignored*. Otherwise if joining indexes
+on indexes or indexes on a column or columns, the index will be passed on.
 
 Parameters
 ----------%s
@@ -153,13 +152,13 @@ how : {'left', 'right', 'outer', 'inner'}, default 'inner'
     Type of merge to be performed.
 
     * left: use only keys from left frame, similar to a SQL left outer join;
-      preserve key order
+      preserve key order.
     * right: use only keys from right frame, similar to a SQL right outer join;
-      preserve key order
+      preserve key order.
     * outer: use union of keys from both frames, similar to a SQL full outer
-      join; sort keys lexicographically
+      join; sort keys lexicographically.
     * inner: use intersection of keys from both frames, similar to a SQL inner
-      join; preserve the order of the left keys
+      join; preserve the order of the left keys.
 on : label or list
     Column or index level names to join on. These must be found in both
     DataFrames. If `on` is None and not merging on indexes then this defaults
@@ -172,23 +171,23 @@ right_on : label or list, or array-like
     Column or index level names to join on in the right DataFrame. Can also
     be an array or list of arrays of the length of the right DataFrame.
     These arrays are treated as if they are columns.
-left_index : boolean, default False
+left_index : bool, default False
     Use the index from the left DataFrame as the join key(s). If it is a
     MultiIndex, the number of keys in the other DataFrame (either the index
     or a number of columns) must match the number of levels.
-right_index : boolean, default False
+right_index : bool, default False
     Use the index from the right DataFrame as the join key. Same caveats as
     left_index.
-sort : boolean, default False
+sort : bool, default False
     Sort the join keys lexicographically in the result DataFrame. If False,
     the order of the join keys depends on the join type (how keyword).
-suffixes : 2-length sequence (tuple, list, ...)
+suffixes : tuple of (str, str), default ('_x', '_y')
     Suffix to apply to overlapping column names in the left and right
-    side, respectively. If (False, False), overlapping
-    column names raise an error.
-copy : boolean, default True
+    side, respectively. To raise an exception on overlapping columns use
+    (False, False).
+copy : bool, default True
     If False, avoid copy if possible.
-indicator : boolean or string, default False
+indicator : bool or str, default False
     If True, adds a column to output DataFrame called "_merge" with
     information on the source of each row.
     If string, column with information on source of each row will be added to
@@ -198,7 +197,7 @@ indicator : boolean or string, default False
     "right_only" for observations whose merge key only appears in 'right'
     DataFrame, and "both" if the observation's merge key is found in both.
 
-validate : string, default None
+validate : str, default None
     If specified, checks if merge is of specified type.
 
     * "one_to_one" or "1:1": check if merge keys are unique in both
@@ -213,7 +212,8 @@ validate : string, default None
 
 Returns
 -------
-DataFrame
+df: DataFrame
+    A DataFrame of the two merged objects.
 
 Notes
 -----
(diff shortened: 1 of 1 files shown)


pandas/core/frame.py
```diff
@@ -183,7 +183,8 @@
     the order of the join keys depends on the join type (how keyword).
 suffixes : 2-length sequence (tuple, list, ...)
     Suffix to apply to overlapping column names in the left and right
-    side, respectively.
+    side, respectively. If (False, False), overlapping
+    column names raise an error.
```
jreback: can you add an example in the doc-string to make this even more clear? (we don't even have a suffix example at all......)

jreback (changes requested):

pandas/core/frame.py
elmq0022: @jreback, the doc string example does illustrate the behavior for the default suffix arguments.

Also, there is a more detailed suffix example here: https://pandas.pydata.org/pandas-docs/stable/merging.html#overlapping-value-columns

I suggest adding a specific note or example to the link above for `suffix=(False, False)` instead of the doc string.  

I feel a more experienced user would likely know what to do just from the parameter argument description while a more novice user will probably need to look at the longer documentation with copy/paste examples that work with ipython or jupyter.    

I also suggest the merge, join, and concat methods all have a reference to the more complete documentation in the *see also* section.  I've found this documentation very helpful in the past and still refer to if from time to time.  

Thoughts?

elmq0022: @jreback Also, I'm happy to just update the doc string.  Just wanted to offer a different perspective.  Let me know what you prefer.  Thanks!

jreback: @elmq0022 yeah an updated doc-string would be great

elmq0022:
@jreback, I updated the docstring with a couple suffix examples.

elmq0022:
@jreback any additional feedback?  Thanks!

pandas/core/frame.py
```diff
@@ -254,6 +255,23 @@
 3  foo        5  foo        8
 4  bar        2  bar        6
 5  baz        3  baz        7
+
+>>> A.merge(B, left_on='lkey', right_on='rkey', how='outer',
```
jreback: can you add a comment for both of these example of what they are showing

jreback (changes requested):

elmq0022:
Sure.

pandas/core/frame.py
datapythonista: For the type, I'd have `suffixes : tuple of (str, str), default ('_x', '_y')`. Small detail, but I'd rephrase you addition to sound like `To raise an exception on overlapping columns, use (False, False)`. Being able to use (False, False) to raise looks like a feature to me, and I think it's better to present it this way. May be we could also have it in the `Notes` section.

But it's just an idea, I'm happy with the PR the way it is. It's a good addition.

elmq0022:
@jreback and @datapythonista, I added comments to the examples.

elmq0022:
@datapythonista, I made the updates.  This should be good to go now.  Thanks!

pandas/core/frame.py
```diff
@@ -182,9 +182,10 @@
 sort : boolean, default False
     Sort the join keys lexicographically in the result DataFrame. If False,
     the order of the join keys depends on the join type (how keyword).
-suffixes : 2-length sequence (tuple, list, ...)
+suffixes : tuple of (str, str), default ('_x', '_y')
     Suffix to apply to overlapping column names in the left and right
-    side, respectively.
+    side, respectively. To raise an exption on overlapping columns use
```
datapythonista: typo: s/exption/exception

pandas/core/frame.py
```diff
@@ -254,6 +255,29 @@
 3  foo        5  foo        8
 4  bar        2  bar        6
 5  baz        3  baz        7
+
+Merge DataFrames A and B. Specify the left and right suffix
+to append to the name of any overlapping columns.
+
+>>> A.merge(B, left_on='lkey', right_on='rkey', how='outer',
```
datapythonista: sorry, I didn't see before that in the original docstring the DataFrame names were `A` and `B`. Do you mind changing them to `df1` and `df2`. That's the standard we use, and `A` and `B` in Python should be used for classes, not instances.

Also, we can take care of that in a different PR if you prefer, but it'd be great to have a more real world example (it makes it easier to understand what's going on), and in this case I find the `how='outer'` misleading, as it doesn't add value.

datapythonista (changes requested):
Happy to merge after fixing the typo and the variable names.

But if you want to work on making the docstring follow best practices and our standards, you can also:
- In parameters replace string by str, boolean by bool and integer by int
- In the returns add a description (indented in the next line after the type.
- Make the short summary (first line) shorter to fit in a single line
- run `scripts/validate_docstrings pandas.DataFrame.merge` to see if there is any other issue

Thanks!

pandas/core/frame.py
elmq0022: Yes, I can make those updates.

pandas/core/frame.py
```diff
@@ -254,6 +255,29 @@
...
+...         suffixes=('_left', '_right'))
+  lkey  value_left rkey  value_right
+0  foo           1  foo            5
+1  foo           1  foo            8
+2  foo           5  foo            5
+3  foo           5  foo            8
+4  bar           2  bar            6
+5  baz           3  baz            7
+
```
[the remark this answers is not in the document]
elmq0022: Yes, I was able to verify this using pytest --doctest-modules.

pandas/core/frame.py
elmq0022: Also thanks for the location of the validation scripts and such.  This is very helpful for my first documentation pull request.

elmq0022:
@datapythonista I made the updates.

datapythonista (changes requested):
Just some comments on minor things. For the rest looks good to me.

elmq0022:
@jreback, @datapythonista two items to address:

First, the following Travis job is failing.  

43166.3
Python: 3.5
JOB="2.7, locale, slow, old NumPy" ENV_FILE="ci/travis-27-locale.yaml" LOCALE_OVERRIDE="zh_CN.UTF-8" SLOW=true
3 min 51 sec

The error message is:
ImportError: libgfortran.so.1: cannot open shared object file: No such file or directory.  

I don't think this error is on my end, but let me know if I have to do something differently.  I did just rebase and rebuild all the C extensions.  

Second, I think I get what @jreback was referring to prior.  I am not sure if the pandas.DataFram.merge doctests are necessarily run during CI.  In doctests.sh on line 23 there is no *merge* argument mentioned. 

```bash
pytest --doctest-modules -v pandas/core/frame.py \
        -k"-assign -axes -combine -isin -itertuples -join -nlargest -nsmallest -nunique -pivot_table -quantile -query -reindex -reindex_axis -replace -round -set_index -stack -to_dict -to_stata -transform"
```
Should we add it here between -join and -nlargest?  Not sure where else this would go or what else would kick off the tests.

datapythonista:
@elmq0022 the CI error seems indeed unrelated to your changes. About the doctests, the `-` in `-join -nlargest ...` means that the tests are being skipped. So, that list is the list of tests that are failing now. If `-merge` is not there means that the examples are already correct, so as far as they continue to be correct, nothing needs to be done (if you have a example that breaks or with output different than what presented, the CI should report it).

datapythonista (approved):
lgtm. Later on, we can think on providing more examples about the behavior described in the extended summary, but I'd address that in a separated PR. Thanks @elmq0022

elmq0022:
@datapythonista @jreback, thanks for all your help.  Will need to rebase and resubmit to pass the CI?

datapythonista:
Yes, if you can merge master into your branch and push, that should trigger the CI again.

elmq0022:
@jreback, and @datapythonista  all green.

elmq0022:
@jreback friendly reminder. Thanks!

datapythonista:
Thanks for the work on this @elmq0022

elmq0022:
@datapythonista no problem.  Happy to help.  Thanks for feedback and merging the request.

Issue #22092, opened 2018-07-28
Bool Ops: One bug hides another


jbrockmendel:
Boolean ops (&, |, ^) are less well-tested than most of the others.  Going through some of these, some of them raise for weird reasons:

```
ser = pd.Series([True, False, True])
idx = pd.Index([False, True, True])
```

Starting with `ser & idx`, we'd expect to get `Series([False, False, True])`.  Instead we get a `ValueError` because we accidentally go down the path treating Index as a scalar:
```
>>> ser & idx
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pandas/core/ops.py", line 1481, in wrapper
    res_values = na_op(self.values, other)
  File "pandas/core/ops.py", line 1439, in na_op
    if not isna(y):
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

OK, that's fixed easily enough.  What about the reversed operation `idx & ser`?  That means something entirely different, since `Index.__and__` is an alias for `Index.intersection`, so we'd expect either a `Series` or `Index` containing both `False` and `True`.  Instead we get a `ValueError`:

```
>>> idx & ser
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "pandas/core/indexes/base.py", line 2658, in __and__
    return self.intersection(other)
  File "pandas/core/indexes/base.py", line 2823, in intersection
    if self.is_monotonic and other.is_monotonic:
  File "pandas/core/indexes/base.py", line 1407, in is_monotonic
    return self.is_monotonic_increasing
  File "pandas/core/indexes/base.py", line 1424, in is_monotonic_increasing
    return self._engine.is_monotonic_increasing
  File "pandas/_libs/index.pyx", line 214, in pandas._libs.index.IndexEngine.is_monotonic_increasing.__get__
    self._do_monotonic_check()
  File "pandas/_libs/index.pyx", line 228, in pandas._libs.index.IndexEngine._do_monotonic_check
    values = self._get_index_values()
  File "pandas/_libs/index.pyx", line 244, in pandas._libs.index.IndexEngine._get_index_values
    return self.vgetter()
  File "pandas/core/indexes/base.py", line 1847, in <lambda>
    return self._engine_type(lambda: self._ndarray_values, len(self))
  File "pandas/core/base.py", line 752, in _ndarray_values
    if is_extension_array_dtype(self):
[...]

Issue #22101, opened 2018-07-28
COMPAT: handle numpy 1.15 deprecation warnings


jreback:
I think this is a new warning from numpy in 1.15: https://travis-ci.org/pandas-dev/pandas/jobs/409274058

```
/home/travis/build/pandas-dev/pandas/pandas/core/dtypes/common.py:2026: DeprecationWarning: elementwise == comparison failed; this will raise an error in the future.
    if dtype in [object, np.object_, 'object', 'O']:
```

jreback:
cc @charris

charris:
The `DeprecationWarning` has been around since  NumPy 1.10, what is new is https://github.com/numpy/numpy/pull/10337. Note the Pandas comments in the referenced issue https://github.com/numpy/numpy/issues/5188. I think Pandas may have had a workaround that now fails.

jreback:
closed by #22142

PR #22142, opened 2018-07-31
Fix zillion deprecation warnings in core.dtypes.common

jbrockmendel:
Looking through a recent Travis log I see tons of these:

```
  /home/travis/build/pandas-dev/pandas/pandas/core/dtypes/common.py:2026: DeprecationWarning: elementwise == comparison failed; this will raise an error in the future.
    if dtype in [object, np.object_, 'object', 'O']:
```

This is driven by (probably not desired) cases where `dtype` is array-like at this point (also showed up in #22092).  Easily avoided by checking for hashability before checking for inclusion.


jreback:
this is #22101

jreback:
thanks!

PR #22146, opened 2018-07-31
Replace (private) _checknull with (public) is_nan

jbrockmendel:
`util._checknull(x)` is equivalent to `x is None or isnan(x)`.  On several occasions I have had to go remind myself what `_checknull` covers.  It is much clearer to write out `x is None or util.is_nan(x)`, which is what this PR does.

In at least one place where `_checknull` is currently used, the `is None` part of it can be skipped.

Also gets rid of usages of private functions, and trades a no-docstring func for a nice-docstring func.

Small assorted cleanups along the way.


pandas/_libs/tslibs/util.pxd
```diff
@@ -228,5 +228,16 @@ cdef inline bint is_offset_object(object val):
...
+
+    Parameters
+    ----------
+    val : object
+
+    Returns
+    -------
+    is_nan : bool
+    """
+    return isinstance(val, float) and val != val
```
jreback: should this be is_float_object (e.g. this excludes compex)

jreback (changes requested):

pandas/_libs/tslibs/util.pxd
jbrockmendel: This is functionally equivalent to the current implementation.

(The current version uses `PyFloatCheck(val)` instead of `isinstance(val, float)`, but cython converts the latter to the former)

jreback (approved):

pandas/_libs/tslibs/util.pxd
jreback: does this handle complex as well? I would rather use the same check as we do elsewhere, as this is confusing when to use one vs the other.

jreback (changes requested):
comments

pandas/_libs/tslibs/util.pxd
jbrockmendel: > does this handle complex as well?

No.  This also only checks for python float, as opposed to is_float_object which also checks for numpy float (the examples I've checked with `isinstance(np.float64(2.3), float)` have come back `True`, not sure what the Falsey case is).

`_libs.missing.checknull` begins with the check you might have in mind:

```
if is_float_object(val) or is_complex_object(val):
    return val != val
[...]
```

ATM `_checknull` is used in exactly 19 places, at least one of which is in a float-only branch.  Based on a quick look it isn't obvious which usages (at least one) are specifically-for-float.  I'll take a look.

jbrockmendel: After looking these over, I'm inclined to think it makes sense to use `is_float_object` but continue excluding complex.  At least two of the usages are floating-specific, and complex is sufficiently corner-case-y that I'd rather check for it explicitly-or-not-at-all.

It also looks like `checknull_with_nat` and `is_null_period` may be unnecessary, will look at that in an upcoming pass.

jreback (approved):

jreback:
can you rebase and merge on green (just to be sure).

PR #22147, opened 2018-07-31
use memoryviews instead of ndarrays

jbrockmendel:
For lookups and slicing memoryviews are supposed to be more performant than ndarrays.  I haven't been able to detect any difference.  @xhochy IIRC you've looked at this in some depth; are memoryviews unambiguously better for our purposes?

Getting rid of `from numpy cimport [...]` is necessary (but not sufficient) for getting rid of numpy 1.7 deprecation warnings.


xhochy:
I also had the feeling that they are more performant and flexible than using `ndarray` but they don't support `const object[:]` at the moment which was problematic in my case. I actually have the feeling that this is a bug in the type parsing of Cython as `const object` is not possible but `const object[:]` should be. Asked for clarification at https://github.com/cython/cython/issues/2485 but got no answer yet and I did not fully understand the Cython codebase to answer it myself.

jreback:
so this is not faster, but not slower? is it worth it at all?

jbrockmendel:
> so this is not faster, but not slower? is it worth it at all?

Based on %timeit results, the memoryviews do appear to be every-so-slightly faster.

The other upsides I have in mind are:
- weens us off of being numpy-specific (think: arrow)
- easier to lint (though we're a few steps away from that)
- getting rid of `from numpy cimport ...` is prerequisite to getting rid of deprecation warnings

jreback:
ok thanks

PR #22152, opened 2018-08-01
implement tslibs/src to make tslibs self-contained

jbrockmendel:


jreback:
looks good. a build is failing, maybe rebase and push again.

pandas/_libs/src/ujson/python/objToJSON.c
```diff
@@ -47,8 +47,8 @@ Numeric decoder derived from from TCL library
 #include <numpy/npy_math.h>       // NOLINT(build/include_order)
 #include <stdio.h>                // NOLINT(build/include_order)
 #include <ultrajson.h>            // NOLINT(build/include_order)
-#include <np_datetime.h>          // NOLINT(build/include_order)
-#include <np_datetime_strings.h>  // NOLINT(build/include_order)
```
gfyoung: I'm confused...why are we making JSON functionality dependent on `tslibs` ?

IMO, this is confusing from a maintenance perspective.  I understand that there is an agenda for `tslib`, but I'm wondering if we are sacrificing other parts of the code base for it.

gfyoung (approved):

PR #21957, opened 2018-07-17
BUG: bug in GroupBy.count where arg minlength passed to np.bincount must be None for np<1.13

topper-123:
- [x] closes #21956
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

See #21956 for details.

PR #22155, opened 2018-08-01
Fix deprecation warning in SeriesGroupBy.count

topper-123:
See https://github.com/pandas-dev/pandas/pull/21957#issuecomment-409527005


jreback:
thanks @topper-123

PR #22164, opened 2018-08-02
CLN: Use public method to capture UTC offsets

mroeschke:
xref https://github.com/pandas-dev/pandas/pull/21822#discussion_r207023121

Per @pganssle's suggestion, avoiding using private dateutil methods to access the timezone offset and more succinctly using `utcoffset` instead


jbrockmendel:
Does this mean instead of capturing eg US/Pacific we get -0800? (Never mind that date until doesn’t yet parse US/Pacific)

mroeschke:
Essentially any datetime string with some sort of timezone information (that gets successfully parsed) will be returned as a fixed offset.

gfyoung (approved):

jreback:
thanks @mroeschke and @pganssle

PR #22174, opened 2018-08-02
DOC: update the Series.str.join docstring

feiphoon:
Checklist for the pandas documentation sprint (ignore this if you are doing an unrelated PR):

- [x] PR title is "DOC: update the docstring"
- [x] The validation script passes: scripts/validate_docstrings.py <your-function-or-method>
- [x] The PEP8 style check passes: git diff upstream/master -u -- "*.py" | flake8 --diff
- [x] The html version looks good: python doc/make.py --single <your-function-or-method>

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1110,8 +1110,11 @@ def str_join(arr, sep):
 
     Notes
     -----
-    If any of the lists does not contain string objects the result of the join
+    If any of the list items is not a string object, the result of the join
     will be `NaN`.
+
+    If the list does not contain string objects, performing a join will raise
+    an AttributeError.
 
     See Also
     --------
@@ -1130,10 +1133,6 @@ def str_join(arr, sep):
     2             NaN
     3    [swan, fish]
     dtype: object
-
-    Join all lists using an '-', the lists containing object(s) of types other
-    than str will become a NaN.
-
     >>> s.str.join('-')
     0      l-i-o-n
     1          NaN
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1108,13 +1108,15 @@ def str_join(arr, sep):
         The list entries concatenated by intervening occurrences of the
         delimiter.
 
+    Raises
+    -------
+    AttributeError
+        If the supplied Series contains neither strings nor lists.
+
     Notes
     -----
     If any of the list items is not a string object, the result of the join
     will be `NaN`.
-
-    If the Series does not contain string objects, performing a join will raise
-    an AttributeError.
 
     See Also
     --------
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1123,7 +1123,6 @@ def str_join(arr, sep):
 
     Examples
     --------
-
     Example with a list that contains non-string elements.
 
     >>> s = pd.Series([['lion', 'elephant', 'zebra'],
@@ -1138,6 +1137,10 @@ def str_join(arr, sep):
     3               [cow, 4.5, goat]
     4    [duck, [swan, fish], guppy]
     dtype: object
+
+    Join all lists using a '-'. The lists containing object(s) of types other
+    than str will produce a NaN.
+
     >>> s.str.join('-')
     0    lion-elephant-zebra
     1                    NaN
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1127,10 +1127,10 @@ def str_join(arr, sep):
     Example with a list that contains non-string elements.
 
     >>> s = pd.Series([['lion', 'elephant', 'zebra'],
-    ...            [1.1, 2.2, 3.3],
-    ...            ['cat', np.nan, 'dog'],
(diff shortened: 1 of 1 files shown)


Cheukting:
Looks great, minor problem, the line ([Line 1108:80](https://github.com/pandas-dev/pandas/blob/ea35874b9713964bbce6f494af2eb777d5386746/pandas/core/strings.py#L1108)) is too long so may be break it up in 2 lines

feiphoon:
Tagging @sebryan as pairing partner

pandas/core/strings.py
```diff
@@ -1105,12 +1105,17 @@ def str_join(arr, sep):
...
+        delimiter.
 
     Notes
     -----
-    If any of the lists does not contain string objects the result of the join
+    If any of the list items is not a string object, the result of the join
     will be `NaN`.
 
+    If the Series does not contain string objects, performing a join will raise
+    an AttributeError.
```
datapythonista: I think that's not very clear. Do you mean that if the Series elements are not lists or `NaN` the join operation will raise an AttributeError?

pandas/core/strings.py
```diff
@@ -1122,21 +1127,17 @@ def str_join(arr, sep):
     Example with a list that contains non-string elements.
 
     >>> s = pd.Series([['lion', 'elephant', 'zebra'],
-    ...                [1.1, 2.2, 3.3],
-    ...                ['cat', np.nan, 'dog'],
-    ...                ['cow', 4.5, 'goat']
-    ...                ['duck', ['swan', 'fish'], 'guppy']])
+    ...            [1.1, 2.2, 3.3],
+    ...            ['cat', np.nan, 'dog'],
+    ...            ['cow', 4.5, 'goat'],
```
datapythonista: Good catch of the typo, but I think the indentaion was correct before.

pandas/core/strings.py
```diff
@@ -1122,21 +1127,17 @@ def str_join(arr, sep):
...
     >>> s
     0        [lion, elephant, zebra]
     1                [1.1, 2.2, 3.3]
     2                [cat, nan, dog]
     3               [cow, 4.5, goat]
     4    [duck, [swan, fish], guppy]
     dtype: object
-
-    Join all lists using an '-', the lists containing object(s) of types other
-    than str will become a NaN.
```
datapythonista: I think this comment was useful, any particular reason to remove it?_

datapythonista (changes requested):

pandas/core/strings.py
feiphoon: D'oh! Thank you.

pandas/core/strings.py
feiphoon: Hmm, I agree. How about:
"If the _supplied_ Series does not contain string objects, performing a join will raise an AttributeError.

```
s = pd.Series([1.1, np.nan]) # applying str.join raises an AttributeError
s = pd.Series([np.nan, np.nan]) # ditto
s = pd.Series([1.1, np.nan, ['swan', 'fish']]) # applying str.join produces something.
```

pandas/core/strings.py
feiphoon: We removed it after misdiagnosing an issue, I will reinstate it. Thank you!

pandas/core/strings.py
datapythonista: To be accurate, I'd say something like `If the supplied Series does not contain strings or lists, performing a join will raise an AttributeError`.

```
>>> import pandas
>>> pandas.Series(['foo', 'bar']).str.join('')
0    foo
1    bar
dtype: object
```

And if you think it's useful, you can add an example with a case when it raises the exception, in the `Examples` section.

feiphoon: 👌 We asked the question of mentioning exceptions last night - of the two examples below, which is the preferred method here?

- **Notes** & **Examples**: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.select_dtypes.html
- **Raises** section: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.update.html

Appreciate all your guidance on this, I'm learning a lot! Thank you.

datapythonista: We don't pay so much attention to the `Raises` section, to not add extra complexity, but it's surely a nice to have. So feel free to add it.

`Notes` is mainly used for implementation details. In this cases `Raises` can be more appropriate. I'd use `Notes` for example to explain that adding a `NaN` to a `int` column will convert it to `float`, and this kind of stuff.

`Examples` are usually the best way to illustrate things. But we don't want to get crazy with them either. Specially in docstrings, that being mixed with the code, can be annoying if too long. In this case, feel free to choose whether you think it's useful to show with an example a case where the `AttributeError` is raised. Or to omit it if you don't think it add much value. Up to you.

datapythonista (approved):
lgtm, thanks for the contribution @feiphoon

jreback:
thanks @feiphoon

jreback:
will merge on green

feiphoon:
Thank you @jreback @datapythonista !

PR #22175, opened 2018-08-02
DOC: updating the Series.str.cat docstring

ym-pett:

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -2063,6 +2063,7 @@ class StringMethods(NoNewAttributesMixin):
         See Also
         --------
         split : Split each string in the Series/Index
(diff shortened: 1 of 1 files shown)


amblina:
We have run ``validate_docstrings.py`` and gotten the following error:

```
Errors found:
  Errors in parameters section
       Parameter "join" description should finish with "."
```
This is nothing to do with our PR.  We're great.

Cheukting:
I think the error is because the "." is followed by "`other`" @datapythonista may be you can varify

TomAugspurger:
Pretty sure that error is from the `.. versionadded:: `. We need to update the script to not complain when there isn't an ending period for those.

TomAugspurger:
Thanks!

PR #22176, opened 2018-08-02
DOC: update the Series.str.contains docstring 

nicolab100:
updated see also section with two more examples


TomAugspurger:
Thanks!

jbrockmendel:
Looks like line 296 is causing flake8 failures in CI, not sure why it wasn’t caught here.

Issue #22166, opened 2018-08-02
DOC: backport py2.7 whatsnew warning


jreback:
xref #21944

PR #22178, opened 2018-08-02
0.23.4 backports 1

TomAugspurger:
Closes https://github.com/pandas-dev/pandas/issues/22166

### DIFF
--- a/pandas/_libs/window.pyx
+++ b/pandas/_libs/window.pyx
@@ -14,6 +14,7 @@ cnp.import_array()
 
 
 cdef extern from "../src/headers/cmath" namespace "std":
+    bint isnan(double) nogil
     int signbit(double) nogil
     double sqrt(double x) nogil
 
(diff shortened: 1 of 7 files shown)


pandas/_libs/window.pyx
```diff
@@ -14,6 +14,7 @@ cnp.import_array()
 
 
 cdef extern from "../src/headers/cmath" namespace "std":
```
TomAugspurger: @chris-b1 @jbrockmendel this changed between 0.23.x and master. On master we use

```cython
cdef extern from "src/headers/cmath" namespace "std":
```

But here we use `../src/headers/cmath`. Does this look correct for 0.23.x?

jbrockmendel: This looks like something I would have nit-picked about.  The "../" is inaccurate in this context.  Evidently the cython/gcc compiler doesn't mind either way, but I'm pretty sure the one in master is More Correct.

gfyoung:
> Closes #22166

@TomAugspurger : I thought we were experimenting with a bot that auto-backported?

Separate note: some weird `pytables` failures on Travis and Circle.  Not sure what happened but rerunning to confirm.

TomAugspurger:
> I thought we were experimenting with a bot that auto-backported?

Yes, tagging an issue with a milestone will automatically backport it if possible.

I think the whatsnew changes weren't tagged with the 0.23.4 milestone, because it contained changes that shouldn't be backported.

The window fix had a merge conflict.

PR #22181, opened 2018-08-02
Standardize special case in tz_conversion functions

jbrockmendel:
ATM `_tz_convert_dst` handles iteration and masking differently (and more complicated-ly) than other related functions.  AFAICT the reasoning is that this implementation is performant when the passed `vals` have a high proportion of `iNaT`s.

By changing this special case to conform to the standard pattern, we allow other functions to call this without having to do any special casting, generally simplify things, and open the door to de-duplicating these functions (which this PR also starts doing by implementing `_tz_convert`)

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


pandas/_libs/tslibs/conversion.pyx
```diff
@@ -757,17 +750,53 @@ cpdef int64_t tz_convert_single(int64_t val, object tz1, object tz2):
...
-        arr = np.array([utc_date])
+        arr[0] = utc_date
         # Note: at least with cython 0.28.3, doing a lookup `[0]` in the next
         # line is sensitive to the declared return type of _tz_convert_dst;
         # if it is declared as returning ndarray[int64_t], a compile-time error
         # is raised.
         return _tz_convert_dst(arr, tz2, to_utc=False)[0]
 
 
+cdef inline int64_t[:] _tz_convert(int64_t[:] vals, object tz, bint to_utc):
```
jreback: can you name this something else, it is pretty confusing

jreback:
thanks!

PR #22187, opened 2018-08-03
DOC: Update the Series.str.len docstring

IsvenC:
[X] PR title is "DOC: update the docstring"
[X] The PEP8 style check passes: git diff upstream/master -u -- "*.py" | flake8 --diff
[X] The html version looks good: python doc/make.py --single

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -2801,11 +2801,53 @@ class StringMethods(NoNewAttributesMixin):
         return self._wrap_result(result)
 
     _shared_docs['len'] = ("""
-    Compute length of each string in the Series/Index.
+    Compute length of each element in the Series/Index. The element may be
+    a sequence (such as a string, tuple or list) or a collection
+    (such as a dictionary).
 
     Returns
     -------
-    lengths : Series/Index of integer values
+    Series or Index of integer values
+        A Series or Index of integer values indicating the length of each
+        element in the Series or Index.
+
+    See Also
+    --------
+    str.len : Python built-in function returning the length (number of items) of an object.
+
+    Examples
+    --------
+
+    Returning a series of integer values as floats when `NaN` is returned as a
+    result.
+
+    >>> s = pd.Series(['dog', 5, 'bird', np.nan])
+    >>> s
+    0     dog
+    1     5
+    2     bird
+    3     NaN
+    dtype: object
+    >>> s.str.len()
+    0    3.0
+    1    NaN
+    2    4.0
+    3    NaN
+    dtype: float64
+
+    Returning the length (number of entries) of dictionaries, lists or tuples as integer values.
+
+    >>> s = pd.Series([{'foo':'bar'}, [2,3,5,7], ('one', 'two', 'three')])
+    >>> s
+    0       {'foo': 'bar'}
+    1         [1, 3, 5, 7]
+    2    (one, two, three)
+    dtype: object
+    >>> s.str.len()
+    0    1
+    1    4
+    2    3
+    dtype: int64
     """)
     len = _noarg_wrapper(len, docstring=_shared_docs['len'], dtype=int)
 
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -2801,7 +2801,7 @@ class StringMethods(NoNewAttributesMixin):
         return self._wrap_result(result)
 
     _shared_docs['len'] = ("""
-    Compute length of each element in the Series/Index. The element may be
+    Compute the length of each element in the Series/Index. The element may be
     a sequence (such as a string, tuple or list) or a collection
     (such as a dictionary).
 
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -2801,11 +2801,54 @@ def rindex(self, sub, start=0, end=None):
...
     _shared_docs['len'] = ("""
-    Compute length of each string in the Series/Index.
+    Compute length of each element in the Series/Index. The element may be
+    a sequence (such as a string, tuple or list) or a collection
+    (such as a dictionary).
 
     Returns
     -------
-    lengths : Series/Index of integer values
+    Series or Index of integer values
```
datapythonista: We usually use Python types, `Series or Index of int` would be more consistent with the rest of the docstrings.

pandas/core/strings.py
```diff
@@ -2801,11 +2801,54 @@ def rindex(self, sub, start=0, end=None):
...
+        A Series or Index of integer values indicating the length of each
+        element in the Series or Index.
+
+    See Also
+    --------
+    str.len : Python built-in function returning the length of an object.
```
datapythonista: What do you think about adding `Series.size` here? I think some users could come by mistake to this page looking how to check the length of the Series. And it'd be nice to help them find it.

pandas/core/strings.py
```diff
@@ -2801,11 +2801,54 @@ def rindex(self, sub, start=0, end=None):
...
+
+    Examples
+    --------
+
```
datapythonista: I don't think we need this blank line here

pandas/core/strings.py
```diff
@@ -2801,11 +2801,54 @@ def rindex(self, sub, start=0, end=None):
...
+    Returning a series of integer values as floats when `NaN` is returned as a
+    result.
```
datapythonista: I think that's a useful comment, but not sure if very relevant in this context (the user checking the examples of the str.len). As this is a general pandas "problem", I'd simply omit it here. But if you think it's important in this page, the `Notes` section is probably a better place.

pandas/core/strings.py
```diff
@@ -2801,11 +2801,54 @@ def rindex(self, sub, start=0, end=None):
...
+    0    3.0
+    1    NaN
+    2    4.0
+    3    NaN
+    dtype: float64
+
+    Returning the length (number of entries) of dictionaries, lists or
+    tuples as integer values.
+
+    >>> s = pd.Series([{'foo':'bar'}, [2,3,5,7], ('one', 'two', 'three')])
```
datapythonista: missing spaces after colon and commas.

pandas/core/strings.py
```diff
@@ -2801,11 +2801,54 @@ def rindex(self, sub, start=0, end=None):
...
+    >>> s
+    0       {'foo': 'bar'}
+    1         [1, 3, 5, 7]
+    2    (one, two, three)
+    dtype: object
+    >>> s.str.len()
+    0    1
+    1    4
+    2    3
+    dtype: int64
```
datapythonista: I would just create a single Series `s` with the elements of both examples. I think that would make the example more concise, and it'd still illustrate the same cases.

pandas/core/strings.py
```diff
@@ -2801,11 +2801,54 @@ def rindex(self, sub, start=0, end=None):
...
+    --------
+    str.len : Python built-in function returning the length of an object.
+
+    Examples
+    --------
+
+    Returning a series of integer values as floats when `NaN` is returned as a
+    result.
+
+    >>> s = pd.Series(['dog', 5, 'bird', np.nan])
```
datapythonista: Very minor thing, but feels like if we're trying to confuse the user by not placing the two string values one after the other. And may be I'd show the empty string `''` instead of one of the animals, that they are illustrating the same case.

datapythonista (commented):
Thanks for the PR @IsvenC, looks great. Just added some minor comments.

pandas/core/strings.py
IsvenC: Good idea! Included in new commit.

pandas/core/strings.py
IsvenC: You're right - I have taken it out

pandas/core/strings.py
IsvenC: Fair enough! A user can still see that the values are integers even if they happen to be floats.

pandas/core/strings.py
IsvenC: Ah yes - thanks for spotting! Have changed in new commit.

pandas/core/strings.py
IsvenC: Good point. Changed in new commit.

pandas/core/strings.py
IsvenC: Have changed the structure of the consolidated example to incorporate an empty string.

datapythonista (approved):
lgtm, great docstring. Thanks for it @IsvenC

jreback:
thanks @IsvenC

Issue #22189, opened 2018-08-03
valid parameters check while reading data


shantanuo:
Pandas seems to ignore the extra (invalid) parameters. For e.g.
#### Code Sample, a copy-pastable example if possible

```python
import pandas as pd
df=pd.read_excel('myfile.xlsx', some_dummy_param=True)
```
Note that some_dummy_param does not throw an error.

#### Problem description

Is there any way to make sure only valid parameters are passed to read_excel method?

#### Expected Output
Since there is no such parameter called "some_dummy_param", I should get an error:

> TypeError: __init__() got an unexpected keyword argument 'some_dummy_param'

TomAugspurger:
I thought we had another issue for this (https://github.com/pandas-dev/pandas/issues/17994 was aobut `sheet_name` vs. `sheetname`).

TomAugspurger:
The most difficult part is that kwds is passed through to the underlying engine. We would want to reconcile the common keyword arguments, and then perhaps provide an `engine_kwargs` parameter that's a dict of engine-specific kwargs.

TomAugspurger:
Yes, that issue was specifically for sheetname.

I'm not sure why a decorator would be useful, could you explain?

shantanuo:
I just saw a decorator called "deprecate_kwarg" to deprecate a keyword argument of a function. I thought similar concept can be used for "validate_kwarg". Ignore this comment if that is not the case.

shantanuo:
May be this function 

```
def validate_kwarg():
    def _validate_kwarg(func):
        @wraps(func)
        def wrapper(*args, **kwargs):
            expected_keys=['io', 'sheet_name','header', 'names', 'index_col', 'usecols', 'squeeze', 'dtype', 'engine', 
               'converters', 'true_values',  'false_values', 'skiprows', 'nrows', 'na_values', 'verbose', 'parse_dates',
               'date_parser', 'thousands', 'comment', 'skipfooter', 'convert_float']

            if set(kwargs.keys()).difference(set(expected_keys)): 
                raise ValueError('invalid parameter found')

            return func(*args, **kwargs)
        return wrapper
    return _validate_kwarg

```
and a decorater added in the file pandas/io/excel.py

@validate_kwarg()
@Appender(_read_excel_doc)
@deprecate_kwarg("parse_cols", "usecols")
@deprecate_kwarg("skip_footer", "skipfooter")
def read_excel(io,

TomAugspurger:
[...]

PR #22193, opened 2018-08-03
Added a decorator function to validate parameters 

shantanuo:
read_excel method needs to check if it has received valid parameters only. Use of **kwds allows any parameter that leads to confusion and bugs.

### DIFF
--- a/pandas/util/_decorators.py
+++ b/pandas/util/_decorators.py
@@ -9,11 +9,11 @@ def validate_kwarg():
     def _validate_kwarg(func):
         @wraps(func)
         def wrapper(*args, **kwargs):
-            expected_keys=['io', 'sheet_name','header', 'names', 'index_col', 'usecols', 'squeeze', 'dtype', 'engine', 
+            expected_keys=set(['io', 'sheet_name','header', 'names', 'index_col', 'usecols', 'squeeze', 'dtype', 'engine', 
                'converters', 'true_values',  'false_values', 'skiprows', 'nrows', 'na_values', 'verbose', 'parse_dates',
-               'date_parser', 'thousands', 'comment', 'skipfooter', 'convert_float']
+               'date_parser', 'thousands', 'comment', 'skipfooter', 'convert_float'])
 
-            if set(kwargs.keys()).difference(set(expected_keys)): 
+            if set(kwargs.keys()).difference(expected_keys): 
                 raise ValueError('invalid parameter found')
 
             return func(*args, **kwargs)
--- a/pandas/util/_decorators.py
+++ b/pandas/util/_decorators.py
@@ -5,6 +5,21 @@ import warnings
 from textwrap import dedent, wrap
 from functools import wraps, update_wrapper, WRAPPER_ASSIGNMENTS
 
+def validate_kwarg():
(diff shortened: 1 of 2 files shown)


jbrockmendel:
Take a look at pandas.util._validators.validate_kwargs

WillAyd:
First and foremost this needs tests.

If this is only for `read_excel` then I think the decorator is overkill - would be better served directly within the function body. 

Is this in reference a particular issue #? If so can you update your post to link to it?

shantanuo:
This is not only related to "read_excel" function. All read_* methods accept *kwargs due to which non-valid parameters are allowed. 

https://github.com/pandas-dev/pandas/issues/22189

This is the only issue that is holding me back from using pandas and therefore I tried to fix it :)

pandas/util/_decorators.py
```diff
@@ -5,6 +5,21 @@
 from textwrap import dedent, wrap
 from functools import wraps, update_wrapper, WRAPPER_ASSIGNMENTS
 
+def validate_kwarg():
+    def _validate_kwarg(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            expected_keys=['io', 'sheet_name','header', 'names', 'index_col', 'usecols', 'squeeze', 'dtype', 'engine',
```
WillAyd: Just instantiate this as a set

pandas/util/_decorators.py
```diff
@@ -5,6 +5,21 @@
...
+def validate_kwarg():
+    def _validate_kwarg(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            expected_keys=['io', 'sheet_name','header', 'names', 'index_col', 'usecols', 'squeeze', 'dtype', 'engine', 
+               'converters', 'true_values',  'false_values', 'skiprows', 'nrows', 'na_values', 'verbose', 'parse_dates',
+               'date_parser', 'thousands', 'comment', 'skipfooter', 'convert_float']
+
+            if set(kwargs.keys()).difference(set(expected_keys)): 
+                raise ValueError('invalid parameter found')
```
WillAyd: Would be nice to be explicit about which parameter isn't working. Also want a test

WillAyd (changes requested):
I still think the wrapper is overcomplicating things. As mentioned before tests are the most critical thing here and ideally should come before any other coding changes

TomAugspurger:
I think we should focus our efforts on removing the need for `**kwargs` from the signature, rather than trying to work around it.

shantanuo:
sure. Decorator was my shortcut to incorporate the functionality that I needed desperately :)

jbrockmendel:
Am I missing something?  Why is the existing validate_kwargs decorator not up to the task?

shantanuo:
I checked the existing validate_kwargs decorator, but could not figure out how to use it :) 
Removing **kwargs seems to be ideal though. Any one of the two (that is practical, easy and quick) approach should be adopted.

jbrockmendel:
@shantanuo any plans to add a test to this?  I doubt this implementation will be merged, but a test could help us figure out how to use the existing decorators for this purpose.

jreback:

PR #22195, opened 2018-08-03
 API: Tick arithmetic respects calendar time

### LINKED ISSUE #20633: API: offsets.Day should be one calendar day instead of 24 hours
xref issues

- [ ] #20596
- [ ] #16980 
- [ ] #8774

As I understand the offset classes, they are supposed to respect transitions in "wall time" instead of absolute time. Currently the `Day` offset is currently defined to be 24 hours instead of 1 calendar day which is problematic to respect "wall time" wrt DST transitions.

```
In [25]: foo
Out[25]: Timestamp('2016-10-30 00:00:00+0300', tz='Europe/Helsinki') # DST change on this day

In [26]: foo + pd.tseries.offsets.DateOffset(weeks=1)
Out[26]: Timestamp('2016-11-06 00:00:00+0200', tz='Europe/Helsinki')

In [27]: foo + pd.tseries.offsets.Week()
Out[27]: Timestamp('2016-11-06 00:00:00+0200', tz='Europe/Helsinki') # respects calendar transition

In [28]: foo + pd.Timedelta(weeks=1)
Out[28]: Timestamp('2016-11-05 23:00:00+0200', tz='Europe/Helsinki')

In [29]: foo + pd.tseries.offsets.DateOffset(days=1)
Out[29]: Timestamp('2016-10-31 00:00:00+0200', tz='Europe/Helsinki')

In [30]: foo + pd.tseries.offsets.Day()
Out[30]: Timestamp('2016-10-30 23:00:00+0200', tz='Europe/Helsinki') # does not respects calendar transition

In [31]: foo + pd.Timedelta(days=1)
Out[31]: Timestamp('2016-10-30 23:00:00+0200', tz='Europe/Helsinki')
```

From prior issues, #20596, #16980, #8774, it seems like many are confused that 1 day = 24 hours when the DST transition comes up and defining 1 day = 1 calendar day would be more intuitive and consistent with other offsets. @

mroeschke:
- [x] closes #20633
- [ ] tests added / passed
- [ ] whatsnew entry

This PR changes `Tick` arithmetic to respect calendar time akin to `Datoffset` for `Timestamp` and `DatetimIndex` (and hence `Series`).  There may be some embedded Tick work-around code (I am thinking in `DatetimeIndex._generate` and `resample` that may be refactorable after this change.


pandas/core/arrays/datetimes.py
```diff
@@ -466,22 +466,19 @@ def _sub_datelike_dti(self, other):
...
-            if self.tz is not None:
-                result = result.tz_localize(self.tz)
 
         except NotImplementedError:
             warnings.warn("Non-vectorized DateOffset being applied to Series "
                           "or DatetimeIndex", PerformanceWarning)
             result = self.astype('O') + offset
 
-        return type(self)(result, freq='infer')
+        return type(self)(result, freq='infer', tz=self.tz)
```
mroeschke: This was changed because if the code goes down the `NotImplementedError` path, the tz would be dropped.

pandas/tseries/offsets.py
```diff
@@ -2229,6 +2228,24 @@ def apply(self, other):
...
+        Parameters
+        ----------
+        idx : DatetimeIndex
+
+        Returns
+        -------
+        DatetimeIndex
+        """
+        # TODO: Add a vectorized DatetimeIndex.dst() method
+        ambiguous = np.array([bool(ts.dst()) if ts is not tslibs.NaT else False
```
mroeschke: I could delay this ambiguous calculation until an AmbiguousTimeError is raised?

jbrockmendel:
Definitely +1 on doing this for Day, but it is less clear what this means for the other Tick subclasses.  What's the motivation?

mroeschke:
For the other Ticks, it's mostly to keep them consistent with the definition of `Day` e.g. a user should expect 24 `Hour`s to behave similarly as 1 `Day`. Otherwise the other Tick subclasses would be redundant with Timedeltas as well.

jbrockmendel:
> For the other Ticks, it's mostly to keep them consistent with the definition of Day e.g. a user should expect 24 Hours to behave similarly as 1 Day. Otherwise the other Tick subclasses would be redundant with Timedeltas as well.

I can see some usefulness here, but what about use cases where the frequency _should_ be Timedelta-like?  e.g. if I call `pd.date_range('2016-01-02 03:04', periods=1000, freq='H', tz='US/Pacific')`, I'd be surprised if the resulting index skipped an hour at a DST change.

This is also going to affect `timedeltas.delta_to_nanoseconds`, I'll have to look at this more closely.

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -268,13 +268,18 @@ cdef class _Timestamp(datetime):
...
 
-        elif PyDelta_Check(other) or hasattr(other, 'delta'):
-            # delta --> offsets.Tick
+        elif PyDelta_Check(other):
             nanos = delta_to_nanoseconds(other)
             result = Timestamp(self.value + nanos,
                                tz=self.tzinfo, freq=self.freq)
+            return result
+
+        elif hasattr(other, 'delta'):
```
jbrockmendel: You might be able to import `tslibs.offsets._Tick` here and make this an isinstance check (thought that might cause a circular import, not sure off the top)

mroeschke:
@jbrockmendel 

I think the core confusion for users is the inherent ambiguity when specifying a string frequency like `'D'` on whether the strings mean calendar time or absolute time.

Pandas advertises them as (Date)offsets, but there's an implicit (and not-clearly-documented) consistency throughout the API that string offsets passed to other methods like `resample` or `date_range` respect absolute time which is consistent with `pd.tseries.frequencies.to_offset('D')` returning a `Day` Tick that also respects absolute time. (This is probably why we have a lot of kludgey code dancing around `Day`)

That being said, I would love for all the frequency strings to map to calendar-respecting `Ticks`
/`DateOffsets`, but that would API breaking for `df.resample('D')` with timezones.

For the sake of consistency, I am okay with the API break, but interested to get your initial thoughts
Probably should open this discussion up to other too.

mroeschke:
@jbrockmendel After thinking about it; I am convinced to only change `Day` to respect calendar arithmetic and leaving the rest of the ticks alone.

To do this I was thinking of having `Day` subclass `DateOffset` instead of `Tick` because the arithmetic paths are already determined by checking for `DateOffset` vs `Tick`. Any thoughts/reservations?

jreback:
@mroeschke ok with your suggestion, it *might* make some code simpler (though likely to break some tests I think)

mroeschke:
Thanks for the input @jreback and @jbrockmendel:

In regards to the `Timestamp + Day = Ambiguous Datetime` issue, I am partial toward raising an `AmbiguousTimeError` for "fall back" DST behavior and `NonExistentTimeError` for "spring forward" behavior for now. In the future we can consider adding a special kwarg to `Day.apply` to handle these ambiguous times.

mroeschke:
Closing as this will require a different implimentation

Issue #22122, opened 2018-07-30
STYLE: Additional flake8 checks


mroeschke:
Currently we ignore some flake8 checks:

Checks that (currently) we're okay with skipping
-  E402,  # module level import not at top of file
-  E731,  # do not assign a lambda expression, use a def
-  W503   # line break before binary operator

Checks that should be run 
- [x]  E741,  # do not use variables named 'l', 'O', or 'I' #22795 (#23131)
- [x] C405,  # Unnecessary (list/tuple) literal - rewrite as a set literal.
- [x] C406   # Unnecessary (list/tuple) literal - rewrite as a dict literal.
- [x] C408,  # Unnecessary (dict/list/tuple) call - rewrite as a literal.
- [x] C409,  # Unnecessary (list/tuple) passed to tuple() - (remove the outer call to tuple()/rewrite as a tuple literal).
- [x] C410   # Unnecessary (list/tuple) passed to list() - (remove the outer call to list()/rewrite as a list literal).

Ideally, we should continue to lint our codebase with reasonable flake8 checks.

jbrockmendel:
I’d be +1 on adding E741

jbrockmendel:
I'm noticing in test_offsets `assert not MonthEnd() in _daterange_cache` which I would expect flake8 to complain should be `assert MonthEnd() not in _daterange_cache`.  Any idea why this isn't being chcked?

jbrockmendel:
Recently saw flake8-eradicate, which looks for commented-out code.  It seems to produce a lot of false-positives, so we probably wouldn't want it in the CI, just a one-time run to do a dedicated remove-it-all sweep.

jbrockmendel:
Not sure if this is specifically flake8, but it'd be nice to use a linter to prevent creep in the dependency structure.  Specifically, once a file has zero-runtime-imports, we could have a check to keep it that way.

mroeschke:
Looks like our code checks are linting all the flake8 codes mentioned above. Going to close this issue out.

PR #22202, opened 2018-08-05
STYLE/LINT: Set literals

mroeschke:
xref #22122

Use `{1, 2, 3}` instead of `set([1, 2, 3])` and enforce with a linting rule.

### DIFF
--- a/pandas/tests/dtypes/test_inference.py
+++ b/pandas/tests/dtypes/test_inference.py
@@ -66,7 +66,7 @@ def test_is_sequence():
     "ll",
     [
         [], [1], (1, ), (1, 2), {'a': 1},
-        set([1, 'a']), Series([1]),
+        {1, 'a'}, Series([1]),
         Series([]), Series(['a']).str,
         np.array([2])])
 def test_is_list_like_passes(ll):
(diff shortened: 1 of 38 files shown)


pandas/tests/dtypes/test_inference.py
```diff
@@ -385,7 +385,7 @@ def test_convert_numeric_uint64_nan(self, coerce, arr):
 
     def test_convert_numeric_uint64_nan_values(self, coerce):
         arr = np.array([2**63, 2**63 + 1], dtype=object)
-        na_values = set([2**63])
+        na_values = {2 ** 63}
```
jbrockmendel: What's the convention for whitespace around `**`?  I thought that one usually stays `2**63`.

jbrockmendel:
Pending green, LGTM.

pandas/tests/dtypes/test_inference.py
mroeschke: Looks like there's some ambiguity for the convention around `**` specifically: https://stackoverflow.com/questions/36794533/pep8-e226-recommendation

However this was auto-formatted automatically by PyCharm (which probably takes a strict interpretation of whitespace around arithmetic operators). I'll just convert it back to match the rest of this function (and retrigger a failed build).

jbrockmendel:
I think we’re safe taking this off @jreback’s plate.  Merging.  Thanks @mroeschke.

PR #22209, opened 2018-08-06
API/CLN: Have toplevel pd.pivot mirror pivot instead of pivot_simple

### LINKED ISSUE #22116: API: Have pd.pivot mirror pivot instead of pivot_simple
To pivot, there's currently:

`pd.pivot_table / df.pivot_table`
https://github.com/pandas-dev/pandas/blob/d30c4a0696d5fbdc3c7ce36a9b9b19224a557e09/pandas/core/reshape/pivot.py#L28

`df.pivot`
https://github.com/pandas-dev/pandas/blob/d30c4a0696d5fbdc3c7ce36a9b9b19224a557e09/pandas/core/reshape/reshape.py#L386

`pivot_simple` (Importable at the top level as `pd.pivot`)
https://github.com/pandas-dev/pandas/blob/d30c4a0696d5fbdc3c7ce36a9b9b19224a557e09/pandas/core/reshape/reshape.py#L411

I'm not sure why `pd.pivot` currently doesn't mirror `df.pivot` like how `pd.pivot_table` mirrors `df.pivot_table`, but it would clearer if these were in sync.

mroeschke:
- [x] closes #22116
- [x] tests added / passed
- [x] whatsnew entry

- Moved `pivot` from `pandas/core/reshape/reshape.py` to `pandas/core/reshape/pivot.py`
- Moved `pivot_simple` and `_slow_pivot` from `pandas/core/reshape/reshape.py` to `pandas/core/reshape/pivot.py` (only used by one panel test)
- Remapped top level `pd.pivot` to use `pivot` instead of `pivot_simple` 
- Added some additional tests for `pd.pivot`


doc/source/whatsnew/v0.24.0.txt
```diff
@@ -467,6 +467,7 @@ Other API Changes
 - :meth:`PeriodIndex.tz_convert` and :meth:`PeriodIndex.tz_localize` have been removed (:issue:`21781`)
 - :class:`Index` subtraction will attempt to operate element-wise instead of raising ``TypeError`` (:issue:`19369`)
 - :class:`pandas.io.formats.style.Styler` supports a ``number-format`` property when using :meth:`~pandas.io.formats.style.Styler.to_excel` (:issue:`22015`)
+- Top level function :meth:`pandas.pivot` now maps directly to the instance method :meth:`DataFrame.pivot` (:issue:`22116`)
```
jreback: this is not necessary, you are cleaning internal code and its not user facing

jreback (changes requested):
small comment, ping when green.

mroeschke:
Ping @jreback. All green.

pandas/core/reshape/pivot.py
```diff
@@ -598,3 +623,68 @@ def _get_names(arrs, names, prefix='row'):
...
+                             ' same')
+    if len(index) == 0:
+        return DataFrame(index=[])
+    hindex = MultiIndex.from_arrays([index, columns])
+    series = Series(values.ravel(), index=hindex)
+    series = series.sort_index(level=0)
+    return series.unstack()
+
+
+def _slow_pivot(index, columns, values):
```
jreback: I don't think we are using this *anywhere* but the ``test_panel`` test? if so I would just blow it away & remove the test

jreback (changes requested):

mroeschke:
@jreback removed those unused methods and test and all green.

jreback (approved):

jreback:
thanks @mroeschke nice clean!

jorisvandenbossche:
We cannot simply change this, as this completely breaks existing usage of `pd.pivot`. This needs to go through a deprecation cycle (or need to think how to keep supporting the old usage).

jorisvandenbossche:
That's certainly an option. 
I personally don't really know to what extent it would be useful to also keep the current behaviour of `pivot_simple` (so to not fully deprecate it). Will open an issue to not forget this.

PR #22213, opened 2018-08-06
Fix Python2 test failures in certain locales

### LINKED ISSUE #22129: Python2 test_datetime_name_accessors failure with some locales
#### Problem description

Python2 **TestDatetime64.test_datetime_name_accessors[en_IL]** fails with `ValueError: unknown locale: en_IL`.  The same test passes in Python3.

By default, **tm.get_locales()** should only return locales that can be set without throwing an exception.

Compare the output in Python2:
```$ python2
Python 2.7.15 (default, May  1 2018, 05:55:50) 
[GCC 7.3.0] on linux2
>>> import pandas.util.testing as tm; tm.get_locales()
['af_ZA.UTF-8', 'C', 'en_US.UTF-8', 'de_AT.UTF-8', 'de_BE.UTF-8', 'de_CH.UTF-8', 'de_DE.UTF-8', 'de_IT.utf8', 'de_LI.UTF-8', 'de_LU.UTF-8', 'en_AG.UTF-8', 'en_AG.UTF-8', 'en_AU.UTF-8', 'en_BW.UTF-8', 'en_CA.UTF-8', 'en_DK.UTF-8', 'en_GB.UTF-8', 'en_HK.UTF-8', 'en_IE.UTF-8', 'en_IL', 'en_IL.utf8', 'en_IN.UTF-8', 'en_NG.UTF-8', 'en_NG.UTF-8', 'en_NZ.UTF-8', 'en_PH.UTF-8', 'en_SG.UTF-8', 'en_US.UTF-8', 'en_ZA.UTF-8', 'en_ZM.UTF-8', 'en_ZM.UTF-8', 'fr_CH.UTF-8', 'fr_FR.ISO8859-1', 'fr_FR.ISO8859-1', 'fr_FR.ISO8859-1', 'fr_FR.UTF-8', 'C', '']
```
with that in Python3:
```$ python3
Python 3.6.6 (default, Jun 27 2018, 14:44:17) 
[GCC 8.1.0] on linux
>>> import pandas.util.testing as tm; tm.get_locales()
['af_ZA.UTF-8', 'C', 'en_US.UTF-8', 'de_AT.UTF-8', 'de_BE.UTF-8', 'de_CH.UTF-8', 'de_DE.UTF-8', 'de_IT.UTF-8', 'de_LI.UTF-8', 'de_LU.UTF-8', 'en_AG.UTF-8', 'en_AG.UTF-8', 'en_AU.UTF-8', 'en_BW.UTF-8', 'en_CA.UTF-8', 'en_DK.UTF-8', 'en_GB.UTF-8', 'en_HK.UTF-8', 'en_IE.UTF-8', 'en_IL.UTF-8', 'en_IL.UTF-8', 'en_IN.UTF-8', 'en_NG.UTF-8', 'en_NG.UTF-8', 'en_NZ.UTF-8', 'en_PH.UTF-8', 'en_SG.UTF-8', 'en_US.UTF-8', 'en_ZA.UTF-8', 'en_ZM.UTF-8', 'en_ZM.UTF-8', 'fr_CH.UTF-8', 'fr_FR.ISO8859-1', 'fr_FR.ISO8859-1', 'fr_FR.ISO8859-1', 'fr_FR.UTF-8', 'C', '']
```
The Python2 output contains `'en_IL', 'en_IL.utf8'` while the Python3 output contains `'en_IL.UTF-8', 'en_IL.UTF-8'`.

#### Expected Output

The same locales should be returned in Python2 and Python3.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.6.final.0
python-bits: 64
OS: Linux
OS-release: 4.15.0-29-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_GB.UTF-8
LOCALE: en_GB.UTF-8

pandas: 0.23.3
pytest: 3.3.2
pip: 9.0.1
setuptools: 39.2.0
Cython: 0.28.4
numpy: 1.14.5
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 5.5.0
sphinx: 1.7.5
patsy: None
dateutil: 2.6.1
pytz: 2018.5
blosc: None
bottleneck: None
tables: 3.4.4
numexpr: 2.6.5
feather: None
matplotlib: 2.2.2
openpyxl: 2.4.9
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 0.9.6
lxml: 4.2.1
bs4: 4.6.0
html5lib: 0.999999999
sqlalchemy: 1.2.8
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

ginggs:
Check that we can also get the locale, after setting it, without raising an Exception.

- [x] closes #22129
- [x] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/tests/util/test_util.py
+++ b/pandas/tests/util/test_util.py
@@ -418,6 +418,32 @@ def test_numpy_errstate_is_default():
 
 
 @td.skip_if_windows
+def test_can_set_locale_valid_set():
+    # Setting the default locale should return True
+    assert tm.can_set_locale('') is True
+
+
+@td.skip_if_windows
+def test_can_set_locale_invalid_set():
+    # Setting an invalid locale should return False
+    assert tm.can_set_locale('non-existent_locale') is False
+
+
+@td.skip_if_windows
+def test_can_set_locale_invalid_get(monkeypatch):
+    # In some cases, an invalid locale can be set,
+    # but a subsequent getlocale() raises a ValueError
+    # See GH 22129
+
+    def mockgetlocale():
+        raise ValueError()
+
+    with monkeypatch.context() as m:
+        m.setattr(locale, 'getlocale', mockgetlocale)
+        assert tm.can_set_locale('') is False
+
+
+@td.skip_if_windows
 class TestLocaleUtils(object):
 
     @classmethod
--- a/pandas/tests/util/test_util.py
+++ b/pandas/tests/util/test_util.py
@@ -418,32 +418,6 @@ def test_numpy_errstate_is_default():
 
 
 @td.skip_if_windows
-def test_can_set_locale_valid_set():
-    # Setting the default locale should return True
-    assert tm.can_set_locale('') is True
-
-
-@td.skip_if_windows
-def test_can_set_locale_invalid_set():
-    # Setting an invalid locale should return False
-    assert tm.can_set_locale('non-existent_locale') is False
-
-
-@td.skip_if_windows
-def test_can_set_locale_invalid_get(monkeypatch):
-    # In some cases, an invalid locale can be set,
-    # but a subsequent getlocale() raises a ValueError
-    # See GH 22129
-
-    def mockgetlocale():
-        raise ValueError()
-
-    with monkeypatch.context() as m:
-        m.setattr(locale, 'getlocale', mockgetlocale)
-        assert tm.can_set_locale('') is False
-
-
-@td.skip_if_windows
 class TestLocaleUtils(object):
 
     @classmethod
@@ -458,6 +432,26 @@ class TestLocaleUtils(object):
     def teardown_class(cls):
         del cls.locales
         del cls.current_locale
+
+    def test_can_set_locale_valid_set(self):
+        # Setting the default locale should return True
+        assert tm.can_set_locale('') is True
+
+    def test_can_set_locale_invalid_set(self):
+        # Setting an invalid locale should return False
+        assert tm.can_set_locale('non-existent_locale') is False
+
+    def test_can_set_locale_invalid_get(self, monkeypatch):
+        # In some cases, an invalid locale can be set,
+        # but a subsequent getlocale() raises a ValueError
+        # See GH 22129
+
+        def mockgetlocale():
+            raise ValueError()
+
+        with monkeypatch.context() as m:
+            m.setattr(locale, 'getlocale', mockgetlocale)
+            assert tm.can_set_locale('') is False
 
     def test_get_locales(self):
         # all systems should have at least a single locale
--- a/pandas/util/testing.py
+++ b/pandas/util/testing.py
@@ -504,23 +504,19 @@ def set_locale(new_locale, lc_var=locale.LC_ALL):
 
     try:
         locale.setlocale(lc_var, new_locale)
-
-        try:
-            normalized_locale = locale.getlocale()
-        except ValueError:
-            yield new_locale
+        normalized_locale = locale.getlocale()
+        if com._all_not_none(*normalized_locale):
+            yield '.'.join(normalized_locale)
         else:
-            if com._all_not_none(*normalized_locale):
-                yield '.'.join(normalized_locale)
-            else:
-                yield new_locale
+            yield new_locale
     finally:
         locale.setlocale(lc_var, current_locale)
 
 
 def can_set_locale(lc, lc_var=locale.LC_ALL):
     """
-    Check to see if we can set a locale without raising an Exception.
+    Check to see if we can set a locale, and subsequently get the locale,
+    without raising an Exception.
 
     Parameters
     ----------
@@ -538,7 +534,8 @@ def can_set_locale(lc, lc_var=locale.LC_ALL):
     try:
         with set_locale(lc, lc_var=lc_var):
             pass
-    except locale.Error:  # horrible name for a Exception subclass
+    except (ValueError,
+            locale.Error):  # horrible name for a Exception subclass
         return False
     else:
         return True


WillAyd (changes requested):
Need to add test cases for this

WillAyd:
Should be able to monkey patch getlocale:

https://docs.pytest.org/en/latest/monkeypatch.html

ginggs:
Thanks!  I'll see what I can do.

TomAugspurger:
Hmm, I seem to have reproduced this in my branch migrating us to circleci 2.0: https://circleci.com/gh/pandas-dev/pandas/16656?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link (search for unknown locale).

ginggs:
@TomAugspurger it seems you have a Debian Stretch container there, and the `locales-all` package is installed.

I'm still struggling to come up with a test case that is not too contrived.

TomAugspurger:
These tests look good to me. Merging later today unless @WillAyd has other comments.

pandas/tests/util/test_util.py
```diff
@@ -417,6 +417,29 @@ def test_numpy_errstate_is_default():
...
+    # but a subsequent getlocale() raises a ValueError
+    # See GH 22129
+
+    def mocksetlocale(category, locale):
+        return locale
+
+    def mockgetlocale():
+        raise ValueError()
+
+    monkeypatch.setattr(locale, 'setlocale', mocksetlocale)
```
WillAyd: Can you use the context manager for patching instead? From the monkey patch docs it is the preferred watching of patching stdlib

WillAyd (changes requested):
One minor comment

ginggs:
@WillAyd updated using monkeypatch.context().

WillAyd (commented):
Context manager looks good. One question on what is being patched

ginggs:
@WillAyd sorry, what is the question?

pandas/tests/util/test_util.py
```diff
@@ -417,6 +417,30 @@ def test_numpy_errstate_is_default():
...
+
+    def mocksetlocale(category, locale):
+        return locale
+
+    def mockgetlocale():
+        raise ValueError()
+
+    with monkeypatch.context() as m:
+        m.setattr(locale, 'setlocale', mocksetlocale)
+        m.setattr(locale, 'getlocale', mockgetlocale)
```
WillAyd: Do we need to patch this? From the comment above, shouldn't it naturally raise a ValueError after setting `non-existent_locale` as the locale?

WillAyd (commented):
Trying again - not sure what happened to my comment earlier

pandas/tests/util/test_util.py
ginggs: If locale.Error is raised while setting `non-existent_locale`, then the locale is not changed, and 'getlocale' will return the current locale.

The second test needs to simulate successfully setting `non-existent_locale` **and** raising a ValueError so we can test the changes in `pandas/util/testing.py`.

WillAyd: Hmm so does `non-existent_locale` fail when getting or setting? Reason I ask is because I think there's a better way to isolate the expected behavior here, i.e. we should only need to mock the getter or setter but not both within a given test

WillAyd: Right - so isn't the patch to `setlocale` unnecessary? So long as `getlocale` raises (as your patch causes to happen) the set behavior is more or less irrelevant (?)

ginggs: If we don't patch `setlocale`, then it will raise locale.Error and we don't get to test what happens when `getlocale` raises ValueError.

WillAyd: Only because we are setting an invalid locale though right? If we set the default locale yet mocked the get aspect to it wouldn’t that give the desired effect?

ginggs: Yes, for the second test we could choose a valid locale, and only patch `getlocale` to raise ValueError, but at this point, how do we know the name of a valid locale?

WillAyd: Can you call getlocale prior to the patch and subsequently send that as an argument to can_set_locale?

ginggs: On my system (Ubuntu 18.04), Python2 gives me:
```
>>> import locale; locale.getlocale()
(None, None)
```
Whereas Python3 gives me:
```
>>> import locale; locale.getlocale()
('en_ZA', 'UTF-8')
```

ginggs: This works (in Python2 and Python3):
```
>>> locale.setlocale(locale.LC_ALL, '')
'en_ZA.UTF-8'
```
So shall we rather go with `assert tm.can_set_locale('') is False` and drop the monkeypatch on `setlocale`?

ginggs:
@WillAyd reworked to drop the monkeypatch on `setlocale`, and added another test.

pandas/tests/util/test_util.py
WillAyd: Are you saying `tm.can_set_locale(locale.getlocale())` isn't working across both versions? I believe a tuple is a valid argument. Would be cleaner to just send that value through instead of the empty string

pandas/tests/util/test_util.py
```diff
@@ -417,6 +417,32 @@ def test_numpy_errstate_is_default():
     assert np.geterr() == expected
 
 
+@td.skip_if_windows
```
WillAyd: Have these skips always been in place or is this a new addition in latest push?

WillAyd (commented):
FYI if you merge in your updates instead of rebasing in the future it will maintain the review messages, making this process a little easier

pandas/tests/util/test_util.py
ginggs: The documentation gives this an example.
https://docs.python.org/2/library/locale.html 
```
import locale
locale.setlocale(locale.LC_ALL, '')

```

WillAyd: Ah OK that's fair - thanks for pointing there!

pandas/tests/util/test_util.py
ginggs: They've been there since I first added the tests.

WillAyd: OK thanks. Can you move these into the `TestLocaleUtils` class? It seems like they should logically be grouped in there and that class already has the decorator to skip windows.

Should be the last change - thanks for hanging with the updates!

ginggs: OK, will do. Unfortunately I'll only be able to get to it tomorrow.

WillAyd (approved):
Lgtm @TomAugspurger merge away if satisfied

TomAugspurger:
Note, `monkeypatch.context` is new in pytest 3.6:
https://docs.pytest.org/en/latest/changelog.html#pytest-3-6-0-2018-05-23

We should update our minimum pytest in `ci/environment-dev.yaml`

WillAyd:
@TomAugspurger good point as that was discussed in https://github.com/pandas-dev/pandas/pull/21531#pullrequestreview-137045726 can submit a PR later today to bump

ginggs:
@WillAyd you are welcome, and thanks for your patience!

WillAyd:
Just looking back here we probably want a whatsnew as well - @ginggs do you want to do that in a separate PR? Should be able to leverage the same `Development` section I am creating in #22320

WillAyd:
Probably easiest that way. Hope to have that merged in soon - thanks!

PR #22217, opened 2018-08-06
Add missing period to pandas.Series.interpolate docstring

adamjstewart:
Previously, the _new in version_ message seemed like a run-on sentence:
https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.interpolate.html

Also added code formatting for `BPoly.from_derivatives` reference.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -6098,7 +6098,7 @@ class NDFrame(PandasObject, SelectionMixin):
             * 'index', 'values': use the actual numerical values of the index
             * 'nearest', 'zero', 'slinear', 'quadratic', 'cubic',
               'barycentric', 'polynomial' is passed to
-              :func:`scipy.interpolate.interp1d`. Both 'polynomial' and
+              :class:`scipy.interpolate.interp1d`. Both 'polynomial' and
               'spline' require that you also specify an `order` (int),
               e.g. ``df.interpolate(method='polynomial', order=4)``.
               These use the actual numerical values of the index.
@@ -6111,7 +6111,7 @@ class NDFrame(PandasObject, SelectionMixin):
               and `tutorial documentation
               <http://docs.scipy.org/doc/scipy/reference/tutorial/interpolate.html>`__
             * 'from_derivatives' refers to
-              :func:`scipy.interpolate.BPoly.from_derivatives` which
+              :meth:`scipy.interpolate.BPoly.from_derivatives` which
               replaces 'piecewise_polynomial' interpolation method in
               scipy 0.18
 
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -6110,13 +6110,13 @@ class NDFrame(PandasObject, SelectionMixin):
               <http://docs.scipy.org/doc/scipy/reference/interpolate.html#univariate-interpolation>`__
               and `tutorial documentation
               <http://docs.scipy.org/doc/scipy/reference/tutorial/interpolate.html>`__
-            * 'from_derivatives' refers to BPoly.from_derivatives which
+            * 'from_derivatives' refers to ``BPoly.from_derivatives`` which
               replaces 'piecewise_polynomial' interpolation method in
               scipy 0.18
 
             .. versionadded:: 0.18.1
 
-               Added support for the 'akima' method
+               Added support for the 'akima' method.
                Added interpolate method 'from_derivatives' which replaces
                'piecewise_polynomial' in scipy 0.18; backwards-compatible with
                scipy < 0.18
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -6098,7 +6098,7 @@ class NDFrame(PandasObject, SelectionMixin):
             * 'index', 'values': use the actual numerical values of the index
             * 'nearest', 'zero', 'slinear', 'quadratic', 'cubic',
               'barycentric', 'polynomial' is passed to
-              :py:func:`scipy.interpolate.interp1d`. Both 'polynomial' and
+              :func:`scipy.interpolate.interp1d`. Both 'polynomial' and
               'spline' require that you also specify an `order` (int),
               e.g. ``df.interpolate(method='polynomial', order=4)``.
               These use the actual numerical values of the index.
@@ -6111,7 +6111,7 @@ class NDFrame(PandasObject, SelectionMixin):
               and `tutorial documentation
               <http://docs.scipy.org/doc/scipy/reference/tutorial/interpolate.html>`__
             * 'from_derivatives' refers to
-              :py:func:`scipy.interpolate.BPoly.from_derivatives` which
+              :func:`scipy.interpolate.BPoly.from_derivatives` which
               replaces 'piecewise_polynomial' interpolation method in
               scipy 0.18
 
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -6098,9 +6098,9 @@ class NDFrame(PandasObject, SelectionMixin):
             * 'index', 'values': use the actual numerical values of the index
             * 'nearest', 'zero', 'slinear', 'quadratic', 'cubic',
               'barycentric', 'polynomial' is passed to
-              ``scipy.interpolate.interp1d``. Both 'polynomial' and 'spline'
-              require that you also specify an `order` (int),
-              e.g. df.interpolate(method='polynomial', order=4).
+              :py:func:`scipy.interpolate.interp1d`. Both 'polynomial' and
+              'spline' require that you also specify an `order` (int),
+              e.g. ``df.interpolate(method='polynomial', order=4)``.
               These use the actual numerical values of the index.
             * 'krogh', 'piecewise_polynomial', 'spline', 'pchip' and 'akima'
               are all wrappers around the scipy interpolation methods of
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -6110,13 +6110,13 @@ def replace(self, to_replace=None, value=None, inplace=False, limit=None,
               <http://docs.scipy.org/doc/scipy/reference/interpolate.html#univariate-interpolation>`__
               and `tutorial documentation
               <http://docs.scipy.org/doc/scipy/reference/tutorial/interpolate.html>`__
-            * 'from_derivatives' refers to BPoly.from_derivatives which
+            * 'from_derivatives' refers to ``BPoly.from_derivatives`` which
```
WillAyd: I think we just use double backticks for literals so this is probably better served with just single back ticks. @datapythonista

adamjstewart: A few lines above, I see `scipy.interpolate.interp1d` wrapped in double backticks. I figured we would want to do the same for `scipy.interpolate.BPoly.from_derivatives`. Do you want me to change `scipy.interpolate.interp1d` to single backticks?

P.S. Should these additionally be hyperlinks to the actual scipy docs? Or do the scipy docs reorganize enough that it wouldn't be stable?

datapythonista: Afaik double backticks are mainly used for code, as they are rendered in monospace. While single backticks don't have a specific meaning, as they can be customized. So, for my understanding, we should use single backticks in this case. But I think in many cases we're using double backticks for similar things.

About the links, they are built automatically for other parts of the documentation, by prefixing `:class:`, `:meth:`, `:ref:`... But I'm not sure if for external packages we've got it set up.

May be @TomAugspurger knows. Otherwise, I'll research and see what's the best we can do.

TomAugspurger: Links to scipy should work, since they're in our intersphinx mapping: https://github.com/pandas-dev/pandas/blob/2156431c3a673fbfb720983474d3cb4edc92c332/doc/source/conf.py#L366-L374 so using `:meth:` is probably best.

adamjstewart:
Alright, see if this is better. I added links to the actual scipy docs for those functions. I wasn't sure if you can do something like `:py:func:``foo.bar`` ` to add backticks in addition to the link? rST linking has never been logical or clear to me...

datapythonista:
Not sure about the `:py:func:` I think it should be just `:func:` if they are not methods. Did you try building the documentation, and see if it renders as expected? `./doc/make.py html` should build it, and then you just need to find the `interpolate` method page in the build directory (`doc/source/build/generated` if I'm not wrong).

adamjstewart:
@datapythonista I got the `:py:func:` syntax from http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py:func.

I tried building the docs but got the following error message:
```console
$ ./make.py html
Running Sphinx v1.7.5

Configuration error:
There is a programable error in your configuration file:

  File "/Users/Adam/pandas/pandas/__init__.py", line 26, in <module>
    from pandas._libs import (hashtable as _hashtable,
  File "/Users/Adam/pandas/pandas/_libs/__init__.py", line 4, in <module>
    from .tslibs import (
  File "/Users/Adam/pandas/pandas/_libs/tslibs/__init__.py", line 4, in <module>
    from .conversion import normalize_date, localize_pydatetime, tz_convert_single
ModuleNotFoundError: No module named 'pandas._libs.tslibs.conversion'

  File "/Users/Adam/anaconda3/lib/python3.6/site-packages/sphinx/config.py", line 161, in __init__
    execfile_(filename, config)
  File "/Users/Adam/anaconda3/lib/python3.6/site-packages/sphinx/util/pycompat.py", line 150, in execfile_
    exec_(code, _globals)
  File "conf.py", line 140, in <module>
    import pandas
  File "/Users/Adam/pandas/pandas/__init__.py", line 35, in <module>
    "the C extensions first.".format(module))
ImportError: C extension: No module named 'pandas._libs.tslibs.conversion' not built. If you want to import pandas from the source directory, you may need to run 'python setup.py build_ext --inplace --force' to build the C extensions first.
```
What am I missing here?

WillAyd:
You need to build the C extensions:

```sh
python setup.py build_ext --inplace
```

adamjstewart:
Thanks, but now I'm having troubles with `cython`:
```console
$ python setup.py build_ext --inplace
running build_ext
pandas._libs.algos: -> [['pandas/_libs/algos.c']]
  File "setup.py", line 755, in <module>
    **setuptools_kwargs)
  File "/Users/Adam/anaconda3/lib/python3.6/site-packages/setuptools/__init__.py", line 129, in setup
    return distutils.core.setup(**attrs)
  File "/Users/Adam/anaconda3/lib/python3.6/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/Users/Adam/anaconda3/lib/python3.6/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/Users/Adam/anaconda3/lib/python3.6/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/Users/Adam/anaconda3/lib/python3.6/distutils/command/build_ext.py", line 339, in run
    self.build_extensions()
  File "setup.py", line 369, in build_extensions
    self.check_cython_extensions(self.extensions)
  File "setup.py", line 366, in check_cython_extensions
    """.format(src=src))
Exception: Cython-generated file 'pandas/_libs/algos.c' not found.
                Cython is required to compile pandas from a development branch.
                Please install Cython or download a release package of pandas.
```
I do have `cython` installed though...
```console
$ pip install cython
Requirement already satisfied: cython in /Users/Adam/anaconda3/lib/python3.6/site-packages (0.27.3)
$ python
Python 3.6.5 |Anaconda custom (64-bit)| (default, Apr 26 2018, 08:42:37) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
>>> import cython
>>> cython.__version__
'0.27.3'
```

WillAyd:
Maybe try starting with a clean before building:

```sh
python setup.py clean --all
```

lmk

adamjstewart:
Same error message. I also tried adding the `--force` flag suggested by the previous error message.

WillAyd:
Are you running within an activated virtual environment? Set up looks OK but maybe something surprising going on with path resolution if not

adamjstewart:
No virtual environment. I just updated Cython from 0.27.3 to 0.28.4 and it seems to be working now??

WillAyd:
Makes sense. We just upped our min version dependency to 0.28.2

adamjstewart:
Alright, I ran:
```console
$ python make.py --num-jobs 4 --single Series.interpolate
```
and everything looks good:

screen shot 2018-08-07 at 3 15 40 pm

However, the scipy function links aren't clickable. Is this just because it's a preview?

P.S. I also tried replacing `:py:func:` with `:func:` and it looks identical. If they both work, is `:func:` preferred?

TomAugspurger:
Yes, just :func:

If you're curious, the :py: is a domain
http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html We
use the default Python domain.

datapythonista:
@adamjstewart I think sphinx does not create the links when we build with `--single`. Not sure why, but if you build the whole documentation, it'll take longer, but it should generate the links.

adamjstewart:
@datapythonista I cleaned and rebuilt all of the docs this time but still no links. 🤷‍♂️

datapythonista:
@adamjstewart did you check if the rest of the links (e.g. `See Also` sections) are working? For example, in `Series.astype` you've got links to other pandas methods and also to numpy. Not sure if the problem is in your environment, or in all links to scipy. Checking other links should help identify the problem.

adamjstewart:
The "See Also" links are working. Links to `numpy.where` in `pandas.DataFrame.where` are working. Links to `scipy.stats.gaussian_kde` in `pandas.DataFrame.plot.kde` are working. But the links added in this PR are not.

pandas/core/generic.py
```diff
@@ -6098,9 +6098,9 @@ def replace(self, to_replace=None, value=None, inplace=False, limit=None,
             * 'index', 'values': use the actual numerical values of the index
             * 'nearest', 'zero', 'slinear', 'quadratic', 'cubic',
               'barycentric', 'polynomial' is passed to
-              ``scipy.interpolate.interp1d``. Both 'polynomial' and 'spline'
-              require that you also specify an `order` (int),
-              e.g. df.interpolate(method='polynomial', order=4).
+              :func:`scipy.interpolate.interp1d`. Both 'polynomial' and
```
datapythonista: Seems like `scipy.interpolate.interp1d` is a class. May be that's the problem, as it requires `:class:` instead of `:func:`?

https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.interp1d.html

pandas/core/generic.py
```diff
@@ -6110,13 +6110,14 @@ def replace(self, to_replace=None, value=None, inplace=False, limit=None,
               <http://docs.scipy.org/doc/scipy/reference/interpolate.html#univariate-interpolation>`__
               and `tutorial documentation
               <http://docs.scipy.org/doc/scipy/reference/tutorial/interpolate.html>`__
-            * 'from_derivatives' refers to BPoly.from_derivatives which
+            * 'from_derivatives' refers to
+              :func:`scipy.interpolate.BPoly.from_derivatives` which
```
datapythonista: And this seems to be a method, so `:meth:` instead of `:func:`.

adamjstewart:
@datapythonista That did the trick, thanks! For the record, Sphinx creates the links even with `--single`, so no need to build _all_ the docs to test.

datapythonista (approved):
lgtm.

It's good to know that `--single` creates the links now, I think it didn't before (may be the problem was a different thing).

Thanks @adamjstewart for the contribution, this is a useful change.

jreback:
thanks @adamjstewart

PR #22219, opened 2018-08-06
[CLN] Merge inference into lib.pyx

jbrockmendel:
Continue trimming down _libs/src.

src/inference is a minor hassle, particularly because ATM there are imports/cimports in both it and lib.  Its existence also complicates setup.py a bit.

Evidently parse_helper.h depended on the order in which it was included (I assume this is a code smell in C, but who knows).  Fixed this.

This is almost cut/paste.  The changes are:
 - Arrange (and de-duplicate) imports at the top of lib
 - whitespace fixups
 - change `for i from 0 <= i < n:` to `for i in range(n):` in a few places
 - a couple of comments about potential optimizations in the next pass


jreback (changes requested):
why is this a good idea? i thought we discussed a long time ago to make this a separate module
isn’t that better than making s bigger module? or is this just not possible ?

jbrockmendel:
> why is this a good idea? i thought we discussed a long time ago to make this a separate module
isn’t that better than making s bigger module? or is this just not possible ?

Think of these as orthogonal goals.  This doesn't change the size of the module, just un-splits the file.  ATM I find myself switching between files to make sure I know what's in the namespace, which really doesn't have any upside.  Plus its progress towards killing off src/

There are a few more things we can move out; most of the low-hanging fruit has been picked.

jreback (approved):

jreback:
ok this is fine, ideally can pull out a separate inference module that is orthogonal though.

Issue #14829, opened 2016-12-08
API: allow DataFrame.rename to take a list-like of colums


chris-b1:
This is closely related to #12392, but I think separate issue.  Proposal would be to be able to pass a list-like to `DataFrame.rename` to make method-chaining easier.  I think it would also be consistent with #11980 (Series rename)

```python
df = pd.DataFrame({'a': [1,2], 'b': [3, 4]})

# make this
df = df.rename(columns=['j', 'k'])

# equivalent to 
df.columns = ['j', 'k']
```

jreback:
note that ``Series.rename`` is affecting the ``.name`` and NOT an index. So though I like that this works like ``.set_index`` it might be a bit of an odd duck.

chris-b1:
Yeah, it's a bit odd - I actually could make [5] below work - but at best confusing next to [4]

```python
In [1]: s = pd.Series(['a', 'b'])

In [2]: s
Out[2]: 
0    a
1    b
dtype: object

In [3]: s.rename('my_series')
Out[3]: 
0    a
1    b
Name: my_series, dtype: object

In [4]: s.rename((1,2))
Out[4]: 
0    a
1    b
Name: (1, 2), dtype: object

In [5]: s.rename([1,2])
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
TypeError: Series.name must be a hashable type
```

jorisvandenbossche:
There is also the idea to call this `relabel` (I thought there was an issue about this, but it is mentioned here: https://github.com/pandas-dev/pandas/issues/12392#issuecomment-186308548, https://github.com/pandas-dev/pandas/pull/11980#issuecomment-170273081). 
If we do this, this makes it possible to more easily expand the possibilities here without conflicting with the series name renaming case.

chris-b1:
Thanks @jorisvandenbossche - although I'm a little wary of adding another DataFrame method, I am in favor of calling this `relabel` vs adding functionality to `rename`

TomAugspurger:
From @toobaz in https://github.com/pandas-dev/pandas/issues/16990#issue-243310775:

> My proposal is the following:
> - add a new method .relabel, which changes the content of the axis, as in

```
df.relabel(['l1', 'l2', 'l3']) # changes df.index labels - assuming df.index is flat
df.relabel([['l1a', 'l1b', 'l1c'], [...], [...]]) # changes df.index labels - assuming df.index has 3 levels
[...]

PR #22225, opened 2018-08-06
ENH: Add set_index to Series

### LINKED ISSUE #21684: ENH: set_index for Series
I was surprised to find that `DataFrame.set_index` has no analogue for `Series`. This is especially surprising because I remember reading in several comments here that it is strongly preferred (i.a. by @jreback) that users do not set attributes like `.name` or `.index` directly -- but currently, the only options to change the index on a `Series` is either that, or reconstructing with `pd.Series(s, index=desired_index)`.

This is relevant in many scenarios, but in case someone would like a more concrete example -- I'm currently working on having `.duplicated` be able to return an inverse for `DataFrame / Series / Index`, see #21645. This inverse needs to link two different indexes -- the one of the original object, and the index of the deduplicated one. To reconstruct from the unique values, one needs exactly such a `.set_index` operation, because `.reindex` in itself cannot read a set of indexes and assign them to a different set of indexes in one go (planning an xref here).

```
s = pd.Series(['a', 'b', 'a', 'c', 'a', 'b'])
isdup, inv = s.duplicated(keep='last', return_inverse=True)
isdup
# 0     True
# 1     True
# 2     True
# 3    False
# 4    False
# 5    False
# dtype: bool

inv
# 0    4
# 1    5
# 2    4
# 3    3
# 4    4
# 5    5
# dtype: int64

unique = s.loc[~isdup]
unique
# 3    c
# 4    a
# 5    b
# dtype: object

reconstruct = unique.reindex(inv)
reconstruct 
# 4    a
# 5    b
# 4    a
# 3    c
# 4    a
# 5    b
# dtype: object
```

This object obviously still has the wrong index to be equal to the original. For `DataFrames`, the reconstruction would work as `unique.reindex(inv.values).set_index(inv.index)`, and consequently, this should be available for `Series` as well:

**Desired**:
```
reconstruct = unique.reindex(inv.values).set_index(inv.index)
reconstruct
# 0    a
# 1    b
# 2    a
# 3    c
# 4    a
# 5    b
# dtype: object
```

h-vetinari:
- [x] closes #21684
- [x] tests added / passed
- [x] examples for series.set_index
- [x] whatsnew entry

Following https://github.com/pandas-dev/pandas/issues/21684#issuecomment-401584844, I unified the method to `core/generic.py`. If this is approved in principle, I'll write some tests (both for DF/Series) and the whatsnew. On that note, it seems to me that `df.set_index` - particularly its kwargs - are basically untested. The only tests I found are:
- `pandas/tests/frame/test_indexing.TestDataFrameIndexingDatetimeWithTZ.test_set_reset`
- `pandas/tests/frame/test_indexing.TestDataFrameIndexingUInt64.test_set_reset`

both of which don't test any of the kwargs.


jreback:
look in alter_axes

there are a ton of tests

h-vetinari:
@jreback I cleaned up the tests for `df.set_index` before I moved to the ones for `series.set_index`. Since I'm guessing that this would be preferred as separate PRs, here's just the cleanup for `df`, without any changes to functionality (except some better warnings): #22236

h-vetinari:
Rebased this on #22236 #22526 and split of the changes from #22486 into a separate commit.

h-vetinari:
Rebased this on #22236 #22526 and split of the changes from #22486 into a separate commit.

If you want to review this without #22486, just follow this link: ```https://github.com/pandas-dev/pandas/pull/22225/commits/3edbaeae7731885bf418c5ac50d0a9bcf82c0ed6```

h-vetinari:
@gfyoung @jreback @WillAyd @TomAugspurger 
Could someone please retrigger the failed travis job? Crazy waiting times on Circle/Appveyor currently.

h-vetinari:
Now that all the preps are out of the way, can finally tackle this seemingly simple PR - adding `set_index` to `Series`. I moved the code from `frame.py` to `generic.py`, as mentioned in the issue by @gfyoung.

It's also green (Circle is behind but already passed before the last commit for linting)

pandas/core/generic.py
```diff
@@ -643,6 +645,142 @@ def _set_axis(self, axis, labels):
...
+
+        >>> df.set_index([[1, 2, 3, 4], 'year'])
+                 month  sale
+           year
+        1  2012  1      55
+        2  2014  4      40
+        3  2013  7      84
+        4  2014  10     31
+        """
+        # parameter keys is checked in Series.set_index / DataFrame.set_index!
```
h-vetinari: I'm not duplicating the - different - checks from `Series.set_index` and `DataFrame.set_index` here - is there any chance that someone calls `NDFrame.set_index`?

h-vetinari:
@jreback @gfyoung 
ping :)

gfyoung (approved):
Nice!

cc @jreback

pandas/core/frame.py
```diff
@@ -3980,12 +3991,7 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
         2  2014  4      40
         3  2013  7      84
         4  2014  10     31
-
```
jreback: is the doc-test run for this doc-string?

pandas/core/frame.py
```diff
@@ -4008,65 +4014,10 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
-            del frame[c]
-
-        # clear up memory usage
-        index._cleanup()
-
-        frame.index = index
-
-        if not inplace:
-            return frame
+        vi = verify_integrity
```
jreback: don't use abbreviations

pandas/core/generic.py
```diff
@@ -642,6 +644,142 @@ def _set_axis(self, axis, labels):
         self._data.set_axis(axis, labels)
         self._clear_item_cache()
 
+    def set_index(self, keys, drop=True, append=False, inplace=False,
+                  verify_integrity=False):
+        """
+        Set the index (row labels) using one or more given arrays (or labels).
```
jreback: i am not sure adding a doc-string actually does anything as you have doc-strings in the subclasses. you are welcome / encouraged to use the appender though (e.g. define the template here) and use it in the subclasses, this pattern is already used heavily, e.g. reindex

pandas/core/generic.py
jreback: no

pandas/core/series.py
```diff
@@ -1075,6 +1075,87 @@ def _set_value(self, label, value, takeable=False):
         return self
     _set_value.__doc__ = set_value.__doc__
 
+    def set_index(self, arrays, append=False, inplace=False,
+                  verify_integrity=False):
+        """
+        Set the Series index (row labels) using one or more columns.
```
jreback: same comment about the doc-string

pandas/core/series.py
```diff
@@ -1075,6 +1075,87 @@ def _set_value(self, label, value, takeable=False):
         return self
     _set_value.__doc__ = set_value.__doc__
 
+    def set_index(self, arrays, append=False, inplace=False,
```
jreback: don't change the argument name, it is called keys currently, leave it that way

jreback (changes requested):
looks generally ok, need to fix the duplicated doc-strings

pandas/core/series.py
h-vetinari: The argument currently does not exist - it makes sense for `DataFrame` to have it names `keys`, since the main application (arguably) is setting the index to a given column *key*, but it would be very confusing for `Series`.

The signatures differ already anyway (no `drop` for `Series`), so I'm strongly -1 on naming this parameter `keys`.

h-vetinari (commented):
Thanks for the review. I templated the docstrings as requested.
It looks a bit awkward because I have to unindent the triple-double-quotes, otherwise the docstring would be wonky.

pandas/core/series.py
jreback: that doesn't make any sense, you are conflating 2 different things. change then name to match.

jreback:
actually i just re-read this. This API is completely different than that of DataFrame, yet uses the same name.

This takes an array of values, while the DataFrame does a list of keys. Adding this is going to be very confusing. This is de-facto ``series.index = array``. Which is not the same API.
we don't really have something like this. There were several attempts to have DataFrame.set_index take an array as well, but these never got off the ground.

cc @TomAugspurger @jorisvandenbossche

h-vetinari:
@jreback 
The only difference is that `DataFrame.set_index` *also* takes column keys. But it's completely usable (on master) to pass arrays, list of arrays, or even a mix of arrays and column keys.

So effectively, `DataFrame.set_index` has `arrays_or_keys` and `Series.set_index` only has `arrays`. The other kwargs (`append`, `inplace`, `verify_integrity`) make sense for both as well.

h-vetinari:
@jreback 
> There were several attempts to have DataFrame.set_index take an array as well, but these never got off the ground.

I'm not sure *when*, but they certainly did get off the ground (this is v.0.23, before you asked me to expand the signature in #22486; also check the tests added/expanded in #22236):

```
>>> import pandas as pd
>>> import numpy as np
>>> pd.__version__
'0.23.4'
>>>
>>> df = pd.DataFrame(np.random.randint(0, 10, (4, 4)), columns=list('abcd'))
>>> df.set_index(['a',          # label
...               df.index,     # Index
...               df.b ** 2,    # Series
...               df.d.values,  # ndarray
...               list('ABCD'), # list
...               'c'])         # label again
              b  d
a   b      c
3 0 4  9 A 3  2  9
4 1 1  2 B 3  1  2
8 2 9  5 C 9  3  5
5 3 25 1 D 7  5  1
```

pandas/core/frame.py
```diff
@@ -3920,72 +3920,61 @@ def shift(self, periods=1, freq=None, axis=0):
         return super(DataFrame, self).shift(periods=periods, freq=freq,
                                             axis=axis)
 
+    @Substitution(
+        klass='DataFrame', other_klass='Series',
+        params="""
```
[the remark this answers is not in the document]
h-vetinari: Happy to use that, didn't know it.

pandas/core/series.py
h-vetinari: That, to me, would be an argument to *not* share the implementation in `generic.py`. A parameter named `keys` makes zero sense for series (and neither does `drop` for that matter - or rather it would be very confusing).

h-vetinari:
@TomAugspurger 
switched to `textwrap.dedent`. Since my answer to your question above has been eaten by github almost completely, here it is again:

> Function signatures of subclasses should match the parent. We should also accept drop in the Series version, even if it's not used..

That to me would then be an argument of *not* unifying the methods in `generic.py`. A parameter `keys` makes zero sense for Series (similarly for `drop`, which would be very confusing as well).

h-vetinari:
@jreback @TomAugspurger 
So, do you want me to implement `Series.set_index` separately (which would be duplicating code somewhat), or stay with a changed signature compared to `NDFrame`?

Personally, I don't see why the signatures can't be different - Series and DataFrame *are* different beasts on many levels, so why shouldn't their methods reflect that?

jreback:
@h-vetinari this is simply a different function. we resisted implementing this a long time ago because it is simply confusing. It needs a new name.

h-vetinari:
@jreback 
Have you looked at https://github.com/pandas-dev/pandas/pull/22225#issuecomment-434719598 ?

`DataFrame.set_index` is fully compatible with using arrays. How is setting the index to an array not in line with the name `set_index`? This is exactly what this method does.

IMO the fact that `Series.set_index` does not support labels is a *much* smaller difference compared to the core functionality of setting the index to an array (I recall you being strongly against directly setting attributes like `.index` in general, but there's currently no other option for `Series`).

I don't care if the implementation is shared or not, but `Series` IMO needs a `set_index` with -- at least -- the ability to set the index to a given list-like / arrary.

jreback:
@h-vetinari you are confusing the purpose of ``.set_axis``. It may be that the added tests previously are wrong. The problem with ``.set_index`` on a DataFrame with an array is that it technically can work with an array and not keys. (meaning its not unambiguous)

h-vetinari:
@jreback 
> @h-vetinari you are confusing the purpose of .set_axis.

I think this would not be a very well-named/designed API - it should be easy to avoiding direct setting of attributes. So instead of **set**ting the **`.index`** attribute directly (which works, but should be discouraged)
```
df.index = ['a', 'b', 'c']
```
it should be `df.set_index`.

>  It may be that the added tests previously are wrong.

The above behaviour (`df.set_index(['a', 'b', 'c'])`) works since long ago; the added tests just increased the number of tested combinations.

> The problem with .set_index on a DataFrame with an array is that it technically can work with an array and not keys. (meaning its not unambiguous)

I get (now) that you don't like the mixing of the functionality between `set_index` and `set_axis` but that is a very deeply ingrained reality already (and IMO makes sense from the point of usability/simplicity).

jreback:
> I get (now) that you don't like the mixing of the functionality between set_index and set_axis but that is a very deeply ingrained reality already (and IMO makes sense from the point of usability/simplicity).

I am not sure why you think this is true. exactly how would this be done today?

h-vetinari:
@jreback 
>> I get (now) that you don't like the mixing of the functionality between set_index and set_axis but that is a very deeply ingrained reality already (and IMO makes sense from the point of usability/simplicity).

> I am not sure why you think this is true. exactly how would this be done today?

Let's start as follows:
```
>>> import pandas
>>> import numpy as np
>>> import pandas as pd
>>> pd.__version__
'0.23.4'
>>> df = pd.DataFrame(np.random.randint(0, 10, (3, 3)), columns=list('abc'))
>>> df
   a  b  c
0  3  7  5
1  0  9  4
2  2  8  6
```
IIUC, you're saying that:
* for keys, one should use `df.set_index('a')`
* for arrays, one should use `df.set_axis(df.a.values, inplace=False)`

However, the functionality of `df.set_index` is a proper superset of `df.set_axis(axis=0)`**, in that it's possible to fill `keys` with *only* `Series`/`Index`/`ndarray`/`list` etc.:
```
>>> df.set_index(pd.Index(df.a))  # same result as Series directly below
>>> df.set_index(df.a) 
   a  b  c
a
3  3  7  5
0  0  9  4
2  2  8  6
>>> df.set_index(df.a.values)  # same result as list directly below
>>> df.set_index([[3, 0, 2]])
   a  b  c
3  3  7  5
0  0  9  4
2  2  8  6
```

In fact, it's even possible to mix all of those input types with column keys in `keys`, as I pointed out in in https://github.com/pandas-dev/pandas/pull/22225#issuecomment-434719598.

IMO:
* `set_axis` should be able to do everything that `set_index` does, and just switch between operating on index/columns based on the `axis`-kwarg (after all, `index` and `columns` are the two axes of a DF).
* `set_index` *should* be able to process list-likes (as it currently does)
* it could be considered to add a method `set_columns` on a `DataFrame`
* The `axis`-kwarg of `set_axis` should just switch between the behaviour of `set_index` (i.e. dealing with keys *and* array-likes) and `set_columns`.
* For Series, the `set_index` and `set_axis` methods would/should be exactly the same.

** there is one caveat, in that lists (and only lists; out of all containers) need to be wrapped in another list, i.e. `df.set_index([[3, 0, 2]])` instead of `df.set_index([3, 0, 2])`

jreback:
@h-vetinari I have already commented. what you are doing is completely ambiguous and broken, and was never the intention of ``.set_index()`` to act as ``.set_axis``. It *happens* to work in the particular case you gave. That is exactly the problem. Until / unless you address this, we are not moving forward. You are trying to expand functionaility that is not allowed.

by your definition above this should produce a new Index, but instead this actually does what the doc-string says.
```
In [17]: df.set_index(['a', 'b', 'c'])
Out[17]: 
Empty DataFrame
Columns: []
Index: [(0, 6, 5), (5, 6, 6), (9, 2, 7)]

```

My point is using ``.set_index()`` for actual  list values and NOT column labels is, while not prohitibed not unambiguous.

h-vetinari:
@jreback 

The list-of column-labels is the *only* case that is ambiguous and needs some care.

And using arrays is not ambiguous or broken, even the (v0.23.4)-docstring mentions:
`keys : column label or list of column labels / **arrays**`

So, coming back to your example:
```
>>> df.set_index([['a', 'b', 'c']])
   a  b  c
a  3  7  5
b  0  9  4
c  2  8  6
>>> df.set_index(['a', 'b', 'c'])
Empty DataFrame
Columns: []
Index: [(3, 7, 5), (0, 9, 4), (2, 8, 6)]
```
I mentioned as much above:
> ** there is one caveat, in that lists (and only lists; out of all containers) need to be wrapped in another list, i.e. df.set_index([[3, 0, 2]]) instead of df.set_index([3, 0, 2])

The ambiguity for DataFrame *already* exists, but for `Series` there is *no* ambiguity (since the column keys aspect is disabled). In effect you're asking me to change something you don't like that:
* is already working for DataFrames
* is unchanged by this PR
* is (IMO) neither inconsistent/ambiguous nor broken

I went into all the points you addressed in detail in https://github.com/pandas-dev/pandas/pull/22225#issuecomment-436199850 and previous comments.

h-vetinari:
Both this PR and #23192 have had persistent a `ResourceWarning` the last few runs. I first thought it was a flaky thing like those Warnings used to be, but this time, it stayed, and I can reproduce it locally (not with `pytest pandas/tests/io/test_parquet.py`, but at least with `pytest pandas/tests/io`).

For example in https://travis-ci.org/pandas-dev/pandas/jobs/453820311 or https://travis-ci.org/pandas-dev/pandas/jobs/453822449:
```
sys:1: ResourceWarning: unclosed <socket.socket fd=16, family=AddressFamily.AF_INET, type=2050, proto=0, laddr=('0.0.0.0', 0)>
sys:1: ResourceWarning: unclosed <socket.socket fd=15, family=AddressFamily.AF_INET, type=2050, proto=0, laddr=('0.0.0.0', 0)>
```
and 
```
=============================== warnings summary ===============================
pandas/core/frame.py::pandas.core.frame.DataFrame.to_parquet
  /home/travis/build/pandas-dev/pandas/pandas/io/parquet.py:129: ResourceWarning: unclosed file <_io.BufferedReader name='df.parquet.gzip'>
    **kwargs).to_pandas()
```

There's also a `stderr` (or `stdout`) warning from the parser-tests surfacing somewhere:
```
..............................................................x...........................................................s....
........................................Skipping line 3: Expected 3 fields in line 3, saw 4
.......................................s.......................................................................................
```

I've narrowed it down to the parquet-s3 tests, but couldn't figure out a way to fix it despite *many* tries, so I'm skipping them for now.

h-vetinari:
@jreback 
> @h-vetinari we had quite some discussion about this: #14829 
and never reached resolution. This is an API question.

The discussion you linked is only tangentially related, plus the functionality I'm talking about is the currenct state of the library. Do you want me to open an issue to discuss what (if anything) to do with `df.update`?

h-vetinari:
@jreback
> > @h-vetinari we had quite some discussion about this: #14829 and never reached resolution. This is an API question.

> The discussion you linked is only tangentially related, plus the functionality I'm talking about is the currenct state of the library. Do you want me to open an issue to discuss what (if anything) to do with `df.update`?

Could you please tell me where you want to have this discussion? In a separate issue or in this PR?

h-vetinari:
Since I haven't been able to get an answer on how to proceed, I opened #24046 to discuss the capabilities of `DataFrame.set_index`.

h-vetinari:
@TomAugspurger @jorisvandenbossche 
This would be ready. Any comment before the cut-off?

TomAugspurger:
Apologies for the delay. I think I'm +1 to the idea of this. Taking a look at the diff now.

pandas/core/series.py
```diff
@@ -1171,6 +1171,71 @@ def _set_value(self, label, value, takeable=False):
...
+        b  II     11
+        c  III    12
+        dtype: int64
+        """)
+    )
+    @Appender(generic.NDFrame.set_index.__doc__)
+    def set_index(self, arrays, append=False, inplace=False,
+                  verify_integrity=False):
+
+        if not isinstance(arrays, list) or all(is_scalar(x) for x in arrays):
```
TomAugspurger: What case does the first condition handle here? I'm having trouble finding a case where `Series.set_index([foo])` but `Series.set_index(foo)` doesn't (or wouldn't without the `[]`).

TomAugspurger (approved):

TomAugspurger:
I can't tell if the doc build error from https://dev.azure.com/pandas-dev/pandas/_build/results?buildId=6464 is real, or whether it's a bug in the linter. I can take a look later if you want.

h-vetinari:
@TomAugspurger 
The docstring validation errors were real - some problems with the templating + dedent, plus some plain old oversights that were not yet being tested before. Should hopefully be passing now (locally it is).

pandas/core/series.py
h-vetinari: @TomAugspurger 
The first condition makes sure the second one can execute (i.e. iteration). If it is not true and we end up wrapping it in a list, it will fail in the next step.

The point that is crucial though is that *only* lists are affected by this (and not all list-likes), because `NDFrame.set_index` would try to interpret a list of scalars as a list of keys.

In any case, I added some comments to clarify this behaviour.

h-vetinari (commented):
Added an explanatory comment.

h-vetinari:
@datapythonista 
There's an error in the docstring validation here that I don't understand, even though `Series.set_index` and `DataFrame.set_index` pass validation for me locally (the two docstrings affected by this PR).

My only guess would be that `NDFrame.set_index` is being validated as well, and crashing on the templates. How would you recommend to solve/circumvent that?

As a side note, it would be awesome to see in the log exactly which docstring was causing an error...

pandas/core/generic.py
```diff
@@ -629,6 +631,103 @@ def _set_axis(self, axis, labels):
...
+            method.
+
+        Returns
+        -------
+        %(klass)s
+            The reindexed %(klass)s. Will be None if inplace is True.
+
+        See Also
+        --------
+        %(other_klass)s.set_index: Method adapted for %(other_klass)s.
```
WillAyd: I think your validation error is that there is no space before the colon on this line

WillAyd (changes requested):

pandas/core/generic.py
h-vetinari: @WillAyd 
If that would cause the error then the `[Series|DataFrame].set_index` docstrings should fail too, shouldn't they? Locally they pass, but let's hope you're right. :)

h-vetinari (commented):
Let's see if this works.

pandas/core/generic.py
WillAyd: Hmm yea. So it does fail locally on NDFrame.set_index with the same error as shown in Azure logs

datapythonista:
@h-vetinari the error here is from `numpydoc`, not from our script. We use it to parse the parts of the docstring, and it's failing to parse it here. I guess there are different numpydoc versions in the CI and in your localhost, if it doesn't fail locally.

You can create an issue to capture numpydoc parser errors, and reraise the exception adding the failing docstring, that would be great. It's the first time I see it failing, that's why it's not implemented.

WillAyd:
I think the error is because the docstring is actually created against NDFrame. AFACIT other items would add that docstring to `_shared_docs` which would then be referenced in the `frame` and `series` modules. You might just need to refactor that to get this to pass

jreback:
@h-vetinari I know you keep updating this, but I am -1 on this for all of the reasons I have stated above. This completely changes the semantics of DataFrame.set_index by allowing thru ambiguous cases of arrays. Furthermore the semantics of Series are just plain weird, which *only* accept an array. This is going to lead to lots of confusion down the road. We have set_axis for exactly this purpose, now you are conflating them.

So unless you remove this, and I don't now if this is possible for Series, then this is a no-go.

h-vetinari:
@jreback 
> This completely changes the semantics of DataFrame.set_index by allowing thru ambiguous cases of arrays. 

I know you're very busy, but I get the feeling you don't read what I've repeated time and time again in this thread: this PR changes absolutely *nothing* about the behaviour of `DataFrame.set_index` - as evidenced by the lack of changes in `tests/frame/test_alter_axes.py`.

You have also not commented on #24046, which I opened specifically to discuss your opposition to the *actual / current / existing* capabilities of `DataFrame.set_index`. 

> Furthermore the semantics of Series are just plain weird, which only accept an array

The only difference between `Series.set_index` and `DataFrame.set_index` in this PR is that the former does not allow column keys - which simply do not make sense for `Series`. 

> We have set_axis for exactly this purpose, now you are conflating them.

I have responded to this in detail above and in #24046. I'm not conflating them.

> So unless you remove this, and I don't now if this is possible for Series, then this is a no-go.

I don't know what you're referring to by "this", can you please elaborate?

There are two core devs who have approved this PR (not counting @WillAyd's change request since it was only about docstring validation), and you're objecting on grounds that do not reflect the actual state of affairs - so IMO you're quite off-base in claiming this PR is so outlandish.

jreback:
> I know you're very busy, but I get the feeling you don't read what I've repeated time and time again in this thread: this PR changes absolutely nothing about the behaviour of DataFrame.set_index - as evidenced by the lack of changes in tests/frame/test_alter_axes.py.

I have never said I have a problem with what ``DataFrame.set_index`` does.

I will repeat once again. ``Series.set_index()`` has completely different semantics than ``DataFrame.set_index()`` which accepts keys (which are column names / levels). NOT an array of values. I am not sure are getting this. There for ``Series.set_index()`` is imply not possible the way you have done it here. 

The fact that DataFrame.set_index() happens to accept in a particular case an array proves the point here, it IS ambiguous and has been tried several times before. It is NOT correct.

I have pointed you to ``.set_axis()`` which does exactly what you are proposing for ``Series.set_index()``. The problem is that the name ``.set_index()`` takes *keys* and that is violated for Series, as are there is no such thing as keys.

h-vetinari:
I strongly object to how you're handling this - what does closing this PR achieve? How do you so easily overrule approving reviews (cc @gfyoung @TomAugspurger) and an ongoing discussion?

> > > @jreback This completely changes the semantics of DataFrame.set_index ...

> > @h-vetinari this PR changes absolutely *nothing* about the behaviour of `DataFrame.set_index`

> @jreback I have never said I have a problem with what DataFrame.set_index does.
[...] `Series.set_index()` has completely different semantics than `DataFrame.set_index()` which accepts keys (which are column names / levels). NOT an array of values.

You are simply wrong about that - it does easily and readily accept arrays of values (and this is a fundamental aspect of its purpose of `set`ting the `index`).

jreback:
@h-vetinari you are not listening. If you want to raise an issue or comment feel free.

h-vetinari:
Quoting myself from [this comment](https://github.com/pandas-dev/pandas/issues/24046#issuecomment-476076780):
> @h-vetinari: After not getting a response in #24697, I'd like to ask again how to proceed here. Deprecating lists within lists would have IMO been the most elegant solution, and was the only complaint against adding `Series.set_index`. Now that the decision has been taken to live with this ambiguity, can we please reopen #22225?

Tagging the participants of #24046 and #24697: @jreback @gfyoung @WillAyd @jorisvandenbossche @TomAugspurger @toobaz @jbrockmendel

h-vetinari:
@jreback @gfyoung @WillAyd @jorisvandenbossche @TomAugspurger
Please reopen this PR.

pilkibun:
wow.

Issue #21778, opened 2018-07-06
API: boolean Extension Array


jreback:
xref #20700

PR #22226, opened 2018-08-06
WIP - Bool Extension Array

WillAyd:
This is nowhere near complete as I have a ton of broken tests that need to be resolved, but theoretically progress towards #21778 as I get my feet wet with EAs.

My thought here was to leverage the masking operations used by the Integer EAs to implement an easy Boolean EA on top of that. I've essentially copied over all of the integer tests as well, though someone may have thoughts on a better way to structure all of this.

Any and all direction greatly appreciated

### DIFF
--- a/pandas/core/arrays/boolean.py
+++ b/pandas/core/arrays/boolean.py
@@ -1,0 +1,52 @@
+import numpy as np
+
+from pandas.core.arrays.integer import IntegerArray, _IntegerDtype
+from pandas.core.dtypes.dtypes import registry
+from pandas.util._decorators import cache_readonly
+
+
+class BooleanDtype(_IntegerDtype):
+    name = "EABool"
+    type = np.bool
+    na_value = np.nan
+
+    @cache_readonly
+    def is_signed_integer(self):
+        return False
+
+    @cache_readonly
+    def is_unsigned_integer(self):
+        return False
+
+    @classmethod
+    def construct_array_type(cls):
+        return BooleanArray
+
+
+class BooleanArray(IntegerArray):
+
+    @cache_readonly
+    def dtype(self):
+        return BooleanDtype()
+
+
+def to_boolean_array(values):
+    """
+    Infer and return an integer array of the values.
+
+    Parameters
+    ----------
+    values : 1D list-like
+
+    Returns
+    -------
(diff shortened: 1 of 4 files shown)


jbrockmendel:
Is the idea to make this a bit-per-entry?  If not, I'm not clear on what the benefit of this is.

WillAyd:
No this would be using int8 underneath - I don't think a bit-per-entry is possible since that's not an addressable unit. 

Benefit would be to give users an easy way to cast to and store boolean data with the same masking technique that we are using for integers to denote missing data, albeit the actual implementation underneath uses int8. I figure that would be easier than completely reimplementing this with a dedicated bool subtype, though I'm also looking for feedback on that front

jreback:
haven’t looked at the impl 

but the clear win for this is efficient boolean arrays with missing values

right now you get nice boolean arrays but as soon as you have a NaN you coerce to object (or worse to float)

jbrockmendel:
> Benefit would be to give users an easy way to cast to and store boolean data with the same masking technique that we are using for integers to denote missing data

BoolNA makes sense to me, thanks for clarifying.

> No this would be using int8 underneath - I don't think a bit-per-entry is possible since that's not an addressable unit.

Yah this would takes some behind-the-scenes trickery.  Something like a length-N bool array being backed by a len-N/8 int8 array.

jreback:
i made an issue about using bitarray as an impl detail for integer NA; wouldnobvoiusly be useful here as well (so this would then be really cheap from a memory perspective)

pandas/core/arrays/boolean.py
```diff
@@ -0,0 +1,52 @@
...
+    type = np.bool
+    na_value = np.nan
+
+    @cache_readonly
+    def is_signed_integer(self):
+        return False
+
+    @cache_readonly
+    def is_unsigned_integer(self):
+        return False
```
jschendel: For what it's worth, you should be able to get `is_signed_integer` and `is_unsigned_integer` for free without needing to override since `np.dtype(np.bool).kind` returns `'b'`.  This does save performing a single comparison and is more explicit though.

jreback:
agree this is a nice idea and we should do it but closing as stale

PR #22228, opened 2018-08-07
ENH: option for groupby.hist to match bins

### LINKED ISSUE #22222: ENH: groupby.hist bins don't match

```python
import numpy as np
import pandas as pd
N = 100
df = pd.DataFrame(np.append(np.random.randn(N), np.random.randn(N)/10), columns = ['rand'])
df['group'] = [0]*N + [1]*N
df.groupby('group')['rand'].hist(bins=20, alpha=0.7)
```
#### Problem description

When the dynamic range of two groups are not the same, histogram bins each group according to its own range. Ideally these should match, taking the union of dynamic ranges and binning it for all. 

The output of above code is:
image

#### Expected Output
Histograms with bins of equal widths (or maybe an input flag to `hist` that could do this)

maybe somethings like the output of following code:

```python
bins = np.linspace(-3, 2.5, 20)
df[df['group']==0]['rand'].hist(bins=bins, alpha=0.7)
df[df['group']==1]['rand'].hist(bins=bins, alpha=0.7)
```
image

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.15.0-29-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.23.3
pytest: 3.5.0
pip: 18.0
setuptools: 39.1.0
Cython: 0.28.1
numpy: 1.14.5
scipy: 1.0.1
pyarrow: None
xarray: None
IPython: 6.3.0
sphinx: 1.7.2
patsy: None
dateutil: 2.7.3
pytz: 2018.5
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: 2.2.2
openpyxl: None
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: None
lxml: None
bs4: None
html5lib: 0.9999999
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: 0.2.0
pandas_datareader: None

</details>

javadnoorb:
- [x] closes #22222
- [ ] tests added / passed
- [x] whatsnew entry

Example code and result:

```python
%matplotlib inline
import numpy as np
import pandas as pd

def test_hist_bins_match():
    # GH-22222
    N = 100
    bins = 5

    np.random.seed(0)
    df = pd.DataFrame(np.append(np.random.randn(N), np.random.randn(N) / 10),
                   columns=['rand'])
    df['group'] = [0] * N + [1] * N
    g = df.groupby('group')['rand']
    bin_range = np.linspace(df['rand'].min(), df['rand'].max(), bins + 1)

    ax = g.hist(bins=bins, alpha=0.7, equal_bins=True)
    both_hists_max = g.apply(lambda x: max(
        np.histogram(x, bins=bin_range)[0])).max()
    assert ax.iloc[0].get_ylim()[1] >= both_hists_max

test_hist_bins_match()
```

image


pandas/core/groupby/groupby.py
```diff
@@ -582,6 +582,15 @@ def wrapper(*args, **kwargs):
                kwargs_with_axis['axis'] is None:
                 kwargs_with_axis['axis'] = self.axis
 
+            if (name == 'hist' and
+                    kwargs.pop('equal_bins', False)):
```
WillAyd: Can you get the same behavior without mutating the kwargs dict?

pandas/core/groupby/groupby.py
```diff
@@ -582,6 +582,15 @@ def wrapper(*args, **kwargs):
...
+                # GH-22222
+                bins = kwargs.pop('bins', None)
+                if type(bins) == int:
```
WillAyd: Would be preferable I think to use `is_integer` from `pandas.core.dtypes.inference`

pandas/tests/plotting/test_groupby.py
```diff
@@ -12,6 +12,25 @@
...
+    df = DataFrame(np.append(np.random.randn(N), np.random.randn(N) / 10),
+                   columns=['rand'])
+    df['group'] = [0] * N + [1] * N
+    g = df.groupby('group')['rand']
+    bin_range = np.linspace(df['rand'].min(), df['rand'].max(), bins + 1)
+
+    ax = g.hist(bins=bins, alpha=0.7, equal_bins=True)
+    both_hists_max = g.apply(lambda x: max(
+        np.histogram(x, bins=bin_range)[0])).max()
+    assert ax.iloc[0].get_ylim()[1] >= both_hists_max
```
WillAyd: Is it not possible to make a stronger assertion here about the actual binned values?

WillAyd (changes requested):
Thanks for the PR!

pandas/core/groupby/groupby.py
javadnoorb: Yes, I can pass `equal_bins` to `hist_series`. Do you also mean this for line 588? Just asking, because that might be a bit tricky to pipe through.

WillAyd: Less concerned about 588 since you end up replacing that anyway

javadnoorb: Thanks. I ended up copying kwargs and mutating that.

pandas/core/groupby/groupby.py
javadnoorb: done

pandas/tests/plotting/test_groupby.py
javadnoorb: Good suggestion! I'm comparing bar widths in the new test.

pandas/plotting/_core.py
```diff
@@ -2470,8 +2470,11 @@ def hist_series(self, by=None, ax=None, grid=True, xlabelsize=None,
         bin edges are calculated and returned. If bins is a sequence, gives
         bin edges, including left edge of first bin and right edge of last
         bin. In this case, bins is returned unmodified.
-    bins: integer, default 10
-        Number of histogram bins to be used
```
javadnoorb: This is repeated and seems redundant.

WillAyd: Repeated where? This should still stay, no?

pandas/plotting/_core.py
```diff
@@ -2480,6 +2483,7 @@ def hist_series(self, by=None, ax=None, grid=True, xlabelsize=None,
     matplotlib.axes.Axes.hist : Plot a histogram using matplotlib.
 
     """
+    # TODO: separate docstrings of series and groupby hist functions (GH-22241)
```
WillAyd: Don't need this comment here - the open issue should suffice

pandas/tests/plotting/test_groupby.py
```diff
@@ -12,6 +12,23 @@
...
+    bins = 5
+
+    np.random.seed(0)
+    df = DataFrame(np.append(np.random.randn(N), np.random.randn(N) / 10),
+                   columns=['rand'])
+    df['group'] = [0] * N + [1] * N
+    g = df.groupby('group')['rand']
+    ax = g.hist(bins=bins, alpha=0.7, equal_bins=True)[0]
+    bin_width_group0 = ax.patches[0].get_width()
+    bin_width_group1 = ax.patches[bins].get_width()
```
WillAyd: Instead of using `np.random` can you not define a set of data whereby you can also easily assert that both groups have an equal number of bins across the same X-scale?

pandas/tests/plotting/test_groupby.py
```diff
@@ -12,6 +12,23 @@
 from pandas.tests.plotting.common import TestPlotBase
 
 
+@td.skip_if_no_mpl
+def test_hist_bins_match():
```
WillAyd: Move this function into the class that already exists in the module

pandas/core/groupby/groupby.py
```diff
@@ -578,15 +579,27 @@ def wrapper(*args, **kwargs):
             # a little trickery for aggregation functions that need an axis
             # argument
             kwargs_with_axis = kwargs.copy()
+            kwargs_wo_axis = kwargs.copy()
+
             if 'axis' not in kwargs_with_axis or \
                kwargs_with_axis['axis'] is None:
                 kwargs_with_axis['axis'] = self.axis
 
+            if name == 'hist' and kwargs_wo_axis.pop('equal_bins', False):
```
WillAyd: I don't think another copy of kwargs is the cleanest solution here - can you not just `get` from the kwargs dict instead of popping?

WillAyd (changes requested):

pandas/plotting/_core.py
javadnoorb: `bins` is defined on line 2468 and 2473. It looks like a mistake when writing the docstring. What do you think?

pandas/tests/plotting/test_groupby.py
javadnoorb: I can replace `np.random` with some predefined values. But I think that comparing the number of bins wouldn't do the job. That's the current state of the code. We're here just changing the range without changing the number of bins.

I've written a new test, that compares the x-axis values of the bins, that should be more rigorous. For some reason the functions see kwargs differently in python 2 and 3, so some CI tests fail in the former. I'll make a commit once I figure out what's going on.

pandas/plotting/_core.py
WillAyd: Ah OK didn't see that. This is fine then

pandas/core/groupby/groupby.py
javadnoorb: In that case `matplotltib.pyplot.hist` throws an error that `equal_bins` is not recognized. An alternative solution that I can think of is to pass `equal_bins` as a named argument to `hist_series` in pandas/plotting/_core.py. But then it will be a dummy variable that's not used within the function. Any suggestions?

javadnoorb: To clarify a bit more, in python3 (but for some reason not python2), any extra argument passed to `kwargs`will be read in this line (as `kwds`):

https://github.com/pandas-dev/pandas/blob/0370740034978d3a63d4b8e5e2c96ff54e7e08ba/pandas/plotting/_core.py#L2501

javadnoorb:
There is a python2 versus python3 discrepancy that's causing the some CI tests to fail. I opened #22285 to deal with it.

javadnoorb:
One job failed for no reason. Otherwise it's green.

pandas/plotting/_core.py
```diff
@@ -2443,7 +2443,7 @@ def hist_frame(data, column=None, by=None, grid=True, xlabelsize=None,
 
 def hist_series(self, by=None, ax=None, grid=True, xlabelsize=None,
                 xrot=None, ylabelsize=None, yrot=None, figsize=None,
-                bins=10, **kwds):
+                bins=10, equal_bins=False, **kwds):
```
javadnoorb: I'm trying the alternative approach here. `equal_bins` will be unused within the function. This is just to avoid popping `kwds`.

pandas/tests/plotting/test_groupby.py
```diff
@@ -10,10 +10,54 @@
...
 from pandas.tests.plotting.common import TestPlotBase
+import pytest
 
 
 @td.skip_if_no_mpl
 class TestDataFrameGroupByPlots(TestPlotBase):
+    @pytest.mark.parametrize(
+        'bins, equal_bins',
+        zip([5, None, np.linspace(-3, 3, 10)], [True, False])
+    )
```
javadnoorb: Trying all possible forms of `bins` with and without `equal_bins`

pandas/core/groupby/groupby.py
```diff
@@ -581,6 +582,17 @@ def wrapper(*args, **kwargs):
             if 'axis' not in kwargs_with_axis or \
                kwargs_with_axis['axis'] is None:
                 kwargs_with_axis['axis'] = self.axis
+            if (name == 'hist' and
+                    kwargs.get('equal_bins', False) is True):
```
WillAyd: Do we need `is True` here or can we use the implicit truthiness of objects?

pandas/core/groupby/groupby.py
```diff
@@ -581,6 +582,17 @@ def wrapper(*args, **kwargs):
...
+                # GH-22222
+                bins = kwargs.get('bins')
+                if bins is None:
+                    bins = 10  # use default value used in `hist_series`
+                if is_integer(bins):
```
WillAyd: Could just use `is_scalar` here since we only need to guard against scalar vs sequence

pandas/core/groupby/groupby.py
```diff
@@ -581,6 +582,17 @@ def wrapper(*args, **kwargs):
             if 'axis' not in kwargs_with_axis or \
                kwargs_with_axis['axis'] is None:
                 kwargs_with_axis['axis'] = self.axis
+            if (name == 'hist' and
+                    kwargs.get('equal_bins', False) is True):
+                # GH-22222
+                bins = kwargs.get('bins')
+                if bins is None:
```
WillAyd: What is the purpose of this? Doesn't the keyword already default to 10?

pandas/tests/plotting/test_groupby.py
```diff
@@ -10,10 +10,54 @@
 import numpy as np
 
 from pandas.tests.plotting.common import TestPlotBase
+import pytest
 
 
 @td.skip_if_no_mpl
 class TestDataFrameGroupByPlots(TestPlotBase):
+    @pytest.mark.parametrize(
```
WillAyd: Readability nit but let's add an empty line here

pandas/tests/plotting/test_groupby.py
```diff
@@ -10,10 +10,54 @@
...
+                              ).all()
+        else:
+            # range of values on x-axis (min, max) for each group
+            hist_ranges = DataFrame(
+                [[points[0, 0, 0], points[num_bins - 1, 1, 0]],
+                 [points[num_bins, 0, 0], points[-1, 1, 0]]],
+                index=[0, 1], columns=['min', 'max'])
+            group_ranges = g.agg([min, max])
+            assert np.isclose(group_ranges, hist_ranges).all()
+        tm.close()
```
WillAyd: Not sure what the purpose of this call is?

pandas/tests/plotting/test_groupby.py
```diff
@@ -10,10 +10,54 @@
...
+        N = 100
+        df = DataFrame(np.append(np.linspace(-3, 3, N),
+                                 np.linspace(-1, 1, N)),
+                       columns=['rand'])
+        df['group'] = [0] * N + [1] * N
+        g = df.groupby('group')['rand']
+        axes = g.hist(bins=bins, alpha=0.7, equal_bins=equal_bins)
+        ax = axes[0]
+
+        if bins is None:
```
WillAyd: Just generally avoid this type of code in tests - you aren't asserting anything about the functionality of the code being tested but rather just adding separate logic here in the test case

pandas/tests/plotting/test_groupby.py
WillAyd: Would still prefer if you could be explicit about what is expected here, i.e. add `expected` as a third parameter here and assert that we get the correct number of bins across the two groups

WillAyd (changes requested):

pandas/core/groupby/groupby.py
javadnoorb: It's just a bit safer this way, as it guards against some non-boolean statements by the user, such as `equal_bins=None`. But your suggestion should be fine if we ignore bad inputs from user.

pandas/core/groupby/groupby.py
javadnoorb: The keyword defaults to 10 after the function is called (`curried_axis`), but we need its value before that (on line 593)

pandas/tests/plotting/test_groupby.py
javadnoorb: The test should be slightly different, depending on the type provided for `bins`. Do you recommend dropping `parametrize` decorator and writing separate tests for these instead?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -640,6 +640,7 @@ Plotting
 
 - Bug in :func:`DataFrame.plot.scatter` and :func:`DataFrame.plot.hexbin` caused x-axis label and ticklabels to disappear when colorbar was on in IPython inline backend (:issue:`10611`, :issue:`10678`, and :issue:`20455`)
 - Bug in plotting a Series with datetimes using :func:`matplotlib.axes.Axes.scatter` (:issue:`22039`)
+- The new argument ``equal_bins`` in :func:`SeriesGroupBy.hist` sets histogram bins to equal values (:issue:`22222`)
```
javadnoorb: This is actually `hist` method of a `SeriesGroupBy` object, although that't not available in the docs:

https://pandas.pydata.org/pandas-docs/stable/api.html

javadnoorb: Should it be added?

javadnoorb: This is available in `api.rst` if that suffices:

```
The following methods are available in both ``SeriesGroupBy`` and
``DataFrameGroupBy`` objects, but may differ slightly, usually in that
the ``DataFrameGroupBy`` version usually permits the specification of an
axis argument, and often an argument indicating whether to restrict
application to columns of a specific data type.

.. autosummary::
   :toctree: generated/

...
   DataFrameGroupBy.hist
```

pandas/tests/plotting/test_groupby.py
```diff
@@ -10,11 +10,52 @@
 import numpy as np
 
 from pandas.tests.plotting.common import TestPlotBase
+import pytest
 
 
 @td.skip_if_no_mpl
 class TestDataFrameGroupByPlots(TestPlotBase):
 
+    @pytest.mark.parametrize('equal_bins, bins, expected1, expected2', (
```
javadnoorb: Need two `expected` variables for the two groups. They correspond to the leftmost part of each histogram bar on the x-axis.

pandas/tests/plotting/test_groupby.py
```diff
@@ -10,11 +10,52 @@
...
+        # x-axis (leftmost) hist bar coordinates:
+        # group0: points[:num_bins]
+        # group1: points[num_bins:]
+        num_bins = len(expected1)
+        points = np.array([patch.get_bbox().get_points()
+                           for patch in ax.patches])[:, 0, 0]
+
+        tm.assert_almost_equal(points[:num_bins], np.array(expected1))
+        tm.assert_almost_equal(points[num_bins:], np.array(expected2))
+        tm.close()
```
javadnoorb: this is to close the figure (to avoid overwriting it during `parametrize`)

doc/source/whatsnew/v0.24.0.txt
WillAyd: Just change the reference to DataFrameGroupBy.hist in the whatsnew - should suffice

pandas/core/groupby/groupby.py
```diff
@@ -581,6 +581,16 @@ def wrapper(*args, **kwargs):
             if 'axis' not in kwargs_with_axis or \
                kwargs_with_axis['axis'] is None:
                 kwargs_with_axis['axis'] = self.axis
+            if name == 'hist' and kwargs.get('equal_bins', False):
+                # GH-22222
+                bins = kwargs.get('bins')
```
WillAyd: Hmm wondering if we even need to use .get here - can we just access by index? I'm assuming that 'equal_bins' should always be there so a KeyError when it's not would be fine

pandas/plotting/_core.py
WillAyd: Might have missed the compat issue but if you are using `.get` it shouldn't matter whether this exists as a keyword or not

pandas/core/groupby/groupby.py
```diff
@@ -581,6 +581,16 @@ def wrapper(*args, **kwargs):
             if 'axis' not in kwargs_with_axis or \
                kwargs_with_axis['axis'] is None:
                 kwargs_with_axis['axis'] = self.axis
+            if name == 'hist' and kwargs.get('equal_bins', False):
```
WillAyd: Can just do `kwargs.get('equal_bins')` - no need for the `False` argument

pandas/core/groupby/groupby.py
jreback: do not  add anything here

this is obscuring already extremely fragile code

pandas/core/groupby/groupby.py
```diff
@@ -581,6 +581,16 @@ def wrapper(*args, **kwargs):
...
+                # GH-22222
+                bins = kwargs.get('bins')
+                if bins is None:
```
WillAyd: Can you point me to where `bins` doesn't default to 10?

pandas/tests/plotting/test_groupby.py
```diff
@@ -10,11 +10,52 @@
...
 
 
 @td.skip_if_no_mpl
 class TestDataFrameGroupByPlots(TestPlotBase):
 
+    @pytest.mark.parametrize('equal_bins, bins, expected1, expected2', (
+        (True, 10,
+         [-3., -2.4, -1.8, -1.2, -0.6, 0., 0.6, 1.2, 1.8, 2.4],
+         [-3., -2.4, -1.8, -1.2, -0.6, 0., 0.6, 1.2, 1.8, 2.4]),
+        (True, None,
```
WillAyd: Assuming the default is always 10 for `bins` then `None` is not a valid argument here and should in fact raise an error instead of giving valid results

jreback (changes requested):

pandas/tests/plotting/test_groupby.py
```diff
@@ -10,11 +10,52 @@
...
+        ax = g.hist(bins=bins, alpha=0.7, equal_bins=equal_bins)[0]
+
+        # x-axis (leftmost) hist bar coordinates:
+        # group0: points[:num_bins]
+        # group1: points[num_bins:]
+        num_bins = len(expected1)
+        points = np.array([patch.get_bbox().get_points()
+                           for patch in ax.patches])[:, 0, 0]
+
+        tm.assert_almost_equal(points[:num_bins], np.array(expected1))
```
WillAyd: Would prefer to use `tm.assert_numpy_array_equal` instead here and on line below

WillAyd (changes requested):

pandas/core/groupby/groupby.py
javadnoorb: Do you mean not in the wrapper? Or also beyond that? Is there any other place you'd suggest?

pandas/core/groupby/groupby.py
javadnoorb: For some reason py3 and py2 `kwargs` behavior are different. In py3 `kwargs` defaults are not set here if the parameter is not provided by the user. I opened an issue on this (#22285). I'm not sure where to move the code (per @jreback's comment) to avoid this.

pandas/core/groupby/groupby.py
javadnoorb: same py2/3 issue. In py3 `equal_bins`, `bins`, etc do not take their default values unless the code inside `hist_series` function is being run.

pandas/plotting/_core.py
javadnoorb: `kwds` can only contain `matploltib.pyplot.hist` input args (b/c of line 2504) and throws an error if `equal_bins` is a keyword.

javadnoorb:
Making a new function that does this is pretty straightforward:
```diff
diff --git a/pandas/core/groupby/groupby.py b/pandas/core/groupby/groupby.py
index 3f84fa0..2d324f4 100644
--- a/pandas/core/groupby/groupby.py
+++ b/pandas/core/groupby/groupby.py
@@ -1001,6 +1001,13 @@ class GroupBy(_GroupBy):
     len(grouped) : int
         Number of groups
     """
+    def new_hist(self, bins=10, equal_bins=False, **kwargs):
+        if equal_bins and is_scalar(bins):
+            # share the same numpy array for all group bins
+            bins = np.linspace(self.obj.min(),
+                               self.obj.max(), bins + 1)
+        return self.hist(bins=bins, **kwargs)
+
     def _bool_agg(self, val_test, skipna):
         """Shared func to call any / all Cython GroupBy implementations"""
```

but I don't know how to overload `SeriesGroupBy.hist` without touching the wrappers, or some other sensitive part of the code. I defer this PR to anyone who knows how to do this.

WillAyd:

PR #22229, opened 2018-08-07
ENH: Add strings_as_fixed_length parameter for df.to_records() (#18146)

### LINKED ISSUE #18146: ENH: add to_records() option to output NumPy string dtypes, not objects
`DataFrame.to_records()` outputs string columns with the `object` dtype, which is sometimes not efficient (e.g. for short, similar-length strings, or when storing with `np.save()`).  I wrote the following function to fix this:

```
def to_records_plain(df):
    """Return a NumPy recarray like df.to_records() but with strings stored as bytes, not objects.
    This gives more compact storage and does not require pickling objects when saving to disk.
    Assumes all object arrays in df are strings.

    >>> df = pd.DataFrame({'a': [1, 2], 'b': [0.5, 0.9], 'c': ['x', 'yyy']})
    >>> to_records_plain(df)
    rec.array([(0, 1,  0.5, b'x'), (1, 2,  0.9, b'yyy')], 
              dtype=[('index', '<i8'), ('a', '<i8'), ('b', '<f8'), ('c', 'S3')])
    """
    records = df.to_records()
    descr = records.dtype.descr
    for ii, (name, dtype) in enumerate(descr):
        if dtype == '|O':
            length = df[name].str.len().max()
            descr[ii] = (name, 'S{}'.format(length))

    return records.astype(descr)
```

I suggest exposing something like this as an option in `DataFrame.to_records()`.  An option to convert to Unicode (`'U'`) too would be good too (NumPy's `'S'` is effectively `bytes` in Python 3).

qinghao1:
This option changes DataFrame.to_records() dtype for string arrays
to 'Sx', where x is the length of the longest string, instead of 'O"

- [x] closes #18146
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/pandas/core/dtypes/inference.py
+++ b/pandas/core/dtypes/inference.py
@@ -398,8 +398,11 @@ def is_dict_like(obj):
     >>> is_dict_like([1, 2, 3])
     False
     """
+    for attr in ("__getitem__", "keys", "__contains__"):
+        if not hasattr(obj, attr):
+            return False
 
-    return hasattr(obj, '__getitem__') and hasattr(obj, 'keys')
+    return True
 
 
 def is_named_tuple(obj):
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -1674,6 +1674,15 @@ class DataFrame(NDFrame):
         for i, v in enumerate(arrays):
             index = i
 
+            # When the names and arrays are collected, we
+            # first collect those in the DataFrame's index,
+            # followed by those in its columns.
+            #
+            # Thus, the total length of the array is:
+            # len(index_names) + len(DataFrame.columns).
+            #
+            # This check allows us to see whether we are
+            # handling a name / array in the index or column.
             if index < index_len:
                 dtype_mapping = index_dtypes
                 name = index_names[index]
@@ -1682,6 +1691,11 @@ class DataFrame(NDFrame):
                 dtype_mapping = column_dtypes
                 name = self.columns[index]
 
+            # We have a dictionary, so we get the data type
+            # associated with the index or column (which can
+            # be denoted by its name in the DataFrame or its
+            # position in DataFrame's array of indices or
+            # columns, whichever is applicable.
             if is_dict_like(dtype_mapping):
                 if name in dtype_mapping:
                     dtype_mapping = dtype_mapping[name]
@@ -1690,6 +1704,11 @@ class DataFrame(NDFrame):
                 else:
                     dtype_mapping = None
 
+            # If no mapping can be found, use the array's
+            # dtype attribute for formatting.
+            #
+            # A valid dtype must either be a type or
+            # string naming a type.
             if dtype_mapping is None:
                 formats.append(v.dtype)
             elif isinstance(dtype_mapping, (type, compat.string_types)):
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -35,7 +35,7 @@ from pandas.compat import (range, map, zip, lmap, lzip, StringIO, u,
                            OrderedDict, PY36, raise_with_traceback,
                            string_and_binary_types)
 from pandas.compat.numpy import function as nv
-
+from pandas.api.types import infer_dtype
 from pandas.core.dtypes.cast import (
     maybe_upcast,
     cast_scalar_to_array,
@@ -1540,7 +1540,8 @@ class DataFrame(NDFrame):
 
         return cls(mgr)
 
-    def to_records(self, index=True, convert_datetime64=None):
+    def to_records(self, index=True, convert_datetime64=None,
+                   stringlike_as_fixed_length=False):
         """
         Convert DataFrame to a NumPy record array.
 
@@ -1557,6 +1558,11 @@ class DataFrame(NDFrame):
 
             Whether to convert the index to datetime.datetime if it is a
             DatetimeIndex.
+         stringlike_as_fixed_length : bool, default False
+             .. versionadded:: 0.24.0
+
+             Store string-likes as fixed-length string-like dtypes
+             (e.g. ``S1`` dtype) instead of Python objects (``O`` dtype).
 
         Returns
         -------
@@ -1598,6 +1604,27 @@ class DataFrame(NDFrame):
         >>> df.to_records(index=False)
         rec.array([(1, 0.5 ), (2, 0.75)],
                   dtype=[('A', '<i8'), ('B', '<f8')])
+
+         By default, strings are recorded as dtype 'O' for object:
+
+         >>> df = pd.DataFrame({'A': [1, 2], 'B': ['abc', 'defg']},
+         ...                   index=['a', 'b'])
+         >>> df.to_records()
+         rec.array([('a', 1, 'abc'), ('b', 2, 'defg')],
+                   dtype=[('index', 'O'), ('A', '<i8'), ('B', 'O')])
+
+         This can be inefficient (e.g. for short strings, or when storing with
+         `np.save()`). They can be recorded as fix-length string-like dtypes
+         such as 'S1' for zero-terminated bytes instead:
+
+         >>> df = pd.DataFrame({'A': [1, 2], 'B': ['abc', 'defg']},
+         ...                   index=['a', 'b'])
+         >>> df.to_records(stringlike_as_fixed_length=True)
+         rec.array([('a', 1, 'abc'), ('b', 2, 'defg')],
+                   dtype=[('index', '<U1'), ('A', '<i8'), ('B', '<U4')])
+
+        Notice how the 'B' column is now stored as '<U4' for length-four
+        strings ('S4' for Python 2.x) instead of the 'O' object dtype.
         """
 
         if convert_datetime64 is not None:
@@ -1633,7 +1660,47 @@ class DataFrame(NDFrame):
             arrays = [self[c].get_values() for c in self.columns]
             names = lmap(compat.text_type, self.columns)
 
-        formats = [v.dtype for v in arrays]
+        formats = []
+
+        for v in arrays:
+            if not stringlike_as_fixed_length:
+                formats.append(v.dtype)
+            else:
+                # gh-18146
+                #
+                # For string-like arrays, set dtype as zero-terminated bytes
+                # with max length equal to that of the longest string-like.
+                dtype = infer_dtype(v)
+                symbol = None
+
+                if dtype == "string":
+                    # In Python 3.x, infer_dtype does not
+                    # differentiate string from unicode
+                    # like NumPy arrays do, so we
+                    # specify unicode to be safe.
+                    symbol = "S" if compat.PY2 else "U"
+                elif dtype == "unicode":
+                    # In Python 3.x, infer_dtype does not
+                    # differentiate string from unicode.
+                    #
+                    # Thus, we can only get this result
+                    # in Python 2.x.
+                    symbol = "U"
+                elif dtype == "bytes":
+                    # In Python 2.x, infer_dtype does not
+                    # differentiate string from bytes.
+                    #
+                    # Thus, we can only get this result
+                    # in Python 3.x. However, NumPy does
+                    # not have a fixed-length bytes dtype
+                    # and just uses string instead.
+                    symbol = "S"
+
+                if symbol is not None:
+                    formats.append("{}{}".format(symbol, max(map(len, v))))
+                else:
+                    formats.append(v.dtype)
+
         return np.rec.fromarrays(
             arrays,
             dtype={'names': names, 'formats': formats}
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -35,7 +35,6 @@ from pandas.compat import (range, map, zip, lmap, lzip, StringIO, u,
                            OrderedDict, PY36, raise_with_traceback,
                            string_and_binary_types)
 from pandas.compat.numpy import function as nv
-from pandas.api.types import infer_dtype
 from pandas.core.dtypes.cast import (
     maybe_upcast,
     cast_scalar_to_array,
@@ -1541,7 +1540,7 @@ class DataFrame(NDFrame):
         return cls(mgr)
 
     def to_records(self, index=True, convert_datetime64=None,
-                   stringlike_as_fixed_length=False):
+                   column_dtypes=None, index_dtypes=None):
         """
         Convert DataFrame to a NumPy record array.
 
@@ -1558,11 +1557,20 @@ class DataFrame(NDFrame):
 
             Whether to convert the index to datetime.datetime if it is a
             DatetimeIndex.
-         stringlike_as_fixed_length : bool, default False
-             .. versionadded:: 0.24.0
+        column_dtypes : str, type, dict, default None
+            .. versionadded:: 0.24.0
 
-             Store string-likes as fixed-length string-like dtypes
-             (e.g. ``S1`` dtype) instead of Python objects (``O`` dtype).
+            If a string or type, the data type to store all columns. If
+            a dictionary, a mapping of column names and indices (zero-indexed)
+            to specific data types.
+        index_dtypes : str, type, dict, default None
+            .. versionadded:: 0.24.0
+
+            If a string or type, the data type to store all index levels. If
+            a dictionary, a mapping of index level names and indices
+            (zero-indexed) to specific data types.
+
+            This mapping is applied only if `index=True`.
 
         Returns
         -------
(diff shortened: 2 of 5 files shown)


pandas/core/frame.py
```diff
@@ -1351,6 +1352,9 @@ def to_records(self, index=True, convert_datetime64=None):
 
             Whether to convert the index to datetime.datetime if it is a
             DatetimeIndex.
+        strings_as_bytes : boolean, default False
+            Store strings as bytes (``S`` dtype) instead of Python objects
+            (``O`` dtype)
```
gfyoung: Add a `..versionadded` line right below your docstring description.

pandas/core/frame.py
```diff
@@ -1436,7 +1458,17 @@ def to_records(self, index=True, convert_datetime64=None):
             arrays = [self[c].get_values() for c in self.columns]
             names = lmap(compat.text_type, self.columns)
 
-        formats = [v.dtype for v in arrays]
+        if strings_as_bytes:
+            is_string = np.vectorize(lambda s: type(s) is str)
```
gfyoung: I would rather use `compat.string_types` for this check.

qinghao1:
Thanks for the comments! Made the changes you requested.

pandas/core/frame.py
```diff
@@ -1335,7 +1335,8 @@ def from_records(cls, data, index=None, exclude=None, columns=None,
 
         return cls(mgr)
 
-    def to_records(self, index=True, convert_datetime64=None):
+    def to_records(self, index=True, convert_datetime64=None,
```
jreback: so this is an extremely misleading parameter name. These are NOT bytes, rather they are fixed length strings. maybe ``strings_as_fixed_length``

pandas/tests/frame/test_convert_to.py
```diff
@@ -186,6 +186,18 @@ def test_to_records_with_categorical(self):
                                 dtype=[('index', '=i8'), ('0', 'O')])
         tm.assert_almost_equal(result, expected)
 
+    def test_to_records_with_strings_as_bytes(self):
+
+        # GH18146
+
+        df = DataFrame({'A': [1, 2], 'B': ['abc', 'defg']},
```
jreback: can you test with some mixed object dtypes that include strings, e.g. ``[1, '1']``, this should preserve object dtype.

jreback: maybe just add another column to your test frame

jreback (changes requested):

qinghao1:
@jreback Thanks for the feedback! I've made the changes.

pandas/core/frame.py
```diff
@@ -1436,7 +1460,21 @@ def to_records(self, index=True, convert_datetime64=None):
             arrays = [self[c].get_values() for c in self.columns]
             names = lmap(compat.text_type, self.columns)
 
-        formats = [v.dtype for v in arrays]
+        if strings_as_fixed_length:
+            def is_string(arr):
```
jreback: don't do this,  rather use

``lib.infer_dtype(arr) == 'string'`` (there is also a unicode variant, so I guess use 'U' for that)
for that matter if you actually have bytes it will return that as well.

Anything else would trigger 'O', note that this routine will correctly infer all dtypes

pandas/tests/frame/test_convert_to.py
```diff
@@ -186,6 +186,18 @@ def test_to_records_with_categorical(self):
                                 dtype=[('index', '=i8'), ('0', 'O')])
         tm.assert_almost_equal(result, expected)
 
+    def test_to_records_with_strings_as_fixed_length(self):
+
+        # GH18146
+
```
jreback: so test with strings, unicode, bytes

pandas/core/frame.py
jreback: on second though call this ``stringlike_as_fixed_length``, to encompass string, unicode, bytes

pandas/core/frame.py
jreback: e.g.

```
# this is just the public interface
In [3]: from pandas.api.types import infer_dtype

NameError: name 'ifner_dtype' is not defined

In [5]: infer_dtype(['foo', 'bar'])
Out[5]: 'string'

In [6]: infer_dtype([b'foo', b'bar'])
Out[6]: 'bytes'

# in PY2 this would be unicode
In [7]: infer_dtype([u'foo', u'bar'])
Out[7]: 'string'

In [8]: infer_dtype([u'foo', b'bar'])
Out[8]: 'mixed'
```

jreback (changes requested):

pandas/core/frame.py
qinghao1: @jreback Sorry for the late reply. I think the original issue #18146 was meant for ASCII-encoded strings, and your comment has made me realize that this wouldn't work for strings with unicode characters. (Numpy throws an error like `UnicodeEncodeError: 'ascii' codec can't encode character '\u212b' in position 0: ordinal not in range(128)`).

I propose that this feature be implemented for `bytes` objects instead, which are still encoded as `O` dtype but can be more efficiently encoded as `S` dtype. And the option would be called `bytes_as_fixed_length`. What do you think?

jreback:
can you rebase

jreback:
can you merge master

jreback:
@gfyoung would you be able to merge master and update this

gfyoung:
@jreback : I rebased and addressed all of the comments, including leveraging `infer_dtype` more extensively in the enhancement + adding a lot more tests.

That being said, I'm not 100% certain if the design decisions made were the best.  I commented pretty extensively in the `frame.py` diff regarding what I did.

pandas/core/frame.py
```diff
@@ -1569,7 +1596,47 @@ def to_records(self, index=True, convert_datetime64=None):
...
+                    symbol = "U"
+                elif dtype == "bytes":
+                    # In Python 2.x, infer_dtype does not
+                    # differentiate string from bytes.
+                    #
+                    # Thus, we can only get this result
+                    # in Python 3.x. However, NumPy does
+                    # not have a fixed-length bytes dtype
+                    # and just uses string instead.
+                    symbol = "S"
```
gfyoung: Again, not sure this is best, but is the best fit I could find with how `numpy` and `Python 2.x` works.

gfyoung:
@jreback : The rebase worked, and all is green.  I left some comments for consideration during review.  PTAL

gfyoung:
@jreback : The rebase worked, and all is green.  I left some comments for consideration during review.  PTAL

gfyoung:
@jreback : Could you take a look at this?  Would need some feedback on some of the implementation choices I made here...

pandas/core/frame.py
```diff
@@ -1476,7 +1476,8 @@ def from_records(cls, data, index=None, exclude=None, columns=None,
 
         return cls(mgr)
 
-    def to_records(self, index=True, convert_datetime64=None):
+    def to_records(self, index=True, convert_datetime64=None,
+                   stringlike_as_fixed_length=False):
```
jreback: maybe we should instead have a ``dtype=`` parameter? that accepts a dict?

jreback (changes requested):

pandas/core/frame.py
gfyoung: In place of the currently added parameter?  That would also work, as then we don’t have to deal with the messy dtype inference ourselves...

I’ll give that implementation a shot.

gfyoung: In place of the currently added parameter?  That would also work, as then we don’t have to deal with the messy dtype inference ourselves...

I’ll give that implementation a shot.

pandas/core/frame.py
```diff
@@ -1569,7 +1596,47 @@ def to_records(self, index=True, convert_datetime64=None):
...
+                # with max length equal to that of the longest string-like.
+                dtype = infer_dtype(v)
+                symbol = None
+
+                if dtype == "string":
+                    # In Python 3.x, infer_dtype does not
+                    # differentiate string from unicode
+                    # like NumPy arrays do, so we
+                    # specify unicode to be safe.
+                    symbol = "S" if compat.PY2 else "U"
```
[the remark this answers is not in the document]
gfyoung: This is now moot.

pandas/core/frame.py
gfyoung: This is now moot.

pandas/core/frame.py
gfyoung: Because we can specify `dtype` for both columns and indices, added parameters for each.

Also considered just one giant `dtype` parameter to encompass both, but that felt a little cumbersome (e.g. you would always have nested dictionaries), especially people are probably used to generally specifying just column dtypes (e.g. `read_csv`).  Allowing the specification of `index` dtypes is a nice addition (and consistent with the original changes, which affected both indices and columns) nonetheless.

pandas/core/frame.py
```diff
@@ -1598,6 +1612,18 @@ def to_records(self, index=True, convert_datetime64=None):
         >>> df.to_records(index=False)
         rec.array([(1, 0.5 ), (2, 0.75)],
                   dtype=[('A', '<i8'), ('B', '<f8')])
+
+        Data types can be specified for the columns:
+
+        >>> df.to_records(column_dtypes={"A": "int32"})
```
jreback: can you add an example which looks something like the OP, e.g. 

``df.to_records(index_dtypes="<S{}".format(df.index.str.len().max()))``

jreback (changes requested):

pandas/core/frame.py
gfyoung: Sure, done.

gfyoung:
@jreback : Addressed your comments, and all is green again.  PTAL.

pandas/core/frame.py
```diff
@@ -1557,6 +1557,20 @@ def to_records(self, index=True, convert_datetime64=None):
 
             Whether to convert the index to datetime.datetime if it is a
             DatetimeIndex.
+        column_dtypes : str, type, dict, default None
+            .. versionadded:: 0.24.0
+
+            If a string or type, the data type to store all columns. If
```
jreback: why is this anything but a dict of column_name -> dtype? what does the indices have to do with this?

jreback (changes requested):
are you testing for a multi-index for the index? I think this is ok
multi-index on the columns? does a rec-array even work with this?

pandas/core/frame.py
gfyoung: You can map column indices to dtypes e.g.

```
{0: "int32"}
```

maps `df.columns[0]` to the `int32` `dtype`.

This design was taken from `read_csv`, where we do something similar.

gfyoung: > are you testing for a multi-index for the index? I think this is ok
multi-index on the columns? does a rec-array even work with this?

gfyoung: Added tests for MI, both in the columns and indices.

gfyoung:
@jreback : Added more tests, and everything is finally all green again.  PTAL.

jreback:
@TomAugspurger thoughts on this.

pandas/core/frame.py
```diff
@@ -1620,20 +1651,55 @@ def to_records(self, index=True, convert_datetime64=None):
...
+
+            if index < index_len:
+                dtype_mapping = index_dtypes
+                name = index_names[index]
+            else:
+                index -= index_len
+                dtype_mapping = column_dtypes
+                name = self.columns[index]
+
+            if isinstance(dtype_mapping, dict):
```
TomAugspurger: Don't hardcode dict here. Use `compat.Mapping`.

TomAugspurger (commented):
+1 to the idea in general.

The implementation looks good aside from hardcoding dict.

My only quibble is with splitting `index_dtypes` and `column_dtypes`. Elsewhere in pandas (e.g. merge, groupby) we've allowed targeting index (levels) with keywords that were previously column-only. API-wise, I'd find that cleaner here. The downside would be cases where you want to apply a dtype all all the columns, and a different dtype to the index. In that case you would need to build a dict, rather than `(index_dtypes=int, column_dtypes=str).

pandas/core/frame.py
jreback: actually use ``is_dict_like``

pandas/core/frame.py
```diff
@@ -1557,6 +1557,20 @@ def to_records(self, index=True, convert_datetime64=None):
 
             Whether to convert the index to datetime.datetime if it is a
             DatetimeIndex.
+        column_dtypes : str, type, dict, default None
```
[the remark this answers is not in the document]
gfyoung: Yep, that’s why I use `type` and not `numpy.dtype`.

pandas/core/frame.py
gfyoung: Ah, good catch.  I’ll fix that.

gfyoung: One thing though: `is_dict_like` does not guarantee existence of the `__contains__` method, which is what my implementation also relies on.

I could do some try-except logic to make it more generic, or I could make the `is_dict_like` check slightly stricter.

Thoughts?

TomAugspurger:
collections.abc.Mapping ensures that. Seems like exactly what we want here.

Messing with the various is_*_like methods makes me nervous.

pandas/core/frame.py
jreback: I would try and see if you can make is_dict_like stronger I think its ok here

jreback: I really hate to use other things because they are completely unmaintanable in the codebase, hey just pick your favorite thing to check (dict, collections.Mapping), oh and you need something else, just reinvent it. 

This is the reason we have the is_* things.

TomAugspurger:
Unfortunately, isinstance checks on ABCs are relatively slow, so I don't
think it should be used everywhere.

pandas/core/frame.py
gfyoung: Okay, sounds good.  I was also leaning towards expanding the dict-like check but wanted to get some consensus.

I do understand the appeal of the collections checking, but as we already have the inference utilities, we should use them in the interest of maintainability.

gfyoung:
@TomAugspurger : I actually commented about this earlier, but not seeing it here in the comments...weird...

I actually prefer to keep separate, in part because I suspect the use case in the wild  will largely focus on the columns, not the indices.  Thus, it makes sense to keep separate to facilitate this use case, while still making it possible to specify dtypes for the indices if so desired.

TomAugspurger:
Fair enough.

gfyoung:
@jreback @TomAugspurger : Managed to expand `is_dict_like` without issues, and I added a test in `to_records` to ensure that generic dict-like's are accepted.  PTAL.

TomAugspurger (approved):

pandas/core/dtypes/inference.py
```diff
@@ -398,8 +398,11 @@ def is_dict_like(obj):
     >>> is_dict_like([1, 2, 3])
     False
     """
+    for attr in ("__getitem__", "keys", "__contains__"):
```
jreback: not a big deal, but can you expand the is_dict_like test to account for this?

gfyoung: Sure thing.  I’ll do that and merge afterwards.

gfyoung: Sure thing.  I’ll do that and merge afterwards.

gfyoung: Sure thing.  I’ll do that and merge afterwards.

gfyoung:
@jreback : Added a test for the expanded `is_dict_like`, which is all green.  PTAL.

jreback (approved):

pandas/core/frame.py
```diff
@@ -1620,20 +1652,55 @@ def to_records(self, index=True, convert_datetime64=None):
...
             elif index_names[0] is None:
                 index_names = ['index']
+
             names = (lmap(compat.text_type, index_names) +
                      lmap(compat.text_type, self.columns))
         else:
             arrays = [self[c].get_values() for c in self.columns]
             names = lmap(compat.text_type, self.columns)
+            index_names = []
+
```
jreback: can you sprinkle a couple of comments in this to indicate what the flow is.

pandas/tests/dtypes/test_inference.py
```diff
@@ -178,6 +178,33 @@ def test_is_dict_like_fails(ll):
     assert not inference.is_dict_like(ll)
 
 
+@pytest.mark.parametrize("has_keys", [True, False])
+@pytest.mark.parametrize("has_getitem", [True, False])
+@pytest.mark.parametrize("has_contains", [True, False])
+def test_is_dict_like_duct_type(has_keys, has_getitem, has_contains):
```
jreback: duct->duck

jreback (approved):
@gfyoung lgtm. just a couple of minor things. merge on green.

pandas/core/frame.py
gfyoung: Yep, done.

pandas/tests/dtypes/test_inference.py
gfyoung: Good catch 🙂

gfyoung:
Thanks for the help, @qinghao1 !

PR #22235, opened 2018-08-07
Assorted _libs cleanups

jbrockmendel:
Modernize for-loop syntax
Modernize string formatting
Use memoryviews in a few places
Docstrings, change a couple of funcs in tslibs.parsing to cdef

Follow-up cleanup in tslibs.period
Move some unneeded stuff out of tslibs.util to wrap up the self-contained goal.


pandas/_libs/tslibs/period.pyx
```diff
@@ -163,6 +111,31 @@ cdef extern from *:
...
 
 
+cdef int max_value(int left, int right) nogil:
+    if left > right:
+        return left
+    return right
+
+
+cdef int min_value(int left, int right) nogil:
+    if left < right:
```
jreback: grr how many times do we have this i the code base :)

pandas/_libs/tslibs/period.pyx
```diff
@@ -163,6 +111,31 @@ cdef extern from *:
...
+    return right
+
+
+cdef int64_t get_daytime_conversion_factor(int from_index, int to_index) nogil:
+    cdef:
+        int row = min_value(from_index, to_index)
+        int col = max_value(from_index, to_index)
+    # row or col < 6 means frequency strictly lower than Daily, which
+    # do not use daytime_conversion_factors
+    if row < 6:
```
jreback: can u use the enumerated here?

pandas/_libs/tslibs/period.pyx
jbrockmendel: We could probably use np_datetime.cmp_scalar here, would need to cast from int to int64_t

pandas/_libs/tslibs/period.pyx
jbrockmendel: huh?  This is just moving the existing C code into cython.

jreback: I get it, but the 6 is pretty opaque

pandas/_libs/src/skiplist.h
```diff
@@ -20,18 +20,7 @@ Python recipe (http://rhettinger.wordpress.com/2010/02/06/lost-knowledge/)
...
-#define PANDAS_INLINE static __inline__
-#elif defined(_MSC_VER)
-#define PANDAS_INLINE static __inline
-#elif defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L
-#define PANDAS_INLINE static inline
-#else
-#define PANDAS_INLINE
-#endif
-#endif
+#include "helper.h"
```
jreback: can we rename helper to something more informative

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -1103,7 +1102,7 @@ class Timestamp(_Timestamp):
 
 
 # Add the min and max fields at the class level
-cdef int64_t _NS_UPPER_BOUND = INT64_MAX
```
jreback: future TODO, we look up these limits all over the place. maybe make a ``limits.pyx`` somewhere

jreback (changes requested):
can you start a cython section in internals.rst adding best practices

also I think a lint rule for not using ``0 <= j < n`` might be appropriate

jreback (approved):

jreback:
lgtm. merge on green.

jbrockmendel:
@pandas-dev/pandas-core do we have any windows users who can help troubleshoot appveyor-only build failures?

gfyoung:
@jbrockmendel : I wasn't able to reproduce the AppVeyor failure unfortunately.  Very hard to believe that the removal of the profiler magic comment is the cause of the issue.

jbrockmendel:
>  I wasn't able to reproduce the AppVeyor failure unfortunately. 

Darn, thanks for trying.

> But I have to ask: what happens if you put the magic comment back?

Its worth a try, but I expect the issue isn't with resolution.pyx but in one of its dependencies.

chris-b1:
I can reproduce - likely dependent on the 2.7 toolchain - will see what I can figure out

chris-b1:
Made a pull on your PR - long story short, it appears you were depending in a couple spots on an implicit include of `stdint.h`, which on MSCV doesn't define `int64_t`, so need a couple workarounds.

jreback:
thanks @chris-b1 

ok lgtm, merge on green (is the appveyor tests sufficient for this?)

gfyoung:
@jreback : The AppVeyor test should be sufficient since it was a build issue.  Also, I would have been much more concerned had it occurred on Python 3.x (instead of 2.x in this case).

gfyoung:
Thanks @jbrockmendel !

PR #21899, opened 2018-07-13
TST/CLN: series.duplicated; parametrisation; fix warning

h-vetinari:
Splitting up #21645

Added tests for `duplicated`, parametrized two tests for `drop_duplicates`, fixed a warning from #21614.

PR #22236, opened 2018-08-07
TST/CLN: break up & parametrize tests for df.set_index

h-vetinari:
While working on #22225 I had the strong urge to clean up the tests for `df.set_index` (to be able to work off of them for `series.set_index`)

Since the diff is pretty busted, here's a description. In `tests/frame/test_alter_axes.py`, I did:
- externalised an often-used df as a fixture
- broke up `test_set_index2` into several pieces
- renamed `test_set_index_bug` to `test_set_index_after_mutation` (following corresponding GH issue)
- renamed `test_set_index_nonuniq` to `test_set_index_verify_integrity` (also added a MI-case)
- strongly parametrized `test_set_index_pass_arrays` and `test_set_index_duplicate_names`, including several combinations and cases that were not tested before
- replaced everything `pd.` with direct imports (best practice according to review in #21899):
  > don’t import pd, directly import instead
- replaced `assert_series_equal` etc. with `tm.assert_series_equal` (best practice according to review in #21899):
  > use tm; don’t import assert_series_equal
- cleaned up the other tests a bit too, including several open TODOs and getting rid of the `check_names=False` in several cases.

I also added better warnings in case there are duplicate column labels in `keys`, and some corresponding tests.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3893,13 +3893,16 @@ class DataFrame(NDFrame):
         if not isinstance(keys, list):
             keys = [keys]
 
+        # collect elements from "keys" that are not allowed array types
         col_labels = [x for x in keys
                       if not isinstance(x, (Series, Index, MultiIndex,
                                             list, np.ndarray))]
         if any(x not in self for x in col_labels):
+            # if there are any invalid labels for self, we raise a KeyError
             missing = [x for x in col_labels if x not in self]
             raise KeyError('{}'.format(missing))
         elif len(set(col_labels)) < len(col_labels):
+            # if all are valid labels, but there are duplicates
             dup = Series(col_labels)
             dup = list(dup.loc[dup.duplicated()])
             raise ValueError('Passed duplicate column names '
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3898,7 +3898,7 @@ class DataFrame(NDFrame):
                       if not isinstance(x, (Series, Index, MultiIndex,
                                             list, np.ndarray))]
         if any(x not in self for x in col_labels):
-            # if there are any invalid labels for self, we raise a KeyError
+            # if there are any labels that are invalid, we raise a KeyError
             missing = [x for x in col_labels if x not in self]
             raise KeyError('{}'.format(missing))
         elif len(set(col_labels)) < len(col_labels):
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3888,9 +3888,24 @@ class DataFrame(NDFrame):
         -------
         dataframe : DataFrame
         """
-        inplace = validate_bool_kwarg(inplace, 'inplace')
+        from pandas import Series
+
         if not isinstance(keys, list):
             keys = [keys]
+
+        col_labels = [x for x in keys
+                      if not isinstance(x, (Series, Index, MultiIndex,
+                                            list, np.ndarray))]
+        if any(x not in self for x in col_labels):
+            missing = [x for x in col_labels if x not in self]
+            raise KeyError('{}'.format(missing))
+        elif len(set(col_labels)) < len(col_labels):
+            dup = Series(col_labels)
+            dup = list(dup.loc[dup.duplicated()])
+            raise ValueError('Passed duplicate column names '
+                             'to keys: {dup}'.format(dup=dup))
+
+        inplace = validate_bool_kwarg(inplace, 'inplace')
 
         if inplace:
             frame = self
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3888,28 +3888,9 @@ class DataFrame(NDFrame):
         -------
         dataframe : DataFrame
         """
-        from pandas import Series
-
+        inplace = validate_bool_kwarg(inplace, 'inplace')
         if not isinstance(keys, list):
             keys = [keys]
-
-        # collect elements from "keys" that are not allowed array types
-        col_labels = [x for x in keys
-                      if not isinstance(x, (Series, Index, MultiIndex,
-                                            list, np.ndarray))]
-        if any(x not in self for x in col_labels):
-            # if there are any labels that are invalid, we raise a KeyError
-            missing = [x for x in col_labels if x not in self]
-            raise KeyError('{}'.format(missing))
-
-        elif len(set(col_labels)) < len(col_labels):
-            # if all are valid labels, but there are duplicates
-            dup = Series(col_labels)
-            dup = list(dup.loc[dup.duplicated()])
-            raise ValueError('Passed duplicate column names '
-                             'to keys: {dup}'.format(dup=dup))
-
-        inplace = validate_bool_kwarg(inplace, 'inplace')
 
         if inplace:
             frame = self
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -3901,6 +3901,7 @@ class DataFrame(NDFrame):
             # if there are any labels that are invalid, we raise a KeyError
             missing = [x for x in col_labels if x not in self]
             raise KeyError('{}'.format(missing))
+
         elif len(set(col_labels)) < len(col_labels):
             # if all are valid labels, but there are duplicates
             dup = Series(col_labels)
--- a/pandas/tests/frame/common.py
+++ b/pandas/tests/frame/common.py
@@ -103,15 +103,6 @@ class TestData(object):
         return pd.DataFrame(arr, columns=['one', 'two', 'three'],
                             index=['a', 'b', 'c'])
 
-    @cache_readonly
-    def dummy(self):
-        df = pd.DataFrame({'A': ['foo', 'foo', 'foo', 'bar', 'bar'],
-                           'B': ['one', 'two', 'three', 'one', 'two'],
-                           'C': ['a', 'b', 'c', 'd', 'e'],
-                           'D': np.random.randn(5),
-                           'E': np.random.randn(5)})
-        return df
-
 # self.ts3 = tm.makeTimeSeries()[-5:]
 # self.ts4 = tm.makeTimeSeries()[1:-1]
 
--- a/pandas/tests/frame/common.py
+++ b/pandas/tests/frame/common.py
@@ -103,6 +103,15 @@ class TestData(object):
         return pd.DataFrame(arr, columns=['one', 'two', 'three'],
                             index=['a', 'b', 'c'])
 
+    @cache_readonly
+    def dummy(self):
+        df = pd.DataFrame({'A': ['foo', 'foo', 'foo', 'bar', 'bar'],
+                           'B': ['one', 'two', 'three', 'one', 'two'],
+                           'C': ['a', 'b', 'c', 'd', 'e'],
+                           'D': np.random.randn(5),
+                           'E': np.random.randn(5)})
+        return df
+
 # self.ts3 = tm.makeTimeSeries()[-5:]
 # self.ts4 = tm.makeTimeSeries()[1:-1]
 
--- a/pandas/tests/frame/conftest.py
+++ b/pandas/tests/frame/conftest.py
@@ -1,0 +1,121 @@
+import pytest
+
+import numpy as np
+
+from pandas import compat
+import pandas.util.testing as tm
+from pandas import DataFrame, date_range, NaT
+
+
+@pytest.fixture
+def frame():
+    return DataFrame(tm.getSeriesData())
+
+
+@pytest.fixture
+def frame2():
+    return DataFrame(tm.getSeriesData(), columns=['D', 'C', 'B', 'A'])
+
+
+@pytest.fixture
+def intframe():
+    df = DataFrame({k: v.astype(int)
+                   for k, v in compat.iteritems(tm.getSeriesData())})
+    # force these all to int64 to avoid platform testing issues
+    return DataFrame({c: s for c, s in compat.iteritems(df)}, dtype=np.int64)
+
+
+@pytest.fixture
+def tsframe():
+    return DataFrame(tm.getTimeSeriesData())
+
+
+@pytest.fixture
+def mixed_frame():
+    df = DataFrame(tm.getSeriesData())
+    df['foo'] = 'bar'
+    return df
+
+
+@pytest.fixture
+def mixed_float():
+    df = DataFrame(tm.getSeriesData())
+    df.A = df.A.astype('float16')
+    df.B = df.B.astype('float32')
+    df.C = df.C.astype('float64')
+    return df
+
+
+@pytest.fixture
+def mixed_float2():
+    df = DataFrame(tm.getSeriesData())
+    df.D = df.D.astype('float16')
+    df.C = df.C.astype('float32')
+    df.B = df.B.astype('float64')
+    return df
+
+
+@pytest.fixture
+def mixed_int():
+    df = DataFrame({k: v.astype(int)
+                   for k, v in compat.iteritems(tm.getSeriesData())})
+    df.A = df.A.astype('uint8')
+    df.B = df.B.astype('int32')
+    df.C = df.C.astype('int64')
+    df.D = np.ones(len(df.D), dtype='uint64')
+    return df
+
+
+@pytest.fixture
+def all_mixed():
+    return DataFrame({'a': 1., 'b': 2, 'c': 'foo',
+                      'float32': np.array([1.] * 10, dtype='float32'),
+                      'int32': np.array([1] * 10, dtype='int32')},
+                     index=np.arange(10))
+
+
+@pytest.fixture
+def tzframe():
+    df = DataFrame({'A': date_range('20130101', periods=3),
+                    'B': date_range('20130101', periods=3,
+                                    tz='US/Eastern'),
+                    'C': date_range('20130101', periods=3,
+                                    tz='CET')})
+    df.iloc[1, 1] = NaT
+    df.iloc[1, 2] = NaT
+    return df
+
+
+@pytest.fixture
+def empty():
+    return DataFrame({})
+
+
+@pytest.fixture
+def ts1():
+    return tm.makeTimeSeries(nper=30)
+
+
+@pytest.fixture
+def ts2():
+    return tm.makeTimeSeries(nper=30)[5:]
+
+
+@pytest.fixture
+def simple():
+    arr = np.array([[1., 2., 3.],
+                    [4., 5., 6.],
+                    [7., 8., 9.]])
+
+    return DataFrame(arr, columns=['one', 'two', 'three'],
+                     index=['a', 'b', 'c'])
+
+
+@pytest.fixture
+def frame_of_index_cols():
+    df = DataFrame({'A': ['foo', 'foo', 'foo', 'bar', 'bar'],
+                    'B': ['one', 'two', 'three', 'one', 'two'],
+                    'C': ['a', 'b', 'c', 'd', 'e'],
+                    'D': np.random.randn(5),
+                    'E': np.random.randn(5)})
+    return df
--- a/pandas/tests/frame/conftest.py
+++ b/pandas/tests/frame/conftest.py
@@ -7,46 +7,32 @@ import pandas.util.testing as tm
 from pandas import DataFrame, date_range, NaT
 
 
-# This module is the start of transitioning from attributes of
-# pandas/tests/frame/common.TestData towards fixtures (GH22471).
-# Until all modules have been transitioned, it is advised not to change
-# the (admittedly suboptimal) names of these fixtures.
-
 @pytest.fixture
-def frame():
+def float_frame():
     """
     Fixture for DataFrame of floats with index of unique strings
 
-    After completing the fixturization of the frame tests (GH 22471), this
-    fixture will be renamed to: float_frame_string_index
-
-    Columns are ['A', 'B', 'C', 'D'], see pandas.util.testing.getSeriesData
+    Columns are ['A', 'B', 'C', 'D'].
     """
     return DataFrame(tm.getSeriesData())
 
 
 @pytest.fixture
-def frame2():
+def float_frame2():
     """
     Fixture for DataFrame of floats with index of unique strings
 
-    After completing the fixturization of the frame tests (GH 22471), this
-    fixture will be renamed to: float_frame_string_index2
-
-    Columns are ['D', 'C', 'B', 'A']. See pandas.util.testing.getSeriesData
+    Columns are ['D', 'C', 'B', 'A']
     """
     return DataFrame(tm.getSeriesData(), columns=['D', 'C', 'B', 'A'])
 
 
 @pytest.fixture
-def intframe():
+def int_frame():
     """
     Fixture for DataFrame of ints with index of unique strings
 
-    After completing the fixturization of the frame tests (GH 22471), this
-    fixture will be renamed to: int_frame
-
-    Columns are ['A', 'B', 'C', 'D']. See pandas.util.testing.getSeriesData
+    Columns are ['A', 'B', 'C', 'D']
     """
     df = DataFrame({k: v.astype(int)
                    for k, v in compat.iteritems(tm.getSeriesData())})
@@ -55,28 +41,21 @@ def intframe():
 
 
 @pytest.fixture
-def tsframe():
+def datetime_frame():
     """
     Fixture for DataFrame of floats with DatetimeIndex
 
-    After completing the fixturization of the frame tests (GH 22471), this
-    fixture will be renamed to: float_frame_datetime_index
-
-    Columns are ['A', 'B', 'C', 'D']. See pandas.util.testing.getTimeSeriesData
+    Columns are ['A', 'B', 'C', 'D']
     """
     return DataFrame(tm.getTimeSeriesData())
 
 
 @pytest.fixture
-def mixed_frame():
+def float_string_frame():
     """
-    Fixture for DataFrame of floats and strings with string index
-
-    After completing the fixturization of the frame tests (GH 22471), this
-    fixture will be renamed to: float_string_frame
+    Fixture for DataFrame of floats and strings with index of unique strings
 
     Columns are ['A', 'B', 'C', 'D', 'foo'].
-    See pandas.util.testing.getSeriesData
     """
     df = DataFrame(tm.getSeriesData())
     df['foo'] = 'bar'
@@ -84,14 +63,11 @@ def mixed_frame():
 
 
 @pytest.fixture
-def mixed_float():
+def mixed_float_frame():
     """
-    Fixture for DataFrame of different float types with string index
+    Fixture for DataFrame of different float types with index of unique strings
 
-    After completing the fixturization of the frame tests (GH 22471), this
-    fixture will be renamed to: mixed_float_frame
-
-    Columns are ['A', 'B', 'C', 'D']. See pandas.util.testing.getSeriesData
+    Columns are ['A', 'B', 'C', 'D'].
     """
     df = DataFrame(tm.getSeriesData())
     df.A = df.A.astype('float16')
@@ -101,14 +77,11 @@ def mixed_float():
 
 
 @pytest.fixture
-def mixed_float2():
+def mixed_float_frame2():
     """
-    Fixture for DataFrame of different float types with string index
+    Fixture for DataFrame of different float types with index of unique strings
 
-    After completing the fixturization of the frame tests (GH 22471), this
-    fixture will be renamed to: mixed_float_frame2
-
-    Columns are ['A', 'B', 'C', 'D']. See pandas.util.testing.getSeriesData
+    Columns are ['A', 'B', 'C', 'D'].
     """
     df = DataFrame(tm.getSeriesData())
     df.D = df.D.astype('float16')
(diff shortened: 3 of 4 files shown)


jreback (changes requested):

pandas/core/frame.py
```diff
@@ -3973,10 +3973,25 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
         -------
         dataframe : DataFrame
         """
-        inplace = validate_bool_kwarg(inplace, 'inplace')
+        from pandas import Series
+
         if not isinstance(keys, list):
             keys = [keys]
 
+        col_labels = [x for x in keys
```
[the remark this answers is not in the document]
h-vetinari: I wrote that:
> I also added better warnings in case there are duplicate column labels in keys, and some corresponding tests.

Currently, `df.set_index(['A', 'A'])` yields something like:
```ValueError: Duplicated level name: "A", assigned to level 1, is already used for level 0.```
which was
- untested
- not a very clear warning
- bad for perf, because all arrays get processed already and only upon creation of the `MultiIndex` is the error raised.

I fixed that with the code block above ('better warnings" is also in the title).

File: pandas/tests/frame/common.py [hunk 2]
```diff
@@ -103,6 +103,15 @@ def simple(self):
         return pd.DataFrame(arr, columns=['one', 'two', 'three'],
                             index=['a', 'b', 'c'])
 
+    @cache_readonly
```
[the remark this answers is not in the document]
h-vetinari: Not 100% sure, but don't the attributes of `TestData` have the roles of fixtures for class-based tests?

pandas/tests/frame/test_alter_axes.py
```diff
@@ -10,262 +10,288 @@
...
     is_interval_dtype)
-import pandas as pd
-
-from pandas.util.testing import assert_series_equal, assert_frame_equal
 
 import pandas.util.testing as tm
 
 from pandas.tests.frame.common import TestData
 
+key = lambda x: x.name
```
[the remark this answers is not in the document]
h-vetinari: they are "container"-types for testing, in the sense that -- applied to `df['A']` -- they give the correct type of container. They are to be thought of like `Series` or `Index`, but since the bare `MultiIndex()`- constructor does not take a vector, I wrote it as a lambda. Makes testing all allowed cases nicely parametrisable.

pandas/tests/frame/test_alter_axes.py
```diff
@@ -10,262 +10,288 @@
...
+mi = lambda x: MultiIndex.from_arrays([x])
+
 
 class TestDataFrameAlterAxes(TestData):
 
+    def test_set_index_manually(self):
```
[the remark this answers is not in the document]
h-vetinari: I really think `df.index = idx` deserves the name "manual". It should also be discouraged, that's why there's `set_index` in the first place. `df.set_index(idx)` is then the API-supported (i.e. "non-manual") way of doing it

pandas/tests/frame/test_alter_axes.py
```diff
@@ -10,262 +10,288 @@
...
+        df = self.mixed_frame.copy()
```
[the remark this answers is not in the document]
h-vetinari: This is a class-based test module. `mixed_frame` already existed before, and has - for all intents and purposes - the role of a fixture.

pandas/tests/frame/test_alter_axes.py
```diff
@@ -10,262 +10,288 @@
...
-        index = MultiIndex.from_arrays([df['A'], df['B']], names=['A', 'B'])
+        df2 = df.set_index(df.index.astype(np.int32))
+        tm.assert_frame_equal(df, df2)
+
+    # A has duplicate values, C does not
+    @pytest.mark.parametrize('keys', ['A', 'C', ['A', 'B']])
+    @pytest.mark.parametrize('inplace', [True, False])
+    @pytest.mark.parametrize('drop', [True, False])
+    def test_set_index_drop_inplace(self, drop, inplace, keys):
+        df = self.dummy.copy()
```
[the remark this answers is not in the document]
h-vetinari: See above.

pandas/tests/frame/test_alter_axes.py
```diff
@@ -10,262 +10,288 @@
...
+                             Index(['b', 'c'], name='key'))
 
-        df3 = df.copy()
-        df3.set_index(['A', 'B'], drop=False, inplace=True)
-        assert_frame_equal(df3, expected_nodrop)
+        df2 = df.loc[df.index.map(lambda indx: indx >= 1)]
+        result = df2.set_index('key')
+        tm.assert_frame_equal(result, expected)
+
+    @pytest.mark.parametrize('container', [Series, Index, np.array, mi])
```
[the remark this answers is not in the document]
h-vetinari: I think that "box" is less clear than "container", but OK

pandas/tests/frame/test_alter_axes.py
```diff
@@ -10,262 +10,288 @@
...
+                             (True, 'A'), (True, 'test'), (False, None)])
+    @pytest.mark.parametrize('drop', [True, False])
+    def test_set_index_pass_arrays_duplicate(self, drop, append, df_index_name,
+                                             elem1, elem2):
+        df = self.dummy.copy()
+        df.index.name = df_index_name
+
+        keys = [elem1(df['A']), elem2(df['A'])]
+
+        # == gives ambiguous Boolean for Series
```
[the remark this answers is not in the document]
h-vetinari: I said
> strongly parametrized test_set_index_pass_arrays and test_set_index_duplicate_names, including several combinations and cases that were not tested before

This is (essentially) a very beefed-up version of `test_set_index_duplicate_names`. It test appending duplicate arrays in various forms (and with various kwargs, e.g. against `drop`), and tests for the error of passing duplicate column keys directly.

pandas/tests/frame/test_alter_axes.py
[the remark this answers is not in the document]
h-vetinari: Nevermind, how do you like "directly"? ;-)

pandas/tests/frame/common.py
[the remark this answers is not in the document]
jreback: no that is a very bad pattern. we don't want to do that. pls use fixtures instead

pandas/tests/frame/test_alter_axes.py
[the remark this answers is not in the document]
jreback: this is obtuse, I can't tell what you are doing. if you need something create it where you need it.

pandas/tests/frame/test_alter_axes.py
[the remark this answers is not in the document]
jreback: see above as well

jreback (changes requested):

pandas/tests/frame/common.py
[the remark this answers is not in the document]
h-vetinari: I don't think it's fair to blame me for following an existing pattern - just look at

```
    @cache_readonly
    def simple(self):
        arr = np.array([[1., 2., 3.],
                        [4., 5., 6.],
                        [7., 8., 9.]])

        return pd.DataFrame(arr, columns=['one', 'two', 'three'],
                            index=['a', 'b', 'c'])
```

directly above.

Do you want me to change all of `test_alter_axes` from classes to functions? That's the only (clean) way I can think of.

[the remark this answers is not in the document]
h-vetinari: I don't think it's fair to blame me for following an existing pattern - just look at

```
    @cache_readonly
    def simple(self):
        arr = np.array([[1., 2., 3.],
                        [4., 5., 6.],
                        [7., 8., 9.]])

        return pd.DataFrame(arr, columns=['one', 'two', 'three'],
                            index=['a', 'b', 'c'])
```

directly above.

Do you want me to change all of `test_alter_axes` from classes to functions? That's the only (clean) way I can think of.

pandas/tests/frame/test_alter_axes.py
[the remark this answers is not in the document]
h-vetinari: Refactored. Hope you like it better like this - the advantage is that it stays local and that the generated test names are understandable

pandas/tests/frame/test_alter_axes.py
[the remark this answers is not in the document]
h-vetinari: Refactored. Hope you like it better like this - the advantage is that it stays local and that the generated test names are understandable

h-vetinari:
@jreback Don't know where my comments went, they're not showing here at least - copying them to the thread.

> no that is a very bad pattern. we don't want to do that. pls use fixtures instead

I don't think it's fair to blame me for following an existing pattern - just look at

```
    @cache_readonly
    def simple(self):
        arr = np.array([[1., 2., 3.],
                        [4., 5., 6.],
                        [7., 8., 9.]])

        return pd.DataFrame(arr, columns=['one', 'two', 'three'],
                            index=['a', 'b', 'c'])
```
directly above (https://github.com/pandas-dev/pandas/blob/v0.23.4/pandas/tests/frame/common.py#L97-104)

Do you want me to change all of test_alter_axes from classes to functions? That's the only (clean) way I can think of.

> this is obtuse, I can't tell what you are doing. if you need something create it where you need it.

Refactored. Hope you like it better like this - the advantage is that it stays local and that the generated test names are understandable

pandas/core/frame.py
```diff
@@ -3973,10 +3973,25 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
+                      if not isinstance(x, (Series, Index, MultiIndex,
+                                            list, np.ndarray))]
+        if any(x not in self for x in col_labels):
+            missing = [x for x in col_labels if x not in self]
```
jreback: are there explict tests for each of the branches. pls add a comment to each one indicating what you are checking.

pandas/tests/frame/common.py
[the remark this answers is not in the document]
jreback: not blaming you, asking you to use fixtures.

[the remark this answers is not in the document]
jreback: that is orthogonal to switching from classes.

pandas/tests/frame/test_alter_axes.py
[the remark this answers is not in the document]
jreback: yes this is better (still pls make dummy a fixture)

pandas/tests/frame/test_alter_axes.py
```diff
@@ -28,244 +25,287 @@
...
 
-        df3 = df.copy()
-        df3.set_index(['A', 'B'], drop=False, inplace=True)
-        assert_frame_equal(df3, expected_nodrop)
+        df2 = df.loc[df.index.map(lambda indx: indx >= 1)]
+        result = df2.set_index('key')
+        tm.assert_frame_equal(result, expected)
+
+    # also test index name if append=True (name is duplicate here for B)
+    @pytest.mark.parametrize('box', [Series, Index, np.array, 'MultiIndex'])
```
jreback: dont' use a string, just directly put the lambda here. ideally we have NO if/else in test functions, sometimes its unavoidable, but making tests as simple as possible is the key

pandas/tests/frame/test_alter_axes.py
```diff
@@ -28,244 +25,287 @@
...
+                                     list, 'MultiIndex'])
+    @pytest.mark.parametrize('append, index_name',
+                             [(True, None), (True, 'A'), (True, 'B'),
+                              (True, 'test'), (False, None)])
+    @pytest.mark.parametrize('drop', [True, False])
+    def test_set_index_pass_arrays(self, drop, append, index_name, box):
+        df = self.dummy.copy()
+        df.index.name = index_name
+
+        # update constructor in case of MultiIndex
```
jreback: same comment

pandas/tests/frame/test_alter_axes.py
```diff
@@ -28,244 +25,287 @@
...
+    @pytest.mark.parametrize('append, index_name', [(True, None),
+                             (True, 'A'), (True, 'test'), (False, None)])
+    @pytest.mark.parametrize('drop', [True, False])
+    def test_set_index_pass_arrays_duplicate(self, drop, append, index_name,
+                                             box1, box2):
+        df = self.dummy.copy()
+        df.index.name = index_name
+
+        # transform strings to correct box constructor
+        def rebox(x):
```
jreback: this is not good at all, pls don't do this inside the test function, just use a lambda in the box itself

pandas/tests/frame/test_alter_axes.py
```diff
@@ -28,244 +25,287 @@
...
+        df['B'] = idx
         result = df['B']
-        assert_series_equal(result, expected, check_names=False)
-        assert result.name == 'B'
+        tm.assert_series_equal(result, expected)
 
-        # keep the timezone
-        result = i.to_series(keep_tz=True)
-        assert_series_equal(result.reset_index(drop=True), expected)
+        # convert to series while keeping the timezone
```
jreback: why did this change?

jreback (changes requested):

pandas/tests/frame/common.py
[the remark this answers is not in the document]
h-vetinari: Ok, got it

pandas/tests/frame/test_alter_axes.py
h-vetinari: The whole test is about converting a DatetimeIndex to a Series, so I renamed the test as such.
I renamed `i` to `idx` for better readability.

Finally, `idx.to_series(keep_tz=True)` yields:
```
B
2013-01-01 13:00:00-08:00   2013-01-01 13:00:00-08:00
2013-01-02 14:00:00-08:00   2013-01-02 14:00:00-08:00
Name: B, dtype: datetime64[ns, US/Pacific]
```
so needs an index change to fit with `expected`. I just found that using the `index`-kwarg of `to_series` is cleaner and more understandable than using `reset_index` - and since it's the *conversion* that's being tested and not the manner of the index reset, I changed it.

h-vetinari (commented):
Thanks for the detailed review. Should be getting close now.

pandas/core/frame.py
```diff
@@ -3973,10 +3973,28 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
+        col_labels = [x for x in keys
+                      if not isinstance(x, (Series, Index, MultiIndex,
+                                            list, np.ndarray))]
+        if any(x not in self for x in col_labels):
+            # if there are any labels that are invalid, we raise a KeyError
+            missing = [x for x in col_labels if x not in self]
+            raise KeyError('{}'.format(missing))
+        elif len(set(col_labels)) < len(col_labels):
+            # if all are valid labels, but there are duplicates
+            dup = Series(col_labels)
```
jreback: rather use a set difference operation here

pandas/tests/frame/test_alter_axes.py
```diff
@@ -10,262 +10,309 @@
...
-
-from pandas.util.testing import assert_series_equal, assert_frame_equal
 
 import pandas.util.testing as tm
 
 from pandas.tests.frame.common import TestData
 
 
+@pytest.fixture
+def frame_of_index_cols():
```
jreback: pls move to ``pandas/tests/frame/conftest.py`` need to start this.

pandas/tests/frame/test_alter_axes.py
```diff
@@ -10,262 +10,309 @@
...
+                    'C': ['a', 'b', 'c', 'd', 'e'],
+                    'D': np.random.randn(5),
+                    'E': np.random.randn(5)})
+    return df
+
+
 class TestDataFrameAlterAxes(TestData):
 
+    def test_set_index_directly(self):
+        df = self.mixed_frame.copy()
```
jreback: add this to the conftest as well as a fixture

jreback (changes requested):
ok looks closer a few more comments

pandas/core/frame.py
h-vetinari: maybe I'm not seeing it, but set difference isn't gonna show duplicates (because as a set, they'll be the same)...

h-vetinari:
@jreback 

I translated all the "fixture"-attributes from `TestData` to `conftest.py`, but I only replaced them in `test_alter_axes`, because they occur just too frequently to do it in one go. ;-)

h-vetinari:
@jreback All green, and should be good to go.

h-vetinari:
@jreback Another ping :)

jreback:
i will look soon
we have gotten a ton of PRs lately

pandas/core/frame.py
```diff
@@ -3973,10 +3973,28 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
         """
-        inplace = validate_bool_kwarg(inplace, 'inplace')
+        from pandas import Series
+
         if not isinstance(keys, list):
             keys = [keys]
 
+        # collect elements from "keys" that are not allowed array types
+        col_labels = [x for x in keys
+                      if not isinstance(x, (Series, Index, MultiIndex,
```
jreback: use the ABC forms here

jreback: why isn't this just ``isinstance(x, tuple) or is_scalar(x)``?

pandas/core/frame.py
```diff
@@ -3973,10 +3973,28 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
+                                            list, np.ndarray))]
+        if any(x not in self for x in col_labels):
+            # if there are any labels that are invalid, we raise a KeyError
+            missing = [x for x in col_labels if x not in self]
+            raise KeyError('{}'.format(missing))
+        elif len(set(col_labels)) < len(col_labels):
```
jreback: blank line here

pandas/core/frame.py
jreback: subtract the sets of all - others

pandas/tests/frame/test_alter_axes.py
```diff
@@ -28,244 +25,284 @@
 
 class TestDataFrameAlterAxes(TestData):
```
jreback: you shouldn't need TestData any longer

jreback (changes requested):
ok, pls open a new issue that refs this, to remove use of TestData in favor of fixtures

pandas/core/frame.py
h-vetinari: These are the cases (except valid column keys)  that currently *don't* raise a KeyError. I've kept the error reporting exactly the same as before:
```
pd.DataFrame([[0,1], [2,3]]).set_index(map(str,[1,2]))
# KeyError: <map object at 0x000001D380115550>
```

Could be changed, but then I need to know the desired allowed signature / error reporting

pandas/core/frame.py
h-vetinari: this is not about the set of all inputs - it's strictly about column labels. E.g.
```
df.set_index(['A', 'A', 'B', df.A, some_series, some_index])
[...]
col_labels = ['A', 'A', 'B']  # at the line of your comment (in this case)
```
No set operation that I can think of would yield `A`, which is what I want to raise. (multisets would work, but that's hardly more intuitive than `duplicated`, IMO)

pandas/core/frame.py
h-vetinari: Tried the ABC Versions, but getting errors that indexes are not hashable. Left it as is for the moment.
```
____________________ TestDataFrameAlterAxes.test_set_index ____________________

self = <pandas.tests.frame.test_alter_axes.TestDataFrameAlterAxes object at 0x0000018FC05ED400>
mixed_frame =                    A         B         C         D  foo
lem22AYh2f -0.599136 -0.811715 -0.135242  0.764100  bar
iBogSI...05  bar
vDdWVTLtoj -0.433490 -1.416721  0.460720 -0.539860  bar
Pw5bUHt4sR  0.180016  1.651358 -1.041539 -0.832112  bar

    def test_set_index(self, mixed_frame):
        df = mixed_frame
        idx = Index(np.arange(len(df))[::-1])

>       df = df.set_index(idx)

pandas\tests\frame\test_alter_axes.py:41:
pandas\core\frame.py:3874: in set_index
    if any(x not in self for x in col_labels):
pandas\core\frame.py:3874: in <genexpr>
    if any(x not in self for x in col_labels):
pandas\core\generic.py:1654: in __contains__
    return key in self._info_axis
pandas\core\indexes\base.py:1994: in __contains__
    hash(key)

self = Int64Index([29, 28, 27, 26, 25, 24, 23, 22, 21, 20, 19, 18, 17, 16, 15, 14, 13,
            12, 11, 10,  9,  8,  7,  6,  5,  4,  3,  2,  1,  0],
           dtype='int64')

    def __hash__(self):
>       raise TypeError("unhashable type: %r" % type(self).__name__)
E       TypeError: unhashable type: 'Int64Index'

pandas\core\indexes\base.py:2021: TypeError
_________________ TestDataFrameAlterAxes.test_set_index_cast __________________

self = <pandas.tests.frame.test_alter_axes.TestDataFrameAlterAxes object at 0x0000018FCA8EE668>

    def test_set_index_cast(self):
        # issue casting an index then set_index
        df = DataFrame({'A': [1.1, 2.2, 3.3], 'B': [5.0, 6.1, 7.2]},
                       index=[2010, 2011, 2012])
>       df2 = df.set_index(df.index.astype(np.int32))

pandas\tests\frame\test_alter_axes.py:50:
pandas\core\frame.py:3874: in set_index
    if any(x not in self for x in col_labels):
pandas\core\frame.py:3874: in <genexpr>
    if any(x not in self for x in col_labels):
pandas\core\generic.py:1654: in __contains__
    return key in self._info_axis
pandas\core\indexes\base.py:1994: in __contains__
    hash(key)

self = Int64Index([2010, 2011, 2012], dtype='int64')

    def __hash__(self):
>       raise TypeError("unhashable type: %r" % type(self).__name__)
E       TypeError: unhashable type: 'Int64Index'

pandas\core\indexes\base.py:2021: TypeError
_______________ TestDataFrameAlterAxes.test_set_index_timezone ________________

self = <pandas.tests.frame.test_alter_axes.TestDataFrameAlterAxes object at 0x0000018FCAD1AEF0>

    def test_set_index_timezone(self):
        # GH 12358
        # tz-aware Series should retain the tz
        idx = to_datetime(["2014-01-01 10:10:10"],
                          utc=True).tz_convert('Europe/Rome')
        df = DataFrame({'A': idx})
>       assert df.set_index(idx).index[0].hour == 11

pandas\tests\frame\test_alter_axes.py:343:
pandas\core\frame.py:3874: in set_index
    if any(x not in self for x in col_labels):
pandas\core\frame.py:3874: in <genexpr>
    if any(x not in self for x in col_labels):
pandas\core\generic.py:1654: in __contains__
    return key in self._info_axis
pandas\core\indexes\base.py:1994: in __contains__
    hash(key)

self = DatetimeIndex(['2014-01-01 11:10:10+01:00'], dtype='datetime64[ns, Europe/Rome]', freq=None)

    def __hash__(self):
>       raise TypeError("unhashable type: %r" % type(self).__name__)
E       TypeError: unhashable type: 'DatetimeIndex'

pandas\core\indexes\base.py:2021: TypeError
______________ TestDataFrameAlterAxes.test_dti_set_index_reindex ______________

self = <pandas.tests.frame.test_alter_axes.TestDataFrameAlterAxes object at 0x0000018FCA7FDD68>

    def test_dti_set_index_reindex(self):
        # GH 6631
        df = DataFrame(np.random.random(6))
        idx1 = date_range('2011/01/01', periods=6, freq='M', tz='US/Eastern')
        idx2 = date_range('2013', periods=6, freq='A', tz='Asia/Tokyo')

>       df = df.set_index(idx1)

pandas\tests\frame\test_alter_axes.py:413:
pandas\core\frame.py:3874: in set_index
    if any(x not in self for x in col_labels):
pandas\core\frame.py:3874: in <genexpr>
    if any(x not in self for x in col_labels):
pandas\core\generic.py:1654: in __contains__
    return key in self._info_axis
pandas\core\indexes\base.py:1994: in __contains__
    hash(key)

self = DatetimeIndex(['2011-01-31 00:00:00-05:00', '2011-02-28 00:00:00-05:00',
               '2011-03-31 00:00:00-04:00', '... '2011-05-31 00:00:00-04:00', '2011-06-30 00:00:00-04:00'],
              dtype='datetime64[ns, US/Eastern]', freq='M')

    def __hash__(self):
>       raise TypeError("unhashable type: %r" % type(self).__name__)
E       TypeError: unhashable type: 'DatetimeIndex'
```

h-vetinari (commented):
Thanks for the review; opened #22471 as requested

pandas/core/frame.py
jreback: huh? pls use the ABC version, we do this everywhere else. To be very honest I would split this PR up into 2 parts as I think you are actually changing something here but its very very hard to tell.

jreback (changes requested):

h-vetinari:
@jreback 
> huh? pls use the ABC version, we do this everywhere else. To be very honest I would split this PR up into 2 parts as I think you are actually changing something here but its very very hard to tell.

Fair enough, this was actually a good intuition. I'll split out all the new things into a new issue/PR (xref coming), and reverted all the new warnings here.

pandas/tests/frame/conftest.py
```diff
@@ -0,0 +1,121 @@
+import pytest
```
gfyoung: A couple of things about this file:

* Do you use all of these fixtures in your changes?
* I would prefer if the naming is a little more consistent e.g.:
- You have the word "frame" is some of your fixture names but not othres
- You have underscores between words in some names but not others

h-vetinari: For this PR, I turned an often-used DF into a fixture - together with the other attributes of `TestData`. @jreback then told me to start a `conftest.py` and put it there, together with the other `TestData`-attributes used in this module (`frame`, `mixed_frame`).

I translated all the attributes into `conftest.py` *without* renaming them, so that they can be replaced on a per-module-basis as laid out in #22471. The names of the fixtures are clearly suboptimal, but following up on fixturizing the other modules would be much harder if I start renaming now.

gfyoung: That's fair.  Let's save for a follow-up then.

h-vetinari:
@jreback 

> > huh? pls use the ABC version, we do this everywhere else. To be very honest I would split this PR up into 2 parts as I think you are actually changing something here but its very very hard to tell.

> Fair enough, this was actually a good intuition. I'll split out all the new things into a new issue/PR (#22484 #22486), and **reverted all the new warnings here**.

All green now.

h-vetinari:
@jreback 
Hope there's less open questions around this PR ;-)
Also blocking a couple follow-ups...

jreback:
you can rebase this off of #22550 if you want or rebase off of master after merge.

h-vetinari:
@jreback
> you can rebase this off of #22550 if you want or rebase off of master after merge.

The number you referenced is an issue, I guess you mean #22526? In any case, I can easily rebase afterwards, but there won't be a conflict because these `conftest` files live in different folders (`series` vs. `frame`).

h-vetinari:
@jreback 

#22526 is merged, and there's no conflict here.

pandas/tests/frame/conftest.py
```diff
@@ -0,0 +1,243 @@
...
+
+import numpy as np
+
+from pandas import compat
+import pandas.util.testing as tm
+from pandas import DataFrame, date_range, NaT
+
+
+# This module is the start of transitioning from attributes of
```
jreback: actually no, let's change them to something reasonable here the form, its easier to change them once and then adjust on file migration.

``int_frame``, ``ts_frame`` etc is reasonable, e.g. _ separate and they all have the word frame in them.

jreback (changes requested):

h-vetinari:
@jreback 

The only thing left should be if you like the fixture names I chose.

h-vetinari:
@jreback 

Had to rebase due to unrelated test failures (`ResourceWarning` again...), but there's only one new commit, and not much in it ```https://github.com/pandas-dev/pandas/pull/22236/commits/0802144b18a5436306a5c36843053808c1b77e09```.

Double-checking those fixture name choices should be a triviality - and then we can finally get this over the finish line. :)

pandas/tests/frame/conftest.py
h-vetinari: @gfyoung:

Since the names now changed after all, I thought I ask for your opinion on the naming/consistency.

gfyoung: @h-vetinari : This looks great!

h-vetinari: @gfyoung 
Thanks. I just added a few more consistency fixes (but squashed to simplify reviewing), but only the first three fixture names were affected.

h-vetinari:
@jreback 

Green.

> Double-checking those fixture name choices should be a triviality - and then we can finally get this over the finish line. :)

h-vetinari:
@jreback 
Can we get this over the line?
Things are getting tight for v0.24 with this -> #22486 -> #22225 (-> #21645)

jreback:
@h-vetinari i will look at some point

h-vetinari:
@jreback 
I truly understand that review time is scarce (and that you're probably getting annoyed at my pings here). Still, we've been at
> ok looks closer a few more comments

since over a month, stuck in the following pattern: push trivial fix, wait ~2 weeks for a comment, push trivial fix, wait ~2 weeks for a comment, push trivial fix, etc., and I can't understand why following through on this (within that timeframe) is such a hurdle. Renaming some (still unused!) fixtures is about as simple as it gets.

still applicable:
> I've been responding quickly to any and all reviews, and happy to make it smaller/bigger/split it, whatever - just wanna get this done ASAP.

jreback (approved):

jreback:
lgtm. can you rebase to make sure, ping on green for merge.

h-vetinari:
I've had a `ResourceWarning` twice, and then restarted a third time because AppVeyor isn't triggering for some reason. It didn't start a second time, so I've left it like that ATM.

h-vetinari:
@jreback Green (after 5 retriggers...)

jreback:
thanks. in the future, not necessary to push multiple times, just needed a single one, if the failure is unrelated.

PR #22163, opened 2018-08-02
dispatch scalar DataFrame ops to Series

jbrockmendel:
Many issues closed; will track them down and update.  Will also need whatsnew.

closes #18874
closes #20088
closes #15697
closes #13128
closes #8554
closes #8932
closes #21610 
closes #22005 
closes #22047
closes #22242

This will be less verbose after #22068 implements `ops.dispatch_to_series`.

This still only dispatches a subset of ops.  #22019 dispatches another (disjoint) subset.  After that is another easy-ish case where alignment is known.  Saved for last are cases with ambiguous alignment that is currently done in an ad-hoc best-guess way.

PR #22237, opened 2018-08-07
Collect datetime64 and PeriodDtype arithmetic tests

jbrockmendel:
As suggested, these are split by dtype in a new tests/arithmetic directory.

Started parametrizing a few of these over `box` classes, but for the most part this is just collecting tests from their existing locations.  There is going to be a lot of parametrization and de-duplication coming up.

Holding off on moving the existing timedelta64-dtype tests until #22163 is handled to avoid merge conflict.


jreback:
thanks.

thought was moving ``pandas/tests/test_arithmetic.py`` as well?

maybe ought to move some of the fixtures to the top-level conftest

jbrockmendel:
> thought was moving pandas/tests/test_arithmetic.py as well?

That's coming up, but I decided to wait on that until after #22163 to avoid merge conflicts

PR #22249, opened 2018-08-08
DOC: Fix dateutil link in install instructions

jarmar:
This should be a rather straightforward review. :) Some errant slashes were preventing the python-dateutil link from working.


WillAyd:
Thanks @jarmar

PR #22264, opened 2018-08-09
DEPR: list-likes of list-likes in str.cat

### LINKED ISSUE #21950: DEPR: list of lists in Series.str.cat
The `.str.cat`-method is the only one in the `str`-accessor that takes another Series as an argument, and as such, is a bit of a special case (e.g. it had no index alignment until v0.23).

It makes sense to support lists of objects which get concatenated sequentially, and list of lists have been supported since at least v0.17, see https://pandas.pydata.org/pandas-docs/version/0.17.0/generated/pandas.Series.str.cat.html

When I wrote #20347, I tried very hard to keep signature backwards-compatible, and the example from the v0.17-22 docs working:
```
>>> Series(['a', 'b']).str.cat([['x', 'y'], ['1', '2']], sep=',')
0    a,x,1
1    b,y,2
dtype: object
```
However, this added lots of complexity, and I think that this should be simplified, especially in light of @TomAugspurger's comment in #21894
> As a reminder, the plan is to have no new deprecations in 0.25.x and 1.0.0. So this [v0.24] is the last round of deprecations before 1.0.

My suggestion is to modify the allowed combinations (as of v0.23) as follows:
```
Type of "others"                        |  action  |  comment
---------------------------------------------------------------------
list-like of strings                    |   keep   |  as before; mimics behavior elsewhere,
                                                      cf.: pd.Series(range(3)) + [2,4,6]
Series                                  |   keep   |
np.ndarray (1-dim)                      |   keep   |
DataFrame                               |   keep   |  sequential concatenation
np.ndarray (2-dim)                      |   keep   |  sequential concatenation
list-like of
    Series/Index/np.ndarray (1-dim)     |   keep   |  sequential concatenation
list-like containing list-likes (1-dim)
    other than Series/Index/np.ndarray  |   DEPR   |  sequential concatenation
```

In other words, if the user wants sequential concatenation, there are many possibilities available, and list-of-lists does not have to be one of them, IMO. This would substantially simplify (post-deprecation) the code for `str.cat._get_series_list`, which is currently a bit complicated. https://github.com/pandas-dev/pandas/blob/v0.23.3/pandas/core/strings.py#L2089

Finally, for completeness, the example from the v0.17-22 docs has been removed for v0.23, but there are two examples in https://pandas.pydata.org/pandas-docs/stable/text.html#concatenating-a-series-and-many-objects-into-a-series that would fall under the deprecation I'm suggesting.

h-vetinari:
- [x] closes #21950
- [x] tests modified / passed
- [x] whatsnew entry

As mentioned in #21950, my suggestion is to modify the allowed combinations (as of v0.23) as follows:
```
Type of "others"                        |  action  |  comment
---------------------------------------------------------------------
list-like of strings                    |   keep   |  as before; mimics behavior elsewhere,
                                                      cf.: pd.Series(range(3)) + [2,4,6]
Series                                  |   keep   |
np.ndarray (1-dim)                      |   keep   |
DataFrame                               |   keep   |  sequential concatenation
np.ndarray (2-dim)                      |   keep   |  sequential concatenation
list-like of
    Series/Index/np.ndarray (1-dim)     |   keep   |  sequential concatenation
list-like containing list-likes (1-dim)
    other than Series/Index/np.ndarray  |   DEPR   |  sequential concatenation
```

In other words, if the user wants sequential concatenation, there are many possibilities available, and list-of-lists does not have to be one of them, IMO. This would substantially simplify (post-deprecation) the code for `str.cat._get_series_list`: 

```
def _get_series_list(self, others):
    """
    Auxiliary function for :meth:`str.cat`. Turn potentially mixed input
    into a list of Series (elements without an index must match the length
    of the calling Series/Index).

    Parameters
    ----------
    others : Series, DataFrame, np.ndarray, list-like or list-like of
        objects that are either Series, Index or np.ndarray (1-dim)

    Returns
    -------
    list : others transformed into list of Series
    """
    from pandas import Index, Series, DataFrame

    # self._orig is either Series or Index
    idx = self._orig if isinstance(self._orig, Index) else self._orig.index

    # Generally speaking, all objects without an index inherit the index
    # `idx` of the calling Series/Index - i.e. must have matching length.
    # Objects with an index (i.e. Series/Index/DataFrame) keep their own.
    if isinstance(others, Series):
        return [others]
    elif isinstance(others, Index):
        return [Series(others.values, index=others)]
    elif isinstance(others, DataFrame):
        return [others[x] for x in others]
    elif isinstance(others, np.ndarray) and others.ndim == 2:
        others = DataFrame(others, index=idx)
        return [others[x] for x in others]
    elif is_list_like(others):
        others = list(others)  # ensure iterators do not get read twice etc

        # in case of list-like `others`, all elements must be
        # either Series/Index/np.ndarray (1-dim)...
        if all(isinstance(x, (Series, Index))
               or (isinstance(x, np.ndarray) and x.ndim == 1) for x in others):
            los = []
            while others:  # iterate through list and append each element
                los = los + self._get_series_list(others.pop(0))
            return los
        # ... or just strings
        elif all(not is_list_like(x) for x in others):
            return [Series(others, index=idx)]
    raise TypeError('others must be Series, Index, DataFrame, np.ndarrary or '
                    'list-like (either containing only strings or containing '
                    'only objects of type Series/Index/np.ndarray[1-dim])')
```

### DIFF
--- a/doc/source/text.rst
+++ b/doc/source/text.rst
@@ -306,7 +306,7 @@ The same alignment can be used when ``others`` is a ``DataFrame``:
 Concatenating a Series and many objects into a Series
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-All one-dimensional list-likes can be arbitrarily combined in a list-like container (including iterators, ``dict``-views, etc.):
+All one-dimensional list-likes can be combined in a list-like container (including iterators, ``dict``-views, etc.):
 
 .. ipython:: python
 
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -515,7 +515,7 @@ Deprecations
 - :meth:`Categorical.from_codes` has deprecated providing float values for the ``codes`` argument. (:issue:`21767`)
 - :func:`pandas.read_table` is deprecated. Instead, use :func:`pandas.read_csv` passing ``sep='\t'`` if necessary (:issue:`21948`)
 - :meth:`Series.str.cat` has deprecated using arbitrary list-likes *within* list-likes. A list-like container may still contain
-  many ``Series``, ``Index`` or 1-dimenstional ``np.ndarray``, or alternatively, only scalar values. (:issue:`21950`)
+  many ``Series``, ``Index`` or 1-dimensional ``np.ndarray``, or alternatively, only scalar values. (:issue:`21950`)
 
 .. _whatsnew_0240.prior_deprecations:
 
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -506,13 +506,15 @@ Other API Changes
 Deprecations
 ~~~~~~~~~~~~
 
-- :meth:`DataFrame.to_stata`, :meth:`read_stata`, :class:`StataReader` and :class:`StataWriter` have deprecated the ``encoding`` argument.  The encoding of a Stata dta file is determined by the file type and cannot be changed (:issue:`21244`).
-- :meth:`MultiIndex.to_hierarchical` is deprecated and will be removed in a future version  (:issue:`21613`)
+- :meth:`DataFrame.to_stata`, :meth:`read_stata`, :class:`StataReader` and :class:`StataWriter` have deprecated the ``encoding`` argument. The encoding of a Stata dta file is determined by the file type and cannot be changed (:issue:`21244`)
+- :meth:`MultiIndex.to_hierarchical` is deprecated and will be removed in a future version (:issue:`21613`)
 - :meth:`Series.ptp` is deprecated. Use ``numpy.ptp`` instead (:issue:`21614`)
 - :meth:`Series.compress` is deprecated. Use ``Series[condition]`` instead (:issue:`18262`)
 - The signature of :meth:`Series.to_csv` has been uniformed to that of doc:meth:`DataFrame.to_csv`: the name of the first argument is now 'path_or_buf', the order of subsequent arguments has changed, the 'header' argument now defaults to True. (:issue:`19715`)
 - :meth:`Categorical.from_codes` has deprecated providing float values for the ``codes`` argument. (:issue:`21767`)
 - :func:`pandas.read_table` is deprecated. Instead, use :func:`pandas.read_csv` passing ``sep='\t'`` if necessary (:issue:`21948`)
+- :meth:`Series.str.cat` has deprecated using arbitrary list-likes *within* list-likes. A list-like container may still contain
+  many ``Series``, ``Index`` or 1-dimenstional ``np.ndarray``, or alternatively, only scalar values. (:issue:`21950`)
 
 .. _whatsnew_0240.prior_deprecations:
 
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1930,8 +1930,8 @@ class StringMethods(NoNewAttributesMixin):
 
         Parameters
         ----------
-        others : Series, DataFrame, np.ndarray, list-like or list-like of
-            objects that are either Series, np.ndarray (1-dim) or list-like
+        others : Series, Index, DataFrame, np.ndarray, list-like or list-like
+            of objects that are Series, Index or np.ndarray (1-dim)
         ignore_index : boolean, default False
             Determines whether to forcefully align others with index of caller
 
@@ -1986,11 +1986,20 @@ class StringMethods(NoNewAttributesMixin):
             # either one-dimensional list-likes or scalars
             if all(is_list_like(x) for x in others):
                 los = []
-                warn = False
+                join_warn = False
+                depr_warn = False
                 # iterate through list and append list of series for each
                 # element (which we check to be one-dimensional and non-nested)
                 while others:
                     nxt = others.pop(0)  # nxt is guaranteed list-like by above
+
+                    # GH 21950 - DeprecationWarning
+                    # only allowing Series/Index/np.ndarray[1-dim] will greatly
+                    # simply this function post-deprecation.
+                    if not (isinstance(nxt, (Series, Index)) or
+                            (isinstance(nxt, np.ndarray) and nxt.ndim == 1)):
+                        depr_warn = True
+
                     if not isinstance(nxt, (DataFrame, Series,
                                             Index, np.ndarray)):
                         # safety for non-persistent list-likes (e.g. iterators)
@@ -2013,8 +2022,15 @@ class StringMethods(NoNewAttributesMixin):
                     nxt, wnx = self._get_series_list(nxt,
                                                      ignore_index=ignore_index)
                     los = los + nxt
-                    warn = warn or wnx
-                return (los, warn)
+                    join_warn = join_warn or wnx
+
+                    if depr_warn:
+                        warnings.warn('list-likes other than Series, Index, '
+                                      'or np.ndarray WITHIN another list-like '
+                                      'are deprecated and will be removed in '
+                                      'a future version.',
+                                      FutureWarning, stacklevel=3)
+                return (los, join_warn)
             elif all(not is_list_like(x) for x in others):
                 return ([Series(others, index=idx)], False)
         raise TypeError(err_msg)
@@ -2037,8 +2053,8 @@ class StringMethods(NoNewAttributesMixin):
             Series/Index/DataFrame) if `join` is not None.
 
             If others is a list-like that contains a combination of Series,
-            np.ndarray (1-dim) or list-like, then all elements will be unpacked
-            and must satisfy the above criteria individually.
+            Index or np.ndarray (1-dim), then all elements will be unpacked and
+            must satisfy the above criteria individually.
 
             If others is None, the method returns the concatenation of all
             strings in the calling Series/Index.
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -2024,12 +2024,11 @@ class StringMethods(NoNewAttributesMixin):
                     los = los + nxt
                     join_warn = join_warn or wnx
 
-                    if depr_warn:
-                        warnings.warn('list-likes other than Series, Index, '
-                                      'or np.ndarray WITHIN another list-like '
-                                      'are deprecated and will be removed in '
-                                      'a future version.',
-                                      FutureWarning, stacklevel=3)
+                if depr_warn:
+                    warnings.warn('list-likes other than Series, Index, or '
+                                  'np.ndarray WITHIN another list-like are '
+                                  'deprecated and will be removed in a future '
+                                  'version.', FutureWarning, stacklevel=3)
                 return (los, join_warn)
             elif all(not is_list_like(x) for x in others):
                 return ([Series(others, index=idx)], False)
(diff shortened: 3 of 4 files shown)


doc/source/text.rst
```diff
@@ -306,20 +306,20 @@ The same alignment can be used when ``others`` is a ``DataFrame``:
 Concatenating a Series and many objects into a Series
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-All one-dimensional list-likes can be arbitrarily combined in a list-like container (including iterators, ``dict``-views, etc.):
+Several objects of type ``Series``, ``Index`` or ``np.ndarray`` can be arbitrarily combined in a list-like container (including iterators, ``dict``-views, etc.):
```
gfyoung: "Several objects of type `Series`, `Index`, or `np.ndarray`" - are you sure that's what you meant to say?  I think that should be reworded.

pandas/core/strings.py
```diff
@@ -2013,8 +2022,15 @@ def _get_series_list(self, others, ignore_index=False):
                     nxt, wnx = self._get_series_list(nxt,
                                                      ignore_index=ignore_index)
                     los = los + nxt
-                    warn = warn or wnx
-                return (los, warn)
+                    join_warn = join_warn or wnx
+
+                    if depr_warn:
+                        warnings.warn('list-likes other than Series, Index or '
```
gfyoung: Nit: add a comma after "Index"

doc/source/text.rst
```diff
@@ -306,20 +306,20 @@ The same alignment can be used when ``others`` is a ``DataFrame``:
...
 
-All one-dimensional list-likes can be arbitrarily combined in a list-like container (including iterators, ``dict``-views, etc.):
+An arbitrary combination of ``Series``, ``Index`` and ``np.ndarray`` (1-dimensional) can be passed in a list-like container (including iterators, ``dict``-views, etc.):
 
 .. ipython:: python
 
     s
     u
-    s.str.cat([u.values, ['A', 'B', 'C', 'D'], map(str, u.index)], na_rep='-')
+    s.str.cat([u.values, u, u.index.astype(str)], na_rep='-')
```
jreback: can you save these in another variable, rather than doing the combination in-line. its easier to see it printed out (with comments)

doc/source/text.rst
```diff
@@ -306,20 +306,20 @@ The same alignment can be used when ``others`` is a ``DataFrame``:
...
 
 All elements must match in length to the calling ``Series`` (or ``Index``), except those having an index if ``join`` is not None:
 
 .. ipython:: python
 
     v
-    s.str.cat([u, v, ['A', 'B', 'C', 'D']], join='outer', na_rep='-')
+    s.str.cat([u, v, u.values], join='outer', na_rep='-')
```
jreback: same here

jreback (changes requested):
looks fine, ex comments on the text.rst

doc/source/text.rst
h-vetinari: What do you mean by "comments"?

doc/source/text.rst
h-vetinari: In this case, `u` and `u.values` will have been shown 3 lines above, and `v` directly above. I think this should be plenty, no?

doc/source/text.rst
```diff
@@ -306,20 +306,23 @@ The same alignment can be used when ``others`` is a ``DataFrame``:
...
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-All one-dimensional list-likes can be arbitrarily combined in a list-like container (including iterators, ``dict``-views, etc.):
+An arbitrary combination of ``Series``, ``Index`` and ``np.ndarray`` (1-dimensional) can be passed in a list-like container (including iterators, ``dict``-views, etc.):
 
 .. ipython:: python
 
     s
     u
-    s.str.cat([u.values, ['A', 'B', 'C', 'D'], map(str, u.index)], na_rep='-')
```
jreback: put some comments here, this is just too much code in a row

doc/source/text.rst
jreback: use w as you defined above

jreback (changes requested):

doc/source/text.rst
h-vetinari: From  my POV, it's just one line of code, while showing all the ingredients. But I'm gonna change it back to just using variants of `u` in different "boxes".

I think it's fair enough to show just `u` and not `u.values` and `pd.Index(u)` as well.

doc/source/text.rst
h-vetinari: The point here is to show how `join` works with objects that have no index, so can't use `w` (which has an index).

jreback:
maybe would be helpful to show a rendered version here of text.rst

h-vetinari:
@jreback here you go:

h-vetinari:
@jreback @TomAugspurger ping :)

doc/source/text.rst
jreback: so pls show u.values, I have no idea what it actually is, and as I said, pls either add comments or split up showing s, u, u.values, this is just too long

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -474,12 +474,14 @@ Other API Changes
 Deprecations
 ~~~~~~~~~~~~
 
-- :meth:`DataFrame.to_stata`, :meth:`read_stata`, :class:`StataReader` and :class:`StataWriter` have deprecated the ``encoding`` argument.  The encoding of a Stata dta file is determined by the file type and cannot be changed (:issue:`21244`).
-- :meth:`MultiIndex.to_hierarchical` is deprecated and will be removed in a future version  (:issue:`21613`)
+- :meth:`DataFrame.to_stata`, :meth:`read_stata`, :class:`StataReader` and :class:`StataWriter` have deprecated the ``encoding`` argument. The encoding of a Stata dta file is determined by the file type and cannot be changed (:issue:`21244`)
```
jreback: pls don't edit other things. The longer the PR the more time things will take. I guess since its already done leave it.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -474,12 +474,14 @@ Other API Changes
...
+- :meth:`MultiIndex.to_hierarchical` is deprecated and will be removed in a future version (:issue:`21613`)
 - :meth:`Series.ptp` is deprecated. Use ``numpy.ptp`` instead (:issue:`21614`)
 - :meth:`Series.compress` is deprecated. Use ``Series[condition]`` instead (:issue:`18262`)
 - :meth:`Categorical.from_codes` has deprecated providing float values for the ``codes`` argument. (:issue:`21767`)
 - :func:`pandas.read_table` is deprecated. Instead, use :func:`pandas.read_csv` passing ``sep='\t'`` if necessary (:issue:`21948`)
+- :meth:`Series.str.cat` has deprecated using arbitrary list-likes *within* list-likes. A list-like container may still contain
+  arbitrarily many ``Series``, ``Index`` or 1-dimenstional ``np.ndarray``, or alternatively, only scalar values. (:issue:`21950`)
```
jreback: remove the word arbitrarily (and in text.rst)

pandas/core/strings.py
```diff
@@ -1930,8 +1930,8 @@ def _get_series_list(self, others, ignore_index=False):
 
         Parameters
         ----------
-        others : Series, DataFrame, np.ndarray, list-like or list-like of
-            objects that are either Series, np.ndarray (1-dim) or list-like
+        others : Series, Index, DataFrame, np.ndarray, list-like or list-like
+            of objects that are Series, Index, np.ndarray (1-dim)
```
jreback: or np.ndarray

jreback (changes requested):

doc/source/text.rst
h-vetinari: @jreback `u` is shown directly above...?! Also in the screenshot I posted - in context it's IMO very clear (especially as it's at the end of the concatenation section, where the reader has seen the same objects 10 times already).

I can of course show **both** `u` and `u.values`, but I thought that understanding the relationship between a Series and its `.values` would be safe to expect of a users pandas-fu.

h-vetinari:
@jreback Incorporated feedback, except the `u.values` part, where my comment above doesn't show up (so copying it here):

> so pls show u.values, I have no idea what it actually is, and as I said, pls either add comments or split up showing s, u, u.values, this is just too long

`u` is shown directly above...?! Also in the screenshot I posted. I can of course show **both** `u` and `u.values`, but I thought that understanding the relationship between a Series and its `.values` would be safe to expect of a users' pandas-fu.

In context it's IMO very clear (especially as it's at the end of the concatenation section, where the reader has seen the same objects 10 times already).

h-vetinari:
@jreback Same screenshot as above. IMO the relationship `u` <--> `u.values` does not need to be explained - if you disagree with that, please let me know. (side note: the example in the current docs is even more complicated; but ultimately its easy to grok from the result what's the content of the inputs -- see `[70]` below)

1_alt

jreback (approved):

jreback:
ok this is fine. can you rebase once more and ping on green.

h-vetinari:
@jreback All green.

h-vetinari:
@jreback 
> ok this is fine. can you rebase once more and ping on green.

All green. Just fixed an indentation error on the way.

gfyoung (approved):
cc @jreback

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -507,13 +507,15 @@ Other API Changes
...
+- :meth:`MultiIndex.to_hierarchical` is deprecated and will be removed in a future version (:issue:`21613`)
 - :meth:`Series.ptp` is deprecated. Use ``numpy.ptp`` instead (:issue:`21614`)
 - :meth:`Series.compress` is deprecated. Use ``Series[condition]`` instead (:issue:`18262`)
 - The signature of :meth:`Series.to_csv` has been uniformed to that of doc:meth:`DataFrame.to_csv`: the name of the first argument is now 'path_or_buf', the order of subsequent arguments has changed, the 'header' argument now defaults to True. (:issue:`19715`)
 - :meth:`Categorical.from_codes` has deprecated providing float values for the ``codes`` argument. (:issue:`21767`)
 - :func:`pandas.read_table` is deprecated. Instead, use :func:`pandas.read_csv` passing ``sep='\t'`` if necessary (:issue:`21948`)
+- :meth:`Series.str.cat` has deprecated using arbitrary list-likes *within* list-likes. A list-like container may still contain
+  many ``Series``, ``Index`` or 1-dimenstional ``np.ndarray``, or alternatively, only scalar values. (:issue:`21950`)
```
jreback: dimensional

jreback (approved):
tiny typo, ping on push.

h-vetinari:
@jreback 
> tiny typo, ping on push.

Side note: `[ci skip]` does not seem to be respected by the new CircleCI.

h-vetinari:
@jreback

This is all done, approved and green.

jreback:
thanks @h-vetinari

h-vetinari:
@jreback 

> > Side note: [ci skip] does not seem to be respected by the new CircleCI.

> I don't like these skip's generally. they are not necessary as all of the CI's cancel prior builds anyhow.

This was not about cancelling prior builds but not having to run the full CI suite if (say) a single comment (like for the GH number) gets added. I view the CI as somewhat limited resources (appveyor recently having something like 12h waiting time), so `[ci skip]` can be useful for avoiding a CI run where it's pointless (any code/doc changes should of course run through).

PR #22266, opened 2018-08-09
ENH: Add support for excluding the index from Parquet files (GH20768)

### LINKED ISSUE #20768: to_parquet method should accept index=False option

```python
# for example, index should be accepted, like so:
df.to_parquet( 'my_parquet_file.parquet', compression='gzip', index=False )
```
#### Problem description

[this should explain **why** the current behaviour is a problem and why the expected output is a better solution.]

**Note**: We receive a lot of issues on our GitHub tracker, so it is very possible that your issue has been posted before. Please check first before submitting so that we do not have to handle and close duplicates!

**Note**: Many problems can be resolved by simply upgrading `pandas` to the latest version. Before submitting, please check if that solution works for you. If possible, you may want to check if `master` addresses this issue, but that is not necessary.

For documentation-related issues, you can check the latest versions of the docs on `master` here:

https://pandas-docs.github.io/pandas-docs-travis/

If the issue has not been resolved there, go ahead and file it in the issue tracker.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.4.final.0
python-bits: 64
OS: Linux
OS-release: 3.16.0-77-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.22.0
pytest: 3.2.5
pip: 9.0.1
setuptools: 39.0.1
Cython: 0.28.1
numpy: 1.14.2
scipy: 0.19.1
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: 1.7.2
patsy: 0.5.0
dateutil: 2.7.2
pytz: 2018.3
blosc: None
bottleneck: None
tables: None
numexpr: 2.6.2
feather: None
matplotlib: 2.0.2
openpyxl: 2.5.1
xlrd: 1.0.0
xlwt: None
xlsxwriter: None
lxml: 4.2.1
bs4: 4.6.0
html5lib: 0.999999999
sqlalchemy: 1.2.5
pymysql: None
jinja2: 2.10
s3fs: None
fastparquet: 0.1.4
pandas_gbq: None
pandas_datareader: None

</details>

dargueta:
- [x] closes #20768
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -4644,9 +4644,7 @@ columns in the output file. Thus, this code:
 
 .. ipython:: python
 
-    import pandas
-
-    df = pandas.DataFrame({'a': [1, 2], 'b': [3, 4]})
+    df = pd.DataFrame({'a': [1, 2], 'b': [3, 4]})
     df.to_parquet('test.parquet', engine='pyarrow')
 
 creates a parquet file with *three* columns if you use ``pyarrow`` for serialization:
@@ -4662,9 +4660,6 @@ If you want to omit a dataframe's indexes when writing, pass ``index=False`` to
 
 .. ipython:: python
 
-    import pandas
-
-    df = pandas.DataFrame({'a': [1, 2], 'b': [3, 4]})
     df.to_parquet('test.parquet', index=False)
 
 This creates a parquet file with just the two expected columns, ``a`` and ``b``.
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -4570,6 +4570,8 @@ dtypes, including extension dtypes such as datetime with tz.
 Several caveats.
 
 * Duplicate column names and non-string columns names are not supported.
+* The index is included in the output by default, which can cause problems with non-Pandas consumers that are
+  not expecting that extra column. You can, however, omit indexes. (See below)
 * Index level names, if specified, must be strings.
 * Categorical dtypes can be serialized to parquet, but will de-serialize as ``object`` dtype.
 * Non supported types include ``Period`` and actual Python object types. These will raise a helpful error message
@@ -4632,6 +4634,40 @@ Read only certain columns of a parquet file.
    import os
    os.remove('example_pa.parquet')
    os.remove('example_fp.parquet')
+
+
+Omitting Indexes
+''''''''''''''''
+
+Dumping a ``DataFrame`` to parquet includes the implicit index as one or more
+columns in the output file. Thus, this code:
+
+.. ipython:: python
+
+    import pandas
+
+    df = pandas.DataFrame({'a': [1, 2], 'b': [3, 4]})
+    df.to_parquet('test.parquet')
+
+creates a parquet file with *three* columns: ``a``, ``b``, and ``__index_level_0__``.
+
+This unexpected extra column causes some databases like Amazon Redshift to reject
+the file, because that column doesn't exist in the target table.
+
+If you want to omit a dataframe's indexes when writing, pass ``index=False`` to
+:func:`~pandas.DataFrame.to_parquet`:
+
+.. ipython:: python
+
+    import pandas
+
+    df = pandas.DataFrame({'a': [1, 2], 'b': [3, 4]})
+    df.to_parquet('test.parquet', index=False)
+
+This creates a parquet file with just the two expected columns, ``a`` and ``b``.
+If your ``DataFrame`` has a custom index, you won't get it back when you load
+this file into a ``DataFrame``.
+
 
 .. _io.sql:
 
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -4640,8 +4640,8 @@ Read only certain columns of a parquet file.
 Handling Indexes
 ''''''''''''''''
 
-Dumping a ``DataFrame`` to parquet may include the implicit index as one or more
-columns in the output file. Thus, this code:
+Serializing a ``DataFrame`` to parquet may include the implicit index as one or
+more columns in the output file. Thus, this code:
 
 .. ipython:: python
 
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -4636,10 +4636,10 @@ Read only certain columns of a parquet file.
    os.remove('example_fp.parquet')
 
 
-Omitting Indexes
+Handling Indexes
 ''''''''''''''''
 
-Dumping a ``DataFrame`` to parquet includes the implicit index as one or more
+Dumping a ``DataFrame`` to parquet may include the implicit index as one or more
 columns in the output file. Thus, this code:
 
 .. ipython:: python
@@ -4647,9 +4647,12 @@ columns in the output file. Thus, this code:
     import pandas
 
     df = pandas.DataFrame({'a': [1, 2], 'b': [3, 4]})
-    df.to_parquet('test.parquet')
+    df.to_parquet('test.parquet', engine='pyarrow')
 
-creates a parquet file with *three* columns: ``a``, ``b``, and ``__index_level_0__``.
+creates a parquet file with *three* columns if you use ``pyarrow`` for serialization:
+``a``, ``b``, and ``__index_level_0__``. If you're using ``fastparquet``, the
+index `may or may not <https://fastparquet.readthedocs.io/en/latest/api.html#fastparquet.write>`_
+be written to the file.
 
 This unexpected extra column causes some databases like Amazon Redshift to reject
 the file, because that column doesn't exist in the target table.
@@ -4667,6 +4670,9 @@ If you want to omit a dataframe's indexes when writing, pass ``index=False`` to
 This creates a parquet file with just the two expected columns, ``a`` and ``b``.
 If your ``DataFrame`` has a custom index, you won't get it back when you load
 this file into a ``DataFrame``.
+
+Passing ``index=True`` will *always* write the index, even if that's not the
+underlying engine's default behavior.
 
 
 .. _io.sql:
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -4570,8 +4570,9 @@ dtypes, including extension dtypes such as datetime with tz.
 Several caveats.
 
 * Duplicate column names and non-string columns names are not supported.
-* The index is included in the output by default, which can cause problems with non-Pandas consumers that are
-  not expecting that extra column. You can, however, omit indexes. (See below)
+* The ``pyarrow`` engine always writes the index to the output, but ``fastparquet`` only writes non-default
+  indexes. This extra column can cause problems for non-Pandas consumers that are not expecting it. You can
+  force including or omitting indexes with the ``index`` argument, regardless of the underlying engine.
 * Index level names, if specified, must be strings.
 * Categorical dtypes can be serialized to parquet, but will de-serialize as ``object`` dtype.
 * Non supported types include ``Period`` and actual Python object types. These will raise a helpful error message
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -17,6 +17,8 @@ New features
 
 - ``ExcelWriter`` now accepts ``mode`` as a keyword argument, enabling append to existing workbooks when using the ``openpyxl`` engine (:issue:`3441`)
 
+- ``DataFrame.to_parquet()`` now accepts ``index`` as a keyword argument, allowing the user to optionally omit the dataframe's indexes from the resulting Parquet file. (:issue:`20768`)
+
 .. _whatsnew_0240.enhancements.extension_array_operators:
 
 ``ExtensionArray`` operator support
@@ -157,6 +159,45 @@ array, but rather an ``ExtensionArray``:
 
 This is the same behavior as ``Series.values`` for categorical data. See
 :ref:`whatsnew_0240.api_breaking.interval_values` for more.
+
+
+Omitting Indexes in ``to_parquet()``
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+You now have the option to omit a dataframe's indexes when writing to Parquet
+files with ``to_parquet``, just like ``to_csv`` (:issue:`20768`).
+
+*Previous Behavior*:
+
+Dumping a ``DataFrame`` to Parquet would include the implicit index as a column
+in the output file. Thus, this code:
+
+.. ipython:: python
+
+    import pandas
+
+    df = pandas.DataFrame({'a': [1, 2], 'b': [3, 4]})
+    df.to_parquet('test.parquet')
+
+would create a Parquet file with *three* columns: ``a``, ``b``, and ``__index_level_0__``.
+
+This unexpected extra column causes some databases like Amazon Redshift to reject
+the file, because that column doesn't exist in the target table.
+
+*New Behavior*:
+
+If you want to omit a dataframe's indexes when writing, pass ``index=False`` to
+``to_parquet()``:
+
+.. ipython:: python
+
+    import pandas
+
+    df = pandas.DataFrame({'a': [1, 2], 'b': [3, 4]})
+    df.to_parquet('test.parquet', index=False)
+
+This creates a Parquet file with just the two expected columns, ``a`` and ``b``.
+The preexisting behavior is still the default.
 
 
 .. _whatsnew_0240.enhancements.other:
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -17,7 +17,9 @@ New features
 
 - ``ExcelWriter`` now accepts ``mode`` as a keyword argument, enabling append to existing workbooks when using the ``openpyxl`` engine (:issue:`3441`)
 
-- :func:`DataFrame.to_parquet` now accepts ``index`` as an argument, allowing the user to override the engine's default behavior and include or omit the dataframe's indexes from the resulting Parquet file. (:issue:`20768`)
+- :func:`DataFrame.to_parquet` now accepts ``index`` as an argument, allowing
+the user to override the engine's default behavior to include or omit the
+dataframe's indexes from the resulting Parquet file. (:issue:`20768`)
 
 .. _whatsnew_0240.enhancements.extension_array_operators:
 
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -17,7 +17,7 @@ New features
 
 - ``ExcelWriter`` now accepts ``mode`` as a keyword argument, enabling append to existing workbooks when using the ``openpyxl`` engine (:issue:`3441`)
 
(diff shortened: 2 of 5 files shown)


pandas/core/frame.py
```diff
@@ -1982,6 +1982,9 @@ def to_parquet(self, fname, engine='auto', compression='snappy',
             'pyarrow' is unavailable.
         compression : {'snappy', 'gzip', 'brotli', None}, default 'snappy'
             Name of the compression to use. Use ``None`` for no compression.
+        index : bool
```
gfyoung: Say "bool, default True"

gfyoung:
Good start!  Going to need tests as well as a `whatsnew` entry (potentially a mini-section).

cc @jreback

dargueta:
> Going to need tests as well as a whatsnew entry (potentially a mini-section).

I won't have much time until later today but yeah, I'll finish that off.

pandas/core/frame.py
dargueta: A bit of a nitpicky question, but do you mean

```
bool, default ``True``
```
or 
```
bool, default True
```
?

Difference is pretty much just formatting in the rendered docs, but I tend to be a stickler for convention.

gfyoung: Good question.  The instances that I have seen had no double ticks.

dargueta:
@gfyoung I'm not sure I completely understand what's going on in the unit tests. I _think_ I've written a test that handles the expected cases for both engines but I'm not sure.

pandas/tests/io/test_parquet.py
```diff
@@ -368,6 +368,12 @@ def test_multiindex_with_columns(self, pa_ge_070):
             check_round_trip(df, engine, read_kwargs={'columns': ['A', 'B']},
                              expected=df[['A', 'B']])
 
+    def test_write_ignoring_index(self):
+        """Ensure index=False omits the index from the written Parquet file."""
```
gfyoung: * This should be a comment, not a docstring.
* Reference the issue number in the comment.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -17,6 +17,8 @@ New features
 
 - ``ExcelWriter`` now accepts ``mode`` as a keyword argument, enabling append to existing workbooks when using the ``openpyxl`` engine (:issue:`3441`)
 
+- ``DataFrame.to_parquet()`` now accepts ``index`` as a keyword argument, allowing the user to optionally omit the dataframe's indexes from the resulting Parquet file. (:issue:`20768`)
+
```
gfyoung: We generally do either the `whatsnew` entry OR the mini-section but not both.

cc @jreback

pandas/tests/io/test_parquet.py
dargueta: Done!

dargueta:
Do you have any idea what's causing these seemingly unrelated tests to fail on Travis? The failing appveyor test seems to be a bit more relevant.

WillAyd:
Just glancing at the Travis failures on 3.6 they seem related - looks like there is a difference in the name of the index before / after (`None` vs `index`)

dargueta:
So why isn't it failing on CircleCI? Isn't it the same code?

Also, my changes should _not_ be causing `TestPythonParser.test_no_header_prefix` to suddenly start failing, right? That's another one failing on Travis.

gfyoung:
@dargueta : Good question!  Not quite.  The coverage for each build / platform is slightly different.  That's why we use multiple platforms, for better or worse. 🙂

gfyoung:
Why can't you pass in an argument for `read_kwargs` ?

dargueta:
This feature was intended to be only for writing, but I suppose I could add it for reading. Not sure how well that'd be supported by the Parquet libraries but I can check.

pandas/tests/io/test_parquet.py
```diff
@@ -368,6 +368,41 @@ def test_multiindex_with_columns(self, pa_ge_070):
...
+    def test_write_ignoring_index(self, engine):
+        # ENH 20768
+        # Ensure index=False omits the index from the written Parquet file.
+        df = pd.DataFrame({'a': [1, 2, 3], 'b': ['q', 'r', 's']})
+
+        # Because we're dropping the index, we expect the loaded dataframe to
+        # have the default integer index.
+        expected = df.reset_index(drop=True)
+
+        check_round_trip(df, engine, write_kwargs={'index': False},
```
jorisvandenbossche: reason for the failing test with `RuntimeError: Compression 'snappy' not available` is that `check_round_trip` normally sets `write_kwargs` to `{'compression': None}` is it is not specified (because `fastparquet` does not use the snappy compression). So to solve this you can add `'compression': None` here to the dict as well.

pandas/tests/io/test_parquet.py
```diff
@@ -368,6 +368,41 @@ def test_multiindex_with_columns(self, pa_ge_070):
...
+        # ENH 20768
+        # Ensure index=False omits multiindexes as well.
+        arrays = [['bar', 'bar', 'baz', 'baz', 'foo', 'foo', 'qux', 'qux'],
+                  ['one', 'two', 'one', 'two', 'one', 'two', 'one', 'two']]
+        df = pd.DataFrame({'one': [i for i in range(8)],
+                           'two': [-i for i in range(8)]}, index=arrays)
+
+        expected = df.reset_index(drop=True)
+        check_round_trip(df, engine, write_kwargs={'index': False},
+                         expected=expected)
```
jorisvandenbossche: Those tests with a custom index are failing, but those are actual test failures you need to solve. Currently we raise an error for older pyarrow versions if there is a non-default index (because older pyarrow could not handle that). But when dropping the index, this error is not needed. So the code that checks this can be updated (see https://github.com/pandas-dev/pandas/blob/ad19d73c0c929add3b127bc3f0a81b509de44ca5/pandas/io/parquet.py#L141)

jorisvandenbossche (commented):
Already looking good!

Concerning the failures related to index names: one possible cause is the `write_index` keyword for fastparquet. Now you added that in the `fastparquet.write` call with a default of `True`. However, its default is not True in fastparquet, so this actually changes behaviour: https://github.com/dask/fastparquet/blob/8db811cc4701d5ae100a8e5c95685daec9e24c6b/fastparquet/writer.py#L793-L795

Only not fully sure what the best solution is here. We could restore the current behaviour (not write index by default for a default index, i.e. `write_index=None`), but it would also be nice to have it consistent between the different engines.

pandas/tests/io/test_parquet.py
dargueta: Great, thanks!

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -17,6 +17,8 @@ New features
 
 - ``ExcelWriter`` now accepts ``mode`` as a keyword argument, enabling append to existing workbooks when using the ``openpyxl`` engine (:issue:`3441`)
 
+- ``DataFrame.to_parquet()`` now accepts ``index`` as a keyword argument, allowing the user to optionally omit the dataframe's indexes from the resulting Parquet file. (:issue:`20768`)
```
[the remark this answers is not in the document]
dargueta: Done.

pandas/tests/io/test_parquet.py
```diff
@@ -368,6 +368,41 @@ def test_multiindex_with_columns(self, pa_ge_070):
...
+        # have the default integer index.
+        expected = df.reset_index(drop=True)
+
+        check_round_trip(df, engine, write_kwargs={'index': False},
+                         expected=expected)
+
+    def test_write_ignoring_custom_index(self, engine):
+        # ENH 20768
+        # Ensure index=False omits the index from the written Parquet file,
+        # even if we're using a custom one.s
```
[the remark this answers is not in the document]
dargueta: Done.

dargueta:
> Only not fully sure what the best solution is here. We could restore the current behaviour (not write index by default for a default index, i.e. write_index=None)

The problem is, `pyarrow` _always_ writes the index. Different engines would have different behavior. :|

pandas/tests/io/test_parquet.py
dargueta: I _think_ I fixed it correctly.

doc/source/io.rst
```diff
@@ -4633,6 +4635,46 @@ Read only certain columns of a parquet file.
...
+
+Handling Indexes
+''''''''''''''''
+
+Dumping a ``DataFrame`` to parquet may include the implicit index as one or more
+columns in the output file. Thus, this code:
+
+.. ipython:: python
+
+    import pandas
```
jorisvandenbossche: no need to import pandas here. Pandas is already imported as pd, so you can use that on the line below

doc/source/io.rst
```diff
@@ -4633,6 +4635,46 @@ Read only certain columns of a parquet file.
...
+the file, because that column doesn't exist in the target table.
+
+If you want to omit a dataframe's indexes when writing, pass ``index=False`` to
+:func:`~pandas.DataFrame.to_parquet`:
+
+.. ipython:: python
+
+    import pandas
+
+    df = pandas.DataFrame({'a': [1, 2], 'b': [3, 4]})
```
jorisvandenbossche: also no need to recreate the dataframe. The state is preserved between the different code snippets, so the one of the previous code block can be reused.

pandas/core/frame.py
```diff
@@ -1881,9 +1881,10 @@ def to_parquet(self, fname, engine='auto', compression='snappy',
             'pyarrow' is unavailable.
         compression : {'snappy', 'gzip', 'brotli', None}, default 'snappy'
             Name of the compression to use. Use ``None`` for no compression.
-        index : bool, default True
+        index : bool, default None
             If ``True``, include the dataframe's index(es) in the file output.
-            If ``False``, they will not be written to the file.
+            If ``False``, they will not be written to the file. If ``None``,
+            the behavior depends on the chosen engine.
```
jorisvandenbossche: should we maybe shortly mention that if it is a non-default index, it will always be written with this default behaviour?

pandas/tests/io/test_parquet.py
```diff
@@ -368,6 +368,41 @@ def test_multiindex_with_columns(self, pa_ge_070):
...
+        expected = df.reset_index(drop=True)
+
+        check_round_trip(df, engine, write_kwargs=write_kwargs,
+                         expected=expected)
+
+        # Ignore custom index
+        df = pd.DataFrame({'a': [1, 2, 3], 'b': ['q', 'r', 's']},
+                          index=['zyx', 'wvu', 'tsr'])
+
+        expected = df.reset_index(drop=True)
```
jorisvandenbossche: expected is the same expected as above?

jorisvandenbossche (commented):
Tests seem to be passing now!

doc/source/io.rst
dargueta: Oh wow, I didn't realize this wasn't _just_ code snippets!

pandas/tests/io/test_parquet.py
dargueta: Nice catch!

pandas/core/frame.py
dargueta: I deliberately left that out since if we add support for another Parquet engine, or the implementation of one of the engines changes, our documentation would be out of date.

dargueta:
Last of the (so far) requested fixes are in!

doc/source/whatsnew/v0.24.0.txt
dargueta: Done.

doc/source/io.rst
```diff
@@ -4570,6 +4570,8 @@ dtypes, including extension dtypes such as datetime with tz.
 Several caveats.
 
 * Duplicate column names and non-string columns names are not supported.
```
jreback: don't say see below, just say you can pass ``index=False``

jreback: you can add a ref to the Index section if you want

doc/source/io.rst
```diff
@@ -4633,6 +4635,41 @@ Read only certain columns of a parquet file.
    os.remove('example_pa.parquet')
    os.remove('example_fp.parquet')
 
+
+Handling Indexes
+''''''''''''''''
```
jreback: not really sure you need this doc section at all. ``index=False`` is a pretty common option in pandas.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -17,6 +17,8 @@ New features
 
 - ``ExcelWriter`` now accepts ``mode`` as a keyword argument, enabling append to existing workbooks when using the ``openpyxl`` engine (:issue:`3441`)
 
+- :func:`DataFrame.to_parquet` now accepts ``index`` as an argument, allowing the user to override the engine's default behavior and include or omit the dataframe's indexes from the resulting Parquet file. (:issue:`20768`)
```
jreback: say the default is to *include* the index

pandas/core/frame.py
```diff
@@ -1881,6 +1881,10 @@ def to_parquet(self, fname, engine='auto', compression='snappy',
             'pyarrow' is unavailable.
         compression : {'snappy', 'gzip', 'brotli', None}, default 'snappy'
             Name of the compression to use. Use ``None`` for no compression.
+        index : bool, default None
+            If ``True``, include the dataframe's index(es) in the file output.
+            If ``False``, they will not be written to the file. If ``None``,
+            the behavior depends on the chosen engine.
```
jreback: add a versionadded tag. Why is there a ``None`` option?

pandas/io/parquet.py
```diff
@@ -250,11 +259,15 @@ def to_parquet(df, path, engine='auto', compression='snappy', **kwargs):
         'pyarrow' is unavailable.
     compression : {'snappy', 'gzip', 'brotli', None}, default 'snappy'
         Name of the compression to use. Use ``None`` for no compression.
+    index : bool, default None
+        If ``True``, include the dataframe's index(es) in the file output. If
+        ``False``, they will not be written to the file. If ``None``, the
+        engine's default behavior will be used.
```
jreback: add a versionadded

jreback (changes requested):
why is there a ``None`` option here?

doc/source/whatsnew/v0.24.0.txt
dargueta: > say the default is to _include_ the index

Not necessarily if you're using the `fastparquet` engine. According to the [docs](https://fastparquet.readthedocs.io/en/latest/api.html#fastparquet.write), the index is _only_ written if it's not an integer sequence 0..._n_.

jorisvandenbossche: The default depends on the engine (that's the reason it is None by default, to answer your other question). See the discussion above, eg https://github.com/pandas-dev/pandas/pull/22266#pullrequestreview-146069083. 

The default of None is certainly not the only possibility. As noted above, we can also make it `index=True` by default to have consistency for both engines. But, this would mean a change in behaviour for fastparquet engine. 
On the one hand, I certainly favor this consistency, but on the other hand, I think the default of fastparquet actually makes sense ..

pandas/core/frame.py
dargueta: `fastparquet` doesn't write the index if it's an integer sequence 0-n, but pyarrow always writes the index by default. To preserve backwards compatibility, there's a `None` option that maintains the original engine-dependent behavior.

dargueta:
> Why is there a `None` option here?

`fastparquet` doesn't write the index if it's an integer sequence 0-n, but `pyarrow` always writes the index by default. To preserve backwards compatibility, there's a None option that maintains the original engine-dependent behavior.

jreback:
> fastparquet doesn't write the index if it's an integer sequence 0-n, but pyarrow always writes the index by default. To preserve backwards compatibility, there's a None option that maintains the original engine-dependent behavior.

I c.

This makes fp not idempotent though, which is not a great situation here. I wonder if should make ``index=True/False`` the only option, deprecating ``None`` to preserve back compat with fp. Of course this would make it show on practically every use.

jorisvandenbossche:
> This makes fp not idempotent though, which is not a great situation here

Do you mean that it would not round-trip faithfully? 
Because that is still the case, as it only does not write the index for a default, non-named index. So when reading such parquet file without an explicit index, the created index will be no different.

jreback:
> Do you mean that it would not round-trip faithfully?

is there some magic metadata that it does? The problem is that this round-trip is *only* good for fp-fp?

I think we should standardize on the pandas side with ``index=True``

dargueta:
> I think we should standardize on the pandas side with `index=True`

That's technically an abrupt backwards-incompatible change. Personally, for minimal disruption I believe we should release this feature with `None` as the default, with an explicit note that `True` will be the default in the next backwards-incompatible release.

dargueta:
@jreback and @jorisvandenbossche is there anything else you think needs to be addressed?

doc/source/whatsnew/v0.24.0.txt
dargueta: > I think the default of fastparquet actually makes sense

Agreed. It doesn't really add information to the archive since the order of the rows is already known.

jorisvandenbossche (approved):
This looks good to me. 

@jreback any further comments?

We can always discuss later if we want to change the default of `index` to always be True or to something else (eg use the fastparquet logic ourselves)

doc/source/io.rst
```diff
@@ -4633,6 +4636,41 @@ Read only certain columns of a parquet file.
...
+
+Dumping a ``DataFrame`` to parquet may include the implicit index as one or more
```
jreback: can u add a version changed tag

jreback: not sure we use the word Dumping anywhere
use serialize

pandas/io/parquet.py
```diff
@@ -103,19 +103,27 @@ def __init__(self):
...
-        if self._pyarrow_lt_070:
+
+        # Only validate the index if we're writing it.
+        if self._pyarrow_lt_070 and index is not False:
             self._validate_write_lt_070(df)
         path, _, _, _ = get_filepath_or_buffer(path, mode='wb')
 
+        if index is None:
+            from_pandas_kwargs = {}
+        else:
```
jreback: can u always pass this? and just set index= if it’s None or does that change things?

jorisvandenbossche: I am not sure if `preserve_index=None` is a valid value for pyarrow (at least not according to the docstring)

jreback:
can you rebase

pandas/io/parquet.py
dargueta: I'd rather avoid messing with pyarrow like that, it's fairly finicky in my experience.

doc/source/io.rst
dargueta: Done

doc/source/io.rst
dargueta: True, but the goal of this section is mostly to illustrate use cases and how engines differ.

dargueta:
@jreback rebase is done

jorisvandenbossche:
@dargueta Thanks a lot!

PR #22273, opened 2018-08-10
Comments to understand code!

santosh2702:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@
 
 
 
-## What is it
+## What is it?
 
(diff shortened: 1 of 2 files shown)


WillAyd:
@sam2702 I don't think these comments add any value as they are pretty self explanatory. I also think you have mixed in code from another change.

If you want to make any changes to the code base you are better served opening an issue first so it can be discussed as needed by contributors

sam2702:
Ok from next time i will remember it.

PR #18761, opened 2017-12-13
[WIP] implement tests using hypothesis

jbrockmendel:
[skipci]

Related: #17978 

Exposes a whole bunch of broken cases.  Many of them look like they are not currently caught because the large majority of offset tests use tz-naive pydatetime inputs.

Note this does _not_ fix these bugs.  The goal here is a Proof Of Concept for using hypothesis and discuss if/how we can make it useful.

It also doesn't add hypothesis to the requirements because I have no idea what the appropriate file is for that.

PR #22280, opened 2018-08-11
Add initial property-based tests using Hypothesis

### LINKED ISSUE #17978: Investigate using Hypothesis for some tests

Zac-HD:
This pull request has no features or bugfixes for Pandas itself (though it does have a whatsnew entry for contributors), but adds support for and actually adds a few tests using Hypothesis.  The new section of the contributing guide below should explain what this means and why it's desirable.  There are basically three contributions here:

1. Contributor documentation and all the dependency setup to support Hypothesis tests in CI.
2. A pleasantly concise pair of passing tests in `test_ticks.py`.
3. Three *failing* tests in `test_offsets_properties.py` based on #18761 by @jbrockmendel.  I've made these idiomatic from the Hypothesis side, but while it seems likely that at least `test_on_offset_implementations` might be finding real bugs fixing them is beyond me at the moment.  I thought it worth including them anyway, with an `xfail` decorator.

Future work on Hypothesis tests could focus on such fruitful areas as round-tripping data via any of the serialisation formats Pandas offers (I tried to write a demo of this and couldn't get it to pass), or the melt and merge tests that @TomAugspurger mentioned.

Closes #17978, after which more specific issues could be opened with suggested tests.

### DIFF
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -820,6 +820,8 @@ Tests that we have ``parametrized`` are now accessible via the test name, for ex
    test_cool_feature.py::test_series[int8] PASSED
 
 
+.. _using-hypothesis:
+
 Using ``hypothesis``
 ~~~~~~~~~~~~~~~~~~~~
 
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -820,6 +820,44 @@ Tests that we have ``parametrized`` are now accessible via the test name, for ex
    test_cool_feature.py::test_series[int8] PASSED
 
 
+Using ``hypothesis``
+~~~~~~~~~~~~~~~~~~~~
+
+Hypothesis is a library for property-based testing.  Instead of explicitly
+parametrizing a test, you can describe *all* valid inputs and let Hypothesis
+try to find a failing input.  Even better, no matter how many random examples
+it tries, Hypothesis always reports a single minimal counterexample to your
+assertions - often an example that you would never have thought to test.
+
+See `Getting Started with Hypothesis <https://hypothesis.works/articles/getting-started-with-hypothesis/>`_
+for more of an introduction, then `refer to the Hypothesis documentation
+for details <https://hypothesis.readthedocs.io/en/latest/index.html>`_.
+
+.. code-block:: python
+
+    import json
+    from hypothesis import given, strategies as st
+
+    any_json_value = st.deferred(lambda: st.one_of(
+        st.none(), st.booleans(), st.floats(allow_nan=False), st.text(),
+        st.lists(any_json_value), st.dictionaries(st.text(), any_json_value)
+    ))
+
+    @given(value=any_json_value)
+    def test_json_roundtrip(value):
+        result = json.loads(json.dumps(value))
+        assert value == result
+
+This test shows off several useful features of Hypothesis, as well as
+demonstrating a good use-case: checking properties that should hold over
+a large or complicated domain of inputs.
+
+To keep the Pandas test suite running quickly, parametrized tests are
+preferred if the inputs or logic are simple, with Hypothesis tests reserved
+for cases with complex logic or where there are too many combinations of
+options or subtle interactions to test (or think of!) all of them.
+
+
 Running the test suite
 ----------------------
 
--- a/pandas/conftest.py
+++ b/pandas/conftest.py
@@ -450,3 +450,37 @@ def mock():
         return importlib.import_module("unittest.mock")
     else:
         return pytest.importorskip("mock")
+
+
+# ----------------------------------------------------------------
+# Global setup for tests using Hypothesis
+
+from hypothesis import strategies as st
+
+# Registering these strategies makes them globally available via st.from_type,
+# which is use for offsets in tests/tseries/offsets/test_offsets_properties.py
+for name in 'MonthBegin MonthEnd BMonthBegin BMonthEnd'.split():
+    cls = getattr(pd.tseries.offsets, name)
+    st.register_type_strategy(cls, st.builds(
+        cls,
+        n=st.integers(-99, 99),
+        normalize=st.booleans(),
+    ))
+
+for name in 'YearBegin YearEnd BYearBegin BYearEnd'.split():
+    cls = getattr(pd.tseries.offsets, name)
+    st.register_type_strategy(cls, st.builds(
+        cls,
+        n=st.integers(-5, 5),
+        normalize=st.booleans(),
+        month=st.integers(min_value=1, max_value=12),
+    ))
+
+for name in 'QuarterBegin QuarterEnd BQuarterBegin BQuarterEnd'.split():
+    cls = getattr(pd.tseries.offsets, name)
+    st.register_type_strategy(cls, st.builds(
+        cls,
+        n=st.integers(-24, 24),
+        normalize=st.booleans(),
+        startingMonth=st.integers(min_value=1, max_value=12)
+    ))
--- a/pandas/tests/tseries/offsets/test_offsets_properties.py
+++ b/pandas/tests/tseries/offsets/test_offsets_properties.py
@@ -1,0 +1,133 @@
+# -*- coding: utf-8 -*-
+"""
+Behavioral based tests for offsets and date_range.
+
+This file is adapted from https://github.com/pandas-dev/pandas/pull/18761 -
+which was more ambitious but less idiomatic in its use of Hypothesis.
+
+You may wish to consult the previous version for inspiration on further
+tests, or when trying to pin down the bugs exposed by the tests below.
+"""
+
+import pytest
+from hypothesis import given, assume, strategies as st
+from hypothesis.extra.pytz import timezones as pytz_timezones
+from hypothesis.extra.dateutil import timezones as dateutil_timezones
+
+import pandas as pd
+
+from pandas.tseries.offsets import (
+    Hour, Minute, Second, Milli, Micro, Nano,
+    MonthEnd, MonthBegin, BMonthEnd, BMonthBegin,
+    QuarterEnd, QuarterBegin, BQuarterEnd, BQuarterBegin,
+    YearEnd, YearBegin, BYearEnd, BYearBegin,
+)
+
+
+tick_classes = [Hour, Minute, Second, Milli, Micro, Nano]
+yqm_classes = [MonthBegin, MonthEnd, BMonthBegin, BMonthEnd,
+               QuarterBegin, QuarterEnd, BQuarterBegin, BQuarterEnd,
+               YearBegin, YearEnd, BYearBegin, BYearEnd]
+
+# ----------------------------------------------------------------
+# Helpers for generating random data
+
+gen_date_range = st.builds(
+    pd.date_range,
+    start=st.datetimes(
+        # TODO: Choose the min/max values more systematically
+        min_value=pd.Timestamp(1900, 1, 1).to_pydatetime(),
+        max_value=pd.Timestamp(2100, 1, 1).to_pydatetime()
+    ),
+    periods=st.integers(min_value=2, max_value=100),
+    freq=st.sampled_from('Y Q M D H T s ms us ns'.split()),
+    tz=st.one_of(st.none(), dateutil_timezones(), pytz_timezones()),
+)
+
+gen_random_datetime = st.datetimes(
+    min_value=pd.Timestamp.min.to_pydatetime(),
+    max_value=pd.Timestamp.max.to_pydatetime(),
+    timezones=st.one_of(st.none(), dateutil_timezones(), pytz_timezones())
+)
+
+# Register the various offset classes so st.from_type can create instances.
+# We *could* just append the strategies to a list, but this provides a nice
+# demo and enables future tests to use a simple e.g. `from_type(Hour)`.
+for cls in tick_classes + [MonthBegin, MonthEnd, BMonthBegin, BMonthEnd]:
+    st.register_type_strategy(cls, st.builds(
+        cls,
+        n=st.integers(-99, 99),
+        normalize=st.booleans(),
+    ))
+
+for cls in [YearBegin, YearEnd, BYearBegin, BYearEnd]:
+    st.register_type_strategy(cls, st.builds(
+        cls,
+        n=st.integers(-5, 5),
+        normalize=st.booleans(),
+        month=st.integers(min_value=1, max_value=12),
+    ))
+
+for cls in [QuarterBegin, QuarterEnd, BQuarterBegin, BQuarterEnd]:
+    st.register_type_strategy(cls, st.builds(
+        cls,
+        n=st.integers(-24, 24),
+        normalize=st.booleans(),
+        startingMonth=st.integers(min_value=1, max_value=12)
+    ))
+
+# This strategy can generate any kind of Offset in `tick_classes` or
+# `yqm_classes`, with arguments as specified directly above in registration.
+gen_yqm_offset = st.one_of([st.from_type(cls) for cls in yqm_classes])
+
+
+# ----------------------------------------------------------------
+# Offset-specific behaviour tests
+
+
+# Based on CI runs: Always passes on OSX, fails on Linux, sometimes on Windows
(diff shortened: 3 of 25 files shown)


pandas/tests/tseries/offsets/test_offsets_properties.py
```diff
@@ -0,0 +1,129 @@
+# -*- coding: utf-8 -*-
+"""
+Behavioral based tests for offsets and date_range.
+
+This file is adapted from https://github.com/pandas-dev/pandas/pull/18761 -
+which was more
```
jbrockmendel: you stopped talking in the middle of a

jbrockmendel:
This looks really nice, definitely prettier than #18761.

Do the xfailed tests actually fail or is that just being defensive?  Ideally we'd like to have `strict=True` in xfails.

Zac-HD:
Thanks!  I have an an unfair advantage in Hypothesis idioms, but I couldn't have written anything like this without your example to follow on the Pandas side :smile: 

The tests really do fail - I've added `strict=True` to the xfails.  Hopefully someone who knows more about Pandas internals than I do can either improve the tests or actually fix the bugs!

TomAugspurger:
Agreed this looks quite nice, thanks @Zac-HD .

For better or worse, our test dependencies are all optional other than pytest. I think hypothesis should follow that pattern. If others agree with that policy, I suppose the easiest way to do that would be to keep tests using hypothesis in a separate file, and `pytest.importorskip('hypothesis')`, or add a helper in `pandas.util.testing`.

I'll look at those xfailing tests later.

Zac-HD:
In principle I don't have any particular objections to making Hypothesis optional and keeping it in particular files - that's how it works for Xarray, and creating a new sub-directory `tests/hypothesis/` isn't hard.

The only catch is that they should be executed as part of *every* test run in CI!  Otherwise you miss all the platform-specific issues that it could turn up, including the thing where `test_on_offset_implementations` has to be marked `strict=False` because it only fails on *some* operating systems.

If you feel that property-based tests are going to be valuable - and there are obvious applications to testing e.g. round-tripping serialization that I couldn't get to pass - personally I'd just make Hypothesis a lightweight required dependency for development.  Happy to let you decide either way though!

gfyoung:
@TomAugspurger : I think `hypothesis` could be quite valuable for testing purposes.  However, I would play conservative first and make this optional as you suggested.  If it proves itself to be very important for testing purposes, then we could consider making it required later on.

Zac-HD:
OK, optional it is then! Do I just list the dependency in `ci/requirements-optional-conda.txt`?  Or is there something more involved?  

And just to make this *more* fun, it seems like the conda-forge package doesn't work for Python3 under OSX... cc @jochym for help please 😨

TomAugspurger:
@Zac-HD namespace collision I think :/ Seems to be `hypothesis-python`, at least on conda-forge.

Our CI setup is a bit messy, but I think

1. Include `hypothesis-python` in `ci/environment-dev.yaml`
2. Include a rename in `scripts/convert_deps.py` from `hypothesis-python` to `hypothesis`
3. Run `python scripts/convert_deps.py` and stage / commit the files
4.  Update some of the CI files to include `hypothesis-python`
  a. circle-27-compat.yaml
  b. circle-36-locale.yaml
  c. travis-27.yaml
  c. travis-35-osx.yaml
  d. travis-37.yaml
  e. appveyor-27.yaml
  f. appveyor-36.yaml

doc/source/contributing.rst
```diff
@@ -795,6 +795,44 @@ Tests that we have ``parametrized`` are now accessible via the test name, for ex
    test_cool_feature.py::test_series[int8] PASSED
```
jreback: add a sub-section ref

pandas/tests/tseries/offsets/test_offsets_properties.py
```diff
@@ -0,0 +1,133 @@
...
+
+import pytest
+from hypothesis import given, assume, strategies as st
+from hypothesis.extra.pytz import timezones as pytz_timezones
+from hypothesis.extra.dateutil import timezones as dateutil_timezones
+
+import pandas as pd
+
+from pandas.tseries.offsets import (
+    Hour, Minute, Second, Milli, Micro, Nano,
```
jreback: these already exist as fixtures, i'd like to avoid specifically importing things like this and use the fixture / generating fixture instead

pandas/tests/tseries/offsets/test_offsets_properties.py
```diff
@@ -0,0 +1,133 @@
...
+    YearEnd, YearBegin, BYearEnd, BYearBegin,
+)
+
+
+tick_classes = [Hour, Minute, Second, Milli, Micro, Nano]
+yqm_classes = [MonthBegin, MonthEnd, BMonthBegin, BMonthEnd,
+               QuarterBegin, QuarterEnd, BQuarterBegin, BQuarterEnd,
+               YearBegin, YearEnd, BYearBegin, BYearEnd]
+
+# ----------------------------------------------------------------
```
jreback: this is the problem I had with the original PR. There is a lot of code to create the things that need testing (e.g. the case generating statements).

Putting all of this inside a specific test file makes this very hard to re-use / document. Is there a standard like a conftest.py where these typically exist? We can create infrastructure to do this, but a) want to share as much of this code as possible, and b) put it in standard, well-defined, easy to discover locations.

TomAugspurger:
> Hmm, I've just checked and I don't think the name is the issue after all

Ah, sorry I misread the travis output. Ignore that.

> And it's currently listed in ci/environment-dev.yaml

That's perfect. Could you also run `python scripts/convert_deps.py` and include the output (for our contributors using pip).

pandas/tests/tseries/offsets/test_ticks.py
```diff
@@ -35,6 +36,39 @@ def test_delta_to_tick():
     assert (tick == offsets.Day(3))
 
 
+@given(cls=st.sampled_from(tick_classes),
```
jreback: you should be able to remove tests which this duplicates.

pandas/tests/tseries/offsets/test_ticks.py
```diff
@@ -35,6 +36,39 @@ def test_delta_to_tick():
...
+    expected = cls(n + m)
+
+    assert left + right == expected
+    assert left.apply(right) == expected
+
+    expected = cls(n - m)
+    assert left - right == expected
+
+
+@given(cls=st.sampled_from(tick_classes),
```
jreback: do these generate individual tests like pytest params?

jreback (changes requested):

pandas/tests/tseries/offsets/test_ticks.py
Zac-HD: No, it's only one test at runtime - `@given(...)` creates a wrapper which handles the details of repeatedly calling the inner test.  It also shrinks each error it sees to the minimal example, and will report multiple errors if it saw more than one (distinguished by exception type and line of code).

pandas/tests/tseries/offsets/test_ticks.py
Zac-HD: Sure; I'll convert them to explicit inputs with the `@example(...)` decorator :smile:

pandas/tests/tseries/offsets/test_offsets_properties.py
Zac-HD: This is actually a non-trivial trade-off!  Because we can't call fixtures outside of a test, we'll still need to import these types to register the strategy for each of them in `conftest.py` (see next comment).

In the test file, we have two options:
1. Import them, and let Hypothesis decide which to try.
2. Use a fixture which creates a test for each, and have Hypothesis try cases for each independently.

(1) is faster at runtime because Hypothesis is running one instead of *n* tests, while (2) is more explicit if a little less elegant.  Your call.

pandas/tests/tseries/offsets/test_offsets_properties.py
Zac-HD: - Global setup using `register_type_strategy` could easily be moved to `conftest.py` (heuristic: "equivalent to defining fixtures"), with a comment at each end explaining where the other part is.

- For the `gen_*` variables, I'd actually leave them right where they are - they look fairly specialised to these tests, and strategies would be idiomatically constructed within the call to `@given` if they were shorter or unique.
  Common bits can be extracted out as they get identified, but TBH we'd love to supply anything useful downstream in `hypothesis.extra.pandas` so some of that will hopefully vanish again :wink:

doc/source/contributing.rst
Zac-HD: Done.

jbrockmendel:
@Zac-HD Is there anything else from #18761 worth trying to salvage?

Zac-HD:
@jbrockmendel - it looks like you had some neat ideas for further tests of e.g. timedeltas, but they weren't developed enough for me to adopt here.  That would make a fantastic follow-up PR though, and I'd be happy to advise on Hypothesis idioms (a standing offer for any open source project).

Zac-HD:
If someone could restart the errored Travis jobs that would be great - it looks like they failed to download the Python 3.5 bundle and didn't even start running the tests 😭

pandas/conftest.py
```diff
@@ -453,3 +453,37 @@ def mock():
         return importlib.import_module("unittest.mock")
     else:
         return pytest.importorskip("mock")
+
+
+# ----------------------------------------------------------------
+# Global setup for tests using Hypothesis
+
+from hypothesis import strategies as st
```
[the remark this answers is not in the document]
Zac-HD: My read of the reviews so far is that @jreback was in favor of a mandatory dependency (also my recommendation), and you're now in favor too.

I've therefore made the relevant changes and it's all ready to go :tada: 

(though one build on Travis has errored out, the tests passed until the timeout)

TomAugspurger:
Merged master. The travis timeout should be fixed now.

jreback:
let me have a look

pandas/conftest.py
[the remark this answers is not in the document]
jreback: so, I still think we need to a) remove hypothesis from 1 build (the same one we have removed ``moto`` from is good). and use ``pyimportor.skip('hypthoesis')``. The reason is not for our CI really, rather so when a user does ``pd.test()`` is doesn't fail, rather it will just skip those tests.

pandas/util/_tester.py
```diff
@@ -12,6 +12,10 @@ def test(extra_args=None):
         import pytest
     except ImportError:
         raise ImportError("Need pytest>=3.0 to run tests")
+    try:
```
jreback: remove this here

jreback (changes requested):

pandas/conftest.py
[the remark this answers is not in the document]
Zac-HD: @jreback there are two problems with making Hypothesis optional for `pd.test()`:

1. It makes adding further Hypothesis tests - eg for serialisation round-trips, timedeltas, or reshaping logic - much harder.  They'd have to be in separate files, guard any global setup and configuration, handle import-or-skips, etc.
2. It forces us to choose to either duplicate tests, or skip them at runtime.

That doesn't make it completely unreasonable, I'd prefer to just have the dependency - and I've been using Pandas for much longer than Hypothesis!

TLDR - what's wrong with putting Hypothesis in the same category as pytest?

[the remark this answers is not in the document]
TomAugspurger: I think moto is a bit different, since it's relatively unimportant to mainline pandas, and so is easy to work around.

IMO, hypothesis should be treated the same as pytest.

Zac-HD:
@jreback my and Tom's view is that we're best off with Hypothesis as a hard dependency to run tests.  Can you live with this and give an approving review?  If not, what problems do you see?

(if possible I'd really like to merge this Friday or Saturday so I can talk about it at PyCon Australia and get some follow-up work done at the sprints)

jreback:
it still can be a hard dependency for our testing

but as i said users code will immediately fail if they only have pytest

for user code that’s all that we can / should require 

I am not sure why hisbis a problem
you simply put a skip if in the conftest itself
if it’s not installed

TomAugspurger:
What's the issue with having it as a required test dependency? Having to write tests around that constraint seems unnecessarily burdensome on contributors / maintainers.

jreback:
I guess its ok to add as a dep for testing. can you update the install.rst as well.

Zac-HD:
@TomAugspurger - I'm quite certain these build issues are unrelated to the pull.  Any idea on how to make it green?

TomAugspurger:
Hopefully fixed by https://github.com/pandas-dev/pandas/pull/22499

Zac-HD:
:tada: - it's working!  @TomAugspurger, if you can hit 'merge' in the next, uh, five hours (!) that will make my slides for PyCon Australia a little more precise :wink:

TomAugspurger:
Thanks!

h-vetinari:
After rebasing #22236, I'm getting a failure related to this PR in the `circleci/py36_locale` job:

```
=================================== FAILURES ===================================
___________________________ test_tick_add_sub[Nano] ____________________________

cls = <class 'pandas.tseries.offsets.Nano'>

    @pytest.mark.parametrize('cls', tick_classes)
>   @example(n=2, m=3)
    @example(n=800, m=300)
    @example(n=1000, m=5)
    @given(n=st.integers(-999, 999), m=st.integers(-999, 999))
    def test_tick_add_sub(cls, n, m):
E   hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 2 valid examples in 1.09 seconds (0 invalid ones and 0 exceeded maximum size). Try decreasing size of the data you're generating (with e.g.max_size or max_leaves parameters).
E   See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.too_slow to the suppress_health_check settings for this test.

pandas/tests/tseries/offsets/test_ticks.py:40: FailedHealthCheck
---------------------------------- Hypothesis ----------------------------------
You can add @seed(183585162026757523236170973153487462479) to this test or run pytest with --hypothesis-seed=183585162026757523236170973153487462479 to reproduce this failure.
```

h-vetinari:
And again, this time in another test, but same build job (https://circleci.com/gh/pandas-dev/pandas/18150):

```
=================================== FAILURES ===================================
___________________________ test_tick_add_sub[Milli] ___________________________

cls = <class 'pandas.tseries.offsets.Milli'>

    @pytest.mark.parametrize('cls', tick_classes)
>   @example(n=2, m=3)
    @example(n=800, m=300)
    @example(n=1000, m=5)
    @given(n=st.integers(-999, 999), m=st.integers(-999, 999))
    def test_tick_add_sub(cls, n, m):
E   hypothesis.errors.FailedHealthCheck: Data generation is extremely slow: Only produced 2 valid examples in 1.14 seconds (0 invalid ones and 0 exceeded maximum size). Try decreasing size of the data you're generating (with e.g.max_size or max_leaves parameters).
E   See https://hypothesis.readthedocs.io/en/latest/healthchecks.html for more information about this. If you want to disable just this health check, add HealthCheck.too_slow to the suppress_health_check settings for this test.

pandas/tests/tseries/offsets/test_ticks.py:40: FailedHealthCheck
---------------------------------- Hypothesis ----------------------------------
You can add @seed(188896409658629737425924686489321258929) to this test or run pytest with --hypothesis-seed=188896409658629737425924686489321258929 to reproduce this failure.
```

I think this healthcheck is too restrictive. If the server is busy for that single second that the samples are generated, it might already be triggered... The error message already mentions how to disable it, maybe this should be considered as a follow-up...?

Zac-HD:
Ah, looks like a CI clock issue.  I'll open a follow-up PR in the next day with the following in conftest.py:

```python
from hypothesis import settings, HealthCheck

settings.default.suppress_health_check = (HealthCheck.TooSlow,)
```

PR #22282, opened 2018-08-11
BUG: Swap byteorder in constructor of Block Manager (#4737)

### LINKED ISSUE #4737: ER: should constructors check and raise if big-endian dtypes are detected?
related #2330, #2821

pretty easy to do; doesn't really take much time either (can be done in the block manager)

http://stackoverflow.com/questions/18599579/python-pandas-multi-index

could also easily convert them (which in theory could be expensive so should be a method)

MHC03:
- [x] closes #4737
- [x] tests added / passed (at least the test_internals.py)
- [ ] whatsnew entry

This is one possible approach to directly change the dtypes to the native byteorder in the constructor of the BlockManager. This code does it automatically, without the user deciding a swap, so I am not sure if this is a wise design decision.
Another concern is the speed of conversion into a pandas object. I didn't quite grasp the asv stuff, so maybe someone else can check for me (sorry!).
I will add something to whatsnew, after you have confirmed, that this method of doing it is okay.

Please tell me if I forgot to do something.


pandas/tests/internals/test_internals.py
```diff
@@ -832,6 +832,12 @@ def test_validate_bool_args(self):
             with pytest.raises(ValueError):
                 bm1.replace_list([1], [2], inplace=value)
 
+    def test_swap_to_native_byteorder(self):
+        native_byteorder = '='
```
gfyoung: Reference issue number right above this line.

pandas/core/internals/managers.py
```diff
@@ -106,7 +107,12 @@ class BlockManager(PandasObject):
 
     def __init__(self, blocks, axes, do_integrity_check=True):
         self.axes = [ensure_index(ax) for ax in axes]
-        self.blocks = tuple(blocks)
+
+        non_native_byteorder = '>' if sys.byteorder == 'little' else '<'
```
jreback: this comparison should be in pandas/compat and done at import time

pandas/core/internals/managers.py
```diff
@@ -106,7 +107,12 @@ class BlockManager(PandasObject):
...
+        self.blocks = tuple(block.astype(block.dtype.newbyteorder('='))
+                            if non_native_byteorder in str(block.dtype)
```
jreback: does this affect all block types? I would rather have a single check to avoid perf issues for this

jreback (changes requested):
does this actually work on non-little endian systems? we have NO testing of this at all.

pandas/core/internals/managers.py
MHC03: It won't affect the block types which are from custome pandas dtypes and all non-numeric dtypes like boolean. I stumbled upon the error in this issue, because I read a big endian formatted binary file, so everything is basically big endian. In those cases, a single check would actually suffice. Here I am going for the worst case, in case someone put in some mixed typed (little and big endian) object to convert it into a pandas object. The letter case is very unlikely, so I could just do a single check, if you want.

pandas/core/internals/managers.py
MHC03: Should I just add a function like get_non_native_byteorder() containing this line in the init of pandas/compat? Or should I rather use the is_platform_little_endian() function from compat here?

MHC03:
The code is not specific for little endian. I am not sure how we should test this on non-little-endian systems, is it possible through Travis or other ci's?

pandas/core/internals/managers.py
jreback: ok just do this slightly differently

```
if is_platform_little_endian():
     self.blocks = .....
else:
         # original code here
    self.blocks = ...
```

trying to avoid *any* perf hit for little endian

jreback (changes requested):

jreback:
closing as stale. ping if you'd like to continue.

Issue #21841, opened 2018-07-10
inplace option for series.reindex()


jgerlach93:
is there a specific reason why this convenient option doesnt exist?

Issue #21859, opened 2018-07-11
DEPR: combine_first (replace with update(..., join='outer'); for both Series/DF)


h-vetinari:
I always found the mechanics of `combine_first` very unintuitive, and constantly need to look into the docs to see what's happening. I haven't checked the git history, but it seems that the method was a direct response from wesm to a SO question (https://stackoverflow.com/a/9794891). In particular, I think this would be much more intuitive to do with `df.update`, which is a subset of what #21855 proposes -- it introduces `join='outer'` for `DataFrame.update` (currently, only `'left'` is supported, but even the source code notes `# TODO: Support other joins`).

With that new option, `df1.combine_first(df2)` would be the same as `df1.update(df2, join='outer', overwrite=False)`, only that `combine_first` has much fewer options and controls (i.e. `filter_func` and `raise_conflict)`. The only difference is that `df.update` currently returns not, see #21858.

Since it's quite a well-established function, the deprecation cycle would maybe have to be longer than usual, but I think the `update` path much cleaner and general than this single-purpose function.

gfyoung:
cc @jreback

shoyer:
One downside of `DataFrame.update()` vs `combine_first()` is that it modifies the dataframe inplace rather than having a return value.

Just as a point of reference, in xarray we implement `combine_first()` in terms of an outer join + `fillna`:
```python
    def combine_first(self, other):
        """Combine two DataArray objects, with union of coordinates.

        This operation follows the normal broadcasting and alignment rules of
        ``join='outer'``.  Default to non-null values of array calling the
        method.  Use np.nan to fill in vacant cells after alignment.

        Parameters
        ----------
        other : DataArray
            Used to fill all matching missing values in this array.

        Returns
        -------
        DataArray
        """
        return ops.fillna(self, other, join="outer")
```

h-vetinari:
@shoyer 
> One downside of DataFrame.update() vs combine_first() is that it modifies the dataframe inplace rather than having a return value.

I agree, which is why I opened #21858 and liked it with the issue here... :)

jorisvandenbossche:
While looking through https://github.com/pandas-dev/pandas/issues/22812, I was also thinking about how `combine_first` is basically `fillna`, but without the outer alignment. So basically what @shoyer said :) 

Adding a join or align keyword to `fillna` might also be an option.

jorisvandenbossche:
[...]

PR #22286, opened 2018-08-11
ENH: add inplace-kwarg to df.update

### LINKED ISSUE #21858: ENH: add inplace-kwarg to update
Currently, `df.update` follows the convention of `dict.update` to return None and update inplace. This is against the prevailing trend (and philosopy?) of pandas to move away from `inplace`. See @TomAugspurger's response in #21841:
> Generally, we're moving away from inplace operations. It's confusing whether inplace means no copy or not. Reindex, by definition, can't be inplace unless the index is the same.
We recommend chaining your method calls, and hope to provide better memory control in the future.

The `update`-method of perfoms an important function regardless of whether it returns the object or None, and so should IMO be enabled to work in chained operations as well.

First step there would be adding an `inplace`-argument with default `False`, and then transitioning with a longish deprecation cycle towards `inplace=False`.

Relevant xrefs: #21855

h-vetinari:
- [x] closes #21858
- [x] tests added / passed
- [x] whatsnew entry

There wasn't a lot of discussion yet in #21858, so I thought I'd speed things up with a PR. So far:
* @gfyoung @toobaz (in https://github.com/pandas-dev/pandas/issues/21855#issuecomment-411778262) are +1 
* @TomAugspurger @Liam3851 sounded more or less -1.

While it may not be desirable to break the python dict-default of `update` being inplace (this PR leaves the default unchanged), I think it's very relevant for chaining, and (IMO) pandas' philosophy in general.

In #21858, I quoted @TomAugspurger (commenting in #21841 about *not* adding an option to inplace), which I find a good statement:
> Generally, we're moving away from inplace operations. It's confusing whether inplace means no copy or not. [...] We recommend chaining your method calls, and hope to provide better memory control in the future.

I find this applies just as much to `df.update`, hence the PR.


pandas/core/frame.py
```diff
@@ -5214,6 +5214,9 @@ def update(self, other, join='left', overwrite=True, filter_func=None,
         raise_conflict : bool, default False
             If True, will raise a ValueError if the DataFrame and `other`
             both contain non-NA data in the same place.
+        inplace : bool, default True
+            If True, follows the convention by ``dict`` of updating inplace. If
```
jreback: versionadded tag

remove the commentary about dict

pandas/tests/frame/test_combine_concat.py
```diff
@@ -207,7 +210,10 @@ def test_update(self):
         other = DataFrame([[3.6, 2., np.nan],
                            [np.nan, np.nan, 7]], index=[1, 3])
 
-        df.update(other)
+        if inplace:
+            df.update(other, inplace=inplace)
+        else:
+            df = df.update(other, inplace=inplace)
```
jreback: this is not an actual test, you need to rename to something else and assert that the original is not changed

jreback (changes requested):

jreback:
I am really -1 on ``inplace`` generally. To the extent that this is actually doing inplace operations, I guess explict is better than implicit.

pandas/tests/frame/test_combine_concat.py
h-vetinari: that's a valid objection - though I haven't seen it anywhere on the tests I've worked on that had an inplace-kwarg.

h-vetinari:
> I am really -1 on `inplace` generally. To the extent that this is actually doing inplace operations, I guess explict is better than implicit.

This is one of the few(?) methods that's inplace by default - so to move *away* from that, the first step is introducing an inplace-kwarg in the first place... ;-)

h-vetinari:
@jreback  All green. I opened a follow-up for giving the same capabilities to `Series.update`, #22358.

Liam3851:
Going towards my comments on #21858 and the name "update" seeming wrong for a non-inplace method. It seems to me that there are a lot of methods which perform variations on the same functionality (join, combine, combine_first, update). It seems to me that non-inplace `update` is basically a slightly-modified version of `join` or `combine`. Perhaps deprecation of `update` and the addition of its specific features (e.g. filter_func) as optional parameters to one of the other non-inplace methods would be more appropriate?

h-vetinari:
@Liam3851 I think we see the world quite differently on that. To me, the essential functionality of `update` (like any other method) is the intended *result*, in this case a DF suitably updated with another one.

Whether that operation is inplace or not is purely a semantic distinction, and (IMHO) down to your habits - meaning no offense, of course. To illustrate my point, the exact same argument could be used to argue that `fillna` must work in place (hypothetical quote):
> To me, `fillna` connotes `inplace` -- you're filling the "holes" of *this* DataFrame, why should this return another object?

Regarding your suggestion to wrap this into other methods, I don't think it would work. `join` does something different, `combine_first` should be deprecated (#21859), and `combine` needs a `func`-argument. If there were a default function that would allow using `combine` as `update`, I could maybe see it work - all the existing keywords of `update` would work, plus it could be without `inplace` from the get-go. But I tried quickly writing such a function, and it essentially boils down to rewriting the essential functionality of `update` manually.

Finally, I find "combine" to be a much too general term, when the thought process a user might try to materialise is simply "update this DF with that DF, but don't overwrite" (or maybe do...). For that reason, I never used `combine` (resp. needed to), because `update` was always the more obvious choice.

Liam3851:
@h-vetinari To me "update" has a distinct connotation in terms of both dicts and SQL as changing the underlying data, but I agree the point is largely semantic. 

My personal experience as a longtime user is that I've hardly ever used `update`, while `combine_first` is bread-and-butter and has an analog in `xarray`, so I'm a pretty strong -1 on deprecating the latter (#21859) and replacing it with the former.

h-vetinari:
@jreback I incorporated the requested changes a week ago.

h-vetinari:
@jreback Two weeks of green. ;-)
Would appreciate some progress on this as it's blocking PRs for #22358 and #21855

jreback:
@h-vetinari there are a number of open questions on this API generally. maybe @TomAugspurger @jorisvandenbossche have some thoughts

h-vetinari:
@jreback OK cool, let's have a discussion. Can you specify which open questions you see?

h-vetinari:
@jreback @tomaugspurger @jorisvandenbossche

Do you see any downside to adding an `inplace` kwarg to `update`? That in and of itself should not be controversial - explicit is better than implicit, no?

The discussion about the joins can be continued in #21855.

h-vetinari:
@jreback @TomAugspurger @jorisvandenbossche @toobaz 

This PR has been green for 4 weeks and hasn't moved forward due to lack of input (the issue has been open another month on top of that with little discussion) - can I please have some guidance of how to proceed (then I'd also offer to tackle PRs for #22358 and #21855)?

As this PR suggests, I'd prefer to add an `inplace`-kwarg to `update`, but could imagine another option as well (quoting myself from the issue):

> Alternatively, if `update` is such a reserved name, one could think of having the required functionality (fusing two dataframes with given precedence and requirement for output dimensions) in a separate method called e.g. `df.coalesce`?

TomAugspurger:
I don't think my thoughts have changed since
https://github.com/pandas-dev/pandas/issues/21858#issuecomment-410675641

I don't think we should add an inplace keyword argument.

I'm fine with returning `self`, the updated dataframe, to aid with method
chaining.

h-vetinari:
@TomAugspurger 
Thanks for the response! I responded to your remark in the issue with the suggestion quoted above (of then having a different method for that functionality) - any comment about that? Especially in chained calls, I find it *not* desirable to have inplace operations (whether returning `self` or not).

TomAugspurger:
Why do you prefer not-inplace ops in method chains, or more precisely, why
does it matter?
Other than the first operation in a method chain, you aren't going to have
a reference to the output
of the previous step, so it shouldn't matter.

If you're starting with `.update`, maybe start with `.copy` as a workaround?

TomAugspurger:
Thanks. I think I would recommend your 3rd solution. I prefer that to adding additional keyword arguments or new methods.

h-vetinari:
@TomAugspurger 
> Thanks. I think I would recommend your 3rd solution. I prefer that to adding additional keyword arguments or new methods.

I get your point, just two comments:
* the `inplace`-ing would/will get even more complicated with allowing different joins for `update` (#21855)
* but perhaps even more importantly:
  > [...] does any `inplace` operation currently return something other than `None`?
  
  This was the reason I didn't understand your comment in the issue - I didn't get that you meant that the method should return something *despite* being `inplace`, as this goes against all (my) experience in python/pandas.

jreback:
@h-vetinari I don't mind the code changes you did, they look fine. The issue is that ``inplace=True`` is no-where the default anywhere in pandas. Changing this is simply not-tenable, w/o a deprecation cycle (e.g. adding the keyword defaulting to None and then showing a warning if its not passed).

So we have ``.update`` (in-place defaults) and ``.combine_first`` which is not very standard terminology. 

In an ideal world I think adding ``.coalesce`` is probably the right thing to do (does R use this term?).
which is basically a rename of ``.combine_first``, and deprecate ``.update``.

h-vetinari:
> yes and that is exactly the problem. We can't simply add an `inplace=True` keyword here, which I think would be way more confusing

I disagree that this would be more confusing (just more explicit), but this is moot anyway if we follow the `.coalesce` route. Should I open an issue to introduce `.coalesce` (with the capabilities of `update`; to be able to eventually deprecate `combine_first` and `update`)?

jreback:
@h-vetinari i think that would be ok, to get some more commentary on this, esp from @jorisvandenbossche and @TomAugspurger (and some off-line discussions that I had with @cpcloud )

jreback:
closing this. there are many issues about this, and this PR is stale.

h-vetinari:
It wasn't stale just stalled - it's a PR I haven't pushed because there are more pressing issues to me.

You mention "many" issues, but so far, no-one has given an actual example (aside from personal preference that `.update` should *always* mean inplace).

In any case, I'm letting this be, but tried to restart the conversation in #22812. Would appreciate if you could comment there. :)

PR #22287, opened 2018-08-11
[CLN] More Misc Cleanups in _libs

jbrockmendel:
cpdef --> cdef in a few cases
modernize string formatting
modernize for loop syntax

Should be orthogonal to other outstanding PR(s)


pandas/_libs/lib.pyx
```diff
@@ -647,7 +646,7 @@ def row_bool_subset(ndarray[float64_t, ndim=2] values,
                 out[pos, j] = values[i, j]
             pos += 1
 
-    return out
+    return out.base  # `.base` to access underlying np.ndarray
```
chris-b1: I know elsewhere we do `np.asarray(<memoryview>)` - are those equivalent?

jbrockmendel: Very close to equivalent.  It's a bit roundabout.

Above `out` is typed as a `float64_t[:, :]` memoryview, but then instantiated with 

```
out = np.empty((mask.sum(), k), dtype=np.float64)
```

In executing that line, cython creates the numpy array, then casts it as a memoryview `out`, and attaches the original ndarray as `out.base`.  So `out.base` is an attribute lookup to return an existing np.ndarray.

The performance penalty to `np.asarray` must be tiny; it might be worth it to do it that way to avoid confusion.

jreback: ahh yes, can we change these (separate PR)

jreback (changes requested):
looks fine. are you linting these? (obviously how you made some of the changes), is this thru an IDE? I know I have asked this a bunch of times.

jbrockmendel:
>  are you linting these? (obviously how you made some of the changes)

No, just periodically reading through them.  My vague plan for achieving linting is to get the code (at least some pyx modules) "close enough" to valid python to trick the linter.

jreback (changes requested):
if you'd fix the .base things, otherwise lgtm.

pandas/_libs/lib.pyx
```diff
@@ -329,10 +329,10 @@ def fast_zip(list ndarrays):
             Py_INCREF(val)
             PyArray_ITER_NEXT(it)
 
-    return result
+    return result.base  # `.base` to access underlying np.ndarray
```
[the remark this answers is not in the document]
jbrockmendel: Sure

jbrockmendel:
Had to revert some edits to fix persistent ResourceWarning in one Travis environment.  Will try to narrow down the culprit in the next pass.

jbrockmendel:
Ping (got “LGTM” a few days ago)

jreback (approved):

jreback:
thanks

PR #22289, opened 2018-08-12
Remove unused param/method

alimcmaster1:
- [y] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

-Remove unused param "op" in `_validate_for_numeric_unaryop`. ( Param remove from single call of function)

-Static method `_from_element` appears to be unused

### DIFF
--- a/ci/print_skipped.py
+++ b/ci/print_skipped.py
@@ -10,7 +10,7 @@ def parse_results(filename):
     root = tree.getroot()
     skipped = []
 
-    current_class = old_class = ''
+    current_class = ''
     i = 1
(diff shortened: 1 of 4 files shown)


gfyoung:
IMO the changes are semantically distinct enough that they should be broken up into separate PR's for evaluation.  @alimcmaster1 : what do you think?

cc @jreback

alimcmaster1:
Sure @gfyoung , happy to break it up makes it easier to review. 

See new PR's here:
#22321
#22323 
#22322
#22324

Ill close this one :)

PR #22298, opened 2018-08-12
Use a more helpful error message for invalid correlation methods in DataFrame.corr

dsaxton:
`DataFrame.corr` currently returns a `KeyError` for invalid correlation methods.  The proposed change would instead return a `ValueError` with an error message reminding the user of the valid correlation methods.

```
import pandas as pd
import numpy as np

df = pd.DataFrame(np.random.normal(size=(10, 3)))
df.corr(method="turkey")
```


TomAugspurger:
Thanks. Could you add a release note in 0.24.0, and a test to ensure that code is hit for bad inputs?

dsaxton:
I'm working on some additional modifications to the DataFrame correlation methods to address some potential bugs I found and close another existing issue, is it okay if I close this pull request and reopen another with all the changes?

jreback:
no pls don’t

put independent changes in independent PRs

pandas/core/frame.py
```diff
@@ -6682,6 +6682,9 @@ def corr(self, method='pearson', min_periods=1):
                         c = corrf(ac, bc)
                     correl[i, j] = c
                     correl[j, i] = c
+        else:
+            raise ValueError("method must be either 'pearson', "
+                             "'spearman' or 'kendall'")
```
jreback: add ``....{method} was supplied'.format(method=method)``

pandas/tests/frame/test_analytics.py
```diff
@@ -130,6 +130,11 @@ def test_corr_cov_independent_index_column(self):
             assert result.index is not result.columns
             assert result.index.equals(result.columns)
 
+    def test_corr_invalid_method(self):
+        df = pd.DataFrame(np.random.normal(size=(10, 2)))
+        with pytest.raises(ValueError):
```
jreback: can you check this with an assert_raises_regex (e.g. match the error)

add the gh issue number as a comment

jreback (changes requested):
can you also test / fix this on Series as well.

pandas/tests/frame/test_analytics.py
dsaxton: Will do.  Is the convention here to use the most specific regex that matches the error message, or just one that is reasonably specific?

jreback: reasonable is fine

jreback (changes requested):
small comments. ping on green.

dsaxton:
@jreback Looks like the tests passed.  Let me know if there are any other changes that should be made, if not, thanks for all your help.

TomAugspurger:
Thanks @dsaxton!

dsaxton:
@TomAugspurger Happy to help!

PR #22307, opened 2018-08-13
DOC: fixes to DataFrame.plot.area docstring

jorisvandenbossche:
Follow-up on https://github.com/pandas-dev/pandas/pull/20170

### DIFF
--- a/pandas/plotting/_core.py
+++ b/pandas/plotting/_core.py
@@ -3385,25 +3385,13 @@ class FramePlotMethods(BasePlotMethods):
         .. plot::
             :context: close-figs
 
-            >>> df = pd.DataFrame({
-            ...     'sales': [3, 2, 3, 9, 10, 6],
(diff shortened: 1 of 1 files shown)


pandas/plotting/_core.py
```diff
@@ -3414,9 +3402,8 @@ def area(self, x=None, y=None, **kwds):
             >>> df = pd.DataFrame({
             ...     'sales': [3, 2, 3],
             ...     'visits': [20, 42, 28],
-            ...     'day': ['Monday', 'Tuesday', 'Wednesday'],
-            ... }, index=pd.date_range(start='2018/01/01', end='2018/07/01',
-            ...                        freq='M'))
+            ...     'day': [1, 2, 3],
```
jorisvandenbossche: This is a bit unfortunate (because not that a useful example I think), but the string column did not work for me: it didn't add any actual labels to the x-axis.

PR #22318, opened 2018-08-13
DEPR: join_axes-kwarg in pd.concat

### LINKED ISSUE #21951: DEPR: join_axes-kwarg for pd.concat
The `join_axes` kwarg of `pd.concat` is not very clearly documented (took me several tries to get it to work), and its name is not very clear either -- it's actually about restricting the axes that are not being concatenated (i.e. would be `'outer'`-joined normally).

In particular, it is basically irrelevant with the deprecation of `Panel`, since there are no more ax_e_s (plural), only one non-concatenation ax_i_s.

Finally, with `reindex` and `reindex_like`, it is redundant as well:
```
one = pd.DataFrame([[0, 1], [2, 3]], columns=list('ab'))
two = pd.DataFrame([[10, 11], [12, 13]], index=[1, 2], columns=list('bc'))

## simulating 'right'-join for the non-concatenation axis
pd.concat([one, two], join='outer', axis=1, join_axes=two.index) # cryptic error message!
# AssertionError: length of join_axes must not be equal to 1

## only works with list-like join_axes
pd.concat([one, two], join='outer', axis=1, join_axes=[two.index])
#      a    b   b   c
# 1  2.0  3.0  10  11
# 2  NaN  NaN  12  13

## cleaner with reindex?
pd.concat([one, two], join='outer', axis=1).reindex(two.index)
#      a    b     b     c
# 1  2.0  3.0  10.0  11.0
# 2  NaN  NaN  12.0  13.0
```

Note that the dtype changes due to the intermediate object having `NaN`s in the rows, but this will be fixed by #21160 anyway. Only question is if performance would be much worse, if concatenating huge Series/DFs before selecting small index-subset.

h-vetinari:
- [x] closes #21951 
- [x] tests added / passed
- [x] whatsnew entry


pandas/core/reshape/concat.py
```diff
@@ -219,6 +220,11 @@ def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
         ...
     ValueError: Indexes have overlapping values: ['a']
     """
+    if join_axes is not None:
```
jreback: you would need to remove its use inside the function.

further this is called in several places in the code base they would need to be changed. Are current tests showing a deprecation warning?

jreback (changes requested):

pandas/core/reshape/concat.py
h-vetinari: > you would need to remove its use inside the function.

Um, isn't there a deprecation cycle between issuing the warning and removing the keyword?

> further this is called in several places in the code base they would need to be changed. Are current tests showing a deprecation warning?

It was called in 3 places. Changed those, now there shouldn't be any warnings anymore (will check against log of current travis run)

jreback: of course there is a deprecation cycle. I am not talking about the public API, rather the private usage of this.

h-vetinari: @jreback Then what do you want me to remove? I don't know all the possible uses that someone might have used `join_axes`, (e.g. replacing the call in `pandas/core/generic.py` with `reindex` would change the order of the index and then fail some tests), so replacing it internally should wait until complete deprecation, no?

jreback: if its a useless argument then let's actually remove it internally. If its really really hard then pls indicate that. But much better to completely remove the usage of it if possible.

h-vetinari: @jreback It's not useless per se, just redundant, obscure, and not well documented. I *could* replace its internal uses, but:
- I might silently break code
- the deprecation suggests to use `reindex` on the *result*, so changing something *within* the methdod is IMO pointless until it's ripped out completely.

h-vetinari: @jreback I still think that removing the internal uses might break things. Shouldn't the complete removal wait until 1.0?

pandas/core/frame.py
```diff
@@ -6614,11 +6614,11 @@ def _join_compat(self, other, on=None, how='left', lsuffix='', rsuffix='',
             if can_concat:
                 if how == 'left':
                     how = 'outer'
-                    join_axes = [self.index]
+                    return concat(frames, axis=1, join=how,
+                                  verify_integrity=True).reindex(self.index)
```
jreback: you can add ``copy=False`` to the ``.reindex``

pandas/core/generic.py
```diff
@@ -8848,7 +8848,7 @@ def describe_1d(data):
                 if name not in names:
                     names.append(name)
 
-        d = pd.concat(ldesc, join_axes=pd.Index([names]), axis=1)
+        d = pd.concat([x.reindex(names) for x in ldesc], axis=1, sort=False)
```
jreback: add ``copy=False``

pandas/core/groupby/generic.py
```diff
@@ -518,8 +518,10 @@ def _transform_general(self, func, *args, **kwargs):
                 applied.append(res)
 
         concat_index = obj.columns if self.axis == 0 else obj.index
-        concatenated = concat(applied, join_axes=[concat_index],
-                              axis=self.axis, verify_integrity=False)
+        other_axis = (self.axis + 1) % 2  # switches from 0 to 1 or from 1 to 0
```
jreback: don't use this, just use an if else

pandas/core/groupby/generic.py
```diff
@@ -518,8 +518,10 @@ def _transform_general(self, func, *args, **kwargs):
...
+        concatenated = concat(applied, axis=self.axis,
+                              verify_integrity=False).reindex(concat_index,
+                                                              axis=other_axis)
         return self._set_result_index_ordered(concatenated)
```
jreback: write this as 
```
concatenated = concat(....)
concatenated = concatenated.reindex(...)
```
use copy=False

pandas/core/reshape/concat.py
jreback: no if you are deprecating a public function, then we want to remove the implemenation now and its uses internally.

pandas/core/reshape/concat.py
```diff
@@ -219,6 +220,11 @@ def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
...
+        warnings.warn('The join_axes-keyword is deprecated. Use .reindex or '
+                      '.reindex_like on the result to achieve the same '
```
jreback: you don't need the 2nd sentence (performance)

pandas/tests/reshape/test_concat.py
```diff
@@ -729,6 +729,25 @@ def test_concat_categorical_empty(self):
         tm.assert_series_equal(pd.concat([s2, s1], ignore_index=True), exp)
         tm.assert_series_equal(s2.append(s1, ignore_index=True), exp)
 
+    def test_concat_join_axes(self, axis):
+        one = pd.DataFrame([[0., 1.], [2., 3.]], columns=list('ab'))
```
jreback: issue number

rename test to test_concat_join_axes_deprecated

jreback (changes requested):

pandas/core/reshape/concat.py
```diff
@@ -219,23 +220,39 @@ def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
...
+                      'functionality.', FutureWarning, stacklevel=2)
+        ndim = res.ndim
+        if len(join_axes) != ndim - 1:
+            raise AssertionError("join_axes must be a list of indexes of "
+                                 "length {length}".format(length=ndim - 1))
+        if ndim == 2:
+            other_axis = 1 if axis == 0 else 0  # switches between 0 & 1
+            res = res.reindex(join_axes[0], axis=other_axis)
+        else:  # 3 for Panel; Panel4D already deprecated
```
jreback: no need to mention Panel4D.

use elif ndim == 3 to be explicit

jreback (changes requested):
small comment, lgtm. ping on green.

jreback:
@jorisvandenbossche you had some comments on the issue. I am all for making things simpler.

h-vetinari:
@jorisvandenbossche 
I'm answering your comment from #21951 here, as this is further ahead:

>"but this will be fixed by 21160 anyway." -> that is long off to be the default integer type, so I don't think we should use that as an argument now

Fair enough, that was a misunderstanding on my part. But the dtype-question is a side-issue here, IMO

> "with reindex and reindex_like, it is redundant as well" -> I don't know the implementation, but I would assume that reindexing after the fact can be less performant? (assuming that with `join_axes` it reindexes each input before concatenating)

I posed this question in the OP ("Only question is if performance would be much worse, if concatenating huge Series/DFs before selecting small index-subset."), and it's a valid point. The user *could* apply a reindex directly to the arguments of `concat` (I originally had something to that effect in the `FutureWarning` in this PR), but I see that it would be good to keep having a way to do this easily.

> It is certainly true the way it is explained and spelled (eg the fact that you need to pass a list) is certainly outdated now Panel is removed. But we could also consider improving it.

In my opinion, the `join_axes`-keyword (as a name, not necessarily as functionality) should definitely be deprecated. The question of what to replace it with is very similar to the discussion in https://github.com/pandas-dev/pandas/issues/21855#issuecomment-413103937.

There might be a good solution for replacing the `join_axes`-functionality with something like the following (conceptually):
```
concat(objs, axis=0, join='outer', index=None, columns=None, ignore_index=False,
       keys=None, levels=None, names=None, verify_integrity=False, sort=None, copy=True)
[...]
if (axis == 0 and index is not None) or (axis == 1 and columns is not None):
   raise ValueError('Cannot set index of concatenation axis; '
                    'can only set index of non-concatenation axis')
index = join if index is None else index
columns = join if columns is None else columns
[...]
obs = [x.reindex(index/columns, axis=other_axis) for x in obs]
[...]
# actual concatenation
```
Then index the non-concatenation index of the result could be set as desired (essentially replacing `join_axes` with a cleaner API, now that `Panel` will be deprecated).

> (note that I am not married to the keyword, I have never used it myself, but just think we should have a bit more discussion about it. It would be interesting to search for usage on github/SO)

There's not much on SO - https://www.google.com/search?q=stackoverflow+pandas+concat+join_axes yields only one (sorta) relevant hit: https://stackoverflow.com/q/27391081

Neither did I find much of substance here: https://github.com/search?p=2&q=join_axes&type=Issues

h-vetinari:
@jreback 
> small comment, lgtm. ping on green.

All green (rebased due to persistent test failures).

h-vetinari:
@jreback

Green and all feedback incorporated. Anything missing?

pandas/core/generic.py
```diff
@@ -8930,7 +8930,8 @@ def describe_1d(data):
                 if name not in names:
                     names.append(name)
 
-        d = pd.concat(ldesc, join_axes=pd.Index([names]), axis=1)
+        d = pd.concat([x.reindex(names) for x in ldesc], axis=1,
```
jreback: use the ``copy=False`` on . the ``.reindex``. remove copy= from the ``pd.concat`` it is not respected misleading here.

pandas/core/groupby/generic.py
```diff
@@ -518,8 +518,11 @@ def _transform_general(self, func, *args, **kwargs):
                 applied.append(res)
 
         concat_index = obj.columns if self.axis == 0 else obj.index
-        concatenated = concat(applied, join_axes=[concat_index],
-                              axis=self.axis, verify_integrity=False)
+        other_axis = 1 if self.axis == 0 else 0  # switches between 0 & 1
+        concatenated = concat(applied, axis=self.axis,
```
jreback: remove copy= from the concat

pandas/core/frame.py
```diff
@@ -6513,12 +6513,13 @@ def _join_compat(self, other, on=None, how='left', lsuffix='', rsuffix='',
             # join indexes only using concat
             if can_concat:
                 if how == 'left':
-                    how = 'outer'
-                    join_axes = [self.index]
+                    res = concat(frames, axis=1, join='outer',
```
jreback: remove copy= from concat

pandas/core/reshape/concat.py
```diff
@@ -219,23 +220,39 @@ def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
...
                        keys=keys, levels=levels, names=names,
-                       verify_integrity=verify_integrity,
-                       copy=copy, sort=sort)
-    return op.get_result()
+                       verify_integrity=verify_integrity, copy=copy, sort=sort)
+
+    res = op.get_result()
+
+    if join_axes is not None:
+        warnings.warn('The join_axes-keyword is deprecated. Use .reindex or '
```
jreback: add the issue numbere here as a comment

pandas/core/reshape/concat.py
```diff
@@ -219,23 +220,39 @@ def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
...
+                      '.reindex_like on the result to achieve the same '
+                      'functionality.', FutureWarning, stacklevel=2)
+        ndim = res.ndim
+        if len(join_axes) != ndim - 1:
+            raise AssertionError("join_axes must be a list of indexes of "
+                                 "length {length}".format(length=ndim - 1))
+        if ndim == 2:
+            other_axis = 1 if axis == 0 else 0  # switches between 0 & 1
+            res = res.reindex(join_axes[0], axis=other_axis)
```
jreback: add copy=False to the .reindex

jreback (changes requested):
some comments. ping on green.

h-vetinari:
@jorisvandenbossche 

I answered above (https://github.com/pandas-dev/pandas/pull/22318#issuecomment-415448620) to your remarks/questions in the issue. Any further comments, API- or otherwise?

h-vetinari:
@jreback 
> some comments. ping on green.

ping on green

jreback:
@jorisvandenbossche i am ok with this
any objections

jorisvandenbossche:
@h-vetinari sorry for my slow follow-up, didn't have much time the last weeks

I also saw that the discussion here (about an alternative way to do it) is related to the joining discussions in other PRs/issues (you linked https://github.com/pandas-dev/pandas/issues/21855). I will try to take a look at those discussions one of the coming days.

I personally think we should first decide on an alternative (or decide upon if we want one) before actually deprecating this, but no strong opinion.

BTW, if we keep the PR as is, the documentation certainly needs an update (there are example of `join_axes` in the tutorial docs).

h-vetinari:
@jorisvandenbossche 
> I will try to take a look at those discussions one of the coming days.

Any luck with that?

Barring a decision on what to replace the API with, which tutorial docs do you mean? I didn't find anything in https://pandas.pydata.org/pandas-docs/stable/tutorials.html, but there's a bit in `doc/source/merging.rst`.

jreback (approved):

jreback:
looks ok to me, @jorisvandenbossche has some questions.

h-vetinari:
@jreback @jorisvandenbossche 
> Barring a decision on what to replace the API with, which tutorial docs do you mean? I didn't find anything in https://pandas.pydata.org/pandas-docs/stable/tutorials.html, but there's a bit in doc/source/merging.rst.

h-vetinari:
@jorisvandenbossche 
Can we pick this up again?

h-vetinari:
@gfyoung 
> @h-vetinari : FYI, instead of pushing commits to trigger entirely new builds, just ping one of us. We can restart individual builds for you.

Could you retrigger that timed-out travis job please? ;-)

jreback:
revisiting, can you rebase.

h-vetinari:
@jreback @jorisvandenbossche 
Rebased and green

h-vetinari:
@jreback 
> revisiting, can you rebase.

Merged master and green.

jreback:
@jorisvandenbossche last look. This has been in the queue for a while and is a valid deprecation.

h-vetinari:
@jreback @jorisvandenbossche 
Last ping (in case you consider this relevant for the RC).

WillAyd:
@h-vetinari can you merge master on this?

jorisvandenbossche:
Comment from above:

> BTW, if we keep the PR as is, the documentation certainly needs an update (there are example of `join_axes` in the tutorial docs).

Did you address that already?

As I said above (https://github.com/pandas-dev/pandas/pull/22318#issuecomment-419139173), I might prefer a conclusion first to the align axes discussion, but since that is also not very active / close to conclusion, I won't hold up this PR if the other core devs are in favor of merging it.

h-vetinari:
@jorisvandenbossche

I answered that exact point further up the thread:

> @h-vetinari: Barring a decision on what to replace the API with, which tutorial docs do you mean? I didn't find anything in https://pandas.pydata.org/pandas-docs/stable/tutorials.html, but there's a bit in doc/source/merging.rst.

pandas/core/reshape/concat.py
```diff
@@ -221,23 +223,40 @@ def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
...
+                      '.reindex_like on the result to achieve the same '
+                      'functionality.', FutureWarning, stacklevel=2)
+        ndim = res.ndim
+        if len(join_axes) != ndim - 1:
+            raise AssertionError("join_axes must be a list of indexes of "
+                                 "length {length}".format(length=ndim - 1))
+        if ndim == 2:
+            other_axis = 1 if axis == 0 else 0  # switches between 0 & 1
+            res = res.reindex(join_axes[0], axis=other_axis, copy=False)
+        elif ndim == 3:  # Panel
```
jreback: you can remove this section as Panel is removed

jreback (changes requested):

jorisvandenbossche:
> but there's a bit in doc/source/merging.rst.

Yes, exactly that (sorry, my "tutorial docs" was meaning the written rst docs (like merging.rst) to contrast it to docstrings, which was probably not clear :-))

h-vetinari:
@jorisvandenbossche 
Changed the part in `merging.rst`, PTAL.

jreback:
this deprecation looks ok to me. pls merge master.

jreback:
if @pandas-dev/pandas-core wants to finish this up

h-vetinari:
@jreback
Couldn't respond to your last ping at the time and then forgot. Merged master.

h-vetinari:
@jreback

This should be done. Could you please restart the timed-out travis job?

doc/source/user_guide/merging.rst
```diff
@@ -88,7 +88,9 @@ some configurable handling of "what to do with the other axes":
   concatenation axis does not have meaningful indexing information. Note
   the index values on the other axes are still respected in the join.
 * ``join_axes`` : list of Index objects. Specific indexes to use for the other
```
jreback: you can completely remove the join_axes here; its deprecated and does not need to appear here at all; its in the doc-string which is enough

pandas/core/frame.py
```diff
@@ -6911,12 +6911,13 @@ def _join_compat(self, other, on=None, how='left', lsuffix='', rsuffix='',
...
+                                 verify_integrity=True)
+                    res = res.reindex(self.index, copy=False)
```
jreback: you can just: ``return res.index(....)``

pandas/tests/resample/test_resample_api.py
```diff
@@ -233,6 +233,7 @@ def test_apply_without_aggregation():
         assert_series_equal(result, test_series)
 
 
+@pytest.mark.skipif(not PY36, reason='Insertion order for Python>=3.6')
```
jreback: is this needed?

jreback (changes requested):
small changes, otherwise lgtm. ping on green.

jorisvandenbossche:
From a discussion with @jreback: it might be useful to add a `join='left'` option for one use case of `join_axes` to use the index of the first dataframe in the list.

doc/source/user_guide/merging.rst
```diff
@@ -202,7 +203,7 @@ DataFrame:
 
 .. ipython:: python
 
-   result = pd.concat([df1, df4], axis=1, join_axes=[df1.index])
+   result = pd.concat([df1, df4], axis=1).reindex(df1.index)
```
jorisvandenbossche: Should we show the example of doing the reindex before the concat as well?

```
pd.concat([df1, df4.reindex(df1.index)], axis=1).reindex(df1.index)
```

pandas/core/reshape/concat.py
```diff
@@ -221,23 +223,35 @@ def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
...
+        warnings.warn('The join_axes-keyword is deprecated. Use .reindex or '
+                      '.reindex_like on the result to achieve the same '
+                      'functionality.', FutureWarning, stacklevel=2)
+        ndim = res.ndim
+        if len(join_axes) != ndim - 1:
+            raise AssertionError("join_axes must be a list of indexes of "
+                                 "length {length}".format(length=ndim - 1))
+        if ndim == 2:
+            other_axis = 1 if axis == 0 else 0  # switches between 0 & 1
+            res = res.reindex(join_axes[0], axis=other_axis, copy=False)
```
jorisvandenbossche: Should we move this reindexing to before passing the objs to `_Concatenator` ?

pandas/tests/resample/test_resample_api.py
jorisvandenbossche: How does the changes in this PR affect this case?

h-vetinari:
> @jorisvandenbossche: From a discussion with @jreback: it might be useful to add a `join='left'` option for one use case of `join_axes` to use the index of the first dataframe in the list.

I'm thinking - why don't we deprecate `join_axes` and *introduce* `join_axis` (note "e" vs "i"), which takes a list-like for the Index? This way, we can transparently transform legal `join_axEs`, and avoid the perf-hit, resp. usability loss of having to reindex before or after.

I would find this better than introducing `'left'` for something where one has to be creative about what "left" means in `pd.concat([df1, df2, df3, df4, df5])`. Is it first-left, or all-but-one-left, or whatever...?

About `@pytest.mark.skipif(not PY36, reason='Insertion order for Python>=3.6')`:
> > @jreback: is this needed?

> @jorisvandenbossche: How does the changes in this PR affect this case?

As far as I can see, the test has been passing by chance so far on PY35, because
```
expected = r[['A', 'B', 'C']].agg({'r1': 'mean', 'r2': 'sum'})
result = r.agg({'r1': 'mean', 'r2': 'sum'})
```
is not guaranteed to have the same order in PY35. I haven't touched anything regarding that code, but it did start failing [just now](https://dev.azure.com/pandas-dev/pandas/_build/results?buildId=13524&view=logs&jobId=d6a36677-9147-59f9-684d-001fc5c44c5e). All I've been doing on this PR is just merge master - and it's been passing for months...

h-vetinari:
That codecov-failure is clearly garbage. Let's see if a rebase (after 15x merging master) helps.

jorisvandenbossche:
@h-vetinari workflow related comment: can you not force push but only add commits? That makes reviewing and seeing what has changes easier for us.

jorisvandenbossche:
> As far as I can see, the test has been passing by chance so far on PY35, because

OK, I seem to remember seeing the failure before sometimes. But, if it is needed, let's not do that in this PR.

doc/source/user_guide/merging.rst
jreback: can you update this

doc/source/whatsnew/v0.25.0.rst
```diff
@@ -595,6 +595,7 @@ Other deprecations
 
 - The deprecated ``.ix[]`` indexer now raises a more visible ``FutureWarning`` instead of ``DeprecationWarning`` (:issue:`26438`).
 - Deprecated the ``units=M`` (months) and ``units=Y`` (year) parameters for ``units`` of :func:`pandas.to_timedelta`, :func:`pandas.Timedelta` and :func:`pandas.TimedeltaIndex` (:issue:`16344`)
+- :meth:`pandas.concat` has deprecated the ``join_axes``-keyword. Instead, use :meth:`DataFrame.reindex` or :meth:`DataFrame.reindex_like` on the result (:issue:`21951`)
```
jreback: or on the inputs

pandas/core/reshape/concat.py
```diff
@@ -221,23 +223,35 @@ def concat(objs, axis=0, join='outer', join_axes=None, ignore_index=False,
...
 class _Concatenator:
     """
     Orchestrates a concatenation operation for BlockManagers
     """
 
-    def __init__(self, objs, axis=0, join='outer', join_axes=None,
-                 keys=None, levels=None, names=None,
-                 ignore_index=False, verify_integrity=False, copy=True,
-                 sort=False):
+    def __init__(self, objs, axis=0, join='outer', keys=None, levels=None,
```
jreback: need to update the pd.concat doc-string as well

jreback (changes requested):

h-vetinari:
> @jorisvandenbossche: @h-vetinari workflow related comment: can you not force push but only add commits? That makes reviewing and seeing what has changes easier for us.

That's what I've been doing since last August here. But since now codecov is part of the CI suite and it gave a spurious failure, I rebased (without introducing new changes).

> > @h-vetinari: As far as I can see, the test has been passing by chance so far on PY35, because [...]

> @jorisvandenbossche: OK, I seem to remember seeing the failure before sometimes. But, if it is needed, let's not do that in this PR.

Why not? It's a single line of code and we just reviewed it together...

> @jreback: can you update this

Yes, I will respond to the review. But before that, would  you mind responding to my question [above](https://github.com/pandas-dev/pandas/pull/22318#issuecomment-506813819):
> I'm thinking - why don't we deprecate `join_axes` and *introduce* `join_axis` (note "e" vs "i"), which takes a list-like for the Index? This way, we can transparently transform legal `join_axEs`, and avoid the perf-hit, resp. usability loss of having to reindex before or after.

jorisvandenbossche:
> That's what I've been doing since last August here. But since now codecov is part of the CI suite and it gave a spurious failure, I rebased (without introducing new changes).

Hadn't seen your comment about before I posted that, sorry. Anyway, that should not matter for codecov. But eg if not all CI builds pass, that can also give failures on codecov I think.

> > @jorisvandenbossche: OK, I seem to remember seeing the failure before sometimes. But, if it is needed, let's not do that in this PR.
>
> Why not? It's a single line of code and we just reviewed it together...

separation of concerns. If it is not related, it doesn't need to be in this PR, otherwise you will get the confusion that this PR is causing a failure that needed to be skipped.

h-vetinari:
> jorisvandenbossche: Hadn't seen your comment about before I posted that, sorry. Anyway, that should not matter for codecov. But eg if not all CI builds pass, that can also give failures on codecov I think.

All the CI builds had passed, except for codecov (I imagine because it had forgotten the coverage of the commit where this PR had originally branched off)

> separation of concerns.

Fair enough...

jreback:
@h-vetinari 

in regards to: https://github.com/pandas-dev/pandas/pull/22318#issuecomment-506813819

this is a separate and distinct change. let's treat it as such. Keeping PRs nice and simple is the way to go.

h-vetinari:
Fair point, but without someone doing a PR, we'll have just a deprecation, without the replacement. I think it's equally fair to say that that replacement is part of the deprecation itself (to uphold existing usability where necessary). I'll be gone the next two weeks., but then, this PR is not in a hurry...

jreback:
@h-vetinari IMHO we don't need a replace at all.

jreback:
In any event, if you can respond to the remaining comments (soon) we can merge this.

pandas/core/reshape/concat.py
jreback: @h-vetinari can you update for this point

h-vetinari:
> @jreback: @h-vetinari can you update for this point

This PR needs updating in several different ways (also review from @jorisvandenbossche), and as I mentioned I'm currently away for two weeks.

I'd like to clean this up when I come back (realising that it'll be after 0.25), and properly replace the removed functionality with something like `join='outer'|'inner'|index-instance`.

jreback:
> > @jreback: @h-vetinari can you update for this point
> 
> This PR needs updating in several different ways (also review from @jorisvandenbossche), and as I mentioned I'm currently away for two weeks.
> 
> I'd like to clean this up when I come back (realising that it'll be after 0.25), and properly replace the removed functionality with something like `join='outer'|'inner'|index-instance`.

@h-vetinari that is a separate PR that may not be accepted 
pls just stick to single purpose things

TomAugspurger:
What's the status here? Is this happening for 0.25.0?

jreback:
prob needs a small amount of fixup; let me see if I can do this in the next couple of hours.

TomAugspurger:
Thanks (I'm roughly pushing for an RC tomorrow, maybe cutting it tomorrow
night)

h-vetinari:
@jreback @TomAugspurger

I wouldn't merge this before 0.25. It's been open for almost a year, so it's obviously not urgent. And I disagree that the direct replacement for the functionality we're deprecating should be a separate PR. If this gets merged before 0.25 there won't be replacement for that version (except the workaround of using `.reindex` manually).

OTOH, with a direct replacement in the same PR as the deprecation, one could formulate a clear path for updating.

jreback (approved):

jreback:
thanks @h-vetinari 

again, 1 PR 1 thing. The reason things take so long (not in this case), but generally is that they are too many things in 1 place.

jorisvandenbossche:
I have to say I agree with @h-vetinari that *if* we want something that replaces this functionality, we should not include this in 0.25.0 (unless we get the replacement in 0.25.0 as well). It is quite annoying to already start warning for something before a replacement is available (again, *if* we want to add such a replacement).

jreback:
there already *is* a very good alternative, ``.reindex()``, no need to re-invent the wheel here. This is less than useful in the non-Panel world. SO I am -1 on adding any replacement at all.

jorisvandenbossche:
It was from a discussion between the both of us that I commented https://github.com/pandas-dev/pandas/pull/22318#issuecomment-506781733 with the idea of a `join='first'` or `join='left'`, which is basically a replacement for the number 1 use case of `join_axes`. 
I thought you were OK with that idea.

But still, it is something to discuss. There is also the idea of introducing a `join_axis`, but I personally think that the join one covers most use cases of `join_axi/es`.

jreback:
> It was from a discussion between the both of us that I commented [#22318 (comment)](https://github.com/pandas-dev/pandas/pull/22318#issuecomment-506781733) with the idea of a `join='first'` or `join='left'`, which is basically a replacement for the number 1 use case of `join_axes`.
> I thought you were OK with that idea.
> 

I mean *maybe*, but I'd like to see an actual usecase. Removing API is easy, adding new should have a high bar.

h-vetinari:
@jreback
I get the maintainer POV, but from a usability POV it's the other way around - removing API is costly, while additions are easy. And it's not even a new API - it'd just be changing the name/semantics of existing functionality.

PR #22323, opened 2018-08-13
CLN: Remove unused param

alimcmaster1:
Param no longer used. Can remove it from single caller of this function

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -4695,7 +4695,7 @@ class Index(IndexOpsMixin, PandasObject):
         """ Update Index attributes (e.g. freq) depending on op """
         return attrs
 
-    def _validate_for_numeric_unaryop(self, op, opstr):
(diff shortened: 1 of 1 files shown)


pandas/core/indexes/base.py
```diff
@@ -4776,7 +4776,7 @@ def _make_evaluate_unary(op, opstr):
 
             def _evaluate_numeric_unary(self):
 
-                self._validate_for_numeric_unaryop(op, opstr)
+                self._validate_for_numeric_unaryop(opstr)
                 attrs = self._get_attributes_dict()
                 attrs = self._maybe_update_attributes(attrs)
                 return Index(op(self.values), **attrs)
```
jreback: op is used right here.

jreback (changes requested):
so rather than do this, I would make the signature more similar to ``....numeric_binop`` and pass op, then construct the ``opstr`` inside.

jreback:
closing as this is dependent on #22336 which needs to be fixed first.

Issue #6581, opened 2014-03-09
DEPR: Clean up list of deprecations from prior versions


jsexauer:
Log for previous deprecations, once they are actually removed, move the issue to #13777 

We try to keep these for *three* major versions as actual deprecations. e.g. deprecate in 0.17, 0.18 & 0.19 get the warning, removed in 0.20.

## 0.24.0
- [x] #24596 `__array__` for tzaware Series/Index (0.24.0)

jsexauer:
I was going to start working on the 0.16 deprecations soon.  Wanted to check in and see if you still wanted everything that is listed in 0.16 to be deprecated. @jreback

jorisvandenbossche:
I think everything in the list is good to go.

jreback:
I updated all of the to be deprecated issues. I think think anything <= 0.14.0 should be done in 0.17.0. All require mentions in the ``Previous Deprecations`` section. Some may need a doc note if its 'important' enough (e.g. this was removed in version ???) and/or more prominent mention in the whatsnew (e.g. i the highlites)

jreback:
Ok with doing the 0.15.0 ones to be honest as well.

jreback:
anyone interested in putting up a PR for any of these deprecation removals. Anything < 0.14 is ok for sure. anything else pls lmk.

jreback:
@TomAugspurger 

can we do a PR for this? (first one up top)
```
 boxplot #7096 (0.14.0)
 change default of return_type from None to 'axes'
 update return_type section in visualization.rst
```

TomAugspurger:
Sure thing, will do a PR tonight or Sunday.

jorisvandenbossche:
I updated the list at the top, and there are some deprecations from 0.15 that can be removed in 0.19.0 (and some of 0.16.0 are maybe also worth considering)

gfyoung:
@jorisvandenbossche , @jreback : What do the "maybe" deprecation removals represent?  Are they ones that could be done in `0.19.0` but are not necessary?  Or that we're not sure yet about them?

gfyoung:
@jorisvandenbossche , @jreback : What do the "maybe" deprecation removals represent?  Are they ones that could be done in `0.19.0` but are not necessary?  Or that we're not sure yet about them?

jreback:
I think let's try to prioritize the ones that are marked for 0.19.0, then do the maybes if possible (and other ones on a case-by-case).

[...]

PR #22324, opened 2018-08-13
CLN: Remove deprecated method

alimcmaster1:
These params are no longer in the method, therefore can remove the docs strings

### DIFF
--- a/doc/source/timeseries.rst
+++ b/doc/source/timeseries.rst
@@ -426,10 +426,11 @@ Custom Frequency Ranges
 .. warning::
 
    This functionality was originally exclusive to ``cdate_range``, which is
-   deprecated as of version 0.21.0 in favor of ``bdate_range``.  Note that
-   ``cdate_range`` only utilizes the ``weekmask`` and ``holidays`` parameters
-   when custom business day, 'C', is passed as the frequency string. Support has 
-   been expanded with ``bdate_range`` to work with any custom frequency string.
+   deprecated as of version 0.21.0 (deleted in 0.24.0) in favor of
+   ``bdate_range``. Note that ``cdate_range`` only utilizes the ``weekmask``
+   and ``holidays`` parameters when custom business day, 'C', is passed
+   as the frequency string. Support has been expanded with ``bdate_range``
+   to work with any custom frequency string.
 
 .. versionadded:: 0.21.0
 
--- a/doc/source/timeseries.rst
+++ b/doc/source/timeseries.rst
@@ -423,15 +423,6 @@ resulting ``DatetimeIndex``:
 Custom Frequency Ranges
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-.. warning::
-
-   This functionality was originally exclusive to ``cdate_range``, which is
-   deprecated as of version 0.21.0 (deleted in 0.24.0) in favor of
-   ``bdate_range``. Note that ``cdate_range`` only utilizes the ``weekmask``
-   and ``holidays`` parameters when custom business day, 'C', is passed
-   as the frequency string. Support has been expanded with ``bdate_range``
-   to work with any custom frequency string.
-
 .. versionadded:: 0.21.0
 
 ``bdate_range`` can also generate a range of custom frequency dates by using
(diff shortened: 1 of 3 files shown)


pandas/core/indexes/datetimes.py
```diff
@@ -1815,11 +1815,6 @@ def cdate_range(start=None, end=None, periods=None, freq='C', tz=None,
         Normalize start/end dates to midnight before generating date range
     name : string, default None
         Name of the resulting DatetimeIndex
-    weekmask : string, Default 'Mon Tue Wed Thu Fri'
-        weekmask of valid business days, passed to ``numpy.busdaycalendar``
-    holidays : list
-        list/array of dates to exclude from the set of valid business days,
-        passed to ``numpy.busdaycalendar``
```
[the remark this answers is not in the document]
alimcmaster1: Ahh apologies I looked at `bdate_range` and saw they where optional params there opposed to kwargs. It's a deprecated method anyway, but how about documenting it something like below So its clear they are kwargs?

    closed : string, default None
        Make the interval closed with respect to the given frequency to
        the 'left', 'right', or both sides (None)
    **weekmask : string, Default 'Mon Tue Wed Thu Fri'
        weekmask of valid business days, passed to ``numpy.busdaycalendar``
    **holidays : list
        list/array of dates to exclude from the set of valid business days,
        passed to ``numpy.busdaycalendar``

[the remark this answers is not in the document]
gfyoung: Actually, I'll do you one better.  Let's just remove it (it's been long enough).  How does that sound?

doc/source/timeseries.rst
```diff
@@ -426,10 +426,11 @@ Custom Frequency Ranges
...
    This functionality was originally exclusive to ``cdate_range``, which is
-   deprecated as of version 0.21.0 in favor of ``bdate_range``.  Note that
-   ``cdate_range`` only utilizes the ``weekmask`` and ``holidays`` parameters
-   when custom business day, 'C', is passed as the frequency string. Support has 
-   been expanded with ``bdate_range`` to work with any custom frequency string.
+   deprecated as of version 0.21.0 (deleted in 0.24.0) in favor of
+   ``bdate_range``. Note that ``cdate_range`` only utilizes the ``weekmask``
+   and ``holidays`` parameters when custom business day, 'C', is passed
+   as the frequency string. Support has been expanded with ``bdate_range``
+   to work with any custom frequency string.
```
gfyoung: Do we still need to reference `cdate_range` in the docs?  Or can we rewrite to remove it?

alimcmaster1: Im thinking of just removing that warning altogether since it just mentions some subtleties of `cdate_range`. Alternatively potential rewrite could look like:
`
This functionality was originally exclusive to ``cdate_range``, which is
deprecated as of version 0.21.0 (deleted in 0.24.0) in favor of
``bdate_range`` which works with any custom frequency string.
`
Thoughts?

gfyoung: I would just remove it altogether.

alimcmaster1: Makes sense! updated

doc/source/timeseries.rst
```diff
@@ -423,14 +423,6 @@ resulting ``DatetimeIndex``:
...
 
-.. warning::
-
-   This functionality was originally exclusive to ``cdate_range``, which is
-   deprecated as of version 0.21.0 in favor of ``bdate_range``.  Note that
-   ``cdate_range`` only utilizes the ``weekmask`` and ``holidays`` parameters
-   when custom business day, 'C', is passed as the frequency string. Support has 
-   been expanded with ``bdate_range`` to work with any custom frequency string.
-
 .. versionadded:: 0.21.0
```
gfyoung: I think this line can be removed as well.

jreback:
we are not removing 0.21.0 deprecations yet, see the list here: https://github.com/pandas-dev/pandas/issues/6581

things deprecated in 0.20.0 are fair game however.

alimcmaster1:
Okay no worries, thanks @jreback for the info!

gfyoung:
@jreback : That was my mistake.  I was under the impression that `0.21.0` was fair game.

@alimcmaster1 : Sorry about that!  I think given we'll remove this method soon, let's leave the docs as is.  We can always make this removal later.

PR #22329, opened 2018-08-14
CLN: de-duplicate index validation code

jbrockmendel:
There are currently 3 nearly-identical versions of this code.  I'm pretty sure the strictest one is the most-correct, so that is made into the only one.

### DIFF
--- a/pandas/_libs/index.pyx
+++ b/pandas/_libs/index.pyx
@@ -50,11 +50,7 @@ cpdef get_value_at(ndarray arr, object loc, object tz=None):
 
 
 cpdef object get_value_box(ndarray arr, object loc):
-    cdef:
-        Py_ssize_t i
-
-    i = util.validate_indexer(arr, loc)
-    return get_value_at(arr, i, tz=None)
+    return get_value_at(arr, loc, tz=None)
 
 
 # Don't populate hash tables in monotonic indexes larger than this
--- a/pandas/_libs/index.pyx
+++ b/pandas/_libs/index.pyx
@@ -51,21 +51,9 @@ cpdef get_value_at(ndarray arr, object loc, object tz=None):
 
 cpdef object get_value_box(ndarray arr, object loc):
     cdef:
-        Py_ssize_t i, sz
+        Py_ssize_t i
 
-    if util.is_float_object(loc):
-        casted = int(loc)
-        if casted == loc:
-            loc = casted
-    i = <Py_ssize_t> loc
-    sz = cnp.PyArray_SIZE(arr)
-
-    if i < 0 and sz > 0:
-        i += sz
-
-    if i >= sz or sz == 0 or i < 0:
-        raise IndexError('index out of bounds')
-
+    i = util.validate_indexer(arr, loc)
     return get_value_at(arr, i, tz=None)
 
 
--- a/pandas/_libs/util.pxd
+++ b/pandas/_libs/util.pxd
@@ -44,7 +44,7 @@ ctypedef fused numeric:
     cnp.float64_t
 
 
-cdef inline Py_ssize_t validate_indexer(ndarray arr, object loc) except? -1:
+cdef inline Py_ssize_t validate_indexer(ndarray arr, object loc) except -1:
     """
     Cast the given indexer `loc` to an integer.  If it is negative, i.e. a
     python-style indexing-from-the-end indexer, translate it to a
--- a/pandas/_libs/util.pxd
+++ b/pandas/_libs/util.pxd
@@ -44,23 +44,50 @@ ctypedef fused numeric:
     cnp.float64_t
 
 
-cdef inline object get_value_at(ndarray arr, object loc):
+cdef inline Py_ssize_t validate_indexer(ndarray arr, object loc) except? -1:
+    """
+    Cast the given indexer `loc` to an integer.  If it is negative, i.e. a
+    python-style indexing-from-the-end indexer, translate it to a
+    from-the-front indexer.  Raise if this is not possible.
+
+    Parameters
+    ----------
+    arr : ndarray
+    loc : object
+
+    Returns
+    -------
+    idx : Py_ssize_t
+
+    Raises
+    ------
+    IndexError
+    """
     cdef:
-        Py_ssize_t i, sz
+        Py_ssize_t idx, size
         int casted
 
     if is_float_object(loc):
         casted = int(loc)
         if casted == loc:
             loc = casted
-    i = <Py_ssize_t> loc
-    sz = cnp.PyArray_SIZE(arr)
 
-    if i < 0 and sz > 0:
-        i += sz
-    elif i >= sz or sz == 0:
+    idx = <Py_ssize_t>loc
+    size = cnp.PyArray_SIZE(arr)
+
+    if idx < 0 and size > 0:
+        idx += size
+    if idx >= size or size == 0 or idx < 0:
         raise IndexError('index out of bounds')
 
+    return idx
+
+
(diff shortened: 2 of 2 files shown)


pandas/_libs/util.pxd
```diff
@@ -44,23 +44,50 @@ ctypedef fused numeric:
     cnp.float64_t
 
 
-cdef inline object get_value_at(ndarray arr, object loc):
+cdef inline Py_ssize_t validate_indexer(ndarray arr, object loc) except? -1:
```
jorisvandenbossche: I don't think you need the question mark in `except?` ? (since -1 can never be actually returned from the function without it being an error)

jorisvandenbossche (commented):
You mention you used the "strictest" one: what is exactly the difference between the current versions?

Can you also do a quick check of performance of a call that hits those functions?

jreback:
yeah these are in criticial paths of perf. pls do a check.

jbrockmendel:
> yeah these are in criticial paths of perf. pls do a check.

Running asv now.

> I don't think you need the question mark in except? ? (since -1 can never be actually returned from the function without it being an error)

The `?` is there purely out of habit.  Is there a cost to having it there?

> what is exactly the difference between the current versions?

Corner case handling.  e.g. if `len(arr) == 10` and `i = -4` they will all increment `i += len(arr)` and be OK.  But if `i = -14` only the strict one will check that `i + len(arr)` is still negative.

jbrockmendel:
asv results:

```
asv continuous -f 1.1 -E virtualenv master HEAD -b frame
[...]
     [ffae1587]       [7d2e4330]
+      9.97±0.5ms       16.1±0.3ms     1.61  frame_methods.Apply.time_apply_lambda_mean

asv continuous -f 1.1 -E virtualenv master HEAD -b frame
[...]
     [ffae1587]       [7d2e4330]
+        894±80μs      2.19±0.04ms     2.46  frame_ctor.FromRecords.time_frame_from_records_generator(1000)

taskset 8 time asv continuous -f 1.1 -E virtualenv master HEAD -b frame
[...]
     [ffae1587]       [7d2e4330]
-        172±10ms        946±100μs     0.01  frame_ctor.FromRecords.time_frame_from_records_generator(None)

time taskset 4 asv continuous -f 1.1 -E virtualenv master HEAD -b frame
[...]
     [ffae1587]       [7d2e4330]
-     2.20±0.02ms      1.01±0.01ms     0.46  frame_ctor.FromRecords.time_frame_from_records_generator(1000)

```

jorisvandenbossche:
The asv is clearly not really relevant. Can you do a quick check of the impacted functions with a direct call? Eg `get_value_box` is only used in `Index.get_value` and is eg hit when doing

```
In [4]: idx = pd.Index(['a', 'b', 'c', 'd'])

In [5]: s = pd.Series(range(len(idx)), index=idx)

In [6]: idx.get_value(s, 1)
Out[6]: 1
```

You can do a quick `%timeit` before/after (but with bigger index) to see if there is any significant change.

> The ? is there purely out of habit. Is there a cost to having it there?

There is an extra check whether there is an error raised or not. I don't assume this is costly, but since we know that it will always be an error, I think it is more cleanly code-wise to reflect this in the `except`

jorisvandenbossche:
>  But if i = -14 only the strict one will check that i + len(arr) is still negative.

OK, that sounds like the good change!

jbrockmendel:
I’ll remove the question mark when possible.  At the moment the power
company has decided to do some surprise maintenance, so that might be a
while...

As to the asv, yah it’s not clearly relevant, but it’s low-level code that
gets used a lot so I guessed that running all frame benchmarks would hit
it.  I’ll try your suggestion when the power comes back.

jbrockmendel:
Indistinguishable:

Master:
```
In [2]: idx = pd.Int64Index(range(10**6))

In [3]: ser = pd.Series(idx, index=idx)

In [4]: %timeit idx.get_value(ser, 1000)
10000 loops, best of 3: 26.6 µs per loop

In [5]: %timeit idx.get_value(ser, 1000)
10000 loops, best of 3: 26.7 µs per loop

```

PR:
```
In [2]: idx = pd.Int64Index(range(10**6))

In [3]: ser = pd.Series(idx, index=idx)

In [4]: %timeit idx.get_value(ser, 1000)
10000 loops, best of 3: 26.8 µs per loop

In [5]: %timeit idx.get_value(ser, 1000)
10000 loops, best of 3: 25.6 µs per loop
```

jbrockmendel:
```
In [2]: idx = pd.Index([str(x) for x in range(10**6)])

In [3]: ser = pd.Series(range(len(idx)), index=idx)

In [4]: %timeit idx.get_value(ser, 1000)
```

Master:
```
The slowest run took 7246.54 times longer than the fastest. This could mean that an intermediate result is being cached.
1 loop, best of 3: 37.9 µs per loop

The slowest run took 5.44 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 19.5 µs per loop

The slowest run took 5.11 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 18.8 µs per loop

The slowest run took 5.25 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 18.4 µs per loop

The slowest run took 4.27 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 23.6 µs per loop
```

PR:
```
The slowest run took 9550.30 times longer than the fastest. This could mean that an intermediate result is being cached.
1 loop, best of 3: 29.1 µs per loop

The slowest run took 5.77 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 19.1 µs per loop

The slowest run took 5.49 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 19.3 µs per loop

10000 loops, best of 3: 25.7 µs per loop

The slowest run took 4.35 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 18.2 µs per loop
```

jreback (commented):
can u add these as asvs?

jorisvandenbossche:
@jreback I am not sure that is needed. It should already be covered by other indexing benchmarks that use `get_value` under the hood 
(I mainly here wanted to check the very specific code path to be sure we are checking the correct thing, as a sanity check)

pandas/_libs/index.pyx
```diff
@@ -51,21 +51,9 @@ cpdef get_value_at(ndarray arr, object loc, object tz=None):
...
-    sz = cnp.PyArray_SIZE(arr)
-
-    if i < 0 and sz > 0:
-        i += sz
-
-    if i >= sz or sz == 0 or i < 0:
-        raise IndexError('index out of bounds')
+        Py_ssize_t i
 
+    i = util.validate_indexer(arr, loc)
```
jorisvandenbossche: Since `get_value_at` (the util version below, which is called by the `get_value_at` in this file) now has the same validation, is it then not unnecessary to call the validation here as well?

jbrockmendel: I think you’re right, will update.

jbrockmendel:
Following this and #22344 it tentatively looks like we'll be ready to get rid of numpy_helper (and chunks of util) altogether.

jorisvandenbossche (approved):

jbrockmendel:
@jreback gentle ping.  After this we can get rid of a bunch of old numpy_helper code.

jreback:
thanks!

yeah lots of PRs!

PR #22325, opened 2018-08-13
SparseArray is an ExtensionArray

TomAugspurger:
Closes #21978 
Closes #19506 
Closes https://github.com/pandas-dev/pandas/issues/22835

High-level summary: SparseArray is an ExtensionArray. It's no longer an ndarray subclass. The actual data model hasn't changed at all, it's still an array and a `sparse_index`. Only now the sparse values are `self.sparse_values`, rather than `self`.

This isn't really close to being ready yet. I'm going to go through and self-review a bunch of things right now, will call out for others' opinions in specific places.

API discussions:

- [x] Why is the default kind inconsistent between SparseSeries and SparseArray?
    - Possibly get rid of block in a future PR
- [x] Should `.astype(np_dtype)` be sparse or dense?
    - sparse
- [x] What should the inferred type of an empty SparseArray be? `SparseArray([]).dtype`? NumPy defaults to float (`Sparse[float64]`), pandas typically uses object (`Sparse[object]`)
    - sparse
- [x] Policy for warning when converting to dense (e.g. https://github.com/pandas-dev/pandas/pull/22325/files#diff-71caf9627e9687e837e4b1f86ecc6271R390). In the past, pandas would coerce to an ndarray all over the place. Now we at least have a *chance* of knowing when we're doing a bad thing, and warning about it.
    - warn when it's implicit

PR #22343, opened 2018-08-14
API: dispatch to EA.astype

### LINKED ISSUE #21296: Series[EA].astype always casts intermediate to NumPy array
Calling `pd.Series(EA).astype(object)` will always generate an intermediate NumPy array and never delegate to the `astype` method of the `ExtensionArray`.

Callstack is as follows:

1. `pd.Series(EA).astype()`
2. (omitting some intermediates)
3. `pandas.core.internals.Block._astype` calls `self.get_values()` in https://github.com/pandas-dev/pandas/blob/4274b840e64374a39a0285c2174968588753ec35/pandas/core/internals.py#L661
    (introducted by https://github.com/pandas-dev/pandas/pull/20581/files)
4. `pandas.core.internals.ExtensionBlock.get_values` then casts the ExtensionArray to a `numpy.array`: https://github.com/pandas-dev/pandas/blob/4274b840e64374a39a0285c2174968588753ec35/pandas/core/internals.py#L1937

Expected behaviour would have been that `astype` is called on the `ExtensionArray` which then can do the casting on its own. Currently I have the problem that my underlying storage (ExtensionArray backed by Arrow) is not numpy-compatible and thus everything turns into `np.array(…, dtype=object)` before it is casted.

Happy to fix this on my own but I would need a pointer on what the correct approach is, i.e. where one should delegate to `ExtensionArray.astype`.

Ran into this with 0.23.0 but the code has not changed in master in this area.

<details>

```
INSTALLED VERSIONS                                                                                                                                    [6/1805]
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.0-112-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: de_DE.UTF-8
LOCALE: de_DE.UTF-8

pandas: 0.23.0
pytest: 3.6.0
pip: 9.0.3
setuptools: 39.2.0
Cython: None
numpy: 1.14.2
scipy: None
pyarrow: 0.9.0
xarray: None
IPython: 6.4.0
sphinx: None
patsy: None
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: 1.0.1
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
```

</details>

TomAugspurger:
Closes https://github.com/pandas-dev/pandas/issues/21296

split from https://github.com/pandas-dev/pandas/pull/22325

Note that this is just a partial fix to get sparse working. We have some issues:

for better or worse, categorical is still special cased. `Series[EA].astype('category')` will not yet dispatch to `EA.astype`. That's good non-pandas EAs won't have to worry about specially handling categorical. It's bad, since it's a special case.

And that's the second issue, in general, how should we handle astyping from EA1 -> EA2? In principle, we always have `ndarray[object]` as the lowest common denominator. So something like `EA2(EA.astype(object))` may work. But that won't necessarily be efficient. I'm not sure what's best here.

cc @xhochy

### DIFF
--- a/pandas/core/arrays/integer.py
+++ b/pandas/core/arrays/integer.py
@@ -8,6 +8,7 @@ from pandas.util._decorators import cache_readonly
 from pandas.compat import u, range
 from pandas.compat import set_function_name
 
+from pandas.core.dtypes.cast import astype_nansafe
 from pandas.core.dtypes.generic import ABCSeries, ABCIndexClass
 from pandas.core.dtypes.common import (
     is_integer, is_scalar, is_float,
@@ -391,7 +392,7 @@ class IntegerArray(ExtensionArray, ExtensionOpsMixin):
 
         # coerce
         data = self._coerce_to_ndarray()
-        return data.astype(dtype=dtype, copy=False)
+        return astype_nansafe(data, dtype, copy=None)
 
     @property
     def _ndarray_values(self):
--- a/pandas/core/dtypes/cast.py
+++ b/pandas/core/dtypes/cast.py
@@ -647,7 +647,17 @@ def coerce_to_dtypes(result, dtypes):
 
 def astype_nansafe(arr, dtype, copy=True):
     """ return a view if copy is False, but
-        need to be very careful as the result shape could change! """
+        need to be very careful as the result shape could change!
+
+    Parameters
+    ----------
+    arr : ndarray
+    dtype : np.dtype
+    copy : bool or None, default True
+        Whether to copy during the `.astype` (True) or
+        just return a view (False). Passing `copy=None` will
+        attempt to return a view, but will copy if necessary.
+    """
 
     # dispatch on extension dtype if needed
     if is_extension_array_dtype(dtype):
@@ -735,7 +745,16 @@ def astype_nansafe(arr, dtype, copy=True):
 
     if copy:
         return arr.astype(dtype, copy=True)
-    return arr.view(dtype)
+    else:
+        try:
+            return arr.view(dtype)
+        except TypeError:
+            if copy is None:
+                # allowed to copy if necessary (e.g. object)
+                return arr.astype(dtype, copy=True)
+            else:
+                raise
+
 
 
 def maybe_convert_objects(values, convert_dates=True, convert_numeric=True,
--- a/pandas/core/dtypes/cast.py
+++ b/pandas/core/dtypes/cast.py
@@ -653,10 +653,6 @@ def astype_nansafe(arr, dtype, copy=True):
     ----------
     arr : ndarray
     dtype : np.dtype
-    copy : bool or None, default True
-        Whether to copy during the `.astype` (True) or
-        just return a view (False). Passing `copy=None` will
-        attempt to return a view, but will copy if necessary.
     """
 
     # dispatch on extension dtype if needed
@@ -745,15 +741,7 @@ def astype_nansafe(arr, dtype, copy=True):
 
     if copy:
         return arr.astype(dtype, copy=True)
-    else:
-        try:
-            return arr.view(dtype)
-        except TypeError:
-            if copy is None:
-                # allowed to copy if necessary (e.g. object)
-                return arr.astype(dtype, copy=True)
-            else:
-                raise
+    return arr.view(dtype)
 
 
 def maybe_convert_objects(values, convert_dates=True, convert_numeric=True,
--- a/pandas/core/dtypes/cast.py
+++ b/pandas/core/dtypes/cast.py
@@ -653,6 +653,7 @@ def astype_nansafe(arr, dtype, copy=True):
     ----------
     arr : ndarray
     dtype : np.dtype
+    copy : bool, default True
     """
 
     # dispatch on extension dtype if needed
--- a/pandas/core/dtypes/cast.py
+++ b/pandas/core/dtypes/cast.py
@@ -654,6 +654,8 @@ def astype_nansafe(arr, dtype, copy=True):
     arr : ndarray
     dtype : np.dtype
     copy : bool, default True
+        If False, a view will be attempted but may fail, if
+        e.g. the itemsizes don't align.
     """
 
     # dispatch on extension dtype if needed
--- a/pandas/core/dtypes/cast.py
+++ b/pandas/core/dtypes/cast.py
@@ -739,8 +739,10 @@ def astype_nansafe(arr, dtype, copy=True):
                       FutureWarning, stacklevel=5)
         dtype = np.dtype(dtype.name + "[ns]")
 
-    if copy:
+    if copy or is_object_dtype(arr) or is_object_dtype(dtype):
+        # Explicit copy, or required since NumPy can't view from / to object.
         return arr.astype(dtype, copy=True)
+
     return arr.view(dtype)
 
 
--- a/pandas/core/dtypes/cast.py
+++ b/pandas/core/dtypes/cast.py
@@ -756,7 +756,6 @@ def astype_nansafe(arr, dtype, copy=True):
                 raise
 
 
-
 def maybe_convert_objects(values, convert_dates=True, convert_numeric=True,
                           convert_timedeltas=True, copy=True):
     """ if we have an object dtype, try to coerce dates and/or numbers """
--- a/pandas/core/internals/blocks.py
+++ b/pandas/core/internals/blocks.py
@@ -637,22 +637,25 @@ class Block(PandasObject):
             # force the copy here
             if values is None:
 
-                if issubclass(dtype.type,
-                              (compat.text_type, compat.string_types)):
-
-                    # use native type formatting for datetime/tz/timedelta
-                    if self.is_datelike:
-                        values = self.to_native_types()
-
-                    # astype formatting
-                    else:
-                        values = self.get_values()
-
+                if self.is_extension:
+                    values = self.values.astype(dtype)
                 else:
-                    values = self.get_values(dtype=dtype)
+                    if issubclass(dtype.type,
+                                  (compat.text_type, compat.string_types)):
 
-                # _astype_nansafe works fine with 1-d only
-                values = astype_nansafe(values.ravel(), dtype, copy=True)
+                        # use native type formatting for datetime/tz/timedelta
+                        if self.is_datelike:
+                            values = self.to_native_types()
+
+                        # astype formatting
+                        else:
+                            values = self.get_values()
+
+                    else:
+                        values = self.get_values(dtype=dtype)
+
+                    # _astype_nansafe works fine with 1-d only
+                    values = astype_nansafe(values.ravel(), dtype, copy=True)
 
                 # TODO(extension)
                 # should we make this attribute?
--- a/pandas/tests/extension/decimal/array.py
+++ b/pandas/tests/extension/decimal/array.py
@@ -15,6 +15,17 @@ class DecimalDtype(ExtensionDtype):
     name = 'decimal'
     na_value = decimal.Decimal('NaN')
 
+    def __init__(self, context=None):
+        self.context = context or decimal.getcontext()
+
+    def __eq__(self, other):
+        if isinstance(other, type(self)):
+            return self.context == other.context
+        return super(DecimalDtype, self).__eq__(other)
+
+    def __repr__(self):
+        return 'DecimalDtype(context={})'.format(self.context)
+
     @classmethod
     def construct_array_type(cls):
         """Return the array type associated with this dtype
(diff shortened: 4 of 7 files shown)


pandas/core/arrays/integer.py
```diff
@@ -391,7 +392,7 @@ def astype(self, dtype, copy=True):
 
         # coerce
         data = self._coerce_to_ndarray()
-        return data.astype(dtype=dtype, copy=False)
+        return astype_nansafe(data, dtype, copy=None)
```
TomAugspurger: This was necessary so that 

```python
In [8]: s.values
Out[8]: IntegerArray([0, nan, 2], dtype='Int8')

In [9]: s.astype('uint32')
```

still raises. Previously, it did

```python
In [4]: pd.core.arrays.IntegerArray([1, None, 2], dtype='uint8').astype('uint32')
Out[4]: array([         1, 2415919104,          2], dtype=uint32)
```

which I don't think we want. This was only tested at the `Series[IntegerArray].astype` level, which never called EA.astype

pandas/tests/extension/decimal/array.py
```diff
@@ -15,6 +15,17 @@ class DecimalDtype(ExtensionDtype):
     name = 'decimal'
     na_value = decimal.Decimal('NaN')
 
+    def __init__(self, context=None):
```
TomAugspurger: Seemed like a good idea to have a parameterized ExtensionDtype in the tests folder. For `.astype`, we're testing that `Series[decimal].astype(DecimalDtype(new_context))` is passed all the way through.

pandas/tests/extension/decimal/array.py
```diff
@@ -15,6 +15,17 @@ class DecimalDtype(ExtensionDtype):
...
+        self.context = context or decimal.getcontext()
+
+    def __eq__(self, other):
```
TomAugspurger: Our default __eq__ is dangerous for parametrized dtypes, depending on whether the parameters appear in the name :/

pandas/tests/extension/decimal/test_decimal.py
```diff
@@ -205,6 +205,24 @@ def test_dataframe_constructor_with_dtype():
     tm.assert_frame_equal(result, expected)
 
 
+@pytest.mark.parametrize("frame", [True, False])
+def test_astype_dispatches(frame):
+    data = pd.Series(DecimalArray([decimal.Decimal(2)]), name='a')
```
jorisvandenbossche: can you add a comment here why this test is added here (instead of as generic extension test)

pandas/core/arrays/integer.py
jorisvandenbossche: Can you add a test for this?

jorisvandenbossche:
Code changes look good!

Regarding the discussion items you mention, I am wondering if we need some kind of "negotation protocol" between (EA) dtypes for this. Basically something like:

- consider ``Series(EA).astype(dtype)``
- first try ``EA.astype(dtype)`` -> if the EA knows how to handle the specific `dtype`, return a new EA/array; if it does not know how to convert itself to `dtype`, return NotImplemented
- if the above returned NotImplemented, ask `dtype` if it knows how to convert the `EA` values to its own dtype -> if it knows, return a new EA/array; otherwise return NotImplemented
- if both don't know (above return NotImplemented again), either raise an error or fall back to the base implementation (`np.array(self, dtype=dtype,`, which might also raise an error)

That would in principle enable `EA.astype(EA2)`, and might also mitigate the 'category' problem you mention.

jorisvandenbossche:
> I worry a bit about a divergence between Series.astype(dtype) and Series.values.astype(dtype)

That's a good point, yeah, that would be strange. Such a negotiation does not really fit for a method, only for functions probably ..

TomAugspurger:
Perhaps. We'll leave that for another issue though I think, if we discover a need for it in practice.

pandas/core/internals/blocks.py
```diff
@@ -637,22 +637,25 @@ def _astype(self, dtype, copy=False, errors='raise', values=None,
             # force the copy here
             if values is None:
 
-                if issubclass(dtype.type,
-                              (compat.text_type, compat.string_types)):
+                if self.is_extension:
```
jreback: @TomAugspurger are you really really sure this is needed? 
all of this already works with IntegerArray so not t really sure what problem you are trying to solve

jreback: we already do this is my point

TomAugspurger: Not on master? That hit's https://github.com/pandas-dev/pandas/blob/master/pandas/core/internals/blocks.py#L652, which converts to an ndarray, before ever calling the extension array's astype.

TomAugspurger: At least, my new tests fail on master without these changes. I'm not sure if / how IntegerArray is being handled differently.

jorisvandenbossche: We don't already "do" this (depending on what you mean with "do" .. :-))

The reason this already works correctly for integer dtypes, is because we are converting the values to an array, and then pass this array to the dtypes array constructor (in case of an EA dtype). 

But as Tom tries to argue, always converting to an ndarray (which is what `get_values` does) is not necessarily a good idea. It might not work (in case of sparse), or can also less efficient (what is actually the case for the integers -> eg when going from Int64 -> Int32, we first convert the Int64 EA to object ndarray to then convert (and infer missing values) again to EA of Int32. This could be done much more efficient).

jorisvandenbossche: > [about Int EA astyping] This could be done much more efficient

And note that this (the efficient shortcut for going from one int EA to another) is actually implemented in `IntegerArray.astype`. We only never call this in case of `Series(EA).astype(..)` due to the current implementation.

jreback: i am not arguing that we shouldn’t do this
the implementation is way too convoluted currently

jorisvandenbossche: do you mean the current code in master or this PR?

jreback: this PR

there is a ton of code in astype to dispatch to extension types already
this adds yet another branch

jorisvandenbossche: Can you point to this "ton of code"? I don't see any other dispatch to EAs. 
What is done is the generic conversion to array and then the use of `astype_nansafe` (which has a dispatch to EAs, but for the target EA dtype, not for the calling EA dtype)

TomAugspurger: It's a 2 line change, this extra if condition. Which of those two lines is the convoluted one?

pandas/core/dtypes/cast.py
```diff
@@ -735,7 +745,15 @@ def astype_nansafe(arr, dtype, copy=True):
 
     if copy:
         return arr.astype(dtype, copy=True)
-    return arr.view(dtype)
+    else:
+        try:
+            return arr.view(dtype)
+        except TypeError:
+            if copy is None:
```
jreback: why is this necessary? what breaks this?

pandas/core/dtypes/cast.py
```diff
@@ -735,7 +745,15 @@ def astype_nansafe(arr, dtype, copy=True):
...
+                # allowed to copy if necessary (e.g. object)
+                return arr.astype(dtype, copy=True)
+            else:
```
jreback: don't need an else

pandas/core/internals/blocks.py
jreback: hmm, I could swear I add an ``.is_extension`` branch already, that's why this looks weird. In any event this has lots of if/then condition. Please make a note / issue to clean this up.

jreback (changes requested):

pandas/core/dtypes/cast.py
TomAugspurger: There was some object astype that failed (not sure if it's to or from object). I've reverted this change, we'll see what fails.

jorisvandenbossche:
```
In [10]: np.array([1, 2, 3], dtype='int64').view(object)
...
TypeError: Cannot change data-type for object array.
```

pandas/core/dtypes/cast.py
```diff
@@ -653,10 +653,6 @@ def astype_nansafe(arr, dtype, copy=True):
     ----------
     arr : ndarray
     dtype : np.dtype
-    copy : bool or None, default True
-        Whether to copy during the `.astype` (True) or
-        just return a view (False). Passing `copy=None` will
-        attempt to return a view, but will copy if necessary.
```
jorisvandenbossche: the copy keyword still exists, so I would not remove it completely

pandas/core/dtypes/cast.py
TomAugspurger: Ah, yes. thanks. Pushed a simpler fix that hopefully works.

pandas/tests/extension/decimal/test_decimal.py
```diff
@@ -205,6 +205,27 @@ def test_dataframe_constructor_with_dtype():
     tm.assert_frame_equal(result, expected)
 
 
+@pytest.mark.parametrize("frame", [True, False])
```
jreback: why don't we add a test in base which raises NotImplementedError so that authors are forced to cover this?

jreback (changes requested):
small test comment. is the references closing issue at the top-of-the PR correct? (it is the original EA dtype registry)?

TomAugspurger:
Fixed the referenced issue.

pandas/core/dtypes/cast.py
```diff
@@ -647,7 +647,14 @@ def conv(r, dtype):
...
 def astype_nansafe(arr, dtype, copy=True):
     """ return a view if copy is False, but
-        need to be very careful as the result shape could change! """
+        need to be very careful as the result shape could change!
+
+    Parameters
+    ----------
+    arr : ndarray
+    dtype : np.dtype
+    copy : bool, default True
```
jorisvandenbossche: Can you specify here what happens with False? (will always do view (except for object), even if the itemsize is incompatible)

pandas/tests/extension/decimal/test_decimal.py
jorisvandenbossche: But is it useful to force EA authors to do it? This is basically checking the pandas dispatch (which we can do here), not the actual EA.astype implementation

TomAugspurger: Or how about something like

```python
@pytest.mark.parametrize('typ, check', [
    ('category', 'is_categorical_dtype'),
    ...
])
def test_astype_category(self, data):
    assert check(self.astype(dtype))
```

Would that make sense as a base test? I think our default implementation would need to be updated to not fail that.

My main concern is that it would be difficult to override (e.g. skip) just some of the dtypes, so maybe we would have to write those as separate tests?

pandas/tests/extension/integer/test_integer.py
```diff
@@ -567,6 +567,14 @@ def test_astype(self, all_data):
         expected = pd.Series(np.asarray(mixed))
         tm.assert_series_equal(result, expected)
 
+    def test_astype_nansafe(self):
+        # https://github.com/pandas-dev/pandas/pull/22343
+        arr = IntegerArray([np.nan, 1, 2], dtype="Int8")
+
+        with tm.assert_raises_regex(
+                ValueError, 'cannot convert float NaN to integer'):
+            arr.astype('uint32')
```
jorisvandenbossche: Can you put this as a stand-alone test at the bottom of this file? (it's not overwriting a base test, but it is a custom integer-specific test)

TomAugspurger:
Ugh, too early. https://github.com/pandas-dev/pandas/issues/21296 is the one.

pandas/tests/extension/decimal/test_decimal.py
TomAugspurger: I'm mildly concerned that EA authors will not correctly handle extension types. Our base implementation currently fails to handle them. Although we document it as `Cast to a NumPy array with 'dtype'.`. Should we make the return type Union[ndarray, ExtensionArray]`?

jorisvandenbossche: I would leave such a test until we better figure out how to handle those cross EA/non-EA astype calls (the discussion we were having above in this PR )

> Should we make the return type Union[ndarray, ExtensionArray]?

Yeah, probably yes in any case (since EAs authors that provide multiple dtypes already do that, like IntegerArray)

jorisvandenbossche (approved):

TomAugspurger:
@jreback was your comment about the wrong issue number all that was left from your point of view? If so, this should be good to go.

pandas/tests/extension/decimal/test_decimal.py
jreback: I would add an issue to make this test, not blocking this PR on it.

jreback (approved):

jreback:
thanks @TomAugspurger 

maybe need to modify text in pandas/arrays/base/ExtensionArray (the doc-string) to add astype)?

PR #22345, opened 2018-08-14
API: ExtensionDtype._is_numeric

### LINKED ISSUE #22290: ExtensionBlock.is_numeric is always False
#### Code Sample/Problem Description

Currently `ExtensionBlock.is_numeric` always returns `False`.  This can be problematic for extension arrays that are numeric, as this is used under the hood in places to filter to numeric columns in a `DataFrame`.  I'll be using `IntegerArray` as an example, but this in principle applies to any numeric extension array, e.g. `DecimalArray` in the testing suite, an extension array for units/uncertainties, etc.

Setup:
```python
In [2]: df = pd.DataFrame({'group': list('aaabbb'),
   ...:                    'val1': IntegerArray([0, 1, 2, np.nan, 3, 4]),
   ...:                    'val2': np.arange(6)})
   ...:                    

In [3]: df
Out[3]: 
  group val1  val2
0     a    0     0
1     a    1     1
2     a    2     2
3     b  NaN     3
4     b    3     4
5     b    4     5
```

The `IntegerArray` column is ignored by `DataFrame._get_numeric_data()`:
```python
In [4]: df._get_numeric_data()
Out[4]: 
   val2
0     0
1     1
2     2
3     3
4     4
5     5
```

This leads some numeric routines, such as `DataFrame.corr` ignoring the `IntegerArray` column:
```python
In [5]: df.corr()
Out[5]: 
      val2
val2   1.0
```

Likewise, `groupby` uses `ExtensionBlock.is_numeric` to filter to numeric columns for some operations, leading to the `IntegerArray` column being ignored, even if explicitly requested:
```python
In [6]: df.groupby('group').sum()
Out[6]: 
       val2
group      
a         3
b        12

In [7]: df.groupby('group')['val1', 'val2'].sum()
Out[7]: 
       val2
group      
a         3
b        12
```
#### Expected Output
I'd expect `ExtensionBlock.is_numeric` to return `True` when appropriate, and for behavior to be consistent with non-extension numeric dtypes.

My first impression is that this should be an attribute of the `ExtensionArray` or `ExtensionDtype` class that defaults to `False`, with numeric implementations setting the attribute to `True`, and `ExtensionBlock.is_numeric` would read the value from there.

<details>

INSTALLED VERSIONS
------------------
commit: 0370740034978d3a63d4b8e5e2c96ff54e7e08ba
python: 3.6.5.final.0
python-bits: 64
OS: Linux
OS-release: 4.14.29-galliumos
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.24.0.dev0+456.g0370740
pytest: 3.5.1
pip: 10.0.1
setuptools: 39.1.0
Cython: 0.28.2
numpy: 1.14.3
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: 1.7.4
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.2.1
tables: 3.4.3
numexpr: 2.6.5
feather: None
matplotlib: 2.2.2
openpyxl: 2.5.3
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.4
lxml: 4.2.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.7
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
gcsfs: None

</details>

TomAugspurger:
closes #22290 

split from https://github.com/pandas-dev/pandas/pull/22325

It's not clear what else we should be testing, since I'm not sure what all uses `Block.is_numeric`.

cc @jschendel


pandas/core/internals/blocks.py
```diff
@@ -669,7 +669,9 @@ def _astype(self, dtype, copy=False, errors='raise', values=None,
             newb = self.copy() if copy else self
 
         if newb.is_numeric and self.is_numeric:
-            if newb.shape != self.shape:
+            # use values.shape, rather than newb.shape, as newb.shape
+            # may be incorrect for ExtensionBlocks.
+            if values.shape != self.shape:
```
jschendel: Yeah, I ran into this same issue when I was testing out a similar implementation.

I got around it by passing the argument `ndim=self.ndim` to the `make_block` function on line 664 above, which seems to get the job done and passes all relevant tests, though I didn't run the entire test suite.  Not familiar enough with this code though to say if that's a better (or even adequate) workaround.

For what it's worth, it looks like this `ndim` inference in `NonConsolidatableMixIn` is where things start to go wrong, at least for the length 1 `Series` case:
https://github.com/pandas-dev/pandas/blob/cf70d11a25ec59a5b630921d86759c7fa49884db/pandas/core/internals/blocks.py#L1780-L1785

Again, not familiar enough with this code to immediately see a fix, but maybe this is helpful to someone with more knowledge of this code than me?

jschendel (approved):
Thanks! LGTM unless my comment illuminates anything.

I've also pushed one additional test that I wrote for `DataFrame._get_numeric_data`, which is what I believe is called by all `DataFrame` methods that filter to numeric dtypes.

pandas/core/internals/blocks.py
TomAugspurger: >  Not familiar enough with this code though to say if that's a better (or even adequate) workaround.

That seems better.

pandas/core/dtypes/base.py
```diff
@@ -94,6 +94,18 @@ def is_dtype(cls, dtype):
...
 
+    @property
+    def _is_numeric(self):
+        # type: () -> bool
+        """
+        Whether columns with this dtype should be considered numeric.
+
+        By default ExtensionDtypes are assumed to be non-numeric.
+        They'll be excluded from operations that exclude non-numeric
+        columns, like groupby reductions, plotting, etc.
```
jorisvandenbossche: maybe put 'gropuby' in parantheses, as the same holds for normal reductions (in a dataframe)

pandas/tests/extension/base/groupby.py
```diff
@@ -67,3 +67,16 @@ def test_groupby_extension_apply(self, data_for_grouping, op):
         df.groupby("B").A.apply(op)
         df.groupby("A").apply(op)
         df.groupby("A").B.apply(op)
+
+    def test_in_numeric_groupby(self, data_for_grouping):
+        df = pd.DataFrame({"A": [1, 1, 2, 2, 3, 3, 1, 4],
+                           "B": data_for_grouping,
+                           "C": [1, 1, 1, 1, 1, 1, 1, 1]})
+        result = df.groupby("A").sum().columns
```
jorisvandenbossche: how does this test pass? (I thought reductions did not yet work for EAs?)

pandas/tests/extension/base/interface.py
```diff
@@ -67,3 +67,7 @@ def test_no_values_attribute(self, data):
         # code, disallowing this for now until solved
         assert not hasattr(data, 'values')
         assert not hasattr(data, '_values')
+
+    def test_is_numeric_honored(self, data):
+        result = pd.Series(data)
+        assert result._data.blocks[0].is_numeric is data.dtype._is_numeric
```
jorisvandenbossche: Maybe you can also test `df._get_numeric_data` ?

It would be nice if this would also work for `df.select_dtypes`, but I suppose that is yet another issue

jorisvandenbossche: > Maybe you can also test df._get_numeric_data ?

ah, that is done below

pandas/tests/extension/base/groupby.py
TomAugspurger: I wonder if it's going through the fallback `np.mean(arr)`? Will investigate.

jorisvandenbossche: Yeah, so reductions for Series fail (no `_reduce`), but for dataframe it seems to use some ndarray fallback

jreback:
lgtm. @jorisvandenbossche not sure if you comments are resolved? if so, pls merge away.

TomAugspurger:
@jorisvandenbossche all good?

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

Issue #9744, opened 2015-03-28
BUG: utils.testing.get_locales()


dstephens99:
Line 334: raw_locales = [] deletes the data received from locale_getter, so the return from this function is always a blank list.

```python
  try:
        raw_locales = locale_getter()
    except:
        return None

    try:
        # raw_locales is "\n" seperated list of locales
        # it may contain non-decodable parts, so split
        # extract what we can and then rejoin.
        raw_locales = []
        for x in raw_locales:
            try:
                raw_locales.append(str(x, encoding=pd.options.display.encoding))
            except:
                pass
    except TypeError:
        pass

    if prefix is None:
        return _valid_locales(raw_locales, normalize)

    found = re.compile('%s.*' % prefix).findall('\n'.join(raw_locales))
    return _valid_locales(found, normalize)
```
https://github.com/pydata/pandas/blob/8d2818e32d0bbb50e183ccb5724c391e4f604670/pandas/util/testing.py#L334

jreback:
it DOES run, but you have to override the locale; see the ``.travis.yml`` which sets the env variable which travis interprets.

jreback:
see the install section: https://travis-ci.org/pydata/pandas/jobs/55924400

jreback:
hmm, looking at the code it does seem that it is overriden...

dstephens99:
I tried that in a test branch.  The problem is the code in pandas.utils.get_locales().

You get the locales:
raw_locales = locale_getter()
Next line that runs is:
raw_locales = []

So the function never returns any locales.

I made the fix in:
https://github.com/dstephens99/pandas/commit/4b1e35deb27d1795a5d269751d034f3196d6bff9

And the Travis test has a bunch of errors in TestGoogle (because it wasn't running before).
https://travis-ci.org/dstephens99/pandas/jobs/56178961

dstephens99:
I'm fixing up the data-readers code, when I finish I'll make a pull request to update the pandas data / test_data code, then a pull request to fix this bug.

jreback:
@dstephens99 ok, gr8!

PR #22349, opened 2018-08-14
Drop redundant TestLocale

ginggs:
We have the `TestLocaleUtils` class in `tests/util/test_util.py`
https://github.com/pandas-dev/pandas/blob/master/pandas/tests/util/test_util.py#L421
which is more comprehensive and includes `test_get_locales` which is functionally identical.

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/tests/util/test_testing.py
+++ b/pandas/tests/util/test_testing.py
@@ -848,18 +848,6 @@ class TestRNGContext(object):
             assert np.random.randn() == expected0
 
 
-class TestLocale(object):
-
-    def test_locale(self):
-        if sys.platform == 'win32':
-            pytest.skip(
-                "skipping on win platforms as locale not available")
-
-        # GH9744
-        locales = tm.get_locales()
-        assert len(locales) >= 1
-
-
 def test_datapath_missing(datapath, request):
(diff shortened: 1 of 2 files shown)


gfyoung (approved):
Nice catch!

ginggs:
Should we move the reference to #9744 over to `test_get_locales` ?

gfyoung:
@ginggs : Yeah, why not.  Let's do that.

WillAyd:
Hmm I suppose the only difference here is that the previous test was only skipped on 32 bit windows but assumedly held for 64 bit. With the refactor it is skipped for both.

My knowledge of Windows locale is nil, but any concern about the change here for 64 bit Windows? Maybe the target class' skip decorator should only apply to 32 bit Windows?

ginggs:
Apparently `sys.platform` is `win32` even on 64-bit Windows:
https://mail.python.org/pipermail/patches/2000-May/000648.html

WillAyd (approved):
Shows how much I know - thanks @ginggs !

jreback:
thanks!

PR #22350, opened 2018-08-14
~Finish Collecting Arithmetic Tests

jbrockmendel:
De-duplicate fixtures and put nearly all of them in conftest


pandas/tests/arithmetic/conftest.py
```diff
@@ -30,12 +30,21 @@ def zero(request):
...
-                        pd.UInt64Index(np.arange(5, dtype='uint64'))],
+                        pd.UInt64Index(np.arange(5, dtype='uint64')),
+                        pd.RangeIndex(5)],
                 ids=lambda x: type(x).__name__)
 def idx(request):
     return request.param
 
 
+@pytest.fixture
+def tdser():
```
gfyoung: Perhaps `td_series` ?  At least an underscore between `td` and `ser` would be good.

jbrockmendel: Good idea, will do.

gfyoung (approved):

pandas/tests/arithmetic/conftest.py
```diff
@@ -47,6 +56,76 @@ def scalar_td(request):
...
+    Several timedelta-like and DateOffset instances that are _not_
+    compatible with Daily frequencies.
+    """
+    return request.param
+
+
+@pytest.fixture(params=[np.timedelta64(365, 'D'),
+                        pd.Timedelta(365).to_pytimedelta(),
+                        pd.Timedelta(days=365)] + _common_mismatch)
+def mismatched(request):
```
jreback: can you make this more descriptive

pandas/tests/arithmetic/conftest.py
```diff
@@ -47,6 +56,76 @@ def scalar_td(request):
...
+
+_common_mismatch = [pd.offsets.YearBegin(2),
+                    pd.offsets.MonthBegin(1),
+                    pd.offsets.Minute()]
+
+
+@pytest.fixture(params=[pd.Timedelta(minutes=30).to_pytimedelta(),
+                        np.timedelta64(30, 's'),
+                        pd.Timedelta(seconds=30)] + _common_mismatch)
+def not_hourly(request):
```
jreback: I would add _fixture to these

pandas/tests/arithmetic/test_numeric.py
```diff
@@ -712,6 +733,51 @@ def check(series, other):
         check(tser, 5)
 
 
+class TestUFuncCompat(object):
```
jreback: let's think about putting these in a separate file (test_funcs)? (may want to make numeric a subdir, its already getting long).

jreback (changes requested):
comments on the fixture names. Also want to ensure that individual test files are completely obvious to what they are testing, so may want to add a level of nesting as things grow longer.

pandas/tests/arithmetic/test_numeric.py
jbrockmendel: I'll give that some thought.  Preference is to do quick turnarounds with small PRs here since there are lots of bugs to fix.

pandas/tests/arithmetic/conftest.py
jbrockmendel: docstring doesn't get the job done?

jreback:
can you rebase

jbrockmendel:
Rebased.  After this the tests are pretty much all consolidated, so the next planned follow-up is to paramatrize/de-duplicate.

Also since#22390 just went in, upcoming PR to fix _most_ of the outstanding xfails

gfyoung:
@jbrockmendel : Have you addressed all of the comments here?  I know you were waiting on @jreback to review this again, but at first glance, I can't tell if you have addressed everything he mentioned earlier.

jbrockmendel:
> Have you addressed all of the comments here? 

I think so, yes.  But since this is a big enough PR that it is hard to keep track, I made #22559 as a subset of this in the hopes it would be easier in smaller pieces.

gfyoung:
> I think so, yes.

@jbrockmendel : What about this one [here](https://github.com/pandas-dev/pandas/pull/22350#discussion_r210548778)?  Just double checking! 🙂 

So should we be reviewing / merging this?  Or should we go after #22559 first?

jbrockmendel:
> What about this one here? Just double checking!

The suggestion on the fixture names was to stop fiddling for now and address in the next round.  Since there were multiple comments about fixture names, its easy to lose track when one response is aimed at multiple comments.

> So should we be reviewing / merging this? Or should we go after #22559 first?

Best case would be that everyone is happy here and #22559 is unnecessary.  If this is going to be a slog, then I'd like to get #22559 in as this is a blocker for other work.

gfyoung:
ping @jreback - could you review this again?

pandas/tests/arithmetic/conftest.py
```diff
@@ -30,12 +30,21 @@ def zero(request):
 
 @pytest.fixture(params=[pd.Float64Index(np.arange(5, dtype='float64')),
                         pd.Int64Index(np.arange(5, dtype='int64')),
-                        pd.UInt64Index(np.arange(5, dtype='uint64'))],
+                        pd.UInt64Index(np.arange(5, dtype='uint64')),
+                        pd.RangeIndex(5)],
                 ids=lambda x: type(x).__name__)
-def idx(request):
+def numeric_index(request):
```
jreback: can you docstring

pandas/tests/arithmetic/conftest.py
jreback: can you do this

pandas/tests/arithmetic/conftest.py
jreback: no, prefer this to be a more descriptive name

pandas/tests/arithmetic/test_numeric.py
```diff
@@ -42,6 +33,30 @@ def test_operator_series_comparison_zerorank(self):
         expected = 0.0 > pd.Series([1, 2, 3])
         tm.assert_series_equal(result, expected)
 
+    def test_df_numeric_cmp_dt64_raises(self):
+        # GH#8932, GH#22163
+        ts = pd.Timestamp.now()
+        df = pd.DataFrame({'x': range(5)})
+        with pytest.raises(TypeError):
+            df > ts
```
jreback: commented in another PR these should be in test_datetime

jreback (changes requested):

jreback:
can you rebsae

jbrockmendel:
Closing in favor of #22645.

PR #22359, opened 2018-08-15
DOC: Expose ExcelWriter as part of the Generated API

### LINKED ISSUE #21835: Expose ExcelWriter as part of the Generated API
It has a docstring already (though it could use some examples) and would cover usage questions like the below:

https://stackoverflow.com/questions/46173710/where-can-i-find-the-excelwriter-documentation

newinh:
- [x] closes #21835 
- N/A tests added / passed
- N/A whatsnew entry


WillAyd (changes requested):
Thanks for the PR! Can you also add an examples section to the docstring so people know how to use it?

I think we should also prevent the abstract properties and methods from being rendered, but copying @datapythonista for input

newinh:
@WillAyd yeap i'll try to add an example:)

jorisvandenbossche:
> I think we should also prevent the abstract properties and methods from being rendered

Yes, and I would also add a note in the docstring explicitly saying that none of the methods/properties are considered public? (only 'public' usage is what is in the example?) 
To avoid this rendering with sphinx of all methods, you need to do something like this:

```
.. autosummary::
   :toctree: generated/
   :template: autosummary/class_without_autosummary.rst

   ExcelWriter
```

WillAyd:
@newinh this is really shaping up! When I built your latest commit locally I was still seeing the methods and attributes we are trying to hide - were you not seeing these on your end? I think there is still something that needs to be tweaked there.

One other thing that would make this nice - the `mode` argument is getting added in v0.24. If you can expand your example to show how to subsequently append to an Excel file that would cover all the use cases here.

jorisvandenbossche:
> When I built your latest commit locally I was still seeing the methods and attributes we are trying to hide

Looking at how the `Int64Index` docstring is done (there everything is hidden in previous docs: https://pandas.pydata.org/pandas-docs/version/0.22.0/generated/pandas.Int64Index.html (in new docs, you get None listed, but will open a new issue for that, let's follow here the same practice)), I think this needs to be added to the docstring (a bit ugly ...):

```
Attributes
----------
None

Methods
-------
None
```

jorisvandenbossche:
Can you also add a note in the docstring explicitly saying that none of the methods/properties are considered public?

newinh:
@WillAyd Indeed, I've seen a abstract properties that should have appeared in docs page. But I thought it is problem of my local environment. (I first built the document)

Thanks to @jorisvandenbossche, I hid it.

newinh:
Added example cases and more details to note.

As you know, there may be some mistakes because of my lack of English. If so, Please let me know:)

pandas/io/excel.py
```diff
@@ -824,8 +824,56 @@ class ExcelWriter(object):
 
     Notes
     -----
+    None of methods and properties are considered public.
```
WillAyd: "None of methods" -> "None of the methods"

pandas/io/excel.py
```diff
@@ -824,8 +824,56 @@ class ExcelWriter(object):
...
+    Examples
+    --------
+    Using ExcelWriter, some settings can be added.
+
+    Default usage.
+
+    >>> with ExcelWriter('path_to_file.xlsx') as writer:
+    ...     df.to_excel(writer)
+
+    If you want to set engine that can manipulate Excel,
```
WillAyd: I think this second example is better served in `to_excel` as it generic functionality. Would be OK if you removed here and added that to the `to_excel` docs in a separate change

pandas/io/excel.py
```diff
@@ -824,8 +824,56 @@ class ExcelWriter(object):
...
     Notes
     -----
+    None of methods and properties are considered public.
+
     For compatibility with CSV writers, ExcelWriter serializes lists
     and dicts to strings before writing.
+
+    Examples
+    --------
+    Using ExcelWriter, some settings can be added.
```
WillAyd: Remove this line

pandas/io/excel.py
```diff
@@ -824,8 +824,56 @@ class ExcelWriter(object):
...
+
+    Default usage.
```
WillAyd: Instead of a "." use ":"

pandas/io/excel.py
```diff
@@ -824,8 +824,56 @@ class ExcelWriter(object):
...
+    In order to write separate DataFrames to separate sheets
+    in a single Excel file, one can pass an ExcelWriter.
+
+    >>> with ExcelWriter('path_to_file.xlsx') as writer:
+    ...     df1.to_excel(writer, sheet_name='Sheet1')
+    ...     df2.to_excel(writer, sheet_name='Sheet2')
+
+    You can set date format or datetime format
+
+    >>> with ExcelWriter('path_to_file.xlsx',
```
WillAyd: I am actually surprised that we allow date / date time formatting with ExcelWriter but don't have the same usage in `to_excel` - if you are up to it aligning that functionality would be ideal (separate change)

pandas/io/excel.py
```diff
@@ -824,8 +824,56 @@ class ExcelWriter(object):
...
+    ...     df.to_excel(writer)
+
+    If you want to set engine that can manipulate Excel,
+    pass keyword argument named engine. Actually
+    engine is automatically chosen by file extension.
+
+    >>> with ExcelWriter('path_to_file.xlsx', engine='openpyxl') as writer:
+    ...     df.to_excel(writer)
+
+    In order to write separate DataFrames to separate sheets
```
WillAyd: Just say "To write to separate sheets in a single file:"

pandas/io/excel.py
```diff
@@ -824,8 +824,56 @@ class ExcelWriter(object):
...
+    ...     df2.to_excel(writer, sheet_name='Sheet2')
+
+    You can set date format or datetime format
+
+    >>> with ExcelWriter('path_to_file.xlsx',
+                          date_format='YYYY-MM-DD',
+                          datetime_format='YYYY-MM-DD HH:MM:SS') as writer:
+    ...     df.to_excel(writer)
+
+    It also supports append mode.
```
WillAyd: "You can also append to an existing Excel file:"

WillAyd (changes requested):
Minor edits. There is also a note in the docstring to "See DataFrame.to_excel for typical usage" Can you instead create a "See Also" section that references that method?

newinh:
I do not know why the Travis CI build failed. Where should i look to fix it?

WillAyd (changes requested):
The test failure is because you changed `pd.ExcelWriter` to `ExcelWriter` in the example so Python doesn't know what you are referencing. If you revert that back should be OK (couple other edits needed as commented)

pandas/io/excel.py
```diff
@@ -824,8 +824,45 @@ class ExcelWriter(object):
...
+
+    >>> with ExcelWriter('path_to_file.xlsx',
+                          date_format='YYYY-MM-DD',
+                          datetime_format='YYYY-MM-DD HH:MM:SS') as writer:
+    ...     df.to_excel(writer)
+
+    You can also append to an existing Excel file:
+
+    >>> with ExcelWriter('path_to_file.xlsx', mode='a') as writer:
+    ...     df.to_excel(writer)
```
jorisvandenbossche: I would add here a sheet name? (I think you typically want to add a new sheet to an existing file?)

pandas/core/generic.py
```diff
@@ -1917,18 +1917,24 @@ def _repr_latex_(self):
     ...                   columns=['col 1', 'col 2'])
     >>> df1.to_excel("output.xlsx")
 
+    If you want to set engine that can manipulate Excel,
+    pass keyword argument named engine. Actually
+    engine is automatically chosen by file extension:
```
jorisvandenbossche: Suggestion with some changed language:

```
To set the library that used to write the Excel file, you can pass the `engine` keyword (the default engine is automatically chosen depending on the file extension):
```

pandas/io/excel.py
newinh: It seems more to do that:)

pandas/io/excel.py
```diff
@@ -824,8 +824,45 @@ class ExcelWriter(object):
...
+                          date_format='YYYY-MM-DD',
+                          datetime_format='YYYY-MM-DD HH:MM:SS') as writer:
+    ...     df.to_excel(writer)
+
+    You can also append to an existing Excel file:
+
+    >>> with ExcelWriter('path_to_file.xlsx', mode='a') as writer:
+    ...     df.to_excel(writer, sheet_name='Sheet3')
+
+    .. versionadded:: 0.24.0
```
jorisvandenbossche: I think this one can be removed

jorisvandenbossche (approved):
Minor comment, looks good to me!

pandas/core/generic.py
```diff
@@ -1924,11 +1924,17 @@ def _repr_latex_(self):
...
     necessary to specify an ExcelWriter object:
 
-    >>> writer = pd.ExcelWriter('output2.xlsx', engine='xlsxwriter')
-    >>> df1.to_excel(writer, sheet_name='Sheet1')
-    >>> df2 = df1.copy()
-    >>> df2.to_excel(writer, sheet_name='Sheet2')
-    >>> writer.save()
+    >>> with pd.ExcelWriter('output.xlsx') as writer:
+    ...     df1.to_excel(writer, sheet_name='Sheet_name_1')
+    ...     df2 = df1.copy()
```
WillAyd: Can you do this copy outside of the context manager? On initial glance I assumed it to be some requirement with the context manager itself, though that's clearly not the case

WillAyd (changes requested):

WillAyd (approved):
I think this looks good and slightly prefer the example with append mode but am largely indifferent.
@jorisvandenbossche merge when satisfied.

jorisvandenbossche:
>  slightly prefer the example with append mode but am largely indifferent

@WillAyd what do you mean exactly?

jreback:
thanks @newinh

PR #22363, opened 2018-08-15
DOC: update doc-string in Index.rename

### LINKED ISSUE #20787: DOC: clean Index.set_name / .rename doc-strings
xref #20548 

MultiIndex.rename directly uses the .set_names doc-string (which is ok), but should be a touch more linked (e.g. See Also, maybe some more examples). Alt could de-couple these and make them more informative.

Moons08:
updated doc-strings with examples 
- [x] closes #20787
- N/A tests added / passed
- N/A whatsnew entry

### DIFF
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -1284,12 +1284,12 @@ class Index(IndexOpsMixin, PandasObject):
 
         Parameters
         ----------
-        names : label or list of labels
+        names : label or list of label
             Name(s) to set.
-        level : int, label or list of labels, optional
+        level : int, label or list of int or label, optional
             If the index is a MultiIndex, level(s) to set (None for all
             levels). Otherwise level must be None.
-        inplace : boolean, default False
+        inplace : bool, default False
             Modifies the object directly, instead of creating a new Index or
             MultiIndex.
 
--- a/pandas/core/indexes/base.py
+++ b/pandas/core/indexes/base.py
@@ -1278,39 +1278,52 @@ class Index(IndexOpsMixin, PandasObject):
 
     def set_names(self, names, level=None, inplace=False):
         """
-        Set new names on index. Defaults to returning new index.
+        Set Index or MultiIndex name.
+
+        Able to set new names partially and by level.
 
         Parameters
         ----------
-        names : str or sequence
-            name(s) to set
-        level : int, level name, or sequence of int/level names (default None)
-            If the index is a MultiIndex (hierarchical), level(s) to set (None
-            for all levels).  Otherwise level must be None
-        inplace : bool
-            if True, mutates in place
+        names : label or list of labels
+            Name(s) to set.
+        level : int, label or list of labels, optional
+            If the index is a MultiIndex, level(s) to set (None for all
+            levels). Otherwise level must be None.
+        inplace : boolean, default False
+            Modifies the object directly, instead of creating a new Index or
+            MultiIndex.
 
         Returns
         -------
-        new index (of same type and class...etc) [if inplace, returns None]
+        Index
+            The same type as the caller or None if inplace is True.
+
+        See Also
+        --------
+        Index.rename : Able to set new names without level.
 
         Examples
         --------
-        >>> pd.Index([1, 2, 3, 4]).set_names('foo')
-        Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
-        >>> pd.Index([1, 2, 3, 4]).set_names(['foo'])
-        Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
-        >>> idx = pd.MultiIndex.from_tuples([(1, u'one'), (1, u'two'),
-                                            (2, u'one'), (2, u'two')],
-                                            names=['foo', 'bar'])
-        >>> idx.set_names(['baz', 'quz'])
-        MultiIndex(levels=[[1, 2], [u'one', u'two']],
-                   labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
-                   names=[u'baz', u'quz'])
-        >>> idx.set_names('baz', level=0)
-        MultiIndex(levels=[[1, 2], [u'one', u'two']],
-                   labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
-                   names=[u'baz', u'bar'])
+        >>> idx = pd.Index([1, 2, 3, 4])
+        >>> idx
+        Int64Index([1, 2, 3, 4], dtype='int64')
+        >>> idx.set_names('quarter')
+        Int64Index([1, 2, 3, 4], dtype='int64', name='quarter')
+
+        >>> idx = pd.MultiIndex.from_product([['python', 'cobra'],
+        ...                                   [2018, 2019]])
+        >>> idx
+        MultiIndex(levels=[['cobra', 'python'], [2018, 2019]],
+                   labels=[[1, 1, 0, 0], [0, 1, 0, 1]])
+        >>> idx.set_names(['kind', 'year'], inplace=True)
+        >>> idx
+        MultiIndex(levels=[['cobra', 'python'], [2018, 2019]],
+                   labels=[[1, 1, 0, 0], [0, 1, 0, 1]],
+                   names=['kind', 'year'])
+        >>> idx.set_names('species', level=0)
+        MultiIndex(levels=[['cobra', 'python'], [2018, 2019]],
+                   labels=[[1, 1, 0, 0], [0, 1, 0, 1]],
+                   names=['species', 'year'])
         """
 
         from .multi import MultiIndex
@@ -1319,7 +1332,8 @@ class Index(IndexOpsMixin, PandasObject):
 
         if level is not None and not is_list_like(level) and is_list_like(
                 names):
-            raise TypeError("Names must be a string")
+            msg = "Names must be a string when a single level is provided."
+            raise TypeError(msg)
 
         if not is_list_like(names) and level is None and self.nlevels > 1:
             raise TypeError("Must pass list-like as `names`.")
@@ -1339,18 +1353,48 @@ class Index(IndexOpsMixin, PandasObject):
 
     def rename(self, name, inplace=False):
         """
-        Set new names on index. Defaults to returning new index.
+        Alter Index or MultiIndex name.
+
+        Able to set new names without level. Defaults to returning new index.
+        Length of names must match number of levels in MultiIndex.
 
         Parameters
         ----------
-        name : str or list
-            name to set
-        inplace : bool
-            if True, mutates in place
+        name : label or list of labels
+            Name(s) to set.
+        inplace : boolean, default False
+            Modifies the object directly, instead of creating a new Index or
+            MultiIndex.
 
         Returns
         -------
-        new index (of same type and class...etc) [if inplace, returns None]
+        Index
+            The same type as the caller or None if inplace is True.
+
+        See Also
+        --------
+        Index.set_names : Able to set new names partially and by level.
+
+        Examples
+        --------
+        >>> idx = pd.Index(['A', 'C', 'A', 'B'], name='score')
+        >>> idx.rename('grade')
+        Index(['A', 'C', 'A', 'B'], dtype='object', name='grade')
+
+        >>> idx = pd.MultiIndex.from_product([['python', 'cobra'],
+        ...                                   [2018, 2019]],
+        ...                                   names=['kind', 'year'])
+        >>> idx
+        MultiIndex(levels=[['cobra', 'python'], [2018, 2019]],
+                   labels=[[1, 1, 0, 0], [0, 1, 0, 1]],
+                   names=['kind', 'year'])
+        >>> idx.rename(['species', 'year'])
+        MultiIndex(levels=[['cobra', 'python'], [2018, 2019]],
+                   labels=[[1, 1, 0, 0], [0, 1, 0, 1]],
+                   names=['species', 'year'])
+        >>> idx.rename('species')
+        Traceback (most recent call last):
+        TypeError: Must pass list-like as `names`.
         """
         return self.set_names([name], inplace=inplace)
 
--- a/pandas/tests/indexes/multi/test_get_set.py
+++ b/pandas/tests/indexes/multi/test_get_set.py
@@ -360,7 +360,7 @@ def test_set_levels_labels_names_bad_input(idx):
     with tm.assert_raises_regex(ValueError, 'Length of names'):
         idx.set_names(names[0], level=[0, 1])
 
-    with tm.assert_raises_regex(TypeError, 'string'):
+    with tm.assert_raises_regex(TypeError, 'Names must be a'):
         idx.set_names(names, level=0)
 
 


pandas/core/indexes/base.py
```diff
@@ -1340,17 +1340,39 @@ def set_names(self, names, level=None, inplace=False):
...
         -------
         new index (of same type and class...etc) [if inplace, returns None]
+
+        Examples
+        --------
+        >>> idx = Index([1, 2, 3, 4], name='foo')
+        Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
+        >>> idx.rename('bar')
+        Int64Index([1, 2, 3, 4], dtype='int64', name='bar')
+        >>> idx = MultiIndex.from_tuples([(1, u'one'), (1, u'two'),
```
jbrockmendel: Should this have a newline to denote a separate example?

pandas/core/indexes/base.py
```diff
@@ -1340,17 +1345,40 @@ def set_names(self, names, level=None, inplace=False):
     def rename(self, name, inplace=False):
         """
         Set new names on index. Defaults to returning new index.
+        Length of names must match number of levels in MultiIndex.
```
jorisvandenbossche: Leave a blank line between the two lines above (the first line of the docstring should be a single line, see http://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html#docstring for details)

pandas/core/indexes/base.py
```diff
@@ -1340,17 +1345,40 @@ def set_names(self, names, level=None, inplace=False):
...
         inplace : bool
             if True, mutates in place
 
         Returns
         -------
         new index (of same type and class...etc) [if inplace, returns None]
+
+        Examples
+        --------
+        >>> idx = Index([1, 2, 3, 4], name='foo')
```
jorisvandenbossche: Can you use `pd.` for pandas objects? (but no need to import it)

pandas/core/indexes/base.py
```diff
@@ -1340,17 +1345,40 @@ def set_names(self, names, level=None, inplace=False):
...
+        Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
+        >>> idx.rename('bar')
+        Int64Index([1, 2, 3, 4], dtype='int64', name='bar')
+
+        >>> idx = MultiIndex.from_tuples([(1, u'one'), (1, u'two'),
+                                          (2, u'one'), (2, u'two')],
+                                          names=['foo', 'bar'])
+        >>> idx.rename(['bar', None], inplace=True)
```
jorisvandenbossche: I would personally not show the inplace argument

pandas/core/indexes/base.py
```diff
@@ -1340,17 +1345,40 @@ def set_names(self, names, level=None, inplace=False):
     def rename(self, name, inplace=False):
         """
         Set new names on index. Defaults to returning new index.
+        Length of names must match number of levels in MultiIndex.
 
         Parameters
         ----------
-        name : str or list
-            name to set
+        name : str or sequence
```
jorisvandenbossche: is there a reason you changed 'list' to 'sequence' ? 
I know it technically doest accept any list, but in index names there *can* be a difference between a list (multiple names) and a tuple (single name), so I would rather keep it as list. 
But it is true that a name is not limited to a string, so maybe we should use the more general "label" (and then "label, or list of labels")

jorisvandenbossche (commented):
Thanks for the PR!

Can you check your docstring with `python scripts/validate_docstrings.py pandas.Index.rename` (and paste the output of it here in a comment) ?

pandas/core/indexes/base.py
Moons08: The reason why I used ‘sequence’ is that rename function takes not only list but also tuple in multiple names. 
I am not sure it is intended, but it is.

jorisvandenbossche: Yes, it is intended. But as I tried to explain, a tuple can also be a single name (so you can eg give a list of tuples). Therefore, I don't think the change is an improvement in clarity for the user. 
I would strictly use 'list' for referring to the ability to give multiple names.

Moons08: but it works like this.
```python
>>> idx = pd.MultiIndex(levels=[[1, 2], ['one', 'two']],\
...                     labels=[[0, 0, 1, 1], [0, 1, 0, 1]],\
...                     names=['bar', None])
>>> idx
MultiIndex(levels=[[1, 2], ['one', 'two']],
           labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
           names=['bar', None])
>>> idx.rename(('foo', 'bar'))
MultiIndex(levels=[[1, 2], ['one', 'two']],
           labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
           names=['foo', 'bar'])
```

jorisvandenbossche: Yes, it know it works, but we should not encourage that. 
For a single Index, there is a difference:

```
In [7]: idx = pd.Index([1, 2, 3, 4], name='foo')

In [8]: idx.rename(('bar', 'test'))
Out[8]: Int64Index([1, 2, 3, 4], dtype='int64', name=('bar', 'test'))

In [9]: idx.rename(['bar', 'test'])
...
TypeError: Int64Index.name must be a hashable type
```

So I would clearly state in our docstring one should use a list for multiple names (and not a sequence in general), to not confuse between those.

pandas/core/indexes/base.py
Moons08: I changed some line with your suggestion. 
and I put `\` in the MultiIndex example to avoid error. and what happened in validation error.. 
thank you so much.

```python
####################### Docstring (pandas.Index.rename)  #######################

Set new names on index. Defaults to returning new index.

Length of names must match number of levels in MultiIndex.

Parameters
----------
name : str or sequence
    name(s) to set
inplace : bool
    if True, mutates in place

Returns
-------
new index (of same type and class...etc) [if inplace, returns None]

Examples
--------
>>> idx = pd.Index([1, 2, 3, 4], name='foo')
>>> idx.rename('bar')
Int64Index([1, 2, 3, 4], dtype='int64', name='bar')

>>> idx = pd.MultiIndex.from_tuples([(1, u'one'), (1, u'two'),                                            (2, u'one'), (2, u'two')],                                            names=['foo', 'bar'])
>>> idx.rename(['bar', None], inplace=True)
>>> idx
MultiIndex(levels=[[1, 2], ['one', 'two']],
           labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
           names=['bar', None])
>>> idx.rename(['bar'])
ValueError: Length of names must match number of levels in MultiIndex.

See also
--------
set_names : able to set new names partially and by level

Errors found:
	Errors in parameters section
		Parameter "name" description should start with capital letter
		Parameter "name" description should finish with "."
		Parameter "inplace" description should start with capital letter
		Parameter "inplace" description should finish with "."
	Examples do not pass tests

Line 29, in pandas.Index.rename
Failed example:
    idx.rename(['bar'])
Exception raised:
      File "/home/mk/anaconda3/lib/python3.6/doctest.py", line 1330, in __run
        compileflags, 1), test.globs)
      File "<doctest pandas.Index.rename[5]>", line 1, in <module>
        idx.rename(['bar'])
      File "/home/mk/documents/dev/pandas-mk/pandas/core/indexes/base.py", line 1341, in set_names
        idx._set_names(names, level=level)
      File "/home/mk/documents/dev/pandas-mk/pandas/core/indexes/multi.py", line 672, in _set_names
        raise ValueError('Length of names must match number of levels in '
    ValueError: Length of names must match number of levels in MultiIndex.
```

pandas/core/indexes/base.py
Moons08: oh, I got it. so `list` means not only data type. thank you.

pandas/core/indexes/base.py
jorisvandenbossche: Can you fix the "Errors found:" ? (capitalization and punctuation)

pandas/core/indexes/base.py
```diff
@@ -1340,17 +1345,40 @@ def set_names(self, names, level=None, inplace=False):
...
+        Examples
+        --------
+        >>> idx = Index([1, 2, 3, 4], name='foo')
+        Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
+        >>> idx.rename('bar')
+        Int64Index([1, 2, 3, 4], dtype='int64', name='bar')
+
+        >>> idx = MultiIndex.from_tuples([(1, u'one'), (1, u'two'),
+                                          (2, u'one'), (2, u'two')],
+                                          names=['foo', 'bar'])
```
jorisvandenbossche: To get this passing for the doctests, you need to add `...` at the beginning of the line, as you see it in the console:

```
        >>> idx = MultiIndex.from_tuples([(1, u'one'), (1, u'two'),
        ...                               (2, u'one'), (2, u'two')],
        ...                               names=['foo', 'bar'])
```

pandas/core/indexes/base.py
```diff
@@ -1340,17 +1345,40 @@ def set_names(self, names, level=None, inplace=False):
...
+        >>> idx.rename(['bar', None], inplace=True)
+        >>> idx
+        MultiIndex(levels=[[1, 2], ['one', 'two']],
+                   labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
+                   names=['bar', None])
+        >>> idx.rename(['bar'])
+        ValueError: Length of names must match number of levels in MultiIndex.
```
jorisvandenbossche: To get this passing the doctests, see http://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html#tips-for-getting-your-examples-pass-the-doctests

Moons08:
I fixed some details in `Index.rename` and `Index.set_names`

```python
####################### Docstring (pandas.Index.rename)  #######################

Set new names on index. Defaults to returning new index.

Length of names must match number of levels in MultiIndex.

Parameters
----------
name : str or list
    Name(s) to set.
inplace : bool
    If True, mutates in place.

Returns
-------
new index (of same type and class...etc) [if inplace, returns None]

Examples
--------
>>> idx = pd.Index([1, 2, 3, 4], name='foo')
>>> idx.rename('bar')
Int64Index([1, 2, 3, 4], dtype='int64', name='bar')

>>> idx = pd.MultiIndex.from_tuples([(1, u'one'), (1, u'two'),
...                                  (2, u'one'), (2, u'two')],
...                                  names=['foo', 'bar'])
>>> idx.rename(['bar', None])
MultiIndex(levels=[[1, 2], ['one', 'two']],
           labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
           names=['bar', None])
>>> idx.rename(['bar'])
ValueError: Length of names must match number of levels in MultiIndex.

See also
--------
set_names : able to set new names partially and by level

Docstring for "pandas.Index.rename" correct. :)
```

```python
###################### Docstring (pandas.Index.set_names) ######################

Set new names on index. Defaults to returning new index.

Parameters
----------
names : str or sequence
    Name(s) to set.
level : int, level name, or sequence of int/level names (default None)
    If the index is a MultiIndex (hierarchical), level(s) to set (None
    for all levels).  Otherwise level must be None.
inplace : bool
    If True, mutates in place.

Returns
-------
new index (of same type and class...etc) [if inplace, returns None]

Examples
--------
>>> pd.Index([1, 2, 3, 4]).set_names('foo')
Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
>>> pd.Index([1, 2, 3, 4]).set_names(['foo'])
Int64Index([1, 2, 3, 4], dtype='int64', name='foo')

>>> idx = pd.MultiIndex.from_tuples([(1, u'one'), (1, u'two'),
...                                  (2, u'one'), (2, u'two')],
...                                  names=['foo', 'bar'])
>>> idx.set_names(['baz', 'quz'])
MultiIndex(levels=[[1, 2], ['one', 'two']],
           labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
           names=['baz', 'quz'])
>>> idx.set_names('baz', level=0)
MultiIndex(levels=[[1, 2], ['one', 'two']],
           labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
           names=['baz', 'bar'])

See also
--------
rename : able to set new names without level

Errors found:
	No extended summary found
```

jorisvandenbossche (approved):
Thanks a lot for the updates!

pandas/core/indexes/base.py
```diff
@@ -1283,12 +1283,12 @@ def set_names(self, names, level=None, inplace=False):
         Parameters
         ----------
         names : str or sequence
-            name(s) to set
+            Name(s) to set.
         level : int, level name, or sequence of int/level names (default None)
```
datapythonista: We try to have the types more than description in this line. Something like the next would be more standard:
`level: int, str, list of int or str, optional`

In the description you can expand on explaining that it is the level index or name.

https://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html#section-3-parameters

pandas/core/indexes/base.py
```diff
@@ -1283,12 +1283,12 @@ def set_names(self, names, level=None, inplace=False):
...
             If the index is a MultiIndex (hierarchical), level(s) to set (None
-            for all levels).  Otherwise level must be None
+            for all levels).  Otherwise level must be None.
         inplace : bool
-            if True, mutates in place
+            If True, mutates in place.
 
         Returns
         -------
```
datapythonista: Do you mind changing the Return to a type too. Instead of "new index...", the type `Index` in the first line, and the description in the next.

https://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html#section-4-returns-or-yields

pandas/core/indexes/base.py
```diff
@@ -1300,17 +1300,22 @@ def set_names(self, names, level=None, inplace=False):
...
         >>> idx.set_names('baz', level=0)
-        MultiIndex(levels=[[1, 2], [u'one', u'two']],
+        MultiIndex(levels=[[1, 2], ['one', 'two']],
                    labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
-                   names=[u'baz', u'bar'])
+                   names=['baz', 'bar'])
+
+        See also
+        --------
+        rename : able to set new names without level
```
datapythonista: See also should have a capital `A` (`See Also`) and should be placed before the examples. Also, the description should start with upper case and finish with a period.

Is `rename` the method of `Index`? For sphinx to create a link to it, it should be `Index.rename`. Just rename would be understood as `pandas.rename` which does not exist.

pandas/core/indexes/base.py
```diff
@@ -1341,16 +1346,39 @@ def rename(self, name, inplace=False):
...
+        MultiIndex(levels=[[1, 2], ['one', 'two']],
+                   labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
+                   names=['bar', None])
+        >>> idx.rename(['bar'])
+        Traceback (most recent call last):
+        ValueError: Length of names must match number of levels in MultiIndex.
+
+        See also
+        --------
+        set_names : able to set new names partially and by level
```
datapythonista: Same as before.

pandas/core/indexes/base.py
```diff
@@ -1300,17 +1300,22 @@ def set_names(self, names, level=None, inplace=False):
         Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
         >>> pd.Index([1, 2, 3, 4]).set_names(['foo'])
         Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
```
datapythonista: These 2 are duplicated, aren't they?

pandas/core/indexes/base.py
```diff
@@ -1300,17 +1300,22 @@ def set_names(self, names, level=None, inplace=False):
...
+
         >>> idx = pd.MultiIndex.from_tuples([(1, u'one'), (1, u'two'),
-                                            (2, u'one'), (2, u'two')],
-                                            names=['foo', 'bar'])
+        ...                                  (2, u'one'), (2, u'two')],
+        ...                                  names=['foo', 'bar'])
```
datapythonista: I think `names` is overindented by one space.

pandas/core/indexes/base.py
```diff
@@ -1300,17 +1300,22 @@ def set_names(self, names, level=None, inplace=False):
         Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
         >>> pd.Index([1, 2, 3, 4]).set_names(['foo'])
         Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
+
         >>> idx = pd.MultiIndex.from_tuples([(1, u'one'), (1, u'two'),
-                                            (2, u'one'), (2, u'two')],
-                                            names=['foo', 'bar'])
+        ...                                  (2, u'one'), (2, u'two')],
```
datapythonista: I think we can get rid of the unicode `u` at the beginning of the strings.

datapythonista (changes requested):
Thanks for the PR, added some minor comments.

Moons08:
I fixed detailed you said. and I changed one error msg to a proper one. 
And I got one question. 

now `Index.rename` get `level` parameter just `like set_names`. 
```python
>>> idx = pd.MultiIndex.from_tuples([(1, 'one'), (1, 'two'),
...                                  (2, 'one'), (2, 'two')],
...                                  names=['foo', 'bar'])
>>> idx.rename(1, level=0)
MultiIndex(levels=[[1, 2], ['one', 'two']],
           labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
           names=[1, 'bar'])
```
should we prevent rename from getting level parameter?

pandas/core/indexes/base.py
```diff
@@ -1278,39 +1278,46 @@ def _set_names(self, values, level=None):
...
-        names : str or sequence
-            name(s) to set
-        level : int, level name, or sequence of int/level names (default None)
-            If the index is a MultiIndex (hierarchical), level(s) to set (None
-            for all levels).  Otherwise level must be None
-        inplace : bool
-            if True, mutates in place
+        names : int, str, list-like, optional
+            Name(s) to set.
+        level : int, str, list-like, optional, default None
```
datapythonista: No need for the `default None` if we have `optional`

pandas/core/indexes/base.py
```diff
@@ -1278,39 +1278,46 @@ def _set_names(self, values, level=None):
...
+            If the index is a MultiIndex(hierarchical), level(s) to set (None
```
datapythonista: I don't find the `hierarchical` necessary.

pandas/core/indexes/base.py
```diff
@@ -1278,39 +1278,46 @@ def _set_names(self, values, level=None):
...
+            for all levels). Otherwise level must be None.
+        inplace : boolean, default False
+            Modifies the object directly, instead of creating a new Index or
+            MultiIndex.
 
         Returns
         -------
-        new index (of same type and class...etc) [if inplace, returns None]
+        renamed : Index, MultiIndex or None
```
datapythonista: Just the type is enough, not need for the `renamed`

pandas/core/indexes/base.py
```diff
@@ -1278,39 +1278,46 @@ def _set_names(self, values, level=None):
...
+            The same type as the caller or None if inplace is True.
+
+        See also
+        --------
+        Index.rename : Able to set new names without level
```
datapythonista: Can you add a period at the end?

pandas/core/indexes/base.py
```diff
@@ -1278,39 +1278,46 @@ def _set_names(self, values, level=None):
...
+            Modifies the object directly, instead of creating a new Index or
+            MultiIndex.
 
         Returns
         -------
-        new index (of same type and class...etc) [if inplace, returns None]
+        renamed : Index, MultiIndex or None
+            The same type as the caller or None if inplace is True.
+
+        See also
```
datapythonista: Capital `A` (`See Also`)

pandas/core/indexes/base.py
```diff
@@ -1278,39 +1278,46 @@ def _set_names(self, values, level=None):
...
         Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
-        >>> pd.Index([1, 2, 3, 4]).set_names(['foo'])
-        Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
-        >>> idx = pd.MultiIndex.from_tuples([(1, u'one'), (1, u'two'),
-                                            (2, u'one'), (2, u'two')],
-                                            names=['foo', 'bar'])
+
+        >>> idx = pd.MultiIndex.from_tuples([(1, 'one'), (1, 'two'),
+        ...                                  (2, 'one'), (2, 'two')],
+        ...                                  names=['foo', 'bar'])
```
datapythonista: I think in this case would help having a more real world example. At least for me is not so immediate to understand what we are dong, as replacing `foo` by `baz` is arbitrary. Having something like `['year', 'company']` changed to `['period', 'institution']` or something like this, IMO would make the example easier to understand.

pandas/core/indexes/base.py
```diff
@@ -1278,39 +1278,46 @@ def _set_names(self, values, level=None):
...
+            The same type as the caller or None if inplace is True.
+
+        See also
+        --------
+        Index.rename : Able to set new names without level
 
         Examples
         --------
         >>> pd.Index([1, 2, 3, 4]).set_names('foo')
         Int64Index([1, 2, 3, 4], dtype='int64', name='foo')
```
datapythonista: I'd assign the index to a variable `idx`, display it, and then display it after setting the name. This example gives the impression that this is the right way to create an `Index` with a name.

pandas/core/indexes/base.py
```diff
@@ -1319,7 +1326,8 @@ def set_names(self, names, level=None, inplace=False):
 
         if level is not None and not is_list_like(level) and is_list_like(
                 names):
-            raise TypeError("Names must be a string")
+            msg = "Length of names must match number of levels in MultiIndex."
```
datapythonista: I think the previous error message was more appropriate. May be better to say something like `Names must be a string when one level is provided`. But the new error message gives the impression that names should be a list, which is the not right.

pandas/core/indexes/base.py
```diff
@@ -1339,18 +1347,44 @@ def set_names(self, names, level=None, inplace=False):
...
 
         Returns
         -------
-        new index (of same type and class...etc) [if inplace, returns None]
+        renamed : Index, MultiIndex or None
+            The same type as the caller or None if inplace is True.
+
+        See also
+        --------
+        Index.set_names : Able to set new names partially and by level
```
datapythonista: Same as before for `renamed`, `See Also` and the period.

pandas/core/indexes/base.py
```diff
@@ -1339,18 +1347,44 @@ def set_names(self, names, level=None, inplace=False):
...
+        >>> idx = pd.MultiIndex.from_tuples([(1, 'one'), (1, 'two'),
+        ...                                  (2, 'one'), (2, 'two')],
+        ...                                  names=['foo', 'bar'])
+        >>> idx.rename(['bar', None])
+        MultiIndex(levels=[[1, 2], ['one', 'two']],
+                   labels=[[0, 0, 1, 1], [0, 1, 0, 1]],
+                   names=['bar', None])
+        >>> idx.rename(['bar'])
+        Traceback (most recent call last):
+        ValueError: Length of names must match number of levels in MultiIndex.
```
datapythonista: Same as before regarding the examples.

datapythonista (changes requested):
Great work. Added some comments that IMO would make the docstring better.

pandas/core/indexes/base.py
```diff
@@ -1278,39 +1278,46 @@ def _set_names(self, values, level=None):
...
         Parameters
         ----------
-        names : str or sequence
-            name(s) to set
-        level : int, level name, or sequence of int/level names (default None)
-            If the index is a MultiIndex (hierarchical), level(s) to set (None
-            for all levels).  Otherwise level must be None
-        inplace : bool
-            if True, mutates in place
+        names : int, str, list-like, optional
```
jorisvandenbossche: I don't think we should try to list the possible types here (`int, str`), as in principle an index name can be anything hashable, so not only ints or strings but also timestamps, tuples, ... 
I seem to remember that we had such a discussion during the documentation sprint recently, and we ended up using the generic "label" (and then it can be "label, or list of labels")

BTW, this one is not optional I think?

Moons08: right, it was ended with the generic "label". but @datapythonista said that we should follow this. 
https://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html#section-3-parameters

So I was not sure that using "label" in there is right. which one do you think better?
and yes, it is not optional. it because of my bad misunderstanding. I will delete it.

pandas/core/indexes/base.py
Moons08: good. I will go for it

pandas/core/indexes/base.py
jorisvandenbossche: Yeah, sorry, we should update that text

datapythonista: Yes, sorry, I meant `object` instead of `str`, this is what we've been using for what I know. We tried to keep them as Python types, but I'm happy using `label`, it may be clearer. But if we use that, I'd prefer `list of label` to `list of labels`. This is consistent with `list of str` or `list of int`, and it will make it easier to parse and detect types automatically, if at some point that is useful.

datapythonista: In this case I think `names` expects `label or list of label` if we go for the `label` option. Or `object or list of object` if we prefer the Python type. Am I right?

Moons08: I think `label or list of label` looks better

Moons08:
I am really sorry for this... I am gonna fix it

pandas/core/indexes/base.py
```diff
@@ -1278,39 +1278,49 @@ def _set_names(self, values, level=None):
...
-        names : str or sequence
-            name(s) to set
-        level : int, level name, or sequence of int/level names (default None)
-            If the index is a MultiIndex (hierarchical), level(s) to set (None
-            for all levels).  Otherwise level must be None
-        inplace : bool
-            if True, mutates in place
+        names : label or list of labels
+            Name(s) to set.
+        level : int, str or list of labels, optional
```
datapythonista: `label` instead of `str`

pandas/core/indexes/base.py
```diff
@@ -1278,39 +1278,49 @@ def _set_names(self, values, level=None):
...
-        MultiIndex(levels=[[1, 2], [u'one', u'two']],
+        >>> idx = pd.Index([1, 2, 3, 4])
+        >>> idx
+        Int64Index([1, 2, 3, 4], dtype='int64')
+        >>> idx.set_names('quarter')
+        Int64Index([1, 2, 3, 4], dtype='int64', name='quarter')
+
+        >>> idx = pd.MultiIndex.from_tuples([(1, 'one'), (1, 'two'),
+        ...                                  (2, 'one'), (2, 'two')],
+        ...                                  names=['year', 'company'])
```
datapythonista: I'd display `idx` after creating it. That will make it easier to compare with the result of `.set_names()`.

If we have `year` and `company`, I'd have something like `2018`, `Acme` in the values. I know I suggested it myself, but as I don't think we want to list company names, may be you can use something else?

pandas/core/indexes/base.py
```diff
@@ -1319,7 +1329,8 @@ def set_names(self, names, level=None, inplace=False):
 
         if level is not None and not is_list_like(level) and is_list_like(
                 names):
-            raise TypeError("Names must be a string")
+            msg = "Names must be a string when one level is provided."
```
datapythonista: "when a single level" sounds clearer to me.

datapythonista (commented):
Looks great, thanks for all the changes. Just couple of small things.

pandas/core/indexes/base.py
Moons08: right, maybe we can use programming languages such as python, java, etc.

datapythonista: I'd avoid any data that can lead to controversy, even if it's small. I use animals most of the times. :)

datapythonista:
Thanks @Moons08, very good work, very clear docstring.

Issue #21824, opened 2018-07-08
Segfault on clean-up with count example from docstrings


jorisvandenbossche:
In https://github.com/pandas-dev/pandas/pull/19952 I have problems with a segfault when running doctests, and I could isolate at least the following code snippet that segfaults on clean-up (so only when exiting the process; an explicit del + gc.collect does not yet trigger the segfault).

```
import numpy as  np
import pandas as pd

df = pd.DataFrame({"Person": ["John", "Myla", None, "John", "Myla"],
                   "Age": [24., np.nan, 21., 33, 26],
                   "Single": [False, True, True, True, False]})
                  
res = df.set_index(["Person", "Single"]).count(level="Person")
```

jorisvandenbossche:
OK, and the cause is the missing value in the 'Person' level. 

Simplified example:

```
In [1]: df = pd.DataFrame({"A": ["a", "b", None, "a", "b"],
   ...:                    "B": 1,
   ...:                    "C": range(5)})
   ...:                   
   ...: res = df.set_index(["A", "B"]).count(level="A")

In [2]: res
Out[2]: 
   B
A   
a  2
b  2
```

is causing the segfault (and without the None it does not). 
But, when selecting the column, it does not produce a segfault, but it also gives a different result (the NaN is included in index):

```
In [1]: df = pd.DataFrame({"A": ["a", "b", None, "a", "b"],
   ...:                    "B": 1,
   ...:                    "C": range(5)})
   ...:                   
   ...: res = df.set_index(["A", "B"])['C'].count(level="A")

In [2]: res
Out[2]: 
A
a      2
b      2
NaN    1
Name: C, dtype: int64
``` 

Here the segfault does not happen, but it seems buggy behaviour, as count should not count the NaNs ?

jreback:
is this example fixed? is this just a tracking issue for finding the cause?

@pandas-dev/pandas-core

gfyoung:
@jreback : Not really sure?  The examples don't seem to segfault for me...

cc @jorisvandenbossche

gfyoung:
Actually, I take that back.  They do.  Just as I exited the Python console.

gfyoung:
@jorisvandenbossche : This issue has gone quiet for a little while, but I tried running your examples again, and it seems like the second example you gave, while the first and third ones are still buggy...

That's quite odd.

gfyoung:
> as count should not count the NaNs?

[...]

PR #22368, opened 2018-08-15
DOC: edit docstring example to prevent segfault (#21824)

jorisvandenbossche:
See https://github.com/pandas-dev/pandas/issues/21824 for the segfault description, changing the example so running the doctests will work: https://github.com/pandas-dev/pandas/pull/19952

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -6896,21 +6896,21 @@ class DataFrame(NDFrame):
         Constructing DataFrame from a dictionary:
 
         >>> df = pd.DataFrame({"Person":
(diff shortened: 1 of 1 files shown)


jbrockmendel:
Nice to see this fixed.  Did you figure out what about the existing version was causing the problem?

jorisvandenbossche:
> Did you figure out what about the existing version was causing the problem?

No, see the details in #21824 for the segfault. This PR is simply changing the example to not run into the problem ..

jorisvandenbossche:
Merged this already (I don't think it is in any way controversial, and only doc change), as I want to test if now the doc tests works

PR #22375, opened 2018-08-16
ENH: Enable DataFrame.corrwith to compute rank correlations

### LINKED ISSUE #21925: Allow different methods of correlation when using corrwith

### LINKED ISSUE #22328: DataFrame.corrwith has unintuitive behavior
```python
import numpy as np
import pandas as pd

np.random.seed(2357)

df1 = pd.DataFrame(np.random.normal(size=(100, 3)))
df2 = pd.DataFrame(np.random.normal(size=(100, 3)))

# not clear what this means
df1.corrwith(df2)
# simply renaming columns produces nans
df2.columns = ["a", "b", "c"]
df1.corrwith(df2)
# this should arguably be the same as df1.corr()
df1.corrwith(df1)
```
#### Problem description

The docstring for `corrwith` describes its behavior as:

    Compute pairwise correlation between rows or columns of two DataFrame
    objects.

My interpretation of this would be that `corrwith` finds *all* pairwise correlations between the two DataFrames along the given axis (it's also defined for Series input which should probably be a bit more explicit in the docstring).  That is (for axis=0) df1.corrwith(df2) would be functionally equivalent to

```
df2.apply(lambda x: df1.corrwith(x))
```

However, as implemented the method returns a single Series which appears to give specific pairwise correlations, and how the pairs are chosen does not appear to be documented.

Defining `corrwith` recursively as above seems more useful (you still retain all the information from the current implementation) and interpretable, fixes the apparent bug caused when the method tries to align columns, and also allows for an easy way of closing issue #21925 (since ultimately the method falls back on Series.corr which permits control of the correlation type).  I can start to work on a pull request if others agree that this would be preferred.

#### Expected Output

```python
# df1.corrwith(df2) should be equal to this in my opinion
df2.apply(lambda x: df1.corrwith(x))
# this yields the ordinary correlation matrix as expected, although it is slower than just df1.corr()
df1.apply(lambda x: df1.corrwith(x))
```

<details>
INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Darwin
OS-release: 17.6.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

pandas: 0.23.0
pytest: 3.7.1
pip: 18.0
setuptools: 39.1.0
Cython: 0.28.2
numpy: 1.14.4
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: 1.7.4
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.2.1
tables: 3.4.3
numexpr: 2.6.5
feather: None
matplotlib: 2.2.2
openpyxl: 2.5.3
xlrd: 1.1.0
xlwt: 1.2.0
xlsxwriter: 1.0.4
lxml: 4.2.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.7
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

dsaxton:
This PR is to enable `DataFrame.corrwith` to calculate rank correlations in addition to Pearson's correlation (and should hopefully be fully backwards compatible), as well as clarifying functionality in the docstring .  An error message is generated if the user specifies a form of correlation that isn't implemented.  Tests were added for the new behavior.

closes #22328 
closes #21925

### DIFF
--- a/asv_bench/benchmarks/stat_ops.py
+++ b/asv_bench/benchmarks/stat_ops.py
@@ -106,6 +106,7 @@ class Correlation(object):
             from pandas.core import nanops
             nanops._USE_BOTTLENECK = use_bottleneck
         self.df = pd.DataFrame(np.random.randn(1000, 30))
+        self.df2 = pd.DataFrame(np.random.randn(1000, 30))
         self.s = pd.Series(np.random.randn(1000))
         self.s2 = pd.Series(np.random.randn(1000))
 
@@ -114,6 +115,12 @@ class Correlation(object):
 
     def time_corr_series(self, method, use_bottleneck):
         self.s.corr(self.s2, method=method)
+
+    def time_corrwith_cols(self, method, use_bottleneck):
+        self.df.corrwith(self.df2, method=method)
+
+    def time_corrwith_rows(self, method, use_bottleneck):
+        self.df.corrwith(self.df2, axis=1, method=method)
 
 
 class Covariance(object):
--- a/doc/source/whatsnew/v0.24.0.rst
+++ b/doc/source/whatsnew/v0.24.0.rst
@@ -413,6 +413,7 @@ Other Enhancements
 - The ``scatter_matrix``, ``andrews_curves``, ``parallel_coordinates``, ``lag_plot``, ``autocorrelation_plot``, ``bootstrap_plot``, and ``radviz`` plots from the ``pandas.plotting`` module are now accessible from calling :meth:`DataFrame.plot` (:issue:`11978`)
 - :class:`IntervalIndex` has gained the :attr:`~IntervalIndex.is_overlapping` attribute to indicate if the ``IntervalIndex`` contains any overlapping intervals (:issue:`23309`)
 - :func:`pandas.DataFrame.to_sql` has gained the ``method`` argument to control SQL insertion clause. See the :ref:`insertion method <io.sql.method>` section in the documentation. (:issue:`8953`)
+- :meth:`DataFrame.corrwith` now supports Spearman's rank correlation, Kendall's tau as well as callable correlation methods. (:issue:`21925`)
 
 .. _whatsnew_0240.api_breaking:
 
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -7084,14 +7084,16 @@ class DataFrame(NDFrame):
 
             correl = num / dom
 
-        else:
+        elif method in ['kendall', 'spearman'] or callable(method):
             def c(x):
-                return Series(x[0]).corr(Series(x[1]),
-                                         method=method)
+                return nanops.nancorr(x[0], x[1], method=method)
 
             correl = Series(map(c,
                                 zip(left.values.T, right.values.T)),
                             index=left.columns)
+
+        else:
+            raise ValueError('Invalid method')
 
         if not drop:
             raxis = 1 if axis == 0 else 0
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -7093,7 +7093,10 @@ class DataFrame(NDFrame):
                             index=left.columns)
 
         else:
-            raise ValueError('Invalid method')
+            raise ValueError("Invalid method {method} was passed, "
+                             "valid methods are: 'pearson', 'kendall', "
+                             "'spearman', or callable".
+                             format(method=str(method)))
 
         if not drop:
             raxis = 1 if axis == 0 else 0
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -7096,15 +7096,19 @@ class DataFrame(NDFrame):
             raise ValueError("Invalid method {method} was passed, "
                              "valid methods are: 'pearson', 'kendall', "
                              "'spearman', or callable".
-                             format(method=str(method)))
+                             format(method=method))
 
         if not drop:
+            # Find non-matching labels along the given axis
+            # and append missing correlations (GH 22375)
             raxis = 1 if axis == 0 else 0
             result_index = (this._get_axis(raxis).
                             union(other._get_axis(raxis)))
             idx_diff = result_index.difference(correl.index)
-            correl = correl.append(Series([np.nan] * len(idx_diff),
-                                          index=idx_diff))
+
+            if len(idx_diff) > 0:
+                correl = correl.append(Series([np.nan] * len(idx_diff),
+                                              index=idx_diff))
 
         return correl
 
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -7103,8 +7103,8 @@ class DataFrame(NDFrame):
             result_index = (this._get_axis(raxis).
                             union(other._get_axis(raxis)))
             idx_diff = result_index.difference(correl.index)
-            correl = correl.append(pd.Series([np.nan] * len(idx_diff),
-                                             index=idx_diff))
+            correl = correl.append(Series([np.nan] * len(idx_diff),
+                                          index=idx_diff))
 
         return correl
 
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -7102,7 +7102,9 @@ class DataFrame(NDFrame):
             raxis = 1 if axis == 0 else 0
             result_index = (this._get_axis(raxis).
                             union(other._get_axis(raxis)))
-            correl = correl.reindex(result_index)
+            idx_diff = result_index.difference(correl.index)
+            correl = correl.append(pd.Series([np.nan] * len(idx_diff),
+                                             index=idx_diff))
 
         return correl
 
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -7056,11 +7056,6 @@ class DataFrame(NDFrame):
         -------
         DataFrame.corr
         """
-        if method not in ['pearson', 'spearman', 'kendall']:
-            raise ValueError("method must be either 'pearson', "
-                             "'spearman', or 'kendall', '{method}' "
-                             "was supplied".format(method=method))
-
         axis = self._get_axis_number(axis)
         this = self._get_numeric_data()
 
@@ -7080,10 +7075,6 @@ class DataFrame(NDFrame):
             left = left + right * 0
             right = right + left * 0
 
-            if axis == 1:
-                left = left.T
-                right = right.T
-
             # demeaned data
             ldem = left - left.mean()
             rdem = right - right.mean()
@@ -7092,12 +7083,6 @@ class DataFrame(NDFrame):
             dom = (left.count() - 1) * left.std() * right.std()
 
             correl = num / dom
-
-            if not drop:
-                raxis = 1 if axis == 0 else 0
-                result_index = (this._get_axis(raxis).
-                                union(other._get_axis(raxis)))
-                correl = correl.reindex(result_index)
 
         else:
             def c(x):
@@ -7108,8 +7093,11 @@ class DataFrame(NDFrame):
                                 zip(left.values.T, right.values.T)),
                             index=left.columns)
 
-            if drop:
-                correl.dropna(inplace=True)
+        if not drop:
+            raxis = 1 if axis == 0 else 0
+            result_index = (this._get_axis(raxis).
+                            union(other._get_axis(raxis)))
+            correl = correl.reindex(result_index)
 
         return correl
 
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -6868,6 +6868,11 @@ class DataFrame(NDFrame):
               dogs cats
         dogs   1.0  0.3
         cats   0.3  1.0
+
+        See Also
+        -------
+        DataFrame.corrwith
+        Series.corr
         """
         numeric_df = self._get_numeric_data()
         cols = numeric_df.columns
@@ -7021,10 +7026,11 @@ class DataFrame(NDFrame):
 
         return self._constructor(baseCov, index=idx, columns=cols)
 
-    def corrwith(self, other, axis=0, drop=False):
+    def corrwith(self, other, axis=0, drop=False, method='pearson'):
         """
-        Compute pairwise correlation between rows or columns of two DataFrame
-        objects.
+        Compute pairwise correlation between rows or columns of DataFrame
+        with rows or columns of Series or DataFrame.  DataFrames are first
+        aligned along both axes before computing the correlations.
 
         Parameters
         ----------
@@ -7032,43 +7038,78 @@ class DataFrame(NDFrame):
         axis : {0 or 'index', 1 or 'columns'}, default 0
             0 or 'index' to compute column-wise, 1 or 'columns' for row-wise
         drop : boolean, default False
-            Drop missing indices from result, default returns union of all
+            Drop missing indices from result
+        method : {'pearson', 'kendall', 'spearman'} or callable
+            * pearson : standard correlation coefficient
+            * kendall : Kendall Tau correlation coefficient
+            * spearman : Spearman rank correlation
+            * callable: callable with input two 1d ndarrays
+                and returning a float
+
+            .. versionadded:: 0.24.0
 
         Returns
         -------
         correls : Series
+
+        See Also
+        -------
+        DataFrame.corr
         """
+        if method not in ['pearson', 'spearman', 'kendall']:
+            raise ValueError("method must be either 'pearson', "
+                             "'spearman', or 'kendall', '{method}' "
+                             "was supplied".format(method=method))
+
         axis = self._get_axis_number(axis)
         this = self._get_numeric_data()
 
         if isinstance(other, Series):
-            return this.apply(other.corr, axis=axis)
+            return this.apply(lambda x: other.corr(x, method=method),
+                              axis=axis)
 
         other = other._get_numeric_data()
-
         left, right = this.align(other, join='inner', copy=False)
-
-        # mask missing values
-        left = left + right * 0
-        right = right + left * 0
 
         if axis == 1:
             left = left.T
             right = right.T
 
-        # demeaned data
-        ldem = left - left.mean()
-        rdem = right - right.mean()
+        if method == 'pearson':
+            # mask missing values
+            left = left + right * 0
+            right = right + left * 0
 
-        num = (ldem * rdem).sum()
-        dom = (left.count() - 1) * left.std() * right.std()
+            if axis == 1:
+                left = left.T
+                right = right.T
 
-        correl = num / dom
+            # demeaned data
+            ldem = left - left.mean()
+            rdem = right - right.mean()
 
-        if not drop:
-            raxis = 1 if axis == 0 else 0
-            result_index = this._get_axis(raxis).union(other._get_axis(raxis))
-            correl = correl.reindex(result_index)
+            num = (ldem * rdem).sum()
+            dom = (left.count() - 1) * left.std() * right.std()
+
+            correl = num / dom
+
+            if not drop:
+                raxis = 1 if axis == 0 else 0
+                result_index = (this._get_axis(raxis).
+                                union(other._get_axis(raxis)))
+                correl = correl.reindex(result_index)
+
+        else:
+            def c(x):
+                return Series(x[0]).corr(Series(x[1]),
+                                         method=method)
+
+            correl = Series(map(c,
+                                zip(left.values.T, right.values.T)),
+                            index=left.columns)
+
+            if drop:
+                correl.dropna(inplace=True)
 
         return correl
 
--- a/pandas/tests/frame/test_analytics.py
+++ b/pandas/tests/frame/test_analytics.py
@@ -466,6 +466,7 @@ class TestDataFrameAnalytics():
         expected = pd.Series(data=corrs, index=['a', 'b'])
         tm.assert_series_equal(result, expected)
 
+    @pytest.mark.xfail
     def test_corrwith_dup_cols(self):
         # GH 21925
         df1 = pd.DataFrame(np.vstack([np.arange(10)] * 3).T)
--- a/pandas/tests/frame/test_analytics.py
+++ b/pandas/tests/frame/test_analytics.py
@@ -466,7 +466,6 @@ class TestDataFrameAnalytics():
         expected = pd.Series(data=corrs, index=['a', 'b'])
         tm.assert_series_equal(result, expected)
 
-    @pytest.mark.xfail
     def test_corrwith_dup_cols(self):
         # GH 21925
         df1 = pd.DataFrame(np.vstack([np.arange(10)] * 3).T)
--- a/pandas/tests/frame/test_analytics.py
+++ b/pandas/tests/frame/test_analytics.py
@@ -466,6 +466,41 @@ class TestDataFrameAnalytics():
         expected = pd.Series(data=corrs, index=['a', 'b'])
         tm.assert_series_equal(result, expected)
 
+    def test_corrwith_dup_cols(self):
+        # GH 21925
+        df1 = pd.DataFrame(np.vstack([np.arange(10)] * 3).T)
+        df2 = df1.copy()
+        df2 = pd.concat((df2, df2[0]), axis=1)
+
+        result = df1.corrwith(df2).values
+        expected = np.ones(4)
+        tm.assert_almost_equal(result, expected)
+
+    @td.skip_if_no_scipy
+    def test_corrwith_spearman(self):
+        # GH 21925
+        df = pd.DataFrame(np.random.random(size=(100, 3)))
+        result = df.corrwith(df**2, method="spearman")
+        expected = Series(np.ones(len(result)))
+        tm.assert_series_equal(result, expected)
+
+    @td.skip_if_no_scipy
+    def test_corrwith_kendall(self):
+        # GH 21925
+        df = pd.DataFrame(np.random.random(size=(100, 3)))
+        result = df.corrwith(df**2, method="kendall")
+        expected = Series(np.ones(len(result)))
+        tm.assert_series_equal(result, expected)
+
+    def test_corrwith_invalid_method(self):
+        # GH 21925
+        df = pd.DataFrame(np.random.normal(size=(10, 2)))
+        s = pd.Series(np.random.randn(10))
+        msg = ("method must be either 'pearson', 'spearman', "
+               "or 'kendall'")
+        with tm.assert_raises_regex(ValueError, msg):
+            df.corrwith(s, method="____")
+
     def test_bool_describe_in_mixed_frame(self):
         df = DataFrame({
             'string_data': ['a', 'b', 'c', 'd', 'e'],
--- a/pandas/tests/frame/test_analytics.py
+++ b/pandas/tests/frame/test_analytics.py
@@ -493,15 +493,6 @@ class TestDataFrameAnalytics():
         expected = Series(np.ones(len(result)))
         tm.assert_series_equal(result, expected)
 
-    def test_corrwith_invalid_method(self):
-        # GH 21925
-        df = pd.DataFrame(np.random.normal(size=(10, 2)))
-        s = pd.Series(np.random.randn(10))
-        msg = ("method must be either 'pearson', 'spearman', "
-               "or 'kendall'")
-        with tm.assert_raises_regex(ValueError, msg):
-            df.corrwith(s, method="____")
-
     def test_bool_describe_in_mixed_frame(self):
         df = DataFrame({
             'string_data': ['a', 'b', 'c', 'd', 'e'],
--- a/pandas/tests/frame/test_analytics.py
+++ b/pandas/tests/frame/test_analytics.py
@@ -473,9 +473,9 @@ class TestDataFrameAnalytics():
         df2 = df1.copy()
         df2 = pd.concat((df2, df2[0]), axis=1)
 
-        result = df1.corrwith(df2).values
-        expected = np.ones(4)
-        tm.assert_almost_equal(result, expected)
+        result = df1.corrwith(df2)
+        expected = pd.Series(np.ones(4), index=[0, 0, 1, 2])
+        tm.assert_series_equal(result, expected)
 
     @td.skip_if_no_scipy
     def test_corrwith_spearman(self):


jreback (changes requested):
since you are changing the impl can you run the asv for perf check

dsaxton:
I'm having some trouble finding a relevant benchmark for `corrwith` (a `grep` on the benchmarks folder returns no hits).  Should one be added to the `Correlation` class in `stats_ops.py`?

dsaxton:
Here is one I put together which seems to suggest around a 33-50% speed-up with the new implementation.

```python
import numpy as np
from pandas import DataFrame, Series

np.random.seed(2357)

def corrwith_old(df1, df2, axis=0, drop=False):
    axis = df1._get_axis_number(axis)
    this = df1._get_numeric_data()

    if isinstance(df2, Series):
        return this.apply(df2.corr, axis=axis)

    df2 = df2._get_numeric_data()

    left, right = this.align(df2, join='inner', copy=False)

    # mask missing values
    left = left + right * 0
    right = right + left * 0

    if axis == 1:
        left = left.T
        right = right.T

    # demeaned data
    ldem = left - left.mean()
    rdem = right - right.mean()

    num = (ldem * rdem).sum()
    dom = (left.count() - 1) * left.std() * right.std()

    correl = num / dom

    if not drop:
        raxis = 1 if axis == 0 else 0
        result_index = this._get_axis(raxis).union(df2._get_axis(raxis))
        correl = correl.reindex(result_index)

    return correl


def corrwith_new(df1, df2, axis=0, drop=False, method='pearson'):
    if method not in ['pearson', 'spearman', 'kendall']:
        raise ValueError("method must be either 'pearson', "
                         "'spearman', or 'kendall', '{method}' "
                         "was supplied".format(method=method))

    axis = df1._get_axis_number(axis)
    this = df1._get_numeric_data()

    if isinstance(df2, Series):
        return this.apply(lambda x: df2.corr(x, method=method),
                          axis=axis)

    df2 = df2._get_numeric_data()
    left, right = this.align(df2, join='inner', copy=False)

    if axis == 1:
        left = left.T
        right = right.T

    correl = (left.apply(lambda x:
                         x.corr(right[x.name], method=method)))

    if drop:
        correl.dropna(inplace=True)

    return correl


data1 = DataFrame(np.random.normal(size=(10**6, 10)))
data2 = DataFrame(np.random.normal(size=(10**6, 10)))

%timeit corrwith_old(data1, data2)
%timeit corrwith_new(data1, data2)
```

jreback:
can u add this as an asv

in particular i suspect you method will slow down when used with a larger number of columns 

so pls show that as well

dsaxton:
I've added the new benchmark function `time_corrwith` to `stats_ops.py` within the `Correlation` class.  I'll try to run the benchmark locally (although I botched the setup of `asv` on my main computer and it seems to be broken).

dsaxton:
I think your suspicion may have been accurate regarding the dimension of the inputs.  In any case, regarding the tests it looks like there's an issue with `scipy` during the travis build.  Am I missing an import somewhere perhaps?

WillAyd:
@dsaxton we test quite a few different configurations, not all of which have SciPy installed. If your tests depend on that package you should use the `skip_if_no_scipy` decorate which you'll see used in other places in that same module

asv_bench/benchmarks/stat_ops.py
```diff
@@ -109,6 +109,10 @@ class Correlation(object):
 
     def setup(self, method):
         self.df = pd.DataFrame(np.random.randn(1000, 30))
+        self.df2 = pd.DataFrame(np.random.randn(1000, 30))
 
     def time_corr(self, method):
         self.df.corr(method=method)
+
+    def time_corrwith(self, method):
+        self.df.corrwith(df2, method=method)
```
mroeschke: `df2` should be `self.df2` here.

dsaxton: Good catch, thank you.

dsaxton:
@jreback If a certain benchmark is actually part of the PR, what would you say is the most straightforward way to show that there isn't a degradation in performance?  Here the `time_corrwith` function will not work for the current implementation because it uses all three correlation types.  I suppose it could temporarily be modified to look only at Pearson?

WillAyd:
You should have for all three. The two new ones will fail since they don't exist on master but that's OK - still sets a baseline going forward and we can still get insights out of the Pearson benchmark

dsaxton:
screen shot 2018-08-16 at 4 31 22 pm

Here are the results I get running the `stat_ops.Correlation` benchmarks (all `DataFrames` involved are comprised of 1000 x 30 Gaussian arrays).  The Pearson calculation somehow took longer than that of `DataFrame.corr`, even though the latter should actually be doing more work since it computes the full 30 x 30 correlation matrix.  I'll need to run it on the current `corrwith` to see how that performs.

dsaxton:
I modified the body of the function to mimic how the current calculation is being done in the special case where `method=pearson` (in the event that there was somehow an overhead associated with repeatedly calling `Series.corr`) and I saw essentially the same performance as before (shown below).  My guess then is that `DataFrame.corr` gets its speed from `nancorr` within `pandas._libs.algos`.  Could using functions from `algos` inside `corrwith` as well make sense?

```python
        if method in ['spearman', 'kendall']:
            correl = (left.apply(lambda x:
                                 x.corr(right[x.name], method=method)))
        else:
            correl = (((right - right.mean()) * (left - left.mean())).mean()
                      / right.std() / left.std())
```

screen shot 2018-08-16 at 6 52 16 pm

jreback (changes requested):
pls add a whatsnew note in other enhancements

pandas/core/frame.py
```diff
@@ -6655,6 +6655,11 @@ def corr(self, method='pearson', min_periods=1):
         Returns
         -------
         y : DataFrame
+
+        See Also
+        -------
+        DataFrame.corrwith
```
jreback: can you add Series.corr here as well

jreback (changes requested):
what do the asv's show ?

jreback (changes requested):
pls add a whatsnew as well (with both issues)

pandas/tests/frame/test_analytics.py
```diff
@@ -248,6 +248,31 @@ def test_corrwith_mixed_dtypes(self):
         expected = pd.Series(data=corrs, index=['a', 'b'])
         tm.assert_series_equal(result, expected)
 
+    @td.skip_if_no_scipy
+    def test_corrwith_spearman(self):
+        # GH PR #22375
```
jreback: change these to the issue numbers, rather than the PR number

jreback (changes requested):

jreback:
don't post pictures of the asv's just copy-paste the text pls.

dsaxton:
`asv` results for the `Correlation` class (`DataFrames` are built from `np.random.randn(1000, 30)`):

```
> asv continuous -f 1.1 origin/corrwith-dev HEAD -b stat_ops.Correlation
· Creating environments
· Discovering benchmarks
·· Uninstalling from conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt.
·· Building a511a498 for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt........................................
·· Installing into conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt..
· Running 6 total benchmarks (2 commits * 1 environments * 3 benchmarks)
[  0.00%] · For pandas commit hash a511a498:
[  0.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[  0.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[  0.00%] ··· Running (stat_ops.Correlation.time_corr--)...
[ 16.67%] ··· stat_ops.Correlation.time_corr                                                                                         ok
[ 16.67%] ··· ========== =============
                method                
              ---------- -------------
               spearman     93.6±2ms  
               kendall      219±4ms   
               pearson    2.14±0.09ms 
              ========== =============

[ 33.33%] ··· stat_ops.Correlation.time_corrwith_cols                                                                                ok
[ 33.33%] ··· ========== ============
                method               
              ---------- ------------
               spearman    51.8±1ms  
               kendall     20.6±1ms  
               pearson    17.0±0.7ms 
              ========== ============

[ 50.00%] ··· stat_ops.Correlation.time_corrwith_rows                                                                                ok
[ 50.00%] ··· ========== ==========
                method             
              ---------- ----------
               spearman   758±30ms 
               kendall    493±7ms  
               pearson    258±7ms  
              ========== ==========

[ 50.00%] · For pandas commit hash a511a498:
[ 50.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[ 50.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[ 50.00%] ··· Running (stat_ops.Correlation.time_corr--)...
[ 66.67%] ··· stat_ops.Correlation.time_corr                                                                                         ok
[ 66.67%] ··· ========== =============
                method                
              ---------- -------------
               spearman     101±6ms   
               kendall      226±9ms   
               pearson    2.16±0.03ms 
              ========== =============

[ 83.33%] ··· stat_ops.Correlation.time_corrwith_cols                                                                                ok
[ 83.33%] ··· ========== ============
                method               
              ---------- ------------
               spearman    52.1±4ms  
               kendall     21.1±2ms  
               pearson    16.5±0.6ms 
              ========== ============

[100.00%] ··· stat_ops.Correlation.time_corrwith_rows                                                                                ok
[100.00%] ··· ========== ==========
                method             
              ---------- ----------
               spearman   752±20ms 
               kendall    490±4ms  
               pearson    266±3ms  
              ========== ==========


BENCHMARKS NOT SIGNIFICANTLY CHANGED.
> 
```

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -182,6 +182,7 @@ Other Enhancements
 - :func:`to_timedelta` now supports iso-formated timedelta strings (:issue:`21877`)
 - :class:`Series` and :class:`DataFrame` now support :class:`Iterable` in constructor (:issue:`2193`)
 - :class:`DatetimeIndex` gained :attr:`DatetimeIndex.timetz` attribute. Returns local time with timezone information. (:issue:`21358`)
+- :meth:`DataFrame.corrwith` now supports Spearman's rank correlation as well as Kendall's tau. (:issue:`21925`)
```
TomAugspurger: This makes is sound like Knedall's tau was the previously supported method. Maybe "... supports Spearman's rank correlation and Kendall's tau for `method`"

pandas/core/frame.py
```diff
@@ -6804,54 +6810,58 @@ def cov(self, min_periods=None):
...
             right = right.T
 
-        # demeaned data
-        ldem = left - left.mean()
-        rdem = right - right.mean()
-
-        num = (ldem * rdem).sum()
-        dom = (left.count() - 1) * left.std() * right.std()
+        correl = (left.apply(lambda x:
+                             x.corr(right[x.name], method=method)))
```
TomAugspurger: Does this break when left or right have duplicate columns? Can you add a test ensuring that it works? It seems like we support that on master.

doc/source/whatsnew/v0.24.0.txt
dsaxton: Good point, will make that change.

pandas/core/frame.py
dsaxton: I'll test that out.  What would you say the expected behavior should be in this case?  Would it make sense to first drop duplicate columns from `self` and `other`?

TomAugspurger:
Duplicate column names doesn't necessarily mean duplicate values.

We handle this correctly on master, so you should be able to add a test on
master that passes, and ensure it still passes on your branch.

pandas/core/frame.py
dsaxton: Yes, I believe this breaks with duplicate columns.  There might be a way to still rely on `apply`, but if not I'll try to think of something else.  In the worst case a `for` loop should work, but I'd like to avoid that if possible.

dsaxton: @TomAugspurger I think this is actually an error on `master`, but I may have a fix.

```
import numpy as np
import pandas as pd

df1 = pd.DataFrame(np.random.normal(size=(10, 3)))
df2 = pd.DataFrame(np.random.normal(size=(10, 3)))
df2 = pd.concat((df2, pd.DataFrame(np.random.randn(10))), axis=1)

df1.corrwith(df2)
```

A possible fix within `corrwith` is to create `correl` by applying a `map` to the `zipped` columns

```
pd.Series(map(lambda x: pd.Series(x[0]).corr(pd.Series(x[1]), method="spearman"),
              zip(left.values.T, right.values.T)),
          index=left.columns)
```

I can commit this as a change if you think it makes sense.

dsaxton:
@TomAugspurger I just pushed some changes to that included the above (`zip` / `map` instead of `apply`, more clear `whatsnew`), and also added a relatively simple test to check if the method can handle duplicate columns (below).  I may have mauled the `git` history a bit with all the merging, so a `rebase` might be needed.

Regarding the test, I am just creating a `DataFrame` with three columns equal to the integers `0, 1, ... , 9`, then creating another which adds an additional column of the same form with a duplicate column name.  The result of `corrwith` should be a `Series` of four ones (and should not generate an error).  Please let me know if there's a more straightforward way to simply check that the method doesn't err out.

```python
    def test_corrwith_dup_cols(self):
        # GH 21925
        df1 = pd.DataFrame(np.vstack([np.arange(10)] * 3).T)
        df2 = df1.copy()
        df2 = pd.concat((df2, df2[0]), axis=1)

        result = df1.corrwith(df2).values
        expected = np.ones(4)
        tm.assert_almost_equal(result, expected)
```

jreback:
can you rebase

jreback (changes requested):
what is the perf diff with this (also add something with a lot of columns, like 100) to see what effect that has

jreback:
see #22684, a nice alternative

dsaxton:
@jreback  Should this PR still be implemented, or does #22684 make it obsolete?  (Since #22684 seems to be changing the implementation of `corr`, does it make sense to extend it to `corrwith`?)

dsaxton:
I'm not entirely sure how to show a perf diff, but here is something I put together that seems to show that the new implementation is faster than `corrwith` on `master` for Pearson's correlation, at least on my local machine. (Here we just have two scripts that only perform the part of the calculation that is different between `master` and this branch.  Each constructs two 100,000 x 100 `DataFrames` and then finds the correlation `Series` returned by `corrwith`.)

```
Daniels-MacBook-Pro:~ dsaxton$ cat perf1.py 
#!/usr/bin/env python3

import numpy as np
from pandas import DataFrame, Series

left = DataFrame(np.random.normal(size=(100000, 100)))
right = DataFrame(np.random.normal(size=(100000, 100)))

def c(x):
    return Series(x[0]).corr(Series(x[1]),
                  method="pearson")

correl = Series(map(c,
                    zip(left.values.T, right.values.T)),
                index=left.columns)
Daniels-MacBook-Pro:~ dsaxton$ cat perf2.py 
#!/usr/bin/env python3

import numpy as np
from pandas import DataFrame, Series

left = DataFrame(np.random.normal(size=(100000, 100)))
right = DataFrame(np.random.normal(size=(100000, 100)))

ldem = left - left.mean()
rdem = right - right.mean()

num = (ldem * rdem).sum()
dom = (left.count() - 1) * left.std() * right.std()

correl = num / dom
Daniels-MacBook-Pro:~ dsaxton$ time ./perf1.py 

real	0m2.025s
user	0m1.877s
sys	0m0.139s
Daniels-MacBook-Pro:~ dsaxton$ time ./perf2.py 

real	0m2.291s
user	0m2.128s
sys	0m0.740s
```

jreback:
pls look at the contributing docs on running the performance suite

the new impl does not scale with the number of columns - pls show this

dsaxton:
I think you're correct that the proposed implementation is not scaling.  Could it be due to `zip` and `map` not being as efficient as pure `numpy`?

```
Daniels-MacBook-Pro:asv_bench dsaxton$ asv continuous -f 1.1 upstream/master corrwith-dev -b stat_ops.Correlation
· Creating environments
· Discovering benchmarks
·· Uninstalling from conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt.
·· Installing into conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt..
· Running 8 total benchmarks (2 commits * 1 environments * 4 benchmarks)
[  0.00%] · For pandas commit hash e19b3f62:
[  0.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[  0.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[  0.00%] ··· Running (stat_ops.Correlation.time_corr--)....
[ 12.50%] ··· stat_ops.Correlation.time_corr                                                                                                                         ok
[ 12.50%] ··· ========== ============
                method               
              ---------- ------------
               spearman    93.5±4ms  
               kendall     206±20ms  
               pearson    1.95±0.2ms 
              ========== ============

[ 25.00%] ··· stat_ops.Correlation.time_corrwith                                                                                                                     ok
[ 25.00%] ··· ========== ==========
                method             
              ---------- ----------
               spearman   39.6±5ms 
               kendall    35.0±2ms 
               pearson    35.9±3ms 
              ========== ==========

[ 37.50%] ··· stat_ops.Correlation.time_corrwith_cols                                                                                                                ok
[ 37.50%] ··· ========== ==========
                method             
              ---------- ----------
               spearman   115±6ms  
               kendall    19.5±1ms 
               pearson    36.4±2ms 
              ========== ==========

[ 50.00%] ··· stat_ops.Correlation.time_corrwith_rows                                                                                                                ok
[ 50.00%] ··· ========== ==========
                method             
              ---------- ----------
               spearman   731±20ms 
               kendall    458±30ms 
               pearson    237±7ms  
              ========== ==========

[ 50.00%] · For pandas commit hash 1a12c41d:
[ 50.00%] ·· Building for conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt...
[ 50.00%] ·· Benchmarking conda-py3.6-Cython-matplotlib-numexpr-numpy-openpyxl-pytables-pytest-scipy-sqlalchemy-xlrd-xlsxwriter-xlwt
[ 50.00%] ··· Running (stat_ops.Correlation.time_corr--)....
[ 62.50%] ··· stat_ops.Correlation.time_corr                                                                                                                         ok
[ 62.50%] ··· ========== =============
                method                
              ---------- -------------
               spearman     85.9±5ms  
               kendall      200±4ms   
               pearson    2.03±0.07ms 
              ========== =============

[ 75.00%] ··· stat_ops.Correlation.time_corrwith                                                                                                                     ok
[ 75.00%] ··· ========== ============
                method               
              ---------- ------------
               spearman   19.8±0.5ms 
               kendall    19.4±0.3ms 
               pearson    19.6±0.5ms 
              ========== ============

[ 87.50%] ··· stat_ops.Correlation.time_corrwith_cols                                                                                                            failed
[ 87.50%] ··· ========== ========
                method           
              ---------- --------
               spearman   failed 
               kendall    failed 
               pearson    failed 
              ========== ========

[100.00%] ··· stat_ops.Correlation.time_corrwith_rows                                                                                                            failed
[100.00%] ··· ========== ========
                method           
              ---------- --------
               spearman   failed 
               kendall    failed 
               pearson    failed 
              ========== ========

     [1a12c41d]       [e19b3f62]
+      19.8±0.5ms         39.6±5ms     2.00  stat_ops.Correlation.time_corrwith('spearman')
+      19.6±0.5ms         35.9±3ms     1.83  stat_ops.Correlation.time_corrwith('pearson')
+      19.4±0.3ms         35.0±2ms     1.80  stat_ops.Correlation.time_corrwith('kendall')

```

jreback:
numpy is using a fully vectorizer soln

your impl is basically linear in the number of columns and thus doesn’t scale

dsaxton:
Got it, I will work on a numpy replacement for map

jreback:
nothing to do with map

the issue is a column by column impl is simply not scalable, that’s why it was written like this to begin with

dsaxton:
@jreback In the latest commit I'm retaining the current implementation of the Pearson calculation so there should be no difference in performance for this case, and only using the new approach for other correlation types. (While perhaps not perfectly scalable yet, I figure it will be a reasonable first effort.)  Also now that #22684 has been merged a nice side effect of this PR would be that `corrwith` can accept callable methods as well.

jreback:
can you merge master and see if you can get this to pass

dsaxton:
@TomAugspurger I added a commit with a merge (should I have rebased?  It seems many unrelated changes were added to the commit), a few fixes (the test failure was apparently caused by accidentally taking a transpose twice, and there was some error-catching code that wasn't accounting for callable methods) and an xfail added to the duplicate columns test

jreback:
you need to merge master. changes should only be your changes, not including other ones.

pandas/tests/frame/test_analytics.py
```diff
@@ -466,6 +466,42 @@ def test_corrwith_mixed_dtypes(self):
...
+
+    @td.skip_if_no_scipy
+    def test_corrwith_kendall(self):
+        # GH 21925
+        df = pd.DataFrame(np.random.random(size=(100, 3)))
+        result = df.corrwith(df**2, method="kendall")
+        expected = Series(np.ones(len(result)))
+        tm.assert_series_equal(result, expected)
+
+    def test_corrwith_invalid_method(self):
```
dsaxton: The relevant code here was actually removed because it didn't consider callable methods.  Should this test then be taken out as well?

jreback:
can you merge master

jreback:
can you merge master and update to comments

doc/source/whatsnew/v0.24.0.rst
```diff
@@ -370,8 +370,10 @@ Other Enhancements
 - :meth:`Index.difference` now has an optional ``sort`` parameter to specify whether the results should be sorted if possible (:issue:`17839`)
 - :meth:`read_excel()` now accepts ``usecols`` as a list of column names or callable (:issue:`18273`)
 - :meth:`MultiIndex.to_flat_index` has been added to flatten multiple levels into a single-level :class:`Index` object.
-- :meth:`DataFrame.to_stata` and :class:`pandas.io.stata.StataWriter117` can write mixed sting columns to Stata strl format (:issue:`23633`)
-- :meth:`DataFrame.between_time` and :meth:`DataFrame.at_time` have gained the an ``axis`` parameter (:issue:`8839`)
+- :meth:`DataFrame.to_stata` and :class:` pandas.io.stata.StataWriter117` can write mixed sting columns to Stata strl format (:issue:`23633`)
+- :ref:`Series.resample` and :ref:`DataFrame.resample` have gained the :meth:`Resampler.quantile` (:issue:`15023`).
```
jreback: are these existing lines changed for some reason?

doc/source/whatsnew/v0.24.0.rst
```diff
@@ -1260,7 +1262,8 @@ Performance Improvements
 Documentation Changes
 ~~~~~~~~~~~~~~~~~~~~~
 
--
+- Added sphinx spelling extension, updated documentation on how to use the spell check (:issue:`21079`)
```
jreback: why was this changed? I think these are just old / rebase issues.

pandas/core/frame.py
```diff
@@ -7021,53 +7026,77 @@ def cov(self, min_periods=None):
...
+
+            num = (ldem * rdem).sum()
+            dom = (left.count() - 1) * left.std() * right.std()
 
-        num = (ldem * rdem).sum()
-        dom = (left.count() - 1) * left.std() * right.std()
+            correl = num / dom
+
+        else:
+            def c(x):
```
jreback: I think its reasonable to simply do
```
def c(x):
    a, b = x
    return nanops.nancorr(x, y, method=method)
```
here as these are already aligned, and already arrays and don't need the extra checking

pandas/core/frame.py
```diff
@@ -7021,53 +7026,77 @@ def cov(self, min_periods=None):
...
+            rdem = right - right.mean()
+
+            num = (ldem * rdem).sum()
+            dom = (left.count() - 1) * left.std() * right.std()
 
-        num = (ldem * rdem).sum()
-        dom = (left.count() - 1) * left.std() * right.std()
+            correl = num / dom
+
+        else:
```
jreback: do elif method in ['.....], and add an else clause that raises (wrong method is passed)

pandas/tests/frame/test_analytics.py
```diff
@@ -466,6 +466,33 @@ def test_corrwith_mixed_dtypes(self):
         expected = pd.Series(data=corrs, index=['a', 'b'])
         tm.assert_series_equal(result, expected)
 
+    @pytest.mark.xfail
+    def test_corrwith_dup_cols(self):
+        # GH 21925
+        df1 = pd.DataFrame(np.vstack([np.arange(10)] * 3).T)
```
jreback: can you add an example with an empty frame.

jreback (changes requested):

doc/source/whatsnew/v0.24.0.rst
TomAugspurger: We definitely don't want :ref: here. Maybe just a bad merge conflict that needs to be reverted.

pandas/tests/frame/test_analytics.py
```diff
@@ -466,6 +466,33 @@ def test_corrwith_mixed_dtypes(self):
         expected = pd.Series(data=corrs, index=['a', 'b'])
         tm.assert_series_equal(result, expected)
 
+    @pytest.mark.xfail
```
TomAugspurger: Perhaps I'm missing something, but why is this xfailed?

doc/source/whatsnew/v0.24.0.rst
dsaxton: This was a bad merge that I'm going to revert

pandas/tests/frame/test_analytics.py
dsaxton: I had xfailed this because master was giving an error when columns were duplicated (`ValueError: cannot reindex from a duplicate axis`)

TomAugspurger: Could you rewrite the test to not have duplicate labels?

But I notice now that on master we do allow duplicates in `DataFrame.corrwith`

```python
In [6]: df = pd.DataFrame([[1, 2], [3, 4]], columns=[0, 0], index=[1, 1])

In [7]: df.corrwith(df)
Out[7]:
0    1.0
0    1.0
dtype: float64
```

so we may need to adjust the implementation to not regress on that. It'd be good to add some tests for that if they aren't already present.

doc/source/whatsnew/v0.24.0.rst
dsaxton: Looks like the revert didn't work as intended so I'm going to need to fix that as well

pandas/tests/frame/test_analytics.py
dsaxton: Hmm, I wonder why that code sample works but this one gives an error:

```
import numpy as np
import pandas as pd

df1 = pd.DataFrame(np.random.random(size=(10, 2)), columns=["a", "b"])
df2 = pd.DataFrame(np.random.random(size=(10, 2)), columns=["a", "a"])

df1.corrwith(df2)
```

doc/source/whatsnew/v0.24.0.rst
```diff
@@ -1260,7 +1261,7 @@ Performance Improvements
 Documentation Changes
 ~~~~~~~~~~~~~~~~~~~~~
 
--
+- Improved clarity of `corrwith` docstring, added See Also entries for `corrwith` and `corr` (:issue:`22328`)
```
jreback: this is not needed

pandas/core/frame.py
```diff
@@ -7021,53 +7026,79 @@ def cov(self, min_periods=None):
...
+        elif method in ['kendall', 'spearman'] or callable(method):
+            def c(x):
+                return nanops.nancorr(x[0], x[1], method=method)
+
+            correl = Series(map(c,
+                                zip(left.values.T, right.values.T)),
+                            index=left.columns)
+
+        else:
+            raise ValueError('Invalid method')
```
jreback: make this ``'Invalid method {} was passed, valid methods are: ......'``

jreback (changes requested):

pandas/core/frame.py
jreback: I *think* that is already done in the Series method, but certainly check

pandas/tests/frame/test_analytics.py
```diff
@@ -466,6 +466,33 @@ def test_corrwith_mixed_dtypes(self):
...
+    def test_corrwith_dup_cols(self):
+        # GH 21925
+        df1 = pd.DataFrame(np.vstack([np.arange(10)] * 3).T)
+        df2 = df1.copy()
+        df2 = pd.concat((df2, df2[0]), axis=1)
+
+        result = df1.corrwith(df2).values
+        expected = np.ones(4)
+        tm.assert_almost_equal(result, expected)
```
jreback: can you use ``assert_series_equal`` to make sure the indexes are correct

jreback (changes requested):

pandas/core/frame.py
dsaxton: Yes, seems you're correct (although the error message doesn't mention that callable methods are allowed; is this worth updating?)

pandas/tests/frame/test_analytics.py
dsaxton: @TomAugspurger By duplicate labels do you mean the values within the `DataFrame` themselves (i.e., not the indices or column names)?

pandas/tests/frame/test_analytics.py
dsaxton: A test to ensure that the output is an empty `Series` (with the proper index)?

jreback:
can you merge master, also update to handle duplicates

pandas/core/frame.py
```diff
@@ -7102,7 +7102,9 @@ def c(x):
             raxis = 1 if axis == 0 else 0
             result_index = (this._get_axis(raxis).
                             union(other._get_axis(raxis)))
-            correl = correl.reindex(result_index)
```
dsaxton: Since the issue with duplicates occurs when we try to use the `reindex` method, my thought was to simply add in the missing values directly by appending to the existing `Series`

jreback (changes requested):
minor comment. ping on green.

jreback (approved):

jreback:
lgtm. ping on green.

dsaxton:
@jreback Looks like it's passing.  Thanks for your help and patience!

jreback:
thanks @dsaxton

PR #22378, opened 2018-08-16
Fix and test scalar extension dtype op corner case

jbrockmendel:
Fixes the following behavior in master:

```
ser = pd.Series(['a', 'b', 'c'])

>>> ser + "category"
  File "pandas/core/ops.py", line 1233, in wrapper
    return dispatch_to_extension_op(op, left, right)
  File "pandas/core/ops.py", line 1163, in dispatch_to_extension_op
    res_values = op(new_left, new_right)
TypeError: can only concatenate list (not "str") to list


>>> ser + "Int64"
  File "pandas/core/ops.py", line 1233, in wrapper
    return dispatch_to_extension_op(op, left, right)
  File "pandas/core/ops.py", line 1163, in dispatch_to_extension_op
    res_values = op(new_left, new_right)
TypeError: can only concatenate list (not "str") to list
```

and the same for the reversed ops.

### DIFF
--- a/pandas/core/ops.py
+++ b/pandas/core/ops.py
@@ -1229,7 +1229,7 @@ def _arith_method_SERIES(cls, op, special):
 
         elif (is_extension_array_dtype(left) or
                 (is_extension_array_dtype(right) and not is_scalar(right))):
-            # disallow scalar to exclude e.g. "category", "Int64"
+            # GH#22378 disallow scalar to exclude e.g. "category", "Int64"
             return dispatch_to_extension_op(op, left, right)
 
         elif is_datetime64_dtype(left) or is_datetime64tz_dtype(left):
--- a/pandas/core/ops.py
+++ b/pandas/core/ops.py
@@ -1228,8 +1228,8 @@ def _arith_method_SERIES(cls, op, special):
                             "{op}".format(typ=type(left).__name__, op=str_rep))
 
         elif (is_extension_array_dtype(left) or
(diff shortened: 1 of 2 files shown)


pandas/core/ops.py
```diff
@@ -1228,8 +1228,8 @@ def wrapper(left, right):
                             "{op}".format(typ=type(left).__name__, op=str_rep))
 
         elif (is_extension_array_dtype(left) or
-                is_extension_array_dtype(right)):
-            # TODO: should this include `not is_scalar(right)`?
+                (is_extension_array_dtype(right) and not is_scalar(right))):
```
jorisvandenbossche: Is checking that it is a string a bit more logical?

jorisvandenbossche (commented):
Good catch :)

pandas/core/ops.py
jbrockmendel: I guess that would work.  Elsewhere the same check is done with `not is_scalar`

jorisvandenbossche: then ignore my comment and go for consistency

jbrockmendel:
@jorisvandenbossche think this is mergeable?  I'm looking through the queue for stuff sufficiently small/easy to take off @jreback's plate.

jorisvandenbossche (approved):

jorisvandenbossche:
Yes, merged, thanks!

Issue #22386, opened 2018-08-16
Support ExtensionArrarys for shift


TomAugspurger:
This is needed for sparse. I have a separate PR incoming

```python
In [2]: import pandas as pd

In [3]: pd.Series([1, 2], dtype='Int8').shift()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-3-8c2f15affd64> in <module>()
----> 1 pd.Series([1, 2], dtype='Int8').shift()

~/sandbox/pandas/pandas/core/series.py in shift(self, periods, freq, axis)
   3476     @Appender(generic._shared_docs['shift'] % _shared_doc_kwargs)
   3477     def shift(self, periods=1, freq=None, axis=0):
-> 3478         return super(Series, self).shift(periods=periods, freq=freq, axis=axis)
   3479
   3480     def reindex_axis(self, labels, axis=0, **kwargs):

~/sandbox/pandas/pandas/core/generic.py in shift(self, periods, freq, axis)
   8088         block_axis = self._get_block_manager_axis(axis)
   8089         if freq is None:
-> 8090             new_data = self._data.shift(periods=periods, axis=block_axis)
   8091         else:
   8092             return self.tshift(periods, freq)

~/sandbox/pandas/pandas/core/internals/managers.py in shift(self, **kwargs)
    545
    546     def shift(self, **kwargs):
--> 547         return self.apply('shift', **kwargs)
    548
    549     def fillna(self, **kwargs):

~/sandbox/pandas/pandas/core/internals/managers.py in apply(self, f, axes, filter, do_integrity_check, consolidate, **kwargs)
    421
    422             kwargs['mgr'] = self
--> 423             applied = getattr(b, f)(**kwargs)
    424             result_blocks = _extend_blocks(applied, result_blocks)
    425

~/sandbox/pandas/pandas/core/internals/blocks.py in shift(self, periods, axis, mgr)
   1289
   1290         # make sure array sent to np.roll is c_contiguous
-> 1291         f_ordered = new_values.flags.f_contiguous
   1292         if f_ordered:
   1293             new_values = new_values.T

AttributeError: 'IntegerArray' object has no attribute 'flags'

```

PR #22387, opened 2018-08-16
Support NDFrame.shift with EAs

TomAugspurger:
Uses take internally.

Closes https://github.com/pandas-dev/pandas/issues/22386

One question: do we want to add `shift` to the interface, and dispatch to that? This just does the right thing at the block level, by writing a `shift` that'll work with an ExtensionArray.

CategoricalBlock still overrides `.shift` to call `Categorical.shift`, which uses a slightly faster implementation based around the codes and `np.roll` (~1.5-2x faster based on microbenchmarks)`.

We could add move `ExtensionBlock.shift `to `ExtensionArray.shift`, and just call that from ExtensionBlock. Then we can delete CategoricalBlock.shift.

I'm 50/50 on whether it's worth adding to the interface.


jreback:
i think it’s worth adding to the interface itself

jorisvandenbossche:
Wouldn't it be possible to do this without take, but with slicing?

Something like (very crude implementation for positive shifts):

```
def shift(a, p):
    a_sliced = a[:-p]
    a_NA = a._from_sequence([a.dtype.na_value]*p, dtype=a.dtype)
    return a._concat_same_type([a_NA, a_sliced])
```

Speed wise I would expect this to be very similar to the current `Categorical.shift` (which only uses the `np.roll` to basically do this instead of manually concatenating both parts)

jorisvandenbossche:
> Speed wise I would expect this to be very similar to the current Categorical.shift

Categorical was apparently a very bad example as there it is like 5x slower than the existing `Categorical.shift`, but that is more because generating the missing values (`_from_sequence)` and `_concat_same_type` are very slow for Categorical (and could certainly be optimized), for IntegerArray it was as fast as I expected.

TomAugspurger:
That seems preferable to me, as it avoids allocating the `indexer` array.

pandas/core/arrays/base.py
```diff
@@ -400,6 +400,36 @@ def dropna(self):
...
+        periods : int, default 1
+            The number of periods to shift. Negative values are allowed
+            for shifting backwards.
+
+        Returns
+        -------
+        shifted : ExtensionArray
+        """
+        if periods == 0:
+            return self.copy()
```
TomAugspurger: Thoughts on this? non-zero periods will necessarily be a copy.

TomAugspurger: Although `Series.shift(0)` doesn't make a copy, so let's follow that?

jorisvandenbossche: I personally find that strange behaviour to not copy then (it would be more logical to know that the result is always a copy, regardless the value of periods IMO)

TomAugspurger: Opened https://github.com/pandas-dev/pandas/issues/22397 for that. Will leave this as is then.

TomAugspurger:
Just as a note, I couldn't quite use this `shift` implementation for `SparseArray`.

When `ExtensionArray.shift`creates the empty SparseArray, it ends up with NaN  for the fill value. When `shift`ing a SparseArray with, say, 0 as the fill value, you end up concatenating the empty (with fill of NaN) and the shifted (with a fill of 0) and that ends up with a fill of NaN.

The root issue is that SparseArray stores fill_value on the array, rather than on the dtype. I haven't looked into whether it *should* go on the dtype or not, it's not clear to me.

jorisvandenbossche:
From that it seems to makes sense to have this on the dtype then (but not that familiar with the rest of the code)

TomAugspurger:
Indeed... On the other hand, it complicates things like equality a bit...

I'll push it onto the dtype and see what breaks.

TomAugspurger:
Any objections to merging?

pandas/core/arrays/base.py
```diff
@@ -400,6 +400,36 @@ def dropna(self):
 
         return self[~self.isna()]
 
+    def shift(self, periods=1):
```
jreback: I think update in the ExtensionArray doc-string?

jreback (changes requested):
shouldn't we be able to change the impl of interval / categorical shift for this?

jorisvandenbossche:
> shouldn't we be able to change the impl of interval / categorical shift for this?

As I mentioned above (https://github.com/pandas-dev/pandas/pull/22387#issuecomment-413658486), if we implement Categorical.shift as is done here (eg by simply inheriting this base method), it gets much slower because the creating of the NaNs and concatting same typed-categoricals is very slow (at least slower as it should be). 
`Interval` did not yet have any shift implementation.

pandas/core/arrays/base.py
jorisvandenbossche: Currently, in the class docstring we only mention the methods that either needs to implemented (because they raise AbstractMethodError otherwise) or either have a suboptimal implementation because it does the object ndarray roundtrip. 
This is not the case here (which is not saying we couldn't also list other methods that can be overriden for specific reasons)

pandas/tests/extension/base/methods.py
```diff
@@ -138,3 +138,28 @@ def test_combine_add(self, data_repeated):
...
             orig_data1._from_sequence([a + val for a in list(orig_data1)]))
         self.assert_series_equal(result, expected)
+
+    @pytest.mark.parametrize('frame', [True, False])
+    @pytest.mark.parametrize('periods, indices', [
+        (-2, [2, 3, 4, -1, -1]),
+        (0, [0, 1, 2, 3, 4]),
+        (2, [-1, -1, 0, 1, 2]),
+    ])
+    def test_container_shift_negative(self, data, frame, periods, indices):
```
jorisvandenbossche: Very minor comment, but is there a reason this is called `test_container_shift_negative`? (because it does not only test negative periods, and 'container' seems a bit superfluous). I would simply do `test_shift`

jorisvandenbossche (approved):

TomAugspurger:
It seems that `IntervalArray` may need some updates,

```python
In [22]: idx = IntervalArray.from_breaks(pd.date_range('2017', periods=10))

In [23]: idx.shift()
<ipython-input-23-1b2c2192e1e6> in <module>()
----> 1 idx.shift()

~/sandbox/pandas/pandas/core/arrays/base.py in shift(self, periods)
    422             return self.copy()
    423         empty = self._from_sequence([self.dtype.na_value] * abs(periods),
--> 424                                     dtype=self.dtype)
    425         if periods > 0:
    426             a = empty

~/sandbox/pandas/pandas/core/arrays/interval.py in _from_sequence(cls, scalars, dtype, copy)
    193     @classmethod
    194     def _from_sequence(cls, scalars, dtype=None, copy=False):
--> 195         return cls(scalars, dtype=dtype, copy=copy)
    196
    197     @classmethod

~/sandbox/pandas/pandas/core/arrays/interval.py in __new__(cls, data, closed, dtype, copy, fastpath, verify_integrity)
    138
    139         return cls._simple_new(left, right, closed, copy=copy, dtype=dtype,
--> 140                                verify_integrity=verify_integrity)
    141
    142     @classmethod

~/sandbox/pandas/pandas/core/arrays/interval.py in _simple_new(cls, left, right, closed, copy, dtype, verify_integrity)
    156                 raise TypeError(msg.format(dtype=dtype))
    157             elif dtype.subtype is not None:
--> 158                 left = left.astype(dtype.subtype)
    159                 right = right.astype(dtype.subtype)
    160

~/sandbox/pandas/pandas/core/indexes/numeric.py in astype(self, dtype, copy)
    313             msg = ('Cannot convert Float64Index to dtype {dtype}; integer '
    314                    'values are required for conversion').format(dtype=dtype)
--> 315             raise TypeError(msg)
    316         elif is_integer_dtype(dtype) and self.hasnans:
    317             # GH 13149

TypeError: Cannot convert Float64Index to dtype datetime64[ns]; integer values are required for conversion

In [24]: idx = IntervalArray.from_breaks(range(10))

In [25]: idx.shift()
<ipython-input-25-1b2c2192e1e6> in <module>()
----> 1 idx.shift()

~/sandbox/pandas/pandas/core/arrays/base.py in shift(self, periods)
    422             return self.copy()
    423         empty = self._from_sequence([self.dtype.na_value] * abs(periods),
--> 424                                     dtype=self.dtype)
    425         if periods > 0:
    426             a = empty

~/sandbox/pandas/pandas/core/arrays/interval.py in _from_sequence(cls, scalars, dtype, copy)
    193     @classmethod
    194     def _from_sequence(cls, scalars, dtype=None, copy=False):
--> 195         return cls(scalars, dtype=dtype, copy=copy)
    196
    197     @classmethod

~/sandbox/pandas/pandas/core/arrays/interval.py in __new__(cls, data, closed, dtype, copy, fastpath, verify_integrity)
    138
    139         return cls._simple_new(left, right, closed, copy=copy, dtype=dtype,
--> 140                                verify_integrity=verify_integrity)
    141
    142     @classmethod

~/sandbox/pandas/pandas/core/arrays/interval.py in _simple_new(cls, left, right, closed, copy, dtype, verify_integrity)
    156                 raise TypeError(msg.format(dtype=dtype))
    157             elif dtype.subtype is not None:
--> 158                 left = left.astype(dtype.subtype)
    159                 right = right.astype(dtype.subtype)
    160

~/sandbox/pandas/pandas/core/indexes/numeric.py in astype(self, dtype, copy)
    316         elif is_integer_dtype(dtype) and self.hasnans:
    317             # GH 13149
--> 318             raise ValueError('Cannot convert NA to integer')
    319         return super(Float64Index, self).astype(dtype, copy=copy)
    320

ValueError: Cannot convert NA to integer
```

The basic issue is 

```
    423         empty = self._from_sequence([self.dtype.na_value] * abs(periods),
--> 424                                     dtype=self.dtype)
 ```

The container (ndarray[self.dtype]) can't actually hold `self.dtype.na_value` (nan).

Does that change our opinion on this implementation?

**Comment by jorisvandenbossche** [comment]
I think that is rather a bug in IntervalArray? (or maybe better a current limitation) 

Anyway, if the point is that it (currently) cannot hold NaNs, it can also not implement `shift`, as it requires NaNs (so don't know if that changes anything about the shift implementation?)

**Comment by TomAugspurger** [comment]
Well, it can implement `.shift`, as long as it's allowed to change dtypes (upcast from int to float). The current implementation assumes / requires that the dtype not change.

**Comment by jorisvandenbossche** [comment]
Basically, the problem is that for IntervalArray, the data is stored in `left` and `right` arrays, and so it depends on the dtype of those arrays whether IntervalArray can have NaNs or not? 
So eg float and datetime64 can hold missing values, ints not.

For the datetime error above, that is actually solvable with the current model, because if the data is not all-NaN it works:

```
In [75]: IntervalArray([pd.Interval(pd.Timestamp('2017-01-01'), pd.Timestamp('2017-01-02'))], dtype='interval[datetime64[ns]]')
Out[75]: 
IntervalArray([(2017-01-01, 2017-01-02]],
              closed='right',
              dtype='interval[datetime64[ns]]')

In [76]: IntervalArray([pd.Interval(pd.Timestamp('2017-01-01'), pd.Timestamp('2017-01-02')), np.nan], dtype='interval[datetime64[ns]]')
Out[76]: 
IntervalArray([(2017-01-01, 2017-01-02], nan],
              closed='right',
              dtype='interval[datetime64[ns]]')

In [78]: IntervalArray([np.nan], dtype='interval[datetime64[ns]]').left
...
TypeError: Cannot convert Float64Index to dtype datetime64[ns]; integer values are required for conversion
```

**Comment by jorisvandenbossche** [comment]
> Well, it can implement .shift, as long as it's allowed to change dtypes

Ah, yes, the same happens with `Series([int]).shift()` of course.

I think ideally, we should improve the missing values handling on the IntervalArray level (since it's a composite array, we have the ability to do it properly, not requiring such upcasts), but that's certainly out of scope for this PR :-)

**Comment by jorisvandenbossche** [comment]
So using the `take` implementation, that does already do the upcasting for Interval, so the `shift` would then work for intervals. 

I don't really care much to whether to switch back to that or leave the implementation now as is (with the consequence shift doesn't work for Interval; although given the `take` precedent, upcasting the dtype seems somewhat consistent. 
But I do think we should discuss the missing value handling for IntervalArray independently from this (eg using a mask instead).

**Comment by TomAugspurger** [comment]
> I think ideally, we should improve the missing values handling on the IntervalArray level

That's my preference. Opened https://github.com/pandas-dev/pandas/issues/22428 for that.

**Review by jreback — CHANGES_REQUESTED** [review]
lgtm. minor doc-comments.

File: pandas/core/internals/blocks.py [hunk 4]
```diff
@@ -2068,6 +2068,18 @@ def interpolate(self, method='pad', axis=0, inplace=False, limit=None,
                                  limit=limit),
             placement=self.mgr_locs)
 
+    def shift(self, periods, axis=0, mgr=None):
+        """
+        Shift the block by `periods`.
+
+        Dispatches to underlying ExtensionArray and re-boxes in an
+        ExtensionBlock.
+        """
```
jreback: bonus points for the Parameters :> (future PR ok too)

jreback (approved):
lgtm. minor comments. merge away

jreback:
thanks @TomAugspurger

Issue #20088, opened 2018-03-09
DataFrame[timedelta64] / timedelta64 or pydatetime has wrong dtype and wrong values


nmusolino:
#### Code Sample

```
In [1]: import pandas, numpy, datetime

In [2]: df = pandas.DataFrame({'x': numpy.timedelta64(1, 'ms') * numpy.arange(0, 5)})

In [3]: df
Out[3]:
                x
0        00:00:00
1 00:00:00.001000
2 00:00:00.002000
3 00:00:00.003000
4 00:00:00.004000

In [4]: df.dtypes
Out[4]:
x    timedelta64[ns]
dtype: object

In [5]: df['x'] / datetime.timedelta(milliseconds=1)
Out[5]:
0    0.0
1    1.0
2    2.0
3    3.0
4    4.0
Name: x, dtype: float64

In [6]: df[['x']] / datetime.timedelta(milliseconds=1)
Out[6]:
                x
0        00:00:00
1 00:00:00.000000
2 00:00:00.000000
3 00:00:00.000000
4 00:00:00.000000

In [8]: df[['x']] / numpy.timedelta64(1, 'ms')
Out[8]:
                x
0        00:00:00
1 00:00:00.000000
2 00:00:00.000000
3 00:00:00.000000
4 00:00:00.000000
```

#### Problem description

When performing true division on a dataframe containing timedelta64 values, and dividing by a datetime.timedelta object or a timedelta64, there are two problems:
1.  the resulting values are incorrect; and 
2.  the dtype (`timedelta64[ns]`) in the resulting dataframe is not consistent with the results of the same operation on the pandas `Series` or the numpy array.  (In those cases, the result is a float series or array.)

#### Expected Output
The dataframe should contain a `float64` column, with values equal to `df['x'] / numpy.timedelta64(1, 'ms')`:
```
In [11]: df.apply(lambda s: s.values / numpy.timedelta64(1, 'ms'))
Out[11]:
     x
0  0.0
1  1.0
2  2.0
3  3.0
4  4.0
```

#### Output of ``pd.show_versions()``

<details>

[paste the output of ``pd.show_versions()`` here below this line]
INSTALLED VERSIONS
------------------
commit: None
python: 3.4.5.final.0
python-bits: 64
OS: Windows
OS-release: 7
machine: AMD64
processor: Intel64 Family 6 Model 79 Stepping 1, GenuineIntel
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
[...]

PR #22390, opened 2018-08-16
Fix arithmetic errors with timedelta64 dtypes

jbrockmendel:
Kind of surprising, but I don't see any Issues for these bugs.  Might be because many of them were found after parametrizing arithmetic tests, so instead of Issues we had xfails.

xref #20088 
xref #22163

- [x] tests added / passed
- [x] whatsnew entry


doc/source/whatsnew/v0.24.0.txt
```diff
@@ -576,7 +576,10 @@ Datetimelike
 - Bug in :class:`DataFrame` with mixed dtypes including ``datetime64[ns]`` incorrectly raising ``TypeError`` on equality comparisons (:issue:`13128`,:issue:`22163`)
 - Bug in :meth:`DataFrame.eq` comparison against ``NaT`` incorrectly returning ``True`` or ``NaN`` (:issue:`15697`,:issue:`22163`)
 - Bug in :class:`DataFrame` with ``timedelta64[ns]`` dtype division by ``Timedelta``-like scalar incorrectly returning ``timedelta64[ns]`` dtype instead of ``float64`` dtype (:issue:`20088`,:issue:`22163`)
```
jreback: can you split out to a timedelta bug section (as this is getting kind of long)?

can be in a future PR

pandas/core/indexes/range.py
```diff
@@ -596,6 +597,9 @@ def _evaluate_numeric_binop(self, other):
                     # GH#19333 is_integer evaluated True on timedelta64,
                     # so we need to catch these explicitly
                     return op(self._int64index, other)
+                elif is_timedelta64_dtype(other):
```
jreback: doesn't the fall thru raise TypeError?

pandas/core/ops.py
```diff
@@ -1244,6 +1244,32 @@ def wrapper(left, right):
                                     index=left.index, name=res_name,
                                     dtype=result.dtype)
 
+        elif is_timedelta64_dtype(right) and not is_scalar(right):
+            # i.e. exclude np.timedelta64 object
+            # GH#22390 Unfortunately we need to special-case right-hand
+            # timedelta64 dtypes because numpy casts integer dtypes to
+            # timedelta64 when operating with timedelta64
+            if isinstance(right, np.ndarray):
+                # upcast to TimedeltaIndex before dispatching
```
jreback: uhh this seems like a lot of special casing, any way to make this simpler

jreback (changes requested):

doc/source/whatsnew/v0.24.0.txt
jbrockmendel: Sounds good

pandas/core/ops.py
jbrockmendel: The only simplification I considered but didn't use was something like:

```
def maybe_upcast_for_op(obj):
     if type(obj) is timedelta:
        return pd.Timedelta(obj)
    if isinstance(obj, np.ndarray) and is_timedelta64_dtype(obj):
        return pd.TimedeltaIndex(obj)
    return obj
```

which would go right after `get_op_result_name`.

jbrockmendel:
This will probably need rebasing after #22350 goes through

pandas/core/indexes/range.py
jbrockmendel: I don’t understand the question

pandas/core/ops.py
```diff
@@ -1244,6 +1244,32 @@ def wrapper(left, right):
...
+
+            # Note: we cannot use dispatch_to_index_op because
+            # that may incorrectly raise TypeError when we
+            # should get NullFrequencyError
+            result = op(pd.Index(left), right)
+            return construct_result(left, result,
+                                    index=left.index, name=res_name,
+                                    dtype=result.dtype)
+
+        elif type(right) is datetime.timedelta:
```
jreback: this case you can include above (before what you added), no?

```
if isintance(right, (datetime.timedelta, Timedelta)):
  right = Timedelta(right)
....
```

pandas/core/ops.py
jreback: maybe re-route the scalars to a separate path. This is getting way special casey here.

pandas/core/indexes/range.py
jreback: your comment is at odds with the checking i think. pls revise comments and/or checks.

jreback (changes requested):

pandas/core/indexes/range.py
jbrockmendel: The comment below this line `# must be an np.ndarray; GH#22390` along with the check above `elif is_timedelta64_dtype(other)` is just saying this is an `ndarray["timedelta64['ns']"]`.  I don't think these are at odds.

pandas/core/ops.py
jbrockmendel: I'll implement the maybe_upcast idea and see if that is prettier

pandas/core/ops.py
jbrockmendel: The version code above casts to `TimedeltaIndex`, _not_ `Timedelta`.

doc/source/whatsnew/v0.24.0.txt
jreback: I *think* we now have a separate section?

jreback (approved):
minor doc-comment. rebase and ping on green.

jbrockmendel:
ping

jreback:
thanks! nice cleaning

PR #22394, opened 2018-08-17
Bug in DataFrame.drop_duplicates for empty DataFrame throws error

### LINKED ISSUE #20516: Calling drop_duplicates method for empty pandas dataframe throws error
#### Code Sample

```python
>>> pd.DataFrame().drop_duplicates()
  File "/home/analytical-monk/miniconda3/lib/python3.6/site-packages/pandas/core/frame.py", line 3098, in drop_duplicates
    duplicated = self.duplicated(subset, keep=keep)
  File "/home/analytical-monk/miniconda3/lib/python3.6/site-packages/pandas/core/frame.py", line 3144, in duplicated
    labels, shape = map(list, zip(*map(f, vals)))
ValueError: not enough values to unpack (expected 2, got 0)
```
#### Problem description

Currently, calling the drop_duplicates method for an empty dataframe object (simply pd.DataFrame()) throws an error.
Ideally it should return back the empty dataframe just liked it does when at least one column is present.

#### Expected Output

```
>>> pd.DataFrame().drop_duplicates()

Empty DataFrame
Columns: []
Index: []
```

<details>

```
>>> pd.show_versions()

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.1.final.0
python-bits: 64
OS: Linux
OS-release: 4.8.0-58-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_IN
LOCALE: en_IN.ISO8859-1

pandas: 0.20.3
pytest: None
pip: 9.0.1
setuptools: 36.6.0
Cython: None
numpy: 1.13.1
scipy: None
xarray: None
IPython: None
sphinx: None
patsy: None
dateutil: 2.6.1
pytz: 2017.2
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: 4.6.0
html5lib: 1.0b10
sqlalchemy: 1.1.14
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
pandas_gbq: None
pandas_datareader: None
```
</details>

HyunTruth:
- [x] closes #20516
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -711,7 +711,7 @@ Reshaping
 - Bug in :func:`get_dummies` with Unicode attributes in Python 2 (:issue:`22084`)
 - Bug in :meth:`DataFrame.replace` raises ``RecursionError`` when replacing empty lists (:issue:`22083`)
 - Bug in :meth:`Series.replace` and meth:`DataFrame.replace` when dict is used as the `to_replace` value and one key in the dict is is another key's value, the results were inconsistent between using integer key and using string key (:issue:`20656`)
-- Bug in :meth:`DataFrame.drop_duplicates` for empty DataFrame throws error (:issue:`20516`)
+- Bug in :meth:`DataFrame.drop_duplicates` for empty ``DataFrame`` which incorrectly raises error (:issue:`20516`)
 
 Build Changes
 ^^^^^^^^^^^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -711,7 +711,7 @@ Reshaping
 - Bug in :func:`get_dummies` with Unicode attributes in Python 2 (:issue:`22084`)
 - Bug in :meth:`DataFrame.replace` raises ``RecursionError`` when replacing empty lists (:issue:`22083`)
 - Bug in :meth:`Series.replace` and meth:`DataFrame.replace` when dict is used as the `to_replace` value and one key in the dict is is another key's value, the results were inconsistent between using integer key and using string key (:issue:`20656`)
--
+- Bug in :meth:`DataFrame.drop_duplicates`for empty DataFrame throws error (:issue:`20516`)
 
 Build Changes
 ^^^^^^^^^^^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -711,7 +711,7 @@ Reshaping
 - Bug in :func:`get_dummies` with Unicode attributes in Python 2 (:issue:`22084`)
 - Bug in :meth:`DataFrame.replace` raises ``RecursionError`` when replacing empty lists (:issue:`22083`)
 - Bug in :meth:`Series.replace` and meth:`DataFrame.replace` when dict is used as the `to_replace` value and one key in the dict is is another key's value, the results were inconsistent between using integer key and using string key (:issue:`20656`)
-- Bug in :meth:`DataFrame.drop_duplicates`for empty DataFrame throws error (:issue:`20516`)
+- Bug in :meth:`DataFrame.drop_duplicates` for empty DataFrame throws error (:issue:`20516`)
 
 Build Changes
 ^^^^^^^^^^^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -711,7 +711,7 @@ Reshaping
 - Bug in :func:`get_dummies` with Unicode attributes in Python 2 (:issue:`22084`)
 - Bug in :meth:`DataFrame.replace` raises ``RecursionError`` when replacing empty lists (:issue:`22083`)
 - Bug in :meth:`Series.replace` and meth:`DataFrame.replace` when dict is used as the `to_replace` value and one key in the dict is is another key's value, the results were inconsistent between using integer key and using string key (:issue:`20656`)
-- Bug in :meth:`DataFrame.drop_duplicates` for empty ``DataFrame`` which incorrectly raises error (:issue:`20516`)
+- Bug in :meth:`DataFrame.drop_duplicates` for empty ``DataFrame`` which incorrectly raises an error (:issue:`20516`)
 
 Build Changes
 ^^^^^^^^^^^^^
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4336,7 +4336,7 @@ class DataFrame(NDFrame):
         deduplicated : DataFrame
         """
         if self.empty:
-            return self
+            return self.copy()
 
         inplace = validate_bool_kwarg(inplace, 'inplace')
         duplicated = self.duplicated(subset, keep=keep)
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4369,6 +4369,9 @@ class DataFrame(NDFrame):
         from pandas.core.sorting import get_group_index
         from pandas._libs.hashtable import duplicated_int64, _SIZE_HINT_LIMIT
 
+        if self.empty:
+            return Series()
+
         def f(vals):
             labels, shape = algorithms.factorize(
                 vals, size_hint=min(len(self), _SIZE_HINT_LIMIT))
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4335,6 +4335,9 @@ class DataFrame(NDFrame):
         -------
         deduplicated : DataFrame
         """
+        if self.empty:
+            return self
+
         inplace = validate_bool_kwarg(inplace, 'inplace')
         duplicated = self.duplicated(subset, keep=keep)
 
--- a/pandas/tests/frame/test_duplicates.py
+++ b/pandas/tests/frame/test_duplicates.py
@@ -286,7 +286,6 @@ def test_drop_duplicates_empty():
     tm.assert_frame_equal(result, expected)
 
 
-
 def test_drop_duplicates_NA():
     # none
     df = DataFrame({'A': [None, None, 'foo', 'bar',
--- a/pandas/tests/frame/test_duplicates.py
+++ b/pandas/tests/frame/test_duplicates.py
@@ -273,8 +273,9 @@ def test_drop_duplicates_empty():
     result = expected.drop_duplicates()
     tm.assert_frame_equal(result, expected)
 
-    expected = DataFrame(columns=['A', 'B', 'C'])
-    result = expected.drop_duplicates()
+    df = DataFrame(columns=['A', 'B', 'C'])
+    result = df.drop_duplicates()
+    expected = DataFrame(columns=[])  # The column infos are not carrying over
     tm.assert_frame_equal(result, expected)
 
     expected = DataFrame(index=[])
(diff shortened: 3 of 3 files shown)


doc/source/whatsnew/v0.24.0.txt
```diff
@@ -711,7 +711,7 @@ Reshaping
 - Bug in :func:`get_dummies` with Unicode attributes in Python 2 (:issue:`22084`)
 - Bug in :meth:`DataFrame.replace` raises ``RecursionError`` when replacing empty lists (:issue:`22083`)
 - Bug in :meth:`Series.replace` and meth:`DataFrame.replace` when dict is used as the `to_replace` value and one key in the dict is is another key's value, the results were inconsistent between using integer key and using string key (:issue:`20656`)
--
+- Bug in :meth:`DataFrame.drop_duplicates`for empty DataFrame throws error (:issue:`20516`)
```
WillAyd: Does this link render? Curious if a space is required after the end backtick

HyunTruth: Thank you for noticing! Fixed it

datapythonista (approved):

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -711,7 +711,7 @@ Reshaping
 - Bug in :func:`get_dummies` with Unicode attributes in Python 2 (:issue:`22084`)
 - Bug in :meth:`DataFrame.replace` raises ``RecursionError`` when replacing empty lists (:issue:`22083`)
 - Bug in :meth:`Series.replace` and meth:`DataFrame.replace` when dict is used as the `to_replace` value and one key in the dict is is another key's value, the results were inconsistent between using integer key and using string key (:issue:`20656`)
--
+- Bug in :meth:`DataFrame.drop_duplicates` for empty DataFrame throws error (:issue:`20516`)
```
jreback: use double backticks on DataFrame

throws error -> which incorrectly raises

pandas/tests/frame/test_duplicates.py
```diff
@@ -263,6 +263,13 @@ def test_drop_duplicates_tuple():
     tm.assert_frame_equal(result, expected)
 
 
+def test_drop_duplicates_empty():
```
jreback: can you parameterize this on columns with ``[], ['A', 'B','C']`` as the values
IOW test the ``DataFrame(columns=[])`` and ``DataFrame(columns=['A', 'B', 'C'])``

jreback (changes requested):

pandas/tests/frame/test_duplicates.py
jreback: actually also need a test with no columns but an index e.g. ``DataFrame(index=[1, 2])``

so don't need to parameterize but test all of these cases (where the result is the input)

doc/source/whatsnew/v0.24.0.txt
HyunTruth: Okay, will apply the changes right away

HyunTruth:
Test assert failed due to the following issue: #22409. Posted the details regarding this issue. Will try to update on it if I have time.

HyunTruth:
@jreback @WillAyd @datapythonista  Should I just not include the test for columns until the #22409 is closed (as it will constantly tell that the shape is different due to column info is not there) or put this PR on hold until the above issue is closed?

WillAyd:
See note for the referenced issue but I don't think that is related (?). As @jreback mentioned you should parametrize the test case you have created. If so it will help more easily identify which if any case is failing

WillAyd (changes requested):
Change looks OK but the test needs to be parametrized still. Take a look at some of the other functions in the same module if unsure how to do that

pandas/tests/frame/test_duplicates.py
```diff
@@ -263,6 +263,30 @@ def test_drop_duplicates_tuple():
...
+    result = expected.drop_duplicates()
+    tm.assert_frame_equal(result, expected)
+
+    expected = DataFrame(columns=[])
+    result = expected.drop_duplicates()
+    tm.assert_frame_equal(result, expected)
+
+    df = DataFrame(columns=['A', 'B', 'C'])
+    result = df.drop_duplicates()
+    expected = DataFrame(columns=[])  # The column infos are not carrying over
```
jreback: I don't find this comment useful, rather put a comment about testing with empty columns, and below about an empty index

jreback (changes requested):
small comments, ping on green.

pandas/tests/frame/test_duplicates.py
```diff
@@ -263,6 +263,21 @@ def test_drop_duplicates_tuple():
...
+    DataFrame(columns=['A', 'B', 'C']),
+    DataFrame(index=[]),
+    DataFrame(index=['A', 'B', 'C'])
+])
+def test_drop_duplicates_empty(df):
+    # GH 20516
+    result = df.drop_duplicates()
+    if df.columns.empty is False:
+        df = DataFrame(columns=[])
+    tm.assert_frame_equal(result, df)
```
datapythonista: This tests feels a bit tricky to me. So, when there are columns, we set them to an empty list. So, that would be the same as simply using as expected value `DataFrame(columns=[])`, or am I missing something?

But not sure if I'm missing something, but I'd expect `DataFrame(columns=['A', 'B', 'C']).drop_duplicates()` to keep the columns. Any reason to drop them?

HyunTruth: Please check the closed issue of #22409. I've been told that it is an expected behavior.

datapythonista: I don't think the problem may be your approach implementing this story. I agree on what @WillAyd said in #22409, but not in the side effect in this PR.

I guess `df[pandas.Series()]` is the same as `df[[]]`, so you're telling the DataFrame to select an empty list of columns, so none of them. So, if we have a DataFrame with 4 columns and 10 rows, I expect this to return the 10 rows, but 0 columns.

But the case here with `drop_duplicates`, if I understand correctly, is that you have a DataFrame with 4 columns and 0 rows, you remove the duplicates, that there are none, as the DataFrame has 0 rows, so I should still have 4 columns and 0 rows.

And in your test you're asserting that the return will be 0 columns and 0 rows.

Am I right? Does it makes sense?

HyunTruth: Yeah. However, if we want the column to carry over, then we'd need to change the drop_duplicates itself to not use the logic of df[~duplicated], as it currently does, as this is what is causing this problem at least at this moment with the empty return of duplicated... Maybe returning empty series by itself wasn't a good idea. Completely stuck then.

pandas/core/frame.py
```diff
@@ -4335,6 +4335,9 @@ def drop_duplicates(self, subset=None, keep='first', inplace=False):
         -------
         deduplicated : DataFrame
         """
+        if self.empty:
+            return self
```
jreback: return self.copy() here

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -711,7 +711,7 @@ Reshaping
 - Bug in :func:`get_dummies` with Unicode attributes in Python 2 (:issue:`22084`)
 - Bug in :meth:`DataFrame.replace` raises ``RecursionError`` when replacing empty lists (:issue:`22083`)
 - Bug in :meth:`Series.replace` and meth:`DataFrame.replace` when dict is used as the `to_replace` value and one key in the dict is is another key's value, the results were inconsistent between using integer key and using string key (:issue:`20656`)
--
+- Bug in :meth:`DataFrame.drop_duplicates` for empty ``DataFrame`` which incorrectly raises error (:issue:`20516`)
```
jreback: raises an error

jreback (changes requested):
minor comments. ping on green.

pandas/core/frame.py
```diff
@@ -4335,6 +4335,9 @@ def drop_duplicates(self, subset=None, keep='first', inplace=False):
         -------
         deduplicated : DataFrame
         """
+        if self.empty:
+            return self.copy()
```
datapythonista: instead of this, can you try to replace the last line `return self[-duplicated]` by `return self.iloc[-duplicated]`, for the reasons I mentioned earlier?

pandas/tests/frame/test_duplicates.py
```diff
@@ -263,6 +263,19 @@ def test_drop_duplicates_tuple():
...
+    DataFrame(),
+    DataFrame(columns=[]),
+    DataFrame(columns=['A', 'B', 'C']),
+    DataFrame(index=[]),
+    DataFrame(index=['A', 'B', 'C'])
+])
+def test_drop_duplicates_empty(df):
+    # GH 20516
+    result = df.drop_duplicates()
+    tm.assert_frame_equal(result, df)
```
datapythonista: can you add a test case with `inplace=True` please?

datapythonista (changes requested):

pandas/core/frame.py
HyunTruth: Then wouldn't index be lost in the case of `DataFrame(index=['A', 'B', 'C'])? As it will be `self.iloc[[]]`?

pandas/tests/frame/test_duplicates.py
HyunTruth: Sure.

pandas/core/frame.py
HyunTruth: If we use `self.iloc[~duplicated]`, as you said, we will be able to maintain the columns if we have 4 columns and 0 rows. However, if we have 0 columns and 4 rows, we have the same problem - the outcome is `self.iloc[[]]`, which means that the rows won't be selected, for the same reason with using `self[[]]` for columns, thus ending up with 0 columns and 0 rows, once again.

datapythonista: I'm not sure whether it'd make sense for a dataframe with rows but no columns to drop the empty rows as duplicate.

But as @jreback if happy with this implementation, just leave it like this. I didn't see his comment earlier.

HyunTruth: I tried a test, and here's the result:
```
import pandas as pd

a = pd.DataFrame(index=['A', 'B'])
b = a.iloc[[]]
a.shape # (2, 0)
b.shape # (0, 0)
```

WillAyd (approved):
Outside of changes requested by other reviewers this lgtm

pandas/core/frame.py
datapythonista: Yes, that makes sense to me. If we consider that empty rows are equal among them, then `pd.DataFrame(index=['A', 'B']).duplicated()` would return `False, True`, and `.iloc[-duplicated]` would return the first row.

But as I said, I'm happy to keep the original DataFrame as is for this case, as @jreback is happy with it.

HyunTruth: Oh, if you see the empty rows as equals, then it does make sense. I haven't thought of it that way. Thanks.

jreback (approved):

jreback:
lgtm. @datapythonista merge when satisfied.

datapythonista:
Thanks @HyunTruth

HyunTruth:
@jreback @datapythonista @WillAyd Thank you all

PR #22403, opened 2018-08-17
Feat/bubble plot

### LINKED ISSUE #16827: Scatter plot with colour_by and size_by variables

VincentAntoine:
- [x] Closes #16827 partially : makes bubble plots possible by passing a column name as an argument ''s''
- [x] Replaces PR #20572 (updated and improved)
- [x] 2 tests added & passed
- [x] Passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] Whatsnew entry added for 0.24.0


VincentAntoine:
Here are two examples of the bubble plots produced with numeric and categorical data.

**Numeric bubble size data:**

```python
df = pd.DataFrame(np.random.rand(50, 3), columns=['a', 'b', 'c'])
df.plot.scatter(x='a', y='b', s='c')
```
image

The auto generated legend adapts to the maximum value to be represented and always has round values.

---
**Ordered categorical bubble size data:**

```python
surf_area = np.concatenate([40.0 + 80.0 * np.random.rand(30),
                            80.0 + 160.0 * np.random.rand(20),
                            100.0 + 200.0 * np.random.rand(10)])

types = np.array(30*['Flat'] + 20*['House'] + 10*['Castle'])
prices = 0.01 * surf_area * (np.random.rand(60) + 1.5) / 2
prices *= np.array([1]*30 + [1.4]*20 + [2]*10)
property_types = pd.Categorical(types, categories=['Flat', 'House', 'Castle'], ordered=True)
df = pd.DataFrame({
   'Surface area (sqm)': surf_area,
   'Price (M€)': prices,
   'Property type': property_types
})

df.loc[[0, 1, 30, 31, 50, 51], :]
```

image


```python
df.plot.scatter(x='Surface area (sqm)', y='Price (M€)',
                s='Property type', alpha=.5);
```

image

VincentAntoine:
@TomAugspurger Since you followed this from the start your comments would be highly appreciated! Thx Vincent

pandas/plotting/_core.py
```diff
@@ -861,11 +863,42 @@ def _plot_colorbar(self, ax, **kwds):
...
-    def __init__(self, data, x, y, s=None, c=None, **kwargs):
+    def __init__(self, data, x, y, s=None, c=None, size_factor=1, **kwargs):
         if s is None:
-            # hide the matplotlib default for size, in case we want to change
-            # the handling of this argument later
-            s = 20
+            # Set default size if no argument is given.
+            s = 20 * size_factor
+        elif is_hashable(s) and s in data.columns:
+            # Handle the case where s is a label of a column of the df.
```
TomAugspurger: Can you separate the bubble type and size logic from the getting / setting on self. I'd like for testing that the bubble sizes are computed correctly to be easier.

So a standalone function like

```python
def get_bubbles(size_data, bubble_points):
    ...
    return bubble_dtype, s_data_max, s
```

pandas/plotting/_core.py
```diff
@@ -861,11 +863,42 @@ def _plot_colorbar(self, ax, **kwds):
...
+                    s = self.bubble_points * size_factor * \
+                        size_data_codes**2 / size_data_codes.max()**2
+                else:
+                    raise TypeError(
+                        "'s' must be numeric or ordered categorical dtype")
+            elif is_numeric_dtype(size_data):
+                self.bubble_dtype = 'numeric'
+                self.size_title = s
+                self.s_data_max = size_data.max()
+                self.size_factor = size_factor
```
TomAugspurger: Why is size factor only set here? Ideally it would be set regardless.

pandas/plotting/_core.py
```diff
@@ -861,11 +863,42 @@ def _plot_colorbar(self, ax, **kwds):
...
+                s = self.bubble_points * size_factor * size_data / \
```
TomAugspurger: Style: we prefer parenthesis to break long lines.

TomAugspurger (commented):
Just gave a quick look, will need to follow up.

Also needs a release note for 0.24.0

pandas/plotting/_core.py
VincentAntoine: Thanks Tom for the review,
I separated the bubble logic into two functions, one for the bubbles in the plot, one for the bubbles in the legend. For the bubbles in the plot, the sizes are passed to the parent class as an argument `s`. For the legend, I stored the sizes and labels as attributes of the scatter plot which are used by the legend building functions. Let me know if that's what you had in mind.

pandas/plotting/_core.py
VincentAntoine: OK

pandas/plotting/_core.py
VincentAntoine: OK, this was due to the way the bubble logic was strangely split between the ``__init__`` and ``_make_legend`` functions. I reorganized it.

doc/source/visualization.rst
```diff
@@ -581,26 +581,57 @@ each point:
...
+given to specify a multiplication factor to bubble sizes displayed on the graph:
 
 .. ipython:: python
 
-   @savefig scatter_plot_bubble.png
-   df.plot.scatter(x='a', y='b', s=df['c']*200);
+   @savefig scatter_plot_bubble_with_size_factor.png
+   df.plot.scatter(x='a', y='b', s='c', size_factor=0.2);
+
+The keyword ''s'' can also be of ordered categorical data type.
```
TomAugspurger: Backticks instead of quotes.

doc/source/visualization.rst
```diff
@@ -581,26 +581,57 @@ each point:
...
+
+.. ipython:: python
+
+   surf_area = np.concatenate([40.0 + 80.0 * np.random.rand(30),
+                               80.0 + 160.0 * np.random.rand(20),
+                               100.0 + 200.0 * np.random.rand(10)])
+
+   types = np.array(30*['Flat'] + 20*['House'] + 10*['Castle'])
```
TomAugspurger: preferably pep8 here. (spaces around `*`)

doc/source/visualization.rst
```diff
@@ -581,26 +581,57 @@ each point:
...
+
+   prices = 0.01 * surf_area * (np.random.rand(60) + 1.5) / 2
+   prices *= np.array([1]*30 + [1.4]*20 + [2]*10)
+
+   property_types = pd.Categorical(types, categories=['Flat', 'House', 'Castle'], ordered=True)
```
TomAugspurger: Line looks a bit long.

doc/source/visualization.rst
```diff
@@ -581,26 +581,57 @@ each point:
...
+.. ipython:: python
+
+   surf_area = np.concatenate([40.0 + 80.0 * np.random.rand(30),
+                               80.0 + 160.0 * np.random.rand(20),
+                               100.0 + 200.0 * np.random.rand(10)])
+
+   types = np.array(30*['Flat'] + 20*['House'] + 10*['Castle'])
+
+   prices = 0.01 * surf_area * (np.random.rand(60) + 1.5) / 2
+   prices *= np.array([1]*30 + [1.4]*20 + [2]*10)
```
TomAugspurger: pep8

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -12,6 +12,8 @@ v0.24.0 (Month XX, 2018)
 
 New features
 ~~~~~~~~~~~~
+- The ``DataFrame`` method :func:`plot.scatter()` now accepts column names as an argument ``s`` to produce bubble plots in which the data in the corresponding column is represented by bubble sizes. (:issue:`16827`)
```
TomAugspurger: The `:func:` won't work. Needs to be what's listed in our api.rst like 

    :func:`DataFrame.plot.scatter`

TomAugspurger: "bubble plots" -> "scatter plots". Or maybe repharse the entire second half as "a plot where the marker sizes reflect the values in the column."

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -159,6 +161,24 @@ This is the same behavior as ``Series.values`` for categorical data. See
 :ref:`whatsnew_0240.api_breaking.interval_values` for more.
 
 
+.. _whatsnew_0240.enhancements.bubble_plots:
+
+Bubble plots
```
TomAugspurger: Maybe "Automatic Marker Sizes" or something like that. Don't want to confuse people with "bubble" vs "scatter" plot.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -159,6 +161,24 @@ This is the same behavior as ``Series.values`` for categorical data. See
...
+^^^^^^^^^^^^
+
+Bubble plots can now be produced with the :func:`plot.scatter()` method of ``DataFrame`` objects by passing a column name as an argument ``s`` to provide the data for bubble sizes. Bubble sizes automatically adjust to the maximum of the data, and a legend for bubble sizes also is automatically produced. (:issue:`22441`)
```
TomAugspurger: Fix the ref to plot.scatter.

TomAugspurger: Perhaps, "and the legend reflects the marker sizes".

pandas/plotting/_core.py
TomAugspurger: Mainly, I'm interested in ease of testing. I'd like to have a dedicated method that, given an array of values, tells me what the marker size should be for each point.

pandas/plotting/_core.py
```diff
@@ -861,11 +863,20 @@ def _plot_colorbar(self, ax, **kwds):
...
-            s = 20
+            # Set default size if no argument is given.
+            s = 20 * size_factor
+        elif is_hashable(s) and s in data.columns:
+            # Handle the case where s is a label of a column of the df.
+            # The data is normalized to 200 * size_factor.
+            self.size_title = s
+            self.bubble_points = 200
+            self.size_factor = size_factor
+            size_data = data[s]
```
TomAugspurger: What happens `s` is a label with duplicates? This would be a dataframe. We should error gracefully in that case.

pandas/plotting/_core.py
```diff
@@ -861,11 +863,20 @@ def _plot_colorbar(self, ax, **kwds):
...
+            s = self._get_plot_bubbles(size_data)
+            self.bubble_legend_sizes, self.bubble_legend_labels = \
```
TomAugspurger: Backslash -> parenthesis if possible.

```python
self.bubble_legend_sizes, self.bubble_legend_labels = (
    self._get_legend_bubbles(size_data)
)
```

pandas/plotting/_core.py
```diff
@@ -861,11 +863,20 @@ def _plot_colorbar(self, ax, **kwds):
...
-            # hide the matplotlib default for size, in case we want to change
-            # the handling of this argument later
-            s = 20
+            # Set default size if no argument is given.
+            s = 20 * size_factor
+        elif is_hashable(s) and s in data.columns:
+            # Handle the case where s is a label of a column of the df.
+            # The data is normalized to 200 * size_factor.
+            self.size_title = s
+            self.bubble_points = 200
```
TomAugspurger: Rename to `n_bubble_points` for clarity on whether it's an int or an array?

pandas/plotting/_core.py
```diff
@@ -919,6 +929,105 @@ def _make_plot(self):
             ax.errorbar(data[x].values, data[y].values,
                         linestyle='none', **err_kwds)
 
+    def _get_plot_bubbles(self, size_data):
```
TomAugspurger: For ease of testing, I think this should be a static method and take `(size_data, n_bubble_points=200, size_factor=1)`. I don't care whether it's a static method or a standalone function in the module.

pandas/plotting/_core.py
```diff
@@ -3486,6 +3595,12 @@ def scatter(self, x, y, s=None, c=None, **kwds):
               recursively. For instance, when passing [2,14] all points size
               will be either 2 or 14, alternatively.
 
+            - .. versionadded:: 0.24.0
```
TomAugspurger: I'm not sure how sphinx will handle this... I would say just do

```
s : int, str, scalar or array_like, optional
    The size of each point....
    ...
    - a column name containing numeric or ordered categorical data...
    
    .. versionchanged:: 0.24.0
       `s` can now be a column name. 
```

pandas/plotting/_core.py
```diff
@@ -3500,6 +3615,12 @@ def scatter(self, x, y, s=None, c=None, **kwds):
             - A column name or position whose values will be used to color the
               marker points according to a colormap.
 
+        size_factor : scalar, optional
+            A multiplication factor to change the size of bubbles
```
TomAugspurger: Does this only apply when `s` is a column name? If so, state that.

pandas/plotting/_core.py
```diff
@@ -3537,7 +3658,8 @@ def scatter(self, x, y, s=None, c=None, **kwds):
             ...                       c='species',
             ...                       colormap='viridis')
         """
```
TomAugspurger: Could you add an example here? I'd say just repeat the previous one with `s='species'`, even if it doesn't make a lot of sense (will need to convert species to an ordered categorical).

pandas/tests/plotting/test_frame.py
```diff
@@ -1237,6 +1237,41 @@ def test_scatter_colors(self):
...
+                        [0.4, 3.4],
+                        [4.4, 4.9],
+                        [2.7, 6.2]])
+        df = DataFrame(data, columns=['x', 'y'])
+        df['z'] = pd.Categorical(['a', 'b', 'c', 'a', 'b', 'c'], ordered=True)
+        ax = df.plot.scatter(x='x', y='y', s='z', size_factor=4)
+        bubbles = ax.collections[0]
+        bubble_sizes = bubbles.get_sizes()
+        max_data = df['z'].cat.codes.max() + 1.0
+        expected_sizes = 200.0 * 4 * (df['z'].cat.codes.values + 1)**2 / \
```
TomAugspurger: Wrap long lines with parans.

pandas/tests/plotting/test_frame.py
```diff
@@ -1237,6 +1237,41 @@ def test_scatter_colors(self):
...
+            max_data**2
+        tm.assert_numpy_array_equal(bubble_sizes, expected_sizes)
+
```
TomAugspurger: Can you add a test that directly calls `_get_plot_bubbles`? The previous test is good, but it'd be nice to avoid having to go through matplotlib's collections to get the expected value. Ideally we could hard-code one in the test.

doc/source/visualization.rst
VincentAntoine: OK

doc/source/visualization.rst
VincentAntoine: OK

doc/source/visualization.rst
VincentAntoine: OK

doc/source/visualization.rst
VincentAntoine: OK

doc/source/whatsnew/v0.24.0.txt
VincentAntoine: OK - I change it to your formulation.

TomAugspurger:
I would tend to follow matplotlib's lead here.

https://matplotlib.org/api/_as_gen/matplotlib.pyplot.plot.html does say
scatter plots are "sometimes also called bubble chart" in an aside. We can
do something similar.

doc/source/whatsnew/v0.24.0.txt
VincentAntoine: My suggestion : 
- For the title of the section in the whatnew entry we keep "bubble plots"
- the first sentence of that section starts with "Bubble plots (i.e. scatter plots in which individual marker sizes vary and reflect data) can now be produced..."

doc/source/whatsnew/v0.24.0.txt
VincentAntoine: OK

doc/source/whatsnew/v0.24.0.txt
VincentAntoine: In fact "bubble plots" already appear clearly in the main documentation in the visualization section. It probably does not matter too much how call these graphs in the whatsnew entry after all... I do as you suggest.

pandas/plotting/_core.py
VincentAntoine: It applies to all cases, not only when `s` is a column name. Which is also useful for simple scatter plots with constant marker size by the way, as it is probably more intuitive to write `size_factor=2` than `s=40` if you want to double the size of markers.

pandas/plotting/_core.py
VincentAntoine: OK

pandas/plotting/_core.py
VincentAntoine: OK

pandas/plotting/_core.py
VincentAntoine: OK. I guess it makes sense to also change the `_get_legend_bubbles` function to a static method with the same signature (`size_data, n_bubble_points=200, size_factor=1`)?

TomAugspurger (approved):
Overall, this looks good I think. Would be nice to have an example in the plot.scatter docstring, but not a dealbreaker if you don't have time.

pandas/tests/plotting/test_frame.py
VincentAntoine: Shall I *replace* this test by a test that calls _get_plot_bubbles, or *add* such a test and keep both?

TomAugspurger:
Keep both I think. They're both valuable in different ways.

pandas/plotting/_core.py
```diff
@@ -3477,7 +3591,7 @@ def scatter(self, x, y, s=None, c=None, **kwds):
         y : int or str
             The column name or column position to be used as vertical
             coordinates for each point.
-        s : scalar or array_like, optional
+        s : int, str, scalar or array_like, optional
```
datapythonista:
```suggestion
        s : str, scalar or array-like, optional
```

pandas/plotting/_core.py
datapythonista: +1

pandas/plotting/_core.py
```diff
@@ -3500,6 +3620,12 @@ def scatter(self, x, y, s=None, c=None, **kwds):
             - A column name or position whose values will be used to color the
               marker points according to a colormap.
 
+        size_factor : scalar, optional
+            A multiplication factor to change the size of bubbles
```
datapythonista:
```suggestion
            A multiplication factor to change the size of points.
```

datapythonista (commented):
@VincentAntoine do you have time to update the PR?

VincentAntoine:
@datapythonista not at the moment. When is the new version scheduled?

datapythonista:
Was scheduled for September I think. No worries, I'll see if I have time to push the changes myself, as it's almost done.

jreback:
rebase. @VincentAntoine @datapythonista @TomAugspurger if you'd have a look.

datapythonista:
I've been checking this, and personally I don't think we should add these changes to pandas. If I'm not wrong, generating the desired plots in pandas is possible with `s=df['col']` and adding the legend after the plot has been generated. For a bit of syntactic sugar of being able to use `s='col'` and implementing a legend that won't be customizable, we are adding a lot of complexity (including a function `_sci_notation` that shouldn't be in the scatter, or in plots in general).

TomAugspurger:
> I'm not wrong, generating the desired plots in pandas is possible with
s=df['col'] and adding the legend after the plot has been generated.

Not quite. That generates an exact mapping between the value in `df['col']`
and the radius of the marker. This isn't user-friendly since
you need to scale the data to something reasonable for matplotlib before
doing the plot. This may be a complex non-linear transformation,
depending on the data.

This PR implements a kind of binning (which libraries like ggplot do),
which can make the visualization easier to understand.

FWIW, I notice now that seaborn has a scatterplot method:
https://seaborn.pydata.org/generated/seaborn.scatterplot.html

datapythonista:
I missed that part, thanks for clarifying.

Wouldn't be a better approach to simply have a method that does the binning, and apply it always to `s` (I assume that we want that `s='col'` and `s=df['col']` returns the same thing).

I think that is much simpler than what's implemented here.

jreback:
closing. if you want to continue, pls ping. needs to merge master and update to comments.

PR #20241, opened 2018-03-10
DOC: update the DataFrame.mode method docstring

mkoo21:
Checklist for the pandas documentation sprint (ignore this if you are doing
an unrelated PR):

- [x] PR title is "DOC: update the <your-function-or-method> docstring"
- [x] The validation script passes: `scripts/validate_docstrings.py <your-function-or-method>`
- [x] The PEP8 style check passes: `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] The html version looks good: `python doc/make.py --single <your-function-or-method>`
- [x] It has been proofread on language by another sprint participant

Please include the output of the validation script below between the "```" ticks:

```
################################################################################
###################### Docstring (pandas.DataFrame.mode)  ######################
################################################################################

Get the mode(s) of each element along the axis selected.

Adds a row for each mode per label, filling gaps with NaN.

Parameters
----------
axis : {0 or 'index', 1 or 'columns'}, default 0
    The axis to iterate over while searching for the mode.
    To find the mode for each column, iterate over rows (``axis=0``
    default behaviour).
    To find the mode for each row, iterate over columns (``axis=1``).
numeric_only : boolean, default False
    If True, only apply to numeric dimensions.

Returns
-------
modes : DataFrame (sorted)
    A DataFrame containing the modes
    If ``axis=0``, there will be one column per column in the original
    DataFrame, with as many rows as there are modes.
    If ``axis=1``, there will be one row per row in the original
    DataFrame, with as many columns as there are modes.

Notes
-----
There may be multiple values returned for the selected
axis (when more than one item share the maximum frequency), which is
the reason why a dataframe is returned. If you want to impute missing
values with the mode in a dataframe ``df``, you can just do this:
``df.fillna(df.mode().iloc[0])``.

See Also
--------
Series.mode : Return the highest frequency value in a Series.
Series.value_counts : Returns a Series with all occuring values as
    indices and the number of occurences as values.

Examples
--------

``mode`` returns a DataFrame with multiple rows if there is more than
one mode. Missing entries are imputed with NaN.

>>> grades = pd.DataFrame({
...     'Science': [80, 70, 80, 75, 80, 75, 85, 90, 80, 70],
...     'Math': [70, 70, 75, 75, 80, 80, 85, 85, 90, 90]
... })
>>> grades.apply(lambda x: x.value_counts())
    Science  Math
70        2     2
75        2     2
80        4     2
85        1     2
90        1     2
>>> grades.mode()
   Science  Math
0     80.0    70
[...]

Issue #20298, opened 2018-03-12
DOC: docstring validation script improvements


jorisvandenbossche:
Based on the experience from the sprint, some known issues:

- args, kwargs (or kwds) issue
- no punctuation needed at end of explanation it last line is a `versionadded`
- not working for some methods that are not imported in pandas like `pandas.core.resample.Resampler.backfill`

cc @TomAugspurger @datapythonista @WillAyd others that you noticed?

TomAugspurger:
There were some like

```
param : {'a', 'b'}
    - a : do this
    - b : do thhat
```

that complained about the explanation not starting with a capital letter. Maybe we're OK with that to force a general explanation before going into individual params.

Tangentially related: whitespace normalization for doctests (https://github.com/pandas-dev/pandas/pull/20284)

TomAugspurger:
Under the see also, we could check for and error if they have `pandas.DataFrame.apply` instead of `DataFrame.apply`.

It'd be nice to somehow know if the target is valid (a decent amount of `DataFrame.groupby.head`-like references, but perhaps that's expensive to check?

jorisvandenbossche:
I think we can easily check if the first character is a `-` (sometimes adding such a first sentence can be superfluous, and you then have to leave a blank line)

> Tangentially related: whitespace normalization for doctests (#20284)

We need to do this for pytest, but it is already handled separately in the validation_script

datapythonista:
If I'm not wrong, the validation wasn't checking much on the `Returns` section, at least I've reviewed many PRs without description, or with name+type...

CCing @kynan and @mariocj89, if I'm not wrong they did some work on the `kwargs` problem during the sprint.

datapythonista:
Also, I think it could be good to warn if the user is importing pandas or numpy in the examples, or if uses `string`, `boolean`, `integer`, or `list-like` in the type descriptions.

I can take care of this ticket if nobody is working on it already, I've got couple of fixes to the generation of the csv that I didn't send it yet.

jorisvandenbossche:
Yes, certainly go ahead with doing a PR

WillAyd:
@datapythonista if it helps I started on a test for the validation script in #20061 - I can put some time into wrapping that up today so we can be sure we don't fix one thing and break another with this going forward

TomAugspurger:
The section order should be validated. I believe there's a numpydoc issue about making the order required.

TomAugspurger:
[...]

PR #22404, opened 2018-08-17
Updating `DataFrame.mode` docstring.

datapythonista:
Supersedes #20241 (source branch does not exist anymore).


pandas/core/frame.py
```diff
@@ -7213,38 +7213,88 @@ def _get_agg_axis(self, axis_num):
...
+        Notes
+        -----
+        Every column or row of the resulting DataFrame contains all its modes.
+        And possibly NaN values at the end (if other columns or rows have a
+        higher number of modes).
 
         Examples
         --------
-        >>> df = pd.DataFrame({'A': [1, 2, 1, 2, 1, 2, 3]})
+
```
jbrockmendel: Is it standard to have a newline after the section title?

datapythonista: No, it's not, good catch. I added it to the TODO list for the validation script in #20298 (it happens very often).

pandas/core/frame.py
```diff
@@ -7213,38 +7213,87 @@ def _get_agg_axis(self, axis_num):
...
         ----------
         axis : {0 or 'index', 1 or 'columns'}, default 0
-            * 0 or 'index' : get mode of each column
-            * 1 or 'columns' : get mode of each row
-        numeric_only : boolean, default False
-            if True, only apply to numeric columns
-        dropna : boolean, default True
+            The axis to iterate over while searching for the mode.
+            To find the mode for each column, use ``axis='index'``.
+            To find the mode for each row, use ``axis='columns'``.
```
gfyoung: I like the first line of your summary.  However, I would put back the two bullet points.

datapythonista: Agree, fixed. And we should probably define a standard and be consistent when documenting `axis` (added it to #20298)

pandas/core/frame.py
```diff
@@ -7217,38 +7217,88 @@ def _get_agg_axis(self, axis_num):
...
+        ...                   index=('falcon', 'horse', 'spider', 'ostrich'),
+        ...                   columns=('species', 'legs', 'wings'))
+        >>> df
+                   species  legs  wings
+        falcon        bird     2    2.0
+        horse       mammal     4    NaN
+        spider   arthropod     8    0.0
+        ostrich       bird     2    NaN
+
+        By default, missing values are not considered, and the mode of winds
```
WillAyd: Think you mean to say `wings` here and not `winds`; side note - ostriches do still have wings :-)

pandas/core/frame.py
```diff
@@ -7217,38 +7217,88 @@ def _get_agg_axis(self, axis_num):
...
+        are both 0 and 2. The second row of species and legs contains NaN,
```
WillAyd: Double backticks for `NaN`?

pandas/core/frame.py
```diff
@@ -7217,38 +7217,88 @@ def _get_agg_axis(self, axis_num):
...
+            The modes of each column or row.
+
+        See Also
+        --------
+        Series.mode : Return the highest frequency value in a Series.
+        Series.value_counts : Return the counts of values in a Series.
+
+        Notes
+        -----
+        Every column or row of the resulting DataFrame contains all its modes.
```
WillAyd: May just be me but I don't understand the Notes section - is this necessary or better explained via examples?

pandas/core/frame.py
```diff
@@ -7217,38 +7217,88 @@ def _get_agg_axis(self, axis_num):
...
-        for each mode per label, fills in gaps with nan.
+        Get the mode(s) of each element along the selected axis.
 
-        Note that there could be multiple values returned for the selected
-        axis (when more than one item share the maximum frequency), which is
-        the reason why a dataframe is returned. If you want to impute missing
-        values with the mode in a dataframe ``df``, you can just do this:
-        ``df.fillna(df.mode().iloc[0])``
+        The mode of a set of values is the value that appears most often.
+        It can be multiple values.
```
WillAyd: Can we add an example to show this?

WillAyd (changes requested):

datapythonista (commented):
addressed changes from code review.

datapythonista:
@WillAyd can you take a look and merged this, if the comments from your review are addressed correctly. Thanks!

PR #22408, opened 2018-08-18
Changes to validate_docstring script to be able to check all docstrings at once

datapythonista:
- [ ] closes #xxxx
- [X] tests added / passed
- [ ] whatsnew entry

This PR includes two main changes:
1. When calling the `scripts/validate_docstrings.py` script without parameters, it returns a JSON with the validation of all docstrings.
2. Refactoring the structure of the code, mainly to be able to implement 1.

The previous behavior when calling the script with parameters was to generate a csv with basic information about each docstring (not the full validation, but basic things like whether sections existed, or the file where the function is defined).

The main change is that `validate_one()` now returns the results as Python objects, instead of printing directly to the console. Few other changes to make the code cleaner are implemented (mainly moving `_load_obj` and `_original_callable` inside the `Docstring` class.


scripts/validate_docstrings.py
```diff
@@ -505,42 +466,107 @@ def validate_one(func_name):
...
-        return validate_all()
+        return '\n{full_line}\n{title_line}\n{full_line}\n\n'.format(
+            full_line=full_line, title_line=title_line)
+
+    if func_name is None:
+        json_doc = validate_all()
+        fd.write(json.dumps(json_doc))
     else:
-        return validate_one(function)
+        doc_info = validate_one(func_name)
```
WillAyd: Is the reason why we write different output for one vs all simply console readability? Wondering if we shouldn't align the output regardless of how this is invoked (maybe a separate PR depending on complexity)

WillAyd (changes requested):

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -490,25 +490,25 @@ def _import_path(self, klass=None, func=None):
         return base_path
 
     def test_good_class(self):
-        assert validate_one(self._import_path(  # noqa: F821
-            klass='GoodDocStrings')) == 0
+        assert len(validate_one(self._import_path(  # noqa: F821
```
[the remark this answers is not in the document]
jorisvandenbossche: In general that is true, however in this case I think it is good to test it is actually a zero-length list and not something else Falsey?

jorisvandenbossche (commented):
Can you give a quick summary of the changes? (it seems more a bit of refactor than new functionality? Or are there user-visible changes?)

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -546,7 +546,6 @@ def test_bad_generic_functions(self, func):
                      marks=pytest.mark.xfail)
     ])
     def test_bad_examples(self, capsys, klass, func, msgs):
-        validate_one(self._import_path(klass=klass, func=func))  # noqa:F821
-        err = capsys.readouterr().err
+        res = validate_one(self._import_path(klass=klass, func=func))  # noqa:F821
         for msg in msgs:
-            assert msg in err
+            assert msg in ''.join(res['errors'])
```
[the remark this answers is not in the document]
datapythonista: If I'm not wrong, the messages we're checking are no the whole error message, but part of it. So, this is equivalent to something like:

`assert 'error in summary' in ''.join(['there is a error in summary for this doc', 'some other error'])`

I think it makes more sense to use the `join`, as the error message will be clearer if this fails. Compared to checking all in a comprehension and them doing an `any` (which I guess will just say `False` is not `True`).

scripts/validate_docstrings.py
datapythonista: I see your point, but I think the lack of consistency is necessary in this case.

If the script is called for one object (a user validating the docstring they are changing), the "pretty" output in the console is what makes sense to return.

If the script is called for all objects (we as maintainers analyzing the status), a JSON with the information to be analyzed with pandas makes more sense than any console formatting.

Does it make sense?

scripts/validate_docstrings.py
```diff
@@ -44,86 +44,153 @@
...
-    if 'module' not in locals():
-        raise ImportError('No module can be imported '
-                          'from "{}"'.format(obj_name))
+    previous_line = current_section = current_subsection = ''
+    position = None
+    with open(api_fname) as f:
+        for line in f:
+            line = line.strip()
+            if len(line) == len(previous_line):
+                if set(line) == set('-'):
```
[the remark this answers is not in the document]
datapythonista: I somehow agree with what you say, but I think in this case there is no way (at least I couldn't find it) to merge both use cases without increasing the amount of code and its complexity.

datapythonista:
Thanks for the great feedback. I addressed most of the points, and commented on the others.

@jorisvandenbossche agree some more info about the PR had to be provided. I edited the description to include what's the goal, and what else has been changed.

scripts/validate_docstrings.py
```diff
@@ -44,86 +44,183 @@
...
-        func_name_split = obj_name.rsplit('.', maxsplit)
-        module = func_name_split[0]
-        func_parts = func_name_split[1:]
-        try:
-            obj = importlib.import_module(module)
-        except ImportError:
-            pass
-        else:
-            continue
+def get_api_items():
```
WillAyd: Thanks for the docstring it helps a lot! Is there a way to make a minimal test for this? Don't want to go overboard with tests here but I am worried that if this were to somehow break in the future it could allow subtle failures in subsequent tests to go through unnoticed

jorisvandenbossche:
> I edited the description to include what's the goal, and what else has been changed.

OK, that sounds good to me!

scripts/validate_docstrings.py
datapythonista: Makes sense to me, but I'd address it in a separate PR

datapythonista:
@WillAyd let me know if there is any blocker to merge this

datapythonista:
Not critical, but it'll take me few days to have the time to write those tests. And as the PR modifies the whole script, it is blocking (or will generate conflicts) with any other change to it.

#22423 is already a case, and I'd like to start creating issues for tasks in #20298, as they will save us a lot of time when reviewing the docstring updates.

But it can surely wait a couple of weeks if you think it's worth adding the tests here.

WillAyd:
Thanks @datapythonista - I'd prefer tests up front. Hate to be a blocker on this, but realistically the longer we wait for test coverage the harder it becomes to ever get there

datapythonista:
@WillAyd I added the requested tests, let me know if this can be merge now

datapythonista:
@jorisvandenbossche can you take a look at this? it's the PR that makes the docstring validation script generate the list with the errors in all docstrings.

Once this is merged I'll start creating individual tickets for all the points, so we start validating all possible errors in the docstrings (and with some luck we can add it to the CI soon).

jreback:
cc @jorisvandenbossche @WillAyd

scripts/tests/test_validate_docstrings.py
```diff
@@ -594,7 +603,82 @@ def test_bad_generic_functions(self, func):
                      marks=pytest.mark.xfail)
     ])
     def test_bad_examples(self, capsys, klass, func, msgs):
-        validate_one(self._import_path(klass=klass, func=func))  # noqa:F821
-        err = capsys.readouterr().err
+        result = validate_one(self._import_path(klass=klass, func=func))  # noqa:F821
         for msg in msgs:
-            assert msg in err
+            assert msg in ' '.join(result['errors'])
```
WillAyd: Minor nit but what does this add? Can't we just do the normal inclusion check against the list instead of joining into a string?

scripts/tests/test_validate_docstrings.py
```diff
@@ -594,7 +603,82 @@ def test_bad_generic_functions(self, func):
...
+
+
+class TestApiItems(object):
```
WillAyd: Also minor but since this test isn't used for any testing (i.e. not collected by the runner) can we remove the `Test` prefix?

scripts/tests/test_validate_docstrings.py
```diff
@@ -594,7 +603,82 @@ def test_bad_generic_functions(self, func):
...
+    randint
+''')
+
+    @pytest.mark.parametrize('idx,name', [(0, 'itertools.cycle'),
+                                          (1, 'itertools.count'),
+                                          (2, 'itertools.chain'),
+                                          (3, 'random.seed'),
+                                          (4, 'random.randint')])
+    def test_item_name(self, idx, name):
+        res = list(validate_docstrings.get_api_items(self.api_doc))
```
WillAyd: Can you use `result` instead of `res`?

scripts/validate_docstrings.py
```diff
@@ -45,86 +45,186 @@
...
-        func_parts = func_name_split[1:]
-        try:
-            obj = importlib.import_module(module)
-        except ImportError:
-            pass
-        else:
-            continue
+def get_api_items(api_doc_fd):
+    """
+    Parse api.rst file from the documentation, and extract all the functions,
```
WillAyd: Should we make the first sentence one line?

scripts/validate_docstrings.py
```diff
@@ -517,42 +511,109 @@ def validate_one(func_name):
...
+            'file_line': doc.source_file_def_line,
+            'github_link': doc.github_url,
+            'errors': errs,
+            'warnings': wrns,
+            'examples_errors': examples_errs}
+
 
-    if not errs:
-        sys.stderr.write('Docstring for "{}" correct. :)\n'.format(func_name))
+def validate_all():
```
WillAyd: Assuming this to be the main function of the change I was thinking the test cases would cover this but don't see that - oversight on my end or something we can test?

WillAyd (changes requested):
Sorry missed the last ping

scripts/tests/test_validate_docstrings.py
datapythonista: `msg in result['errors']` is different than this, for example `'a' in 'foo bar'` is `True`, but `a in ['foo', 'bar']` is `False`.

I think the `.join()` is simpler than another loop, is it what you would do, or you were thinking on the previous case?

WillAyd: No but I thought `msg` would explicitly match one of the `errors`; will take a look more deeply on next review if that assumption is incorrect

scripts/validate_docstrings.py
datapythonista: This function is just boilerplate code, I don't think it adds any value to test that, as we would be mocking everything.

datapythonista:
@WillAyd addressed or commented your review, can you take another look please?

WillAyd (changes requested):
Thanks - just some more comments / detail

scripts/validate_docstrings.py
```diff
@@ -44,86 +44,153 @@
...
 
     @property
     def source_file_name(self):
-        fname = inspect.getsourcefile(self.method_obj)
-        if fname:
-            fname = os.path.relpath(fname, BASE_PATH)
-            return fname
+        try:
+            fname = inspect.getsourcefile(self.code_obj)
+        except TypeError:
```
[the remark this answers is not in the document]
datapythonista: what exactly? the `inspect.getsourcefile()` returns the source file where a Python object is implemented. The `except TypeError` is explain in the comment below.

scripts/validate_docstrings.py
WillAyd: My mistake - thanks for clarifying

scripts/validate_docstrings.py
```diff
@@ -45,86 +45,188 @@
...
+        self.obj = obj
+        self.code_obj = self._to_original_callable(obj)
+        self.raw_doc = obj.__doc__ or ''
+        self.clean_doc = pydoc.getdoc(obj)
         self.doc = NumpyDocString(self.clean_doc)
 
     def __len__(self):
         return len(self.raw_doc)
 
+    @staticmethod
```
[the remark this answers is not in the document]
WillAyd: All of the calls within the code base are using an instance of the class (save the example in the docstring). With that being the case would prefer to remove the decorator unless there's actually a use case for it; otherwise just adds fluff

WillAyd (changes requested):
One last comment from me then I'll leave to others for review :-)

scripts/validate_docstrings.py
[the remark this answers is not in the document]
datapythonista: Not a problem at all to make it a regular method, I leave often static methods as regular methods, as it makes almost no difference.

But I don't understand your comment about using an instance of the class. Do you mean that it's not something that we usually use?

The way I see it is that methods of classes in general access the attributes and other methods of the class. As they are encapsulate a set of data and code that somehow works together. Functions in the other hand are somehow independent. In few cases, there are functions that while they do not access anything else from the class, they make more sense inside the class than outside. This is the case here, as `_load_obj` is only used by other methods of the class, and living outside at the same level of the class `Docstring`, IMO makes it look more important in the code than what really is, as conceptually it makes more sense at the same level of the other methods.

Assuming that as a method of the class is a better decision, having it as an staticmethod makes very quick to understand that it's a function inside of a class, and not a method that can access other methods or attributes. I personally find this information very valuable when reading code. Normal methods can access anything in the class, adding an extra complexity in the way that you can't see them as pure input->output units, but that can instead use somehow global variables. A static method doesn't have this complexity, and I know I can expect it to use only whatever parameter it receives, return a result, and not modify anything else. So, it gives a lot of information without having to read its code. The cost is just having the text `@staticmethod` instead of `self`. I think the few characters of difference is a very good price for the amount of information it provides. And I don't think that not using them often in the pandas code itself is a good reason for not using it here. May be we should start using them more there.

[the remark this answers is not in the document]
jorisvandenbossche: \+1 on explicitly using staticmethod, if it is actually one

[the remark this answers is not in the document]
WillAyd: My point is that I don't see the benefit of just adding things like this without any accompanying test regardless of how trivial the change may seem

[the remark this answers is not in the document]
datapythonista: I'm a bit lost on what's the connection between the staticmethod and the testing here.

The benefit of this PR is that makes the code more organized and better. And it implements a way to extract detailed information of the errors in the docstrings. So I can quickly generate the list of functions that fail a specific validation, fix the problem in all them, and start checking in the CI those validations. So the errors don't happen again, and I don't need to waste time on manually reviewing them in all the docstrings PRs.

A huge benefit for me. :)  So would be great to have this merged. Is there any blocker, or can I merge on green?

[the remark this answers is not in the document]
WillAyd: I've blocked long enough so if others are fine with it majority rules :-) 

Just generally trying to minimize PRs and keep them directed towards one clear goal at a time, hence the pushback on this one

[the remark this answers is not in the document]
datapythonista: hehe, I'm trying to maximize PRs, so hopefully we'll end up in a good equilibrium. ;) After this is merged I'll start creating issues for the points in #20298, so we have better validation of the docstrings.

@jorisvandenbossche happy to merge this?

jorisvandenbossche:
Sorry forgot about this one. Thanks for moving this forward!

PR #22413, opened 2018-08-19
BLD: Install scripts tests only during inplace

gfyoung:
The `scripts` directory is not visible if the installation is not inplace.

Follow-up to #20061.

cc @WillAyd @datapythonista

### DIFF
--- a/pandas/tests/scripts/test_validate_docstrings.py
+++ b/pandas/tests/scripts/test_validate_docstrings.py
@@ -448,22 +448,37 @@ class TestValidator(object):
     @pytest.fixture(autouse=True, scope="class")
     def import_scripts(self):
         """
-        Because the scripts directory is above the top level pandas package
-        we need to hack sys.path to know where to find that directory for
-        import. The below traverses up the file system to find the scripts
-        directory, adds to location to sys.path and imports the required
-        module into the global namespace before as part of class setup,
-        reverting those changes on teardown.
+        Import the validation scripts from the scripts directory.
+
+        Because the scripts directory is above the top level pandas package,
+        we need to modify `sys.path` so that Python knows where to find it.
+
+        The code below traverses up the file system to find the scripts
+        directory, adds the location to `sys.path`, and imports the required
+        module into the global namespace before as part of class setup.
+
+        During teardown, those changes are reverted.
         """
+
         up = os.path.dirname
+        global_validate_one = "validate_one"
         file_dir = up(os.path.abspath(__file__))
-        script_dir = os.path.join(up(up(up(file_dir))), 'scripts')
+
+        script_dir = os.path.join(up(up(up(file_dir))), "scripts")
         sys.path.append(script_dir)
-        from validate_docstrings import validate_one
-        globals()['validate_one'] = validate_one
+
+        try:
+            from validate_docstrings import validate_one
+            globals()[global_validate_one] = validate_one
+        except ImportError:
+            # Import will fail if the pandas installation is not inplace.
+            raise pytest.skip("pandas/scripts directory does not exist")
+
         yield
+
+        # Teardown.
         sys.path.pop()
-        del globals()['validate_one']
+        del globals()[global_validate_one]
 
     def _import_path(self, klass=None, func=None):
         """


WillAyd:
Would the other option be to just skip these tests if the scripts folder isn't included? I actually wonder if that would be preferable especially in cases where users install a binary and try to run `pd.test()` in the REPL

gfyoung:
@WillAyd : I don't see why we should be packaging these tests when we do a release of pandas to the general public.  That's one reason why I would prefer exclusion.

jreback (commented):
shouldn’t you just put a skip in the tests themselves?

gfyoung:
@jreback : See my answer [above](https://github.com/pandas-dev/pandas/pull/22413#issuecomment-414094889).  The only time these tests should be running is if we do an inplace install.  Otherwise, why should we be packaging those tests in the distribution (e.g. the public releases).

jorisvandenbossche:
> Would the other option be to just skip these tests if the scripts folder isn't included?

\+1

jorisvandenbossche:
@gfyoung I understand your argument, but I think the added complexity (I would expect to find logic about whether the tests are ran/skipped in the tests itself, our `setup.py` is already very complex) is not worth the few kb's for those tests.

gfyoung:
Fair enough.  IMHO, these validation tests should not have been located in the `pandas/tests` directory, but oh well. 🤷‍♂️

datapythonista:
Not sure what it involves in the CI configuration, but I'm +1 on moving the tests for the scripts to `scripts/tests` too.

gfyoung:
@jreback @jorisvandenbossche : Thoughts on moving the tests?  As much as I know both of you don't want the extra complexity in `setup.py` of "skipping" the copying of certain files during distribution, I am also not a fan of seeing tests fail locally just because the `scripts` isn't packaged.

jorisvandenbossche:
> I am also not a fan of seeing tests fail locally just because the scripts isn't packaged.

They wouldn't fail, the idea is that they would be skipped.

That said, I am also fine with having the tests live in `scripts/tests/, if that doesn't complicate the travis set-up too much to also run those (I suppose now we everywhere only test `pandas` directory?)

WillAyd:
I'm -1 on moving to `scripts/tests`. If we are worried about the added size of the new module then we are making micro-optimizations whose benefit would pale in comparison to the effort making sure that we can account for tests that live outside of the `tests` directory

gfyoung:
> whose benefit would pale in comparison to the effort making sure that we can account for tests that live outside

@WillAyd : I'm not sure we can make this assertion yet without giving it a shot first. 😉

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -448,22 +448,37 @@ class TestValidator(object):
...
-        Because the scripts directory is above the top level pandas package
-        we need to hack sys.path to know where to find that directory for
-        import. The below traverses up the file system to find the scripts
-        directory, adds to location to sys.path and imports the required
-        module into the global namespace before as part of class setup,
-        reverting those changes on teardown.
+        Import the validation scripts from the scripts directory.
+
+        Because the scripts directory is above the top level pandas package,
+        we need to hack `sys.path` to know where to find it for import.
```
jreback: avoid the term 'hack'

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -448,22 +448,37 @@ class TestValidator(object):
...
         sys.path.append(script_dir)
-        from validate_docstrings import validate_one
-        globals()['validate_one'] = validate_one
+
+        try:
+            from validate_docstrings import validate_one
+            globals()[global_validate_one] = validate_one
+        except ImportError:
+            # Import will fail if the pandas installation is not inplace.
+            raise pytest.skip("scripts directory does not exist")
```
jreback: can you say pandas/scripts

jreback (changes requested):

pandas/tests/scripts/test_validate_docstrings.py
gfyoung: Done.

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -448,22 +448,37 @@ class TestValidator(object):
...
         sys.path.append(script_dir)
-        from validate_docstrings import validate_one
-        globals()['validate_one'] = validate_one
+
+        try:
+            from validate_docstrings import validate_one
+            globals()[global_validate_one] = validate_one
+        except ImportError:
+            # Import will fail if the pandas installation is not inplace.
+            raise pytest.skip("pandas/scripts directory does not exist")
```
WillAyd: Not sure if this causes an issue but from what I've tested locally if using the imperative skip `validate_docstrings` stays in globals as does the modification to system path. Perhaps we should ensure that code gets executed in the `except` block here?

gfyoung: * Fair point regarding system path modification.
* Not sure how `validate_docstrings` stays in globals if the import fails...

jreback (approved):

jreback:
let's see if this fixes the wheels

jreback:
ok wheels fixed: https://travis-ci.org/MacPython/pandas-wheels/builds/418513881

pandas/tests/scripts/test_validate_docstrings.py
gfyoung: @WillAyd : Addressed the change in follow-up (#22456)

PR #12915, opened 2016-04-17
BUG: Enforce parse_dates as bool when scalar

gfyoung:
Title is self-explanatory.  Closes #5636.

PR #22418, opened 2018-08-19
Remove weird comment in parse_dates.py

gfyoung:
Was added back in #12915...not sure why I did that. 😕


pandas/tests/io/parser/parse_dates.py
```diff
@@ -365,7 +365,7 @@ def test_parse_tz_aware(self):
         assert stamp.minute == 39
         try:
             assert result.index.tz is pytz.utc
-        except AssertionError:  # hello Yaroslav
+        except AssertionError:
```
jreback: actually why do we have an assertion error check  here at all?

gfyoung: Not sure...that code is at least 2 years old.

I imagine that `pytz` compatibility has improved since then...

jreback: can u try to remove this and see

PR #22421, opened 2018-08-19
Bug22373: Enhance documentation for RangeIndex

### LINKED ISSUE #22373: version 0.23: RangeIndex argument order changed, docs stale
```python
>>> import pandas as pd
>>> pd.__version__
'0.23.4'
>>> pd.RangeIndex(0, 3, 1, 'name')
  File "/home/mpfluege/.local/miniconda3/envs/pitchwalk/lib/python3.7/site-packages/pandas/core/indexes/range.py", line 74, in __new__
    cls._validate_dtype(dtype)
  File "/home/mpfluege/.local/miniconda3/envs/pitchwalk/lib/python3.7/site-packages/pandas/core/indexes/range.py", line 161, in _validate_dtype
    if not (dtype is None or is_int64_dtype(dtype)):
  File "/home/mpfluege/.local/miniconda3/envs/pitchwalk/lib/python3.7/site-packages/pandas/core/dtypes/common.py", line 991, in is_int64_dtype
    tipo = _get_dtype_type(arr_or_dtype)
  File "/home/mpfluege/.local/miniconda3/envs/pitchwalk/lib/python3.7/site-packages/pandas/core/dtypes/common.py", line 1872, in _get_dtype_type
    return _get_dtype_type(np.dtype(arr_or_dtype))
TypeError: data type "name" not understood
```
#### Problem description

[Documentation of RangeIndex](https://pandas-docs.github.io/pandas-docs-travis/generated/pandas.RangeIndex.html) states that it has the arguments `start`, `stop`, `step`, `name` and `copy`. However, the actual signature is `RangeIndex(start=None, stop=None, step=None, dtype=None, copy=False, name=None, fastpath=False)`, which has `dtype` as fourth argument. This breaks scripts that relied on the fourth argument being the name, which used to be the case:
```python
>>> import pandas as pd
>>> pd.__version__
'0.22.0'
>>> pd.RangeIndex(0, 3, 1, 'name')
RangeIndex(start=0, stop=3, step=1, name='name')
```

#### Expected Output

Either revert to the old argument order or at least update the documentation.

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.7.0.final.0
python-bits: 64
OS: Linux
OS-release: 4.4.140-62-default
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: de_DE.UTF-8
LOCALE: de_DE.UTF-8

pandas: 0.23.4
pytest: None
pip: 10.0.1
setuptools: 40.0.0
Cython: None
numpy: 1.15.0
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: None
sphinx: None
patsy: None
dateutil: 2.7.3
pytz: 2018.5
blosc: None
bottleneck: None
tables: None
numexpr: 2.6.7
feather: None
matplotlib: 2.2.3
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: None
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

mikapfl:
Corrects documentation for RangeIndex and enhances documentation for RangeIndex and RangeIndex.from_range
closes #22373
passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

### DIFF
--- a/pandas/core/indexes/range.py
+++ b/pandas/core/indexes/range.py
@@ -40,12 +40,13 @@ class RangeIndex(Int64Index):
     Parameters
     ----------
     start : int or other RangeIndex instance, default 0
-        If of type `int` and `stop` is not given, interpreted as `stop` instead.
+        If of type `int` and `stop` is not given, interpreted as `stop`
+        instead.
     stop : int, default 0
     step : int, default 1
     dtype : numpy dtype or pandas type, default int64
-        Only the default `int64` is supported, argument accepted for homogeneity
-        with other index types.
+        Only the default `int64` is supported, argument accepted for
+        homogeneity with other index types.
     copy : bool, default False
         Unused, accepted for homogeneity with other index types.
     name : object, optional
@@ -134,8 +135,8 @@ class RangeIndex(Int64Index):
         name : object
              Name to be stored in the index
         dtype : numpy dtype or pandas type, default int64
-            Only the default int64 is supported, argument accepted for homogeneity
-            with other index types.
+            Only the default int64 is supported, argument accepted for
+            homogeneity with other index types.
         **kwargs
             These parameters will be passed to :class:`~pandas.RangeIndex`.
         """
--- a/pandas/core/indexes/range.py
+++ b/pandas/core/indexes/range.py
@@ -46,10 +46,10 @@ class RangeIndex(Int64Index):
     dtype : numpy dtype or pandas type, default int64
         Only the default int64 is supported, argument accepted for homogeneity
         with other index types.
-    name : object, optional
-        Name to be stored in the index
     copy : bool, default False
         Unused, accepted for homogeneity with other index types.
+    name : object, optional
+        Name to be stored in the index
 
     See Also
     --------
--- a/pandas/core/indexes/range.py
+++ b/pandas/core/indexes/range.py
@@ -28,10 +28,19 @@ from pandas.core.indexes.numeric import Int64Index
 class RangeIndex(Int64Index):
 
     """
-    Immutable Index implementing a monotonic integer range.
+    Immutable :class:`~pandas.Index` implementing a monotonic integer range.
 
-    RangeIndex is a memory-saving special case of :class:`~pandas.Int64Index`
-    limited to representing monotonic ranges. Using RangeIndex may in some
+    Like a python `range`, a `RangeIndex` contains values from `start`
+    (inclusive) to `stop` (exclusive)
+    with increments of size `step`. This means that for a positive `step`,
+    the contents of a `RangeIndex` r are determined by
+    the formula ``r[i] = start + step*i where i >= 0 and r[i] < stop``.
+    Accordingly, for a negative `step`, the contents of the `RangeIndex` are
+    still determined by the formula ``r[i] = start + step*i``, but the constraints
+    are ``i >= 0 and r[i] > stop``.
+
+    `RangeIndex` is a memory-saving special case of :class:`~pandas.Int64Index`
+    limited to representing monotonic ranges. Using `RangeIndex` may in some
     instances improve computing speed.
 
     This is the default index type used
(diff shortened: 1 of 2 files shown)


gfyoung (approved):

pandas/core/indexes/range.py
```diff
@@ -30,23 +30,30 @@ class RangeIndex(Int64Index):
...
-        If int and "stop" is not given, interpreted as "stop" instead.
-    stop : int (default: 0)
-    step : int (default: 1)
-    name : object, optional
-        Name to be stored in the index
+    start : int or other RangeIndex instance, default 0
+        If of type `int` and `stop` is not given, interpreted as `stop`
+        instead.
+    stop : int, default 0
+    step : int, default 1
```
datapythonista: It'f be good to have these documented. Also, `start`, which would be good to have a description, besides the behavior described.

datapythonista (commented):
Thanks for the changes, looking good. If you want to add examples as part of this PR, that would be great (we can do it in a separate PR too).

Also, if you can add periods at the end of the `See Also` items description, that would be great.

pandas/core/indexes/range.py
mikapfl: Re documenting `start`, `stop` and `step`: I also thought that it would be nice to document them, and figured we could just recycle the [python documentation for `range`](https://docs.python.org/3/library/stdtypes.html#typesseq-range). However, the documentation for the arguments is very basic in python as well: "`stop`: The value of the stop parameter". Maybe there is not much else to say?

But I'll try to adapt the rest of the python `range` docs (i.e. the long description) and try to come up with some useful examples.

datapythonista: I think it can be useful to say whether the `stop` value is included or not. It's not obvious to me that `range(0, 10)` doesn't include the `10`.

Also, for beginner users, something as simple as  `start : The first value of the generated range.`  can be useful. `step` can be tricky, do we support negative values, as `range`?

I do think there is a lot to say. :)

jreback: > think it can be useful to say whether the stop value is included or not. It's not obvious to me that range(0, 10) doesn't include the 10.

this is standard python behavior to NOT include the rhs on ranges.

pandas/core/indexes/range.py
```diff
@@ -30,23 +30,30 @@ class RangeIndex(Int64Index):
...
+    dtype : numpy dtype or pandas type, default int64
+        Only the default `int64` is supported, argument accepted for
+        homogeneity with other index types.
     copy : bool, default False
         Unused, accepted for homogeneity with other index types.
+    name : object, optional
+        Name to be stored in the index
```
jreback: this constructor should be discouraged for using integers on ranges, this is the reason we have ``.from_range``. Actually folks should generally NOT use specific constructors anyhow. ``pd.Index`` dispatches in most cases. Only if you have very very specific requreiments do you need to use this constructor.

jreback (changes requested):

datapythonista:
Agree that not including the stop of the range is standard, my point is that it's not obvious unless you know about this standard. So, I think it's worth mentioning it in the documentation.

If using the constructor is discouraged, we should also say it in the description I guess.

mikapfl:
@datapythonista I have added documentation and examples showing the half-inclusive behaviour of RangeIndex. The documentation is based on the python range documentation.

@jreback I have added langeuage and examples to Index and RangeIndex to guide users towards using Index(range(…)) instead of RangeIndex(…).

datapythonista (commented):
Can you change the default of `tupleize_cols` from `(default: True)` to `default True

And finish the comment in `Notes` with a period.

alimcmaster1:
@mikapfl could you rebase and merge master if you are still looking to take this on? If not I can potentially take a look

datapythonista:
@alimcmaster1, if you want to open a new PR with this. Closing as this seems abandoned.

PR #22423, opened 2018-08-19
BUG: Ignore versionadded directive when checking for periods at docstring end

### LINKED ISSUE #22405: Consider directives when validating docstrings parameters
When running the docstring validation script, errors are mistakenly reported if parameters have a `deprecated`, `versionadded` or `versionchanged` directive.

For example, when running `./scripts/validate_docstrings.py pandas.DataFrame.mode`, the next error is returned by the script:

```

Errors found:
	Errors in parameters section
		Parameter "dropna" description should finish with "."
```
But the `dropna` parameter description does finish with a period:

```
dropna : bool, default True
    Don't consider counts of NaN/NaT.

    .. versionadded:: 0.24.0
```

The problem is that the `.. versionadded::` directive is not considered.

The script should ignore the directive when validating, and shouldn't report the error in this case.

bengineer19:
Will ignore the `versionadded` directive when checking for `'.'` at the end of descriptions.  

- [x] closes #22405
- [x] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/tests/scripts/test_validate_docstrings.py
+++ b/pandas/tests/scripts/test_validate_docstrings.py
@@ -193,6 +193,44 @@ class GoodDocStrings(object):
         """
         pass
 
+    def mode(self, axis=0, numeric_only=False, dropna=True):
+        """
+        Get the mode(s) of each element along the axis selected. Adds a row
+        for each mode per label, fills in gaps with nan.
+
+        Note that there could be multiple values returned for the selected
+        axis (when more than one item share the maximum frequency), which is
+        the reason why a dataframe is returned. If you want to impute missing
+        values with the mode in a dataframe ``df``, you can just do this:
+        ``df.fillna(df.mode().iloc[0])``
+
+        Parameters
+        ----------
+        axis : {0 or 'index', 1 or 'columns'}, default 0
+            Describe axis.
+        numeric_only : boolean, default False
+            Describes numeric_only.
+        dropna : boolean, default True
+            This param tests that the versionadded directive doesn't break the
+            checks for the ending period.
+            Don't consider counts of NaN/NaT.
+
+            .. versionadded:: 0.24.0
+
+        Returns
+        -------
+        modes : DataFrame (sorted)
+
+        Examples
+        --------
+        >>> df = pd.DataFrame({'A': [1, 2, 1, 2, 1, 2, 3]})
+        >>> df.mode()
+           A
+        0  1
+        1  2
+        """
+        pass
+
 
 class BadGenericDocStrings(object):
     """Everything here has a bad docstring
@@ -374,6 +412,18 @@ class BadParameters(object):
            Doesn't end with a dot
         """
 
+    def no_description_period_with_version_added(self, kind):
+        """
+        Forgets to add a period, and also includes a directive.
+
+        Parameters
+        ----------
+        kind : str
+           Doesn't end with a dot
+
+           .. versionadded:: 0.00.0
+        """
+
     def parameter_capitalization(self, kind):
         """
         Forgets to capitalize the description.
@@ -495,7 +545,7 @@ class TestValidator(object):
 
     @pytest.mark.parametrize("func", [
         'plot', 'sample', 'random_letters', 'sample_values', 'head', 'head1',
-        'contains'])
+        'contains', 'mode'])
     def test_good_functions(self, func):
         assert validate_one(self._import_path(   # noqa: F821
             klass='GoodDocStrings', func=func)) == 0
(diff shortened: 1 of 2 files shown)


WillAyd:
Can you add a test case for this? We also have a versionchanged directive which we should account for here

TomAugspurger:
And the `deprecated` directive as well probably.

scripts/validate_docstrings.py
```diff
@@ -465,7 +466,17 @@ def validate_one(func_name):
...
+            period_check_index = -1
+            for directive in DIRECTIVES:
+                if directive in doc.parameter_desc(param):
+                    # Get index of character before start of directive
+                    index = doc.parameter_desc(param).index(directive) - 1
+                    # If this directive is closest to the description, use it.
+                    if index < period_check_index or period_check_index is -1:
+                        period_check_index = index
+
+            if doc.parameter_desc(param)[period_check_index] != '.':
```
datapythonista: What do you think about creating a new property in the class `Docstring` that returns the `parameter_desc` but without directives?

I think the code will be much more readable if we have this logic there, and in this part of the code where all validations happen we simply have something like `if doc.parameter_desc_without_directives(param)[-1] != '.':`

scripts/validate_docstrings.py
```diff
@@ -42,6 +42,7 @@
 
 
 PRIVATE_CLASSES = ['NDFrame', 'IndexOpsMixin']
+DIRECTIVES = ['.. versionadded', '.. versionchanged', '.. deprecated']
```
datapythonista: afaik, all sphinx directives start with `.. ` so I think it's better to have just the names here.

WillAyd (changes requested):

scripts/validate_docstrings.py
bengineer19: Yes, I agree that would be more useful and readable; I'll make a property.

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -193,6 +193,53 @@ def contains(self, pat, case=True, na=np.nan):
         """
         pass
 
+    def mode(self, axis=0, numeric_only=False, dropna=True):
```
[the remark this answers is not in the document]
bengineer19: Sure thing, sounds good.

scripts/validate_docstrings.py
```diff
@@ -236,6 +237,15 @@ def parameter_type(self, param):
     def parameter_desc(self, param):
         return self.doc_parameters[param][1]
 
+    def parameter_desc_without_directives(self, param):
```
WillAyd: What happens if you just use this logic in `parameter_desc` instead of as a separate method? I can't think of a case where we want the directives to be considered part of the description, and the way you've coded this works well for the missing period but may not be generalizable to other issues that could come up

datapythonista: That's a good point. I'd probably keep a `raw_parameter_desc` with what `parameter_desc` has now, as in the future I'd probably like to have the values of the directives (for example to find the deprecated parameters we need to delete in the next version).

scripts/validate_docstrings.py
```diff
@@ -233,9 +234,19 @@ def correct_parameters(self):
     def parameter_type(self, param):
         return self.doc_parameters[param][0]
 
-    def parameter_desc(self, param):
+    def raw_parameter_desc(self, param):
```
WillAyd: @datapythonista I know you wanted this but I would prefer not to include unless it serves a purpose. May even be better served as a keyword argument going forward instead of a dedicated method so would rather not go down this path until needed.

Outside of that I think this change looks good

pandas/tests/scripts/test_validate_docstrings.py
```diff
@@ -193,6 +193,25 @@ def contains(self, pat, case=True, na=np.nan):
...
+        axis : str
+            Sentence ending in period, followed by single directive.
+
+            .. versionchanged:: 0.1.2
+
+        numeric_only : boolean
+            Sentence ending in period, followed by multiple directives.
+
+            .. versionadded:: 0.1.2
+            .. deprecated:: 0.00.0
```
datapythonista: I'm not sure about `versionadded` and `versionchanged`, but `deprecated` can have a description after if, for example:

```
          .. deprecated:: 0.21.0
              Use :func:`pandas.read_csv` instead.
```

And it can be even multiline. Do you mind adding a test for that? I'm not sure if this is working with the current implementation.

datapythonista: Also, if you check the `convert_datetime64` of `to_records`, there are cases where the directives come before the description. I'm happy if we consider only valid having them in one place (before or after the description). But, can we make the script generate a descriptive error for it? I guess with the current implementation we'll report that the parameter has no description.

datapythonista (changes requested):
Looks good, but I was checking some docstrings, and there are couple of cases we didn't consider. Can you take care of them?

WillAyd (approved):
This looks good to me

datapythonista (approved):
This will fix most cases, happy to merge on green, and take care of other cases in separate PRs. Thanks @bengineer19

Issue #22427, opened 2018-08-20
CI: Travis tests timing out


TomAugspurger:
E.g. https://travis-ci.org/pandas-dev/pandas/jobs/417983438

```
........................................................................ [ 95%]
.......................
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

TomAugspurger:
Seems to have first appeared in https://travis-ci.org/pandas-dev/pandas/jobs/418084453. It actually failed the PR build: https://travis-ci.org/pandas-dev/pandas/jobs/418011554, https://github.com/pandas-dev/pandas/pull/22424.

Trying to reproduce locally now.

TomAugspurger:
Unable to reproduce. I'm going to temporarily revert #22424.

TomAugspurger:
cc @webartifex if you have any clues as to why this may be, though I don't suspect you do. This is a strange one :)

jorisvandenbossche:
Sorry about merging the PR. I actually saw that travis was failing, but since it was a time-out, just thought it was a temporary glitch .. Should have restarted the failing build first before merging.

TomAugspurger:
Ah indeed I should have looked more closely at which one was failing. I'll dig deeper in the logs to see what packages changed.

TomAugspurger:
```diff
diff --git a/b b/a
index d2ff0d5..5e1e74c 100644
--- a/b
+++ b/a
@@ -15,7 +15,7 @@ backports_abc             0.5                      py27_0
 beautifulsoup4            4.6.3                    py27_0  
 blas                      1.0                    openblas  
 blinker                   1.4                      py27_0  
-blosc                     1.14.3               hdbcaa40_0  
+blosc                     1.14.4               hdbcaa40_0  
 boto                      2.49.0                   py27_0  
 boto3                     1.7.62                   py27_1  
 botocore                  1.10.62                  py27_0  
@@ -52,7 +52,7 @@ freetype                  2.9.1                h8a8886c_0
 funcsigs                  1.0.2                    py27_0  
 functools32               3.2.3.2                  py27_1  
 futures                   3.2.0                    py27_0  
-gcsfs                     0.1.1                      py_0    conda-forge
+gcsfs                     0.1.2                      py_0    conda-forge
[...]

PR #22424, opened 2018-08-19
CI: add missing tzlocal dependency (rpy2, doc build)

webartifex:
- [x] closes #22412
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`
- [x] whatsnew entry

PR #22429, opened 2018-08-20
Pin blosc to 1.14.3 for travis-27 build

TomAugspurger:
This reverts commit 52e186af3e0fed9415a2db9e0c8d7d75d1267a54.

xref https://github.com/pandas-dev/pandas/issues/22427 and https://github.com/pandas-dev/pandas/pull/22424

### DIFF
--- a/pandas/tests/extension/integer/test_integer.py
+++ b/pandas/tests/extension/integer/test_integer.py
@@ -784,7 +784,7 @@ def test_groupby_mean_included():
 
 def test_astype_nansafe():
     # https://github.com/pandas-dev/pandas/pull/22343
-    arr = IntegerArray([np.nan, 1, 2], dtype="Int8")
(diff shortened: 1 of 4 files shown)


pandas/tests/extension/integer/test_integer.py
```diff
@@ -770,7 +770,7 @@ def test_groupby_mean_included():
     df = pd.DataFrame({
         "A": ['a', 'b', 'b'],
         "B": [1, None, 3],
-        "C": IntegerArray([1, None, 3], dtype='Int64'),
```
jorisvandenbossche: I would do this in a separate PR

TomAugspurger: Keeping it here, now that we know that the doc PR isn't related.

PR #22436, opened 2018-08-21
BUG #19860 Corrected use of mixed indexes with .at

### LINKED ISSUE #19860: BUG?: .at not working on object indexes containing some integers
Version 0.22.0

#### Problem description
Using the .at - Method on an Index which contains Integers as well as str/objects raises an Error. This used to be possible using the ``.get_value()``-Method. As ``.at`` is the designated successor (#15269) the same behaviour should be supported.
I also noticed that ``.get_value`` is approx. twice as fast as ``.at``. Is there a specific reason to stick with ``.at``?  (see again #15269 for a speed comparison)

#### Code Sample
```python
import pandas as pd
import numpy as np

data = np.random.randn(10, 5)
df = pd.DataFrame(data, columns=['a', 'b', 'c', 1, 2])
df.at[0, 1]
```

Raises:
```python
  File "/home/thielc/anaconda3/lib/python3.6/site-packages/pandas/core/indexing.py", line 1868, in __getitem__
    key = self._convert_key(key)
  File "/home/thielc/anaconda3/lib/python3.6/site-packages/pandas/core/indexing.py", line 1915, in _convert_key
    raise ValueError("At based indexing on an non-integer "
ValueError: At based indexing on an non-integer index can only have non-integer indexers
```

mariuspot:
`.at` incorrectly disallowed the use of integer indexes when a mixed index was used
```
s = Series([1,2,3,4,5],index=['a','b','c',1,2])
s.at['a'] # returns 1
s.at[1]   # returns 4
s.at[4]   # raises KeyError, doesn't do fallback indexing
```

Made sure fallback indexing doesn't happen on mixed indexes

- [X] closes #19860
- [X] tests added / passed
- [X] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -659,11 +659,8 @@ Indexing
 - Fixed ``DataFrame[np.nan]`` when columns are non-unique (:issue:`21428`)
 - Bug when indexing :class:`DatetimeIndex` with nanosecond resolution dates and timezones (:issue:`11679`)
 - Bug where indexing with a Numpy array containing negative values would mutate the indexer (:issue:`21867`)
-<<<<<<< HEAD
 - Bug where mixed indexes wouldn't allow integers for ``.at`` (:issue:`19860`)
-=======
 - ``Float64Index.get_loc`` now raises ``KeyError`` when boolean key passed. (:issue:`19087`)
->>>>>>> upstream/master
 
 Missing
 ^^^^^^^
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -656,6 +656,7 @@ Indexing
 - Fixed ``DataFrame[np.nan]`` when columns are non-unique (:issue:`21428`)
 - Bug when indexing :class:`DatetimeIndex` with nanosecond resolution dates and timezones (:issue:`11679`)
 - Bug where indexing with a Numpy array containing negative values would mutate the indexer (:issue:`21867`)
+- Bug where mixed indexes wouldn't allow integers for ``.at`` (:issue:`19860`)
 
 Missing
 ^^^^^^^
--- a/pandas/tests/indexing/test_indexing.py
+++ b/pandas/tests/indexing/test_indexing.py
@@ -610,6 +610,22 @@ class TestFancy(Base):
     def test_index_not_contains(self, index, val):
         assert val not in index
 
+    @pytest.mark.parametrize("index,val", [
+        (Index([0, 1, '2']), 0),
+        (Index([0, 1, '2']), '2'),
+    ])
+    def test_mixed_index_contains(self, index, val):
+        # GH 19860
+        assert val in index
+
+    @pytest.mark.parametrize("index,val", [
+        (Index([0, 1, '2']), '1'),
+        (Index([0, 1, '2']), 2),
+    ])
+    def test_mixed_index_not_contains(self, index, val):
+        # GH 19860
+        assert val not in index
+
     def test_index_type_coercion(self):
 
         with catch_warnings(record=True):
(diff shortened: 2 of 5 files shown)


pandas/tests/indexing/test_indexing.py
```diff
@@ -666,6 +666,20 @@ def test_index_type_coercion(self):
                     idxr(s2)['0'] = 0
                     assert s2.index.is_object()
 
+    @pytest.mark.parametrize("index,val", [
```
jreback: can you add a comment with the issue number

jreback: these tests should be in pandas/tests/indexes/test_base (look for other contains tests)

pandas/tests/indexing/test_indexing.py
```diff
@@ -710,6 +724,27 @@ def test_float_index_at_iat(self):
         for i in range(len(s)):
             assert s.iat[i] == i + 1
 
+    def test_mixed_index_at_iat(self):
```
jreback: move these to pandas/tests/indexing/test_scalar.py (see where other ones are)

jreback: add a gh comment with the issue number

jreback: can you add the test from the OP (was on dataframe)

pandas/tests/indexing/test_indexing.py
```diff
@@ -710,6 +724,27 @@ def test_float_index_at_iat(self):
...
+        s = Series([1, 2, 3, 4, 5], index=['a', 'b', 'c', 1, 2])
+        for el, item in s.iteritems():
+            assert s.at[el] == item
+        for i in range(len(s)):
```
jreback: can you check that we have similar tests for .loc (I think so), if not can you add a loc version (inside same test is ok). maybe even just add the .loc tests and move from elsewhere.

you can parameterize these over .at / .loc

jreback (changes requested):
looks pretty good. can you add a whatsnew note in bugfixes indexing for 0.24

jreback:
also there *might* be other .at issues which this closes. pls have a look if you can.

cc @toobaz

pandas/tests/indexing/test_indexing.py
mariuspot: `test_index_not_contains` and `test_index_contains` are in the same file a few lines above, should I move these up or create pandas/tests/indexes/test_base and move them there?

jreback:
looks good. can you rebase

jreback:
looks like you picked up some other commits, do this
```git merge upstream/master`` and push again

**Comment by mariuspot** [comment]
`git merge upstream/master` didn't do anything
Not sure what I did now :/
Hope it's ok now

**Comment by mariuspot** [comment]
Ok, I figured out what went wrong with the merge. Should be ok now

File: doc/source/whatsnew/v0.24.0.txt [hunk 4]
```diff
@@ -659,7 +659,11 @@ Indexing
 - Fixed ``DataFrame[np.nan]`` when columns are non-unique (:issue:`21428`)
 - Bug when indexing :class:`DatetimeIndex` with nanosecond resolution dates and timezones (:issue:`11679`)
 - Bug where indexing with a Numpy array containing negative values would mutate the indexer (:issue:`21867`)
+<<<<<<< HEAD
```
jreback: hve a rebase issue here

jreback (changes requested):
lgtm. looks like some extraneous rebase issue. ping on green.

jreback (approved):

jreback:
lgtm ping on green.

mariuspot:
Looks like travis failed with a segfault on the linker

mariuspot:
@jreback The travis build had a segfault with the linker while setting up the environment. What is the process for that. Is there a way for me to rerun the build or should I make a change and do a new commit?

jreback:
thanks!

Issue #22440, opened 2018-08-21
BUG: IntegerArray cannot cast from int to uint


jorisvandenbossche:
```
In [25]: s = pd.Series([1, 2, 3, 4], dtype='Int64')

In [26]: s
Out[26]: 
0    1
1    2
2    3
3    4
dtype: Int64

In [27]: s.astype('UInt32')
...
TypeError: Cannot cast array from dtype('int64') to dtype('uint32') according to the rule 'same_kind'
```

with numpy dtypes this works:

```
In [30]: s = pd.Series([1, 2, 3, 4], dtype='int64')

In [31]: s.astype('uint32')
Out[31]: 
0    1
1    2
2    3
3    4
dtype: uint32
```

PR #22441, opened 2018-08-21
BUG fix IntegerArray.astype int -> uint

jorisvandenbossche:
Closes https://github.com/pandas-dev/pandas/issues/22440

### DIFF
--- a/pandas/core/arrays/integer.py
+++ b/pandas/core/arrays/integer.py
@@ -409,8 +409,7 @@ class IntegerArray(ExtensionArray, ExtensionOpsMixin):
 
         # if we are astyping to an existing IntegerDtype we can fastpath
         if isinstance(dtype, _IntegerDtype):
-            result = self._data.astype(dtype.numpy_dtype,
-                                       casting='same_kind', copy=False)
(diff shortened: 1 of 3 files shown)


pandas/core/arrays/integer.py
```diff
@@ -409,8 +409,7 @@ def astype(self, dtype, copy=True):
 
         # if we are astyping to an existing IntegerDtype we can fastpath
         if isinstance(dtype, _IntegerDtype):
-            result = self._data.astype(dtype.numpy_dtype,
-                                       casting='same_kind', copy=False)
```
jorisvandenbossche: @jreback do you remember what the reason was to explicitly put `casting='same_kind'` ?

jreback: iirc was to make sure we were
only coercing actual ints (and would raise on floats)

jorisvandenbossche: But since we have the `if isinstance(dtype, _IntegerDtype):`, we are sure about that? (so it can never be a float dtype)

jreback: it might have been an artificat.

jreback:
thanks!

PR #22454, opened 2018-08-21
use fused types for reshape

jbrockmendel:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


topper-123:
I know very little of Cython (or fused), but looked up in the docs, and [saw this warning](http://docs.cython.org/en/latest/src/userguide/fusedtypes.html)

    
```
Note

Support is still somewhat experimental, there may be bugs! 
```

Are we sure this is mature enough?

jbrockmendel:
> Are we sure this is mature enough?

This is a reasonable concern.  The docs don't say what version that warning was added in, but later on that page it suggests that the feature was available in 0.20.0 (... I think) in which case its been around for awhile.

In porting the tempita functions one-by-one I am finding a few places where there might be bugs, but I'm not at all ready to rule out the possibility that the problems are on my end.

Even if we don't go in this direction, hopefully we can produce some good bug reports for cython.  In my experience they've been very responsive/helpful.

gfyoung:
> Maybe consult with the Cython team whether fused types can be considered stable now?

@scoder : Hello, e were planning on using fused types in place of tempita for some of our source code.  However, there's a note in the docs from >5 years ago regarding the instability of fused types.

scoder:
Thanks for asking, I removed that note from the docs.

gfyoung:
@topper-123 : Does that alleviate your concerns?

gfyoung (approved):
LGTM!

cc @jreback

pandas/_libs/reshape.pyx
```diff
@@ -1,15 +1,95 @@
...
+    width : int
+    new_values : typed ndarray
+        result array
+    new_mask : boolean ndarray
+        result mask
+    """
+    cdef:
+        Py_ssize_t i, j, w, nulls, s, offset
+
+    if reshape_t is not object:
```
jreback: what if we make the inner a function instead?

jreback (changes requested):

pandas/_libs/reshape.pyx
jbrockmendel: No dice; we'd need a "nogil" version and a not-nogil version.

jbrockmendel: I think like long-term cython may implement a feature discussed in #22452 that may let us clear this up further.  For now this is much more readable than the existing tempita implementation.

jbrockmendel:
@jreback gentle ping

jreback (approved):

jreback:
ok this is fine (rebase the other fused impls after this merge).

Issue #22459, opened 2018-08-22
DOC: ongoing fixing of doctests


jorisvandenbossche:
Starting with https://github.com/pandas-dev/pandas/pull/19952, we run (some) doctests on Travis to ensure the example are kept up to date. 
Currently, we only test the docstrings for Series and DataFrame, and reshape-related top level functions, *and* we also skip a lot of the tests.

So PRs to incrementally test more of the docstrings are very welcome, and can be good bite-sized newbie issues.

The workflow would be like this:
- Take a certain function or a certain module

- Fix the examples in the docstring(s):
  - To run the doctest for a single method, you can either use the validation script, eg:

    ```
    python scripts/validate_docstrings.py pandas.DataFrame.mean
    ```

    to run a full module or a subset of them, you can use something like:

    ```
    pytest --doctest-modules pandas/core/frame.py -k DataFrame
    ```

  - It is certainly welcome to also fix other problems with the docstring (all other errors and warnings that the validation script gives), or to actually improve the example. But I think it is also fine to do PRs just fixing the doctests as they are, so we at least can test them (so if that then somebody else improves the examples, we are sure those new examples are actually tested).
  
  - Some more details can be found here: http://pandas-docs.github.io/pandas-docs-travis/contributing.html#updating-a-pandas-docstring. And specifically, some tips on getting doctests to pass: http://pandas-docs.github.io/pandas-docs-travis/contributing_docstring.html#tips-for-getting-your-examples-pass-the-doctests
  
- When the doctests pass for a certain function or module, add those (or remove the skip) to the ones that are tested on Travis. 
  For that, you need to edit the `ci/code_checks.sh` file. For example, for the Series tests, it looks like:
  
  https://github.com/pandas-dev/pandas/blob/cf11f71e423be65e9c2ae83ea1bab8b69ee79ca4/ci/code_checks.sh#L125-L126
  
  Here, you could remove one of the skipped ones (the `-k"-agg"` syntax is to skip the doctests for Series.agg and Series.aggregate). Or add such a new line for a new module.
  
  

Good docstrings to start with are the ones for Series and DataFrame (the ones located in `pandas/core/series.py`, `pandas/core/frame.py` and `pandas/core/generic.py`). 
After that, adding new modules like the indexing ones would be a possibility.

Roald87:
[...]

PR #22474, opened 2018-08-23
DOC: Updating Series.agg docstring

alimcmaster1:
With reference to: #22459. I have fixed up the agg/map doc strings for `pandas/core/series.py` so they now pass our doc tests. CC @jorisvandenbossche

Thanks,

### DIFF
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -1132,7 +1132,6 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         Examples
         --------
-
         >>> s = pd.Series([1, 2, 3, 4], name='foo',
         ...               index=pd.Index(['a', 'b', 'c', 'd'], name='idx'))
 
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
Thanks for the PR! 
Just a question, how does this fix the `map` docstring? (I only see changes to the agg docstring? Or that one was already passing?)

pandas/core/series.py
```diff
@@ -3046,21 +3046,21 @@ def _gotitem(self, key, ndim, subset=None):
     Examples
     --------
 
-    >>> s = pd.Series(np.random.randn(10))
+    >>> s = pd.Series(range(1,10,1))
```
datapythonista: As it doesn't make a difference for the example, I'd use `[1, 2, 3, 4]` instead of range (it makes the example easier to understand). In any case, please check pep8 in the examples, there are missing spaces after the commas.

datapythonista: Can you also show `>>> s` after creating it, and remove the blank line between the examples title and the code.

pandas/core/series.py
```diff
@@ -3046,21 +3046,21 @@ def _gotitem(self, key, ndim, subset=None):
...
+    dtype: int64
 
     See also
     --------
-    pandas.Series.apply
-    pandas.Series.transform
-
+    pandas.Series.apply : Invoke function on a Series.
+    pandas.Series.transform : Transform function producing
+     a Series with like indexes.
```
datapythonista: I think this should have the standard 4 spaces indentation.

datapythonista (changes requested):
Thanks for the changes, just couple of styling things.

alimcmaster1:
Thanks both for the comments. Yes @jorisvandenbossche map doc string was already passing. @datapythonista update as per your comments, let me know if there is anything else, thanks

jorisvandenbossche:
The failure on travis is an unrelated build issue (and already happening on master as well), so merging

Thanks a lot @alimcmaster1 !

alimcmaster1:
Thanks both!!

PR #22475, opened 2018-08-23
DOC: Fix DataFrame.to_csv docstring and add an example. GH22459

Moisan:
- [ ] closes #xxxx
- [x] tests added / passed
- [ ] whatsnew entry

Fix the DataFrame.to_csv docstring to match `scripts/validate_docstrings.py` as explained in #22459. I also added an example. Is the whatsnew entry needed for documentation too?


Moisan:
Travis fails due to a curl error.

pandas/core/generic.py
```diff
@@ -9372,68 +9376,80 @@ def to_csv(self, path_or_buf=None, sep=",", na_rep='', float_format=None,
...
             False do not print fields for index names. Use index_label=False
-            for easier importing in R
+            for easier importing in R.
         mode : str
-            Python write mode, default 'w'
+            Python write mode, default 'w'.
         encoding : string, optional
             A string representing the encoding to use in the output file,
             defaults to 'ascii' on Python 2 and 'utf-8' on Python 3.
-        compression : {'infer', 'gzip', 'bz2', 'zip', 'xz', None},
```
WillAyd: Think the domain of potential values should remain here - was this causing some type of validation error?

WillAyd (changes requested):

pandas/core/generic.py
Moisan: In fact the length of the potential values plus the default value span on two lines with the 80 characters limit. This leads to the following error:
`Parameter "compression" description should start with a capital letter`.

pandas/core/generic.py
```diff
@@ -9372,68 +9376,80 @@ def to_csv(self, path_or_buf=None, sep=",", na_rep='', float_format=None,
...
-        .. versionchanged:: 0.24.0
-            The order of arguments for Series was changed.
+        Returns
+        -------
+        If path_or_buf is None, returns the resulting csv format as a string.
+        Otherwise returns None.
+
+        See Also
+        --------
+        pandas.read_csv : load a CSV file into a DataFrame
```
datapythonista: can you also add `DataFrame.to_excel`?

And make the `See Also` description start with capital and finish with period.

datapythonista (changes requested):
Do you mind changing the parameter types, so they are Python types when possible (`str`, `bool`, `int`...) please? In case of list you can use something like `list of str`. The default can stay at the end, but things like that the length of a string should be 1, should go into the description.

pandas/core/generic.py
```diff
@@ -9360,80 +9360,98 @@ def to_csv(self, path_or_buf=None, sep=",", na_rep='', float_format=None,
...
+        decimal : str, default '.'
             Character recognized as decimal separator. E.g. use ',' for
-            European data
+            European data.
 
-        .. versionchanged:: 0.24.0
-            The order of arguments for Series was changed.
+        Returns
+        -------
+        If path_or_buf is None, returns the resulting csv format as a string.
```
datapythonista: Can you add a line first `None or str` and then this description in the next line indented

pandas/core/generic.py
```diff
@@ -9360,80 +9360,98 @@ def to_csv(self, path_or_buf=None, sep=",", na_rep='', float_format=None,
...
+
+        See Also
+        --------
+        pandas.read_csv : Load a CSV file into a DataFrame.
+        pandas.to_excel: Load an Excel file into a DataFrame.
+
+        Examples
+        --------
+        >>> df = pd.DataFrame({'name': ['Raphael', 'Donatello'],
+        ... 'mask': ['red', 'purple'], 'weapon': ['sai', 'bo staff']})
```
datapythonista: can you indent `mask` at the level of `name` and `weapon` in the next line with the same indentation

datapythonista (commented):
Looks great, just two small things.

datapythonista (approved):
Looks great, thanks for fixing the docsting @Moisan

Moisan:
circleci `py27_compat` fails after a `libgfortran` import.

datapythonista:
@WillAyd please merge when you're happy with this

WillAyd (approved):

WillAyd:
Thanks @Moisan

jorisvandenbossche:
BTW, sphinx requires a blank line *before* a `versionadded`/`versionchanged`/`deprecated` directive (unless it is the first thing in a parameter description), which was removed for some here. Did a quick fix: https://github.com/pandas-dev/pandas/pull/22816

PR #22479, opened 2018-08-23
BUG: fix Series(extension array) + extension array values addition

### LINKED ISSUE #22478: BUG: Series(extension array) + extension array values addition cause runtime exception

```python
In [41]: s
Out[41]:
0    1
1    2
2    3
dtype: Int64

In [42]: s.values
Out[42]: IntegerArray([1, 2, 3], dtype='Int64')

In [43]: s + s.values
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-43-c3f015376225> in <module>()
----> 1 s + s.values

C:\Users\Public\pandas-peter\pandas\core\ops.py in wrapper(left, right)
   1231                 (is_extension_array_dtype(right) and not is_scalar(right
))):
   1232             # GH#22378 disallow scalar to exclude e.g. "category", "Int6
4"
-> 1233             return dispatch_to_extension_op(op, left, right)
   1234
   1235         elif is_datetime64_dtype(left) or is_datetime64tz_dtype(left):

C:\Users\Public\pandas-peter\pandas\core\ops.py in dispatch_to_extension_op(op,
left, right)
   1152             new_right = list(new_right)
   1153         elif is_extension_array_dtype(right) and type(left) != type(righ
t):
-> 1154             new_right = list(new_right)
   1155         else:
   1156             new_right = right

UnboundLocalError: local variable 'new_right' referenced before assignment

```
#### Problem description
Extension array add itself cause problems

#### Expected Output
In [2]: s + s.values
Out[2]:
0    2
1    4
2    6
dtype: Int64

<details>
INSTALLED VERSIONS
------------------
commit: 2be2ba570453ab09f0c8da46dfd8e1a179771fee
python: 3.6.4.final.0
python-bits: 64
OS: Windows
OS-release: 7
machine: AMD64
byteorder: little
LC_ALL: None
LANG: zh_CN.UTF-8
LOCALE: None.None

pandas: 0.24.0.dev0+506.g2be2ba570
pytest: 3.7.1
pip: 9.0.1
setuptools: 38.4.0
Cython: 0.28.4
numpy: 1.14.0
scipy: 1.0.0
pyarrow: None
xarray: None
IPython: 6.2.1
sphinx: 1.6.7
patsy: 0.5.0
dateutil: 2.6.1
pytz: 2017.3
blosc: None
bottleneck: 1.2.1
tables: 3.4.2
numexpr: 2.6.4
feather: None
matplotlib: 2.1.2
openpyxl: 2.4.10
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.2
lxml: 4.1.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.1
pymysql: 0.7.11.None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
gcsfs: None

</details>

peterpanmj:
- [x] closes #22478
- [x] tests added / passed
- [x] whatsnew entry


peterpanmj:
It is a simple fix, but not sure where to put additional tests.

jorisvandenbossche:
> It is a simple fix, but not sure where to put additional tests.

I would look in `pandas/tests/extension/base/ops.py`, there is a `test_arith_series_with_array` that tests Series + Series, you could add in that test to do a `Series + array` (with a similar `self.check_opname` call). But make sure you also test that it actually fails without your fix.

peterpanmj:
pandas/tests/extension/base/ops.py Can not be discovered by pytest. How to test it ? @jorisvandenbossche

jorisvandenbossche:
@peterpanmj the actual tests are in one directory up (they inherit from those base classes), so if you do `pytest pandas/tests/extension`, they should all be run.

peterpanmj:
I will compose a test shortly @jreback

pandas/tests/arrays/test_integer.py
```diff
@@ -587,6 +587,13 @@ def test_cross_type_arithmetic():
     tm.assert_series_equal(result, expected)
 
 
+def test_arith_extension_array_values():
+    s = pd.Series([1, 2, 3], dtype='Int64')
```
jreback: can you add the issue as a comment

jreback (changes requested):
can you add this issue number onto the IntegerArray whatsnew section

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -493,6 +493,7 @@ ExtensionType Changes
 - :meth:`Series.combine()` with scalar argument now works for any function type (:issue:`21248`)
 - :meth:`Series.astype` and :meth:`DataFrame.astype` now dispatch to :meth:`ExtensionArray.astype` (:issue:`21185:`).
 - Added :meth:`pandas.api.types.register_extension_dtype` to register an extension type with pandas (:issue:`22664`)
```
TomAugspurger: W can remove this I think... integerArray is new in 0.24

TomAugspurger (approved):
@jreback is there an issue for not listifying right? That seems unnecessary  when new_left and right are both EAs of the same type.

jreback:
u can add this issue number to the integer_array whatsnew

TomAugspurger:
Shouldn’t be necessary since there isn’t a released version with this bug.

jreback:
that’s not the point
it provides a reference to all issues with this feature

just add the number (on the existing note)

TomAugspurger:
Sorry I don't see the need. If this bug hadn't been there when you wrote IntegerArray, would you have made a note saying that ops between Series[IntegerArray] and IntegeryArray works? I don't see why we should distract readers going through the release notes with the fact that this was implemented across two PRs.

jreback:
i think u have missed my point

we list all issues that are relevent if possible . all u do is add it to the list of issues. no extra note is necessary. 

the problem with ignoring it is if you only look at the listed issues you will miss things. it’s irrelevant whether it’s a bug or additional features. it’s all a part of the single what’s new note. 

this has been practice for many years

this way you have a complete record of what has changed.

peterpanmj:
How about we leave the bug there in 0.24 . Fix it later in 0.24.1 and update the whatsnew then.

pandas/tests/arrays/test_integer.py
```diff
@@ -587,6 +587,14 @@ def test_cross_type_arithmetic():
     tm.assert_series_equal(result, expected)
 
 
+def test_arith_extension_array_values():
```
jorisvandenbossche: Can you move the test to `pandas/tests/extension/base/ops.py`, as this is not a specific failure for IntegerArray, it is relevant for all extension types.

jorisvandenbossche (dismissed):

jorisvandenbossche:
> How about we leave the bug there in 0.24 . Fix it later in 0.24.1 and update the whatsnew then.

No, we certainly want to get this fixed in master / in the first coming release.

---

I agree with Tom, this is something that only failed in master, for a new feature that doesn't exist in 0.23. So no whatsnew note is needed, and adding the issue number there only adds noise IMO. If it would have been another issue that already existed before that was also resolved by the same fix, yes, then it would be good to list the issue number in the whatsnew.

(but let's further discuss this on gitter / mail to not bother this PR?)

---

> @jreback is there an issue for not listifying right? That seems unnecessary when new_left and right are both EAs of the same type.

I didn't find one. Opened https://github.com/pandas-dev/pandas/issues/22922

pandas/tests/extension/test_integer.py
```diff
@@ -143,6 +143,12 @@ def test_error(self, data, all_arithmetic_operators):
         # other specific errors tested in the integer array specific tests
         pass
 
+    @pytest.mark.xfail(reason="EA is listified. GH-22922", strict=True)
```
TomAugspurger: So, here's what's happening for IntegerArray with missing values,

Because of #22922, the ExtensionArray is listified, so we do `IntegerArray + List`. Then, in `IntegerArray.__add__` we do `np.asarray(other)`. Since we have missing values, we naturally end up with a float ndarray.

Fixing #22922 will fix this specific issue, but doesn't handle the general problem of `IntegerArray + List`. i.e. is this a bug?

```python
In [3]: arr = pd.Series([1, None, 3], dtype='Int8').values

In [4]: arr
Out[4]: IntegerArray([1, nan, 3], dtype='Int8')

In [5]: arr + list(arr)
Out[5]: array([ 2., nan,  6.])
```

Should `Out[5]` be an `IntegerArray([2, nan, 6], dtype='Int8')`? If so, I'll open a new issue.

TomAugspurger:
Note that this is still tested via decimal.

pandas/tests/extension/test_integer.py
jreback: yes i think you are right

not sure this is hitting the right path

TomAugspurger (approved):
No new CI failures (just s3 and numpydev, which I'm hoping to resolve tomorrow).

pandas/tests/extension/test_integer.py
TomAugspurger: https://github.com/pandas-dev/pandas/issues/22962

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @peterpanmj and @TomAugspurger !

jreback:
this was not fully reviewed

need tests for the reverse op as well, eg

data + series

ideally we would test using the fixture for all ops as well

TomAugspurger:
@jreback can you open a new issue with specifics?

> need tests for the reverse op as well, eg

That hits a different code path, so would be a different bug if it doesn't work (what's the expected result? Right now IntegerArray + Series[IntegerArray] returns an IntegerArray)

> ideally we would test using the fixture for all ops as well

I meant to include a note with my push, but was distracted by the float array issue. In this case, I opted for a simpler test that didn't include all the indirection of `_check_op_name`.  This test was about covering our dispatching, so just the add is sufficient.

jorisvandenbossche:
> what's the expected result? Right now IntegerArray + Series[IntegerArray] returns an IntegerArray

That should be a Series instead IMO (that's what we do with Series/array as well, and Series is "higher in the hierarchy")

TomAugspurger:
https://github.com/pandas-dev/pandas/issues/22974

Issue #22471, opened 2018-08-22
TST/CLN: remove TestData from frame-tests; replace with fixtures


h-vetinari:
Following review in #22236: 
> ok, pls open a new issue that refs this, to remove use of `TestData` in favor of fixtures

Started the process in that PR by creating a `conftest.py` that translates all the current attributes of `TestData` to fixtures. Need to incrementally replace their usages in `pandas/tests/frame/`.

h-vetinari:
@gfyoung 
I think this issue and #22550 should have a "good first issue" and potentially "difficulty: novice" tag. What do you think?

gfyoung:
@h-vetinari : Seems reasonable to me.  Done.

gfyoung:
@akulagrawal : Go for it!

SaturnFromTitan:
Seems like the following files don't contain TestData (anymore) and can be checked or removed from the issue description:
- test_constructors.py
- test_dtypes.py
- test_period.py
- test_quantile.py
- test_rank.py
- test_sort_values_level_as_str.py
- test_timezones.py
- test_validate.py

I'll take a look at the others

PR #22486, opened 2018-08-23
API: better error-handling for df.set_index

### LINKED ISSUE #22484: API: better error-handling for df.set_index
splitting up #22236.

Let's have
```df = pd.DataFrame(np.random.randn(5, 5), columns=list('ABCDE'))```

The error handling of `df.set_index` can be improved in at least three cases:
1. ```df.set_index(['A', 'A'], drop=False)``` works, while
  ```df.set_index(['A', 'A'], drop=True)``` yields
  ```KeyError: 'A'```
1. Objects of unknown type yield `KeyError` instead of `TypeError`:
  ```df.set_index(map(str, df.A))```
  ```KeyError: "None of [Index([...], dtype='object')] are in the [columns]"```
1. ```df.set_index(['foo', 'bar', 'baz'])``` only shows one missing key
   ```KeyError: 'foo'``` (in a huge stacktrace)

Better would be:
1. gracefully handle duplicate column names when `drop=True`
1. raise better error message, e.g. `TypeError: only allowed types are: ...`
1. Show *all* missing keys: ```KeyError: "['foo', 'bar', 'baz']"```

### PR DESCRIPTION
**Pull request by h-vetinari**
- [x] closes #22484
- [x] tests added / passed
- [x] whatsnew entry

split off of #22236, and so builds on top of it.

### CODE REVIEW

File: pandas/core/frame.py [hunk 1]
```diff
@@ -3862,10 +3863,27 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
+        # raise error for elements of keys that are neither valid column keys
+        # or an array of allowed type
+        if any(not (is_scalar(x) or isinstance(x, tuple)) and
+               not isinstance(x, (ABCSeries, ABCIndexClass, ABCMultiIndex,
+                                  list, np.ndarray)) for x in keys):
+            raise TypeError('keys may only contain a combination of the '
+                            'following: valid column keys, Series, Index, '
+                            'MultiIndex, list or np.ndarray')
+
+        inplace = validate_bool_kwarg(inplace, 'inplace')
```
gfyoung: Leave the `inplace` invalidation where it was.

pandas/core/frame.py
```diff
@@ -3862,10 +3863,27 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
+            raise KeyError('{}'.format(missing))
+
+        # raise error for elements of keys that are neither valid column keys
+        # or an array of allowed type
+        if any(not (is_scalar(x) or isinstance(x, tuple)) and
+               not isinstance(x, (ABCSeries, ABCIndexClass, ABCMultiIndex,
+                                  list, np.ndarray)) for x in keys):
+            raise TypeError('keys may only contain a combination of the '
+                            'following: valid column keys, Series, Index, '
+                            'MultiIndex, list or np.ndarray')
```
gfyoung: This looks like a lot of iterations over the `keys` (iterating over `col_labels` is equivalent time-complexity-wise).  Could we have a single for-loop where we iterate over `keys` and fail fast?

pandas/tests/frame/conftest.py
```diff
@@ -0,0 +1,121 @@
+import pytest
```
gfyoung: A couple of things about this file:

* Do you use all of these fixtures in your changes?
* I would prefer if the naming is a little more consistent e.g.:
- You have the word "frame" is some of your fixture names but not othres
- You have underscores between words in some names but not others

h-vetinari:
@gfyoung, thanks for the review. This is based on #22236, and so it will be easier to review once that is in.

About the number of times iterating over the columns, it would IMO be very complicated and unreadable to do it in one loop, and - to me - not worth the performance gain, because the number of columns used for an index is generally so negligibly small, that traversing that list is a non-issue performance-wise.

pandas/tests/frame/conftest.py
gfyoung: > This is based on #22236, and so it will be easier to review once that is in.

I'm not sure I follow you here.  How does adding currently unused fixtures in this PR get explained in #22236?  I'm sensing you might want to break some of the fixtures for a separate PR.

h-vetinari: It's buried in the review there and related to #22471 - replacing the attributes of `TestData` with fixtures. I did it all in one go as a start towards #22471. It's a direct translation of the attributes, including their names.

pandas/core/frame.py
h-vetinari: Not bad, but doesn't handle the duplicate column keys yet.

pandas/tests/frame/conftest.py
gfyoung: Hmm...I'm sorry, but I think this is a little too much for one PR.  Keep in mind that the purpose of the PR, per your title, is better error reporting for `df.set_index`.  This is a lot of new code that is arguably not pertinent to that aim.

pandas/core/frame.py
gfyoung: I don't quite see the difference between what you wrote, and what I did.  Where do you handle duplicate column keys in your changes above?

h-vetinari: Nevermind. Forgot that this case had been removed now by this PR

pandas/tests/frame/conftest.py
h-vetinari: The diff of this PR will be tiny after #22236. Feel free to comment there about the changes of that PR

pandas/core/frame.py
gfyoung: In that case, unless there are other objections, I would suggest using the single for-loop instead.

pandas/tests/frame/conftest.py
gfyoung: FYI, I literally have the same questions in that PR as I do in this one for that file.

h-vetinari: I'll answer in more detail in the other thread. But I already mentioned what it's about:

> It's [...] related to #22471 - replacing the attributes of `TestData` with fixtures. I did it all in one go as a start towards #22471. It's a direct translation of the attributes, including their names.

h-vetinari:
@gfyoung 
> In that case, unless there are other objections, I would suggest using the single for-loop instead. It seems pretty readable IMO.

Incorporated your feedback, coming with the next commit after rebasing on top of #22236

pandas/core/frame.py
```diff
@@ -3901,37 +3918,34 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
                 names.extend(col.names)
-            elif isinstance(col, Series):
-                level = col._values
+            elif isinstance(col, ABCIndexClass):
+                # Index but not MultiIndex (treated above)
+                arrays.append(col)
                 names.append(col.name)
-            elif isinstance(col, Index):
-                level = col
+            elif isinstance(col, ABCSeries):
```
h-vetinari: The ABC forms are a left-over from the review of #22236 by @jreback (https://github.com/pandas-dev/pandas/pull/22236#discussion_r212255994):
> huh? pls use the ABC version, we do this everywhere else.

h-vetinari (commented):
Now that #22236 is in, we can continue here. :)

pandas/tests/frame/test_alter_axes.py
```diff
@@ -186,18 +186,19 @@ def test_set_index_pass_arrays_duplicate(self, frame_of_index_cols, drop,
 
         # == gives ambiguous Boolean for Series
         if drop and keys[0] is 'A' and keys[1] is 'A':
-            with tm.assert_raises_regex(KeyError, '.*'):
-                df.set_index(keys, drop=drop, append=append)
+            # can't drop same column twice
+            first_drop = False
```
jreback: why are you changing these?

these tests are not very hard to interpret

pandas/tests/frame/test_alter_axes.py
```diff
@@ -229,13 +230,24 @@ def test_set_index_verify_integrity(self, frame_of_index_cols):
...
 
         # non-existent key in list with arrays
-        with tm.assert_raises_regex(KeyError, '.*'):
+        with tm.assert_raises_regex(KeyError, 'X'):
             df.set_index([df['A'], df['B'], 'X'], drop=drop, append=append)
 
+        rgx = 'keys may only contain a combination of the following:.*'
+        # forbidden type, e.g. set
+        with tm.assert_raises_regex(TypeError, rgx):
+            df.set_index(set(df['A']), drop=drop, append=append)
```
jreback: why are you singling out an iterable (set) is there some reason?

pandas/core/frame.py
```diff
@@ -3892,6 +3893,22 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
         if not isinstance(keys, list):
             keys = [keys]
 
+        missing = []
+        for x in keys:
+            if not (is_scalar(x) or isinstance(x, tuple)):
```
jreback: why this overly complicated condition? can u show a simple example which fails now?

jreback (changes requested):

pandas/core/frame.py
h-vetinari: This came out of the 2./3. outlined in #22484 and then the feedback from @gfyoung further up the thread

pandas/tests/frame/test_alter_axes.py
h-vetinari: This changes because now we're not failing for `drop=True` and `keys=['A', 'A']` (for example), see #22484

pandas/tests/frame/test_alter_axes.py
h-vetinari: No, just any of the types that are not allowed (I've tried to indicate as much by using "e.g.")

pandas/core/frame.py
```diff
@@ -3892,6 +3893,22 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
+                if not isinstance(x, (ABCSeries, ABCIndexClass, ABCMultiIndex,
+                                      list, np.ndarray)):
+                    raise TypeError('keys may only contain a combination of '
```
jreback: this is not true. either all have to be scalars, or all list-likes (yes they don't have to be the same, just list-convertible). 

Please make this more readable from a user perspective. If I got this I would have no idea what to do.

pandas/tests/frame/test_alter_axes.py
jreback: this tests is very confusing. pls make it simpler.

jreback (changes requested):
ok, theme is reasonable, but need to simplify the code to avoid special casing things

h-vetinari:
> this is not true. either all have to be scalars, or all list-likes (yes they don't have to be the same, just list-convertible).

This is wrong on both counts. On master, a mix of column labels and a list:

```
>>> pd.DataFrame({'a': [1, 2], 'b': [11, 12]})
   a   b
0  1  11
1  2  12
>>> df = pd.DataFrame({'a': [1, 2], 'b': [11, 12]})
>>> df.set_index(['a', [101, 102]])
        b
a
1 101  11
2 102  12
```

Regarding the argument types, the list I wrote (and test) is exhaustive (see https://github.com/pandas-dev/pandas/blob/master/pandas/core/frame.py#L3908) - currently only `Series/Index/MultiIndex/list/ndarray` and everything else gets tried as a frame key. This is one of the points in #22484.
```
>>> from pandas.core.dtypes.common import is_list_like
>>> LL = iter([21, 22])
>>> is_list_like(LL)
True
>>> df.set_index(LL)
KeyError: "None of [Int64Index([21, 22], dtype='int64')] are in the [columns]"
```

pandas/core/frame.py
h-vetinari: See longer comment in the main thread.

Tuples are allowed because they might be valid column labels.

pandas/tests/frame/test_alter_axes.py
h-vetinari: The diff might be confusing, but it doesn't look so bad in practice: https://github.com/pandas-dev/pandas/blob/e861dcdafb0b2674f6be311b611cf5988c17b1e5/pandas/tests/frame/test_alter_axes.py#L187

It doesn't get much simpler - to test against previously tested behaviour, I have to do the `set_index`-calls sequentially, but it's not possible to drop the same key twice (which this test is changing).

h-vetinari:
@jreback 
Green

h-vetinari:
@jreback 
While you're at it with the reviewing, please don't forget this one. :)

pandas/core/frame.py
```diff
@@ -3891,6 +3892,22 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
         if not isinstance(keys, list):
             keys = [keys]
 
+        missing = []
+        for x in keys:
+            if not (is_scalar(x) or isinstance(x, tuple)):
+                if not isinstance(x, (ABCSeries, ABCIndexClass, ABCMultiIndex,
```
jreback: you can use is_list_like here

pandas/tests/frame/test_alter_axes.py
```diff
@@ -229,13 +226,24 @@ def test_set_index_verify_integrity(self, frame_of_index_cols):
...
+        with tm.assert_raises_regex(KeyError, "['foo', 'bar', 'baz']"):
+            # column names are A-E
             df.set_index(['foo', 'bar', 'baz'], drop=drop, append=append)
 
         # non-existent key in list with arrays
-        with tm.assert_raises_regex(KeyError, '.*'):
+        with tm.assert_raises_regex(KeyError, 'X'):
             df.set_index([df['A'], df['B'], 'X'], drop=drop, append=append)
 
+        rgx = 'The parameter "keys" may only contain a combination of.*'
```
jreback: rgx -> msg

jreback (changes requested):

pandas/core/frame.py
h-vetinari: This would also change the capabilities of `set_index`. e.g. an iterator has `is_list_like = True` - do you want to support that? Currently, only the list that I'm instance-checking (plus column keys) are allowed.

pandas/tests/frame/test_alter_axes.py
h-vetinari: @jreback Changed. Please let me know what you want me to do about the allowed argument signature. Allowing all list-likes also means checking `ndim==1` for `np.ndarrays`, etc.

See also:
https://github.com/pandas-dev/pandas/pull/22486#discussion_r219705345
https://github.com/pandas-dev/pandas/pull/22486#discussion_r219705993

pandas/core/frame.py
jreback: I don't want to explicity list types here as its verbose and fragile, use is_list_like and not is_iterator if you must (though to be honest an iterator is actually ok)

jreback (changes requested):

pandas/core/frame.py
h-vetinari: > I don't want to explicity list types here as its verbose and fragile

I have a feeling the implementation you request will be much more verbose, but OK

> OK. How to deal with tuples then. Can be valid column keys and valid list-likes.

You want to try tuples as column keys first and *then* as an array (would make the most sense to me), or something else?

h-vetinari:
@jreback 

I implemented the list-likes, like you requested. I'm excluding two-dimensional data and sets (didn't know they are even list-like, as they have no determined order?!). I also added a column with a tuple label to the fixture in conftest to test that case.

BTW, the test results are not as grim as they look, there were 2x `CondaHTTPError: HTTP 504 GATEWAY_TIMEOUT` in CircleCI, and one doctest-failure in a file I didn't touch.

h-vetinari:
@jreback 

PTAL

pandas/core/frame.py
```diff
@@ -3951,6 +3953,24 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
 
+        missing = []
+        for x in keys:
+            if (is_scalar(x) or isinstance(x, tuple)) and x in self:
+                continue
+            elif is_scalar(x) and x not in self:
+                # a tuple gets tried as column key first;
+                # otherwise considered as a list-like; i.e. not missing
+                missing.append(x)
+            elif (not is_list_like(x) or isinstance(x, set)
```
jreback: shouldn't ``not is_like(x) or getattr(x, 'ndim', 1) > 1`` work here

pandas/core/frame.py
```diff
@@ -3951,6 +3953,24 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
         if not isinstance(keys, list):
             keys = [keys]
 
+        missing = []
+        for x in keys:
+            if (is_scalar(x) or isinstance(x, tuple)) and x in self:
```
jreback: add a comment here

pandas/core/frame.py
```diff
@@ -3951,6 +3953,24 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
         if not isinstance(keys, list):
             keys = [keys]
 
+        missing = []
+        for x in keys:
```
jreback: can you rename x -> col for consistency with the loop below

jreback (changes requested):

pandas/core/frame.py
```diff
@@ -3952,6 +3953,25 @@ def set_index(self, keys, drop=True, append=False, inplace=False,
...
+        for col in keys:
+            if (is_scalar(col) or isinstance(col, tuple)) and col in self:
+                # tuples can be both column keys or list-likes
+                # if they are valid column keys, everything is fine
+                continue
+            elif is_scalar(col) and col not in self:
+                # tuples that are not column keys are considered list-like,
+                # not considered missing
+                missing.append(col)
+            elif (not is_list_like(col) or isinstance(col, set)
```
jreback: is this isinstance check for set necessary?

pandas/tests/frame/test_alter_axes.py
```diff
@@ -183,21 +198,20 @@ def test_set_index_pass_arrays_duplicate(self, frame_of_index_cols, drop,
...
-            expected = expected.set_index([keys[1]], drop=drop, append=True)
+        # if either box was iter, the content has been consumed; re-read it
+        keys = [box1(df['A']), box2(df['A'])]
 
-            tm.assert_frame_equal(result, expected)
+        # to test against already-tested behaviour, we add sequentially,
+        # hence second append always True; must wrap keys in list, otherwise
+        # box = list would be illegal; need to adapt drop for case that both
+        # keys are 'A' -- can't drop the same column twice
+        expected = df.set_index([keys[0]], drop=(False if (keys[0] is 'A'
```
jreback: can you make the drop a separate statement, e.g.

```
drop=(False.....)
expected = df.set_index(....
```

pandas/tests/frame/test_alter_axes.py
```diff
@@ -229,13 +243,24 @@ def test_set_index_verify_integrity(self, frame_of_index_cols):
...
             df.set_index([df['A'], df['B'], 'X'], drop=drop, append=append)
 
+        msg = 'The parameter "keys" may only contain a combination of.*'
+        # forbidden type, e.g. set
+        with tm.assert_raises_regex(TypeError, msg):
+            df.set_index(set(df['A']), drop=drop, append=append)
+
+        # forbidden type in list, e.g. set
+        with tm.assert_raises_regex(TypeError, msg):
+            df.set_index(['A', df['A'], set(df['A'])],
```
jreback: see my comment above, I am not sure a set is specifically excluded.

jreback (changes requested):

pandas/core/frame.py
h-vetinari: I didn't know sets are considered list-like. IMO they shouldn't be, because they have no defined order (which is important for an index, especially when `append=True`). Do we really want to support this option of letting users shoot themselves in the foot?
```
>>> s = pd.Series([1, 2, 3])
>>> s.set_index({'a', 'c', 'e'}, append=True)
0  c    1
1  e    2
2  a    3
dtype: int64
```

pandas/tests/frame/test_alter_axes.py
h-vetinari: That's what I had before (with `first_drop`). I can't overwrite `drop` because it's needed for the second `set_index`.

pandas/tests/frame/test_alter_axes.py
h-vetinari: I am proposing to exclude sets (see above). Irrespective of that, I need to test raising the TypeError for forbidden types (which type exactly is less relevant)

h-vetinari (commented):
Thanks for the review.

pandas/core/frame.py
jreback: i am not averse to excluding, but i also don't want ad-infinitem special cases here. This code is already too complex. We accept an iterable, so this is a valid input, if the user wants to do it that's there issue; these are not currently excluded.

pandas/tests/frame/test_alter_axes.py
```diff
@@ -126,21 +131,29 @@ def test_set_index_pass_single_array(self, frame_of_index_cols,
         df.index.name = index_name
 
         key = box(df['B'])
-        # np.array and list "forget" the name of B
-        name = [None if box in [np.array, list] else 'B']
+        if box == list:
```
jreback: why exactly is a list interpreted differently here? this makes this test really really odd. I am worried something changed here, as this appears to work just fine in the master.

jreback (changes requested):

pandas/core/frame.py
h-vetinari: The larger issue IMO is that `is_list_like(<set>)` should be False and not True. It's the only exclusion here, and it makes sense. On many issues, pandas prides itself on enforcing sensible defaults and behaviour. Why would we want to enable something that's broken by design?

I may also add that **this is not something that's currently allowed**, and it shouldn't be IMO

pandas/tests/frame/test_alter_axes.py
h-vetinari: You will see above (line 123)  that `list` wasn't tested - for precisely this reason. `df.set_index(['A', 'B'])` interprets a `A` & `B` as column keys, so (assuming this `df` was length 2) it would not use the list `['A', 'B']` as the index. To do that, one would have to pass `[['A', 'B']]`. This PR proposes to add tests for the current behaviour.

h-vetinari (commented):
Thanks for review; added responses

pandas/tests/frame/test_alter_axes.py
h-vetinari: @jreback 
In case the above was not very clearly worded, this corresponds exactly to behaviour on `master`:
```
>>> df = pd.DataFrame(np.random.randn(2,3))
>>> df.set_index(['A', 'B'])
KeyError: 'A'
>>> df.set_index([['A', 'B']])
          0         1         2
A  1.962370 -1.150770  0.843600
B -0.417274  0.509781 -0.697802
```

jreback:
@h-vetinari why don't you try (separate PR) excluding set from is_list_like and see what the implications of that are.

h-vetinari:
@jreback 
> @h-vetinari why don't you try (separate PR) excluding set from is_list_like and see what the implications of that are.

I did, seems doable: #23061 #23065

h-vetinari:
@jreback 
After merging and incorporating #23065, all comments here should now be addressed.

jreback (approved):

jreback:
thanks!

PR #22457, opened 2018-08-22
BUG: Retain timezone information in to_datetime if box=False

mroeschke:
- [x] closes #22446
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

Don't think this requires a whatsnew entry since this was introduced in (and part of) #21822 which is slated for v0.24.0

PR #22494, opened 2018-08-24
CLN: Simplify read_csv tz offset parsing

mroeschke:
I _think_ after  #22457  we do not need the `np.asarray` calls that were introduced in #22380


gfyoung:
If tests pass, I'm +1 for this.

mroeschke:
Looks like tests are passing besides a travis build termination.

jreback:
thanks!

PR #22499, opened 2018-08-24
CI: Bump to NumPy compat to 1.9.3

TomAugspurger:
1.9.2 doesn't seem to be available in `/pkgs/main`. We're seeing errors like

```
ImportError: libgfortran.so.1: cannot open shared object file: No such file or directory
```

when importing numpy.

Switching to 1.9.3 (which is available in main) solves that for me locally.


TomAugspurger:
I'm not sure why appveyor wasn't run, but in theory it's unaffected by this.

Merging to get master passing again.

jorisvandenbossche:
Thanks!

PR #22510, opened 2018-08-26
DOC: Update series apply docstring. GH22459

eldritchideen:
- [X] refs #22459
- [x] tests added / passed
- [ ] whatsnew entry

Updated `Series.apply` docstring to resolve errors raised from scripts/validate_docstrings.py from #22459.


pandas/core/series.py
```diff
@@ -3104,17 +3104,19 @@ def apply(self, func, convert_dtype=True, args=(), **kwds):
...
-        convert_dtype : boolean, default True
+            Python function or NumPy ufunc.
+        convert_dtype : bool, default True
             Try to find better dtype for elementwise function results. If
-            False, leave as dtype=object
+            False, leave as dtype=object.
         args : tuple
-            Positional arguments to pass to function in addition to the value
-        Additional keyword arguments will be passed as keywords to the function
+            Positional arguments to pass to func in addition to the value.
```
gfyoung: "in addition to the value" - can we clarify what that means?

pandas/core/series.py
```diff
@@ -3104,17 +3104,19 @@ def apply(self, func, convert_dtype=True, args=(), **kwds):
...
+        **kwds : dict
```
gfyoung: @datapythonista : I wonder if we should standardize and use `**kwargs` everywhere.

datapythonista: In terms of documentation, the standard we defined was to always document `*args` and `**kwargs` if present, showing the stars in the name, but without the type (it'll always be tuple and dict).

Not sure if you mean using `kwargs` instead of `kwds`, or which standardization are you proposing here, but I'm +1 on being consistent everywhere.

I see that in this case `args` needs to be provided as a tuple (not as positional arguments) and `kwds` needs to be provided as keyword arguments (not as a dict). Not sure if there is a reason for this, but if it's not, I'd be in favor of changing it to be consistent (in a separate PR). @jreback ?

datapythonista (commented):
Changes look good to me, after considering @gfyoung comments, and removing the type (dict) of `**kwds`. Thanks @eldritchideen

pandas/core/series.py
gfyoung: @datapythonista : Indeed, I was referring to using `kwargs` across the board.  I'm also in favor of consistency in general.

eldritchideen:
@datapythonista I removed the type from the **kwds parameter. I also agree that it should be renamed to **kwargs to be more consistent with the rest of the code base.

datapythonista (commented):
Thanks for the changes @eldritchideen . Can you add the clarification to the `args` parameter @gfyoung asked? I don't think it's clear enough either.

Also, it'd be great if you can change couple of few things to make the docstring follow all our guidelines:
- Remove the colon after `**kwds`
- In the `Returns` section, remove the `y :` and leave just `Series or DataFrame` adding the description (and clarifications about the type) to the next line (indented)
- Add a period at the end of each `See Also` item (and capitalize the `A` in `Also` in the title)
- Remove the blank line after the `Examples` title.
- Rename the variable `series` by `s` in the examples.

Thank you!

eldritchideen:
I have updated the Series.apply docstring based on review feedback. Thanks @datapythonista and @gfyoung

pandas/core/series.py
```diff
@@ -3104,36 +3104,38 @@ def apply(self, func, convert_dtype=True, args=(), **kwds):
...
         Invoke function on values of Series. Can be ufunc (a NumPy function
         that applies to the entire Series) or a Python function that only works
-        on single values
+        on single values.
 
         Parameters
         ----------
         func : function
-        convert_dtype : boolean, default True
+            Python function or NumPy ufunc.
```
datapythonista: may be we can say "Python function or NumPy ufunc **to apply**."?

pandas/core/series.py
```diff
@@ -3104,36 +3104,38 @@ def apply(self, func, convert_dtype=True, args=(), **kwds):
...
 
-        See also
+        See Also
         --------
-        Series.map: For element-wise operations
-        Series.agg: only perform aggregating type operations
-        Series.transform: only perform transforming type operations
+        Series.map: For element-wise operations.
+        Series.agg: only perform aggregating type operations.
+        Series.transform: only perform transforming type operations.
```
datapythonista: Can you capitalize all the first letters of the descriptions

pandas/core/series.py
```diff
@@ -3144,7 +3146,7 @@ def apply(self, func, convert_dtype=True, args=(), **kwds):
 
         >>> def square(x):
         ...     return x**2
```
datapythonista: Can you fix the PEP-8 of all examples please? There are missing spaces around all the `**` , one `-` and one `+=`. I know it's from the original code, but let's get it fixed.

pandas/core/series.py
```diff
@@ -3104,36 +3104,38 @@ def apply(self, func, convert_dtype=True, args=(), **kwds):
...
+        Series.transform: only perform transforming type operations.
 
         Examples
         --------
-
         Create a series with typical summer temperatures for each city.
 
-        >>> series = pd.Series([20, 21, 12], index=['London',
+        >>> s = pd.Series([20, 21, 12], index=['London',
         ... 'New York','Helsinki'])
```
datapythonista: I think it's more readable if we have all `index` in the next line.

pandas/core/series.py
```diff
@@ -3104,36 +3104,38 @@ def apply(self, func, convert_dtype=True, args=(), **kwds):
         """
         Invoke function on values of Series. Can be ufunc (a NumPy function
         that applies to the entire Series) or a Python function that only works
-        on single values
+        on single values.
```
datapythonista: Can you have a single line description first, and the rest later. You have the documentation about this here:
https://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html

If you run `./scripts/validate_docstrings.py pandas.Series.apply` should report an error about it.

datapythonista (commented):
Looking better, there are some more changes that we can do (mainly to the original code) to make it look perfect.

jorisvandenbossche:
@eldritchideen do you have time to update this PR based on the comments?

WillAyd:
@datapythonista I see the self assignment on this - are you actually working this one?

datapythonista:
I forgot about it, but fixed the pending issues now. Can you take a look and merge if everything is ok (or fix any outstanding issue, as the PR is discontinued)?

WillAyd (approved):
Thanks @datapythonista will merge on green

datapythonista:
thanks @eldritchideen

PR #22516, opened 2018-08-26
DOC: Fix to_latex docstring.

Moisan:
- [x] tests added / passed

Fix the DataFrame.to_latex docstring to match `scripts/validate_docstrings.py` as explained in #22459  and add an example.

The docstring was previously in a variable that was only used in to_latex. I put it in the method docstring instead. The `@Substitution` wasn't matching anything, I suspect this dates back to the common docstring in `io/formats/format.py`.

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -2536,9 +2536,9 @@ class NDFrame(PandasObject, SelectionMixin):
 
         Parameters
         ----------
-        buf : StringIO-like, optional
-            Buffer to write to.
-        columns : sequence, optional, default None
+        buf : file descriptor or None
+            Buffer to write to. If None, the output is returned as a string.
+        columns : list of label, optional
             The subset of columns to write. Writes all columns by default.
         col_space : int, optional
             The minimum width of each column.
@@ -2549,18 +2549,19 @@ class NDFrame(PandasObject, SelectionMixin):
             Write row names (index).
         na_rep : str, default 'NaN'
             Missing data representation.
-        formatters : list or dict of one-param. functions, optional
+        formatters : list of functions or dict of {str: function}, optional
             Formatter functions to apply to columns' elements by position or
             name. The result of each function must be a unicode string.
             List must be of length equal to the number of columns.
-        float_format : str, default None
+        float_format : str, optional
             Format string for floating point numbers.
         sparsify : bool, optional, default None
             Set to False for a DataFrame with a hierarchical index to print
-            every multiindex key at each row.
-        index_names : bool, optional, default True
+            every multiindex key at each row. If None, the default will be
+            read from the config module.
+        index_names : bool, default True
             Prints the names of the indexes.
-        bold_rows : boolean, default False
+        bold_rows : bool, default False
             Make the row labels bold in the output.
         column_format : str, default None
             The columns format as specified in `LaTeX table format
@@ -2570,8 +2571,8 @@ class NDFrame(PandasObject, SelectionMixin):
             When set to None, the value will default from the pandas config
             module. Use a longtable environment instead of tabular. Requires
             adding a \usepackage{longtable} to your LaTeX preamble.
-        escape : bool, default will be read from the pandas config module
-            Default: True.
+        escape : bool, default None
+            If None, default will be read from the pandas config module.
             When set to False prevents from escaping latex special
             characters in column names.
         encoding : str, default None
@@ -2604,8 +2605,9 @@ class NDFrame(PandasObject, SelectionMixin):
 
         See Also
         --------
-        DataFrame.to_csv : Write a DataFrame to CSV format.
-        DataFrame.to_excel : Write a DataFrame to an Excel file.
+        DataFrame.to_string : Render a DataFrame to a console-friendly
+            tabular output.
+        DataFrame.to_html : Render a DataFrame as an HTML table.
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -2521,69 +2521,102 @@ def to_xarray(self):
...
+            name. The result of each function must be a unicode string.
+            List must be of length equal to the number of columns.
+        float_format : str, default None
+            Format string for floating point numbers.
+        sparsify : bool, optional, default None
+            Set to False for a DataFrame with a hierarchical index to print
+            every multiindex key at each row.
+        index_names : bool, optional, default True
+            Prints the names of the indexes.
         bold_rows : boolean, default False
```
datapythonista: `bool` instead of `boolean`

pandas/core/generic.py
```diff
@@ -2521,69 +2521,102 @@ def to_xarray(self):
...
-            Use a longtable environment instead of tabular. Requires adding
-            a \\usepackage{longtable} to your LaTeX preamble.
-        escape : boolean, default will be read from the pandas config module
+            <https://en.wikibooks.org/wiki/LaTeX/Tables>`__ e.g. 'rcl' for 3
+            columns.
+        longtable : bool, default None
+            When set to None, the value will default from the pandas config
+            module. Use a longtable environment instead of tabular. Requires
+            adding a \usepackage{longtable} to your LaTeX preamble.
+        escape : bool, default will be read from the pandas config module
```
datapythonista: can you leave here simple `default None`, and explain about the config in the description?

pandas/core/generic.py
```diff
@@ -2521,69 +2521,102 @@ def to_xarray(self):
...
+        str or None
+            If buf is None, returns the resulting LateX format as a
+            string. Otherwise returns None.
 
-            .. versionadded:: 0.20.0
-            """
+        See Also
+        --------
+        DataFrame.to_csv : Write a DataFrame to CSV format.
+        DataFrame.to_excel : Write a DataFrame to an Excel file.
```
datapythonista: I'd have `to_string` and `to_html` in this case, which to me are conceptually more similar than `to_csv` or `to_excel` (formatting to present vs formatting to export).

datapythonista (commented):
Great contribution, the docstring really needed some care. Added just some minor comments mainly about the parameter types.

pandas/core/generic.py
```diff
@@ -2521,69 +2521,102 @@ def to_xarray(self):
...
+            Missing data representation.
+        formatters : list or dict of one-param. functions, optional
+            Formatter functions to apply to columns' elements by position or
+            name. The result of each function must be a unicode string.
+            List must be of length equal to the number of columns.
+        float_format : str, default None
+            Format string for floating point numbers.
+        sparsify : bool, optional, default None
+            Set to False for a DataFrame with a hierarchical index to print
+            every multiindex key at each row.
```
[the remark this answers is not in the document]
Moisan: They are not the same, in `formats.py`:
````
if sparsify is None:
    sparsify = get_option("display.multi_sparse")

File: pandas/core/generic.py [hunk 5]
```diff
@@ -2521,69 +2521,104 @@ def to_xarray(self):
...
-            columns
-        longtable : boolean, default will be read from the pandas config module
-            Default: False.
-            Use a longtable environment instead of tabular. Requires adding
-            a \\usepackage{longtable} to your LaTeX preamble.
-        escape : boolean, default will be read from the pandas config module
-            Default: True.
+            <https://en.wikibooks.org/wiki/LaTeX/Tables>`__ e.g. 'rcl' for 3
+            columns.
+        longtable : bool, default None
```
jeffyancey: should it be `bool, optional, default None`, like on line 2558?

pandas/core/generic.py
```diff
@@ -2521,69 +2521,104 @@ def to_xarray(self):
...
+            When set to None, the value will default from the pandas config
+            module. Use a longtable environment instead of tabular. Requires
+            adding a \usepackage{longtable} to your LaTeX preamble.
+        escape : bool, default None
```
jeffyancey: same question as above.

pandas/core/generic.py
datapythonista: In this case it should be `bool, optional`. We never use `whatever_type, optional, default None`. We use only one of them.

When the `None` default value is being used, we use `default None`, imagine for example `.fillna(value=None)` where the `None` is the value used to impute.

When the `None` is just a flag, then we use `optional`. For example, in this case the `longtable` won't get the value `None` itself, but a value from the config. Meaning that it's optional to provide a `longtable` value, as we can use that.

datapythonista (commented):
Can you run `./scripts/validate_docstrings.py pandas.DataFrame.to_latex`?

I think it will complain that the first line should be a single line (this is important for the page with the list of methods)

When the script says everything is all right, I'm happy with it. Really nice change, much better docstring now.

datapythonista:
That's an error in the script them. Can you post the output of the script to see how the docstring is veing rendered, and see if it helps to see what's wrong with the script.

And after that fix the docstring, so the first line (short summary) fits in a single line, as described here: https://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html

datapythonista (approved):
lgtm, thanks for the contribution @Moisan 

If you can open the issue for the validation, that would be great.

jreback:
thanks @Moisan

Issue #17135, opened 2017-08-01
still doesn't work with boto3 StreamingBody


uiur:
ref: https://github.com/pandas-dev/pandas/issues/16135 https://github.com/pandas-dev/pandas/pull/16150

#### Example
```py
s3_object = client.get_object(Bucket=bucket, Key=key)
result = read_csv(s3_object["Body"])
# ValueError: Invalid file path or buffer object type: <class 'botocore.response.StreamingBody'>
```

#### Problem description
issue: is_file_like requirements are too strict for boto3 S3 objects https://github.com/pandas-dev/pandas/issues/16135
pull request: https://github.com/pandas-dev/pandas/pull/16150

This issue was closed but it's not working. I've found that the test is skipped and has a bug.
I'll send a pull request to reproduce the behavior.

I think it's because `botocore.response.StreamingBody` doesn't have `__iter__` so `is_file_like` returns `False`.

<details>

INSTALLED VERSIONS
------------------------
commit: c8dcf194c7a5264ed6d9161d5807e9da972cab4c
python: 2.7.10.final.0
python-bits: 64
OS: Darwin
OS-release: 14.5.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL:
LANG: en_US.UTF-8
LOCALE: None.None

pandas: 0.21.0.dev+317.gc8dcf19
pytest: 3.1.3
pip: 7.1.2
setuptools: 36.2.6
Cython: 0.26
numpy: 1.13.1
scipy: None
pyarrow: None
xarray: None
IPython: 5.0.0
sphinx: None
patsy: None
dateutil: 2.6.1
pytz: 2017.2
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: None
s3fs: 0.1.2
pandas_gbq: None
pandas_datareader: None
</details>

uiureo:
cc @gfyoung

gfyoung:
@uiureo : Thanks for the report!  From our conversation back in that issue, we got the impression that the PR would address this issue.  Network tests are more difficult to reproduce because they are out of our control, but that's a little unfortunate that the purported patch failed to address this.

gfyoung:
[...]

PR #22520, opened 2018-08-27
TST: Streaming of S3 files

kokes:
Issue #17135 was resolved by an upstream change, this is just a unit test to avoid regressions.

### DIFF
--- a/ci/environment-dev.yaml
+++ b/ci/environment-dev.yaml
@@ -9,6 +9,7 @@ dependencies:
   - flake8-comprehensions
   - hypothesis>=3.58.0
   - moto
+  - botocore>=1.10.47
   - pytest>=3.6
   - python-dateutil>=2.5.0
   - python=3
--- a/ci/environment-dev.yaml
+++ b/ci/environment-dev.yaml
@@ -9,7 +9,6 @@ dependencies:
   - flake8-comprehensions
   - hypothesis>=3.58.0
   - moto
-  - botocore>=1.10.47
   - pytest>=3.6
   - python-dateutil>=2.5.0
   - python=3
--- a/pandas/tests/io/test_s3.py
+++ b/pandas/tests/io/test_s3.py
@@ -1,3 +1,7 @@
+from six import BytesIO
+
+from botocore.response import StreamingBody
+from pandas import read_csv
 from pandas.io.common import is_s3_url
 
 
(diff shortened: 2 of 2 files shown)


kokes:
CI will crash and burn as I didn't add `botocore` as a dependency properly. Any help on this appreciated.

TomAugspurger:
I think the easiest thing to do is to ensure the test is skipped without botocore, and rely on `s3fs` pulling it in as a dependency.

ci/environment-dev.yaml
```diff
@@ -9,6 +9,7 @@ dependencies:
   - flake8-comprehensions
   - hypothesis>=3.58.0
   - moto
+  - botocore>=1.10.47
```
TomAugspurger: So this can be removed.

pandas/tests/io/test_s3.py
```diff
@@ -1,3 +1,7 @@
+from six import BytesIO
+
+from botocore.response import StreamingBody
```
TomAugspurger: And remove this.

pandas/tests/io/test_s3.py
```diff
@@ -6,3 +10,13 @@ class TestS3URL(object):
     def test_is_s3_url(self):
         assert is_s3_url("s3://pandas/somethingelse.com")
         assert not is_s3_url("s4://pandas/somethingelse.com")
+
+
+def test_streaming_s3_objects():
+    data = [
```
TomAugspurger: Then here you can do `botocore_response = pytest.importorskip('botocore.response', minversion='1.10.47')`

pandas/tests/io/test_s3.py
```diff
@@ -1,3 +1,7 @@
+from six import BytesIO
```
TomAugspurger: `from pandas.compat import BytesIO`

pandas/tests/io/test_s3.py
kokes: Can we do this when the version pertains to the base module, not the imported class? I just tried it in a dummy file and all I got was `Skipped: module 'botocore.response' has __version__ None, required is: '1.10.47'`. So should I do something like this?

```
import pytest

def my_test_fn():
    pytest.importorskip('botocore', minversion='1.10.47')
    from botocore.response import StreamingBody
    # test
```

pandas/tests/io/test_s3.py
```diff
@@ -6,3 +10,16 @@ class TestS3URL(object):
     def test_is_s3_url(self):
         assert is_s3_url("s3://pandas/somethingelse.com")
         assert not is_s3_url("s4://pandas/somethingelse.com")
+
+
+def test_streaming_s3_objects():
+    pytest.importorskip('botocore', minversion='1.10.47')
```
jreback: can you add the issue number here as a comment

jreback (changes requested):
doc comment ping on green.

pandas/tests/io/test_s3.py
kokes: yup, done

jreback (approved):

jreback:
thanks!

PR #22531, opened 2018-08-29
TST: Move tests/scripts to scripts/tests

gfyoung:
Moves the `tests/scripts` directory to `scripts/tests`, as these tests don't really belong in the top-level `pandas` directory (they test the `scripts` directory, not `pandas`).

### DIFF
--- a/scripts/tests/conftest.py
+++ b/scripts/tests/conftest.py
@@ -1,0 +1,3 @@
+def pytest_addoption(parser):
+    parser.addoption("--strict-data-files", action="store_true",
+                     help="Unused. For compat with setup.cfg.")
--- a/scripts/tests/test_validate_docstrings.py
+++ b/scripts/tests/test_validate_docstrings.py
@@ -1,8 +1,10 @@
-import os
-import sys
-
-import numpy as np
+import string
+import random
 import pytest
+import numpy as np
+
+import validate_docstrings
+validate_one = validate_docstrings.validate_one
 
 
 class GoodDocStrings(object):
@@ -44,7 +46,7 @@ class GoodDocStrings(object):
         float
             Random number generated.
         """
-        return random.random()  # noqa: F821
+        return random.random()
 
     def random_letters(self):
         """
@@ -60,9 +62,8 @@ class GoodDocStrings(object):
         letters : str
             String of random letters.
         """
-        length = random.randint(1, 10)  # noqa: F821
-        letters = ''.join(random.choice(string.ascii_lowercase)  # noqa: F821
-                          for i in range(length))
+        length = random.randint(1, 10)
+        letters = "".join(random.sample(string.ascii_lowercase, length))
         return length, letters
 
     def sample_values(self):
@@ -78,7 +79,7 @@ class GoodDocStrings(object):
             Random number generated.
         """
         while True:
-            yield random.random()  # noqa: F821
+            yield random.random()
 
     def head(self):
         """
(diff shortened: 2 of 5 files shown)


scripts/tests/test_validate_docstrings.py
```diff
@@ -1,8 +1,10 @@
-import os
-import sys
-
-import numpy as np
 import pytest
+import string
+import random
```
datapythonista: standard library imports should go before third-party module imports

https://www.python.org/dev/peps/pep-0008/#imports

scripts/tests/test_validate_docstrings.py
```diff
@@ -1,8 +1,10 @@
...
+import numpy as np
+
+validate_docstrings = pytest.importorskip("validate_docstrings")
```
datapythonista: I guess I'm missing something obvious, but why is this needed? can't we simply `import validate_docstrings`?

scripts/tests/test_validate_docstrings.py
```diff
@@ -60,9 +62,9 @@ def random_letters(self):
         letters : str
             String of random letters.
         """
-        length = random.randint(1, 10)  # noqa: F821
-        letters = ''.join(random.choice(string.ascii_lowercase)  # noqa: F821
-                          for i in range(length))
+        length = random.randint(1, 10)
+        letters = "".join(random.choice(string.ascii_lowercase)
+                          for _ in range(length))
```
datapythonista: I know this comes from the original code, but we don't need a comprehension here, we can simply do `''.join(random.sample(string.ascii_lowercase, length))`

datapythonista (approved):
Good changes, I think they make things simpler and better organized. Thanks!

scripts/tests/test_validate_docstrings.py
gfyoung: That's vestigial paranoia from when we skipped tests if we couldn't see `validate_docstrings`.  I agree that we can just import this without the protection.

.travis.yml
```diff
@@ -57,6 +57,9 @@ matrix:
     - dist: trusty
       env:
         - JOB="3.6, coverage" ENV_FILE="ci/travis-36.yaml" TEST_ARGS="--skip-slow --skip-network" PANDAS_TESTING_MODE="deprecate" COVERAGE=true DOCTEST=true
+    - dist: trusty
```
jreback: i don't think we need another build

scripts/tests/conftest.py
```diff
@@ -0,0 +1,3 @@
+def pytest_addoption(parser):
```
jreback: what is this for?

jreback (changes requested):

WillAyd:
Think we also need to update the Makefile to ensure this still gets linted (?)

.travis.yml
jreback: you can just put on an existing build

jreback (changes requested):

.travis.yml
gfyoung: That's fair, and I agree we can do that.  I don't have access to my computer ATM (which is why I haven't addressed the comments).

scripts/tests/conftest.py
gfyoung: As it says in the code: it's needed to placate `setup.cfg`.

scripts/tests/test_validate_docstrings.py
gfyoung: Indeed, they should!  Done.

scripts/tests/test_validate_docstrings.py
gfyoung: Fair enough.  Done.

gfyoung:
> Think we also need to update the Makefile to ensure this still gets linted (?)

We don't use the `Makefile` when doing the linting on Travis.  That being said, I don't see why we don't add it as another command.

gfyoung:
@jreback @datapythonista @WillAyd : I've addressed all of your comments, and everything is still green!  PTAL.

jreback (approved):

gfyoung:
@jreback : Is this good to merged since you approved it?

jreback:
yep

PR #22535, opened 2018-08-29
DEPR: deprecate integer add/sub with DTI/TDI/PI/Timestamp/Period

### LINKED ISSUE #21939: [DISC] DatetimeIndex/TimedeltaIndex ops with integers should deprecated
An example of the behavior that should be deprecated:

```
In [2]: dti = pd.date_range('2016-01-01', periods=3)
In [3]: dti + 1
Out[3]: DatetimeIndex(['2016-01-02', '2016-01-03', '2016-01-04'], dtype='datetime64[ns]', freq='D')
In [4]: dti[1] + 100
Out[4]: Timestamp('2016-04-11 00:00:00', freq='D')

In [5]: tdi = pd.timedelta_range('1D', periods=3)
In [6]: tdi - 1
Out[6]: TimedeltaIndex(['0 days', '1 days', '2 days'], dtype='timedelta64[ns]', freq='D')
In [7]: tdi[-1] - 1
<ipython-input-7-ac830ffbe976> in <module>()
----> 1 tdi[-1] - 1

TypeError: unsupported operand type(s) for -: 'Timedelta' and 'int'
```

This `TypeError` occurs because `TimedeltaIndex._box_func` does not pass `self.freq` to `Timedelta`, and is one of many ways that this behavior introduces inconsistencies into the code.

- Integer add/sub for DTI/DTI just passes through to `self.shift(n)`, so there is a ready alternative.
- This behavior is inconsistent between `DatetimeIndex` and `Series[datetime64]` (#20937) (same for TDI)
- `freq` is not really a useful part of the Timestamp or Timedelta abstraction (#15146).

- The existence of a `Timestamp.freq` gives many users the incorrect impression that `Period.freq` behaves the same way (#4591, #4878, #5091, #12379)
- to_datetime fails to preserve `freq` in some cases (need to track down the issue)
- Timestamp and Timedelta `__eq__` ignores `freq`

Bottom line: DTI/TDI/Timestamp/Timedelta arithmetic with integers should be deprecated to a) cut out a bunch of hard-to-maintain code and b) avoid internal inconsistencies.

jbrockmendel:
Discussed in #21939; these operations cause headaches and internal-inconsistencies (e.g. ATM `Timedelta` is the only one of the related classes that doesn't support integer ops)

- [x] closes #21939
- [x] tests added / passed
- [x] whatsnew entry


jreback (changes requested):

pandas/core/arrays/datetimelike.py
```diff
@@ -466,6 +466,15 @@ def _addsub_int_array(self, other, op):
         result : same class as self
         """
         assert op in [operator.add, operator.sub]
+
+        if self.freq is not None:
+            # case with freq of None will raise
+            warnings.warn("Addition/subtraction of integer array from {cls} "
+                          "is deprecated, will be removed in version "
+                          "0.26.0 (or 1.0, whichever comes first).  "
```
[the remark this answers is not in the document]
jbrockmendel: Will update.

mroeschke:
There are some reference to integer arithmetic in `timeseries.rst` (e.g. https://pandas.pydata.org/pandas-docs/stable/timeseries.html#period). It's probably a good idea to scrub them so we don't advertise depreciated functionality

jbrockmendel:
Is there a usage of `tm.assert_produces_warning` to use when we expect the affected code to produce two warnings?

jreback:
> Is there a usage of tm.assert_produces_warning to use when we expect the affected code to produce two warnings?

I think you can pass a list? (you can to the underlying machinery, so maybe this just works, if not it should)

jreback:
@jbrockmendel pls add to the list in the deprecation issue (#6581) , when we merge can check the box

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -514,6 +514,8 @@ Deprecations
 - The signature of :meth:`Series.to_csv` has been uniformed to that of doc:meth:`DataFrame.to_csv`: the name of the first argument is now 'path_or_buf', the order of subsequent arguments has changed, the 'header' argument now defaults to True. (:issue:`19715`)
 - :meth:`Categorical.from_codes` has deprecated providing float values for the ``codes`` argument. (:issue:`21767`)
 - :func:`pandas.read_table` is deprecated. Instead, use :func:`pandas.read_csv` passing ``sep='\t'`` if necessary (:issue:`21948`)
+- Addition or subtraction of integers with :class:`Timestamp`, :class:`Period`, :class:`DatetimeIndex`, :class:`TimedeltaIndex`, :class:`PeriodIndex` is deprecated, will be removed in a future version (:issue:`21939`)
+- Addition or subtraction of integer-dtyped arrays with:class:`DatetimeIndex`, :class:`TimedeltaIndex`, :class:`PeriodIndex` is deprecated, will be removed in a future version (:issue:`21939`)
```
jreback: space after with

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -250,14 +250,23 @@ cdef class _Timestamp(datetime):
...
+        cdef:
+            int64_t other_int, nanos
 
         if is_timedelta64_object(other):
             other_int = other.astype('timedelta64[ns]').view('i8')
             return Timestamp(self.value + other_int,
                              tz=self.tzinfo, freq=self.freq)
 
         elif is_integer_object(other):
+            if self.freq is None:
```
jreback: is it necessary for freq to be none here? (we don't have this check elsewhere)

jreback (changes requested):

pandas/_libs/tslibs/timestamps.pyx
jbrockmendel: Should be not-None, good catch.

jbrockmendel:
Will update as indicated.  BTW gentle ping re #22350; many upcoming PRs touch those files, so it'd be helpful to get it out of the way.

jbrockmendel:
> I think you can pass a list? (you can to the underlying machinery, so maybe this just works, if not it should)

Doesn't look like it, I'll fix that.

jbrockmendel:
> I think you can pass a list? (you can to the underlying machinery, so maybe this just works, if not it should)

Trying to implement this, I've got most of it working, but can find no combination of `stacklevel=` parameters that don't cause the "not set with correct stacklevel" assertion to fail.  Who is the team expert on this?

jreback:
don't worry about checking the stacklevel too much, the idea is to raise it to a reasonable level, e.g. if you run the tests, it should point to the top-level of the code where its most releveant (rather than a lower level call). Just pick something that looks reasonable, and you can turn off the stacklevel check itself.

jreback (approved):

pandas/_libs/tslibs/timestamps.pyx
jreback: does this need to be changed?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -556,6 +556,8 @@ Deprecations
 - :func:`pandas.read_table` is deprecated. Instead, use :func:`pandas.read_csv` passing ``sep='\t'`` if necessary (:issue:`21948`)
 - :meth:`Series.str.cat` has deprecated using arbitrary list-likes *within* list-likes. A list-like container may still contain
   many ``Series``, ``Index`` or 1-dimensional ``np.ndarray``, or alternatively, only scalar values. (:issue:`21950`)
+- Addition or subtraction of integers with :class:`Timestamp`, :class:`Period`, :class:`DatetimeIndex`, :class:`TimedeltaIndex`, :class:`PeriodIndex` is deprecated, will be removed in a future version (:issue:`21939`)
```
jreback: pls make a sub-section and show what is being deprecated and what is the alternative. this is actually a rather large change.

jreback (changes requested):

doc/source/whatsnew/v0.24.0.txt
jbrockmendel: I'll try to get to this tomorrow.

jbrockmendel: @jreback just added this section; wasn't sure on where in the doc to put it and the ipython/code-block syntax is a shot in the dark.  Please advise.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -575,6 +575,57 @@ Deprecations
...
+    Out[8]: TimedeltaIndex(['-1 days', '1 days'], dtype='timedelta64[ns]', freq=None)
+
+    In [9]: dti = pd.date_range('2001-01-01', periods=2, freq='7D')
+    In [10]: dti + pd.Index([1, 2])
+    Out[10]: DatetimeIndex(['2001-01-08', '2001-01-22'], dtype='datetime64[ns]', freq=None)
+
+Current Behavior:
+
+.. ipython:: python
+
```
jreback: these should be executable statements, e.g. just show the input, the output is run

pandas/_libs/tslibs/period.pyx
```diff
@@ -1645,6 +1646,13 @@ cdef class _Period(object):
             elif other is NaT:
                 return NaT
             elif util.is_integer_object(other):
+                warnings.warn("Addition of integers to {cls} is "
```
jreback: can you call a function for this warning (not sure exatcly where), rather than repeating each time, prob need to pass in the level

jreback (changes requested):
looks good, pls rebase as well

pandas/_libs/tslibs/period.pyx
jbrockmendel: Not sure what you have in mind.  Just a one-liner containing this warning?

doc/source/whatsnew/v0.24.0.txt
jbrockmendel: Updated.  One day I'll get the hang of this doc system...

jbrockmendel:
Since last review:

Edits to pass CI env that fails on unhandled warnings, so lots of `tm.assert_produces_warning`s added.

Identified a bug in PeriodIndex addition #22998.  The place this shows up here is in `PeriodIndex.to_timestamp` which filters out the FutureWarning and continues using the deprecated behavior until 22998 is fixed.

in tseries.offsets, many DateOffset subclasses have `apply_index` implementations that use the deprecated behavior.  They are not hard to update, but that would come with a non-trivial performance penalty.  For now they filter out the warning and continue using the deprecated behavior.  I think the thing to do is to make a non-method helper function (possibly in tslibs.offsets) to take the place of this behavior.

doc/source/whatsnew/v0.24.0.txt
jreback: if you can update these

pandas/_libs/tslibs/period.pyx
jreback: yeah a module level function just seems better than copy-pasting

pandas/tests/arithmetic/test_period.py
```diff
@@ -857,47 +875,53 @@ def test_pi_ops_nat(self):
...
-        exp = PeriodIndex(['2010-12', '2010-12', 'NaT', '2010-12'],
-                          freq='M', name='idx')
-        self._check(idx, f, exp)
-
-        f = lambda x: np.subtract(x, np.array([3, 2, 3, -2]))
-        exp = PeriodIndex(['2010-10', '2010-12', 'NaT', '2011-06'],
-                          freq='M', name='idx')
-        self._check(idx, f, exp)
+        with tm.assert_produces_warning(FutureWarning, check_stacklevel=False,
+                                        clear=[pd.core.arrays.datetimelike]):
```
jreback: prob can parametrize lots of this (future PR of course)

pandas/tseries/offsets.py
```diff
@@ -1319,7 +1338,12 @@ def _end_apply_index(self, dtindex):
         else:
             roll = self.n
 
-        base = (base_period + roll).to_timestamp(how='end')
+        with warnings.catch_warnings(record=True):
+            # integer-array addition on PeriodIndex is deprecated,
+            #  but still used internally for performance
```
jreback: why can't you turn roll into an offset here, rather than catchng the warning? (and all other cases in this module), e.g. since these are not tests, would rather actually fix the issue

jreback (changes requested):
also rebase

jreback:
also prob need to check the docs and change / remove the cases which causes warnings.

pandas/tseries/offsets.py
jbrockmendel: Using the deprecated behavior amounts to adding two int64 ndarrays.  The non-deprecated version here `base_period + roll * offset` turns `roll*offset` into an object-array, so the whole thing is slow.

I've got a nice solution for this in mind.  This and a few other parts of this PR can be addressed more gracefully after #23031.

doc/source/whatsnew/v0.24.0.txt
jreback: you don't need the ipython line numbers

pandas/_libs/tslibs/period.pyx
jreback: can you do this

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -315,14 +315,23 @@ cdef class _Timestamp(datetime):
...
+        cdef:
+            int64_t other_int, nanos
 
         if is_timedelta64_object(other):
             other_int = other.astype('timedelta64[ns]').view('i8')
             return Timestamp(self.value + other_int,
                              tz=self.tzinfo, freq=self.freq)
 
         elif is_integer_object(other):
+            if self.freq is not None:
```
jreback: so I would put the warning function in this module and just call it from higher levels

jreback (changes requested):
need to fix the docs and consolidate the warnings to a single function.

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -40,8 +40,19 @@ from timezones cimport (
 _zero_time = datetime_time(0, 0)
 _no_input = object()
 
+
 # ----------------------------------------------------------------------
 
+def int_op_deprecated(obj):
```
jreback: ok this looks good. can you rename to 
``maybe_integer_op_deprecated`

jreback (changes requested):
small change, otherwise lgtm. ping on green.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -955,6 +955,53 @@ Deprecations
...
+    per = pd.Period('2016Q1')
+    per + 3 * per.freq
+
```
jreback: you might want to show the deprecation warning itself on 1 case here (you will need to add ``:okwarning:`` though)

pandas/core/arrays/period.py
```diff
@@ -778,7 +784,7 @@ def _add_delta_tdi(self, other):
         assert isinstance(self.freq, Tick)  # checked by calling function
 
         delta = self._check_timedeltalike_freq_compat(other)
-        return self._addsub_int_array(delta, operator.add).asi8
```
jreback: huh? what are you adding a new kwarg for?

pandas/tests/arithmetic/test_datetime64.py
```diff
@@ -1066,14 +1067,16 @@ def test_dti_iadd_int(self, tz_naive_fixture, one):
                             periods=10, tz=tz)
         expected = pd.date_range('2000-01-01 10:00', freq='H',
                                  periods=10, tz=tz)
-        rng += one
+        with tm.assert_produces_warning(FutureWarning, check_stacklevel=False):
```
jreback: why can't you check the stacklevel anymore?

jreback (changes requested):

pandas/tests/arithmetic/test_datetime64.py
jbrockmendel: We basically never did.  I never got the stacklevel checks to pass in any consistent way.

pandas/_libs/tslibs/timestamps.pyx
jbrockmendel: sure

pandas/core/arrays/period.py
jbrockmendel: We still call `_addsub_int_array` internally from a few places (here and in DateOffset `apply_index` methods) and want to suppress the warning.

But actually now that the warning is a one-liner, it is less cumbersome to put it directly in `__add__/__sub__`.  Will change.

pandas/tseries/offsets.py
```diff
@@ -536,13 +539,21 @@ def apply_index(self, i):
         time = i.to_perioddelta('D')
         # to_period rolls forward to next BDay; track and
         # reduce n where it does when rolling forward
-        shifted = (i.to_perioddelta('B') - time).asi8 != 0
+        asper = i.to_period('B')
         if self.n > 0:
+            shifted = (i.to_perioddelta('B') - time).asi8 != 0
+
+            # Integer-array addition is deprecated, so we use
```
jreback: this comment is out of date, but can change later

jreback (approved):

jreback:
thanks @jbrockmendel sometimes they take a lot of iterations :-D

jbrockmendel:
It's all about the power of Teamwork.

jorisvandenbossche:
@jbrockmendel what was actually the rationale of deprecating it for Periods as well? 
The issue that is closed by this (https://github.com/pandas-dev/pandas/issues/21939) discusses it for DatetimeIndex/TimedeltaIndex, but I think says to keep it for PeriodIndex for now. 

In any case, for PeriodIndex, this was an explicitly documented behaviour, and the documentation is now no longer up to date. 
But instead of updating the docs, I think we should also reconsider the deprecation.

(and to be clear: I fully agree with the deprecation for DatetimeIndex / TimedeltaIndex!)

jbrockmendel:
> I think we should also reconsider the deprecation

I'm open to that.  The ones I care about are the TS/TD/DTI/TDI

jorisvandenbossche:
After commenting here I opened an issue with the same content: https://github.com/pandas-dev/pandas/issues/24305. Can you post it there?

PR #22539, opened 2018-08-30
DOC: Improve the docstring of DataFrame.equals()

### LINKED ISSUE #22462: DOC: Update docstring of DataFrame/Series .equals() to explain the behavior with indices
Some work and some discussion happened at #19213.

`df1.equals(df2)` returns a single boolean value on whether the two dataframes contain the same values and indices. This includes types for the data, but not for the indices.

The docstring should be improved to follow all our standards, including:
- A `See Also` section that at least links to `.eq()`
- `Parameters`
- `Examples`
- `Notes` with the mentioned behavior about indices

seantchan:
- [X] closes #22462
- [ ] tests added / passed
- [ ] whatsnew entry

```
##################### Docstring (pandas.DataFrame.equals)  #####################

Test whether two DataFrame objects contain the same elements.

This function allows two DataFrame objects to be compared against
each other to see if they same shape and elements. NaNs in the same
location are considered equal. The column headers do not need to have
the same type, but the elements within the columns must be
the same dtype.

Parameters
----------
other : DataFrame
    The other DataFrame to be compared with the first.

Returns
-------
bool
    True if all elements are the same in both DataFrames, False
    otherwise.

See Also
--------
pandas.Series.eq : Compare two Series objects of the same length
    and return a Series where each element is True if the element
    in each Series is equal, False otherwise.

numpy.array_equal : Return True if two arrays have the same shape
    and elements, False otherwise.

Notes
-----
This function requires that the elements have the same dtype as their
respective elements in the other DataFrame. However, the indices do
not need to have the same type, as long as they are still considered
equal.

Examples
--------
>>> a = pd.DataFrame({1:[0], 0:[1]})
>>> b = pd.DataFrame({1.0:[0], 0.0:[1]})

DataFrames a and b have the same element types and values, but have
different types for the indices, which will still return True.

>>> a.equals(b)
True

DataFrames a and c have different types for the same values for their
elements, and will return False even though the indices are the same
values and types.

>>> c = pd.DataFrame({1:[0.0], 0:[1.0]})
>>> a.equals(c)
False

Docstring for "pandas.DataFrame.equals" correct. :)
```

### DIFF
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -1303,58 +1303,80 @@ class NDFrame(PandasObject, SelectionMixin):
 
     def equals(self, other):
         """
-        Test whether two DataFrame objects contain the same elements.
+        Test whether two NDFrame objects contain the same elements.
 
-        This function allows two DataFrame objects to be compared against
-        each other to see if they same shape and elements. NaNs in the same
-        location are considered equal. The column headers do not need to have
-        the same type, but the elements within the columns must be
-        the same dtype.
+        This function allows two NDFrame objects to be compared against
+        each other to see if they have the same shape and elements. NaNs in
+        the same location are considered equal. The column headers do not
+        need to have the same type, but the elements within the columns must
+        be the same dtype.
 
         Parameters
         ----------
-        other : DataFrame
-            The other DataFrame to be compared with the first.
+        other : NDFrame
+            The other NDFrame to be compared with the first.
 
         Returns
         -------
         bool
-            True if all elements are the same in both DataFrames, False
+            True if all elements are the same in both NDFrames, False
             otherwise.
 
         See Also
         --------
-        pandas.Series.eq : Compare two Series objects of the same length
+        Series.eq : Compare two Series objects of the same length
             and return a Series where each element is True if the element
             in each Series is equal, False otherwise.
-
+        DataFrame.eq : Compare two DataFrame objects of the same shape and
+            return a DataFrame where each element is True if the respective
+            element in each DataFrame is equal, False otherwise.
         numpy.array_equal : Return True if two arrays have the same shape
             and elements, False otherwise.
 
         Notes
         -----
         This function requires that the elements have the same dtype as their
-        respective elements in the other DataFrame. However, the indices do
-        not need to have the same type, as long as they are still considered
-        equal.
+        respective elements in the other DataFrame. However, the column labels
+        do not need to have the same type, as long as they are still
+        considered equal.
 
         Examples
         --------
-        >>> a = pd.DataFrame({1:[0], 0:[1]})
-        >>> b = pd.DataFrame({1.0:[0], 0.0:[1]})
+        >>> df = pd.DataFrame({1:[0], 0:[1]})
+        >>> df
+           1  0
+        0  0  1
 
-        DataFrames a and b have the same element types and values, but have
-        different types for the indices, which will still return True.
+        DataFrames df and exactly_equal have the same types and values for
+        their elements and column labels, which will return True.
 
-        >>> a.equals(b)
+        >>> exactly_equal = pd.DataFrame({1:[0], 0:[1]})
+        >>> exactly_equal
+           1  0
+        0  0  1
+        >>> df.equals(exactly_equal)
         True
 
-        DataFrames a and c have different types for the same values for their
-        elements, and will return False even though the indices are the same
-        values and types.
+        DataFrames df and different_column_label have the same element
+        types and values, but have different types for the column labels,
+        which will still return True.
 
-        >>> c = pd.DataFrame({1:[0.0], 0:[1.0]})
-        >>> a.equals(c)
+        >>> different_column_label = pd.DataFrame({1.0:[0], 0.0:[1]})
+        >>> different_column_label
+           1.0  0.0
+        0    0    1
+        >>> df.equals(different_column_label)
+        True
+
+        DataFrames df and different_data_type have different types for the
+        same values for their elements, and will return False even though
+        their column labels are the same values and types.
+
+        >>> different_data_type = pd.DataFrame({1:[0.0], 0:[1.0]})
+        >>> different_data_type
+             1    0
+        0  0.0  1.0
+        >>> df.equals(different_data_type)
         False
         """
         if not isinstance(other, self._constructor):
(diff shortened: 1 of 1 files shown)


pandas/core/generic.py
```diff
@@ -1303,8 +1303,59 @@ def __invert__(self):
 
     def equals(self, other):
         """
-        Determines if two NDFrame objects contain the same elements. NaNs in
-        the same location are considered equal.
+        Test whether two DataFrame objects contain the same elements.
+
+        This function allows two DataFrame objects to be compared against
+        each other to see if they same shape and elements. NaNs in the same
```
datapythonista: they have the same shape...

pandas/core/generic.py
```diff
@@ -1303,8 +1303,59 @@ def __invert__(self):
...
+
+        Returns
+        -------
+        bool
+            True if all elements are the same in both DataFrames, False
+            otherwise.
+
+        See Also
+        --------
+        pandas.Series.eq : Compare two Series objects of the same length
```
datapythonista: No need for the `pandas.` prefix, just `Series.eq` is enough.

I think we need `eq` and `equals` of both `Series` and `DataFrame`, not just `Series.eq`.

pandas/core/generic.py
```diff
@@ -1303,8 +1303,59 @@ def __invert__(self):
...
+            and return a Series where each element is True if the element
+            in each Series is equal, False otherwise.
+
+        numpy.array_equal : Return True if two arrays have the same shape
```
datapythonista: no need for the blank line before this

pandas/core/generic.py
```diff
@@ -1303,8 +1303,59 @@ def __invert__(self):
...
+        -----
+        This function requires that the elements have the same dtype as their
+        respective elements in the other DataFrame. However, the indices do
+        not need to have the same type, as long as they are still considered
+        equal.
+
+        Examples
+        --------
+        >>> a = pd.DataFrame({1:[0], 0:[1]})
+        >>> b = pd.DataFrame({1.0:[0], 0.0:[1]})
```
datapythonista: to be consistent with other docstrings, it'd be good to use `df` for the main DataFrame the ones in which we'll call the method `df.equals()`, and something more descriptive for the other (e.g. `exactly_equal`, `different_data_type`, `different_index_type`...)

After creating the data, it'd be nice to display the content (i.e. `>>> df`)

Not sure in this case if it'd be better or worse, but in general we try to have somehow real-world (meaningful) data for the examples.

pandas/core/generic.py
```diff
@@ -1303,8 +1303,59 @@ def __invert__(self):
...
+
+        DataFrames a and b have the same element types and values, but have
+        different types for the indices, which will still return True.
```
datapythonista: "indices" are the row labels, which are the same, in this case the different type is in the column labels.

pandas/core/generic.py
```diff
@@ -1303,8 +1303,59 @@ def __invert__(self):
 
     def equals(self, other):
         """
-        Determines if two NDFrame objects contain the same elements. NaNs in
-        the same location are considered equal.
+        Test whether two DataFrame objects contain the same elements.
```
datapythonista: This docstring is used for both Series and DataFrame, we should be generic.

datapythonista (commented):
Great docstring, added some comments.

pandas/core/generic.py
seantchan: Would it be redundant to mention `equals` since this is the docstring for that method?

pandas/core/generic.py
```diff
@@ -1303,8 +1303,81 @@ def __invert__(self):
 
     def equals(self, other):
         """
-        Determines if two NDFrame objects contain the same elements. NaNs in
-        the same location are considered equal.
+        Test whether two NDFrame objects contain the same elements.
```
datapythonista: I think you can leave `Test whether two objects contain the same elements.` for now.

The problem is that NDFrame is not a class we expect users to know about (it's private). So, it shouldn't be used in the documentation. `Series or DataFrame` or `object` are probably what we want to use in most cases.

pandas/core/generic.py
```diff
@@ -1303,8 +1303,81 @@ def __invert__(self):
...
+        See Also
+        --------
+        Series.eq : Compare two Series objects of the same length
+            and return a Series where each element is True if the element
+            in each Series is equal, False otherwise.
+        DataFrame.eq : Compare two DataFrame objects of the same shape and
+            return a DataFrame where each element is True if the respective
+            element in each DataFrame is equal, False otherwise.
+        numpy.array_equal : Return True if two arrays have the same shape
+            and elements, False otherwise.
```
datapythonista: We add the method being documented to the `See Also`, only in cases where the docstring is being reused by many methods, and we want those methods to reference each other. But this is not the case here. May be you can still add the `pandas.utils.testing.assert_frame_equal` and `assert_series_equal` here, but I don't think it adds value to reference `.equals` itself.

pandas/core/generic.py
```diff
@@ -1303,8 +1303,81 @@ def __invert__(self):
...
+
+        Notes
+        -----
+        This function requires that the elements have the same dtype as their
+        respective elements in the other DataFrame. However, the column labels
```
datapythonista: in the other Series or DataFrame

pandas/core/generic.py
```diff
@@ -1303,8 +1303,81 @@ def __invert__(self):
...
+        do not need to have the same type, as long as they are still
+        considered equal.
+
+        Examples
+        --------
+        >>> df = pd.DataFrame({1:[0], 0:[1]})
```
datapythonista: missing spaces after colons. I think using 0 -> 1 and 1 -> 0 is a bit confusing, may be 1 -> 10 and 2 -> 20 makes the example easier to read?

pandas/core/generic.py
```diff
@@ -1303,8 +1303,81 @@ def __invert__(self):
...
+           1  0
+        0  0  1
+        >>> df.equals(exactly_equal)
+        True
+
+        DataFrames df and different_column_label have the same element
+        types and values, but have different types for the column labels,
+        which will still return True.
+
+        >>> different_column_label = pd.DataFrame({1.0:[0], 0.0:[1]})
```
datapythonista: `different_column_type` seems a clearer name to me

datapythonista (changes requested):
The docstring looks really good. Just some comments about technicalities.

datapythonista (approved):
lgtm, really good docstring, thanks for the work on this @seantchan

PR #22544, opened 2018-08-30
DOC: Fix copy-pasted name in `.day_name` docstring

Peque:
Just a small change... Feel free to close and include this change in any other typo-fixing commit if you like. :blush:

### DIFF
--- a/pandas/core/arrays/datetimes.py
+++ b/pandas/core/arrays/datetimes.py
@@ -793,17 +793,27 @@ class DatetimeArrayMixin(dtl.DatetimeLikeArrayMixin):
         """
         Return the month names of the DateTimeIndex with specified locale.
 
+        .. versionadded:: 0.23.0
+
         Parameters
         ----------
-        locale : string, default None (English locale)
-            locale determining the language in which to return the month name
+        locale : str, optional
+            Locale determining the language in which to return the month name.
+            Default is English locale.
 
         Returns
         -------
-        month_names : Index
-            Index of month names
+        Index
+            Index of month names.
 
-        .. versionadded:: 0.23.0
+        Examples
+        --------
+        >>> idx = pd.DatetimeIndex(start='2018-01', freq='M', periods=3)
+        >>> idx
+        DatetimeIndex(['2018-01-31', '2018-02-28', '2018-03-31'],
+                      dtype='datetime64[ns]', freq='M')
+        >>> idx.month_name()
+        Index(['January', 'February', 'March'], dtype='object')
         """
         if self.tz is not None and self.tz is not utc:
             values = self._local_timestamps()
@@ -819,17 +829,27 @@ class DatetimeArrayMixin(dtl.DatetimeLikeArrayMixin):
         """
         Return the day names of the DateTimeIndex with specified locale.
 
+        .. versionadded:: 0.23.0
+
         Parameters
         ----------
-        locale : string, default None (English locale)
-            locale determining the language in which to return the day name
+        locale : str, optional
+            Locale determining the language in which to return the day name.
+            Default is English locale.
 
         Returns
         -------
-        month_names : Index
-            Index of day names
+        Index
+            Index of day names.
 
-        .. versionadded:: 0.23.0
+        Examples
+        --------
+        >>> idx = pd.DatetimeIndex(start='2018-01-01', freq='D', periods=3)
+        >>> idx
+        DatetimeIndex(['2018-01-01', '2018-01-02', '2018-01-03'],
+                      dtype='datetime64[ns]', freq='D')
+        >>> idx.day_name()
+        Index(['Monday', 'Tuesday', 'Wednesday'], dtype='object')
         """
         if self.tz is not None and self.tz is not utc:
             values = self._local_timestamps()


WillAyd:
You can simply remove `day_names` and `month_names` from the docstrings since they don't reference any variable that actually exists. It's OK for a returns to simply describe the type of object being returned.

See https://python-sprints.github.io/pandas/guide/pandas_docstring.html#section-4-returns-or-yields

Peque:
@WillAyd Actually, I prefer it that way, did not know it was allowed in Pandas documentation. :innocent: 

Changed it and force-pushed the commit.

WillAyd:
Do these pass the validate_docstring script? I think there is at least a missing period at the end of the descriptions that we should probably just clean up while touching. Maybe a couple more

WillAyd (changes requested):
As mentioned above if you could run these docstrings against the validation script and fix any issues would be ideal. Should be easy enough to do in one pass

datapythonista (commented):
The changes look great.

But as @WillAyd said, if you don't mind taking care of the whole docstring, that would be very useful.

Meaning:
- Replace the parameter type from `string` to `str`
- Capitalize the first letter of the parameter description, and finish it with period
- Move that `None` means English locale from the type line, to the description of the parameter line.
- Finish the returns description with a period
- Add a simple example

Thanks!

Peque:
Thanks all for your feedback. Updated the docstrings with some of the @datapythonista suggestions (did not add any examples though).

I was unable to run `validate_docstrings.py` (got a `ModuleNotFoundError: No module named 'pandas._libs.window'` exception). :confused: Did not spend much time trying to fix it, I must admit.

As a side note, did you ever think about releasing that docstring checker as a separate package? I bet many other projects could benefit from it. :blush:

WillAyd:
Looks like you need to build the C extensions

```sh
python setup.py build_ext --inplace -j 4
```

Not aware of a conversation to have as a separate package but @gfyoung might have some thoughts on that as well

Peque:
@WillAyd Yeah, I was only missing a C++ compiler.

Validated the docstrings. No errors found. :+1:

WillAyd (changes requested):

pandas/core/arrays/datetimes.py
```diff
@@ -795,13 +795,14 @@ def month_name(self, locale=None):
 
         Parameters
         ----------
-        locale : string, default None (English locale)
-            locale determining the language in which to return the month name
+        locale : str
+            Locale determining the language in which to return the month name.
+            Default is `None` (English locale).
```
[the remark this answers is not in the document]
datapythonista: Sorry my previous comment was misleading, let me add more detail.

In the first line, after the name of the parameter and the colon, we have the (Python) type or types, and then the default value, and nothing else (e.g. `param_name : int or str, default 0`). In case the default value is `None` and this `None` is just a flag to indicate something, we use instead of `default None` the word `optional`. As a example, if `fillna(value=None)` means that the `None` is the value used to impute, we will still use `default None`. If `None` meant that we won't impute anything, we  would have `optional` instead.

What we need to have in this case is that `None` means that 'English locale' will be used. But anything except the format mentioned before should go in the description. Besides consistency, the reason is that at some point it'd be nice to be able to parse all the information in the line of the type. And if we have explanations there, they'll be a problem.

You have more documentation here: https://python-sprints.github.io/pandas/guide/pandas_docstring.html#section-3-parameters

pandas/core/arrays/datetimes.py
```diff
@@ -821,13 +822,14 @@ def day_name(self, locale=None):
...
 
         Returns
         -------
-        month_names : Index
-            Index of day names
+        Index
+            Index of day names.
 
         .. versionadded:: 0.23.0
```
datapythonista: I think the `versionadded` would be more useful / standard after the description, not here at the end. @WillAyd what do you think?

datapythonista (commented):
Thanks for the changes @Peque, looking better.

Besides those comments, do you think you can add a simple example? I think it'd add a lot of value to let people understand easier what the method does.

Peque:
I uploaded some changes to better match the preferred style and add simple examples (this time I read the style guide :innocent:).

PS: [in the style guide that you shared with me](https://python-sprints.github.io/pandas/guide/pandas_docstring.html), the first example is missing dots after descriptions. That may be misleading for the lazy (like me) that only read the first example and "infer" the rules. :stuck_out_tongue_winking_eye:

pandas/core/arrays/datetimes.py
```diff
@@ -793,17 +793,24 @@ def month_name(self, locale=None):
...
 
+        .. versionadded:: 0.23.0
+
         Parameters
         ----------
-        locale : string, default None (English locale)
-            locale determining the language in which to return the month name
+        locale : str, optional
+            Locale determining the language in which to return the month name.
+            Default is English locale.
```
WillAyd: Where is pandas making the default English? Haven't looked in detail but assumed this would be system dependent

Peque: I would expect that to be system independent (when programming, I'd rather have predictable results).

Code is implemented in `get_date_name_field()`:

https://github.com/pandas-dev/pandas/blob/86e8f23be6d8496cb39ee836b5b02f5c91fda0ba/pandas/_libs/tslibs/fields.pyx#L102-L103
https://github.com/pandas-dev/pandas/blob/86e8f23be6d8496cb39ee836b5b02f5c91fda0ba/pandas/_libs/tslibs/fields.pyx#L117-L118

`DAYS_FULL` and `MONTHS_FULL` are hard-coded in Pandas:

https://github.com/pandas-dev/pandas/blob/86e8f23be6d8496cb39ee836b5b02f5c91fda0ba/pandas/_libs/tslibs/ccalendar.pyx#L39-L48

WillAyd (approved):
lgtm - @datapythonista merge away

pandas/core/arrays/datetimes.py
mroeschke: The idea was for `day_name` to replace the depreciated `weekday_name` which defaulted to hardcoded english names (without localization support).

jreback: agree this should say the default is based on the locale

jreback (changes requested):

Peque:
@jreback But it is not based on the locale...

pandas/core/arrays/datetimes.py
Peque: @jreback As you can see in the code snippets above, it is not based on the locale. The default is always English.

jreback: sure it does, just pass a locale name

Peque: @jreback Sorry, I do not understand your last comment. Are you still saying when no locale is passed as parameter then the default behavior is to use the system locale instead of always English? Or are you saying the default locale (English) should be specified instead of `locale=None` in the method declaration?

I am confused because it seems you first asked to change the docstring to state that the default was the system locale instead of English (which is not true).

jreback: why don't you try both ways and see

Peque: @jreback Which both ways? Could you be more explicit about what you are trying to say? Explicit is better than implicit and words are free. :stuck_out_tongue_winking_eye: 

I tried changing the system locale and verified the default behavior was to use English. Maybe I did something wrong, in which case could you provide a better way to verify it? Also, I searched through the Pandas code and it seemed to me (as shown above), that the default is English, not system locale.

Peque: @jreback Yeah, the tests seem to verify the same behavior: when locale is `None`, the default is always English, not system locale. I still do not get what you are trying to tell me then... :cry:

jreback: can you show here what you are running exactly

Peque: @jreback I do not think the code I am running is relevant here. Unless you refer to a trivial example like the one I added in the docstring (but why would you need that?):

```python
idx = pd.DatetimeIndex(start='2018-09-02', freq='D', periods=3)
idx.day_name()
```

Can you explain why you still say the default when `locale=None` is based on the system's locale? I checked the tests, the code and the behavior and everything seems to prove otherwise: the default when `locale=None` is English, and hence the proposed docstring. Did you check anything?

pandas/core/arrays/datetimes.py
```diff
@@ -793,17 +793,24 @@ def month_name(self, locale=None):
...
-            Index of month names
+        Index
+            Index of month names.
 
-        .. versionadded:: 0.23.0
+        Examples
+        --------
+        >>> idx = pd.DatetimeIndex(start='2018-09', freq='M', periods=3)
+        >>> idx.month_name()
+        Index(['September', 'October', 'November'], dtype='object')
```
datapythonista: If you don't mind, can you first show `idx` (i.e. `>>> idx` after creating the index). Also, if we use '2018-01' it'd make the example more obvious.

pandas/core/arrays/datetimes.py
```diff
@@ -819,17 +826,24 @@ def day_name(self, locale=None):
...
-            Index of day names
+        Index
+            Index of day names.
 
-        .. versionadded:: 0.23.0
+        Examples
+        --------
+        >>> idx = pd.DatetimeIndex(start='2018-09-02', freq='D', periods=3)
+        >>> idx.day_name()
+        Index(['Sunday', 'Monday', 'Tuesday'], dtype='object')
```
datapythonista: Same as before.

datapythonista (approved):
lgtm, added couple of suggestions that I think could make the examples a bit simpler.

Regarding the locale I think you're just misunderstanding each other.

pandas/core/arrays/datetimes.py
Peque: Sure, will change it and upload it later today. :+1:

About the locale discussion, I too think it must be a misunderstanding. With an average of 10 words per comment, I am having a hard time trying to understand what he tries to say. :sweat_smile: 

If you understand both of us, could you help us? :joy:

Peque:
@datapythonista Updated the PR showing the `idx` values and using a different `start`.

pandas/core/arrays/datetimes.py
Peque: @jreback If you tell me exactly what changes to the current docstring you have on your head, I may better understand what you are saying (it seems other reviewers are okay with it). :heart:

mroeschke: @Peque @jreback The default locale is English as `day_name` is meant to replace `weekday_name` which defaulted to english. So the docstring here is correct.

@Peque feel free to open up a new issue suggesting that `day_name` and `month_name` should default to system locale. This would be an API change.

Peque: @mroeschke Yeah, but I do prefer the current behavior, so I would rather not open that issue (and note I am not a native English speaker nor live in an English country).

When programming, I prefer everything in English by default. :stuck_out_tongue_winking_eye:

mroeschke: Gotcha @Peque. I hold a similar opinion for API transition sake.

Therefore, I think there's no action with this docstring here as far as I'm concerned.

jreback (approved):

jreback:
thanks @Peque

PR #22547, opened 2018-08-30
CLN: use 'codes' rather than 'values' internally in Categorical

topper-123:
In some places a variable has the name ``values``, where ``codes`` would be more logical, as we're dealing with codes.

This makes it a bit easier to understand what we're looking at when reading IMO,

### DIFF
--- a/pandas/core/indexes/category.py
+++ b/pandas/core/indexes/category.py
@@ -460,9 +460,7 @@ class CategoricalIndex(Index, accessor.PandasDelegate):
             other = self._na_value
         values = np.where(cond, self.values, other)
 
-        cat = Categorical(values,
-                          categories=self.categories,
-                          ordered=self.ordered)
+        cat = Categorical(values, dtype=self.dtype)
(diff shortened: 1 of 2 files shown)


pandas/core/indexes/category.py
```diff
@@ -460,9 +460,7 @@ def where(self, cond, other=None):
             other = self._na_value
         values = np.where(cond, self.values, other)
 
-        cat = Categorical(values,
-                          categories=self.categories,
-                          ordered=self.ordered)
+        cat = Categorical(values, dtype=self.dtype)
```
alimcmaster1: Hey- why can we get rid of the categories and ordered params out of interest?

topper-123: The dtype is a ``CategoricalDtype``, which contains attributes ``categories`` and ``ordered``. So we're  not really removing them, only passinf the dtype around instead, which is clearer IMO.

jbrockmendel:
+1 on moving away from ‘values’, too many variants of that exist at the moment

jreback:
thanks!

PR #22549, opened 2018-08-31
Fix incorrect DTI/TDI indexing; warn before dropping tzinfo

### LINKED ISSUE #20464: BUG: DatetimeIndex allows indexing with Timedelta and likewise for TimedeltaIndex/Timestamp
A `Series` or `DataFrame` with `DatetimeIndex` can currently be indexed with a `Timedelta`:
```python-traceback
In [2]: s = pd.Series(list('abc'), pd.date_range(0, periods=3))

In [3]: s
Out[3]:
1970-01-01    a
1970-01-02    b
1970-01-03    c
Freq: D, dtype: object

In [4]: s.loc[pd.Timedelta(0)]
Out[4]: 'a'

In [5]: s.loc[pd.Timedelta('0 days'):pd.Timedelta('1 day')]
Out[5]:
1970-01-01    a
1970-01-02    b
Freq: D, dtype: object
```

Likewise for `TimedeltaIndex` and `Timestamp`:
```python-traceback
In [6]: s = pd.Series(list('abc'), pd.timedelta_range(0, periods=3))

In [7]: s
Out[7]:
0 days    a
1 days    b
2 days    c
Freq: D, dtype: object

In [8]: s.loc[pd.Timestamp(0)]
Out[8]: 'a'

In [9]: s.loc[pd.Timestamp('1970-01-01'):pd.Timestamp('1970-01-02')]
Out[9]:
0 days    a
1 days    b
Freq: D, dtype: object
```
#### Problem description
Matches are being returned for labels that are not in the index.

Similar to #20432, but doesn't seem as widespread.

#### Expected Output
I'd expect a `KeyError` to be raised.

<details>

INSTALLED VERSIONS
------------------
commit: 85817a72897be808fb6d311f65fc6031e0216603
python: 3.6.1.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.23.0.dev0+667.g85817a7
pytest: 3.1.2
pip: 9.0.1
setuptools: 27.2.0
Cython: 0.25.2
numpy: 1.13.3
scipy: 1.0.0
pyarrow: 0.6.0
xarray: 0.9.6
IPython: 6.1.0
sphinx: 1.5.6
patsy: 0.4.1
dateutil: 2.6.0
pytz: 2017.2
blosc: None
bottleneck: None
tables: 3.4.2
numexpr: 2.6.4
feather: 0.4.0
matplotlib: 2.0.2
openpyxl: 2.4.8
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 0.9.8
lxml: 3.8.0
bs4: None
html5lib: 0.999
sqlalchemy: 1.1.13
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
fastparquet: 0.1.0
pandas_gbq: None
pandas_datareader: None
</details>

### LINKED ISSUE #21333: to_period is dropping timezone information
```python
localSeries = pd.date_range("2018-7-1", periods=1, freq="30T", tz="Europe/London")
print(localSeries.to_period("30T")[0].start_time)
#Timestamp('2018-07-01 00:00:00')
print(localSeries.to_period("30T")[0].start_time.tz)
#None
print(localSeries.to_period("30T")[0].end_time.tz)
#None
print(localSeries[0].tz)
#<DstTzInfo 'Europe/London' BST+1:00:00 DST>
```

While this may be useful for monthly/quarterly granularity data, it causes a problem for half-hourly data or similar around clock change, because the local times are used but the time-zone knowledge is dropped without warning.

jbrockmendel:
Slight hodge-podge.  

#21333 is about `to_period` method on DatetimeIndex silently dropping timezone info.  This PR makes `DatetimeIndex.to_period` and `Timestamp.to_period` issue a warning when losing timezone info.

#20464 is entirely unrelated, notes that `pd.date_range('1970-01-01', period=2).get_loc(pd.Timedelta(0))` returns `0` instead of raising; ditto for indexing a TDI with a Timestamp.

- [x] closes #21333
- [x] closes #20464
- [x] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/pandas/_libs/tslibs/timestamps.pyx
+++ b/pandas/_libs/tslibs/timestamps.pyx
@@ -737,6 +737,11 @@ class Timestamp(_Timestamp):
         """
         from pandas import Period
 
+        if self.tz is not None:
+            # GH#21333
+            warnings.warn("Converting to Period representation will "
+                          "drop timezone information.")
+
         if freq is None:
             freq = self.freq
 
--- a/pandas/_libs/tslibs/timestamps.pyx
+++ b/pandas/_libs/tslibs/timestamps.pyx
@@ -740,7 +740,8 @@ class Timestamp(_Timestamp):
         if self.tz is not None:
             # GH#21333
             warnings.warn("Converting to Period representation will "
-                          "drop timezone information.")
+                          "drop timezone information.",
+                          UserWarning)
 
         if freq is None:
             freq = self.freq
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -2,7 +2,7 @@
 from __future__ import division
 import operator
 import warnings
-from datetime import time, datetime
+from datetime import time, datetime, timedelta
 
 import numpy as np
 from pytz import utc
@@ -727,6 +727,10 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
         """
         from pandas.core.indexes.period import PeriodIndex
 
+        if self.tz is not None:
+            warnings.warn("Converting to PeriodIndex representation will "
+                          "drop timezone information.")
+
         if freq is None:
             freq = self.freqstr or self.inferred_freq
 
@@ -737,7 +741,7 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
 
             freq = get_period_alias(freq)
 
-        return PeriodIndex(self.values, name=self.name, freq=freq, tz=self.tz)
+        return PeriodIndex(self.values, name=self.name, freq=freq)
 
     def snap(self, freq='S'):
         """
@@ -1200,6 +1204,12 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
             # needed to localize naive datetimes
             key = Timestamp(key, tz=self.tz)
             return Index.get_loc(self, key, method, tolerance)
+
+        if isinstance(key, timedelta):
+            # GH#20464
+            raise TypeError("Cannot index {cls} with {other}"
+                            .format(cls=type(self).__name__,
+                                    other=type(key).__name__))
 
         if isinstance(key, time):
             if method is not None:
--- a/pandas/core/indexes/datetimes.py
+++ b/pandas/core/indexes/datetimes.py
@@ -729,7 +729,7 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
 
         if self.tz is not None:
             warnings.warn("Converting to PeriodIndex representation will "
-                          "drop timezone information.")
+                          "drop timezone information.", UserWarning)
 
         if freq is None:
             freq = self.freqstr or self.inferred_freq
@@ -1205,7 +1205,7 @@ class DatetimeIndex(DatetimeArrayMixin, DatelikeOps, TimelikeOps,
             key = Timestamp(key, tz=self.tz)
             return Index.get_loc(self, key, method, tolerance)
 
-        if isinstance(key, timedelta):
+        elif isinstance(key, timedelta):
             # GH#20464
             raise TypeError("Cannot index {cls} with {other}"
                             .format(cls=type(self).__name__,
--- a/pandas/core/indexes/timedeltas.py
+++ b/pandas/core/indexes/timedeltas.py
@@ -489,7 +489,10 @@ class TimedeltaIndex(TimedeltaArrayMixin, DatetimeIndexOpsMixin,
         loc : int
         """
         if is_list_like(key) or (isinstance(key, datetime) and key is not NaT):
-            # GH#20464 for datetime case
+            # GH#20464 datetime check here is to ensure we don't allow
+            #   datetime objects to be incorrectly treated as timedelta
+            #   objects; NaT is a special case because it plays a double role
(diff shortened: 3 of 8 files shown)


pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -737,6 +737,11 @@ class Timestamp(_Timestamp):
         """
         from pandas import Period
 
+        if self.tz is not None:
+            # GH#21333
+            warnings.warn("Converting to Period representation will "
+                          "drop timezone information.")
```
gfyoung: Warning type and stack level?

gfyoung: This questions applies to everywhere where you placed warnings.

jbrockmendel: No idea what warning type to use.  Suggestions?

As to stack level, I tried a bunch to get that to work with tm.assert_produces_warning and eventually threw in the towel.

pandas/core/indexes/datetimes.py
```diff
@@ -1201,6 +1205,12 @@ def get_loc(self, key, method=None, tolerance=None):
             key = Timestamp(key, tz=self.tz)
             return Index.get_loc(self, key, method, tolerance)
 
+        if isinstance(key, timedelta):
```
jreback: if/elif here?

pandas/core/indexes/timedeltas.py
```diff
@@ -487,7 +488,8 @@ def get_loc(self, key, method=None, tolerance=None):
         -------
         loc : int
         """
-        if is_list_like(key):
+        if is_list_like(key) or (isinstance(key, datetime) and key is not NaT):
```
jreback: not sure if we use isna else for NaT checking?

pandas/core/indexes/timedeltas.py
```diff
@@ -487,7 +488,8 @@ def get_loc(self, key, method=None, tolerance=None):
...
+            # GH#20464 for datetime case
```
jreback: make this more explicit, meaning datetime dtype for all-NaT

pandas/tests/indexes/datetimes/test_astype.py
```diff
@@ -266,81 +270,95 @@ def test_to_period_tz_pytz(self):
...
         xp = date_range('1/1/2000', '4/1/2000').to_period()
 
         ts = date_range('1/1/2000', '4/1/2000', tz=pytz.timezone('US/Eastern'))
 
-        result = ts.to_period()[0]
-        expected = ts[0].to_period()
+        with tm.assert_produces_warning(UserWarning):
+            # warning that timezone info will be lost
+            result = ts.to_period()[0]
+            expected = ts[0].to_period()
```
jreback: in future PR these should be parameterized

pandas/tests/indexes/timedeltas/test_indexing.py
```diff
@@ -41,6 +41,14 @@ def test_getitem(self):
             tm.assert_index_equal(result, expected)
             assert result.freq == expected.freq
 
+    def test_timestamp_invalid_key(self):
```
jreback: can you parameterize over a datetime as well

pandas/tests/indexes/timedeltas/test_indexing.py
```diff
@@ -41,6 +41,14 @@ def test_getitem(self):
...
+        # GH#20464
+        tdi = pd.timedelta_range(0, periods=10)
```
jreback: do we have a test that indexes with NaT? (both Timedelta and Datetime dtype)

jreback (changes requested):

pandas/core/indexes/timedeltas.py
jbrockmendel: I think in this context `isna` is less clear.  Since there is a specific na-like object we are catching here, we should be explicit about it.

pandas/tests/indexes/timedeltas/test_indexing.py
jbrockmendel: We have one for timedelta, just added one for datetime

pandas/_libs/tslibs/timestamps.pyx
jorisvandenbossche: Since the warnings are raised in direct user-api methods, normally putting a `stacklevel=2` should do the correct thing. 
What did you not get working in the tests? (which kind of code sample)

jbrockmendel: > What did you not get working in the tests?

IIRC the problem is that the affected code paths issues both the new `FutureWarning` and in some cases also a `PerformanceWarning`.  `tm.assert_produces_warning` doesn't support multiple expected warnings, and my attempt to modify it led to tests failing the stacklevel checks.

jbrockmendel:
Looks like circleCI is a hypothesis timeout

pandas/core/indexes/timedeltas.py
jreback: ok this comment looks good

jreback:
lgtm. rebase after @mroeschke CalendarDay change. ping on green.

jreback (approved):

jreback:
thanks!

PR #22552, opened 2018-08-31
Correct assert_frame_equal doc string

pablojim:
Correct default values in assert_frame_equal doc string

### DIFF
--- a/pandas/util/testing.py
+++ b/pandas/util/testing.py
@@ -1317,10 +1317,10 @@ def assert_frame_equal(left, right, check_dtype=True,
         Second frame to compare.
     check_dtype : bool, default True
         Whether to check the DataFrame dtype is identical.
-    check_index_type : bool / string {'equiv'}, default 'equiv'
+    check_index_type : {'equiv'} or bool, default 'equiv'
         Whether to check the Index class, dtype and inferred_type
         are identical.
-    check_column_type : bool / string {'equiv'}, default 'equiv'
+    check_column_type : {'equiv'} or bool, default 'equiv'
         Whether to check the columns class, dtype and inferred_type
         are identical.
     check_frame_type : bool, default True
--- a/pandas/util/testing.py
+++ b/pandas/util/testing.py
@@ -1309,12 +1309,17 @@ def assert_frame_equal(left, right, check_dtype=True,
     """
     Check that left and right DataFrame are equal.
 
+    This function is intended to compare two DataFrames and output any
+    differences. Is is mostly intended for use in unit tests.
+    Additional parameters allow varying the strictness of the
+    equality checks performed.
+
     Parameters
     ----------
     left : DataFrame
-        First frame to compare.
+        First DataFrame to compare.
     right : DataFrame
-        Second frame to compare.
+        Second DataFrame to compare.
     check_dtype : bool, default True
         Whether to check the DataFrame dtype is identical.
     check_index_type : {'equiv'} or bool, default 'equiv'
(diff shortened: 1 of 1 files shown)


pandas/util/testing.py
```diff
@@ -1314,13 +1314,13 @@ def assert_frame_equal(left, right, check_dtype=True,
     right : DataFrame
     check_dtype : bool, default True
         Whether to check the DataFrame dtype is identical.
-    check_index_type : bool / string {'equiv'}, default False
+    check_index_type : bool / string {'equiv'}, default 'equiv'
```
WillAyd: The type here seems rather verbose. Wondering if it's better served as just `bool or 'equiv'` since 'equiv' appears to be the only allowable string. @datapythonista

WillAyd (commented):
Can you check that this passes the docstring validation script?

pablojim:
@WillAyd 
```################################################################################

Warnings found:
        No extended summary found
        See Also section not found
        No examples section found
Docstring for "pandas.util.testing.assert_frame_equal" correct. :)
```

WillAyd (changes requested):

pandas/util/testing.py
```diff
@@ -1306,33 +1306,36 @@ def assert_frame_equal(left, right, check_dtype=True,
...
-        If int, then specify the digits to compare
+        If int, then specify the digits to compare.
     check_names : bool, default True
         Whether to check that the `names` attribute for both the `index`
         and `column` attributes of the DataFrame is identical, i.e.
 
         * left.index.names == right.index.names
         * left.columns.names == right.columns.names
-
+        .
```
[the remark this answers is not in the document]
datapythonista: it was already in #20298, but clarified it a bit

pandas/util/testing.py
```diff
@@ -1306,33 +1306,36 @@ def assert_frame_equal(left, right, check_dtype=True,
...
     Parameters
     ----------
     left : DataFrame
+        First frame to compare.
     right : DataFrame
+        Second frame to compare.
     check_dtype : bool, default True
         Whether to check the DataFrame dtype is identical.
-    check_index_type : bool / string {'equiv'}, default False
+    check_index_type : bool / string {'equiv'}, default 'equiv'
```
datapythonista: I think what we've been using in these cases is `{'equiv'} or bool`

pandas/util/testing.py
```diff
@@ -1345,10 +1348,10 @@ def assert_frame_equal(left, right, check_dtype=True,
     check_like : bool, default False
         If True, ignore the order of index & columns.
         Note: index labels must match their respective rows
-        (same as in columns) - same labels must be with the same data
+        (same as in columns) - same labels must be with the same data.
     obj : str, default 'DataFrame'
         Specify object name being compared, internally used to show appropriate
-        assertion message
+        assertion message.
     """
```
datapythonista: I think it's worth adding a `See Also` section, with `DataFrame.equals`, `testing.assert_series_equal` and may be others.

pandas/util/testing.py
pablojim: IMO that is more confusing. e.g. it might imply I pass a set. I'm not sure it's worth a special case for when there is only one possible string value.

pandas/util/testing.py
pablojim: Added

pandas/util/testing.py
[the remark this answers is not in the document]
pablojim: removed

pandas/util/testing.py
datapythonista: what is your suggestion, just `'equiv' or bool, default 'equiv'` or something else? Besides being more consistent for the user, using the curly brackets in all cases would simplify parsing the types and adding validation and extracting stats. But if you are strongly in favor of not using them, I'm happy to merge this with it now, and see later on what's best when we implement that validation.

pandas/util/testing.py
```diff
@@ -1345,10 +1347,15 @@ def assert_frame_equal(left, right, check_dtype=True,
...
+        (same as in columns) - same labels must be with the same data.
     obj : str, default 'DataFrame'
         Specify object name being compared, internally used to show appropriate
-        assertion message
+        assertion message.
+
+    See Also
+    --------
+    assert_series_equal: equivalent method for asserting Series equality
+    DataFrame.equals: check DataFrame equality
```
datapythonista: Do you think it makes sense adding a example?

pandas/util/testing.py
pablojim: I think as it is above is good. It's maybe a little verbose but is very clear.

```check_index_type : bool / string {'equiv'}, default 'equiv'```

Happy to revisit if a standard emerges.

pandas/util/testing.py
pablojim: I'm not sure an example is particularly useful in this case.

pandas/util/testing.py
datapythonista: You have the standard here: https://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html#parameter-types

If you can use the first format I suggested, that would be great.

pandas/util/testing.py
datapythonista: We've got users of all levels, and things like comparing dataframes with different types are worth showing with examples.

Besides that, can you add a space before the colons in the see also section, capitalize the first letter of the description, and finish the description with a period? If you generate the html `./doc/make.py html --single pandas.util.testing.assert_frame_equal`, I don't think without the space this is being rendered correcty.

datapythonista (changes requested):
There are formatting issues, and I think examples, and also an extended summary that explains that this is mainly used in unit tests, would make the docstring much more useful to users.

pandas/util/testing.py
pablojim: Added examples and corrected formatting as mentioned.

pandas/util/testing.py
pablojim: Thanks. I have used your suggestion.

datapythonista (approved):
lgtm, great changes. Thanks @pablojim

PR #22560, opened 2018-09-01
ENH/BUG: Add is_dst method to DatetimeIndex and Timestamp to solve AmbiguousTimeError

### LINKED ISSUE #18885: AmbiguousTimeError merging two timezone-aware DataFrames with DST change
When merging two DataFrames by a timezone-aware datetime column, if the datetime values doesn't include a DST change, there's no problem:

```python

df1 = pd.DataFrame([pd.to_datetime('2017-10-30 02:00:00+01:00'),
                    pd.to_datetime('2017-10-30 03:00:00+01:00'),
                    pd.to_datetime('2017-10-30 04:00:00+01:00')],columns=['date'])
df1['date'] = df1['date'].dt.tz_localize('UTC').dt.tz_convert('Europe/Madrid')
df1['value'] = 1

df2 = pd.DataFrame([pd.to_datetime('2017-10-30 04:00:00+01:00'),
                    pd.to_datetime('2017-10-30 05:00:00+01:00'),
                    pd.to_datetime('2017-10-30 06:00:00+01:00')],columns=['date'])
df2['date'] = df2['date'].dt.tz_localize('UTC').dt.tz_convert('Europe/Madrid')
df2['value'] = 2

pd.merge(df1, df2, how='outer', on='date')

```

### Output

```
                       date  value_x  value_y
0 2017-10-30 02:00:00+01:00      1.0      NaN
1 2017-10-30 03:00:00+01:00      1.0      NaN
2 2017-10-30 04:00:00+01:00      1.0      2.0
3 2017-10-30 05:00:00+01:00      NaN      2.0
4 2017-10-30 06:00:00+01:00      NaN      2.0
```

This is correct. But if the datetime values include a date with DST change, we get an AmbiguousTimeError exception:

```python
df1 = pd.DataFrame([pd.to_datetime('2017-10-29 02:00:00+02:00'),
                    pd.to_datetime('2017-10-29 02:00:00+01:00'),
                    pd.to_datetime('2017-10-29 03:00:00+01:00')],columns=['date'])
df1['date'] = df1['date'].dt.tz_localize('UTC').dt.tz_convert('Europe/Madrid')
df1['value'] = 1

df2 = pd.DataFrame([pd.to_datetime('2017-10-29 03:00:00+01:00'),
                    pd.to_datetime('2017-10-29 04:00:00+01:00'),
                    pd.to_datetime('2017-10-29 05:00:00+01:00')],columns=['date'])
df2['date'] = df2['date'].dt.tz_localize('UTC').dt.tz_convert('Europe/Madrid')
df2['value'] = 2

pd.merge(df1, df2, how='outer', on='date')
```

### Expected output

```
                       date  value_x   value_y
0 2017-10-29 02:00:00+02:00      1.0       NaN
1 2017-10-29 02:00:00+01:00      1.0       NaN
2 2017-10-29 03:00:00+01:00      1.0       2.0
3 2017-10-29 04:00:00+01:00      NaN       2.0
4 2017-10-29 05:00:00+01:00      NaN       2.0
```

### Actual output

```
  File "(...)/venv/lib/python3.6/site-packages/pandas/core/reshape/merge.py", line 58, in merge
    return op.get_result()
  File "(...)//venv/lib/python3.6/site-packages/pandas/core/reshape/merge.py", line 604, in get_result
    self._maybe_add_join_keys(result, left_indexer, right_indexer)
  File "(...)//venv/lib/python3.6/site-packages/pandas/core/reshape/merge.py", line 715, in _maybe_add_join_keys
    key_col = Index(lvals).where(~mask, rvals)
  File "(...)//venv/lib/python3.6/site-packages/pandas/core/indexes/datetimelike.py", line 809, in where
    result = self._ensure_localized(result)
  File "(...)//venv/lib/python3.6/site-packages/pandas/core/indexes/datetimelike.py", line 230, in _ensure_localized
    result = result.tz_localize(self.tz)
  File "(...)//venv/lib/python3.6/site-packages/pandas/util/_decorators.py", line 118, in wrapper
    return func(*args, **kwargs)
  File "(...)//venv/lib/python3.6/site-packages/pandas/core/indexes/datetimes.py", line 1858, in tz_localize
    errors=errors)
  File "pandas/_libs/tslib.pyx", line 3593, in pandas._libs.tslib.tz_localize_to_utc
pytz.exceptions.AmbiguousTimeError: Cannot infer dst time from Timestamp('2017-10-29 02:00:00'), try using the 'ambiguous' argument
```

<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.2.final.0
python-bits: 64
OS: Darwin
OS-release: 17.3.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: es_ES.UTF-8
LOCALE: es_ES.UTF-8

pandas: 0.21.1
pytest: 3.2.5
pip: 9.0.1
setuptools: 36.8.0
Cython: None
numpy: 1.13.3
scipy: None
pyarrow: None
xarray: None
IPython: None
sphinx: 1.5.3
patsy: None
dateutil: 2.6.1
pytz: 2017.3
blosc: None
bottleneck: None
tables: None
numexpr: None
feather: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: 0.9.6
lxml: None
bs4: None
html5lib: None
sqlalchemy: None
pymysql: None
psycopg2: None
jinja2: 2.9.6
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None

</details>

### LINKED ISSUE #18946: AmbiguousTimeError in floor() operation
This is probably related to #18885 

Given a DataFrame with a DST change on it:

```python
df1=pd.DataFrame([pd.to_datetime('2017-10-29 02:00:00+02:00'),pd.to_datetime('2017-10-29 02:00:00+01:00'),pd.to_datetime('2017-10-29 03:00:00+01:00')],columns=['date'])
df1['date'] = df1['date'].dt.tz_localize('UTC').dt.tz_convert('Europe/Madrid')
df1['value'] = 1
```

When we try to do a `floor` operation, we get an AmbiguousTimeError exception:

```python
df1.date.dt.floor('H')
```

### Expected output

```
0   2017-10-29 02:00:00+02:00
1   2017-10-29 02:00:00+01:00
2   2017-10-29 03:00:00+01:00
```

### Actual output

```
  File "(...)/venv/lib/python3.6/site-packages/pandas/core/accessor.py", line 115, in f
    return self._delegate_method(name, *args, **kwargs)
  File "(...)/venv/lib/python3.6/site-packages/pandas/core/indexes/accessors.py", line 131, in _delegate_method
    result = method(*args, **kwargs)
  File "(...)/venv/lib/python3.6/site-packages/pandas/core/indexes/datetimelike.py", line 118, in floor
    return self._round(freq, np.floor)
  File "(...)/venv/lib/python3.6/site-packages/pandas/core/indexes/datetimelike.py", line 110, in _round
    self._shallow_copy(result, **attribs))
  File "(...)/venv/lib/python3.6/site-packages/pandas/core/indexes/datetimelike.py", line 230, in _ensure_localized
    result = result.tz_localize(self.tz)
  File "(...)/venv/lib/python3.6/site-packages/pandas/util/_decorators.py", line 118, in wrapper
    return func(*args, **kwargs)
  File "(...)/venv/lib/python3.6/site-packages/pandas/core/indexes/datetimes.py", line 1858, in tz_localize
    errors=errors)
  File "pandas/_libs/tslib.pyx", line 3593, in pandas._libs.tslib.tz_localize_to_utc
pytz.exceptions.AmbiguousTimeError: Cannot infer dst time from Timestamp('2017-10-29 02:00:00'), try using the 'ambiguous' argument
```

mroeschke:
- [x] closes #18885
- [x] closes #18946
- [x] tests added / passed
- [x] whatsnew entry

The issues above required a new method to keep track if each timestamp in the `DatetimeIndex` was previously in daylight savings time in order to pass into `tz_localize(..., ambiguous=...)`

Therefore, added a `is_dst` method to `DatetimeIndex` and `Timestamp`.

### DIFF
--- a/pandas/_libs/tslibs/nattype.pyx
+++ b/pandas/_libs/tslibs/nattype.pyx
@@ -260,6 +260,20 @@ class NaTType(_NaT):
     def is_year_end(self):
         return False
 
+    def is_dst(self):
+        """
+        Returns a boolean indicating if the Timestamp is in daylight savings
+        time. Naive timestamps are considered not to be in daylight savings
+        time.
+
+        Returns
+        -------
+        Boolean
+            True if the Timestamp is in daylight savings time
+            False if the Timestamp is naive or not in daylight savings time
+        """
+        return False
+
     def __rdiv__(self, other):
         return _nat_rdivide_op(self, other)
 
@@ -353,6 +367,7 @@ class NaTType(_NaT):
     strptime = _make_error_func('strptime', datetime)
     strftime = _make_error_func('strftime', datetime)
     isocalendar = _make_error_func('isocalendar', datetime)
+    dst = _make_error_func('dst', datetime)
     ctime = _make_error_func('ctime', datetime)
     time = _make_error_func('time', datetime)
     toordinal = _make_error_func('toordinal', datetime)
--- a/pandas/_libs/tslibs/nattype.pyx
+++ b/pandas/_libs/tslibs/nattype.pyx
@@ -353,7 +353,6 @@ class NaTType(_NaT):
     strptime = _make_error_func('strptime', datetime)
     strftime = _make_error_func('strftime', datetime)
     isocalendar = _make_error_func('isocalendar', datetime)
-    dst = _make_error_func('dst', datetime)
     ctime = _make_error_func('ctime', datetime)
     time = _make_error_func('time', datetime)
     toordinal = _make_error_func('toordinal', datetime)
--- a/pandas/_libs/tslibs/timezones.pyx
+++ b/pandas/_libs/tslibs/timezones.pyx
@@ -187,9 +187,10 @@ cdef object get_utc_trans_times_from_dateutil_tz(object tz):
     return new_trans
 
 
-cdef int64_t[:] unbox_utcoffsets(object transinfo, dst):
+cdef int64_t[:] unbox_utcoffsets(object transinfo, bint dst):
     """
-    Unpack the offset information from pytz timezone objects
+    Unpack the offset information from the _transition_info attribute of pytz
+    timezones
 
     Parameters
     ----------
@@ -206,9 +207,10 @@ cdef int64_t[:] unbox_utcoffsets(object transinfo, dst):
 
     sz = len(transinfo)
     arr = np.empty(sz, dtype='i8')
-    key = int(dst)
     for i in range(sz):
-        arr[i] = int(transinfo[i][key].total_seconds()) * 1000000000
+        # If dst == True, extract the DST shift in nanoseconds
+        # If dst == False, extract the UTC offset in nanoseconds
+        arr[i] = int(transinfo[i][dst].total_seconds()) * 1000000000
 
     return arr
 
@@ -217,15 +219,15 @@ cdef int64_t[:] unbox_utcoffsets(object transinfo, dst):
 # Daylight Savings
 
 
-cdef object get_dst_info(object tz, dst):
+cdef object get_dst_info(object tz, bint dst):
     """
     Return DST info from a timezone
 
     Parameters
     ----------
     tz : object
-        timezone
-    dst : bool
+        timezone object
+    dst : bint
         True returns the DST specific offset and will NOT store the results in
         dst_cache. dst_cache is reserved for caching UTC offsets.
         False returns the UTC offset
@@ -333,10 +335,12 @@ def is_dst(int64_t[:] values, object tz):
         return result
     transitions, offsets, typ = get_dst_info(tz, True)
     offsets = np.array(offsets)
+
     # Fixed timezone offsets do not have DST transitions
     if typ not in {'pytz', 'dateutil'}:
         return result
     positions = transitions.searchsorted(values, side='right') - 1
+
     # DST has nonzero offset
     result = offsets[positions] != 0
     return result
--- a/pandas/_libs/tslibs/timezones.pyx
+++ b/pandas/_libs/tslibs/timezones.pyx
@@ -108,7 +108,8 @@ def _p_tz_cache_key(tz):
     return tz_cache_key(tz)
 
 
-# Timezone data caches, key is the pytz string or dateutil file name.
+# Timezone data (UTC offset) caches
+# key is the pytz string or dateutil file name.
 dst_cache = {}
 
 
@@ -225,7 +226,8 @@ cdef object get_dst_info(object tz, dst):
     tz : object
         timezone
     dst : bool
-        True returns the DST specific offset
+        True returns the DST specific offset and will NOT store the results in
+        dst_cache. dst_cache is reserved for caching UTC offsets.
         False returns the UTC offset
         Specific for pytz timezones only
 
@@ -246,7 +248,7 @@ cdef object get_dst_info(object tz, dst):
                 np.array([num], dtype=np.int64),
                 None)
 
-    if cache_key not in dst_cache:
+    if cache_key not in dst_cache or dst:
         if treat_tz_as_pytz(tz):
             trans = np.array(tz._utc_transition_times, dtype='M8[ns]')
             trans = trans.view('i8')
@@ -298,12 +300,14 @@ cdef object get_dst_info(object tz, dst):
             deltas = np.array([num], dtype=np.int64)
             typ = 'static'
 
+        if dst:
+            return trans, deltas, typ
         dst_cache[cache_key] = (trans, deltas, typ)
 
     return dst_cache[cache_key]
 
 
-def _is_dst(int64_t[:] values, object tz):
+def is_dst(int64_t[:] values, object tz):
     """
     Return a boolean array indicating whether each epoch timestamp is in
     daylight savings time with respect with the passed timezone.
@@ -322,9 +326,6 @@ def _is_dst(int64_t[:] values, object tz):
     """
     cdef:
         Py_ssize_t n = len(values)
-        # Cython boolean memoryviews are not supported yet
-        # https://github.com/cython/cython/issues/2204
-        # bint[:] result
         object typ
 
     result = np.zeros(n, dtype=bool)
@@ -335,9 +336,9 @@ def _is_dst(int64_t[:] values, object tz):
     # Fixed timezone offsets do not have DST transitions
     if typ not in {'pytz', 'dateutil'}:
         return result
-    positions = transitions.searchsorted(values, side='right')
-    # DST has 0 offset
-    result = offsets[positions] == 0
+    positions = transitions.searchsorted(values, side='right') - 1
+    # DST has nonzero offset
+    result = offsets[positions] != 0
     return result
 
 
--- a/pandas/_libs/tslibs/timezones.pyx
+++ b/pandas/_libs/tslibs/timezones.pyx
@@ -305,9 +305,8 @@ cdef object get_dst_info(object tz, dst):
 
 def _is_dst(int64_t[:] values, object tz):
     """
-    Return a boolean array where True indicates a value that lies in
-    daylight savings time and False indicates a value that does not lie in
-    daylight savings time
+    Return a boolean array indicating whether each epoch timestamp is in
+    daylight savings time with respect with the passed timezone.
 
     Parameters
     ----------
@@ -318,8 +317,8 @@ def _is_dst(int64_t[:] values, object tz):
 
     Returns
     -------
-    ndarray
-        Booleans
+    ndarray of booleans
+        True indicates daylight savings time
     """
     cdef:
         Py_ssize_t n = len(values)
--- a/pandas/_libs/tslibs/timezones.pyx
+++ b/pandas/_libs/tslibs/timezones.pyx
@@ -186,16 +186,28 @@ cdef object get_utc_trans_times_from_dateutil_tz(object tz):
     return new_trans
 
 
-cdef int64_t[:] unbox_utcoffsets(object transinfo):
+cdef int64_t[:] unbox_utcoffsets(object transinfo, dst):
+    """
+    Unpack the offset information from pytz timezone objects
+
+    Parameters
+    ----------
+    transinfo : list of tuples
+        Each tuple contains (UTC offset, DST offset, tz abbreviation)
+    dst : boolean
+        True returns an array of the DST offsets
+        False returns an array of UTC offsets
+    """
     cdef:
         Py_ssize_t i, sz
         int64_t[:] arr
+        int key
 
     sz = len(transinfo)
     arr = np.empty(sz, dtype='i8')
-
+    key = int(dst)
     for i in range(sz):
-        arr[i] = int(transinfo[i][0].total_seconds()) * 1000000000
+        arr[i] = int(transinfo[i][key].total_seconds()) * 1000000000
 
     return arr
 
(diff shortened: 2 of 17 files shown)


pandas/_libs/tslibs/nattype.pyx
```diff
@@ -353,7 +353,6 @@ class NaTType(_NaT):
     strptime = _make_error_func('strptime', datetime)
     strftime = _make_error_func('strftime', datetime)
     isocalendar = _make_error_func('isocalendar', datetime)
-    dst = _make_error_func('dst', datetime)
```
jbrockmendel: Why removing?

mroeschke:
> Is this related to the “fold” arg added in py3?

Yup fairly similar. The `ambiguous=True/False` argument is essentially similar to the `fold=0/1` argument in py3. 

These `AmbiguousTimeError`s stem from internal routines that transform a timestamp to an ambiguous time. I think it's best we try to return a sensible time based on the original timeseries instead of leaving the user with an error they cannot control.

mroeschke:
Ready for another look @jbrockmendel when you get the chance.

pandas/_libs/tslibs/timezones.pyx
```diff
@@ -186,16 +187,28 @@ cdef object get_utc_trans_times_from_dateutil_tz(object tz):
     return new_trans
 
 
-cdef int64_t[:] unbox_utcoffsets(object transinfo):
+cdef int64_t[:] unbox_utcoffsets(object transinfo, dst):
```
jreback: can you add bint to type dst

pandas/_libs/tslibs/timezones.pyx
```diff
@@ -186,16 +187,28 @@ cdef object get_utc_trans_times_from_dateutil_tz(object tz):
...
+    """
     cdef:
         Py_ssize_t i, sz
         int64_t[:] arr
+        int key
 
     sz = len(transinfo)
     arr = np.empty(sz, dtype='i8')
-
+    key = int(dst)
```
jreback: you may not need this if typing bint

pandas/_libs/tslibs/timezones.pyx
```diff
@@ -186,16 +187,28 @@ cdef object get_utc_trans_times_from_dateutil_tz(object tz):
...
     for i in range(sz):
```
jreback: can you add a comment about what you are extracting

pandas/_libs/tslibs/timezones.pyx
```diff
@@ -204,9 +217,23 @@ cdef int64_t[:] unbox_utcoffsets(object transinfo):
...
+cdef object get_dst_info(object tz, dst):
     """
-    return a tuple of :
+    Return DST info from a timezone
+
+    Parameters
+    ----------
+    tz : object
+        timezone
+    dst : bool
```
jreback: type with bitn

pandas/_libs/tslibs/timezones.pyx
```diff
@@ -204,9 +217,23 @@ cdef int64_t[:] unbox_utcoffsets(object transinfo):
...
-cdef object get_dst_info(object tz):
+cdef object get_dst_info(object tz, dst):
     """
-    return a tuple of :
+    Return DST info from a timezone
+
+    Parameters
+    ----------
+    tz : object
+        timezone
```
jreback: I think we accept only timezone objects and not strings here?

pandas/_libs/tslibs/timezones.pyx
```diff
@@ -273,11 +300,48 @@ cdef object get_dst_info(object tz):
...
+    cdef:
+        Py_ssize_t n = len(values)
+        object typ
+
+    result = np.zeros(n, dtype=bool)
+    if tz is None:
+        return result
+    transitions, offsets, typ = get_dst_info(tz, True)
+    offsets = np.array(offsets)
+    # Fixed timezone offsets do not have DST transitions
```
jreback: new line here

pandas/_libs/tslibs/timezones.pyx
```diff
@@ -273,11 +300,48 @@ cdef object get_dst_info(object tz):
...
+    if typ not in {'pytz', 'dateutil'}:
+        return result
+    positions = transitions.searchsorted(values, side='right') - 1
+    # DST has nonzero offset
```
jreback: same

pandas/tests/indexes/datetimes/test_timezones.py
```diff
@@ -1012,6 +1012,22 @@ def test_iteration_preserves_nanoseconds(self, tz):
         for i, ts in enumerate(index):
             assert ts == index[i]
 
+    @pytest.mark.parametrize('arg, expected_arg', [
+        [[], []],
+        [date_range('2018-11-04', periods=4, freq='H', tz='US/Pacific'),
```
jreback: can you try with a dateutil as well (I think we default to pytz)

pandas/tests/scalar/timestamp/test_timezones.py
```diff
@@ -307,3 +307,14 @@ def test_timestamp_timetz_equivalent_with_datetime_tz(self,
         expected = _datetime.timetz()
 
         assert result == expected
+
+    def test_timestamp_is_dst(self):
+        ts_naive = Timestamp('2018-11-04')
+        assert ts_naive.is_dst() is False
+
+        ts_aware = ts_naive.tz_localize('US/Pacific')
+        assert ts_aware.is_dst() is True
```
jreback: can you test dateutil

jreback (changes requested):

pandas/tests/indexes/datetimes/test_timezones.py
mroeschke: @pganssle Do dateutil timezones provide a way to access the possible daylight savings time shifts like `pytz.timezone._transition_info` does?

```
# I'm interested in the 2nd element of the tuples below
In [4]: tz._transition_info
Out[4]:
[(datetime.timedelta(-1, 58020), datetime.timedelta(0), 'LMT'),
 (datetime.timedelta(-1, 57600), datetime.timedelta(0), 'PST'),
 (datetime.timedelta(-1, 61200), datetime.timedelta(0, 3600), 'PDT'),
...
```
Context: I am trying to create a `DatetimeIndex.is_dst` method that indicates whether a particular timestamp is in daylight savings time. I know it's possible to do `bool(datetime.dst())` on each timestamp of the `DatetimeIndex`, but I am curious if there's a more efficient way determine DST for a dateutil timezone than iterating.

pganssle: @mroeschke Yes and no. Yes in the sense that this is Python and obviously `dateutil` has to store that information *somewhere*. No in the sense that neither `dateutil` *nor* `pytz` exposes that information publicly. Note that `_transition_info` is a private member.

That said:

1. I'm not entirely sure why you want to implement a "fast isdst" - I'm not sure I've ever heard of a case where it was useful to know whether or not a time had or did not have DST.
2. I don't really see what this would buy you - the member you list looks like what you would get from `utcoffset`, `dst` and `tzname`, respectively. How do you expect to make use of this information?

pganssle:
I would recommend against adding an `is_dst` method. If this is for resolving ambiguous times, it would be preferable to use something more akin to a `fold` method, since there are several situations where an ambiguity cannot be resolved by querying DST or not (base offset shifts).

In general DST/not DST is only a proxy for the real information you want in all cases I can think of - the offset, the tzname or which side of an ambiguity you are on. All of those can be independently queried and/or specified without mentioning DST.

jbrockmendel:
> These AmbiguousTimeErrors stem from internal routines that transform a timestamp to an ambiguous time. 

I'm not clear on what problem this is solving.  Can you give an example of the internal routines that produce ambiguous times?

@pganssle and I don't always see eye-to-eye on how to treat public vs private attributes, but in general if he's got a strong opinion, it is worth listening to.

jorisvandenbossche:
I also wanted to mention that we deliberately added an `ambiguous` keyword and deprecated `is/infer_dst` because not all ambiguities are DST issues, if I remember well (https://github.com/pandas-dev/pandas/pull/7963, there was some discussion about it, but I am not very familiar with timezone handling details), so we should maybe consider that issue here as well. But I see that @pganssle in the meantime already noted something similar :)

mroeschke:
@jbrockmendel 
> Can you give an example of the internal routines that produce ambiguous times?

When rounding a tz-aware date time to a particular frequency, essentially the steps are:

* convert the date to a naive timestamp
* round the i8 value to the frequency
* relocalize the i8 value to the timezone

You can't necessarily convert to UTC first because it can affect the rounding (unless I am missing something).

So in #18946, a date that sits on the DST border `2017-10-29 02:00:00+02:00` that's rounded to an hour (`'H'`):

* Convert to `2017-10-29 02:00:00`
* Round to the nearest hour (`2017-10-29 02:00:00`)
* Relocalize to `2017-10-29 02:00:00+02:00` or `2017-10-29 02:00:00+01:00`?

I guess #18946 can be solved by added an `ambiguous` kwarg to `round` also. However the routine above is the cause for other issues as well #18885, but that _may_ be solved by converting to UTC first.

I guess using DST information may be a too hand-wavy solution to solve both #18946 and #18885. I had also though having a `is_dst` would be a nice-to-have enhancement for DatetimeIndexes.

pganssle:
@mroeschke I'm guessing that rounding issues in UTC is mainly at the day-or-higher level, right? Since rounding to the nearest minute or hour will cause no problems.

I think the correct solution here is to introduce the `fold` concept into pandas, and start deprecating `pytz`-isms.

I see a few good ways forward:

1. Write `fold` compatible wrappers for `pytz` and start deprecating `pytz`-isms like `localize` and `normalize`.
2. Write wrappers for `dateutil` time zones that expose some `pytz`-isms (which you can deprecate), and stop using `pytz` entirely.
3. Implement a third time zone library optimized for array types and drop both `pytz` and `dateutil`.

In all cases, I think `fold` semantics is the best way to go.

I am very strongly hoping that within a year I'll be able to get a fast, compiled Rust or C backend for many of `dateutil`'s modules, including `tz`, which should make it a lot cheaper to make bulk function calls (particularly if I expose a C API or similar so you don't have any Python function call overhead), but number 3 might still be the best option since it's likely that supporting something that would work *really* well for array types in `dateutil` *and* work really well for normal Python implementations would be more complicated than just maintaining two separate libraries.

jbrockmendel:
> You can't necessarily convert to UTC first because it can affect the rounding (unless I am missing something).

I think there may be an implicit assumption in here that should be made explicit.  Does the user want the rounding to be based on wall time or actual time?  (My intuition is actual time)

```
ts = pd.Timestamp('2017-10-29 02:59:00+0200', tz='Europe/Madrid')
m = pd.Timedelta(minutes=1)

>>> ts + m
Timestamp('2017-10-29 02:00:00+0100', tz='Europe/Madrid')
>>> ts - 59*m
Timestamp('2017-10-29 02:00:00+0200', tz='Europe/Madrid')
```

If I were rounding `ts` to the nearest hour, it seems clear I'd want the one 1 minute away rather than 59 minutes away, despite the fact they have the same difference in wall time.

Is this intuition not applicable to the case you're considering?

pganssle:
@jbrockmendel I think for time scales of hours or less it's probably actual time, but time scale of days they almost certainly want wall time.

Imagine rounding to midnight UTC, then going back to PST/PDT. No one wants to round to 17:00 sometimes and 16:00 others.

To be honest it's probably always wall time, but on sufficiently small time scales the two are equivalent.

mroeschke:
I agree with @pganssle and would imagine users would want to round based on wall time in general.

Thanks for the feedback guys. I'm convinced I need to surface a `fold` (or `ambiguous` in pandas terms) argument in `round` then. Closing.

PR #22562, opened 2018-09-01
DOC: fix return type of str.extract

zhujun98:
- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -854,8 +854,8 @@ def str_extract(arr, pat, flags=0, expand=True):
     pat : string
         Regular expression pattern with capturing groups.
     flags : int, default 0 (no flags)
-        ``re`` module flags, e.g. ``re.IGNORECASE``.
-        See :mod:`re`
+        One of the ``re`` module flags, e.g. ``re.IGNORECASE``.
+        See :mod:`re`.
     expand : bool, default True
         If True, return DataFrame with one column per capture group.
         If False, return a Series/Index if there is one capture group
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -854,8 +854,8 @@ def str_extract(arr, pat, flags=0, expand=True):
     pat : string
         Regular expression pattern with capturing groups.
     flags : int, default 0 (no flags)
-        ``re`` module flags, e.g. ``re.IGNORECASE``.
-        See :mod:`re`
+        One of the ``re`` module flags, e.g. ``re.IGNORECASE``.
+        See :mod:`re`.
```
datapythonista: Do you mind briefly explaining what these flags are about. Nothing long, but something a bit more descriptive than what we've got now just linking to the `re` documentation. Something like "Flags that modify regular expression matching for things like case, spaces, etc."

datapythonista (commented):
Great changes. Adding some things that could be improved in the original docstring.

datapythonista (approved):
lgtm

datapythonista:
Thanks for the changes @zhujun98 

No need to address the comments in a PR in separate commits, that's rarely done. Feel free to fix all them in a single commit, unless separating them adds value to you.

Can you merge master in your branch and push it, so the CI runs again, and hopefully this time all is green please.

zhujun98:
Thanks for the explanation @datapythonista 

I am a little bit confused about the next step. To trigger the CI again, do you mean 'git merge upstream/master' or 'git rebase upstream/master'?

By the way, I checked other PRs (e.g. #22571) and it seems there is some issue with Anaconda now.

gfyoung:
Anaconda is back to normal, so I restarted all of your failed builds.  In terms of re-triggering builds, you have two options:

* Ping us: then we can restart the failing builds instead of re-triggering the entire build.
* "Redo" your last commit as follows:

~~~
git commit --amend
git push -f
~~~

zhujun98:
Thank you for the information @gfyoung!

Finally we got all the green lights @datapythonista!  Thank you very much for guiding me through the whole workflow of  contributing to Pandas docstrings! I really enjoyed it

datapythonista:
Another option to what @gfyoung said is `git fetch upstream && git merge upstream/master`, which is what I initially proposed. As pandas is quite active it rarely won't have anything to merge, and it has the advantage that if there is any conflict, you'll detect it directly.

Changes lgtm, but will let someone else merge it, so we have an extra pair of eyes on the changes.

PR #22563, opened 2018-09-01
DOC: Deleted 'replaced' from Returns docstring

davidmvalente:
Deleted the word 'replaced' from the Returns section of the docstring, since it is an unnecessary local variable name.

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -558,7 +558,7 @@ def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
 
     Returns
     -------
-    replaced : Series/Index of objects
+    Series/Index of objects
 
     Raises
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -558,7 +558,7 @@ def str_replace(arr, pat, repl, n=-1, case=None, flags=0, regex=True):
 
     Returns
     -------
-    replaced : Series/Index of objects
+    Series/Index of objects
```
datapythonista: I think we usually document this as `Series or Index of object` (note I also change objects to object).

Also, If you can add a description in the next line, indented with four spaces, explaining what this method returns, that would be great.

datapythonista:
Thanks for the changes @davidmvalente

PR #22566, opened 2018-09-01
DOC: Updating str_repeat docstring

JesperDramsch:
- [ x] tests passed
- [ x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

Updated the Docstring with examples for str.repeat()

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -703,7 +703,7 @@ def str_repeat(arr, repeats):
     1    bb
     2    cc
 
-    Array of int repeats corresponding string in Series
+    Sequence of int repeats corresponding string in Series
 
     >>> s.str.repeat(repeats=[1, 2, 3])
     0      a
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -680,12 +680,33 @@ def str_repeat(arr, repeats):
 
     Parameters
     ----------
-    repeats : int or array
+    repeats : int or array of int
         Same value for all (int) or different value per (array)
 
     Returns
     -------
-    repeated : Series/Index of objects
+    Series  or Index of objects
+
+    Examples
+    --------
+    
+    >>> s = pd.Series(['a', 'b', 'c'])
+    >>> s
+    0    a
+    1    b
+    2    c
+    
+    Single int repeats string in Series
+    >>> s.str.repeat(2)
+    0    aa
+    1    bb
+    2    cc
+
+    Array of int repeats corresponding string in Series 
+    >>> s.str.repeat([1, 2, 3])
+    0      a
+    1     bb
+    2    ccc
     """
     if is_scalar(repeats):
 
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -689,7 +689,6 @@ def str_repeat(arr, repeats):
 
     Examples
     --------
-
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -680,12 +680,33 @@ def str_repeat(arr, repeats):
 
     Parameters
     ----------
-    repeats : int or array
+    repeats : int or array of int
```
datapythonista: I think array can give the impression we want a specific structure like numpy array. Same if we say list. I'd say we've been using sequence in cases like this.

pandas/core/strings.py
```diff
@@ -680,12 +680,33 @@ def str_repeat(arr, repeats):
...
         Same value for all (int) or different value per (array)
 
     Returns
     -------
-    repeated : Series/Index of objects
+    Series  or Index of objects
+
```
datapythonista: Two spaces after Series. Use `object` as the Python type instead of objects (like using `list of int` instead of `list of ints`.

Can you add a description in the line after the type, on what is being returned please?

pandas/core/strings.py
```diff
@@ -680,12 +680,33 @@ def str_repeat(arr, repeats):
...
+    Examples
+    --------
+
```
datapythonista: I don't think we need this blank line.

pandas/core/strings.py
```diff
@@ -680,12 +680,33 @@ def str_repeat(arr, repeats):
...
+    >>> s = pd.Series(['a', 'b', 'c'])
+    >>> s
+    0    a
+    1    b
+    2    c
+
+    Single int repeats string in Series
+    >>> s.str.repeat(2)
```
datapythonista: I think we need a blank line between the text and the code. Or at least we've been leaving it, and it'd be good to be consistent.

Can you provide the argument as a keyword argument? I think `.repeat(repeats=2)` is more explicit thatn `.repeat(2)` (may be would make sense to call the parameter `.repeat(times=2)` but that's for another issue.

pandas/core/strings.py
```diff
@@ -680,12 +680,33 @@ def str_repeat(arr, repeats):
...
+    0    aa
+    1    bb
+    2    cc
+
+    Array of int repeats corresponding string in Series
+    >>> s.str.repeat([1, 2, 3])
```
datapythonista: Same as above about using the "array" word, and about leaving a blank line.

datapythonista (commented):
Great changes, very clear example. Added some comments about minor things.

pandas/core/strings.py
```diff
@@ -680,12 +680,35 @@ def str_repeat(arr, repeats):
...
-    repeats : int or array
-        Same value for all (int) or different value per (array)
+    repeats : int or sequence of int
+        Same value for all (int) or different value per (sequence)
 
     Returns
     -------
-    repeated : Series/Index of objects
+    Series or Index of object
+        Series or Index of repeated string object specified by input parameter repeats
```
jorisvandenbossche: I think this line will be too long (pep8 80 chars)

also "string object" -> "string objects"

jorisvandenbossche (commented):
Nice example!

JesperDramsch:
I messed up. It goes on here: #22571

PR #22570, opened 2018-09-01
DOC: Updating str_pad docstring

JesperDramsch:
Added Examples to docstring.
Added documentation to each part of docstring

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1364,13 +1364,6 @@ def str_pad(arr, width, side='left', fillchar=' '):
     0    -caribou--
     1    --tiger---
     dtype: object
-
-    >>> s = pd.Series(["caribou", 0, np.NaN])
-    >>> s.str.pad(width=10, fillchar='-')
-    0    ---caribou
-    1           NaN
-    2           NaN
-    dtype: object
     """
     if not isinstance(fillchar, compat.string_types):
         msg = 'fillchar must be a character, not {0}'
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1314,8 +1314,7 @@ def str_index(arr, sub, start=0, end=None, side='left'):
 
 def str_pad(arr, width, side='left', fillchar=' '):
     """
-    Pad strings in the Series/Index with additional characters on
-    specified side to fill up to specified width.
+    Pad strings in the Series/Index up to width
 
     Parameters
     ----------
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1321,14 +1321,16 @@ def str_pad(arr, width, side='left', fillchar=' '):
     ----------
     width : int
         Minimum width of resulting string; additional characters will be filled
-        with spaces
+        with character defined in fillchar
     side : {'left', 'right', 'both'}, default 'left'
+        Side from which to fill resulting string
     fillchar : str, default ' '
         Additional character for filling, default is whitespace
 
     Returns
     -------
-    Series or Index of objects
+    Series or Index of object
+        Returns Series or Index with minimum number of char in object
 
     Examples
     --------
@@ -1337,15 +1339,15 @@ def str_pad(arr, width, side='left', fillchar=' '):
     0    panda
     1      fox
 
-    >>> s.str.pad(10)
+    >>> s.str.pad(width=10)
     0         panda
     1           fox
 
-    >>> s.str.pad(10, 'right', '-')
+    >>> s.str.pad(width=10, side='right', fillchar='-')
     0    panda-----
     1    fox-------
 
-    >>> s.str.pad(10, 'both', '-')
+    >>> s.str.pad(width=10, side='both', fillchar='-')
     0    --panda---
     1    ---fox----
     """
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -1331,6 +1331,17 @@ def str_pad(arr, width, side='left', fillchar=' '):
     Series or Index of object
         Returns Series or Index with minimum number of char in object.
 
+    See Also
+    --------
+    Series.str.rjust: Fills the left side of strings with an arbitrary
+        character.
+    Series.str.ljust: Fills the right side of strings with an arbitrary
+        character.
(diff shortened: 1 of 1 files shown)


pandas/core/strings.py
```diff
@@ -1314,21 +1314,42 @@ def str_index(arr, sub, start=0, end=None, side='left'):
 
 def str_pad(arr, width, side='left', fillchar=' '):
     """
-    Pad strings in the Series/Index with an additional character to
-    specified side.
+    Pad strings in the Series/Index with additional characters on
+    specified side to fill up to specified width.
```
gfyoung: Since you're trying to improve the docstring, can we try to make this initial description a single line (per convention)?  If you need to provide more details, you can do so afterwards i.e.

~~~python
<one-line summary>

<description>
~~~

JesperDramsch: Thanks. I'll keep this in mind and correct it now.

datapythonista (commented):
Looks good, added couple of comments.

pandas/core/strings.py
```diff
@@ -1314,21 +1314,41 @@ def str_index(arr, sub, start=0, end=None, side='left'):
...
+        Side from which to fill resulting string.
+    fillchar : str, default ' '
+        Additional character for filling, default is whitespace.
 
     Returns
     -------
-    padded : Series/Index of objects
+    Series or Index of object
+        Returns Series or Index with minimum number of char in object
+
```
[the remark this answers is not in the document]
JesperDramsch: Sure thing.

pandas/core/strings.py
```diff
@@ -1314,21 +1314,41 @@ def str_index(arr, sub, start=0, end=None, side='left'):
...
+    Examples
+    --------
+    >>> s = pd.Series(["panda", "fox"])
```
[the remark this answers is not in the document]
JesperDramsch: Done

pandas/core/strings.py
```diff
@@ -1314,21 +1314,65 @@ def str_index(arr, sub, start=0, end=None, side='left'):
...
+    Series.str.rjust: Fills the left side of strings with an arbitrary
+        character.
+    Series.str.ljust: Fills the right side of strings with an arbitrary
+        character.
+    Series.str.pad: Fills the specified sides of strings with an arbitrary
+        character.
+    Series.str.center: Fills boths sides of strings with an arbitrary
+        character.
+    Series.str.zfill:  Pad strings in the Series/Index by prepending '0'
+        character.
```
datapythonista: `See Also` should be placed before the `Examples`. You have the documentation about docstrings here: https://pandas.pydata.org/pandas-docs/stable/contributing_docstring.html

As `pad` is this same method, I wouldn't include it in the list of related methods.

Did you try rendering the page with `./doc/make.py html --single pandas.Series.str.pad`? I think there is a missing space before the colon that will break the format.

Also, if you run `./scripts/validate_docstrings.py pandas.Series.str.pad` will probably report it.

More than explaining `rjust`... I'd just say it's equivalent to `pad(side='left')`, so users don't need to guess how they are different and which they should use.

pandas/core/strings.py
```diff
@@ -1314,21 +1314,65 @@ def str_index(arr, sub, start=0, end=None, side='left'):
...
+    0    -caribou--
+    1    --tiger---
+    dtype: object
+
+    >>> s = pd.Series(["caribou", 0, np.NaN])
+    >>> s.str.pad(width=10, fillchar='-')
+    0    ---caribou
+    1           NaN
+    2           NaN
+    dtype: object
```
datapythonista: Not sure if I'd include the `0` and `np.nan` in the original Series, so we don't need this last example.

We usually use `np.nan` instead of `np.NaN`, I'd change it for consistency.

pandas/core/strings.py
```diff
@@ -1314,21 +1314,65 @@ def str_index(arr, sub, start=0, end=None, side='left'):
 
 def str_pad(arr, width, side='left', fillchar=' '):
     """
-    Pad strings in the Series/Index with an additional character to
-    specified side.
+    Pad strings in the Series/Index up to width.
```
datapythonista: I think in this case we could have a extended summary, providing a bit more information on what pad means, and when it can be useful.

datapythonista (commented):
Nice docstring, added couple of comments that I think could improve it further.

jreback:
@datapythonista

datapythonista:
@JesperDramsch do you have time to address the previous comments?

JesperDramsch:
> @JesperDramsch do you have time to address the previous comments?

I'll get to it this weekend, I think.

JesperDramsch:
I think I addressed all comments and validated it. @datapythonista

pandas/core/strings.py
```diff
@@ -1314,23 +1314,57 @@ def str_index(arr, sub, start=0, end=None, side='left'):
...
     -------
-    padded : Series/Index of objects
-    """
+    Series or Index of object
+        Returns Series or Index with minimum number of char in object.
 
+    See Also
+    --------
+    Series.str.rjust: Fills the left side of strings with an arbitrary
+        character. Equivalent to pd.str.pad(side='left').
```
datapythonista: it's not really `pd.str.pad`, but `Series.str.pad`. Can you also add double backticks around it so it renders in the website as code. Same for all cases.

pandas/core/strings.py
```diff
@@ -1314,23 +1314,57 @@ def str_index(arr, sub, start=0, end=None, side='left'):
...
-    Pad strings in the Series/Index with an additional character to
-    specified side.
+    Pad strings in the Series/Index up to width.
 
     Parameters
     ----------
     width : int
         Minimum width of resulting string; additional characters will be filled
-        with spaces
+        with character defined in fillchar.
```
datapythonista: Can you add backticks around `fillchar`, so it's explicit that is the parameter name.

datapythonista (commented):
Just couple of additional comments. Besides those, looks great.

jorisvandenbossche (approved):

jorisvandenbossche:
@JesperDramsch Pushed some last small changes. Thanks a lot for the PR!

PR #22534, opened 2018-08-29
CLN: use dispatch_to_series where possible

jbrockmendel:
A bunch of PRs touching DataFrame ops have gone through recently.  This does some follow-up cleanup to unify the way things are done across a few different methods.

PR #22572, opened 2018-09-01
Use dispatch_to_series where possible

jbrockmendel:
Broken off from #22534, since that ended up surfacing a bug elsewhere.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -4806,13 +4806,14 @@ class DataFrame(NDFrame):
             return ops.dispatch_to_series(this, other, _arith_op)
         else:
             result = _arith_op(this.values, other.values)
-
-        return self._constructor(result, index=new_index, columns=new_columns,
-                                 copy=False)
+            return self._constructor(result,
+                                     index=new_index, columns=new_columns,
+                                     copy=False)
 
     def _combine_match_index(self, other, func, level=None):
         left, right = self.align(other, join='outer', axis=0, level=level,
                                  copy=False)
+        assert left.index.equals(right.index)
         new_data = func(left.values.T, right.values).T
         return self._constructor(new_data,
                                  index=left.index, columns=self.columns,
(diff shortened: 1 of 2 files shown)


pandas/core/frame.py
```diff
@@ -4806,13 +4806,14 @@ def _arith_op(left, right):
...
-        return self._constructor(result, index=new_index, columns=new_columns,
-                                 copy=False)
+            return self._constructor(result,
+                                     index=new_index, columns=new_columns,
+                                     copy=False)
 
     def _combine_match_index(self, other, func, level=None):
         left, right = self.align(other, join='outer', axis=0, level=level,
                                  copy=False)
+        assert left.index.equals(right.index)
```
gfyoung: Why do we need these assert statements?  Are they tested anywhere?

jbrockmendel: These assertions are always true.  I added these so I don't have to keep looking back at core.ops (the only place from which these are called) to double-check what's going on.

gfyoung: So this is just a sanity check really?  Perhaps we should comment that then?

pandas/core/ops.py
```diff
@@ -1638,6 +1638,7 @@ def dispatch_to_series(left, right, func):
     """
     # Note: we use iloc to access columns for compat with cases
     #       with non-unique columns.
+    right = lib.item_from_zerodim(right)
```
jreback: is there a reason you needed to add this? (e.g. do we actually hit this in the new code path and not in the original)?

jreback (changes requested):

pandas/core/ops.py
jbrockmendel: Yes.  This PR has `DataFrame._combine_const` send zero-dim arrays down this path, which without this line don't get handled correctly.

jreback: ok great

jreback (approved):

PR #14645, opened 2016-11-13
Standardize function signatures

gfyoung:
Standardize the following function signatures:

1) `repeat(reps, *args, **kwargs)`
2) `searchsorted(value, side='left', sorter=None)`

Closes #12662.

PR #22574, opened 2018-09-02
CLN: Rename 'n' to 'repeats' in .repeat methods

gfyoung:
For `Index` and `MultiIndex`.

xref #14645.

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -527,6 +527,7 @@ Removal of prior version deprecations/changes
 - Several private functions were removed from the (non-public) module ``pandas.core.common`` (:issue:`22001`)
 - Removal of the previously deprecated module ``pandas.core.datetools`` (:issue:`14105`, :issue:`14094`)
 - Strings passed into :meth:`DataFrame.groupby` that refer to both column and index levels will raise a ``ValueError`` (:issue:`14432`)
+- :meth:`Index.repeat` and :meth:`MultiIndex.repeat` have renamed the ``n`` argument to ``repeats``(:issue:`14645`)
 -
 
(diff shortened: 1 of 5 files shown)


doc/source/whatsnew/v0.24.0.txt
```diff
@@ -527,6 +527,7 @@ Removal of prior version deprecations/changes
 - Several private functions were removed from the (non-public) module ``pandas.core.common`` (:issue:`22001`)
 - Removal of the previously deprecated module ``pandas.core.datetools`` (:issue:`14105`, :issue:`14094`)
 - Strings passed into :meth:`DataFrame.groupby` that refer to both column and index levels will raise a ``ValueError`` (:issue:`14432`)
+- :meth:`Index.repeat` and :meth:`MultiIndex.repeat` have renamed the ``n`` argument to ``repeats``(:issue:`14645`)
 -
```
topper-123: ``n`` is being removed though, not renamed? Maybe

> :meth:`Index.repeat` and :meth:`MultiIndex.repeat` have the ``n`` synonym to the ``repeats`` argument removed (deprecated since v.0.20.0) ). (:issue:`14645`)

topper-123 (changes requested):

doc/source/whatsnew/v0.24.0.txt
topper-123: The way I see it is that in 0.20 ``n`` was renamed and deprecated, and now the deprecated param ``n`` is being removed...

Anyway, could just be a difference of perspective.

gfyoung: > Anyway, could just be a difference of perspective.

Agreed.

jreback:
thanks!

PR #22575, opened 2018-09-02
CLN: tests for str.cat

h-vetinari:
Cleaning up some left-overs from #20347 and preparing the tests for changing the default of `join` to `'left'` in 1.0 (which would break some tests that assume no alignment currently). As a side benefit, this has:
* nicer parametrization
* nicer switches (and reduced usage of them within the tests)
* now follows `result / expected` pattern instead of calculating result within `assert_..._equal`

### DIFF
--- a/pandas/tests/test_strings.py
+++ b/pandas/tests/test_strings.py
@@ -164,27 +164,27 @@ class TestStringMethods(object):
 
         # single array
         result = s.str.cat()
-        exp = 'aabbc'
-        assert result == exp
+        expected = 'aabbc'
+        assert result == expected
 
         result = s.str.cat(na_rep='-')
-        exp = 'aabbc-'
-        assert result == exp
+        expected = 'aabbc-'
+        assert result == expected
 
         result = s.str.cat(sep='_', na_rep='NA')
-        exp = 'a_a_b_b_c_NA'
-        assert result == exp
+        expected = 'a_a_b_b_c_NA'
+        assert result == expected
 
         t = np.array(['a', np.nan, 'b', 'd', 'foo', np.nan], dtype=object)
-        exp = box(['aa', 'a-', 'bb', 'bd', 'cfoo', '--'])
+        expected = box(['aa', 'a-', 'bb', 'bd', 'cfoo', '--'])
 
         # Series/Index with array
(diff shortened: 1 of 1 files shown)


pandas/tests/test_strings.py
```diff
@@ -220,122 +199,110 @@ def test_str_cat(self, series_or_index):
...
         s = Index(['a', 'a', 'b', 'a'], dtype=dtype_caller)
-        if series_or_index == 'series':
-            s = Series(s)
+        s = s if box == Index else Series(s, index=s)
         t = Index(['b', 'a', 'b', 'c'], dtype=dtype_target)
 
         exp = Index(['ab', 'aa', 'bb', 'ac'])
-        if series_or_index == 'series':
-            exp = Series(exp)
+        exp = exp if box == Index else Series(exp, index=s)
```
jreback: use expected

pandas/tests/test_strings.py
```diff
@@ -469,12 +436,11 @@ def test_cat_on_filtered_index(self):
 
         str_year = df.year.astype('str')
         str_month = df.month.astype('str')
-        str_both = str_year.str.cat(str_month, sep=' ', join='left')
```
jreback: why are you removing the join here? is it not needed? should these be parameterized by join type?

jreback (changes requested):
pls don't use abbreviations, eg. always use expected

pandas/tests/test_strings.py
h-vetinari: In the course of #20347, the warnings were originally always issues whenever `join` was not specified - and this test correspondingly adapted. However, the warnings were later reduced to only be emitted in case of different indexes between caller and others, so in this case, it's not necessary.

h-vetinari:
@jreback 
> pls don't use abbreviations, eg. always use expected

This module's full of `exp` (maybe even the majority), but I replaced it in the diff I'm working on now.

h-vetinari:
not sure what's up with these `TestExcelWriter` test failures...? rebasing didn't help.

h-vetinari:
@jreback All green

jreback (approved):

jreback:
thanks @h-vetinari

PR #22580, opened 2018-09-03
Add 'name' as argument for index 'to_frame' method

henriqueribeiro:
- [ ] closes #xxxx
- [x] tests added / passed
- [ ] whatsnew entry

Since series `to_frame` method has a `name` argument, I believe it makes sense for index also have it.


TomAugspurger:
The problem is that a name argument doesn’t make sense for a MultiIndex. Not sure what’s best here.

pandas/core/indexes/base.py
```diff
@@ -1154,10 +1158,17 @@ def to_frame(self, index=True):
...
+        >>> idx.to_frame(index=False, name='zoo')
+            zoo
+        0   Ant
+        1  Bear
+        2   Cow
         """
 
         from pandas import DataFrame
-        name = self.name or 0
+        if not name:
```
TomAugspurger: This would have to be if name is not None, to allow falsey names.

henriqueribeiro: You are right! I will change that

henriqueribeiro:
@TomAugspurger regarding MultiIndex, I'm not seeing how this `name` will affect MultiIndex. Can you enlight me please?

TomAugspurger:
Probably best to keep the signature the same for MI, but indicate that
`name` should be a sequence of strings matching the number of levels.

Can you add a test for that, and implement it if it's broken?

gfyoung:
@henriqueribeiro : Thanks for the contribution!  In the future, if you could open an issue *before* opening the PR, that would allow us to evaluate the idea first before implementing.

henriqueribeiro:
@gfyoung Sure! Sorry for that

henriqueribeiro:
@TomAugspurger I cannot understand why some checks are failing. Can you give me some help please?

TomAugspurger:
There was  a new release of openpyxl that's incompatible with pandas. It's been worked around on master, so you need to fetch upstream and push your changes.

```
git fetch upstream
git merge upstream/master
git push origin
```

pandas/core/indexes/base.py
```diff
@@ -1127,6 +1127,10 @@ def to_frame(self, index=True):
         index : boolean, default True
             Set the index of the returned DataFrame as the original Index.
 
+        name : object, default None
+            The passed name should substitute for the series name (if it has
+            one).
+
```
jreback: need a version added tag (0.24.0)

pandas/core/indexes/multi.py
```diff
@@ -1126,7 +1126,7 @@ def _to_safe_for_reshape(self):
         """ convert to object if we are a categorical """
         return self.set_levels([i._to_safe_for_reshape() for i in self.levels])
 
-    def to_frame(self, index=True):
+    def to_frame(self, index=True, names=None):
         """
```
jreback: can you update the doc-string (and add a versionadded tag)

pandas/core/indexes/multi.py
```diff
@@ -1143,11 +1143,21 @@ def to_frame(self, index=True):
...
 
         from pandas import DataFrame
-        result = DataFrame({(name or level):
-                            self._get_level_values(level)
-                            for name, level in
-                            zip(self.names, range(len(self.levels)))},
-                           copy=False)
+        if names is not None:
+            if len(names) != len(self.levels):
+                raise AssertionError("'names' should have same lenght as "
```
jreback: ValueError

pandas/core/indexes/multi.py
```diff
@@ -1143,11 +1143,21 @@ def to_frame(self, index=True):
...
+                                     "number of levels on index")
+            result = DataFrame({(name):
```
jreback: can you just set the name once here (in the if/else), then construct the frame after, so the construction code is only once

pandas/tests/indexes/common.py
```diff
@@ -81,6 +81,17 @@ def test_to_frame(self):
         df = idx.to_frame(index=False)
         assert df.index is not idx
```
jreback: can you add a comment referencing the issue number

pandas/tests/indexes/multi/test_conversion.py
```diff
@@ -37,6 +37,17 @@ def test_to_frame():
     expected.index = index
     tm.assert_frame_equal(result, expected)
 
+    index = MultiIndex.from_tuples(tuples)
```
jreback: same as above

pandas/tests/indexes/multi/test_conversion.py
```diff
@@ -45,12 +56,20 @@ def test_to_frame():
             1: np.tile(pd.date_range('20130101', periods=3), 5)})
     tm.assert_frame_equal(result, expected)
 
-    index = MultiIndex.from_product([range(5),
-                                     pd.date_range('20130101', periods=3)])
     result = index.to_frame()
     expected.index = index
     tm.assert_frame_equal(result, expected)
 
+    result = index.to_frame(index=False, names=['first', 'second'])
```
jreback: hmm do we have multiple tests for the same thing here?

pandas/core/indexes/multi.py
jreback: this code path is not hit in the tests

jreback (changes requested):

pandas/core/indexes/multi.py
```diff
@@ -1126,7 +1126,7 @@ def _to_safe_for_reshape(self):
         """ convert to object if we are a categorical """
         return self.set_levels([i._to_safe_for_reshape() for i in self.levels])
 
-    def to_frame(self, index=True):
+    def to_frame(self, index=True, names=None):
```
TomAugspurger: This should be `name`, singular right? Changing the signature of a method in a subclass is generally a bad idea.

jreback: oh yes I c. though we do accept name/names in many methods we should just not add to the problem (IIRC we are changing that signature in another issue)

TomAugspurger:
@henriqueribeiro linting error https://travis-ci.org/pandas-dev/pandas/jobs/425773293#L3653 and it looks like you may need to merge master still.

pandas/core/indexes/base.py
```diff
@@ -1115,17 +1115,21 @@ def to_series(self, index=None, name=None):
...
-        .. versionadded:: 0.21.0
+        .. versionadded:: 0.24.0
 
         Parameters
         ----------
         index : boolean, default True
             Set the index of the returned DataFrame as the original Index.
 
+        name : object, default None
+            The passed name should substitute for the series name (if it has
```
TomAugspurger: 'series' -> 'index'

pandas/core/indexes/base.py
```diff
@@ -1153,10 +1157,17 @@ def to_frame(self, index=True):
         0   Ant
         1  Bear
         2   Cow
+
```
TomAugspurger: A bit of prose here explaining:

    To override the name of the resulting column, specify `name`:

pandas/core/indexes/multi.py
```diff
@@ -1126,27 +1126,39 @@ def _to_safe_for_reshape(self):
...
 
-        .. versionadded:: 0.20.0
+        .. versionadded:: 0.24.0
 
         Parameters
         ----------
         index : boolean, default True
             Set the index of the returned DataFrame as the original MultiIndex.
 
+        name : list / sequence of strings or None
```
TomAugspurger: just sequence. We don't list `None` as a type, just put optional.

    name : sequence of strings, optional

pandas/core/indexes/multi.py
```diff
@@ -1126,27 +1126,39 @@ def _to_safe_for_reshape(self):
...
+            The passed names should substitute for the series names (if it has
```
TomAugspurger: 'series' -> 'index level names'. "If it has one" can probably be removed.

pandas/core/indexes/multi.py
```diff
@@ -1126,27 +1126,39 @@ def _to_safe_for_reshape(self):
...
+            one).
+
         Returns
         -------
         DataFrame : a DataFrame containing the original MultiIndex data.
         """
 
         from pandas import DataFrame
+        if name is not None:
+            if len(name) != len(self.levels):
```
TomAugspurger: Should we protect against people accidentally using a single string here, rather than a list of strings? You can use `is_list_like` to check I think.

pandas/core/indexes/multi.py
```diff
@@ -1126,27 +1126,39 @@ def _to_safe_for_reshape(self):
...
+                raise ValueError("'name' should have same lenght as "
```
TomAugspurger: Typo: `lenght`

TomAugspurger:
Needs a release note too, for 0.24

TomAugspurger:
Probably under the "Other Enhancements" section.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -184,6 +184,7 @@ Other Enhancements
 - :class:`DatetimeIndex` gained :attr:`DatetimeIndex.timetz` attribute. Returns local time with timezone information. (:issue:`21358`)
 - :class:`Resampler` now is iterable like :class:`GroupBy` (:issue:`15314`).
 - :ref:`Series.resample` and :ref:`DataFrame.resample` have gained the :meth:`Resampler.quantile` (:issue:`15023`).
+- :meth:Both Index and MultiIndex now support overriding column name(s) when using `to_frame` method (:issue:`22580`).
```
TomAugspurger: I think just

    :meth:`Index.to_frame` now supports overriding column name(s) (:issue:`22580`).

henriqueribeiro: Totally agree

pandas/core/indexes/multi.py
```diff
@@ -1126,27 +1126,42 @@ def _to_safe_for_reshape(self):
...
         Returns
         -------
         DataFrame : a DataFrame containing the original MultiIndex data.
         """
 
         from pandas import DataFrame
+        if name is not None:
+            if not is_list_like(name):
+                raise TypeError("'name' must be a list / sequence "
+                                "of array-likes.")
```
TomAugspurger: Sorry, just noticing. This should be a "list / sequence of column names". Not array-likes.

We can say "strings" instedad of "column names", but I think we allow any hashable value, so "list of names" is probably best.

pandas/tests/indexes/common.py
```diff
@@ -81,6 +81,18 @@ def test_to_frame(self):
...
 
+        # See GH-22580
```
TomAugspurger: Lot's of duplication here. Can you make this a parameterized test?

```
@pytest.mark.parametrize("name", [None, "new_name"]):
    ...
    if new_name:
        assert df.columns[0] == new_name
    else:
        assert df.columns[0] == name
    ....
```

jreback (approved):

gfyoung:
All green, and since @jreback has already approved this, merging.

gfyoung:
Thanks @henriqueribeiro !

henriqueribeiro:
Thanks to @TomAugspurger for this "ping-pong" fixing.

Issue #13349, opened 2016-06-02
API: skipfooter or skip_footer?  read_csv can't seem to decide


gfyoung:
`skipfooter`  : "This line <a href="https://github.com/pydata/pandas/blob/master/pandas/io/parsers.py#L729">here</a> says I am the right choice."

`skip_footer` : "This line <a href=https://github.com/pydata/pandas/blob/master/pandas/io/parsers.py#L519">here</a> says I am the right choice."

Whatever the case is, something should certainly be done to avoid having duplicate arguments for the same thing, as can be seen <a href="https://github.com/pydata/pandas/blob/master/pandas/io/parsers.py#L582">here</a>

jreback:
The full-list is here: https://github.com/pydata/pandas/issues/4988

Yeah should prob fix this. I think the ``_`` are better (though I wouldn't change for example ``usecols``)
and ``skiprows`` is in-line with ``skipfooter``.

We should see what changes to make this reasonable and minimze the deprecations (including maybe doing nothing).

gfyoung:
Yeah, not sure what is best yet.  Just wanted to mention this problem since having duplicate arguments in the API isn't a good thing.  A more careful analysis of the API / code would be needed to make sure that we don't change too much unnecessarily.

gfyoung:
I'm inclined to deprecate `skipfooter` because the `_` is used internally throughout the code.  I'll take a look at that first and change any relevant tests as well to use the "correct" one.

PR #22587, opened 2018-09-04
ENH: Replace skiprows with skip_rows to begin standardizing underscore usage in keyword arguments

palewire:
By [my rough count](https://docs.google.com/spreadsheets/d/1h_iQ5Pexs5SuCT2pfbN8OH6ld6ZyvWqTEaH2W-xkEJE/edit?usp=sharing), the [`read_csv`](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_csv.html) method has nearly 50 keyword arguments.

Of those, 32 arguments are made up or two or more words. Twenty of those multi-word arguments use an underscore to mark the space between words, like  `skip_blank_lines` and `parse_dates`. Twelve do not, like `chunksize` and `lineterminator`. 

It is my opinion this is a small flaw in pandas' API, and the library would benefit by standardizing how spaces are handled. It will make pandas more legible and consistent, and therefore easier for users of all experience levels.

Since the underscore method is more common and more legible, I propose it be adopted.

All existing arguments without an underscore will need to be modified. As a first salvo, I have attempted to change the `skiprows` kwarg to `read_csv` and its sibling methods to `skip_rows`. I have included an experimental deprecation warning that aims to continue to support the old argument for some interval into the future. 

Due to my lack of expertise on pandas' internals, I expect this request includes some flaws that would need to be corrected before inclusion. However, I hope that the maintainers of the library will agree with the overall aims of this patch, which is to begin a process of introducing greater consistency to the style of keyword arguments. 

If you do agree, I would be pleased to lead an effort to gradually standardize the inputs and put in the work to finish the job. 

Thank you for your consideration.


palewire:
I am sympathetic to desire to maintain backwards compatiblility, but the lack of consistency in syntax across pandas has gradually grown into a serious drawback.

I have taught pandas to dozens of newbies across the country and I can testify from experience that small variations in the naming style of commonly used methods introduces unnecessary frustration, and even reduces user confidence in the quality of the overall product.

As a frequent user of pandas, I can also attest that the inconsistencies require me, someone who uses the library daily, to routinely use the documentation to ensure I use the proper kwarg naming style.

All of this can be avoided by selecting a style a d sticking to it, just as an author or professional documents outside of programming rountinely do.

The deprecation warnings, if included, could be temporary, and ultimately removed in a future version, much in the way `sort_values` was introduced.

asuozzo:
Agreed, @palewire! I know a lot of the common two-word arguments for read_csv, but it's always frustrating to not be able to remember off the top of my head whether they have underscores.

emamd:
Agreed with this proposal, I have the same issues as @pallih and @asuozzo. While it would be a headache to add these deprecations, I think increasing consistency would reduce the headache users have in remembering basic keyword arguments and speed up pandas use in general.

gfyoung:
@palewire : Thanks for the contribution!  Could you open an issue for this first (that you can then close with this PR)?  This would actually be much preferred for evaluating and discussing a proposal like this.

xref: #13349

palewire:
@gfyoung, will do.

jreback:
@palewire can you list all of the kwargs that you are attempting to change.

I think *some* of them could be ok to actually deprecate and change (with a suitable deprecation period), but others are just not worth the effort and/or maybe just support both spelllings.

jreback:
@gfyoung I believe we have an issue for this anyhow?

gfyoung:
@jreback : No, we don't.  The closest was #13349.

palewire:
I will file an issue shortly. Sorry for the delay.

WillAyd:
Closing this PR as stale. For any discussion please move to referenced issue

PR #22598, opened 2018-09-04
Fix string format in test runner

rhysparry:
`.format()` was expecting keyword arguments. Updated to match other skips nearby.


alimcmaster1 (commented):
LGTM, mind updating the PR title to something like "Fix string format in test runner" the current one is a bit misleading

TomAugspurger:
If you fetch and merge master the CI failures should be fixed.

gfyoung (approved):
Good catch!

gfyoung:
@rhysparry : Thanks!

PR #22622, opened 2018-09-06
DataFrame.corrwith() method now supports 'how', 'method' and 'min_periods' params

Mottl:
- [x] tests added / passed
- [x] whatsnew entry

A method `DataFrame.corrmatrix()` creates a correlation matrix between
all columns of the first DataFrame object and all columns of the second
DataFrame object. It differs from `corrwith()` in that way it computes correlations between all pairs of first and second columns (not only that has the same names).


jreback:
what is the utility of this? we are loathe to add additional api

Mottl:
@jreback
The utility is simple:  
during the feature selection process it is not desirable to spend time to get a full correlation matrix when the number of features are huge. It's more important to calculate a correlation matrix between M number of features and Y number of targets (we have a variety of different targets and we want to understand how features correlates to every of them) and `corrmatrix()` makes it simple.

So `corrmatrix()` is a simple form of `corr()` but without unnecessary calculations which will be dropped anyway.

Mottl:
This is how most people solve this problem:
https://stackoverflow.com/questions/34896455/how-to-do-pearson-correlation-of-selected-columns-of-a-pandas-data-frame
viewed: 24,989 times!

They calculate the full correlation matrix and drop all the unnecessary columns or make a lambda function, but it works only for correlation between DataFrame and Series, not with two DataFrames.

jreback:
-1 on this api as it will make things even more confusing

Mottl:
@jreback, how about adding new param to an existing `.corrwith()` function to compute all pair correlation between DataFrames instead of name-like?

Mottl:
We could add `kind` param to `DataFrame.corrwith()` method.
`kind='same'` for the current behaviour (returns a Series with pair-wise correlations between columns or rows with the same name/index)
`kind='all'` to produce a correlation matrix (returns a DataFrame with with pair-wise correlations between columns (or rows) of all possible combinations)

Example:
```python
 index = ['a', 'b', 'c', 'd', 'e']
 columns1 = ['one', 'two']
 columns2 = ['three', 'four']
 df1 = pd.DataFrame(np.random.randn(5, 2), index=index, columns=columns1)
 df2 = pd.DataFrame(np.random.randn(4, 2), index=index[:4], columns=columns2)
 df1.corrwith(df2, kind='all')
```

Mottl:
@jreback, what are we doing with `df1.corrwith()`? Changing as I described above?

jreback:
can u show the output of your example above with both options

Mottl:
@jreback

```python
import numpy as np
import pandas as pd
df1 = pd.DataFrame({'a':np.random.randn(10), 'b':np.random.randn(10)})
df2 = pd.DataFrame({'a':np.random.randn(10), 'c':np.random.randn(10)})

df1.corrwith(df2)  # default, which='same'
```
```
a    0.40657
b        NaN
c        NaN
dtype: float64
```
```python
df1.corrwith(df2, which='all')  # new
```
```
          a         c
a  0.406570 -0.005083
b -0.231165 -0.318797
```

Mottl:
@jreback, what should I do with the cancelled Circleci jobs? Is the PR well enough since Travis is green?

jreback:
i cancelled those jobs

Mottl:
I know

Mottl:
> @Mottl ok I think adding a parameter `how='pairwise|all'` might be ok here.
> cc @jorisvandenbossche

@jreback, @jorisvandenbossche
please review

Mottl:
```python
import numpy as np
import pandas as pd
df1 = pd.DataFrame(np.random.randn(1000,30))
df2 = pd.DataFrame(np.random.randn(1000,30))

%timeit df1.corrwith(df2)
23 ms ± 1.52 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)

%timeit df1.corrwith_orig(df2)
34.7 ms ± 2.17 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Actually the **new implementation is 34% faster** than the old one since it uses `pd.Series.corr()` which in turn uses numpy's `corrcoef()` instead of `left - left.mean()`, `right - right.mean()`, etc.

jreback:
try with a large number of columns

Mottl:
Same — **33% faster**:
```python
df1 = pd.DataFrame(np.random.randn(10000,3000))
df2 = pd.DataFrame(np.random.randn(10000,3000))

%timeit df1.corrwith(df2)
5.7 s ± 81.3 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%timeit df1.corrwith_orig(df2)
8.45 s ± 152 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

pandas/core/frame.py
```diff
@@ -6892,56 +6892,93 @@ def cov(self, min_periods=None):
...
         this = self._get_numeric_data()
 
         if isinstance(other, Series):
-            return this.apply(other.corr, axis=axis)
+            return this.apply(other.corr, axis=axis,
+                              method=method, min_periods=min_periods)
 
-        other = other._get_numeric_data()
+        if not isinstance(other, DataFrame):
+            raise TypeError("'other' parameter should be a DataFrame "
```
jreback: is this tested?

pandas/core/frame.py
```diff
@@ -6892,56 +6892,93 @@ def cov(self, min_periods=None):
...
+                            "or a Series object")
 
-        left, right = this.align(other, join='inner', copy=False)
+        other = other._get_numeric_data()
 
-        # mask missing values
-        left = left + right * 0
-        right = right + left * 0
+        if isinstance(this.columns, MultiIndex) \
```
jreback: is this tested?

pandas/core/frame.py
```diff
@@ -6892,56 +6892,93 @@ def cov(self, min_periods=None):
...
-            return this.apply(other.corr, axis=axis)
+            return this.apply(other.corr, axis=axis,
+                              method=method, min_periods=min_periods)
 
-        other = other._get_numeric_data()
+        if not isinstance(other, DataFrame):
+            raise TypeError("'other' parameter should be a DataFrame "
+                            "or a Series object")
 
-        left, right = this.align(other, join='inner', copy=False)
```
jreback: its odd that things didn't break when you removed this, have you tested a left / right that don't align? (IOW different columns / rows)

pandas/core/frame.py
```diff
@@ -6892,56 +6892,93 @@ def cov(self, min_periods=None):
...
         ----------
         other : DataFrame, Series
         axis : {0 or 'index', 1 or 'columns'}, default 0
             0 or 'index' to compute column-wise, 1 or 'columns' for row-wise
+        how : {'pairwise', 'all'}, default 'pairwise' if other is a DataFrame
+            and default 'all' if other is a Series object
+            * 'pairwise' to compute correlation between rows or columns with
+              the same name/index,
+            * 'all' to compute correlation between all possible pairs of rows
+              or columns
```
jreback: need a versionadded tag

pandas/core/frame.py
```diff
@@ -6892,56 +6892,93 @@ def cov(self, min_periods=None):
...
         drop : boolean, default False
             Drop missing indices from result, default returns union of all
+        method : {'pearson', 'kendall', 'spearman'}
+            * pearson : standard correlation coefficient
+            * kendall : Kendall Tau correlation coefficient
```
jreback: same for both of these

pandas/core/frame.py
```diff
@@ -6892,56 +6892,93 @@ def cov(self, min_periods=None):
...
-        rdem = right - right.mean()
+            this = this.transpose()
+            other = other.transpose()
+
+        if how == 'all':
+            corr = np.zeros((this.shape[1], other.shape[1]))
+
+            for i, col1 in enumerate(this.columns):
+                for j, col2 in enumerate(other.columns):
+                    corr[i, j] = this.loc[:, col1].corr(
```
jreback: dont you just to do a single triangle? (and copy to the other)?

use ``this.iloc[:, j]`` and so on. test this with non-uniques columns (you could raise on non-uniques alternatively)

jreback: but still use the iloc pattern here

pandas/core/frame.py
```diff
@@ -6892,56 +6892,93 @@ def cov(self, min_periods=None):
...
+                        other.loc[:, col2],
+                        method=method, min_periods=min_periods)
+            return DataFrame(data=corr, index=this.columns,
+                             columns=other.columns)
+
+        elif how == 'pairwise':
+            index = []
+            corr = []
+            for col in this.columns:
```
jreback: you are defeating the purpose of the alginment above, pls just align as above

pandas/core/frame.py
```diff
@@ -6892,56 +6892,93 @@ def cov(self, min_periods=None):
...
+                if col in other.columns:
+                    index.append(col)
+                    corr.append(this.loc[:, col].corr(other.loc[:, col],
+                                method=method, min_periods=min_periods))
+
+            correl = Series(data=corr, index=index)
+
+            if not drop:  # add missing columns to the resulting Series
```
jreback: huh? what are you trying to do here? you are fighting pandas by doing this

pandas/core/frame.py
```diff
@@ -6892,56 +6892,93 @@ def cov(self, min_periods=None):
...
-
-        correl = num / dom
-
-        if not drop:
-            raxis = 1 if axis == 0 else 0
-            result_index = this._get_axis(raxis).union(other._get_axis(raxis))
-            correl = correl.reindex(result_index)
-
-        return correl
+        else:
```
jreback: this is a valueerror

pandas/tests/frame/test_analytics.py
```diff
@@ -256,6 +256,27 @@ def test_corrwith_mixed_dtypes(self):
         expected = pd.Series(data=corrs, index=['a', 'b'])
         tm.assert_series_equal(result, expected)
 
+    def test_corrwith_how_all(self):
+        df = DataFrame({'a': np.random.randn(10000),
```
jreback: add a comment with the issue number

pandas/tests/frame/test_analytics.py
```diff
@@ -256,6 +256,27 @@ def test_corrwith_mixed_dtypes(self):
...
+                        'b': np.random.randn(10000)})
+        c1 = df.corrwith(df, how='all').loc['a', 'b']
+        c2 = np.corrcoef(df['a'], df['b'])[0][1]
+
+        tm.assert_almost_equal(c1, c2)
+        assert c1 < 1
```
jreback: need a lot more tests here, e.g. to hit all of the error conditions you are now asserting, as well as non-aligned and non-unique frames.

jreback (changes requested):
iterating over the pair of the columns is exactly what the nancorr routines seeks to avoid. can we just directly use that here? this is not scaling very well.

pandas/core/frame.py
Mottl: They get aligned further in `pd.Series.corr()`:
```python
this, other = self.align(other, join='inner', copy=False)
```

Mottl:
>iterating over the pair of the columns is exactly what the nancorr routines seeks to avoid. can we just directly use that here? this is not scaling very well.

You want to duplicate lines from pd.Series.corr() to pd.DataFrame.corrwith()? What is the reason for that?

This is the `pd.Series.corr()` implementation and it looks just fine to reuse it:
```python
def corr(self, other, method='pearson', min_periods=None):
    this, other = self.align(other, join='inner', copy=False)
    if len(this) == 0:
        return np.nan

    if method in ['pearson', 'spearman', 'kendall']:
        return nanops.nancorr(this.values, other.values, method=method,
                              min_periods=min_periods)
```

pandas/core/frame.py
Mottl: I'm adding missing columns in case `drop` param set to `False`. Not sure I understand you.

pandas/core/frame.py
Mottl: You mean the alignment of the resulting columns? They are just the same:
```python
df1 = pd.DataFrame(np.random.randn(100,3), columns=['A','B','C'])
df2 = pd.DataFrame(np.random.randn(10000,3), columns=['B','C','D'])
In [6]: df1.corrwith(df2)
Out[6]: 
A         NaN
B    0.062616
C    0.113172
D         NaN
dtype: float64

In [7]: df1.corrwith_orig(df2)
Out[7]: 
A         NaN
B    0.062616
C    0.113172
D         NaN
dtype: float64
```

Clarify, please

pandas/core/frame.py
```diff
@@ -6892,56 +6892,97 @@ def cov(self, min_periods=None):
...
+            and default 'all' if other is a Series object
+            * 'pairwise' to compute correlation between rows or columns with
+              the same name/index,
+            * 'all' to compute correlation between all possible pairs of rows
+              or columns
+            .. versionadded:: 0.24.0
         drop : boolean, default False
             Drop missing indices from result, default returns union of all
+        method : {'pearson', 'kendall', 'spearman'} or callable
+            * pearson : standard correlation coefficient
```
jreback: put these in a single quote, e.g. `` `pearson` ``, for all fo the strings

pandas/core/frame.py
jreback: this looks extremely fragile. pls align before you do anything like interating. then intersect the indexes first.

pandas/core/frame.py
jreback: pls fix this

pandas/core/frame.py
```diff
@@ -6892,56 +6892,97 @@ def cov(self, min_periods=None):
...
+
+            if not drop:  # add missing columns to the resulting Series
+                result_index = this._get_axis(1).union(other._get_axis(1))
+                correl = correl.reindex(result_index)
+            return correl
 
-        # demeaned data
-        ldem = left - left.mean()
-        rdem = right - right.mean()
-
```
jreback: benchmark you new impl vs the case that hits this. I can't imagine that iterating over the columns is better than this with any real number of columns. eg.. try 1000

jreback (changes requested):

pandas/core/frame.py
Mottl: >this looks extremely fragile. pls align before you do anything like interating. then intersect the indexes first.

What to you want to align? As I said indexes are aligned automatically in `pd.Series.corr()`:
```python
def corr(self, other, method='pearson', min_periods=None):
    this, other = self.align(other, join='inner', copy=False)
    if len(this) == 0:
        return np.nan

    if method in ['pearson', 'spearman', 'kendall']:
        return nanops.nancorr(this.values, other.values, method=method,
                              min_periods=min_periods)
```

pandas/core/frame.py
```diff
@@ -6892,56 +6892,97 @@ def cov(self, min_periods=None):
...
+            * kendall : Kendall Tau correlation coefficient
+            * spearman : Spearman rank correlation
+            * callable : callable with input two 1d ndarrays
+                 and returning a float
+            .. versionadded:: 0.24.0
+        min_periods : int, optional
+            Minimum number of observations required per pair of columns
+            to have a valid result. Currently only available for pearson
+            and spearman correlation
+            .. versionadded:: 0.24.0
```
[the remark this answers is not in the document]
Mottl: done

pandas/core/frame.py
```diff
@@ -6892,56 +6892,97 @@ def cov(self, min_periods=None):
...
         drop : boolean, default False
             Drop missing indices from result, default returns union of all
+        method : {'pearson', 'kendall', 'spearman'} or callable
+            * pearson : standard correlation coefficient
+            * kendall : Kendall Tau correlation coefficient
+            * spearman : Spearman rank correlation
+            * callable : callable with input two 1d ndarrays
+                 and returning a float
+            .. versionadded:: 0.24.0
```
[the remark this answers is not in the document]
Mottl: done

pandas/core/frame.py
```diff
@@ -6892,56 +6892,97 @@ def cov(self, min_periods=None):
...
         other : DataFrame, Series
         axis : {0 or 'index', 1 or 'columns'}, default 0
             0 or 'index' to compute column-wise, 1 or 'columns' for row-wise
+        how : {'pairwise', 'all'}, default 'pairwise' if other is a DataFrame
+            and default 'all' if other is a Series object
+            * 'pairwise' to compute correlation between rows or columns with
+              the same name/index,
+            * 'all' to compute correlation between all possible pairs of rows
+              or columns
+            .. versionadded:: 0.24.0
```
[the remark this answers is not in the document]
Mottl: done

pandas/core/frame.py
Mottl: done

pandas/core/frame.py
Mottl: done

pandas/core/frame.py
Mottl: done

pandas/tests/frame/test_analytics.py
Mottl: Added MultiIndex and non-unique tests

pandas/core/frame.py
Mottl: There is no triangle since columns and rows have different indexes (the case for `how='all'`)
Changed `.loc()` to `.iloc()`

pandas/tests/frame/test_analytics.py
Mottl: done

pandas/core/frame.py
Mottl: 30 columns:
https://github.com/pandas-dev/pandas/pull/22622#issuecomment-427882357

3000 columns:
https://github.com/pandas-dev/pandas/pull/22622#issuecomment-427893191

Mottl:
@jreback 
Please, review

jreback:
can you merge master and update to comments.

Mottl:
@jreback 
merged

doc/source/whatsnew/v0.24.0.rst
```diff
@@ -27,6 +27,7 @@ New features
   the user to override the engine's default behavior to include or omit the
   dataframe's indexes from the resulting Parquet file. (:issue:`20768`)
 - :meth:`DataFrame.corr` and :meth:`Series.corr` now accept a callable for generic calculation methods of correlation, e.g. histogram intersection (:issue:`22684`)
+- :meth:`DataFrame.corrwith` now can compute correlation matrix between all columns/rows of two DataFrame objects using ``how='all'`` parameter. It also supports ``method`` and ``min_periods`` parameters with the same behavior as :meth:`DataFrame.corr`.
```
jreback: add this PR number here; put in a reference to the docs (above)

doc/source/computation.rst
```diff
@@ -169,6 +168,7 @@ DataFrame objects.
    df1 = pd.DataFrame(np.random.randn(5, 4), index=index, columns=columns)
    df2 = pd.DataFrame(np.random.randn(4, 4), index=index[:4], columns=columns)
    df1.corrwith(df2)
+   df1.corrwith(df2, how='all')
```
jreback: can you add a separate sub-section with an example here (and ref the whatsnew to here)

pandas/core/frame.py
```diff
@@ -7022,56 +7022,154 @@ def cov(self, min_periods=None):
...
+        or DataFrame and Series objects.
 
         Parameters
         ----------
         other : DataFrame, Series
         axis : {0 or 'index', 1 or 'columns'}, default 0
             0 or 'index' to compute column-wise, 1 or 'columns' for row-wise
+        how : {'pairwise', 'all'}, default 'pairwise' if other is a DataFrame
+            and default 'all' if other is a Series object
+            * 'pairwise' to compute correlation between rows or columns with
```
jreback: this is a really confusing default

pandas/core/frame.py
```diff
@@ -7022,56 +7022,154 @@ def cov(self, min_periods=None):
...
+                or isinstance(other.columns, MultiIndex):
+            raise ValueError("MultiIndex is not supported")
 
         if axis == 1:
-            left = left.T
-            right = right.T
+            this = this.transpose()
+            other = other.transpose()
+
+        if (len(set(this.columns)) != len(this.columns)) \
```
jreback: don't use line-continuation, use parens instead

pandas/core/frame.py
```diff
@@ -7022,56 +7022,154 @@ def cov(self, min_periods=None):
...
+                or (len(set(other.columns)) != len(other.columns)):
+            raise ValueError("Non-unique columns are not supported")
+
+        if how == 'all':
+            corr = np.zeros((this.shape[1], other.shape[1]))
+
+            for i in range(len(this.columns)):
+                for j in range(len(other.columns)):
+                    corr[i, j] = this.iloc[:, i].corr(
```
jreback: is this double computing? e.g. is [i, j] == [j, i]?

pandas/core/frame.py
```diff
@@ -7022,56 +7022,154 @@ def cov(self, min_periods=None):
...
+
+            if not drop:  # add missing columns to the resulting Series
+                result_index = this._get_axis(1).union(other._get_axis(1))
+                correl = correl.reindex(result_index)
+            return correl
+
+        else:
+            raise ValueError("'how' parameter should be 'pairwise' or 'all'")
+
+    def corrmatrix(self, other, method='pearson', min_periods=1):
```
jreback: what is this?

jreback (changes requested):
let's pare this way down. only do the corrwith changes

pandas/core/frame.py
Mottl: No — It's not a square matrix. Columns != Rows.

doc/source/whatsnew/v0.24.0.rst
Mottl: Done.

doc/source/computation.rst
Mottl: Done.

pandas/core/frame.py
Mottl: Want to break the backward compatibility?

pandas/core/frame.py
Mottl: Done.

doc/source/computation.rst
```diff
@@ -159,17 +159,27 @@ compute the correlation based on histogram intersection:
...
 .. ipython:: python
 
    index = ['a', 'b', 'c', 'd', 'e']
    columns = ['one', 'two', 'three', 'four']
    df1 = pd.DataFrame(np.random.randn(5, 4), index=index, columns=columns)
    df2 = pd.DataFrame(np.random.randn(4, 4), index=index[:4], columns=columns)
-   df1.corrwith(df2)
-   df2.corrwith(df1, axis=1)
+   df1.corrwith(df2) # how='pairwise'
+   df2.corrwith(df1, axis=1) # how='pairwise'
```
jreback: axis=1 example is not useful here, remove the extra comment

doc/source/computation.rst
```diff
@@ -159,17 +159,27 @@ compute the correlation based on histogram intersection:
    frame.corr(method=histogram_intersection)
 
 A related method :meth:`~DataFrame.corrwith` is implemented on DataFrame to
-compute the correlation between like-labeled Series contained in different
-DataFrame objects.
+compute the correlation between another DataFrame or Series object with
+the columns of the same names (``how='pairwise'`` by default):
```
jreback: remove the default its not explained here and is in the doc-string

doc/source/computation.rst
```diff
@@ -159,17 +159,27 @@ compute the correlation based on histogram intersection:
...
    df2 = pd.DataFrame(np.random.randn(4, 4), index=index[:4], columns=columns)
-   df1.corrwith(df2)
-   df2.corrwith(df1, axis=1)
+   df1.corrwith(df2) # how='pairwise'
+   df2.corrwith(df1, axis=1) # how='pairwise'
+
+.. _computation.corrwith_all:
+.. versionadded:: 0.24.0
+
+... or compute the correlation matrix between another DataFrame with all possible
```
jreback: make this a sentence, what is the '...' for?

doc/source/computation.rst
```diff
@@ -159,17 +159,27 @@ compute the correlation based on histogram intersection:
...
+   df2.corrwith(df1, axis=1) # how='pairwise'
+
+.. _computation.corrwith_all:
+.. versionadded:: 0.24.0
+
+... or compute the correlation matrix between another DataFrame with all possible 
+columns combinations (``how='all'``):
+
+.. ipython:: python
+   df1.corrwith(df2, how='all')
```
jreback: need a blank line

doc/source/computation.rst
```diff
@@ -159,17 +159,27 @@ compute the correlation based on histogram intersection:
...
+   df1.corrwith(df2, axis=1, how='all')
```
jreback: don't need axis=1 example

pandas/core/frame.py
jreback: you haven't answered the question

pandas/core/frame.py
```diff
@@ -7022,56 +7022,101 @@ def cov(self, min_periods=None):
...
-            left = left.T
-            right = right.T
-
-        # demeaned data
-        ldem = left - left.mean()
-        rdem = right - right.mean()
+            this = this.transpose()
+            other = other.transpose()
+
+        if (len(set(this.columns)) != len(this.columns)
```
jreback: use ``.is_unique`` property

pandas/core/frame.py
jreback: add a comment explaining this

pandas/core/frame.py
```diff
@@ -7022,56 +7022,101 @@ def cov(self, min_periods=None):
...
+            for i in range(len(this.columns)):
+                for j in range(len(other.columns)):
+                    corr[i, j] = this.iloc[:, i].corr(
+                        other.iloc[:, j],
+                        method=method, min_periods=min_periods)
+            return DataFrame(data=corr, index=this.columns,
+                             columns=other.columns)
+
+        elif how == 'pairwise':
+            index = []
```
jreback: use the original method if ``min_periods`` is None as this has a performance cost

pandas/core/frame.py
```diff
@@ -7022,56 +7022,101 @@ def cov(self, min_periods=None):
...
+            corr = []
+            for i, col in enumerate(this.columns):
```
jreback: I would rather you call .align

pandas/tests/frame/test_analytics.py
```diff
@@ -466,6 +466,58 @@ def test_corrwith_mixed_dtypes(self):
...
         tm.assert_series_equal(result, expected)
 
+    def test_corrwith_multindex(self):
+        # PR #22622
+        df = pd.DataFrame(np.random.random((4,4)))
+        df.columns = pd.MultiIndex.from_product([[1,2],['A','B']])
+
+        # must raise an exception if columns use MultiIndex
+        try:
+            df.corrwith(df)
```
jreback: these try/except's don't make any sense, use ``with pytest.raises`` if you are trying to assert that it raises

pandas/tests/frame/test_analytics.py
```diff
@@ -466,6 +466,58 @@ def test_corrwith_mixed_dtypes(self):
         expected = pd.Series(data=corrs, index=['a', 'b'])
         tm.assert_series_equal(result, expected)
 
+    def test_corrwith_multindex(self):
+        # PR #22622
+        df = pd.DataFrame(np.random.random((4,4)))
```
jreback: you have lots of pep8 issues, run ci/code_checks.sh

jreback (changes requested):
you have no tests for Series added, its is unclear what adding the ``method=`` does for Series, and it is a confusing default.

Mottl:
@jreback
Let's do the things simpler — just fix what you want.

jreback:
@Mottl the point is that the community does these PRs
if you don’t want to complete then close this

Issue #16615, opened 2017-06-06
read_sas fails when values in SAS dataset contains a new line symbol


Grinvydas:
#### Problem description

I was trying to read a SAS dataset with pandas 0.19.2. It was not successful, with an error: ValueError('Length of values does not match length of ' 'index').

After some research I came up with the idea, that new line symbol in one of the character values creates this error.

I removed new line and carriage return symbols from column values in SAS data and read_sas finished without errors after that. I assume that read_sas treats any new line symbol it encounters as new line of a table.

#### Expected Output

read_sas could translate new line symbols found in column values to space and finish without an error.

chris-b1:
Can you provide a data set that reproduces the issue?  Either something existing online, or you can attach to the issue.

Grinvydas:
As always, when I try to re-create, it reads the file fine. I'll try to fetch the original file tomorrow and update.

Grinvydas:
After some further investigation I think the problem could be elsewhere, not in New line or carriage return symbol. Actually all I needed is just to re-create the file with simple data step and after that the new data set is read properly by read_sas

data new_file;
set old;
run;

I'm attaching the problematic file which gives the error below:

Traceback (most recent call last):
  File "isItemDsUpdated.py", line 27, in <module>
    status_log = pd.read_sas(sasfile, encoding='latin1').dropna()
  File "/projects/karegr01/anaconda3/lib/python3.6/site-packages/pandas/io/sas/sasreader.py", line 61, in read_sas
    data = reader.read()
  File "/projects/karegr01/anaconda3/lib/python3.6/site-packages/pandas/io/sas/sas7bdat.py", line 614, in read
    rslt = self._chunk_to_dataframe()
  File "/projects/karegr01/anaconda3/lib/python3.6/site-packages/pandas/io/sas/sas7bdat.py", line 663, in _chunk_to_dataframe
    rslt[name] = self._string_chunk[js, :]
  File "/projects/karegr01/anaconda3/lib/python3.6/site-packages/pandas/core/frame.py", line 2419, in __setitem__
    self._set_item(key, value)
  File "/projects/karegr01/anaconda3/lib/python3.6/site-packages/pandas/core/frame.py", line 2485, in _set_item
    value = self._sanitize_column(key, value)
  File "/projects/karegr01/anaconda3/lib/python3.6/site-packages/pandas/core/frame.py", line 2656, in _sanitize_column
    value = _sanitize_index(value, self.index, copy=False)
[...]

PR #22628, opened 2018-09-07
BUG: Some sas7bdat files with many columns are not parseable by read_sas

troels:
- [X] tests added / passed
- [X] whatsnew entry

The reason is that column definitions may be split up into different pages.
Allow column information to be parsed from different pages
and add a test for it.


gfyoung:
@troels : Thanks for the report!  Do you mind sharing in this PR what output you get without your patch for reference?  A small, reproducible example would be great.

pandas/tests/io/sas/test_sas7bdat.py
```diff
@@ -183,6 +183,22 @@ def test_date_time(datapath):
...
+    fname = datapath("io", "sas", "data", "many_columns.csv")
+    df0 = pd.read_csv(fname, encoding='latin-1',
+                      dtype={
+                          'DATASRC': np.dtype('S'),
+                          'PDDOCID': np.dtype('S'),
+                          'VISIT_NO': np.dtype('S'),
+                          'ecgrtxt': np.dtype('S'),
+                          'ecgrtime': np.dtype('d'),
+                          'nvitl1': np.dtype('d')
+                      })
```
gfyoung: Generally, we would prefer to construct the expected DataFrame from scratch.  This is one of the motivations for having a small, reproducible example.

doc/source/whatsnew/v0.23.5.txt
```diff
@@ -49,3 +49,5 @@ Bug Fixes
 **I/O**
 
 - Bug in :func:`read_csv` that caused it to raise ``OverflowError`` when trying to use 'inf' as ``na_value`` with integer index column (:issue:`17128`)
+
+- Bug in :func:`read_sas` that caused it to fail parsing some sas7bdat files with many columns
```
gfyoung: * This should go in `0.24.0`
* Reference your PR as the issue number

pandas/io/sas/sas7bdat.py
```diff
@@ -102,6 +103,15 @@ def __init__(self, path_or_buf, index=None, convert_dates=True,
...
         self._parse_metadata()
 
+    def column_data_lengths(self):
+        return np.asarray(self._column_data_lengths, dtype=np.int64)
+
+    def column_data_offsets(self):
+        return np.asarray(self._column_data_offsets, dtype=np.int64)
+
+    def column_types(self):
+        return np.asarray(self._column_types, dtype=np.dtype('S1'))
```
gfyoung: Can we have docstrings for these methods?

troels:
Hi @gfyoung 

So the test case I added will fail in the current version of pandas, with the following error (these files can't be parsed):

pandas/tests/io/sas/test_sas7bdat.py ............F.                                              [100%]

=============================================== FAILURES ===============================================
__________________________________________ test_many_columns ___________________________________________

datapath = <function datapath.<locals>.deco at 0x7f53f5a47f28>

    def test_many_columns(datapath):
        fname = datapath("io", "sas", "data", "many_columns.sas7bdat")
>       df = pd.read_sas(fname, encoding='latin-1')

pandas/tests/io/sas/test_sas7bdat.py:188: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pandas/io/sas/sasreader.py:68: in read_sas
    data = reader.read()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pandas.io.sas.sas7bdat.SAS7BDATReader object at 0x7f53f5899860>, nrows = 3

    def read(self, nrows=None):
    
        if (nrows is None) and (self.chunksize is not None):
            nrows = self.chunksize
        elif nrows is None:
            nrows = self.row_count
    
        if len(self.column_types) == 0:
            self.close()
>           raise EmptyDataError("No columns to parse from file")
E           pandas.errors.EmptyDataError: No columns to parse from file

pandas/io/sas/sas7bdat.py:604: EmptyDataError
================================= 1 failed, 13 passed in 18.88 seconds =================================

pandas/tests/io/sas/test_sas7bdat.py
troels: Hi @gfyoung, I'd generally agree, but in this case, I don't have a good way of constructing another test example. I could perhaps use a more type-safe format than csv, as the reference file, do you have any good ideas?

gfyoung: Gotcha.  Let's see what you have here: one header and three rows IIUC.  Do you need that many columns to construct a breaking example?

troels:
Hi @gfyoung 

I added a commit fixing #16615 too. I hope it's ok to continue using the same PR, as the two pieces of code build upon each other.

pandas/tests/io/sas/test_sas7bdat.py
troels: Yeah. The problem occurs as a consequence of the number of columns. Otherwise SAS will save all column info on one page, and the problem will not be replicated.

gfyoung: Hmm...I see.  Alright, let's leave this as is for now then. 👍

gfyoung:
> I added a commit fixing #16615 too. I hope it's ok to continue using the same PR, as the two pieces of code build upon each other.

@troels : In this case, it should be fine.  Thanks for letting me know!

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -690,7 +690,8 @@ I/O
 - :func:`read_html()` no longer ignores all-whitespace ``<tr>`` within ``<thead>`` when considering the ``skiprows`` and ``header`` arguments. Previously, users had to decrease their ``header`` and ``skiprows`` values on such tables to work around the issue. (:issue:`21641`)
 - :func:`read_excel()` will correctly show the deprecation warning for previously deprecated ``sheetname`` (:issue:`17994`)
 - :func:`read_csv()` will correctly parse timezone-aware datetimes (:issue:`22256`)
--
+- :func:`read_sas` will correctly parse sas7bdat files with many columns (:issue:`22628`)
+- :func:`read_sas` will correctly parse sas7bdat files with odd data page types (:issue:`16615`)
```
jreback: can you expand on 'odd' here?

pandas/io/sas/sas.pyx
```diff
@@ -437,7 +437,7 @@ cdef class Parser(object):
             elif column_types[j] == column_type_string:
                 # string
                 string_chunk[js, current_row] = np.array(source[start:(
-                    start + lngt)]).tostring().rstrip()
+                    start + lngt)]).tostring().rstrip(b"\x00 ")
```
jreback: there is a space here as well?

pandas/tests/io/sas/test_sas7bdat.py
```diff
@@ -183,6 +183,28 @@ def test_date_time(datapath):
...
+    tm.assert_frame_equal(df, df0)
+
+
+def test_inconsistent_number_of_rows(datapath):
```
jreback: can you add the issue number as a comment

pandas/tests/io/sas/test_sas7bdat.py
```diff
@@ -183,6 +183,28 @@ def test_date_time(datapath):
     tm.assert_frame_equal(df, df0)
 
 
+def test_many_columns(datapath):
+    fname = datapath("io", "sas", "data", "many_columns.sas7bdat")
```
jreback: same

jreback (changes requested):

pandas/io/sas/sas.pyx
troels: Hi @jreback 

Yes, there is supposed to be a space, just like many other places in sas7bdat.py. 

Apparently SAS can both pad with zeros and spaces to the right. (Maybe depending on version or something else).

troels:
The test failure here is quite unrelated to the PR.

troels:
Hi @jreback 

Anything else preventing this from being merged?

gfyoung (approved):
I'm not @jreback , but I don't think there are any barriers to merging this.  LGTM!

jreback (approved):

jreback:
thanks!

troels:
Thanks both of you :)

Issue #22180, opened 2018-08-02
[DISC] Use type annotations in cython


jbrockmendel:
Regardless of whether run in py2/py3, cython supports modern type annotations.  A lot of _libs could be made valid python by using these.

Upsides:
- linting!
- get experience with these, will help in deciding if we want to use them in the non-cython code-base.
- may lower barrier-to-entry for contributors not comfortable with cython

Downsides:
- If it works, don't fix it
- Until fully implemented, we'd have mix-and-match conventions
- cython's pure-python mode has more rough edges than its regular mode
- in some cases the valid-python versions get pretty verbose:

```
cdef inline bint is_integer_object(object obj) nogil:
    cdef:
         int foo

vs.

@cython.cfunc
@cython.inline
@cython.nogil
def is_integer_object(tz: object) -> bint:
    foo: int
```

The main upside I see is the linting.  My inclination is to try to implement it in a subset of files where we can get close to pure-python as a test run.

Thoughts?

WillAyd:
Is this a new feature? If it's as advertised I'm +1 as it would probably even make it easier for newcomers to understand the Cython code base

gfyoung:
+1 from me as well.

jbrockmendel:
Opened cython#2529 to discuss some syntax combinations that are parsed incorrectly (or not supported, but not made explicit in the docs).  I think this is worth revisiting a couple of cython releases down the road, but not just yet.  Closing.

PR #22638, opened 2018-09-08
pythonize cython code

jbrockmendel:
Use python-style type annotations instead of cython-style in a few places.

Use python-style isinstance checks in cases where cython will automatically optimize them into C calls.

Part of the hope is that we can get the code close enough to valid python that we can trick flake8 into working on it.


jreback:
thanks. maybe add a note somewhere that we *can* use type annotation in cython files (*maybe* at the top of each file?)

jorisvandenbossche:
> Use python-style type annotations instead of cython-style in a few places.

Sorry, why do we think this is an improvement? Now we have two ways of doing things? 
Can you provide a bit more background / reasoning?

jbrockmendel:
Some discussion in #22180.

Main motivation: I _think_ that if we can get the code close enough to valid python, we can trick flake8 into running on it.

PR #22641, opened 2018-09-08
DOC: improve doc string for .aggregate and .transform

topper-123:
Since #21224, operations using ``axis=1`` in df.aggregate and df.transform now work the same as when axis=0.

This PR updates the methods' doc strings to reflect the new reality. For example, we can now pass a dict to DataFrame.agg/transform when ``axis=1`` also, and ``DataFrame.transform`` now has an ``axis`` parameter.

There's a minor API change, as ``Series.transform`` should have a ``axis=0`` parameter to have the same API as ``Series.aggregate``.

Also some related minor clarifications.

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -5995,33 +5995,6 @@ class DataFrame(NDFrame):
 
     agg = aggregate
 
-    _transform_doc = dedent("""
-    Examples
-    --------
-    >>> df = pd.DataFrame(np.random.randn(10, 2), columns=['A', 'B'],
-    ...                   index=pd.date_range('1/1/2000', periods=10))
-    >>> df.iloc[3:7] = np.nan
-
-    >>> df.transform(lambda x: (x - x.mean()) / x.std())
-                       A         B
-    2000-01-01  0.579457  1.236184
-    2000-01-02  0.370357 -0.605875
-    2000-01-03  1.455756 -0.277446
-    2000-01-04       NaN       NaN
-    2000-01-05       NaN       NaN
-    2000-01-06       NaN       NaN
-    2000-01-07       NaN       NaN
-    2000-01-08 -0.498658  1.274522
-    2000-01-09 -0.540524 -1.012676
-    2000-01-10 -1.366388 -0.614710
-
-    See also
-    --------
-    pandas.DataFrame.aggregate
-    pandas.DataFrame.apply
-    """)
-
-    @Appender(_transform_doc)
     @Appender(_shared_docs['transform'] % _shared_doc_kwargs)
     def transform(self, func, axis=0, *args, **kwargs):
         axis = self._get_axis_number(axis)
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -5995,6 +5995,33 @@ class DataFrame(NDFrame):
 
     agg = aggregate
 
+    _transform_doc = dedent("""
+    Examples
+    --------
+    >>> df = pd.DataFrame(np.random.randn(10, 2), columns=['A', 'B'],
+    ...                   index=pd.date_range('1/1/2000', periods=10))
+    >>> df.iloc[3:7] = np.nan
+
+    >>> df.transform(lambda x: (x - x.mean()) / x.std())
+                       A         B
+    2000-01-01  0.579457  1.236184
+    2000-01-02  0.370357 -0.605875
+    2000-01-03  1.455756 -0.277446
+    2000-01-04       NaN       NaN
+    2000-01-05       NaN       NaN
+    2000-01-06       NaN       NaN
+    2000-01-07       NaN       NaN
+    2000-01-08 -0.498658  1.274522
+    2000-01-09 -0.540524 -1.012676
+    2000-01-10 -1.366388 -0.614710
+
+    See also
+    --------
+    pandas.DataFrame.aggregate
+    pandas.DataFrame.apply
+    """)
+
+    @Appender(_transform_doc)
     @Appender(_shared_docs['transform'] % _shared_doc_kwargs)
     def transform(self, func, axis=0, *args, **kwargs):
         axis = self._get_axis_number(axis)
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -109,10 +109,9 @@ from pandas.core.config import get_option
 _shared_doc_kwargs = dict(
     axes='index, columns', klass='DataFrame',
     axes_single_arg="{0 or 'index', 1 or 'columns'}",
-    axis="""
-    axis : {0 or 'index', 1 or 'columns'}, default 0
-        - 0 or 'index': apply function to each column.
-        - 1 or 'columns': apply function to each row.""",
+    axis="""axis : {0 or 'index', 1 or 'columns'}, default 0
+        If 0 or 'index': apply function to each column.
+        If 1 or 'columns': apply function to each row.""",
     optional_by="""
         by : str or list of str
             Name or list of names to sort by.
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4545,17 +4545,16 @@ class NDFrame(PandasObject, SelectionMixin):
 
     Parameters
     ----------
-    func : function, string, dictionary, or list of string/functions
+    func : function, string, list of string/functions or dictionary
         Function to use for aggregating the data. If a function, must either
-        work when passed a %(klass)s or when passed to %(klass)s.apply. For
-        a DataFrame, can pass a dict, if the keys are DataFrame column names.
+        work when passed a %(klass)s or when passed to %(klass)s.apply.
 
         Accepted combinations are:
 
-        - string function name.
-        - function.
-        - list of functions.
-        - dict of column names -> functions (or list of functions).
+        - string function name
+        - function
+        - list of functions and/or function names
+        - dict of axis labels -> functions, function names or list of such
     %(axis)s
     *args
         Positional arguments to pass to `func`.
@@ -4581,43 +4580,32 @@ class NDFrame(PandasObject, SelectionMixin):
 
     Parameters
     ----------
-    func : callable, string, dictionary, or list of string/callables
-        To apply to column
+    func : function, string, list of string/functions or dictionary
+        Function to use for transforming the data. If a function, must either
+        work when passed a %(klass)s or when passed to %(klass)s.apply.
+        The function (or each function in a list/dict) must return an
+        object with the same length for the provided axis as the
+        calling %(klass)s.
 
-        Accepted Combinations are:
+        Accepted combinations are:
 
         - string function name
         - function
-        - list of functions
-        - dict of column names -> functions (or list of functions)
+        - list of functions and/or function names
+        - dict of axis labels -> functions, function names or list of such
+    %(axis)s
+    *args
+        Positional arguments to pass to `func`.
+    **kwargs
+        Keyword arguments to pass to `func`.
 
     Returns
     -------
     transformed : %(klass)s
 
-    Examples
-    --------
-    >>> df = pd.DataFrame(np.random.randn(10, 3), columns=['A', 'B', 'C'],
-    ...                   index=pd.date_range('1/1/2000', periods=10))
-    df.iloc[3:7] = np.nan
-
-    >>> df.transform(lambda x: (x - x.mean()) / x.std())
-                       A         B         C
-    2000-01-01  0.579457  1.236184  0.123424
-    2000-01-02  0.370357 -0.605875 -1.231325
-    2000-01-03  1.455756 -0.277446  0.288967
-    2000-01-04       NaN       NaN       NaN
-    2000-01-05       NaN       NaN       NaN
-    2000-01-06       NaN       NaN       NaN
-    2000-01-07       NaN       NaN       NaN
-    2000-01-08 -0.498658  1.274522  1.642524
-    2000-01-09 -0.540524 -1.012676 -0.828968
-    2000-01-10 -1.366388 -0.614710  0.005378
-
-    See also
-    --------
-    pandas.%(klass)s.aggregate
-    pandas.%(klass)s.apply
+    Raises
+    ------
+    ValueError: if the returned %(klass)s has a different length than self.
     """)
 
     # ----------------------------------------------------------------------
@@ -9401,7 +9389,7 @@ class NDFrame(PandasObject, SelectionMixin):
 
         cls.ewm = ewm
 
-    @Appender(_shared_docs['transform'] % _shared_doc_kwargs)
+    @Appender(_shared_docs['transform'] % dict(axis="", **_shared_doc_kwargs))
     def transform(self, func, *args, **kwargs):
         result = self.agg(func, *args, **kwargs)
         if is_scalar(result) or len(result) != len(self):
--- a/pandas/core/generic.py
+++ b/pandas/core/generic.py
@@ -4545,7 +4545,7 @@ class NDFrame(PandasObject, SelectionMixin):
 
     Parameters
     ----------
-    func : function, string, list of string/functions or dictionary
+    func : function, string, list of functions and/or strings or dict
         Function to use for aggregating the data. If a function, must either
         work when passed a %(klass)s or when passed to %(klass)s.apply.
 
@@ -4563,7 +4563,7 @@ class NDFrame(PandasObject, SelectionMixin):
 
     Returns
     -------
-    aggregated : %(klass)s
+    pandas.%(klass)s
 
     Notes
     -----
@@ -4573,14 +4573,14 @@ class NDFrame(PandasObject, SelectionMixin):
     """)
 
     _shared_docs['transform'] = ("""
-    Call function producing a like-indexed %(klass)s
-    and return a %(klass)s with the transformed values
+    Call ``func`` on self producing a %(klass)s with transformed values
+    and that has the same axis length as self.
 
     .. versionadded:: 0.20.0
 
     Parameters
     ----------
-    func : function, string, list of string/functions or dictionary
+    func : function, string, list of functions and/or strings or dict
         Function to use for transforming the data. If a function, must either
         work when passed a %(klass)s or when passed to %(klass)s.apply.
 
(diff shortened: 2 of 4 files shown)


pandas/core/series.py
```diff
@@ -3100,6 +3100,39 @@ def aggregate(self, func, axis=0, *args, **kwargs):
...
+
+    See also
+    --------
+    pandas.Series.aggregate
+    pandas.Series.apply
+    """)
+
+    @Appender(_transform_doc)
+    @Appender(generic._shared_docs['transform'] % _shared_doc_kwargs)
+    def transform(self, func, axis=0, *args, **kwargs):
```
WillAyd: Hmm generally not sure its worth changing actual implementation for docstrings. If this is solely to isolate the various Examples I'd think it preferable to just have one shared Example docstring that covers Series and DataFrame rather than making code changes like this

pandas/core/generic.py
```diff
@@ -4581,43 +4580,32 @@ def pipe(self, func, *args, **kwargs):
 
     Parameters
     ----------
-    func : callable, string, dictionary, or list of string/callables
-        To apply to column
+    func : function, string, list of string/functions or dictionary
+        Function to use for transforming the data. If a function, must either
+        work when passed a %(klass)s or when passed to %(klass)s.apply.
+        The function (or each function in a list/dict) must return an
```
WillAyd: IMO would be better served in a Notes section along with description of what happens if something is returned that doesn't meet this requirement

WillAyd (changes requested):

pandas/core/generic.py
topper-123: Hmm, it is already described in the Raises section, so I could just drop this part here...

pandas/core/series.py
topper-123: The doc string is currently inherited from NDFrame, but it's not pretty IMO, see [here](http://pandas-docs.github.io/pandas-docs-travis/generated/pandas.Series.transform.html)

The issues are:
* The current doc string discussed NDFrames, not Series, which is confusing,
* the links in the SeeAlso don't work,
* ``Series.transform``'s signature is in master different from the signature for ``Series.agg`` and ``Series.apply`` in that it misses the ``axis`` parameter that the other two have. This may/may not be a problem (I actually couldn't produce a bug caused by this), but the signature should be consistent amongst the three methods, and I agree with the current design for ``agg`` and ``apply`` (i.e. to have an axis parameter).

Thoughts? I can remove this if there's not consensus.

WillAyd: I'd say just use substitution for the class name (i.e. Series or DataFrame) and update the See Also links to point to both the Series and DataFrame methods. One of those will obviously be self referencing, but we've done this in other places as well

topper-123: Ok, I see what you mean. I've made a new commit with common examples but the SeeAlso can actually be made correct if Series gets it's own transform method (which is needed for the signature issue to be resolved).

pandas/core/series.py
```diff
@@ -3100,6 +3100,12 @@ def aggregate(self, func, axis=0, *args, **kwargs):
 
     agg = aggregate
 
+    @Appender(generic._shared_docs['transform'] % _shared_doc_kwargs)
+    def transform(self, func, axis=0, *args, **kwargs):
+        # Validate the axis parameter
+        self._get_axis_number(axis)
```
WillAyd: What's the point of this statement?

topper-123: This checks that the value passed to axis is 0 or “index”, else an exception is raised. So, a minor check for consistency.

topper-123:
all tests have passed. Is this ok?

WillAyd (approved):
@datapythonista can you take a look?

pandas/core/generic.py
```diff
@@ -4545,17 +4545,16 @@ def pipe(self, func, *args, **kwargs):
 
     Parameters
     ----------
-    func : function, string, dictionary, or list of string/functions
+    func : function, string, list of string/functions or dictionary
```
datapythonista: We try to use only Python types in this row, and be consistent with the format. `func : function, str, list or dict` would be the preferred format (or specifying the list types, like `list of str, list of function`).

pandas/core/generic.py
```diff
@@ -4581,38 +4580,61 @@ def pipe(self, func, *args, **kwargs):
 
     Parameters
     ----------
-    func : callable, string, dictionary, or list of string/callables
-        To apply to column
+    func : function, string, list of string/functions or dictionary
```
datapythonista: same as before

pandas/core/generic.py
```diff
@@ -4581,38 +4580,61 @@ def pipe(self, func, *args, **kwargs):
...
+        - dict of axis labels -> functions, function names or list of such
+    %(axis)s
+    *args
+        Positional arguments to pass to `func`.
+    **kwargs
+        Keyword arguments to pass to `func`.
 
     Returns
     -------
     transformed : %(klass)s
```
datapythonista: The `transformed` name doesn't add a lot of value. Can you just leave the type here, and add a description in the next line. Same for `aggregate`.

pandas/core/generic.py
```diff
@@ -4581,38 +4580,61 @@ def pipe(self, func, *args, **kwargs):
...
 
+    Raises
+    ------
+    ValueError: if the returned %(klass)s has a different length than self.
```
datapythonista: Not sure if sphinx requires a space before the colon. Did you render the docstring? Also, if you can capitalize the sentence `If the returned...`.

Then, I don't find the description very clear on what the user did wrong, and how they should fix the problem. Do you think you can add a bit more information?

pandas/core/generic.py
```diff
@@ -4581,38 +4580,61 @@ def pipe(self, func, *args, **kwargs):
...
+    >>> s = pd.Series(range(5))
+    >>> s.transform([np.sqrt, np.exp])
+           sqrt        exp
+    0  0.000000   1.000000
+    1  1.000000   2.718282
+    2  1.414214   7.389056
+    3  1.732051  20.085537
+    4  2.000000  54.598150
 
     See also
```
datapythonista: `See Also` should be placed before the examples, and should have a capital `A` in `Also`.

pandas/core/generic.py
```diff
@@ -4581,38 +4580,61 @@ def pipe(self, func, *args, **kwargs):
...
-    2000-01-04       NaN       NaN       NaN
-    2000-01-05       NaN       NaN       NaN
-    2000-01-06       NaN       NaN       NaN
-    2000-01-07       NaN       NaN       NaN
-    2000-01-08 -0.498658  1.274522  1.642524
-    2000-01-09 -0.540524 -1.012676 -0.828968
-    2000-01-10 -1.366388 -0.614710  0.005378
+                       A         B
+    2000-01-01 -1.143001  1.143001
+    2000-01-02 -0.889001  0.889001
```
datapythonista: I think this example could be simplified, so users don't need to spend a lot of time understanding what's going on. I think 3 or 4 rows should be enough to illustrate `transform`. Also, standardizing the values may be a more real-world example, but I don't think anybody is able to do the mental math, to compare what they think the function is doing, with what we show here. Also, I would use the default index, as using dates eems to have a meaning, and it's misleading.

So, I'd do:
- A much shorter DataFrame (e.g. 3 rows)
- A much simpler function (e.g. `lambda x: x + 1`)
- Use the default index

pandas/core/generic.py
```diff
@@ -4581,38 +4580,61 @@ def pipe(self, func, *args, **kwargs):
...
+    2000-01-05       NaN       NaN
+    2000-01-06       NaN       NaN
+    2000-01-07       NaN       NaN
+    2000-01-08  0.635001 -0.635001
+    2000-01-09  0.889001 -0.889001
+    2000-01-10  1.143001 -1.143001
+
+    It is only required for the axis specified in the ``axis`` parameter
+    to have the same length for output and for self. The other axis may have a
+    different length:
```
datapythonista: I don't quite understand what you mean here. We're not specifying the `axis` parameter in the example.

datapythonista (changes requested):
The changes look great. I added some comments, mainly related to the original docstring, that I think would make them follow our standards and be more clear.

Besides the comments, I think the previous examples were failing the doctests, and yours pass, so we should be able to stop skipping these docstrings in `ci/doctest.sh`.

Also, did you run `./scripts/validate_docstrings.py pandas.Series.transform`...?

pandas/core/generic.py
topper-123: Alright, but the list can contain bot functions and strings (``[np.exp, 'sqrt']``). Maybe ``func : function, string, list of functions and/or strings or dict``?

pandas/core/generic.py
topper-123: Yeah, that's a good point, changed.  For the Series, I've also shortened it, but kept ``s.transform([np.sqrt, np.exp])``, as I think that also is quite simple.

pandas/core/generic.py
topper-123: Code-wise ``transform`` is the same as ``aggregate``, except there's a check that the result has the same length as self. I try to bring out this requirement for ``transform``.

I've tried to word the doc string differently to highlight this better.

topper-123:
@datapythonista , I'm not familiar with ``ci/doctest.sh``. Are you saying I should remove the two ``-transform`` parameters from that file?

Wrt. other stuff, I've made various changes according to the comments (great comments BTW).

pandas/core/generic.py
datapythonista: At some point I'd like to validate that all the types provided in that line are from a subset, to avoid typos and inconsistencies. I understand your point about mixing both, but I'd prefer `function, str, list or dict` (note `str`, the Python type, instead of `string`), and then provide the details about mixing strings and everything else in the description.

pandas/core/generic.py
```diff
@@ -4564,7 +4563,7 @@ def pipe(self, func, *args, **kwargs):
 
     Returns
     -------
-    aggregated : %(klass)s
+    pandas.%(klass)s
```
datapythonista: I don't think we prefix `Series` and `DataFrame` with `pandas`. I'd just leave with the class for consistency.

pandas/core/generic.py
```diff
@@ -4545,17 +4545,16 @@ def pipe(self, func, *args, **kwargs):
...
 
         Accepted combinations are:
 
-        - string function name.
-        - function.
-        - list of functions.
-        - dict of column names -> functions (or list of functions).
+        - string function name
+        - function
+        - list of functions and/or function names
```
datapythonista: may be you can add the example you wrote in the comments? I think that would make much clearer that we can mix both

pandas/core/generic.py
```diff
@@ -4574,50 +4573,61 @@ def pipe(self, func, *args, **kwargs):
...
+    Call ``func`` on self producing a %(klass)s with transformed values
+    and that has the same axis length as self.
 
     .. versionadded:: 0.20.0
 
     Parameters
     ----------
-    func : callable, string, dictionary, or list of string/callables
-        To apply to column
+    func : function, string, list of functions and/or strings or dict
```
datapythonista: `str` instead of `string`

pandas/core/generic.py
```diff
@@ -4574,50 +4573,61 @@ def pipe(self, func, *args, **kwargs):
...
         - function
-        - list of functions
-        - dict of column names -> functions (or list of functions)
+        - list of functions and/or function names
+        - dict of axis labels -> functions, function names or list of such.
+    %(axis)s
+    *args
+        Positional arguments to pass to `func`.
+    **kwargs
+        Keyword arguments to pass to `func`.
```
datapythonista: If you prefer, numpydoc also accepts having both in one line:
```
*args, **kwargs
    Arguments to pass to `func`.
```

pandas/core/generic.py
```diff
@@ -4574,50 +4573,61 @@ def pipe(self, func, *args, **kwargs):
...
 
     Returns
     -------
-    transformed : %(klass)s
+    pandas.%(klass)s
```
datapythonista: Same as before.

pandas/core/generic.py
```diff
@@ -4574,50 +4573,61 @@ def pipe(self, func, *args, **kwargs):
...
+        A %(klass)s that must have the same length as self.
 
-    Examples
+    Raises
+    ------
+    ValueError : if the returned %(klass)s has a different length than self.
```
datapythonista: Capital `I` in `if`.

pandas/core/generic.py
```diff
@@ -4574,50 +4573,61 @@ def pipe(self, func, *args, **kwargs):
...
-    2000-01-05       NaN       NaN       NaN
-    2000-01-06       NaN       NaN       NaN
-    2000-01-07       NaN       NaN       NaN
-    2000-01-08 -0.498658  1.274522  1.642524
-    2000-01-09 -0.540524 -1.012676 -0.828968
-    2000-01-10 -1.366388 -0.614710  0.005378
-
-    See also
+    pandas.%(klass)s.agg : only perform aggregating type operations
+    pandas.%(klass)s.apply : Invoke function on a Series
```
datapythonista: No need for `pandas.`. Capital `O` in `only`. Descriptions finishing with period.

pandas/core/generic.py
```diff
@@ -4574,50 +4573,61 @@ def pipe(self, func, *args, **kwargs):
...
+
+    Examples
     --------
-    pandas.%(klass)s.aggregate
-    pandas.%(klass)s.apply
+    >>> df = pd.DataFrame({'A': range(3), 'B': range(1, 4)})
+    >>> df.transform(lambda x: x + 1)
```
datapythonista: I would display `df` first before the transform. So, it's immediate to compare the original data and the transformed data, even without reading the constructor of the DataFrame.

pandas/core/generic.py
```diff
@@ -4574,50 +4573,61 @@ def pipe(self, func, *args, **kwargs):
...
+       A  B
+    0  1  2
+    1  2  3
+    2  3  4
+
+    Even though the resulting %(klass)s must have the length as the input
+    %(klass)s, it is possible to provide several input functions:
+
+    >>> s = pd.Series(range(3))
+    >>> s.transform([np.sqrt, np.exp])
```
datapythonista: I'd also show `s` before the transform.

datapythonista (commented):
Looks great, just added few comments about the formatting.

pandas/core/generic.py
topper-123: > may be you can add the example you wrote in the comment?

Not sure I understand whst you mean here, could you expand?

pandas/core/generic.py
topper-123: Alright, I did that, though my own preference would be to not use star arguments and use ``args=None, kwargs=None`` instead to better have a distinction what gets passed to ``func`` and what doesn't get passed on. But that's for a whole another discussion, and may be too late now :-)

pandas/core/generic.py
```diff
@@ -4574,50 +4571,68 @@ def pipe(self, func, *args, **kwargs):
...
+    Call ``func`` on self producing a %(klass)s with transformed values
+    and that has the same axis length as self.
 
     .. versionadded:: 0.20.0
 
     Parameters
     ----------
-    func : callable, string, dictionary, or list of string/callables
-        To apply to column
+    func : function, str, list of functions and/or strings or dict
```
datapythonista: can you leave it as the previous?

pandas/core/generic.py
```diff
@@ -4545,26 +4545,23 @@ def pipe(self, func, *args, **kwargs):
...
-        Positional arguments to pass to `func`.
-    **kwargs
-        Keyword arguments to pass to `func`.
+    *args, **kwargs
+        Arguments to pass to `func`.
 
     Returns
     -------
-    aggregated : %(klass)s
+    %(klass)s
```
datapythonista: If you don't mind adding a description to what is returned, that would be great.

pandas/core/generic.py
```diff
@@ -4574,50 +4571,68 @@ def pipe(self, func, *args, **kwargs):
...
-    2000-01-05       NaN       NaN       NaN
-    2000-01-06       NaN       NaN       NaN
-    2000-01-07       NaN       NaN       NaN
-    2000-01-08 -0.498658  1.274522  1.642524
-    2000-01-09 -0.540524 -1.012676 -0.828968
-    2000-01-10 -1.366388 -0.614710  0.005378
-
-    See also
+    %(klass)s.agg : Only perform aggregating type operations
+    %(klass)s.apply : Invoke function on a Series
```
datapythonista: Can you finish these descriptions with a period? I think we need to add this validation to the script, but it's what we've been doing.

datapythonista (commented):
Looks great, I think the new docstrings are very clear and informative. There are couple of minor things regarding conventions we follow, but otherwise lgtm.

pandas/core/generic.py
topper-123: The return type  is actually a bit more complex than expected:

```python
>>> df = pd.DataFrame({'A': range(5), 'B': 5})
>>> df.agg(['sum', 'mean'])  -> a DataFrame
>>>df.agg('sum')  -> a Series
>>> df.A.agg(['sum', 'mean'])  -> a Series
>>>df.A.agg('sum')  -> a scalar
```

I suggest we should do a ``%(return_value)s`` in the Returns section and add as appropriate in Series.agg and DataFrame.agg appenders.

datapythonista: I'd simply have `DataFrame, Series or scalar` as the type, and then in the return description explain a bit better what is being returned in each case. May be there are better options, but this is what I've been doing, and I personally prefer to not overcomplicate the docstrings with variables, unless it really adds significant value. But that's my opinion, it's you call.

pandas/core/frame.py
```diff
@@ -109,10 +109,9 @@
...
     axes='index, columns', klass='DataFrame',
     axes_single_arg="{0 or 'index', 1 or 'columns'}",
-    axis="""
-    axis : {0 or 'index', 1 or 'columns'}, default 0
-        - 0 or 'index': apply function to each column.
-        - 1 or 'columns': apply function to each row.""",
+    axis="""axis : {0 or 'index', 1 or 'columns'}, default 0
+        If 0 or 'index': apply function to each column.
+        If 1 or 'columns': apply function to each row.""",
     optional_by="""
```
topper-123: ``axis`` doesn't line up with the previous line in master, so generated doc string doesn't show axis etc, see [here in master](https://pandas.pydata.org/pandas-docs/stable/generated/pandas.Series.transform.html).

The above change lines it properly up, so it shows up as expected.

datapythonista (approved):
lgtm. Thanks for all the work on this @topper-123

jreback:
thanks @topper-123

PR #22644, opened 2018-09-09
API/ENH: tz_localize handling of nonexistent times: rename keyword + add shift option

### LINKED ISSUE #8917: Timestamp.tz_localize() NonExistentTimeError handling
I'm trying to use pandas to speed up timezone conversions on many datetimes and I can't get around `NonExistentTimeErrors`. Pandas `tz_localize()` seems to ignore the `ambiguous` argument for the non existent time case.

Example:

``` python
import pytz
import pandas

tz = pytz.timezone('Europe/Warsaw')
non_existent = datetime.datetime(2015, 3, 29, 2, 30)

tz.normalize(tz.localize(non_existent))
#2015-03-29 03:30:00+02:00

tz.normalize(tz.localize(non_existent, is_dst=False))
#2015-03-29 03:30:00+02:00

pandas.Timestamp(non_existent).tz_localize(tz)
# NonExistentTimeError: 2015-03-29 02:30:00

pandas.Timestamp(non_existent).tz_localize(tz, ambiguous=0)
# NonExistentTimeError: 2015-03-29 02:30:00
```

It would be nice if the `ambiguous` argument worked the same as `is_dst` in pytz. As it is now, it's impossible afaik to reliably localize a series of datetimes using pandas, since any of them might cause a `NonExistentTimeError` and there's no way of guiding pandas what to do with such datetimes.

```
INSTALLED VERSIONS
------------------
commit: None
python: 2.7.5.final.0
python-bits: 64
OS: Linux
OS-release: 3.13.0-39-generic
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8

pandas: 0.15.1
nose: 1.3.4
Cython: None
numpy: 1.9.1
scipy: None
statsmodels: None
IPython: 2.3.1
sphinx: 1.2.2
patsy: None
dateutil: 2.2
pytz: 2013.9
bottleneck: None
tables: None
numexpr: None
matplotlib: None
openpyxl: None
xlrd: None
xlwt: None
xlsxwriter: None
lxml: None
bs4: None
html5lib: None
httplib2: 0.9
apiclient: 1.3.1
rpy2: None
sqlalchemy: None
pymysql: None
psycopg2: 2.5.4 (dt dec pq3 ext)
```

mroeschke:
- [x] closes #8917
- [x] tests added / passed
- [x] whatsnew entry

Currently, users do not have any control over nonexistent datetime handling when `tz_localize`ing like they do ambiguous times. This adds a new keyword `nonexistent` to `tz_localize` so that users now can:

``'raise'``: Raise an error (default)
``'NaT'``: Replace nonexistent times with ``'NaT'``
``'shift'``: Shift nonexistent times forward to the closest existing time


pandas/_libs/tslibs/nattype.pyx
```diff
@@ -535,6 +535,13 @@ class NaTType(_NaT):
             - 'NaT' will return NaT for an ambiguous time
             - 'raise' will raise an AmbiguousTimeError for an ambiguous time
 
+        nonexistent : 'shift', 'NaT', default 'raise'
+            - 'shift' will shift the nonexistent time forward to the closest
+              existing time
+            - 'NaT' will return NaT where there are nonexistent times
```
jreback: add a ersionadded tag

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -884,6 +885,13 @@ class Timestamp(_Timestamp):
             - 'NaT' will return NaT for an ambiguous time
             - 'raise' will raise an AmbiguousTimeError for an ambiguous time
 
+        nonexistent : 'shift', 'NaT', default 'raise'
```
jreback: hmm, this is now a bit confusing, e.g. there is overlap between errors and nonexistent. Is it possible to use errors instead here? (and not add a new keyword)?

jreback (changes requested):

mroeschke:
@jreback re: overlap between `errors` and `nonexistent`

1) The original issue mentions https://github.com/pandas-dev/pandas/issues/8917#issuecomment-65082306 having the ability to control over ambiguous times vs nonexistent times independently.
1a) `nonexistent` and `ambiguous` can handle their own errors independently
2) This PR has the ability to `shift` the nonexistent time to a real time (like how `ambiguous` can take `True/False`

So I would propose that eventually we can depreciate `errors` and keep both `ambiguous` and `nonexistent`

jreback:
> So I would propose that eventually we can depreciate errors and keep both ambiguous and nonexistent

can you do that in this PR?

mroeschke:
@jreback went ahead and depreciated the `errors` parameter in `tz_localize`

doc/source/timeseries.rst
```diff
@@ -2350,6 +2350,31 @@ constructor as well as ``tz_localize``.
    # tz_convert(None) is identical with tz_convert('UTC').tz_localize(None)
    didx.tz_convert('UCT').tz_localize(None)
 
+.. _timeseries.timezone_nonexsistent:
+
+Nonexistent Times when Localizing
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
jreback: this is too short, needs to be same length as title

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -191,6 +191,8 @@ Other Enhancements
 - :meth:`Series.resample` and :meth:`DataFrame.resample` have gained the :meth:`Resampler.quantile` (:issue:`15023`).
 - :meth:`Index.to_frame` now supports overriding column name(s) (:issue:`22580`).
 - New attribute :attr:`__git_version__` will return git commit sha of current build (:issue:`21295`).
+- :meth:`Timestamp.tz_localize`, :meth:`DatetimeIndex.tz_localize`, and :meth:`Series.tz_localize` have gained the ``nonexistent`` argument for alternative handling of nonexistent times (:issue:`8917`)
```
jreback: can you add a pointer to the docs

pandas/_libs/tslibs/conversion.pyx
```diff
@@ -837,6 +839,8 @@ def tz_localize_to_utc(ndarray[int64_t] vals, object tz, object ambiguous=None,
     tz : tzinfo or None
     ambiguous : str, bool, or arraylike
         If arraylike, must have the same length as vals
+    nonexistent : str, bool, or arraylike
```
jreback: versionadded

pandas/_libs/tslibs/conversion.pyx
```diff
@@ -849,16 +853,21 @@ def tz_localize_to_utc(ndarray[int64_t] vals, object tz, object ambiguous=None,
...
         npy_datetimestruct dts
         bint infer_dst = False, is_dst = False, fill = False
+        bint shift = False, fill_nonexist = False
         bint is_coerce = errors == 'coerce', is_raise = errors == 'raise'
 
     # Vectorized version of DstTzInfo.localize
-
     assert is_coerce or is_raise
 
+    if is_coerce:
```
jreback: can we do the warning at a higher level? eg. not in the cython code? IOW refactor to avoid passing errors at all

pandas/_libs/tslibs/conversion.pyx
```diff
@@ -849,16 +853,21 @@ def tz_localize_to_utc(ndarray[int64_t] vals, object tz, object ambiguous=None,
...
+        warnings.warn("the errors argument is deprecated, will be removed "
+                      "in a future release. Use the ambiguous or nonexistent "
+                      "argument instead.", DeprecationWarning)
```
jreback: use FutureWarning

pandas/tests/scalar/timestamp/test_timezones.py
```diff
@@ -86,7 +86,8 @@ def test_tz_localize_nonexistent(self, stamp, tz):
             ts.tz_localize(tz)
         with pytest.raises(NonExistentTimeError):
             ts.tz_localize(tz, errors='raise')
-        assert ts.tz_localize(tz, errors='coerce') is NaT
+        with tm.assert_produces_warning(DeprecationWarning):
```
jreback: change these to FutureWarning

jreback (changes requested):

jreback (changes requested):
can you assert that all of the 3 deprecated ones are showing a FutureWarning (Timestamp, DTI, .dt)

jreback (changes requested):

mroeschke:
Hmm good point @jorisvandenbossche. I think this was a complete oversight on my part. Thanks.

I had thought `errors` affected _any_ localization errors this entire time, but you're right, it appears to specifically target nonexistent times.

I guess the consolation this PR brings is:

1) IMO `nonexistent` is a better argument name than `errors` (demonstrated by my oversight)
2) Addition of the `shift` method (as you mentioned).
3) Brings nonexistent time handling to `Series.tz_localize` (akin to adding errors to that method)

I can only really hang my hat on 1, but I understand if its not worth the API churn. Thoughts?

jreback:
i don’t think api churn is really much of an issue here

this is a pretty niche use of the kwarg ; when u need it it’s very important but most of the time u don’t care

rather have the correct handling with rationale named things

jorisvandenbossche:
I agree `nonexistent` is a better name than `errors`, but I am not fully sure it is that much better to warrant a deprecation. This is indeed a rather niche use case, but if the parameter description clearly indicates "How to handle non-existent time errors: ...", I think that also might be sufficient.

The addition of shift and addition of the functionality to `Series.tz_localize` are good enhancements for sure. Since for `Series.tz_localize` it is a new keyword, there is no backward compatibility problem in that case of course ..

It just feels a bit stupid that we are deprecating a keyword that we only relatively recently added for the exact usecase in this PR (apart from adding the shift option of course)

> rather have the correct handling with rationale named things

The concern I raise is not about not having "correct handling" or so, apart from the additional features added here in this PR, it is a pure keyword-rename for the rest.

jreback:
> It just feels a bit stupid that we are deprecating a keyword that we only relatively recently added for the exact usecase in this PR (apart from adding the shift option of course)

sure, but would have rather have the new, descriptive keyword in a year than not. I don't think the prior deprecation period is relevant in the long term, though sure can produce churn. 

I don't see a reason not to proceed in this relatively niche case here. If it were a major impacting change (e.g. ``skiprows`` -> ``skip_rows``) or some such, then sure maybe it is more impactful, though that shoulnd't stop a good deprecation.

pandas/_libs/tslibs/conversion.pyx
```diff
@@ -837,7 +836,14 @@ def tz_localize_to_utc(ndarray[int64_t] vals, object tz, object ambiguous=None,
     tz : tzinfo or None
     ambiguous : str, bool, or arraylike
         If arraylike, must have the same length as vals
-    errors : {"raise", "coerce"}, default "raise"
+    nonexistent : str, bool, or arraylike
+        If arraylike, must have the same length as vals
+
+        .. versionadded:: 0.24.0
+
+    errors : {"raise", "coerce"}, default None
```
jreback: why do you actually need this arg to be passed?

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -913,14 +914,23 @@ class Timestamp(_Timestamp):
...
+            - 'shift' will shift the nonexistent time forward to the closest
+              existing time
+            - 'NaT' will return NaT where there are nonexistent times
+            - 'raise' will raise an NonExistentTimeError if there are
+              nonexistent times
+
+            .. versionadded:: 0.24.0
+
+        errors : 'raise', 'coerce', default None
```
jreback: this is a top-level yes? (e.g. user callable)

pandas/_libs/tslibs/timestamps.pyx
```diff
@@ -934,13 +944,27 @@ class Timestamp(_Timestamp):
...
+            warnings.warn("The errors argument is deprecated and will be "
+                          "removed in a future release. Use the ambiguous or "
+                          "nonexistent argument instead.", FutureWarning,
+                          stacklevel=2)
+            if errors == 'coerce':
+                nonexistent = 'NaT'
+            elif errors == 'raise':
+                nonexistent = 'raise'
+            else:
+                raise ValueError("The errors argument must be either coerce "
```
jreback: is this tested?

jreback (changes requested):

mroeschke:
Addressed your comments @jreback

pandas/_libs/tslibs/conversion.pyx
```diff
@@ -888,39 +887,43 @@ def tz_localize_to_utc(ndarray[int64_t] vals, object tz, object ambiguous=None,
                              "the same size as vals")
         ambiguous_array = np.asarray(ambiguous)
 
+    if is_string_object(nonexistent):
+        if nonexistent == 'NaT':
+            fill_nonexist = True
+        elif nonexistent == 'shift':
+            shift = True
+
```
jreback: in theory should assert the possible values of nonexistent (e.g. what If I pass 'foo'),

jreback (approved):
lgtm. minor comment, pls merge on green after that push.

jorisvandenbossche (commented):
Added some doc related comments

mroeschke:
Thanks @jorisvandenbossche. Addressed your comments.

jreback:
lgtm. @jorisvandenbossche merged when satisified.

jorisvandenbossche (commented):
Some minor comments on the new changes

pandas/tests/indexes/datetimes/test_timezones.py
```diff
@@ -312,9 +312,13 @@ def test_dti_tz_localize_nonexistent_raise_coerce(self):
...
 
         with pytest.raises(pytz.NonExistentTimeError):
-            index.tz_localize(tz=tz, errors='raise')
+            with tm.assert_produces_warning(FutureWarning,
+                                            check_stacklevel=False):
+                index.tz_localize(tz=tz, errors='raise')
 
-        result = index.tz_localize(tz=tz, errors='coerce')
+        with tm.assert_produces_warning(FutureWarning, check_stacklevel=False,
+                                        clear=FutureWarning):
```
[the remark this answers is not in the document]
mroeschke: I test for another `FutureWarning` above this one, so as I understand it, I need `clear=FutureWarning` to test for another `FutureWarning` in the same test.

jreback:
lgtm @jorisvandenbossche any more comments?

pandas/_libs/tslibs/nattype.pyx
```diff
@@ -565,6 +565,8 @@ class NaTType(_NaT):
             - 'raise' will raise an AmbiguousTimeError for an ambiguous time
 
         nonexistent : 'shift', 'NaT', default 'raise'
+            A nonexistent time doesn't not exist in a particular timezone
+            where clocks moved forward due to DST.
             - 'shift' will shift the nonexistent time forward to the closest
```
jorisvandenbossche: Sorry, rst formatting nitpick: there needs to be a blank line between the first sentences, and the start of this list ... (getting rst right can be annoying ..)

pandas/tests/indexes/datetimes/test_timezones.py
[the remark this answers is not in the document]
jorisvandenbossche: I would expect the previous `assert_produces_warning` to take care of this to not leak this outside the context manager (but my experience is that with testing warnings expectations are often wrong .. :-)) 

Does it actually raise an error if you remove it?

mroeschke:
Thanks @jorisvandenbossche. Added those blank lines for rendering.

pandas/tests/indexes/datetimes/test_timezones.py
```diff
@@ -574,6 +578,38 @@ def test_dti_tz_localize_bdate_range(self):
         localized = dr.tz_localize(pytz.utc)
         tm.assert_index_equal(dr_utc, localized)
 
+    @pytest.mark.parametrize('tz', ['Europe/Warsaw', 'dateutil/Europe/Warsaw'])
+    @pytest.mark.parametrize('method, exp', [
+        ['shift', '2015-03-29 03:00:00'],
+        ['NaT', pd.NaT],
```
jreback: do you have tests that exericse the assertion when you pass a ``nonexistent`` keyword that is invalid?

mroeschke:
Thanks for catching that typo @jorisvandenbossche

jreback:
one more rebase and I think ok to go

jreback:
thanks @mroeschke

PR #22646, opened 2018-09-09
TST: Avoid DeprecationWarnings

jbrockmendel:
Avoid a boatload of these:

```
pandas/tests/plotting/test_frame.py::TestDataFramePlots::()::test_plot
  /home/travis/build/pandas-dev/pandas/pandas/core/indexes/multi.py:1556: DeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
    if lab[key] == -1:

  /home/travis/build/pandas-dev/pandas/pandas/io/formats/style.py:1039: DeprecationWarning: In future, it will be an error for 'np.bool_' scalars to be interpreted as an index
    return css_bar(min(x, zero), max(x, zero), colors[x > zero])

pandas/tests/series/test_analytics.py::TestCategoricalSeriesAnalytics::()::test_drop_duplicates_categorical_non_bool[False-datetime64[D]]
  source:2442: DeprecationWarning: invalid escape sequence \(
```

That last one was a nice little adventure tracking down to pd.util.testing.

Does not silence these, as I'm not confident that is The Right Thing To Do:
```
pandas/tests/tslibs/test_parsing.py::TestGuessDatetimeFormat::()::test_guess_datetime_format_nopadding[2011-1-1 0:0:0-%Y-%m-%d %H:%M:%S]
  /home/travis/miniconda3/envs/pandas/lib/python3.6/site-packages/dateutil/parser/__init__.py:46: DeprecationWarning: _timelex is a private class and may break without warning, it will be moved and or renamed in future versions.
    warnings.warn(msg, DeprecationWarning)
```

Does not silence these, as I'm not sure how:
```
pandas/tests/io/test_sql.py::TestSQLiteAlchemyConn::()::test_read_table
  /home/travis/miniconda3/envs/pandas/lib/python3.6/site-packages/_pytest/fixtures.py:795: RemovedInPytest4Warning: Fixture setup_method called directly. Fixtures are not meant to be called directly, are created automatically when test functions request them as parameters. See https://docs.pytest.org/en/latest/fixture.html for more information.
    res = next(it)
```

### DIFF
--- a/pandas/core/indexes/multi.py
+++ b/pandas/core/indexes/multi.py
@@ -1551,6 +1551,10 @@ class MultiIndex(Index):
 
     def __getitem__(self, key):
         if is_scalar(key):
+            if lib.is_float(key) and key == int(key):
+                # Avoid DeprecationWarning for non-integer number indexer
+                key = int(key)
+
             retval = []
             for lev, lab in zip(self.levels, self.labels):
                 if lab[key] == -1:
--- a/pandas/core/indexes/multi.py
+++ b/pandas/core/indexes/multi.py
@@ -1551,9 +1551,7 @@ class MultiIndex(Index):
 
     def __getitem__(self, key):
         if is_scalar(key):
-            if lib.is_float(key) and key == int(key):
-                # Avoid DeprecationWarning for non-integer number indexer
-                key = int(key)
+            key = com.cast_scalar_indexer(key)
 
             retval = []
             for lev, lab in zip(self.levels, self.labels):
--- a/pandas/core/common.py
+++ b/pandas/core/common.py
@@ -122,6 +122,24 @@ def is_bool_indexer(key):
     return False
 
 
+def cast_scalar_indexer(val):
+    """
+    To avoid numpy DeprecationWarnings, cast float to integer where valid.
+
+    Parameters
+    ----------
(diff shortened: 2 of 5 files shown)


pandas/core/indexes/multi.py
```diff
@@ -1551,6 +1551,10 @@ def __setstate__(self, state):
 
     def __getitem__(self, key):
         if is_scalar(key):
+            if lib.is_float(key) and key == int(key):
+                # Avoid DeprecationWarning for non-integer number indexer
```
jreback: we do this a number of times in the indexing code

i think we have a function

maybe_convert_to_int_indexer iirc

if not can u make one

jreback (changes requested):

pandas/core/indexes/multi.py
jbrockmendel: Index._convert_scalar_indexer looks like a candidate; need to check where it is called.

jbrockmendel: I take it back.  core.common is right next to `is_bool_indexer` is the place for this.

jreback (approved):

jreback:
thanks! some more if you have a chance :>

PR #22515, opened 2018-08-26
Fix test_sql pytest fixture warnings

alimcmaster1:
- [y] closes #22338
- [y] tests added / passed
- [y] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

See Travis CI Python 3.6 build [before](https://travis-ci.org/pandas-dev/pandas/jobs/419432794) and [after](https://travis-ci.org/pandas-dev/pandas/jobs/420824121) my changes get rid of the warning for calling fixtures directly.

You can see this by searching for the string "pandas/tests/io/test_sql.py" to see that well over 100 warnings have been removed.

PR #22651, opened 2018-09-09
BUG: Make sure that sas7bdat parsers memory is initialized to 0 (#21616)

### LINKED ISSUE #21616: read_sas does not handle numeric variables stored with fewer than 8 bytes in SAS datasets
This problem has been reported twice on stackoverflow, but I cannot find any issue raised here to address it.

https://stackoverflow.com/questions/49059421/pandas-fails-with-correct-data-type-while-reading-a-sas-file
https://stackoverflow.com/questions/51005244/how-can-i-preserve-the-data-type-of-a-column-when-using-pandas-read-sas/51005427#51005427

read_sas() is not handling SAS numeric values that are stored using fewer than the full 8 bytes required for floating point numbers.  It appears that instead of padding the short values with binary zeros that read_sas is using some other, perhaps randomly chosen, bytes to fill out the values.

In the first stackover flow example you can see that SAS has stored the value of the number 8.  This would be represented in IEEE floating point as the 8 bytes represented by the hex string '40 20 00 00 00 00 00 00'.  In the sas7bdat file SAS stored only the 3 most significant bytes.  Instead of padding it with 5 bytes of zeros before converting it it looks like read_sas padded it with bytes represented by the hex codes '06 07 80 FD C1' instead.  So that instead of reading the value as 8 it was read as 8.00046 .

troels:
Memory for numbers in sas7bdat-parsing was not initialized properly to 0.
For sas7bdat files with numbers smaller than 8 bytes this made the
least significant part of the numbers essentially random.
Fix it by initializing memory correctly.

- [X] closes #21616
- [X] tests added / passed
- [X] whatsnew entry


pandas/tests/io/sas/test_sas7bdat.py
```diff
@@ -183,6 +183,18 @@ def test_date_time(datapath):
...
 
+def test_compact_numerical_values(datapath):
+    # Regression test for #21616
+    fname = datapath("io", "sas", "data", "cars.sas7bdat")
+    df = pd.read_sas(fname, encoding='latin-1')
+    fname = datapath("io", "sas", "data", "cars.csv")
+    df0 = pd.read_csv(fname)
+    # The two columns CYL and WGT have column
+    # width < 8 in cars.sas7bdat
+    assert (df['WGT'] == df0['WGT']).all()
```
WillAyd: Should use `tm.assert_frame_equal` here instead

pandas/tests/io/sas/data/cars.csv
```diff
@@ -0,0 +1,393 @@
+MPG,CYL,ENG,WGT
```
WillAyd: Can you make the test data smaller? Shouldn't need close to 400 lines of data IIUC

pandas/tests/io/sas/test_sas7bdat.py
```diff
@@ -183,6 +183,18 @@ def test_date_time(datapath):
     tm.assert_frame_equal(df, df0)
 
 
+def test_compact_numerical_values(datapath):
+    # Regression test for #21616
+    fname = datapath("io", "sas", "data", "cars.sas7bdat")
```
WillAyd: Using the fixture in this fashion will generate warnings. We just went through an exercise to clean these up in #22515 - can you take a look at that and adjust here accordingly?

WillAyd (changes requested):

pandas/tests/io/sas/data/cars.csv
troels: Well, the sas7bdat file is from the test-case from issue, and I can't produce a new one. Also these kind of tests, testing for random bytes in uninitialized memory are somewhat more sensitive with larger test-data. (We are after all dependent on someone writing something nonzero to the memory before we get it, so more is better). 

I suppose I might be able to check that all numbers are integral or close to instead of comparing them to a csv. Would that be preferable?

pandas/tests/io/sas/test_sas7bdat.py
troels: If I understood the discussion on the pytest-tracker properly, datapath used this way is in fact not something that will be deprecated.

datapath is a "factory as a fixture" as described here:
https://docs.pytest.org/en/latest/fixture.html#factories-as-fixtures

Which is something distinct from what is beeing talked about here:
https://github.com/pytest-dev/pytest/issues/3661

Have I misunderstood?

pandas/tests/io/sas/data/cars.csv
troels: Ok, I decided to check that the floats are integral instead of having a csv fixture, so I resolve this.

pandas/tests/io/sas/test_sas7bdat.py
troels: Changed to tm.assert_series_equal now.

pandas/tests/io/sas/test_sas7bdat.py
troels: And it produces no warnings, as far as I can tell.

pandas/tests/io/sas/test_sas7bdat.py
```diff
@@ -183,6 +183,17 @@ def test_date_time(datapath):
...
+    df = pd.read_sas(fname, encoding='latin-1')
+    # The two columns CYL and WGT in cars.sas7bdat have column
+    # width < 8 and only contains integral values. Test
+    # that pandas doesn't corrupt the less significant bits.
+    tm.assert_series_equal(df['WGT'], df['WGT'].round(), check_exact=True)
```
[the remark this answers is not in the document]
troels: We aren't comparing to self. We are making sure, that floats with no decimal part in the sas7bdat-file doesn't get a decimal part because of a bug in pandas. That is exactly what this bug is about, after all. 

(See: https://stackoverflow.com/questions/49059421/pandas-fails-with-correct-data-type-while-reading-a-sas-file)

Before:
1. Memory was not initialized.
2. The decimal part of the float was taken from random uninitialized memory.
3. Test fails most of the time.

Now:
1. Memory is zeroed out.
2. The decimal part of the float which is not read (because it's implicit in the file format, that it should be zero) is therefore still zero in pandas.
3. Test succeeds.

WillAyd (approved):
@TomAugspurger any thoughts on this one?

pandas/tests/io/sas/test_sas7bdat.py
TomAugspurger: This usage is fine I think.

But, do you need to include a binary file in the git repo to reproduce the error on master? I'd like to avoid adding new ones if possible.

TomAugspurger:
cc @kshedden and @Winand if you have time to give this a review.

pandas/tests/io/sas/test_sas7bdat.py
troels: Hi @TomAugspurger 

I am adding a sas7bdat file for testing the parsing of sas7bdat files. How can that be done more sensibly?

TomAugspurger:
Not sure. Do any of the test files already in the repo have this problem.

We can add it if necessary, just trying to avoid bloating the git
repository more than it already is.

troels:
@TomAugspurger: 

No, none of the test-files has the problem. If any of the already existing test-files had numeric columns less than 8 bytes, the tests would probably have been unstable.

Usually one wants to avoid binary files in git repositories for two reasons:
1. They tend to be large
2. Git's diff has a tendency to every time a binary files has changed store the complete file as the delta. (as git's diff is clumsy with binary content) 
That makes e.g. large images that are edited and committed often rather toxic. 

A small file like cars.sas7bdat, which will never be edited and to top it off is highly compressible is not going to take very much space at all, neither now nor in the future.

pandas/tests/io/sas/test_sas7bdat.py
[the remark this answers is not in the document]
jreback: can you use
```
result = 
expected = 
```

to make this easier to read

why do we need to .round(), rather than just using ``check_less_precision`` ?

jreback (changes requested):

pandas/tests/io/sas/test_sas7bdat.py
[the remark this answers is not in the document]
troels: I can also do:
```
msg = "Expected df['WGT'] to be full of integers"

assert df['WGT'] == df['WGT'].astype('int'), msg
tm.assert_series_equal(df['WGT'], df['WGT'].astype('int'), check_exact=True)
assert all(f.is_integer() for f in df['WGT']), msg
assert any(f - int(f) != 0 for f in df['WGT']), msg
assert all(f == int(f) for f in df['WGT']), msg
```
or any other way of saying the same thing you would prefer...

jreback (approved):

jreback:
thanks @troels

PR #22652, opened 2018-09-09
ASV: more for str.cat

h-vetinari:
Planning to do a clean-up of the `Series.str.cat` internals and wanted to expand ASV-coverage before doing so.

### DIFF
--- a/asv_bench/benchmarks/strings.py
+++ b/asv_bench/benchmarks/strings.py
@@ -87,18 +87,19 @@ class Repeat(object):
 class Cat(object):
 
     goal_time = 0.2
-    params = ([None, 5], [None, ','], [None, '-'])
-    param_names = ['others', 'sep', 'na_rep']
+    params = ([None, 5], [None, ','], [None, '-'], [0.0, 1e-4, 0.1])
+    param_names = ['others', 'sep', 'na_rep', 'na_frac']
 
-    def setup(self, others, sep, na_rep):
+    def setup(self, others, sep, na_rep, na_frac):
         N = int(5e5)
-        mask_gen = lambda: np.random.choice([True, False], N, p=[0.9, 0.1])
+        mask_gen = lambda: np.random.choice([True, False], N,
+                                            p=[1 - na_frac, na_frac])
         self.s = Series(tm.makeStringIndex(N)).where(mask_gen())
         self.others = (DataFrame({i: tm.makeStringIndex(N).where(mask_gen())
                                   for i in range(others)})
                        if others is not None else None)
 
-    def time_cat(self, others, sep, na_rep):
+    def time_cat(self, others, sep, na_rep, na_frac):
         self.s.str.cat(self.others, sep=sep, na_rep=na_rep)
 
 
--- a/asv_bench/benchmarks/strings.py
+++ b/asv_bench/benchmarks/strings.py
@@ -87,24 +87,24 @@ class Repeat(object):
 class Cat(object):
 
     goal_time = 0.2
-    params = ([None, 5], [None, ','], [None, '-'], [0.0, 1e-4, 0.1])
-    param_names = ['others', 'sep', 'na_rep', 'na_frac']
+    params = ([0, 3], [None, ','], [None, '-'], [0.0, 0.001, 0.15])
+    param_names = ['other_cols', 'sep', 'na_rep', 'na_frac']
 
-    def setup(self, others, sep, na_rep, na_frac):
-        N = int(5e5)
+    def setup(self, other_cols, sep, na_rep, na_frac):
+        N = 10 ** 5
         mask_gen = lambda: np.random.choice([True, False], N,
                                             p=[1 - na_frac, na_frac])
         self.s = Series(tm.makeStringIndex(N)).where(mask_gen())
         self.others = (DataFrame({i: tm.makeStringIndex(N).where(mask_gen())
-                                  for i in range(others)})
-                       if others is not None else None)
+                                  for i in range(other_cols)})
+                       if other_cols > 0 else None)
 
-    def time_cat(self, others, sep, na_rep, na_frac):
-        # before the concatenation (one caller + others columns), the total
+    def time_cat(self, other_cols, sep, na_rep, na_frac):
+        # before the concatenation (one caller + other_cols columns), the total
         # expected fraction of rows containing any NaN is:
-        #     reduce(lambda t, _: t + (1 - t) * na_frac, range(others + 1), 0)
-        # for others=5 and na_frac=0.1, this works out to ~47%
-        self.s.str.cat(self.others, sep=sep, na_rep=na_rep)
+        # reduce(lambda t, _: t + (1 - t) * na_frac, range(other_cols + 1), 0)
+        # for other_cols=3 and na_frac=0.15, this works out to ~48%
+        self.s.str.cat(others=self.others, sep=sep, na_rep=na_rep)
 
 
 class Contains(object):
--- a/asv_bench/benchmarks/strings.py
+++ b/asv_bench/benchmarks/strings.py
@@ -1,7 +1,7 @@
 import warnings
 
 import numpy as np
-from pandas import Series
+from pandas import Series, DataFrame
 import pandas.util.testing as tm
 
 
@@ -11,9 +11,6 @@ class Methods(object):
 
     def setup(self):
         self.s = Series(tm.makeStringIndex(10**5))
-
-    def time_cat(self):
-        self.s.str.cat(sep=',')
 
     def time_center(self):
         self.s.str.center(100)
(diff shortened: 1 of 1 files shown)


h-vetinari:
@TomAugspurger 

ping because you managed the CirlceCI transition from 1.0 to 2.0. It seems that new setup does not respect `[ci skip]` - I view the CI as a limited resource (appveyor recently fluctuating up to something like 12h waiting time), so `[ci skip]` can be useful for avoiding a CI run where it's pointless (like this PR, which changes nothing in code/docs; and PEP8 runs anyway).

TomAugspurger:
Last I saw CircleCI had an issue open for ci-skip. Not sure what the status
is.

h-vetinari:
@TomAugspurger 

not sure where you saw it, but https://discuss.circleci.com/search?q=ci%20skip is pretty messy with lots of duplicates (due to their auto-closing policy for issues after a while with no activity). One of the closer ones seems to be https://discuss.circleci.com/t/ci-skip-does-not-work-in-pull-requests/5308

The 2.0 docs (https://circleci.com/docs/2.0/skip-build/) mention that
> By default, CircleCI automatically builds a project whenever you push changes to a version control system (VCS). You can override this behavior by adding a [ci skip] or [skip ci] tag anywhere in a commit’s title or description.
[...]
**Note**: This feature is not supported for fork PRs.

which is weird, because it *did* work for 1.0 (side note: the docs for that don't exist anymore: https://circleci.com/docs/1.0/skip-a-build/).

On an unrelated note: would you care to comment about the PR itself? :)

TomAugspurger:
https://discuss.circleci.com/t/ci-skip-does-not-work-in-pull-requests/5308/10 is the one I saw.

Agreed it's unfortunate. If a maintainer is around you can ask them to cancel it :/

No strong thoughts on how the PR, other than a consideration of runtime. It seems like these should run pretty quickly.

h-vetinari:
@TomAugspurger 
> Agreed it's unfortunate. If a maintainer is around you can ask them to cancel it :/

Maintainer time is an even more precious resource than CI time, so I think I won't bother you guys. ;-)

> No strong thoughts on how the PR, other than a consideration of runtime. It seems like these should run pretty quickly.

Here's the output of an ASV run against master. Of course, `N` (and the number of columns for `others`) could be tuned as desired, but I don't think it's so bad (and due to decrease a fair amount - hopefully - by an upcoming PR).

```
     [0976e126]       [2e70311e]
            594ms            609ms     1.03  strings.Cat.time_cat(5, ',', '-', 0.0)
            594ms            609ms     1.03  strings.Cat.time_cat(5, ',', '-', 0.0001)
            594ms            594ms     1.00  strings.Cat.time_cat(5, ',', '-', 0.1)
            375ms            375ms     1.00  strings.Cat.time_cat(5, ',', None, 0.0)
            359ms            344ms     0.96  strings.Cat.time_cat(5, ',', None, 0.0001)
            297ms            312ms     1.05  strings.Cat.time_cat(5, ',', None, 0.1)
           failed            453ms      n/a  strings.Cat.time_cat(5, None, '-', 0.0)
            438ms            438ms     1.00  strings.Cat.time_cat(5, None, '-', 0.0001)
            453ms            453ms     1.00  strings.Cat.time_cat(5, None, '-', 0.1)
            344ms            359ms     1.05  strings.Cat.time_cat(5, None, None, 0.0)
            344ms            344ms     1.00  strings.Cat.time_cat(5, None, None, 0.0001)
           failed            297ms      n/a  strings.Cat.time_cat(5, None, None, 0.1)
         46.9±0ms       46.9±0.7ms     1.00  strings.Cat.time_cat(None, ',', '-', 0.0)
         54.7±4ms       46.9±0.7ms    ~0.86  strings.Cat.time_cat(None, ',', '-', 0.0001)
         54.7±3ms         62.5±4ms    ~1.14  strings.Cat.time_cat(None, ',', '-', 0.1)
         44.3±3ms         44.3±3ms     1.00  strings.Cat.time_cat(None, ',', None, 0.0)
         62.5±4ms         62.5±4ms     1.00  strings.Cat.time_cat(None, ',', None, 0.0001)
         70.3±1ms         78.1±1ms    ~1.11  strings.Cat.time_cat(None, ',', None, 0.1)
           failed       41.7±0.7ms      n/a  strings.Cat.time_cat(None, None, '-', 0.0)
       41.7±0.7ms         46.9±4ms    ~1.12  strings.Cat.time_cat(None, None, '-', 0.0001)
         54.7±0ms         54.7±0ms     1.00  strings.Cat.time_cat(None, None, '-', 0.1)
           failed         62.5±1ms      n/a  strings.Cat.time_cat(None, None, None, 0.0)
           failed         58.6±5ms      n/a  strings.Cat.time_cat(None, None, None, 0.0001)
           failed         54.7±0ms      n/a  strings.Cat.time_cat(None, None, None, 0.1)
```

asv_bench/benchmarks/strings.py
```diff
@@ -87,6 +84,29 @@ def time_repeat(self, repeats):
...
 
 
+class Cat(object):
+
+    goal_time = 0.2
+    params = ([None, 5], [None, ','], [None, '-'], [0.0, 1e-4, 0.1])
+    param_names = ['others', 'sep', 'na_rep', 'na_frac']
+
+    def setup(self, others, sep, na_rep, na_frac):
+        N = int(5e5)
```
WillAyd: Can you see how we do this in other benchmarks? I think we usually just pick 100_000 so for consistency might be nice to reflect that here

asv_bench/benchmarks/strings.py
```diff
@@ -87,6 +84,29 @@ def time_repeat(self, repeats):
...
+        mask_gen = lambda: np.random.choice([True, False], N,
```
WillAyd: Do we need the lambda here? Figured we could just use the returned array in the subsequent call

asv_bench/benchmarks/strings.py
```diff
@@ -87,6 +84,29 @@ def time_repeat(self, repeats):
...
+                                            p=[1 - na_frac, na_frac])
+        self.s = Series(tm.makeStringIndex(N)).where(mask_gen())
+        self.others = (DataFrame({i: tm.makeStringIndex(N).where(mask_gen())
+                                  for i in range(others)})
+                       if others is not None else None)
```
WillAyd: Any reason you don't just use `0` instead of `None` as the parametrized argument? Would get rid of the need for a condition here

WillAyd (changes requested):

asv_bench/benchmarks/strings.py
h-vetinari: No problem, was trying to get the results near `goal_time`, but maybe I'm misinterpreting what that is supposed to be

asv_bench/benchmarks/strings.py
h-vetinari: Yes, it's essential IMO that each column gets a different mask to compound the NaNs (for the `na_rep=None` case) and have a reasonable similarity to real-life data.

asv_bench/benchmarks/strings.py
h-vetinari: Changed to `0` for consistency/legibility, but still need the condition, as `pd.DataFrame({'i' : ... for i in range(0)})` would yield a DataFrame (but `str.cat` needs `others=None` to self-concatenate).

h-vetinari (commented):
Thanks for the review.

asv_bench/benchmarks/strings.py
```diff
@@ -87,6 +84,29 @@ def time_repeat(self, repeats):
...
+    param_names = ['other_cols', 'sep', 'na_rep', 'na_frac']
+
+    def setup(self, other_cols, sep, na_rep, na_frac):
+        N = 10 ** 5
+        mask_gen = lambda: np.random.choice([True, False], N,
+                                            p=[1 - na_frac, na_frac])
+        self.s = Series(tm.makeStringIndex(N)).where(mask_gen())
+        self.others = (DataFrame({i: tm.makeStringIndex(N).where(mask_gen())
+                                  for i in range(other_cols)})
+                       if other_cols > 0 else None)
```
WillAyd: Better but I think the condition inside the comprehension just makes this tougher to read than it needs to be. Would rather take the condition out of the comprehension and as a subsequent statement evaluate the truthiness of the expression, assigning None where False

WillAyd (changes requested):
Request on readability otherwise lgtm

h-vetinari (commented):
Thx, changed

WillAyd (approved):
lgtm @TomAugspurger

h-vetinari:
@WillAyd 
FWIW, @TomAugspurger said above:
> No strong thoughts on how the PR, other than a consideration of runtime. It seems like these should run pretty quickly.

TomAugspurger:
Thanks!

PR #22654, opened 2018-09-10
ENH: Support writing timestamps with timezones with to_sql

### LINKED ISSUE #9086: ENH/SQL: support writing timestamps with timezone in to_sql
This currently works a bit for datetime values in the index (not in the columns, reason for this is that is then are object columns: #9085). 
I say 'a bit', as it does give a 'timestamp with time zone' type (example for postgresql), but it is the time localized in your local timezone (I suppose because this is the way numpy works on input/output of datetimes).

In any case, also the code here (https://github.com/pydata/pandas/blob/v0.15.2/pandas/io/sql.py#L916) is not working (as columns never have a `tzinfo` attribute), and there are also no tests for this.

So in practice, timezones are not yet supported.

### LINKED ISSUE #23510: to_sql method turns datetime64 index to time zone aware in postgres
I have a `pandas` `Dataframe` with an index which is of type `datetime64[ns]`.
when I use the `to_sql` method i expect the index to be created as a `timestamp` postgres column, however it is creates a `timestamptz` column.

Note that when the `datetime64[ns]` column is not an index but rather a normal column, this doesn't happen.

pandas 0.23.4, psycopg2 2.7.4, sqlalchemy 1.2.7, PostgreSQL 9.6.6

---
*example*

```python
dates = pd.date_range('2018-01-01', periods=5, freq='6h')
df_test = pd.DataFrame({'nums': range(5)}, index=dates)
```
<details>

|                     | nums |
|---------------------|------|
| 2018-01-01 00:00:00 | 0    |
| 2018-01-01 06:00:00 | 1    |
| 2018-01-01 12:00:00 | 2    |

</details>

inserting to postgres
```python
df_test.to_sql('foo_table',postgres_uri,schema='data_test',index_label='info_date')
```
when reading from the database I get the index with different type
```python
df_db = pd.read_sql_table('foo_table',postgres_uri,schema='data_test',index_col='info_date')
```
<details>

|                           | nums |
|---------------------------|------|
| info_date                 |      |
| 2018-01-01 00:00:00+00:00 | 0    |
| 2018-01-01 06:00:00+00:00 | 1    |
| 2018-01-01 12:00:00+00:00 | 2    |

</details>
As you can see `info_date` is of type `datetime64[ns, UTC]`

not sure if it's a sqlalchemy or pandas question.
reported because of my [question](https://stackoverflow.com/questions/52986158/pandas-to-sql-turns-datetime64-index-to-time-zone-aware) on SO

mroeschke:
- [x] closes #9086
- [x] closes #23510
- [x] tests added / passed
- [x] whatsnew entry


jorisvandenbossche (commented):
Nice! this was long overdue :-)

pandas/tests/io/test_sql.py
```diff
@@ -1354,9 +1355,32 @@ def check(col):
...
+
+        result = sql.read_sql_query(
+            'SELECT * FROM test_datetime_tz', self.conn
+        )
+        result = result.drop('index', axis=1)
+        tm.assert_frame_equal(result, expected)
+
     def test_date_parsing(self):
         # No Parsing
-        df = sql.read_sql_table("types_test_data", self.conn)
```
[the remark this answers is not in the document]
mroeschke: Oh I see. My cursory glance thought it was duplicate of the line below; you're right, I will try to add an assert to this result as well

pandas/tests/io/test_sql.py
```diff
@@ -1354,9 +1355,32 @@ def check(col):
         df = sql.read_sql_table("types_test_data", self.conn)
         check(df.DateColWithTz)
 
+    def test_datetime_with_timezone_writing(self):
+        # GH 9086
+        if self.flavor != 'postgresql':
+            msg = "{} does not support datetime with time zone"
+            pytest.skip(msg.format(self.flavor))
```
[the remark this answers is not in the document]
mroeschke: That's a good point. From earlier tests databases that don't support `timestamp with timezone` read back data as naive (don't know if this is local naive or converted to UTC first and then returned naive though). Will add some tests.

mroeschke:
@jorisvandenbossche so looks like based on the passing tests, databases without timezone support store those the data as naive local time.

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -179,6 +179,7 @@ Other Enhancements
 - :class:`IntervalIndex` has gained the :meth:`~IntervalIndex.set_closed` method to change the existing ``closed`` value (:issue:`21670`)
 - :func:`~DataFrame.to_csv`, :func:`~Series.to_csv`, :func:`~DataFrame.to_json`, and :func:`~Series.to_json` now support ``compression='infer'`` to infer compression based on filename extension (:issue:`15008`).
   The default compression for ``to_csv``, ``to_json``, and ``to_pickle`` methods has been updated to ``'infer'`` (:issue:`22004`).
+- :func:`to_sql` now supports writing ``TIMESTAMP WITH TIME ZONE`` columns (:issue:`9086`)
```
TomAugspurger: Should be

    :meth:`DataFrame.to_sql`

mroeschke:
Thanks @TomAugspurger. Addressed your comment.

mroeschke:
If you could have a look as well @jreback

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -179,7 +179,7 @@ Other Enhancements
 - :class:`IntervalIndex` has gained the :meth:`~IntervalIndex.set_closed` method to change the existing ``closed`` value (:issue:`21670`)
 - :func:`~DataFrame.to_csv`, :func:`~Series.to_csv`, :func:`~DataFrame.to_json`, and :func:`~Series.to_json` now support ``compression='infer'`` to infer compression based on filename extension (:issue:`15008`).
   The default compression for ``to_csv``, ``to_json``, and ``to_pickle`` methods has been updated to ``'infer'`` (:issue:`22004`).
-- :func:`to_sql` now supports writing ``TIMESTAMP WITH TIME ZONE`` columns (:issue:`9086`)
+- :meth:`DataFrame.to_sql` now supports writing ``TIMESTAMP WITH TIME ZONE`` columns (:issue:`9086`)
```
jorisvandenbossche: I think it would be good to mention what this means in practice.

So if one had now a tz-aware column, it was converted to naive UTC, and you got back naive UTC. So the difference is now that you will get back tz-aware UTC ? (and of course that the schema of the sql table now includes 'with time zone')

But so for databases that don't support timezones. Previously we converted to UTC so users got also back naive UTC, while now they will get back naive local? 

(trying to fully understand the consequences)

pandas/tests/io/test_sql.py
```diff
@@ -1361,9 +1362,42 @@ def check(col):
...
+
+        if self.flavor == 'postgresql':
+            # SQLalchemy "timezones" (i.e. offsets) are coerced to UTC
+            expected['A'] = expected['A'].dt.tz_convert('UTC')
+        else:
+            # Otherwise, timestamps are returned as local, naive
+            expected['A'] = expected['A'].dt.tz_localize(None)
+
+        result = sql.read_sql_table('test_datetime_tz', self.conn)
+        result = result.drop('index', axis=1)
```
jorisvandenbossche: you can put `index=False` in the `to_sql` call, then you don't need to drop the index here I think

jorisvandenbossche (commented):
Thanks for the updates! Tests look good now. 

Added some more questions.

I am wondering if it would be worth adding a small section to the sql part of io.rst explaining some of the behaviours of working with datetime data / timezones.

jreback:
lgtm. @jorisvandenbossche merge when ready.

mroeschke:
@jorisvandenbossche took your suggestions and added a section in io.rst explaining datetime/timezone io with SQL

doc/source/io.rst
```diff
@@ -4777,7 +4777,7 @@ writes ``data`` to the database in batches of 1000 rows at a time:
     data.to_sql('data_chunked', engine, chunksize=1000)
 
 SQL data types
-++++++++++++++
```
jreback: do the docs look ok after this?

pandas/core/generic.py
```diff
@@ -2312,6 +2312,13 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
         --------
         pandas.read_sql : read a DataFrame from a table
 
+        Notes
+        -----
+        Timezone aware datetime columns will be written as
+        ``Timestamp with timezone`` type with SQLAlchemy if supported by the
+        database. Otherwise, the datetimes will stored in the closest available
+        type as timezone unaware timestamps local to the original timezone.
```
jreback: add a versionadded / changed

doc/source/io.rst
jorisvandenbossche: This is moving it one level up (since this is the only 'subsubsection' in the while sql section, I think that is fine). @mroeschke what was your reasoning here?

doc/source/io.rst
```diff
@@ -4809,6 +4809,30 @@ default ``Text`` type for string columns:
     Because of this, reading the database table back in does **not** generate
     a categorical.
 
+Datetime data types
+'''''''''''''''''''
+
+Using SQLAlchemy, :func:`~pandas.DataFrame.to_sql` is capable of writing
+datetime data that is timezone naive or timezone aware. However, the resulting
+data stored in the database ultimately depends on the corresponding,
+supported data type for datetime data.
```
jorisvandenbossche: With "corresponding", you mean depending on the database system? 
If so, I think something like "this ultimately depends on the supported data type for datetime data of the database system being used" is clearer. 

I would also add that the following table lists some common systems (but that it can still be different for other database dialects)

doc/source/io.rst
```diff
@@ -4809,6 +4809,30 @@ default ``Text`` type for string columns:
...
+MySQL         ``TIMESTAMP`` or ``DATETIME``                  No
+PostgreSQL    ``TIMESTAMP`` or ``TIMESTAMP WITH TIME ZONE``  Yes
+===========   =============================================  ===================
+
+When writing timezone aware data to databases that do not support timezones,
+the data will be written as timezone naive timestamps that are in local time
+with respect to the timezone.
+
+:func:`~pandas.read_sql_table` is also capable of reading datetime data that is
+timezone aware of naive. When reading ``TIMESTAMP WITH TIME ZONE`` types, pandas
```
jorisvandenbossche: of -> or

pandas/core/generic.py
```diff
@@ -2312,6 +2312,13 @@ def to_sql(self, name, con, schema=None, if_exists='fail', index=True,
         --------
         pandas.read_sql : read a DataFrame from a table
 
+        Notes
+        -----
+        Timezone aware datetime columns will be written as
+        ``Timestamp with timezone`` type with SQLAlchemy if supported by the
+        database. Otherwise, the datetimes will stored in the closest available
```
jorisvandenbossche: stored -> be stored

I think "closest available type" is a bit unclear. I would just say that they will be stored as "timezone unaware timestamps local ..."

jorisvandenbossche (commented):
Thanks for the updates! Another round of small comments

jorisvandenbossche:
Since I can't seem to answer to my previous inline comment, repeating it here:

> I think it would be good to mention what this means in practice.
>
>So if one had now a tz-aware column, it was converted to naive UTC, and you got back naive UTC. So the difference is now that you will get back tz-aware UTC ? (and of course that the schema of the sql table now includes 'with time zone')
>
>But so for databases that don't support timezones. Previously we converted to UTC so users got also back naive UTC, while now they will get back naive local?

Can you answer this one explicitly? Is that last paragraph correct? If so, I think we should mention this in the whatsnew (the behavioural change; the fact itself that you get back naive local is already mentioned in the new io.rst section you added)

doc/source/io.rst
mroeschke: Ah. I thought this was rendering incorrectly when I built the docs, but I was looking at the wrong section. I'll revert it.

mroeschke:
Your last paragraph is correct @jorisvandenbossche. I have added that point to the whatsnew section.

mroeschke:
Any additional comments?

TomAugspurger (approved):

jreback:
ping @jorisvandenbossche

mroeschke:
This one all good @jorisvandenbossche?

mroeschke:
Will push this in tomorrow if there are no other comments.

jorisvandenbossche:
@mroeschke really sorry about the slow reply (I am lagging a bit behind in following up PRs)

I still think we need to make it much more clear what the consequence for users is (and not refer to the user guide which only mentions the current behaviour and not the change in behaviour compared to earlier releases)

jorisvandenbossche:
Some more concrete feedback:

- let's make it an actual section to give it more weight than just a bullet point in a list
- I would include an example to illustrate it better: if you have a database with a table with timestamp with timezone column, what did it return in 0.23? and what in 0.24?

Such an example will not work in sqlite, so probably need to run it locally with eg postgresql and paste it verbatim in the whatsnew.

jorisvandenbossche:
@mroeschke unless I am mistaken about the consequence of course .. 
Because I was now trying to write up an example myself, and I am not so sure any more of what I thought would change. 

So above I said the following:

>  >   So if one had now a tz-aware column, it was converted to naive UTC, and you got back naive UTC. So the difference is now that you will get back tz-aware UTC ? (and of course that the schema of the sql table now includes 'with time zone')
> >
> >   But so for databases that don't support timezones. Previously we converted to UTC so users got also back naive UTC, while now they will get back naive local?
>
> Can you answer this one explicitly? Is that last paragraph correct? If so, I think we should mention this in the whatsnew (the behavioural change; the fact itself that you get back naive local is already mentioned in the new io.rst section you added)

on which you answered that this was correct. 

So therefore I assumed there is a significant change in behaviour, which I don't see reflected in the whatsnew note.

But:

- I one has a tz-aware column now, and you try to write it, it actually errors instead of converting to naive UTC as I said above?
- And if one previously read from a database that did support timezones, you already got utc data back, so nothing changed here as well?

mroeschke:
So the only significant consequence is described here in the `Datetimelike API Changes`:

https://github.com/pandas-dev/pandas/blob/de6278863571a10b2f35c1e65cd25801a98622aa/doc/source/whatsnew/v0.24.0.txt#L617

Roundtrips Before:
* tz-aware DataFrame --> database that doesn't support timezones --> tz-naive UTC DataFrame
* tz-aware DataFrame --> database that does support timezones --> tz-naive UTC DataFrame

Roundtrips After:
* tz-aware DataFrame --> database that doesn't support timezones --> tz-naive local DataFrame
* tz-aware DataFrame --> database that does support timezones --> tz-aware UTC DataFrame

So addressing your last comment's bullet points:

* No error is raised. The data will now be written as naive local instead of naive UTC
* Correct. Before, pandas read tz aware data from databases as UTC and that behavior has not changed.

mroeschke:
Yeah given this discussion; it would be wise to highlight this change as its own section in the whatsnew.

mroeschke:
@jorisvandenbossche added a new whatsnew section in API detailing the before and after implications of this change.

jorisvandenbossche:
So yesterday I tried to make a small illustrative example showing the changed round-trip, but did it actually work before? 
Because I get:

```
In [1]: import sqlalchemy

In [3]: engine = sqlalchemy.create_engine('sqlite:///:memory:')

In [4]: df = pd.DataFrame({'a':[1,2,3], 'b': pd.date_range('2016-01-01', periods=3, tz='Europe/Brussels')})

In [5]: df.to_sql('table', engine)
...
TypeError: Cannot cast DatetimeIndex to dtype datetime64[us]

In [6]: pd.__version__
Out[6]: '0.23.4'
```

So that is my source of confusion. I thought it worked before (and you seemed to confirm it), so therefore I assumed that there is a *change* in behaviour. But is that actually the case? 
Do you have a running example of the cases that you list in the table above?

mroeschke:
Hmm interesting. I made a (bad) assumption that the tests covered this case, but I actually don't see a test case where to_sql is tested with timezones. So the behavior you're seeing is probably correct.

I'll try to confirm tonight on other dbs as well, but I guess it looks like I fixed a bug too :)

mroeschke:
I ran the same test on postgres @jorisvandenbossche and I got the same result.

Therefore you're correct in that there really isn't a change in behavior since previously `to_sql` and timezone aware datetimes raised an error. I reflected this in a separate whatsnew entry in the io bug section.

TomAugspurger:
Small issue with the new test I think https://travis-ci.org/pandas-dev/pandas/jobs/451739169#L1986

```
_____ TestMySQLAlchemy.test_naive_datetimeindex_roundtrip[load_iris_data0] _____
self = <pandas.tests.io.test_sql.TestMySQLAlchemy object at 0x7febe71173d0>
    def test_naive_datetimeindex_roundtrip(self):
        # GH 23510
        # Ensure that a naive DatetimeIndex isn't converted to UTC
        dates = date_range('2018-01-01', periods=5, freq='6H')
        expected = DataFrame({'nums': range(5)}, index=dates)
        expected.to_sql('foo_table', self.conn, index_label='info_date')
        result = sql.read_sql_table('foo_table', self.conn
                                    index_col='info_date')
>       tm.assert_frame_equal(result, expected)
E       AssertionError: DataFrame.index are different
E       
E       Attribute "names" are different
E       [left]:  [u'info_date']
E       [right]: [None]
```

looks good otherwise.

mroeschke:
Thanks @TomAugspurger. Fixed that test (I don't think index name is expected to roundtrip in that test)

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -1275,6 +1246,9 @@ MultiIndex
 I/O
 ^^^
 
+- Bug in :meth:`to_sql` when writing timezone aware data (``datetime64[ns, tz]`` dtype) would raise a ``TypeError`` (:issue:`9086`)
```
jorisvandenbossche: I would move this to the "enhancements" though, I would say that timezones were simply never supported, so it is a nice enhancement that we will now actually support it.

pandas/io/sql.py
```diff
@@ -852,8 +852,13 @@ def _sqlalchemy_type(self, col):
         if col_type == 'datetime64' or col_type == 'datetime':
             # GH 9086: TIMESTAMP is the suggested type if the column contains
             # timezone information
-            if col.dt.tz is not None:
-                return TIMESTAMP(timezone=True)
+            try:
+                if col.dt.tz is not None:
+                    return TIMESTAMP(timezone=True)
+            except AttributeError:
+                # The column is actually a DatetimeIndex
```
jorisvandenbossche: just for my understanding: where was this attribute error catched before?

doc/source/whatsnew/v0.24.0.txt
jorisvandenbossche: Ah, now looking at the full diff (and not only what changed recently), and see you actually already have that. It's a bit duplicated now, but I am fine with keeping it in both places.

pandas/io/sql.py
jorisvandenbossche: Whoop, sorry, again misunderstanding from not looking at the full diff :-)

jorisvandenbossche (approved):

jorisvandenbossche:
Lets's finally merge this! @mroeschke thanks for the endurance :-)

Issue #15019, opened 2016-12-31
BUG: to_html misses truncation indicators (...) when index=False


jorisvandenbossche:
When the html output is truncated (by using `mmax_rows=..`), you get `...` to indicate this, but not when using ``index=False``:

```python
df = pd.DataFrame(np.random.randn(5,2))
df.to_html(max_rows=4, index=False)
```

![screenshot from 2016-12-31 11-02-49](https://cloud.githubusercontent.com/assets/1020496/21576981/ecdd1b1a-cf48-11e6-85a9-a3cd50500cdc.png)

cc @TomAugspurger follow-up from #14999

simonjayhawkins:
> to_html misses truncation indicators (...) when index=False

and when `index=True` and `max_rows=1` which is inconsistent with `max_cols=1` behavior:

![image](https://user-images.githubusercontent.com/13159005/45191109-4aa4e280-b239-11e8-9325-8dc696282efd.png)

simonjayhawkins:
Reproducible code sample to use as test case:

#### Code Sample

```python
np.random.seed(seed=0)
df = pd.DataFrame(np.random.randn(5,2))
df.to_html(max_rows=4, index=False)
```

#### Output

<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th>0</th>
      <th>1</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>1.764052</td>
      <td>0.400157</td>
    </tr>
    <tr>
      <td>0.978738</td>
      <td>2.240893</td>
    </tr>
    <tr>
      <td>0.950088</td>
      <td>-0.151357</td>
    </tr>
    <tr>
      <td>-0.103219</td>
      <td>0.410599</td>
    </tr>
  </tbody>
</table>

#### Expected Output

<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th>0</th>
      <th>1</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>1.764052</td>
      <td>0.400157</td>
    </tr>
    <tr>
      <td>0.978738</td>
      <td>2.240893</td>
    </tr>
    <tr>
      <td>...</td>
      <td>...</td>
    </tr>
    <tr>
      <td>0.950088</td>
      <td>-0.151357</td>
    </tr>
    <tr>
      <td>-0.103219</td>
      <td>0.410599</td>
    </tr>
  </tbody>
</table>

PR #22655, opened 2018-09-10
BUG: output formatting with to_html(), index=False and/or index_names=False (#22579, #22747)

### LINKED ISSUE #22579: Column Offset Bug with to_html(index=False) with MultiIndex Columns and Index
```python
import pandas as pd
from IPython.display import HTML
df = pd.DataFrame({'a': range(10), 'b': range(10,20), 'c': range(10,20), 'd': range(10,20)})
df.columns = pd.MultiIndex.from_product([['a', 'b'], ['c', 'd']])
df.index = pd.MultiIndex.from_product([['a', 'b'], ['c', 'd', 'e', 'f', 'g']])
HTML(df.to_html(index=False))
```

#### Output:

<table border="1" class="dataframe">
  <thead>
    <tr>
      <th></th>
      <th></th>
      <th colspan="2" halign="left">a</th>
      <th colspan="2" halign="left">b</th>
    </tr>
    <tr>
      <th></th>
      <th></th>
      <th>c</th>
      <th>d</th>
      <th>c</th>
      <th>d</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>0</td>
      <td>10</td>
      <td>10</td>
      <td>10</td>
    </tr>
    <tr>
      <td>1</td>
      <td>11</td>
      <td>11</td>
      <td>11</td>
    </tr>
    <tr>
      <td>2</td>
      <td>12</td>
      <td>12</td>
      <td>12</td>
    </tr>
    <tr>
      <td>3</td>
      <td>13</td>
      <td>13</td>
      <td>13</td>
    </tr>
    <tr>
      <td>4</td>
      <td>14</td>
      <td>14</td>
      <td>14</td>
    </tr>
    <tr>
      <td>5</td>
      <td>15</td>
      <td>15</td>
      <td>15</td>
    </tr>
    <tr>
      <td>6</td>
      <td>16</td>
      <td>16</td>
      <td>16</td>
    </tr>
    <tr>
      <td>7</td>
      <td>17</td>
      <td>17</td>
      <td>17</td>
    </tr>
    <tr>
      <td>8</td>
      <td>18</td>
      <td>18</td>
      <td>18</td>
    </tr>
    <tr>
      <td>9</td>
      <td>19</td>
      <td>19</td>
      <td>19</td>
    </tr>
  </tbody>
</table>

#### Problem description

The code sample above is taken from #8452 but with a MultiIndex applied to the index as well as the columns. This seems to reproduce the column offset bug as seen in the original issue. The desired behaviour is observed if either one of the axes is a MultiIndex but not both.
 
<details>

INSTALLED VERSIONS
------------------
commit: None
python: 3.6.5.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
processor: Intel64 Family 6 Model 69 Stepping 1, GenuineIntel
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

pandas: 0.23.0
pytest: 3.5.1
pip: 10.0.1
setuptools: 39.1.0
Cython: 0.28.2
numpy: 1.14.3
scipy: 1.1.0
pyarrow: None
xarray: None
IPython: 6.4.0
sphinx: 1.7.4
patsy: 0.5.0
dateutil: 2.7.3
pytz: 2018.4
blosc: None
bottleneck: 1.2.1
tables: 3.4.3
numexpr: 2.6.5
feather: None
matplotlib: 2.2.2
openpyxl: 2.5.3
xlrd: 1.1.0
xlwt: 1.3.0
xlsxwriter: 1.0.4
lxml: 4.2.1
bs4: 4.6.0
html5lib: 1.0.1
sqlalchemy: 1.2.7
pymysql: None
psycopg2: None
jinja2: 2.10
s3fs: None
fastparquet: None
pandas_gbq: None
pandas_datareader: None
</details>

### LINKED ISSUE #22747: Columns Index Name with to_html(index_names=False) is displayed.
#### Code Sample

```python
df = pd.DataFrame(np.zeros((2, 2), dtype=int))
df.columns.name = 'columns.name'
df.to_html(index_names=False)
```

#### Output

<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th>columns.name</th>
      <th>0</th>
      <th>1</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>0</th>
      <td>0</td>
      <td>0</td>
    </tr>
    <tr>
      <th>1</th>
      <td>0</td>
      <td>0</td>
    </tr>
  </tbody>
</table>

#### Problem description

from the docs:

index_names : bool, optional
Prints the names of the indexes, default True

The problem occurs with both standard and MultiIndex columns index. The expected behavior is observed on the (row) Index, if it has names they are not displayed.

#### Expected Output

<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>0</th>
      <th>1</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>0</th>
      <td>0</td>
      <td>0</td>
    </tr>
    <tr>
      <th>1</th>
      <td>0</td>
      <td>0</td>
    </tr>
  </tbody>
</table>

<details>

</details>

simonjayhawkins:
- [x] closes #22579
- [x] closes #22747
- [x] tests added / passed
- [x] whatsnew entry


pandas/tests/io/formats/test_to_html.py
```diff
@@ -1905,6 +1905,636 @@ def test_to_html_multiindex_max_cols(self):
         </table>""")
         assert result == expected
 
+    def test_to_html_multi_indices_index_false(self):
```
WillAyd: Haven't looked in detail at all of these if only due to the sheer volume of them, but is there not a way for you to parametrize these instead?

WillAyd (changes requested):

simonjayhawkins:
@WillAyd : `index_names=False` and `index=False` issues can be separated. Since there is an issue on the `index_names=False` with `index=True` codepath, I'll raise this as a separate issue and a separate PR.

We can then move forward on the test methodology.

In order to resolve the alignment issues for this PR,  I'll need to add some additional tests to the basic alignment checks that I've done so far to ensure that the truncation indicators are also aligned correctly after shifting the header.

simonjayhawkins:
@WillAyd : looks like I will need to fix #15019 first before i can check the misalignment is corrected on a truncated DataFrame.

WillAyd:
@simonjayhawkins please ping when ready for review

simonjayhawkins:
@WillAyd If you could take another look that'll be great. Thanks.

> @WillAyd : index_names=False and index=False issues can be separated. Since there is an issue on the index_names=False with index=True codepath, I'll raise this as a separate issue and a separate PR.

This PR is closing both these issues. I was originally planning to raise a separate PR to reduce the scope of this one. Please let me know if you think that will still be necessary after having a quick look.

I'll leave this PR marked as WIP for the moment because of potential merge conflicts with #22726

pandas/tests/io/formats/test_to_html.py
```diff
@@ -415,6 +431,85 @@ def test_to_html_multiindex_max_cols(self, datapath):
...
+                        'd': range(10, 20)})
+        df.columns = MultiIndex.from_product([['a', 'b'], ['c', 'd']])
+        df.index = MultiIndex.from_product([['a', 'b'],
+                                            ['c', 'd', 'e', 'f', 'g']])
+        result = df.to_html(index=False)
+        expected = expected_html(datapath, 'gh22579_expected_output')
+        assert result == expected
+
+    @pytest.mark.parametrize(
+        ('idx_flag', 'named_index', 'expected_output'),
```
WillAyd: Can you make the first argument to `parametrize` a single string? I haven't seen a Tuple passed like this before in other parts of the codebase nor in their documentation

pandas/tests/io/formats/test_to_html.py
```diff
@@ -415,6 +431,85 @@ def test_to_html_multiindex_max_cols(self, datapath):
...
+        names = ['foo', None, 'baz'] if named_index else None
+        index = MultiIndex.from_product([['a', 'b'], ['c', 'd'], ['e', 'f']],
+                                        names=names)
+        df = DataFrame(np.arange(64).reshape(8, 8), index=index, columns=index)
+        result = df.to_html(max_rows=4, max_cols=4, index=idx_flag)
+        expected = expected_html(datapath, expected_output)
+        assert result == expected
+
+    @pytest.mark.parametrize('col_idx_type',
+                             [Index([0, 1]),
```
WillAyd: Style nit but can you move the opening bracket up one line here? Should allow for better indentenation of the arguments

pandas/tests/io/formats/test_to_html.py
```diff
@@ -415,6 +431,85 @@ def test_to_html_multiindex_max_cols(self, datapath):
...
+                                                             'index.name.1'])
+                              ],
+                             ids=['index_unnamed_standard',
+                                  'index_named_standard',
+                                  'index_unnamed_multi',
+                                  'index_named_multi'])
+    def test_to_html_basic_alignment(self, datapath, idx_type, col_idx_type,
+                                     index, index_names):
+        # GH 22747, GH 22579
+        def _exp_name(idx_type, index, index_names):
```
WillAyd: This is a rather complicated test. Can you not alternately inspect the ids of the fixture to get what you want instead?

WillAyd (changes requested):

pandas/tests/io/formats/test_to_html.py
simonjayhawkins: the only methods i've found on getting the `id` inside the test function all appear a bit hacky. the 'id' seems to be mainly to make the test output more readable.

i originally coded the index paramatrization as a fixture, but since it is only used for this test moved to a `parametrize` decorator. If i don't have different names for column and row indexes in the tests then a single fixture could be used for both the column and row indexes.

One way to simplify this is to use the the `id` as the parameter and then construct the index inside the test function. That would remove the need to introspect the index. The test would still need to construct the mapping from the test parameters to the expected output so would not remove the apparent complexity while making it more complicated to move the indexes into a fixture if they could be used in another test.

The mapping could be simplified by creating the map explicitly but this would result in a dict with  64 (4 x 4 x 2 x 2) loc. I am planning to add header parametrization to this test in a future PR to fix a couple of `header=False` issues. the explicit map would then be 128 loc.

I have tried all these methods at some point and this solution I believe is the simplest without contaminating the module namespace. Adding a `is_multiindex` and `has_names` helper function into the module namespace would make the code more readable.

```python
def _has_names(index):
    if index.nlevels > 1:
        return com._any_not_none(*index.names)
    else:
        return index.name is not None
```

pandas/tests/io/formats/test_to_html.py
simonjayhawkins: much better! thanks.

pandas/tests/io/formats/test_to_html.py
simonjayhawkins: issue #23788 raised for the header=False issues which is why the header parameterisation has been removed from the basic alignment checks for this PR but will be added back in when this issue is addressed.

WillAyd: If inspecting the id isn't an option couldn't you just pass in the name as an argument instead?

pandas/tests/io/formats/test_to_html.py
```diff
@@ -11,6 +11,22 @@
 from pandas.compat import u, lrange, StringIO
 from pandas.util import testing as tm
 import pandas.io.formats.format as fmt
+import pandas.core.common as com
+
+
+@pytest.fixture(params=[True, False], ids=['index_true', 'index_false'])
```
WillAyd: I don't think this warrants a separate fixture; fine to just pass True / False via parametrization to the functions where required

WillAyd (changes requested):
Sorry missed your last update. Still looking for simplification of tests here

WillAyd (changes requested):
Generally having a tough time with readability. Might have a few more rounds of edits but headed in the right direction

pandas/tests/io/formats/test_to_html.py
```diff
@@ -415,6 +415,91 @@ def test_to_html_multiindex_max_cols(self, datapath):
...
+         'index_named_multi')
+    ],
+        ids=['index_unnamed_standard',
+             'index_named_standard',
+             'index_unnamed_multi',
+             'index_named_multi'])
+    def test_to_html_basic_alignment(self, datapath, idx_type, col_idx_type,
+                                     index, index_names):
+        # GH 22747, GH 22579
+        idx_type, idx_type_id = idx_type
```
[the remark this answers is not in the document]
simonjayhawkins: > the only methods i've found on getting the id inside the test function all appear a bit hacky. the 'id' seems to be mainly to make the test output more readable.

> If inspecting the id isn't an option couldn't you just pass in the name as an argument instead?

in response to your comment the ids are passed along with the index.. using a tuple is a hack to pass the id along with the index, they are not 2 separate parameters. introspecting the indexes was a better solution in my opinion.

if i pass just the id as a parameter then the index would need to be constructed within the test. with header parametrization we create 256 indexes instead of 8 and even more if we add additional tests when a fixture could have been used.

pandas/tests/io/formats/test_to_html.py
```diff
@@ -415,6 +415,91 @@ def test_to_html_multiindex_max_cols(self, datapath):
...
+        names = ['foo', None, 'baz'] if named_index else None
+        index = MultiIndex.from_product([['a', 'b'], ['c', 'd'], ['e', 'f']],
+                                        names=names)
+        df = DataFrame(np.arange(64).reshape(8, 8), index=index, columns=index)
+        result = df.to_html(max_rows=4, max_cols=4, index=idx_flag)
+        expected = expected_html(datapath, expected_output)
+        assert result == expected
+
+    @pytest.mark.parametrize(
+        'index_names', [True, False], ids=['index_names_true',
```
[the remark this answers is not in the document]
WillAyd: Understood but we are sacrificing readability of the code base for readability of the log file and I'm not sure it's worth it. Especially for the True/False items it's not terribly difficult to map those failures back to the items that failed and we aren't used `ids` that liberally elsewhere in the tests

pandas/io/formats/html.py
```diff
@@ -195,7 +198,11 @@ def write_result(self, buf):
 
     def _write_header(self, indent):
         truncate_h = self.fmt.truncate_h
-        row_levels = self.frame.index.nlevels
+        if self.fmt.index:
+            row_levels = self.frame.index.nlevels
+        else:
+            row_levels = 1 if self.show_col_idx_names else 0
```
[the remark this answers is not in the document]
simonjayhawkins: row_levels is used to determine the number of header tags, number of blank cells before the column header etc. it is the column number(of the html table) where the data starts.

if showing the (row) index (if self.fmt.index) the number of cells in the html table before the data is the number of levels in the row index (row_levels = self.frame.index.nlevels).

if not showing the (row) index, the number of cells in the html table before the data is 1 (blank) if the column names are displayed and none if the column names are not displayed or the column index is not named (self.show_col_idx_names) 

`        self.show_col_idx_names = all((self.fmt.has_column_names,
                                      self.fmt.show_index_names,
                                      self.fmt.header))
`

pandas/tests/io/formats/test_to_html.py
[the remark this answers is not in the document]
simonjayhawkins: since the indexes have been moved from a fixture into the parametrize decorator for this PR, then with

`@pytest.mark.parametrize('col_idx_type, col_idx_type_id', [
        (Index([0, 1]),
         'columns_unnamed_standard'),`

`def test_to_html_basic_alignment(
            self, datapath, idx_type, idx_type_id, col_idx_type,
            col_idx_type_id, index, index_names):`

it is possible for this PR to remove the unpacking but this would prevent easily making it a fixture in the future.

pandas/tests/io/formats/test_to_html.py
```diff
@@ -415,6 +415,91 @@ def test_to_html_multiindex_max_cols(self, datapath):
...
+                                           'index_names_false'])
+    @pytest.mark.parametrize(
+        'index', [True, False], ids=['index_true', 'index_false'])
+    @pytest.mark.parametrize('col_idx_type', [
+        (Index([0, 1]),
```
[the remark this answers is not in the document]
simonjayhawkins: i'm probably not following you here but if i think i understand then it's again related to the readability of the test log files, having four parameters instead of six..

 pandas/tests/io/formats/test_to_html.py::TestToHTML::test_to_html_basic_alignment[True-False-True-True-False-False] PASSED [ 66%]

simonjayhawkins:
@WillAyd I'll make the changes as you have suggested since if the tests are all passing the readability of the test logs is irrelevant. however, while coding the bug fixes the improved readability of the test logs were invaluable.

pandas/tests/io/formats/test_to_html.py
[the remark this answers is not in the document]
simonjayhawkins: a side-effect of adding the type as an extra parameter is the test log file is readable again.. nice!

pandas/tests/io/formats/test_to_html.py::TestToHTML::test_to_html_basic_alignment[row_index1-named_standard-column_index3-named_multi-False-False] PASSED [ 66%]

pandas/tests/io/formats/test_to_html.py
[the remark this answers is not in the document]
simonjayhawkins: thats OK for a single failure, but a bit mind boggling when looking at many failures across 128 tests.

pytest -x to the rescue here.

pandas/tests/io/formats/test_to_html.py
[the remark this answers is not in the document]
simonjayhawkins: following this PR i was hoping to extend the parametric alignment checking to the other output formatting methods. hence much refactoring here was to create fixtures separate from the actual test which could have easily been copy and pasted to conftest.py in the next PR. we'll cross that bridge...

pandas/tests/io/formats/test_to_html.py
```diff
@@ -415,6 +431,85 @@ def test_to_html_multiindex_max_cols(self, datapath):
         expected = expected_html(datapath, 'gh6131_expected_output')
         assert result == expected
 
+    def test_to_html_multi_indexes_index_false(self, datapath):
+        # GH 22579
+        df = DataFrame({'a': range(10), 'b': range(10, 20), 'c': range(10, 20),
```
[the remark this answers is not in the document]
simonjayhawkins: noted.

doc/source/whatsnew/v0.24.0.rst
```diff
@@ -1389,6 +1389,8 @@ Notice how we now instead output ``np.nan`` itself instead of a stringified form
 - :func:`read_sas()` will correctly parse sas7bdat files with data page types having also bit 7 set (so page type is 128 + 256 = 384) (:issue:`16615`)
 - Bug in :meth:`detect_client_encoding` where potential ``IOError`` goes unhandled when importing in a mod_wsgi process due to restricted access to stdout. (:issue:`21552`)
 - Bug in :func:`to_html()` with ``index=False`` misses truncation indicators (...) on truncated DataFrame (:issue:`15019`, :issue:`22783`)
+- Bug in :func:`to_html()` with ``index=False`` when both columns and row index are MultiIndex (:issue:`22579`)
```
jreback: can you add a double backticks on MultiIndex

pandas/io/formats/html.py
```diff
@@ -264,12 +271,13 @@ def _write_header(self, indent):
                         values = (values[:ins_col] + [u('...')] +
                                   values[ins_col:])
 
-                name = self.columns.names[lnum]
-                row = [''] * (row_levels - 1) + ['' if name is None else
-                                                 pprint_thing(name)]
-
-                if row == [""] and self.fmt.index is False:
-                    row = []
+                row = [''] * (row_levels - 1)
```
jreback: might be worth it to make this a function / method as it appears to repeat below.

can you add a comment on what this is doing

jreback (changes requested):

pandas/io/formats/html.py
simonjayhawkins: the fixes for both the issues with the standard index and and issues with the multiindex have been purposefully made to use similar code to make the subsequent refactor simpler.

imo the refactor should create a `_get_formatted_column_labels` method for code parity with the `DataFrameFormatter` class but that refactor is outside the scope of this PR. the `DataFrameFormatter` class also has a `_get_column_name_list` which could be used.

However, i've not yet done this refactor because I want to get the bugs fixed first and the parametric tests in place for added assurance. I really wanted to limit the scope of this PR to the bug fixes since there is also a couple of issues with the header=False and sparsify=False which really need sorting before refactoring.

I will happily add a function but it would only be temporary. please let me know your thoughts.

jreback: ok that’s fine for then
pls add comments though and a TODO here as well

pandas/io/formats/html.py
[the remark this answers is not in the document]
simonjayhawkins: @WillAyd 

> What gets shown when this evaluates to True?

row_levels=1 in `_write_header` only affects the placement of the truncation indicators for a single level named columns index.

Changing the 1 to a 0 does not fail any tests so i've added an additional test to gh-22783 to ensure we get a test failure if `row-levels` is not 1.

simonjayhawkins:
@WillAyd @jreback comments addressed. ptal.

pandas/io/formats/html.py
```diff
@@ -42,6 +42,15 @@ def __init__(self, formatter, classes=None, notebook=False, border=None,
         self.border = border
         self.table_id = table_id
         self.render_links = render_links
+        # see gh-22579
+        # Column misalignment also occurs for
+        # a standard index when the columns index is named.
+        # Determine if ANY column names need to be displayed
+        # since if the row index is not displayed a column of
+        # blank cells need to be included before the DataFrame values.
+        self.show_col_idx_names = all((self.fmt.has_column_names,
```
WillAyd: Sorry for questions but still trying to wrap my head around implementation. Based off of the comment, why is this `all` here and not `any`? Wouldn't any of these require there to be a cell where a column index name would be placed?

simonjayhawkins: `index=False` with a single level row index and multi-level columns index with named columns but not all named...

```python
index = pd.MultiIndex.from_product([['a','b'], ['c','d'], ['e','f']], names=[
                 'foo',None, 'baz'])
df = pd.DataFrame(np.arange(64).reshape(8,8), columns=index)
result = df.to_html(max_rows=4, max_cols=4, index=False)
print(result)
```

<table border="1" class="dataframe">
  <thead>
    <tr>
      <th>foo</th>
      <th colspan="2" halign="left">a</th>
      <th>...</th>
      <th colspan="2" halign="left">b</th>
    </tr>
    <tr>
      <th colspan="2" halign="left">c</th>
      <th>...</th>
      <th colspan="2" halign="left">d</th>
    </tr>
    <tr>
      <th>baz</th>
      <th>e</th>
      <th>f</th>
      <th>...</th>
      <th>e</th>
      <th>f</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>0</td>
      <td>1</td>
      <td>6</td>
      <td>7</td>
    </tr>
    <tr>
      <td>8</td>
      <td>9</td>
      <td>14</td>
      <td>15</td>
    </tr>
    <tr>
      <td>48</td>
      <td>49</td>
      <td>54</td>
      <td>55</td>
    </tr>
    <tr>
      <td>56</td>
      <td>57</td>
      <td>62</td>
      <td>63</td>
    </tr>
  </tbody>
</table>

Note: missing truncation indicators in data now fixed in master.

the misalignment of the column names is due to the logic being applied within the level generating loop..

https://github.com/pandas-dev/pandas/blob/d43ac979fdf3038c713de93c7b27b4140be6c9c5/pandas/io/formats/html.py#L270-L275

hence class-level variable needed to check if ANY names need to be displayed to determine alignment.

ALL condition is to determine in ANY names should be displayed given the `to_html` parameters and uses similar logic as `to_string` etc.

https://github.com/pandas-dev/pandas/blob/d43ac979fdf3038c713de93c7b27b4140be6c9c5/pandas/io/formats/format.py#L796-L803

and the rows in `to_html`..

https://github.com/pandas-dev/pandas/blob/d43ac979fdf3038c713de93c7b27b4140be6c9c5/pandas/io/formats/html.py#L307-L309

There is currently no test to explicitly cover this example. so i think the best way forward is to fully parameterize the truncation tests in line with the parametrized basic_alignment tests for added assurance.

i'll make `show_col_idx_names` a class property for clarity and add a note to refactor and 'inherit' from `DataFrameFormatter` class. inherit quoted since `HTMLFormatter` class is not directly inherited from `DataFrameFormatter`. in the first refactor just use mock inheritence like..

https://github.com/pandas-dev/pandas/blob/d43ac979fdf3038c713de93c7b27b4140be6c9c5/pandas/io/formats/html.py#L46-L48

simonjayhawkins:
@jreback @WillAyd with the additional parameterization of the truncation tests, we now have test coverage for multi-indexes with more than 2 rows, missing column index names and truncation with standard row indexes. There is now test coverage in place allowing the refactoring of `row-levels` to class property in this PR for use by `_write_header` and `_write_regular_rows`. I've added a TODO in `_write_hierarchical_rows` to refactor after #22887 is fixed

simonjayhawkins:
@WillAyd @jreback Could you please take another look. Thanks.

jreback:
I would make a sub-dir of data/html to hold all of this test data (and move the original .html files as well).

jreback (approved):

jreback:
@WillAyd over to you

WillAyd (approved):

WillAyd:
Thanks @simonjayhawkins !

jreback:
thanks @simonjayhawkins !

PR #22658, opened 2018-09-10
DOC iteritems docstring update and examples

Ecboxer:
Updated iteritems docstring to start with an infinitive and added a short example

- [ ] closes #xxxx
- [ ] tests added / passed
- [ ] whatsnew entry


pandas/core/frame.py
```diff
@@ -726,13 +726,32 @@ def style(self):
...
         See also
         --------
         iterrows : Iterate over DataFrame rows as (index, Series) pairs.
         itertuples : Iterate over DataFrame rows as namedtuples of the values.
 
+        Examples
+        --------
+
+        >>> df = pd.DataFrame({'col1': [1, 2], 'col2': [0.1, 0.2]},
+                             index=['a', 'b'])
```
TomAugspurger: Continuation lines need to start with `...` (right below the `>>>`)

TomAugspurger: Also, make sure the `i` in `index` is right below the opening `{` on the line above.

TomAugspurger:
You can run the validation script locally FYI. `python scripts/validate_docstring.py pandas.DataFrame.iteritems` I think.

pandas/core/frame.py
```diff
@@ -726,13 +726,32 @@ def style(self):
 
     def iteritems(self):
         """
-        Iterator over (column name, Series) pairs.
```
mroeschke: Technically this returns an `Iterator` (or more specifically a `Generator`). So I would leave this as `Iterator` or change this to `Generator`.

TomAugspurger:
@Ecboxer do you have time to update?

Ecboxer (commented):
> @Ecboxer do you have time to update?

The edited commit passed my local tests, I would like to try to update though. Can you direct me on how to understand/correct the issues in the checks?

pandas/core/frame.py
```diff
@@ -728,11 +728,35 @@ def iteritems(self):
...
+
+        Returns
+        -------
+        it : generator
+            A generator that iterates over the columns of the frame.
+
         See also
         --------
         iterrows : Iterate over DataFrame rows as (index, Series) pairs.
         itertuples : Iterate over DataFrame rows as namedtuples of the values.
```
datapythonista: `A` in `See Also` should be capitalized.

Use `DataFrame.iterrows` instead of `iterrows` and same for `itertuples`.

pandas/core/frame.py
```diff
@@ -728,11 +728,35 @@ def iteritems(self):
...
 
+        Examples
+        --------
+        >>> df = pd.DataFrame({'col1': [1, 2], 'col2': [0.1, 0.2]},
+        ...                   index=['a', 'b'])
```
datapythonista: Can you use something that looks more like a real world example? I think it makes things easier to read.

See for example the example DataFrame in https://pandas-docs.github.io/pandas-docs-travis/generated/pandas.DataFrame.reset_index.html

pandas/core/frame.py
```diff
@@ -728,11 +728,35 @@ def iteritems(self):
...
+        >>> df
+           col1 col2
+        a     1  0.1
+        b     2  0.2
+        >>> for col in df.iteritems():
+        ...     print(col)
```
datapythonista: I think something like next would make things easier to understand:
```
for label, content in df.iteritems():
    print('label:', label)
    print('content:' content)
```

datapythonista (changes requested):
Added some comments

pandas/core/frame.py
Ecboxer: Thanks for the edits! I rephrased the long description, changed the Yields section to cover label and content, prepended DataFrame. to the See Also examples, and reworked the Examples section to hopefully be more engaging.

datapythonista: did you push the changes?

Ecboxer: I got my git commands mixed up *sheepishly*, but it's pushed now

pandas/core/frame.py
```diff
@@ -781,11 +781,50 @@ def iteritems(self):
...
+        >>> df
+                species   pop
+        black 	bear 	  300000
+        brown 	bear 	  200000
+        panda 	bear 	  1864
+        polar 	bear 	  22000
+        koala 	marsupial 80000
+        >>> for label, content in df.iteritems():
+        ...     print('label:', label)
+        ...     print('content:', content)
```
datapythonista: I think for this second print we could have `print('content:', content, sep='\n', end='\n')` and the output would be easier to read.

And use `population` istead of `pop`.

Other than that, looks great, thank you!

Ecboxer: Gotcha. I didn't know that print could take sep and end args, that's neat!

datapythonista:
Can you run the validation script? I think the output in the tests is not accurate.

You can do it by `./scripts/validate_docstrings.py pandas.DataFrame.iteritems` (you need pandas compiled), and it should say everything is ok.

Ecboxer:
> Can you run the validation script? I think the output in the tests is not accurate.
> 
> You can do it by `./scripts/validate_docstrings.py pandas.DataFrame.iteritems` (you need pandas compiled), and it should say everything is ok.

When I run scripts/validate_docstrings.py pandas.DataFrame.iteritems I get the following error under #Validation#
Errors found:
	No Returns section found
(When I renamed Yields to Returns, I would get No Yields section found). Should I make some sort of empty Returns section?

datapythonista:
If that's the only error, ignore it, it may be an error in the validation script.

I'd just make a quick update if you don't mind. Instead of using `\\n` in the code to escape the backslash, I think it'd be more readable if we set the docstring as a raw string with (`r"""` at the beginning). And may be we can set `end=`\n\n`? that would add a blank line between the two items (I didn't realised that `end='\n'` is actually the default, so it's actually doing nothing now). If you think the `\n\n` it's adding too much complexity, just get rid of the `end` parameter.

Thanks!

Ecboxer:
> If that's the only error, ignore it, it may be an error in the validation script.
> 
> I'd just make a quick update if you don't mind. Instead of using `\\n` in the code to escape the backslash, I think it'd be more readable if we set the docstring as a raw string with (`r"""` at the beginning). And may be we can set `end=`\n\n`? that would add a blank line between the two items (I didn't realised that`end='\n'`is actually the default, so it's actually doing nothing now). If you think the`\n\n`it's adding too much complexity, just get rid of the`end` parameter.
> 
> Thanks!

I tried it with the end='\n\n' but there was another Validation error for 
Errors found:
I've dropped the end parameter, because I'm not sure how to get around that test.

datapythonista (approved):
Ok, I think this is very clear anyway. Looks great, really useful docstring. Thanks a lot @Ecboxer

pandas/core/frame.py
```diff
@@ -778,14 +778,55 @@ def style(self):
         return Styler(self)
 
     def iteritems(self):
-        """
+        r"""
         Iterator over (column name, Series) pairs.
 
-        See also
+        Iterates over the DataFrame columns, returning a tuple with the column name and the content as a Series.
```
jorisvandenbossche: this doesn't fit within 80 chars line length limit (PEP8), I assume the CI is failing due to that.

There are also some other lines below that are too long. I would recommend to activate a flake8 / linter plugin in your editor, or run `flake8` locally. You can also check the output of the failing build on Travis.

pandas/core/frame.py
```diff
@@ -778,14 +778,55 @@ def style(self):
...
-        iterrows : Iterate over DataFrame rows as (index, Series) pairs.
-        itertuples : Iterate over DataFrame rows as namedtuples of the values.
+        DataFrame.iterrows : Iterate over DataFrame rows as (index, Series) pairs.
+        DataFrame.itertuples : Iterate over DataFrame rows as namedtuples of the values.
 
+        Examples
+        --------
+        >>> df = pd.DataFrame({'species': ['bear', 'bear', 'bear', 'bear', 'marsupial'],
+        ...                   'population': [300000, 200000, 1864, 22000, 80000]},
+        ...                   index=['black', 'brown', 'panda', 'polar', 'koala'])
```
jorisvandenbossche: I would make it a little bit shorter (eg 3 rows seems enough), that will make the output below more clear I think

pandas/core/frame.py
```diff
@@ -778,14 +778,55 @@ def style(self):
...
 
-        See also
+        Iterates over the DataFrame columns, returning a tuple with the column name and the content as a Series.
+
+        Yields
+        ------
+        label : object
+            The column names for the DataFrame being iterated over.
+        content : Series
+            The column entries belonging to each label, as a Series.
```
jorisvandenbossche: @datapythonista I think in this case, the above is actually a bit confusing. Typically, we use the formatting above if there are actually two return values (so if you could do `label, content = df.iteritems()`), which is not the case here. 
So I think the original single item was better, but we could try to make it clearer that it the values of the iterator consist of those two items.

jorisvandenbossche (commented):
@Ecboxer thanks for working on this! Added some more comments

pandas/core/frame.py
datapythonista: Looks clear to me, not sure why it doesn't to you. In this case you can do `for label, content in df.iteritems():` which is equivalent to what you said.

Not a big deal changing this to a `Returns` saying it's a generator returning tuples. But I don't think that would be clearer to me, and feels a bit inconsistent.

What do you think is clearer for you @Ecboxer? Also, may be @WillAyd want to give an opinion, and he's doing a lot with the docstrings? Happy with whatever option is clearer to most people.

Ecboxer: Let me know what you think of the rephrased it under Yields. It may be too wordy?

jorisvandenbossche: `for label, content in df.iteritems()` is *not* the same as `label, content = df.iteritems()` ..

The thing is that otherwise we are using the same visual formatting to mean two different things. I would prefer that a user can know from the return type if there is a single or multiple return values (but maybe I am overestimating our users?)

We can maybe still combine both, something like:
```
Iterator over (label, content) pairs
    label : object
        The column names for the DataFrame being iterated over.
    content : Series
        The column entries belonging to each label, as a Series.
```

or does that only make it more complicated?

jorisvandenbossche: Ah sorry, I missed that it was a "Yields" section, and not a "Returns" section. In that case, it is correct that it yields two values in each iteration!

jorisvandenbossche: @Ecboxer sorry, you can change it back to how it was before I commented :-)

datapythonista: hehe, I see what you meant now. Cool then. :)

Ecboxer: Changed it back :)

jorisvandenbossche (approved):

jorisvandenbossche:
@Ecboxer Thanks!

Issue #22664, opened 2018-09-11
API: Add extension type registry to API


TomAugspurger:
In `pandas.api.extensions` probably.

TomAugspurger:
Actually, I prefer something like

```
def register(cls):
    """Class decorator to register an ExtensionType with pandas.

    This enables operations like ``.astype(name)`` for the name
    of the ExtensionDtype.

    Examples
    --------
    >>> from pandas.api.extensions import register, ExtensionDtype
    >>> @register
    ... class MyExtensionDtype(ExtensionDtype):
    ...     pass
    """
    registry.register(cls)
    return cls

```

That gives us a bit more freedom in changing how `registry` is implemented.

PR #22666, opened 2018-09-11
API: register_extension_dtype class decorator

TomAugspurger:
Closes https://github.com/pandas-dev/pandas/issues/22664

### DIFF
--- a/pandas/tests/extension/base/dtype.py
+++ b/pandas/tests/extension/base/dtype.py
@@ -1,4 +1,3 @@
-import pytest
 import numpy as np
 import pandas as pd
 
@@ -50,10 +49,6 @@ class BaseDtypeTests(BaseExtensionTests):
 
     def test_array_type(self, data, dtype):
         assert dtype.construct_array_type() is type(data)
-
-    def test_array_type_with_arg(self, data, dtype):
-        with pytest.raises(NotImplementedError):
-            dtype.construct_array_type('foo')
 
     def test_check_dtype(self, data):
         dtype = data.dtype
--- a/pandas/tests/extension/test_categorical.py
+++ b/pandas/tests/extension/test_categorical.py
@@ -77,9 +77,7 @@ def data_for_grouping():
 
 
 class TestDtype(base.BaseDtypeTests):
-
-    def test_array_type_with_arg(self, data, dtype):
-        assert dtype.construct_array_type() is Categorical
+    pass
 
 
 class TestInterface(base.BaseInterfaceTests):
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -2559,6 +2559,7 @@ objects.
 .. autosummary::
    :toctree: generated/
 
+   api.extensions.register_extension_dtype
    api.extensions.register_dataframe_accessor
    api.extensions.register_series_accessor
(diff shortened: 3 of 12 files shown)


pandas/tests/extension/base/dtype.py
```diff
@@ -51,10 +50,6 @@ def test_eq_with_numpy_object(self, dtype):
     def test_array_type(self, data, dtype):
         assert dtype.construct_array_type() is type(data)
 
-    def test_array_type_with_arg(self, data, dtype):
```
TomAugspurger: This test could never pass or fail correctly. It would always raise a TypeError since the method doesn't take an argument.

pandas/tests/extension/test_categorical.py
```diff
@@ -77,9 +77,7 @@ def data_for_grouping():
 
 
 class TestDtype(base.BaseDtypeTests):
-
-    def test_array_type_with_arg(self, data, dtype):
```
jreback: this is a legit test

jreback (changes requested):

jreback:
also can you change the registration of the pandas EA types to use the new register function (e.g. Categorical, Interval etc)

pandas/tests/extension/test_categorical.py
TomAugspurger: What's it testing? That code wasn't ever run on master.

TomAugspurger:
Updated for categorical and interval. Couldn't do the same for the various Integer dtypes, since they're generated dynamically.

jreback:
> Couldn't do the same for the various Integer dtypes, since they're generated dynamically.

https://github.com/pandas-dev/pandas/blob/master/pandas/core/arrays/integer.py#L621

just a 1-liner to change

TomAugspurger:
Right... fixed.

TomAugspurger:
All green (this was split from the SparseArray PR fyi. I think I have one or two more things to split before)

jreback:
lgtm

though i think should re-evaluate the test (could be that wasn’t testing anything), but this should have a test

pandas/tests/extension/test_categorical.py
jorisvandenbossche: What do you mean with "That code wasn't ever run on master." ? That the test is never run, or that what it is testing is never used anywhere in the code? 

Those tests are certainly running, but it is true that the base class test is not very useful (and also wrong with the string argument). 
But I think it is still potentially useful to assert that the dtypes `construct_array_type` is giving an ExtensionArray class object

jorisvandenbossche (approved):

pandas/tests/extension/test_categorical.py
TomAugspurger: The test was overridden by every subclass. Note that we still have https://github.com/pandas-dev/pandas/pull/22666/files/9b646e28656ca3c61ee8a221e16ad74bba2610c3#diff-32e4b328fc01507825a6249caac0cb21R50, which tests this properly.

jorisvandenbossche: > Note that we still have .., which tests this properly.

Ah, yes, that is indeed what the other overridden ones were also doing, but then manually.

PR #22667, opened 2018-09-11
is_bool_dtype for ExtensionArrays

### LINKED ISSUE #22326: Clarify is_bool_indexer for Extension dtypes
What do we want here?

```python
In [1]: import pandas as pd

In [2]: pd.core.common.is_bool_indexer(pd.Categorical([True, True]))
Out[2]: False
```

working around this in https://github.com/pandas-dev/pandas/pull/22325/

### LINKED ISSUE #22665: BUG: CategoricalIndex can't be a boolen mask
```python
In [1]: import pandas as pd

In [2]: idx = pd.CategoricalIndex([True, False, True])

In [3]: pd.Series(range(3))[idx]
Out[3]:
True    NaN
False   NaN
True    NaN
dtype: float64
```

Expected:

```python
In [4]: pd.Series(range(3))[idx.astype(object)]
Out[4]:
0    0
2    2
dtype: int64
```

Fix coming shortly.

TomAugspurger:
Closes #22326
Closes #22665 

Two commits:  https://github.com/pandas-dev/pandas/commit/fe3500248dec1b21469e94e054f636dc7707fae0 partially implements an Arrow-backed ExtensionArray. We need this as we don't currently have any boolean-values EAs.

https://github.com/pandas-dev/pandas/commit/04029ac94d1e53fbec18264beabccb49c1e002b1 has the fix for the bug.


TomAugspurger:
The bot is wrong, right?

jschendel:
Looks like there's an inconsistency in `is_bool_indexer` with how `NaN` is handled.

Returns `True` for a `Categorical` with `NaN`:
```python
In [2]: cat = pd.Categorical([True, False, np.nan])

In [3]: cat
Out[3]:
[True, False, NaN]
Categories (2, object): [False, True]

In [4]: pd.core.common.is_bool_indexer(cat)
Out[4]: True
```

Raises for an `Index` with `NaN`:
```python
In [5]: idx = pd.Index([True, False, np.nan])

In [6]: idx
Out[6]: Index([True, False, nan], dtype='object')

In [7]: pd.core.common.is_bool_indexer(idx)
ValueError: cannot index with vector containing NA / NaN values
```

Also, it doesn't look like we have any existing tests for `is_bool_indexer`?  Might be nice to have some tests around it, but can probably create a separate issue/pr for that.

jorisvandenbossche (commented):
Should we document how we decide something is considered boolean? (I think it is now the dtype.kind == 'b' check?)

TomAugspurger:
Yes, fixed.

pandas/core/dtypes/base.py
```diff
@@ -169,6 +169,9 @@ def kind(self):
         the extension type cannot be represented as a built-in NumPy
         type.
 
+        This affect whether the ExtensionArray can be used as a boolean
+        mask. ExtensionArrays with ``kind == 'b'`` can be boolean masks.
```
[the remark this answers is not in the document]
jreback: e.g. maybe we should add ``_is_boolean`` as a property to EA base class (and default False), similar to how we have numeric introspection via ``_is_numeric=True`` for Integer & Decimal types

jreback (changes requested):

jreback (commented):
this more directly tests the EA array (alternatively these could be on the Dtype itself), however I am -1 on using ``.kind == 'b'`` directlry for this purpose (and would prefer a method / property for this

pandas/core/dtypes/common.py
```diff
@@ -1626,6 +1640,9 @@ def is_bool_dtype(arr_or_dtype):
         # guess this
         return (arr_or_dtype.is_object and
                 arr_or_dtype.inferred_type == 'boolean')
+    elif is_extension_array_dtype(arr_or_dtype):
+        dtype = getattr(arr_or_dtype, 'dtype', arr_or_dtype)
```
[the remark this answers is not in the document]
TomAugspurger: This is `is_bool_dtype` 😄

TomAugspurger (commented):
> however I am -1 on using .kind == 'b' directlry for this purpose

Why's that? It's the stated purpose of `.kind`. To be clear, I kind of share your opinion. My main reason for maybe not using `.kind` is if the typecodes used by NumPy aren't flexible enough for our needs. But I'm not sure.

So I'm not against adding a `_is_boolean` property, but would like a bit more convincing that it isn't covered by `.kind` already :)

TomAugspurger:
@jreback changed `.kind` to use a new `_is_boolean` property (False by default)

TomAugspurger:
All green.

pandas/core/common.py
```diff
@@ -100,17 +102,45 @@ def maybe_box_datetimelike(value):
...
+    bool
+
+    Raises
+    ------
+    ValueError
+        When the array is an object-dtype ndarray or ExtensionArray
+        and contains missing values.
+    """
+    na_msg = 'cannot index with vector containing NA / NaN values'
+    if (isinstance(key, (ABCSeries, np.ndarray, ABCIndex)) or
```
jreback: is this if clause necessary? e.g. an EA type cannot match key.dtype == np.object_ (which actually should be ``is_object_dtype``

pandas/core/dtypes/common.py
```diff
@@ -1618,6 +1625,13 @@ def is_bool_dtype(arr_or_dtype):
...
         return False
 
+    if isinstance(arr_or_dtype, (ABCCategorical, ABCCategoricalIndex)):
+        arr_or_dtype = arr_or_dtype.dtype
+
+    if isinstance(arr_or_dtype, CategoricalDtype):
+        arr_or_dtype = arr_or_dtype.categories
+        # now we use the special definition for Index
+
     if isinstance(arr_or_dtype, ABCIndexClass):
```
jreback: could be elif here

pandas/tests/extension/arrow/test_bool.py
```diff
@@ -0,0 +1,48 @@
+import numpy as np
+import pytest
+import pandas as pd
+import pandas.util.testing as tm
+from pandas.tests.extension import base
+
+pytest.importorskip('pyarrow', minversion="0.10.0")
```
jreback: woa didn't realize this worked now :>

pandas/tests/extension/arrow/bool.py
```diff
@@ -0,0 +1,102 @@
...
+    def nbytes(self):
+        return sum(x.size for chunk in self._data.chunks
+                   for x in chunk.buffers()
+                   if x is not None)
+
+    def isna(self):
+        return pd.isna(self._data.to_pandas())
+
+    def take(self, indices, allow_fill=False, fill_value=None):
+        from pandas.core.algorithms import take
```
jreback: could put import at the top

pandas/tests/extension/arrow/bool.py
```diff
@@ -0,0 +1,102 @@
+import copy
```
jreback: can you add a module doc-string

jreback (changes requested):
looks good. small question.

pandas/core/common.py
TomAugspurger: It my be redundant with the `is_array_like(key)`. But IIRC the tests here were fairly light, and I don't want to risk breaking the old behavior.

pandas/core/dtypes/common.py
TomAugspurger: Has to be an `if` so that if someone passes a `Categorical` we go `Categorical` -> `Categorical.categories (index)` to this block. We want to go down here since Index has special rules.

jorisvandenbossche (approved):

TomAugspurger:
Going to fix the whatsnew conflict on merge.

Speaking of which... are we going to wait around for circleci to catch up today, or are people OK with merging smallish PRs that haven't finished there?

jorisvandenbossche:
I would go ahead and merge

PR #22673, opened 2018-09-12
DOC: Fix DataFrame.to_xarray doctests and allow the CI to run it.

Moisan:
- [x] tests added / passed

Fix the docstring for DataFrame.to_xarray as described in #22459 . I removed the pandas.panel example since it was deprecated and I replaced it with a MultiIndex.


pandas/core/generic.py
```diff
@@ -2492,17 +2492,25 @@ def to_xarray(self):
         """
         Return an xarray object from the pandas object.
 
+        Return the xarray equivalent of the pandas object. `xarray
+        <http://xarray.pydata.org/en/stable/>`__ is a
+        Python package that allows to handle N-dimensional data.
+
```
gfyoung: I'm not sure if we need to mention this here.

Moisan: I was trying to extend the previous short summary. Do you suggest to simply remove it or is there something else we should mention here?

gfyoung: I think we can remove it.  We describe `xarray` elsewhere in the docs anyhow.

gfyoung (approved):
cc @jreback @datapythonista @WillAyd

datapythonista (commented):
Thanks for the changes @Moisan 

I added some comments about the original docstring, that if you implement them, I think the examples won't only pass the tests, but will be much clearer.

pandas/core/generic.py
```diff
@@ -2494,80 +2494,91 @@ def to_xarray(self):
...
+        >>> index = pd.MultiIndex(levels=[[pd.to_datetime("2018-01-01"),
+        ... pd.to_datetime("2015-05-23"), pd.to_datetime("2015-06-06"),
+        ... pd.to_datetime("2011-02-13"), pd.to_datetime("2014-07-06")],
+        ...    ['one', 'two']],
+        ...   labels=[[0, 0, 1, 1, 2, 2, 3, 3], [0, 1, 0, 1, 0, 1, 0, 1]],
+        ...   names=['first', 'second'])
+
+        >>> s = pd.Series(np.arange(8), index=index)
+        >>> s
```
datapythonista: I find this too complicated for what we need to show. To have a Series with a multiindex with a datetime level, we can have something like:

```
import pandas as pd

df = pd.DataFrame({'date': pd.to_datetime(['2018-01-01', '2018-01-01', '2018-01-02', '2018-01-02']),
                   'animal': ['falcon', 'parrot', 'falcon', 'parrot'],
                   'speed': [350, 18, 361, 15]}).set_index(['date', 'animal'])
df['speed']
```

I haven't used much xarray myself, and not sure what makes sense to show here. May be:
- Series.to_xarray()
- DataFrame.to_xarray()
- DataFrame(with multiindex including datetime).to_xarray()

If that makes sense, I think with the first example, we can have `df.to_xarray()` and `df['max_speed'].to_xarray()`, and then a example like the one I wrote.

@jreback does this make sense?

Sorry for requesting the changes @Moisan, but my I find like the current version gives the idea that we're trying to show something more complex than what we are actually showing.

datapythonista (commented):
Looks really great. Just a couple of comments that if I'm right, should simplify the examples.

pandas/core/generic.py
Moisan: No problem, I'm happy to make the examples more relevant :).

datapythonista (approved):
lgtm, thanks @Moisan 

@jreback I think you comments should be addressed properly, let us know if that's not the case

TomAugspurger:
Thanks!

PR #22681, opened 2018-09-12
TST Replace unittest.TestCase.fail calls, use pytest.raises instead of legacy constructs

rth:
Similarly to https://github.com/numpy/numpy/pull/11933,
> Following pytest migration there are unit tests that used to call unittest.TestCase.fail and now will produce a misleading error message on failure because the fail method no longer exists (test classes just inherit from object).
> 
> This fixes this issue, by raising an exception on failure explicitly.

In addition this uses `pytest.raises` instead of more complicated legacy constructions of the form,
```py
try:
   something
except ExceptedException:
   pass
else:
   raise AssertionError(error_message)
```
A few additional minor issues are discussed in comments below.

### DIFF
--- a/pandas/tests/plotting/test_frame.py
+++ b/pandas/tests/plotting/test_frame.py
@@ -588,17 +588,13 @@ class TestDataFramePlots(TestPlotBase):
     @pytest.mark.slow
     def test_subplots_warnings(self):
         # GH 9464
-        warnings.simplefilter('error')
-        try:
+        with tm.assert_produces_warning(None):
             df = DataFrame(np.random.randn(100, 4))
             df.plot(subplots=True, layout=(3, 2))
 
             df = DataFrame(np.random.randn(100, 4),
                            index=date_range('1/1/2000', periods=100))
             df.plot(subplots=True, layout=(3, 2))
-        except Warning as w:
-            self.fail(w)
-        warnings.simplefilter('default')
 
     @pytest.mark.slow
     def test_subplots_multiple_axes(self):
--- a/pandas/tests/series/test_constructors.py
+++ b/pandas/tests/series/test_constructors.py
@@ -465,7 +465,7 @@ class TestSeriesConstructors(TestData):
         # test that construction of a Series with an index of different length
         # raises an error
         msg = 'Length of passed values is 3, index implies 4'
-        with pytest.raises(ValueError, message=msg):
+        with pytest.raises(ValueError, match=msg):
             Series(input, index=np.arange(4))
 
     def test_constructor_numpy_scalar(self):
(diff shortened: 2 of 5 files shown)


pandas/tests/plotting/test_frame.py
```diff
@@ -588,17 +588,13 @@ def test_subplots_layout(self):
     @pytest.mark.slow
     def test_subplots_warnings(self):
         # GH 9464
-        warnings.simplefilter('error')
-        try:
+        with pytest.warns(Warning):
```
rth: Asserting that a warning is raised is too general. Currently this passes, but no warning related to matplotlib are raise (at least locally for me), instead I get,
```
The list of emitted warnings is: 
[DeprecationWarning("Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working"),
 DeprecationWarning("Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working")].
```

Just FYI, it's an unrelated issue.

pandas/tests/series/test_constructors.py
```diff
@@ -465,7 +465,7 @@ def test_constructor_index_mismatch(self, input):
         # test that construction of a Series with an index of different length
         # raises an error
         msg = 'Length of passed values is 3, index implies 4'
-        with pytest.raises(ValueError, message=msg):
+        with pytest.raises(ValueError, match=msg):
```
rth: I think there might have been a confusion between `message` and `match` here: message is the error printed on failure, `match` is the error message we check in the caught exception.

pandas/tests/plotting/test_frame.py
jreback: use tm.assert_produces_warning

pandas/tests/series/test_constructors.py
gfyoung: That does appear to be the case.  Good catch!

pandas/tests/plotting/test_frame.py
rth: Thanks for the review!

Sorry, I misread the code, it meant to check that no warning are raised, please ignore my comment above. Fixed it now.

jorisvandenbossche (approved):

jreback:
can you rebase on master, ping on green.

rth:
@jreback I rebased -- Travis CI passed but there is still one failure in `pandas.tests.io.parser.test_parsers.TestPythonParser` in CircleCI py35_ascii. I don't think this PR changed that file, though this doesn't seem to be happening on master...

jreback:
thanks @rth

PR #19884, opened 2018-02-24
ENH: Let initialisation from dicts use insertion order for python >= 3.6 (part III)

topper-123:
- [x] closes #19018
- [x] xref #19830 and #19859
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

This PR lets pandas allow initialisation using insertion order for python>=3.6. E.g. ``Series({'B': 1, 'A': 2})`` will retain order and not convert to ``Series({'A': 2, 'B': 1})``. As such this changes the API soemwhat fundamentally (but in a very good way, IMO).

Some questions:
* Should this change be at the top of the whatsnew, in order to increase visibility to this change?
* Should an option be added to ``pd.options``, so user can revert back to the old behaviour? I'm thinking here of users who have made tests based on dict sort-by-values, and whose tests will break by this new change
* Should there be a deprecation procedure, as asked by @gfyoung ?

I also need to look thorugh the documentation if changes should be made there, but would appreciate getting the basic PR approved first.

PR #22687, opened 2018-09-13
Only sort dict columns in from_records for py < 3.6

llawall:
- [ ] closes #xxxx (no existing issue)
- [x] tests passed (no new test added - can add one if necessary)
- [ ] whatsnew entry

DataFrame.from_records does not respect dictionary order on Python 3.6+. For example:

```
>>> df = pd.DataFrame.from_records({'C':[1], 'A':[2]})
>>> df
   A  C
0  2  1
```

This change respects dict ordering on Python 3.6+ in line with 0.23 pandas behaviour of constructor and from_dict:

```
>>> df = pd.DataFrame.from_records({'C':[1], 'A':[2]})
>>> df
   C  A
0  1  2
```

There appears to be further potential to reduce duplication if from_records called from_dict?

### DIFF
--- a/pandas/core/frame.py
+++ b/pandas/core/frame.py
@@ -1293,7 +1293,8 @@ class DataFrame(NDFrame):
 
         if isinstance(data, dict):
             if columns is None:
-                columns = arr_columns = ensure_index(sorted(data))
+                columns = arr_columns = ensure_index(
+                    com.dict_keys_to_ordered_list(data))
                 arrays = [data[k] for k in columns]
             else:
                 arrays = []
--- a/pandas/tests/frame/test_constructors.py
+++ b/pandas/tests/frame/test_constructors.py
@@ -2104,6 +2104,15 @@ class TestDataFrameConstructors(TestData):
         for r in results:
             tm.assert_frame_equal(r, df)
 
+    @pytest.mark.skipif(not PY36, reason='Insertion order for Python>=3.6')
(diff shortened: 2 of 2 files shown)


WillAyd:
Would at the very least need tests bundled with this, but can you also open an issue to discuss? The existing behavior is an explicit sort, not necessarily an undefined order for Python < 3.6 so it's not as if this is a bug being fixed but rather a behavioral change that could use discussion

TomAugspurger:
Also, this would be somewhat unique as you have the potential for dicts of different orders in `records`. I believe in all our other changes to respect dict ordering, the data are columnar so you don't have this potential ambiguity.

llawall:
@WillAyd : 
* I'll add a test for this - that is no problem. 
* With respect to "explicit sort", what makes this sort any different to the change to the change in [_init_dict](https://github.com/pandas-dev/pandas/pull/19884/files#diff-1e79abbbdd150d4771b91ea60a4e1cc7) used by constructor? Previously an OrderedDict was special cased but the columns were otherwise always sorted too. 

@TomAugspurger :
* I'm not quite sure I understand what you mean by "dicts of different orders in `records`"? 

I've opened #22708 for further discussion about this. Let's move the discussion there.

WillAyd:
Can you merge master here?

jreback:
closing as stale. if you want to continue, pls ping.

PR #22694, opened 2018-09-13
TST: Test for bug fixed during #22534 discussion

jbrockmendel:
In the course of profiling #22534 a bug was found, but that PR was made redundant and the bug fixed in master.  This just adds a test for that bug.

### DIFF
--- a/pandas/tests/frame/test_arithmetic.py
+++ b/pandas/tests/frame/test_arithmetic.py
@@ -100,6 +100,16 @@ class TestFrameComparisons(object):
 # Arithmetic
 
 class TestFrameFlexArithmetic(object):
+    def test_df_add_td64_columnwise(self):
+        # GH#22534 Check that column-wise addition broadcasts correctly
+        dti = pd.date_range('2016-01-01', periods=10)
+        tdi = pd.timedelta_range('1', periods=10)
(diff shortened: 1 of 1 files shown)


pandas/tests/frame/test_arithmetic.py
```diff
@@ -100,6 +100,18 @@ def test_df_flex_cmp_constant_return_types_empty(self, opname):
...
+    def test_df_add_td64_columnwise(self):
+        # GH#22534 Check that column-wise addition broadcasts correctly
+        dti = pd.date_range('2016-01-01', periods=10)
+        tdi = pd.timedelta_range('1', periods=10)
+        tser = pd.Series(tdi)
+        df = pd.DataFrame({0: dti, 1: tdi})
+
+        result = df.add(tser, axis=0)
+        expected = pd.DataFrame({0: dti + tdi,
+                                 1: tdi + tdi})
```
gfyoung: I generally prefer that the `expected` be as explicit as possible i.e., write out the actual result of those "sums" instead of the expression.  You can add a comment if needed to describe that we expected `dti + tdi` and `tdi + tdi`.

jreback: I agree with you generally, but these types of tests its ok.

PR #22695, opened 2018-09-13
BUG: Incorrect addition of Week(weekday=6) to DatetimeIndex

### LINKED ISSUE #22465: Change in behavior of DatetimeIndex + Offset
```python
import pandas as pd

offset = pd.tseries.offsets.Week(weekday=6)
idx = pd.DatetimeIndex(['1999-12-26', '2000-05-14']) 

assert (idx[0] + offset) == (idx + offset)[0]
```

bisected to https://github.com/pandas-dev/pandas/pull/18952

Before that the assertion was true, after it's false

```
# before
>>> idx + offset
DatetimeIndex(['2000-01-02', '2000-05-21'], dtype='datetime64[ns]', freq=None)

# after
>>> idx + offset
DatetimeIndex(['1999-12-26', '2000-05-14'], dtype='datetime64[ns]', freq=None)
```

cc @reidy-p is this an expected change due to #18952, or a new bug? If it's intentional, we should document that.

reidy-p:
- [ ] closes #22465
- [ ] tests added / passed
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -629,7 +629,6 @@ Datetimelike
 - Bug in :meth:`DataFrame.eq` comparison against ``NaT`` incorrectly returning ``True`` or ``NaN`` (:issue:`15697`, :issue:`22163`)
 - Bug in :class:`DatetimeIndex` subtraction that incorrectly failed to raise ``OverflowError`` (:issue:`22492`, :issue:`22508`)
 - Bug in :class:`DatetimeIndex` incorrectly allowing indexing with ``Timedelta`` object (:issue:`20464`)
-- Bug in :class:`DatetimeIndex` where addition of Week offset with weekday=6 had no effect (:issue:`22465`)
 
 Timedelta
 ^^^^^^^^^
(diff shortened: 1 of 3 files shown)


doc/source/whatsnew/v0.24.0.txt
```diff
@@ -629,7 +629,7 @@ Datetimelike
 - Bug in :meth:`DataFrame.eq` comparison against ``NaT`` incorrectly returning ``True`` or ``NaN`` (:issue:`15697`, :issue:`22163`)
 - Bug in :class:`DatetimeIndex` subtraction that incorrectly failed to raise ``OverflowError`` (:issue:`22492`, :issue:`22508`)
 - Bug in :class:`DatetimeIndex` incorrectly allowing indexing with ``Timedelta`` object (:issue:`20464`)
--
+- Bug in :class:`DatetimeIndex` where addition of Week offset with weekday=6 had no effect (:issue:`22465`)
```
TomAugspurger: Was this bug present in 0.23.x? I think it wasn't, in which case this entry can be removed as users never saw it.

reidy-p: No I don't think this bug was in 0.23.x because the changes causing this bug are also part of 0.24.0 so I guess we can remove this whatsnew entry

TomAugspurger:
Great. Since CI is passing, we can just remove it when merging.

TomAugspurger (approved):

doc/source/whatsnew/v0.24.0.txt
gfyoung: Tests are passing.  Don't forget to do this!  Either @reidy-p or @TomAugspurger

TomAugspurger:
Thanks!

Issue #22020, opened 2018-07-23
COMPAT: Import abstract classes from collections.abc


holymonson:
> Changed in version 3.3: Moved Collections Abstract Base Classes to the collections.abc module. For backwards compatibility, they continue to be visible in this module through Python 3.7. Subsequently, they will be removed entirely.

Numpy team has done it. https://github.com/numpy/numpy/pull/10743
```python
if PY33:
    import collections.abc as collections_abc
else:
    import collections as collections_abc
```
A problem is `collections_abc` being too long, `isinstance(something, collections_abc.Iterable)` occupies half line, but `from pandas.compat.collections_abc import Iterable` might pollute namespace.

---
BTW, `pandas/compat/__init__.py` seems messy. Since *Much of the code in this module comes from Benjamin Peterson's six library*,  why not require six like many other libraries do?

gfyoung:
@holymonson :

* The `collections_abc` proposal seems reasonable.
* As for `six`, I imagine it's because we only needed that chunk of code for `__init__` and not everything else, though probably someone like @jreback would know more...

jbrockmendel:
re six: it is an implicit dependency because dateutil requires it.  It would simplify some of the intra-pandas dependencies if we got imports from six instead of pandas.compat.

holymonson:
Since six provide some but not all compat things, keeping all import from pandas.compat seems more unified. What I suggest is from six import what's needed to pandas.compat, that will clean up code in pandas.compat.

holymonson:
> Being the last of the 2.x series, 2.7 will have an extended period of maintenance. Specifically, 2.7 will receive bugfix support until January 1, 2020. All 2.7 development work will cease in 2020.

The End Of Life date of Python 2 is 2010-01-01, let's forget six and clean up the code in the future.

TomAugspurger:
We're planning to remove 2.7 support at the end of the year. If somewhat absolutely needs Python 3.8 support before then, they're welcome to update `pandas/compat/__init__.py` to handle this. Otherwise, I'd recommend waiting until January when we remove Python 2 compatibility code, and we can just update the imports to `collections.abc`.

Issue #19677, opened 2018-02-13
CI: Make uncaught warnings raise as errors


TomAugspurger:
By adding

```
filterwarnings =
    error
```

to our `setup.cfg`, we can easily make uncaught warnings turn into errors. What are people's thoughts on this? The only downside I see is that contributor's PR will fail due to unrelated changes to a downstream package, but we can quickly fix those.

It seems like uses of

```python
with warnings.catch_warnings(record=True):
   ...
```

does not work with however pytest is asserting that warnings haven't gotten through. We need to actually use `tm.assert_produces_warning` or `pytest.warns`, so it'll be a bit of work to transition.

Issue #16481, opened 2017-05-24
TST/CLN: Warnings from pytest 3.1+


TomAugspurger:
New versions of pytest print a useful summary of warnings that aren't caught during the test run.
Here's a log from a couple days ago

<details>

```

=============================== warnings summary ===============================
pandas/tests/test_downstream.py::test_dask
  /Users/taugspurger/Envs/pandas-dev/lib/python3.6/site-packages/dask/dataframe/__init__.py:3: DeprecationWarning: pandas.util.decorators is deprecated and will be removed in a future version, import from pandas.util
    from .core import (DataFrame, Series, Index, _Frame, map_partitions,

pandas/tests/test_downstream.py::test_xarray
  /Users/taugspurger/Envs/pandas-dev/lib/python3.6/site-packages/xarray/core/formatting.py:16: FutureWarning: The pandas.tslib module is deprecated and will be removed in a future version.
    from pandas.tslib import OutOfBoundsDatetime

pandas/tests/test_downstream.py::test_statsmodels
  /Users/taugspurger/Envs/pandas-dev/lib/python3.6/importlib/_bootstrap.py:205: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
    return f(*args, **kwds)
  /Users/taugspurger/Envs/pandas-dev/lib/python3.6/site-packages/statsmodels/statsmodels/compat/pandas.py:62: FutureWarning: The pandas.core.datetools module is deprecated and will be removed in a future version. Please use the pandas.tseries module instead.
    from pandas.core import datetools
  /Users/taugspurger/Envs/pandas-dev/lib/python3.6/site-packages/patsy/util.py:652: DeprecationWarning: pandas.core.common.is_categorical_dtype is deprecated. import from the public API: pandas.api.types.is_categorical_dtype instead
    return safe_is_pandas_categorical_dtype(data.dtype)

pandas/tests/test_downstream.py::test_scikit_learn
  /Users/taugspurger/Envs/pandas-dev/lib/python3.6/importlib/_bootstrap.py:205: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
    return f(*args, **kwds)
  /Users/taugspurger/Envs/pandas-dev/lib/python3.6/importlib/_bootstrap.py:205: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
    return f(*args, **kwds)

[...]

PR #22699, opened 2018-09-13
TST: Fail on warning

TomAugspurger:
Sets our pytest config to fail on unhandled warnings.

Fixes a bunch of tests to not fail.

This still has a few TODOs.

Closes https://github.com/pandas-dev/pandas/issues/16481
Closes https://github.com/pandas-dev/pandas/issues/19677
Closes https://github.com/pandas-dev/pandas/issues/13962
Closes https://github.com/pandas-dev/pandas/issues/22020


h-vetinari:
If it turns *all* warnings into errors, it might close #13962 as well.

TomAugspurger:
@rth thanks for the context, that's helpful. I think that elevating warnings to errors only on worker with the latest dependencies makes the most sense. I'll probably keep at this until things are passing, and then limit the elevation once things are passing.

> If it turns all warnings into errors, it might close #13962 as well.

Thanks for the link. It should close that.

pandas/conftest.py
```diff
@@ -31,6 +32,16 @@ def pytest_addoption(parser):
                      help="Fail if a test is skipped for missing data file.")
 
 
+def pytest_collection_modifyitems(items):
```
TomAugspurger: I don't think we'll want this permanently. This is just to track down #22675 in case it happens on this branch.

TomAugspurger:
There's a new warning coming through from numpy dev.

```
In [9]: pd.DataFrame(x).median(1)
/Users/taugspurger/miniconda3/envs/travis-37-numpydev/lib/python3.7/site-packages/numpy/lib/function_base.py:3267: RuntimeWarning: All-NaN slice encountered
  r = func(a, **kwargs)
Out[9]:
0    -0.601975
1    -0.572101
2     0.405565
3     0.848596
4     0.856747
5          NaN
6          NaN
7          NaN
8          NaN
9          NaN
10    0.333711
11    0.115346
12    0.193118
...
````

I assume the user should not see that warning?

File: pandas/compat/__init__.py [hunk 2]
```diff
@@ -38,6 +38,7 @@
 import struct
 import inspect
 from collections import namedtuple
+import collections
```
TomAugspurger: Using ABCs from `collections` is deprecated. Need to use `collections.abc` instead (which isn't in Py2).

pandas/core/arrays/integer.py
```diff
@@ -147,6 +147,11 @@ def coerce_to_array(values, dtype, mask=None, copy=False):
             dtype = values.dtype
 
     if dtype is not None:
+        if (isinstance(dtype, string_types) and
+                (dtype.startswith("Int") or dtype.startswith("UInt"))):
+            # Avoid DeprecationWarning from NumPy about np.dtype("Int64")
+            # https://github.com/numpy/numpy/pull/7476
+            dtype = dtype.lower()
```
TomAugspurger: @jreback could you take a close look here. I think you ran into this when writing a test.

TomAugspurger (commented):
FYI @jreback, @jbrockmendel, @jorisvandenbossche, @gfyoung, @mroeschke, @toobaz, @jschendel, @datapythonista, @chris-b1, and other reviewers I'm forgetting, the main thing we want to avoid doing in tests / code is

```python
with warnings.catch_warnings(record=True):
    do_operation_that_maybe_has_warning()
```

While say `record=True` does prevent the warning from going to stdout, it still ends up *somewhere* that python / pytest considers having happened, so if you're in `-W error` that will be an error.

The correct way to do it in code is

```python
with warnings.catch_warnings():
    warnings.simplefilter(...)
    # or warnings.filterwarnings(...)
    do_operation_that_maybe_has_warning()
```

and in tests, I think we should recommend using

1. `tm.assert_produces_warning` for assert that a warning was emitted
2. `@pytest.mark.filterwarnings(...)` for ignoring warnings we don't care about.

If a test *really* needs to (e.g. because it needs different warnings handling for different parts of the same test), it can use the `warnings.catch_warnings` and a `warningsfilter` directly within the test.

---

Currently, the PR turns all underhanded warnings into errors on our numpydev build (which I've updated to 3.7, to catch warnings from the latest Python). This will occasionally cause unrelated failures on PRs and NumPy / dateutil add warnings. I think we should see how that works in practice for a bit, and revisit the policy if things break too often.

doc/source/contributing.rst
```diff
@@ -859,6 +859,32 @@ preferred if the inputs or logic are simple, with Hypothesis tests reserved
...
+
+By default, pandas test suite will fail if any unhandled warnings are emitted.
+
+If your change involves checking that a warning is actually emitted, use
+``tm.assert_produces_warning(ExpectedWarning)``. We prefer this to pytest's
+``pytest.warns`` context manager because ours checks that the warning's stacklevel
+is set correctly.
+
+If you have a test that would emit a warning, but you aren't actually testing the
+warning it self (say because it's going to be removed in the future, or because we're
```
gfyoung: it self &rarr; itself

TomAugspurger:
All green, aside from the doc build timeout.

Are people generally OK with turning unhandled warnings into errors on the dev build? This will be warnings from

1. Python 3.7 (shouldn't be any new ones)
2. NumPy & python-dateutil dev
3. (most importantly) us

Does anyone expect to have a chance to review this today? The fact that I haven't seen a ResourceWarning turning into an exception suggests that we *may* have fixed that random failure on this branch. It's a lot of small changes, but hopefully harmless.

jorisvandenbossche (commented):
Thanks for this huge effort!

Should we add

```
filterwarnings =
    always
```

to print all warnings always? (to the extent this is equivalent to `-W always`?)

Not needed for this PR, but do we want to make a `with tm.ignore_warnings(DeprecationWarning): ...` utility context manager to replace the two-line `with catch_warnings: simplefiltter("ignore", DeprecationWarning)` ?

jorisvandenbossche:
> Are people generally OK with turning unhandled warnings into errors on the dev build? This will be warnings from

So the errors will only happen at the numpy dev build (which is in allowed failures) ?

Is there some way to let the warnings from pandas itself to error always? (I think those would be fine to error on all builds, not only numpy-dev)

TomAugspurger:
> So the errors will only happen at the numpy dev build (which is in allowed failures) ?

Yes, but I thought our policy was "allowed failures aren't allowed", but I've never understood why it's in the allowed failures section.

> Is there some way to let the warnings from pandas itself to error always?

Not that I know of.

pandas/core/arrays/datetimelike.py
```diff
@@ -59,6 +59,7 @@ def cmp_method(self, other):
         # numpy will show a DeprecationWarning on invalid elementwise
         # comparisons, this will raise in the future
         with warnings.catch_warnings(record=True):
+            warnings.filterwarnings("ignore", "elementwise", FutureWarning)
```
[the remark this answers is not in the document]
TomAugspurger: Added to https://github.com/pandas-dev/pandas/issues/22698

pandas/conftest.py
WillAyd: Do we need this as part of the ultimate change or no?

pandas/core/arrays/integer.py
TomAugspurger: This is a deprecation warning from NumPy right now, which is filtered interactive, but pytest removes that filter by default. Some of the `integer_array` tests were hitting this.

TomAugspurger:
> but do we want to make a with tm.ignore_warnings(DeprecationWarning): ... utility context manager to replace the two-line with catch_warnings: simplefiltter("ignore", DeprecationWarning) ?

I hope that we wont really need `catch_warnings` going forward. Ideally tests are small / focused enough that a `pytest.mark.filterwarnings` is appropriate.

pandas/core/arrays/integer.py
```diff
@@ -508,6 +513,7 @@ def cmp_method(self, other):
             # numpy will show a DeprecationWarning on invalid elementwise
             # comparisons, this will raise in the future
             with warnings.catch_warnings(record=True):
```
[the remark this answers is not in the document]
TomAugspurger: No, record isn't needed here (and should be avoided). It should only be used when you need to make some assertion about the value yielded by the contextmanager.

pandas/compat/__init__.py
WillAyd: Assuming the subsequent changes didn't generate any failures does this point to a potential lack of compatibility test coverage?

pandas/core/indexes/base.py
```diff
@@ -98,6 +98,7 @@ def cmp_method(self, other):
             # numpy will show a DeprecationWarning on invalid elementwise
             # comparisons, this will raise in the future
             with warnings.catch_warnings(record=True):
+                warnings.filterwarnings("ignore", "elementwise", FutureWarning)
```
WillAyd: What's the point in using `warnings.filterwarnings` instead of `pytest.mark.filterwarnings`? Wondering if we shouldn't always use the latter for consistency

pandas/tests/frame/test_indexing.py
```diff
@@ -364,6 +364,7 @@ def test_getitem_ix_mixed_integer(self):
         assert_frame_equal(result, expected)
 
         with catch_warnings(record=True):
+            simplefilter("ignore", DeprecationWarning)
```
WillAyd: Same comment as before but just wonder if perhaps the `pytest.mark.filterwarnings` decorator could give flexibility to apply at function level to minimize diff

pandas/tests/indexes/timedeltas/test_ops.py
```diff
@@ -334,7 +334,7 @@ def test_freq_setter_errors(self):
             idx.freq = '5D'
 
         # setting with a non-fixed frequency
-        msg = '<2 \* BusinessDays> is a non-fixed frequency'
+        msg = r'<2 \* BusinessDays> is a non-fixed frequency'
```
WillAyd: Hmm what failure was this causing? Change certainly looks correct just not sure how it's related to warnings

WillAyd (commented):
Didn't make it through the entire diff yet but some comments so far which may be helpful

pandas/tests/extension/base/dtype.py
```diff
@@ -67,7 +69,11 @@ def test_check_dtype(self, data):
             expected = pd.Series([True, True, False, False],
                                  index=list('ABCD'))
 
-        result = df.dtypes == str(dtype)
+        with warnings.catch_warnings():
+            # XXX: This should probably be *fixed* not ignored.
+            # See libops.scalar_compare
+            warnings.simplefilter("ignore", DeprecationWarning)
+            result = df.dtypes == str(dtype)
```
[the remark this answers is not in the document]
TomAugspurger: The problem is in https://github.com/pandas-dev/pandas/blob/8a1c8ad4bc9d152db1e4b8d2f304bc5077392a2d/pandas/_libs/ops.pyx#L84-L87

With `-W error`, a TypeError is raised on `np.dtype("Int64")`. I'm not really sure how best to fix that right now...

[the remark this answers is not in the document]
TomAugspurger: To clarify, the `np.dtype("Int64")` is (I assume) called as part of `np.dtype.__eq__`.

pandas/tests/indexes/datetimes/test_tools.py
```diff
@@ -1587,12 +1589,15 @@ def units_from_epochs():
...
 
 
-@pytest.fixture(params=[epoch_1960(),
-                        epoch_1960().to_pydatetime(),
-                        epoch_1960().to_datetime64(),
-                        str(epoch_1960())])
-def epochs(request):
-    return request.param
+@pytest.fixture(params=['timestamp', 'pydatetime', 'datetime64'])
+def epochs(epoch_1960, request):
```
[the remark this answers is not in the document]
TomAugspurger: The motivation for changing was pytest's deprecation of calling fixtures (we called them in the `@pytest.fixture` up above.

Removing the string was an accident, thanks for catching. Restored.

pandas/tests/indexes/timedeltas/test_ops.py
TomAugspurger: Python 3.something (6ish?) added DeprecationWarnings for invalid escape sequences in strings.

```
bash-4.4$ cat bad.py
# file: bad.py
x = "<2 \* Day>"
bash-4.4$ python3 -W error bad.py
  File "bad.py", line 2
    x = "<2 \* Day>"
       ^
SyntaxError: invalid escape sequence \*
```

pandas/tests/indexing/test_floats.py
```diff
@@ -57,6 +60,7 @@ def f():
                 s.iloc[3.0] = 0
             pytest.raises(TypeError, f)
 
+    @ignore_ix
```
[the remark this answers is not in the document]
TomAugspurger: Yes (laziness 😄).

pandas/tests/series/indexing/test_indexing.py
```diff
@@ -390,6 +390,8 @@ def test_setslice(test_data):
     assert sl.index.is_unique
 
 
+# FutureWarning from NumPy about [slice(None, 5).
+@pytest.mark.filterwarnings("ignore:Using a non-tuple:FutureWarning")
```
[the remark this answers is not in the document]
TomAugspurger: I think so... but not sure. My thinking was we're taking user-providing input and passing it straight through to numpy, so we want to directly match NumPy's behavior. *We* don't want to filter the warning in pandas, we want users to see it if they're providing bad input. So we only ignore it in the test.

pandas/tests/test_downstream.py
```diff
@@ -104,13 +110,17 @@ def test_pandas_datareader():
         'F', 'quandl', '2017-01-01', '2017-02-01')
 
 
+# importing from pandas, Cython mport warning
```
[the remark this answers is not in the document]
TomAugspurger: Yes (later ones I added have a reference to the issue).

> Should we have an issue to make sure to remove those filters again at some point?

Maybe, but I think that's largely out of our control. IIUC, it would require a release from each of those libraries with their extensions compiled by a new enough Cython.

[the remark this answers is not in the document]
TomAugspurger: > we should not generally ignore all warnings here

I'm restricting these now, but we may want to ignore all these warnings. Fixing e.g. geopandas required

1. ignore Cython warning
2. ignore geopandas importing pandas.util.decorators (shall  I open an issue?)
3. ignore fiona using `Logger.warn` instead of `Logger.warning`

It seems like we're exposing ourself to a lot of spurious failures if we don't filter these heavily.

pandas/conftest.py
TomAugspurger: I'm tempted to leave it in master for a week or two to see if we catch any of the failures in action. I've made myself a reminder to remove it.

pandas/compat/__init__.py
TomAugspurger: I don't follow 100%, but I don't think so. On 3.7, you basically couldn't import pandas when `-W error:DeprecationWarning` is set, so this was definitely being hit..

pandas/core/indexes/base.py
TomAugspurger: laziness / not wanting to introduce copy-paste errors. New code should essentially always use marks, and I can make the changes here if you want. I mostly cleaned up Panel to remove these `catch_warnings` / `filterwarnings`, but introduced one copy-paste error in the process (caught before pushing I think), so I'm inclined to not make unnecessary changes.

TomAugspurger:
I wasn't able to get the suggestion of 

```filterwarnings =
error:::pandas[.*]
```

to work reliably. In my small test of

```
def test_warns():
    pd.Panel()
    pd.DataFrame({"A": [1, 2]}).ix[0]
```

the panel warning was not elevated to an error, but the `.ix` was. I didn't investigate too deeply.

pandas/core/indexes/base.py
WillAyd: Makes sense. Maybe not a hard rule on this change but it would be helpful to use this instead in functions where there is a big diff

pandas/tests/indexes/timedeltas/test_ops.py
WillAyd: Was CI not hitting this before this change?

TomAugspurger: It was, e.g. https://travis-ci.org/pandas-dev/pandas/jobs/429647889#L2261

pandas/conftest.py
TomAugspurger: sighhhhhhhhhhhhhhhhhhhhhhhhhhhh. we got one: https://travis-ci.org/pandas-dev/pandas/jobs/429657921#L2280, but that doesn't actually fail the right test, because the warning is in the file object's `__del__`, just prints out to stderr rather than raising an exception (to fail the test).

I'm going to do some debugging on the branch.

TomAugspurger:
Those interested in the ResourceWarning failures may want to take a look at https://github.com/pandas-dev/pandas/pull/22699/commits/cfb5ae2d4ac8b24b871da69f24e09960d8868cba. I tried to squash all the places we may have had unclosed files sitting around. If I had to guess, either https://github.com/pandas-dev/pandas/pull/22699/commits/cfb5ae2d4ac8b24b871da69f24e09960d8868cba#diff-0d7b5a2c72b4dfc11d80afe159d45ff8R389 or https://github.com/pandas-dev/pandas/pull/22699/commits/cfb5ae2d4ac8b24b871da69f24e09960d8868cba#diff-9d7323cc6b0cfa3f4482bc8d4268966dR213 is going to be the one that fixes it, but who knows.

pandas/tests/series/indexing/test_datetime.py
```diff
@@ -383,6 +383,8 @@ def test_getitem_setitem_periodindex():
     assert_series_equal(result, ts)
 
 
+# FutureWarning from NumPy.
+@pytest.mark.filterwarnings("ignore:Using a non-tuple:FutureWarning")
```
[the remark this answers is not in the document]
TomAugspurger: Yes, I think so. AFAICT, it's equivalent to `s.values[indexer]`, so we should match that behavior up until NumPy enforces the deprecation warning.

TomAugspurger:
I think I've addressed all the inline comments

Sorry to push on this relatively quickly, but I want to get our CI in a healthy state again. If possible, I'd like to merge this today.

I currently have two followup issues for warnings that need some care

1. https://github.com/pandas-dev/pandas/issues/22698 (Series ctor)
2. https://github.com/pandas-dev/pandas/issues/22712 (median with all-NA slice)

@jreback can you clarify our current policy on "allowed failures"? I don't recall why they're in the allowed-failures section.

doc/source/contributing.rst
```diff
@@ -859,6 +867,55 @@ preferred if the inputs or logic are simple, with Hypothesis tests reserved
...
+If your change involves checking that a warning is actually emitted, use
+``tm.assert_produces_warning(ExpectedWarning)``.
+
+
+.. code-block:: python
+
+   with tm.assert_prodcues_warning(FutureWarning):
+       df.some_operation()
+
+We prefer this to the ``pytest.warns`` context manager because ours checks that the warning's
```
jreback: you may want to mention that we will actually fail if pytest.warns will fail the linter :>

pandas/core/arrays/integer.py
jreback: oh I didn't know numpy did this:

```
In [1]: np.dtype('Int64')
/Users/jreback/miniconda3/envs/pandas/bin/ipython:1: DeprecationWarning: Numeric-style type codes are deprecated and will result in an error in the future.
  #!/Users/jreback/miniconda3/envs/pandas/bin/python
Out[1]: dtype('int64')
```

I guess so, thanks

pandas/core/groupby/generic.py
```diff
@@ -758,7 +758,7 @@ def aggregate(self, func_or_funcs, *args, **kwargs):
         if isinstance(func_or_funcs, compat.string_types):
             return getattr(self, func_or_funcs)(*args, **kwargs)
 
-        if isinstance(func_or_funcs, collections.Iterable):
+        if isinstance(func_or_funcs, compat.Iterable):
```
jreback: in theory we could add a lint rule to avoid using ``collection.Iterable``, and instead just use our compat

pandas/core/internals/blocks.py
```diff
@@ -3490,6 +3490,7 @@ def _putmask_smart(v, m, n):
 
         # we ignore ComplexWarning here
         with warnings.catch_warnings(record=True):
+            warnings.simplefilter("ignore", np.ComplexWarning)
```
jreback: does the context manager restore all of the warnings filters after? (I guess that is the point?)

pandas/tests/api/test_api.py
```diff
@@ -175,23 +174,23 @@ def test_get_store(self):
 
 class TestParser(object):
 
+    @pytest.mark.filterwarnings("ignore")
     def test_deprecation_access_func(self):
```
jreback: should these actually be ``tm.assert_produces_warning(FutureWarning)``?

jreback: nvm I see your comment below

pandas/tests/frame/test_indexing.py
```diff
@@ -1134,19 +1175,22 @@ def test_ix_frame_align(self):
         df = df_orig.copy()
 
         with catch_warnings(record=True):
+            simplefilter("ignore", DeprecationWarning)
```
jreback: grr....we should just remove all of this .ix bizness

pandas/tests/groupby/test_groupby.py
```diff
@@ -508,30 +507,30 @@ def test_frame_multi_key_function_list():
```
jreback: we could prob split the panel tests to a separate file (and just ignore warnings there), maybe make an issue

pandas/tests/indexes/datetimes/test_tools.py
```diff
@@ -1587,12 +1589,18 @@ def units_from_epochs():
...
 
 
-@pytest.fixture(params=[epoch_1960(),
-                        epoch_1960().to_pydatetime(),
-                        epoch_1960().to_datetime64(),
-                        str(epoch_1960())])
-def epochs(request):
-    return request.param
+@pytest.fixture(params=['timestamp', 'pydatetime', 'datetime64', 'str_1960'])
+def epochs(epoch_1960, request):
```
jreback: can you add a doc-string

pandas/tests/indexing/test_floats.py
```diff
@@ -10,6 +10,9 @@
 import pandas.util.testing as tm
 
 
+ignore_ix = pytest.mark.filterwarnings("ignore:\\n.ix:DeprecationWarning")
```
jreback: maybe move this to the top-level conftest and use it above (a number of files where catching all of the ix warnings)

pandas/tests/indexing/test_iloc.py
```diff
@@ -388,52 +388,61 @@ def test_iloc_getitem_frame(self):
 
         result = df.iloc[2]
         with catch_warnings(record=True):
+            filterwarnings("ignore", "\\n.ix", DeprecationWarning)
```
jreback: use ignore_ix?

jreback (changes requested):
@TomAugspurger awesome effort here. This was for sure a sore point. Some minor comments.

jreback:
> @jreback can you clarify our current policy on "allowed failures"? I don't recall why they're in the allowed-failures section.

- so numpy-dev is allowed to fail because there may be a change which breaks us but is not pushed to public (ideally would fail, except for when numpy / dateutil breaks things *everything* breaks and all builds would then fail)
- slow tests are there just to make the build green faster
- clipboard test uses to fail pretty unexpectedly
- doc build takes a long time.

so they are really there to get a probable green faster. I don't really have a problem moving them to required builds actually.

pandas/core/arrays/integer.py
TomAugspurger: Yeah. I'm waiting to hear back on https://github.com/numpy/numpy/pull/7476#issuecomment-422358051

If `np.dtype("int") == "Int64"` errors in the future, we may have to rework some things.

pandas/core/groupby/generic.py
TomAugspurger: I think we're close enough to py3-only that we'll be OK. Writing that regex would be a bit fiddly since we'd need to enumerate all the ABCs.

pandas/core/internals/blocks.py
TomAugspurger: Yeah, that's the idea. Ideally libraries shouldn't modify the global warnings registry without good cause.

pandas/tests/indexing/test_floats.py
TomAugspurger: Hmm do you know how to use marks defined in the top-level `conftest.py`? I'm not actually sure that's possible.

pandas/tests/indexing/test_iloc.py
TomAugspurger: That filters the warning for the entire test. I was trying to exactly match the tests previous intent, of only filtering within the `catch_warnings` blocks.

pandas/tests/indexing/test_floats.py
jreback: i think you can just import it

TomAugspurger:
Thanks. So I think it's best to treat numpy / dateutil dev as "allowed failures" but as soon as we see a failure we should open an issue that would block the next release (and maybe report upstream if it looks like an unintentional break).

We might consider moving the doc build to azure. Apparently the workers more powerful than Travis.

pandas/core/groupby/generic.py
jreback: true, certainly a follow up issue is ok

jreback:
> Thanks. So I think it's best to treat numpy / dateutil dev as "allowed failures" but as soon as we see a failure we should open an issue that would block the next release (and maybe report upstream if it looks like an unintentional break).

yep, that's the idea

> We might consider moving the doc build to azure. Apparently the workers more powerful than Travis.

certainly

jreback (approved):

jreback:
@TomAugspurger merge when you are ready

TomAugspurger:
👍 pushed some doc fixes. Will merge on green.

TomAugspurger:
Travis all passed. Merging.

PR #22700, opened 2018-09-13
Remove --strict-data-files on pytest by default

### LINKED ISSUE #22683: missing files on test
Hi, testing with "pytest pandas" from a pypi package returns a bunch of errors related to missing files.
I understand that those files are removed on propose ?
Shouldn't the tests skip the use of those files if they aren't present ?

marianielias:
- closes #22683 
Explanation:
pytest gets the configuration from setup.cfg, witch forces the option "--strict-data-files" by default. This makes the test to fail on missing files that are not packaged (on pypi and in the github tarball).
Removing this line fixes the issue and the tests are skipped.

Adding missing newline also...

Cheers.
Elias.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -32,5 +32,4 @@ markers =
     slow: mark a test as slow
     network: mark a test as network
     high_memory: mark a test as a high-memory only
-addopts = --strict-data-files
 doctest_optionflags= NORMALIZE_WHITESPACE IGNORE_EXCEPTION_DETAIL


TomAugspurger:
We wan this to be the default.

Should we instead remove setup.cfg from the distributions?

FlingeR:
> We want strict to be the default.
> 
> Should we instead remove setup.cfg from the distributions?

Assuming that the file is not being used by something else... LGTM.
Let me test without the file just in case.

Issue #8508, opened 2014-10-08
Allow writing to S3 paths


hunterowens:
It would be really great if `to_(filetype)` supported writing to S3. 

Here is an example upload to s3 function that takes in a local file and places it on a s3 bucket.

``` python
def upload_to_s3(local_file_path, file_name, bucket_name, s3_directory):
    """
    Returns
    ----------
    Uploads local file to appropriate s3 key, and prints status 

    Parameters
    ----------
    local_file_path : str
        ex. 'my/local/path'
    file_name : str
        ex. 'cleaned_data.csv'
    bucket_name : str
        ex. 'dsapp-edu-data'
    s3_directory : str
        ex. 'NC-Cabarrus/cleaned_data'
    """

    def percent_cb(complete, total):
        """
        Helper function that prints progress
        """
        sys.stdout.write('.')
        sys.stdout.flush()

    conn = boto.connect_s3()
    bucket = conn.get_bucket(bucket_name)
    full_key_name = os.path.join(s3_directory, file_name)
    k = bucket.new_key(full_key_name)
    full_filepath = os.path.join(local_file_path, file_name)
    k.set_contents_from_filename(full_filepath, cb=percent_cb, num_cb=10)

    return None 
```

andrewgiessel:
any PRs open on this?  I thought of this too and think it'd be great.  I can take a shot at it if not.

TomAugspurger:
Go for it!

TomAugspurger:
https://github.com/wrobstory/pgshift/blob/master/pgshift/pgshift.py#L99 could be a helpful template

swarajban:
I also wanted the ability to write a DataFrame as a CSV to S3 (acronym overload...), and wrote the following snippet:

```
# Write dataframe to buffer
csv_buffer = StringIO()
df.to_csv(csv_buffer, index=False)

# Upload CSV to S3
s3_key = 'test.csv'
s3_resource = aws_session.resource('s3')
s3_resource.Object(s3_bucket, s3_key).put(Body=csv_buffer.getvalue())
```

Hope this helps anyone trying to do the same thing. IMO, this features shouldn't be added to pandas. I think this snippet (or a better version) should simply be documented

maximveksler:
Any hope of supporting writing to S3 for the new release? Now that [parquet is supported](https://pandas.pydata.org/pandas-docs/stable/io.html#io-parquet) this becomes double interesting.

jreback:
[...]

PR #22704, opened 2018-09-14
Support writing CSV to GCS

### LINKED ISSUE #23094: Allow writing to GCS paths
This is quite a bit simpler than #8508 since `gcsfs` allows for writing text-type blobs so I think it makes sense to track separately.

bnaul:
- [x] fixes #23094 
- [x] tests added / passed
- [x] whatsnew entry

This seems to work as-is and doesn't break any of the IO tests; as I mentioned in https://github.com/pandas-dev/pandas/issues/8508#issuecomment-421184687 getting S3 to work is a little more complicated but maybe still not bad. But this would be a step in the right direction regardless.

cc @TomAugspurger

### DIFF
--- a/pandas/tests/io/test_gcs.py
+++ b/pandas/tests/io/test_gcs.py
@@ -27,6 +27,21 @@ def test_read_csv_gcs(mock):
 
 
 @td.skip_if_no('gcsfs')
+def test_to_csv_gcs(mock):
+    df1 = DataFrame({'int': [1, 3], 'float': [2.0, np.nan], 'str': ['t', 's'],
+                     'dt': date_range('2018-06-18', periods=2)})
+    with mock.patch('gcsfs.GCSFileSystem') as MockFileSystem:
+        s = StringIO()
+        instance = MockFileSystem.return_value
+        instance.open.return_value = s
+
+        df1.to_csv('gs://test/test.csv', index=True)
+        df2 = read_csv(StringIO(s.getvalue()), parse_dates=['dt'], index_col=0)
+
+    assert_frame_equal(df1, df2)
+
+
+@td.skip_if_no('gcsfs')
 def test_gcs_get_filepath_or_buffer(mock):
     df1 = DataFrame({'int': [1, 3], 'float': [2.0, np.nan], 'str': ['t', 's'],
                      'dt': date_range('2018-06-18', periods=2)})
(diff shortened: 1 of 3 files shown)


pandas/tests/io/test_gcs.py
```diff
@@ -26,6 +26,21 @@ def test_read_csv_gcs(mock):
     assert_frame_equal(df1, df2)
 
 
+@td.skip_if_no('gcsfs')
+def test_to_csv_gcs(mock):
+    df1 = DataFrame({'int': [1, 3], 'float': [2.0, np.nan], 'str': ['t', 's'],
+                     'dt': date_range('2018-06-18', periods=2)})
+    with mock.patch('gcsfs.GCSFileSystem') as MockFileSystem:
```
WillAyd: Might be missing the point but if you are patching this what is actually getting tested for gcs?

bnaul: Yeah, it's kind of the same problem that we discussed in #20729. This does at least test the logic that I touched here; I think all the mocks assume is that `gcsfs.GCSFileSystem` can `read`/`write` strings and everything else is using the real `pandas` methods.

pandas/tests/io/test_gcs.py
```diff
@@ -26,6 +26,21 @@ def test_read_csv_gcs(mock):
...
+        s = StringIO()
+        instance = MockFileSystem.return_value
+        instance.open.return_value = s
+
+        df1.to_csv('gs://test/test.csv', index=True)
```
WillAyd: Any particular reason you are explicitly stating `index=True` here instead of using the default?

pandas/tests/io/test_gcs.py
```diff
@@ -26,6 +26,21 @@ def test_read_csv_gcs(mock):
...
+        df2 = read_csv(StringIO(s.getvalue()), parse_dates=['dt'], index_col=0)
```
WillAyd: Related to above comment

WillAyd (commented):
Thanks for the clarification on the mock

WillAyd:
Could also use a related issue and whatsnew note for v0.24

TomAugspurger:
When reading S3, we have to wrap it in a TextIOWrapper. Do we need to do the same for writing?

pandas/tests/io/test_gcs.py
bnaul: `df.to_csv(f)` and `pd.read_csv(f)` handle the index differently so I wanted to be extra clear that the index is also being checked in the round tripping

pandas/tests/io/test_gcs.py
bnaul: See above

bnaul:
Hi @WillAyd @TomAugspurger,

- Added a whatsnew/related issue
- The `TextIOWrapper` thing would indeed be needed for `s3fs` for but `gcsfs` it isn't; I'd like to keep this small so for now I'm going to take care of the easier case first

If anyone else wants to take a look at this I would be grateful since touching anything related to `to_csv` makes me a bit nervous 😬

TomAugspurger (approved):
Feel free to merge if you're comfortable @WillAyd.

WillAyd (approved):

5amfung:
@bnaul Looks like there's no documentation on how to use this feature.  Hate to see this being implemented but no one is aware of it at all.

TomAugspurger:
@5amfung can you submit a PR?

PR #22710, opened 2018-09-14
BUG: Empty CategoricalIndex fails with boolean categories

### LINKED ISSUE #22702: ValueError in Categorical Constructor with empty data and boolean categories
This works,

```python
In [15]: pd.Categorical([], categories=['a', 'b'])
Out[15]: [], Categories (2, object): [a, b]
```

This doesn't

```pytb
In [16]: pd.Categorical([], categories=[True, False])
<ipython-input-16-8e79cd310199> in <module>()
----> 1 pd.Categorical([], categories=[True, False])

~/sandbox/pandas/pandas/core/arrays/categorical.py in __init__(self, values, categories, ordered, dtype, fastpath)
    426
    427         else:
--> 428             codes = _get_codes_for_values(values, dtype.categories)
    429
    430         if null_mask.any():

~/sandbox/pandas/pandas/core/arrays/categorical.py in _get_codes_for_values(values, categories)
   2449     (_, _), cats = _get_data_algo(categories, _hashtables)
   2450     t = hash_klass(len(cats))
-> 2451     t.map_locations(cats)
   2452     return coerce_indexer_dtype(t.lookup(vals), cats)
   2453

~/sandbox/pandas/pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.PyObjectHashTable.map_locations()
   1330             raise KeyError(key)
   1331
-> 1332     def map_locations(self, ndarray[object] values):
   1333         cdef:
   1334             Py_ssize_t i, n = len(values)

ValueError: Buffer dtype mismatch, expected 'Python object' but got 'unsigned long'
```

the `values` there is `array([], dtype=object)`. It should be int dtype by this point.

pganssle:
Fixes #22702.

This bug was introduced in [7818486859d1aba53](https://github.com/pandas-dev/pandas/commit/7818486859d1aba53ce359b93cfc772e688958e5); per [my comment](https://github.com/pandas-dev/pandas/issues/22702#issuecomment-421364087), the problem is here:

```python
    if not is_dtype_equal(values.dtype, categories.dtype):
        values = ensure_object(values)
        categories = ensure_object(categories)

    (hash_klass, vec_klass), vals = _get_data_algo(values, _hashtables)
    (_, _), cats = _get_data_algo(categories, _hashtables)
```

When `categories` is `Index([True], dtype='object')` and `values` is `array([], dtype='object')`, the `ensure_object` call is bypassed, but in `_get_data_algo`, an `Index` consisting entirely of boolean objects will be coerced to `uint64`, which violates the assumption that `values` and `categories` have the same dtype.

I felt that retrieving the underlying numpy arrays (if any exist) is the safest way to handle this without having too many wide-reaching effects across the rest of the codebase, but there might be a better way to enforce that these are not coerced into different data types.

- [x] closes #xxxx
- [x] tests added 
- [x] whatsnew entry

### DIFF
--- a/doc/source/whatsnew/v0.24.0.txt
+++ b/doc/source/whatsnew/v0.24.0.txt
@@ -616,6 +616,7 @@ Categorical
 ^^^^^^^^^^^
 
 - Bug in :meth:`Categorical.from_codes` where ``NaN`` values in ``codes`` were silently converted to ``0`` (:issue:`21767`). In the future this will raise a ``ValueError``. Also changes the behavior of ``.from_codes([1.1, 2.0])``.
+- Constructing a :class:`pd.CategoricalIndex` with empty values and boolean categories was raising a ``ValueError`` after a change to dtype coercion (:issue:`22702`).
 
 Datetimelike
 ^^^^^^^^^^^^
--- a/pandas/core/arrays/categorical.py
+++ b/pandas/core/arrays/categorical.py
@@ -2439,9 +2439,13 @@ def _get_codes_for_values(values, categories):
     """
     utility routine to turn values into codes given the specified categories
     """
-
     from pandas.core.algorithms import _get_data_algo, _hashtables
-    if not is_dtype_equal(values.dtype, categories.dtype):
+    if is_dtype_equal(values.dtype, categories.dtype):
+        # To prevent erroneous dtype coercion in _get_data_algo, retrieve
+        # the underlying numpy array. gh-22702
+        values = getattr(values, 'values', values)
+        categories = getattr(categories, 'values', categories)
+    else:
         values = ensure_object(values)
         categories = ensure_object(categories)
 
--- a/pandas/tests/arrays/categorical/test_constructors.py
+++ b/pandas/tests/arrays/categorical/test_constructors.py
@@ -42,7 +42,6 @@ class TestCategoricalConstructors(object):
         expected = pd.Int64Index([1, 2, 3])
         tm.assert_index_equal(c.categories, expected)
 
-    @pytest.mark.xfail
     def test_constructor_empty_boolean(self):
         # see gh-22702
         cat = pd.Categorical([], categories=[True, False])
--- a/pandas/tests/arrays/categorical/test_constructors.py
+++ b/pandas/tests/arrays/categorical/test_constructors.py
@@ -42,6 +42,13 @@ class TestCategoricalConstructors(object):
         expected = pd.Int64Index([1, 2, 3])
         tm.assert_index_equal(c.categories, expected)
 
+    @pytest.mark.xfail
+    def test_constructor_empty_boolean(self):
+        # see gh-22702
+        cat = pd.Categorical([], categories=[True, False])
+        categories = sorted(cat.categories.tolist())
+        assert categories == [False, True]
+
     def test_constructor_tuples(self):
         values = np.array([(1,), (1, 2), (1,), (1, 2)], dtype=object)
         result = Categorical(values)
--- a/pandas/tests/indexes/test_category.py
+++ b/pandas/tests/indexes/test_category.py
@@ -136,6 +136,13 @@ class TestCategoricalIndex(Base):
         result = CategoricalIndex(idx, categories=idx, ordered=True)
         tm.assert_index_equal(result, expected, exact=True)
 
+    @pytest.mark.xfail
+    def test_construction_empty_with_bool_categories(self):
+        # see gh-22702
+        cat = pd.CategoricalIndex([], categories=[True, False])
+        categories = sorted(cat.categories.tolist())
+        assert categories == [False, True]
+
     def test_construction_with_categorical_dtype(self):
         # construction with CategoricalDtype
         # GH18109
--- a/pandas/tests/indexes/test_category.py
+++ b/pandas/tests/indexes/test_category.py
@@ -136,7 +136,6 @@ class TestCategoricalIndex(Base):
         result = CategoricalIndex(idx, categories=idx, ordered=True)
         tm.assert_index_equal(result, expected, exact=True)
 
-    @pytest.mark.xfail
     def test_construction_empty_with_bool_categories(self):
         # see gh-22702
         cat = pd.CategoricalIndex([], categories=[True, False])


gfyoung (approved):
Nice!

cc @TomAugspurger @jreback

doc/source/whatsnew/v0.23.5.txt
```diff
@@ -26,7 +26,7 @@ Fixed Regressions
 - Calling :meth:`DataFrameGroupBy.rank` and :meth:`SeriesGroupBy.rank` with empty groups
   and ``pct=True`` was raising a ``ZeroDivisionError`` due to `c1068d9
   <https://github.com/pandas-dev/pandas/commit/c1068d9d242c22cb2199156f6fb82eb5759178ae>`_ (:issue:`22519`)
--
+- Constructing a :class:`pd.CategoricalIndex` with empty values and boolean categories was raising a ``ValueError`` after a change to dtype coercion in `78184868 <https://github.com/pandas-dev/pandas/commit/7818486859d1aba53ce359b93cfc772e688958e5>`_ (:issue:`22702`).
```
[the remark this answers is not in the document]
pganssle: OK, moved.

pandas/tests/arrays/categorical/test_constructors.py
```diff
@@ -42,6 +42,12 @@ def test_constructor_empty(self):
         expected = pd.Int64Index([1, 2, 3])
         tm.assert_index_equal(c.categories, expected)
 
+    def test_constructor_empty_boolean(self):
+        # see gh-22702
+        cat = pd.CategoricalIndex([], categories=[True, False])
```
jschendel: Should this be `Categorical` instead of `CategoricalIndex`?  If you want to test `CategoricalIndex`, I think the test should go in `tests/indexes/test_category.py`.

pandas/core/arrays/categorical.py
```diff
@@ -2439,11 +2439,16 @@ def _get_codes_for_values(values, categories):
...
-
     from pandas.core.algorithms import _get_data_algo, _hashtables
     if not is_dtype_equal(values.dtype, categories.dtype):
         values = ensure_object(values)
         categories = ensure_object(categories)
+    else:
+        # These may be Index, in which case their dtype would be coerced
+        # as part of _get_data_algo; in that case, we should retrieve the
+        # underlying numpy array instead.
+        values = getattr(values, 'values', values)
```
jreback: I think you can just use ``np.asarray`` here as we have to re-infer anyhow. But pls check perf of this. The comment can be a bit simpler, and reference the gh issue number

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -613,6 +613,7 @@ Categorical
 ^^^^^^^^^^^
 
 - Bug in :meth:`Categorical.from_codes` where ``NaN`` values in ``codes`` were silently converted to ``0`` (:issue:`21767`). In the future this will raise a ``ValueError``. Also changes the behavior of ``.from_codes([1.1, 2.0])``.
+- Constructing a :class:`pd.CategoricalIndex` with empty values and boolean categories was raising a ``ValueError`` after a change to dtype coercion in `78184868 <https://github.com/pandas-dev/pandas/commit/7818486859d1aba53ce359b93cfc772e688958e5>`_ (:issue:`22702`).
```
jreback: you don't need to be so explict here (e.g. the commit is not really necessary)

jreback (changes requested):

pandas/core/arrays/categorical.py
pganssle:
```python
In [4]: categories = pd.Index([True, False], dtype='object')

In [5]: values = np.array([], dtype='object')

In [6]: %timeit np.asarray(values)
260 ns ± 1.41 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [7]: %timeit np.asarray(categories)
2.2 µs ± 21.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [8]: %timeit getattr(values, 'values', values)
312 ns ± 1.26 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [9]: %timeit getattr(categories, 'values', categories)
410 ns ± 4.23 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [10]: categories = pd.Index(list('abcdefghijklmnopqrstuvwxyz'), dtype='object')

In [11]: categories
Out[11]: 
Index(['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n',
       'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z'],
      dtype='object')

In [12]: %timeit np.asarray(categories)
2.09 µs ± 9.76 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [13]: %timeit getattr(categories, 'values', categories)
409 ns ± 6.44 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each
```

Looks like the way I've done it is faster, but `np.asarray` doesn't seem to be scaling with the number of categories, so I don't think it's a huge deal to use `np.asarray`. That said, I don't love the idea of using `np.asarray` in this situation, because although I can't *think* of a time when it would present a problem, `np.asarray` is allowed to infer `dtype`. The way I've done it you're guaranteed to either get the original thing back or the underlying numpy array, which is pretty much as light a touch as I can imagine.

doc/source/whatsnew/v0.24.0.txt
pganssle: Will remove it. This is a remnant from when I had this under the `0.23.5.txt` release notes under "regressions fixed".

pandas/tests/arrays/categorical/test_constructors.py
pganssle: Added a test into `tests/indexes/test_category.py` and also changed this over to use `pd.Category`, since the bug is present in both the index and the Category itself. Sharp eye @jschendel!

pganssle:
I've pushed new changes that address all comments (though I did not switch over to `asarray`).

jreback (changes requested):
tiny code change, otherwise lgtm. ping on green.

pandas/core/arrays/categorical.py
```diff
@@ -2439,11 +2439,15 @@ def _get_codes_for_values(values, categories):
     """
     utility routine to turn values into codes given the specified categories
     """
-
     from pandas.core.algorithms import _get_data_algo, _hashtables
     if not is_dtype_equal(values.dtype, categories.dtype):
         values = ensure_object(values)
```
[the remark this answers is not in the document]
pganssle: I can do that if you want. One thing to note is that I believe `if not is_dtype_equal` case is probably the "common case", since it doesn't occur with many `dtype`s like integers or strings.

I don't *think* there's any speed advantage to being in the `if` branch vs. the `else` branch, but depending on convention you may prefer the common case to be the first block and the rare case to be the `else` block. Up to you.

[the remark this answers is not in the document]
jreback: its not about speed or anything else, its about consistency in our code of how we do these kinds of tests.

pganssle:
All issues should be addressed now. I'm guessing the azure pipeline failure is unrelated. I can rebase against master if it's fixed in master.

TomAugspurger:
Currently working on the azure stuff, you can safely ignore it.

jreback (approved):

jreback:
lgtm. merging after azure updates by @TomAugspurger

TomAugspurger:
Sorry about all the noise, your PR is my guinea pig :)

Things should be OK now. Ping on green.

pganssle:
@TomAugspurger Green now.

TomAugspurger:
Thanks!

PR #22715, opened 2018-09-14
TST: Mock clipboard IO

TomAugspurger:
No idea if this will work on not.


TomAugspurger:
This failed.

Unless anyone objects, I'm going to just mock the actual clipboard layer part of things.

jreback:
yeah the clipboard tests are annoying

pandas/tests/io/test_clipboard.py
```diff
@@ -76,10 +76,41 @@ def df(request):
         raise ValueError
 
 
+# our local clipboard for tests
+_mock_data = {}
```
jreback: it’s really odd to keep state outside of this fixture when it’s a function level

why don’t u just make it a module level?
or it seems that function level might be ok with state inside

pandas/tests/io/test_clipboard.py
```diff
@@ -76,10 +76,41 @@ def df(request):
...
+
+
+@pytest.fixture
+def mock_clipboard(mock, request):
+    def _mock_set(data):
+        _mock_data[request.node.name] = data
```
jreback: can u add a docstring

pandas/tests/io/test_clipboard.py
```diff
@@ -76,10 +76,41 @@ def df(request):
...
+
 @pytest.mark.single
+@pytest.mark.clipboard
 @pytest.mark.skipif(not _DEPS_INSTALLED,
                     reason="clipboard primitives not installed")
+@pytest.mark.usefixtures("mock_clipboard")
 class TestClipboard(object):
+
+    @pytest.mark.clipboard
+    def test_mock_clipboard(self):
```
jreback: can’t u use monkey patch here?

jreback (changes requested):

TomAugspurger:
Moved the mock clipboard to be function-scope, rather than global.

jreback:
sure

PR #22723, opened 2018-09-15
TST/CLN: remove duplicate data file used in tests (unicode_series.csv)

simonjayhawkins:
- [N/A ] closes #xxxx
- [x] tests added / passed
- [ N/A] whatsnew entry


WillAyd:
Can you open an issue for this? Not entirely clear of the context

pandas/tests/io/conftest.py
```diff
@@ -14,6 +14,12 @@ def jsonl_file(datapath):
     return datapath('io', 'parser', 'data', 'items.jsonl')
 
 
+@pytest.fixture
+def unicode_series_file(datapath):
```
WillAyd: Hmm not so sure about this being a dedicated fixture - what's the thought in doing this instead of just resolving `datapath` in the functions that need it?

pandas/tests/io/parser/dtypes.py
```diff
@@ -125,13 +125,14 @@ def test_categorical_dtype_high_cardinality_numeric(self):
             np.sort(actual.a.cat.categories), ordered=True)
         tm.assert_frame_equal(actual, expected)
 
-    def test_categorical_dtype_encoding(self, datapath):
+    def test_categorical_dtype_encoding(self, datapath, unicode_series_file):
```
WillAyd: Does this still need the `datapath` fixture with how you've configured?

WillAyd (changes requested):

simonjayhawkins:
@WillAyd 
> Can you open an issue for this? Not entirely clear of the context
there is no issue. context: just came across duplicate data file and thought it could be de-duplicated!

pandas/tests/io/parser/dtypes.py
simonjayhawkins: the test uses another file so the datapath fixture is still required.

pandas/tests/io/conftest.py
simonjayhawkins: sorted.

pandas/tests/io/parser/dtypes.py
simonjayhawkins: no longer applicable.

TomAugspurger:
Some context: previously, we had a function `get_data_path` that did some black magic to get the directory of the currently running test, so we had to duplicate files for each tree that needed the file :) 

@simonjayhawkins did you notice any other duplicates?

WillAyd (approved):
One Travis build timed out which I restarted. Another failed due to clipboard tests which I believe were resolved in #22715 so lgtm so long as the restart goes green

jreback:
thanks @simonjayhawkins

PR #22725, opened 2018-09-16
CLN/ERR: str.cat internals

### LINKED ISSUE #22721: str.cat not working with binary data on Python3
In Python 2, the following works:
```
s = pd.Series(np.array(list('abc'), 'S1').astype(object))
t = pd.Series(np.array(list('def'), 'S1').astype(object))
s.str.cat(t)
# 0    b'ad'
# 1    b'be'
# 2    b'cf'
# dtype: object
```

whereas in Python 3, `s.str.cat(t)` yields:
```
TypeError: sequence item 0: expected str instance, bytes found
```

Clearly, Python 3 fundamentally changed the separation between bytes and strings, but in principle, there is no reason (other than implementation) why this should not have the same result in Python 3 as in Python 2.

### LINKED ISSUE #22722: Improve TypeError message for str.cat
Currently, 

```
s = pd.Series(['a', 'b', 'c'])
s.str.cat([1, 2, 3])
```
yields
```
  File "C:\Users\Axel Obermeier\eclipse-workspace\pddev\pandas\core\strings.py", line 2222, in cat
    res = str_cat(data, others=others, sep=sep, na_rep=na_rep)
  File "C:\Users\Axel Obermeier\eclipse-workspace\pddev\pandas\core\strings.py", line 111, in str_cat
    cats = [sep.join(tup) for tup in tuples]
  File "C:\Users\Axel Obermeier\eclipse-workspace\pddev\pandas\core\strings.py", line 111, in <listcomp>
    cats = [sep.join(tup) for tup in tuples]
TypeError: sequence item 1: expected str instance, int found
```

IMO, this should be improved to have a better error message, and shallower stack trace.

h-vetinari:
This is mainly a clean-up of internal methods for `str.cat` that I didn't want to touch within #20347.

~~As a side benefit of changing the implementation, this also solves #22721. Finally, I've also added a better message for TypeErrors (closes #22722)~~

~~closes #22721~~
~~closes #22722~~
- [x] tests modified / passed
- [x] whatsnew entry

Here's the ASV output (the original implementation of this PR (see first commit) that used more higher-level pandas-functions like `fillna`, `drop_na`, etc. was up to three times slower, so I tweaked it some more, and actually believe that the last solution with `interleave_sep` is the most elegant anyway):
```
     [37455764]       [4d1710f1]
         10.9±1ms         9.11±1ms    ~0.83  strings.Cat.time_cat(0, ',', '-', 0.0)
         9.55±1ms         10.9±0ms    ~1.15  strings.Cat.time_cat(0, ',', '-', 0.001)
         12.5±2ms         14.1±2ms    ~1.12  strings.Cat.time_cat(0, ',', '-', 0.15)
         9.94±1ms       9.23±0.7ms     0.93  strings.Cat.time_cat(0, ',', None, 0.0)
         14.3±2ms         8.68±1ms    ~0.61  strings.Cat.time_cat(0, ',', None, 0.001)
-        13.7±1ms       11.7±0.8ms     0.86  strings.Cat.time_cat(0, ',', None, 0.15)
       9.11±0.7ms         7.81±2ms    ~0.86  strings.Cat.time_cat(0, None, '-', 0.0)
       9.38±0.8ms         10.9±1ms    ~1.17  strings.Cat.time_cat(0, None, '-', 0.001)
         11.4±2ms         10.9±2ms     0.96  strings.Cat.time_cat(0, None, '-', 0.15)
         13.4±2ms       9.38±0.6ms    ~0.70  strings.Cat.time_cat(0, None, None, 0.0)
         9.23±2ms         11.7±1ms    ~1.27  strings.Cat.time_cat(0, None, None, 0.001)
         10.2±2ms         10.9±1ms     1.08  strings.Cat.time_cat(0, None, None, 0.15)
-        70.3±4ms         54.7±4ms     0.78  strings.Cat.time_cat(3, ',', '-', 0.0)
         62.5±4ms        46.9±20ms    ~0.75  strings.Cat.time_cat(3, ',', '-', 0.001)
        93.8±10ms        66.4±10ms    ~0.71  strings.Cat.time_cat(3, ',', '-', 0.15)
         62.5±4ms         62.5±4ms     1.00  strings.Cat.time_cat(3, ',', None, 0.0)
+        46.9±4ms         85.9±4ms     1.83  strings.Cat.time_cat(3, ',', None, 0.001)
         52.1±2ms         50.8±6ms     0.97  strings.Cat.time_cat(3, ',', None, 0.15)
         50.8±8ms         54.7±6ms     1.08  strings.Cat.time_cat(3, None, '-', 0.0)
         54.7±4ms         62.5±4ms    ~1.14  strings.Cat.time_cat(3, None, '-', 0.001)
         62.5±5ms         54.7±3ms    ~0.88  strings.Cat.time_cat(3, None, '-', 0.15)
         46.9±4ms         39.1±0ms    ~0.83  strings.Cat.time_cat(3, None, None, 0.0)
         46.9±8ms         58.6±6ms    ~1.25  strings.Cat.time_cat(3, None, None, 0.001)
         46.9±9ms         54.7±4ms    ~1.17  strings.Cat.time_cat(3, None, None, 0.15)
                                                                  ^  ^     ^     ^
                                                                  |  |     |     |
                                                         other_cols  |   na_rep  |
                                                                     |           |
                                                                    sep        na_frac
```

There's a bunch of noise in there, but by and large, things don't look so bad IMO. Especially, when one excludes the not-so-common worst-case scenario of a very small but non-zero amount of NaNs (`na_frac=0.001`):
```
     [37455764]       [4d1710f1]
         10.9±1ms         9.11±1ms    ~0.83  strings.Cat.time_cat(0, ',', '-', 0.0)
         12.5±2ms         14.1±2ms    ~1.12  strings.Cat.time_cat(0, ',', '-', 0.15)
         9.94±1ms       9.23±0.7ms     0.93  strings.Cat.time_cat(0, ',', None, 0.0)
-        13.7±1ms       11.7±0.8ms     0.86  strings.Cat.time_cat(0, ',', None, 0.15)
       9.11±0.7ms         7.81±2ms    ~0.86  strings.Cat.time_cat(0, None, '-', 0.0)
         11.4±2ms         10.9±2ms     0.96  strings.Cat.time_cat(0, None, '-', 0.15)
         13.4±2ms       9.38±0.6ms    ~0.70  strings.Cat.time_cat(0, None, None, 0.0)
         10.2±2ms         10.9±1ms     1.08  strings.Cat.time_cat(0, None, None, 0.15)
-        70.3±4ms         54.7±4ms     0.78  strings.Cat.time_cat(3, ',', '-', 0.0)
         62.5±4ms        46.9±20ms    ~0.75  strings.Cat.time_cat(3, ',', '-', 0.001)
        93.8±10ms        66.4±10ms    ~0.71  strings.Cat.time_cat(3, ',', '-', 0.15)
         62.5±4ms         62.5±4ms     1.00  strings.Cat.time_cat(3, ',', None, 0.0)
         52.1±2ms         50.8±6ms     0.97  strings.Cat.time_cat(3, ',', None, 0.15)
         50.8±8ms         54.7±6ms     1.08  strings.Cat.time_cat(3, None, '-', 0.0)
         62.5±5ms         54.7±3ms    ~0.88  strings.Cat.time_cat(3, None, '-', 0.15)
         46.9±4ms         39.1±0ms    ~0.83  strings.Cat.time_cat(3, None, None, 0.0)
         46.9±9ms         54.7±4ms    ~1.17  strings.Cat.time_cat(3, None, None, 0.15)
                                                                  ^  ^     ^     ^
                                                                  |  |     |     |
                                                         other_cols  |   na_rep  |
                                                                     |           |
                                                                    sep        na_frac
```

### DIFF
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -2210,11 +2210,11 @@ class StringMethods(NoNewAttributesMixin):
         # concatenate Series/Index with itself if no "others"
         if others is None:
             data = ensure_object(data)
-            mask = isna(data)
-            if na_rep is None and mask.any():
-                data = data[~mask]
-            elif na_rep is not None and mask.any():
-                data = np.where(mask, na_rep, data)
+            na_mask = isna(data)
+            if na_rep is None and na_mask.any():
+                data = data[~na_mask]
+            elif na_rep is not None and na_mask.any():
+                data = np.where(na_mask, na_rep, data)
             return sep.join(data)
 
         try:
@@ -2254,8 +2254,8 @@ class StringMethods(NoNewAttributesMixin):
             others = [others[x] for x in others]  # again list of Series
 
         all_cols = [ensure_object(x) for x in [data] + others]
-        masks = np.array([isna(x) for x in all_cols])
-        union_mask = np.logical_or.reduce(masks, axis=0)
+        na_masks = np.array([isna(x) for x in all_cols])
+        union_mask = np.logical_or.reduce(na_masks, axis=0)
 
         if na_rep is None and union_mask.any():
             # no na_rep means NaNs for all rows where any column has a NaN
@@ -2268,8 +2268,8 @@ class StringMethods(NoNewAttributesMixin):
                                           sep)
         elif na_rep is not None and union_mask.any():
             # fill NaNs with na_rep in case there are actually any NaNs
-            all_cols = [np.where(mask, na_rep, col)
-                        for mask, col in zip(masks, all_cols)]
+            all_cols = [np.where(nm, na_rep, col)
+                        for nm, col in zip(na_masks, all_cols)]
             result = cat_core(all_cols, sep)
         else:
             # no NaNs - can just concatenate
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -18,7 +18,6 @@ from pandas.core.dtypes.common import (
 import pandas.core.common as com
 from pandas.core.algorithms import take_1d
 import pandas.compat as compat
-from pandas.compat.numpy import _np_version_under1p11
 from pandas.core.base import NoNewAttributesMixin
 from pandas.util._decorators import Appender
 import re
@@ -38,15 +37,15 @@ _cpython_optimized_decoders = _cpython_optimized_encoders + (
 _shared_docs = dict()
 
 
-def cat_core(all_cols, sep):
+def cat_core(list_of_columns, sep):
     """
     Auxiliary function for :meth:`str.cat`
 
     Parameters
     ----------
-    all_cols : two-dimensional numpy array
-        array of columns to be concatenated with sep;
-        this array may not contain NaNs!
+    list_of_columns : list of numpy arrays
+        List of arrays to be concatenated with sep;
+        these arrays may not contain NaNs!
     sep : string
         The separator string for concatenating the columns
 
@@ -55,15 +54,9 @@ def cat_core(all_cols, sep):
     nd.array
         The concatenation of list_of_columns with sep
     """
-    list_of_columns = np.split(all_cols, all_cols.shape[1], axis=1)
     list_with_sep = [sep] * (2 * len(list_of_columns) - 1)
     list_with_sep[::2] = list_of_columns
-    res = np.sum(list_with_sep, axis=0)
-    if not (_np_version_under1p11 and len(res) == 0):
-        # np.split splits into arrays of shape (N, 1); NOT (N,)
-        # need to reduce dimensionality of result
-        res = res[:, 0]
-    return res
+    return np.sum(list_with_sep, axis=0)
 
 
 def _na_map(f, arr, na_result=np.nan, dtype=object):
@@ -2246,21 +2239,21 @@ class StringMethods(NoNewAttributesMixin):
                           "'outer'|'inner'|'right'`. The future default will "
                           "be `join='left'`.", FutureWarning, stacklevel=2)
 
-        # concatenate others into DataFrame; need to add keys for uniqueness in
-        # case of duplicate columns (for join is None, all indexes are already
-        # the same after _get_series_list, which forces alignment in this case)
-        others = concat(others, axis=1,
-                        join=(join if join == 'inner' else 'outer'),
-                        keys=range(len(others)), copy=False)
+        # if join is None, _get_series_list already force-aligned indexes
+        join = 'left' if join is None else join
 
         # align if required
-        if not data.index.equals(others.index):
+        if any(not data.index.equals(x.index) for x in others):
+            # Need to add keys for uniqueness in case of duplicate columns
+            others = concat(others, axis=1,
+                            join=(join if join == 'inner' else 'outer'),
+                            keys=range(len(others)), copy=False)
             data, others = data.align(others, join=join)
+            others = [others[x] for x in others]  # again list of Series
 
-        # collect all columns
-        all_cols = ensure_object(concat([data, others], axis=1, copy=False))
-        na_masks = isna(all_cols)
-        union_mask = np.logical_or.reduce(na_masks, axis=1)
+        all_cols = [ensure_object(x) for x in [data] + others]
+        na_masks = np.array([isna(x) for x in all_cols])
+        union_mask = np.logical_or.reduce(na_masks, axis=0)
 
         if na_rep is None and union_mask.any():
             # no na_rep means NaNs for all rows where any column has a NaN
@@ -2269,10 +2262,13 @@ class StringMethods(NoNewAttributesMixin):
             np.putmask(result, union_mask, np.nan)
 
             not_masked = ~union_mask
-            result[not_masked] = cat_core(all_cols[not_masked], sep)
+            result[not_masked] = cat_core([x[not_masked] for x in all_cols],
+                                          sep)
         elif na_rep is not None and union_mask.any():
             # fill NaNs with na_rep in case there are actually any NaNs
-            result = cat_core(np.where(na_masks, na_rep, all_cols), sep)
+            all_cols = [np.where(nm, na_rep, col)
+                        for nm, col in zip(na_masks, all_cols)]
+            result = cat_core(all_cols, sep)
         else:
             # no NaNs - can just concatenate
             result = cat_core(all_cols, sep)
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -18,6 +18,7 @@ from pandas.core.dtypes.common import (
 import pandas.core.common as com
 from pandas.core.algorithms import take_1d
 import pandas.compat as compat
+from pandas.compat.numpy import _np_version_under1p11
 from pandas.core.base import NoNewAttributesMixin
 from pandas.util._decorators import Appender
 import re
@@ -57,9 +58,12 @@ def cat_core(all_cols, sep):
     list_of_columns = np.split(all_cols, all_cols.shape[1], axis=1)
     list_with_sep = [sep] * (2 * len(list_of_columns) - 1)
     list_with_sep[::2] = list_of_columns
-    # np.split splits into arrays of shape (N, 1); NOT (N,)
-    # need to reduce dimensionality of result
-    return np.sum(list_with_sep, axis=0)[:, 0]
+    res = np.sum(list_with_sep, axis=0)
+    if not (_np_version_under1p11 and len(res) == 0):
+        # np.split splits into arrays of shape (N, 1); NOT (N,)
+        # need to reduce dimensionality of result
+        res = res[:, 0]
+    return res
 
 
 def _na_map(f, arr, na_result=np.nan, dtype=object):
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -5,6 +5,7 @@ from pandas.compat import zip
 from pandas.core.dtypes.generic import ABCSeries, ABCIndex
 from pandas.core.dtypes.missing import isna
 from pandas.core.dtypes.common import (
+    ensure_object,
     is_bool_dtype,
     is_categorical_dtype,
     is_object_dtype,
@@ -36,13 +37,13 @@ _cpython_optimized_decoders = _cpython_optimized_encoders + (
 _shared_docs = dict()
 
 
-def interleave_sep(all_cols, sep):
-    '''
+def interleave_sep(list_of_columns, sep):
+    """
     Auxiliary function for :meth:`str.cat`
 
     Parameters
     ----------
-    all_cols : list of numpy arrays
+    list_of_columns : list of numpy arrays
         List of arrays to be concatenated with sep
     sep : string
         The separator string for concatenating the columns
@@ -51,12 +52,12 @@ def interleave_sep(all_cols, sep):
     -------
     list
         The list of arrays interleaved with sep; to be fed to np.sum
-    '''
+    """
     if sep == '':
         # no need to add empty strings
-        return all_cols
-    result = [sep] * (2 * len(all_cols) - 1)
-    result[::2] = all_cols
+        return list_of_columns
+    result = [sep] * (2 * len(list_of_columns) - 1)
+    result[::2] = list_of_columns
     return result
 
 
@@ -2207,12 +2208,12 @@ class StringMethods(NoNewAttributesMixin):
 
         # concatenate Series/Index with itself if no "others"
         if others is None:
-            data = data.astype(object).values
+            data = ensure_object(data)
             mask = isna(data)
-            if mask.any():
-                if na_rep is None:
-                    return sep.join(data[~mask])
-                return sep.join(np.where(mask, na_rep, data))
+            if na_rep is None and mask.any():
+                data = data[~mask]
+            elif na_rep is not None and mask.any():
+                data = np.where(mask, na_rep, data)
             return sep.join(data)
 
         try:
@@ -2251,11 +2252,13 @@ class StringMethods(NoNewAttributesMixin):
             data, others = data.align(others, join=join)
             others = [others[x] for x in others]  # again list of Series
 
-        all_cols = [x.astype(object).values for x in [data] + others]
+        all_cols = [ensure_object(x) for x in [data] + others]
         masks = np.array([isna(x) for x in all_cols])
         union_mask = np.logical_or.reduce(masks, axis=0)
 
         if na_rep is None and union_mask.any():
+            # no na_rep means NaNs for all rows where any column has a NaN
+            # only necessary if there are actually any NaNs
             result = np.empty(len(data), dtype=object)
             np.putmask(result, union_mask, np.nan)
 
@@ -2264,11 +2267,12 @@ class StringMethods(NoNewAttributesMixin):
 
             result[not_masked] = np.sum(all_cols, axis=0)
         elif na_rep is not None and union_mask.any():
-            # fill NaNs
-            all_cols = [np.where(masks[i], na_rep, all_cols[i])
-                        for i in range(len(all_cols))]
+            # fill NaNs with na_rep in case there are actually any NaNs
+            all_cols = [np.where(mask, na_rep, col)
+                        for mask, col in zip(masks, all_cols)]
             result = np.sum(interleave_sep(all_cols, sep), axis=0)
-        else:  # no NaNs
+        else:
+            # no NaNs - can just concatenate
             result = np.sum(interleave_sep(all_cols, sep), axis=0)
 
         if isinstance(self._orig, Index):
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -37,28 +37,29 @@ _cpython_optimized_decoders = _cpython_optimized_encoders + (
 _shared_docs = dict()
 
 
-def interleave_sep(list_of_columns, sep):
+def cat_core(list_of_columns, sep):
     """
     Auxiliary function for :meth:`str.cat`
 
     Parameters
     ----------
     list_of_columns : list of numpy arrays
-        List of arrays to be concatenated with sep
+        List of arrays to be concatenated with sep;
+        these arrays may not contain NaNs!
     sep : string
         The separator string for concatenating the columns
 
     Returns
     -------
-    list
-        The list of arrays interleaved with sep; to be fed to np.sum
+    nd.array
+        The concatenation of list_of_columns with sep
     """
     if sep == '':
         # no need to add empty strings
-        return list_of_columns
-    result = [sep] * (2 * len(list_of_columns) - 1)
-    result[::2] = list_of_columns
-    return result
+        return np.sum(list_of_columns, axis=0)
+    list_with_sep = [sep] * (2 * len(list_of_columns) - 1)
+    list_with_sep[::2] = list_of_columns
+    return np.sum(list_with_sep, axis=0)
 
 
 def _na_map(f, arr, na_result=np.nan, dtype=object):
@@ -2263,17 +2264,16 @@ class StringMethods(NoNewAttributesMixin):
             np.putmask(result, union_mask, np.nan)
 
             not_masked = ~union_mask
-            all_cols = interleave_sep([x[not_masked] for x in all_cols], sep)
-
-            result[not_masked] = np.sum(all_cols, axis=0)
+            result[not_masked] = cat_core([x[not_masked] for x in all_cols],
+                                          sep)
         elif na_rep is not None and union_mask.any():
             # fill NaNs with na_rep in case there are actually any NaNs
             all_cols = [np.where(mask, na_rep, col)
                         for mask, col in zip(masks, all_cols)]
-            result = np.sum(interleave_sep(all_cols, sep), axis=0)
+            result = cat_core(all_cols, sep)
         else:
             # no NaNs - can just concatenate
-            result = np.sum(interleave_sep(all_cols, sep), axis=0)
+            result = cat_core(all_cols, sep)
 
         if isinstance(self._orig, Index):
             result = Index(result, name=self._orig.name)
--- a/pandas/core/strings.py
+++ b/pandas/core/strings.py
@@ -3,7 +3,7 @@ import numpy as np
 
 from pandas.compat import zip
 from pandas.core.dtypes.generic import ABCSeries, ABCIndex
-from pandas.core.dtypes.missing import isna, notna
+from pandas.core.dtypes.missing import isna
 from pandas.core.dtypes.common import (
     is_bool_dtype,
     is_categorical_dtype,
@@ -36,114 +36,28 @@ _cpython_optimized_decoders = _cpython_optimized_encoders + (
 _shared_docs = dict()
 
 
-def _get_array_list(arr, others):
-    """
-    Auxiliary function for :func:`str_cat`
+def interleave_sep(all_cols, sep):
+    '''
+    Auxiliary function for :meth:`str.cat`
 
     Parameters
     ----------
-    arr : ndarray
-        The left-most ndarray of the concatenation
-    others : list, ndarray, Series
-        The rest of the content to concatenate. If list of list-likes,
-        all elements must be passable to ``np.asarray``.
+    all_cols : list of numpy arrays
+        List of arrays to be concatenated with sep
+    sep : string
+        The separator string for concatenating the columns
 
     Returns
     -------
     list
-        List of all necessary arrays
-    """
-    from pandas.core.series import Series
-
-    if len(others) and isinstance(com.values_from_object(others)[0],
-                                  (list, np.ndarray, Series)):
-        arrays = [arr] + list(others)
-    else:
-        arrays = [arr, others]
-
-    return [np.asarray(x, dtype=object) for x in arrays]
-
-
-def str_cat(arr, others=None, sep=None, na_rep=None):
-    """
-    Auxiliary function for :meth:`str.cat`
-
-    If `others` is specified, this function concatenates the Series/Index
-    and elements of `others` element-wise.
-    If `others` is not being passed then all values in the Series are
-    concatenated in a single string with a given `sep`.
-
-    Parameters
-    ----------
-    others : list-like, or list of list-likes, optional
-        List-likes (or a list of them) of the same length as calling object.
-        If None, returns str concatenating strings of the Series.
-    sep : string or None, default None
-        If None, concatenates without any separator.
-    na_rep : string or None, default None
-        If None, NA in the series are ignored.
-
-    Returns
-    -------
-    concat
-        ndarray containing concatenated results (if `others is not None`)
-        or str (if `others is None`)
-    """
-    if sep is None:
-        sep = ''
-
-    if others is not None:
-        arrays = _get_array_list(arr, others)
-
-        n = _length_check(arrays)
-        masks = np.array([isna(x) for x in arrays])
-        cats = None
-
-        if na_rep is None:
-            na_mask = np.logical_or.reduce(masks, axis=0)
-
-            result = np.empty(n, dtype=object)
-            np.putmask(result, na_mask, np.nan)
-
-            notmask = ~na_mask
-
-            tuples = zip(*[x[notmask] for x in arrays])
-            cats = [sep.join(tup) for tup in tuples]
-
-            result[notmask] = cats
-        else:
-            for i, x in enumerate(arrays):
-                x = np.where(masks[i], na_rep, x)
-                if cats is None:
-                    cats = x
-                else:
-                    cats = cats + sep + x
-
-            result = cats
-
-        return result
-    else:
-        arr = np.asarray(arr, dtype=object)
-        mask = isna(arr)
-        if na_rep is None and mask.any():
-            if sep == '':
-                na_rep = ''
-            else:
-                return sep.join(arr[notna(arr)])
-        return sep.join(np.where(mask, na_rep, arr))
-
-
-def _length_check(others):
-    n = None
-    for x in others:
-        try:
-            if n is None:
-                n = len(x)
-            elif len(x) != n:
-                raise ValueError('All arrays must be same length')
-        except TypeError:
-            raise ValueError('Must pass arrays containing strings to str_cat')
-    return n
+        The list of arrays interleaved with sep; to be fed to np.sum
+    '''
+    if sep == '':
+        # no need to add empty strings
+        return all_cols
+    result = [sep] * (2 * len(all_cols) - 1)
+    result[::2] = all_cols
+    return result
 
 
 def _na_map(f, arr, na_result=np.nan, dtype=object):
@@ -2283,6 +2197,8 @@ class StringMethods(NoNewAttributesMixin):
 
         if isinstance(others, compat.string_types):
             raise ValueError("Did you mean to supply a `sep` keyword?")
+        if sep is None:
+            sep = ''
 
         if isinstance(self._orig, Index):
             data = Series(self._orig, index=self._orig)
@@ -2291,9 +2207,13 @@ class StringMethods(NoNewAttributesMixin):
 
         # concatenate Series/Index with itself if no "others"
         if others is None:
-            result = str_cat(data, others=others, sep=sep, na_rep=na_rep)
-            return self._wrap_result(result,
-                                     use_codes=(not self._is_categorical))
+            data = data.astype(object).values
+            mask = isna(data)
+            if mask.any():
+                if na_rep is None:
+                    return sep.join(data[~mask])
+                return sep.join(np.where(mask, na_rep, data))
+            return sep.join(data)
 
         try:
             # turn anything in "others" into lists of Series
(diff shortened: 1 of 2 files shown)


gfyoung:
@WillAyd @jreback : The conversations went stale a little in the original issues, and I'm not sure how well aligned these changes with what you guys were suggesting or saying in them.

h-vetinari:
@gfyoung @WillAyd @jreback 
I opened #22721 because the behaviour was changed through this refactor. As I said in the issue, I'm happy to disable it, but that would need adding in some check against binary data (because `np.sum` -- contrary to the previous implementation -- doesn't throw an error for binary data).

#22722 is equally something that I'm not going to fight for. I got some less than ideal error messages while testing, and decided this could/should be improved. If you disagree, it's easy to remove the offending lines.

gfyoung:
@h-vetinari : Not a problem.  I was only reading through the issues and your PR and matching up what you were doing with what was being said.  Just pinging to get their eyes on this.

WillAyd:
I'm still very much against setting the expectation that the `.str` accessor will work with byte objects.

Also generally don't think that a "cleanup" should be performed in a same PR that changes the expected functionality of the codebase. Would be much easier to stick to one thing at a time, i.e. a clean up that doesn't introduce or change any existing behavior

h-vetinari:
@WillAyd 
> Would be much easier to stick to one thing at a time, i.e. a clean up that doesn't introduce or change any existing behavior

I reverted the (unrelated) fix for #22721, but the situation for #22722 is more delicate. The genesis was as follows:
1. refactor
1. see that behaviour changed
1. (sorta) agree with new behaviour
1. open issue so that changed behaviour in PR is explained

Turns out, that the situation is even a bit more delicate, as currently on master, bytes can successfully concatenated *as long as* `sep` is explicitly set:
```
(pandas-dev) C:\Users\[...]\pddev>python
Python 3.6.6 |Anaconda, Inc.| (default, Jun 28 2018, 11:27:44) [MSC v.1900 64 bit (AMD64)] on win32
>>> import pandas as pd
>>> import numpy as np
>>> pd.__version__
'0.24.0.dev0+586.g8a1c8ad4b'
>>> s = pd.Series(np.array(list('abc'), 'S1').astype(object))
>>> t = pd.Series(np.array(list('def'), 'S1').astype(object))
>>> s.str.cat(t)
TypeError: sequence item 0: expected str instance, bytes found
>>> s.str.cat(t, sep=b'')
0    b'ad'
1    b'be'
2    b'cf'
dtype: object
>>> s.str.cat(t, sep=b',')
0    b'a,d'
1    b'b,e'
2    b'c,f'
dtype: object
```

The problem was that `tests/frame/test_strings.test_method_on_bytes` only tests `sep=None`.

h-vetinari:
This is the sort of thing I was talking about with the missing string dtype. Without one, there are legitimate off-label uses for the `.str`-accessor, and concatenating bytes workes like a charm *already*. The only real change here then would be that `sep=None` does not automatically trigger the `TypeError`.

jreback (changes requested):
comments

h-vetinari:
@jreback
> We do NOT handle bytes in .str

Yes you (currently) do. Just try the code I posted above.

> pls pls pls 1 thing per PR.

The whatsnew-note notwithstanding, this PR only changes the implementation (you'll see in the test that I've not changed anything substantial)

I understand that you don't want people using `.str` for byte data, but it works currently. The problem is that there's no good dtype distinction, and inspecting every element of a Series when calling `.str` would come with a big perf hit.

jreback:
> Yes you (currently) do. Just try the code I posted above.

It may happen to work. Instead of refactoring this as I said above, would prefer tests / and better error messages with bytes inputs.

h-vetinari:
@jreback 
> It may happen to work. Instead of refactoring this as I said above, would prefer tests / and better error messages with bytes inputs.

The point is that this PR *does not change the current behaviour* and should stand on its merits, unrelated to the fact that you'd like to disallow `.str` on bytes.

jreback:
@h-vetinari you removed tests, so clearly you are changing things.

h-vetinari:
@jreback 
> you removed tests, so clearly you are changing things.

I removed *one* test for the internal method that has been factored away. Furthermore, this removed test (`test_cat`) is exactly replicated in the test directly below `test_str_cat`.

pandas/tests/test_strings.py
```diff
@@ -97,53 +97,6 @@ def test_iter_object_try_string(self):
         assert i == 100
         assert s == 'h'
 
-    def test_cat(self):
-        one = np.array(['a', 'a', 'b', 'b', 'c', NA], dtype=np.object_)
-        two = np.array(['a', NA, 'b', 'd', 'foo', NA], dtype=np.object_)
-
-        # single array
-        result = strings.str_cat(one)
```
h-vetinari: @jreback 
> I removed one test for the internal method that has been factored away

Please have a look here - this is directly importing the internal method and testing it (*not* `str.cat`)

pandas/tests/test_strings.py
```diff
@@ -97,53 +97,6 @@ def test_iter_object_try_string(self):
...
-
-        result = strings.str_cat(one, two)
-        exp = np.array(['aa', NA, 'bb', 'bd', 'cfoo', NA], dtype=np.object_)
-        tm.assert_almost_equal(result, exp)
-
-        # error for incorrect lengths
-        rgx = 'All arrays must be same length'
-        three = Series(['1', '2', '3'])
-
-        with tm.assert_raises_regex(ValueError, rgx):
```
h-vetinari: > Furthermore, this removed test (`test_cat`) is exactly replicated in the test directly below (`test_str_cat`).

I can't mark lines that are not in the diff, but check out https://github.com/pandas-dev/pandas/blob/fcd11b5ed53c82e60e7121f90651fa0d72d908c0/pandas/tests/test_strings.py#L119

I replicated the removed test (acting on `strings.test_cat`) as a test acting on `str.cat` within #20347.

h-vetinari (commented):
@jreback 

PTAL

h-vetinari:
@jreback 
While you're at it with the reviewing, please don't forget this one. :)

pandas/core/strings.py
```diff
@@ -36,114 +36,28 @@
...
-    else:
-        arrays = [arr, others]
-
-    return [np.asarray(x, dtype=object) for x in arrays]
-
-
-def str_cat(arr, others=None, sep=None, na_rep=None):
-    """
+def interleave_sep(all_cols, sep):
+    '''
```
jreback: use triple-double quotes

jreback: all_cols -> list_of_columns

pandas/core/strings.py
```diff
@@ -2180,9 +2096,13 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
 
         # concatenate Series/Index with itself if no "others"
         if others is None:
-            result = str_cat(data, others=others, sep=sep, na_rep=na_rep)
-            return self._wrap_result(result,
-                                     use_codes=(not self._is_categorical))
+            data = data.astype(object).values
```
jreback: why is this astype needed?

pandas/core/strings.py
```diff
@@ -2180,9 +2096,13 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
...
+            mask = isna(data)
+            if mask.any():
+                if na_rep is None:
+                    return sep.join(data[~mask])
+                return sep.join(np.where(mask, na_rep, data))
+            return sep.join(data)
```
jreback: can we do a single sep.join, and just have the branches mask the data as needed

pandas/core/strings.py
```diff
@@ -2209,23 +2129,42 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
...
             others = concat(others, axis=1,
                             join=(join if join == 'inner' else 'outer'),
-                            keys=range(len(others)))
+                            keys=range(len(others)), copy=False)
             data, others = data.align(others, join=join)
             others = [others[x] for x in others]  # again list of Series
 
-        # str_cat discards index
-        res = str_cat(data, others=others, sep=sep, na_rep=na_rep)
+        all_cols = [x.astype(object).values for x in [data] + others]
```
jreback: why do you need the astype, much prefer ``ensure_object`` generally

pandas/core/strings.py
```diff
@@ -2209,23 +2129,42 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
...
+        masks = np.array([isna(x) for x in all_cols])
+        union_mask = np.logical_or.reduce(masks, axis=0)
+
+        if na_rep is None and union_mask.any():
```
jreback: comment on these cases

pandas/core/strings.py
```diff
@@ -2209,23 +2129,42 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
...
+            result = np.empty(len(data), dtype=object)
+            np.putmask(result, union_mask, np.nan)
+
+            not_masked = ~union_mask
+            all_cols = interleave_sep([x[not_masked] for x in all_cols], sep)
+
+            result[not_masked] = np.sum(all_cols, axis=0)
+        elif na_rep is not None and union_mask.any():
+            # fill NaNs
+            all_cols = [np.where(masks[i], na_rep, all_cols[i])
```
jreback: use zip(masks, all_cols)

pandas/core/strings.py
```diff
@@ -36,114 +36,28 @@
...
-    return n
+    list
+        The list of arrays interleaved with sep; to be fed to np.sum
+    '''
+    if sep == '':
+        # no need to add empty strings
+        return all_cols
+    result = [sep] * (2 * len(all_cols) - 1)
+    result[::2] = all_cols
+    return result
```
jreback: I would simply do ``np.sum(result)`` here, no?

pandas/tests/test_strings.py
jreback: ok

pandas/tests/test_strings.py
```diff
@@ -3136,7 +3089,7 @@ def test_method_on_bytes(self):
         lhs = Series(np.array(list('abc'), 'S1').astype(object))
         rhs = Series(np.array(list('def'), 'S1').astype(object))
         if compat.PY3:
-            pytest.raises(TypeError, lhs.str.cat, rhs)
+            pytest.raises(TypeError, lhs.str.cat, rhs, sep=',')
```
jreback: is this the bytes concat?

jreback (changes requested):

pandas/core/strings.py
h-vetinari: OK, that's reasonable. Refactored the function as necessary

pandas/core/strings.py
h-vetinari: I used it because data may be categorical, and then values is not necessarily a numpy array. Changed to `ensure_object` which you mentioned below, hope this is better.

pandas/core/strings.py
h-vetinari: Done

pandas/core/strings.py
h-vetinari: Done

pandas/core/strings.py
h-vetinari: Added comments

pandas/tests/test_strings.py
h-vetinari: Yes

h-vetinari (commented):
Thanks for review; pushed new commits

h-vetinari:
@jreback 

PTAL

h-vetinari:
@jreback 

PTAL

jreback (approved):

jreback:
lgtm. @jorisvandenbossche @TomAugspurger @WillAyd if someone else would have a look

WillAyd:
Away for a few days but can take a look Monday. If someone approves in the interim feel free to merge

gfyoung (approved):

pandas/core/strings.py
```diff
@@ -36,114 +37,29 @@
...
-    others : list-like, or list of list-likes, optional
-        List-likes (or a list of them) of the same length as calling object.
-        If None, returns str concatenating strings of the Series.
-    sep : string or None, default None
-        If None, concatenates without any separator.
-    na_rep : string or None, default None
-        If None, NA in the series are ignored.
+    list_of_columns : list of numpy arrays
+        List of arrays to be concatenated with sep;
+        these arrays may not contain NaNs!
```
WillAyd: What is with this documented limitation? Believe in master that `NaN` is valid and returns `NaN` when passed in

WillAyd (changes requested):

pandas/core/strings.py
h-vetinari: @WillAyd 
the new helper function `cat_core` has nothing to do with the existing (essentially fully-fledged up to index-handling) `str_cat`. It's an internal function to avoid lots of copied code, and (as it just wraps `np.sum`) is not nan-safe for string values.

h-vetinari: @WillAyd 
Does that answer your question? what would you like to see for this docstring?

WillAyd:
Something seems off here (points to a gap in test coverage). Take the following as an example on your branch:

```python
In [4]: import pandas as pd

In [5]: ser1 = pd.Series(list(range(3)), dtype=object)

In [6]: ser2 = pd.Series(list(range(3, 6)), dtype=object)

In [7]: ser1.str.cat(ser2)
Out[7]: 
0    3
1    5
2    7
dtype: object
```

On master this raises:

```python
TypeError: sequence item 0: expected str instance, int found
```

Can you see what may be causing that difference in behavior? The new behavior is incorrect

h-vetinari:
@WIllAyd,

good catch. Will have to bite the bullet and always add `sep`, even if it's `''`, as this is the only way to catch these cases (barring a proper string dtype).

I have another PR corresponding to the part of this PR that I removed for the purpose of stream-lining it, which will raise when the entered columns have the wrong dtype (but wouldn't catch ints masquerading as objects).

h-vetinari:
Build failure unrelated - an error due to hypthesis internals/health checks:
```
______________ TestDataFrameAggregate.test_frequency_is_original _______________
[gw1] linux -- Python 3.7.0 /home/travis/miniconda3/envs/pandas/bin/python
self = <pandas.tests.frame.test_apply.TestDataFrameAggregate object at 0x7f9b7025db70>
    @given(index=indices(5), num_columns=integers(0, 5))
>   def test_frequency_is_original(self, index, num_columns):
pandas/tests/frame/test_apply.py:1159: 
[...]
E           hypothesis.errors.HypothesisDeprecationWarning: Test took 295.55ms to run.
In future the default deadline setting will be 200ms, which will make this an error.
You can set deadline to an explicit value of e.g. 300 to turn tests slower than this into
an error, or you can set it to None to disable this check entirely.
```

pandas/core/strings.py
```diff
@@ -2283,6 +2196,8 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
 
         if isinstance(others, compat.string_types):
             raise ValueError("Did you mean to supply a `sep` keyword?")
+        if sep is None:
```
WillAyd: Not anything you need to change in this PR but if we just overwrite `None` here it would probably make more sense to change the default function signature to simply be `''`

pandas/core/strings.py
```diff
@@ -2320,23 +2239,45 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
                           "'outer'|'inner'|'right'`. The future default will "
                           "be `join='left'`.", FutureWarning, stacklevel=2)
 
+        # if join is None, _get_series_list already aligned indexes
```
WillAyd: Perhaps as a follow up but if the alignment behavior is going to be scattered across these two methods would probably make more sense as a dedicated private method which can be called rather than having various arguments trigger the behavior in various functions

pandas/core/strings.py
```diff
@@ -2320,23 +2239,45 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
...
             others = concat(others, axis=1,
                             join=(join if join == 'inner' else 'outer'),
-                            keys=range(len(others)))
+                            keys=range(len(others)), copy=False)
             data, others = data.align(others, join=join)
             others = [others[x] for x in others]  # again list of Series
 
-        # str_cat discards index
-        res = str_cat(data, others=others, sep=sep, na_rep=na_rep)
+        all_cols = [ensure_object(x) for x in [data] + others]
```
WillAyd: Assuming the index is aligned here can we alternately just concat the columns together and call `ensure_object` on the entire frame instead of using a list comprehension?

pandas/core/strings.py
```diff
@@ -2320,23 +2239,45 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
...
+        na_masks = np.array([isna(x) for x in all_cols])
```
WillAyd: Similar comment above but should be able to use `pd.isnull` instead of a list comp with `isna`

pandas/core/strings.py
```diff
@@ -2320,23 +2239,45 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
...
+            # only necessary if there are actually any NaNs
+            result = np.empty(len(data), dtype=object)
+            np.putmask(result, union_mask, np.nan)
+
+            not_masked = ~union_mask
+            result[not_masked] = cat_core([x[not_masked] for x in all_cols],
+                                          sep)
+        elif na_rep is not None and union_mask.any():
+            # fill NaNs with na_rep in case there are actually any NaNs
+            all_cols = [np.where(nm, na_rep, col)
```
WillAyd: Readability nit but can you change `nm` to `na_mask`? Makes it clearer what is going on (on initial glance misread `nm` to be something `name` related)

WillAyd (changes requested):
Pretty close. Just a few idiomatic / readability things but otherwise looking good

h-vetinari (commented):
@WillAyd 
I refactored `cat_core` to use arrays, so that all the operations you asked for can be done on array level.
This is actually more elegant, and allows to get rid of almost all the list-comprehensions.

(forgot to push changes this morning...)

WillAyd (approved):
lgtm - @jreback

WillAyd (changes requested):
Hmm just noticed CircleCI failed for Py27 - @h-vetinari can you check the errors for compat issues?

h-vetinari:
@WillAyd 

No idea why, the other numpy1.9 didn't fail. In any case, I skipped right to bigger guns (#23062) to solve this. ;-)

pandas/core/strings.py
```diff
@@ -36,114 +37,29 @@
...
-    for x in others:
-        try:
-            if n is None:
-                n = len(x)
-            elif len(x) != n:
-                raise ValueError('All arrays must be same length')
-        except TypeError:
-            raise ValueError('Must pass arrays containing strings to str_cat')
-    return n
+    list_of_columns = np.split(all_cols, all_cols.shape[1], axis=1)
```
WillAyd: FYI I think this is where your compatibility issues start. Running on Py27 with np 1.9 yields:

```
FutureWarning: in the future np.array_split will retain the shape of arrays with a zero size, instead of replacing them by `array([])`, which always has a shape of (0,).
```

WillAyd: Checking their source it looks like this was handled in version 1.11 so if the other change goes first makes sense to resolve the build failures

jreback:
> No idea why, the other numpy1.9 didn't fail. In any case, I skipped right to bigger guns (#23062) to solve this. ;-)

@h-vetinari we generally like to fix issues independently, you can wait on this if you want, but changing deps PR is far from ready.

h-vetinari:
@jreback 
> @h-vetinari we generally like to fix issues independently, you can wait on this if you want, but changing deps PR is far from ready.

IMO no point to spend time working around something that will be obsolete very soon.

WillAyd:
@h-vetinari why don't you just use the np less than 1.11 check here and resize the empty arrays as appropriate for the lower version? 

The other PR seems to be a larger scope than when it first started so may take time. Using a condition here allows this to move independently and is an easy change on the other PR when the time is right

WillAyd (commented):
Looks pretty good. Can you do one last post of ASV results against master to make sure last few updates haven't negatively impacted performance?

WillAyd (approved):
So long as ASVs don't show a regression this lgtm

h-vetinari:
@WillAyd 

ASVs are in line:
```
All benchmarks:

     [296c251d]       [317e042a]
                      <bump_numpy>
         10.9±2ms         12.5±0ms    ~1.14  strings.Cat.time_cat(0, ',', '-', 0.0)
         15.6±0ms       12.5±0.6ms    ~0.80  strings.Cat.time_cat(0, ',', '-', 0.001)
         15.6±2ms       14.1±0.6ms    ~0.90  strings.Cat.time_cat(0, ',', '-', 0.15)
         15.6±6ms       10.9±0.6ms    ~0.70  strings.Cat.time_cat(0, ',', None, 0.0)
         15.6±0ms         15.6±0ms     1.00  strings.Cat.time_cat(0, ',', None, 0.001)
         15.6±0ms         15.6±0ms     1.00  strings.Cat.time_cat(0, ',', None, 0.15)
       10.9±0.8ms       12.5±0.6ms    ~1.14  strings.Cat.time_cat(0, None, '-', 0.0)
       10.9±0.6ms       10.9±0.8ms     1.00  strings.Cat.time_cat(0, None, '-', 0.001)
         15.6±0ms       15.6±0.6ms     1.00  strings.Cat.time_cat(0, None, '-', 0.15)
       14.1±0.8ms       12.5±0.8ms    ~0.89  strings.Cat.time_cat(0, None, None, 0.0)
         15.6±0ms       12.5±0.8ms    ~0.80  strings.Cat.time_cat(0, None, None, 0.001)
         12.5±0ms       14.1±0.6ms    ~1.12  strings.Cat.time_cat(0, None, None, 0.15)
         93.8±0ms          109±8ms    ~1.17  strings.Cat.time_cat(3, ',', '-', 0.0)
         93.8±0ms         93.8±0ms     1.00  strings.Cat.time_cat(3, ',', '-', 0.001)
          109±6ms          102±8ms     0.93  strings.Cat.time_cat(3, ',', '-', 0.15)
         78.1±8ms         78.1±8ms     1.00  strings.Cat.time_cat(3, ',', None, 0.0)
         62.5±6ms         62.5±8ms     1.00  strings.Cat.time_cat(3, ',', None, 0.001)
         62.5±8ms         62.5±6ms     1.00  strings.Cat.time_cat(3, ',', None, 0.15)
         78.1±8ms         78.1±6ms     1.00  strings.Cat.time_cat(3, None, '-', 0.0)
         62.5±8ms         70.3±8ms    ~1.12  strings.Cat.time_cat(3, None, '-', 0.001)
        93.8±10ms        93.8±10ms     1.00  strings.Cat.time_cat(3, None, '-', 0.15)
        78.1±10ms         78.1±8ms     1.00  strings.Cat.time_cat(3, None, None, 0.0)
        78.1±10ms         62.5±8ms    ~0.80  strings.Cat.time_cat(3, None, None, 0.001)
         62.5±0ms         62.5±6ms     1.00  strings.Cat.time_cat(3, None, None, 0.15)
```

Compared to before, there's a bit of speed lost by uselessly appending `''` if `sep=None`, but I know how to deal with that in a follow-up.

h-vetinari:
Nevermind the previous post I immediately deleted, it ran on the wrong branch... Ugh.

There's gonna be a bit of speed lost due to always adding in `sep`, even if it's `''`, but I know how to deal with that in a follow-up (or *maybe* predecessor)...

jreback:
ok this looks fine. pending another asv run, can merge.

h-vetinari:
@WillAyd @jreback 
Unfortunately, I have bad news. I started out in this PR with a very idiomatic solution (see the first couple commits), and it was just too slow.

Here's the ASV for the last commit:
```
All benchmarks:

     [b28cf5aa]       [a97fe67e]
     <master>         <cln_str_cat>
         9.38±0ms         9.38±0ms     1.00  strings.Cat.time_cat(0, ',', '-', 0.0)
       9.38±0.8ms         10.9±0ms    ~1.17  strings.Cat.time_cat(0, ',', '-', 0.001)
         10.9±0ms       12.5±0.6ms    ~1.14  strings.Cat.time_cat(0, ',', '-', 0.15)
       9.38±0.6ms       8.59±0.8ms     0.92  strings.Cat.time_cat(0, ',', None, 0.0)
-      12.5±0.8ms         10.9±0ms     0.87  strings.Cat.time_cat(0, ',', None, 0.001)
-        14.1±0ms         10.9±0ms     0.78  strings.Cat.time_cat(0, ',', None, 0.15)
         9.38±0ms       9.38±0.8ms     1.00  strings.Cat.time_cat(0, None, '-', 0.0)
         9.38±0ms       9.38±0.8ms     1.00  strings.Cat.time_cat(0, None, '-', 0.001)
       10.9±0.8ms       12.5±0.8ms    ~1.14  strings.Cat.time_cat(0, None, '-', 0.15)
         10.9±2ms       7.81±0.8ms    ~0.71  strings.Cat.time_cat(0, None, None, 0.0)
         7.81±0ms       10.9±0.8ms    ~1.40  strings.Cat.time_cat(0, None, None, 0.001)
         10.9±2ms       10.9±0.8ms     1.00  strings.Cat.time_cat(0, None, None, 0.15)
         78.1±8ms         93.8±8ms    ~1.20  strings.Cat.time_cat(3, ',', '-', 0.0)
+        62.5±8ms          109±0ms     1.75  strings.Cat.time_cat(3, ',', '-', 0.001)
+        78.1±8ms          125±8ms     1.60  strings.Cat.time_cat(3, ',', '-', 0.15)
+        46.9±8ms         93.8±0ms     2.00  strings.Cat.time_cat(3, ',', None, 0.0)
         62.5±8ms          109±0ms    ~1.75  strings.Cat.time_cat(3, ',', None, 0.001)
+        46.9±8ms          102±8ms     2.17  strings.Cat.time_cat(3, ',', None, 0.15)
         62.5±0ms         78.1±6ms    ~1.25  strings.Cat.time_cat(3, None, '-', 0.0)
+        46.9±8ms         93.8±0ms     2.00  strings.Cat.time_cat(3, None, '-', 0.001)
+        62.5±8ms          125±6ms     2.00  strings.Cat.time_cat(3, None, '-', 0.15)
+        62.5±8ms         85.9±8ms     1.38  strings.Cat.time_cat(3, None, None, 0.0)
         46.9±6ms         93.8±0ms    ~2.00  strings.Cat.time_cat(3, None, None, 0.001)
         46.9±0ms         93.8±0ms    ~2.00  strings.Cat.time_cat(3, None, None, 0.15)
```
So especially when `others` is not None (all the `pd.concat` and dealing with `DataFrames`) we lose perf.
As a comparison, here's the ASV before the idiomatic changes @WillAyd requested:

```
All benchmarks:

     [b28cf5aa]       [0d3c6d21]
     <master>         <cln_str_cat~2>
         9.38±0ms         9.38±0ms     1.00  strings.Cat.time_cat(0, ',', '-', 0.0)
         9.38±0ms         10.9±0ms    ~1.17  strings.Cat.time_cat(0, ',', '-', 0.001)
         10.9±0ms         12.5±0ms    ~1.14  strings.Cat.time_cat(0, ',', '-', 0.15)
       9.38±0.6ms         9.38±0ms     1.00  strings.Cat.time_cat(0, ',', None, 0.0)
       14.1±0.8ms         10.9±0ms    ~0.78  strings.Cat.time_cat(0, ',', None, 0.001)
       14.1±0.6ms       11.7±0.8ms    ~0.83  strings.Cat.time_cat(0, ',', None, 0.15)
         9.38±0ms         9.38±0ms     1.00  strings.Cat.time_cat(0, None, '-', 0.0)
       9.38±0.6ms       10.9±0.6ms    ~1.17  strings.Cat.time_cat(0, None, '-', 0.001)
         10.9±0ms         12.5±0ms    ~1.14  strings.Cat.time_cat(0, None, '-', 0.15)
         10.9±0ms         7.81±2ms    ~0.71  strings.Cat.time_cat(0, None, None, 0.0)
         9.38±0ms       10.9±0.8ms    ~1.17  strings.Cat.time_cat(0, None, None, 0.001)
       10.9±0.8ms       11.7±0.8ms     1.07  strings.Cat.time_cat(0, None, None, 0.15)
         78.1±0ms         78.1±8ms     1.00  strings.Cat.time_cat(3, ',', '-', 0.0)
         70.3±8ms         78.1±0ms    ~1.11  strings.Cat.time_cat(3, ',', '-', 0.001)
         93.8±8ms         93.8±0ms     1.00  strings.Cat.time_cat(3, ',', '-', 0.15)
         46.9±6ms         78.1±0ms    ~1.67  strings.Cat.time_cat(3, ',', None, 0.0)
         46.9±8ms         78.1±0ms    ~1.67  strings.Cat.time_cat(3, ',', None, 0.001)
         46.9±6ms         62.5±8ms    ~1.33  strings.Cat.time_cat(3, ',', None, 0.15)
         54.7±8ms         54.7±8ms     1.00  strings.Cat.time_cat(3, None, '-', 0.0)
         46.9±8ms         62.5±8ms    ~1.33  strings.Cat.time_cat(3, None, '-', 0.001)
         62.5±0ms        78.1±10ms    ~1.25  strings.Cat.time_cat(3, None, '-', 0.15)
         62.5±6ms         46.9±8ms    ~0.75  strings.Cat.time_cat(3, None, None, 0.0)
         46.9±6ms         62.5±0ms    ~1.33  strings.Cat.time_cat(3, None, None, 0.001)
         46.9±0ms         62.5±6ms    ~1.33  strings.Cat.time_cat(3, None, None, 0.15)
```

This isn't great, but not too bad IMO. Obviously it costs us to uselessly add in `sep=''` just to catch TypeErrors that should already be caught in the `.str`-accessor. I have something in mind there as well.
The direct comparison:
```
                        HEAD~2 vs. master  HEAD vs. master HEAD vs. HEAD~2 HvH2 increase
(3, ',', '-', 0.0)                   1.00             1.20            1.20       +20.00%
(3, ',', '-', 0.001)                 1.11             1.75            1.58       +57.66%
(3, ',', '-', 0.15)                  1.00             1.60            1.60       +60.00%
(3, ',', None, 0.0)                  1.67             2.00            1.20       +19.76%
(3, ',', None, 0.001)                1.67             1.75            1.05        +4.79%
(3, ',', None, 0.15)                 1.33             2.17            1.63       +63.16%
(3, None, '-', 0.0)                  1.00             1.25            1.25       +25.00%
(3, None, '-', 0.001)                1.33             2.00            1.50       +50.38%
(3, None, '-', 0.15)                 1.25             2.00            1.60       +60.00%
(3, None, None, 0.0)                 0.75             1.38            1.84       +84.00%
(3, None, None, 0.001)               1.33             2.00            1.50       +50.38%
(3, None, None, 0.15)                1.33             2.00            1.50       +50.38%
```

Finally, as a general warning about the results right after the run with `SOME BENCHMARKS CHANGED SIGNIFICANTLY` etc.: all **benchmarks with a `~` in their ratio are falsely omitted** from those results. I've opened airspeed-velocity/asv#752 for that.

WillAyd:
IIUC you are saying the last batch of changes requested are causing performance to suffer anywhere between 20-80%? I've been wrong before but at the same time I've never seen instances where applying a function via list comprehension would be significantly faster than applying to the entire frame. Would be helpful if you could profile and debug further

h-vetinari:
> IIUC you are saying the last batch of changes requested are causing performance to suffer anywhere between 20-80%? I've been wrong before but at the same time I've never seen instances where applying a function via list comprehension would be significantly faster than applying to the entire frame.

You do understand correctly. The list-comps themselves (i.e. not counting what goes on inside) are pretty fast, but the more important part is staying in numpy-land, and only going to pandas-land where absolutely necessary. You can check some of the earlier commits (and the ASVs at the top) yourself. In short: working on pandas objects like `DataFrame`, `pd.concat` is expensive compared to pure numpy.

> Would be helpful if you could profile and debug further

I did it in the beginning of this PR already, with the above conclusion. Since IMO "Non-idiomatic" < PERF (by a large margin), this case is settled for me.

pandas/core/strings.py
```diff
@@ -2246,21 +2239,21 @@ def cat(self, others=None, sep=None, na_rep=None, join=None):
                           "'outer'|'inner'|'right'`. The future default will "
                           "be `join='left'`.", FutureWarning, stacklevel=2)
 
-        # concatenate others into DataFrame; need to add keys for uniqueness in
-        # case of duplicate columns (for join is None, all indexes are already
-        # the same after _get_series_list, which forces alignment in this case)
-        others = concat(others, axis=1,
```
h-vetinari: @WillAyd, this is the main reason for the slow-down. For working with a `DataFrame` below (as you wished), we first need to create it with `pd.concat` (expensive). Before, we were only using `pd.concat` if the indices need to be aligned (which they don't in the benchmarks).

h-vetinari (commented):
@WillAyd 
Some further explanation in the diff of the last commit:
```https://github.com/pandas-dev/pandas/pull/22725/commits/e58ec9dfa82a459d9b316b678b77d50fc4901e9e```

**Comment by h-vetinari** [comment]
@jreback @WillAyd 
Can we sacrifice the idiomatic code for perf? Or how do we proceed here?

**Comment by WillAyd** [comment]
`pd.concat` is not expensive. In fact here's a small comparison of the initial part of both code branches:

```python
In [50]: sers = [pd.Series(np.arange(100_000)) for x in range(10)] 
 
In [57]: %%timeit  
    ...: all_cols = [ensure_object(x) for x in sers]                                                                     
38.2 ms ± 228 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [60]: %%timeit  
    ...: df = pd.concat(sers, axis=1) 
    ...: all_cols_df = ensure_object(df)                                                                 
37.8 ms ± 419 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

I *think* the problem may be that `ensure_object` against a DataFrame returns an ndarray of ndarrays with shape of (100_000, 10) whereas the shape of `all_cols` has a shape of (10, 100_000). If you can inspect more closely would be helpful

Unless one of the other devs objects, I would *really* prefer this to be idiomatic from a *pandas* perspective. And just to be clear on what that actually means, list comprehensions over 2D data are NOT idiomatic when operations can be performed against a DataFrame instead. We can always optimize operations with the latter but are limited in regards to the former.

h-vetinari:
There's not much to go on - the last commit shows how little changed: ```https://github.com/pandas-dev/pandas/pull/22725/commits/e58ec9dfa82a459d9b316b678b77d50fc4901e9e``` (I make this format as code to prevent github from mangling the actual comparison url).

* `cat_core` doesn't have to `np.split`
* one less `pd.concat` (because it's now only necessary for different indexes)
* list comp instead of DataFrame operations

I honestly don't see where it's coming from if not the useless concatenating and then splitting again (because, for interleaving `sep` in `cat_core`, we *need* a list of Series anyway). Of course it'd be nicer to have idiomatic code (I tried that right off the bat, but it was 2-3x slower), but ultimately perf should dictate this. All of the cython code isn't pandas-idiomatic either. ;-)

jreback:
@h-vetinari code maintainablinity is actually the most important property of changes
pls use more idiomatic constructions as @WillAyd indicates

jreback:
looks ok to me; @WillAyd merge when satisfied

WillAyd:
Thanks @h-vetinari !

PR #22726, opened 2018-09-16
CLN: io/formats/html.py: refactor

simonjayhawkins:
- [N/A ] follow on from #20341
- [x] tests added / passed
- [N/A ] whatsnew entry


gfyoung:
@simonjayhawkins : Not sure where this was discussed before, but could you explain in your PR description why making `indent` a stateful entity makes sense than leaving it as a parameter to individual methods?

gfyoung:
> i'll report back when i have explored this avenue further.

This sounds to me like we need an issue for this to discuss more.  Could you do that?

simonjayhawkins:
quote from #20341

> ... PR splits the various formatters into different files so that it's easier to refactor a specific formatter.

simonjayhawkins:
@gfyoung : i've reverted the changes to indentation for the moment and added some simpler refactorings to start.

pandas/io/formats/html.py
```diff
@@ -79,7 +79,7 @@ def _write_cell(self, s, kind='td', indent=0, tags=None):
         self.write(u'{start}{rs}</{kind}>'
                    .format(start=start_tag, rs=rs, kind=kind), indent)
 
-    def write_tr(self, line, indent=0, indent_delta=4, header=False,
+    def write_tr(self, line, indent=0, indent_delta=0, header=False,
```
gfyoung: Changing the default argument is an API change.  Could you explain why you are doing this?

simonjayhawkins: I think that the only public method of `HTMLFormatter` should be `write_result()`.

The other methods are currently a mix of private and public. There does not seem to be any consistency.

If `write_tr` should be private, the API change would not be an issue. The `indent_delta` is not left to the default by any calls in the current implementation. `indent_delta=0` would be a cleaner default.

Following on from the discussion regarding making `indent` a stateful entity rather than leaving it as a parameter to individual methods: The `write_tr` method accesses the `bold_rows` instance variable. The value should probably be passed through the method's `nindex_levels` parameter which was probably the intention of adding that parameter to the method.

The other HTML formatting methods `write`, `write_th`, `write_td` and `_write_cell` do not access instance variables with the exception that `write` needs access to `elements` which effectively the output buffer.

jreback:
does this actually change any output?

simonjayhawkins:
@jreback : Are you referring to the comment regarding `indent_delta`, or the removal of the unreachable code and subsequent refactoring?

I've added some additional tests, but not all paths can be tested until the fixes for #22655 have been implemented.

All the tests in `pandas\tests\io\formats\test_to_html.py` currently test the full `df.to_html()` path. Adding additional tests is quite time consuming. I still need to add some more tests to #22655 before I'm happy that the permutations of input parameters with DataFrame truncation has been fully tested.

maybe the way forward is to add unit tests for `HTMLFormatter`?

pandas/io/formats/html.py
jreback: yeah this is ok, as its internal. I agree its less that clear what is external / internal (and we don't usually use _ prefixed). If you want to indicate in the doc-string of an external API would be ok.

jreback:
lgtm. @WillAyd if you'd have a quick look.

pandas/tests/io/formats/test_to_html.py
```diff
@@ -1905,6 +2013,20 @@ def test_to_html_multiindex_max_cols(self):
         </table>""")
         assert result == expected
 
+    @pytest.mark.parametrize('header', [True])
```
WillAyd: If these are always just True what's the point of parametrization?

pandas/tests/io/formats/test_to_html.py
```diff
@@ -1905,6 +2013,20 @@ def test_to_html_multiindex_max_cols(self):
...
+    @pytest.mark.parametrize('index', [True])
+    @pytest.mark.parametrize('index_names', [True])
+    def test_to_html_index_names(self, expected_output,
+                                 idx_type, col_idx_type, index, header,
+                                 index_names):
+        df = DataFrame(np.zeros((2, 2), dtype=int),
+                       index=idx_type, columns=col_idx_type)
+        result = df.to_html(index=index, header=header,
+                            index_names=index_names)
+        assert result == expected_output(idx_type, col_idx_type,
```
WillAyd: Can you do `expected = expected_output(...` on the line preceding this then change the assertion to simply `result == expected`?

pandas/tests/io/formats/test_to_html.py
```diff
@@ -22,6 +23,113 @@
...
+    else:
+        return index.name is not None
+
+
+def _index_init_params(index):
+    return (index.nlevels > 1,
+            _has_names(index))
+
+
+@pytest.fixture
```
WillAyd: Is this going to be used outside this module? If not then can probably just be a global function instead of a fixture. If so then I think would make more sense to move fixtures to a conftest for this package

pandas/tests/io/formats/test_to_html.py
```diff
@@ -22,6 +23,113 @@
...
+         'columns_unnamed_multi',
+         'columns_named_multi'])
+def col_idx_type(request):
+    # standard and multiIndex columns index, named and unnamed
+    # used for basic table alignment tests
+    return request.param
+
+
+@pytest.fixture
+def expected_output(expected_html):
```
WillAyd: Similar comment as for `expected_html`

WillAyd (changes requested):

pandas/tests/io/formats/test_to_html.py
simonjayhawkins: The tests added here are the same tests added for regression testing of #22655. If either the `index`, `header` or `index_names` parameter is `False`, the tests will currently fail.

So the tests for regression testing of the refactoring only covers the `index=True`, `header=True`, `index_names=True` codepath.

I appreciate that the changes to the tests are not ready for merging into master but the tests can be removed from this PR before approving the refactoring.

WillAyd: I see. If that's the case just go ahead and remove here and keep as part of #22655. Relatively minor but would keep a cleaner and more logical history of changes that way

pandas/tests/io/formats/test_to_html.py
simonjayhawkins: It will not be of use outside this module. It uses `read_file` which uses the `datapath` fixture. So does it not need to be a fixture or am I doing something wrong?

pandas/tests/io/formats/test_to_html.py
simonjayhawkins: no problem. will apply this change to #22655.

pandas/tests/io/formats/test_to_html.py
simonjayhawkins: I think it would be a good idea if the methodology used here, if approved, is then used to move all the inline HTML from the other tests into separate files. Would be easier to check the expected output of other tests. Would be easier to re-use the expect outputs of other tests. And would allow much larger outputs/comprehensive tests to be used in future tests without cluttering the test file.

I'll move it all into one fixture for the moment. It can always be separated out again in the future.

simonjayhawkins: In  #22655, I've added a issue specific test as well as the parametric regression tests. So I am already using `expected_html` in more than one test and will need to leave it as it is.

jreback:
can you merge master on this.

jreback:
@WillAyd comments?

pandas/io/formats/html.py
```diff
@@ -201,26 +201,6 @@ def _write_header(self, indent):
...
-                row = [''] * (self.frame.index.nlevels - 1)
-            else:
-                row = []
-
-            if isinstance(self.columns, ABCMultiIndex):
-                if self.fmt.has_column_names and self.fmt.index:
-                    row.append(single_column_table(self.columns.names))
-                else:
-                    row.append('')
-                style = "text-align: {just};".format(just=self.fmt.justify)
```
WillAyd: Relatively nuanced but in the case of a MultiIndex isn't this formatting the very first column row different from the rest? If so is that accounted for in the refactor?

WillAyd (changes requested):
Generally looks good - one question around the handling of MI but overall a lot simpler. Nice job!

pandas/io/formats/html.py
simonjayhawkins: This `if` clause is unreachable since `_column_header()` was only called from the `else` clause of another `if isinstance(self.columns, ABCMultiIndex):` so the styling is not accounted for in this refactor.

WillAyd (approved):
@jreback back to you

jreback:
thanks @simonjayhawkins

PR #22731, opened 2018-09-17
DOC: Fix Series nsmallest and nlargest docstring/doctests

Moisan:
- [x] tests added / passed

Based on #22459. Fix the docstring for Series.nsmallest and Series.nlargest. I did both together since the same example could be used. I removed both from the skipped doctests in `ci/doctests.sh`.

I also found out about the "all" option for the `keep` parameter which was surprisingly not documented.

### DIFF
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -2874,7 +2874,7 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         Nauru        11300
         Tuvalu       11300
         Anguilla     11300
-        Iceland     337780
+        Iceland     337000
         dtype: int64
 
         >>> s.nsmallest(3)
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -2745,7 +2745,7 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         ----------
         n : int, default 5
             Return this many descending sorted values.
-        keep : str, default 'first'
+        keep : {'first', 'last', 'all'}, default 'first'
             When there are duplicate values that cannot all fit in a
             Series of `n` elements:
             - ``first`` : take the first occurrences based on the index order
@@ -2766,6 +2766,8 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         See Also
         --------
         Series.nsmallest: Get the `n` smallest elements.
+        Series.sort_values: Sort Series by values.
+        Series.head: Return the first `n` rows.
 
         Examples
         --------
@@ -2836,9 +2838,9 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         Parameters
         ----------
-        n : int
+        n : int, default 5
             Return this many ascending sorted values.
-        keep : str, default 'first'
+        keep : {'first', 'last', 'all'}, default 'first'
             When there are duplicate values that cannot all fit in a
             Series of `n` elements:
             - ``first`` : take the first occurrences based on the index order
@@ -2859,6 +2861,8 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
         See Also
         --------
         Series.nlargest: Get the `n` largest elements.
+        Series.sort_values: Sort Series by values.
+        Series.head: Return the first `n` rows.
 
         Examples
         --------
--- a/pandas/core/series.py
+++ b/pandas/core/series.py
@@ -2743,17 +2743,20 @@ class Series(base.IndexOpsMixin, generic.NDFrame):
 
         Parameters
         ----------
-        n : int
-            Return this many descending sorted values
-        keep : {'first', 'last'}, default 'first'
-            Where there are duplicate values:
-            - ``first`` : take the first occurrence.
-            - ``last`` : take the last occurrence.
+        n : int, default 5
+            Return this many descending sorted values.
+        keep : str, default 'first'
+            When there are duplicate values that cannot all fit in a
+            Series of `n` elements:
+            - ``first`` : take the first occurrences based on the index order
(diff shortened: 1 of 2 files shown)


pandas/core/series.py
```diff
@@ -2762,23 +2765,56 @@ def nlargest(self, n=5, keep='first'):
...
+
+        >>> s.nlargest()
+        France      65000000
+        Italy       59000000
+        Malta         434000
+        Maldives      434000
+        Brunei        434000
+        dtype: int64
+
+        >>> s.nlargest(3)
```
WillAyd: I think it would be nice to have just a quick one-liner to highlight the difference between this and the subsequent example

WillAyd (changes requested):
Minor comments

pandas/core/series.py
Moisan: Do you mean a comment at the end of the line?

WillAyd: No just some text in between the examples to call out what the user should be looking at

pandas/core/series.py
```diff
@@ -2743,17 +2743,20 @@ def nlargest(self, n=5, keep='first'):
...
-            - ``last`` : take the last occurrence.
+        n : int, default 5
+            Return this many descending sorted values.
+        keep : str, default 'first'
+            When there are duplicate values that cannot all fit in a
+            Series of `n` elements:
+            - ``first`` : take the first occurrences based on the index order
+            - ``last`` : take the last occurrences based on the index order
+            - ``all`` : keep all occurrences. This can result in a Series of
+                size larger than `n`.
```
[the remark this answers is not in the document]
Moisan: I confirm that the validation fails if the last period is not present.

pandas/core/series.py
```diff
@@ -2762,23 +2765,69 @@ def nlargest(self, n=5, keep='first'):
...
+        Brunei will be kept since it is the last with value 434000 based on
+        the index order.
+
+        >>> s.nlargest(3, keep='last')
+        France      65000000
+        Italy       59000000
+        Brunei        434000
+        dtype: int64
+
+        The n largest elements where n=3 with all duplicates kept. Note that the
```
TomAugspurger: Looks like this line is too long:  https://travis-ci.org/pandas-dev/pandas/jobs/429733867#L3699

pandas/core/series.py
```diff
@@ -2762,23 +2765,69 @@ def nlargest(self, n=5, keep='first'):
...
+        Maldives      434000
+        Brunei        434000
+        Iceland       337000
+        Nauru          11300
+        Tuvalu         11300
+        Anguilla       11300
+        Monserat        5200
+        dtype: int64
+
+        The n largest elements where n=5 by default.
```
TomAugspurger: single backticks around `n`. Double backticks around `n=5`.

pandas/core/series.py
```diff
@@ -2762,23 +2765,69 @@ def nlargest(self, n=5, keep='first'):
...
+
+        >>> s.nlargest()
+        France      65000000
+        Italy       59000000
+        Malta         434000
+        Maldives      434000
+        Brunei        434000
+        dtype: int64
+
+        The n largest elements where n=3. Default keep value is 'first' so
```
TomAugspurger: Same. Single backticks around `keep`.

pandas/core/series.py
```diff
@@ -2762,23 +2765,69 @@ def nlargest(self, n=5, keep='first'):
...
+        Malta will
+        be kept.
```
TomAugspurger: Can be on the same line as "Malta will"

pandas/core/series.py
```diff
@@ -2762,23 +2765,69 @@ def nlargest(self, n=5, keep='first'):
...
+
+        >>> s.nlargest(3)
+        France    65000000
+        Italy     59000000
+        Malta       434000
+        dtype: int64
+
+        The n largest elements where n=3 and keeping the last duplicates.
```
TomAugspurger: backticks.

WillAyd (approved):
Think this is a nice update so lgtm

pandas/core/series.py
```diff
@@ -2789,16 +2837,19 @@ def nsmallest(self, n=5, keep='first'):
         Parameters
         ----------
         n : int
```
datapythonista: can you add the default here?

pandas/core/series.py
```diff
@@ -2743,17 +2743,20 @@ def nlargest(self, n=5, keep='first'):
...
         ----------
-        n : int
-            Return this many descending sorted values
-        keep : {'first', 'last'}, default 'first'
-            Where there are duplicate values:
-            - ``first`` : take the first occurrence.
-            - ``last`` : take the last occurrence.
+        n : int, default 5
+            Return this many descending sorted values.
+        keep : str, default 'first'
```
datapythonista: I think it's better to keep `{'first', 'last', 'all'}`, as I don't think there is any other value allowed. That applies to both docstrings.

pandas/core/series.py
```diff
@@ -2762,23 +2765,68 @@ def nlargest(self, n=5, keep='first'):
 
         See Also
         --------
-        Series.nsmallest
+        Series.nsmallest: Get the `n` smallest elements.
```
datapythonista: I'd add `sort_values` and `head` here, as they're mentioned in the notes. In both docstrings.

datapythonista (commented):
Great changes, suggested couple of small things, but other than that looks really good.

datapythonista (approved):
lgtm. Excellent work @Moisan, thank you!

PR #22730, opened 2018-09-17
CLN: res/exp and GH references in frame tests

h-vetinari:
EDIT2: reverted import-related clean-up due to https://github.com/pandas-dev/pandas/pull/22730#issuecomment-423844217

EDIT: after review by @WillAyd, the fixturization was split out to #22735, #22736 and #22738. Now contains only the following clean-ups, also for #22733:
* Replace instances of pd. with direct imports (from review https://github.com/pandas-dev/pandas/pull/21899#discussion_r202524353):
  > don’t import pd, directly import instead
* Replace direct use of `assert_..._equal` with the `tm.assert_..._equal`
    (from review https://github.com/pandas-dev/pandas/pull/21899#discussion_r202524374):
  > use tm; don’t import assert_series_equal
* Clean up usage of `exp/res` (from review https://github.com/pandas-dev/pandas/pull/22733#discussion_r218306730):
  > Here you can use `result` and `expected` in spite of the existing code using `exp`
* Unify references to github issues

Original post below. [END EDIT]

- [x] 3 modules of #22471
- [x] tests added / passed
- [x] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`

I split the commits per module, and (where possible) into fixturizing and cleaning up the imports. Other notes:
* For `test_apply`, I added an another often-used fixture
* In translating the quasi-fixtures from `TestData` to `conftest` in #22236, I sorted the dtypes for the columns of `mixed_float_frame` and `mixed_int_frame`, which turns out to have been a mistake. This is reverted here to be a true translation of the attribute of `TestData`. Otherwise, tests in the newly fixturized `test_arithmetic.py` would fail.

En passant, I also unified GH issue references.

PR #22733, opened 2018-09-17
TST/CLN: Fixturize frame/test_analytics

h-vetinari:
- [x] 1 step closer towards #22471
- [x] tests added / passed

This module is **much** harder to fixturize than e.g. #22236 or #22730, mainly due to the class methods `_check_stat_op` and `_check_bool_op`, which, despite having an argument for the frame they're testing, are *also* testing on other quasi-fixtures of `TestData`. Since I can't import directly from `frame/conftest` without getting `RemovedInPytest4Warnings`, I made theses fixtures explicit arguments of the respective methods.

Furthermore, I extracted two fixtures that basically correspond to those methods being called without a `frame` argument, and added them to `conftest`.

The larger question is how to avoid all these redundant calls being made (e.g. in `test_max`), and how `_check_stat_op` / `_check_stat_op` should be properly split up into several tests/parametrizations. So I don't view this PR as ready, but needing discussion regarding how to best proceed.


pandas/tests/frame/test_analytics.py
```diff
@@ -12,65 +12,58 @@
...
-        if not check_minp:
-            correls = self.frame.corr(method=method)
-            exp = self.frame['A'].corr(self.frame['C'], method=method)
-            tm.assert_almost_equal(correls['A']['C'], exp)
-        else:
-            result = self.frame.corr(min_periods=len(self.frame) - 8)
-            expected = self.frame.corr()
-            expected.loc['A', 'B'] = expected.loc['B', 'A'] = nan
-            tm.assert_frame_equal(result, expected)
+    def _check_method(self, frame, method='pearson'):
```
h-vetinari: I had a typo in the else-branch of `check_minp`, and all tests still ran through, so I'm suggesting to remove this unused branch.

pandas/tests/frame/test_analytics.py
```diff
@@ -815,59 +826,54 @@ def wrapper(x):
...
                                    check_less_precise=check_less_precise)
 
         # check dtypes
         if check_dtype:
-            lcd_dtype = frame.values.dtype
+            lcd_dtype = main_frame.values.dtype
             assert lcd_dtype == result0.dtype
             assert lcd_dtype == result1.dtype
 
-        # result = f(axis=1)
```
h-vetinari: Uncommenting these lines leads to some failures in `test_sum`, `test_median`, and `test_sem`, but they seem to be tested more thoroughly directly above (with the correct wrappers and kwargs), so I think they should be removed.

pandas/tests/frame/test_analytics.py
```diff
@@ -815,59 +826,54 @@ def wrapper(x):
...
-        # comp = frame.apply(alternative, axis=1).reindex(result.index)
-        # assert_series_equal(result, comp)
-
         # bad axis
         tm.assert_raises_regex(ValueError, 'No axis named 2', f, axis=2)
         # make sure works on mixed-type frame
-        getattr(self.mixed_frame, name)(axis=0)
-        getattr(self.mixed_frame, name)(axis=1)
+        getattr(float_string_frame, name)(axis=0)
```
h-vetinari: From here on I'm thinking these tests should live in a separate class method, or be broken out some other way.

pandas/tests/frame/test_analytics.py
```diff
@@ -543,107 +537,124 @@ def test_stat_operators_attempt_obj_array(self, method):
...
-    def test_max(self):
+    def test_max(self, float_frame_with_na, int_frame,
+                 float_frame, float_string_frame):
         with warnings.catch_warnings(record=True):
-            self._check_stat_op('max', np.max, check_dates=True)
-        self._check_stat_op('max', np.max, frame=self.intframe)
+            self._check_stat_op('max', np.max, float_frame_with_na,
+                                float_frame, float_string_frame,
+                                check_dates=True)
+        self._check_stat_op('max', np.max, int_frame, float_frame,
```
h-vetinari: In tests like this one, all the parts of `_check_stat_op` that depend on `float_frame` and `float_string_frame`  get tested twice unnecessarily

WillAyd (changes requested):
Similar comments to #22730 - if you can keep PR focused on fixtures and keep changes to name spacing / comments for another PR would be very helpful for review process

pandas/tests/frame/test_analytics.py
```diff
@@ -23,54 +23,47 @@
...
-            tm.assert_almost_equal(correls['A']['C'], exp)
-        else:
-            result = self.frame.corr(min_periods=len(self.frame) - 8)
-            expected = self.frame.corr()
-            expected.loc['A', 'B'] = expected.loc['B', 'A'] = nan
-            tm.assert_frame_equal(result, expected)
+    def _check_method(self, frame, method='pearson'):
+        correls = frame.corr(method=method)
+        exp = frame['A'].corr(frame['C'], method=method)
+        tm.assert_almost_equal(correls['A']['C'], exp)
```
WillAyd: Here you can use `result` and `expected` in spite of the existing code using `exp`

pandas/tests/frame/test_analytics.py
```diff
@@ -2078,9 +2079,6 @@ def test_n_error(self, df_main_dtypes, nselect_method, columns):
         col = columns[1]
         error_msg = self.dtype_error_msg_template.format(
             column=col, method=nselect_method, dtype=df[col].dtype)
-        # escape some characters that may be in the repr
-        error_msg = (error_msg.replace('(', '\\(').replace(")", "\\)")
```
WillAyd: Why was this removed?

pandas/tests/frame/test_analytics.py
```diff
@@ -784,17 +799,13 @@ def alt(x):
         assert kurt.name is None
         assert kurt2.name == 'bar'
 
-    def _check_stat_op(self, name, alternative, frame=None, has_skipna=True,
+    def _check_stat_op(self, name, alternative, main_frame, float_frame,
```
WillAyd: Is there any risk here in using arguments identical to the fixture names? Assuming since this method itself is not a test that pytest won't try to inject the fixture here but not sure if that will always be the case

WillAyd (changes requested):
Looking much better - thanks!

pandas/tests/frame/test_analytics.py
h-vetinari: Um, first you ask me to remove all orthogonal changes, and then you ask me to add... orthogonal changes? Same happened with the import clean-up other PRs, but when I preempted them here, you ask me to remove them - can you understand how exasperating this back-and-forth is?

pandas/tests/frame/test_analytics.py
h-vetinari: Changed. Did you seem my comment about breaking up this function?

pandas/tests/frame/test_analytics.py
h-vetinari: I removed it as part of trying to hunt down some DeprecationWarnings for unescaped brackets (which I believe are somehow coming from pytest), and the tests still ran through. It is IMO an unnecessary misdirection that makes the test harder to read, but I reverted it.

h-vetinari:
@gfyoung @jreback @WillAyd @TomAugspurger
Weird error on Appveyor, could some one please restart?

```
Build started
git clone -q https://github.com/pandas-dev/pandas.git C:\projects\pandas
git fetch -q origin +refs/pull/22733/merge:
git checkout -qf FETCH_HEAD
Running Install scripts
if ($env:APPVEYOR_PULL_REQUEST_NUMBER -and $env:APPVEYOR_BUILD_NUMBER -ne ((Invoke-RestMethod ` https://ci.appveyor.com/api/projects/$env:APPVEYOR_ACCOUNT_NAME/$env:APPVEYOR_PROJECT_SLUG/history?recordsNumber=50).builds | ` Where-Object pullRequestId -eq $env:APPVEYOR_PULL_REQUEST_NUMBER)[0].buildNumber) { ` throw "There are newer queued builds for this pull request, failing early." }
Cannot index into a null array.
At line:1 char:5
+ if ($env:APPVEYOR_PULL_REQUEST_NUMBER -and $env:APPVEYOR_BUILD_NUMBER ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : NullArray
 
Command executed with exception: Cannot index into a null array.
```

h-vetinari:
@WillAyd Green :)

WillAyd (approved):
lgtm - @jreback

pandas/tests/frame/test_analytics.py
```diff
@@ -507,19 +501,20 @@ def test_nunique(self):
...
-                            has_numeric_only=True, check_dtype=False,
-                            check_less_precise=True)
+                            mixed_float_frame.astype('float32'), float_frame,
+                            float_string_frame, has_numeric_only=True,
+                            check_dtype=False, check_less_precise=True)
 
-    @pytest.mark.parametrize(
-        "method", ['sum', 'mean', 'prod', 'var',
-                   'std', 'skew', 'min', 'max'])
+    @pytest.mark.parametrize('method', ['sum', 'mean', 'prod', 'var',
```
jreback: in follow, can use these fixtures (may need to make small changes) when #22762 is merged

pandas/tests/frame/test_analytics.py
jreback: hmm, can you have a look at the git history and see if you can figure out why. otherwise ok to remove.

pandas/tests/frame/test_analytics.py
```diff
@@ -788,17 +803,14 @@ def alt(x):
         assert kurt.name is None
         assert kurt2.name == 'bar'
 
-    def _check_stat_op(self, name, alternative, frame=None, has_skipna=True,
+    # underscores added to distinguish argument names from fixture names
```
jreback: can this just be a module level function now? maybe rename to assert_stat_ops

pandas/tests/frame/test_analytics.py
```diff
@@ -1191,9 +1199,13 @@ def test_idxmax(self):
     # ----------------------------------------------------------------------
     # Logical reductions
 
-    def test_any_all(self):
-        self._check_bool_op('any', np.any, has_skipna=True, has_bool_only=True)
-        self._check_bool_op('all', np.all, has_skipna=True, has_bool_only=True)
+    def test_any_all(self, bool_frame_with_na, float_string_frame):
```
jreback: can parameterize on ['all', 'any'] (use ``getattr(np, name)`` in side

jreback (changes requested):
looks good, a couple of comments

pandas/tests/frame/test_analytics.py
h-vetinari: I made this a module-level function, and broke it up into `assert_stat_op_calc` and `assert_stat_op_api`. This also removes the redundancy mentioned above.

h-vetinari (commented):
Broke up `_check_stat_op` and `_check_bool_op` into two module-level functions each, which should be much cleaner in terms of needed fixtures/kwargs.

There is still some redundancy between the new `assert_stat_op_calc` and `assert_bool_op_calc`, but that's also true of the current state and therefore something for a follow-up -- this PR should (as far as possible) just be about fixturization.

h-vetinari:
@WillAyd @jreback @gfyoung 

Reverted the underscored for the function again after cross-review in #22730:
@jreback: 
> actually don't think you need to do this as pytest doesn't care

@h-vetinari: 
> This is from #22733, where this was an review request from @WillAyd.
>> Is there any risk here in using arguments identical to the fixture names? Assuming since this method itself is not a test that pytest won't try to inject the fixture here but not sure if that will always be the case

@gfyoung:
> Fixture injection should only happen when we have a function called `test_*`. pytest wouldn't even collect that function at all. Thus, I agree with @jreback to drop underscores.
If we're wrong, IMO that's a design flaw.

pandas/tests/frame/test_analytics.py
```diff
@@ -23,54 +23,192 @@
 
 import pandas.util.testing as tm
 import pandas.util._test_decorators as td
-from pandas.tests.frame.common import TestData
 
 
-class TestDataFrameAnalytics(TestData):
+def assert_stat_op_calc(opname, alternative, main_frame, has_skipna=True,
```
jreback: is there a reason you split up these to calc/api tests now? its very hard to tell if anything actually changed. I am concerned that something DID change even accidently. can you just do this as a copy-paste change of the original function.

jreback (changes requested):

pandas/tests/frame/test_analytics.py
h-vetinari: Yes, because they perform different checks, need different fixtures, and introduced the redundancy mentioned further up the thread (of checking the api-part twice in e.g. `test_max`).

I did not remove any single line of the functions.

> can you just do this as a copy-paste change of the original function.

you mean as a separate commit?

h-vetinari: > no i mean do a pre-cursor PR that moves the functions, but doesn't change them. Then a followon to show those changes. moving & changing need to be separate steps. It should be easy to see what actually changed.

Why not a separate commit...? You can review commits individually. Plus, this is essentially linked with the whole fixturization effort, which would have to come first (and your request to make `_check_stat_op` a module-level function is strictly speaking orthogonal to this PR).

I'll make some nice modular commits for you to review in sequence.

jreback: you can try, but its much easier to review a separate PR to be honest

h-vetinari: I pushed some commits that just go until moving `_check_stat_op` and `_check_bool_op` and making them run. I'll push the rest of the commits once the CI passes.

h-vetinari:
@jreback 
So the CI was successful except a `CondaHTTPError: HTTP 504 GATEWAY TIME-OUT` for `py36_locale_slow`. Pushing the rest of the commits.

h-vetinari:
@jreback 
PTAL

I know you wanted separation (just moving the functions to module-level), but that's not possible without the fixturization that's carried out in this PR. However, I've really rolled out the red carpet here, the commits are super modular and easy to digest...

h-vetinari:
@jreback

If you start here (```https://github.com/pandas-dev/pandas/pull/22733/commits/7ac476ec50d3a9b44112c078a61f9455efe93c07```), the commits are very easy to follow, I promise. :)

h-vetinari:
@jreback 
>  I really prefer to do the diff using the tools which git provides

If you don't like the github-tools that allow comparing separate commits (which I linked; and where you can step through one by one), there's always `git diff <hash1> <hash2>`.

Your request of moving the function was orthogonal to this PR - I still accomodated it. If anything could be split off, it's that (i.e. revert the last 4 commits). Similarly, the whole fixturization effort was your request in #22236 (orthogonal there as well) - it's not something I'm gonna fight for. Feel free to reopen if you wanna have a look again.

pandas/tests/frame/test_analytics.py
```diff
@@ -1839,9 +1841,10 @@ def test_clip_against_series(self, inplace):
         (0, [[2., 2., 3.], [4., 5., 6.], [7., 7., 7.]]),
         (1, [[2., 3., 4.], [4., 5., 6.], [5., 6., 7.]])
     ])
-    def test_clip_against_list_like(self, inplace, lower, axis, res):
+    def test_clip_against_list_like(self, simple_frame,
+                                    inplace, lower, axis, res):
         # GH #15390
-        original = self.simple.copy(deep=True)
+        original = simple_frame.copy(deep=True)
```
jreback: note for later to clean up the names of these fixtures to be meaningful (e.g. float_frame is, simple_frame not so much, prob need to wait to see how much more things ike this are used)

jreback:
@h-vetinari ok this looks ok thanks.

jreback (approved):

h-vetinari:
@jreback 
Thanks.

> note for later to clean up the names of these fixtures to be meaningful (e.g. float_frame is, simple_frame not so much, prob need to wait to see how much more things ike this are used)

OK, this was part of the process in #22236. Originally it was just called `simple` - `simple_frame` was the closest in spirit. I'd argue it should be renamed ASAP (and update the "translation guide" in #22471), before more modules get translated. What name would you like to have?

PR #22736, opened 2018-09-17
Fixturize tests/frame/test_arithmetic

h-vetinari:
Split off from #22730 as per review from @WillAyd 

The changes in `conftest.py` are due to the following:
> In translating the quasi-fixtures from `TestData` to `conftest` in #22236, I sorted the dtypes for the columns of `mixed_float_frame` and `mixed_int_frame`, which turns out to have been a mistake. This is reverted here to be a true translation of the attribute of `TestData`. Otherwise, tests in the newly fixturized `test_arithmetic.py` would fail.


pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,97 @@ def test_df_add_flex_filled_mixed_dtypes(self):
              'B': ser * 2})
         tm.assert_frame_equal(result, expected)
 
-    def test_arith_flex_frame(self):
-        seriesd = tm.getSeriesData()
-        frame = pd.DataFrame(seriesd).copy()
+    @pytest.mark.parametrize('op', ['add', 'sub', 'mul', 'div', 'truediv',
```
jreback: can you change these to use the fixture: all_arithmetic_operators

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,97 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
-                                  'B': np.ones(len(intframe), dtype='uint64'),
-                                  'C': intframe['C'].copy().astype('uint8'),
-                                  'D': intframe['D'].copy().astype('int64')})
-
-        # force these all to int64 to avoid platform testing issues
-        intframe = pd.DataFrame({c: s for c, s in intframe.items()},
-                                dtype=np.int64)
-
-        ops = ['add', 'sub', 'mul', 'div', 'truediv', 'pow', 'floordiv', 'mod']
         if not PY3:
```
jreback: you prob don't need this (when using the fixture)

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,97 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
+        exp = f(float_frame, 2 * float_frame)
+        tm.assert_frame_equal(result, exp)
+
+        # vs mix float
+        result = getattr(mixed_float_frame, op)(2 * mixed_float_frame)
+        exp = f(mixed_float_frame, 2 * mixed_float_frame)
+        tm.assert_frame_equal(result, exp)
+        _check_mixed_float(result, dtype=dict(C=None))
+
+        # vs mix int
```
jreback: might be more clear to move these to a separate test (from here down)

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,97 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
-        # res_div = frame.div(2 * frame)
-
-        # tm.assert_frame_equal(res_add, frame + frame)
-        # tm.assert_frame_equal(res_sub, frame - frame)
-        # tm.assert_frame_equal(res_mul, frame * frame)
-        # tm.assert_frame_equal(res_div, frame / (2 * frame))
-
-        const_add = frame.add(1)
-        tm.assert_frame_equal(const_add, frame + 1)
+        # ndim >= 3
```
jreback: here and down could be in the top-section, e.g. the first test as they test all ops (or another tests is even better, with the fixture this becomes easy)

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,97 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
+        ndim_5 = np.ones(float_frame.shape + (3, 4, 5))
```
jreback: can you add a comment here on what this is testing

jreback (changes requested):

jreback:
@h-vetinari ping when updated

pandas/tests/frame/test_arithmetic.py
h-vetinari: will do

pandas/tests/frame/test_arithmetic.py
h-vetinari: I don't want to get into orthogonal changes here. I haven't changed anything about this except removing the indentation. This goes for the test split requested below as well.

pandas/tests/frame/test_arithmetic.py
h-vetinari: See above - I did not write the test; just removed indentation

h-vetinari:
@jreback 

I though your request for `all_arithmetic_operators` would not change anything, but it includes more operators than previously tested - I get errors:
```
AttributeError: module 'operator' has no attribute '__rmod__'  [etc.]
```

I've skipped those tests now.

h-vetinari:
@ jreback
> might be more clear to move these to a separate test (from here down)

> can you add a comment here on what this is testing

OK, I relented, and split up and cleaned up that huge test. Also the `skips` for the `r`-versions of the ops aren't necessary anymore. They were tested for `add/sub/mult` before, and I extended that to all ops now.

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,92 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
-                    # no overflow in the uint
-                    dtype = None
-                    if op in ['sub']:
-                        dtype = dict(B='uint64', C=None)
-                    elif op in ['add', 'mul']:
-                        dtype = dict(C=None)
-                    tm.assert_frame_equal(result, exp)
-                    _check_mixed_int(result, dtype=dtype)
-
-                    # rops
```
h-vetinari: rops were only tested within that if-block. Now they're all tested

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,92 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
-                    result = getattr(mixed_float, op)(2 * mixed_float)
-                    exp = f(mixed_float, 2 * mixed_float)
-                    tm.assert_frame_equal(result, exp)
-                    _check_mixed_float(result, dtype=dict(C=None))
-
-                    result = getattr(intframe, op)(2 * intframe)
-                    exp = f(intframe, 2 * intframe)
-                    tm.assert_frame_equal(result, exp)
-
-                    # vs mix int
```
h-vetinari: For some reason, this part was tested twice. only once in cleaned-up version

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,92 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
+        exp = f(float_frame, 2 * float_frame)
+        tm.assert_frame_equal(result, exp)
+
+        # vs mix float
+        result = getattr(mixed_float_frame, op)(2 * mixed_float_frame)
+        exp = f(mixed_float_frame, 2 * mixed_float_frame)
+        tm.assert_frame_equal(result, exp)
+        _check_mixed_float(result, dtype=dict(C=None))
+
+    @pytest.mark.parametrize('op', ['__add__', '__sub__', '__mul__'])
```
h-vetinari: Broke up the test like you wanted

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,92 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
+        result = getattr(int_frame, op)(2 * int_frame)
+        exp = f(int_frame, 2 * int_frame)
+        tm.assert_frame_equal(result, exp)
+
+    def test_arith_flex_frame_corner(self, all_arithmetic_operators,
+                                     float_frame):
+
+        op = all_arithmetic_operators
+
+        # Check that arrays with dim >= 3 raise
```
h-vetinari: > can you add a comment here on what this is testing

Added (and expanded)

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,92 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
+        for dim in range(3, 6):
+            arr = np.ones((1,) * dim)
             msg = "Unable to coerce to Series/DataFrame"
             with tm.assert_raises_regex(ValueError, msg):
-                f(frame, ndim_5)
```
h-vetinari: This part is redundant (via the `f` of the `op`) is testing the same thing as getting the `op` from the frame directly

h-vetinari (commented):
@ jreback
> might be more clear to move these to a separate test (from here down)

> can you add a comment here on what this is testing

OK, I relented, and split up and cleaned up that huge test. Also the `skips` for the `r`-versions of the ops aren't necessary anymore. They were tested for `add/sub/mult` before, and I extended that to all ops now.

h-vetinari:
@jreback Green

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,92 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
-                            dtype = dict(C=None)
-                        tm.assert_frame_equal(result, exp)
-                        _check_mixed_int(result, dtype=dtype)
-            except:
-                printing.pprint_thing("Failing operation %r" % op)
-                raise
-
-            # ndim >= 3
-            ndim_5 = np.ones(frame.shape + (3, 4, 5))
+        op = all_arithmetic_operators  # one instance of parametrized fixture
```
jreback: put comments on the prior line

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,92 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
+        tm.assert_frame_equal(result, exp)
+        _check_mixed_float(result, dtype=dict(C=None))
+
+    @pytest.mark.parametrize('op', ['__add__', '__sub__', '__mul__'])
+    def test_arith_flex_frame_mixed(self, op, int_frame, mixed_int_frame,
+                                    float_frame, mixed_float_frame):
+
+        if op.startswith('__r'):
+            # get op without "r" and invert it
+            tmp = getattr(operator, op[:2] + op[3:])
```
jreback: make this a named function to avoid the tmp something like

```
def f(x, y):
    # comment
    op = op.replace('__r', '__')
    return getattr(operator, op)

**Review by jreback — CHANGES_REQUESTED** [review]
ok test looks a lot nicer. small change, ping on green.

File: pandas/tests/frame/test_arithmetic.py (hunk 12 above)
Comment (jreback) [reply pandas/tests/frame/test_arithmetic.py]: this comment applies to the prior test actually. I don't think you need this here at all, as op is just add, mul, sub.

did you mean to test the reverse of these as well?

**Review by jreback — CHANGES_REQUESTED** [review]

File: pandas/tests/frame/test_arithmetic.py (hunk 12 above)
Comment (h-vetinari) [reply pandas/tests/frame/test_arithmetic.py]: > make this a named function to avoid the tmp something like [...]

Best I can come up with is the following. Is this what you meant?
```
[...]
# one instance of parametrized fixture
op = all_arithmetic_operators

def f(x, y):
    if op.startswith('__r'):
        # get op without "r" and invert it
        return getattr(operator, op.replace('__r', '__'))(y, x)
    return getattr(operator, op)(x, y)
[...]
```

Comment (h-vetinari) [reply pandas/tests/frame/test_arithmetic.py]: > this comment applies to the prior test actually. I don't think you need this here at all, as op is just add, mul, sub. did you mean to test the reverse of these as well?

That was indeed not necessary anymore. Removed

Comment (jreback) [reply pandas/tests/frame/test_arithmetic.py]: ok this is fine (your change). note that I suspect we do this in other places, maybe see if you can make a more generalized version of this that we can then include in the pandas.util.testing (but followon-PR)

File: pandas/tests/frame/test_arithmetic.py [hunk 13]
```diff
@@ -127,132 +126,86 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
-            ndim_5 = np.ones(frame.shape + (3, 4, 5))
+    def test_arith_flex_frame(self, all_arithmetic_operators, float_frame,
+                              mixed_float_frame):
+        # one instance of parametrized fixture
+        op = all_arithmetic_operators
+
+        def f(x, y):
+            if op.startswith('__r'):
+                # get op without "r" and invert it
+                return getattr(operator, op.replace('__r', '__'))(y, x)
```
jreback: @jbrockmendel pls have a look here

jreback (approved):

jreback:
ok by me. @jbrockmendel if you'd have a look and merge if ok.

pandas/tests/frame/test_arithmetic.py
jbrockmendel: This is fine.  The alternative would be to do a lookup in core.ops, but I think for a reader who is unfamiliar with core.ops this is clearer.

pandas/tests/frame/test_arithmetic.py
jbrockmendel: Can you separate out these dimension-specific tests?

pandas/tests/frame/test_arithmetic.py
```diff
@@ -127,132 +126,86 @@ def test_df_add_flex_filled_mixed_dtypes(self):
...
-
-        # res_add = frame.add(frame)
-        # res_sub = frame.sub(frame)
-        # res_mul = frame.mul(frame)
-        # res_div = frame.div(2 * frame)
-
-        # tm.assert_frame_equal(res_add, frame + frame)
-        # tm.assert_frame_equal(res_sub, frame - frame)
-        # tm.assert_frame_equal(res_mul, frame * frame)
-        # tm.assert_frame_equal(res_div, frame / (2 * frame))
```
jbrockmendel: Did this commented-out stuff go somewhere else?  Or just determined to be redundant?

pandas/tests/frame/test_arithmetic.py
h-vetinari: Split them out in a separate test

pandas/tests/frame/test_arithmetic.py
h-vetinari: Determined to be redundant with `test_arith_flex_frame`, which checks all arith_ops, and much more thoroughly.

pandas/tests/frame/test_arithmetic.py
h-vetinari: Modified comment slightly

jreback:
thanks @h-vetinari 

so this is a good example of a change. you moved / parameterized things, but its a single change that is quite straightforward for a reviewer to grok / comment on w/o diving into things too much.

PR #22738, opened 2018-09-17
Fixturize tests/frame/test_api and tests/sparse/frame/test_frame

h-vetinari:
- [x] split off from #22730 
- [x] tests added / passed

`tests/frame/test_api` was easy enough to fixturize, but I ran into problems because some tests (notably those in `SharedWithSparse`) are also used in `tests/sparse/frame/test_frame`, and there, the corresponding fixtures were not available. Therefore, this PR fixturizes both modules simultaneously, which also necessitates a new `conftest.py` for the sparse tests.

Some more notes:
* happy to hear feedback on the sparse fixture names. I haven't worked with `SparseDataFrame` so far, and couldn't find much documentation for (e.g. `default_kind`), so the names I chose are best effort
* I would have loved to replace `_check_all` with parametrization, but since fixtures can't be parametrized (see pytest-dev/pytest#349), I had to replace it with direct calls.

### DIFF
--- a/pandas/tests/frame/test_api.py
+++ b/pandas/tests/frame/test_api.py
@@ -24,8 +24,6 @@ from pandas.util.testing import (assert_almost_equal,
 
 import pandas.util.testing as tm
 
-from pandas.tests.frame.common import TestData
-
 
 class SharedWithSparse(object):
     """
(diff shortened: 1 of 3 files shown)


jreback:
lgtm @TomAugspurger is this going to completely mess up SparseEA refactor?

TomAugspurger:
At a glance, it doesn't look like the merge will be *too* bad. I'd recommend merging this first.

jreback:
thanks

PR #22601, opened 2018-09-05
BLD: Fix openpyxl to 2.5.5

gfyoung:
`2.5.5` --> `2.5.6` broke compatibility with pandas `Timestamp` objects.

Closes #22595.

PR #22749, opened 2018-09-18
Corrected to necessary pip requirements format

vadakattu:
- [x] Addresses bug introduced by #22601

Requirements format for building pip env requires '==' instead of '='

### DIFF
--- a/ci/requirements-optional-pip.txt
+++ b/ci/requirements-optional-pip.txt
@@ -14,7 +14,7 @@ lxml
 matplotlib
 nbsphinx
 numexpr
-openpyxl=2.5.5
+openpyxl==2.5.5
 pyarrow
 pymysql
 tables


alimcmaster1:
Looks good to me, thanks
CC @gfyoung

TomAugspurger:
That file is autogenerated. We need to update `scripts/convert_deps.py` to turn `=` into `==` for pip.

TomAugspurger:
And then run `scripts/convert_deps.py` to regenerate the file.

gfyoung:
Second what @TomAugspurger said above.  Good catch, @vadakattu !

vadakattu:
I do. Back from vacation next week, will action.

- Krishna

vadakattu:
Seems to have been resolved in the meantime by #22889

PR #22751, opened 2018-09-18
WIP: Use dispatch_to_series for combine_const

jbrockmendel:
This and #22019 get rid of the only two usages of `BlockManager.eval` and `Block.eval`.  Getting rid of those will be a good day.

This is pretty ugly at the moment.  There are some design issues (namely, what broadcasting rules are intended to be supported) that need to be cleared up before this can be made pretty.


pandas/core/frame.py
```diff
@@ -4867,10 +4867,34 @@ def _combine_const(self, other, func, errors='raise', try_cast=True):
...
-                                   errors=errors,
-                                   try_cast=try_cast)
-        return self._constructor(new_data)
+        elif (np.ndim(other) == 1 and isinstance(other, np.ndarray) and
+              len(other) == len(self.columns)):
+            right = np.broadcast_to(other, self.shape)
+            return ops.dispatch_to_series(self, right, func)
+
+        elif (np.ndim(other) == 1 and
+              isinstance(other, (tuple, np.ndarray)) and
```
jbrockmendel: There is a test for that goes through this path with a tuple, but none with a list.  IIRC, no tests fail if list is included here.

Similarly, in the case above, only np.ndarray cases are tested.  Should tuple and/or list be allowed?

pandas/core/frame.py
```diff
@@ -4867,10 +4867,34 @@ def _combine_const(self, other, func, errors='raise', try_cast=True):
...
+            return ops.dispatch_to_series(self, right, func)
+
+        elif np.ndim(other) == 1:
+            raise ValueError("Shape incompatible")
+
+        elif np.ndim(other) == 2 and other.shape == self.shape:
+            return ops.dispatch_to_series(self, other, func)
+
+        elif (np.ndim(other) == 2 and isinstance(other, np.ndarray) and
+              other.shape[0] == 1 and other.shape[1] == len(self.columns)):
```
jbrockmendel: Should the transposed case be supported?

jbrockmendel:
Likely the easiest way to address the inconsistencies (and ugliness that goes with it) would be to call `_align_method_FRAME` in `ops._comp_method_FRAME`.  It is called in the other two FRAME functions but not that one.

jorisvandenbossche:
I am personally a bit hesitant to do much refactoring on something rather fragile as broadcasting behaviour without some better testing/documenting of the actual/desired broadcasting behaviour (but it might be a good opportunity to actually do that!).

jbrockmendel:
> I am personally a bit hesitant to do much refactoring on something rather fragile as broadcasting 

I definitely agree.  In case it is not clear, this PR is _not_ intended to be merged in its present form.  The  idea behind the ugly, verbose implementation here is to show precisely what behavior is currently tested.  The discussion I'm hoping for is to clarify what the "spec" is supposed to be so that can be documented/tested and then implemented cleanly.

jorisvandenbossche:
OK, good to hear. 

> The idea behind the ugly, verbose implementation here is to show precisely what behavior is currently tested. The discussion I'm hoping for is to clarify what the "spec" is supposed to be so that can be documented/tested and then implemented cleanly.

I think it would be a good idea to try to translate the 'ugly code' you wrote now to a written explanation of the broadcasting rules.

PR #22759, opened 2018-09-19
ENH: Making header_style a property of ExcelFormatter #22758

### LINKED ISSUE #22758: Make header_style a Property of ExcelFormatter
Right now the header style of excel exports is defined as global to the `pandas/io/formats/excel.py` module:

https://github.com/pandas-dev/pandas/blob/1a12c41d201f56439510e683fadfed1218ea9067/pandas/io/formats/excel.py#L37

Because this exists outside of the `ExcelFormatter` class where it is used it is way more difficult to override the default than it needs to be. If this was moved from global scope to instance scope it would allow users to theoretically subclass the `ExcelFormatter` and easily define their own style for the header

This could also eliminate the need for any work on #19386

dannyhyunkim:
- [y] closes #22758
- [y] tests added / passed
- [y] passes `git diff upstream/master -u -- "*.py" | flake8 --diff`


WillAyd:
Nice job! Can you add a whatsnew note for 0.25?

WillAyd:
This is totally my mistake but I meant v0.24 (a whatsnew should already exist). 

You did a great job in adding one but unfortunately that’s not what we want here. Sorry!

dannyhyunkim:
No worries at all. I should have clarified earlier.

WillAyd (commented):
I think this looks good but @gfyoung care to take a look?

doc/source/whatsnew/v0.24.0.txt
```diff
@@ -185,6 +185,7 @@ Other Enhancements
 - :class:`Resampler` now is iterable like :class:`GroupBy` (:issue:`15314`).
 - :meth:`Series.resample` and :meth:`DataFrame.resample` have gained the :meth:`Resampler.quantile` (:issue:`15023`).
 - :meth:`Index.to_frame` now supports overriding column name(s) (:issue:`22580`).
+- :class:`ExcelFormatter` now has a `header_style` property. Was previously in global scope (:issue:`22758`).
```
WillAyd: Might not actually need this since the ExcelFormatter isn't part of the exposed API

TomAugspurger: Should it be? If we want to support use-cases like https://github.com/pandas-dev/pandas/issues/22758 then it should be. OTOH, I don't think we consider things like HTMLFormatter to be public.

TomAugspurger: That sounds right to me. (and there's nothing stopping people from using it anyway).

gfyoung:
@WillAyd : Thanks for the ping.  Will take a look later today.

doc/source/whatsnew/v0.24.0.txt
WillAyd: Opened #22773 to further that discussion; @dannyhyunkim if you can revert the change to whatsnew would be great as this in its current state is more of an internal refactor. Sorry for the back and forth from me on that!

gfyoung (approved):
LGTM!

cc @jreback @TomAugspurger

WillAyd (approved):
lgtm as well

